PR #532, opened 2024-05-06
Explosion testing

Tiduzz:
Fixed explosion moves not killing user when missing, neither displaying the animation.

Moves affected: Explosion, Self-Destruct and Misty-Explosion

This is my first pull request (in like, ever?) to the repo, please advise if anything has been done incorrectly.


bennybroseph:
Was this tested against Damp?

Tiduzz:
> Was this tested against Damp?

It was not, i'll test this as soon as i'm able to. Good call

src/phases.ts
```diff
@@ -2254,7 +2254,10 @@ export class MovePhase extends BattlePhase {
       
       if ((moveQueue.length && moveQueue[0].move === Moves.NONE) || !targets.length) {
         moveQueue.shift();
-        this.cancel();
+        if (this.move.moveId === 153 || this.move.moveId === 120 || this.move.moveId === 802){ //Currently all explosions, AKA, kamikaze-attacks. There might be a better way to do it except to hardcode it, but i haven't found how yet.
```
benjam1337, line 2257:
```suggestion
        //Currently all explosions, AKA, kamikaze-attacks. There might be a better way to do it except to hardcode it, but i haven't found how yet.
        if (!(this.move.moveId in [120,153,802])){
            this.cancel();
        }
```

Tiduzz: That's nice. I added it to the code, but instead of the ID i used Moves.MOVE, it's easier to read.

PR #542, opened 2024-05-06
arena trap and other similar abilities 

Reldnahc:
have applyCheckTrappedAbAttrs() send the pokemon that is being prevented from switching also.

implement conditional arena traps.

implement magnet pull.

correct arena trap to only effect grounded pokemon.

correct shadow tag to only effect non-ghost types.

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2215,7 +2215,7 @@ export class ConditionalArenaTrapAbAttr extends ArenaTrapAbAttr {
   }
 
   applyCheckTrapped(pokemon: Pokemon, opponent: Pokemon, passive: boolean, trapped: Utils.BooleanHolder, args: any[]): boolean {
-    if (this.condition(opponent)) {
+    if (this.condition(opponent) && !opponent.isOfType(Type.GHOST)) {
       trapped.value = true;
       return true;
     }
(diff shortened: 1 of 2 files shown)


src/data/ability.ts
```diff
@@ -2733,7 +2750,7 @@ export function initAbilities() {
     new Ability(Abilities.INTIMIDATE, 3)
       .attr(PostSummonStatChangeAbAttr, BattleStat.ATK, -1, false, true),
     new Ability(Abilities.SHADOW_TAG, 3)
-      .attr(ArenaTrapAbAttr),
+      .attr(ConditionalArenaTrapAbAttr, (opponent: Pokemon) => !opponent.isOfType(Type.GHOST)),
```
TempsRay, line 2753: The check for ghost type can be applied in ConditionalArenaTrapAbAttr since all ghosts are immune to being trapped

src/data/ability.ts
```diff
@@ -2866,7 +2881,7 @@ export function initAbilities() {
       .attr(PostSummonWeatherChangeAbAttr, WeatherType.SUNNY)
       .attr(PostBiomeChangeWeatherChangeAbAttr, WeatherType.SUNNY),
     new Ability(Abilities.ARENA_TRAP, 3)
-      .attr(ArenaTrapAbAttr)
+      .attr(ConditionalArenaTrapAbAttr, (opponent: Pokemon) => opponent.isGrounded())
```
TempsRay, line 2884: I believe the `isGrounded` might be a little bit more complicated than just checking a boolean since it needs to account for a lot of factors, but that might be more of a different PR to account for all the intricacies of being grounded

src/data/ability.ts
Reldnahc: I didn't know. I will fix this!

TempsRay:
Oh yeah, apparently the ability run away lets you run away (but not switch out) even if arena trap'd or shadow tag'd

bennybroseph:

PR #544, opened 2024-05-06
Repels Implementation

ojukeihpos:
Implements all types of Repels (Regular, Super, and Max) with functionality as described on the [Trello Board](https://trello.com/c/1TSW0HSz/29-add-repels). Repels last for (5/10/25) waves and prevent double battles from occurring (including trainer doubles) by checking if the player has the modifier active during `newBattle()`, and overriding the double parameter to be false.

Also modifies the weightings of both repels and lures so that they do not show up in the shop if the player already has one of either.

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -4,7 +4,7 @@ import { NextEncounterPhase, NewBiomeEncounterPhase, SelectBiomePhase, MessagePh
 import Pokemon, { PlayerPokemon, EnemyPokemon } from './field/pokemon';
 import PokemonSpecies, { PokemonSpeciesFilter, allSpecies, getPokemonSpecies, initSpecies, speciesStarters } from './data/pokemon-species';
 import * as Utils from './utils';
-import { Modifier, ModifierBar, ConsumablePokemonModifier, ConsumableModifier, PokemonHpRestoreModifier, HealingBoosterModifier, PersistentModifier, PokemonHeldItemModifier, ModifierPredicate, DoubleBattleChanceBoosterModifier, FusePokemonModifier, PokemonFormChangeItemModifier, TerastallizeModifier, overrideModifiers, overrideHeldItems } from './modifier/modifier';
+import { Modifier, ModifierBar, ConsumablePokemonModifier, ConsumableModifier, PokemonHpRestoreModifier, HealingBoosterModifier, PersistentModifier, PokemonHeldItemModifier, ModifierPredicate, DoubleBattleChanceBoosterModifier, DoubleBattleChancePreventerModifier, FusePokemonModifier, PokemonFormChangeItemModifier, TerastallizeModifier, overrideModifiers, overrideHeldItems } from './modifier/modifier';
 import { PokeballType } from './data/pokeball';
 import { initCommonAnims, initMoveAnim, loadCommonAnimAssets, loadMoveAnimAssets, populateAnims } from './data/battle-anims';
 import { Phase } from './phase';
@@ -861,6 +861,8 @@ export default class BattleScene extends SceneBase {
 					this.applyModifiers(DoubleBattleChanceBoosterModifier, true, doubleChance);
 					playerField.forEach(p => applyAbAttrs(DoubleBattleChanceAbAttr, p, null, doubleChance));
 					doubleTrainer = !Utils.randSeedInt(doubleChance.value);
+					if (this.getModifiers(DoubleBattleChancePreventerModifier).length != 0)
+						doubleTrainer = false;
 				}
 				newTrainer = trainerData !== undefined ? trainerData.toTrainer(this) : new Trainer(this, trainerType, doubleTrainer ? TrainerVariant.DOUBLE : Utils.randSeedInt(2) ? TrainerVariant.FEMALE : TrainerVariant.DEFAULT);
 				this.field.add(newTrainer);
@@ -880,6 +882,9 @@ export default class BattleScene extends SceneBase {
 
 		if (Overrides.DOUBLE_BATTLE_OVERRIDE)
 			newDouble = true;
+
+		if (this.getModifiers(DoubleBattleChancePreventerModifier).length != 0)
+			newDouble = false;
 
 		const lastBattle = this.currentBattle;
 
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -407,11 +407,21 @@ export class DoubleBattleChanceBoosterModifierType extends ModifierType {
   }
 }
 
+/**
+ * Class representing a ModifierType for a Modifier that prevents Double Battles from occurring.
+ * @extends ModifierType
+ */
 export class DoubleBattleChancePreventerModifierType extends ModifierType {
   public battleCount: integer;
 
+  /**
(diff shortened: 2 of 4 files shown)


src/battle-scene.ts
```diff
@@ -870,6 +870,9 @@ export default class BattleScene extends SceneBase {
 	}
 
 	newBattle(waveIndex?: integer, battleType?: BattleType, trainerData?: TrainerData, double?: boolean): Battle {
+		if (this.getModifiers(DoubleBattleChancePreventerModifier).length != 0) {
+			double = false;
```
Flashfyre, line 874: I don't think this is the right place to put this. This is for initialization and whether it's set or not depends on whether data is being loaded.

Flashfyre (changes requested):
Avoiding running a formatter on the code next time, it makes changes hard to read.

src/battle-scene.ts
```diff
@@ -821,10 +821,11 @@ export default class BattleScene extends SceneBase {
 				if (trainerConfigs[trainerType].doubleOnly)
 					doubleTrainer = true;
 				else if (trainerConfigs[trainerType].hasDouble) {
-					const doubleChance = new Utils.IntegerHolder(newWaveIndex % 10 === 0 ? 32 : 8);
+					let doubleChance = new Utils.IntegerHolder(newWaveIndex % 10 === 0 ? 32 : 8);
```
Flashfyre, line 824: This was fine as const

src/battle-scene.ts
```diff
@@ -833,10 +834,11 @@ export default class BattleScene extends SceneBase {
 
 		if (double === undefined && newWaveIndex > 1) {
 			if (newBattleType === BattleType.WILD && !this.gameMode.isWaveFinal(newWaveIndex)) {
-				const doubleChance = new Utils.IntegerHolder(newWaveIndex % 10 === 0 ? 32 : 8);
+				let doubleChance = new Utils.IntegerHolder(newWaveIndex % 10 === 0 ? 32 : 8);
```
Flashfyre, line 837: Same as above.

src/battle-scene.ts
```diff
@@ -833,10 +834,11 @@ export default class BattleScene extends SceneBase {
...
 				this.applyModifiers(DoubleBattleChanceBoosterModifier, true, doubleChance);
 				playerField.forEach(p => applyAbAttrs(DoubleBattleChanceAbAttr, p, null, doubleChance));
 				newDouble = !Utils.randSeedInt(doubleChance.value);
+				if (this.getModifiers(DoubleBattleChancePreventerModifier).length != 0) newDouble = false;
```
Flashfyre, line 841:
```suggestion
				if (this.getModifiers(DoubleBattleChancePreventerModifier).length != 0)
					newDouble = false;
```

src/modifier/modifier-type.ts
```diff
@@ -332,6 +332,17 @@ export class DoubleBattleChanceBoosterModifierType extends ModifierType {
   }
 }
 
+export class DoubleBattleChancePreventerModifierType extends ModifierType {
+  public battleCount: integer;
+
+  constructor(name: string, battleCount: integer) {
+    super(name, `Prevents the chance of an encounter being a double battle for ${battleCount} battles`, (_type, _args) => new Modifiers.DoubleBattleChancePreventerModifier(this, this.battleCount),
```
Flashfyre, line 339:
```suggestion
    super(name, `Prevents double battles for ${battleCount} battles`, (_type, _args) => new Modifiers.DoubleBattleChancePreventerModifier(this, this.battleCount),
```

src/modifier/modifier-type.ts
```diff
@@ -983,7 +994,15 @@ const modifierPool: ModifierPool = {
       const thresholdPartyMemberCount = Math.min(party.filter(p => p.hp && p.getMoveset().filter(m => (m.getMove().pp - m.ppUsed) <= 5).length).length, 3);
       return thresholdPartyMemberCount;
     }, 3),
-    new WeightedModifierType(modifierTypes.LURE, 2),
+    new WeightedModifierType(modifierTypes.REPEL, (party: Pokemon[]) => {
+      const existingItem = party[0].scene.findModifier(m => m.type instanceof DoubleBattleChancePreventerModifierType || m.type instanceof DoubleBattleChanceBoosterModifierType);
+      //return existingItem ? 0 : 2;
```
Flashfyre, line 999: Debug code left

src/overrides.ts
```diff
@@ -4,7 +4,7 @@ import { Biome } from "./data/enums/biome";
 import { Moves } from "./data/enums/moves";
 import { WeatherType } from "./data/weather";
 
-export const SEED_OVERRIDE = '';
+export const SEED_OVERRIDE = 'b3DoubDG5i4izShAO06Z2V2y';
```
Flashfyre, line 7: Debug code left

Flashfyre (changes requested):

ojukeihpos:
For whatever reason trying to grab the latest commits from the main repo completely broke whatever I had, I'll fix it and reopen

package-lock.json
bennybroseph, line 1: These changes need to be reverted

bennybroseph (changes requested):
Could we get a video of this move working in game?

package-lock.json
ojukeihpos: hi Benny, I'm not sure what changes you mean. I didn't make any changes to `package-lock.json`, do you mean that I should pull the version from the main repo? I also don't know how I would make a video of the item working in game as the main function of the item is to prevent the occurrence of something happening.

I will add comments as per the documentation.

bennybroseph: I'm unsure, but changes to package-lock.json are committed here and need to be removed.

As for the video, you are still able to show the item itself in the shop, the description, etc.

src/overrides.ts
bennybroseph, line 1: We don't commit changes to the override

bennybroseph (changes requested):

src/overrides.ts
ojukeihpos: Yeah, I'm trying to clean up the commits. I'll make a comment when I have things cleaned up, but to be quite honest, things are kind of a mess here and the dev-corner channel is closed

ojukeihpos:
https://github.com/pagefaultgames/pokerogue/assets/10471964/e699a9e9-ed9f-4ff5-a7e8-edc103394804

Video is subtitled, explanation of implementation is written here for easy access:

The implementation was tested by setting the double battle override to true, forcing double battles to always occur.
Normal Repel weighting was set to 10000 to ensure it appears in the store for demonstration purposes.

When `newBattle()` is called, there is a check for if a modifier of `DoubleBattleChanceModifierType` exists in the player's modifiers list. If it does, then one of the following occurs:

If the `newBattleType` is a Trainer, the `doubleTrainer` value is overwritten to always be `false`. This does not apply if the Trainer can ONLY be a double battle.
Otherwise, `newDouble`'s value is overwritten to always be false as long as the player has a Repel.

This also overrides the `DOUBLE_BATTLE_OVERRIDE` value. Let me know if that shouldn't be the case.

Tempo-anon:
Closing due to inactivity and merge conflicts. We're also not sure if we want repels right now.

PR #547, opened 2024-05-06
Fixed Grafaiai animation

gericocross:
Now the last frame of Grafaiai animation is no longer an invisible frame.


Flashfyre:
I appreciate the effort but the floating sprite is part of the issue so this will just need to be reprocessed anyway. I'm not sure if it's worth a partial fix.

Flashfyre:
On the other hand, it costs me 10 seconds to merge the work you already did so I guess I'll just put it in.

Flashfyre (approved):

PR #485, opened 2024-05-05
add simplifed  chinese locale, translation is complete now.

Jane620:
only for the code part and a small portion of text testing.

PR #549, opened 2024-05-06
add Simplified Chinese Translation.

ZXZhuang:
translate some file to Simplified Chinese :
ability.ts
move.ts
pokemon.ts


TempsRay:
Is this still a WIP? Also I believe there's another open PR #485 for Chinese localization

PR #550, opened 2024-05-06
Redo moveset generation

Xavion3:
Starting this partially to get feedback too and make it more obvious it can be tested by people. This redoes how random movesets are generated.

This generally tweaks the weighting of moves, replacing the old easing function with a weighting that's biased based on level, avoiding the old issue where you'd get things like multiple level 1 moves being different weights based on the order they were in in the level up list. This also makes evo moves not literally the least likely moves to be picked but instead weights them fairly highly, and also uses a basic check for "move reminder" moves and gives them a weight bump too assuming they'll be unusually important parts of a pokemons moveset.

For bosses, this additionally tries to force them get a stab move which is merely hoped for by the current approach instead of forced.

For trainers, this does the above and also adds some basic prioritisation to move selection. Trainers will normally have at least one status move, and prefer to diversify their damage types within a moveset and have STAB moves. They also gain a small chance to start having TM moves and egg moves at higher levels.

All the weighting is tweakable, I've just picked some values which seem to produce reasonable results to me.


src/field/pokemon.ts
```diff
@@ -1121,38 +1123,114 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+          if (tmPoolTiers[moveId] === ModifierTier.COMMON && this.level >= 15)
+            movePool.push([moveId, 4]);
+          else if (tmPoolTiers[moveId] === ModifierTier.GREAT && this.level >= 30)
+            movePool.push([moveId, 8]);
+          else if (tmPoolTiers[moveId] === ModifierTier.ULTRA && this.level >= 50)
+            movePool.push([moveId, 14]);
+        }
+      }
+
+      if (this.level >= 25) { // No egg moves below level 25
```
TempsRay, line 1166: I'm still iffy about including egg moves but that's probably a different conversation for a different time/place. There is a funny bit of gameplay/mechanic segregation here though as the rival's Starter will suddenly be able to have egg moves in their movepool later on while players can't memory mushroom back egg moves. Then again, Ivy is the type of person to suddenly give her Pokemon a passive halfway through a run

src/field/pokemon.ts
```diff
@@ -1121,38 +1123,114 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+            break;
+          }
+        }
+        if (compatible && !movePool.some(m => m[0] === moveId) && !allMoves[moveId].name.endsWith(' (N)')) {
+          if (tmPoolTiers[moveId] === ModifierTier.COMMON && this.level >= 15)
+            movePool.push([moveId, 4]);
+          else if (tmPoolTiers[moveId] === ModifierTier.GREAT && this.level >= 30)
+            movePool.push([moveId, 8]);
+          else if (tmPoolTiers[moveId] === ModifierTier.ULTRA && this.level >= 50)
+            movePool.push([moveId, 14]);
```
TempsRay, line 1162: Are the 4/8/14 arbitrarily chosen weights?

src/field/pokemon.ts
```diff
@@ -1121,38 +1123,114 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
         break;
-      if (movePool.indexOf(levelMove[1]) === -1) {
-        if (!allMoves[levelMove[1]].name.endsWith(' (N)'))
-          movePool.push(levelMove[1]);
-        else
-          movePool.unshift(levelMove[1]);
+      let weight = levelMove[0];
+      if (weight === 0) // Evo Moves
+        weight = 50;
+      if (weight === 1 && allMoves[levelMove[1]].power >= 80) // Assume level 1 moves with 80+ BP are "move reminder" moves and bump their weight
```
TempsRay, line 1129: Is this the only instance where base power is considered for weighting? I feel like base power should also be a factor for weighting especially in the later game. Even if learned at the same level lava plume should have twice the weight of ember for example

src/field/pokemon.ts
Flashfyre: I think egg moves should probably only generate in special cases.

src/field/pokemon.ts
```diff
@@ -1121,38 +1123,114 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+        weight = 40;
+      if (allMoves[levelMove[1]].name.endsWith(' (N)'))
+        weight /= 100; // Unimplemented level up moves are possible to generate, but 1% of their normal chance.
```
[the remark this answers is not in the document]
Xavion3: They were allowed in the old code so figured might be a reason to not ignore them.

src/field/pokemon.ts
Xavion3: All the weights are kinda arbitrarily chosen but yeah.

src/field/pokemon.ts
Xavion3: Effectively yes, only here. Adding an additional check based on power means we have to decide how much power matters, which means we also have to decide how much power matters in comparison to being a status move.

src/field/pokemon.ts
Xavion3: I like the idea of egg moves myself, as rare things which can pop up occasionally. They can help tie back into that mechanic potentially too, a taste of what you can unlock.

Xavion3: I've decreased the weight of egg moves a bit, they did seem a bit too common in my tests.

Flashfyre (approved):

PR #552, opened 2024-05-06
New spanish translations and literals for all langs

ArceUseless:
Spanish translations and literals for all langs, they affect starter-select-ui-handler


Flashfyre (changes requested):
I think we need to separate starter-select-ui-handler into its own localization file.

ArceUseless:
I kind of messed up the pull request but I think everything should be in order now

Flashfyre (approved):

PR #557, opened 2024-05-06
Implements Supersweet Syrup

lucfd:


Madmadness65 (changes requested):
This ability is somewhat different from other stat changes on summon, in that it activates only once per battle, unlike something like Intimidate, which activates every time the ability holder is sent out.

PR #558, opened 2024-05-06
Added Pokemon renaming hook to the Summary

roberika:
The hook is done. On Submit it should be able to set the Pokemon's name to its new name saved in variable newName, and on Cancel is sets newName back to the Pokemon's current name. It also saves capitalization state in variable isCapitalized, and will automatically capitalizes the first letter.

But, the text input UI itself is still mising and there's still no actual way to input the text. Ideally it'd be a cursor input system like traditional handheld Pokemon games so controller player's could also rename their Pokemon, but to make it there needs to be a UI first so moving the cursor variable is consistent to moving the cursor UI.

If anyone wants to, feel free to take over adding the feature since making good UI is not my cup of tea.

### DIFF
--- a/src/ui/summary-ui-handler.ts
+++ b/src/ui/summary-ui-handler.ts
@@ -29,7 +29,8 @@ enum Page {
 
 export enum SummaryUiMode {
   DEFAULT,
-  LEARN_MOVE
+  LEARN_MOVE,
+  RENAME,
 }
 
 export default class SummaryUiHandler extends UiHandler {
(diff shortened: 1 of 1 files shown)


Flashfyre (commented):
I hadn't considered adding a rename mode. What would this look like? Should we actually need to go to the summary at all for this?

roberika:
I kinda just defaulted to placing it on the summary because of Unbound. I think this is the best place to put it though since it'll be hard to fit it anywhere else. The only other place that might fit is adding it as an option to the Party UI, but that might bloat the menu. Placing it in Summary UI is nice and tucked away so some players that doesn't use them often don't need to scroll an additional menu, but is accessible to the players that do want to use it.

bennybroseph:

PR #559, opened 2024-05-06
Update French starter-select-ui-handler.ts

Adri1:

PR #564, opened 2024-05-06
Fix overflowing fr locale

Dakurei:
## Changes
- For lack of space, a literal translation can't be integrated, so replace "cycle" verbs with an "arrow".
  A change to a verb translation may be reconsidered if more space becomes available in the future. (Fix #559)

Before:
image

After:
image

- Plus minor corrections


Flashfyre (approved):

Adri1:
Thanks for ce corrections 👍

PR #565, opened 2024-05-06
double battle and passive overrides

Reldnahc:
add override to force double battles.
add override for passive abilities.


Greenlamp2 (commented):
I've tested it, everything seems working as intended

Flashfyre (approved):

PR #567, opened 2024-05-06
fix scrappy and mindseye working as passive ability

contra1337:
- change the attribute apply to more properly use the cancelled param
- apply the attribute for active and passive abilities
- mark minds eye as partially implemented

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2483,11 +2483,11 @@ export class IgnoreTypeImmunityAbAttr extends AbAttr {
   }
 
   apply(pokemon: Pokemon, passive: boolean, cancelled: Utils.BooleanHolder, args: any[]): boolean {
-    if (this.defenderType === (args[1] as Type)) {
-      cancelled.value = this.allowedMoveTypes.some(type => type === (args[0] as Type));
+    if (this.defenderType === (args[1] as Type) && this.allowedMoveTypes.some(type => type === (args[0] as Type))) {
+      cancelled.value = true;
+      return true;
     }
-
-    return cancelled.value;
+    return false;
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/field/pokemon.ts
```diff
@@ -901,14 +901,16 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       return this.isTerastallized() ? 2 : 1;
     const types = this.getTypes(true, true);
 
-    const ignorableImmunities = source?.getAbility()?.getAttrs(IgnoreTypeImmunityAbAttr) || [];
-    const cancelled = new Utils.BooleanHolder(false);
+    let multiplier = types.map(defType => {
+      if (source && source.hasAbilityWithAttr(IgnoreTypeImmunityAbAttr)) {
```
contra1337, line 905: i think hasAbilityWithAttr might be redundant with how applyAbAttrs works, can remove if desired but was suggested to use it

src/data/ability.ts
```diff
@@ -2483,11 +2483,11 @@ export class IgnoreTypeImmunityAbAttr extends AbAttr {
   }
 
   apply(pokemon: Pokemon, passive: boolean, cancelled: Utils.BooleanHolder, args: any[]): boolean {
-    if (this.defenderType !== (args[1] as Type)) {
-      return false;
+    if (this.defenderType === (args[1] as Type) && this.allowedMoveTypes.some(type => type === (args[0] as Type))) {
```
Xavion3, line 2486: This could just be an includes right, doesn't have to be a some?

src/field/pokemon.ts
Xavion3: Yeah, the apply handles this so this is redundant.

Xavion3 (changes requested):

src/data/ability.ts
contra1337: yep good point thanks

Xavion3 (approved):

Madmadness65 (approved):

PR #568, opened 2024-05-06
Future Sight Distiction

bennybroseph:
Marked Future Sight as Partial (P). I think we all know this is the case, I'm just marking it as such explicitly.


Reldnahc:
Should probably mark Doom Desire as (p) as well since they use the same code. I can also take a look at fixing these. I could implement them like how I did wish and that should fix the problems I think.

PR #569, opened 2024-05-06
Implement innards out

TempsRay:
Reusing a lot of code from Aftermath and `PostFaintContactDamageAbAttr`. Passing in lastAttack.damage into the unused `args[]` of `applyPostFaintAbAttrs` since that value is capped by the HP of the fainted Pokemon at the time.

Not sure about the message, but the for aftermath is pretty descriptive so I also reused that.

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2281,7 +2281,6 @@ export class PostFaintHPDamageAbAttr extends PostFaintAbAttr {
   }
 
   applyPostFaint(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: PokemonMove, hitResult: HitResult, args: any[]): boolean {
-    console.log(args);
     let damage = args[0];
     attacker.damageAndUpdate((damage), HitResult.OTHER);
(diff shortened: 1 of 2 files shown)


destinytrigger:
Does lastAttack.damage need to be passed in args? Should be able to get it from pokemon.turnData.attacksReceived[0].damage in applyPostFaint, yeah?

TempsRay:
Oh that's neat, didn't know that was a thing. Will index 0 be the first or last actual hit?

destinytrigger:
Should be the last. Above your change to phases.ts to pass lastAttack.damage is:

      const lastAttack = pokemon.turnData.attacksReceived[0];

So you'll be retrieving the same data at least. I just thought this was cleaner. This code base is a heck of a thing.

I haven't gotten this set up locally to test yet, I just started looking at the code base and heard someone mention Innards Out, so I took a whack at it. But it looks like you're already there!

bennybroseph (dismissed):
Looking for TSDoc documentation for this commit

Xavion3 (approved):

PR #572, opened 2024-05-06
Starter filter tab

jmlevesque333:
Added a filtering tab when starter selection. This tab can be used to add more filtering options in the future, like IV thresholds and whatnot.

This is my first experience with phaser so do not hesitate to give me feedback. The ui for the tabs itself is pretty rough, did not know how to achieve what I was looking for.

Here are some screenshots:
image
image
image
image

### DIFF
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -12,7 +12,7 @@ import { AbilityAttr, DexAttr, DexAttrProps, DexEntry, Passive as PassiveAttr, S
 import * as Utils from "../utils";
 import PokemonIconAnimHandler, { PokemonIconAnimMode } from "./pokemon-icon-anim-handler";
 import { StatsContainer } from "./stats-container";
-import { addWindow } from "./ui-theme";
+import { WindowVariant, addWindow } from "./ui-theme";
 import { Nature, getNatureName } from "../data/nature";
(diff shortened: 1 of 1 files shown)


victorlapa:
personally i liked the changes, i think it also adds room for further improvements like filtering by name, ivs, moves, etc.

bennybroseph:
I feel like this would need a lot more UI to stay consistent. I feel this should show all pokemon of all gens and filter to only those that match your filtered requirements. Currently it looks like it just fades them out which looks too similar to how it is when you are out of points.

jmlevesque333:
> I feel like this would need a lot more UI to stay consistent. I feel this should show all pokemon of all gens in a single window and filter to only those that match your filtered requirements. Currently it looks like it just fades them out which looks too similar to how it is when you are out of points.

I have designed it to not hide pokemons hat dont fit a filter because in my opinion, being able to see pokemons that dont fit a filter and still select them can be as useful as seeing those that do fit the filter. I cam be wromg on this though obviously.

bennybroseph:
Yeah, I think it just adds clutter. I'd like to see it only display pokemon that fit the filter since there are so many pages to get through and the whole point of the filter would be to get to things quickly.

flx-sta:
@jmlevesque333 are you still actively working on this?

bennybroseph:
Closing due to inactivity

PR #573, opened 2024-05-06
Show number of Pokeballs in inventory when selecting a pokeball in the reward phase

happinyz:
QOL enhancement to show the user how many of the selected Pokeball they have while in the rewards phase.

image

### DIFF
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -144,13 +144,14 @@ class AddPokeballModifierType extends ModifierType implements Localizable {
   }
 
   localize(): void {
+    // TODO: Actually use i18n to localize this description.
     this.name = `${this.count}x ${getPokeballName(this.pokeballType)}`;
     this.description = `Receive ${getPokeballName(this.pokeballType)} x${this.count} (Inventory: {AMOUNT}) \nCatch Rate: ${getPokeballCatchMultiplier(this.pokeballType) > -1 ? `${getPokeballCatchMultiplier(this.pokeballType)}x` : 'Certain'}`;
   }
   
   getDescription(scene: BattleScene): string {
     this.localize();
-    return this.description.replace('{AMOUNT}', scene.pokeballCounts[this.pokeballType].toLocaleString('en-US'));
+    return this.description.replace('{AMOUNT}', scene.pokeballCounts[this.pokeballType].toString());
   }
 
 }
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -145,8 +145,14 @@ class AddPokeballModifierType extends ModifierType implements Localizable {
 
   localize(): void {
     this.name = `${this.count}x ${getPokeballName(this.pokeballType)}`;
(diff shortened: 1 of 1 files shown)


src/modifier/modifier-type.ts
```diff
@@ -145,8 +145,14 @@ class AddPokeballModifierType extends ModifierType implements Localizable {
 
   localize(): void {
     this.name = `${this.count}x ${getPokeballName(this.pokeballType)}`;
-    this.description = `Receive ${getPokeballName(this.pokeballType)} x${this.count}\nCatch Rate: ${getPokeballCatchMultiplier(this.pokeballType) > -1 ? `${getPokeballCatchMultiplier(this.pokeballType)}x` : 'Certain'}`;
+    this.description = `Receive ${getPokeballName(this.pokeballType)} x${this.count} (Inventory: {AMOUNT}) \nCatch Rate: ${getPokeballCatchMultiplier(this.pokeballType) > -1 ? `${getPokeballCatchMultiplier(this.pokeballType)}x` : 'Certain'}`;
   }
+  
+  getDescription(scene: BattleScene): string {
+    this.localize();
+    return this.description.replace('{AMOUNT}', scene.pokeballCounts[this.pokeballType].toLocaleString('en-US'));
```
TempsRay, line 153: Does this need to be en-US? Isn't it just a number?

src/modifier/modifier-type.ts
```diff
@@ -145,8 +145,14 @@ class AddPokeballModifierType extends ModifierType implements Localizable {
 
   localize(): void {
     this.name = `${this.count}x ${getPokeballName(this.pokeballType)}`;
-    this.description = `Receive ${getPokeballName(this.pokeballType)} x${this.count}\nCatch Rate: ${getPokeballCatchMultiplier(this.pokeballType) > -1 ? `${getPokeballCatchMultiplier(this.pokeballType)}x` : 'Certain'}`;
+    this.description = `Receive ${getPokeballName(this.pokeballType)} x${this.count} (Inventory: {AMOUNT}) \nCatch Rate: ${getPokeballCatchMultiplier(this.pokeballType) > -1 ? `${getPokeballCatchMultiplier(this.pokeballType)}x` : 'Certain'}`;
```
TempsRay, line 148: Does Inventory need to be localized?

TempsRay (commented):
LGTM besides some questions I had on localization. Good QOL update

src/modifier/modifier-type.ts
happinyz: Good point, changed to use `toString`

src/modifier/modifier-type.ts
happinyz: I'm not super familiar with i18n, but it seems like here we weren't doing any localization in the first place, despite the `localize` function existing. I agree we should be localizing this, but it's out of the scope of this PR. I added a comment to remind us to actually use i18n in the future.

TempsRay (approved):

Madmadness65 (approved):

PR #238, opened 2024-04-22
Implement Heal Bell + Aromatherapy

shayebeadling:
1. Adds a PartyStatusCureAttr to handle Heal Bell + Aromatherapy;
Originally I created a more general use StatusCureAttr that hooked into our targeting properly, however the current targeting system can not handle pokemon not currently on the field. So without refactoring it, these abilities needed their own way of handling the party targeting.

2. Adds a PartyStatusCurePhase to handle the unique case of modifying pokemon without a BattlerIndex;
Takes a condition that specifies what would disqualify the pokemon from being cured. Heal Bell = Soundproof, Aromatherapy = Sap Sipper. These need to be provided since we're circumventing the targeting system. 

3. Modifies pokemon.resetStatus() to have an optional flag to not revive. Simple override for this case.

NOTE: This implementation is hacky because of the current targeting system. If anyone has any suggestions for improvements let me know, or maybe this warrants a refactor of the targeting system for these special cases.

PR #575, opened 2024-05-07
Implement Heal Bell, Aromatherapy, Sparkly Swirl and added status effect options to overrides

innerthunder:
This is a cleaner version of my last pull request...

A new attribute `HealPartyStatusAttr` implements the status healing effects of these moves:
- Aromatherapy
- Heal Bell
- Sparkly Swirl

These moves are also implemented/updated with the existing `HealStatusEffectAttr` attribute:
- Purify
- Refresh (should now wake up the user if used via Sleep Talk)

New overrides are also added to `overrides.ts`. `STATUS_OVERRIDE` initalizes player Pokemon with the specified status effect, and `OPP_STATUS_OVERRIDE` does the same for enemy Pokemon.


.gitignore
```diff
@@ -35,4 +35,6 @@ src/data/battle-anim-raw-data*.ts
 src/data/battle-anim-data.ts
 src/overrides.ts
 
-coverage
\ No newline at end of file
+coverage
+package-lock.json
+package.json
```
contra1337, line 40: these files are necessary for the project

package.json
```diff
@@ -22,12 +24,13 @@
...
 		"pokenode-ts": "^1.20.0",
 		"typescript": "^5.0.3",
-		"vite": "^4.5.0",
+		"vite": "^4.5.3",
 		"vite-plugin-fs": "^0.4.4",
 		"vitest": "^1.4.0",
 		"vitest-canvas-mock": "^0.3.3"
 	},
 	"dependencies": {
+		"@koa/router": "^12.0.1",
```
contra1337, line 33: why is this being added? doesn't seem relevant or imported anywhere either

innerthunder: Sorry for the confusion, I had an error when I first ran the dev build related to koa/router, and I had to manually install a later version of npm to work around it. It modified the json files and I figured those changes shouldn't be included

Xavion3 (changes requested):
This should be two seperate PRs. Don't include both the engine changes and package changes in the same PR. Also don't include override values which actually affect things in a PR.

bennybroseph:
Already implemented here https://github.com/pagefaultgames/pokerogue/pull/238

PR #578, opened 2024-05-07
Fix attacks not working against semi-invulnerable enemies

Tiduzz:
Added an exception in phases.ts to bypass premature cancellation of attacks against semiInvulnerable enemies (fly, bounce, dive, dig, etc)

Worked properly with all moves i tried, but i encourage whoever is interested to test with other moves or point out moves which might not work with this change.

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -3206,8 +3206,8 @@ export function initAbilities() {
     new Ability(Abilities.HARVEST, 5)
       .unimplemented(),
     new Ability(Abilities.TELEPATHY, 5)
-      .ignorable()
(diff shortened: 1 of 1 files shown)


Tiduzz:
This is currently, how attacks interact with semi-invulnerable targets
https://github.com/pagefaultgames/pokerogue/assets/42784723/4763d3ff-0229-4521-b74a-b84ede7446d7
---------
This is how the attacks interact with semi-invulnerable targets AFTER the fix being applied
https://github.com/pagefaultgames/pokerogue/assets/42784723/b441f046-0740-4ff0-a310-56a9963abcea

PR #586, opened 2024-05-07
Added hyperlinks to Moves, Abilities, and Species enums

bennybroseph:
Yes, I'm aware of how bad this looks like in the file itself, but it enables a nice neat little hyperlink shown below to the bulbapedia page when you hover over a move. 

image

I feel like the added convenience is worth making the file harder to read considering nobody is ever going to touch that file anyway. I plan on adding a python script to finish the rest of them, but didn't want to actually sink time into this if we absolutely despise it.


Flashfyre:
Do you happen to know if this gets filtered out of the compiled build so it doesn't increase the file size?

contra1337:
> Do you happen to know if this gets filtered out of the compiled build so it doesn't increase the file size?

looks like these comments are all stripped out in the bundled js file

bennybroseph:
added a python script to automate it

Madmadness65 (approved):

PR #593, opened 2024-05-07
Implement Beat Up

happinyz:
This PR implements the move Beat Up. It's a pretty unique move with mechanics not found anywhere so a lot of this code is tightly coupled together.

[How Beat Up works](https://pokemondb.net/move/beat-up)

https://github.com/pagefaultgames/pokerogue/assets/59787978/fd0c4692-1b64-4488-ab5a-9ffc58a33e93

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -804,6 +804,7 @@ export enum MultiHitType {
   _3,
   _3_INCR,
   _1_TO_10,
+  BEAT_UP,
 }
 
 export class HealAttr extends MoveEffectAttr {
@@ -1069,6 +1070,11 @@ export class MultiHitAttr extends MoveAttr {
             hitTimes = 10;
         }
         break;
+      case MultiHitType.BEAT_UP:
+        // No status means the ally pokemon can contribute to Beat Up
+        hitTimes = user.scene.getParty().reduce((total, pokemon) => {
+          return total + (pokemon.id === user.id ? 1 : pokemon?.status && pokemon.status.effect !== StatusEffect.NONE ? 0 : 1)
+        }, 0);
     }
     (args[0] as Utils.IntegerHolder).value = hitTimes;
     return true;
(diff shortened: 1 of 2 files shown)


happinyz:
@bennybroseph @Madmadness65 Could I get a review on this? Thanks!

bennybroseph (changes requested):

happinyz:
@bennybroseph Could you take another look at this? thanks!

src/data/move.ts
```diff
@@ -1866,6 +1872,45 @@ export class MovePowerMultiplierAttr extends VariablePowerAttr {
...
+ * @returns The base power of the Beat Up hit.
+ */
+const beatUpFunc = (user: Pokemon, allyIndex: number): number => {
+  const party = user.scene.getParty();
+
+  for (let i = allyIndex; i < party.length; i++) {
+    const pokemon = party[i];
+
+    // The user contributes to Beat Up regardless of status condition.
+    // Allies can contribute only if they do not have a non-volatile status condition.
```
bennybroseph, line 1888: this is referenced in two places, is this by mistake?

src/data/move.ts
```diff
@@ -1069,6 +1070,11 @@ export class MultiHitAttr extends MoveAttr {
             hitTimes = 10;
         }
         break;
+      case MultiHitType.BEAT_UP:
+        // No status means the ally pokemon can contribute to Beat Up
+        hitTimes = user.scene.getParty().reduce((total, pokemon) => {
+          return total + (pokemon.id === user.id ? 1 : pokemon?.status && pokemon.status.effect !== StatusEffect.NONE ? 0 : 1)
+        }, 0);
```
bennybroseph, line 1077: here as well

bennybroseph (changes requested):

src/data/move.ts
happinyz: No, we do this in line 1075 to determine how many times Beat Up can hit, then do it again here to determine which party member's turn it is to hit

bennybroseph (approved):

PR #596, opened 2024-05-07
corrected u turn and flip turn crash

### LINKED ISSUE #273: U-turn causes game freeze repeatedly (cannot read properties of undefined (reading scene)
pr
Browser : Brave (chromium)

Situation : 
- Battle start
- Meowscarada uses U-turn against Skuntank
- Switching from Meowscarada to any other pokemon
- Skuntank uses explosion
- Either my pokemon dies or not, game freeze + scene is undefined in console

Tried +10 times, happenned everytime.
Tried with default configuration (normal speed, english language, ...) : it happenned aswell

### LINKED ISSUE #584: Iron Barbs make the game crash
### Situation :
Here is the current situation. Floor 194, classic run. My Ferrothorn VS Foe Iron Bundle. I take him low with Power whip (or anything else) and next turn he hit Flip Turn. 

### Issue :
He go back and the game freeze (crashed, forced to reload the page).

### Conclusion :
I think the game was trying to deal the Iron Barbs Damages to Iron Bundle despite of it was'nt on the battlefield. So the game don't understand what to do and is trying to deal damage at something that can not recieve damages.

YounesM:
Fixes #273 
Fixes #584

### DIFF
--- a/src/field/damage-number-handler.ts
+++ b/src/field/damage-number-handler.ts
@@ -13,7 +13,7 @@ export default class DamageNumberHandler {
   add(target: Pokemon, amount: integer, result: DamageResult | HitResult.HEAL = HitResult.EFFECTIVE, critical: boolean = false): void {
     const scene = target.scene;
 
-    if (!scene.damageNumbersMode)
+    if (!scene?.damageNumbersMode)
       return;
 
     const battlerIndex = target.getBattlerIndex();


YounesM:
Not sure if it's the entended behavior but Flip Turn / U-turn makes the **wild** pokemon using it flee the fight. If it's not the entended behavior there are some more changes that need to be added to these two move. Nonetheless, this should fix the crashes at least

chacolah:
Testing Done:
-Tested Ferrothorn with Iron Barbs being hit with both U-Turn and Flip Turn (issue #584)
-Tested pokemon with U-Turn swapping when being targeted with a suicide move (issue #273)

No error logs or crashes present.

src/field/damage-number-handler.ts
```diff
@@ -13,7 +13,7 @@ export default class DamageNumberHandler {
   add(target: Pokemon, amount: integer, result: DamageResult | HitResult.HEAL = HitResult.EFFECTIVE, critical: boolean = false): void {
     const scene = target.scene;
 
-    if (!scene.damageNumbersMode)
+    if (!scene?.damageNumbersMode)
```
bennybroseph, line 16: I'm worried that value being null is causing more issues down the line. Why would scene be null/undefined here?

bennybroseph (changes requested):

src/field/damage-number-handler.ts
YounesM: Not 100% sure about what I'm going to say since I'm travelling and don't have direct access to my PC but when I looked a it seemed like that since Flip Turn and U-Turn causes the Wild Pokemon to flee, the scene unloads before the damage is applied (the pokemon is remove from the UI)

As I said in my previous comment I'm not sure if that's the intended behavior of those moves. I believe that these moves do not make wild pokémon leave the battle in the main line games.

It's maybe preferable to change UTurn / Filp turn accordingly so but this seemed to be a quickfix that doesn't change much for the time being

bennybroseph: Yeah, that's a tough call. I think we should fix the root cause. This may cover it up and cause other problems to be hard to identify later.

YounesM: I did some more investigating and it seems like what's happening is :

- Flip Turn / U-Trun is used and `ForceSwitchOutAttr` triggers, damage is applied, pokemon flees (scene is destroyed here).
- Damage is taken by pokemon with Iron Barbs / Rough Skin and the ability triggers `PostDefendContactDamageAbAttr`
- `PostDefendContactDamageAbAttr` tries to apply damage to a pokemon that has already fled, but doesn't find the scene.

The only thing I can think of to fix this is to find a way to trigger the move in two steps :
Step 1 : Move deals damage
(Ability triggers and deals damage)
Step 2 : Move switches out the Pokémon (Maybe split the animation in 2 phases too?)

I don't know if it is possible right know looking at how the code is structured to have a move to pause and wait until the damage phase is done to decide what happens next. This seems to be a very specific requirement for these two moves.

On the other hand it's very weird to me that the game tries to display damage numbers without checking if a scene is present. Isn't it better to check if a scene is present before trying to display anything anyway?

bennybroseph (approved):

PR #597, opened 2024-05-07
Make status damage skip endure

ElMoustacho:
Set **preventEndure** flag to **true** when pokémon takes damage from status

Fixes [Focus Band keeping pokemon alive through status](https://trello.com/c/NmVfN8xy/22-focus-band-keeping-pokemon-alive-through-status-damage)


src/phases.ts
```diff
@@ -2961,7 +2961,7 @@ export class PostTurnStatusEffectPhase extends PokemonPhase {
             break;
         }
         if (damage) {
-          this.scene.damageNumberHandler.add(this.getPokemon(), pokemon.damage(damage));
+          this.scene.damageNumberHandler.add(this.getPokemon(), pokemon.damage(damage, false, true));
```
bennybroseph, line 2964: Can we add a comment stating why this function passes the arguments that it does?

bennybroseph (changes requested):

bennybroseph (approved):

PR #598, opened 2024-05-07
Override - Modifiers & Held items

Greenlamp2:
chrome_Uhuddx6dQd


src/overrides.ts
```diff
@@ -11,8 +11,11 @@ export const STARTING_LEVEL_OVERRIDE = 0;
 export const STARTING_WAVE_OVERRIDE = 0;
 export const STARTING_BIOME_OVERRIDE = Biome.TOWN;
 export const STARTING_MONEY_OVERRIDE = 0;
+export const STARTING_MODIFIER_OVERRIDE = []; // ['EXP_SHARE', 'GOLDEN_EXP_CHARM']
```
molker, line 14: instead of separate const for modifier list and quantity list can you just make it an array of objects with a modifier value and quantity value

bennybroseph (changes requested):
Some of these comments are short enough to go off to the side of the line of code. Let's clean those up

bennybroseph (approved):

bennybroseph:
Looks good to me. Only adds items when populated. Works on both enemy pokemon in a double battle. Applies on every floor. Very useful for dev

Some notes for a future pull request of v 1.1:
- Needs a way to autocomplete the names of items/modifiers. Simple suggestion is an enum of the strings in a file like overrides-enums.ts or whatever
  - A more elegant solution may a possible v 1.2
- Currently applies the same held items to both pokemon in a double battle. needs a second override for the other pokemon

molker:
Implementing as part of https://github.com/pagefaultgames/pokerogue/pull/603

PR #614, opened 2024-05-07
Fix charge

TempsRay:
1. Charge accidentally had the `failOnOverlap` field as true instead of false which meant you would not gain the special defense boost if you were already charged
2. The charged status was incorrectly trying to be removed on the attacked Pokemon instead of the attacker


bennybroseph:
Looks good to me. Opponent charge never stopped my own. Had a discussion about if it was boosting damage, and sorted it out.

Madmadness65 (approved):

PR #619, opened 2024-05-07
Implement Ball Fetch

Fabpharos:
Could have some niche use. This implementation tracks the most recently thrown type of Pokeball that failed to catch a Pokemon, letting a Pokemon with Ball Fetch fetch it. Triggers once per battle, at the end of the turn. Tested with all types of Pokeballs, besides Master Balls obviously, and tested a mon switching in when a ball was thrown on a previous turn.

Pokemon can't hold Pokeballs so they just put it directly into your inventory.

https://m.bulbapedia.bulbagarden.net/wiki/Ball_Fetch_(Ability)


src/data/ability.ts
```diff
@@ -2106,6 +2107,22 @@ export class PostTurnFormChangeAbAttr extends PostTurnAbAttr {
   }
 }
 
+export class FetchBallAbAttr extends PostTurnAbAttr {
+  constructor() {
+    super();
+  }
+  applyPostTurn(pokemon: Pokemon, passive: boolean, args: any[]): boolean {
+    let lastUsed = pokemon.scene.currentBattle.lastUsedPokeball;
+    if(lastUsed != null && pokemon.isPlayer) {
```
TempsRay, line 2116: I believe you can check if lastUsed is null with just
```suggestion
    if(!lastUsed && pokemon.isPlayer) {
```

TempsRay (approved):
lgtm

src/data/ability.ts
bennybroseph: Nah, I think that's backwards. You would remove the exclamation mark.

bennybroseph (changes requested):

src/data/ability.ts
TempsRay: I think you're right. Can't believe I made such a typo 🤦

src/data/ability.ts
```diff
@@ -2106,6 +2107,22 @@ export class PostTurnFormChangeAbAttr extends PostTurnAbAttr {
   }
 }
 
+export class FetchBallAbAttr extends PostTurnAbAttr {
```
TempsRay, line 2110: Not sure the exact doc requirements, but here's something

```ts
/**
 * Grabs the last failed Pokeball used
 * @extends PostTurnAbAttr
 * @see {@link applyPostTurn}
 */
export class FetchBallAbAttr extends PostTurnAbAttr {
...
  /**
   * Adds the last used Pokeball back into the player's inventory
   * @param pokemon the Pokemon with this ability
   * @param passive n/a
   * @param args n/a
   * @returns true if player has used a pokeball and this pokemon is owned by the player
   */
  applyPostTurn(pokemon: Pokemon, passive: boolean, args: any[]): boolean
```

TempsRay (commented):
Hope the doc I added helps

PR #622, opened 2024-05-08
Candy/Ribbon UI Improvement

Akuma-Reiki:
Added candy numbers to the summary screen
Added ribbon to summary screen
Moved ribbon on starter select screen

### DIFF
--- a/src/ui/summary-ui-handler.ts
+++ b/src/ui/summary-ui-handler.ts
@@ -50,6 +50,8 @@ export default class SummaryUiHandler extends UiHandler {
   private candyShadow: Phaser.GameObjects.Sprite;
   private candyIcon: Phaser.GameObjects.Sprite;
   private candyOverlay: Phaser.GameObjects.Sprite;
+  private candyCountText: Phaser.GameObjects.Text;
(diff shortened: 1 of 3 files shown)


src/ui/summary-ui-handler.ts
```diff
@@ -275,6 +289,9 @@ export default class SummaryUiHandler extends UiHandler {
       this.splicedIcon.on('pointerout', () => (this.scene as BattleScene).ui.hideTooltip());
     }
 
+    if(this.scene.gameData.starterData[this.pokemon.species.getRootSpeciesId()].classicWinCount > 0)
```
TempsRay, line 292: Bit of an edge case here. I think it makes sense to check both root species.
```suggestion
    if(this.scene.gameData.starterData[this.pokemon.species.getRootSpeciesId()].classicWinCount > 0 && this.scene.gameData.starterData[this.pokemon.species.getRootSpeciesId(true)].classicWinCount > 0)
```

TempsRay (approved):
LGTM, just one minor comment about checking both root species before displaying the ribbon

Issue #481, opened 2024-05-05
Opponent pokemon sprite disappears after using Roar


PaulBeslin:
When an opponent pokemon (Trainer or wild) uses Roar, its sprite disappears. The pokemon is still here and behaves properly, except for the sprite.
Other moves such as whirlwind do not seem to be affected.
Using Roar as the player works as intended.

To reproduce, override the opponent used move to be Roar.

![image](https://github.com/pagefaultgames/pokerogue/assets/71925152/ef3c55a5-043e-4a46-a3c9-49a444a17147)

happinyz:
PR for the fix: https://github.com/pagefaultgames/pokerogue/pull/625

happinyz:
> PR for the fix: #625

I suspect this is something wrong with the animation file, rather than any move application/attribute logic. When changing Roar's attributes to the exact same ones as Whirlwind's, this bug still happens. However, when I switch Roar's animation to use Whirlwind's, this bug does not appear.

I tried playing around with the animation file a bit but I'm not super familiar with how the animation framework works. I've closed my PR for now. I welcome anyone who's more familiar with the animation system to tackle this bug.

pom-eranian:
Copy pasted from discord since the issue was brought up there. 

I already tried changing or removing the anim file's parameters one by one and it still happened.
The behavior I saw every single time is that no matter who used the move, player or opponent, the opponent disappeared after the anim ends and before the player team UI shows up.
Then I tried extending and shortening the amount of frames, and every time the opponent would disappear after the anim was done.

_However having another move do the same thing is interesting.  Does anyone know if baneful bunker already did this or is it a new bug?_ It's new.

The only part that can be fixed through the anim file is the little blip at the end like this: https://github.com/pagefaultgames/pokerogue/pull/1660#issuecomment-2154936069

Snailman11:
Link to Discord (Bug report deleted so chat link) https://discord.com/channels/1125469663833370665/1125894949020381285/1267314975953780808

flx-sta:
After some debugging I noticed that only `Roar` and `Dragon Tail` seem to do a false-swap of `user` and `target` during the animation. 
I was fighting with `Charmander` vs a wild `Surfskit`
I'm going to dump some debug messages in here to show what I mean

------

### Roar

```
anim:  PRAS- Roar 
    User Charmander  | this.user: Surskit 
    Target Surskit  | this.target: Charmander
```

------

### Dragon Tail

```
[...]

PR #625, opened 2024-05-08
Fix opponent sprite disappearing when using Roar

happinyz:
This PR addresses the issue raised here: https://github.com/pagefaultgames/pokerogue/issues/481

https://github.com/pagefaultgames/pokerogue/assets/59787978/8f899953-d9c2-4fa7-93dc-376e0ee3f615

Verified that this is also working correctly when the player pokemon uses roar

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -4167,8 +4167,7 @@ export function initMoves() {
       .target(MoveTarget.ALL_NEAR_ENEMIES),
     new StatusMove(Moves.ROAR, Type.NORMAL, -1, 20, -1, -6, 1)
       .attr(ForceSwitchOutAttr)
-      .soundBased()
-      .hidesTarget(),
+      .soundBased(),
     new StatusMove(Moves.SING, Type.NORMAL, 55, 15, -1, 0, 1)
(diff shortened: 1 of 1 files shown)


PaulBeslin:
Removing `hidesTarget` makes the target reappear for a frame

https://github.com/pagefaultgames/pokerogue/assets/71925152/1aeaa3fb-187a-4dc3-bdb5-3c10defe86aa

happinyz:
> Removing `hidesTarget` makes the target reappear for a frame
> 
>  2024-05-08.12-04-08.mp4

Oh, thanks for the catch. Didn't notice because I was on 5x animation speed. I'll update the PR accordingly

happinyz:
I suspect this is something wrong with the animation file, rather than any mov application/attribute logic. When changing Roar's attributes to the exact same ones as Whirlwind's, this bug still happens. However, when I switch Roar's animation to use Whirlwind's, this bug does not appear.

I tried playing around with the animation file a bit but I'm not super familiar with how the animation framework works. I'm closing this PR for now, and updating the original issue with my finds. Hopefully this helps someone if they decide to pick it up, if not I might come back to this in the future.

PR #633, opened 2024-05-08
Implements bad dreams

Sbug98:
Hi,
This is my attempt at implementing Bad Dreams.
Hopefully the video showcases some worst case scenarios of the ability and it's enough proof of testing. I'm a bit worried of the interaction with healer, I wasn't able to test it as it is not yet implemented, but if the priority is consistent the order of activation should handle the rest. 
https://github.com/pagefaultgames/pokerogue/assets/68464158/2683bb74-1998-4cd1-a3a4-0ec3c2d25a91


src/data/ability.ts
```diff
@@ -2106,6 +2106,22 @@ export class PostTurnFormChangeAbAttr extends PostTurnAbAttr {
...
 }
 
+export class PostTurnHurtIfSleepingAbAttr extends PostTurnAbAttr {
+
+  applyPostTurn(pokemon: Pokemon, passive: boolean, args: any[]): boolean | Promise<boolean> {
+    let hadEffect: boolean = false;
+    for(let oppo of pokemon.getOpponents()) {
+      if(oppo.status !== undefined && oppo.status.effect == StatusEffect.SLEEP) {
+        oppo.damageAndUpdate(Math.ceil(oppo.getMaxHp() * (1 / 8)), HitResult.OTHER);
+        pokemon.scene.queueMessage(getPokemonMessage(oppo, ' is tormeted!'));
```
PinsonJonas, line 2116:
```suggestion
        pokemon.scene.queueMessage(getPokemonMessage(oppo, ' is tormented!'));
```

PinsonJonas (changes requested):

PinsonJonas (approved):

src/data/ability.ts
```diff
@@ -2106,6 +2106,22 @@ export class PostTurnFormChangeAbAttr extends PostTurnAbAttr {
   }
 }
 
+export class PostTurnHurtIfSleepingAbAttr extends PostTurnAbAttr {
+
+  applyPostTurn(pokemon: Pokemon, passive: boolean, args: any[]): boolean | Promise<boolean> {
+    let hadEffect: boolean = false;
+    for(let oppo of pokemon.getOpponents()) {
+      if(oppo.status !== undefined && oppo.status.effect == StatusEffect.SLEEP) {
```
TempsRay, line 2114: nit: usually I see `opp` instead of `oppo` but that's minor. Also most comparisons use 3 equals signs rather than just 2

src/data/ability.ts
```diff
@@ -2106,6 +2106,22 @@ export class PostTurnFormChangeAbAttr extends PostTurnAbAttr {
...
+        oppo.damageAndUpdate(Math.ceil(oppo.getMaxHp() * (1 / 8)), HitResult.OTHER);
```
TempsRay, line 2115: Smogon says it rounds down so it should be Math.floor rather than Math.ceil. Also, for the edge case of wonder guard this should probably do a minimum of one damage.
`oppo.damageAndUpdate(Math.Max(1, Math.floor(oppo.getMaxHp() / 8)), HitResult.OTHER);`

TempsRay (changes requested):
Some minor changes but otherwise looks fine

bennybroseph (changes requested):

src/data/ability.ts
```diff
@@ -2106,6 +2106,27 @@ export class PostTurnFormChangeAbAttr extends PostTurnAbAttr {
...
+      //check if anyone of them is sleeping
+      if(opp.status !== undefined && opp.status.effect === StatusEffect.SLEEP) {
+        //if so, deal a min of 1 damage and queue a message
+        opp.damageAndUpdate(Math.floor(Math.max(1, opp.getMaxHp() / 8)), HitResult.OTHER);
+        pokemon.scene.queueMessage(getPokemonMessage(opp, ' is tormented!'));
+        hadEffect = true;
+      }
+    
+    }
+    //returns true only if one of the opponents were sleeping
```
TempsRay, line 2124: I don't think this is what Benny meant by comments. Here's what he means by tsdoc
```ts
/**
 * Attribute used for abilities (Nightmare) that damages the opponents for being asleep
 */
export class PostTurnHurtIfSleepingAbAttr extends PostTurnAbAttr {

/**
 * Deals damage to all sleeping opponents equal to 1/8 of their max hp (min 1)
 * @param {Pokemon} pokemon Pokemon that has this ability 
 * @param {boolean} passive N/A
 * @param {any[]} args N/A
 * @returns {boolean} true if any opponents are sleeping
 */
  applyPostTurn(pokemon: Pokemon, passive: boolean, args: any[]): boolean | Promise<boolean> {
```

TempsRay (approved):
Code looks good to me but your tsdoc doesn't seem quite right. I've commented what the appropriate TSdoc looks like. Your // comments aren't needed since your code is easy enough to understand

src/data/ability.ts
Sbug98: Ah I see, that makes sense. I'll update the PR with your docs. Thanks

src/data/ability.ts
```diff
@@ -2106,6 +2106,33 @@ export class PostTurnFormChangeAbAttr extends PostTurnAbAttr {
...
+   * @param {boolean} passive N/A
+   * @param {any[]} args N/A
+   * @returns {boolean} true if any opponents are sleeping
+   */
+  applyPostTurn(pokemon: Pokemon, passive: boolean, args: any[]): boolean | Promise<boolean> {
+    let hadEffect: boolean = false;
+    for(let opp of pokemon.getOpponents()) {
+      if(opp.status !== undefined && opp.status.effect === StatusEffect.SLEEP) {
+        opp.damageAndUpdate(Math.floor(Math.max(1, opp.getMaxHp() / 8)), HitResult.OTHER);
+        pokemon.scene.queueMessage(getPokemonMessage(opp, ' is tormented!'));
```
TempsRay, line 2126: Here's where the localization files are https://github.com/pagefaultgames/pokerogue/blob/main/src/locales/en/ability-trigger.ts

Sbug98: Using the getTriggerMessage method I'm not sure how to trigger the message for each opponent since I can only give one value as the return. Could you suggest an approach to this problem? I can't seem to find any reference using localization for more than one Pokémon.

Sbug98: I added translations but they may be inaccurate from the games.

bennybroseph:
has conflicts currently. need to resync and resolve

bennybroseph (approved):

Sbug98 (commented):
I've resolved the conflicts and resynced my PR branch

PR #641, opened 2024-05-08
Make max and z moves unable to be chosen with metronome

Xavion3:
We never actually set these to not be choosable with metronome whoops.


bennybroseph:
For this change, would it be better to instead have .unimplemented() automatically do this? Is there any reason an unimplemented move should not have this attribute?

Xavion3:
Unimplemented does do this incidentally, but not all of these are unimplemented because some of them are technically implemented, just meant to be unusable.

Madmadness65 (approved):

PR #642, opened 2024-05-08
Prevent doubles trainers generating with one pokemon

Xavion3:
Difficult to test, but as far as I can tell this is safe. Fixes this longstanding bug by just forcing doubles trainers to have at least two pokemon.


bennybroseph:
Tested on waves 5 and 16 over and over with double battles override on. Generated the proper number of pokemon each time. If it still occurs on live after this, we'll need more data.

Madmadness65 (approved):

Issue #627, opened 2024-05-08
ESLint errors


Sylux6:
As the title says: ESlint returns errors. As of today, the existence of ESLint in the project looks useless since there is no lint check against the code. Also I have noticed code style inconsistencies across the project files, so having a code formatter would help the contributors to agree on the same rules.

I suggest to:
- Add a code formatter tool
- Add a lint step in the github workflow

Sylux6:
@Greenlamp2 Thank you so much, I stayed away from the codebase solely for that reason because the code style was a real mess

PR #643, opened 2024-05-08
Replace ESLint by biome

Sylux6:
This PR adresses https://github.com/pagefaultgames/pokerogue/issues/627
It replaces the ESLint dependency by [biome ](https://biomejs.dev/) to format and lint the code (it basically does the job of ESlint + prettier). Automatic changes have been applied to the whole project while files with unsafe changes have been ignored to prevent unexpected new bugs. Moreover, a new step has been introduced in the github deploy workflow to prevent future merges with malformed code that should pass the biome check.

The commit with the big changes has been ignored (see `.git-blame-ignore-revs`) to not lose the original authors and keep the git history clean.

The next steps are to remove one by one the files from the ignore list in `biome.json` and make the necessary changes manually. Also, having a pre-commit hook script to auto format code on commit may be useful for the contributors.


biome.json
```diff
@@ -0,0 +1,160 @@
+{
+  "$schema": "https://biomejs.dev/schemas/1.7.3/schema.json",
+  "organizeImports": {
+    "enabled": true
+  },
+  "files": {
+    "maxSize": 2097152,
```
Sylux6, line 7: 1024 * 2048 (2MB)

bennybroseph:

PR #653, opened 2024-05-08
Added repels :

Manon-Oreins:
* Prevents double battle from happening by setting the maximum range of the random  managing the double battle chances at Number.MAX_SAFE_INTEGER.
* Does not prevents lure appearance as I think it allows more play options
* Renamed the Modifier class DoubleBattleChanceBoosterModifier into DoubleBattleChanceModifier using refactor
* Added an integer argument to DoubleBattleChanceModifier class called 'modifier' (maybe change the name), that indicates to apply the repel effects if negative, or the lure effect if positive. It allows to possibly use it at a multiplicative modifier if you do not want repels to make the chances fall to 0 but instead diminish the chances by 2 or 3 depending on repel type. Same goes for the lures.

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -4,7 +4,7 @@ import { NextEncounterPhase, NewBiomeEncounterPhase, SelectBiomePhase, MessagePh
 import Pokemon, { PlayerPokemon, EnemyPokemon } from './field/pokemon';
 import PokemonSpecies, { PokemonSpeciesFilter, allSpecies, getPokemonSpecies, initSpecies, speciesStarters } from './data/pokemon-species';
 import * as Utils from './utils';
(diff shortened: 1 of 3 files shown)


PaulBeslin:
Hi, there's already a PR opened for repels implementation: #544 
Not sure which one we should keep though, so probably don't close this one yet until someone provides an opinion

ojukeihpos:
Hi, I wrote the PR for #544 .

I tested your PR locally, and specifically checked `doubleBattleChance.value` as it was being sent as a parameter to `Utils.randSeedInt`. By setting the maximum range to `Number.MAX_SAFE_INTEGER`, `Utils.randSeedInt` will still have a chance (however statistically unlikely) of returning 0, which is what causes the double battles to occur. My implementation overrides the value returned by `Utils.randSeedInt` if it finds that the player has a Repel in their modifiers list, which prevents it entirely.

My implementation also creates a separate modifier and modifier type named `DoubleBattleChancePreventerModifier(Type)`, which can be changed to suit the same purpose as your last point. I don't think it's a bad idea to possibly merge the modifiers into one, though.

bennybroseph:

PR #662, opened 2024-05-09
Spectral Thief stat-stealing added (Still Partial)

notpatchmaybe:
Spectral Thief now steals stats and outputs a message similar to the original game's if it does. Still marked partial as the move needs to have the phase where it deals damage moved to after the stats are stolen, not before, but I'm not familiar enough with the project yet to do this.


notpatchmaybe:
Woops, deleted a line by mistake for checking if the stats changed, ill close and reopen when I've fixed it

PR #665, opened 2024-05-09
Issue: Should be able to set hazards when target hidden

mattrossdev:
Previously if a target was in fly/dig/dive etc. and you used a hazard move the hazard would not get set and it would just skip over the move. 

Testing: 
- Can set hazard with hazard status move if target is in fly 
- Hazard move fails if already fully stacked (one stealth rock, 3 spikes etc.)
- Attacking moves that set hazards still miss and do not set hazard (ex: stone axe)


mattrossdev:
Need to resolve merge conflicts

mattrossdev:
Closing this to revisit later. After discussing, we want hazard moves to not require a target (they should get set for future waves if you use the move after all targets have fainted).

PR #668, opened 2024-05-09
Gorilla Tactics implementation

pixelizedgaming:
Still a few rough edges here and there I feel but I've tested it quite a bit and it works for all I can tell. Sorry if this is a bit messy, I've never written javascript before


bennybroseph:
Looks like this needs to be remade. There's a ton of non-code related changes. Feel free to make a new request once this is resolved. Keep in mind we have new comment standards that you can check out here: https://discord.com/channels/1125469663833370665/1239957703820836975

PR #672, opened 2024-05-09
Add wordwrap to Chinese words (debug)

mercurius-00:
Add wordwrap to Chinese words


TempsRay:
Looks like this PR can be closed since 683 added lien breaks

PR #673, opened 2024-05-09
Added local system save to reduce requests made

GamrayW:
Base implementation of what has been discussed in the [discord thead](https://discord.com/channels/1125469663833370665/1237843658628141256).  
TLDR: Make the system save less often to reduce the amount of requests to the server by saving the system data locally, and syncing with the server every 5 waves / on logout (no impact on session requests for now)

Currently, it's only the base implementation, **no encryption** is made to the local save, this needs to be tested more to identify edge cases before adding encryption on top, since encryption will need more code changes.
I don't think it should be merged to main right now, maybe in a separate branch so we can work on it more. (because right now no encryption = easily modifiable)

## What I have tested
✅ Normal game 
✅ Reload mid wave (= local data will be fetched)
✅ Reload after sync (= remote data will be fetched)
✅ Logout (= local data is synced)
✅ New login (=local data is overwrote with the new system data)
✅ Login on previous account (= remote data is synced)

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -116,13 +116,26 @@ export class LoginPhase extends Phase {
         }
         return null;
       } else {
-        this.scene.gameData.loadSystem().then(success => {
-          if (success || bypassLogin)
-            this.end();
-          else {
-            this.scene.ui.setMode(Mode.MESSAGE);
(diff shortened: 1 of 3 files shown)


GamrayW:
closing this as it's out of date

PR #676, opened 2024-05-09
Bouncy Bubble Description Update

paulcazac:
Updated the description of the move Bouncy Bubble to state the correct heal amount as per the move update a little while ago.
~~50%~~ -> **100%**


Flashfyre (changes requested):
This ought to be changed for all languages so it doesn't get left alone and forgotten about. Should be easy enough to translate with DeepL if necessary.

PR #670, opened 2024-05-09
Add wordwrap to Chinese words

mercurius-00:
Since Phaser.Text's wordwrap doesn't work with Chinese words, ''\n' has been added to move.ts and ability.ts

Before:
image

After:
image

PR #683, opened 2024-05-09
Modified Chinese translation and added line breaks.

Somod1:
Modified some Chinese mistranslations, added line breaks to tutorials, move, etc. to ensure the correct display of text.
Before:
before
After:
after


justhil:
Someone has already done it.

zjw0231:
> Someone has already done it.

but it still have this problem
5d876781e4853852c02abdc64ee086a

Somod1:
> Someone has already done it.

#670 Only does line breaks for skill descriptions.
#672  Already did abitlity and move part, but other .ts files in zh_CN didn't do it, such as tutorial.ts, menu-ui-handlerde.ts, etc. In addition to that, some Pokémon mistranslations were fixed, such as ‘保母曼波’ should be ‘保姆曼波’, ‘火爆猴’ should be ‘火暴猴’, and the descriptions of some move.ts and ability.ts have been added. Other than that, #672 is currently in the passed phase and has not yet been merged.
Before:
before1
before2


After:
after1
after2

Flashfyre (approved):

justhil:
I think some changes to the translations of Pokémon names may not be entirely necessary. For example, "拉达" and "阿罗拉拉达" The 阿罗拉 prefix might be necessary for some people. Also, names like "谜拟q" and "谜拟丘" where the former is more popular for me and many others.

justhil:
the addition of newline characters is a great modification！

Somod1:
> I think some changes to the translations of Pokémon names may not be entirely necessary. For example, "拉达" and "阿罗拉拉达" The 阿罗拉 prefix might be necessary for some people. Also, names like "谜拟q" and "谜拟丘" where the former is more popular for me and many others.

If add a prefix it will be displayed incompletely in the game, and the English version just doesn't have a prefix. The other name changes are mainly for consistency with the Pokémon Chinese wiki.

PR #688, opened 2024-05-09
Implements Grudge + some changes to how data is cleared on faint

shayebeadling:
Implements Grudge

1. Adds a GrudgeTag in battler tags that lasts 1 turn and will trigger it's pp-drain if the tag holder is koed by an attack move.
2. Delays clearing summonData of pokemon until after lapsing FAINT battle tags, previously these tags were not working because the mons tags were already cleared when lapseTags was called.

It may be worth checking other tag related moves, especially ones that hook into onRemove to function.


bennybroseph (changes requested):

flx-sta:
@shayebeadling are you still working on this?

flx-sta:
No response within more than a month. 
Considered stale

Issue #591, opened 2024-05-07
Reviver seed not proccing off self damage.


DarmokJaladd:
I used wood hammer on my rillaboom with reviver seed, it died from recoil and did not revive. Unsure if this is intended, but I wanted to mention it regardless.

TempsRay:
Revival seed only procs if you get KO'd by a move by the opponent. It will not save you from weather, hazards, or recoil

JoannaFalkowska:
> Revival seed only procs if you get KO'd by a move by the opponent. It will not save you from weather, hazards, or recoil

...Why though? Isn't this a bug? It doesn't make much sense why would it proc off direct hits but not off other things.

Tempo-anon:
That's just the way it's implemented right now as a design choice. Lets you preserve your revival seed by fainting to a non direct attack

PR #689, opened 2024-05-09
Fix Issue #591 The Pokemon is now revived when it holds a reviver seed an…

CodeTappert:
…d kills itself with something like explosion

This fixes https://github.com/pagefaultgames/pokerogue/issues/591 by checking if a pokemon is fainted, then checking if it holds a reviver seed and triggering the effect like with the prevent endure.

Also added to the faint phase that it only triggers if the pokemon is actually fainted. (Otherwise the effect would trigger but the pokemon would still faint)

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -3077,8 +3077,28 @@ export class DamagePhase extends PokemonPhase {
             this.getPokemon().updateInfo().then(() => this.end());
         }
       });
-    } else
+    } else {
       this.getPokemon().updateInfo().then(() => this.end());
+    }
+
+    // Check if the damage was enough to faint the pokemon
+    if (this.getPokemon().isFainted()) {
+      // Check if the pokemon has a reviver_seed
+      const hasReviverSeed = this.getPokemon().getHeldItems().find(i => i.type.id === 'REVIVER_SEED');
+
+        // If the Pokémon has a reviver seed, consume it and revive the Pokémon
+      if (hasReviverSeed) {
+        const instantReviveModifier = this.scene.applyModifier(PokemonInstantReviveModifier, this.player, this.getPokemon()) as PokemonInstantReviveModifier;
+
+
+        if (instantReviveModifier) {
+          if (!--instantReviveModifier.stackCount)
+            this.scene.removeModifier(instantReviveModifier);
+          this.scene.updateModifiers(this.player)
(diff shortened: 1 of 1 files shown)


bennybroseph:
I believe this is intentional.

CodeTappert:
> I believe this is intentional.

But this doesnt make sense no? The item only says that it revives a pokemon when the hp drop to 0. Either the description needs to be changed or we should implement the fix i made

bennybroseph:
Explosion doesn't cause your HP to drop to 0, it causes you to faint.

CodeTappert:
Well

`  REVIVER_SEED: () => new PokemonHeldItemModifierType('Reviver Seed', 'Revives the holder for 1/2 HP upon fainting',` - https://github.com/pagefaultgames/pokerogue/blob/a27822b6243a9dacd418a32a9b36da02b6c79f04/src/modifier/modifier-type.ts#L802

i looked it up and this is in the code for the reviver seed. So it should work on any faint.

bennybroseph:
We should probably have a discussion on what wording to use for this item then.

CodeTappert:
> Collaborator

Something llike:

`Revives the holder for 1/2 HP upon reaching 0hp by a enemy attack`

Of course in better english than that.

Mani7602:
I think the revive seed should work as its name applies in every "way" a mon can die. This also applies to causes of fainting like explosion or sandstorm. This would essentially make this item good in the long run. Weather hazards are the only big threat in a long endless run since most people rely on a spliced mon with sturdy and/ or focus band which both is denied by weather hazards upon activation. So the only purpose the reviver seed can serve is in this case reviving the mon that has fainted due to this.

bennybroseph:
After discussion we've decided the item should prevent fainting on all moves that you could normally use the move [Endure](https://bulbapedia.bulbagarden.net/wiki/Endure_(move)) on and survive with 1 HP.

From my testing in pokerogue this is how it compares to Endure currently:
0KO doesn't revive you which is incorrect. 
Burn/Poison revives you which is incorrect

Weather damage doesn't revive you which is correct.
Self-Destruct doesn't revive you with is correct.

If you can provide source on any of this information, please do so. I'm going to close out this request, but you are more than welcome to reopen it with these changes in mind or submit a new one.

PR #691, opened 2024-05-09
Sacrifical Moves (that dont require a target like explosion or self d…

CodeTappert:
…estruct) now also work if the target is flying, diving etc.

There is also a new catagorie of moves. "SacrificalMovesOnHit" for all moves that need to hit for them to be sacrifical like MEMENTO

Fixes https://discord.com/channels/1125469663833370665/1238179985571975300/1238179985571975300


CodeTappert:
Solved the conflict with main

bennybroseph (changes requested):
Needs TSDoc documentation on all new classes and functions and code comments where applicable. Also needs some formatting fixes as there's a lot of open space between lines.

CodeTappert:
So i added the changes you requested to the best of my ability. (I dont know how a TSDoc for a class should look like. Please provide an example).

CodeTappert:
> So i added the changes you requested to the best of my ability. (I dont know how a TSDoc for a class should look like. Please provide an example).

@bennybroseph

src/data/move.ts
```diff
@@ -738,21 +738,75 @@ export class RecoilAttr extends MoveEffectAttr {
   }
 }
 
+
+/**
+ * Attribute used for moves which self KO the user regardless if the move hits a target
+ * */
```
TempsRay, line 744: nit: the bottom of the comment should be `**/` without the space in between

src/data/move.ts
```diff
@@ -738,21 +738,75 @@ export class RecoilAttr extends MoveEffectAttr {
...
     super(true, MoveEffectTrigger.PRE_APPLY);
   }
 
+    /**
+     * Applies the sacrificial effect to the user
+     * @param user Pokemon that used the move
+     * @param target The target of the move
+     * @param move Move with this attribute
+     * @param args N/A
+     */
```
TempsRay, line 756: This seems like it's 1 tabbed in further than it should be. I'd also suggest changing the main `Applies the sacrificial effect to the user` to be more concise like `Deals damage to the user equal to their current hp`. You're also missing the `@return` in the doc and the comment should be ended with `**/`

src/data/move.ts
```diff
@@ -738,21 +738,75 @@ export class RecoilAttr extends MoveEffectAttr {
...
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    user.damageAndUpdate(user.hp, HitResult.OTHER, false, true, true);
+	user.turnData.damageTaken += user.hp;
+
+    return true;
+  }
+
+  /**
+   *
```
TempsRay, line 765: This part should describe what's happening (the AI is calculating how good the move is). Bosses will always return -20 while otherwise it scales based on the user's hp ratio. The doc here is also a bit mis-tabbed I think

src/data/move.ts
```diff
@@ -738,21 +738,75 @@ export class RecoilAttr extends MoveEffectAttr {
...
+  /**
+   * Applies the sacrificial effect to the user
+   * @param user Pokemon that used the move
+   * @param target The target of the move
+   * @param move Move with this attribute
+   * @param args N/A
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+
+    // If the move hits a target, the user will self KO
```
TempsRay, line 794: Doesn't this line mean the opposite? If the move fails then the user does not faint and the function returns false

CodeTappert: You are correct.

src/data/move.ts
CodeTappert: Fixed. But tabbing was right imo (same as at other places)

src/data/move.ts
CodeTappert: Yes your idea for a description is better indeed!

src/data/move.ts
CodeTappert: Fixed

bennybroseph:
This makes weird behavior happen. Because you're targeting yourself, you take damage in parts. First from the move and then to fully faint. I'm about to mess around with this.

https://github.com/pagefaultgames/pokerogue/assets/13838608/048c622a-f82d-4cfd-8cf8-9c360e8556ca

bennybroseph:
Working to make this even better, but as I thought you definitely don't need to force target yourself. Self Inflicted damage is now applied after all other move effects and does not rely on a target in order to do so.


https://github.com/pagefaultgames/pokerogue/assets/13838608/9979a5fa-e610-434d-a0d2-927353bbd5a1

bennybroseph (approved):

bennybroseph:
I tried a few things, but any other changes were just out of scope for this PR. Changes are needed to how the system handles Fly, Dig, Dive, etc. before this works is implemented correctly. Everything works like it should though.

PR #692, opened 2024-05-09
Ability: Implemented pre-gen9 Protean/Libero.

paulcazac:
Added the partial tag (P) since the gen 9 update/nerf to Protean + Libero only allows it to happen once per switch in, which I was struggling to get working right now. 
Currently works the same as pre-gen9 nerf (pokemon's type changes every time it uses a new move).

Otherwise should be fully working as per Bulbapedia: 
[https://bulbapedia.bulbagarden.net/wiki/Protean_(Ability)](url)

Also implements a new move flag that was required for these abilities, which are moves that call other moves: 
[https://bulbapedia.bulbagarden.net/wiki/Category:Moves_that_call_other_moves](url)

Also works with the Toxic use-case of making toxic a guarantee hit, as per the Bulbapedia page.

(Not tested with Tera as there were no overrides in overrides.ts, but should work as Tera just overrides the pokemon's type anyway)

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -964,6 +964,34 @@ export class FieldPreventExplosiveMovesAbAttr extends AbAttr {
   }
 }
 
+export class PokemonTypeChangeAttr extends PreAttackAbAttr {
+  private condition: PokemonAttackCondition;
+
+  constructor(condition: PokemonAttackCondition){
+    super(true);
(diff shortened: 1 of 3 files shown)


bennybroseph (changes requested):

bennybroseph:

PR #697, opened 2024-05-09
Add Documentation to All Partial (P) Moves and Abilities

f-raZ0R:
This pull request adds comments to every partially implemented move or ability that does not already have one, explaining what about the move or ability still needs to be worked on.

Additionally, this adds the partial tag to Prankster and Pastel Veil, as both of these abilities do still need an aspect of them implemented (prankster not working on dark types, and pastel veil also affecting allies in a double battle).


happinyz:
Hmm if we're doing this, maybe we allow passing a string argument into `partial()` that explains what has not been implemented yet, and then we can include that info in the description of the move. Since we're already telling users if a move is partially implemented, I think it would be fine to also augment the move description with what still needs to be implemented. Maybe this would be exposing too much information to the user, would love to get some thoughts on this.

hassanoza22:
This is actually great if reviewed to be accurate

TempsRay (approved):
Pretty helpful. Thanks. Could use some more detail in specific abilities/moves though.

bennybroseph:
After discussion with @f-raZ0R, better to assist with updating [this](https://discord.com/channels/1125469663833370665/1238277103145713754/1239055850337996872) document instead. Closing this out.

PR #698, opened 2024-05-10
Added punctuation to several item descriptions.

### LINKED ISSUE #247: Minor text corrections for item descriptions.
Most of the items that are unique to PokeRogue or have had their descriptions altered in some way are missing periods at the end of said descriptions.

Additionally, in Healing Charm's description, "Revives" should be capitalized since it's referencing an actual item name.

I understand this is extremely nit-picky and not exactly a priority, but I always felt spotless text presentation to be a good marker of polish in a game and IMO it's sort of "in the spirit of Pokemon" to be relatively clean on this front. I've updated the descriptions on the wiki for how they would look if it makes my point more clear.

On the off chance you choose to entertain me, thank you. Your game is wonderful.

grencham:
Adds punctuation to several item descriptions. Closes #247


Flashfyre (commented):
This is the second time this changeset has been made, but this is intentional punctuation, differentiating sentences and sentence fragments.

PR #699, opened 2024-05-10
Implemented Power Split and Guard Split

XGSleepWalker:
* Implemented Power Split
* Implemented Guard Split


src/field/pokemon.ts
```diff
@@ -543,7 +543,12 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
   }
 
   getStat(stat: Stat): integer {
-    return this.stats[stat];
+    if (!this.summonData)
+    {
+      return this.stats[stat];
+    }
+
+    return this.summonData.stats[stat];
```
bennybroseph, line 551: We use Egyptian style braces in this project, let's get these updated

src/field/pokemon.ts
```diff
@@ -1552,6 +1557,11 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     return healAmount;
   }
 
+  changeSummonStat(stat: Stat, value: integer) : void
+  {
+    this.summonData.stats[stat] = value;
+  }
```
bennybroseph, line 1563: We use Egyptian style braces in this project, let's get these updated

bennybroseph (changes requested):

src/data/move.ts
```diff
@@ -1755,6 +1755,59 @@ export class HpSplitAttr extends MoveEffectAttr {
...
+      target.changeStat(Stat.SPDEF, specialDefenseValue);
+      infoUpdates.push(target.updateInfo());
+
+
+      return Promise.all(infoUpdates).then(() => resolve(true));
+    });
+  }
+
+}
+
```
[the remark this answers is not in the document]
XGSleepWalker: Documentation added

src/field/pokemon.ts
XGSleepWalker: Braces fixed.

src/field/pokemon.ts
XGSleepWalker: Braces fixed.

bennybroseph (approved):

src/data/terrain.ts
bennybroseph, line 1: Is this change needed?

bennybroseph (changes requested):

src/data/move.ts
```diff
@@ -1842,6 +1842,83 @@ export class HpSplitAttr extends MoveEffectAttr {
...
+   * @returns {boolean} True if power split is applied successfully.
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): Promise<boolean> {
+    return new Promise(resolve => {
+      if (!super.apply(user, target, move, args))
+        return resolve(false);
+
+      const infoUpdates = [];
+
+      const defenseValue = Math.floor((target.getStat(Stat.DEF) + user.getStat(Stat.DEF)) / 2);
```
bennybroseph, line 1903: Is there a source that states these values are rounded down? If not we should be using Math.round here.

bennybroseph (changes requested):

src/data/move.ts
XGSleepWalker: From Smogon: https://www.smogon.com/dex/sv/moves/guard-split/
https://www.smogon.com/dex/sv/moves/power-split/

'The user and the target have their Attack/Defense and Special Attack/Defense stats set to be equal to the average of the user and the target's Attack and Special Attack stats, respectively, rounded down. Stat stage changes are unaffected.'

src/data/terrain.ts
XGSleepWalker: Leftover from a bad push, removed.

src/field/pokemon.ts
bennybroseph: Didn't realize these were brand new functions. need these documented

src/field/pokemon.ts
bennybroseph: Didn't realize these were functions. need these documented

bennybroseph (changes requested):

src/field/pokemon.ts
XGSleepWalker: Documentation added.

src/field/pokemon.ts
XGSleepWalker: Documentation added.

bennybroseph (approved):

Issue #681, opened 2024-05-09
Lycanrock dusk form evolution not possible


j-diefenbach:
Rockruff doesn't evolve into lycanrock during dusk w appropriate level up, only evolves during day/night into midday/midnight forms.

Madmadness65:
You need to catch an Own Tempo Rockruff (or a Dusk Lycanroc, both will make Own Tempo Rockruff available) to be able to evolve it into the Dusk form.

j-diefenbach:
> You need to catch an Own Tempo Rockruff (or a Dusk Lycanroc, both will make Own Tempo Rockruff available) to be able to evolve it into the Dusk form.

It had own tempo, im going off of the discord post. Looks like there's an issue in pokemon-evolutions.ts line 1396. Currently it is 
`    new SpeciesFormEvolution(Species.LYCANROC, '', 'dusk', 25, null, new SpeciesEvolutionCondition(p =>  p.scene.getSpeciesFormIndex(p.species) === 1), null),`
There's no TimeOfDay.DUSK.
Working on replicating it but I'll come back to it tmr

Madmadness65:
That evolution method is literally just "Evolve a form index 1 Rockruff (Own Tempo) at level 25", setting the time of day evolution requirement too was deemed too restrictive, so the form index is the only requirement for it.

j-diefenbach:
not sure if i've read the code wrong but there seems to be no actual functionality for checking that form. When getSpeciesFormIndex(p.species) is called it checks for special IDs and ignoreArena, of which rockruff will ignore. Then it calls this.arena.getSpeciesFormIndex which again is pretty much just hardcoded. It looks like its mostly just returning cases e.g. lycanrock it returns the time of day. There's no rockruff switch case so I'll see if that can be added in, could just return 1 if it finds OwnTempo as an ability, with the current function itll just be returning 0 since it defaults

j-diefenbach:
![image](https://github.com/pagefaultgames/pokerogue/assets/105332964/4d059bbc-0a60-49f9-897d-c222e4419e0b)
Have done some more testing currently rockruff evolves into day form even with own tempo. Have changed the code to directly check for own tempo rather than the form and doing more testing atm

PR #700, opened 2024-05-10
Rockruff dusk form not evolving

j-diefenbach:
This PR addresses an issue raised where rockruff with own tempo is unable to evolve into dusk form. During dusk it does not evolve, during day or night it evolves instead into the midday/midnight forms.
Replaced inappropriate 'getSpeciesFormIndex' (which is not able to check rockruff's species since it doesn't have access to the pokemon's information, only scene info. It defaults to zero so rockruffs form index will always be zero) with 'hasAbility' for own tempo. In order to prevent an own tempo rockruff from evolving otherwise I've added a check for own tempo. Plus relevant import from enum abilities.
Screenshots of bug (#681)
image
note that getSpeciesFormIndex might be worth updating in future for more complex forms however this approach seems to be suitable and avoids messing with getSpeciesFormIndex unnecessarily


src/data/pokemon-evolutions.ts
```diff
@@ -1392,9 +1393,10 @@ export const pokemonEvolutions: PokemonEvolutions = {
     new SpeciesEvolution(Species.CRABOMINABLE, 1, EvolutionItem.ICE_STONE, null)
   ],
   [Species.ROCKRUFF]: [
-    new SpeciesFormEvolution(Species.LYCANROC, '', 'midday', 25, null, new SpeciesEvolutionCondition(p => p.scene.arena.getTimeOfDay() === TimeOfDay.DAY), null),
-    new SpeciesFormEvolution(Species.LYCANROC, '', 'dusk', 25, null, new SpeciesEvolutionCondition(p =>  p.scene.getSpeciesFormIndex(p.species) === 1), null),
-    new SpeciesFormEvolution(Species.LYCANROC, '', 'midnight', 25, null, new SpeciesEvolutionCondition(p => p.scene.arena.getTimeOfDay() === TimeOfDay.NIGHT), null)
+    new SpeciesFormEvolution(Species.LYCANROC, '', 'midday', 25, null, new SpeciesEvolutionCondition(p => (p.scene.arena.getTimeOfDay() === TimeOfDay.DAY) && (p.hasAbility(Abilities.OWN_TEMPO) === false)), null),
+    new SpeciesFormEvolution(Species.LYCANROC, '', 'dusk', 25, null, new SpeciesEvolutionCondition(p =>  p.hasAbility(Abilities.OWN_TEMPO)), null),
```
Xavion3, line 1397: You should be using the second and third arguments here as false and true respectively. This will fail if the ability is changed/suppressed/etc. right now.

Xavion3 (changes requested):

src/data/pokemon-evolutions.ts
Xavion3: Actually more complex, you'd want to somehow check the actual ability. Fusions complicated this as fusing a base rockruff with something with own tempo shouldn't let you evolve, and fusing an own tempo rockruff with something  else causing it to lose own tempo won't work either.

j-diefenbach:
Since the getSpeciesFormIndex doesn't work for rockruff, formIndex will accurately track whether it was originally an owntempo rockruff or not (otherwise we would be creating additional functions or modifying getFormIndex which was intended for wild pokemon gen). If it fuses to any other pokemon it still keeps its formIndex the same since it takes the first pokemon's form. So far rockruff still evolves correctly, an OT rockruff fused w any pokemon still evolves into dusk form, a regular rockruff fused with an OT rockruff (or any other pokemon that overwrites the ability as own tempo) will turn into day/night forms still.

Regular rockruff fused with own tempo rockruff evolves into day form correctly:
image
OT rockruff fused with another pokemon (replacing OT) still evolves into dusk form:
image

src/overrides.ts
```diff
@@ -23,7 +23,7 @@ import { Stat } from './data/pokemon-stat';
 export const SEED_OVERRIDE: string = '';
 export const WEATHER_OVERRIDE: WeatherType = WeatherType.NONE;
 export const DOUBLE_BATTLE_OVERRIDE: boolean = false;
-export const STARTING_WAVE_OVERRIDE: integer = 0;
+export const STARTING_WAVE_OVERRIDE: integer = 15;
```
TempsRay, line 26: override changes should not be committed

src/data/pokemon-evolutions.ts
```diff
@@ -1392,9 +1393,9 @@ export const pokemonEvolutions: PokemonEvolutions = {
     new SpeciesEvolution(Species.CRABOMINABLE, 1, EvolutionItem.ICE_STONE, null)
   ],
   [Species.ROCKRUFF]: [
-    new SpeciesFormEvolution(Species.LYCANROC, '', 'midday', 25, null, new SpeciesEvolutionCondition(p => p.scene.arena.getTimeOfDay() === TimeOfDay.DAY), null),
-    new SpeciesFormEvolution(Species.LYCANROC, '', 'dusk', 25, null, new SpeciesEvolutionCondition(p =>  p.scene.getSpeciesFormIndex(p.species) === 1), null),
-    new SpeciesFormEvolution(Species.LYCANROC, '', 'midnight', 25, null, new SpeciesEvolutionCondition(p => p.scene.arena.getTimeOfDay() === TimeOfDay.NIGHT), null)
+    new SpeciesFormEvolution(Species.LYCANROC, '', 'midday', 25, null, new SpeciesEvolutionCondition(p => (p.scene.arena.getTimeOfDay() === TimeOfDay.DAY) && (p.formIndex === 0)), null),
+    new SpeciesFormEvolution(Species.LYCANROC, '', 'dusk', 25, null, new SpeciesEvolutionCondition(p =>  p.formIndex === 1), null),
```
TempsRay, line 1397: Doesn't dusk form require a specific time of day as well?

TempsRay (changes requested):

Xavion3 (approved):

PR #702, opened 2024-05-10
Adding accuracy to in move display window in battle

td76099:
Exactly what it sounds like - displays move accuracy in battle window next to PP and Power.

Was considering adding info about target selection using `MoveTarget` enum so that info was displayed in a 2x2 instead of a row of 2 then a centered single item.

@piefrenzy on Discord

Some examples

Brave Bird

Synthesis

Tackle


Dakurei:
If I didn't do it when I added power to the display
It's because lining up this kind of information in so little space will prevent translation in some languages because of the length of certain words

Moreover, you've modified the layout, which breaks the localization function. I'm against this PR

image

PR #704, opened 2024-05-10
Update of german translation

Nekiritan:


bennybroseph:
Currently out of date. If conflicts are resolved and translations are updated from what we currently have please feel free to reopen this.

PR #706, opened 2024-05-10
Bouncy bubble description update + DeepL translations

paulcazac:
Updated Bouncy Bubble description (and added DeepL translations) to reflect the move changes.
~~50%~~ -> **100%**

@Flashfyre Closed the previous PR as it accidentally took commits from others into it.
DeepL used the en version as the base description. Might need it to be looked over by the translators.

### DIFF
--- a/src/locales/de/move.ts
+++ b/src/locales/de/move.ts
@@ -2931,7 +2931,7 @@ export const move: MoveTranslationEntries = {
   },
   "bouncyBubble": {
     name: "Blubbsauger",
-    effect: "Evoli greift mit Wasserblasen an. Evolis KP werden um die Hälfte des vom Wasser angerichteten Schadens geheilt."
+    effect: "Der Anwender greift an, indem er Wasserblasen auf sein Ziel schießt. Er absorbiert dann Wasser und stellt seine HP um 100% des vom Ziel erlittenen Schadens wieder her."
   },
   "buzzyBuzz": {
     name: "Knisterladung",
--- a/src/locales/de/move.ts
+++ b/src/locales/de/move.ts
@@ -2931,7 +2931,7 @@ export const move: MoveTranslationEntries = {
   },
   "bouncyBubble": {
     name: "Blubbsauger",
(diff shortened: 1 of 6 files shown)


src/locales/de/move.ts
```diff
@@ -2931,7 +2931,7 @@ export const move: MoveTranslationEntries = {
   },
   "bouncyBubble": {
     name: "Blubbsauger",
-    effect: "Evoli greift mit Wasserblasen an. Evolis KP werden um die Hälfte des vom Wasser angerichteten Schadens geheilt."
+    effect: "Der Anwender greift an, indem er Wasserblasen auf sein Ziel schießt. Er absorbiert dann Wasser und stellt seine HP um 100% des vom Ziel erlittenen Schadens wieder her."
```
CodeTappert, line 2934: Why did you change this? 
What was there before was the offical german text.  

I get that you want to remove the evoli part. 

So it should be 
"Der Anwender greift mit Wasserblasen an. Seine KP werden um die Hälfte des vom Wasser angerichteten Schadens geheilt."

CodeTappert, line 2934: While this is probably a correct german sentence (i am a native speaker). It doesnt feel correct. 

I would honestly go with the original and only replace the evoli part (and change it to 100%).

So

`Der Anwender greift mit Wasserblasen an. Seine KP werden in Höhe des vom Wasser angerichteten Schadens geheilt.`


Either way (if you want to use my provided sentence or not. You definitly have to change HP to KP because this is the correc german term for it.

paulcazac: Thanks for the input! 
I originally uploaded this with _only_ the en changes, but @Flashfyre suggested I change all the included languages so that it doesn't get left behind by accident. The suggested way was to use DeepL which is what I used to translate the English directly into the other languages. 

I did say in the description that some translators for the other languages should probably have a look at it and make any changes, as I don't know those languages natively. Thanks for doing that! If this gets approved, someone can easily change it as needed.

CodeTappert: I mean you can still add it. Just change it in your forked repo and push again to the branch. Would be better if you do it in the same pr. 

If you need help on how to do it feel free to ask

bennybroseph:
Not willing to trust AI over manual translations at the moment unless this directly grabs from in game sources. Feel free to update this request if more info can be given on how these are acquired.

PR #711, opened 2024-05-10
Adds 'accuracy' to the move information in the fight ui

Dakurei:
## Changes
- Adds `accuracy` to the move information in the fight ui
> In a safer way for translations, compared to what was proposed in #702
image
- Add new locales field `accuracy` in fight-ui-handler with default english value for translation
- Capitalize `accuracy` & `power` fields

### DIFF
--- a/src/locales/de/fight-ui-handler.ts
+++ b/src/locales/de/fight-ui-handler.ts
@@ -2,5 +2,6 @@ import { SimpleTranslationEntries } from "#app/plugins/i18n";
 
 export const fightUiHandler: SimpleTranslationEntries = {
     "pp": "PP",
-    "power": "POWER",
+    "power": "Power",
+    "accuracy": "Accuracy",
 } as const;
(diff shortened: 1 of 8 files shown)


Flashfyre:
I don't think this will cut it. The info window wasn't meant to go this high and it's covering the EXP bar. I think using a shorter label and making it closer to the other PR would be better.

Dakurei:
This was the purpose of having a slightly larger window for the duration of the display, which is what I specified in the message.

After doing a test in duo combat, it masks part of the HP bar, and for that, I agree that it can't do it.
image

But the exp bar argument didn't really have any value, it's not relevant information in combat unlike HP


As far as the other PR is concerned, it's done in such a way that the word "POWER" gets through and those for whom the translation used to get through don't.

I don't see this as an "enhancement" for my part 😞

Dakurei:
@Flashfyre And assuming that the interface is like this to avoid breaking the translation already in place, couldn't it do it given that it's visible in both PC and mobile mode ?

> image
image
image

Flashfyre (approved):

PR #718, opened 2024-05-10
Acrobatics does not treat vitamins as held items

Tiduzz:
Changed the move Acrobatics so it doesn't include non-held items on it's damage calculations

How it currently is:

Main_Acrobatics

How it is post-fix:

Fix_Acrobatics

Fix_Acrobatics_test1

Fix_Acrobatics_test2


Xavion3:
Yeah, vitamins count as held items in pokerogue. This was known and Sam suggested just using anything which counts as a held item, though this is potentially open for change.

TempsRay (approved):
Code looks good to me. Can you change the title of the commit so it's more descriptive though? "Acrobatics does not treat vitamins as held items"

src/locales/en/move.ts
```diff
@@ -2047,7 +2047,7 @@ export const move: MoveTranslationEntries = {
   },
   "acrobatics": {
     name: "Acrobatics",
-    effect: "The user nimbly strikes the target. If the user is not holding an item, this attack inflicts massive damage."
+    effect: "The user nimbly strikes the target. The fewer held items, higher the damage."
```
molker, line 2050: nitpick but wouldn't it be "The few the held items, the higher the damage." ? 
The idea is good but the wording feels off
Maybe something like: "The fewer items the user holds, the higher the damage it inflicts."

TempsRay: I like "The fewer items the user holds, the higher damage it inflicts" but should the in game formula be represented as well? It's sort of hard to word concisely though. Mentally when I think of acrobatics I just think it's 110bp - 11bp per item stack (max 5)

Tiduzz: I don't think any description really has specific values, but i'm willing to change the description to wathever is deemed alright

Issue #523, opened 2024-05-06
Facade does not ignore Burned status


LaukkaE:
Facade should ignore the attack halving effect of burn as of Generation 6.

LaukkaE:
Fixed in #530

PR #530, opened 2024-05-06
Facade should ignore burn damage reduction

mattrossdev:
Facade should ignore the burn damage reduction of physical moves: https://bulbapedia.bulbagarden.net/wiki/Facade_(move)

I didn't add a move flag for this because it only applies to this move, but I can rewrite it that way if that's preferred. 

Testing: 
- non facade contact moves do half damage when burned 
- facade does 2x damage if burned

PR #719, opened 2024-05-10
Issue #523 Facade should now ignore the attack reduction of beeing bu…

CodeTappert:
…rned

Fixes: https://github.com/pagefaultgames/pokerogue/issues/523


happinyz:
Seems like a PR was already opened for this: https://github.com/pagefaultgames/pokerogue/pull/530

CodeTappert:
@happinyz But those fix two different places in the code. 

Did i do the wrong place? Or should we combine both?

happinyz:
I think your implementation also works fine. I'd tend to use @mattrossdev's implementation just because we can exit the if block earlier and avoid having to do unnecessary calculations

CodeTappert:
Okay well then i will close this. It is a duplicate of #530

PR #720, opened 2024-05-10
[FIX] Give functionality to the RETURN TO TITLE button before a new run

DeadCode1010:
This pr is to bring functionality to the Return to Title menu button before starting a new run. (Such as the Starter Select Screen)

Steps to reproduce the bug: 
1. Launch the game. 
2. Select the New Game button. 
3. Press Esc to enter the Menu. 
4. Navigate to the Return to Title button. 
5. Select the button. 
6. Observe the screen. 

Result: 
The user remains at the Starter Select screen. 

Expected Results: 
I expected to be taken back to the Title Menu. 

return_to_tilte_button_issue

### DIFF
--- a/src/ui/menu-ui-handler.ts
+++ b/src/ui/menu-ui-handler.ts
@@ -298,6 +298,11 @@ export default class MenuUiHandler extends MessageUiHandler {
           success = true;
           break;
         case MenuOptions.RETURN_TO_TITLE:
+          if (this.scene.getParty().length === 0) {
+            success = true;
+            this.scene.reset(true);
+          }
+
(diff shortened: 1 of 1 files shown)


Flashfyre (commented):
We shouldn't allow returning to title on the actual title. From the starter screen you can just go back to it with x. I don't see why we need it to work everywhere.

DeadCode1010:
> We shouldn't allow returning to title on the actual title. From the starter screen you can just go back to it with x. I don't see why we need it to work everywhere.

For sure 👍  
Just wanted to see if that was working as expected. 

Thanks for checking it out! 

Closing PR.

Issue #708, opened 2024-05-10
Sturdy does not work properly in double battles


umbresp:
If two pokemon attack a pokemon with Sturdy it will endure both hits

DustinLin:
Can't seem to reproduce, testing with a set up of geodude with sturdy and 2 opposing swampert with surf, endures the first surf but dies to the second

DustinLin:
nvm, PR #604 resolves this

DustinLin:
can someone close this issue? PR #604 is merged

PR #604, opened 2024-05-07
Fix sturdy

Tiduzz:
Stops sturdy ability from activating everytime the user that has it gets hit, even though he doesn't stay at 1hp.

Not the fix in mind, but attacks like double hit popped sturdy on the first hit, and fainted the target on the second. I saw people commenting on discord it was a bugged, and sturdy was working like endure. I couldn't reproduce it with those changes, but might not be related.

PR #722, opened 2024-05-10
Fixing sturdy ability conidtion

DustinLin:
Sturdy should only apply when the current hp is full

resolves issue #708


DustinLin:
PR #604 fixes this

PR #723, opened 2024-05-10
Allow selection of item quantity when transferring items

dielle000:
This PR expands the item transfer functionality allowing to select how many items to transfer instead of being forced to transfer the entire stack. 

The transfer menu is also changed so that after a single transfer operation the user is not kicked out of the menu. This is particularly helpful when moving around a lot of items.

## Changes

* Added UI for selection of item quantity if `stackCount > 1`. Introduced `PartyUiMode.MODIFIER_TRANSFER_QUANTITY`, `PartyOption.TRANSFER_QUANTITY` and private variables `transferQuantity` and `transferQuantityMax` in `party-ui-handler.ts`;

* Changed method `tryTransferHeldItemModifier` parameter `transferStack` (boolean) into `transferQuantity` (integer, default value 1). Every call to this method has been changed to keep the behavior consistent. This change has been tested with the move Covet, abilities Magician and Pickpocket, the item Baton and the fusion mechanic (items of both Pokemon are kept);

* Changed type `PartyModifierTransferSelectCallback` to include the `itemQuantity`;

* Changed method `startTransfer` in `party-ui-handler.ts` so that it doesn't automatically update the variable `transferOptionCursor`. Without this change, the transfer would always involve the first item in the list because of the addition of the new UI element (which has only one option).

## Screenshot of the new UI
transfer_quantity

I hope that this is helpful and that I didn't break anything. Have a nice day!


Flashfyre:
I appreciate the contribution, but this UI is a bit inconsistent. There's a lot of empty space around the quantity select and it isn't centered either. I'd like to see this fixed before it gets in. Also, having this available from the item select itself would be even better, if that can be made to look good.

dielle000:
Thank you for the feedback. Selecting the item and the quantity at the same time would definitely work better, so I'll make the change.

dielle000:
Now the UI is the same as before but it is possible to change the quantity by pressing left or right if the cursor is on an item with stack bigger than 1

Greenlamp2:
Could you post some screenshots/videos to show the latest changes please ?

dielle000:
Sure, here you go!

https://github.com/pagefaultgames/pokerogue/assets/168655502/65c9a9ce-cfb8-489f-976c-4e42d011d66d

molker:
I'd like it if there was any indication for players that they can increment/decrement. I see you had "<" and ">" on either side before, it might be good to have the cursor on either side mimicking that to imply you can change it

bennybroseph:
> I see you had "<" and ">" on either side before, it might be good to have the cursor on either side mimicking that to imply you can change it

This sounds a little clunky. Can we just add height to the existing message window on the left?

image

bennybroseph:
looks like an improper merge or rebase messed this pull request up pretty badly. either revert the commit or make a new fresh one based on main.

dielle000:
Yeah, I'll make a new PR

PR #724, opened 2024-05-10
Adds Light Clay item / modifier that extends screens from 5 -> 8 turns

shayebeadling:
So the item function properly, but I didn't add it to the loot pool since I wanted to hear some thoughts on it first. 

It hooks into the addArenaTagAttr on moves, as far as I know screens can only be summoned via moves. If that assumption is wrong let me know.

Also, I implemented it as a persistent 'party wide' item like the healing charm, as it already seems pretty situational anyway.


shayebeadling:
In order to cleanly check if a mon in the party had a screen move, I did a few things

1. Adds a screenMove flag for the 5 screen moves. If this is too small of a set to warrant a flag, let me know. However, this allowed for some very clean implementation.
2. Adds a hasMoveWithFlag() method on pokemon to check their movesets for a move with a certain flag, this makes it very easy to check for certain moves in a moveset while generating the weights for items.
3. Adds Light Clay as an Ultra Tier item that has a weight of 10 if a mon in the party has a screen move, 0 if not. 10 weight is the same as the MoveBoosting items like Charcoal, Soft Sand, etc.

src/modifier/modifier-type.ts
```diff
@@ -918,6 +918,9 @@ export const modifierTypes = {
   HEALING_CHARM: () => new ModifierType('Healing Charm', 'Increases the effectiveness of HP restoring moves and items by 10% (excludes Revives)',
     (type, _args) => new Modifiers.HealingBoosterModifier(type, 1.1), 'healing_charm'),
   CANDY_JAR: () => new ModifierType('Candy Jar', 'Increases the number of levels added by Rare Candy items by 1', (type, _args) => new Modifiers.LevelIncrementBoosterModifier(type)),
+  
+  LIGHT_CLAY: () => new ModifierType('Light Clay', 'Aurora Veil, Light Screen, or Reflect lasts 8 turns instead of 5.',
```
DustinLin, line 922: Feel like text makes more sense if worded:

"Aurora Veil, Light Screen, **and** Reflect 8 turns instead of 5."

src/modifier/modifier.ts
```diff
@@ -918,6 +918,35 @@ export class PreserveBerryModifier extends PersistentModifier {
...
+    return super.shouldApply(args) && args[0] instanceof Pokemon && args[1] instanceof Utils.IntegerHolder;
+  }
+
+  apply(args: any[]): boolean {
+    if ((args[1] as Utils.IntegerHolder).value)
+      (args[1] as Utils.IntegerHolder).value = (args[1] as Utils.IntegerHolder).value + 3;
+
+    return true;
+  }
+
```
DustinLin, line 944: would be nice to add some documentation/comments in this class, perhaps JSDoc style for the functions

WingofaGriffin:
Nice, was looking into this myself. I feel like this can easily be refactored into a Terrain Extender and Weather Rock case to extend the duration of those as well, let me know if that makes sense.

amatsune:
Alternatively I could see this item being stackable, increasing duration by 1 or 2 per stack, limited to 5 or 3, respectively.

Regarding weight, if you have a Pokémon in your party that can set up screens I think it should have a slightly higher weight than regular type boosting items, seeing how situational it is already. Moreso if it's a stacking item, since effects would be more tame

flx-sta:
Considered stale.

PR #726, opened 2024-05-10
Initial Commit for Japanese Locales

Adri1:
Initial commit for a Japanese translation.
It's very far from being complete, I did what I could with my limited knowledge in Japanese, so feel free to improve it as much as you want and can.

- Aside from this initial commit for a Japanese localization, I updated the **pokemon-emerald-pro.ttf** font where the Japanese kanas were missing, using the style used in Japanese Pokémon Emerald to keep consistency. Note that **pkmnems.tff** font is not adapted yet, but is a work in progress.
- Kanjis are not included in this edited font, as Pokémon games before Pokémon BW were only written in kanas, so that's why the few translations I made are in kanas only. If kanjis are written, it would display them using the (far too big) font by defaut, as it's currently the case in the Simplified Chinese locales.

image
image
image
image


src/locales/fr/battle.ts
```diff
@@ -31,7 +31,7 @@ export const battle: SimpleTranslationEntries = {
   "learnMoveNotLearned": "{{pokemonName}} n’a pas appris\n{{moveName}}.",
   "learnMoveForgetQuestion": "Quelle capacité doit être oubliée ?",
   "learnMoveForgetSuccess": "{{pokemonName}} oublie comment\nutiliser {{moveName}}.",
-  "levelCapUp": "La limite de niveau\na été augmentée à {{levelCap}} !",
+  "levelCapUp": "La limite de niveau\nest augmentée à {{levelCap}} !",
```
CodeTappert, line 34: Why did you update the french locale if this pull request is about the japanese local?

(I dont know if what you changed makes sense but i would rather have it in seperate pull requests tbh)

Adri1: Oops that's an error, I meant to put it in a diffrent pull request

bennybroseph:
I really want to add this, but there's so much being added. There's some french image files, there's new fonts, there's files that don't seem to change anything but show a full diff, etc. If there's any way this can be cleaned up, please feel free to reopen this but for now it's too much for me to go through.

PR #727, opened 2024-05-11
Impl: Poison Heal

kd8lvt:
I probably did this in a really weird way, but I got Poison Heal working.

I used [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Poison_Heal_(Ability)) as reference, and I'm glad I did, because I was today years old when I learned that Toxic's turn count still increases regardless of Poison Heal. Certainly a niche use of Gastro Acid, but sure GameFreak, why not.

In the future when Heal Block is implemented, Poison Heal's healing (not the poison immunity, just the healing) will have to be negated by it, but until then, this works.


src/data/ability.ts
```diff
@@ -2398,6 +2398,29 @@ export class ReduceStatusEffectDurationAbAttr extends AbAttr {
   }
 }
 
+export class HealFromStatusAbAttr extends AbAttr {
+  private healRatio: integer
+  private statusEffect: StatusEffect;
+  private statusEffect2: StatusEffect;
```
happinyz, line 2404: `private statusEffects: StatusEffect[]` would be better here imo, just in case in the future we implement heal conditions for possible 1 or 3+ statuses.

src/data/ability.ts
```diff
@@ -2398,6 +2398,29 @@ export class ReduceStatusEffectDurationAbAttr extends AbAttr {
...
+  constructor(healRatio: integer,statusEffect: StatusEffect,statusEffect2: StatusEffect) {
+    super();
+    this.healRatio = healRatio;
+    this.statusEffect = statusEffect;
+    this.statusEffect2 = statusEffect2;
+  }
+
+  apply(pokemon: Pokemon, passive: boolean, cancelled: Utils.BooleanHolder, args: any[]): boolean | Promise<boolean> {
+    if (pokemon.status.effect === this.statusEffect || pokemon.status.effect === this.statusEffect2) {
```
happinyz, line 2413: `if (this.statusEffects.includes(pokemon.status.effect)` if we change this to take in a list of status effects instead.

kd8lvt:
Ah, yeah that's a much better implementation. No idea why I didn't think of that.

src/data/ability.ts
kd8lvt: 1 was covered by `StatusEffect, null` but yeah, like I said that's a much better implementation

src/data/ability.ts
```diff
@@ -2398,6 +2398,27 @@ export class ReduceStatusEffectDurationAbAttr extends AbAttr {
   }
 }
 
+export class HealFromStatusAbAttr extends AbAttr {
+  private healRatio: integer
+  private statusEffects: StatusEffect[];
+  constructor(healRatio: integer,statusEffects: StatusEffect[]) {
```
happinyz, line 2404:
```suggestion
  constructor(healRatio: integer, statusEffects: StatusEffect[]) {
```

src/data/ability.ts
```diff
@@ -2398,6 +2398,27 @@ export class ReduceStatusEffectDurationAbAttr extends AbAttr {
...
+    this.healRatio = healRatio;
+    this.statusEffects = statusEffects;
+  }
+
+  apply(pokemon: Pokemon, passive: boolean, cancelled: Utils.BooleanHolder, args: any[]): boolean | Promise<boolean> {
+    if (this.statusEffects.includes(pokemon.status.effect)) {
+      let healAmount: integer = Math.floor(pokemon.getMaxHp()*(this.healRatio/8));
+      pokemon.heal(healAmount);
+      pokemon.status.incrementTurn();
+      cancelled.value = true;
```
happinyz, line 2415: What does setting this to true do?

kd8lvt: It prevents the status effect from triggering

happinyz (approved):
lgtm, we should get feedback from a contributor with write access first though

bennybroseph (changes requested):

Also, does this show any sort of animation for the heal?

bennybroseph:

PR #728, opened 2024-05-11
Implemented Tera Shell

Zedarflight:
Implemented Tera Shell - https://bulbapedia.bulbagarden.net/wiki/Tera_Shell_(Ability)


happinyz:
Looks like @jaimefd already made a PR implementing this move: https://github.com/pagefaultgames/pokerogue/pull/695

I think this PR has a much cleaner way of implementing this, however.

jaimefd:
Does this implementation reduce all hits from a multi-hit move? Bulbapedia doesn't mention that for some reason, but it's a thing, Smogon does: https://www.smogon.com/dex/sv/abilities/tera-shell/

flx-sta:
@Madmadness65 i think we should close this PR in favor of #695 which has an active development

Madmadness65:
As this is a duplicate of #695 , and not being actively maintained, this PR will be closed for now.

Zedarflight:
Completely fair, work got real busy for me so I'm glad to let someone with time work on this

PR #731, opened 2024-05-11
Bugfix - Ghost types are no longer trapped by abilities like Shadow Tag or moves like Fire Spin

td76099:
@Piefrenzy in Discord

Originally reported in Discord Bug Reports channel, Ghost type Pokemon were being trapped by Shadow Tag as well as moves like Fire Spin/Whirlpool/etc. despite documentation saying [trapping abilities](https://bulbapedia.bulbagarden.net/wiki/Ghost_(type)#Interacting_with_the_Ghost_type) shouldn't work and [trapping moves](https://bulbapedia.bulbagarden.net/wiki/Fire_Spin_(move)#Generation_VI_onwards) shouldn't work.

The function didn't even take in the Pokemon trying to switch/run in the first place, so it now gets passed in and checked to see if it's a Ghost type.

Here is a Ghost type swapping through Shadow Tag and a non-Ghost type getting stuck:

https://github.com/pagefaultgames/pokerogue/assets/85713900/d017534b-fd46-45f5-ae0c-d3d5ca22e414

Ghost type successfully swapping through Fire Spin and a non-Ghost type getting stuck:

https://github.com/pagefaultgames/pokerogue/assets/85713900/5897b8d0-a9af-4a85-9270-4e6d3de6794a

Working on right side:

https://github.com/pagefaultgames/pokerogue/assets/85713900/2f39ad3a-4c85-4e6b-9d2d-32f2af60c0de

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2259,13 +2259,17 @@ export class CheckTrappedAbAttr extends AbAttr {
     super(false);
   }
   
-  applyCheckTrapped(pokemon: Pokemon, passive: boolean, trapped: Utils.BooleanHolder, args: any[]): boolean | Promise<boolean> {
+  applyCheckTrapped(pokemon: Pokemon, passive: boolean, trapped: Utils.BooleanHolder, otherPokemon: Pokemon[], args: any[]): boolean | Promise<boolean> {
     return false;
   }
 }
 
 export class ArenaTrapAbAttr extends CheckTrappedAbAttr {
-  applyCheckTrapped(pokemon: Pokemon, passive: boolean, trapped: Utils.BooleanHolder, args: any[]): boolean {
+  applyCheckTrapped(pokemon: Pokemon, passive: boolean, trapped: Utils.BooleanHolder, otherPokemon: Pokemon[], args: any[]): boolean {
+    if (otherPokemon[0].getTypes().includes(Type.GHOST)){
+      trapped.value = false;
+      return false;
+    }
     trapped.value = true;
     return true;
   }
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2263,20 +2263,11 @@ export class CheckTrappedAbAttr extends AbAttr {
   constructor() {
     super(false);
   }
-  /**
-   * Checks if enemy Pokemon is trapped by an Arena Trap-esque ability
-   * @param pokemon N/A
-   * @param passive N/A
-   * @param trapped N/A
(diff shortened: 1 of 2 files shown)


src/data/ability.ts
```diff
@@ -2259,13 +2259,17 @@ export class CheckTrappedAbAttr extends AbAttr {
...
-  applyCheckTrapped(pokemon: Pokemon, passive: boolean, trapped: Utils.BooleanHolder, args: any[]): boolean | Promise<boolean> {
+  applyCheckTrapped(pokemon: Pokemon, passive: boolean, trapped: Utils.BooleanHolder, otherPokemon: Pokemon[], args: any[]): boolean | Promise<boolean> {
     return false;
   }
 }
 
 export class ArenaTrapAbAttr extends CheckTrappedAbAttr {
-  applyCheckTrapped(pokemon: Pokemon, passive: boolean, trapped: Utils.BooleanHolder, args: any[]): boolean {
+  applyCheckTrapped(pokemon: Pokemon, passive: boolean, trapped: Utils.BooleanHolder, otherPokemon: Pokemon[], args: any[]): boolean {
+    if (otherPokemon[0].getTypes().indexOf(Type.GHOST) != -1){
```
kd8lvt, line 2269: Could be simplified as
```suggestion
    if (otherPokemon[0].getTypes().includes(Type.GHOST)){
```

kd8lvt: Also, how does this work with more than one Pokemon on the field? From a cursory glance it looks like you're only checking the first party member?

kd8lvt (changes requested):

src/data/ability.ts
td76099: > Could be simplified as

Agreed - made change

td76099: > Also, how does this work with more than one Pokemon on the field? From a cursory glance it looks like you're only checking the left party member?

`otherPokemon`, while an array, ever only contains a single Pokemon. It's the Pokemon that is being targeted by a trapping move or the Pokemon that is trying to run/switch. That's not something I wanted to do; that is how it is in `phases.ts` and it's just passed on to `ability.ts`.

Here's it working on right side Pokemon too
https://github.com/pagefaultgames/pokerogue/assets/85713900/540e985b-6981-4da0-8edc-9b8f813bbb69

td76099: Figured out why it was passing in as an array and it is now updated to not do that 👍

bennybroseph (approved):

PR #365, opened 2024-04-30
implement triple axel and triple kick.

Reldnahc:
Changed MultiHitType._3_INCR to handle the move missing. Should this be renamed to something like MultiHitType.1_to_3_INCR?

Apply a damage multiplier depending on hit count to the two moves.

PR #734, opened 2024-05-11
Update MultiHitAttr power calculation for power-up attack

XGSleepWalker:
Updated MultiHitAttr power calculation for power-up attack

Based on the current formula, both Triple Kick and Triple Axel can be calculated by multiplying the number of hits with the base power.


src/field/pokemon.ts
```diff
@@ -1409,6 +1409,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       case MoveCategory.SPECIAL:
         const isPhysical = moveCategory === MoveCategory.PHYSICAL;
         const power = new Utils.NumberHolder(move.power);
+        if(move.getAttrs(MultiHitAttr).length && move.getAttrs(MultiHitAttr).find(attr => (attr as MultiHitAttr).multiHitType === MultiHitType._3_INCR)){
+          power.value *= this.battleData.hitCount + 1;
+        }
```
happinyz, line 1414: Most move-specific logic belongs in the move's `apply` function.

https://github.com/pagefaultgames/pokerogue/blob/main/src/data/move.ts#L1041-L1043

happinyz (changes requested):

happinyz:
Sorry, should've linked this earlier, forgot I had seen this the other day: https://github.com/pagefaultgames/pokerogue/pull/365

XGSleepWalker:
I'm abandoning this PR for the sake of #365

PR #742, opened 2024-05-11
Use getBattleStat instead of getStat in BattleStatRatioPowerAttr

JakubHanko:
Usage of getStat causes moves like Electro Ball to use incorrect Stat.SPD (SPD before it gets halved).


Flashfyre (approved):

Flashfyre (changes requested):
Should this also pass opponent in for abilities like Unaware?

JakubHanko:
> Should this also pass opponent in for abilities like Unaware?

I have two questions since my Pokémon skills are rusty:

1. correct me if I am wrong but Unaware should still respect the speed changes in case of Gyro Ball and Electro Ball
2. right now there aren't any moves in the game that fall into the BattleStatRatio category - are you aware of any moves that are somehow take the ratio of your and opponent's stats into consideration?

TempsRay (approved):
Unaware doesn't ignore speed so this should work fine.

Should probably rename the `BattleStatRatioPowerAttr` attr though, as it's only ever being used for 1 stat (speed) and 2 moves that have different calculations

JakubHanko:
Decided to go with two classes, each dedicated to its respective move since both calculations are fairly specific to their use cases.

TempsRay (commented):
Looks good and I guess it makes sense to separate out these as two classes. Please add some documentation to the new classes though

JakubHanko:
TSDocs have been added.

Flashfyre (approved):

PR #744, opened 2024-05-11
Ability Corrosion

Brandon-Rodgers:
Implemented Corrosion Partially.
Tested against:
- Poison Powder
- Toxic
- Dire Claw
- Sludge Bomb
- Psycho Shift

They all work as expected

Missing ability Magic Bounce to test against.


bennybroseph (changes requested):

bennybroseph (changes requested):

bennybroseph:
So, it's doing what it is supposed to, but there's no indication of what's happening here. Corrosion can't go through an Immunity Pokemon but neither ability pops up to say that's why it's failing nor does it say "But it failed!" or similar. I will check showdown to see what the message is there, but you should so some research as well

https://github.com/pagefaultgames/pokerogue/assets/13838608/85c2dae4-214c-48a2-bcf1-21141ff21701

bennybroseph (approved):

Brandon-Rodgers:
https://github.com/pagefaultgames/pokerogue/assets/9848614/5693a796-8b78-440f-96b0-d3a00f21c41f

Toxic - Affects Steelix and triggers Corrosion pop-up
Poison Sting - Does not affect Steelix due to it being a damaging move

PR #748, opened 2024-05-11
Adds safety goggles, grants team immunity to weather damage (Sandstorm and Hail)

shayebeadling:
With how popular the trap moves are (mostly because of how they work with boss hp bars) I thought making the damaging weather moves better would be cool. 

1. Adds a WeatherDamageImmunityModifier that prevents the team from taking damage from hazardous weather.
2. Hooks into the existing WeatherEffectPhase to check for the modifier and adjust accordingly.

TODO: 
Add them to the drop pool, I was thinking they would be a very low weight in the rogue tier pool, and the weight is boosted if the user has a damaging weather move.


shayebeadling:
Alright I've added them to the rogue pool with a normal weight of 1 (very low)
But the weight is 10 if you have a relevant move or ability (Sandstorm, Hail, Sand Spit, Sand Stream)

EmoUsedHM01:
Can you fix the conflicts and make these a held item instead of team-wide?

bennybroseph:

PR #751, opened 2024-05-11
Completed Throat Chop 

JobSoltero:
I modified the disabled moves functionality to be able to disable a certain type of moves. I avoided using an object so its simpler to check if pokemon has disabled move through disabled moves length


bennybroseph:

PR #753, opened 2024-05-11
Finish Sweet Veil and Pastel Veil

Brandon-Rodgers:
Implemented the ally check for Sweet Veil and Pastel Veil On status check the Pokémon checks if it's ally has an ability to prevent the status


bennybroseph (changes requested):
Needs TSDoc documentation on all new classes and functions

Brandon-Rodgers:
Shortened code to better match PR #754.

PR #758, opened 2024-05-11
Add ignoresProtect to Life Dew

XGSleepWalker:
When using Life Dew, pokemons that use Protect should still be healed from Life Dew coming from an ally.

As per https://bulbapedia.bulbagarden.net/wiki/Life_Dew_(move)


Tiduzz (commented):
LGTM!

bennybroseph (approved):

PR #759, opened 2024-05-12
Fixes some variable damage moves not working with tinted lens

mattrossdev:
Updated Description: Some variable power moves like magnitude and dragon energy were not getting boosted by tinted lens. This PR adds a new class for abilities that boost the damage of moves rather than the base power (currently this appears to just be tinted lens). 

Testing: 
- Magnitude does double damage on a not very effective target when attacker has tinted lens 
- Magnitude does normal damage on a not very effective target when attacker does not have tinted lens 
- Neutral damage move does same damage with/without tinted lens

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -988,6 +988,12 @@ export class MoveTypeChangeAttr extends PreAttackAbAttr {
   }
 }
 
+/**
+ * Class for abilities that boost the damage of moves
+ * For abilities that boost the base power of moves, see VariableMovePowerAbAttr
+ * @param damageMultiplier the amount to multiply the damage by 
+ * @param condition the condition for this ability to be applied 
+ */
 export class DamageBoostAbAttr extends PreAttackAbAttr {
   private damageMultiplier: number;
   private condition: PokemonAttackCondition;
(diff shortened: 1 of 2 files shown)


mattrossdev:
Closing this for now, I think a better approach is that there should be a different class for tinted lens since it modifies the damage dealt and not the base power. Also need to check if there are other abilities this is true for. 
https://bulbapedia.bulbagarden.net/wiki/Power
https://bulbapedia.bulbagarden.net/wiki/Damage

src/data/ability.ts
```diff
@@ -3066,7 +3102,7 @@ export function initAbilities() {
       .attr(IgnoreOpponentStatChangesAbAttr)
       .ignorable(),
     new Ability(Abilities.TINTED_LENS, 4)
-      .attr(MovePowerBoostAbAttr, (user, target, move) => target.getAttackTypeEffectiveness(move.type, user) <= 0.5, 2),
+      .attr(DamageBoostAbAttr, 2, (user, target, move) => target.getAttackTypeEffectiveness(move.type, user) <= 0.5),
```
mattrossdev, line 3105: I changed the order of the multiplier/condition to match StabBoostAbAttr

mattrossdev:
Updated now. I checked and tinted lens and flash fire appear to be the only abilities that modify the damage of an attack rather than the base power, and flash fire has a battlertag so doesn't need to be  added here.

src/field/pokemon.ts
```diff
@@ -1409,6 +1409,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
               if (!burnDamageReductionCancelled.value)
                 damage.value = Math.floor(damage.value / 2);
             }
+
+            applyPreAttackAbAttrs(DamageBoostAbAttr, source, this, battlerMove, damage)
```
bennybroseph, line 1413: missing a semi-colon here

bennybroseph (changes requested):

bennybroseph:
Functionality looks good for this. 

https://github.com/pagefaultgames/pokerogue/assets/13838608/5e634691-7bc1-4664-a849-8cbf9e37264b

Magnitude

https://github.com/pagefaultgames/pokerogue/assets/13838608/615d2ac0-2013-4430-ae27-44a46489ff67

bennybroseph (approved):

PR #761, opened 2024-05-12
Implement Torment

f-raZ0R:
This is a draft PR that I am making as I progress; I do not feel confident enough in my implementation, and there may still be some issues. I am opening this now, so others would be able to look at it. Once I feel like torment is doing everything it should, I will work on cleaning up my code.

Things this PR does so far:
-Adds an entirely new field to pokemon summonData, for moves affected by torment/gigaton hammer/blood moon or whatever, for moves that can not be selected to be used, but will still work if forced to be used for whatever reason.
-Adds a battle tag for being tormented.
-Adds the Encore-Disable interaction of forcing a pokemon to struggle, and extends that to torment as well.
-Adds a field to Command.FIGHT, for whether or not the player was given the option to pick a move themselves. There's also a lot of Bad Code in this part. I don't like it and its only the way it is because i spent so long fiddling with to even get torment to behave. Hopefully this section will not Stay like this.
-Has absolutely 0 comments or documentation yet


f-raZ0R:
Okay, I feel like this PR is currently good enough for a Partial Implementation, but I would still like to have some other eyes on this and people testing things in case i missed something.

Current things that I know are not implemented yet but I do not feel like tackling myself at the moment, so they are under the umbrella of "eh its good enough for a Partial Implementation":

-Moves that call other moves (e.g. metronome) are not properly disabled by torment, and instead disable the move caused.


thats it thats all i've found that i havent fixed so far

f-raZ0R:
how do i mark this as a draft again i just realized i broke struggle

f-raZ0R:
Never mind i fixed it (for both cases)

bennybroseph:

PR #766, opened 2024-05-12
Added Portuguese (Brazil) translations

josericardo-fo:
Every translation was made with interpretation and adaptation, as there is not an official source of portuguese Pokémon material. Me and @Hareon, owner of [Elite dos Quatro Traduções](https://e4t.com.br/), used translations from Pokémon Unite, Anime, TCG, Pokémon GO.

- Moves and abilities not translated?
A small poll was made with the brazillian community and we chose to not translate the names of moves and abilities. The original games never had an official portugues translation, so many portuguese fans grew up playing those games in English and got used to it. If needed, we have the translations of every name of moves and abilities to follow the pattern of other languages.


src/system/settings.ts
```diff
@@ -210,6 +210,10 @@ export function setSetting(scene: BattleScene, setting: Setting, value: integer)
                 label: '简体中文',
                 handler: () => changeLocaleHandler('zh_CN')
               },
+              {
+                label: 'Português',
```
CodeTappert, line 214: Maybe change this to include the "Brazil" Part since there is also european portugese. (Otherwise there will be confusion later)

CodeTappert (changes requested):

src/system/settings.ts
josericardo-fo: Fixed

Madmadness65 (approved):

Issue #779, opened 2024-05-12
Curse bug


travbp:
Used Curse with Torterra in a double battle and it asked for a target selection (as if it were being used by a ghost type). All enemies were KOed before Torterra could move, and when he used Curse, no stat boost was given. Tried using Curse in battles afterwards and it had no effect and still asked for target selection

Issue #537, opened 2024-05-06
Non-ghost curse fails against invulnerable moves and good as gold


Tempo-anon:
Currently there is no way to make a move target conditional before the move is chosen. For the implementation of curse it targets but has a hardcoded branch logic that does the stat change if the user is not ghost type. Because it is technically classified as a targeting status move, it fails completely when the opponent is in the invulnerable phase of a move like dig/fly/dive or if they have good as gold because the move is cancelled immediately

PR #782, opened 2024-05-12
Fix Curse targeting for non ghost types

jaimefd:
Added a new MoveTarget type, MoveTarget.CURSE especifically for the move, so that it only makes you choose target if you are a ghost type (or tera into ghost). I don't believe many moves with variable targeting will be added so I think it's okay to fix it like this.

This should fix the bugs related to this move, such as the ones mentioned in issues https://github.com/pagefaultgames/pokerogue/issues/537 and https://github.com/pagefaultgames/pokerogue/issues/779

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -51,7 +51,8 @@ export enum MoveTarget {
   USER_SIDE,
   ENEMY_SIDE,
   BOTH_SIDES,
-  PARTY
+  PARTY,
+  CURSE
 }
 
 export enum MoveFlags {
@@ -4096,7 +4097,7 @@ export function getMoveTargets(user: Pokemon, move: Moves): MoveTargetSet {
   switch (moveTarget) {
     case MoveTarget.USER:
     case MoveTarget.PARTY:
-      set = [ user];
+      set = [ user ];
       break;
     case MoveTarget.NEAR_OTHER:
     case MoveTarget.OTHER:
@@ -4131,6 +4132,9 @@ export function getMoveTargets(user: Pokemon, move: Moves): MoveTargetSet {
     case MoveTarget.BOTH_SIDES:
       set = [ user, user.getAlly() ].concat(opponents);
       multiple = true;
+      break;
+    case MoveTarget.CURSE:
+      set = user.getTypes(true).includes(Type.GHOST) ? (opponents.concat([ user.getAlly() ])) : [ user ];
       break;
   }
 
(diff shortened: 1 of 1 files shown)


CodeTappert:
But isnt curse also beeing able to be used by non ghost types?

jaimefd:
> But isnt curse also beeing able to be used by non ghost types?
> 
> https://bulbapedia.bulbagarden.net/wiki/Curse_(move)
> 
> Maybe i am missunderstanding somehting

Yes, it still allows non ghost types to use it, but they don't have to choose a target anymore, since it becomes a self targeting move and gives +1 Attack, Defense and -1 Speed

src/data/move.ts
```diff
@@ -4132,6 +4133,9 @@ export function getMoveTargets(user: Pokemon, move: Moves): MoveTargetSet {
       set = [ user, user.getAlly() ].concat(opponents);
       multiple = true;
       break;
+    case MoveTarget.CURSE:
+        set = user.getTypes(true).includes(Type.GHOST) ? (opponents.concat([ user.getAlly() ])) : [ user];
```
TempsRay, line 4137: nit: this is a bit too tabbed forward and there is an extra space in `[ user]`

TempsRay (commented):
Nice! Didn't consider making a new MoveTarget category since this is such a minor edge case that I think only curse has conditional targeting. Have you tested that this works?

turtlesboulder:
Doesn't tera starstorm have variable targeting? If terapagos is not terrastalized then it is single target, and if it is terastellized then it is multi target and should skip asking for a target.

src/data/move.ts
jaimefd: Thank you, I fixed the tab, and the space in user was just to keep it consistent with the rest of the code since it's the same up there, but I just fixed both instead

jaimefd:
> Nice! Didn't consider making a new MoveTarget category since this is such a minor edge case that I think only curse has conditional targeting. Have you tested that this works?

Yes, it does work perfectly, though I haven't tested every edge situation, just did a double battle with a ghost type and a non ghost type and the first could choose target and the second couldn't. The animations also played as expected.

jaimefd:
> Doesn't tera starstorm have variable targeting? If terapagos is not terrastalized then it is single target, and if it is terastellized then it is multi target and should skip asking for a target.

You are right, I may do another PR for that, though I want to keep this one for Curse

bennybroseph (approved):

PR #787, opened 2024-05-12
[CONCEPT] Conditional boss ur

TempsRay:
Proof of concept for conditional biome spawns.
* Create a new rarity in a tier BOSS_CONDITIONAL_ULTRA_RARE (but this can be extended to other tiers as well)
* So right now the logic for generating a boss is 0 = UR, 1-5 = SR, 6-19 R, 20-63 C
* With  BOSS_CONDITIONAL_ULTRA_RARE generate a number from 0-128 where 0 = CUR, 1 = UR, 2-11 = SR, 12 - 39 = R, 40+ C
* If the condition fails then it defaults to BOSS_ULTRA_RARE
* Allows for more puzzle concepts (ie needing relicanth+wailord in the party to unlock something) and new encounters
* An extra biome tier allows weighting in tiers. For example, by setting the volcano condition to always be true and have the volcano conditional_ur tier have [reshiram, arceus-fire] for example, then the URs in volcano would be weighted 2:1 reshiram to arceus-fire

## Testing
* Set Zekrom as a ur boss and Reshiram as a conditional ur boss in town
* The condition in town is having a mewtwo in the party
* Override to generate a boss with random number 0
* [no mewtwo in party leads to zekrom](https://cdn.discordapp.com/attachments/1176874654015684739/1239273739028267098/image.png?ex=6642531b&is=6641019b&hm=aaa0b38d5d0ebaadb60d44295b8aa7a77c17b67d57705050d2b41e48f7f2a903&)
* [mewtwo in party leads to reshiram](https://cdn.discordapp.com/attachments/1176874654015684739/1239273739225665686/image.png?ex=6642531c&is=6641019c&hm=ef4800241f8bed29e26862f5be6351178dc8e72df915e6897721853da5ebe977&)

Ignore the innards out commits in here, that's some git confusion on my end.

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2281,7 +2281,6 @@ export class PostFaintHPDamageAbAttr extends PostFaintAbAttr {
   }
 
   applyPostFaint(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: PokemonMove, hitResult: HitResult, args: any[]): boolean {
-    console.log(args);
     let damage = args[0];
     attacker.damageAndUpdate((damage), HitResult.OTHER);
     attacker.turnData.damageTaken += damage;
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2275,6 +2275,9 @@ export class PostFaintContactDamageAbAttr extends PostFaintAbAttr {
   }
 }
 
+/** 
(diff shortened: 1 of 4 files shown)


TempsRay:
Closing this because of actual implementation here https://github.com/pagefaultgames/pokerogue/pull/1242

PR #793, opened 2024-05-12
add pokeball override

molker:
added override check into loading session and reset so it can be done on new game or continuing an existing save


TempsRay (approved):
Thanks for this lol

CodeTappert:
NIce! Now it is easier to test things

bennybroseph (approved):

PR #798, opened 2024-05-12
Hide leaderboard and player count in offline mode

Admiral-Billy:
As the title implies, this will hide the leaderboard and player count when the game is offline (as determined by isLocal in utils). I can't quite test that this works properly when online, but based on the condition being checked, I don't see why it would have any issues.

image

Feel free to go one step further and put something else there if you want heh


CodeTappert (approved):
Honestly simple but effective change. 

Code also looks good,

bennybroseph:
Does this affect the dev environment as well?

Admiral-Billy:
We talked in Discord about this but just to be thorough, yeah this should affect anything using the isLocal flag.

bennybroseph:
If we can fix isLocal affecting Dev we can revisit this, but for now we'll close this out

PR #799, opened 2024-05-12
Fixed clear freeze in offline mode

Nexllon:
Currently when in offline mode, api "savedata/newclear" gets called to check dailyRun seed every handleGameOver().
Added a simple isLocal check to prevent that since it was causing freezes when playing classic offline.

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -3611,9 +3611,13 @@ export class GameOverPhase extends BattlePhase {
       });
     };
     if (this.victory) {
-      Utils.apiFetch(`savedata/newclear?slot=${this.scene.sessionSlotId}`, true)
+      if (Utils.isLocal) {
+        doGameOver(true);
+      } else {
+        Utils.apiFetch(`savedata/newclear?slot=${this.scene.sessionSlotId}`, true)
         .then(response => response.json())
         .then(newClear => doGameOver(newClear));
+      }
     } else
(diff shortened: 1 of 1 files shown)


Admiral-Billy:
Oh cool, you did that for me! I'll note that this does fix the problem but I think this also lets you redo dailies to get the same reward multiple times.

Of course, if you're playing in offline mode you can already do far crazier shenanigans if you want to cheese the system, and the daily run isn't even a consideration unless some other code is changed to add it to offline, but it's worth noting regardless.

scw6538 (approved):
LGTM

bennybroseph (changes requested):
```ts
export const isLocal = window.location.hostname === 'localhost' || window.location.hostname === '';
```
How easy is this to spoof to be false

Nexllon:
> ```ts
> export const isLocal = window.location.hostname === 'localhost' || window.location.hostname === '';
> ```
> 
> How easy is this to spoof to be false

As easy as anything else in the game client (pretty)

Nexllon:
@Flashfyre since the issue with this PR is correlated to [#808](https://github.com/pagefaultgames/pokerogue/pull/808) that is not yet implemented (and will be fixed). Could we push this and fix in the other issue?
Or just close this one and I'll merge everything in the other one too.

PR #804, opened 2024-05-12
Feature: Adds the ability for users to spend candy to max a random IV

td76099:
@Piefrenzy on Discord

Right now, once you've reduced cost twice and unlocked the passive there is nothing to do with your candy on a particular Pokemon. For a 3 cost, that's 110 candy which can be done in under 10 runs of classic (personally got there in about 8).

The goal of this PR is to provide something else to spend candy on - upgrading IVs. Still keeps eggs valuable seeing as that's how you get egg moves + shinies, but makes it so it's slightly easier to use different Pokemon on endless because it'll increase the max stack of a specific vitamin they can use.

I made the cost identical to first cost reduction, but numbers are not necessarily set in stone if they're too low/high. Alternatively, it could be a random amount (like 5-15 for an avg of 10?) for a lower cost.

The text is the same across all the languages. I don't speak any other language but didn't want to not include it in those respective files. Hope that's okay until someone that speaks those languages can provide a translation.

Also pulls up/refreshes the IV page whenever you max an IV so you can see what IV you rolled to max.

https://github.com/pagefaultgames/pokerogue/assets/85713900/1cdff3c2-b4fb-4d48-99a5-95c1bc39d273

### DIFF
--- a/src/locales/de/starter-select-ui-handler.ts
+++ b/src/locales/de/starter-select-ui-handler.ts
@@ -28,5 +28,6 @@ export const starterSelectUiHandler: SimpleTranslationEntries = {
     "cycleNature": "N: Wesen Ändern",
     "cycleVariant": "V: Seltenheit ändern",
     "enablePassive": "Passiv-Skill aktivieren",
-    "disablePassive": "Passiv-Skill deaktivieren"
+    "disablePassive": "Passiv-Skill deaktivieren",
+    "maxIV": "Max a Random IV"
 }
--- a/src/locales/en/starter-select-ui-handler.ts
+++ b/src/locales/en/starter-select-ui-handler.ts
@@ -28,5 +28,6 @@ export const starterSelectUiHandler: SimpleTranslationEntries = {
     "cycleNature": 'N: Cycle Nature',
     "cycleVariant": 'V: Cycle Variant',
     "enablePassive": "Enable Passive",
-    "disablePassive": "Disable Passive"
+    "disablePassive": "Disable Passive",
+    "maxIV": "Max a Random IV"
 }
--- a/src/locales/es/starter-select-ui-handler.ts
+++ b/src/locales/es/starter-select-ui-handler.ts
@@ -28,5 +28,6 @@ export const starterSelectUiHandler: SimpleTranslationEntries = {
     "cycleNature": 'N: Cambiar Naturaleza',
     "cycleVariant": 'V: Cambiar Variante',
     "enablePassive": "Activar Pasiva",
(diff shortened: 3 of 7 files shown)


j-diefenbach:
I think rather than maxing a random stat it could increase one of your choosing by 5 or so. Then the cost to increase could go up, this way it allows u to make mons more usable / better but without going straight to max stat for an IV, encouraging players to still try to catch them to max it (unless they want to spend more candies)

LiamCrsd:
Maybe also bought hidden ability with candy can be an other option too

td76099:
> I think rather than maxing a random stat it could increase one of your choosing by 5 or so. Then the cost to increase could go up, this way it allows u to make mons more usable / better but without going straight to max stat for an IV, encouraging players to still try to catch them to max it (unless they want to spend more candies)

Couple things about this:
1. I am not necessarily against it being cheaper and having it increase by 5, but the point of this was not to be an additional choice to cost reduce/passive, but instead a sink for when you unlock all of those things.
2. Making it cost more per upgrade would probably require DB change, struct change on the API layer, and an object change on the front end. Not necessarily something I wanted to do without making sure it would be accepted, and it's also more of a core developer change than just a random guy like me.

td76099:
> Maybe also bought hidden ability with candy can be an other option too

The community seems pretty against being able to buy shinies, but pretty okay with being able to buy IVs. Buying HAs is somewhere in-between, but I think it ultimately goes too far. The point of making it IVs was so it was something pretty trivial, but still feels okay spending candy on.

td76099:
> I like the idea of unique upgrades for candies. Passives and Cost Reduction are things you cannot earn by playing the game where IVs are. I feel like we need a different solution to excess candies, although I do agree it is something to work on.

Generally agree - that's why it's IVs and not shiny, egg moves, hidden ability, nature, etc. Ultimately the idea of it was to be a cost sink for after you purchase the more unique ones. So in that regard I'd actually be okay with bumping the cost up. I am not sure what else you could spend candies on that would be unique/low committal as IVs. A third tier of cost reductions? Doesn't really sound super exciting tbh. I suppose it could interact with other game modes that are being worked on, but idk.

Also think of this scenario - you hatch a Latios (or whatever mon you can think of) with -SPA nature and 4 SPA IV. It's entirely possible you can play 200 hours and never see another one. With current mechanics, you can somewhat quickly fix the nature, but you'll never fix the IVs.

Flashfyre:
A solution for excess candies is already well planned out. I'm not convinced at this point to add this in, at least until that feature is in.

bennybroseph:

PR #808, opened 2024-05-13
adds daily run to offline mode

Nexllon:
Fixed fetchDailyRunSeed() to reject properly, had unhandled errors.
Made a function to handle daily generation inside initDailyRun() and added a isLocal check to run based on that:
online -> generate based on seed from fetchDailyRunSeed()
offline -> generated based on a simple base64 string date

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -333,15 +333,15 @@ export class TitlePhase extends Phase {
           this.end();
         });
       };
-
-      if (!Utils.isLocal){
+      // If Online, calls seed fetch from db to generate daily run. If Offline, generates a daily run based on current date.
+      if (!Utils.isLocal) {
         fetchDailyRunSeed().then(seed => {
           generateDaily(seed);
         }).catch(err => {
           console.error("Failed to load daily run:\n", err);
         });
       } else {
-        generateDaily(btoa(new Date().toISOString().substring(0, 10)))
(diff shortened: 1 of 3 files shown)


src/phases.ts
```diff
@@ -332,9 +332,17 @@ export class TitlePhase extends Phase {
           this.scene.sessionPlayTime = 0;
           this.end();
         });
-      }).catch(err => {
-        console.error("Failed to load daily run:\n", err);
-      });
+      };
+
+      if (!Utils.isLocal){
```
bennybroseph, line 337: Would like a very simple comment above this line explaining this addition

Admiral-Billy:
I can confirm offline works as expected, cannot verify online although it *looks* fine to me heh

bennybroseph (changes requested):

src/phases.ts
Nexllon: Sure, just did!

bennybroseph (approved):

Flashfyre:
Daily runs on offline mode bypass the once per day reward. This is a change in balance between online and offline.

Nexllon:
> Daily runs on offline mode bypass the once per day reward. This is a change in balance between online and offline.

That's true, I was thinking about how easily moddable everything is and forgot about the normal players who want/need the same experience.
I'll try to make an offline check for that using local storage as the rest of the offline info

PR #811, opened 2024-05-13
Added Ability Triggers Localization

bennybroseph:
Adds the base files needed to start translating Ability trigger messages. Provides one as an example.

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -3,7 +3,7 @@ import { Type } from "./type";
 import * as Utils from "../utils";
 import { BattleStat, getBattleStatName } from "./battle-stat";
 import { PokemonHealPhase, ShowAbilityPhase, StatChangePhase } from "../phases";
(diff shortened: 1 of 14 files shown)


Adri1 (approved):

Adri1:
While you are on the topic, it is possible to same with the Passive, which displays the same way as regular abilities? Or it is too different?

Flashfyre (commented):
Do we need a broader category for this? Feels a bit wasteful as-is.

Madmadness65 (approved):

bennybroseph:
There's going to be a lot of these. Certainly more than other localization files for menu items. There's ~30-40 already in the game that need to be added to this new file and I'm sure there's plenty more abilities that have trigger messages that will needed to be added once they're implemented. It's really just about how many lines we're okay with per file. I'm assuming it makes translating easier when the number is lower, but maybe that's backwards.

PR #816, opened 2024-05-13
Implement Trump Card move

hayuna:
# without pressure
## 1st attack (PP left: 4, move power: 40)
Screenshot 2024-05-14 at 17 32 13

## 2nd attack (PP left: 3, move power: 50)
Screenshot 2024-05-14 at 17 32 23

## 3rd attack (PP left: 2, move power: 60)
Screenshot 2024-05-14 at 17 32 40

## 4th attack (PP left: 1, move power: 80)
Screenshot 2024-05-14 at 17 32 54

## 5th attack (PP left: 0, move power: 200)
Screenshot 2024-05-14 at 17 33 08

# with pressure
## 1st attack (PP left: 3, move power: 50)
Screenshot 2024-05-14 at 17 47 50

## 2nd attack (PP left: 1, move power: 80)
Screenshot 2024-05-14 at 17 48 01

## 3rd attack (PP left: 0, move power: 200)
Screenshot 2024-05-14 at 17 48 46


src/data/move.ts
```diff
@@ -1885,6 +1885,35 @@ export class VariablePowerAttr extends MoveAttr {
   }
 }
 
+
+export class LessPPMorePowerAttr extends VariablePowerAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
```
TempsRay, line 1890: Add TSdoc for new classes/apply's

src/data/move.ts
```diff
@@ -1885,6 +1885,35 @@ export class VariablePowerAttr extends MoveAttr {
...
+    const ppMax = move.pp;
+    const ppUsed = user.moveset.find((m) => m.moveId === move.id).ppUsed;
+    const ppRemains = ppMax - ppUsed;
+    const power = args[0] as Utils.NumberHolder;
+
+    switch (ppRemains) {
```
TempsRay, line 1896: nit: feels a bit weird going from case 3->0 instead of case 0->3 since the numbers should all be ascending or descending

TempsRay (commented):
Code looks mostly good to me. Did you test and verify it works like it's supposed to?

src/data/move.ts
```diff
@@ -1885,6 +1885,35 @@ export class VariablePowerAttr extends MoveAttr {
   }
 }
 
+
+export class LessPPMorePowerAttr extends VariablePowerAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    const ppMax = move.pp;
+    const ppUsed = user.moveset.find((m) => m.moveId === move.id).ppUsed;
```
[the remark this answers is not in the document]
TempsRay: If you are 5/5 pp and use trump card against a pressure opponent the pp will drop to 3 before calculating bp (50bp)

src/data/move.ts
```diff
@@ -1904,6 +1904,49 @@ export class VariablePowerAttr extends MoveAttr {
...
+   * @param args Utils.NumberHolder for calculating power
+   * @returns true if the function succeeds
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    const ppMax = move.pp;
+    let ppUsed = user.moveset.find((m) => m.moveId === move.id).ppUsed;
+    
+    if(target.hasAbility(Abilities.PRESSURE)) {
+      ppUsed += 1;
+    }
```
TempsRay, line 1922: Just straight up adding 1 will result in an edge case if the move is at 1/5 right?

TempsRay (changes requested):
Can you just do a quick test to verify how this works with and without pressure? You can get some pretty simple screenshots by opening the console (F12) and showing examples since you can log the base power (which is different from damage) of the moves you use.

src/data/move.ts
hayuna: If we use this twice, we will have 1PP, then to avoid negative numbers, I reduced PP to 0

hayuna:
Added description with screenshots of using this move with and without pressure

TempsRay (approved):

bennybroseph:
would like the comments updated to match [this write-up](https://discord.com/channels/1125469663833370665/1239957703820836975)

hayuna:
@bennybroseph Done

bennybroseph (approved):

PR #818, opened 2024-05-13
Fixes status immunity abilities not showing on UI

notpatchmaybe:
Status abilities ( Limber, Insomnia, Immunity, Magma Armour, Water Veil, Vital Spirit, Leaf Guard, Sweet Veil, Water Bubble, Pastel Veil, Thermal Exchange, Purifying Salt ) will now show on the left ability/passive pane when they successfully immune a status, and correctly send their chat box message. [Example](https://imgur.com/aJ5aMSj)

The change in the applyPreSetStatusAbAttrs removes the !simulated arg so that the message doesn't get suppressed by .attr.
The change in getTriggerMessage is because after getting the trigger message to actually show, turns out it never worked and just produced undefined. Instead it accesses the immuneEffects array.

One bug still remains in the sleep prevention abilities, as yawn is handled by BattlerTagImmunityAbAttr which would need another fix to properly send a chat message.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,6 +2,7 @@
 <html lang="en">
 
 <head>
+	<meta charset="UTF-8" />
 	<title>PokéRogue</title>
 	<meta name="description" content="A Pokémon fangame heavily inspired by the roguelite genre. Battle endlessly while gathering stacking items, exploring many different biomes, and reaching Pokémon stats you never thought possible." />
 	<meta name="theme-color" content="#da3838" />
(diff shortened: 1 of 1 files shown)


CodeTappert:
Could you maybe think already about localization?

notpatchmaybe:
> Could you maybe think already about localization?

What is there for me to think about? None of this PR even affects plain text...

CodeTappert:
> > Could you maybe think already about localization?
> 
> What is there for me to think about? None of this PR even changes plain text?

Oh sorry. I was not aware that the hardcoded english text was already there before your PR: Then nevermind

Xavion3:
What if you just change it from `!simulated` to `simulated`? Rewriting a function signature kinda suss too, but doing that and entirely deleting an argument's functionality just to patch over slightly faulty logic seems like a bad idea.

PR #825, opened 2024-05-13
Draft: Feature for Hidden Ability Gacha machine

td76099:
@Piefrenzy on Discord

Quick draft for a gacha machine that has a higher chance of giving hidden abilities.

Doesn't exclude Pokemon that don't have hidden abilities - just raises it from a 1/256 to 1/128 that it is generated with a Hidden Ability.

There seems to be some level of interest for something like this, but wanted to check if something like this would be approved before doing all of it + testing.


bennybroseph:

PR #826, opened 2024-05-13
Update Form Change Item Check

Brandon-Rodgers:
Updated item check for when applying item to a Pokemon

[PR #826](<https://github.com/pagefaultgames/pokerogue/pull/826>)

### What this fix does:
This allows an item to be given to valid Species that can use the given item even if not immediately.

Ex: Shaymin cannot be given Gracidea unless it is Daytime, this allows you to give it to Shaymin even at night, then when day time comes around Shaymin will change form

Just like Rayquaza and a mega stone which currently works. You can give a mega stone even if they do not know Dragon Ascent. This fix should apply this same logic to all Pokemon that use Evolution items. (Currently Shaymin is the only known Pokemon that this does not work for, and is the only one with an Item & Time of day requirement)

### What this fix does not do:
This will _not_ let you use incorrect items on species. So you will not be able to give max mushrooms to a Pokemon if that current species cannot g-max, or evolution stones like Thunder Stone cannot be given to Pichu because Pichu specific species cannot evolve with a thunder stone.

### How it fixes the issue:
It uses similar logic to the shop when it chooses if an evolution item should spawn or not. 
The shop checks if any current species in party has SpeciesFormEvolution with an evolution item requirement.
This fix checks if the Pokemon you are applying to has a SpeciesFormEvolution with an item or items and maps it to an Array, then checks if the applied item is in that array. 
The old process was trying to do this shorthand, but for some reason on Shaymin it was not detecting an evolution item, so I just changed it to use a deeper check.

https://github.com/pagefaultgames/pokerogue/assets/9848614/b1d78204-6a1d-45dc-872f-6303b82e697b


bennybroseph:
> In general this should allow any item to be given to your Pokemon whether they can evolve yet or not. As long as it shows in the shop, it can be given to the proper Pokemon

That doesn't sound like the solution we're looking for. The shop should only show items that are valid and when applying an item, it should only do so if it is also valid. What happens when I have a Pichu and a Pikachu? Can I give Pichu a Max Mushroom?

Brandon-Rodgers:
This should work like Rayquaza, where you can give the megastone even when Dragon Ascent is not learned. But it cannot be given to other Pokemon if they cannot use it. 
It uses the same check logic as the shop where a Thunderstone only shows up if a Pokemon can use it to evolve. 
So if you try to apply it to Pichu it will check if Pichu has SpeciesFormEvolution attribute with the EvolutionItem.THUNDER_STONE. Since Pichu does not have that attribute with that item, then it will say it will not affect Pichu

Brandon-Rodgers:
https://github.com/pagefaultgames/pokerogue/assets/9848614/80141eb2-178d-4f87-bb89-9e33da5a708d

Video of it not affecting Pichu and only Pikachu

src/modifier/modifier-type.ts
```diff
@@ -745,8 +745,11 @@ export class FormChangeItemModifierType extends PokemonModifierType implements G
   constructor(formChangeItem: FormChangeItem) {
     super('', FormChangeItem[formChangeItem].toLowerCase(), (_type, args) => new Modifiers.PokemonFormChangeItemModifier(this, (args[0] as PlayerPokemon).id, formChangeItem, true),
     (pokemon: PlayerPokemon) => {
-      if (pokemonFormChanges.hasOwnProperty(pokemon.species.speciesId) && !!pokemonFormChanges[pokemon.species.speciesId].find(fc => fc.trigger.hasTriggerType(SpeciesFormChangeItemTrigger)
-        && (fc.trigger as SpeciesFormChangeItemTrigger).item === this.formChangeItem))
+      if (pokemonFormChanges.hasOwnProperty(pokemon.species.speciesId) 
+        && !!pokemonFormChanges[pokemon.species.speciesId].find(fc => fc.trigger.hasTriggerType(SpeciesFormChangeItemTrigger))
+        && pokemonFormChanges[pokemon.species.speciesId].filter(fc => fc.trigger.hasTriggerType(SpeciesFormChangeItemTrigger))
+          .map(fc => fc.findTrigger(SpeciesFormChangeItemTrigger) as SpeciesFormChangeItemTrigger)
+          .flat().flatMap(fc => fc.item).includes(this.formChangeItem))
```
bennybroseph, line 752: Could we get some basic comments to explain some of these checks and the surrounding code?

bennybroseph (changes requested):

bennybroseph:
Closing due to inactivity

PR #835, opened 2024-05-13
Testing setup

PhilipHayes:
A barebones structure for working with unit testing in vitest.


XGSleepWalker:
A couple of notes here: 

- Currently as the tests are, they generate a bunch of failures. Probably it would be best to mark them as `describe.skip` until someone fills something inside of them.
- Any test we try to kick off starts complaining about modules and undefined behavior. I'm not a js/ts expert, but it looks like this is coming from circular dependencies. We need to resolve those as we start adding tests, but that will require major refactorings of the entire codebase, so we can start out by adding tests on the more simple and isolated files like the utils.ts file.

src/test/utils.test.ts
XGSleepWalker, line 1: Can you add here the tests that already exist in src/utils.test.ts and remove that file?

PhilipHayes: Sure thing.

XGSleepWalker (approved):

bennybroseph (approved):

XGSleepWalker:
Closed for https://github.com/pagefaultgames/pokerogue/pull/1908

PR #840, opened 2024-05-14
Partially implemented mimicry.

Electrolyzer:
Edge cases where other effects also change type may exhibit some unintended behavior (soak, forest's curse).

Streamlined on summon weather/terrain effects by adding an onTerrainChange and onWeatherChange call when switching in and removing the relevant onSummonEffect so that when coding an ability that activates on specific terrain or weather, you only need to implement the onTerrainChange or onWeatherChange attribute. Added a getWeather() function to arena.ts to simplify it (identical to the preceding getTerrain() function).

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2126,10 +2126,21 @@ export class PostTerrainChangeAddBattlerTagAttr extends PostTerrainChangeAbAttr 
   }
 }
 
+/**
+ * When the terrain changes or the pokemon is summoned, change its type according to
+ * a Record passed into the constructor.
+ * @
+ * @extends PostTerrainChangeAbAttr
+ * @see {@link applyPostTerrainChange}
+ */
 export class PostTerrainChangeTypeAttr extends PostTerrainChangeAbAttr {
-  private oldType: Type[];
   private terrainToTypeMap: Record<TerrainType, Type>;
 
+  /**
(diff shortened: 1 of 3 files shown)


src/data/ability.ts
```diff
@@ -2126,6 +2126,43 @@ export class PostTerrainChangeAddBattlerTagAttr extends PostTerrainChangeAbAttr
...
+      pokemon.summonData.types = null;
+      pokemon.updateInfo();
+      pokemon.scene.queueMessage(getPokemonMessage(pokemon, ` transformed\nback into its original type!`));
+      return true;
+    }
+
+    pokemon.summonData.types = [this.terrainToTypeMap[terrain]]
+    pokemon.updateInfo();
+    pokemon.scene.queueMessage(getPokemonMessage(pokemon, ` transformed\ninto the ${Utils.toReadableString(Type[pokemon.getTypes()[0]])} type!`));
+    return true;
```
bennybroseph, line 2162: Could we get some comments explaining this function?

bennybroseph (changes requested):

src/data/ability.ts
Electrolyzer: Added comments. Let me know if there is anything else or something in particular I should explain more. Thank you!

Electrolyzer: (If the question was about terrainToTypeMap that was already explained above in the comment on the constructor)

flx-sta:
Considered stale.

PR #403, opened 2024-05-02
feat: added localisation for eggs and french translations

TeKrop:
Added localisation support for eggs (list and gacha). Also added french translation.
Tested in both English and French. Italian, Spanish and German translations are copied from English.

Here are a few changes which could be discussed if needed :
- Gacha types french descriptions (for french native speakers or translators)
- Pull options text refactoring, in order to facilitate changes
- Pull options text resizing to make the french translation visually OK (see following screenshots)

Pull options today
image

Pull options after changes
image

Pull options in french
image

Eggs list in french
image
image
image
image

Eggs rarity after pull
image
image

Too many eggs
image

PR #841, opened 2024-05-14
Added Egg List Localization

josericardo-fo:


josericardo-fo:
PR #403 already did this for the Egg List and additionally to Egg Gatcha

Issue #839, opened 2024-05-14
Sinistea (antique form) can't evolve with chipped pot ("it will have no effect")


j-diefenbach:
Related to #636 as evolution items now check formIndex however sinistea antique form is index 1. The chipped pot can still spawn however attempting to use it gives the 'but it will have no effect' message
Cracked pot working as intended (using overrides to have phony form):

https://github.com/pagefaultgames/pokerogue/assets/105332964/3ada93bb-3721-47ef-a1ea-3baa3982a5c4

Using chipped pot on antique form which should prompt evolution (overriden to have antique form):

https://github.com/pagefaultgames/pokerogue/assets/105332964/26590c22-b675-44e9-9178-906085ac9ce8

I imagine this also effects poltchageist and its counterfeit/artisan forms

bennybroseph:
Good catch. Not something to revert the commit over though as this edge case is not as bad as what the commit fixed. If you already have a solution in mind for this, I'm open to suggestions.

j-diefenbach:
Yeah i think using pokemon.GetFormKey() !== SpeciesFormKey.GIGANTAMAX will cover literally all the possibly cases where evo items can be used on non-base-form pokemon but we could also throw on the rest of the SpeciesFormKey cases

PR #636, opened 2024-05-08
Issue#613 Prevent evo item use on none base form Pokes

JackGilham:
Issue#613.

Have added additional EvolutionItemModifierType condition that prevents non-base form mons from taking Evo Items.

Fixes scenarios like where you may have a GMAX Pikachu and some other Thunderstone evo mon on your team:
1) Shop offers Thunderstone because of other evo mon
2) You are able to use Thunderstone on GMAX Pikachu because no form check.
3) GMAX Pikachu eats Thunderstone and nothing happens due checks/fails during evo.

PR #844, opened 2024-05-14
Evolution item check for g-max form instead of alternate forms (allows sinistea antique form to evolve)

j-diefenbach:
Modifies changes from #636 for compatibility with pokemon with alternate forms that also use evolution items. E.g. sinistea, poltchageist. The alternate forms are treated as formIndex 1 so getFormKey() and comparing against GIGANTAMAX seems more appropriate, see #839.
Since there are no other cases i.e. megas are always final form, eternamax, primal, origin, gignatamax single/rapid all are very limited and only apply to legendaries / one pokemon that don't use evo items I think it's safe to only check GIGANTAMAX (I double checked).
See #839 for main behaviour (antique form can't evolve with chipped pot it just says it will have no effect).

Previous pull changes still working i.e. thunder stone won't apply for g-max pikachu and g-max pikachu fused with other pokemon.

ScreenRecording2024-05-14at11 41 15-ezgif com-video-to-gif-converter
ScreenRecording2024-05-14at11 47 34-ezgif com-video-to-gif-converter

Thunder stone will work on normal evolution item pokemon fused with a g-max pokemon (e.g. normal pikachu x g-max pikachu):

ScreenRecording2024-05-14at11 52 17-ezgif com-video-to-gif-converter

Also testing some other g-max interactions

Polteagist evolution via chipped pot working as intended
ScreenRecording2024-05-14at12 13 45-ezgif com-video-to-gif-converter


j-diefenbach:
Will fix up the videos 💀

bennybroseph (approved):

Issue #810, opened 2024-05-13
The window is not centered when playing on non-mobile platforms with resolutions other than 16:9


Odizinne:
On steam deck, or any 16/10 resolution, since this game is built around 16/9, there is a huge bottom bar to fill empty space.

![image](https://github.com/pagefaultgames/pokerogue/assets/102679854/636f24cb-35b2-4d9c-87dd-e92c0b4a58ca)

While it is necessary, it does look a bit weird.

It would be nice to have the main window centered, with thinner top / bottom bar, or at least to have an option for this.

Thank's for the hard work on this project.

torranx:
Opened a new pull request.
[As mentioned from a previous closed PR](https://github.com/pagefaultgames/pokerogue/pull/849#issuecomment-2109235066), creating an option in game settings is not ideal

PR #849, opened 2024-05-14
Added [Center Window Vertically] option in Game Settings

torranx:
Resolves enhancement request/suggestion on #810 

- Added _Center Window Vertically_ option in Game Settings


bennybroseph:
I don't think this should be a whole setting, it should just do it automatically if you aren't on mobile

torranx:
> I don't think this should be a whole setting, it should just do it automatically if you aren't on mobile

yeah i actually agree, my problem was the user preference when playing but i just realize i can check the device orientation instead to automatically do this, will update the PR accordingly, thanks

PR #850, opened 2024-05-14
Do not trigger focus band effect on status damage

GriffinDZ:


GriffinDZ:
https://trello.com/c/NmVfN8xy/22-focus-band-keeping-pokemon-alive-through-status-damage

Files used for reproduction / testing the fix attached


[overridests.txt](https://github.com/pagefaultgames/pokerogue/files/15302238/overridests.txt)

(Just modifies the focus band to always trigger)
[modifierts.txt](https://github.com/pagefaultgames/pokerogue/files/15302239/modifierts.txt)

GriffinDZ:
Not sure if there's an optimal way to handle the parameters here, was hoping to be able to only pass two with a named param, but looks like TypeScript only does that through object passing

src/field/pokemon.ts
```diff
@@ -1657,7 +1657,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     return result;
   }
 
-  damage(damage: integer, ignoreSegments: boolean = false, preventEndure: boolean = false): integer {
+  damage(damage: integer, ignoreSegments: boolean = false, preventEndure: boolean = false, statusDamage: boolean = false): integer {
```
knispeja, line 1660: Instead of creating a new param, can't you just reuse `preventEndure`? It appears to me that Endure/Sturdy are also incorrectly triggering on status damage, if I'm reading this correctly

GriffinDZ: Let me check to see if they also are triggering incorrectly, I separated them because I figured there isn't a scenario where you could be taken from 100% to 0% with status damage, but now that I think about it maybe you could with healing + high toxic stacks

GriffinDZ: It looks like Endure/Sturdy work correctly because of something in the `lapseTag` logic (tested with a shedinja w/ sturdy set to one health manually, and it dies after being toxic'd), but I think we can reuse `preventEndure` here safely actually

bennybroseph (changes requested):
Looks like this has conflicts at the moment.

Also, can we get a video of this being tested?

flx-sta:
Considered stale

PR #851, opened 2024-05-14
Automatically center window vertically when playing on landscape

torranx:
Resolves enhancement #810 

EDIT:

If device is not a mobile device and you'd disable the touch controls, it will center the game window.
- this will be ideal for those playing with tablets along with external controller, and handheld devices like steam deck and the likes
- this will technically not matter on large screen devices like laptops and pc, since height of the game window will fill the screen as much as possible keeping the aspect ratio
- we only want this to happen on landscape orientation

If device is mobile device, this will retain the existing behavior- the window should be at the top of the screen.

**MOBILE DEVICES**
- PORTRAIT
Screenshot 2024-05-15 at 11 08 39 PM

- LANDSCAPE
Screenshot 2024-05-15 at 11 09 40 PM

**NON-MOBILE DEVICES**
- With touchscreen capability (please note in screenrecording, it is set to Desktop (touch) to simulate tablet or other handheld devices like steam deck)

https://github.com/pagefaultgames/pokerogue/assets/68144167/bf3294db-f0ad-46d3-a2c6-805b5e78628b

- Without touchscreen capability like pc, this will always center the window

### DIFF
--- a/src/system/settings.ts
+++ b/src/system/settings.ts
@@ -2,7 +2,7 @@ import SettingsUiHandler from "#app/ui/settings-ui-handler";
 import { Mode } from "#app/ui/ui";
 import i18next from "i18next";
 import BattleScene from "../battle-scene";
-import { hasTouchscreen } from "../touch-controls";
+import { hasTouchscreen, isMobile } from "../touch-controls";
 import { updateWindowType } from "../ui/ui-theme";
 import { PlayerGender } from "./game-data";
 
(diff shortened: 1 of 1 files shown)


grenlith:
I really preferred the implementation here: https://github.com/pagefaultgames/pokerogue/pull/849

I made the changes in that PR to my own instance and it has been fantastic:

I've been playing on iPad as a PWA, mirroring to Apple TV, and honestly really preferred having the option. When attached to a keyboard, I center the game screen and hide the touch controls. If the keyboard isn't attached, I move it back to the top and have touch controls on, and having it at the top helps keep the controls out of the way. 

I never hold my iPad in portrait unless I'm reading something long, so this is all landscape orientation.

torranx:
I'm down for both but it will actually boil down to user preference still and what the devs think

torranx:
For reference, this is what it looks like on a device with landscape orientation using steam deck's screen size:

BEFORE:
Screenshot 2024-05-15 at 3 02 38 PM

AFTER:
Screenshot 2024-05-15 at 3 03 46 PM

Now that I compare them, I personally think the _before_ is actually better user experience for handheld devices that DO NOT have built in controls because of the bottom space it provides (still depending on screen size) and vice versa.

https://github.com/pagefaultgames/pokerogue/assets/68144167/158a834e-5dca-4767-b295-adb03821e661

grenlith:
iPad screenshots

Default
image

Centered with Touch
image

Centered without Touch
image

bennybroseph (approved):

PR #858, opened 2024-05-14
Modify legend gacha to be even split

Xavion3:
Modifies the functionality of the legendary gacha banner to be evenly split between all possible legends, as opposed to a 64 length cycle. This works by creating super cycles where it goes through a shuffled list exactly once before reshuffling and restarting, so it is still random, but each 25 week cycle will always contain every legend exactly once.

Due to the limited seed space this is still technically going to loop, but instead of looping every 64 banners it will loop every 1638400 banners, or specifically every 65536 super cycles which seems more than sufficient.

This is probably retroactive to any eggs anyone currently has but hasn't hatched, but so was the change earlier today to the banner so I'm not sure that's a high concern.

### DIFF
--- a/src/data/egg.ts
+++ b/src/data/egg.ts
@@ -95,9 +95,14 @@ export function getLegendaryGachaSpeciesForTimestamp(scene: BattleScene, timesta
 
   let ret: Species;
 
+  // 604800000 is the number of miliseconds in one week
+  const weekOffset = Utils.getSunday(new Date(timestamp)).getTime(); // Timestamp of current week
+  const offset = Math.floor(Math.floor(weekOffset / 604800000)/legendarySpecies.length); // Cycle number
+  const index = Math.floor(weekOffset / 604800000)%legendarySpecies.length // Index within cycle
+
   scene.executeWithSeedOffset(() => {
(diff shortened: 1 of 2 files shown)


TempsRay:
It's still going to take a week to change?

src/data/egg.ts
```diff
@@ -95,9 +95,14 @@ export function getLegendaryGachaSpeciesForTimestamp(scene: BattleScene, timesta
 
   let ret: Species;
 
+  // 604800000 is the number of miliseconds in one week
+  const weekOffset = Utils.getSunday(new Date(timestamp)).getTime(); // Timestamp of current week
+  const offset = Math.floor(Math.floor(weekOffset / 604800000)/legendarySpecies.length); // Cycle number
+  const index = Math.floor(weekOffset / 604800000)%legendarySpecies.length // Index within cycle
```
TempsRay, line 101: nit: spaces in between the `/` and `%`

TempsRay (approved):
Can always change the length of the rotation later and I'm all for making it shorter than a week but this at least fixes the current issue of some mons not showing up at all or other mons showing up way too often

Flashfyre (approved):

PR #871, opened 2024-05-14
Enable swapping A and B buttons on keyboard controls

jatinkohli:
Determines which of x and z keys are action and cancel based on the "SWAP_A_B" setting, as is done for the respective gamepad buttons. Also rename field in ui-inputs to inputController from inputsController for consistency with battle-scene class field.


Flashfyre:
I think this would cause confusion since it's grouped with the controller settings. Also it would be strange for this to be the only changeable keybind, we'll probably just add keybinds at some point.

bennybroseph:
I believe @Greenlamp2 is already working on a fully rebindable menu option.

PR #881, opened 2024-05-14
Added a champion ribbon on enemy pokemon if they have a classic win.

JonStudders:
Checks using root species.

No Classic Wins:
image

Classic Win Root Species:
image

Classic Win non-root Species:
image

To-do:
Maybe if the pokemon hasn't been caught shift the icon over?

### DIFF
--- a/src/ui/battle-info.ts
+++ b/src/ui/battle-info.ts
@@ -8,6 +8,7 @@ import BattleScene from '../battle-scene';
 import { Type, getTypeRgb } from '../data/type';
 import { getVariantTint } from '#app/data/variant';
 import { BattleStat } from '#app/data/battle-stat';
+import { StarterDataEntry } from '#app/system/game-data';
 
 const battleStatOrder = [ BattleStat.ATK, BattleStat.DEF, BattleStat.SPATK, BattleStat.SPDEF, BattleStat.ACC, BattleStat.EVA, BattleStat.SPD ];
 
@@ -33,6 +34,7 @@ export default class BattleInfo extends Phaser.GameObjects.Container {
   private nameText: Phaser.GameObjects.Text;
   private genderText: Phaser.GameObjects.Text;
   private ownedIcon: Phaser.GameObjects.Sprite;
+  private championRibbon: Phaser.GameObjects.Sprite;
   private teraIcon: Phaser.GameObjects.Sprite;
   private shinyIcon: Phaser.GameObjects.Sprite;
   private fusionShinyIcon: Phaser.GameObjects.Sprite;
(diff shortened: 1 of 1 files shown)


src/ui/battle-info.ts
```diff
@@ -260,10 +268,16 @@ export default class BattleInfo extends Phaser.GameObjects.Container {
     if (!this.player) {
       const dexEntry = pokemon.scene.gameData.dexData[pokemon.species.speciesId];
       this.ownedIcon.setVisible(!!dexEntry.caughtAttr);
+
       const dexAttr = pokemon.getDexAttr();
       if ((dexEntry.caughtAttr & dexAttr) < dexAttr || !(pokemon.scene.gameData.starterData[pokemon.species.getRootSpeciesId()].abilityAttr & Math.pow(2, pokemon.abilityIndex)))
         this.ownedIcon.setTint(0x808080);
 
+      const starterDataEntry: StarterDataEntry = pokemon.scene.gameData.starterData[pokemon.species.getRootSpeciesId()];
```
TempsRay, line 276: The code for checking if a ribbon displays already exists [here](https://github.com/pagefaultgames/pokerogue/blob/main/src/ui/summary-ui-handler.ts#L311). Wonder if it's worth moving it somewhere else since it seems like multiple places will use it.

That being said, the ribbon should check for `getRootSpeciesId(true)` as well as `getRootSpeciesId(false)` so if you have something like Magmortar it checks if both Magby and Magmar have ribbons

Flashfyre:
Gonna be real with you, I don't think this is necessary information to have on the battle screen.

src/ui/battle-info.ts
JonStudders: Very good point raised,
thank you for clarifying in discord.

JonStudders:
> Gonna be real with you, I don't think this is necessary information to have on the battle screen.

Hi Sam,

As someone who is personally attempting to farm vouchers by beating classic multiple times, I am having to resort to a spreadsheet to track which starters I have completed.

If it's an idea of wanting to keep the UI cleaner, how would you feel about the ability to toggle this within settings?

It could be apart of a bigger game setting such like 'Show additional battle information' and include things like:
- Weaknesses
- HP Numbers

src/ui/battle-info.ts
JonStudders: This has been resolved in my most recent PR.

TempsRay:
> Gonna be real with you, I don't think this is necessary information to have on the battle screen.

It's a very, very common and popular request users have

TempsRay (approved):
Looks good and it's a fantastic QOL change lots of users want

TempsRay (approved):
Thanks for adding the classic check and refactor!

src/ui/battle-info.ts
```diff
@@ -260,7 +268,16 @@ export default class BattleInfo extends Phaser.GameObjects.Container {
     if (!this.player) {
       const dexEntry = pokemon.scene.gameData.dexData[pokemon.species.speciesId];
       this.ownedIcon.setVisible(!!dexEntry.caughtAttr);
+
       const opponentPokemonDexAttr = pokemon.getDexAttr();
+      if (pokemon.scene.gameMode.isClassic) {
+        if(pokemon.scene.gameData.starterData[pokemon.species.getRootSpeciesId()].classicWinCount > 0 && pokemon.scene.gameData.starterData[pokemon.species.getRootSpeciesId(true)].classicWinCount > 0) {
+          if (!dexEntry.caughtAttr) {
```
molker, line 275: Why would there be a case where you've beaten the game with a pokemon you haven't caught?

JonStuddersAccess: There may not be, but I felt like the functionality is still good to have. Just incase :)

JonStudders:
Merged + Removed the is captured.

PR #884, opened 2024-05-15
feat: add hide unimplemented moves setting

heretere:
**What Changed?:**
I added a setting to allow completely hiding unimplemented moves from the game

As a result,
- Unimplemented moves aren't prompted on level
- Unimplemented moves aren't shown in the starter select

**Why?:**
For many players, these move are never selected, as they're not implemented. This is just a QOL improvement, so the move swap isn't triggered on level up.

Relevant Feature Vote: https://discord.com/channels/1125469663833370665/1238282579329683499


LeCehlou:
If someone starts a game with this setting, and lager change the settings, they will not be able to get back the unimplemented moves if I'm not wrong

flx-sta:
Considered stale.

Issue #883, opened 2024-05-14
Tinted pokeball false positive from abilities


TempsRay:
More info [here](https://discord.com/channels/1125469663833370665/1240069508127789117) but this affects all Pokemon with only 1 ability and makes the caught icon tinted when they appear with their hidden abilities even if the user owns the hidden ability.

This is because in the dexattr the ability is stored as a binary bitmask where 1 is the first ability, 2 is the second ability, and 4 is the hidden ability.

A Pokemon with only 1 ability has their hidden ability at abilityIndex 1 instead of 2 so when the code checks 2^abilityIndex it's getting a value of 2 instead of 4 and this mismatch causes the caught icon to be tinted even when it shouldn't be.

PR #892, opened 2024-05-15
Fix tinted caught icon false positive

TempsRay:
Issue [here](https://github.com/pagefaultgames/pokerogue/issues/883)

Normally a Pokemon has their hidden ability as abilityIndex 2. However Pokemon with only 1 ability have their hidden ability with abilityIndex 1 instead of 2. This causes a problem when trying to compare the Pokemon's abilityIndex with the starter's abilityAttr where the hidden ability is stored as 4, causing a dark tinted Pokeball to appear even if the player owns the hidden ability.

I added a check so that if a Pokemon only has 1 ability has a hidden ability then it will be treated as being 2 instead of 1. I also went through and renamed some variables to hopefully make the logic a lot more clear and readable.

[Before fix](https://cdn.discordapp.com/attachments/1240069508127789117/1240070681236865155/image.png?ex=66453951&is=6643e7d1&hm=837211f91a5ed1b9d3ee1ee742156a0be6ff38c8ab56db64780081613dc3497f&)

[After fix](https://cdn.discordapp.com/attachments/1240069508127789117/1240108420825354281/image.png?ex=66455c77&is=66440af7&hm=bf4914e8d3c972ab41f279de6aeb30cb57f229bfb5617ac33f3be14ac9510865&)

### DIFF
--- a/src/ui/battle-info.ts
+++ b/src/ui/battle-info.ts
@@ -268,8 +268,8 @@ export default class BattleInfo extends Phaser.GameObjects.Container {
       // If the opposing Pokemon only has 1 ability and is using the hidden ability, it should have the same
       // behavior if it had 2 abilities so it is set to 2 so it lines up with the math checking the proper
       // hidden ability abilityAttr which is 4
-      const opponentPokemonOneAbility = (pokemon.species.getAbilityCount() == 2);
-      const opponentPokemonAbilityIndex = (pokemon.abilityIndex === 1 && opponentPokemonOneAbility) ? 2 : pokemon.abilityIndex;
+      const opponentPokemonOneNormalAbility = (pokemon.species.getAbilityCount() == 2);
+      const opponentPokemonAbilityIndex = (opponentPokemonOneNormalAbility && pokemon.abilityIndex === 1) ? 2 : pokemon.abilityIndex;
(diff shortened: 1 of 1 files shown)


src/ui/battle-info.ts
```diff
@@ -260,8 +260,21 @@ export default class BattleInfo extends Phaser.GameObjects.Container {
...
-      const dexAttr = pokemon.getDexAttr();
-      if ((dexEntry.caughtAttr & dexAttr) < dexAttr || !(pokemon.scene.gameData.starterData[pokemon.species.getRootSpeciesId()].abilityAttr & Math.pow(2, pokemon.abilityIndex)))
+      const opponentPokemonDexAttr = pokemon.getDexAttr();
+
+      // Check if Player owns all genders and forms of the Pokemon
+      const missingDexAttrs = ((dexEntry.caughtAttr & opponentPokemonDexAttr) < opponentPokemonDexAttr); 
+
+      // If the opposing Pokemon only has 1 ability and is using the hidden ability, it should have the same
+      // behavior if it had 2 abilities so it is set to 2 so it lines up with the math checking the proper
+      // hidden ability abilityAttr which is 4
```
bennybroseph, line 270: should use block comments here

bennybroseph (changes requested):

PR #895, opened 2024-05-15
Fixes toxic spikes poison damage turn 0

### LINKED ISSUE #830: Poison ticks at start of fight
No moves have been made yet in this battle, but the poison from toxic spikes ticked.
I cannot further debug the issue due to https://github.com/pagefaultgames/pokerogue/issues/684
I assume the way the game handles turns makes the "1st" turn take place during the swap prompt, so the end of turn hook is called, perhaps other hazards tick as well?
image

Electrolyzer:
Very simple fix for toxic spikes turn 0 damage bug. Fixes #830 
Bug description: If toxic spikes were out and a new pokemon encounter happened, they would take poison damage before battle start.
Fix: Instead of adding the post-turn status phase when adding status, just check for it at the end of each turn so there are no situations where it gets added at the wrong time, or added multiple times (e.g. toxic -> heal bell -> toxic causing 2 procs)

https://github.com/pagefaultgames/pokerogue/assets/107140457/faeef013-eedf-4d9d-890f-bbf5e1ba4665

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -2054,8 +2054,7 @@ export class TurnStartPhase extends FieldPhase {
       this.scene.pushPhase(new WeatherEffectPhase(this.scene, this.scene.arena.weather));
 
     for (let o of order) {
-      if (field[o].status && field[o].status.isPostTurn())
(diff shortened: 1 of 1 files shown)


flx-sta:
Considered stale.

PR #905, opened 2024-05-15
chore: add poisonTypeCondition and HitsTagAttr to moves.toxic

Raksus:
Fix the gen6 interaction of toxic

https://bulbapedia.bulbagarden.net/wiki/Toxic_(move)#Generation_VI_onwards


bennybroseph:
Did you test this? Looks like this just makes Toxic only usable by Poison types and hit Semi-Invul.

Raksus:
> Did you test this? Looks like this just makes Toxic only usable by Poison types and hit Semi-Invul.

Yeah, sorry. I missed how condition works.
I thought it applied to the rules that came afterwards.

bennybroseph:
This is also hardcoding the move to work a certain way that circumvents our current implementation options. I think it would be best to reach out in the Discord for assistance. Magnitude/Earthquake already work like this for Dig, that would be the first place to start.

PR #910, opened 2024-05-15
Update ability.ts

mbroll:
Water bubble should double the damage of water moves. The current multiplier in-game is x1.


Madmadness65 (approved):

TempsRay:
In the future please make the commit title more descriptive ("Fix water bubble") to make it easier for tracking

PR #911, opened 2024-05-15
offline - fixed continue button

Nexllon:
Fixed continue button for offline mode (didn't clear on game win/lose nor got any of the last 3 slots)
It only updates when the game saves: wins/loses, new games, new waves etc


Flashfyre (commented):
I just pushed changes that should fix this.

Nexllon:
I saw it, but you still get the first save found, not the actual last save slot you played
Nor does it fix the win/lose scenario
It's not the root of the problem, it only changed from first 2 slots to all 5, the rest still continues

flx-sta:
Considered stale.

PR #915, opened 2024-05-15
Added current biome text

JonStudders:
Added:
- Biome Text to the WaveCountText 

Screenshots:
Standard:
image
Boss:
image
Next Biome:
image


TempsRay (approved):
I still think colored text _can_ work, but it will have to be a bespoke color per biome to match with the upper right background color. That being said, pure black or white looks pretty nice and this is good information to display for newer players.

src/battle-scene.ts
```diff
@@ -1183,12 +1183,13 @@ export default class BattleScene extends SceneBase {
...
 
-	updateWaveCountText(): void {
+	updateLineOneText(): void {
 		const isBoss = !(this.currentBattle.waveIndex % 10);
-		this.waveCountText.setText(this.currentBattle.waveIndex.toString());
-		this.waveCountText.setColor(!isBoss ? '#404040' : '#f89890');
-		this.waveCountText.setShadowColor(!isBoss ? '#ded6b5' : '#984038');
-		this.waveCountText.setVisible(true);
+		const biomeString: string = getBiomeName(this.arena.biomeType);
+		this.lineOneText.setText( this.currentBattle.waveIndex.toString() + " - " + biomeString);
```
molker, line 1189: small nitpick, can the order be reversed
I think the wave number should stay where it is and the biome can be to the left of it

JonStudders: Resolved.

PR #919, opened 2024-05-15
Fix False Trigger of Corrosion

Brandon-Rodgers:
Issue: Corrosion triggers when moves use getTargetBenefitScore even if the move is not actually used by the opponent.

Fix:
Moved IgnoreTypeStatusEffectImunity to extend from PreSetStatusAbAttr instead of AbAttr so I can use the existing implemented quiet arg.

Extras:
I changed simulated to just use the first arg which is how the current checks pass the quiet arg.

I was able to remove effect from args and have it as a set parameter being passed

https://github.com/pagefaultgames/pokerogue/assets/9848614/0bfc926d-9ddb-4915-ae05-e78a9e989a5d

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -2052,6 +2052,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     switch (effect) {
       case StatusEffect.POISON:
       case StatusEffect.TOXIC:
+        console.log('checking poison immunity', this.name, types, sourcePokemon?.name, sourcePokemon?.id)
         // Check if the Pokemon is immune to Poison/Toxic or if the source pokemon is canceling the immunity
         let poisonImmunity = types.map(defType => {
           // Check if the Pokemon is not immune to Poison/Toxic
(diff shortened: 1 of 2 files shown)


DustinLin:
Just curious I'm a bit confused on the original problem, can you provide an example? Are you saying a non-poison move will trigger the ability? Providing a before fix demo could help the reviewers also

src/field/pokemon.ts
```diff
@@ -2052,6 +2052,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     switch (effect) {
       case StatusEffect.POISON:
       case StatusEffect.TOXIC:
+        console.log('checking poison immunity', this.name, types, sourcePokemon?.name, sourcePokemon?.id)
```
DustinLin, line 2055: not sure if this is needed in the final merge

Brandon-Rodgers: I thought I was thorough in removing my logs. Will remove

Brandon-Rodgers:
> Just curious I'm a bit confused on the original problem, can you provide an example? Are you saying a non-poison move will trigger the ability? Providing a before fix demo could help the reviewers also

Every time the AI's goes they trigger the Corrosion pop-up even if they do not use a move that would trigger it like poison.
This fix makes it so when the AI is trying to select a move they do not trigger the Corrosion pop-up. It will now only trigger when a move actually is used to trigger it. See below for videos of the issue. 

Salazzle has Poison Powder, and 3x Dragon Dance moves. Uses Dragon Dance and corrosion triggers

https://github.com/pagefaultgames/pokerogue/assets/9848614/9c5f63b3-dec2-46da-9059-be9bfc43dee6

Salazzle only has Dragon Dance and no moves that would proc corrosion. Uses Dragon Dance and corrosion does not trigger

https://github.com/pagefaultgames/pokerogue/assets/9848614/76a0a1f1-3bd7-4523-9b49-6f41785fc3b5

PR #929, opened 2024-05-15
Resolve TODO on Sky Drop "Add 2nd Turn Message"

brandonarude:
Currently, moves using the ChargeAttr only show custom messages on the first turn. To address the TODO for the second turn message ("The {Target} was freed from Sky Drop!"), I extended the ChargeAttr class into SkyDropChargeAttr and added a flag that would trigger the message on the second turn of the move.

By extending the class in this way, it also opens the door to implementing the other functions of this move that are found in the mainline games.


TempsRay:
While you're here in sky drop, do you mind implementing the part where it fails against 200+kg opponents?

brandonarude:
I'll certainly try!  This is my first time contributing to an open source project, and I'm still getting my head around the codebase.

brandonarude:
Alright, added the check for weight, and tested on a Ho-oh (199kg) and Crustle (200kg). As expected, it worked on the former and failed on the latter.

bennybroseph (changes requested):

brandonarude:
Alright, I think I got the JSDocs added properly! Let me know if any further changes are needed!

bennybroseph:
Lastly, can we get a video of this being tested? Need a normal interaction and one with a pokemon that is too heavy

brandonarude:
Yep! Here's the success:

[Skydrop Success-1.webm](https://github.com/pagefaultgames/pokerogue/assets/38122826/3116318a-8f69-497a-91b2-af8493a355dc)


And here's the failure:

[Skydrop Failure-1.webm](https://github.com/pagefaultgames/pokerogue/assets/38122826/e79ca27d-bb13-4ae5-80b6-064099da05c1)

bennybroseph:
Seeing multiple issues unfortunately. Sky Drop works similar to Fly/Dive/Dig except for both Pokemon. Ho-oh should have it's sprite hidden and anything that targets it misses just like Fly/etc. Ho-oh should also be immune to the damage from Sky Drop since it's flying type. They should still get hit by gust, thunder, thousand arrows, etc. 

Please read up on the following and let me know which interactions still have to be considered.

[source](https://bulbapedia.bulbagarden.net/wiki/Sky_Drop_(move))

Tempo-anon:
Sky drop is actually much more complex to fully implement. Closing this for now due to inactivity and issues

brandonarude:
Yep! Sorry about the inactivity. Makes sense to close this, bit off a bit more than I could chew!

PR #940, opened 2024-05-16
Add unthaw logic to Steam Eruption, Scorching Sands, Matcha Gotcha and all Damaging Fire moves

OrangeRed:
Quotes from the [Freeze bulbapedia entry](https://bulbapedia.bulbagarden.net/wiki/Freeze_(status_condition))
> Starting in Generation VI, if a frozen Pokémon is successfully hit by Scald, Steam Eruption, Scorching Sands, or Matcha Gotcha, it will be immediately thawed.
>
> All damaging Fire-type moves can now thaw a frozen target, regardless of whether or not they have a chance to burn;

- Add thaw target `HealStatusEffectAttr` for Steam Eruption, Scorching Sands and Matcha Gotcha since they were missing it
- Add comments to the `HealStatusEffectAttr` Class
- Add thaw target `HealStatusEffectAttr` to every attacking fire move inside of the `AttackMove` Class

~~Also added `STATUS_OVERRIDE` and `OPP_STATUS_OVERRIDE` to make testing stuff like this easier~~

---

### Videos of skills properly defrosting

Match Gotcha + Scorched Sands
<video src="https://github.com/pagefaultgames/pokerogue/assets/5421700/4dcf6184-5b0f-4f2c-a1fc-ef71e3264b1c" />

Scald + Steam Eruption
<video src="https://github.com/pagefaultgames/pokerogue/assets/5421700/e8ebe91a-1e41-4bb3-86e6-85ae1d98a2d6" />

Ember + Sizzly Slide
<video src="https://github.com/pagefaultgames/pokerogue/assets/5421700/2b1b592b-4179-44e8-b430-36e1eb20114b" />

Burn up + burn up failure
<video src="https://github.com/pagefaultgames/pokerogue/assets/5421700/660117f4-b9e8-438e-bdaa-c5c6546523b6" />


td76099:
Tested out locally - can confirm it works

LGTM

IceFire03 (approved):

bennybroseph (changes requested):
Looking to get JSDoc documentation as shown here: [Code Comment FAQ](https://discord.com/channels/1125469663833370665/1239957703820836975). Current comments look a little messy

OrangeRed:
> Looking to get JSDoc documentation as shown here: [Code Comment FAQ](https://discord.com/channels/1125469663833370665/1239957703820836975). Current comments look a little messy

@bennybroseph I don't quite understand, what changes are you looking for specifically? I only added comments in two places the if statement on `src/data/move.ts:386` and the class on `src/data/move.ts:1438`. Which of them would you like changed?

src/data/move.ts
```diff
@@ -1416,6 +1424,17 @@ export class RemoveHeldItemAttr extends MoveEffectAttr {
...
+ * @param ...effects - Status Effects to heal
+ * @example
+ * ```
+ * // A move that unthaws target
+ * const thawSelf = new HealStatusEffectAttr(false, StatusEffect.FREEZE)
+ * // Pokemon uses 'Refresh' on itself
+ * const cureAll = new HealStatusEffectAttr(true, StatusEffect.BURN, StatusEffect.PARALYSIS, StatusEffect.POISON)
+ * ```
+ */
 export class HealStatusEffectAttr extends MoveEffectAttr {
```
bennybroseph, line 1438: This class

src/data/move.ts
```diff
@@ -1430,7 +1449,7 @@ export class HealStatusEffectAttr extends MoveEffectAttr {
       return false;
 
     const pokemon = this.selfTarget ? user : target;
-    if (pokemon.status && this.effects.includes(pokemon.status.effect)) {
+    if (pokemon.hp > 0 && pokemon.status && this.effects.includes(pokemon.status.effect)) {
```
bennybroseph, line 1452: Why does hp need to be above 0 to heal a freeze?

bennybroseph (changes requested):

src/data/move.ts
OrangeRed: A pokemon that is going to faint from an attack doesn't need to be unthawed. Not having that check would trigger the unthaw logic / display the message and then the pokemon would immediately faint (and have its status changed to that).

OrangeRed: This also applies not to just freeze of course. A pokemon that is poisoned or burned doesn't need to display that it's been cured when the thing that happens immediately after is it fainting since fainting changes the pokemon's statusEffect.

bennybroseph: I'm not aware of any off hand, but what about abilities or effects that require the user to not be frozen in order to activate? I feel like the thaw should be happening upon getting hit by the move, not taking damage.

bennybroseph: Ah, yeah how does this handle if the pokemon should get burned by the fire move? It can't do so if the pokemon is still considered frozen

src/data/move.ts
OrangeRed: I still don't know exactly what you're looking for but here is a draft of the comments I am considering pushing. Let me know if there is still something I need to change.
```ts
/**
 * Move attribute that signals that that move should cure a status effect
 * @extends MoveEffectAttr
 * @see {@linkcode apply()}
 */
export class HealStatusEffectAttr extends MoveEffectAttr {
  /** Array of Status Effects to cure */
  private effects: StatusEffect[];
  
  /**
   * @param selfTarget - Whether this move targets the user
   * @param ...effects - Array of status effects to cure
   */
  constructor(selfTarget: boolean, ...effects: StatusEffect[]) {
    super(selfTarget);

    this.effects = effects;
  }
// ...
}
```

Also on a related note, should we be adding comments to every single `Class.apply()` inside of child classes? I feel like it would be a lot better doing it once for a given parent and then letting the comments propagate via inheritance.

src/data/move.ts
bennybroseph: Understood. My question rolls back to the original one. Why would a pokemon need to be at above 0 HP in order to be defrosted? I don't think this should be included based on the info provided.

OrangeRed: Again because in either scenario (whether the status is cured and then the pokemon faints or the status is not cured and the pokemon faints) the end result will be the same the pokemon's status will change to `FAINTED`.

So its just a matter of cutting out an unnecessary step in that in the first scenario the pokemon will go from status -> no status -> fainted, whereas in the second the pokemon will go straight from status -> fainted

src/data/move.ts
bennybroseph: Please take a look at the linked information [Code Comment FAQ](https://discord.com/channels/1125469663833370665/1239957703820836975). There are also examples within the code itself that I have already documented if you need.

src/data/move.ts
XGSleepWalker: > So its just a matter of cutting out an unnecessary step in that in the first scenario the pokemon will go from status -> no status -> fainted, whereas in the second the pokemon will go straight from status -> fainted

I don't agree with this part. For now it may be okay to make the jump assuming the pokemon will be fainted anyways, but we may cause issues down the line with new interactions (who knows what comes from Z-A). I am in support of reverting the hp check and applying the effect regardless.

bennybroseph: Agreed, let's leave each step intact. The move thaws the pokemon, so it should do so even if it just faints afterwards.

src/data/move.ts
OrangeRed: Is there something in the code block above that isn't what you're looking for? From my point of view it conforms to what is written in the FAQ. 

Also I believe writing JSDoc on the constructor is a nicety because when creating a new class the intellisense will show you the comments as you're writing the variable.

src/data/move.ts
OrangeRed: I removed the `pokemon.hp > 0` check so the effect will trigger before a pokemon faints

OrangeRed:
@bennybroseph I added comments to the `HealStatusEffectAttr` class that are hopefully easy enough to understand.

src/overrides.ts
bennybroseph, line 1: need to remove this commit

bennybroseph (changes requested):

src/overrides.ts
OrangeRed: Which commit? The one that adds status overrides? Isn't that a useful addition for testing moves / abilities that interact with statuses?

bennybroseph: if we're adding new overrides that should be in a separate PR. it's completely unrelated to this

OrangeRed: Sure that makes sense. I removed the status overrides from this PR and I'll make a new PR with them included.

OrangeRed: > if we're adding new overrides that should be in a separate PR. it's completely unrelated to this

I made a separate PR for this. https://github.com/pagefaultgames/pokerogue/pull/1602

bennybroseph (approved):

PR #944, opened 2024-05-16
Fixed the incorrect translation key for ‘Never-Melt Ice’

imfanhua:
Fixed the issue with the incorrect translation key for ‘Never-Melt Ice’.


Madmadness65:
This translation key is actually still incorrect, as it says "never_**nelt**_ice" instead of "melt".

bennybroseph (approved):

PR #949, opened 2024-05-16
Added german modifier-type.ts localization

CodeTappert:


bennybroseph:
Some of these are kinda long, did you check if they fit in game?

bennybroseph:
This also isn't compiling, you'll need to check your formatting

CodeTappert:
I will test it.

CodeTappert:
image

image

image

Changed also the text of the mints. Because they indeed were to long (certain ones)

Also changed text of PP Max

image

Changed IV Scanner (which was the longest i think):
image

Changed Healing Pin

src/locales/de/modifier-type.ts
```diff
@@ -259,7 +259,7 @@ export const modifierType: ModifierTypeTranslationEntries = {
     "miracle_seed": "Wundersaat",
     "magnet": "Magnet",
     "twisted_spoon": "Krümmlöffel",
-    "never-nelt_ice": "Ewiges Eis",
+    "never-melt_ice": "Ewiges Eis",
```
bennybroseph, line 262: don't modify this, you've re-broken it

bennybroseph (changes requested):

src/locales/de/modifier-type.ts
bennybroseph: `never_melt_ice`

CodeTappert: I changed it TO never-melt_ice in my commit because it was never-nelt_ice before which is wrong. (could have been my error).

In my newest commit i changed it to never_melt_ice like it is in english

bennybroseph (approved):

PR #955, opened 2024-05-16
Added german egg localization

CodeTappert:


CodeTappert:
Draft because i will to make sure it is shown correctly

CodeTappert:
This PR is ready. I made sure that every text is not to long, makes sense (for example the egg hatch text is directly from the offical games) and did not change "pull" on purpose since there isnt a german word with a length that fits there that makes sense.

bennybroseph (approved):

PR #960, opened 2024-05-16
Added German Translations for menu.ts

Gerafique:


src/locales/de/menu.ts
```diff
@@ -12,10 +12,10 @@ export const menu: SimpleTranslationEntries = {
     "loadGame": "Spiel laden",
     "newGame": "Neues Spiel",
     "selectGameMode": "Wähle einen Spielmodus",
-    "logInOrCreateAccount": "Logge dich ein oder erstelle einen Account zum starten. Keine Email nötig!",
+    "logInOrCreateAccount": "Melde dich an oder erstelle einen Account zum starten. Keine Email nötig!",
     "username": "Benutzername",
     "password": "Passwort",
-    "login": "Einloggen",
+    "login": "Anmelden",
```
JonStuddersAccess, line 18: I believe Anmelden is a much broader term and can mean multiple things, including register?

CodeTappert: No. Anmelden is the correct term here. It is always dependend on the context. Sure "Einloggen" is also correct but there isnt anything wrong with "anmelden"

Gerafique: "Einloggen" and "Anmelden" are both fine. The more common way to express a login in German is "Anmelden", feels more natural to me. Nevertheless its just a suggestion on my behalf. :)

Regarding register, in German also "Registrieren", it is used to express a "creation of an account".

CodeTappert (approved):

bennybroseph (approved):

PR #966, opened 2024-05-16
Added difference between the "x wants to fight" for double and single …

CodeTappert:
…battle since in some languages it needs to be (like german)


CodeTappert:
> …battle since in some languages it needs to be (like german)

It is the same for every language but german until they localise it (for now they will have all a english message on a double battle). But i wasnt confident enough to change it in languages i do not speak

bennybroseph (approved):

PR #971, opened 2024-05-16
Adjust all Existing Comments to Match

bennybroseph:
Adjusted all existing comments in move.ts to match. This should help when new Pull Requests are made as there are multiple examples to work from.

There are no functionality changes here, it's all comments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,9 @@ dist-ssr
 *.sln
 *.sw?
 
+# Docummentation
+docs/*
+
 public/images/trainer/convert/*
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -24,6 +24,9 @@ dist-ssr
 *.sln
 *.sw?
 
+# Docummentation
+docs/*
+
```
bennybroseph, line 29: at some point i'll be attempting to create HTML docs again. This just adds these to the gitignore ahead of time.

Flashfyre (approved):

PR #980, opened 2024-05-16
Set the name for the ace trainer female class

### LINKED ISSUE #974: Ace Trainer Female Localization does not work
The male version is shown instead
(@CodeTappert , allow me to poke you)

https://github.com/pagefaultgames/pokerogue/blob/ff086ca4926bb090ac54a043ec02faddfbce753a/src/locales/fr/trainers.ts#L18

CodeTappert:
Fixes #974

(But this is a fundemental problem and needs to be adressed by doing this for all classed that have "hasGender"

### DIFF
--- a/src/data/trainer-config.ts
+++ b/src/data/trainer-config.ts
@@ -596,7 +596,7 @@ function getRandomTeraModifiers(party: EnemyPokemon[], count: integer, types?: T
 
 export const trainerConfigs: TrainerConfigs = {
   [TrainerType.UNKNOWN]: new TrainerConfig(0).setHasGenders(),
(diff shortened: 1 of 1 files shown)


Madmadness65:
It's due to the Male and Female symbols not appearing next to the trainer classes anymore, moving it to localization seems to have broken it in some way.

CodeTappert:
> It's due to the Male and Female symbols not appearing next to the trainer classes anymore, moving it to localization seems to have broken it in some way.

Yes. We just need to write in the "setHasGenders" function the keys for the female version and it will be fixed. How i did with the ace trainer here

bennybroseph:
I feel like the larger issue should be resolved rather than fixing this specific case. Feel free to reopen this with additional changes or make a new request

PR #981, opened 2024-05-16
Type localization instead of TeraType

josericardo-fo:
Before:

image

After:

image

(I didn't translate the rest of the description to ptBR yet lol, but you get it)

Now the type name in the description is translated too. Changed TeraType to Type and where it was used since they are essentially the same.


bennybroseph:
there's a ton of changes to imports in this. Why?

bennybroseph:
Yeah, if we wanna update formatting we should do it in a separate commit. 

With that said though, these shouldn't be different than regular pokemon types. can we just update this so that it pulls from the regular pokemon types instead of creating a whole new section? i.e. delete the teratypes and update the function to grab types from our existing localization

josericardo-fo:
I did this because I didn't find any "type" localization. Where is it?

bennybroseph:
if not, let's put them in `pokemon-stat.ts` and rename the file to `pokemon-info.ts`

josericardo-fo:
Sure. Working on that

josericardo-fo:
Made the changes tou requested @bennybroseph

bennybroseph:
Nice, final thing can we get some screenshots showing this was tested?

josericardo-fo:
**Stats**
ptBR
image
en
image

**Types**
ptBR
image
en
image

bennybroseph (approved):

PR #982, opened 2024-05-16
 add berry, battle-stat zh-CN translate

Rei1mu:
add berry, battle-stat zh-CN translate
changed src/data/ :
berry.ts
battle-stat.ts


Adri1:
Don't forget to add **berry.ts** to the other existing languages, and also to edit **config.ts** and **battle.ts** with the new entries in every language

Rei1mu:
ok,i will try

PR #647, opened 2024-05-08
Updated game info with rewards and more info about game state

XGSleepWalker:
Updated game info with rewards in BattleScene
Updated game info with more info about the game state

PR #983, opened 2024-05-16
Exposed More Info to gameInfo

devingearing:
By exposing more info we can allow for third party extension to read this info to make overlays or whatever they want without needing them to read localStorage.

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -2000,13 +2000,15 @@ export default class BattleScene extends SceneBase {
 			gameMode: this.currentBattle ? this.gameMode.getName() : 'Title',
 			currentPhase: (this.getCurrentPhase())?.constructor.name || "Unknown",
 			biome: this.currentBattle ? getBiomeName(this.arena.biomeType) : '',
+			weather: this.arena.weather,
 			wave: this.currentBattle?.waveIndex || 0,
 			 party: this.party ? this.party.map(p => {
-				return { name: p.name, level: p.level, speciesId: p.species.speciesId, fusionId: p.fusionSpecies?.speciesId || -1, ability: p.abilityIndex, nature:p.nature, ivs: p.ivs, active: p.active, gender: p.gender, pokerus: p.pokerus,fieldPosition: p.fieldPosition};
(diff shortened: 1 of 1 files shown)


src/battle-scene.ts
```diff
@@ -1993,11 +1998,15 @@ export default class BattleScene extends SceneBase {
...
+			currentPhase: (this.getCurrentPhase())?.constructor.name || "Unknown",
 			biome: this.currentBattle ? getBiomeName(this.arena.biomeType) : '',
 			wave: this.currentBattle?.waveIndex || 0,
-			party: this.party ? this.party.map(p => {
-				return { name: p.name, level: p.level };
-			}) : []
+			 party: this.party ? this.party.map(p => {
+				return { name: p.name, level: p.level, speciesId: p.species.speciesId, fusionId: p.fusionSpecies?.speciesId || -1, ability: p.abilityIndex, nature:p.nature, ivs: p.ivs, active: p.active, gender: p.gender, pokerus: p.pokerus,fieldPosition: p.fieldPosition};
+			}) : [],
+			enemyParty: this.getEnemyParty() ? this.getEnemyParty().map(p => {
```
XGSleepWalker, line 2007: https://github.com/pagefaultgames/pokerogue/pull/647/files#diff-4b1a5f9160d1d8aecce0db85774f20e3c254f97c27cc0d0473911d5cbd565396 - Check this as a reference, but the main point is that gameInfo shouldn't expose more info than what the player can see. Exposing enemy IVs goes against the purpose of IV Scanner, and likewise for ability.

devingearing: That is a valid point. If I removed the ivs and nature would this look alright then?

devingearing: Here is video proof


Uploading 2024-05-16 20-17-51 (1).mp4…

Admiral-Billy (approved):
I tested locally and everything seems to work as you'd expect (at least from an output standpoint).

Here's a summary on the discord: https://discord.com/channels/1125469663833370665/1240785598214705192/1240863471483097149

src/battle-scene.ts
```diff
@@ -1533,7 +1533,15 @@ export default class BattleScene extends SceneBase {
...
-
+		if(this.getCurrentPhase()?.constructor.name === 'SwitchSummonPhase'){
+			this.updateGameInfo();
+		}
+		if(this.getCurrentPhase()?.constructor.name === 'SummonPhase'){
+			this.updateGameInfo();
+		}
+		if(this.getCurrentPhase()?.constructor.name === 'TurnInitPhase'){
+			this.updateGameInfo();
+		}
```
Xavion3, line 1544: This could just be an includes, makes it easier to add more later.

flx-sta:
Considered stale.

PR #984, opened 2024-05-16
Fix Entry Hazard Miss

bennybroseph:
Made Entry Hazards bypass the accuracy check as they cannot miss a target. There's still one more bug to fix with this but it requires way more code changes. This change needed to be done anyway, so I'm getting it out first.

https://github.com/pagefaultgames/pokerogue/assets/13838608/96c02cf7-fd6f-4066-9a45-1397469d9d8d

## These are really just sanity checks, nothing in the code affected this anyway

https://github.com/pagefaultgames/pokerogue/assets/13838608/b4109773-d443-4def-b743-0e8af4aa4f69

https://github.com/pagefaultgames/pokerogue/assets/13838608/e441dc1b-4a24-4023-8c36-2f29b5bf1358

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -55,6 +55,7 @@ export enum MoveTarget {
   /** {@link https://bulbapedia.bulbagarden.net/wiki/Category:Moves_that_target_all_Pok%C3%A9mon Moves that target all Pokemon} */
   ALL,
   USER_SIDE,
+  /** {@link https://bulbapedia.bulbagarden.net/wiki/Category:Entry_hazard-creating_moves Entry hazard-creating moves} */
   ENEMY_SIDE,
   BOTH_SIDES,
   PARTY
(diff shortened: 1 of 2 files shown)


TempsRay (approved):
Mind documenting the one last bug that needs to be fixed as just a comment in the code or PR?

bennybroseph:
Yeah, when there are no targets Entry Hazards show a "But it failed!" message which is not correct. Since the move doesn't need a target, it should still succeed. Fixing this requires some more changes though because everything is so tightly connected to having a target, and I'd need to allow moves that aren't actually targeting something to bypass that system. It's an edge case though tbh so it shouldn't be thought of as high priority. Most people won't encounter it or won't even know it's a problem.

XGSleepWalker (approved):

Greenlamp2 (approved):

molker (approved):
lgtm

PR #994, opened 2024-05-16
Check available pkmn before asking to switch, removed cancelled move text

0zuzu:
QoL changes:
- Game now checks if player has available Pokemon before asking to switch at the start of battles.
- Allows player to switch on the first wave (so that the order when building a team doesn't matter)
- Move text won't be displayed if the move was cancelled. (stops players from mistakenly thinking PP was used and removes a text that's unnecessary)

A move is cancelled if there are no available targets and no PP is used, such as damaging moves in a double battle where both foes already fainted from the previous move.

Before: (asks to switch, Charcadet uses ember but move is cancelled) 

https://github.com/pagefaultgames/pokerogue/assets/63069680/ffefac24-fa1e-40e5-bcf8-b540355d64fa

After:

https://github.com/pagefaultgames/pokerogue/assets/63069680/0c5ab3d4-ff6b-4e69-b800-39bc0d679d97

Update: Now multi turn moves like Fly, Outrage and Electro Shot will display failure messages when outside of first use:

https://github.com/pagefaultgames/pokerogue/assets/63069680/cddffe95-35eb-4707-a3d2-8f14baa69c0f


src/phases.ts
```diff
@@ -2316,12 +2322,8 @@ export class MovePhase extends BattlePhase {
...
 
-      if (this.move.moveId)
-        this.showMoveText();
-
       // This should only happen when there are no valid targets left on the field
       if ((moveQueue.length && moveQueue[0].move === Moves.NONE) || !targets.length) {        
-        this.showFailedText();
         this.cancel();
         
         // Record a failed move so Abilities like Truant don't trigger next turn and soft-lock
```
DustinLin, line 2329: maybe move this comment to a better place?

0zuzu: Are you referring to the comment I added? "Display move text as long as it wasn't cancelled"? I can remove it if it's not necessary

bennybroseph:
Not sure about this one. The failed text is there for when there are no targets on the field which is not just when you finish a wild double battle. This also happens when you knock out two pokemon in a trainer battle. Removing the text will be very strange in this case. Also, what about self destruct? This needs more testing for sure

0zuzu:
> Not sure about this one. The failed text is there for when there are no targets on the field which is not just when you finish a wild double battle. This also happens when you knock out two pokemon in a trainer battle. Removing the text will be very strange in this case. Also, what about self destruct? This needs more testing for sure

About double trainer battles, I don't see why it would be strange to remove that text compared to wild battles.
The main reasons for hiding a cancelled move is so that players don't think PP was wasted, and to streamline combat text in grindy wild double battles. 

If the community wants to keep the trainer battles more faithful to the Pokemon games than I could add an exception for trainer battles, but personally I don't find it strange at all.

Self destruct still goes through if you have an ally, the move logic didn't change at all.

bennybroseph:
> The main reasons for hiding a cancelled move is so that players don't think PP was wasted, and to streamline combat text in grindy wild double battles.

What about for Outrage or Fly/Dive/Dig/Bounce/Phantom Force? What about charge moves like Electro Shot?

0zuzu:
> What about for Outrage or Fly/Dive/Dig/Bounce/Phantom Force? What about charge moves like Electro Shot?

Thank you for bringing me attention to those edge cases. I updated the commit so that multi turn moves (when they ignore PP usage) will still display a failure message, I included a third video to showcase how it works.

IceFire03 (approved):

knispeja:
> The main reasons for hiding a cancelled move is so that players don't think PP was wasted, and to streamline combat text in grindy wild double battles.

I think this is a good thought -- but I'm personally in favor of leaving the message in and making it clearer that the move wasn't actually used. Removing the message completely might lead players to believe that the battle ends when all Pokemon on one side are fainted, when it actually plays out the remainder of the turn.

bennybroseph:
I agree, and on top of this I already have plans on fixing moves that should continue even when there are no targets left. This will help to better distinguish these moves. Currently I'm talking about entry hazards/weather/terrain moves as these should be able to be used at the end of a battle or between turns after defeating all enemies in a double battle.

PR #1008, opened 2024-05-17
 update zh-CN trans: berry, select-ui,Transfer , battle-stat

Rei1mu:
some can edit the words order for different Languages.
add files:
locales:
config.ts
berry -> berry.ts
select-ui  -> modifier-select-ui-handler.ts
battle-stat.ts -> battle.ts


bennybroseph:
Berrys have been added already and will need to be removed in order to resolve the conflicts

PR #1010, opened 2024-05-17
Fix Trainer Localization Typos

bennybroseph:
Took a look at some of the entries for trainers and adjusted accordingly. it looks like all of these trainer types were grabbed from an online source as a sort of future-proofing which is good, but resulted in some errors. Some of which have already been fixed, and some which are included here.

Also prevents an issue if there is a missing image and displays a warning in the console instead of a crash.


src/locales/zh_CN/trainers.ts
```diff
@@ -80,8 +80,12 @@ export const trainerClasses: SimpleTranslationEntries = {
     "pokémon_rangers": "宝可梦巡护员组合",
     "ranger": "巡护员",
     "restaurant_staff": "服务生组合",
+    "rich": "Rich",
+    "rich_female": "Rich",
     "rich_boy": "富家少爷",
     "rich_couple": "富豪夫妇",
+    "rich_kid": "Rich Kid",
+    "rich_kid_female": "Rich Kid",
```
TempsRay, line 88: Are rich, rich_female, rich_kid, and rich_kid_female used anywhere? I see there are existing tl's for rich_boy, rich_couple, and lady

bennybroseph: yes, it's an enum

TempsRay (approved):

Madmadness65 (approved):

PR #1011, opened 2024-05-17
Transfer items from released Pokémon to the one that replaces it

knispeja:
First pass at transferring items from a forcibly released Pokémon to the one that replaces it, as described on the feature board. A message indicating the items were transferred would probably improve the experience, but I wasn't that familiar with the UI code.

**Before**:
https://github.com/pagefaultgames/pokerogue/assets/9343160/edf77a71-e140-45af-aeb5-40601e3a6c81

**After**:
https://github.com/pagefaultgames/pokerogue/assets/9343160/8e6560a5-e5e6-4ee0-8b68-8c9030aeea51


j-diefenbach:
How does this interact with stack limits? I understand you're using the try transfer items function but have you checked if that gets interrupted e.g. there is an item already at stack limit on one or both pokemon.

knispeja:
> How does this interact with stack limits? I understand you're using the try transfer items function but have you checked if that gets interrupted e.g. there is an item already at stack limit on one or both pokemon.

Ah that's a good point. The `tryTransferHeldItemModifier()` function does appear to handle that at a glance but I didn't think to test it, I can try it out tomorrow

knispeja:
> > How does this interact with stack limits? I understand you're using the try transfer items function but have you checked if that gets interrupted e.g. there is an item already at stack limit on one or both pokemon.
> 
> Ah that's a good point. The `tryTransferHeldItemModifier()` function does appear to handle that at a glance but I didn't think to test it, I can try it out tomorrow

This appears to work as expected, tried catching a Pokemon with a few different values of Sitrus Berries and replacing a Pokemon with max stacks and the new Pokemon always had max stacks afterwards.

https://github.com/pagefaultgames/pokerogue/assets/9343160/62fbfb36-4014-4b3d-9119-fe99e4da407e

knispeja:
I added a new message to indicate the items were transferred after all:
https://github.com/pagefaultgames/pokerogue/assets/9343160/0aec63b9-c9e3-4ac0-8ab6-e69734813ce6

src/phases.ts
```diff
@@ -4323,12 +4325,24 @@ export class AttemptCapturePhase extends PokemonPhase {
...
-                      addToParty();
-                    else
+                    if (releasedPokemonSlotIndex < 6) {
+                      if (releasedItems != null && releasedItems.length > 0) {
+                        this.scene.ui.showText(`You transferred all items held by ${releasedPokemon.name} to ${pokemon.name}.`, null, () => {
+                          addToParty(releasedItems);
+                        }, 0, true);
+                      }
+                      else {
+                        this.scene.ui.clearText();
```
knispeja, line 4339: I cleared the text here to address a minor issue where the release text remained onscreen.

Before:
https://github.com/pagefaultgames/pokerogue/assets/9343160/816e691d-3068-439c-a3bc-707eb16a334b

After:
https://github.com/pagefaultgames/pokerogue/assets/9343160/cd9c1f67-bb93-4db4-b348-68f75579811c

src/ui/party-ui-handler.ts
```diff
@@ -782,20 +780,26 @@ export default class PartyUiHandler extends MessageUiHandler {
 
   doRelease(slotIndex: integer): void {
     this.showText(this.getReleaseMessage(this.scene.getParty()[slotIndex].name), null, () => {
+
+      if (this.partyUiMode === PartyUiMode.RELEASE) {
+        const selectCallback = this.selectCallback;
+        this.selectCallback = null;
+        selectCallback(this.cursor, PartyOption.RELEASE);
```
knispeja, line 787: It's necessary to call this callback earlier to allow `phases.ts` to gather the held items before the object is destroyed

flx-sta:
Considered stale.

knispeja:
Any chance we can open this as draft @flx-sta? I can do the work to port these changes to the latest beta code, `attempt-capture-phase.ts` etc.

flx-sta:
> Any chance we can open this as draft @flx-sta? I can do the work to port these changes to the latest beta code, `attempt-capture-phase.ts` etc.

I can only re-open it.
You'd have to ask a project admin to put it into "Draft" as I no longer actively work on the project/repo.

fabske0:
First of all, it's nice to see you come back to this PR, @knispeja
With the next version (1.12.0), there will be a modifier/item rework, so you might want to check that out: [Modifier rework branch](https://github.com/pagefaultgames/pokerogue/tree/modifier-rework). It would also be great, if you could move the text to the [Locales repo](https://github.com/pagefaultgames/pokerogue-locales/) and use i18next so we can translate the text

PR #1015, opened 2024-05-17
Implement Friend Guard

JakubHanko:
In this draft pull request, I implement Friend Guard. It functions as follows:

1. At the start of the move, apply tags to all eligible Pokémon (your ally) for 1 turn.
2. Before damage calculation, apply the damage reduction to all tagged Pokémon.
3. Reapply the tag at the start of the next turn if the source of Friend Guard is still alive.

This implementation could be generalized to support any aura effect that applies to multiple Pokémon as long as the source is active.

I attempted to use Arena tags but encountered a limitation: they only support one tag per side without modifications.

Currently, this is a proof of concept with several areas needing further work. For example, abilities like Mold Breaker are not yet accounted for, and TSDocs are missing.


JakubHanko:
Adding two videos: one with Mold Breaker and one without.

https://github.com/pagefaultgames/pokerogue/assets/60473007/1f16f5f7-b47e-44b1-a936-7a7b40256888


https://github.com/pagefaultgames/pokerogue/assets/60473007/fa18c66e-96fa-4e25-8189-f4966c4c65e4

JakubHanko:
TSDocs will follow shortly. I don't know how to make the PR draft again.

flx-sta:
Considered stale.

PR #1016, opened 2024-05-17
Updated Logic when Animating Trainer Sprites

bennybroseph:
Updated the check in place when attempting to play a trainer sprite. Separated it into a separate function to reduce reusing code and to make it easier to exit early when failures occur.


XGSleepWalker (approved):

TempsRay (approved):
Should hopefully fix a bug where trainer animations were stuck in the first frame

Madmadness65 (approved):

PR #1025, opened 2024-05-17
Fix daily run freeze

Admiral-Billy:
For whatever reason, this code was changed as a part of the endings commit, and without it, it seems to cause daily runs to freeze at the end.


bennybroseph (approved):
Worked for me in both modes. May have been an oversight or a change made early in the commit and forgotten about.

https://github.com/pagefaultgames/pokerogue/assets/13838608/29817677-1975-4196-8e7c-f8be446c9232

https://github.com/pagefaultgames/pokerogue/assets/13838608/4249748d-0088-4be8-84b2-a61d5a313284

PR #998, opened 2024-05-16
Fix ingrain for Phantump and Trevenant (#856)

charliegiang:
Fixes issue https://github.com/pagefaultgames/pokerogue/issues/856

Before fix:
https://github.com/pagefaultgames/pokerogue/assets/146129103/f4dccd4b-1cce-4466-918f-ff30fd1d3fcb

After fix: 
https://github.com/pagefaultgames/pokerogue/assets/146129103/36c75dd2-6039-4cbb-be57-3f75f4d6be6d

PR #1030, opened 2024-05-17
Move Ingrain Fix For All Ghosts

Brandon-Rodgers:
### Issue
Ingrain can only be used by Trevenant and Phantump because they are hardcoded. This move should be useable by all Ghost types.
Since we have Splicing, technically any pokemon can learn any move.

### What this fix does
Sets Class TrappedTag, method canAdd back to the original of checking if pokemon is not trapped and not ghost instead of having an override for two specific pokemon
Adds method override canAdd to the IngrainTag and it just checks if the pokemon is already trapped since it should be applicable to any pokemon regardless of type.

### What this fix does not do
This does not trap ghost types when they use ingrain. 
PR #998 already added a check to switching pokemon if the trapped tag is applied by move Ingrain and the pokemon switching is ghost.

### DIFF
--- a/src/data/battler-tags.ts
+++ b/src/data/battler-tags.ts
@@ -15,7 +15,6 @@ import { TerrainType } from "./terrain";
 import { WeatherType } from "./weather";
 import { BattleStat } from "./battle-stat";
 import { allAbilities } from "./ability";
-import { Species } from "./enums/species";
 
 export enum BattlerTagLapseType {
   FAINT,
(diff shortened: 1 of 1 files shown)


Brandon-Rodgers:
Before Fix Trevenant can use Ingrain as normal.

https://github.com/pagefaultgames/pokerogue/assets/9848614/8c01ccef-e5a4-48b4-b836-a2655c76373b

Before Fix Aegislash cannot use Ingrain. From the video you can see they do not get the message of planting their roots

https://github.com/pagefaultgames/pokerogue/assets/9848614/dacf5ffc-4b60-433b-ac43-a0028a2b053a

After Fix Trevenant can still use Ingrain as normal.

https://github.com/pagefaultgames/pokerogue/assets/9848614/65417d23-7936-462d-be16-900337564025

After Fix Aegislash can now use Ingrain

https://github.com/pagefaultgames/pokerogue/assets/9848614/6fcbb09c-e547-41f6-92cc-5128330ab729

bennybroseph (approved):

PR #1033, opened 2024-05-17
update battle-stat,statLevelChangDescription for translate to edit

Rei1mu:
update battle-stat for zh-CN. en. 
battle-stat,and battleStatLevelChangeDescription can be edit for translate at 'src/locales/xx/battle-stat.ts'


src/phases.ts
```diff
@@ -2927,7 +2927,7 @@ export class StatChangePhase extends PokemonPhase {
           : `${relLevelStats.slice(0, -1).map(s => getBattleStatName(s)).join(', ')}${relLevelStats.length > 2 ? ',' : ''} and ${getBattleStatName(relLevelStats[relLevelStats.length - 1])}`;
       } else
         statsFragment = getBattleStatName(relLevelStats[0]);
-      messages.push(getPokemonMessage(this.getPokemon(), `'s ${statsFragment} ${getBattleStatLevelChangeDescription(Math.abs(parseInt(rl)), levels >= 1)}!`));
+        messages.push(getPokemonMessage(this.getPokemon(), i18next.t('battleStat:levelChangeDescripion', { statsFragment: i18next.t(`battleStat:${statsFragment.replace(/\./g, '').replace(/[ ]/g, '_')}`), description: i18next.t(`battleStat:${getBattleStatLevelChangeDescription(Math.abs(parseInt(rl)), levels >= 1).replace(/\./g, '').replace(/[ ]/g, '_')}`) })));
```
bennybroseph, line 2930: what do the `.replace()` calls do here?

bennybroseph (approved):

bennybroseph (changes requested):
what do the .replace() calls do here?

src/phases.ts
bennybroseph, line 2930: ones like this for reference

bennybroseph (changes requested):

Adri1:
Isn't the "!" fragment missing or I misslooked?

Rei1mu:
> Isn't the "!" fragment missing or I misslooked?
> 
> EDIT : It is possible to give to gendered languages the possibility to choose a determiner? Because in French for example before the stat name, you have to put a mandatory determiner like "the" but is either masculine/feminine/genderless Example :
> 
> * "[the (feminine)] Defense"
> * "[the (genderless, without any space between the derterminer and the word)] Attack"
> * "[the (masculine)] critical-hit ratio"

"!" maybe i  missed , but it can edit on 
 ```
"levelChangeDescripion": "'s {{statsFragment}} {{description}}", 
 "levelChangeDescripion": "'s {{statsFragment}} {{description}}!",
```
about [the] 's problem, this is just part of that . 
i think it means will change the 'the'  at getPokemonMessage() ?

```
export function getPokemonMessage(pokemon: Pokemon, content: string): string {
  return `${getPokemonPrefix(pokemon)}${pokemon.name}${content}`;
}
export function getPokemonMessage(pokemon: Pokemon, content: string): string {
  return `The ${getPokemonPrefix(pokemon)}${pokemon.name}${content}`;
}
```
like this ?  sorry i'm not good at languages

Rei1mu:
> what do the .replace() calls do here?

It's update. please check

bennybroseph:
Please address the above concerns

Tempo-anon:
Closing this PR as it's a bit out of date. If this is still required please submit a new PR

PR #1035, opened 2024-05-17
[Implementation] (Power | Guard) Swap moves

SoulsHawk:
- Reworked the SwapStatsAttr used in the Heart Swap moves to implement Power and Guard Swap
~~- Added a SpeedSwapAttr for Speed Swap~~

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -1960,9 +1960,9 @@ export class ResetStatsAttr extends MoveEffectAttr {
 }
 
 /**
- * Attribute used for moves which swap the user and the target's stat changes.
+ * Attribute used for moves which swap the user and the target's battle stat changes.
  */
-export class SwapStatsAttr extends MoveEffectAttr
+export class SwapBattleStatsAttr extends MoveEffectAttr
 {
   public statsToSwap: BattleStat[];
 
@@ -1972,7 +1972,7 @@ export class SwapStatsAttr extends MoveEffectAttr
     this.statsToSwap = statToSwap;
   }
   /**
-   * Swaps the user and the target's stat changes.
+   * Swaps the user and the target's battle stat changes.
    * @param user Pokemon that used the move
    * @param target The target of the move
    * @param move Move with this attribute
@@ -1988,7 +1988,7 @@ export class SwapStatsAttr extends MoveEffectAttr
     const targetBattleStat = target.summonData.battleStats;
 
     this.statsToSwap.forEach(stat => {
-      let userStat = userBattleStat[stat];
+      const userStat = userBattleStat[stat];
       userBattleStat[stat] = targetBattleStat[stat];
       targetBattleStat[stat] = userStat;
     });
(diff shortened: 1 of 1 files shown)


src/data/move.ts
```diff
@@ -1964,30 +1964,73 @@ export class ResetStatsAttr extends MoveEffectAttr {
...
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any []): boolean
+  {
+    if (!super.apply(user, target, move, args))
+      return false;
+
+    const userBattleStat   = user.summonData.battleStats;
+    const targetBattleStat = target.summonData.battleStats;
+
+    this.statsToSwap.forEach(stat => {
+      let userStat = userBattleStat[stat];
```
styrix560, line 1991: Is there a reason why this is not const?

src/data/move.ts
```diff
@@ -1964,30 +1964,73 @@ export class ResetStatsAttr extends MoveEffectAttr {
...
+
+    user.updateInfo();
+    target.updateInfo();
+
+    if (this.statsToSwap.length === 7) {
+      target.scene.queueMessage(getPokemonMessage(user, ' switched stat changes with the target!'));
     }
+    else {
+      const statsName = this.statsToSwap.map(stat => getBattleStatName(stat));
+      target.scene.queueMessage(getPokemonMessage(user, ' switched all changes to its ' + statsName.join(' and ')  + ' with its target!'));
```
styrix560, line 2004: Here, I would say 'switched its [stats which got changed here] with the target'. Also I would not join them by ' and ' but instead by ', ' and only the last one by ' and '.

So something like: 'switched its Attack, Defense and Speed with the target'

src/data/move.ts
```diff
@@ -1964,30 +1964,73 @@ export class ResetStatsAttr extends MoveEffectAttr {
...
+    });
+
+    user.updateInfo();
+    target.updateInfo();
+
+    if (this.statsToSwap.length === 7) {
+      target.scene.queueMessage(getPokemonMessage(user, ' switched stat changes with the target!'));
     }
+    else {
+      const statsName = this.statsToSwap.map(stat => getBattleStatName(stat));
```
styrix560, line 2003: This variable should probably be in plural

src/data/move.ts
```diff
@@ -1964,30 +1964,73 @@ export class ResetStatsAttr extends MoveEffectAttr {
...
+      target.scene.queueMessage(getPokemonMessage(user, ' switched all changes to its ' + statsName.join(' and ')  + ' with its target!'));
+    }
+
+    return true;
+  }
+}
+
+export class SpeedSwapAttr extends MoveEffectAttr
```
styrix560, line 2011: Why is the swapping of Speed extra?

src/data/move.ts
```diff
@@ -1964,30 +1964,73 @@ export class ResetStatsAttr extends MoveEffectAttr {
...
+{
+  /**
+   * Swaps the user and the target's stat changes.
```
styrix560, line 2014: In this case the stat changes of only the speed

styrix560 (changes requested):
General logic looks good, only small things

src/data/move.ts
SoulsHawk: Because my brain thought that the value will change so let but yeah, in a loop so doesn't matter

src/data/move.ts
SoulsHawk: Other swap battle stats, here it changes the pokemon stat

src/data/move.ts
SoulsHawk: There is no move that switch more than 2 battle stats but why not

src/data/move.ts
SoulsHawk: I should remain the first one like "SwapBattleStatsAttr" but did not because it was "legacy"

styrix560: What is the difference between battle stat and pokemon stat?

Either way if there is a good reason for it, it should probably be written in the docs comment.

src/data/move.ts
SoulsHawk: And the "...all changes..." is something that I saw in an official game

src/data/move.ts
SoulsHawk: Battle stats are your "boost" that you can see when pressing 'c'
Pokemon stats are in your pokemon summary
We talking about my attrs docs right ?

src/data/move.ts
```diff
@@ -1964,30 +1964,73 @@ export class ResetStatsAttr extends MoveEffectAttr {
...
+   * @param user Pokemon that used the move
+   * @param target The target of the move
+   * @param move Move with this attribute
+   * @param args N/A
+   * @returns true if the function succeeds
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any []): boolean
```
XGSleepWalker, line 2021: This implementation is going to cause some unforeseen problems that we currently don't have a nice way to tackle. 
For example, this would change Escape in an unintended way: https://bulbapedia.bulbagarden.net/wiki/Escape

My advice is to only include Power and Guard Swap on this PR and leave Speed Swap out for when there's a better plan on how to tackle this.

styrix560 (commented):
code lgtm
I think the commit name should be more descriptive.

Also, I am not sure, whether the speed changes really cause undefined behaviour. However, if we are not sure we can just as well open another issue and leave this one as is.

src/data/move.ts
XGSleepWalker: There's three types of stats if you will. 

- Your raw base stats
- Your raw variable stats (affected by things like Speed Swap or Power Split)
- Your boosted stats (for things like Agility and Dragon Dance)

That code was changing the raw base stats, when it should be changing the raw variable stats (that should live under summonData). Unfortunately the implementation around summonData.stats isn't correct at the moment, so we'll have to work on that before working on this.

flx-sta:
Considered stale.

PR #986, opened 2024-05-16
Spectral Thief Partial Implementation + Marking Electro Shot as Partial (Pre-Damage Stat Change Phase Needed for Both Moves)

SunshineFuture753:
Implements Spectral Thief as much as possible; all that is left is recalculating damage after stats are stolen. This will require examining the stat change phases, figuring out how to make them occur before damage is calculated if necessary (which is the case for Spectral Thief). This PR relies on https://github.com/pagefaultgames/pokerogue/pull/1038 to fix PRE_APPLY, which Spectral Thief utilizes. Electro Shot is also now marked as Partial, given how it has the same issue of not recalculating stats after a mid-turn boost, specifically when used in Rain.

PR #1038, opened 2024-05-17
Fix MoveEffectTrigger.PRE_APPLY

Brandon-Rodgers:
### Issue
Any move class that uses MoveEffectTrigger.PRE_APPLY still has that attribute applied AFTER the move is completed.

### Current Implementation
The move is applied to the target if it is a Physical or Special move.
Once the hit has calculated the damage and returned if it hit to phases.ts, then all move attributes are attempted to be triggered
Due to this implementation Brick Break, Raging Bull, and Psychic Fangs do not Remove Screens like Light Screen, Reflect, and Aurora Veil

### What this fix does
It moves the Triggering of move effects to the Pokemon.apply method.
This means after the apply method calculates if the move will hit or not, then we can trigger all PRE_APPLY move effects once the move has been calculated to hit.
This now allows the above moves to Remove Screens before calculating the damage in the apply method.

### What this fix does not do
Tested with Spectral Thief Implementation from #986 and found it does not fix the bug with the stats not being boosted first.
With the current implementation it queues an unshiftPhase for StatChangePhase and this phase seems to trigger after the move phase is over instead of immediately raising the stats.


Brandon-Rodgers:
Base Damage Values

https://github.com/pagefaultgames/pokerogue/assets/9848614/8925309b-887c-4c7e-b1b9-8076ec419765

Before Fix is Implemented

https://github.com/pagefaultgames/pokerogue/assets/9848614/d70d763a-0b08-42fa-a8bf-c3836227a6e6

Breaking Moves Vs. Reflect

https://github.com/pagefaultgames/pokerogue/assets/9848614/2af986bd-2c50-4889-aeee-e24ed3283b50

Breaking Moves Vs. Light Screen

https://github.com/pagefaultgames/pokerogue/assets/9848614/1b009a6b-0c91-4cdb-a686-621159ad4e59

Spectral Thief Test with Fix - You can see compared to base values video that the damage is roughly the same with +6 attack because it is stolen before the attack, but not actually applied until after the attack

https://github.com/pagefaultgames/pokerogue/assets/9848614/5cd9908c-86e2-4947-95ae-fadbfa3f989f

Brandon-Rodgers:
Spectral Thief with this implementation does not steal stat boosts if it has no effect

https://github.com/pagefaultgames/pokerogue/assets/9848614/d704af1a-c582-4a4c-a307-6dafbf78d2c1

Brandon-Rodgers:
Testing Spectral Thief from PR #986  against normal types with new isTypeImmune change and it works as expected

https://github.com/pagefaultgames/pokerogue/assets/9848614/09795bea-e214-408d-a02c-d73830af881e

Testing Brick Break against a Ghost Type and it does not break the screen since they are immune

https://github.com/pagefaultgames/pokerogue/assets/9848614/f3dba171-5eb9-481c-9716-a37f33ce8e6e

bennybroseph (changes requested):
would like to get some basic comments in and surrounding this change

jaimefd:
Hello, I accidentally started working on Spectral Thief without seeing https://github.com/pagefaultgames/pokerogue/pull/986, and while doing so learned that you can do user.scene.overridePhase instead of user.scene.unshiftPhase for the StatChangePhase. However that won't be enough, you also need to change the logic in StatChangePhase so that the stat changes are applied without having to wait for their animation to show. Not sure how relevant this is for this PR but I hope it can help!

Tempo-anon:
Closing due to inactivity and merge conflicts

PR #1046, opened 2024-05-17
Make the settings display the current set language for the Language Setting

karl-police:
image

This is what the change looks like. If I set my game to German, it shows it as German in German.

**Currently**, it just shows English. And to be honest, while I was exploring the source code for the best location to change it... well, it's a bit of a unorganized setup for this specific setting.

I implemented this inside the ``setup()``, since the game reloads anyways, making it ideal to add this in. For some reason I decided to put ``i18next.language || "en"`` incase someone manually changes ``i18next.language`` and messes up, but whatever, to a regular user, this would never happen.

Then it converts ``"en"`` or whatever to the set language, hopefully.

But then I ended up changing this with **try**, **catch**, because when I tested languages such as ``zH_CN``, these somehow have like something after it, and whatever ``Init.DisplayNames`` does, has it different.

But since there are some other issues, with that in pokemon species...

image

And now it will just show the current set language, except for one. Because there's more languages that have that issue where the locale IDs aren't the same. I don't know if i18next has a implementation of DisplayNames as well.

**If it errors**, it will just use "en" and show "English".

### DIFF
--- a/src/ui/settings-ui-handler.ts
+++ b/src/ui/settings-ui-handler.ts
@@ -6,6 +6,7 @@ import { Mode } from "./ui";
 import UiHandler from "./ui-handler";
 import { addWindow } from "./ui-theme";
 import {Button} from "../enums/buttons";
+import i18next from "i18next";
 
 export default class SettingsUiHandler extends UiHandler {
   private settingsContainer: Phaser.GameObjects.Container;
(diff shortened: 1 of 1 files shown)


karl-police:
also how can I give myself stuff without overrides on local testing

too lazy to convert local storage base64 to table and then put table somewhere to change properties

only way is to set breakpoint to wherever ``this`` is being exposed

Adri1:
Instead of this, why not just having translations of the Settings, in which each language puts its name in place "English"?

karl-police:
> Instead of this, why not just having translations of the Settings menu, in which each language puts its name in place "English"?

"English" is defined in some place where there's barely any import of the other modules.

karl-police:
The issue with the translation files are. If I create one, they want us to put them in all the other locales folder.

idk if i18 can be told to just use the English one if an entry doesn't exist

Adri1:
You dont have to sepcify every languages in every locale. Just the "English" text being replaced the name of the language the translation file is in.
The names of the langages in the language selection menu is are different entries and don't need to be part of any localization file

bennybroseph:
Closing due to inactivity

PR #1055, opened 2024-05-17
So currently if you meet someone with a title it only shows there nam…

CodeTappert:
…e and only then the title (not like trainer classes). So i wanted to change that

Trainerclasses already do it this way. 

Currently (main):
image

With this PR:
image

### DIFF
--- a/src/data/trainer-config.ts
+++ b/src/data/trainer-config.ts
@@ -417,8 +417,7 @@ export class TrainerConfig {
         // Handle name by checking this.name - making it lowercase and replacing spaces with underscores and then calling i18next.t with the name
         const nameForCall = this.name.toLowerCase().replace(/\s/g, '_');
         this.name = i18next.t(`trainerNames:${nameForCall}`);
-        this.name = `${i18next.t("titles:gym_leader")} ${this.name}`;
-        this.title = "";
+        this.setTitle('gym_leader')
         this.setMoneyMultiplier(2.5);
         this.setBoss();
         this.setStaticParty();
@@ -450,8 +449,7 @@ export class TrainerConfig {
         // Handle name by checking this.name - making it lowercase and replacing spaces with underscores and then calling i18next.t with the name
         const nameForCall = this.name.toLowerCase().replace(/\s/g, '_');
         this.name = i18next.t(`trainerNames:${nameForCall}`);
-        this.name = `${i18next.t("titles:elite_four")} ${this.name}`;
-        this.title = "";
+        this.setTitle('elite_four');
         this.setMoneyMultiplier(3.25);
         this.setBoss();
         this.setStaticParty();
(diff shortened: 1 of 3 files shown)


src/data/trainer-config.ts
```diff
@@ -417,10 +417,8 @@ export class TrainerConfig {
         // Handle name by checking this.name - making it lowercase and replacing spaces with underscores and then calling i18next.t with the name
         const nameForCall = this.name.toLowerCase().replace(/\s/g, '_');
         this.name = i18next.t(`trainerNames:${nameForCall}`);
-
-
-
-        this.setTitle("gym_leader");
+        this.name = `${i18next.t("titles:gym_leader")} ${this.name}`;
+        this.title = "";
```
bennybroseph, line 421: I'd also like it to display like this, but what does title do? I'm not sure that I like it being set to a blank string.

bennybroseph (changes requested):

src/data/trainer-config.ts
CodeTappert: As far as my IDE tells me. 

image

image


So it is really only used for this? 

Which doesnt make sense. It needs to be read somewhere

And this is basically done here:

```
getName(trainerSlot: TrainerSlot = TrainerSlot.NONE, includeTitle: boolean = false): string {
    let name = this.config.getTitle(trainerSlot, this.variant);
    let title = includeTitle && this.config.title ? this.config.title : null;


    if (this.name) {
      if (includeTitle)

      // Check if i18n is initialized
        if (!getIsInitialized()) {
          initI18n()
        }
        title = i18next.t(`trainerClasses:${name.toLowerCase().replace(/\s/g, '_')}`);
      if (!trainerSlot) {
        name = this.name;
        if (this.partnerName)
          name = `${name} & ${this.partnerName}`;
      } else
        name = trainerSlot === TrainerSlot.TRAINER ? this.name : this.partnerName || this.name;
    }

    return title ? `${title} ${name}` : name;
  }
```

At least this is what i traced back from the UI that implements the name tag. 

And it seems to be the only place. 

So in my understanding it is totaly fine to leave the title as an empty string. (if it would be set also it would show the title twice)

bennybroseph: shouldn't we populate it with the translation then?

CodeTappert: No. What you propose is how it currently works on main. It then does not show the title in the name tag (as you see in the pictures). 

Maybe we could change it so the name tag also shows the title value and not just the name value if that would be better for you

bennybroseph: I think that makes the most sense logically, rather than setting a value to nothing and then adding what it should do to another value

CodeTappert: Okay i will try to find a way

CodeTappert:
@bennybroseph The Solution was way easier then i thought tbh... The only reason why the title wasnt shown in the nametag is because only the default values were used in the getName function. By setting this to true it is shown. (and also does not break the normal trainers)

See:
image


Well and the titles look like before (second image in first post)

bennybroseph (changes requested):
Don't forget semi colons at the end of lines.

Also, now that you have a better understanding of these files, can you leave some comments for others?

CodeTappert:
> Don't forget semi colons at the end of lines.
> 
> Also, now that you have a better understanding of these files, can you leave some comments for others?

Added semicolons and comments (but i dont know how much sense they make to someone who already knows how to code tbh)

bennybroseph (approved):

PR #314, opened 2024-04-26
New Feature: Move Effectiveness Color Display

paulcazac:
Comes with a toggle in settings. 
DEFAULT: OFF

Set the move text color to change dynamically depending on target based on the target's weaknesses. 
Currently it changes between different shades of green/red depending on effectiveness, and grey if it there no effect. Otherwise it defaults to standard white. 
Currently works in the range of 4x resist - 4x weakness. Niche cases of the 8x range can easily be implemented. 
Devs can change and alter colors to their liking in text.ts.

PR #1061, opened 2024-05-17
Add setting for showing type effectiveness hints

Yentis:
Choice between off, partial & full hints:
brave_JLKnHVGORB

Partial hints show type effectiveness above the affected opponent & add a color to the move:
brave_itbPAWNgGL

Full hints show hints in party select:
- Orange if the opponent(s) have super effective **typings** against your pokemon
- Green if the pokemon has a super effective move against the current opponent(s)
- Blue if the pokemon has a defensive advantage against the current opponent(s)
brave_tJWwRKtTl5

The same hint is also shown in the battle itself on the pokemon's name to easily know whether or not to switch
brave_zT8isRQspk

Color list:
image


j-diefenbach:
I think the 'full' hints mode is pushing it. You're effectively giving the player extra information they wouldn't necessarily have. I like the option of giving some adjustability so players who want to rely on their own knowledge (or have assistance) have the opportunity to do so. But being given indications of the opponent's moveset when it hasn't been revealed yet isn't in the spirit of the game.
Even if you were to only reveal the 'full' hints when you've seen all their moves I think it becomes unnecessary at that point.

j-diefenbach:
Note there is also a similar pull request #314 which does this without the full hints

Yentis:
I can change the full mode to show hints based on the opponent's types and not its moves, how does that sound?

Adri1:
Not a fan of displaying the type effectiveness directly with the name of the move, sometimes move names can be so long that adding extra stuff could make it overlap neighboring move names or going off the borders (as it already happen with (P) and (N)) mentions

Yentis:
How about this?
brave_sCOqPHGGHy

j-diefenbach:
> How about this? The effectiveness is shown under accuracy. brave_sCOqPHGGHy

Imo at this point it becomes a copy of #314 which has likely not been implemented because of conscious decisions from the devs. I think if this is something you're serious about implementing I would make it alot cleaner and I have doubts about adding another column of text to the move info (which has already been cluttered to the max).

Yentis:
No one has even responded to the other PR, so there's no official stance on it. I also don't know what you mean by "make it alot cleaner", do you have actionable feedback?

j-diefenbach:
I mean the red colour(s) are too saturated and would be weird to differentiate 3 levels of green/red-ness (I'm assuming this is supposed to be clarified but text but I'm not confident this text can be well displayed anywhere), its still unclear how dual typing effects the effectiveness indicator on the pokemon names as well as dual battles / how players would actually be able to interpret that (would it even be useful?). In terms of actionables maybe reconsider the additional text under the info section seeing how many people had issues with it having 3 sections of text plus icons I don't see why there should be another line. 

Also looking at the code doesn't the full hints option still give players additional information?

Yentis:
> I'm assuming this is supposed to be clarified by text but I'm not confident this text can be well displayed anywhere

The effect text shows you the actual multiplier the move has and you can then learn what the colors mean in terms of effectiveness.

> its still unclear how dual typing effects the effectiveness indicator on the pokemon names as well as dual battles

In double battles the idea is to look at the effect text as it shows which pokemon the move is effective against, the move's color will be the one for the highest effectiveness. As for the name, that's just the same thing as in the party select screen. It's so you can decide whether to stay in or not without having to open the party menu to check type matchups.

> Also looking at the code doesn't the full hints option still give players additional information?

No, it only uses the opponent's types which can be seen at any time.

j-diefenbach:
> No, it only uses the opponent's types which can be seen at any time.
Ah sorry i misinterpreted `this.getAttackMoveEffectiveness(source, move)`
> It's that big of an issue? Sounds like vocal minority.
This is something that can be brought up in the dev-corner chat once it reopens but there were quite a few people against it (but accuraccy, power and pp were all necessary enough to the devs). I'm sure there will only be more people who are against adding another row. It's not that I'm against it I just think there are many issues that arise from trying to squeeze another section of info into the battle UI

Yentis:
https://github.com/pagefaultgames/pokerogue/assets/17981128/844cb710-b09b-431a-b8ac-426c8d51ba94

Thoughts?

Yentis:
New colors:
brave_Vut16u2SnK

They all range from contrast 4.5 to 5.5 according to https://www.siegemedia.com/contrast-ratio.
I also changed partial hints to only show move effectiveness, as that's a more clear distinction.

CodeTappert:
> New colors: brave_NPPSwOpanf
> 
> They all range from contrast 4.5 to 5.5 according to https://www.siegemedia.com/contrast-ratio. I also changed partial hints to only show move effectiveness, as that's a more clear distinction.

As a person with protoanomalie (red color deficency) i can tell you that the difference between 0.25 and 0.125 isnt that big. They look pretty similar.

But as long as they will have the text next to it (like in the main post) it should be fine

CodeTappert:
Well i just saw that there isnt a number text to the move anymore...Thats bad with how similar the colors are

Yentis:
The numbers are shown above the enemy sprites instead of next to the move as shown in https://github.com/pagefaultgames/pokerogue/pull/1061#issuecomment-2120297419

CodeTappert:
> The numbers are shown above the enemy sprites instead of next to the move as shown in [#1061 (comment)](https://github.com/pagefaultgames/pokerogue/pull/1061#issuecomment-2120297419)

Also in Solo Battles?

Yentis:
yes

CodeTappert:
Then 

> yes

Then it is fine! Good job

brain-frog:
A few things before I review further: 

- can you update the original PR description to have the most up to date description and visuals
- please add documentation (you can reference the [guide](https://github.com/pagefaultgames/pokerogue/blob/main/docs/comments.md)) for any new and updated functions

Yentis:
Updated PR description & added comments where they could be considered meaningful

brain-frog:
The developers have discussed this and don't believe we would like the "Full" option.
Also instead of the effectiveness hovering above the sprite, can it be next to the enemy hp/info box

Yentis:
> Also instead of the effectiveness hovering above the sprite, can it be next to the enemy hp/info box

image
Like this?

Luminiferous348:
One Problem with positioning of the effectiveness box besides the enemy pokemon info box is, that it collides with showing the enemy move information. So it should either move automatically further to the right, when the move information is shown, or it should disappear.

brain-frog:
> One Problem with positioning of the effectiveness box besides the enemy pokemon info box is, that it collides with showing the enemy move information. So it should either move automatically further to the right, when the move information is shown, or it should disappear.

I think its fine if they dont display when move info box is shown, that's a toggleable anyway

> Also, the effectiveness boxes kinda clash the theme of the enemy info box, but I don't feel like thats a big issue.

We'd need new UI boxes if we wanted to do that right, which I dont think is fair to require for your change

Yentis:
https://github.com/pagefaultgames/pokerogue/assets/17981128/d272e0f2-8448-4eee-8d87-3f65bbb1b4ba

Updated to hide the box while move info is showing

src/system/settings/settings.ts
```diff
@@ -143,6 +144,13 @@ export const Setting: Array<Setting> = [
     default: 0,
     type: SettingType.GENERAL
   },
+  {
+    key: SettingKeys.Type_Hints,
+    label: "Type hints",
+    options: OFF_ON,
+    default: 0,
+    type: SettingType.GENERAL
```
brain-frog, line 152: please add to DISPLAY

src/data/type.ts
```diff
@@ -501,6 +501,51 @@ export function getTypeDamageMultiplier(attackType: integer, defType: integer):
...
+    case 0:
+      return "#929292";
+    case 0.125:
+      return "#FF5500";
+    case 0.25:
+      return "#FF7400";
+    case 0.5:
+      return "#FE8E00";
+    case 1:
+      return "white";
```
brain-frog, line 519: move-hint
don't forget about legacy ui

brain-frog (approved):
great work, thanks for all the edits

whatanoob:
This is a nice feature to have but I wish there was an option to only show the colored text. Since I'm someone who only occasionally needs these hints, I'd still like to leave it on, but atm I find myself leaving it off to reduce visual clutter.

DayKev:
Looks like it doesn't understand Terastallization? Or it was cheating for me and telling me they were about to switch to a Ground type (which did happen, a Stunfisk came out after I chose an attack).

image

travbp:
Love this feature so far! I did notice that Revelation Dance doesn't give the right type effectiveness hints, but its type is always shown as Normal so that could be why. image

Issue #1048, opened 2024-05-17
Brave Bird hit 2 times with Multi Lens, second hit recoil removed too much life


jerleroy:
My Staraptor is lv127, the Crobat is lv110+ . I use Brave Bird while my Staraptor has Multi Lens.

The first hit remove 2/3rd of Crobat life, my Staraptor loses ~80hp.

The second hit removed the last 1/3rd of Crobat life, my Staraptor loses ~110hp (in addition to the previous ~80hp)

(Good luck to you and thanks you for your work)

PR #1063, opened 2024-05-17
Calculate recoil damage correctly

rayanzmn:
As #1048 highlighted recoil damage isn't calculated correctly. Currently the recoil damage is a certain % of the total damage the user has dealt during the entire turn, so when recoil is activated within the same turn in which damage was already dealt then recoil damage is higher than expected. This change makes it so that recoil damage takes only into account the damage dealt by the recoil move.

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -779,8 +779,8 @@ export class RecoilAttr extends MoveEffectAttr {
     if (cancelled.value)
       return false;
 
-    const recoilDamage = Math.max(Math.floor((!this.useHp ? user.turnData.damageDealt : user.getMaxHp()) * this.damageRatio),
-      user.turnData.damageDealt ? 1 : 0);
+    const recoilDamage = Math.max(Math.floor((!this.useHp ? user.turnData.currDamageDealt : user.getMaxHp()) * this.damageRatio),
+      user.turnData.currDamageDealt ? 1 : 0);
     if (!recoilDamage)
       return false;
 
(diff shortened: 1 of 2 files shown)


rayanzmn:
*take into account the damage dealt by the recoil move.
So it works even when recoil is activated multiple times in a single turn. Should've worded it better.

TempsRay (approved):
Looks good to me for fixing the bug. Weird that `turnData` stores information for total damage dealt during the entire battle though, when it should only store damage dealt this turn.

rayanzmn:
That's true I also find it weird. I was thinking of changing the name of turnData.damageDealt into turnData.totalDamageDealt since I suppose it could still be useful but I would've broken pretty much everything that uses it.

rayanzmn:
Actually now that I think about it, it does actually store the damage dealt during the turn. The issue is that if damage is dealt twice in a single turn it doesn't differentiate between the damage dealt the first time and the second time, it simply adds them.

rayanzmn:
I have edited the PR description to better describe the bug.

bennybroseph (changes requested):
Can we get a video of how this was tested?

rayanzmn:
This is the current behavior of recoil (bugged):

https://github.com/pagefaultgames/pokerogue/assets/129906641/c57e3b51-0ea2-4f10-a76a-f52a410a65dc

Whereas this is the behavior of recoil after having implemented the fix:

https://github.com/pagefaultgames/pokerogue/assets/129906641/dfb968ff-3a7f-4e39-b55c-3744187ceda5

In both videos I have the console opened because I used the `console.log()` method to show the contents of `user.turnData.damageDealt` and `user.turnData.currDamageDealt` just for the sake of these tests (they're not in the actual PR).

bennybroseph (approved):

PR #1069, opened 2024-05-18
Add a manager for timed events and display event banners during events

IceFire03:
Currently, weekend events are handled in a hacky manner, just adding a date check and shiny multiplier into pokemon.ts; this PR implements a TimedEventManager to handle events in a more standardized and safe manner. New events can be entered via the TimedEvent interface and can be scheduled to occur during certain dates. Then, banners can easily be added to the events folder and will be displayed on the title screen during events. Tested for all screen sizes and for events and edge cases.
Screenshot 2024-05-17 at 7 01 23 PM


bennybroseph (dismissed):

bennybroseph:
I like this as is where it's already placed. Timed events are hard to test, but assuming this works as intended it's a yes from me.

src/loading-scene.ts
```diff
@@ -209,9 +213,6 @@ export class LoadingScene extends SceneBase {
 
     this.loadImage('egg_list_bg', 'ui');
 
-    this.loadImage('end_m', 'cg');
-    this.loadImage('end_f', 'cg');
-
```
TempsRay, line 214: Is this intentional?

IceFire03: No, I think my branch wasn't properly synced. Fixed!

IceFire03:
> I like this as is where it's already placed. Timed events are hard to test, but assuming this works as intended it's a yes from me.

Agreed, I tested the timing extensively by actually changing the date/time on the local server and it worked without a hitch.

brain-frog:
We all of a sudden last minute needed to add this functionality so I very much appreciate the fact that you made this. I implemented it based on this with some adjustments.

Issue #1075, opened 2024-05-18
Enable Discussions in this repository


karl-police:
yes

anything communication related, with the ability to find threads or something that isn't a locked chat channel

then I can ask about uhh 🤔 controller... and UI functions

PR #1080, opened 2024-05-18
#vote to #feature-vote

karl-police:
ok that's the change

now I am seeking the UI function like explanation, yep

help

thanks

do not ask for if UI supports controller, that doesn't help me, cuz if it doesn't support controller basically if you'd draw one of those mindmaps, I'd draw thousands of arrows that point back to --> Aslong I don't understand the UI functions, how one gonna make UI work right

I mentally died going through them, maybe if I give it ago another time, maybe I can figure out, but for now, no.

Hmmmmmm

Check out other great posts:

https://github.com/pagefaultgames/pokerogue/issues/1075

Channel is not named vote anymore or never was. I was seeking for vote but I didn't find vote, so feature-vote

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ PokéRogue is a browser based Pokémon fangame heavily inspired by the roguelite
 
 # Contributing
 ## 🛠️ Development
-If you have the motivation and experience with Typescript/Javascript (or are willing to learn) please feel free to fork the repository and make pull requests with contributions. If you don't know what to work on but want to help, reference the below **To-Do** section or the **#vote** channel in the discord. 
+If you have the motivation and experience with Typescript/Javascript (or are willing to learn) please feel free to fork the repository and make pull requests with contributions. If you don't know what to work on but want to help, reference the below **To-Do** section or the **#feature-vote** channel in the discord. 
 
 ### 💻 Environment Setup
 #### Prerequisites


karl-police:
If @Madmadness65 sees and puts label here without opening this Discord channel

image


I will lose and fail to believe the existance of everything.

IceFire03:
What are you asking for? You can see controlled mappings in /src/configs, which are mapped in /src/inputs-controller.ts to translate any input into a Button. These Button presses are then used to navigate around the UI. It's not very clear to me what you want help with.

IceFire03 (approved):
This part of the PR makes sense, though

karl-police:
> What are you asking for? You can see controller mappings in /src/configs, which are mapped in /src/inputs-controller.ts to translate any input into a Button. These Button presses are then used to navigate around the UI. It's not very clear to me what you want help with.

I seek explanations, just like yours. 👍 

Though how do I test them without a controller 🤔

IceFire03:
It would probably be difficult, and I wouldn't recommend it, but you could definitely simulate the keystrokes, i.e. using [reWASD](https://www.rewasd.com/releases/release-5.2.0) or any other software

karl-police:
Controller can't even hover over the Modifier icons, how do they view the Modifiers descriptions, or can they hover over?

IceFire03:
I'm pretty sure you can't see item descriptions right now on controller.

karl-police:
> I'm pretty sure you can't see item descriptions right now on controller.

3x (1)

Madmadness65 (approved):

PR #1081, opened 2024-05-18
Mostly implement the Harvest ability

OrangeRed:
### Current implementation:
Once per turn there is a 50% to create a random berry that was eaten this battle. Harvested berries are removed from the pokemon's memory until they are eaten again. The proc chance is doubled in sun.

> If the Pokémon has used a Berry, there is a 50% chance of it being restored at the end of each turn. The held Berry is remembered even if it is not restored or if the Pokémon has switched out. Harvest will always activate in harsh sunlight.
> > https://bulbapedia.bulbagarden.net/wiki/Harvest_(Ability)

_This part of Harvest isn't implemented because neither Fling nor Nature's Gift are implemented._

> Harvest can recover Berries consumed by [Fling](https://bulbapedia.bulbagarden.net/wiki/Fling_(move)) and [Natural Gift](https://bulbapedia.bulbagarden.net/wiki/Natural_Gift_(move)); 

https://github.com/pagefaultgames/pokerogue/assets/5421700/5fa12354-a258-4dee-9909-94b770f78238

https://github.com/pagefaultgames/pokerogue/assets/5421700/18fecc72-ff01-4339-8251-47e70920dae6

 ---

### Techincal stuff

- As long as a berry is pushed into the `pokemon.battleData.berriesEaten[]` it will be considered a candidate for regeneration.
- I had to rewrite the `BerryPhase` class and spawn it before the `TurnEndPhase` 
  - The previous implementation would spawn `BerryPhase` and push it into the phase queue during the `TurnEndPhase`. Doing this means that the `PostTurnAbilities` which run during the `TurnEndPhase` receive data that has yet to updated by the `BerryPhase`.
  - Essentially Harvest on the current turn would trigger with `pokemon.battleData.berriesEaten[]` from the previous turn.


molker (changes requested):
This looks good, once you fix the merge conflicts ping me and I can re-review and approve

molker (approved):
lgtm

Issue #1072, opened 2024-05-18
If a pokemon fussion beats classic, only base species gets the ribbon.


ReneGV:
Desired behavior:
- Both species should get the ribbon.
- You should get 2 5x tickets if you complete classic for first time with both species.
- You should get 1 5x tickets if you complete classic for first time with only one of both species.
- You should get 0 5x tickets if you have already completed classic with both species

TempsRay:
This is not a bug since the current behavior is intentional. This is a suggestion.

PR #1082, opened 2024-05-18
Classic Mode - Add ribbon to both pokemon species when there is fusion

ReneGV:
Implementing this issue: https://github.com/pagefaultgames/pokerogue/issues/1072

Add a ribon to both pokemon species whenever a pokemon is a fussion.

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -3638,11 +3638,11 @@ export class GameOverPhase extends BattlePhase {
               if (pokemon.species.getRootSpeciesId() != pokemon.species.getRootSpeciesId(true)) {
                 this.awardRibbon(pokemon.species, true);
               }
-
               if(pokemon.isFusion()){
-                this.awardRibbon(pokemon.fusionSpecies);
-                if (pokemon.fusionSpecies.getRootSpeciesId() != pokemon.fusionSpecies.getRootSpeciesId(true)) {
-                  this.awardRibbon(pokemon.fusionSpecies, true);
+                let fusionSpecies = pokemon.fusionSpecies
+                this.awardRibbon(fusionSpecies);
+                if (fusionSpecies.getRootSpeciesId() != fusionSpecies.getRootSpeciesId(true)) {
+                  this.awardRibbon(fusionSpecies, true);
                 }  
               }
             }
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -3728,7 +3728,6 @@ export class GameOverPhase extends BattlePhase {
       this.firstRibbons.push(getPokemonSpecies(pokemonSpecies.getRootSpeciesId(forStarter)));
     }
   }
-
 }
 
 export class EndCardPhase extends Phase {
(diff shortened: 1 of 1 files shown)


j-diefenbach:
This isn't a bug it's intended behaviour. Otherwise you would expect to get two tickets, one for each pokemon in the fusion so if you win classic with a full party of fusions you get twice the tickets.

ReneGV:
> This isn't a bug it's intended behaviour. Otherwise you would expect to get two tickets, one for each pokemon in the fusion so if you win classic with a full party of fusions you get twice the tickets.
@j-diefenbach 
Yep I agree, however the `awardRibbon` method I used also awards ticket rewards.
So if you beat the game with a party full of fussions you would indeed get the double annount of tickets.

If you fuse pokemon A and B, both texts:
`A beat classic for first time, you received ...`
`B beat classic for first time, you received ...`

j-diefenbach:
> So if you beat the game with a party full of fussions you would indeed get the double amount of tickets.

Yeah my opinion is that this is sorta unbalanced. Obviously it's not game breaking but I think it makes sense you only get the ribbon and ticket for the base pokemon. This has most likely been considered before and I'm sure the devs originally decided not to do it that way.

jaimefd:
I may be missing some strat to get lots of DNA splicers in classic but i rarely ever get one, never even got more than 2 in a single run, I dont think this is unbalanced

TempsRay:
Originally the main reason to prevent this was we didn't want a case where a player would have Pokemon A fused with Pokemon B but be missing Pokemon B's starter dex entry which would result in a weird case of missing a dex entry but having a ribbon. However, since this only applies to classic mode, the only way to own a fusion is to have both component starters owned and DNA splicers are super rare, I'm not completely opposed to this change.

ReneGV:
Thanks for your feedback @TempsRay. It looks there is consensus that this change is not unbalanced 
Please let me know if there is work remaining to do in order to get approval,

bennybroseph:
I don't think this is something that should be added. My reasoning is actually because I don't think capturing a Pokemon mid-run should count either. I know that's way too much of a hot take, but I feel like the ribbon is intended to be a "Hey I started with this Pokemon at floor 1 and beat classic with it!" but really it's kinda just a "Hey I took this Pokemon to Eternatus" and that's it. It doesn't even need to participate in a single battle. 

Again, it's a hot take and too late to really change now, but I'd make it to where the Pokemon must be selected in the starter selection screen. Then you could fuse it or do whatever you want and it'd count so long as it stayed on your team til the final floor.

So yeah, this change moves even further from that so I don't really like it even though it rarely happens anyway.

Tempo-anon:
> Again, it's a hot take and too late to really change now, but I'd make it to where the Pokemon must be selected in the starter selection screen. Then you could fuse it or do whatever you want and it'd count so long as it stayed on your team til the final floor.

Heavily disagree with this since that would make gathering ribbons even more tedious than it currently is for legendary, paradox, and expensive Pokemon, and there's already an in game precedent for how the champion's ribbon works.

As for the PR itself, while this would be a nice QOL change that would not have giant impact, we are planning on adding spliced classic in the future and in such a case it would not make sense for fused pokemon to give double ribbons as that would make spliced classic a much more efficient ribbon/egg farming method. The alternative workaround of only allowing non-spliced classic fusions give double ribbons is quite hacky.

PR #1088, opened 2024-05-18
Implement/Fix Beak Blast

jatinkohli:
# Beak Blast
This is my attempt at a faithful recreation of beak blast, based on [bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Beak_Blast_(move)), [this video](https://www.youtube.com/watch?v=fcBHkRqK3Hc), and showdown testing of various cases. Sorry for the essay of a pr lmao, there is a lot about beak blast I was confused about that I feel needs clarification.

## Beak Blast Mechanics

A hopefully thorough overview of beak blast's bizarre and non-sensical mechanics:

Beak blast has two portions. First, Toucannon's beak heats up, causing pokemon that make contact with it to become burnt. Then, Toucannon unleashes an attack.

- The first portion happens at the very beginning of a turn, outspeeding even helping hand (+5 priority). I tested the order of events in showdown against moves but did not test the order of events against quick claw or other beginning of turn effects.
- The second portion happens with -3 priority.
- Both portions happen in the same turn.
- The first portion happens regardless of status conditions (sleep, freeze, etc.), whereas the second portion acts like a normal move. This means beak blast can burn opponents while Toucannon is asleep, frozen, etc.
- The second portion is affected by status effects (sleep, freeze, flinch, etc.) like any normal move, even if Toucannon becomes affected by a status between the first and second portions.
- The first portion can affect bulletproof pokemon, while the second portion (normal attack damage) cannot.
- Contact moves that move after the second portion (i.e. dragon tail) do not cause that pokemon to be burnt.

## Implemented Features/Fixes

With all of this in mind, I implemented/fixed these features about beak blast:

- Made pokemon that make contact with beak blasting Toucannon between beak blast's first and second portions get burnt (if able to be burnt).
- Pokemon that do damage but do not make contact do not get burnt, pokemon that make contact after the second portion do not get burnt.
- The first portion happens regardless of status while the second portion is affected by status.
- Conversely, the second portion is affected by status (i.e. sleep) inflicted between the first and second phases. This was not the case before, the second portion could go through sleep for instance.
- Fixed a bug where adding the new move phase for the second portion would add the move phase after the end turn phase, slightly affecting timing of move effects like aqua ring.

## Implementation/Testing

The burn mechanic works by giving Toucannon a custom tag (BeakBlastTag) during the first portion. During every MoveEffectPhase, the game attempts to resolve all BeakBlastTags on opposing pokemon after damage is dealt to them. Then, the game removes the BeakBlastTag during the second portion with a custom MoveEffectAttr (RemoveBeakBlastTagArr) on the beak blast move. 

The two weird parts about this implementation are hardcoding a line during every MoveEffectPhase to resolve BeakBlastTag at the right time and flipping the followUp field of MovePhase during the doMove portion of the start method. Flipping followUp allows the first portion to bypass status and the second to be affected by it, inverse of how it is handled for other multi-attack moves. All things considered, these aren't that bad as far as poor programming practices go.

I tested these changes against all of the above requirements of beak blast (burn on contact, don't burn on non-contact, burn only between portions, portion timings and order, bulletproof, beak blast with status in the middle, beak blast with status at start of turn) in PokeRogue enough to be fairly confident that all the mechanics work as intended, but I cannot promise I didn't miss an edge case.

 ## Conclusion 

**Definitely let me know if anything looks wrong, if there is a cleaner way to do things, or if you all need help testing this move.** I have learned way too much about this froot loops mfer and his thermodynamically-impossible george foreman of a beak.


jatinkohli:
# Testing Videos:

Beak Blast has two phases and burns pokemon who use contact moves in between the phases. It does not burn users who use non-contact moves.

https://github.com/pagefaultgames/pokerogue/assets/22225497/3d800831-c66c-4934-b114-4d137539d55e



https://github.com/pagefaultgames/pokerogue/assets/22225497/b144a53f-9c6a-455d-b4cd-82dcc7154475


The second portion of beak blast is affected by status conditions (asleep, paralyzed, frozen, etc.) as normal, but the first portion is immune to them and happens regardless.

https://github.com/pagefaultgames/pokerogue/assets/22225497/64c45a71-a50a-47e4-9ffe-5b055188a164



Beak Blast's first portion moves before all other moves, regardless of priority.

https://github.com/pagefaultgames/pokerogue/assets/22225497/215574cd-717a-47f6-b02b-252c0863ba8d



Beak Blast does not burn pokemon who use contact moves after the second portion

https://github.com/pagefaultgames/pokerogue/assets/22225497/fefc1deb-06e6-4bdc-9c75-77e481e15976



Beak Blast can burn pokemon with bulletproof but cannot damage them.

https://github.com/pagefaultgames/pokerogue/assets/22225497/9519d21b-a3f0-47d0-9098-8adbde68f83b

src/battle-scene.ts
```diff
@@ -1671,6 +1671,20 @@ export default class BattleScene extends SceneBase {
   }
 
   pushMovePhase(movePhase: MovePhase, priorityOverride?: integer): void {
```
DustinLin, line 1673: Can we get some motivation/explanation for this change? Purposely moving phases around may cause some unintended consequences. I'm guessing there are many calls to `pushMovePhase()` throughout the battle

src/data/move.ts
```diff
@@ -584,6 +584,32 @@ export class MoveEffectAttr extends MoveAttr {
   }
 }
 
+export class RemoveBeakBlastTagArr extends MoveEffectAttr {
+  constructor() {
+    super(true, MoveEffectTrigger.PRE_APPLY, false);
+  }
+
+  canApply(user: Pokemon, target: Pokemon, move: Move, args: any[]) {
```
DustinLin, line 592: the devs may want this to be in [JSDoc style](https://discord.com/channels/1125469663833370665/1239957703820836975)

src/battle-scene.ts
jatinkohli: I just double checked, pushMovePhase is only called one time in ChargeAttr's apply, and only for instances of ChargeAttr with the sameTurn flag set. Beak Blast is the only move atm which creates a ChargeAttr with this flag, so essentially the pushMovePhase is only used for beak blast (I believe). 

This change fixes a bug where the second move phase for beak blast happened after the TurnEndPhase, causing effects like aqua ring heal and leech seed drain to happen before Toucannon attacks. The TurnEndPhase has no fields or specific data, so I don't think there is any problem with deleting the original TurnEndPhase and pushing a new one after Toucannon's second move phase, but I would love some feedback if there's something I'm missing. The two move phase system is how beak blast was implemented before, I did not change this.

src/data/move.ts
jatinkohli: Turns out this MoveEffectAttr is unnecessary since I can just have the BeakBlastTag lapse on MoveEffect to remove the tag automatically, thanks for making me take a second look at this.

src/battle-scene.ts
jatinkohli: While I was taking a look at your other comment, I also noticed and fixed pushMovePhase pushing the move phase before other phases that should happen at the end of turn like weather and status effects, so that should be fixed now to. Thanks again!

flx-sta:

-----

PR #1091, opened 2024-05-18
Fix README to match current state of Discord

karl-police:
I was not able to type in ``#dev-corner`` 😔, after asking one Staff. My hopes are shattered. I will still have to explore the source code all by myself and go through pain. And what's the point typing in dev-corner if not all Contributors are Contributors on the Discord. Who will reply back, a ghost? The answer is nobody.

&nbsp;

To avoid anyone else from experiencing this, I simply stroke through the motivating text to join the Discord. Since this so called "lockdown" is prolly gonna stay up for long, and yes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ PokéRogue is a browser based Pokémon fangame heavily inspired by the roguelite
 
 # Contributing
 ## 🛠️ Development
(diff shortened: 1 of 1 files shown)


Madmadness65:
The situation in the Discord is only a temporary one until everything gets restructured under the new ownership. It's not going to be permanently locked.

PR #1074, opened 2024-05-18
[QoL] Make IV-Scanner less annoying, by giving it a toggle and the ability to scan stuff at any time by clicking on the icon

karl-police:
**Edit:** scroll down, I fixed some stuff

I was trying to figure out how to do this for like way too long. I am not happy with the end result.

The freaking code base has a bunch of lines but **lack comments above them**, that explain what is what. So, when I was trying to figure out what the hell ``Mode.MESSAGE`` and the rest is for, I had a mental breakdown.

And the result, is buggy. Except the Toggle button, that one works... So I expect someone to like tell me how to fix, yes?? :) :) :)

&nbsp;

Anyways, let me explain the change.

FIRST, I tried to push back **CommandPhase** dynamically. But that thing would break, here's my freak out post about it: https://github.com/pagefaultgames/pokerogue/issues/1062

So, then I decided to do this more stupid. But eventually, it's not really that bad actually.

&nbsp;

Currently, nobody will freaking choose the IV-Scanner if they've chosen it once, pretty sure. That freaking thing prompts a player if you want to scan a Pokemon every encounter, everytime.

So, I added a Toggle.

You **click** on the Modifier. And yes, I tried to listen to the SOLID Principle and OOP. There's a ``onClick`` method that gets now called onto all the Modifiers in that Modifiers bar, but it doesn't do anything unless you assign it something.

This should even function on mobile.

Eitherways.

&nbsp;

This UI Container will open. It opens like that, because I copy pasted the other existing code and I think the offset is alright too.

The Toggle button is amazing and it toggles stuff.

&nbsp;

&nbsp;

**Then!**, the other option is that you can scan anything anywhere. Ofc only if it's a wild encounter and bla bla bla.

You simply click on it.

Draw back is, you'll be stuck with this box until the end of the turn.

This is where I need help with. And yep. Okay thanks.

PR #1093, opened 2024-05-18
Added A Toggle For The IV Scanner In The Style Of Lock Capsule

Simplyfenton:
Fairly certain this is a feature that a lot of players want (myself included)

Added the toggle to the shop on the bottom right hand side, to mirror that of the Lock Capsule.

It's certainly not the prettiest implementation, and the logic could likely be compacted from what I've got, but in fairness to me, I did start making this at 8am.

(Any feedback and stuff would be great, first time contributing to a public project like this)


CodeTappert:
Please add images showing it.

karl-police:
conflicts with https://github.com/pagefaultgames/pokerogue/pull/1074

Simplyfenton:
> Please add images showing it.

The text being red indicates that the IV Scanner is currently disabled.



Demo


> When I first, got the idea to add that. I realized that adding a Toggle simply isn't enough. What's the point of being able to toggle it. You don't know what Pokemon you encounter next. But giving the ability to Force Scan as well. Now that one, is a pog.

I think a "force scan" is something that would likely be useful, but as it stands, I don't think there's anywhere that it would go within the battle UI that would "fit", and it certainly wouldn't belong in the standard menu. Since neither the Lock Capsule or this IV Scanner setting are saved, refreshing the page would re-enable the IV Scanner at the start of the current battle which, while not ideal, is an option before a wider discussion about the "force scan"'s implementation. I don't know how many people I speak for, but I personally wouldn't mind not having the force scan option - the upside of not having the dialogue boxes at the beggining of every non-trainer battle is then balanced by the downside of not having that information to hand readily.

CodeTappert:
Definitly it can not be that color. 
It is really hard to read that way. (But maybe its my red color deficency). There is not.enough contrast between the text and the Background

karl-police:
> > Please add images showing it.
> 
> The text being red indicates that the IV Scanner is currently disabled.
> 
> Demo
> 
> > When I first, got the idea to add that. I realized that adding a Toggle simply isn't enough. What's the point of being able to toggle it. You don't know what Pokemon you encounter next. But giving the ability to Force Scan as well. Now that one, is a pog.
> 
> I think a "force scan" is something that would likely be useful, but as it stands, I don't think there's anywhere that it would go within the battle UI that would "fit", and it certainly wouldn't belong in the standard menu. Since neither the Lock Capsule or this IV Scanner setting are saved, refreshing the page would re-enable the IV Scanner at the start of the current battle which, while not ideal, is an option before a wider discussion about the "force scan"'s implementation. I don't know how many people I speak for, but I personally wouldn't mind not having the force scan option - the upside of not having the dialogue boxes at the beggining of every non-trainer battle is then balanced by the downside of not having that information to hand readily.

I believe that with additions that allow flexibility in the code itself, to quickly change how it is implemented, shouldn't bring a worry on where it is currently put at.


The IV-Scanner is just one of current things. If I'd toggle the IV-Scanner off, and I encounter a Pokemon that I do want to scan, then one is simply just screwed. But you can re-open the game and then just scan it, that's true.

Then there is the issue at where to put it. I could ask myself, HOW many players, would toggle it off, and never toggle it on again? 🤔 But the toggle button will still show up there.

What if there's other things in the future, other than the IV-Scanner.

Simplyfenton:
> Definitly it can not be that color. It is really hard to read that way. (But maybe its my red color deficency). There is not.enough contrast between the text and the Background

Yeah, I largely just picked the colour so that it'd be obvious that it's off, it's the same white as the rest of the text when the IV Scanner is enabled. Realistically, a dulled grey might work well, wasn't something I particularly looked into though.

Simplyfenton:
Updated the toggled colour so that it now displays in Grey rather than Red, quite a bit more readable now.

Demo2

CodeTappert:
> Updated the toggled colour so that it now displays in Grey rather than Red, quite a bit more readable now.
> 
> Demo2

Way better! :)

Simplyfenton:
I _Believe_ this is following the new linting stuff, wasn't seeing anything wrong on my end as far as I could tell

Simplyfenton:
Just bumping this for any potential feedback/thoughts

Simplyfenton:
Resolved Conflicts in Phases.ts

Simplyfenton:
Fixed linting issues that I evidently didn't fix previously

flx-sta:
Considered stale.

PR #1095, opened 2024-05-18
Game Stats Locales

Adri1:
Localization for Game Stats.
The only entry in the stats I haven't found how to open to translation is "Play Time"


Dakurei:
You've removed question marks in some translations, even though they're mandatory for the interface to work correctly.
They are parsed to hide the field when necessary.

I haven't checked to see if you've blown up any others, but this one is already there.
image
image

image

Adri1:
Oops, I'm on it

Adri1:
Should be okay now

Dakurei:
It looks okay for the display compared to what's already sticking out in the original language
> Original


> Located

Dakurei:
Regarding "Play Time". Which should be translated as "Temps de jeu", it's here:
image

And of course, add the necessary locales
image

Adri1:
Thanks a lot, its been added

Dakurei:
Personally, to avoid touching the question mark parsing system, and to prevent translators from removing them in the same way as you without knowing what they're for, I would have switched these obligatory characters outside the translation

Like this for each field concerned (_And there are a few I know 🤣_):
image
image

image

image
image


**Perhaps someone else's opinion on this change would be more practical than revising the system of how to hide fields**

Adri1:
I didn't knew for what it is but I didn't meant to scrapped them either, it was purely an accident mdrr

Adri1:
Modifications done

Xavion3:
We can stop this abuse of question marks by cleaning up this file. This is untested mind, just from reading the code. But you should be able to remove the entire anonymous function at the end if you convert the displayStat object into into only DisplayStat objects. This would remove the need to append question marks so the localisation file can work correctly so is probably the best approach.

For example, for the mythicalPokemonCaught line you'd turn it into something like this. You just have to go through each of the stats and give it an object with a label, sourceFunc, and hidden entry. This is a bit of work now, but feels like by far the best approach to handling the localisation problems here, particularly as we should not have actual functionality affected by the localisation text.
```ts
mythicalPokemonCaught: {
  label: i18next.t("gameStatsUiHandler:mythicalsCaught"),
  sourceFunc: gameData => gameData.gameStats.mythicalPokemonCaught.toString(),
  hidden: true
},
```

Adri1:
It's updated and it works, however, I don't feel I have the skill to do the necessary clean up, so if anyone is down to do it, feel free to

Dakurei:
It doesn't seem to be totally working

image

Adri1:
Huh weird, there's no error on my side

Dakurei:
I'll take care of the rest of the cleaning, and I'll see how I can offer it on this PR 👍

bennybroseph:
Looks like this was abandoned in favor or #1106

PR #1102, opened 2024-05-18
Added Firebreather trainer class

Alekemon:
As a test to how this engine works I wanted to start by adding a trainer class

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,9 +24,6 @@ dist-ssr
 *.sln
 *.sw?
 
-# Docummentation
-docs/*
-
 public/images/trainer/convert/*
 public/images/battle_anims/input/*.png
 public/images/pokemon/input/*.png
--- a/README.md
+++ b/README.md
@@ -8,13 +8,16 @@ If you have the motivation and experience with Typescript/Javascript (or are wil
 
 ### 💻 Environment Setup
 #### Prerequisites
-- node: 18.3.0
+- node: 20.13.1
 - npm: [how to install](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
 
 #### Running Locally
 1. Clone the repo and in the root directory run `npm install`
     - *if you run into any errors, reach out in the **#dev-corner** channel in discord*
 2. Run `npm run start:dev` to locally run the project in `localhost:8000`
+
+#### Linting
+We're using ESLint as our common linter and formatter. It will run automatically during the pre-commit hook but if you would like to manually run it, use the `npm run eslint` script. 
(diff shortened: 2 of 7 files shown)


chacolah:
Have you tested it at all?

Alekemon:
> Have you tested it at all?

I tried, I checked all errors marked by npm and fixed them but I can't get to open the localhost page

Alekemon:
Ok I opened inspect and notified another error, going to fix it

Alekemon:
Now the game opens in localhost:8000, I'll test when a firebreather battle pops in

Alekemon:
Is there a way to force a trainer battle to ocur in debug to save time?

Alekemon:
Now it works as intended

TempsRay:
Seems okay to me. No dialogue though?

Alekemon:
I've implemented dialogue now

src/data/dialogue.ts
```diff
@@ -414,7 +414,17 @@ export const trainerTypeDialogue = {
         `Time to rev up and leave you in the dust!`
       ],
       victory: [
-        `I'll tune up for the next race.`
+        `If we only had it warmer.`
+      ]
+    }
+  ],,
```
LaukkaE, line 420: Double colon here

Alekemon: Thanks for notifying the oversight! I fixed that

LaukkaE:
Trainer spawns correctly in Volcano, and seems to work right. 
Could use more dialogue if available.

Alekemon:
> Trainer spawns correctly in Volcano, and seems to work right. Could use more dialogue if available.

I'll take a look at dialogues from Pokémon HGSS (where the Firebreather trainer class originates from) to implement into the trainer class

Alekemon:
I've added some more dialogue lines

LaukkaE (approved):

src/data/dialogue.ts
```diff
@@ -418,6 +418,20 @@ export const trainerTypeDialogue = {
...
+    {
+      encounter: [
+        `My flames shall devour you!`,
+        `I'm seriously upset by your cold confusion!`,
+        `I'm going to put on my fire-breathing shows!`
+      ],
+      victory: [
+        `If we only had it warmer.`,
+        `I flamed out.`
+
```
molker, line 431: can we get one more victory dialogue to balance out the encounter and victory count?

Alekemon: Sure, I added a 3rd dialogue now

PR #1104, opened 2024-05-18
Implement Destiny Bond move

JakubHanko:
Implements Destiny Bond by adding a tag to the user.

The tag either runs out before the user's next move or triggers after receiving fatal damage.

https://github.com/pagefaultgames/pokerogue/assets/60473007/2800020d-3cf9-4098-b5c9-5d2fda8ba2d9

https://github.com/pagefaultgames/pokerogue/assets/60473007/7c604cb5-2c32-492a-ab03-451020814712


src/data/move.ts
```diff
@@ -4944,8 +4964,8 @@ export function initMoves() {
       .unimplemented(),
     new SelfStatusMove(Moves.DESTINY_BOND, Type.GHOST, -1, 5, -1, 0, 2)
       .ignoresProtect()
-      .condition(failOnBossCondition)
-      .unimplemented(),
+      .attr(DestinyBondAttr)
+      .condition(failOnBossCondition),
```
TempsRay, line 4968: What happens if it's a double battle against a boss and the the player has their left pokemon use destiny bond on their right pokemon? They'd still get the destiny bond tag which would let them instantly defeat bosses. I think the fail condition should be checked in the battlertag's apply

src/data/battler-tags.ts
```diff
@@ -254,6 +254,39 @@ export class ConfusedTag extends BattlerTag {
...
+   */
+  lapse(pokemon: Pokemon, lapseType: BattlerTagLapseType): boolean {
+    if (lapseType !== BattlerTagLapseType.CUSTOM) {
+      return super.lapse(pokemon, lapseType);
+    }
+    const source = pokemon.scene.getPokemonById(this.sourceId);
+    if (source.isFainted()) {
+      const targetMessage = getPokemonMessage(pokemon, '');
+      pokemon.scene.queueMessage(`${getPokemonMessage(source, ` took\n${targetMessage} down with it!`)}`)
+      pokemon.damageAndUpdate(pokemon.hp, HitResult.ONE_HIT_KO, false, false, true);
```
TempsRay, line 284: I was thinking of doing the boss check here since there might be battles where one enemy is a boss and the other one is not, in which case it seems like it'd make sense to me if destiny bond still works if the non-boss knocked out the destiny bond Pokemon.

JakubHanko: Let's say the destiny bond user is attacked by the boss first and then the non-boss Pokémon. Am I correct to assume that it should not trigger at all? Or should it kill the non-boss pokemon?

TempsRay: I believe destiny bond only tracks the attacker that knocked the Pokemon out. So if your pokemon was against opponent A (who is a boss) and B (who is not)

1. If A attacks, then B attacks and your Pokemon faints => Pokemon B faints
2. If B attacks, and then A attacks and your Pokemon faints => Nothing happens

If your A is not a boss then scenario 1 is unchanged but scenario 2 would result in A fainting.

JakubHanko: You are correct. Moreover, it seems like you cannot knock your ally out.

I will fix it soon.

JakubHanko: Fixed and added docs that should comply with the guidelines.

TempsRay (approved):
Code looks good to me

Issue #1101, opened 2024-05-18
I think my male Kirlia wanted to be a Gardevoir


Thesgtkirby:
Who's to stop him? It looks like it was an intended choice which strikes me as odd. Why would someone add something so specific as gender based evolutions to an evolution that wasn't based on pokemon gender? In every pokemon game since gen 3, a male Kirlia could still evolve into a Gardevoir. Maybe someone finds this very purposely designed mechanic to be annoying for some reason. I feel like a sane and reasonable person shouldn't have an issue with this unless they are petty in nature.

NotSalad-in:
Why are you only focusing on Gardevoir and not also mentioning Gallade being genderlocked to male?

Do you have a thing for Gardevoir?

CodeTappert:
Does Kirlia need an item to evolve to gallade currently (in pokerouge)

If not this might explain why is it set that way.
https://github.com/pagefaultgames/pokerogue/blob/6fa7412465b99d38944c124ec49d52b010d8d5c7/src/data/pokemon-evolutions.ts#L429

CodeTappert:
Okay i looked more at the code. 
I mean we could really implement it the way it is in the games. I dont see a code reason not to do  it....

I will look at it

guest1511:
> Okay i looked more at the code. I mean we could really implement it the way it is in the games. I dont see a code reason not to do it....
> 
> I will look at it

As another player that likes the pokemon. I ask that if that the evolution must change, make it where Gardevoir and Gallade are as controllable as they are now and not with the stone evolution. It being gender locked is an elegant mechanical solution instead of nerfing Gallade behind a stone evolution. Something to note is that the discord server before lockdown did discuss a possible solution being that of having a prompt on evolution for krilia that you could choose whether to have it evolve into Gallade/Gardevoir regardless of gender.

Xavion3:
To summarise why this was done, this and snorunt were changed into split gender evolutions as a buff to these pokemons usability as starters. Kirlia and snorunt are not great pokemon, both really want to evolve to be capable. With the mechanic in the games in addition to the gender you also need to find a dawn stone which meant if you wanted a froslass or gallade you'd often struggle immensely sometimes having to wait 100+ waves hoping to be able to evolve your pokemon. Evolution stone chances are higher than they're used to be, but it can still feel punishing having to wait so long. Kirlia's stats are *really* bad after all. 

[...]

PR #1107, opened 2024-05-18
Change Kirlia/Snorunt evolution methods to match main series

Fontbane:
Before evolution items were common, Kirlia and Snorunt simply had gender exclusive level up evolutions for convenience. As they're now easier to get, they'll instead use the Dawn Stone as in the main series.

### DIFF
--- a/src/data/pokemon-evolutions.ts
+++ b/src/data/pokemon-evolutions.ts
@@ -426,8 +426,8 @@ export const pokemonEvolutions: PokemonEvolutions = {
     new SpeciesEvolution(Species.KIRLIA, 20, null, null)
   ],
   [Species.KIRLIA]: [
-    new SpeciesEvolution(Species.GARDEVOIR, 30, null, new SpeciesEvolutionCondition(p => p.gender === Gender.FEMALE, p => p.gender = Gender.FEMALE)),
-    new SpeciesEvolution(Species.GALLADE, 30, null, new SpeciesEvolutionCondition(p => p.gender === Gender.MALE, p => p.gender = Gender.MALE))
+    new SpeciesEvolution(Species.GARDEVOIR, 30, null, null),
+    new SpeciesEvolution(Species.GALLADE, EvolutionItem.DAWN_STONE, null, new SpeciesEvolutionCondition(p => p.gender === Gender.MALE, p => p.gender = Gender.MALE), SpeciesWildEvolutionDelay.LONG)
   ],
   [Species.SURSKIT]: [
     new SpeciesEvolution(Species.MASQUERAIN, 22, null, null)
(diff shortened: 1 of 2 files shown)


CodeTappert:
As stated in https://github.com/pagefaultgames/pokerogue/issues/1101#issuecomment-2118914942 by Xavion3. 

The devs will discuss what should be done about it and how or if it should be changed. Please wait for that decision

src/data/pokemon-evolutions.ts
```diff
@@ -515,8 +515,8 @@ export const pokemonEvolutions: PokemonEvolutions = {
     new SpeciesEvolution(Species.DUSCLOPS, 37, null, null)
   ],
   [Species.SNORUNT]: [
-    new SpeciesEvolution(Species.GLALIE, 42, null, new SpeciesEvolutionCondition(p => p.gender === Gender.MALE, p => p.gender = Gender.MALE)),
-    new SpeciesEvolution(Species.FROSLASS, 42, null, new SpeciesEvolutionCondition(p => p.gender === Gender.FEMALE, p => p.gender = Gender.FEMALE))
+    new SpeciesEvolution(Species.GLALIE, 42, null, null),
+    new SpeciesEvolution(Species.FROSLASS, EvolutionItem.DAWN_STONE, null, new SpeciesEvolutionCondition(p => p.gender === Gender.FEMALE, p => p.gender = Gender.FEMALE), SpeciesWildEvolutionDelay.VERY_LONG)
```
CodeTappert, line 519: Also this is wrong. 
You added the evolution item to the slot for the level when it can evolve. You need to change this to 1 and then the other null right behind it to the stone

CodeTappert: Soo like

`    new SpeciesEvolution(Species.GALLADE, 1, EvolutionItem.DAWN_STONE, new SpeciesEvolutionCondition(p => p.gender === Gender.MALE, p => p.gender = Gender.MALE))`

guest1511:
The best solution for Snorunt and Krilia would be to make it a prompt on evolution when Raltz/Snorunt hit the appropriate levels, allowing male gardevoirs/frosslasses and female gallades, instead of unnecessarily nerfing frosslas/gallade by tagging them to stone evolution if a change from the gender based evolution is to be made. There are many other evolutions in pokerogue changed from base game to better fit Pokerogue's gameplay and mechancis. If we're trying to go for 1-to-1 accuracy, why not change Runerigus to evolve near temple (_note I think this is an awful idea and shouldn't be done whatsoever_) to match closer to how it evolves in the mainline game?

src/data/pokemon-evolutions.ts
```diff
@@ -426,8 +426,8 @@ export const pokemonEvolutions: PokemonEvolutions = {
     new SpeciesEvolution(Species.KIRLIA, 20, null, null)
   ],
   [Species.KIRLIA]: [
-    new SpeciesEvolution(Species.GARDEVOIR, 30, null, new SpeciesEvolutionCondition(p => p.gender === Gender.FEMALE, p => p.gender = Gender.FEMALE)),
-    new SpeciesEvolution(Species.GALLADE, 30, null, new SpeciesEvolutionCondition(p => p.gender === Gender.MALE, p => p.gender = Gender.MALE))
+    new SpeciesEvolution(Species.GARDEVOIR, 30, null, null),
+    new SpeciesEvolution(Species.GALLADE, null, EvolutionItem.DAWN_STONE, new SpeciesEvolutionCondition(p => p.gender === Gender.MALE, p => p.gender = Gender.MALE), SpeciesWildEvolutionDelay.LONG)
```
CodeTappert, line 430: Why null for the level? I told you to set it to 1.

Fontbane: My bad...

CodeTappert (approved):
Code changes are made correctly. 

They can be merged once the decision is made

bennybroseph:
I don't see a reason to make evolutions more complicated when they are already functioning. I agree that the only real improvement to be made is to have an evolution selection for branching evolutions like this. Adding a stone just makes the process worse regardless of how it functions in the mainline series.

Fontbane:
> I don't see a reason to make evolutions more complicated when they are already functioning. I agree that the only real improvement to be made is to have an evolution selection for branching evolutions like this. Adding a stone just makes the process worse regardless of how it functions in the mainline series.

Is this the consensus of the development team, or your personal opinion on the matter? This topic is currently being discussed and options are being weighed. If you wanted to close it until we figured out how/if we want to address this, I would understand, but you didn't mention that. With all due respect, is your own personal opinion on the matter justification enough to reject the PR?

PR #1108, opened 2024-05-18
Kirlia/Snorunt evolution

Rosalie-A:
In the core games, Kirlia and Snorunt evolve into Gallade and Froslass when a specific gender and when exposed to a stone; otherwise, they evolve into Gardevoir/Glalie. This PR changes this behavior to match the main series.

### DIFF
--- a/src/data/pokemon-evolutions.ts
+++ b/src/data/pokemon-evolutions.ts
@@ -426,8 +426,8 @@ export const pokemonEvolutions: PokemonEvolutions = {
     new SpeciesEvolution(Species.KIRLIA, 20, null, null)
   ],
   [Species.KIRLIA]: [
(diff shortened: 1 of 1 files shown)


guest1511:
The best solution for Snorunt and Krilia would be to make it a prompt on evolution when Raltz/Snorunt hit the appropriate levels, allowing male gardevoirs/frosslasses and female gallades, instead of unnecessarily nerfing frosslas/gallade by tagging them to stone evolution if a change from the gender based evolution is to be made. There are many other evolutions in pokerogue changed from base game to better fit Pokerogue's gameplay and mechanics. If we're trying to go for 1-to-1 accuracy, why not change Runerigus to evolve near temple (note I think this is an awful idea and shouldn't be done whatsoever) to match closer to how it evolves in the mainline game?

(_copy and pasted from someone else's commit for the same change_)

CodeTappert:
Duplicate of https://github.com/pagefaultgames/pokerogue/pull/1107

PR #1114, opened 2024-05-18
[QoL] Automatically center window vertically when playing on landscape 

torranx:
### PLEASE REFER [HERE](https://github.com/pagefaultgames/pokerogue/pull/1114#issuecomment-2128087672) FOR UPDATED DESCRIPTION

--------

This PR includes changes from [851](https://github.com/pagefaultgames/pokerogue/pull/851)

Only added the check if window width < 1920 - this should prevent applying the `align-items: center` property when the game window is scaled up.

There is another cleaner approach to achieving this through phaser but this would _require a reload_. If forcing a reload is okay, then that approach is better, otherwise this is.

- width < 1920
Screenshot 2024-05-19 at 5 46 38 AM

- width >= 1920
Screenshot 2024-05-19 at 5 49 38 AM

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -27,12 +27,6 @@ window.addEventListener("unhandledrejection", (event) => {
   //alert(errorString);
 });
 
-// Reinitialize game with the updated config.scale.autoCenter when 'touchControlsChange' event is dispatched
-window.addEventListener("touchControlsChange", () => {
-  config.scale.autoCenter = shouldCenterWindow() ? Phaser.Scale.CENTER_VERTICALLY : Phaser.Scale.NO_CENTER;
-  reinitializeGame();
-});
-
 /**
  * Determines if the game window should be centered based on the current device and settings.
  *
@@ -46,7 +40,7 @@ window.addEventListener("touchControlsChange", () => {
  */
 const shouldCenterWindow = (): boolean => {
   const touchControlsOptions = Setting.find(s => s.key === SettingKeys.Touch_Controls).options;
-  console.log(touchControlsOptions)
+  console.log(touchControlsOptions);
   const settings = localStorage.hasOwnProperty("settings") ? JSON.parse(localStorage.getItem("settings")) : null;
 
   const isLandscape = window.matchMedia("(orientation: landscape)").matches;
@@ -190,18 +184,6 @@ let game: Phaser.Game;
 const startGame = () => {
   game = new Phaser.Game(config);
   game.sound.pauseOnBlur = false;
-};
-
-/**
- * Reinitializes the game by destroying the current instance and starting a new one.
- * @function reinitializeGame
- * @returns {void}
- */
-const reinitializeGame = (): void => {
-  if (game) {
-    game.destroy(true, false);
-    startGame();
-  }
 };
 
 fetch("/manifest.json")
--- a/src/main.ts
+++ b/src/main.ts
@@ -29,7 +29,7 @@ window.addEventListener("unhandledrejection", (event) => {
 
 // Reinitialize game with the updated config.scale.autoCenter when 'touchControlsChange' event is dispatched
 window.addEventListener("touchControlsChange", () => {
-  config.scale.autoCenter = shouldCenterWindow() ? Phaser.Scale.CENTER_BOTH : Phaser.Scale.NO_CENTER;
+  config.scale.autoCenter = shouldCenterWindow() ? Phaser.Scale.CENTER_VERTICALLY : Phaser.Scale.NO_CENTER;
   reinitializeGame();
 });
 
@@ -65,7 +65,7 @@ const config: Phaser.Types.Core.GameConfig = {
     width: 1920,
     height: 1080,
     mode: Phaser.Scale.FIT,
-    autoCenter: shouldCenterWindow() ? Phaser.Scale.CENTER_BOTH : Phaser.Scale.NO_CENTER
+    autoCenter: shouldCenterWindow() ? Phaser.Scale.CENTER_VERTICALLY : Phaser.Scale.NO_CENTER
   },
   plugins: {
     global: [{
--- a/src/main.ts
+++ b/src/main.ts
@@ -40,7 +40,6 @@ window.addEventListener("unhandledrejection", (event) => {
  */
 const shouldCenterWindow = (): boolean => {
   const touchControlsOptions = Setting.find(s => s.key === SettingKeys.Touch_Controls).options;
-  console.log(touchControlsOptions);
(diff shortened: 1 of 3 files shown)


bennybroseph:
> There is another cleaner approach to achieving this through phaser but this would require a reload. If forcing a reload is okay, then that approach is better, otherwise this is.

When would the reload occur?

torranx:
Sorry forgot to mention on PR description. Reload will happen _when changing Touch Controls settings_, this is because:
- we need to include the check whether the touch controls is enabled or not
- phaser does not allow us to update game config during runtime (as per my research), so we must do it before initialization, hence the reload

Odizinne:
I'm trying to do the same in my electron app for days.

Setting the alignment to center break the login page. The input fields position will change depending of the current resolution.

Also it introduce overflow above 1080p

This is at 1600p
image

and this is 800p
image

Odizinne:
If that helps, from a user perspective i guess triggering a relaod one time when setting an option is not really an issue

torranx:
I just wanna clarify so we're all on the same page, the reload is an actual page reload and not a scene reset (as it is now).

torranx:
> EDIT: Nvm i just tried your branch, and i cannot replicate the overflow issue. It might be on my side.
> 
> But i cannot manage to get the window centered with your branch either (touch disabled, display set at 2560x1600)

Oh just saw this, i thought what made the overflow was my changes so I reverted them lol. 

You should set Desktop (touch) in your devtools too

torranx:
Confirming the overflow on login page, this only happens on viewports < 1920. Will update my branch with the phaser solution

Screenshot 2024-05-23 at 3 22 23 PM

torranx:
Since messing with the CSS _messed_ up a part of the UI (login phase), and the reverted PR #851 also encountered issues on certain viewport sizes, I've decided to update the game window through Phaser config's [autoCenter](https://newdocs.phaser.io/docs/3.55.2/focus/Phaser.Core.Config-autoCenter) instead.
**Note: the trade-off here is we have to require reload when changing touch controls setting.**

Tests done were a success but would appreciate other tests as well.

For reference (please take note of the device type on the upper portion of the recordings):
**Non-mobile devices**
- Portrait (Changing touch controls has no effect on alignment)
https://github.com/pagefaultgames/pokerogue/assets/68144167/b3d06e69-62ac-4d9b-b23a-9add7ee27339

- Landscape (Changing touch controls has effect: Disabled -> centers the game window; Enabled -> no center)
https://github.com/pagefaultgames/pokerogue/assets/68144167/ed2105a7-e403-4e29-9318-096c0cf9abe1

- Desktop (no effects when changing touch control settings)
https://github.com/pagefaultgames/pokerogue/assets/68144167/7d867fa4-c889-49e9-8951-897be2d30cb7

**Mobile devices** (Current behavior applies - you cannot disable touch controls)

No overflow on login phase also:
https://github.com/pagefaultgames/pokerogue/assets/68144167/828f0f8f-c2b9-40e3-9c1b-46828d932419

Odizinne:
Following this logic, on desktop with no touch controls shouldnt the window be centered by default?

torranx:
> Following this logic, on desktop with no touch controls shouldnt the window be centered by default
> 
> If you have no touch control to show in any situation, i guess it make more sense.

Yeah you do actually have a point but this was initially introduced to enhance the UX on handheld devices, so desktops were really not put into consideration and since most of larger devices already use 16:9.

I'm open to remove that condition but I'm also considering the UX on the small number of players accustomed/conditioned to play with what we have right now. I think we'd only want to remove that condition if we get good traction or a lot of users are calling for that change. I actually asked the community on discord and only got 2 replies, 1 yes for mobile (which we do not want), 1 yes generally lol.

torranx:
> It seems logic to me that user with no touchscreen get the same behavior.

I think this is slightly different from our situation/issue now. If that's the case, we want to center the game automatically on non mobile devices with no touchscreen capabilities and not apply the logic of auto aligning the game window when touch controls are enabled/disabled.

> Is there any benefits to not center the window on 16:10 screens without touch screen?

This will entirely depend on user preference.

Odizinne:
If we're going back to original issue #810, creating an ingame setting for this is not the way to go.

Also this issue was not opened with only handhelds in mind, but i think you're right, those are 2 different issues.

So the final goal would be to center the window on desktop / no touchscreen 16:10 devices, and let the touch control option decide for other touchscreen devices that are not mobile?

| Device | Touch control setting | Result |
| --- | --- | --- |
| SteamDeck  | ON | Not centered
| SteamDeck  | OFF | Centered
| 16:10 Laptop with no touchscreen | OFF | Centered by default


Is that correct?

Sorry had a bad night i'm a lil slow to understand right now!

torranx:
> If we're going back to original issue #810, creating an ingame setting for this is not the way to go.

We're not creating another an in game settin.

> So the final goal would be to center the window on desktop / no touchscreen 16:10 devices, and let the touch control option decide for other touchscreen devices that are not mobile?

Yes, but the PR does not address desktop-no touchscreen capability yet.

Odizinne:
Gotcha, 2 different issues

torranx:
I'll try and add this on my PR since there is some dependencies present.

Odizinne:
Thank you for the time taken for explanation

torranx:
Please refer to the following comments on what are the changes/tests made:
- [initial changes/tests](https://github.com/pagefaultgames/pokerogue/pull/1114#issuecomment-2128087672), refer to bullet points
- [auto-centering on desktop by default](https://github.com/pagefaultgames/pokerogue/pull/1114#issuecomment-2130207998)
- [changed centering to vertical only](https://github.com/pagefaultgames/pokerogue/pull/1114/commits/5002a0b1b9ae7683e20fea41db3fc37a10673785)

This is ready for review.

brain-frog:
This would need to be thoroughly tested just to make sure the re-initialization doesn't cause any problems but I think its a good change for mobile players
please get the merge conflicts sorted and let us know when that happens so I can ask for it to be tested

torranx:
> This would need to be thoroughly tested just to make sure the re-initialization doesn't cause any problems but I think its a good change for mobile players please get the merge conflicts sorted and let us know when that happens so I can ask for it to be tested

This is ready for testing now. Recent commits also removed the reinitialization and instead replaced with a page reload - because after the updates on settings and on the `main.ts` file, the reinitialization isn't working anymore and causes the game to crash.

> also in the future please just update the PR description instead of linking to a comment with a new one so its easy to understand the current state of the PR when opening it

Noted on this, I apologize for the messy flow of explanations on this thread.

torranx:
Closing this as @Dakurei have found a better approach on scaling/centering the game (he's yet to open the PR). Please refer to this [discussion](https://discord.com/channels/1125469663833370665/1249827339014115359).

cc @brain-frog

Issue #993, opened 2024-05-16
Download does not account for vitamins


Tempo-anon:
![](https://cdn.discordapp.com/attachments/1240694435952463932/1240694436413968454/image.png?ex=66477e3c&is=66462cbc&hm=afc49657adbaabe1a3b75208aeacf7d83728e3f26e25626421a94b0bb67309a0&) ![](https://media.discordapp.net/attachments/1240694435952463932/1240694436929994822/image.png?ex=66477e3c&is=66462cbc&hm=7cd99f077fea83f224bc38e1e9aa0f6bedeef01edd8dd6e8d7540f74133fe830&=&format=webp&quality=lossless&width=1080&height=585)

TempsRay:
not merged in yet

EvasiveAce:
Issue can be closed now.

PR #1119, opened 2024-05-18
Download Ability Fix

EvasiveAce:
# Fixes Issue #993 

Please let me know if there is anything I could improve.

Default higher SpDef:

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/636d8580-47ee-4d8b-bd55-06ea0485d883
)

Using vitamins to increase Def

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/11a77bf1-dbf7-4c90-8f2d-d0f013de11d0)

Default higher Def:

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/11d0296b-5774-41b7-9fc8-9c6822113623)

Using vitamins to increase SpDef

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/6384c72e-f9d1-496b-8b9d-e5b676168712)


src/data/ability.ts
```diff
@@ -3139,7 +3156,7 @@ export function initAbilities() {
       .attr(TypeImmunityHealAbAttr, Type.WATER)
       .ignorable(),
     new Ability(Abilities.DOWNLOAD, 4)
-      .attr(DownloadAbAttr),
+      .attr(DownloadAbAttr, true),
```
TempsRay, line 3159: Is the boolean being passed in here `passive`? The doc for the class says it's being usused

src/data/ability.ts
```diff
@@ -1529,19 +1530,35 @@ export class PostSummonClearAllyStatsAbAttr extends PostSummonAbAttr {
...
+ * @extends PostSummonAbAttr
+ * @see {applyPostSummon}
+ */
 export class DownloadAbAttr extends PostSummonAbAttr {
   private enemyDef: integer;
   private enemySpDef: integer;
   private stats: BattleStat[];
-
+  
+  /**
```
TempsRay, line 1545: Maybe should add a TODO for it ignoring substitute'd pokemon

src/data/ability.ts
```diff
@@ -1529,19 +1530,35 @@ export class PostSummonClearAllyStatsAbAttr extends PostSummonAbAttr {
...
+   * Checks to see if it is the opening turn (starting a new game), if so, Download won't work. This is because Download takes into account
```
TempsRay, line 1546: Maybe can add an edge case for the first turn of the game so it still works? I assume the intended behavior will still want Download activating on the first floor

EvasiveAce: I tried, but I learned its for like almost every postsummonattr. I will post video proof with intimidate but with Download it should be fine since there aren't any Download abilities in Town.

Let me say: going from 20->21, this bug isn't here. It's literally (from what I gathered) when a new game is created. 

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/14819e90-9b9c-4f33-b508-7284fd00594e)

I think it happens when starting a new game because when the ability procs, the Pokemon, hasn't been summoned at least once.

src/data/ability.ts
EvasiveAce: Fixed, not sure why I set that.

src/data/ability.ts
```diff
@@ -1529,19 +1530,36 @@ export class PostSummonClearAllyStatsAbAttr extends PostSummonAbAttr {
...
   applyPostSummon(pokemon: Pokemon, passive: boolean, args: any[]): boolean {
     this.enemyDef = 0;
     this.enemySpDef = 0;
-	
-    for (let opponent of pokemon.getOpponents()) {
-      this.enemyDef += opponent.stats[BattleStat.DEF];
-      this.enemySpDef += opponent.stats[BattleStat.SPDEF];
-    }
+
+    if(pokemon.getOpponents()[0].summonData != undefined)
```
TempsRay, line 1558: Is accessing zeroth index of something and checking if it's undefined safe? Should also be `!==` for conssitency

TempsRay (approved):
Code looks good to me

src/data/ability.ts
EvasiveAce: It's basically to fix the issue we had above, if pokemon.getOpponents()[0] (opposing pokemon) is undefined, it will not work, and prevent a crash on a new game. I will fix the !==.

EvasiveAce:
Edgecase for Double Download users and single mon

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/90851414-1644-471c-b163-779697f2ce76)

Tempo-anon (approved):

PR #1121, opened 2024-05-19
Implemented the Imprison move

audriddax:
# Summary

I've been enjoying this game, and wanted to start helping out. The first non-implemented move I encountered after this was Imprison, so I took a stab at implementing it. I plan to keep making PR's for other moves I encounter.

Full disclosure, I've never used Imprison in the actual games, so my implementation is based on my research, I think I covered all the nuances of it. I've included all my test scenarios in case there is behavior that isn't actually expected.

# Tests

### Overrides
 
```javascript
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.EMBER, Moves.ACID, Moves.RECOVER, Moves.IMPRISON];
export const OPP_MOVESET_OVERRIDE: Array<Moves> = [Moves.EMBER, Moves.ACID, Moves.LEER];
```

### Party

1. Bulbasaur
2. Charmander

### Tests
* If the imprisoning pokemon wins the battle and the next battle is a wild encounter, imprison is still in effect.
* If the imprisoning pokemon is swapped out, imprison no longer has an effect, even if the imprisoning pokemon is brought back into battle.
* When a trainer battle started, imprisoned was reset, so the trainer could use any move until imprison was used again.
* After a boss battle, imprisoning is reset, so the player needs to use it again.

### Overrides
 
```javascript
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.EMBER, Moves.ACID, Moves.RECOVER, Moves.IMPRISON];
export const OPP_MOVESET_OVERRIDE: Array<Moves> = [Moves.EMBER, Moves.ACID];
```

### Party

1. Bulbasaur
2. Charmander

### Tests
* If the imprisoning pokemon moves first, the opposing pokemon cannot attack since the move it chose is imprisoned when it attacks.
* Once imprisoned is used, the opponent can only use struggle.

## Test Scenario 3: Imprisoning Pokemon faints
 
```javascript
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.EMBER, Moves.IMPRISON, Moves.AGILITY, Moves.LEER];
export const OPP_MOVESET_OVERRIDE: Array<Moves> = [Moves.EMBER, Moves.BUBBLE, Moves.AGILITY, Moves.LEER];
```

### Party

1. Bulbasaur
2. Charmander

### Tests
* Once imprisoned is used, I let Bulbasaur faint, won the battle with Charmander, revived Bulbasaur, then swapped to Bulbasaur for the next batter. Imprisoned was no longer active.


src/field/pokemon.ts
```diff
@@ -3427,8 +3427,16 @@ export class PokemonMove {
     this.virtual = !!virtual;
   }
 
+  #isImprisoned(pokemon: Pokemon) {
```
JakubHanko, line 3430: I suppose the hashtag is an oversight?

audriddax: It's to make it a private function, so it's only callable from this class. I didn't want to bog down the isUsable function with a bunch of logic, so I abstracted it into a private function to make isUsable more readable.

I'm good moving this directly into the isUsable function if that's preferable, this is just my default habit to assist with my interpretation of self-documenting code.

audriddax: For reference on the hashtag:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties#private_methods

Let me know if there's anything else I can help with!

JakubHanko: So it's my lack of knowledge then. Thanks for the explanation!

audriddax: no problem!

tbh...I don't often makes classes in JS, so I always need to lookup how to make private methods when I need it. It's not intuitive.

audriddax: I kinda wish typescript implemented some "private" modifier for a function that would compile to the hashtag

Xavion3: Doesn't the standard private access modifier work on functions?

audriddax: as far as I know, that's not a thing in javascript, but if I'm wrong, please correct me! I would love for that to be a thing, it's so much clearer than the hashtag!

bennybroseph: this is written in typescript, and `private myFucntion()` is valid syntax.

audriddax: omg...how have I not known about this?!?!

I updated the code to use the typescript private syntax.

src/data/enums/battler-tag-type.ts
```diff
@@ -56,5 +56,6 @@ export enum BattlerTagType {
   CHARGED = "CHARGED",
   GROUNDED = "GROUNDED",
   MAGNET_RISEN = "MAGNET_RISEN",
-  MINIMIZED = "MINIMIZED"
+  MINIMIZED = "MINIMIZED",
+  IMPRISONING = "IMPRISONING",
```
bennybroseph, line 60: Should this be IMPRISONED?

bennybroseph (changes requested):

src/data/enums/battler-tag-type.ts
audriddax: I originally did that, but it seemed like the attacker isn't the one imprisoned, they are imprisoning the opposing pokemon.

Do you want me to change that?

bennybroseph: I'm not sure, this just feels off. Have you looked at other moves like this such as Safeguard/Light Screen/etc.? If we have those implemented, I would check and see what we did in those cases and report back. This seems similar as it interacts with the opponents moves but is on the user's side of the field.

audriddax: I'll be honest, I'm in the same boat, it does feel off. To me, it just felt more off to tag the pokemon using it as IMPRISONED since it's really the opponents that are IMPRISONED.

I'll take a look at those moves you listed, thanks for the references!

audriddax:
> Looking to get JSDoc documentation as shown here: [Code Comment FAQ](https://discord.com/channels/1125469663833370665/1239957703820836975)

I have no idea why I didn't do that, I usually work in vanilla js at work and do that for intellisense. My best guess is I just got spoiled by the auto-intellisense of typescript with typing and forget.

I'll get this added!

audriddax:
> Looking to get JSDoc documentation as shown here: [Code Comment FAQ](https://discord.com/channels/1125469663833370665/1239957703820836975)

@bennybroseph ok, I added jsdoc comments, thanks for the callout on that!

src/data/enums/battler-tag-type.ts
audriddax: So I looked at those, and it looks like they aren't quite the same since they apply to an entire side, whereas imprison applies to a specific pokemon. That being said, I think I'm even more confused now on the naming, I'm seeing two different approaches.

For example, safeguard just tags as SAFEGUARD, that's the verb/action being referenced. If you look at other tags just above the new one, you see MINIMIZED and GROUNDED. So it looks like some tags are phrased as "I will" (I will SAFE GUARD) and others are "I am" (I am MINIMIZED).

I can go either way on this, but where I'm at is it should be either IMPRISON or IMPRISONING, but not IMPRISONED.

bennybroseph (changes requested):
Okay, final things. 
- Need a video showing this working. 
- What is your source for the messages used for this move?
- Let's just use IMPRISON

audriddax:
> Okay, final things.
> 
> * Need a video showing this working.
> * What is your source for the messages used for this move?
> * Let's just use IMPRISON

@bennybroseph:
* Ok, that'll take me a bit, I just rebuilt my PC and haven't installed recording software. Once I've got that, is a youtube video what you're looking for?
* I looked up videos of imprison in action, here's a link to one that shows the verbiage: https://youtu.be/Z2-a9YmRIjw?si=iEovHDv54g4ponyR&t=620
* Changed to IMPRISON, thanks!

audriddax:
> > Okay, final things.
> > 
> > * Need a video showing this working.
> > * What is your source for the messages used for this move?
> > * Let's just use IMPRISON
> 
> @bennybroseph:
> 
> * Ok, that'll take me a bit, I just rebuilt my PC and haven't installed recording software. Once I've got that, is a youtube video what you're looking for?
> * I looked up videos of imprison in action, here's a link to one that shows the verbiage: https://youtu.be/Z2-a9YmRIjw?si=iEovHDv54g4ponyR&t=620
> * Changed to IMPRISON, thanks!

@bennybroseph ok, got a video together where I show all my testing scenarios:
https://www.youtube.com/watch?v=cPqg0rt6k-Y

src/data/battler-tags.ts
```diff
@@ -1145,6 +1145,54 @@ export class MagnetRisenTag extends TypeImmuneTag {
...
+/**
+ * A tag used to identify when a pokemon has used imprison since it was summoned.
+ * @see {@linkcode onAdd}
+ * @see {@linkcode onOverlap}
+ * @see {@linkcode lapse}
+ * @see {@linkcode getDescriptor}
+ */
+export class ImprisonTag extends BattlerTag {
+  constructor(sourceMove: Integer) {
+    super(BattlerTagType.IMPRISON, BattlerTagLapseType.CUSTOM, Number.MAX_SAFE_INTEGER, sourceMove);
```
bennybroseph, line 1157: Last thing, I'm confident this works, but I just can't get over this line with `Number.MAX_SAFE_INTEGER`. Don't get me wrong, I know exactly why it's done, but I really don't like it. Can we add a property called `infiniteDuration` or `ignoreTurnCount` or something and ignore the decrementing turn count so that we don't have to use this workaround?

bennybroseph (changes requested):

src/data/battler-tags.ts
audriddax: Yeah, I'll get that updated. I wasn't a big fan of it either. Since this was my first commit to the project, I didn't want to do _to_ much to rock the boat until I had a better handling of the general coding style here.

audriddax: So, I noticed there is this check in the Pokmeon.lapseTags() function:
```javascript
tags.filter(t => lapseType === BattlerTagLapseType.FAINT || ((t.lapseType === lapseType) && !(t.lapse(this, lapseType))) || (lapseType === BattlerTagLapseType.TURN_END && t.turnCount < 1))
```

If I'm reading this correctly, either I need to:
1. Still set the `turnCount` to something positive, like 1
1. Add a check for the new property `ignoreTurnCount` here

I'm not a fan of option 1 because that is still deceptive, but I'm also hesitant to add to the line above because now my change has broader effects on other moves (though it _shouldn't_ make a difference).

I'm leaning towards option 2, adding the `ignoreTurnCount` check as I think that plus a `turnCount = 0` property is the most descriptive and accurate, and then I'll just have to do some spot checking of other moves, but I wanted your opinion on this.

Which approach are you more comfortable with me implementing?

bennybroseph: It is more destructive, but I think option 2 is needed. I'll take a look after you make the changes to see how it looks. We may run into more problems than it seems, but I'd like to try anyway and can always go back to max int if we absolutely have to.

audriddax: Sounds good!

I don't have time to do testing right now, but I pushed the changes so you can review them in case something sticks out when you see it.

I'll try to get testing done tomorrow!

src/overrides.ts
bennybroseph, line 1: Make sure this is removed from commits

bennybroseph (changes requested):

src/overrides.ts
audriddax: Oh wow, can't believe I let that through. Thanks for catching it!

bennybroseph:
Is this still being worked on?

audriddax:
> Is this still being worked on?
sorry, was busy the last couple of weeks, gonna clean this up now!

audriddax:
@bennybroseph -
Any suggestions on what to test outside of Imprison with the new change? Any specific moves you think should be tested, or should I just play for a bit and see if anything weird happens?

Tempo-anon:
Closing due to inactivity and merge conflicts

PR #1122, opened 2024-05-19
Changed How is local is checked and how serverURL is build to prevent…

### LINKED ISSUE #1078: Locally Hosted Configuration has CORS conflict with API
Locally Hosted in this context refers to running from source, using `npm run start:dev`

It's been reported already in discord that the game fails to run the `GameOverPhase`, although the source of the issue was unknown.  I ran into this issue and identified the source/conflict preventing it from loading properly.

Essentially the game is attempting to run the `api.pokerogue.net`, specifically: `Access to fetch at 'https://api.pokerogue.net/savedata/newclear?slot=0' `

The source of this is caused when the individual connects via the Network option instead of localhost.  This causes the `utils.ts` to not recognize it as being run locally, specifically `localhost`

To resolve this issue locally I modified the `src/utils.ts` line 218 to include my hosted IP address in the secondary option of the `window.location.hostname`

This example uses `10.10.10.10` as my hosted IP
```
export const sessionIdKey = 'pokerogue_sessionId';
export const isLocal = window.location.hostname === 'localhost' || window.location.hostname === '10.10.10.10';
export const serverUrl = isLocal ? 'http://localhost:8001' : 'http://10.10.10.10:8001';
export const apiUrl = isLocal ? serverUrl : 'https://api.pokerogue.net';
```

CodeTappert:
… errros in offline mode

Fixes #1078


Admiral-Billy (changes requested):
Not having the `window.location.hostname === ''` check breaks the offline detection for local files, such as from my app :(

CodeTappert:
> Not having the `window.location.hostname === ''` check breaks the offline detection for local files, such as from my app :(

You mean for compiled files? Because this change works fine when run via "npm start dev".

Even if you serve dist folder it should work without problems

Admiral-Billy:
This is indeed back to working. Also yeah, my app is serving compiled files and registers as an empty string for that check for whatever reason; I don't know the details behind it, but that is what happens for it. My Android apk registers normally as localhost though.

CodeTappert:
> This is indeed back to working. Also yeah, my app is serving compiled files and registers as an empty string for that check for whatever reason; I don't know the details behind it, but that is what happens for it. My Android apk registers normally as localhost though.

I readded the check for you

Admiral-Billy:
Yep, that's what I was talking about; it works fine for my app's particular offline now. Now it's just up to the actual devs to see what they think of this change for their purposes heh

athias:
As there were minor changes to the code, I tested this again and can confirm it fixes #1078

Admiral-Billy (approved):
As far as I can tell, this works perfectly fine for both the dev server and my app.

CodeTappert:
This PR is ready to be reviewed/merged by a dev.

CodeTappert:
> This PR is ready to be reviewed/merged by a dev.

Walker says this is fine. See https://discord.com/channels/1125469663833370665/1243948971110694982

src/utils.ts
```diff
@@ -232,8 +232,15 @@ export function executeIf<T>(condition: boolean, promiseFunc: () => Promise<T>):
...
 
 export const sessionIdKey = "pokerogue_sessionId";
-export const isLocal = window.location.hostname === "localhost" || window.location.hostname === "";
-export const serverUrl = isLocal ? "http://localhost:8001" : "";
+// Check if the current hostname is 'localhost' or an IP address, and ensure a port is specified
+export const isLocal = (
+  (window.location.hostname === "localhost" ||
+        /^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$/.test(window.location.hostname)) &&
+    window.location.port !== ""
+) || window.location.hostname === "";
```
bennybroseph, line 240: This needs some better formatting, it looks pretty bad. Please use any kind of consistency with where you're splitting up the line

bennybroseph (changes requested):

src/utils.ts
CodeTappert: To be fair i wouldnt split the line at all....
But then it is pretty long.

bennybroseph (changes requested):
This is just a different kind of bad

CodeTappert:
Now it is consistent and breaks after || or && if it would be to long without the break

bennybroseph (approved):

PR #1128, opened 2024-05-19
Reduce EXP Balance

bennybroseph:
The EXP Balance item is not currently functioning as a useful item. This reduces the item's Max Stack Count temporarily until a better solution can be found. Saves with this item will have them reduces to a max of 2 and shops will no longer carry them once you reach this value.

### With EXP Balance

https://github.com/pagefaultgames/pokerogue/assets/13838608/af4b8ffe-1477-432e-8087-71cfc423e18e

### Without EXP Balance

https://github.com/pagefaultgames/pokerogue/assets/13838608/116b0094-d7ee-43dd-b410-0cacd4df0ead

### DIFF
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -1280,7 +1280,6 @@ const modifierPool: ModifierPool = {
     new WeightedModifierType(modifierTypes.IV_SCANNER, 4),
     new WeightedModifierType(modifierTypes.EXP_CHARM, 8),
     new WeightedModifierType(modifierTypes.EXP_SHARE, 12),
-    new WeightedModifierType(modifierTypes.EXP_BALANCE, 4),
     new WeightedModifierType(modifierTypes.TERA_ORB, (party: Pokemon[]) => Math.min(Math.max(Math.floor(party[0].scene.currentBattle.waveIndex / 50) * 2, 1), 4), 4),
     new WeightedModifierType(modifierTypes.VOUCHER, (party: Pokemon[], rerollCount: integer) => !party[0].scene.gameMode.isDaily ? Math.max(3 - rerollCount, 0) : 0, 3),
   ].map(m => { m.setTier(ModifierTier.ULTRA); return m; }),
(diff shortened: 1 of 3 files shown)


Greenlamp2 (approved):

Greenlamp2:
i've created a save with 5 exp balance, than i've applied your changes, as soon as we load the save, we can already see that the exp balance is no longer there.


Fun fact, if you start a save with 5 exp balance, you got -145 exp on the first wave lmao
image

Gammer0909:
W-Why?

Gammer0909 (changes requested):
This is an unneeded PR, and the item is implemented correctly, therefore there is no need to remove it.

TempsRay:
Straight up just deleting the item? Not sure about that. Could set the weight to 0 for now just to stop it spawning, but ideally I would like to see this reworked completely to be just a positive like "your lowest leveled pokemon gains +10% (per stack) more exp"

MooGoesCow21:
> This is an unneeded PR, and the item is implemented correctly, therefore there is no need to remove it.

Its implemented correctly but as it is atm its a Trap item that sabotages runs

Gammer0909:
> > This is an unneeded PR, and the item is implemented correctly, therefore there is no need to remove it.
> 
> 
> 
> Its implemented correctly but as it is atm its a Trap item that sabotages runs

Then... dont grab the item and reroll? I'm not sure what the issue is here. If you don't need the item, *then don't take it*.

bennybroseph:
The item can't be ignored as it appears in the shop until you have the maximum amount of it. This items does not prevent this:
> 200, 176, 140, 109, 190, 107
That's what EXP Share is already for.

Admiral-Billy:
I think a better solution in the meantime would just be capping it to 4 max instead of 5, or changing it to no longer have a downside but make the upside smaller.

Xavion3 (approved):

zizibaloob:
This change was a significant reduction to the quality of life of my runs. All party members would be at the level cap, and then all party members would immediately grow to the level cap as soon as it was increased.

Now only my party member holding eggs hits the level cap, at which point it stops contributing to the total xp pool (thereby eliminating the benefit of all the eggs). It is now significantly more difficult to keep all party members at the level cap.

Please revert this change.

PR #1131, opened 2024-05-19
Implement Zero to hero

XGSleepWalker:
Context for this PR:

- Implements Zero to Hero for Palafin
- Fixes permanent form for all pokemon (forms should be reset on biome change or trainer battle, essentially on forced recall)
- Fixes PreSwitchOut attributes being erroneously triggered on the forced recall. 
- Changes PostBattleInit attributes to only being triggered on the forced recall.
- Forces the following abilities to be reverted on forced recall: Zen Mode, Shields Down, Schooling, Disguise, Battle Bond, Power Construct, Zero to Hero
- Marks Ogerpon's Embody Aspect as partial. (It will be reworked later on when we go back to rework on the Terastallization mechanics)
- Adds a new bypass for abilities to be forcefully triggered. (Form abilities shouldn't be triggered on death, but they should be able to be reverted regardless on forced recall).


src/overrides.ts
```diff
@@ -54,8 +54,8 @@ export const STARTING_LEVEL_OVERRIDE: integer = 0;
  * default is 0 to not override
  * @example SPECIES_OVERRIDE = Species.Bulbasaur;
  */
-export const STARTER_SPECIES_OVERRIDE: Species | integer = 0;
-export const ABILITY_OVERRIDE: Abilities = Abilities.NONE;
+export const STARTER_SPECIES_OVERRIDE: Species | integer = Species.PALAFIN;
```
JakubHanko, line 57: The overrides shouldn't be changed IMHO

src/data/ability.ts
```diff
@@ -1702,6 +1702,39 @@ export class PreSwitchOutHealAbAttr extends PreSwitchOutAbAttr {
...
+  private formFunc: (p: Pokemon) => integer;
+
+  constructor(formFunc: ((p: Pokemon) => integer)) {
+    super();
+
+    this.formFunc = formFunc;
+  }
+
+  /**
+   *
```
bennybroseph, line 1720: either need doc here or a removal of this space

src/overrides.ts
bennybroseph, line 1: need to remove this commit

bennybroseph (changes requested):
Can we get a video of how this was tested?

LaukkaE:
- [ * ]  Palafin switches to the correct form, and also back correctly.

- [  * ] Works with Self switching moves, as well as opponents force switching moves.

- [  * ]  If form is fusioned with another pokemon, stays in the form indefinitely while fusioned.

- [  * ] No console Errors found while testing.

- [  * ] If another pokemon that has another form should somehow acquire Zero to Hero, the ability will announce itself, but not change the form. 

- [  * ] Not affected by Trace, Gastro Acid, etc.

- [  * ] Fainting will not cause the form to change back.

Feature wise, seems to work correctly.

src/overrides.ts
XGSleepWalker: Reverted overrides back, thanks for noticing.

PR #1134, opened 2024-05-19
Bug Fix with Starter Icon Sprites along with adjustment for the Starter UI

karl-police:
This took way too long 💀 

If this Pull Request gets denied, solely because there are ``//``...

That's why it took so long, I had to figure shit it out. And I don't want people to have to figure too much shit out, so I put comments. I'd have put more _documentation_ but, I kept it related to this Pull Request.

So, if you solely deny this Pull Request because there are ``// comments that explain stuff``. Then something is wrong with you and you hate people. 👍 

&nbsp;

Okay, good.

&nbsp;

### Explanation

Currently, in the game, **when you turn a shiny Pokemon to a non-shiny** then use the arrow keys to move to a different Pokemon. The Sprite resets back to shiny. (Only if you own a shiny)

With selection I mean this thing:
image
image

&nbsp;

So, I changed that, but it wouldn't make sense to solely change the icon not resetting back. If the entire statistics change back.

It's caused by ``lastSpeciesIcon``. Because I have no clue why that existed, I've commented it out incase it was for something. But I doubt.

With that I also had to comment out its ``checkIconId``. I have absolutely no idea, what the hell the purpose of ``checkIconId`` is.

### Changes

The starter selection will also **remember** the last value. So if I go back to, it will remain non-shiny.

Currently, it only remembers changes made to **Shiny**, because I am not sure how many would want the other values to remain when switching back. It shouldn't matter though, because the game, literally selects a Nature based on whatever you have unlocked. So something _default_ doesn't really exist? 🤔 Well, maybe for the Nature since all are the same if you have all. But not for the abilities.

However, I can easily add the other properties as well, since they're all stored in a goofy table called ``customizedProps``. Similar to how starter values are stored when you have them added to the party, but all at one place.

But note that if you have a chosen starter right here:
image

Currently the game will actually keep all the modifications you've set, which I think is nice. Like that you don't have to unselect the starter to make modifications. In the other hand, the selected sprite icon never updates. Which is another bug.

&nbsp;

Regardless though, I've made it so that it will forget all customized properties if you return to the main menu.

### DIFF
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -164,6 +164,13 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
   private starterMovesets: StarterMoveset[] = [];
   private speciesStarterDexEntry: DexEntry;
   private speciesStarterMoves: Moves[];
+  private customizedProps: {
+    [index: number]: {
+      dexAttr?: DexAttrProps;
+      abilityIndex?: number;
+      natureIndex?: number;
+    }
+  } = {};
   private canCycleShiny: boolean;
   private canCycleForm: boolean;
   private canCycleGender: boolean;
@@ -755,6 +762,7 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
         this.updateInstructions();
       } else {
         this.blockInput = true;
+        this.customizedProps = {}; // Clear this
         this.scene.clearPhaseQueue();
         this.scene.pushPhase(new TitlePhase(this.scene));
         this.scene.getCurrentPhase().end();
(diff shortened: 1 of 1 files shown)


Icehawk78:
I think if the selected shiny value is remembered (which I agree that it should be) then it would make the most sense to have all of the other selectable props also be remembered.

Additionally, would it be possible to *not* have the customized properties forgotten by returning to the main menu? This is already how moves work, so it seems better to have all properties that have been changed persist.

karl-police:
> I think if the selected shiny value is remembered (which I agree that it should be) then it would make the most sense to have all of the other selectable props also be remembered.
> 
> Additionally, would it be possible to _not_ have the customized properties forgotten by returning to the main menu? This is already how moves work, so it seems better to have all properties that have been changed persist.

Alright, will change that.

The way I do it seems goofy, cuz I wanna save unnecessary memory. So only changed stuff goes inside the customized prop, but honestly it's creating an entry for every thing you cycle through. I didn't create a Class or Interface for it, the entire thing is already chaos, but I put comments.

karl-police:
> I think if the selected shiny value is remembered (which I agree that it should be) then it would make the most sense to have all of the other selectable props also be remembered.
> 
> Additionally, would it be possible to _not_ have the customized properties forgotten by returning to the main menu? This is already how moves work, so it seems better to have all properties that have been changed persist.

Changed, and now the starter that is selected, also visually changes

Since the logic of the starter gui is already accounting any changes made even when you have a starter already selected.

image
image

flx-sta:
Considered stale.

karl-police:
> No action within more than a month. Considered stale.
> 
> Feel free to re-open it when you are still working on it/want to work on it again

the what?

flx-sta:
> the what?

You want me to re-open it ?
We are trying to keep things a little “tidy” that’s why we close stale PRs

karl-police:
> > the what?
> 
> You want me to re-open it ? We are trying to keep things a little “tidy” that’s why we close stale PRs

probably

PR #1136, opened 2024-05-19
HSTS + XCTO

danieland:
HSTS isn't currently enabled, which can potentially lead to some security issues, probably the most common of these a man-in-the-middle attack, for more info see:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security
Basically this informs the client connecting that all requests should be HTTPS, but you can read more about it above, or feel free to ask!

Alongside this, X-Content-Type-Options header is added too, this can stop things like improper scripts running, more info here:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options

These two alongside CSP is holding back a perfect security score at the moment, CSP can be quite a pain though to implement, these are easy.

There should be zero impact to any current features, but some testing cases could include stuff like ensuring there's no hiccups with a locally hosted pokerogue server or similar connectivity cases

EDIT: to do a somewhat quick security audit yourself - https://observatory.mozilla.org/ is a great resource !


flx-sta:
Considered stale.

PR #1142, opened 2024-05-19
Add support for a simple menu message

Admiral-Billy:
As the title implies, this adds support for a basic menu message that can be used to alert players of shiny events, special updates, giveaways, etc.

If the message is an empty string, it just won't show up at all. If the background color is blank, there will be no background color. if the text color is blank, it'll default to white.

For example this is the result of this code:

```typescript
export const MENU_MESSAGE: string = "Admiral Billy cool mode active!";
export const MENU_MESSAGE_TEXT_COLOR: string = "#FFD700";
export const MENU_MESSAGE_BACKGROUND_COLOR: string = "#FF69B4";
```

image

Feel free to move the variables wherever seems most appropriate; currently they're just at the top of overrides.

### DIFF
--- a/src/overrides.ts
+++ b/src/overrides.ts
@@ -12,10 +12,13 @@ import { Stat } from './data/pokemon-stat';
 import { PokeballCounts } from './battle-scene';
 import { PokeballType } from './data/pokeball';
 
+export const MENU_MESSAGE: { active: boolean, text: string, textColor: string, backgroundColor: string } = {
+    active: true,
+    text: "Shiny event active!",
+    textColor: "#FFD700",
+    backgroundColor: "#FAFAD2"
+}
 
-export const MENU_MESSAGE: string = "";
-export const MENU_MESSAGE_TEXT_COLOR: string = "";
-export const MENU_MESSAGE_BACKGROUND_COLOR: string = "";
 /**
  * Overrides for testing different in game situations
  * if an override name starts with "STARTING", it will apply when a new run begins
--- a/src/overrides.ts
+++ b/src/overrides.ts
@@ -12,6 +12,10 @@ import { Stat } from './data/pokemon-stat';
 import { PokeballCounts } from './battle-scene';
 import { PokeballType } from './data/pokeball';
 
+
+export const MENU_MESSAGE: string = "";
+export const MENU_MESSAGE_TEXT_COLOR: string = "";
+export const MENU_MESSAGE_BACKGROUND_COLOR: string = "";
 /**
  * Overrides for testing different in game situations
  * if an override name starts with "STARTING", it will apply when a new run begins
--- a/src/overrides.ts
+++ b/src/overrides.ts
@@ -12,13 +12,6 @@ import { Stat } from './data/pokemon-stat';
 import { PokeballCounts } from './battle-scene';
 import { PokeballType } from './data/pokeball';
 
-export const MENU_MESSAGE: { active: boolean, text: string, textColor: string, backgroundColor: string } = {
-    active: true,
-    text: "Shiny event active!",
-    textColor: "#FFD700",
-    backgroundColor: "#FAFAD2"
-}
-
 /**
  * Overrides for testing different in game situations
  * if an override name starts with "STARTING", it will apply when a new run begins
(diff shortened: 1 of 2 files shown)


src/overrides.ts
```diff
@@ -12,6 +12,10 @@ import { Stat } from './data/pokemon-stat';
 import { PokeballCounts } from './battle-scene';
 import { PokeballType } from './data/pokeball';
 
+
+export const MENU_MESSAGE: string = "";
+export const MENU_MESSAGE_TEXT_COLOR: string = "";
+export const MENU_MESSAGE_BACKGROUND_COLOR: string = "";
```
molker, line 18: if these are always going to be used together can they be one object MENU_MESSAGE with keys for each of these consts?
and have an "active" key like in the pokeball override so we can use that as the conditional?

molker (changes requested):
this is a great idea, very simple way to display events
one small suggestion

src/overrides.ts
Admiral-Billy: I pushed some changes; are you looking for something like that, or did I misunderstand? The settings currently pushed are for a shiny event, with gold text and a pale yellow background.

src/overrides.ts
```diff
@@ -12,6 +12,13 @@ import { Stat } from './data/pokemon-stat';
 import { PokeballCounts } from './battle-scene';
 import { PokeballType } from './data/pokeball';
 
+export const MENU_MESSAGE: { active: boolean, text: string, textColor: string, backgroundColor: string } = {
+    active: true,
+    text: "Shiny event active!",
```
danieland, line 17: Any thoughts on localization ?

Admiral-Billy: I'm not sure how the localization is handled normally; presumably the text here can be handled the same way as other localization strings if needed?

danieland: Code wise it's pretty straightforward, an example from the current code can be seen here:
https://github.com/pagefaultgames/pokerogue/blob/3671fe4c2f81b8efe097819079408fd4432d699c/src/data/pokeball.ts#L30

`import i18next from '../plugins/i18n';` gives access to the localization library
`i18next.t('pokeball:greatBall');` means get the translation with the name "greatBall" from the file "pokeball"

Personally I think all texts should be localizable, buuut maybe this one doesn't make sense to do since it's a sort of "admin" message that would require admins to be able to translate into 10+ languages ?

molker: It would be nice to be able to translate it, we'd have to run it by a  native speaker or hope for the best with google translate or something
I think for an initial implementation english should be okay until we figure out a process for all translations

src/overrides.ts
molker: That's exactly what I meant, thank you

Admiral-Billy:
The PR still said the changes were needed so I went ahead and re-requested the review.

I can also change the current settings if needed, but they do at least accurately describe what's currently going on even if it isn't pretty.

src/ui/title-ui-handler.ts
```diff
@@ -41,6 +42,16 @@ export default class TitleUiHandler extends OptionSelectUiHandler {
     this.playerCountLabel = addTextObject(this.scene, (this.scene.game.canvas.width / 6) - 2, (this.scene.game.canvas.height / 6) - 90, `? ${i18next.t("menu:playersOnline")}`, TextStyle.MESSAGE, { fontSize: '54px' });
     this.playerCountLabel.setOrigin(1, 0);
     this.titleContainer.add(this.playerCountLabel);
+	
+    if (Overrides.MENU_MESSAGE.active) {
+        this.menuMessage = addTextObject(this.scene, (this.scene.game.canvas.width / 6) - 2, (this.scene.game.canvas.height / 6) - 100, Overrides.MENU_MESSAGE.text, TextStyle.MESSAGE, {
```
molker, line 47: you  need to add menuMessage as a property of titleuihandler

Admiral-Billy: So you'd rather have it there instead of in the overrides? I can make that work.

flx-sta:
Considered stale.

PR #1143, opened 2024-05-20
[QOL] Allow starters to relearn egg moves

Fontbane:
Somewhat simple fix for a QOL issue.

Currently egg moves are mostly balanced around the last stage of a mon's line, including Megas. So occasionally you'll have to bring in an egg move that's dead weight until you reach the stage it's intended for. You might also accidentally delete egg moves during move learning.

This PR allows you to relearn an unlocked egg move with the Memory Mushroom _only if you brought that mon into the run as a starter_. It also adds a function, "getUnlockedEggMoves", which may be useful in the future.


Tempo-anon:
Going to close this PR out since it has conflicts and it seems #2482 implements this but also covers an edge case

Issue #1146, opened 2024-05-20
[Suggestion] Unlocking Egg Moves with Candy


lloyd-delacroix:
Wanted to open a discussion on the possibility of expanding the candy options to include unlocking a new egg move. The cost would have to be high enough to not negate Move Up Egg Gatcha, but provide a way to target favorite 'mons. As a starting point, using the passive unlock costs per unlocked egg might be reasonable, though it might need to be higher for the epic / legendary tiers or have some sort of increasing cost scaling per move unlocked.

lloyd-delacroix:
Added PR with sample implementation. #1147

jsalts:
As an avid player, I cannot wait to unlock egg moves with candy.

JoannaFalkowska:
IIRC Discord is a better place for a suggestion/balancing discussion, if you didn't start a thread there already.

PR #1147, opened 2024-05-20
Add Unlocking Egg Moves with Candy

lloyd-delacroix:
image

Added the ability to purchase a new egg move using candy. Currently set at a fixed cost equal to the passive unlock cost for that tier. This implementation unlocks the egg moves in order rather than randomly so the final rare egg move can only be unlocked after the first three are unlocked.

#1146


src/locales/fr/starter-select-ui-handler.ts
```diff
@@ -30,6 +30,7 @@ export const starterSelectUiHandler: SimpleTranslationEntries = {
     "selectMoveSwapWith": "Sélectionnez laquelle échanger avec",
     "unlockPassive": "Débloquer Passif",
     "reduceCost": "Diminuer le cout",
+    "learnEggMove": "Apprendre l'attaque aux œufs",
```
Greenlamp2, line 33: AHAH this sentence in french is hilarious, here's the correct one:
"capacité Œuf"

otherwise it means: "Learn the move to the eggs"

src/ui/starter-select-ui-handler.ts
```diff
@@ -103,6 +103,27 @@ function getValueReductionCandyCounts(baseValue: integer): [integer, integer] {
   return starterCandyCosts[baseValue - 1].costReduction;
 }
 
+function getEggMoveCandyCount(baseValue: integer): integer {
```
Greenlamp2, line 106: can't you reuse the same logic as the 2 other options ?

```ts
const starterCandyCosts: { passive: integer, costReduction: [integer, integer] }[] = [
  { passive: 50, costReduction: [30, 75] }, // 1
  { passive: 45, costReduction: [25, 60] }, // 2
  { passive: 40, costReduction: [20, 50] }, // 3
  { passive: 30, costReduction: [15, 40] }, // 4
  { passive: 25, costReduction: [12, 35] }, // 5
  { passive: 20, costReduction: [10, 30] }, // 6
  { passive: 15, costReduction: [8, 20] },  // 7
  { passive: 10, costReduction: [5, 15] },  // 8
  { passive: 10, costReduction: [3, 10] },  // 9
  { passive: 10, costReduction: [3, 10] },  // 10
]

function getPassiveCandyCount(baseValue: integer): integer {
  return starterCandyCosts[baseValue - 1].passive;
}

function getValueReductionCandyCounts(baseValue: integer): [integer, integer] {
  return starterCandyCosts[baseValue - 1].costReduction;
}
```

lloyd-delacroix: When I wrote this, it was the same logic as the other options. It's been reworked by someone else since I submitted my PR, but yeah, should be easy to refactor. 

https://github.com/pagefaultgames/pokerogue/commit/ca1ae4b55621a38b6e72778f57be506f0c9f7c3c

Tempo-anon:
Going to close this out since this PR is a bit old and has conflicts, as well as the fact players can buy eggs with candy that can give egg moves.

PR #1151, opened 2024-05-20
update chinese conifg.ts and egg.ts

GhostFlys:


GhostFlys:
image

Madmadness65 (approved):
This one doesn't appear to be crashing now, so it _should_ be fine now

PR #1152, opened 2024-05-20
Improve move summary screen UI

knispeja:
No longer automatically enter move rearrange mode when tabbing to move list. This is quite annoying for me so I updated this for parity with the official games (you must press "confirm" (A) once to begin rearranging moves). Note that I did test learning a move and confirmed this doesn't interfere with that UI logic.

Before:
https://github.com/pagefaultgames/pokerogue/assets/9343160/3c1f7753-bf23-45b1-8921-ccb439f32d88

After:
https://github.com/pagefaultgames/pokerogue/assets/9343160/c7e26d44-b09e-4601-814c-27dea2d02e38


happinyz (approved):
lgtm, an eventual improvement would be to hide the description box when not selecting a move

bennybroseph:
This reverts an intentional design choice

knispeja:
> This reverts an intentional design choice

Feel free to close if you don't think it makes sense. I just personally found it confusing the way it currently works, I thought it was a mistake

knispeja:
> I'm open to feedback, I'm just not sure what the benefit is. The only action you'd want to take in the move select screen is viewing/swapping moves around

The current behavior prevents you from using up/down to cycle through your party while on the move summary until you hit "Cancel", which is why I thought it was a mistake (since you can cancel an action you never prompted). But maybe this is only a me thing -- I do like to cycle through my party on this page often.

j-diefenbach:
> > I'm open to feedback, I'm just not sure what the benefit is. The only action you'd want to take in the move select screen is viewing/swapping moves around
> 
> The current behavior prevents you from using up/down to cycle through your party while on the move summary until you hit "Cancel", which is why I thought it was a mistake (since you can cancel an action you never prompted). But maybe this is only a me thing -- I do like to cycle through my party on this page often.

Understandable but to me cycling through a specific pokemon's moves takes priority over checking all the pokemons' moves. I prefer hitting cancel once to cycle through all over having to press z just to look at specific moves, especially given the fact I opened the summary for that specific mon

bennybroseph:
It's two lines of code, i'm fine with closing this. I don't think we should change the current behavior at this time.

knispeja:
> Understandable but to me cycling through a specific pokemon's moves takes priority over checking all the pokemons' moves. I prefer hitting cancel once to cycle through all over having to press z just to look at specific moves, especially given the fact I opened the summary for that specific mon

Yeah I get why it was done now, I just think it's bad/confusing UI design. You also have to hit cancel twice to exit the menu, which is out of step with how every other summary page works. Automatically "confirming" into move inspection, to me, is like automatically choosing "Fight" for the player at the beginning of a battle because it's the most common option. But again I don't think this is that serious, I'm OK with having this closed, I just want my perspective to be clear :)

PR #1153, opened 2024-05-20
Implement Taunt, Torment, Heal Block, and Aroma Veil

jatinkohli:
(WIP) Implement Taunt, Torment, and Heal Block moves. All these moves are variants of disable which prevent the target from using certain types of moves. Also implement Aroma Veil, which prevents the effects of these moves and a few others.


jatinkohli:
# Torment

As of now, I (mostly) implemented torment. I am still working on taunt and heal block.

## Torment requirements
- Target cannot use the same move twice in a row
- Torment status only is removed when the pokemon leaves the battlefield
- Torment does not take effect until the turn after it is used
- A pokemon with only one usable move will struggle every other turn

## Torment implementation

Torment prevents the user from selecting the same move twice (see below video).
[https://youtu.be/64DZXjgJK0Q](https://youtu.be/64DZXjgJK0Q)

Torment status is removed upon switch (see below video). My abra is tormented with a previous move of growl before switching out. When my abra comes back in, it has no torment status and can use growl again.
[https://youtu.be/a-2MJQOgeEY](https://youtu.be/a-2MJQOgeEY)

Torment causes the target to struggle every other turn if the target has only one usable move (see below video).
[https://youtu.be/rtrZj9B4tLs](https://youtu.be/rtrZj9B4tLs)

Torment does not take effect until the turn after it is used (see below video), meaning a pokemon who was just tormented can use the same move as it did last turn. My shuckle keeps using growl. On the turn it is tormented, it is still able to use growl. I cut the video off too early, but the next turn shuckle cannot select growl just like in the first video.
[https://youtu.be/3Bgwp52Yokc](https://youtu.be/3Bgwp52Yokc)

## What's Next for Torment

### Double Battle Targeting Issues
As shown in the fourth video (about tormenting not taking effect until the turn after), one wild abra uses torment on the other wild abra. This should not happen, since torment is a debuffing move. I was confused about the move scoring system and how to handle this, so some help would be appreciated. <3

### Localization 
Torment has two messages which need localization: the message when a pokemon is tormented ("_pokemon_ was subjected to torment!") and the message when the user tries to select the same move as they did the previous turn under torment ("_pokemon_ can't use the same move twice in a row due to the torment!").

### Unintuitive Struggle
As shown in the third video (about torment causes struggle every other turn in some cases), the UI shows the pokemon's move instead of what will happen, struggle. This could be confusing and/or frustrating, so it would be a good idea to change this move to struggle in this scenario. This also probably applies to other existing implementations of moves like disable.

jatinkohli:
Will work on heal block in the near future, it will be very barebones and will not include many interactions with items and abilities, at least at first.

jatinkohli:
I am also implementing aroma veil, it is lumped in here because its primary purpose is to give immunity to moves like taunt, torment, etc.

jatinkohli:
This marks all the changes I want to make for these features, I will start testing and making updated videos/comments soon

jatinkohli:
# Heal Block/Psychic Noise
I partially implemented the heal block status (just blocks healing moves) for Heal Block and Psychic Noise

## Heal Block status requirements
### Requirements for my partial implementation
- Heal Block status prevents the target from using status moves
- Heal Block status lasts for 5 turns (including the turn it is used) from Heal Block and 2 turns from Psychic Noise
- Heal Block is also removed when the target leaves the battlefield (faint, switch, baton pass, etc.)
- Pokemon with the Aroma Veil ability cannot become heal blocked (detail about aroma veil in another comment in this pr)
  - will still take damage from psychic noise

### Additional Requirements for a full implementation
-

## Taunt implementation
Taunt prevents the target from selecting status moves (see below video). When the target (my Bulbasaur) moves before the user (enemy pokemon), taunt lasts for 4 turns (3 additional turns after the turn it was inflicted). Taunt persists as long as the target remains on the field, even if the user faints.

https://github.com/pagefaultgames/pokerogue/assets/22225497/9d3d023e-a2e0-421f-8361-a305d4e3c1ca



When the target (my Shuckle) moves after the user (enemy pokemon), taunt lasts for 3 turns (2 additional turns after the turn it was inflicted). 

https://github.com/pagefaultgames/pokerogue/assets/22225497/710e9a65-e794-445b-b0c4-9fb756bb01de



Taunt causes the target to struggle if the target has only status moves (see below video).

https://github.com/pagefaultgames/pokerogue/assets/22225497/dc7ef022-8b29-46f3-91a8-e915e78b14e2



Taunt fails on a pokemon with the oblivious ability.

https://github.com/pagefaultgames/pokerogue/assets/22225497/a2e4f84a-f337-4a2b-a676-7aa4fd420b34



## What's Next for Taunt

### Mechanics
Magic guard and magic bounce need to reflect status moves like taunt. Taunt also needs to bypass substitute. I will leave these to whoever takes on the tasks of implementing these.

### Localization
Taunt has three messages which need localization: the message when a pokemon is taunted ("_pokemon_ fell for the taunt!"), the message when a move fails due to taunt or the user attempts to select a status move while taunted ("_pokemon_ can't use _move_ after the taunt!"), and the message when the taunt duration expires ("_pokemon_ shook off the taunt!").

### Unintuitive Struggle
Same point as above in torment comment in this pr.

jatinkohli:
# Disable
I modified the existing implementation for disable to be a BattlerTag like the other move blocking statuses. This comment describes testing the new implementation to ensure everything works properly.

## Disable requirements
- Disable disables the target's previously used move for 4 turns (including the turn the status is applied)
- Disable is also removed when the target leaves the battlefield (faint, switch, baton pass, etc.)
- Disable comes into effect immediately, potentially cancelling the target's move for the turn
- A pokemon with no usable moves as a result will struggle
- Disable fails on Max pokemon
- Pokemon with the Aroma Veil ability cannot be affected by disable (detail about aroma veil in another comment in this pr)

## Disable testing

Disable disables the target's previously used move for 4 turns (including the turn the status is applied). It fails when the target has not yet used a move. It cancel's the target's move for the turn if the user moves before the target and the target used the same move as last turn. It remains in effect as long as the target remains on the battlefield, even if the user faints.

https://github.com/pagefaultgames/pokerogue/assets/22225497/7b59547a-5c22-4d2c-a81b-e5537e51922e


The disable status is removed whenever the user leaves the battlefield (switch, baton pass, faint, etc.).

https://github.com/pagefaultgames/pokerogue/assets/22225497/eefc0967-436c-4879-b215-c39ace9e4d7a


Pokemon with no usable moves as a result of the disable status will struggle.

https://github.com/pagefaultgames/pokerogue/assets/22225497/a35a6aad-fd6a-419d-b052-16b73ed89303


Disable fails on Max pokemon.

https://github.com/pagefaultgames/pokerogue/assets/22225497/52d75521-e0c1-45d7-8d38-a7a4be7abcc1


## What's Next for Disable

### Mechanics
Magic guard and magic bounce need to reflect status moves like disable. Disable does not seem to bypass substitute like taunt and torment, but I am not entirely sure. I will leave these to whoever takes on the tasks of implementing these.

### Unintuitive Struggle
Same point as above comments in this pr.

jatinkohli:
# Torment

I (almost entirely) implemented torment.

## Torment requirements
- Target cannot use the same move twice in a row
- Torment status only is removed when the pokemon leaves the battlefield (faint, switch, baton pass, etc.)
  - I decided to have torment remain between battles and not reset which move was used last, similar to encore's implementation
- Torment does not take effect until the turn after it is used
- A pokemon with no usable moves on a given turn as a result will struggle
- Torment fails on Max pokemon
- Pokemon with the Aroma Veil ability cannot become tormented (detail about aroma veil in another comment in this pr)

## Torment implementation

Torment prevents the user from selecting the same move twice (see below video).

https://github.com/pagefaultgames/pokerogue/assets/22225497/3c9b3a9b-7e11-4ca4-8821-623e10cc6975



Torment status is removed upon switch, baton pass, and faint (see below video). My abra is tormented with a previous move of growl before switching out. When my abra comes back in, it has no torment status and can use growl again.

https://github.com/pagefaultgames/pokerogue/assets/22225497/6e0a3b8b-503e-4935-89bb-63645ebfabbb



Torment causes the target to struggle every turn which is has no usable moves (see below videos).

Torment + encore means the target will have no usable moves every other turn, starting the turn after the encore status is given.

https://github.com/pagefaultgames/pokerogue/assets/22225497/f59a8d4f-e0d5-4a4a-8be2-54e74443ca0a



Torment on a pokemon with only one move in its moveset causes the target to struggle every other turn.

https://github.com/pagefaultgames/pokerogue/assets/22225497/00059851-0d1f-41e7-b8fb-78128185c814



Torment does not take effect until the turn after it is used (see below video), meaning a pokemon who was just tormented can use the same move as it did last turn. My shuckle keeps using growl. On the turn it is tormented, it is still able to use growl. I cut the video off too early, but the next turn shuckle cannot select growl just like in the first video.

https://github.com/pagefaultgames/pokerogue/assets/22225497/6ebffd29-dbf4-4c44-9610-281ced3b8d2b



Torment remains in effect between battles as long as the target remains on the field (see below video).

https://github.com/pagefaultgames/pokerogue/assets/22225497/40f2090c-4d36-4d31-bdda-2c5ce555556c



Torment fails on Max pokemon (see below video).

https://github.com/pagefaultgames/pokerogue/assets/22225497/d378e87e-3b6a-4c53-aa90-0052e8762b7d



## What's Next for Torment

### Mechanics
Magic guard and magic bounce need to reflect status moves like torment. Torment also needs to bypass substitute. I will leave these to whoever takes on the tasks of implementing these.

### Localization 
Torment has two messages which need localization: the message when a pokemon is tormented ("_pokemon_ was subjected to torment!") and the message when the user tries to select the same move as they did the previous turn under torment ("_pokemon_ can't use the same move twice in a row due to the torment!").

### Unintuitive Struggle
As shown in the third video (about torment causes struggle every other turn in some cases), the UI shows the pokemon's move instead of what will happen, struggle. This could be confusing and/or frustrating, so it would be a good idea to change this move to struggle in this scenario. This also probably applies to other existing implementations of moves like disable.

jatinkohli:
A few other things to consider:

- How should scoring work for these moves? Currently they have a fixed score like Encore, but in double battles there may be some targeting issues.
- Should these moves work on bosses? They seem somewhat balanced, so I vote yes.

jatinkohli:
Ready for review

Tempo-anon:
Really appreciate the work and implementation you've done but this PR is a bit out of date and has merge conflicts and also should probably be broken up so each move has their own PR. Closing it for now

PR #1155, opened 2024-05-20
Implemented Plus and Minus

Upos1997:
As the title says, I implemented the Plus and Minus abilities.

This is my first time using pull requests (in general not that familiar with Git), so if there is anything I forgot to do, do let me know.


bennybroseph:
Do you have a video showing this move being tested?

Upos1997:
Not currently, give me a moment to make one

Upos1997:
https://github.com/pagefaultgames/pokerogue/assets/37412723/ddcaebd9-65ce-4aae-9760-a78d452aa6a7

With Plusle and Minun on the field:
Damage numbers: 11, 12, 12
With only Plusle
Damage numbers: 8, 9, 9

src/data/ability.ts
```diff
@@ -3138,9 +3138,11 @@ export function initAbilities() {
     new Ability(Abilities.CUTE_CHARM, 3)
       .attr(PostDefendContactApplyTagChanceAbAttr, 30, BattlerTagType.INFATUATED),
     new Ability(Abilities.PLUS, 3)
-      .unimplemented(),
+      .conditionalAttr(p => p.getAlly() && [Abilities.PLUS, Abilities.MINUS].some(a => p.getAlly().hasAbility(a)), BattleStatMultiplierAbAttr, BattleStat.SPATK, 1.5)
+      .ignorable(),
```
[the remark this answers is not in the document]
Upos1997: the pokemon.getAlly() is also used in other places in the code (like the Healer ability) to check if a pokemon has an ally or not.  It will never reach the .hasAbility() section of the code as typescript stops at the first condition that returns false in an and. So I'm fairly certain this would be functionally correct.

I do agree that pokemon.scene.currentBattle.double is clearer so I changed it to that.

src/data/ability.ts
```diff
@@ -3138,9 +3138,11 @@ export function initAbilities() {
     new Ability(Abilities.CUTE_CHARM, 3)
       .attr(PostDefendContactApplyTagChanceAbAttr, 30, BattlerTagType.INFATUATED),
     new Ability(Abilities.PLUS, 3)
-      .unimplemented(),
+      .conditionalAttr(p => p.scene.currentBattle.double && [Abilities.PLUS, Abilities.MINUS].some(a => p.getAlly().hasAbility(a)), BattleStatMultiplierAbAttr, BattleStat.SPATK, 1.5)
```
TempsRay, line 3141: Won't this trigger if both Pokemon have plus or both Pokemon have minus?

src/data/ability.ts
```diff
@@ -3138,9 +3138,11 @@ export function initAbilities() {
...
+      .ignorable(),
     new Ability(Abilities.MINUS, 3)
-      .unimplemented(),
+      .conditionalAttr(p => p.scene.currentBattle.double && [Abilities.PLUS, Abilities.MINUS].some(a => p.getAlly().hasAbility(a)), BattleStatMultiplierAbAttr, BattleStat.SPATK, 1.5)
```
TempsRay, line 3144: Also I thought the plan was to trigger on itself too (a pokemon with both plus and minus will get the boost)

src/data/ability.ts
Upos1997: That is how the abilities work since gen 5.
https://bulbapedia.bulbagarden.net/wiki/Plus_(Ability)

src/data/ability.ts
Upos1997: I'm not aware of any plans like that.
It's easy enough to change to that if I can get confirmation that this is indeed how they wish plus and minus to work.

src/data/ability.ts
Xavion3: Does this gracefully fail if it's a double battle and you only have one pokemon out?

Upos1997: https://github.com/pagefaultgames/pokerogue/assets/37412723/aa2e1f3b-6468-46f8-9566-b7a06e48e35f

It does.

Upos1997:
I synced the fork and checked that my changes work with the new version of node (they do).

bennybroseph:
Not seeing the Ability Pop-Up on any of these videos. Does the Ability show the pop-up when sent into battle and conditions are met?

Upos1997:
Currently there is no pop up no.
I based it on other conditional stat increasing abilities like Guts, which also have no pop up.

bennybroseph:
I'm gonna have to boot up S/V again aren't I

bennybroseph (approved):

PR #1156, opened 2024-05-20
Prevent Pokemon Info screen from being covered up

knispeja:
Prevents the yes/no confirm menu from getting in the way when the Pokemon Info screen is open when catching a new Pokemon with a full party. Maybe subjective whether or not this is improved but I think it's nicer this way since you can actually read the ability and nature completely.

Before:

https://github.com/pagefaultgames/pokerogue/assets/9343160/97e62850-970f-4a99-8a40-c73ce9439bbf

After:

https://github.com/pagefaultgames/pokerogue/assets/9343160/3ae328a4-5286-4afd-ae1b-0fab0e99e1fc

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -4322,6 +4322,7 @@ export class AttemptCapturePhase extends PokemonPhase {
         if (this.scene.getParty().length === 6) {
           const promptRelease = () => {
             this.scene.ui.showText(`Your party is full.\nRelease a Pokémon to make room for ${pokemon.name}?`, null, () => {
+              this.scene.pokemonInfoContainer.makeRoomForConfirmUi();
               this.scene.ui.setMode(Mode.CONFIRM, () => {
                 this.scene.ui.setMode(Mode.PARTY, PartyUiMode.RELEASE, this.fieldIndex, (slotIndex: integer, _option: PartyOption) => {
                   this.scene.ui.setMode(Mode.MESSAGE).then(() => {
--- a/src/ui/confirm-ui-handler.ts
+++ b/src/ui/confirm-ui-handler.ts
@@ -5,6 +5,9 @@ import i18next from "i18next";
 import {Button} from "../enums/buttons";
 
 export default class ConfirmUiHandler extends AbstractOptionSelectUiHandler {
+
+  public static readonly windowWidth: integer = 48;
+
   private switchCheck: boolean;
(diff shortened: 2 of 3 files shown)


bennybroseph:
Thinking on it, this is actually a good solution. With the window slid to the left you can now also see money/floor/etc. I'd like to see it either occur faster or have the "Yes" "No" prompt tween alpha to make it less jarring. The later is definitely outside the scope of this PR though.

Also, was this tested on different resolutions/mobile?

karl-police:
Animation faster, definitely. Like very fast. It's like u ask me a question and I have to respond with "Yes" or "No", and the response needs to be served NOW, type of thing.

TempsRay:
Whoa sliding the whole window to the left, pretty cool. I know that can help alleviate issues I've heard users complain about such as the HP and defense IVs being covered up by other UI components

knispeja:
> Also, was this tested on different resolutions/mobile?

I did try several window sizes (up to 4k) on desktop and did the same in Chrome's Device Mode thing to emulate mobile -- it seems like this game always renders at the same aspect ratio though so I'm not too worried

> I'd like to see it either occur faster or have the "Yes" "No" prompt tween alpha to make it fit in more.

Good idea, I sped it up quite a bit and experimented with different ease functions. Looking for feedback on which one people like better (or something else, maybe, from [Phaser's easing functions](https://phaser.io/docs/2.6.2/Phaser.Easing.html)):

`Cubic.easeInOut` - this is what was already being used, I just sped it up a lot
https://github.com/pagefaultgames/pokerogue/assets/9343160/91830e37-7376-4dd1-bb36-8bb8d903269c

`Back.easeOut` - a little weird, but this one gives the feeling that the confirm menu "knocks" the Pokemon info screen out of the way, I kind of like it
https://github.com/pagefaultgames/pokerogue/assets/9343160/d666e9a5-642d-4a14-90de-5b46504d849c

bennybroseph:
I think we should go with the sped up `Cubic.easeInOut` . The other one is cool, but none of our other animations match this. Maybe something to consider, but not for this PR

knispeja:
> I think we should go with the sped up `Cubic.easeInOut`

Done! It's currently using that cubic example from my previous comment.

bennybroseph:
man, i love little polish things like this

Issue #1169, opened 2024-05-20
Berries do not proc at the appropriate time.


alchemis:
Currently, all berries proc their effects at the end of the turn, after damage calculation, but this is not the way they work in the mainline games.
The way it should work is that they should proc immediately when their condition is met, status for Lum, super-effective move for enigma, hp threshold for the rest.
Specifically this means that if you have a Lum berry, you cannot suffer the effects of an status, not even once, because the Lum should proc immediately when it is inflicted.
It would also mean that you can heal *during* a multi-hit move, which would nerf them somewhat, but I think the tradeoff is fine.

DayKev:
This is currently intended behavior, though it is being discussed right now.

PR #1176, opened 2024-05-20
made enigma berries and lum berries able to proc more than once per turn

alchemis:
fixes issue #1169
I think this issue is actually a bit up to interpretation, because in mainline pokemon games you cannot have more than 1 berry at once. 
Normally, if you are below 50% hp for example and obtain a Sitrus berry, you will heal immediately.
However I'm not sure if Sitrus berries and stat raising berries proccing multiple times per turn is desirable, so I have added the requirement to them that they can only be procced during the TurnEndPhase.
However, I made Enigma and Lum berries able to proc multiple times per turn, this is currently checked for both the user and the opponent after using a move or being targeted by a move.
Note: This means that (as in cart) if you select a move and are afflicted by an status that would prevent you from using that move, but have a Lum berry, it will clear the status immediately and you will be able to use the move.
There might be edge cases.

### DIFF
--- a/src/data/berry.ts
+++ b/src/data/berry.ts
@@ -1,4 +1,4 @@
-import { PokemonHealPhase, StatChangePhase } from "../phases";
+import { PokemonHealPhase, StatChangePhase, TurnEndPhase } from "../phases";
 import { getPokemonMessage } from "../messages";
 import Pokemon, { HitResult } from "../field/pokemon";
 import { getBattleStatName } from "./battle-stat";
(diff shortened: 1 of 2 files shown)


bennybroseph:
I feel like berries are strong enough on their own, no? We probably need to take a closer look at what we want players to get out of berries before going and buffing them. Also, the change to berries triggering only at the end of the turn was a consideration for the player so that high level pokemon or bosses do not consume 10 berries to heal to full multiple times or fully buff themselves in the middle of the turn.

bennybroseph:
I feel like the lum berry changes attacks a symptom of the problem but not the root. Sure, you'd never get put to sleep/paralyzed/frozen/etc. but what is the point of those tokens then? In endless, you have infinite berries. This would make the tokens non-existent. Maybe that's okay because we shouldn't have them, but I feel like that warrants a larger discussion than is appropriate on a Pull Request

alchemis:
I think that the tokens existing is fine, they encourage the player to build to sustain on berries, specially if they appear early. The problem is that currently they behave as status + flinch tokens, with the current implementation of Lum berries.
This change also affects things like moves that have a chance to inflict a status effect, in a way that I think makes it more intuitive.
In retrospective I should've marked this PR as draft, I apologize.

Tempo-anon:
Closing this due to inactivity and merge conflicts. Also there has been changes to endless and will eventually also be reworked

PR #1181, opened 2024-05-20
Fixed warnings that triggered with ptBR and zhCN

josericardo-fo:
### Explanation:

image
[link](https://www.i18next.com/overview/configuration-options)

### Warnings:
image

### DIFF
--- a/src/plugins/i18n.ts
+++ b/src/plugins/i18n.ts
@@ -93,8 +93,9 @@ export function initI18n(): void {
 
   i18next.use(LanguageDetector).init({
     lng: lang,
+    nonExplicitSupportedLngs: true,
     fallbackLng: 'en',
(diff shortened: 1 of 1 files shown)


CodeTappert:
this wont work like this. 

First of all you would need to rename this everywhere like this.

Otherwise there are 2 kinds of portuguese, 2 kinds of chinese. So it can not be without the BR or CN

CodeTappert:
> this wont work like this.
> 
> First of all you would need to rename this everywhere like this. (even the folders)
> 
> Otherwise there are 2 kinds of portuguese, 2 kinds of chinese. So it can not be without the BR or CN
> 
> Maybe do ptBR and zhCN just without the "_"

Also text.ts needs it changed, starter-select-ui-handler, settings.ts

CodeTappert:
I deleted my comments since i learned what he did here was correct. And i was wrong

danieland (approved):

molker (approved):

PR #1186, opened 2024-05-21
Fix - Missing json sprite + masterlist updated

Greenlamp2:
- Added missing json's sprite from colormap script
- Updated masterlist
- Updated tests

Feel free to double check of course :)

thanks for the knowledge sharing Chacolah !


public/images/pokemon/variant/_masterlist.json
```diff
@@ -1796,8 +1801,8 @@
 	],
 	"890-eternamax": [
 		0,
-		2,
-		2
+		1,
```
molker, line 1804: are we using the palette swap now instead of the sprite sheet replacements?

public/images/pokemon/exp/890-eternamax.png
molker, line 1: does this need an experimental sprite? this looks identical to the one in the main folder

Greenlamp2: it's indeed the identical, it was needed to make the script not crashing on eternamax
image

public/images/pokemon/variant/_masterlist.json
Greenlamp2: i think the script determined the color swap and the sprite sheet returned the same result, here's what i observed:
image

what do you think ? should we revert to using the sprite sheet for eternamax ?

molker (approved):

PR #1188, opened 2024-05-21
Attempt to sort items on summary & battle screen for user's party.

HighMans:
~~This is a PR to sort items on Summary screen, currently sorted by item name and then by type.~~

This is a PR to sort items on Summary screen and battle screen, currently sorted by item name and then by type on user's party.


bennybroseph:
Could we get a video or screenshot showing the before and after of this change?

HighMans:
> Could we get a video or screenshot showing the before and after of this change?

Sorry about that, should've done that to begin with.

Before:
image


After:

image

bennybroseph:
Is it possible to just sort these items in the collection itself so that this shows in the battle UI as well? I like this, but only having it on the summary is kinda weird.

HighMans:
I agree, do you know the name of the file that controls that UI? There's so many and I'm not sure which one is which...

bennybroseph:
I think here is a good place to start

[VS Code Link](https://vscode.dev/github/bennybroseph/pokerogue/blob/Flyout-Add/src/battle-scene.ts#L1622)

HighMans:
Fixed the sorting to actually do sorting, partially implemented sorting of items displayed on screen during battle.

Don't know why it's not working on the enemy side. Need help from someone who knows what they're doing.

Before:
image
image



After:

image
image

bennybroseph:
Enemies just have a separate function to add modifiers i believe. should be in the same file, but a different function

HighMans:
I can see the items are sorted properly for the enemy side in the variable "sortedVisibleIconModifiers" while debugging, but the icons aren't displayed in the right order. It must be applied somewhere else/changed in the code somewhere else, I'm not sure.

bennybroseph:
Wait, why are there new variables for sorted modifiers? You should just sort the original collection.

bennybroseph:
I would try that as I would imagine that's why enemy isn't being updated

HighMans:
Just tried it, no difference.

Code and what it looks like in the debugger.

image

What it looks like on the display.

image

bennybroseph:
Is this still being worked on?

HighMans:
I'm at a stand still on the enemy side and how things are sorted there. You can close it, sorry.

bennybroseph:
Can we instead make the PR for only the player and update the title/code/etc. accordingly?

HighMans:
> Can we instead make the PR for only the player and update the title/code/etc. accordingly?

Done.

bennybroseph (approved):
A great v2 of this would be to add settings for a few different sorting options.

PR #246, opened 2024-04-23
Serene Grace, Shield Dust and Sheer Force partially implemented.

NxKarim:
- Added `MoveEffectChanceMultiplierAbAttr` to modify additional effect chances with multipliers.
- Added `IgnoreMoveEffectsAbAttr`; this sets to zero the chance of foe's moves additional effects.
- Changed `AddArenaTrapTagAttr`, `AddBattlerTagAttr`, and `StatusEffectAttr` to allow for modified additional effect chance.
- Added `getSheerForceHitDisableAbCondition` applied to abilities that get disable when hit by a Sheer Force boosted move.
- Added conditions to Poison Touch, Toxic Chain and Stench to be ignored by Shield Dust.
- Changed the additional effect chance of certain moves to make them immune or susceptible to chance modification as according to the main games.
Moves not affected by chance modification were changed to -1 because their effects will always apply.
Moves that are affected by chance modification were changed to 100, this allows for chance to be modified.
Most of these changes are because of Sheer Force.

Implemented:
- Serene Grace: Boosts moves additional effect chance by x2.
- Sheer Force: Boosts power of moves with additional effect by ~1.3. Disables additional effects on own moves. After boosted hit disables Color Change, PickPocket, Wimp Out, Emergency Exit, Berserk and Anger Shell.
- Shield Dust: Disables foe's moves additional effects and the effects of Poison Touch, Toxic Chain and Stench.
- Serene Grace doubles the chances of King's Rock
- Sheer Force negates King's Rock if the move has a secondary effect
- King's Rock won't applied if the target has Shield Dust

Not yet implemented:
- Sheer Force: Disable Meloetta form change. Disable Eerie Spell and Secret Power effect (Since they all are partially implemented).

Edit:
-Comments added as requested.
-Some new implementations added.

Edit 2:
- More moves set to not be affected by chance modification. 
- Added Order Up interaction with chance modification.

Edit 3:
- King's Rock interaction by Okimin

PR #1189, opened 2024-05-21
Implement Sheer Force

innerthunder:
This is a draft for now, as I still have a lot of work to do to implement and test Sheer Force. For now, I've made a condition function that checks whether or not Sheer Force would apply to a move as per [Bulbapedia's description](https://bulbapedia.bulbagarden.net/wiki/Sheer_Force_(Ability)). I'd like some input on what would be the best way to disable the move effects that this condition is checking for.


td76099:
Just so you know, someone has already started on Sheer Force with #246

PR #1192, opened 2024-05-21
Update moveset logic to account for new sacrifice attr

Xavion3:
The new addition of a variant sacrifice attribute for a few moves wasn't added to the moveset logic, so trainer and boss pokemon weren't getting the reduced rates of these as they were supposed to. Boss pokemon weren't supposed to be able to generate with self KO moves at all.


TempsRay (approved):
Can you update the PR desc? Mainly it's to prevent the AI from using self KO moves after an attr was updated

0zuzu (approved):

bennybroseph (approved):

PR #1195, opened 2024-05-21
Implement Arceus Plates and Silvally Memories

Madmadness65:
This adds every Plate and Memory as individual form change items for Arceus and Silvally respectively.
By these items simply existing, the abilities Multitype and RKS System are now "implemented" as well (no additional coding was needed for them).
Also included is entries for the Blank and Legend Plate, although no form change entries  are given for them yet as the Blank Plate seems redundant and the Legend Plate is complicated.
With the Plates and Memories, Judgment and Multi-Attack are also implemented (sharing code, as they are functionally identical).

Submitting this as a pull request in case we want to change these items to not be 17 individual items each, since I know item clutter is already an issue.


CodeTappert:
German translation:
Source: Pokewiki.de and https://bulbapedia.bulbagarden.net

```
  "FIST_PLATE": "Fausttafel",
    "SKY_PLATE": "Wolkentafel",
    "TOXIC_PLATE": "Gifttafel",
    "EARTH_PLATE": "Erdtafel",
    "STONE_PLATE": "Steintafel",
    "INSECT_PLATE": "Käfertafel",
    "SPOOKY_PLATE": "Spuktafel",
    "IRON_PLATE": "Eisentafel",
    "FLAME_PLATE": "Feuertafel",
    "SPLASH_PLATE": "Wassertafel",
    "MEADOW_PLATE": "Wiesentafel",
    "ZAP_PLATE": "Blitztafel",
    "MIND_PLATE": "Hirntafel",
    "ICICLE_PLATE": "Frosttafel",
    "DRACO_PLATE": "Dracotafel",
    "DREAD_PLATE": "Furchttafel",
    "PIXIE_PLATE": "Feentafel",
    "BLANK_PLATE": "Neutraltafel",
    "LEGEND_PLATE": "Legendentafel",
    "FIGHTING_MEMORY": "Kampf-Disc",
    "FLYING_MEMORY": "Flug-Disc",
    "POISON_MEMORY": "Gift-Disc",
    "GROUND_MEMORY": "Boden-Disc",
    "ROCK_MEMORY": "Gesteins-Disc",
    "BUG_MEMORY": "Käfer-Disc",
    "GHOST_MEMORY": "Geister-Disc",
    "STEEL_MEMORY": "Stahl-Disc",
    "FIRE_MEMORY": "Feuer-Disc",
    "WATER_MEMORY": "Wasser-Disc",
    "GRASS_MEMORY": "Pflanzen-Disc",
    "ELECTRIC_MEMORY": "Elektro-Disc",
    "PSYCHIC_MEMORY": "Psycho-Disc",
    "ICE_MEMORY": "Eis-Disc",
    "DRAGON_MEMORY": "Drachen-Disc",
    "DARK_MEMORY": "Unlicht-Disc",
    "FAIRY_MEMORY": "Feen-Disc",
```

JoshSalame (approved):
Looks good; your point about adding item clutter is a good consideration though. I also wonder if the new FormChangeItemTypeAttr needs individual tests or not (just to ensure one form change item works for Arceus and Silvally)

src/data/move.ts
```diff
@@ -3030,6 +3030,75 @@ export class VariableMoveTypeAttr extends MoveAttr {
   }
 }
 
+export class FormChangeItemTypeAttr extends VariableMoveTypeAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    if ([user.species.speciesId, user.fusionSpecies?.speciesId].includes(Species.ARCEUS || Species.SILVALLY)) {
+      const form = user.species.speciesId === Species.ARCEUS || Species.SILVALLY ? user.formIndex : user.fusionSpecies.formIndex;
+      const type = (args[0] as Utils.IntegerHolder);
+
+      switch (form) {
```
brain-frog, line 3039: couldn't you do `type.value = Type[form]`

src/data/pokemon-forms.ts
```diff
@@ -531,6 +569,25 @@ export const pokemonFormChanges: PokemonFormChanges = {
     new SpeciesFormChange(Species.SHAYMIN, "sky", "land", new SpeciesFormChangeTimeOfDayTrigger(TimeOfDay.DAWN, TimeOfDay.NIGHT)),
     new SpeciesFormChange(Species.SHAYMIN, "sky", "land", new SpeciesFormChangeStatusEffectTrigger(StatusEffect.FREEZE))
   ],
+  [Species.ARCEUS]: [
+    new SpeciesFormChange(Species.ARCEUS, "normal", "fighting", new SpeciesFormChangeItemTrigger(FormChangeItem.FIST_PLATE)),
```
brain-frog, line 573: are they all set to normal so you're forced to deactivate other ones first?

src/data/pokemon-forms.ts
```diff
@@ -86,7 +86,45 @@ export enum FormChangeItem {
...
+  FLAME_PLATE,
+  SPLASH_PLATE,
+  MEADOW_PLATE,
+  ZAP_PLATE,
+  MIND_PLATE,
+  ICICLE_PLATE,
+  DRACO_PLATE,
+  DREAD_PLATE,
+  PIXIE_PLATE,
+  BLANK_PLATE,  // Will this ever be used?
```
brain-frog, line 108: if the idea is to get back to normal you just deactivate, I don't think this is needed
unless the idea is you can catch a different form of arceus and use this to get it back to normal

src/data/pokemon-forms.ts
Madmadness65: Yes, this was the best solution without flooding the Pokemon Forms file with a hundred mix and matched form changes. Ideally new form change items picked up after one is already active should be immediately deactivated, but that's beyond the scope of this PR in particular.

src/data/pokemon-forms.ts
Madmadness65: Adding the Blank Plate was just for future-proofing, I guess. Yeah, it has no use currently, but it was just included (like the Legend Plate and custom Blank Memory) if someone has an idea for them in the future.

src/data/move.ts
Madmadness65: The code here needed to be updated anyways, as apparently type effectiveness on the form's moves broke at some point. It now uses `move.type` instead of `type.value`.

brain-frog: Oh yeah that just happened, fixes some other abilities

src/data/pokemon-forms.ts
brain-frog: fair enough, might figure out a use later
do you wanna switch it to `// TODO: figure out a use for this` just so it matches other todo comments

Madmadness65: Alright, I've changed the comments to match.

src/data/move.ts
brain-frog: Same question, can't you do `move.type = Type[form]` ?

brain-frog (approved):

PR #1197, opened 2024-05-21
Drop Shadow Adjustments

zaccie:
Adding support for x,y axis on dropshadow and adjusting the values to be less bad
and renamed the constants to actually make sense

dropshadow

### DIFF
--- a/src/ui/text.ts
+++ b/src/ui/text.ts
@@ -86,8 +86,8 @@ export function addTextInputObject(scene: Phaser.Scene, x: number, y: number, wi
 function getTextStyleOptions(style: TextStyle, uiTheme: UiTheme, extraStyleOptions?: Phaser.Types.GameObjects.Text.TextStyle): [ Phaser.Types.GameObjects.Text.TextStyle | InputText.IConfig, string, integer ] {
   const lang = i18next.language;
   let shadowColor: string;
-  let shadowXpos = 3;
-  let shadowYpos = 4;
+  let shadowXpos = 4;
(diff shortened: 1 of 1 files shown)


bennybroseph:
I'm not convinced a blanket 3, 4 value for distance is correct. The defaults for each text style wasn't 6, 6 nor 4, 4. Some were closer than that. I'd need a unique value for each style and an accompanying screenshot for what the new text looks like vs. the old one.

zaccie:
> I'm not convinced a blanket 3, 4 value for distance is correct. The defaults for each text style wasn't 6, 6 nor 4, 4. Some were closer than that. I'd need a unique value for each style and an accompanying screenshot for what the new text looks like vs. the old one.

Sure, the others were close enough I assumed it'd translate over well enough but I'll modify values for those specifics and compile some before/after images

zaccie:
left old right new
any other examples (like moveinfo content) is the same as on main
RegularText
TooltipContent

zaccie:
I haven't entirely convinced myself it's a meaningful change to be honest, but maybe others think its a much bigger improvement than I can see

bennybroseph (approved):

TempsRay:
Some are a lot worse now.
Before:

After:

zaccie:
> Some are a lot worse now. Before:  After: 

I didn't increase any numbers so this seems odd but I'll take a look and try fix it

zaccie:
Made some changes to fix above here: https://github.com/pagefaultgames/pokerogue/pull/1220

PR #1201, opened 2024-05-21
Overrides - Gender, Xp multiplier and arena tint (day & night) overrides

Greenlamp2:
i needed these overrides to test shiny sprites.
For example:

- override player pokemon: Genderless
- override opponent pokemon: Genderless
- override arena tint: Day

image


src/overrides.ts
```diff
@@ -57,6 +60,7 @@ export const STARTING_LEVEL_OVERRIDE: integer = 0;
 export const STARTER_SPECIES_OVERRIDE: Species | integer = 0;
 export const ABILITY_OVERRIDE: Abilities = Abilities.NONE;
 export const PASSIVE_ABILITY_OVERRIDE: Abilities = Abilities.NONE;
+export const GENDER_OVERRIDE: Gender = 0;
```
TempsRay, line 63: Hmm. 0 is specifically Gender.MALE so this will always override the player pokemon's gender to being male

src/field/arena.ts
```diff
@@ -442,7 +442,27 @@ export class Arena {
     }
   }
 
+  overrideTint(): [integer, integer, integer] {
+    if (!Overrides.ARENA_TINT_OVERRIDE) return;
+    let ret;
```
TempsRay, line 447: nit: instead of defining `ret` and then returning it, can't you just do a return in each of the `case` statements?

src/field/arena.ts
```diff
@@ -442,7 +442,27 @@ export class Arena {
...
+    switch(Overrides.ARENA_TINT_OVERRIDE) {
+      case TimeOfDay.DUSK:
+        ret = [ 98, 48, 73 ].map(c => Math.round((c + 128) / 2)) as [integer, integer, integer];
```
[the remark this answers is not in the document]
Greenlamp2: it's also hardcoded in the code, and i thought refactoring this would be too much to just this override
image

src/field/arena.ts
Greenlamp2: indeed

src/overrides.ts
Greenlamp2: indeed, it seems it's not following PlayerGender, let me fix that, good catch, thanks !
image
image

TempsRay (approved):
Yeah the hardcoding of tints is a bit weird, but it's not as easy to just straight up grab from existing code. Maybe for some future cleanup

src/overrides.ts
bennybroseph, line 1: Need some comments for how to use these new overrides.

src/field/arena.ts
```diff
@@ -455,7 +455,29 @@ export class Arena {
     }
   }
 
+  overrideTint(): [integer, integer, integer] {
+    if (!Overrides.ARENA_TINT_OVERRIDE) {
```
molker, line 459: is there a reason that's being checked here and before calling the function?

Greenlamp2: indeed, it's not needed since we enter in this method only if the overrides value is not null

molker (approved):

bennybroseph (approved):

PR #685, opened 2024-05-09
Menu - Gamepad menu (v1.0 Default gamepad selection)

Greenlamp2:
First iteration of the gamepad menu, for now, there is only the 2 "old" gamepad related entry + the new default gamepad selection.

I don't know if you prefer multiple small working PR with features added in this menu or if you want a big PR with all the gamepad related feature i'd like to implement:
- Gamepad selection
- Gamepad rebinding
- Activate/Deactivate turbo (holding a button to repeat the input)
- [... will maybe have more idea later ...]

[screen-capture (1).webm](https://github.com/pagefaultgames/pokerogue/assets/44787002/a193030f-c798-42da-9e42-b7a392322f57)

PR #1205, opened 2024-05-21
Menu - Controls Rebind - Gamepad & Keyboard

Greenlamp2:
Original PR: #685

# What are the changes?
This PR aims to enable players to rebind their gamepad buttons and keyboard keys.

# What are the "features"?
- Display the relevant action buttons in the menu according to the controls being used (buttons for gamepads and keys for keyboards).
- A new tab: Gamepad, where we can rebind the gamepad buttons.
- A new tab: Keyboard, where we can rebind the keyboard keys.
- For keyboards, we can have two sets of bindings per action: the main and the alternate one.
- In the Gamepad tab, we can select which connected controller to use and assign bindings to that specific controller.
- When we choose to change a binding for an action, a small window opens and prompts the user to press a button or key. This window auto-closes after 5 seconds if no input is detected.
- For a gamepad, when we choose a binding that is already assigned, it will ask for confirmation to swap the two buttons.
- For the keyboard, when we choose a binding that is already assigned, it will remove this binding from the previous action and assign it to the new one.
- There are some protected bindings that we can't change since they are important:
- For gamepads: D-pad buttons
- For keyboards: Arrows, Esc, Enter, Del, Home, Space
- These are displayed in gray (vs. yellow) and don't have a text saying: "Press action to assign".
- The bindings are persistent on refresh.
- On keyboard only, you can delete a selected binding with the Delete button.
- You can reset all the bindings you've done with the Home button.
- Prevent the deletion of the last binding for an action.
-Display the extended touch button (like on the starter selection screen) on the menu since you need two more buttons to navigate there.

# Screenshots
### Gamepad:
chrome_Fx9MhdXiPq

chrome_0gdqILUuOz

chrome_qHoXq6h4ZP

Playstation controller:
chrome_AxHqqe94OK

### Keyboard:
chrome_N7w97lKskq

chrome_qRvRQk9i1J

chrome_mW3xYw57r6

chrome_m4Ra3Sw21K

## How to test the changes?
You can test them on my test instance:
https://pokerogue.greenlamp.ovh/

### DIFF
--- a/src/data/pokemon-forms.ts
+++ b/src/data/pokemon-forms.ts
@@ -689,10 +689,10 @@ export const pokemonFormChanges: PokemonFormChanges = {
     new SpeciesFormChange(Species.DURALUDON, "", SpeciesFormKey.GIGANTAMAX, new SpeciesFormChangeItemTrigger(FormChangeItem.MAX_MUSHROOMS))
   ],
   [Species.ZACIAN]: [
-    new SpeciesFormChange(Species.ZACIAN, "hero", "crowned", new SpeciesFormChangeItemTrigger(FormChangeItem.RUSTED_SWORD))
+    new SpeciesFormChange(Species.ZACIAN, "hero-of-many-battles", "crowned", new SpeciesFormChangeItemTrigger(FormChangeItem.RUSTED_SWORD))
   ],
   [Species.ZAMAZENTA]: [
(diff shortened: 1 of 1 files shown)


Greenlamp2:
Todo:
- put back the "save the chosen controller"

Issue #1210, opened 2024-05-21
Beast Boost issue


TrainerX493:
Beast Boost (and similar abilities) should only activate when the user knocks out the Pokémon. Currently it triggers when a opponent is ko'ed in any way (Burn, ally in doubles, etc). Soul-Heart is the exception and does activate regardless of the ko source.

Shadowth117:
I've gone ahead and made a PR for this. It's something I saw too while playing.

Snailman11:
Link to the bug reports on Discord (Beast Boost activates after KO by recoil)
https://discord.com/channels/1125469663833370665/1232315301719441530

-Battle Bond Greninja seems to check for a dying target even after having fainted, but it also gets the boost from weather chip and not direct damage

Snailman11:
Still Active Post Late September 2024 Update

https://discord.com/channels/1125469663833370665/1232315301719441530/1290429000921120913

Snailman11:
**Videos are incorrect** need to check with correct replication steps

SirzBenjie:
I tried to replicate this and see if the issue persists.

I tried exactly the setup from the above video. Beast boost does not activate:

<details><summary>Beast boost does not activate after indirect damage</summary>

https://github.com/user-attachments/assets/4df849f0-3541-40ae-86ef-1fadf9f851b1
</details>

<details><summary>Battle bond does not activate after indirect damage</summary>

https://github.com/user-attachments/assets/528e7d05-8404-4b78-ab96-a0842c2b5ddf
</details>

PR #1211, opened 2024-05-21
Correct SoulHeart, Moxie, Beast Boost, etc.

Shadowth117:
Correct SoulHeart, Moxie, Beast Boost, Chilling Neigh, Grim Neigh, and As One abilities to activate the same way as their official versions as noted in this issue: https://github.com/pagefaultgames/pokerogue/issues/1210

The conditions were actually reversed from what the games did:
https://bulbapedia.bulbagarden.net/wiki/Soul-Heart_(Ability)
https://bulbapedia.bulbagarden.net/wiki/Beast_Boost_(Ability)

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -3435,7 +3435,7 @@ export function initAbilities() {
       .attr(PostDefendAbilityGiveAbAttr, Abilities.MUMMY)
       .bypassFaint(),
     new Ability(Abilities.MOXIE, 5)
-      .attr(PostVictoryStatChangeAbAttr, BattleStat.ATK, 1),
+      .attr(PostKnockOutStatChangeAbAttr, BattleStat.ATK, 1),
     new Ability(Abilities.JUSTIFIED, 5)
       .attr(PostDefendStatChangeAbAttr, (target, user, move) => move.type === Type.DARK && move.category !== MoveCategory.STATUS, BattleStat.ATK, 1),
     new Ability(Abilities.RATTLED, 5)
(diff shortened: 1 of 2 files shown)


TempsRay (approved):
LGTM

TempsRay (changes requested):
Actually on second thought, PostKnockOut might not be the right attr

Shadowth117:
> Actually on second thought, PostKnockOut might not be the right attr
Which would you use instead? Previously Soul Heart was using Post Knockout, but that appears to be wrong for it. It does however seem to be correct for the other abilities, which specifically describe that the pokemon needs to themselves directly knock out an opposing mon.

image

LaukkaE (changes requested):
This will not work. PostKnockOut is called for every pokemon that is alive on the battlefield when fainting occurs. This is what currently happens on your branch.

https://github.com/pagefaultgames/pokerogue/assets/73663099/e7238cb4-7a01-451b-b833-da5fbfae696e

Moxie will get boost even though it is not the pokemon that attacked the fainted pokemon.

Shadowth117:
> This will not work. PostKnockOut is called for every pokemon that is alive on the battlefield when fainting occurs. This is what currently happens on your branch.
> 
>  output.mp4 
> Moxie will get boost even though it is not the pokemon that attacked the fainted pokemon.

Ah, I understand, that makes sense.

Shadowth117:
Okay, I think the new commit should be what we want, after testing it a bit. I ended up adding the attacker to postknockout and splitting it to any and direct variation. The direct variation checks specifically for if the attacker of the knocked out pokemon's killing blow was from them and if it was an attack's direct damage that did them in. The other version is what it was, and makes sense for Soul Heart to keep after comparing things.

I assumed the logic in phases.ts had a pretty necessary order and so the new logic there preserves it. I'd imagine you might be able to condense the if checks some, but for the sake of sanity, I left it as is in case it were to be a big deal.

flx-sta:
Considered stale.

PR #1219, opened 2024-05-21
Implements Supreme Overlord and fixes Last Respects

Xavion3:
Currently the count resets every wave, but changing that is beyond the scope of this PR as that requires reworking when the game resets battle data.


TempsRay (approved):
Please also add in the PR title that it fixes last respects used by the opponent

src/battle.ts
```diff
@@ -63,6 +63,8 @@ export default class Battle {
     private battleSeedState: string;
     public moneyScattered: number;
     public lastUsedPokeball: PokeballType;
+    public playerFaints: number;
+    public enemyFaints: number;
```
TempsRay, line 67: wonder if these variables can be ambiguous in their naming, as in are they fainted pokemon on the player/enemy sides or if they're faints the player/enemy caused. Might be a mouthful but `faintedPokemonOnPlayerSide` and `faintedPokemonOnEnemySide` make it very explicit

Xavion3: They're also incorrect though, as the entire point of these variables as that these do *not* track the actual amount of fainted pokemon, but the amount of times pokemon have fainted. So I'm not sure if it's not so explicit its misleading in its own way.

bennybroseph: Ya know what would clear this up. Property Documentation :)

hayuna:
@Xavion3 There is an item called `REVIVER_SEED` which revive 1/2 HP of fainted pokemon. How this implementation is working with this item?

Counter of fainted pokemons is increasing each death. It means that even if pokemon will be fainted and revived, counter is still increasing. Maybe im thinking incorrectly but it should count *only* fainted pokemons

PR #1223, opened 2024-05-22
Allow opening menu in item reroll.

HighMans:
Allows you to enter the menu screen from the item selection screen after battle. Don't know why it's not allowed to begin with, if you go in to the transfer screen, then open the menu up -- it works.

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -3454,7 +3454,7 @@ export class TrainerVictoryPhase extends BattlePhase {
   }
 
   start() {
-    this.scene.disableMenu = true;
+    this.scene.disableMenu = false;
 
     this.scene.playBgm(this.scene.currentBattle.trainer.config.victoryBgm);
 
--- a/src/ui-inputs.ts
+++ b/src/ui-inputs.ts
@@ -110,6 +110,7 @@ export class UiInputs {
             case Mode.FIGHT:
             case Mode.BALL:
             case Mode.TARGET_SELECT:
+            case Mode.MODIFIER_SELECT:
             case Mode.SAVE_SLOT:
             case Mode.PARTY:
             case Mode.SUMMARY:


EvasiveAce:
Does not seem to work when you complete a trainer battle? I'll get a video real quick.

EvasiveAce:
> Does not seem to work when you complete a trainer battle? I'll get a video real quick.

I was trying to enter the menu, but was unable to. 

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/99dbb376-0257-4a73-ba89-4a2a731f7fef)

HighMans:
> > Does not seem to work when you complete a trainer battle? I'll get a video real quick.
> 
> I was trying to enter the menu, but was unable to.
> 
>  2024-05-21.20-55-37.mp4

Let me test that, I didn't check in trainer battles.

HighMans:
It should work with trainer battles now, but I'm going to need input from other devs as to why the menu isn't allowed to be entered after a trainer battle?

EvasiveAce:
> It should work with trainer battles now, but I'm going to need input from other devs as to why the menu isn't allowed to be entered after a trainer battle?

I wanna say the reason that you can't access the menu is because you're able to save and quit, leading to the previous battle to be void. I'm not sure if this is the case, but I would ask around in the Discord for the actual reason. It seems to work for me when I tried it though.

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/d93e2872-1fef-4919-b9ac-a5bef68a8697)

Greenlamp2:
Thanks for the PR, but unfortunately, we would like to enforce the "menu" opening to neutral state. (i.e on a turn beginning for example)
There is indeed still some places when it's not the case but we don't want to add more for sure until we fix these.
Other than facilitating save scumming, it doesn't bring more value and could potentialy bring more UI issues.

Issue #1225, opened 2024-05-22
Scrafty is unable to learn Trailblaze despite it being in its S/V learnset


FrijDom:
Scraggy and Scrafty's S/V learnset includes being able to learn Trailblaze by TM. However, not only will it never show up unless you have other pokemon that can learn it, but it comes up as 'not able' when attempting to teach it to either evolution stage.

PR #1228, opened 2024-05-22
Scrafty line does not have Trailblaze in their TM learnset

EvasiveAce:
# Fixes Issue #1225

Really easy fix; just added Scrafty and Scraggy to the TM list for Trailblaze.

Testing for it:

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/ab97fe05-7b28-405a-9c65-c4a7e4299371)


TempsRay (approved):
Wasn't even a DLC change, it's from vanilla SV. Odd that it was missed.

Madmadness65 (approved):

PR #1229, opened 2024-05-22
No more end to Endless

BatCrowP:
Fixing the crash bug that occurs when Endless runs reach floor 5850

### DIFF
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -1592,15 +1592,20 @@ export function getEnemyBuffModifierForWave(tier: ModifierTier, enemyModifiers: 
   const tierStackCount = tier === ModifierTier.ULTRA ? 5 : tier === ModifierTier.GREAT ? 3 : 1;
   const retryCount = 50;
   let candidate = getNewModifierTypeOption(null, ModifierPoolType.ENEMY_BUFF, tier);
-  let r = 0;
-  let matchingModifier: Modifiers.PersistentModifier;
-  while (++r < retryCount && (matchingModifier = enemyModifiers.find(m => m.type.id === candidate.type.id)) && matchingModifier.getMaxStackCount(scene) < matchingModifier.stackCount + (r < 10 ? tierStackCount : 1))
-    candidate = getNewModifierTypeOption(null, ModifierPoolType.ENEMY_BUFF, tier);
-
-  const modifier = candidate.type.newModifier() as Modifiers.EnemyPersistentModifier;
-  modifier.stackCount = tierStackCount;
-
-  return modifier;
+  if (candidate != null) {
+    let r = 0;
+    let matchingModifier: Modifiers.PersistentModifier;
+    while (++r < retryCount && (matchingModifier = enemyModifiers.find(m => m.type.id === candidate.type.id)) && matchingModifier.getMaxStackCount(scene) < matchingModifier.stackCount + (r < 10 ? tierStackCount : 1))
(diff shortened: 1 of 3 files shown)


MooGoesCow21:
Yeah this would be nice especially since there is a lot of people who have a big advantage with bugged runs that can do this already

Reibd:
This would be nice for people who want to continue their current run past 5850

Greenlamp2:
Hello, could you explain why endless was crashing initially and what did you do to fix it please ?

BatCrowP:
Endless currently crashes upon reaching a full token count. It tries to find a new token to give to players and finds nothing so it returns null and proceeds to crash. 
The change I made was a tweak to getEnemyBuffModifierForWave that checks if the candidate token is a null and doesn't try to add the token if it is null. If the token isn't null it lets the function operate exactly as it currently does.

BatCrowP (commented):
well this was an accident, meant to make a separate PR for the luck based encounters

src/modifier/modifier-type.ts
```diff
@@ -1592,15 +1592,20 @@ export function getEnemyBuffModifierForWave(tier: ModifierTier, enemyModifiers:
...
+    let r = 0;
+    let matchingModifier: Modifiers.PersistentModifier;
+    while (++r < retryCount && (matchingModifier = enemyModifiers.find(m => m.type.id === candidate.type.id)) && matchingModifier.getMaxStackCount(scene) < matchingModifier.stackCount + (r < 10 ? tierStackCount : 1))
+      candidate = getNewModifierTypeOption(null, ModifierPoolType.ENEMY_BUFF, tier);
+    
+    const modifier = candidate.type.newModifier() as Modifiers.EnemyPersistentModifier;
+    modifier.stackCount = tierStackCount;
+  
+    return modifier;
+    } else {
```
Greenlamp2, line 1605: alignment issue + unneeded ELSE since you got a return in the previous IF

Greenlamp2 (changes requested):
In phases.ts, getEnemyBuffModifierForWave is called in this.scene.addEnemyModifier.
what happens there when it's null ?
(did you test your code on a real run ? if yes, could you provide an exported session ?)

src/modifier/modifier-type.ts
BatCrowP: [SessionData.zip](https://github.com/pagefaultgames/pokerogue/files/15418736/SessionData.zip)
Attached are two runs, one that I tested progressing normally and was still getting tokens, the 5852 is one that I tested a run that is currently frozen on 5850. When null is passed to addEnemyModifier then it will resolve after finding that the null is not a terrastilize or enemy modifier. No error messages when testing on either run

Greenlamp2 (approved):

bennybroseph (changes requested):
needs conflicts resolved

src/modifier/modifier-type.ts
Greenlamp2: are you sure these are the good save files ? None of these are stuck on 5850, and whatever i do, on the main branch, i can't make it crash.
image

BatCrowP: [sessionData_Krow.prsv.zip](https://github.com/pagefaultgames/pokerogue/files/15440737/sessionData_Krow.prsv.zip)
Attached is the 5850 file as it currently stands on the main branch.

BatCrowP:
Anything that I can do here or is this dead?

Xavion3:
Yeah, an alternative solution was implemented in #1463

PR #1231, opened 2024-05-22
Make Magician and Pickpocket have a chance to proc to match Covet and other itemsteal mechanics

Fontbane:
As it is, fighting a boss Meowscarada or anything else with Magician is just like going against Mini Black Hole. Having items stolen every hit with Magician is inconsistent with other item stealing mechanics implemented with the game's unique item system in mind, like Covet and Grip Claw. So, this PR gives Magician and Pickpocket a chance to proc, rather than being guaranteed. Magician has a 1/3 chance to steal, and Pickpocket has a 1/2 chance. Pickpocket has a higher chance because there's at least a way to avoid it, by not using contact moves. These numbers can be adjusted if need be.

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -1135,19 +1135,19 @@ export class PostAttackAbAttr extends AbAttr {
 }
 
 export class PostAttackStealHeldItemAbAttr extends PostAttackAbAttr {
-  private condition: PokemonAttackCondition;
+  private chance: integer;
 
-  constructor(condition?: PokemonAttackCondition) {
+  constructor(chance?: integer) {
     super();
 
-    this.condition = condition;
+    this.chance = chance;
   }
 
   applyPostAttack(pokemon: Pokemon, passive: boolean, defender: Pokemon, move: PokemonMove, hitResult: HitResult, args: any[]): Promise<boolean> {
     return new Promise<boolean>(resolve => {
-      if (hitResult < HitResult.NO_EFFECT && (!this.condition || this.condition(pokemon, defender, move.getMove()))) {
+      if (hitResult < HitResult.NO_EFFECT) {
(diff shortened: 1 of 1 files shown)


OrangeRed:
Rather than hard coding these values into the class can you add them as constructor params and use those instead
like this
```ts
constructor(condition?: PokemonAttackCondition, procChance: number) {
  super()

  this.condition = condition
  this.procChance = procChance
}
```

That way the proc chance is tied to the individual ability

Fontbane:
> Rather than hard coding these values into the class can you add them as constructor params and use those instead like this
> 
> ```ts
> constructor(condition?: PokemonAttackCondition, procChance: number) {
>   super()
> 
>   this.condition = condition
>   this.procChance = procChance
> }
> ```
> 
> That way the proc chance is tied to the individual ability

I considered this tbh, I think it would be cleaner if I did actually.

src/data/ability.ts
```diff
@@ -1224,18 +1224,20 @@ export class PostAttackApplyBattlerTagAbAttr extends PostAttackAbAttr {
 
 export class PostDefendStealHeldItemAbAttr extends PostDefendAbAttr {
   private condition: PokemonDefendCondition;
+  private chance: integer; //Proc chance out of 100
 
-  constructor(condition?: PokemonDefendCondition) {
+  constructor(condition?: PokemonDefendCondition, chance?: integer) {
```
OrangeRed, line 1229: I'm pretty sure we use fractions and decimals everywhere for probability. Can you make this consistent with that?

Also we recently started asking everyone to write jsdoc comments so could you do that too?

Fontbane: Actually we use integers for all the other chance parameters for ability attributes. I'll add the comments though.

brain-frog:
I appreciate the added documentation.I do not see a reason to add probability to an ability that is 100% chance normally.

PR #1232, opened 2024-05-22
Add 5x5 option to gacha

Tempo-anon:
## Motivation

The epic egg exclusives are where a lot of Pokemon are gated behind and currently there is no way to guarantee getting one. This issue is exacerbated by the recent addition of more Pokemon in the epic egg pool. With a median egg count of 12960(!) eggs required to get all the epic egg exclusive Pokemon. This change brings it down to 9000.

## Testing
New UI 

Gold still works as expected 

Purple still works as expected 

New 5x5 option 

Plus still works as expected 

10x1 still works as expected 

1x1 still works as expected 

And yes I verified the cancel button functionality still exists


HopsWas:
The players, they yearn for the egg pity! This feature is long awaited!

star-krieg:
Adding here the analysis to what Temp posted (I'm the one who made the epic egg simulation in Python)

Each realization is an independent set of epic eggs pulled until eggsclusives are completed. I just used the available data for egg weights from Wiki (with some corrections because Manaphy is not in epic egg, and there are additions to epic eggs after the costs have been change).

The blue line is the simulation from when Victini and Deoxys were demoted to epic eggs, while the orange line is the 2nd cost update (Regigas etc. to epic eggs)

As it stands, you would need on average, 350 epic eggs to complete everything (roughly 12 000 eggs pulled total). The 10th and 90th percentiles are also included to give context to the possible worst case scenario for some people. Adding some form of pity should decrease the number a bit, depending on what kind of epic egg guarantee is implemented (epic egg rate up for eggsclusives, X number of eggs pulled for guaranteed epics, etc.)

AlphaIce:
Genuinely think this would be a great addition for the future, as right now getting epic eggs is very time consuming and honestly downright dreadful, even when farming vouchers on a daily basis.

c4vv:
This is a very reasonable addition to the game that lessens the grind for the starter dex without trivializing it entirely. Looks wonderful!

Reibd:
Looks like a fair and balanced way to make the grind easier. Great job!

HopsWas:
> Excuse me, I don't think I understand, how this PR reduce the probability of getting epic eggsclusive ? All i read is UI modification to add the use of 5 given tickets. Is this a suite of multiple PR aiming for this ?

This change uses the feature already built into the code, where a golden 25 pull ticket, typically only available via beating classic for the first time, pulls 1 guaranteed epic egg. with this change, you are essentially converting 5 5 pull tickets into one golden 25 pull to gain that guaranteed epic egg.

Greenlamp2:
> > Excuse me, I don't think I understand, how this PR reduce the probability of getting epic eggsclusive ? All i read is UI modification to add the use of 5 given tickets. Is this a suite of multiple PR aiming for this ?
> 
> This change uses the feature already built into the code, where a golden 25 pull ticket, typically only available via beating classic for the first time, pulls 1 guaranteed epic egg. with this change, you are essentially converting 5 5 pull tickets into one golden 25 pull to gain that guaranteed epic egg.

thank you for you answer, i got it now ! (and i guess the 10x single egg voucher is just QOL ?)

RNGezeus:
> thank you for you answer, i got it now ! (and i guess the 10x single egg voucher is just QOL ?)

If i'm reading code right then 10x pull has guaranteed rare+ which is not mutually exclusive to the 25x guarantee so a 25x also contains the 10x guaranteed rare.

As a player this is great for better progress, but from a design perspective I get really annoyed that this makes 5x vouchers strictly better than 10x vouchers, similar to how 1x are better than 5x currently for same reason.

Torch1e:
Hi! I 100% agree with the need for an egg pity system, the system currently is poorly structure and prohibitive, however this system in my opinion is needlessly complicated and poor for new players.

In my mind, the egg pity system exists for two kinds of players primarily, new players looking to get an edge, and old players looking to fill out their dex of eggclusives. I believe this system only really helps the second group, and would create a negative experience for the first.

As a relatively new player, I spent most runs early on getting to sub 100 (90/80) and beating 2 or 3 gym leaders. With this proposed system I would be more or less required to hoard those tickets early until I get two runs. This forces me to have two good runs without pulling on the gacha system, which in my opinion feels bad. Especially early it feels so good to pull, and getting that early shiny, legendary or rare egg move is INSANE for progression. For example, I pulled a sceptrier early and it carried me to 200 instantly, truly opened up my progression. 

In this way, Id prefer a generic pity system that allows players to pull at their leisure and still get the more often epics and the like. I do like the idea of mass pulling x5s, as they can be quite slow, but I think the extra hoop to jump through is prohibitive to new players, and creates a negative experience for them. Either they are hurt for pulling early without the 100% epic chance, or they are hurt by holding onto those tickets and not getting potential power boosts from eggs. Further, the fact that x25 pulls give a 100% chance for an epic or legendary isnt exactly explicitly stated, so many new players wouldn't even be aware they are missing out or making suboptimal choices, which would feel VERY bad once they found out. In my opinion no matter how you look at it, a generic pity system beats it out the 5x5 system for solving the very real epic egg shortage.

jaimefd:
This is a great change, but it made me think about wanting to save 25 normal vouchers too, and that made me think this could all be made easier if we either:
- Allow exchanging vouchers for their higher tier version
- Make pull options independent of vouchers, so you choose 1x 5x 10x or 25x and then vouchers are spent accordingly using higher value ones first

j-diefenbach:
Honestly converting vouchers to automatically upgrade from 5 1Xs to  a 5X, then from 2 5Xs to a 10X sounds like a good idea to me. I feel like 10 eggs is still enough resolution / fine control if people don't want to throw 25 tickets at the same machine (its only 1-2 more pulls if we cap auto upgrade to 10x)

Icehawk78:
One thing to note with this (and related changes) - this will encourage people to hoard tickets, and there is apparently [code on the server side of things](https://github.com/pagefaultgames/rogueserver/blob/b113ffceee027bb3722ae8b3c8bc9bc61c4f919e/api/savedata/update.go#L45) that will auto-ban any account with too many of any given voucher type, presumably as an anti-cheat since it's quite easy to spawn in more vouchers locally.

Just something to keep in mind.

Tsukistar1:
> One thing to note with this (and related changes) - this will encourage people to hoard tickets, and there is apparently [code on the server side of things](https://github.com/pagefaultgames/rogueserver/blob/b113ffceee027bb3722ae8b3c8bc9bc61c4f919e/api/savedata/update.go#L45) that will auto-ban any account with too many of any given voucher type, presumably as an anti-cheat since it's quite easy to spawn in more vouchers locally.
> 
> Just something to keep in mind.

Which seems kinda stupid if true, cause people hoard vouchers for things like specific legendary gachas.

chacolah:
> One thing to note with this (and related changes) - this will encourage people to hoard tickets, and there is apparently [code on the server side of things](https://github.com/pagefaultgames/rogueserver/blob/b113ffceee027bb3722ae8b3c8bc9bc61c4f919e/api/savedata/update.go#L45) that will auto-ban any account with too many of any given voucher type, presumably as an anti-cheat since it's quite easy to spawn in more vouchers locally.
> 
> Just something to keep in mind.

While this is something to worry about, in my experience chatting with people who run endless to grind x5 tickets, they are already hoarding them extensively for specific legend banners, I don't foresee this being as impactful as it could seem at the get go.

Icehawk78:
From testing with others, it seems as though possibly that logic is either not being applied, or has no effect currently, so I think it's not currently an issue, but still something I wanted to point out regardless, since it seems like there was an intent, even if not successful, to do that at some point.

Tempo-anon:
Closing this since egg pity has been added. I spent an hour doing UI work for nothing 😭

Issue #1208, opened 2024-05-21
Enemy's Last Respects scales from players fainted pokemon


LaukkaE:
The check for Last Respect's power always checks players side. (scene.getParty())
This causes AI's side to scale from players fainted pokemon.

![lastrespects](https://github.com/pagefaultgames/pokerogue/assets/73663099/9e5f8619-a9d3-42b0-8e54-3b92e4c50348)

Wild Opponents pokemon is hitting here for 150 bp with two fainted allies.

Shadowth117:
I verified and went ahead and made a pull request for this. 
https://github.com/pagefaultgames/pokerogue/pull/1233

Shadowth117:
Someone did more of an overhaul to that setup for supreme overlord and so this is probably the go forward pr from here:
https://github.com/pagefaultgames/pokerogue/pull/1219

Tempo-anon:
Fixed

PR #1233, opened 2024-05-22
Get the correct faction's party size for Last Respects

Shadowth117:
This is for resolving: 
https://github.com/pagefaultgames/pokerogue/issues/1208

Code change checks for if you're the player or not and gets the appropriate party accordingly. 

Of note is currently for wild battles, the 2 wild ones on the field are part of the same 'party' even though wild pokemon do not normally have a party. In my opinion, this should just count anyways since you figure they're fighting together against the player and don't attack each other, alleging at least a loose comradery. Were horde battles, the 5 v 1 battles seen in the official games, added, I would argue the same for those as well.

If the above note isn't desirable, it should be as simple as checking if the enemy is wild and then setting party count to 1 if it is, though I feel that'd also just be less interesting for gameplay.


TempsRay:
Already done in #1219

Shadowth117:
> Already done in #1219

The other PR integrates it nicely so I think I can just close this then.

PR #1096, opened 2024-05-18
[Draft] Add Flyout Menus to Battle Stats UI

bennybroseph:
Adds a small flyout menu in the battle UI. The original idea was to add enemy movesets here, but there could be other things here as well.

TODO:

- ~~Support Legacy UI~~
- ~~Support Boss UI~~
- ~~Add something to the small flyout~~
- ~~Draft and add a second larger flyout~~ 
  - ~~Current idea is to show Weather/Terrain/Tailwind/Trick Room/etc.~~
  - ~~This would flyout from top to bottom or vice versa so that it is wider and taller~~
- ~~Track move usage to determine if a move has been seen~~
- Track initial Weather/Terrain/Tag duration

https://github.com/pagefaultgames/pokerogue/assets/13838608/9308b324-cb04-4ee1-b68b-5b5873a5541a

image

image

PR #1235, opened 2024-05-22
Add Move Usage Tracking

bennybroseph:
Adds tracking to determine if a move has been used before. This change is specifically for #1096, but I'd like to propose this before hand so that I include as few other changes as possible in the Flyout feature. 

Currently restructuring this.


Greenlamp2 (approved):

TempsRay (commented):
Pretty cool from what I've seen so far. Maybe don't show the PP for the opponent though? Since in classic there's a case of a hardcoded move having only 1 PP

Issue #222, opened 2024-04-21
Weird end-of-wave knockout damage interactions


Eridanus91:
If a Pokemon is let's say, poisoned, it can get knocked out at the end of a wave. This creates some... interesting quirks.

If Pokemon in slot 1 gets knocked out like that in a double battle, the next wave softlocks until you restart the page, because Pokemon in slot 2 does not get properly removed and the game throws back the dead mon in, so the game becomes just enemy spamming their moves to no avail (hey, this sounds like that one glitch from Generation V games!).

![obraz0](https://github.com/pagefaultgames/pokerogue/assets/26799309/f1542d64-1f77-46d1-bec2-99fd501615ad)

If this happens during an area boss fight, it will cause certain graphics to not load, likely because the game forces you to switch out the dead mon in middle of everything.

![obraz](https://github.com/pagefaultgames/pokerogue/assets/26799309/121e8b17-7baa-43a1-bf2c-15ac367c43da)

Lesys:
I got the exact same bug as OP. I don't know if it's because the double battle before (I did have a double battle before the bug occurring) but my Crocalor died from poison, then next fight I summoned Gyarados and the following picture explains what happened.

Fun fact that I was almost in the same level as OP with the same mon that died from poison :D

PS: I waited for the Ninjask to use all its PP and use Struggle, but it doesn't seem to have Struggle. Instead, its passive looped forever. I couldn't use any interaction, even ESC.

![image](https://github.com/pagefaultgames/pokerogue/assets/10529124/4805cc0b-dfdc-4bd9-8af7-1a9289ee2177)

PS 2: Thanks for the game ! Have been really fun the last 2 days I started it.

knispeja:
Duplicated by:
- https://github.com/pagefaultgames/pokerogue/issues/253
- https://github.com/pagefaultgames/pokerogue/issues/894

JoshuaKeegan3:
Fun fact. If the sprites are vanished from say desinty bond or toxic, Destroying a mimikyu disguise freezes the game

flx-sta:
AAAANd this is actually a dupllicate of #64

Issue #894, opened 2024-05-15
Player and pokemon list can sometimes overlap with battle sprites


PLMMJ-FH:
So I was playing casually and one of my mons (I don't remember which) died from poison at the end of a battle, but then I revived it with shop items. The game had me switch out anyway and then my player sprite persisted into battles. This only ended once I beat the boss.
![PokerogueMayBeBroken](https://github.com/pagefaultgames/pokerogue/assets/159862724/9aea5663-0b0e-42aa-83e0-4a8f4c3f8c47)

PLMMJ-FH:
Can confirm this only happens if the trainer sprite comes out.

knispeja:
Reproduced this locally by fainting with poison then reviving the mon. Technically the trainer sprite doesn't overlap in my example, but it's just because it wasn't followed by a trainer battle. I think this is the same bug:

https://github.com/pagefaultgames/pokerogue/assets/9343160/3261ab22-37c2-4b69-b171-8f029ce1b67f

flx-sta:
Duplicate of #222

Issue #253, opened 2024-04-23
Trainer is persistant if a trainer fight start after a pokemen die at the end of a fight from poison


Salixe1214:
This is a bit specific and doesn't affect the playability, but when a pokemon die from poison at the end of a fight and is revive with an item, if the next fight is a trainer fight, you need to summon a new pokemon that will live on top of the trainer. The trainer will now live on the field for all the fight.

![image](https://github.com/pagefaultgames/pokerogue/assets/49760514/b909735d-3daa-49f1-9eea-eaaf72f5942e)
![image](https://github.com/pagefaultgames/pokerogue/assets/49760514/35da293b-e6da-4286-b023-b5c922a3f390)

knispeja:
Technically this does affect playability, it forces you to switch away from the Pokemon you revived for no reason. More detail in my comment here: https://github.com/pagefaultgames/pokerogue/issues/894#issuecomment-2124188396

This pretty much is a duplicate of these:
- https://github.com/pagefaultgames/pokerogue/issues/222
- https://github.com/pagefaultgames/pokerogue/issues/894

flx-sta:
Duplicate of #253

PR #1236, opened 2024-05-22
Skip forced switch if the fainted party member has been revived already

knispeja:
I need to test this further, but it fixes https://github.com/pagefaultgames/pokerogue/issues/222 and duplicates (https://github.com/pagefaultgames/pokerogue/issues/894, https://github.com/pagefaultgames/pokerogue/issues/253) as far as I can tell.

Before:
https://github.com/pagefaultgames/pokerogue/assets/9343160/3261ab22-37c2-4b69-b171-8f029ce1b67f

After:
https://github.com/pagefaultgames/pokerogue/assets/9343160/cf1d2aaa-8d58-4f5c-80c4-f4c03c23bd23


Tempo-anon:
Tested by Starkrieg https://discord.com/channels/1125469663833370665/1244839350919888897/1269222026380574783

snoozbuster:
I was working on https://github.com/pagefaultgames/pokerogue/pull/3321 which changes u-turn and baton pass (etc) to use SwitchPhase instead of a custom function. this PR breaks that code cause a uturn user is not and was never fainted, so the phase ends immediately without switching.

@knispeja - is there somewhere else you were looking at where we could put this fix? I don't want to revert this bug but it does make fixing things the way I was trying to much more complicated

PR #1242, opened 2024-05-22
Let Jirachi be able to be caught in space [Puzzle]

Tempo-anon:
Based on the idea [here](https://github.com/pagefaultgames/pokerogue/pull/787) but simplified the logic. Also refactored some of the chained ternary operations into more readable if/else

Allows Jirachi to be caught in Space at UR odds if the player solves the puzzle.

Hint: kn6-c223DUU


src/data/biomes.ts
```diff
@@ -34,6 +35,50 @@ interface BiomeDepths {
...
+  [Biome.VOLCANO]: () => {return false},
+  [Biome.GRAVEYARD]: () => {return false},
+  [Biome.DOJO]: () => {return false},
+  [Biome.FACTORY]: () => {return false},
+  [Biome.RUINS]: () => {return false},
+  [Biome.WASTELAND]: () => {return false},
+  [Biome.ABYSS]: () => {return false},
+  [Biome.SPACE]: (p) => ( 
+    (!!p.scene.getParty().find(p1 => p1.moveset.find(m1 => m1.getMove().id === 273 || m1.getMove().id === 361))) &&
+    (!!p.scene.getParty().find(p2 => p2.moveset.find(m2 => m2.getMove().id === 47)))
```
TempsRay, line 70: I've added some obfuscation but not sure how futureproof this will be 🤔

Xavion3:
I mentioned this a bit in the discord, but currently I'm not a fan because I believe this fundamentally and totally fails at being a puzzle. The idea of puzzles is fine, but I feel like if we want puzzles we should try to include some actual puzzles, not just an easter egg that is only findable by luck or code reading.

knispeja:
> I mentioned this a bit in the discord, but currently I'm not a fan because I believe this fundamentally and totally fails at being a puzzle. The idea of puzzles is fine, but I feel like if we want puzzles we should try to include some actual puzzles, not just an easter egg that is only findable by luck or code reading.

I tend to agree, maybe some trainer / new evil team dialogue could give hints towards what to do?

ben-lear:
@Tempo-anon you want to turn this into a Mystery Encounter?

Tempo-anon:
Sacrifices must be made in order to get the PR number down. This can be a proof of concept for situational biome spawns.

PR #1245, opened 2024-05-22
Implement Poison Heal Ability

EvasiveAce:
# Implements the [Poison Heal Ability](https://bulbapedia.bulbagarden.net/wiki/Poison_Heal_(Ability))
Please let me know if there is anything I can change or improve on. Thank you.

Video to show that Toxic works as intended:

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/bfe71a71-fd84-4f47-acb8-52748333cf80)

Video to show that Poison Powder works as intended;

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/1bca288f-bb39-4a61-a92c-a5af5766e6d1)


src/data/ability.ts
```diff
@@ -2198,6 +2214,23 @@ export class PostTurnAbAttr extends AbAttr {
...
+   * @param {any[]} args N/A 
+   * @returns Returns true if Poison Heal procs
+   */
+  applyPostTurn(pokemon: Pokemon, passive: boolean, args: any[]): boolean | Promise<boolean> {
+    if (pokemon.status.effect === StatusEffect.TOXIC || pokemon.status.effect === StatusEffect.POISON) {
+      pokemon.heal(Math.max(Math.floor((pokemon.getMaxHp() / 8))));
+      pokemon.updateInfo();
+    }
+    return true;
+  }
```
TempsRay, line 2231: Need a `return false;`?

src/data/ability.ts
```diff
@@ -1889,6 +1889,22 @@ export class BlockNonDirectDamageAbAttr extends AbAttr {
...
 }
 
+export class BlockPoisonToxicDamageAbAttr extends BlockNonDirectDamageAbAttr {
+  /**
+   * This is so we don't have to edit each switch case in phase.ts, if the damage is Toxic or Poison, it will cancel.
+   * @param {Pokemon} pokemon The pokemon with the ability Poison Heal
+   * @param {boolean} passive N/A
+   * @param {Utils.BooleanHolder} cancelled Whether to cancel the Toxic/Poison damage or not
+   * @param {any[]} args N/A
+   * @returns Returns true if the damage from Toxic/Poison is blocked
```
TempsRay, line 1899: then it should look like 
```
if (pokemon.status.effect === StatusEffect.TOXIC || pokemon.status.effect === StatusEffect.POISON) {
  cancelled.value = true;
  return true;
}
return false;
```

TempsRay (commented):
Code looks mostly okay to me, just had some comments.

Can you also test this with things like neutralizing gas, skill swap, and toxic?
> When a Pokémon with Poison Heal is badly poisoned, this has no effect on the Pokémon's HP recovery. However, the bad poison counter still increases each turn even if the Pokémon has Poison Heal; as such, if [Gastro Acid](https://bulbapedia.bulbagarden.net/wiki/Gastro_Acid_(move)) is used on the Pokémon with Poison Heal after being badly poisoned for five turns, it will incur poison damage starting at 5/16 (instead of 1/16).

EvasiveAce:
> Code looks mostly okay to me, just had some comments.
> 
> Can you also test this with things like neutralizing gas, skill swap, and toxic?
> 
> > When a Pokémon with Poison Heal is badly poisoned, this has no effect on the Pokémon's HP recovery. However, the bad poison counter still increases each turn even if the Pokémon has Poison Heal; as such, if [Gastro Acid](https://bulbapedia.bulbagarden.net/wiki/Gastro_Acid_(move)) is used on the Pokémon with Poison Heal after being badly poisoned for five turns, it will incur poison damage starting at 5/16 (instead of 1/16).

I have done some of the updates you thought would be good. @TempsRay 

Video to show the new notification system (only showAbility when actually healing)

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/e397b344-f36e-4fd3-9c0d-feddc635dec6)

Video to show Neutralizing Gas:

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/f35893e2-0479-4a2a-90b7-f7e255d1d049)

Video to show Skill Swap

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/9b9583cf-2cdc-4566-8173-787cbdf61d60)

For Toxic, it should be in the original post, the first video.

src/data/ability.ts
```diff
@@ -2293,6 +2312,29 @@ export class PostTurnAbAttr extends AbAttr {
   }
 }
 
+export class PostTurnPoisonHealAbAttr extends PostTurnAbAttr {
```
Tempo-anon, line 2315: Can you add some documentation on this class and the BlockPoisonToxicDamageAbAttr class to explain at a quick glance what it does and where it's used (Poison Heal)?
```
/**
  * Ability attribute to heal the user if the user is poisoned or toxic'd. Used by Poison Heal
  */
```

EvasiveAce: Is the newest commit better?

EvasiveAce:
## I changed the way healing is done from Heal() to a HealPhase

Video to show how Toxic works with the new phase:

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/11380af1-2291-4eca-a115-db646563cf8a)

Video to show the same thing but with Poison Powder:

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/d2269be1-9af9-4052-8551-b5b45d04e049)

src/data/ability.ts
```diff
@@ -2293,6 +2324,41 @@ export class PostTurnAbAttr extends AbAttr {
...
+   * @param {any[]} args N/A
+   * @returns Returns true if healed from status, false if not
+   */
+  applyPostTurn(pokemon: Pokemon, passive: boolean, args: any[]): boolean | Promise<boolean> {
+    if (this.effects.includes(pokemon.status.effect)) {
+      if (pokemon.getMaxHp() !== pokemon.hp) {
+        const scene = pokemon.scene;
+        const abilityName = (!passive ? pokemon.getAbility() : pokemon.getPassiveAbility()).name;
+        scene.unshiftPhase(new PokemonHealPhase(scene, pokemon.getBattlerIndex(),
+          Math.max(Math.floor(pokemon.getMaxHp() / (8 / 1)), 1), getPokemonMessage(pokemon, `'s ${abilityName}\nrestored its HP a little!`), true));
```
Tempo-anon, line 2354: This should be localized https://github.com/pagefaultgames/pokerogue/blob/main/src/locales/en/ability-trigger.ts

Tempo-anon, line 2354: Is the (8 / 1) needed?

Tempo-anon (approved):

PR #1250, opened 2024-05-22
Give Certain Starters a Reliable Level 1 Damaging Move

Blitz425:
A simple QOL change to certain Pokemon who otherwise didn't start with any sort of damage dealing move, or started out with unreliable moves that made the first biome impossible to get through such as Wrap/Bind or Astonish/Flail. The move choices were simply to help the early game for these Pokemon and a majority of these moves were moves the Pokemon had access to at some point save for some special cases. The following Pokemon were affected: Abra, Gastly, Lickitung, Dunsparce, Slugma, Wailmer, Spoink, Shuppet, Duskull, Burmy, Chingling, Bonsly, Carnivine, Giratina, Darkrai, Throh, Vanillite, Pumpkaboo, Bounsweet, Pyukumuku, Cosmog, Magearna, Applin, Sinistea, Poltchageist.

### DIFF
--- a/src/data/pokemon-level-moves.ts
+++ b/src/data/pokemon-level-moves.ts
@@ -1055,13 +1055,13 @@ export const pokemonSpeciesLevelMoves: PokemonSpeciesLevelMoves = {
   ],
   [Species.ABRA]: [
     [ 1, Moves.TELEPORT ],
+    [ 1, Moves.CONFUSION ],
   ],
   [Species.KADABRA]: [
-    [ 0, Moves.CONFUSION ],
+    [ 0, Moves.PSYBEAM ],
     [ 1, Moves.DISABLE ],
     [ 1, Moves.TELEPORT ],
     [ 1, Moves.KINESIS ],
-    [ 5, Moves.PSYBEAM ],
     [ 10, Moves.REFLECT ],
     [ 15, Moves.ALLY_SWITCH ],
     [ 20, Moves.PSYCHO_CUT ],
@@ -1542,6 +1542,7 @@ export const pokemonSpeciesLevelMoves: PokemonSpeciesLevelMoves = {
   [Species.GASTLY]: [
     [ 1, Moves.CONFUSE_RAY ],
     [ 1, Moves.LICK ],
+    [ 1, Moves.ACID ],
     [ 4, Moves.HYPNOSIS ],
     [ 8, Moves.MEAN_LOOK ],
     [ 12, Moves.PAYBACK ],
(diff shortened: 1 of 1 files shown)


TempsRay (commented):
Please add `// custom` after custom moves as seen [here](https://github.com/pagefaultgames/pokerogue/blob/main/src/data/pokemon-level-moves.ts#L8482)

src/data/pokemon-level-moves.ts
```diff
@@ -1055,13 +1055,13 @@ export const pokemonSpeciesLevelMoves: PokemonSpeciesLevelMoves = {
   ],
   [Species.ABRA]: [
     [ 1, Moves.TELEPORT ],
+    [ 1, Moves.CONFUSION ],
   ],
   [Species.KADABRA]: [
-    [ 0, Moves.CONFUSION ],
+    [ 0, Moves.PSYBEAM ],
```
TempsRay, line 1061: Does Kadabra matter? Kadabra is not a starter

src/data/pokemon-level-moves.ts
```diff
@@ -13742,6 +13761,7 @@ export const pokemonSpeciesLevelMoves: PokemonSpeciesLevelMoves = {
   [Species.MAGEARNA]: [
     [ 1, Moves.HELPING_HAND ],
     [ 1, Moves.GYRO_BALL ],
+    [ 1, Moves.FAIRY_WIND ],
```
TempsRay, line 13764: Does Magearna need this when it has STAB gyro ball? There can be an argument for Giratina based on normal mons in the early routes but I think Magearna is fine

src/data/pokemon-level-moves.ts
Blitz425: I talked to Madness about it, Kadabra's move on evolution you get is Confusion. I gave Abra Confusion and made Psybeam the evolution move as it was a relearn for Kadabra regardless.

TempsRay:
I feel like custom moves prior to level 5 should only apply if there's good reason for doing so like the starter costing a LOT or the early game is unplayable without egg moves. For a Pokemon like Abra I think it's fair game that if you want to use it in your run you will have to babysit it a bit with your other Pokemon.

src/data/pokemon-level-moves.ts
Blitz425: Gyro Ball only has 5 PP and is unreliable since Magearna has a decent speed tier in early game biomes. It is not necessary but that was the justification behind that.

Madmadness65 (approved):

PR #1251, opened 2024-05-22
[Move][Bug] Fix Enemy Fake Out and Force Switch Interaction

### LINKED ISSUE #368: Fake Out fails when a switch is forced
When a pokemon is switched out by Dragon Tail (and possibly other forced switch moves, I haven't tried all of them) and the player  then selects fake out on the switched-in pokemon, the move will fail despite having been selected on the pokemon's first actionable turn.

torranx:
#347 fixes #368 BUT the game right now only checks the last used move of the **PLAYER'S** pokemon. 

This would mean if player used a non-force switch move before a pokemon is forced switched out, the pokemon switched in cannot use fake out since the last move getting evaluated is the player's move.

Scenario 1. Player used Focus Energy then Enemy Used Dragon Tail
	- Expectation: next player pokemon should be able to use Fake Out
	- Current behavior: Focus Energy is evaluated which does not have ForceSwitchAttr, so the move fails

_Please see logs for move used each turn_

https://github.com/pagefaultgames/pokerogue/assets/68144167/b7acd604-d66e-402e-a1ff-5ab07152e022

Scenario 2. Player used Drain Punch then Enemy used U Turn
	- Expectation: next enemy pokemon should be able to use Fake Out
	- Current behavior: Drain Punch is evaluated which does not have ForceSwitchAttr, so the move fails 
	- _I cant get a proper video here because the enemy keeps spamming U Turn x_x_

Fix introduced:
When SwitchSummonPhase is unshifted, we get instead the last move used by the switched out pokemon.


DustinLin:
Was testing this PR and got this interaction, is this intended? I'm assuming heracross is able to use Fake-out in this case

https://github.com/pagefaultgames/pokerogue/assets/39450497/f92cd90e-f177-4224-8632-d4e56b207a88

DustinLin:
Another sample: Seems like your scenario 2 is fixed (enemy Bagon's are able to use Fake-Out), but the heracross still isn't able to use fake-out when switched in. Ignore the first phase of battle (the fake-out fails are intentional)


https://github.com/pagefaultgames/pokerogue/assets/39450497/0253b757-df66-4cab-8baf-18164b240289

torranx:
Added a fix for this @DustinLin , what happened was if the pokemon is fainted thru force switch move, it decreases the turncount to 0, resulting the fake out to fail. Now in addition to checking if the last used move was a force switch move, it should also check whether the last pokemon was fainted or not.

DustinLin:
Tested doubled (seems ok so far), tried singles and found this issue. Seems like when you switch from a faint it doesn't allow you to use the move

https://github.com/pagefaultgames/pokerogue/assets/39450497/f8b1097c-c895-4893-bede-2d4cd00b6425

DustinLin:
Another potential bug: testing out double roars. I think in this case Bagon (which was cycled out through 2 roars) should be able to use Fake-Out? I could be wrong though


https://github.com/pagefaultgames/pokerogue/assets/39450497/37287e01-31b8-485d-8208-1dd40c7956a0

torranx:
> Another potential bug: testing out double roars. I think in this case Bagon (which was cycled out through 2 roars) should be able to use Fake-Out? I could be wrong though
> 
>  Screen.Recording.2024-05-29.at.4.11.04.PM.mov

Hmm I think so too, I'm guessing this is an edge case that is happening on main branch too.

torranx:
> Tested doubled (seems ok so far), tried singles and found this issue. Seems like when you switch from a faint it doesn't allow you to use the move
> 
>  Screen.Recording.2024-05-29.at.3.44.29.PM.mov

Was this tested after my [commit](https://github.com/pagefaultgames/pokerogue/pull/1251/commits/d2adebbe730e6e87b33351350ed23d0f63e6c83b)?

torranx:
> Another potential bug: testing out double roars. I think in this case Bagon (which was cycled out through 2 roars) should be able to use Fake-Out? I could be wrong though
> 
>  Screen.Recording.2024-05-29.at.4.11.04.PM.mov

cant seem to replicate this

DustinLin:
yeah... I've tested both the 2 cases today, as well as when the opponent is using roar and they seem to be fine not sure what happened yesterday sorry about that

src/phases.ts
```diff
@@ -1548,11 +1548,13 @@ export class SwitchSummonPhase extends SummonPhase {
...
     const pokemon = this.getPokemon();
-    const moveId = pokemon.scene.currentBattle.turnCommands[this.fieldIndex]?.move?.move;
+
+    const moveId = this.lastPokemon?.scene.currentBattle.lastMove;
     const lastUsedMove = moveId ? allMoves[moveId] : undefined;
 
     // Compensate for turn spent summoning
-    if (pokemon.scene.currentBattle.turnCommands[this.fieldIndex]?.command === Command.POKEMON || !!lastUsedMove?.findAttr(attr => attr instanceof ForceSwitchOutAttr)) { //check if hard switch OR pivot move was used
+    // Or compensate for force switch move if switched out pokemon is not fainted
+    if (pokemon.scene.currentBattle.turnCommands[this.fieldIndex]?.command === Command.POKEMON || (!!lastUsedMove?.findAttr(attr => attr instanceof ForceSwitchOutAttr) && !this.lastPokemon.isFainted())) {
```
Tempo-anon, line 1557: nit: can you make this `if` line a bit more readable? It's sort of hard to read when it's `if (conditionA || (conditionB) && !conditionC)`

Tempo-anon (approved):

Issue #1248, opened 2024-05-22
Implement shinylock for uncatchable pokemon


Tempo-anon:
Prevent shinies with `trySetShiny` from spawning on
* Enemy trainers (EXCEPT RIVAL_5 and RIVAL_6, but they have the shiny hardcoded)
* Classic paradox pokemon + final boss if the player does not own it
* End biome in Endless

PR #1252, opened 2024-05-22
Shinylock uncatchable Pokemon in endless mode

Tempo-anon:
* #1248 
* Added more documentation
* Some refactoring


src/field/pokemon.ts
```diff
@@ -1141,15 +1165,24 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   */
   generateVariant(): Variant {
     if (!this.shiny || !variantData.hasOwnProperty(this.species.speciesId))
       return 0;
     const rand = Utils.randSeedInt(10);
-    if (rand > 3)
+    if (rand > 3) {
       return 0;
-    if (rand)
+    } else if (rand > 1) {
```
Xavion3, line 1181: Don't change variant chances.

Xavion3 (changes requested):

src/field/pokemon.ts
```diff
@@ -1117,7 +1118,30 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * 
+   * The exact mechanic is that it calculates E as the XOR of the player's trainer ID and secret ID
+   * F is calculated as the XOR of the first 16 bits of the Pokemon's ID with the last 16 bits
+   * The XOR of E and F are then compared to the thresholdOverride to see whether or not to generate a shiny
+   * @param thresholdOverride number that is divided by 2^16 (65536) to get the shiny chance
+   * @returns true if the Pokemon has been set as a shiny, false otherwise
+   */
   trySetShiny(thresholdOverride?: integer): boolean {
+    // Opponent trainers cannot have shinies
+    if (!this.isPlayer() && this.hasTrainer()) {
```
CodeTappert, line 1133: Doesnt this block the rival from having the Shiny Mega Rayquaza? Which is on purpose?

TempsRay: Nah, the rival's shiny is manually set without using the `trySetShiny` function. See [here](https://github.com/pagefaultgames/pokerogue/blob/main/src/data/trainer-config.ts#L1007)

src/field/pokemon.ts
```diff
@@ -1117,7 +1118,30 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      return false;
+    }
+    // Shiny Pokemon should not spawn in the end biome in endless
+    if (!(this.scene.gameMode.modeId === GameModes.CLASSIC) && this.scene.arena.biomeType === Biome.END) {
+      return false;
+    }
+    // Unowned paradox pokemon are also shinylocked in classic
+    if (this.scene.gameMode.modeId === GameModes.CLASSIC && !this.scene.gameData.dexData[this.species.speciesId].caughtAttr) {
+      return false;
+    }
```
TempsRay, line 1143: Talked with Xavion offline and this is the one edge case that actually affects players. Currently, a player can find a shiny paradox in classic mode and then just put the save file on pause until they hatch the paradox Pokemon, at which point they will be allowed to catch it. This change makes it so that uncaught shiny paradoxes never even try to generate. Not sure if "put that save file on hold until you hatch it" is something that should be encouraged though

Tsukistar1:
> Personally, I think just stop shinies spawning in the end in endless as a first step. I like seeing them occasionally from trainers, but also I feel like the complaints mostly just originate from endless where the shiny rate is so much higher.

Definitely a good first step, though I can confirm that there have been people who have been frustrated by random shiny trainer pokemon as well. It's just that - like you said - shiny rates in Endless are higher, and also people grind out Endless more I think.

molker (approved):

PR #1253, opened 2024-05-22
Implement Incinerate

Fractaos:
Followup on #341 (after it has been closed due to inactivity, previous comments available there)

[Bulbapedia link](https://bulbapedia.bulbagarden.net/wiki/Incinerate_(move)).

A proposition of implementation of `Incinerate`'s effect.

- Adds a `RemoveHeldBerryAttr`

Maybe a more generic `Attr` could be used here, I was not totally sure, don't hesitate to make suggestions.
Also, maybe using the `removeModifier` is not the proper way to do this, so again feel free to tell me if there is a better method. (I'm saying this because of the `StealHeldItemChanceAttr` seems to use a more complex process to do so.)


src/data/move.ts
```diff
@@ -1,6 +1,6 @@
 import { Moves } from "./enums/moves";
 import { ChargeAnim, MoveChargeAnim, initMoveAnim, loadMoveAnimAssets } from "./battle-anims";
-import { BattleEndPhase, MoveEffectPhase, MovePhase, NewBattlePhase, PartyStatusCurePhase, PokemonHealPhase, StatChangePhase, SwitchSummonPhase, ToggleDoublePositionPhase } from "../phases";
+import { BattleEndPhase, MoveEffectPhase, MovePhase, NewBattlePhase, PartyStatusCurePhase, PokemonHealPhase, StatChangePhase, SwitchSummonPhase } from "../phases";
```
Fractaos, line 3: Unused import getting removed by my IDE, please let me know if it is supposed to stay for any reason I might not be aware of.

Fractaos:
A more complete implementation (including Knock Off) has been proposed in #1671.

PR #1261, opened 2024-05-22
Add setting to change money format

francktrouillez:
## What are the changes?

This PR adds a new setting to choose the money format. With this PR, the player can choose between:
- `Normal` (e.g. `1,234,567`)
  image
- `Abbreviated` (e.g. `1.23M`)
  image

I'm open to any feedback on the naming of the settings. I'm not sure if `Normal` is the best name for the first setting.

We could also add the scientific notation (e.g. `1.23e6`), but I'm not sure it adds much value since we already have the `Abbreviated` one. What do you think?

**The default setting is `Normal`.**

I think the change introduced in https://github.com/pagefaultgames/pokerogue/pull/1222 is a great addition to the game, but I believe that players should be able to choose the money format they prefer, specially for beginners that don't even reach the point where they have millions of money.

I also saw quite a lot of reactions going towards that direction on the Discord as well.

## What did change?

Before, the format was enforced to `Abbreviated` and players couldn't change it.

Now, players can choose between `Normal` and `Abbreviated` in the settings, and the format actually changes in the game.

### Videos

- Settings in the menu

  https://github.com/pagefaultgames/pokerogue/assets/57403591/0b509864-efd6-43ca-b623-4710dac468f5
- Money in the game

  https://github.com/pagefaultgames/pokerogue/assets/57403591/170ffb3a-5a30-49be-96b0-382f9cd8f837

## How to test the changes?
I currently tested the changes manually, as I currently don't know how easy it would be to write actual tests for this. I'm open to any feedback on how to improve the testing process.

I've been using the [`src/overrides.ts`](https://github.com/pagefaultgames/pokerogue/blob/e66d7026d5daa3ad3433ef16e2233cd4fc519eb3/src/overrides.ts) file to test the changes.

To do so, I just granted myself a lot of money to test the different formats.
https://github.com/pagefaultgames/pokerogue/blob/e66d7026d5daa3ad3433ef16e2233cd4fc519eb3/src/overrides.ts#L31
was changed to
```typescript
export const STARTING_MONEY_OVERRIDE: number = 987654321;
```

I'm open to any feedback or suggestions about this feature.


src/battle-scene.ts
```diff
@@ -1198,9 +1200,14 @@ export default class BattleScene extends SceneBase {
...
 	}
 
-	updateMoneyText(): void {
-		this.moneyText.setText(`₽${Utils.formatLargeNumber(this.money, 1000)}`);
-		this.moneyText.setVisible(true);
+	updateMoneyText(forceVisible: boolean = true): void {
+		if (this.money === undefined)
+			return;
+		const formattedMoney =
+			this.moneyFormat === MoneyFormat.ABBREVIATED ? Utils.formatLargeNumber(this.money, 1000) : this.money.toLocaleString('en-US');
```
CodeTappert, line 1207: While we are at it. Couldnt we make the "en-US" also language specific?

src/battle-scene.ts
```diff
@@ -1250,9 +1252,16 @@ export default class BattleScene extends SceneBase {
...
 
-  updateMoneyText(): void {
-    this.moneyText.setText(`₽${Utils.formatLargeNumber(this.money, 1000)}`);
-    this.moneyText.setVisible(true);
+  updateMoneyText(forceVisible: boolean = true): void {
+    if (this.money === undefined) {
+      return;
+    }
+    const formattedMoney =
+			this.moneyFormat === MoneyFormat.ABBREVIATED ? Utils.formatLargeNumber(this.money, 1000) : this.money.toLocaleString("en-US");
```
francktrouillez, line 1260: @CodeTappert 
> While we are at it. Couldnt we make the "en-US" also language specific?

Sure, I wanted to do that in the first place, but I saw that the locale was explicitly set here, while the [default behavior is to use the user locale](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString#using_tolocalestring). I don't know if this was set for a specific reason.

If this was just by convention, I believe it's ok to use the user locale.
```suggestion
			this.moneyFormat === MoneyFormat.ABBREVIATED ? Utils.formatLargeNumber(this.money, 1000) : this.money.toLocaleString();
```

francktrouillez:
@bennybroseph @Greenlamp2 @CodeTappert Any feedback on this?

TempsRay (approved):
This change is also accessibility for users with dyslexia

PR #1268, opened 2024-05-23
Bugfix: Beat Up checks the user's party instead of always checking player's party

td76099:
@Piefrenzy on Discord

Beat Up worked correctly when the player used it, but when the enemy used it it was looking at the player's party to determine the number of hits. This means that a wild Pokemon in single battles could hit the player up to 6 times when it should only be once.

Now, enemy Pokemon in single battles will hit once, enemy Pokemon in double battle will hit twice (or once if partner dead), and trainer battles hit the expected amount based on Pokemon remaining in either singles/doubles.

Video of it working:

https://github.com/pagefaultgames/pokerogue/assets/85713900/f8740a82-1c8a-406e-b729-460b48ac7e6c


TempsRay (dismissed):
Can you post a video of an enemy trainer using it and hitting more than twice? Code looks fine

td76099:
> Can you post a video of an enemy trainer using it and hitting more than twice? Code looks fine

Good catch - I just assumed it was because I was at 5x with no damage numbers on but I turned the speed down and flipped damage numbers on and sure enough there were some "hits" doing no damage. Found the function responsible so it should be good now :)

Video as requested:
 

https://github.com/pagefaultgames/pokerogue/assets/85713900/f8740a82-1c8a-406e-b729-460b48ac7e6c

bennybroseph (approved):

Issue #1269, opened 2024-05-23
TERA_BLAST is defined as a move constant but no animation exists


sir-vancelot:
[TERA_BLAST](https://github.com/pagefaultgames/pokerogue/blob/8cd166536a14d8cf0fdf603c4616dbc41efed8c7/src/data/enums/moves.ts#L1738) is defined as a move but no animation exists for it. This results in an error where the run freezes as the base html is returned in place of valid json.

![image](https://github.com/pagefaultgames/pokerogue/assets/22999101/26566f93-9f8a-428a-9564-dc9531b25d01)

PR #1270, opened 2024-05-23
Disable tera-blast to prevent its use while there is not an animation

sir-vancelot:
Fixes an issue where the tera-blast.json animation not existing can result in a run freezing

Closes https://github.com/pagefaultgames/pokerogue/issues/1269


Madmadness65:
It'd be better if you just replace `.partial()` with `.unimplemented()` (to prevent it from being used), because commenting out things like this has consequences on existing runs.

sir-vancelot:
> It'd be better if you just replace `.partial()` with `.unimplemented()` (to prevent it from being used), because commenting out things like this has consequences on existing runs.

Done

Madmadness65 (approved):
This is fine for now, until we fix its underlying problem

PR #1273, opened 2024-05-23
Luck based encounters

BatCrowP:
Increases chance to encounter rarer mons based on party luckValue

### DIFF
--- a/src/field/arena.ts
+++ b/src/field/arena.ts
@@ -64,13 +64,27 @@ export class Arena {
     }
   }
 
-  randomSpecies(waveIndex: integer, level: integer, attempt?: integer): PokemonSpecies {
+  randomSpecies(waveIndex: integer, level: integer, attempt?: integer, luckValue?: integer): PokemonSpecies {
     const overrideSpecies = this.scene.gameMode.getOverrideSpecies(waveIndex);
     if (overrideSpecies)
       return overrideSpecies;
     const isBoss = !!this.scene.getEncounterBossSegments(waveIndex, level) && !!this.pokemonPool[BiomePoolTier.BOSS].length
       && (this.biomeType !== Biome.END || this.scene.gameMode.isClassic || this.scene.gameMode.isWaveFinal(waveIndex));
-    const tierValue = Utils.randSeedInt(!isBoss ? 512 : 64);
+      let tierValue = Utils.randSeedInt(!isBoss ? 512 : 64);
+      if (typeof luckValue !== 'undefined') { //if party passed
+        if (luckValue < 6) {
+          //Do nothing at low luck values
+        } else if (luckValue < 10) {
+          tierValue = Math.floor(tierValue * 5.5 / luckValue); 
+        } else {
+          tierValue = Math.floor(tierValue * 5.5 / luckValue);
+          if (isBoss && tierValue >= 20) { //the modifiers to tierValue in this section are arbitrary
+            tierValue += 6; //Slight bump to common rarity bosses to make sure they stay more common than rares
+          } else if (!isBoss && tierValue >= 32) {
+            tierValue += 62; //Slight bump to common rarity enemies to make sure they appear more frequently than uncommons
+          }
+        }
+      }
     let tier = !isBoss
       ? tierValue >= 156 ? BiomePoolTier.COMMON : tierValue >= 32 ? BiomePoolTier.UNCOMMON : tierValue >= 6 ? BiomePoolTier.RARE : tierValue >= 1 ? BiomePoolTier.SUPER_RARE : BiomePoolTier.ULTRA_RARE
       : tierValue >= 20 ? BiomePoolTier.BOSS : tierValue >= 6 ? BiomePoolTier.BOSS_RARE : tierValue >= 1 ? BiomePoolTier.BOSS_SUPER_RARE : BiomePoolTier.BOSS_ULTRA_RARE;
--- a/src/field/arena.ts
+++ b/src/field/arena.ts
@@ -19,6 +19,7 @@ import { Terrain, TerrainType } from "../data/terrain";
 import { PostTerrainChangeAbAttr, PostWeatherChangeAbAttr, applyPostTerrainChangeAbAttrs, applyPostWeatherChangeAbAttrs } from "../data/ability";
 import Pokemon from "./pokemon";
 import * as Overrides from '../overrides';
+import { getPartyLuckValue } from "#app/modifier/modifier-type.js";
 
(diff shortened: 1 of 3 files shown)


BatCrowP:
Numbers not super important on this, main points that I aimed for with the rough plan was 
1) Minimum encounter rate does not change
2) Max UR encounter rate stays close to 2.5x (arbitrary based on hearing that sam planned on that)
image
3) Keep Common as most frequently appearing

Below is a quick spreadsheet of the spawning rates with my proposed numbers at 14 and 10 luck as well as the relative increase from the original encounter chance
image

BatCrowP:
Calculation is done by checking if luckValue is greater than 6, and if so multiplies the tier value by 5.5/luckValue.

src/field/arena.ts
```diff
@@ -64,13 +64,27 @@ export class Arena {
...
     if (overrideSpecies)
       return overrideSpecies;
     const isBoss = !!this.scene.getEncounterBossSegments(waveIndex, level) && !!this.pokemonPool[BiomePoolTier.BOSS].length
       && (this.biomeType !== Biome.END || this.scene.gameMode.isClassic || this.scene.gameMode.isWaveFinal(waveIndex));
-    const tierValue = Utils.randSeedInt(!isBoss ? 512 : 64);
+      let tierValue = Utils.randSeedInt(!isBoss ? 512 : 64);
+      if (typeof luckValue !== 'undefined') { //if party passed
+        if (luckValue < 6) {
+          //Do nothing at low luck values
+        } else if (luckValue < 10) {
```
molker, line 77: instead of an empty if result can you combine this with the first one?
`luckValue < 10 && luckValue >= 6` ?

src/modifier/modifier-type.ts
```diff
@@ -1592,15 +1592,16 @@ export function getEnemyBuffModifierForWave(tier: ModifierTier, enemyModifiers:
...
+    let r = 0;
+    let matchingModifier: Modifiers.PersistentModifier;
+    while (++r < retryCount && (matchingModifier = enemyModifiers.find(m => m.type.id === candidate.type.id)) && matchingModifier.getMaxStackCount(scene) < matchingModifier.stackCount + (r < 10 ? tierStackCount : 1))
+      candidate = getNewModifierTypeOption(null, ModifierPoolType.ENEMY_BUFF, tier);
+    
+    const modifier = candidate.type.newModifier() as Modifiers.EnemyPersistentModifier;
+    modifier.stackCount = tierStackCount;
+  
+    return modifier;
+
```
Xavion3, line 1604: What is actually being changed here?

Xavion3:
While the idea of changing rarity is fine, I personally disagree with the scaling here. We generally have tried to have luck produce diminishing returns, this makes low luck irrelevant and makes luck have increasing returns so like 14 luck is significantly more than twice as strong as 7 luck. I feel like we could make this whole thing much cleaner overall.

src/modifier/modifier-type.ts
TempsRay: Looks like tabbing

src/field/arena.ts
```diff
@@ -64,13 +64,27 @@ export class Arena {
...
+  randomSpecies(waveIndex: integer, level: integer, attempt?: integer, luckValue?: integer): PokemonSpecies {
     const overrideSpecies = this.scene.gameMode.getOverrideSpecies(waveIndex);
     if (overrideSpecies)
       return overrideSpecies;
     const isBoss = !!this.scene.getEncounterBossSegments(waveIndex, level) && !!this.pokemonPool[BiomePoolTier.BOSS].length
       && (this.biomeType !== Biome.END || this.scene.gameMode.isClassic || this.scene.gameMode.isWaveFinal(waveIndex));
-    const tierValue = Utils.randSeedInt(!isBoss ? 512 : 64);
+      let tierValue = Utils.randSeedInt(!isBoss ? 512 : 64);
+      if (typeof luckValue !== 'undefined') { //if party passed
+        if (luckValue < 6) {
```
TempsRay, line 75: Not sure how I feel about the balance of this. I think ideally either it should scale linearly or in such a way that there's diminishing returns for higher levels.

It also seems like it isn't too simple to explain, which I think should be an important factor. My initial thought of `generate the random number from 0 to (512 - luck * 2) for nonboss and 0 to (64 - luck/2) for bosses` for example is a lot simpler and understandable

BatCrowP: I definitely agree that it would be simpler to modify the ceiling but it doesn't give much incentive to actually have any luck at all. At max luck this only ends up increasing UR rarity from 1.56% to 1.75%, any increase is nice but it felt underwhelming looking at the numbers. Below is what the numbers would look like going from live branch to your suggestion. 

The simple equation would probably be for the best but it might be good to rethink the numbers. 
image

src/modifier/modifier-type.ts
Xavion3: Why though? I'm surprised this is even passing eslint since it's making the indentation inconsistent.

src/field/arena.ts
```diff
@@ -64,13 +64,22 @@ export class Arena {
...
+  randomSpecies(waveIndex: integer, level: integer, attempt?: integer, luckValue?: integer): PokemonSpecies {
     const overrideSpecies = this.scene.gameMode.getOverrideSpecies(waveIndex);
     if (overrideSpecies)
       return overrideSpecies;
     const isBoss = !!this.scene.getEncounterBossSegments(waveIndex, level) && !!this.pokemonPool[BiomePoolTier.BOSS].length
       && (this.biomeType !== Biome.END || this.scene.gameMode.isClassic || this.scene.gameMode.isWaveFinal(waveIndex));
-    const tierValue = Utils.randSeedInt(!isBoss ? 512 : 64);
+      let tierValue = Utils.randSeedInt(!isBoss ? 512 : 64);
+      if (typeof luckValue !== 'undefined') { //if luckValue passed in, use it to modify rare encounter odds
+          if (isBoss) {
```
brain-frog, line 75: since the parameter for randSeedInt is also based on `isBoss` can you please make the if statement structure be: 
```
if (isBoss) 
   have a constant for the parameter
   if (luckValue is defined) 
      have a constant for the multiplier to subtract from the parameter defined above

after all if statements 
const tierValue = Utils.randSeedInt(first const - second const)
```

brain-frog (changes requested):
Once comment is addressed and merge conflicts are solved this looks good to me

brain-frog:
closing due to inactivity (but I did implement using your approach)

PR #1275, opened 2024-05-23
Implement Quick Guard and other conditional team protection moves

innerthunder:
With the following changes, this fully implements the moves Quick Guard, Wide Guard, Mat Block, and Crafty Shield:

- A new arena tag class `ConditionalProtectTag` applies a custom condition function to cancel the effects of qualifying moves. The subclasses `QuickGuardTag`, `WideGuardTag`, `MatBlockTag`, and `CraftyShieldTag` supply the condition functions for their respective moves.
- The move Feint is updated with a new attribute `RemoveArenaTagsAttr` that removes arena tags with the specified tag type(s) from a given side of the field. Feint uses this to remove all of the tags listed above from the target's side of the field.
- Pokemon.apply() is updated to apply the tags listed above to cancel incoming moves if conditions are met

### Test Videos (not exhaustive, more to come)

Quick Guard: 

- Charmander uses Quick Guard, then ally Bulbasaur also tries to use Quick Guard and it fails.
- Pawmi uses Quick Attack targeting Bulbasaur, and the attack is blocked by Quick Guard.

https://github.com/pagefaultgames/pokerogue/assets/168692175/41f31a0e-941f-4fa9-b0dc-a6fc899d0ce2

Wide Guard:

- Charmander uses Wide Guard, then ally Bulbasaur also tries to use Wide Guard and it fails.
- Pawmi uses Dazzling Gleam, and both player Pokemon are protected by Wide Guard.

https://github.com/pagefaultgames/pokerogue/assets/168692175/77b2eee8-3ce5-45f0-ad93-0b153839b3f6

### Known Issues

- Conditional protection is not ignored by moves that ignore the effects of Protect (e.g. Quick Guard blocks Feint when it shouldn't). This should be an easy fix in `Pokemon.apply()` **[EDIT: FIXED]**
- Feint does not disable conditional protection on its target if it resolves before the conditional protection move is used (e.g. Mat Block can be applied after Feint is used)


innerthunder:
## Some more test videos:

Feint and Quick Attack vs. Quick Guard:

- The opposing Poochyena uses Quick Guard to protect its team from priority moves.
- Bulbasaur uses Feint on Poochyena, ignoring Quick Guard and dealing damage.
- Charmander uses Quick Attack on Poochyena, dealing damage and showing that Quick Guard's effects have been removed from the field

https://github.com/pagefaultgames/pokerogue/assets/168692175/498a2662-3f67-4d00-bf6a-3f4ba8cf8ea7

Wide Guard vs. Status Move:

- Charmander and Bulbasaur use Wide Guard as in the previous test for Wide Guard.
- Pawmi uses Growl, and Wide Guard protects both Charmander and Bulbasaur from its effects.

https://github.com/pagefaultgames/pokerogue/assets/168692175/3142672b-13e4-4def-a3d1-fb5c3d290627

Mat Block:

- Charmander uses Mat Block, then Bulbasaur uses Crafty Shield.
- Pawmi uses Earthquake, and both Charmander and Bulbasaur are protected by Mat Block.
- The move Mat Block uses the same condition as Fake Out in its code. If any Pokemon were to use Mat Block on the following turn, it would fail.

https://github.com/pagefaultgames/pokerogue/assets/168692175/bbdfb54e-a286-45c1-ba37-0dd78da74d37

Mat Block overlapping with Wide Guard:

- Charmander uses Mat Block, and Bulbasaur uses Wide Guard. Both moves go through without failing.
- Pawmi uses Earthquake, and Wide Guard is shown to protect both Charmander and Bulbasaur.
- Because Earthquake was already cancelled by Wide Guard, the effects and animations of Mat Block are not shown.

https://github.com/pagefaultgames/pokerogue/assets/168692175/ceb1e08c-3844-4862-a340-78d4527b40c6

Crafty Shield:

- Charmander uses Crafty Shield, then Bulbasaur also uses Crafty Shield and it fails.
- Pawmi uses Growl, and Crafty Shield protects both Charmander and Bulbasaur from its effects.

https://github.com/pagefaultgames/pokerogue/assets/168692175/9e999db2-638d-4ea8-b4a3-8a46c268617c

bennybroseph (approved):

PR #1290, opened 2024-05-23
Added nature selection and base-state toggle to starter menu

prime-dialga:
Added persistent nature selection and base-state toggle to starter menu.
[Note: animations and colors for the base-state-chart as well as the translations might require additional work/checks]


bennybroseph:
This needs to be split into multiple PRs. There's too much being done at once for this to be reviewed on top of there being no comments and no screenshot/video showing what this actually does.

PR #1291, opened 2024-05-23
Implement Purify

ReneGV:
From [bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Purify_(move)).

- Move fails if user does not have a non-volatile status effect.
- If user has an non-volatile status, it heals the status and recovers 50% of HP


td76099:
Per [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Status_condition#Non-volatile_status), non-volatile also includes Sleep and Freeze. I understand that this move can't be used while frozen, but could theoretically be called with "Sleep Talk" while asleep.

Also, I would like to see a new function made in `status-effects.ts` that returns a list of non-volatile conditions which you call instead of naming the conditions individually. Would be cleaner and tbh something that is probably needed as I am sure there are other moves/abilities out there that mess with non-volatile statuses that could benefit from it.

Other than that though I tested it out real quick and it worked as expected 👍

ReneGV:
@td76099, thanks for your feedback
Implemented suggested changes and I also add some videos showing the behave for this move, including sleep talk.

Non status effect

https://github.com/pagefaultgames/pokerogue/assets/31145813/79ac1a22-8a0b-4383-bdc5-cbf7ab430b6a

Sleep talk + purify

https://github.com/pagefaultgames/pokerogue/assets/31145813/bc5a836b-c5c7-4a58-9865-8883d2ea874a

Status + HP heal

https://github.com/pagefaultgames/pokerogue/assets/31145813/1a6cdef4-5c74-4efd-87ff-73f47d27fa35

bennybroseph (dismissed):

bennybroseph (dismissed):

bennybroseph (approved):

PR #1292, opened 2024-05-23
Make zh_CN translation easier to understand

zjl88858:
In the previous translation, some words were used in a way that was close to the original, but did not common usage in Chinese (at least in mainland China and Singapore).
For example, we will not use "最大" when comparing two values, we common use "较大", and only use "最大" when there are more than two values.
So I changed it.
I didn't consider line breaks, but since the length of the modified statement is basically same, it shouldn't cause any line break errors. If someone could help me check, it would be great.


TempsRay (approved):
Thought you were talking about the IV scanning terms, in which case I'd suggest sticking with whatever in-game uses but these localizations are elsewhere. 

LGTM. 输入 sounds better than 提供 for example

Madmadness65 (approved):

PR #1293, opened 2024-05-23
Convert reconnect method to exponential backoff timeout

Icehawk78:
This is an updated implementation of the server unavailable check to lighten the load while the servers are either down or flaky. There are two primary changes contained:

* Converting the reconnect call from `setInterval` to `setTimeout`, to prevent overlapping requests from being sent in while the server is slow to respond
* Changing the retry to an exponential backoff, starting at 5 seconds, and doubling until it gets to 15 minutes, each of which are configurable variables in the code.


bennybroseph:
Sorry for the duplicate work on this, but I had already started fixing this up on an older PR by the time this one was created. If you have improvements to make to mine feel free to edit this PR to do so. Otherwise I'll close this one out in a day or so.

CodeTappert:
Retry works as expected. It doubles the time every time. (Honestly i dont want to wit until it reaches 15min but i verified that the code is correct.

If others want to test it. Just change the info endpoint function to

```go
// /account/info - get account info
func Info(username string, uuid []byte) (InfoResponse, error) {
	return InfoResponse{}, fmt.Errorf("fail Test") // We automatically fail the request here
	response := InfoResponse{Username: username, LastSessionSlot: -1}

	highest := -1
	for i := 0; i < defs.SessionSlotCount; i++ {
		data, err := db.ReadSessionSaveData(uuid, i)
		if err != nil {
			continue
		}

		if data.Timestamp > highest {
			highest = data.Timestamp
			response.LastSessionSlot = i
		}
	}

	if response.LastSessionSlot < 0 || response.LastSessionSlot >= defs.SessionSlotCount {
		response.LastSessionSlot = -1
	}

	return response, nil
}

```


Also i added some logging for my test. 

Here the output of my browser console that it works

image

bennybroseph (approved):

PR #628, opened 2024-05-08
Exponential backoff in unavailable-modal-ui-handler

exitmouse:
Currently when the servers are down each player attempts to reconnect every 5 seconds. I'm not sure how much of the DDoS-like traffic the game is getting is due to this retry, but it seems like it can't hurt to add some kind of exponential backoff.
I'm not super comfortable as a typescript dev and I don't know how to get this dialogue to appear when running locally, so use caution.

PR #1295, opened 2024-05-23
Increases the Amount of Time Until a Reconnect up to a Maximum

bennybroseph:
Exponentially increases the amount of time a client waits before attempting a reconnect up to a set maximum.

### DIFF
--- a/src/ui/unavailable-modal-ui-handler.ts
+++ b/src/ui/unavailable-modal-ui-handler.ts
@@ -9,10 +9,10 @@ export default class UnavailableModalUiHandler extends ModalUiHandler {
   private reconnectInterval: number;
   private reconnectCallback: () => void;
 
-  private readonly minTime = 5000;
-  private readonly maxTime = 60000;
+  private readonly minTime = 1000 * 5;
+  private readonly maxTime = 1000 * 60 * 5;
 
-  private readonly randVarianceTime = 10000;
+  private readonly randVarianceTime = 1000 * 10;
 
   constructor(scene: BattleScene, mode?: Mode) {
     super(scene, mode);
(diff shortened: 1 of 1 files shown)


XGSleepWalker (approved):

src/ui/unavailable-modal-ui-handler.ts
```diff
@@ -49,16 +76,7 @@ export default class UnavailableModalUiHandler extends ModalUiHandler {
...
-        updateUserInfo().then(response => {
-          if (response[0] || [200, 400].includes(response[1])) {
-            clearInterval(this.reconnectTimer);
-            this.reconnectTimer = null;
-            this.scene.playSound("pb_bounce_1");
-            this.reconnectCallback();
-          }
-        });
-      }, 5000);
+      this.reconnectTimer = setInterval(() => this.tryReconnect(), this.reconnectInterval);
```
Icehawk78, line 79: Having this be on an interval instead of a timeout will make the delays between slow responding calls less delayed, because `setInterval` will trigger exactly `reconnectInterval` seconds after the previous request, rather than waiting for any response.

bennybroseph: it would if i didn't have `clearInterval(this.reconnectTimer);`

bennybroseph: something to clean up later, but this works

Dakurei:
So I'm putting back the message I'd put on the PR #628 but with the tests of these commits:

I've tested this with a local server, forcing the /info request to return a 500 error, and the delay is well included in the exponential backoff set up, so it looks all right to me.

This should save the server from getting a ton of useless requests during outages, making the situation worse in the event of a provider's ddos block.

> Server view requests
image
image

> Client view requests
image

Greenlamp2 (approved):

molker (approved):

Madmadness65 (approved):

PR #1299, opened 2024-05-23
TM Shop Step 1: Add flags in save data to track TMs unlocked by player

Fontbane:
Just a draft for now. This PR will make the game start to track which TMs players have "unlocked". This is represented by an array of 10 32 bit integers in the save data. Each bit represents a TM that's been unlocked. I also made functions to convert from TM number to Move number and vice versa. The way I implemented it, a TM will be permanently "unlocked" for the account when the player teaches their mon a move using that TM.

Obviously this doesn't implement a TM shop, but I believe this is a good first step, as I think at the very least keeping track of this will be relevant for our implementation.

Personally, I think the TM shop should have a pool of 5 random TMs available as well as any TM the player has unlocked. TMs probably shouldn't be unlocked by buying in the shop. I also think some TMs could be unlocked in other ways, for example by defeating Gym Leaders, or unlocking the "ultimate" moves like Hydro Cannon or Hyper Beam by beating Classic Mode the first time. This can be discussed on the discord.

### DIFF
--- a/src/modifier/modifier.ts
+++ b/src/modifier/modifier.ts
@@ -1204,7 +1204,7 @@ export class TmModifier extends ConsumablePokemonModifier {
   apply(args: any[]): boolean {
     const pokemon = args[0] as PlayerPokemon;
 
-    pokemon.scene.unshiftPhase(new LearnMovePhase(pokemon.scene, pokemon.scene.getParty().indexOf(pokemon), (this.type as ModifierTypes.TmModifierType).moveId));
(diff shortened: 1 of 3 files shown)


bennybroseph:
Very interested in a feature like this, but it seems quite large in scope. Is there any sort of design doc for it? I'm not sure if this change is needed for a TM shop depending on implementation. Don't want this work to go to waste because we decide we want it to work differently.

PR #1306, opened 2024-05-23
Devcontainer added

HelixOW:
Added a Devcontainer configuration to run dev environment on same machine for everyone opting in to use it.

(No more "Works on my machine")


flx-sta:
@HelixOW can you please elaborate the use of this PR and what it's for?

HelixOW:
When you use docker, you can use a dev container to have the development environment for all users. So no more different node versions, etc.

flx-sta:
Thank you for your contribution.


Unfortunately, at this moment we're not focused on enabling docker for the client project so it's something we'll only take a look at in the future

PR #1309, opened 2024-05-24
[Ability] Implement Protean and Libero abilities

OrangeRed:
### What is does

Once per switch in a pokemon will change its type if possible to the first move it selects.

> A Pokémon with Libero/Protean as its Ability will change its type to be the same type as the move it is using. Libero always causes the user to become a single-type Pokémon.
> 
> https://bulbapedia.bulbagarden.net/wiki/Libero_(Ability)

Full text for implementation details

> Because Libero activates prior to executing a move, the user will receive same-type attack bonus on every move it uses (except Struggle) regardless of the move's type and will always hit with the move Toxic.
> 
> When using a move that calls another move (such as Nature Power), the type of the called move will be considered. When using a move that can change type or if the move has been affected by Electrify, the type it changes to will be considered. When using a move that changes the user's type, such as Reflect Type, first Libero changes the user's type to the type of the move, then the move will either fail or change the user's type again. When using Curse (unless its type is changed by Electrify), if the user is not already Ghost-type prior to the activation of Libero, Libero will change the user to Ghost-type, then the user will use the Ghost-type effect of Curse on itself (regardless of Crafty Shield and Follow Me or Rage Powder.
> 
> Libero will not activate if the move fails (unless it fails due to the activation of Libero), if the execution is interrupted entirely (such as by the effect of infatuation or confusion, if the user is already the same single type as the move being used. Libero will activate even if the move misses, is blocked by a protection move, or fails to affect the opponent. Libero will still change the user's type when using Roar, Whirlwind, Trick-or-Treat, and Forest's Curse, regardless of whether the move successfully executes or not.
> 
> The user's new type remains in effect until changed again or until the user is switched out; negating or replacing the Ability does not change the user back to their original type.
>
> If the user Terastallizes, the effects of Libero will be nullified.

<details>

<summary> Implementation Checklist </summary>  

### Base

- [x] Does the pokemon's type change based on selected move?
- [x] Does the ability skip changing the pokemon's type when it is already the same type of the move?
- [x] Does the ability skip changing the pokemon's type when it is Terastallized?
- [x] Does the pokemon's type change correctly for moves that call other moves?  
- [x] Does the pokemon's type change correctly for moves that have multiple / variable types?
- [x] Does the pokemon's type change correctly for multi-turn moves? (Fly, Dig, Dive, etc..)
- [x] Does the pokemon's type change before a move that changes the pokemon's type triggers?
- [x] Does the ability cause the pokemon to Curse itself?
- [x] Does the ability skip changing the pokemon's type correctly even if the move fails?
- [x] Does the pokemon's type change even if the move misses, is protected against or if the opponent is immune?
- [x] Does the pokemon's type change when using Roar, Whirlwind, Trick-or-Treat, and Forest's Curse even if the move fails?

### Specific to Pokerogue

- [x] Does the pokemon's type change when combined with a passive that change's the types of moves?
- [x] Do the pokemon's types stay changed after reload / refresh.

_Let me know if there is anything else that needs to be tested_

</details>


OrangeRed:
I will add recordings of the skill working as I get them since this ticket will probably require a lot of them.

Blitz425:
image (2)
Do you think it's possible to change it to its old functionality? Before Sam left he intended it to be the original Gen 8 versions.

src/phases.ts
```diff
@@ -2572,6 +2572,10 @@ export class MovePhase extends BattlePhase {
   }
 
   getEffectPhase(): MoveEffectPhase {
+    // Trigger pokemon type change before playing the move animation
+    if (this.pokemon.hasAbilityWithAttr(PokemonTypeChangeAbAttr)) {
```
Xavion3, line 2576: Why are you checking if it has the ability before applying?

src/utils.ts
```diff
@@ -298,22 +298,22 @@ export class BooleanHolder {
...
+export class IntegerHolder<T = integer> extends NumberHolder<T> {
+  constructor(value: T) {
     super(value);
   }
 }
 
-export class FixedInt extends IntegerHolder {
-  constructor(value: integer) {
+export class FixedInt<T = integer> extends IntegerHolder<T> {
+  constructor(value: T) {
```
Xavion3, line 316: Why are these being changed if they have nothing to do with anything else in this PR?

src/data/ability.ts
```diff
@@ -982,29 +984,67 @@ export class FieldPreventExplosiveMovesAbAttr extends AbAttr {
...
+
+      if (args[1] && args[1] instanceof Utils.NumberHolder) {
+        args[1].value *= this.powerMultiplier;
+      }
       return true;
     }
 
     return false;
   }
 }
```
Xavion3, line 1008: Why is this being changed when it's not part of protean/libero?

src/data/ability.ts
```diff
@@ -982,29 +984,67 @@ export class FieldPreventExplosiveMovesAbAttr extends AbAttr {
...
+  constructor() {
+    super(true);
+  }
+
+  applyPreAttack(pokemon: Pokemon, passive: boolean, defender: Pokemon, move: PokemonMove, args: any[]): boolean {
+    if (
+      // If the user Terastallizes, the effect will be nullified.
+      !pokemon.isTerastallized() &&
+      // When using a move that calls another move (such as Nature Power), the type of the called move will be considered.
+      !move.getMove().attrs.find((m) => m instanceof OverrideMoveEffectAttr)
```
Xavion3, line 1021: Why are you directly accessing the attrs object instead of using one of the helper functions? Also two turn moves also fall under this attribute, will they be accounted for seperately?

src/data/ability.ts
```diff
@@ -966,7 +966,9 @@ export class MoveTypeChangePowerMultiplierAbAttr extends VariableMoveTypeAbAttr
     const type = (args[0] as Utils.IntegerHolder);
     if (type.value === this.matchType) {
       type.value = this.newType;
-      (args[1] as Utils.NumberHolder).value *= this.powerMultiplier;
+      if (args[1] && args[1] instanceof Utils.NumberHolder) {
+        args[1].value *= this.powerMultiplier;
+      }
```
Xavion3, line 971: Why is this being changed?

OrangeRed: I invoke these three Attr classes to get the move's new type. Prior to this change if I didn't pass in an IntHolder into the those Attr as `args[1]` the game would throw an error so this is my way of preventing that.

src/data/ability.ts
OrangeRed: This is being changed for the same reason as I mentioned in a different comment where I invoke this class in order to figure out what type the pokemon needs to change to but in order to do so without causing a run time error I need to pass a NumberHolder as `args[1]` so this is my way of preventing that error.

Also the change to the condition was warranted (at least in my mind) because the previous condition function had unused variables for all of the abilities I had seen with that Attr

src/utils.ts
OrangeRed: I made these changes because I had seen a lot of manual type casts like this
```ts
// typeof move.type -> Type
const type = Utils.NumberHolder(move.type)
...
// typeof type.value -> number 
// A type conversion is necessary to return it back to type Type
const moveType = type.value as Type
// typeof moveType -> Type
```

By changing the underlying Holders to use generics typescript is smart enough to infer the type from the passed in value so the above code block becomes this.
```ts
// typeof move.type -> Type
const type = Utils.NumberHolder(move.type)
...
// typeof type.value -> Type 
const moveType = type.value 
// typeof moveType -> Type
```

src/phases.ts
OrangeRed: isn't that normal? You check if the pokemon has an ability with the attribute before you apply said ability?

src/data/ability.ts
OrangeRed: You're right that I could use a helper but there `Attr.findAttr()` basically the does the same thing under the hood. When I was writing this I felt like this was easier to read but I can change it to use the helper instead.

src/phases.ts
Xavion3: No, it just doesn't apply abilities which pokemon don't have.

OrangeRed: If that's the case I can remove it.

src/data/ability.ts
Xavion3: I feel like avoiding the usage of the helper and getter functions purely for aesthetics is a bad idea myself.

src/utils.ts
OrangeRed: The important bit with this change is the `<T = number>` and `<T = integer>`

These make it so that these Holders work the same way as they did before by defining a default type thats the same as the previous implementation.

src/data/ability.ts
OrangeRed: > I feel like avoiding the usage of the helper and getter functions purely for aesthetics is a bad idea myself.
That's valid. 

Also the reason I am skipping OverrideMoves is because I need to wait for the result of moves like Nature Power which under the hood use Nature Power and then spawn a new move that triggers. 

I haven't tested charge moves but if they work the same as Nature Power in that they spawn an attacking move afterwords then the type change will happen prior to that. If that isn't how they work then I'll need to go back to the drawing board and figure out how to accommodate them

src/data/ability.ts
Xavion3: I mean, the previous condition had unused arguments because it was following the standard that abilities which modify attacks were consistently following. Yes that meant some abilities had more data than they needed, but I'm not sure I'd call that a bad thing either.

I'm also suspicious this might not be able to conceal other things, because this is really just letting a function fail silently when given invalid arguments. Though this will probably still log an activation of the function. I guess my simulated call stuff should be prioritised a bit higher so things like this can work properly.

OrangeRed: > This is really just letting a function fail silently when given invalid arguments.

The function doesn't fail if the arguments are invalid it just doesn't mutate whatever is in `args[1]` if there is nothing there. The NumberHolder that I'm omitting is normally used as a damage multiplier and since I don't need that to determine the resulting move type I would simply be passing useless information into this class. 

> I mean, the previous condition had unused arguments because it was following the standard that abilities which modify attacks were consistently following. 

I could change it so that the condition is the same as it was before without causing any breaking changes.

src/utils.ts
bennybroseph: should make that a separate PR

OrangeRed: @bennybroseph https://github.com/pagefaultgames/pokerogue/pull/1325

OrangeRed:
## Videos of performed tests

- [x] Does the pokemon's type change based on selected move?

https://github.com/pagefaultgames/pokerogue/assets/5421700/e5be6bd3-4b48-4c03-ad2a-04b274b24bb4

- [x] Does the ability skip changing the pokemon's type when it is already the same type of the move?

https://github.com/pagefaultgames/pokerogue/assets/5421700/eb35e4a0-0645-4384-8141-fcc3ac6cc59b

- [x] Does the ability skip changing the pokemon's type when it is Terastallized?

--

- [x] Does the pokemon's type change correctly for moves that call other moves?
- [x] Does the pokemon's type change correctly for moves that have multiple / variable types?

https://github.com/pagefaultgames/pokerogue/assets/5421700/2e9a43e4-1a23-469d-841f-6d1ad61d34e0

- [x] Does the pokemon's type change correctly for multi-turn moves? (Fly, Dig, Dive, etc..) <br> Charge / Delayed moves and moves that use other moves inherit the same class so they would not trigger protean before 

https://github.com/pagefaultgames/pokerogue/assets/5421700/90f6d16f-99e5-4ddf-b784-81b74e3243d2

- [x] Does the pokemon's type change before a move that changes the pokemon's type triggers?

https://github.com/pagefaultgames/pokerogue/assets/5421700/a667b4bb-b894-4c66-a6a4-1e504c5cdf3c

- [x] Does the ability cause the pokemon to Curse itself?

https://github.com/pagefaultgames/pokerogue/assets/5421700/564b5a56-dfb5-4549-8ae3-0cc6c9df1fbf

- [x] Does the ability skip changing the pokemon's type correctly even if the move fails?

https://github.com/pagefaultgames/pokerogue/assets/5421700/bc6355c0-b727-4ec3-a400-eea8f75c1edf

- [x] Does the pokemon's type change even if the move misses, is protected against or if the opponent is immune?

https://github.com/pagefaultgames/pokerogue/assets/5421700/fafa5c2f-2a66-4af3-89e1-cf0e399253e8

- [x] Does the pokemon's type change when using Roar, Whirlwind, Trick-or-Treat, and Forest's Curse even if the move fails?

https://github.com/pagefaultgames/pokerogue/assets/5421700/8be10aa6-3da8-490d-a244-85f8fe50ea27

## Pokerogue stuff

- [x] Does the pokemon's type change when combined with a passive that change's the types of moves?

https://github.com/pagefaultgames/pokerogue/assets/5421700/4a44b45d-ced6-4266-b8e7-d2bb6e2f5cb4

- [x] Do the pokemon's types stay changed after reload / refresh.

https://github.com/pagefaultgames/pokerogue/assets/5421700/456b6889-da5d-4c1d-8292-5351b28fea5f

src/data/ability.ts
```diff
@@ -1079,28 +1079,78 @@ export class FieldMultiplyBattleStatAbAttr extends AbAttr {
...
+      // TODO remove this copy when phase order is changed so that damage, type, category, etc.
+      // TODO are all calculated prior to playing the move animation.
+      const moveCopy = new Move(move.id, move.type, move.category, move.moveTarget, move.power, move.accuracy, move.pp, move.chance, move.priority, move.generation);
+
+      // Moves like Weather Ball ignore effects of abilities like Normalize and Refrigerate
+      if (move.findAttr(attr => attr instanceof VariableMoveTypeAttr)) {
+        applyMoveAttrs(VariableMoveTypeAttr, pokemon, null, moveCopy);
+      } else {
+        applyPreAttackAbAttrs(MoveTypeChangeAttr, pokemon, null, moveCopy);
+      }
```
OrangeRed, line 1135: There was a recent change that allowed us to directly mutate move data like damage and type and because of that change I can't simulate move type changes on the provided move without altering its data and preventing future ability triggers from working correctly.

src/data/ability.ts
```diff
@@ -1085,28 +1085,79 @@ export class FieldMultiplyBattleStatAbAttr extends AbAttr {
...
+      // TODO are all calculated prior to playing the move animation.
+      const moveCopy = new Move(move.id, move.type, move.category, move.moveTarget, move.power, move.accuracy, move.pp, move.chance, move.priority, move.generation);
+      moveCopy.attrs = move.attrs;
+
+      // Moves like Weather Ball ignore effects of abilities like Normalize and Refrigerate
+      if (move.findAttr(attr => attr instanceof VariableMoveTypeAttr)) {
+        applyMoveAttrs(VariableMoveTypeAttr, pokemon, null, moveCopy);
+      } else {
+        applyPreAttackAbAttrs(MoveTypeChangeAttr, pokemon, null, moveCopy);
+      }
```
OrangeRed, line 1142: There was a recent change that made it so that move data is directly passed into ability attributes which allows us to directly mutate data like damage and type. Because of that change I can't 'simulate' move type changes on the provided move without altering its data which prevents future ability triggers from working correctly so I have to make a copy of the move in order to maintain the original move's data integrity.

DustinLin:
ran tests on local machine and both passed

Tempo-anon (approved):

PR #1311, opened 2024-05-24
Add Memory Mushroom to the Shop on Floors 141+

YozoraNoHoshi:
## Summary
Adds Memory Mushroom to the shop beginning at Floor 141+ to Classic and Endless modes. 
The starting price is 1.5x base price, equal to a Max Potion.

## Test Case
Using local dev server:
Begin run at 139 -> check shop
Continue to 141 -> check shop (140 has no shop)

https://github.com/pagefaultgames/pokerogue/assets/43153371/1e6c3a33-aca6-4617-bf13-0aa96891a31c

### DIFF
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -1671,7 +1671,8 @@ export function getPlayerShopModifierTypeOptionsForWave(waveIndex: integer, base
       new ModifierTypeOption(modifierTypes.MAX_ELIXIR(), 0, baseCost * 2.5)
     ],
     [
-      new ModifierTypeOption(modifierTypes.FULL_RESTORE(), 0, baseCost * 2.25)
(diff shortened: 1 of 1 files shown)


Madmadness65 (changes requested):
Simply adding the item to the shop isn't enough. For starters, the Max Ether should probably be moved up to the first row, that way both rows have an equal amount of items in them when the Sacred Ash is unlocked.
shop
Second, because of how learning moves works, the used Memory Mushroom will only begin learning the move after you've selected one of the items, which isn't ideal, because a player could potentially buy multiple Memory Mushrooms, thinking that it doesn't work, among other issues that may crop up because of the delayed learning of moves.

https://github.com/pagefaultgames/pokerogue/assets/59298170/88dead4c-16c8-4742-8a2a-fb4145e6a414

PR #1318, opened 2024-05-24
Adding the option to have named trainers be able to have a double battle together

CodeTappert:
## What is the purpose of this?
- Honestly it started with the wish to have Tate and Liza to be able to fight together instead of Solo.
- Now every named trainer (gym leader, elite 4, champ) could partner up with any other
- For example red and blue
image

## What is already done?
- Ability to pair up any two  NAMED trainer types (gym leader. elite 4, champs) 
- Ability to set a double title (see image above) (can be localized)
- Ability to set Double name (can be localized)
- - Ability to mix their species togehter
- - Special Double Dialogue for the teams
## What does still need to be added?
- nothing (unless the dialogue localization PR is merged before this one)

Help would be appriciated!

## Testing:
- Special Dialogue for the trainers  (also special double title and double name)
image
- They have half half of the trainer pools
image
- The trainers still work solo 
image
- The double only has a 33% chance of occuring (in classic mode when one of the two trainers appear)
- Doesnt affect the "normal" duos at all
image
- The following pairs were added 
image

If you have question or want me to fix something to be able to merge this feel free to tell me!


CodeTappert:
Added "Alder & Iris", "Blue & Red", "Steven & Wallice", "Tate & Liza" as teams for now.

Next step will be doing/adding dialogue

CodeTappert:
This is ready for review/merge by a dev!

src/locales/en/trainers.ts
```diff
@@ -241,4 +243,14 @@ export const trainerNames: SimpleTranslationEntries = {
...
+
+  // Double Names
+  "blue_red_double": "Blue & Red",
+  "red_blue_double": "Red & Blue",
+  "tate_liza_double": "Tate & Liza",
+  "liza_tate_double": "Liza & Tate",
+  "steven_wallace_double": "Steven & Wallace",
+  "wallace_steven_double": "Wallace & Steven",
+  "alder_iris_double": "Alder & Iris",
+  "iris_alder_double": "Iris & Alder",
```
bennybroseph, line 255: is there *any* way we can make these one entry per double battle? It seems very strange to need to provide the exact same translation twice but with reversed names.

bennybroseph (changes requested):
This is very cool, but I have to ask for this change.

src/locales/en/trainers.ts
CodeTappert: > is there _any_ way we can make these one entry per double battle? It seems very strange to need to provide the exact same translation twice but with reversed names.

Sure. This is done ON purpose. 

If we just set the same name in the
`setHasDouble("tate_liza_double")` for both of them they would use the same locailzation string.

But i had done this with a reason. So first of all the Nametag is orderd like the charakter sprites.

And also that each unique parining (tate & liza vs liza & tate) could have their unique dialogue that starts with the person that is on the left (from player perspective)

I would really like to leave it that way since it is intentional

bennybroseph (approved):

PR #1322, opened 2024-05-24
Consider technician and priority on trainer moves when weighing movesets

Fontbane:
An older branch I realized I never made a PR for. Had the idea when fighting a Scizor with Mega Gardevoir and being confused why he didn't just use Bullet Punch. Multiplies power of technician boosted moves by 1.5 if the mon has Technician for the purposes of power based weighing. Also multiplies priority move power for weighing by 1.5, so priority moves in general are considered more.

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1266,7 +1266,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     // Weight towards higher power moves, by reducing the power of moves below the highest power.
     // Caps max power at 90 to avoid something like hyper beam ruining the stats.
     // This is a pretty soft weighting factor, although it is scaled with the weight multiplier.
-    const maxPower = Math.min(movePool.reduce((v, m) => Math.max(allMoves[m[0]].power, v), 40), 90);
+    // Account for technician boost and priority
+    const hasTechnician = this.hasAbility(Abilities.TECHNICIAN);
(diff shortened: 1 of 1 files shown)


Xavion3:
This decreases their weight, not increases it. Please fix that while this is considered.

Fontbane:
> This decreases their weight, not increases it. Please fix that while this is considered.

After a more careful look at the weighting, I moved it to the more appropriate area below that line. I also made multi hit moves a consideration and adjusted the power with the assumption that they hit at least 2 times. I also included priority moves and "real" multi hit moves (20 base power or above) among the moves guessed to be move reminder moves.

src/field/pokemon.ts
```diff
@@ -1266,8 +1266,10 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     // Weight towards higher power moves, by reducing the power of moves below the highest power.
     // Caps max power at 90 to avoid something like hyper beam ruining the stats.
     // This is a pretty soft weighting factor, although it is scaled with the weight multiplier.
+    // Account for technician boost and multi hit (assume at least 2 hits), compensate for power based weighting on priority moves
+    const hasTechnician = this.hasAbility(Abilities.TECHNICIAN);
     const maxPower = Math.min(movePool.reduce((v, m) => Math.max(allMoves[m[0]].power, v), 40), 90);
```
Xavion3, line 1271: You'll want to modify both max power and actual power, otherwise actual power could be higher than max power which would lead to unusual behavior.

PR #1329, opened 2024-05-24
Fix for the zoom issue (#1326)

### LINKED ISSUE #1326: Accidental zoom on IOS (D-PAD)
I am using Pokerogue on my IOS device and sometimes when I use the D-Pad my device thinks I want to zoom into the page. 
Which results in this perspective:
IMG_BC492384B299-1

The only way to fix it is to restart the app.

I didn't observe this bug before so it might have been added recently..

flx-sta:
adding `user-scalable=no` to the `<meta name="viewport" ...>` element in index.html might fix the zoom bug

fixes #1326


felix-staud:
@torranx what about this alternative solution?

I didn't know where to put the code. 
Feel free to move it somewhere else if that is the solution you want to go with.

index.html
```diff
@@ -13,7 +13,7 @@
 	<link rel="apple-touch-icon" href="./logo512.png" />
 	<link rel="shortcut icon" type="image/png" href="./logo512.png" />
 	<link rel="canonical" href="https://pokerogue.net" />
-	<meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
+	<meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover, user-scalable=no" />
```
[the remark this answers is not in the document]
danieland: This was the exact reason for my earlier comment, glad you guys caught it!

felix-staud:
@danieland see my updated resolution

torranx:
> @torranx what about this alternative solution?
> 
> I didn't know where to put the code. Feel free to move it somewhere else if that is the solution you want to go with.

Code looks good at a glance. Lacks of testing tho. Please provide screenshots/videos.

src/touch-controls.js
```diff
@@ -115,3 +115,29 @@ function bindKey(node, key, buttonMap) {
...
+ *
+ * Prevent zoom on specified element
+ * @param {HTMLElement} element
+ */
+function preventElementZoom(element) {
+  element.addEventListener("touchstart", (event) => {
+    const t2 = event.timeStamp;
+    const t1 = event.currentTarget.dataset.lastTouch || t2;
+    const dt = t2 - t1;
+    const fingers = event.touches.length;
```
torranx, line 130: Can we use more descriptive variable names?

ben-lear:
@felix-staud why change from the original viewport implementation in the index.html? Just curious why you took an event listener approach instead

torranx:
> @felix-staud why change from the original viewport implementation in the index.html? Just curious why you took an event listener approach instead

https://github.com/pagefaultgames/pokerogue/pull/1329#discussion_r1613881717

felix-staud:
> @felix-staud why change from the original viewport implementation in the index.html? Just curious why you took an event listener approach instead

I want to target specific elements rather than the whole page

src/touch-controls.js
felix-staud: @torranx I've updated the variable name to be more descriptive.
I'm still not 100% happy with the absolute random positioning of the code so if there is some ideas for that feel free to share.

torranx: seems fine to me, fits with the current context

src/touch-controls.js
```diff
@@ -115,3 +115,29 @@ function bindKey(node, key, buttonMap) {
...
+    if (!timeStampDifference || timeStampDifference > 500 || fingers > 1) {
+      return;
+    } // not double-tap
+
+    event.preventDefault();
+    event.target.click();
+  });
+}
+
+const dPadCenterRect = document.querySelector("#dpadCenter");
```
torranx, line 142: Shouldn't this be #dpad? So zooming wont be allowed on the empty spaces on the square too, as encountered by other users.

Screenshot 2024-05-28 at 6 20 35 PM

ben-lear:
Just tested, looks like the issue is no longer occurring in any location around the D-pad. Nice job!

ben-lear (approved):

bennybroseph (approved):

PR #1332, opened 2024-05-24
Fix incorrectly unselectable starter formes

Icehawk78:
This PR enables selection of starter Formes that both should be selectable as starters, but that also have in-battle Forme changes, which the current starter selection logic filters out (presumably to prevent users from selecting in-battle Formes as their starter base, like having a Mega-Evo starter).

This currently appears to affect 3 Pokemon: Minior (color selection unavailable), Froakie (Battle Bond selection unavailable), and Zygarde (10%/50% Formes and Ability split selection unavailable).

This change only includes override form keys, rather than modifying the PokemonForm constructor, which would require a significantly larger code change, but this does centralize the logic in the PokemonForm class, rather than on the forme initializations, which is maybe less than ideal. Happy to make the larger code change if moving where the logic is defined is preferred.

Existing behavior for Minior / Froakie / Zygarde
[screen-capture-old.webm](https://github.com/pagefaultgames/pokerogue/assets/1302975/49dc4672-e722-4f9f-bf37-3924385483c9)

New behavior for Minio / Froakie / Zygarde
[screen-capture-new.webm](https://github.com/pagefaultgames/pokerogue/assets/1302975/6500a55e-a4ef-4bca-9616-4892b086530b)

### DIFF
--- a/src/data/pokemon-species.ts
+++ b/src/data/pokemon-species.ts
@@ -794,7 +794,7 @@ export class PokemonForm extends PokemonSpeciesForm {
   public formSpriteKey: string;
 
   // This is a collection of form keys that have in-run form changes, but should still be separately selectable from the start screen
-  private starterSelectableKeys: string[] = ["10", "50", "10-pc", "50-pc", "battle-bond", "red", "orange", "yellow", "green", "blue", "indigo", "violet"];
+  private starterSelectableKeys: string[] = ["10", "50", "10-pc", "50-pc", "red", "orange", "yellow", "green", "blue", "indigo", "violet"];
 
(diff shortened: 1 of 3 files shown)


Madmadness65:
Battle Bond Froakies can already be selected without issue, as it's only Greninja with different forms and a form changing ability, and obviously Greninja cannot be selected as a starter. The others seem to be fine though on first glance.

Icehawk78:
> Battle Bond Froakies can already be selected without issue, as it's only Greninja with different forms and a form changing ability, and obviously Greninja cannot be selected as a starter. The others seem to be fine though on first glance.

Ah, thanks, somehow I missed that in my testing. Is it worth remove that key from the override array?

Madmadness65 (approved):
Tested this, and everything I tried seems to work as intended, so it LGTM now

PR #1333, opened 2024-05-24
Fixed impossible forced switch

0zuzu:
Fixes a bug where the player gets stuck switching after a faint at the start of a double battle, this happens when the only available Pokemon is in the second slot and the previous battle was single.

The bug happens because the party UI knows it's a double battle, so it doesn't allow you to send the only pkmn you have which is in slot 2 (not on the field, since last battle was single). I decided to skip the switch phase entirely in this situation as this is by far the simplest solution I could find.

Here's what changed:
- I added an extra exception for the switch phase to be skipped if the battle hasn't started yet and you only have 1 Pokemon remaining after a faint in the previous battle.

This is what changed from the player's point of view:
- Player no longer sees the switch menu, meaning they don't know that the next battle is a double battle and they no longer get stuck in an impossible switch;
- As a consequence, the player can't pick the final Pokemon in their party before a battle, instead it is sent automatically.  

I have tested, and this change doesn't affect switching in the middle of a battle, sending 2 Pokemons at a time or when you have more than 1 available pkmn.

Video showcasing the bug:

https://github.com/pagefaultgames/pokerogue/assets/63069680/21f39a43-7dfc-414a-95ea-c895104bcce1

After my fix:

https://github.com/pagefaultgames/pokerogue/assets/63069680/3f59aff8-d5e2-457c-bd91-558fbeb44731

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -3286,13 +3286,12 @@ export class FaintPhase extends PokemonPhase {
     }
 
     if (this.player) {
-      const nonFaintedPartyMembers = this.scene.getParty().filter(p => !p.isFainted());
-      const nonFaintedPartyMemberCount = nonFaintedPartyMembers.length;
-      if (!nonFaintedPartyMemberCount)
+      const availablePartyMembers = this.scene.getParty().filter(p => !p.isFainted());
+      if (!availablePartyMembers.length)
         this.scene.unshiftPhase(new GameOverPhase(this.scene));
-      else if (nonFaintedPartyMemberCount >= this.scene.currentBattle.getBattlerCount() || (this.scene.currentBattle.double && !nonFaintedPartyMembers[0].isActive(true)))
+      else if (availablePartyMembers.length >= this.scene.currentBattle.getBattlerCount() || (this.scene.currentBattle.double && !availablePartyMembers[0].isActive(true)))
         this.scene.pushPhase(new SwitchPhase(this.scene, this.fieldIndex, true, false));
-      if (nonFaintedPartyMemberCount === 1 && this.scene.currentBattle.double)
+      if (availablePartyMembers.length === 1 && this.scene.currentBattle.double)
         this.scene.unshiftPhase(new ToggleDoublePositionPhase(this.scene, true));
     } else {
       this.scene.unshiftPhase(new VictoryPhase(this.scene, this.battlerIndex));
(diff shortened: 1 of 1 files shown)


src/phases.ts
```diff
@@ -3286,13 +3286,12 @@ export class FaintPhase extends PokemonPhase {
...
-      if (!nonFaintedPartyMemberCount)
+      const availablePartyMembers = this.scene.getParty().filter(p => !p.isFainted());
+      if (!availablePartyMembers.length)
         this.scene.unshiftPhase(new GameOverPhase(this.scene));
-      else if (nonFaintedPartyMemberCount >= this.scene.currentBattle.getBattlerCount() || (this.scene.currentBattle.double && !nonFaintedPartyMembers[0].isActive(true)))
+      else if (availablePartyMembers.length >= this.scene.currentBattle.getBattlerCount() || (this.scene.currentBattle.double && !availablePartyMembers[0].isActive(true)))
         this.scene.pushPhase(new SwitchPhase(this.scene, this.fieldIndex, true, false));
-      if (nonFaintedPartyMemberCount === 1 && this.scene.currentBattle.double)
+      if (availablePartyMembers.length === 1 && this.scene.currentBattle.double)
         this.scene.unshiftPhase(new ToggleDoublePositionPhase(this.scene, true));
```
0zuzu, line 3295: I renamed "nonFaintedPartyMembers" to "availablePartyMembers" to make it consistent with the rest of the script. There is no change in logic here.
The only reason why I made changes here was because I was trying to implement the fix here initially.

0zuzu, line 3295: I renamed "nonFaintedPartyMembers" to "availablePartyMembers" to make it consistent with the rest of the script. There is no change in logic here.
The only reason why I made these changes was because I was trying to implement the fix here initially.

src/phases.ts
```diff
@@ -3883,17 +3882,18 @@ export class SwitchPhase extends BattlePhase {
 
   start() {
     super.start();
+    const availablePartyMembers = this.scene.getParty().filter(p => !p.isFainted());
 
     // Skip modal switch if impossible
-    if (this.isModal && !this.scene.getParty().filter(p => !p.isFainted() && !p.isActive(true)).length)
+    if (this.isModal && (!availablePartyMembers.filter(p => !p.isActive(true)).length || (!this.scene.currentBattle.started && availablePartyMembers.length == 1)))
       return super.end();
```
0zuzu, line 3889: The only change in logic happens here: I added an extra condition to skip the switchPhase in case the battle hasn't started and player only has 1 pkmn available.

brain-frog:
Can you fix the merge conflicts on this and see if its still an active bug? We can still add this in

flx-sta:
Closed due to inactivity. Mirrored to #3288

PR #1335, opened 2024-05-24
Added if statement to deploy.yml stopping local builds failing.

JonStudders:
image

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   deploy:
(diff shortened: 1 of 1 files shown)


TempsRay:
I assume this is to fix local builds failing? The PR title makes it sound like adding the if statement will cause local builds to fail

JonStudders:
> I assume this is to fix local builds failing? The PR title makes it sound like adding the if statement will cause local builds to fail

fixed clarity of the title.

TempsRay (approved):

PR #1336, opened 2024-05-24
Update running away in Double Battles

molker:
Running with either pokemon now uses the faster pokemon's speed and checks if either have the ability Run Away
If only one pokemon is alive, just use that one.

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -2161,9 +2161,7 @@ export class TurnStartPhase extends FieldPhase {
         if (this.scene.currentBattle.double) {
           const playerActivePokemon = field.filter(pokemon => pokemon.isPlayer() && pokemon.isActive());
           // if only one pokemon is alive, use that one
-          if (playerActivePokemon.length < 2) {
-            runningPokemon = playerActivePokemon[0];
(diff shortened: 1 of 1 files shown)


src/phases.ts
```diff
@@ -2155,13 +2155,21 @@ export class TurnStartPhase extends FieldPhase {
         this.scene.unshiftPhase(new AttemptCapturePhase(this.scene, turnCommand.targets[0] % 2, turnCommand.cursor));
         break;
       case Command.POKEMON:
+        this.scene.unshiftPhase(new SwitchSummonPhase(this.scene, pokemon.getFieldIndex(), turnCommand.cursor, true, turnCommand.args[0] as boolean, pokemon.isPlayer()));
       case Command.RUN:
```
TempsRay, line 2159: Wow was the old phases.ts really having both cases go into the same logic?

molker: sure was. and then using an if statement to do the same check lmao

TempsRay: 💀

TempsRay (approved):

PR #1338, opened 2024-05-24
Add MoveUsedEvent to Listen For When a Move Gets Used

bennybroseph:
No functionality changes and no player facing features. All this does is allow for events to fire related to move usage. The event contains data for:
- The ID of the Pokemon who used the move
- The base Move itself
- The amount of PP used during this turn


Greenlamp2 (commented):
Since you will add more and more events, maybe you should create a folder "events" ?

Greenlamp2 (approved):

PR #1340, opened 2024-05-24
Update Biome Text Labels and Variable Name

bennybroseph:
Updates the new Text which displays the Biome and Wave count variable to a proper name.
Updates Final Destination to ??? to keep the name an easter egg rather than a blatant label.


src/battle-scene.ts
```diff
@@ -1244,12 +1245,13 @@ export default class BattleScene extends SceneBase {
     });
   }
 
-  updateWaveCountText(): void {
+  updateLineOneText(): void {
```
TempsRay, line 1248: This is still `LineOneText` in the green?

bennybroseph: bro chill lol

src/data/biomes.ts
```diff
@@ -21,7 +21,7 @@ export function getBiomeName(biome: Biome | -1) {
   case Biome.SPACE:
     return "Stratosphere";
```
TempsRay, line 22: I would also like this renamed so it matches the biome name too

TempsRay (commented):
Also update commit title to mention renaming biome names

TempsRay (approved):

PR #1341, opened 2024-05-24
Implement Triple Kick/Triple Axel/Population Bomb

xsn34kzx:
Follows up on #365 and #734, where the individual accuracy checks for each hit of the moves were a point of concern in addition to incrementing the base power of the moves for every hit.

A proposed implementation of the following moves: [`Triple Kick`](https://bulbapedia.bulbagarden.net/wiki/Triple_Kick_(move)#Effect), [`Triple Axel`](https://bulbapedia.bulbagarden.net/wiki/Triple_Axel_(move)#Effect), and [`Population Bomb`](https://bulbapedia.bulbagarden.net/wiki/Population_Bomb_(move)#Effect).
- Adds `MultiHitPowerIncrementAttr` extending `VariablePowerAttr`
- Adds `CHECK_ALL_HITS` flag

In regards to grouping these moves together, `Triple Kick` and `Triple Axel` are most readily comparable since they both increment base power per hit and check accuracy for each hit. `Population Bomb` only shares the latter quality, but the change to include this functionality was simple enough to lump it in with the other two. Also, the closed PRs above mentioned the inclusion of the moves, so might as well address all of them at once.

Only thing I'm unsure of is how sparse/variable flags are, since the implementation below makes it pretty simple to set and check without having to do extensive look-ups on the category of the move, etc. Definitely can be substituted though, this was just the cleanest and simplest idea.

### General Demo Video  + Console Log

https://github.com/pagefaultgames/pokerogue/assets/109637146/9245f517-2657-46e2-b033-74cf2f105c94

tripleAxelIncrementVideo

Demonstrates that the accuracy check on the first hit doesn't guarantee all hits.

### Console Log of Full Triple Axel / Triple Kick

tripleAxelIncrement

tripleKickIncrement

Demonstrates that the all hits provide the behavior described on the Bulbapedia pages for both moves.

### Population Bomb Demo Video

https://github.com/pagefaultgames/pokerogue/assets/109637146/bfef1693-251e-4541-aa50-9a32a0aaf3b2

populationBombLog

Demonstrates that the accuracy check on the first hit doesn't guarantee all hits, as described on the move's Bulbapedia page, while also showing that the behavior is not tied to the incremental base power shown for `Triple Kick` and `Triple Axel`.


LaukkaE:
Something to think about is just how absurdly strong Triple Axel would be with this implementation with multi-lens.
If you have three multi-lenses, the base power of the last hit will be 240 (not taking into account the multi-lens decrease). 
If my calcs are right, this would account to a "normal" move hitting once with 390 bp(130 * 3) if you can cap the accuracy.

TempsRay:
Could always cap the BP of it so it will never exceed the max for how it normally should be (60 for triple axel, 30 for triple kick)

xsn34kzx:
The thing about this implementation is that the new Attr, by extending `VariablePowerAttr`, takes advantage of the attribute application [here](https://github.com/pagefaultgames/pokerogue/blob/5e7a9b08724f39bb43c001e26c788eca08e64d41/src/field/pokemon.ts#L1651), so the power increase is done before items are applied, etc. and accounting for that would probably involve another `Attr` and `applyMoveAttrs` call after the fact to affect the final BP. Not to dismiss the balance at all, but code for adjusting/capping the BP would probably be better in a follow-up PR, at least to not throw system-related changes in a PR for implementing the move behavior one-to-one from Bulbapedia. I'm down to do that though after the fact or in this PR if necessary.

xsn34kzx:
> Could always cap the BP of it so it will never exceed the max for how it normally should be (60 for triple axel, 30 for triple kick)

Ended up thinking about and running with this idea and added another parameter to the `MultiHitPowerIncrementAttr` class that caps off the allowed amount of increments. This doesn't hard cap off BP at that value since it is definitely a system thing that I wouldn't want to touch w/o outside opinion, but it does prevent the BP from going out of control or moving past the intended max incremental BP (before `Multi Lens`'s power decreases & any item increases ofc).

It seems to work as is, but early waves aren't really the best to test the move so let me know if anything.
multiLensDemo

jaimefd:
Does the "user.turnData.hitCount" in the apply method have the final number of hits after multi lens is applied? Because if that's the case, you could just simply divide hitsTotal by the max number of hits to get the power of each hit.

For example if you got that "hitCount" has 6 hits and you are using Triple Axel, you know that the user has 1 multi lens and just make it so that the hits are 20, 20, 40, 40, 60, 60.

I don't really like having it like 20, 40, 60, 60, 60, 60... seems pretty broken

molker:
> Does the "user.turnData.hitCount" in the apply method have the final number of hits after multi lens is applied? Because if that's the case, you could just simply divide that value by the max number of hits to figure out if the user has Multi Lens and assign power accordingly
> 
> For example if you got that "hitCount" has 6 hits and you are using Triple Axel, you know that the user has 1 multi lens and just make it so that the hits are 20, 20, 40, 40, 60, 60.
> 
> I don't really like having it like 20, 40, 60, 60, 60, 60... seems pretty broken

I mean multilens is broken in general that's why its a master tier item but I see what you're saying. realistically I think the way to make it be "accurate" to the move is to do `20, 40, 60, 20, 40, 60...` depending on the amount of multi-lens

Xavion3:
Does this actually work with population bomb? At a glance it looks like it'll effectively be applying the accuracy check twice.

Also has this been tested it interacts correctly with skill link? Given the changes to the accuracy code.

xsn34kzx:
> Does this actually work with population bomb? At a glance it looks like it'll effectively be applying the accuracy check twice.
> 
> Also has this been tested it interacts correctly with skill link? Given the changes to the accuracy code.
> 
> On multi lens, it's really how we want it to act. Like if we model it off parental bond, trading lower damage for easier access, then it shouldn't affect moves that are already multi-hit moves.

Before I did the implementation, I pretty much saw this [line](https://github.com/xsn34kzx/pokerogue/blob/dd5d63f4cd580b45d837a627e8cc143126e44c79/src/phases.ts#L2781) through the step debugger and made the flag in response to that essentially. Because as far as I interpreted it, misses are handled by setting `hitCount` and `hitsLeft` to `1`. So, if the first accuracy check ever misses, the whole move stops and that's how all other multi-hits work without the `CHECK_ALL_HITS` flag being set on that move. If `CHECK_ALL_HITS` is set, then it'll never return true at that point, effectively skipping that for multi-hit moves with the flag. Luckily, this if is also before any "ignore accuracy" ability, item, etc. since it's at the very beginning of `hitCheck` and the `AlwaysHitAbAttr` is checked after that line change in the accuracy check function (effectively making things like Skill Link work). Unless there's another if statement that multi-hits effect (or at least related to `hitCount` and `hitsLeft`), then this should just skip the return true for multi-hits solely and nothing else in the accuracy check related to external accuracy-affecting things. As far as I checked debugging after the fact, it just skipped past that if and went into the accuracy check as normal, so it should be fine unless there's other things in the code base that I didn't see or wasn't thorough enough with.

xsn34kzx:
Ended up investigating `Skill Link` and how it's implemented directly, turns out it has no check in the `hitCheck` function at all, which works fine for other multi-hit moves since they implement `Skill Link` to effectively get the maximum number of hits instantly when the `hitCount` is needed. Accuracy doesn't matter for those, so they run for the max amount of hits naturally. Because the three moves here instead already give the max number of hits and rely on accuracy checks to continue, `Skill Link` isn't used at all to determine if they hit. Pretty much just need to check if the ability is there in accuracy check, and that's fine. Another thing is that there wasn't a "10" hit enum for `Population Bomb`, and it was partially implemented as "1 to 10" hits without the separate accuracy checks. I'll throw these two things in as a commit first, and then tackle the revolving "20, 40, 60, 20, 40, 60, ..." BP increments after.

xsn34kzx:
The logic for `Population Bomb` should now be done completely. The only thing that's kinda broken is that `Skill Link` doesn't provide a pop-up in-game for these moves since that's only done [here](https://github.com/xsn34kzx/pokerogue/blob/dd5d63f4cd580b45d837a627e8cc143126e44c79/src/data/move.ts#L1268) and not in the other cases. The "fix" would be to just call the function even if calling `applyAbAttrs` in the other cases with a dummy `hitValue` even if it doesn't affect anything, but I feel like that's wasteful/hacky. Other than that though, `Skill Link` should be fully accounted for now.

xsn34kzx:
> realistically I think the way to make it be "accurate" to the move is to do `20, 40, 60, 20, 40, 60...` depending on the amount of multi-lens

The latest commit implements this behavior now 👍🏼 , tested with max `Multi Lens`.

JoannaFalkowska:
> I mean multilens is broken in general that's why its a master tier item but I see what you're saying. realistically I think the way to make it be "accurate" to the move is to do `20, 40, 60, 20, 40, 60...` depending on the amount of multi-lens

Shouldn't it be the other way around? Doesn't PokeRogue Multi-Lens generally imitate what would happen if you had Parental Bond that affects multi-hit moves?

Bulbapedia [Parental Bond page](https://bulbapedia.bulbagarden.net/wiki/Parental_Bond_(Ability)) says:
>Additionally, there is only one accuracy check, so either both strikes hit or both strikes miss.

I think this suggests that both strikes are supposed to be literally the same, including an accuracy check. So it would make more sense IMO to do:

(accuracy check) 20-20-... (accuracy check) 40-40-... (accuracy check) 60-60-...

Otherwise, if you want to implement a Triple Axel rule that each strike has increased BP, then you'd have to respect the other Triple Axel rule that there must be an accuracy check between each strike. So with three Multi-Lenses you'd have to do something like:

[(accuracy check) 20 (accuracy check) 40 (accuracy check) 60 (accuracy check) 80] * 3 times

Which IMO doesn't seem like that's in the spirit of Parental Bond rules at all.

bennybroseph:
Has anyone simply tried this in Pokemon Showdown. Mega Kangaskhan with Triple Axel. What happens?

Blitz425:
The way Multi Lens currently functions and the way Parental Bond is meant to function is completely different. If you test Bullet Seed with Multi Lens currently and land all 5 Bullet Seeds, you will get an additional 5 Bullet Seeds per Multi Lens. The proposed 20 40 60 20 40 60 would be in line with how Multi Lens currently functions, it is a Rogue tier item for a reason. Another thing is Multi Lens does not check for accuracy, it's only dependent on if the first move landed in the first place. If it isn't already, an interaction with a single multi lens maybe would function something like 20 40 *miss* 20 40.

XGSleepWalker (approved):

PR #1343, opened 2024-05-24
Added base-state toggle to starter menu

prime-dialga:
A toggle to view the base states for a selected starter and its evolutions as well as forms has been added.

Also, menus can have an optional onHover trigger now. (use menu config parameter "supportHover" to enable the function)
[Note: The colors for the base-state-chart as well as the translations might require additional work/checks]

base-stat-1
base-stat-2
base-stat-3
base-stat-4


bennybroseph (changes requested):
I really like this in theory, but it just feels so clunky. 

I would have prefered this had a design doc and proposal before so much effort was put into it. This feels like something that should go into the Pokedex that's being worked on, not directly in this menu. I'm not sure why you'd want to view all of this Pokémon's base stats for every evolution and form it has. Takes up a lot of space for something so niche.

Also the option being called Toggle Base Stats is misleading since it doesn't do that. It brings you to a secondary menu where you can then actually do what's on the tin.

prime-dialga:
I did try it as a pure toggle in an earlier commit. However having overlapping charts for each evo and form felt very cluttered and confusing.
So, this is a hybrid solution of both, it still toggles the base-stat view, but also offers a list to check each form and evolution (if they have different base-stats)
If the user exits this list without toggling it off, then the base-stats view will continue to be displayed for the base form of the pokemon.

Why is this in the starter menu and not the pokedex? I felt like an easy way to access the base stats during the starter selection was required. It's imo necessary to chose the most optimal nature and starter moves for the pokemon. And sure, you can open the pokedex (or a third party site in a new browser tab), but that is not as fast a this sub-menu.

Why is this called a Toggle? It technically still toggles the view and i couldn't find a better name... well, maybe just "Base Stats" and i remove the option to exit the "Base Stats" view from the menu? (the user can still exit it with the "back" button)

prime-dialga:
The name for the menu option has been changed and toggling behavior has been improved.
Using the "Toggle Base-Stats" option will now toggle an IV-like mode, otherwise, when canceling the menu, the previous mode will be restored.

base-stat-5

molker:
As nice of an idea as this is I don't believe this is something we'll want to add, it's a bit unnecessary for main implementation. I think something like this could be a good extension/overlay for personal/modding use

PR #1346, opened 2024-05-25
Changing Biome Text Colour

zaccie:
- On light coloured biomes this change should be similar readability
- In darker biomes it's has much easier readability
- Now becomes more consistent colour with the rest of the game text which just looks better imo
old text is the grey with yellow drop shadow
biometext
biometext2
biometextmobile


TempsRay (approved):
I think ideally it would be a custom color for each biome but I think for the most part white is more legible than black

PR #1348, opened 2024-05-25
Players win if classic and past floor 200

TempsRay:
Added a check in `EncounterPhase` where if the player is in classic mode and they are past floor 200 it automatically goes to the `GameOverPhase` where the run is treated as a victory and rewards are given.


src/phases.ts
```diff
@@ -729,6 +729,11 @@ export class EncounterPhase extends BattlePhase {
 
     this.scene.initSession();
 
+    // Failsafe if players somehow skip floor 200 in classic mode
+    if (this.scene.gameMode.modeId === GameModes.CLASSIC && this.scene.currentBattle.waveIndex > 200) {
```
molker, line 733: nitpick but you can just use gameMode.isClassic

bennybroseph (approved):

PR #1350, opened 2024-05-25
Update zh_CN ability.ts and move.ts

SnowCharmQ:
## What are the changes:

- fixed an error in the translation of the _Frisk_ ability.
- fixed an error in the translation of the _As One_ ability for Glastrier.
- swapped the translated description of the _Embody Aspect_ ability when carrying _Teal Mask_ and _Wellspring Mask_.

### DIFF
--- a/src/locales/zh_CN/ability.ts
+++ b/src/locales/zh_CN/ability.ts
@@ -475,7 +475,7 @@ export const ability: AbilityTranslationEntries = {
   },
   frisk: {
     name: "察觉",
-    description: "出场时，可以察觉对手的持有物。",
+    description: "出场时，可以察觉对手的特性。",
   },
   reckless: {
     name: "舍身",
--- a/src/locales/zh_CN/ability.ts
+++ b/src/locales/zh_CN/ability.ts
@@ -475,7 +475,7 @@ export const ability: AbilityTranslationEntries = {
   },
   frisk: {
     name: "察觉",
-    description: "进入战斗时，神奇宝贝可以检查对方神奇宝贝的能力。",
+    description: "出场时，可以察觉对手的持有物。",
   },
   reckless: {
     name: "舍身",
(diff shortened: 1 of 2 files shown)


src/locales/zh_CN/ability.ts
```diff
@@ -475,7 +475,7 @@ export const ability: AbilityTranslationEntries = {
   },
   frisk: {
     name: "察觉",
-    description: "进入战斗时，神奇宝贝可以检查对方神奇宝贝的能力。",
+    description: "出场时，可以察觉对手的持有物。",
```
TempsRay, line 478: In Pokerogue Frisk checks ability rather than item

src/locales/zh_CN/move.ts
```diff
@@ -2915,7 +2915,7 @@ export const move: MoveTranslationEntries = {
   },
   "zippyZap": {
     name: "电电加速",
-    effect: "The user attacks the target with bursts of electricity at high speed. This move always goes first and raises the user's evasiveness.",
+    effect: "迅猛无比的电击。必定能够\n先制攻击，击中对方的要害。",
```
TempsRay, line 2918: This says it will always crit but the current implementation of zippy zap right now is still boosting evasion

TempsRay (changes requested):
Frisk and zippy zap are currently incorrect with the way Pokerogue has them.

The other changes are fine though

src/locales/zh_CN/ability.ts
SnowCharmQ: I will fix it now. But I think it will be better to keep up with the official in the future :)

src/locales/zh_CN/move.ts
SnowCharmQ: The same. I think it will be better to keep up with the official version in the future.

src/locales/zh_CN/ability.ts
bennybroseph: Frisk cannot work like the mainline series. you already know your opponents items

TempsRay (approved):

src/locales/zh_CN/move.ts
bennybroseph: > Generation VIII onwards
Zippy Zap cannot be selected in a battle. However, it has been modified: its PP was changed to 10 while its power was changed to 80. Additionally, its secondary effect is raising evasion by one stage, instead of a guaranteed critical hit.

It is

[source](https://bulbapedia.bulbagarden.net/wiki/Zippy_Zap_(move))

SnowCharmQ: Sorry, I just saw the version in Generation VII. It is correct now.

PR #1351, opened 2024-05-25
Made Lab 25% instead of 12.5% to encounter after Factory

damocleas:
This biome is too rare, this is just a temporary change until biomes are redone.

It was also the only biome that rare, despite the specific path needed to get there.


Reibd:
This sounds great because lab is basically non-existent now

XGSleepWalker (approved):

Issue #1352, opened 2024-05-25
Reminder for small tasks in Ability.ts


Tempo-anon:
* Illuminate should have the IgnoreOpponentEvasionAbAttr (fix comments on this abattr)
* Check if unaware only ignores atk, spA, def, sdef, eva, and acc, should not ignore speed. (Also add comments)
* Implement honey gather (maybe make it like pickup but just give money since honey can only be sold for money in SV)

PR #1358, opened 2024-05-25
Fix Unaware + Illuminate and Implement Honey Gather

JakubHanko:
Done small tasks mentioned in #1352 

- Made Unaware ignore speed changes
- Added `IgnoreOpponentEvasionAbAttr` to Illuminate (Ability was changed in SV) and fixed 
- Implemented Honey Gather which behaves, right now, like Pay day move.

https://github.com/pagefaultgames/pokerogue/assets/60473007/762b375f-c908-4b3e-b221-d96c311db988

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -1377,11 +1377,11 @@ export class IgnoreOpponentEvasionAbAttr extends AbAttr {
     super(false);
   }
   /**
-   * Checks if enemy Pokemon is trapped by an Arena Trap-esque ability
+   * Changes the evasion stat to 0 during hit check computation.
    * @param pokemon N/A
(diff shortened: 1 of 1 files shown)


JakubHanko:
After some thinking while AFK:

IgnoreOpponentStatChanges ability attribute is and will be used apparently only with Unaware. I think it might make sense to rename it to UnawareAbAttr.

Another thing is that Honey Gather could maybe use a custom message? As in "Pokemon has picked up some Honey which was sold for XY."

bennybroseph:
we should split this into multiple PR if it tackles multiple issues.

JakubHanko:
Okay, will do right away

PR #1360, opened 2024-05-25
Implement Honey Gather

JakubHanko:
Implements honey gather ability.

Right now, it behaves like pickup (i.e. it gives you money post battle) and the amount of money is same as one Pay Day attack.

https://github.com/pagefaultgames/pokerogue/assets/60473007/fe5c93e0-c21e-482b-a766-824a1d9918af


bennybroseph:
I suppose this is fine for now. Better than doing literally nothing. Should update the ability text to match though

JakubHanko:
Agree with the ability text change. Should I try to keep some flavour and change it to "Pokemon gathers some Honey to be sold after a battle." Or just straight up "Pokemon picks up some money after a battle"?

I dont feel competent enough to do translations a well

bennybroseph:
"The Pokémon may gather Honey after a battle." > "The Pokémon gathers Honey after a battle. The Honey is then sold for money."

bennybroseph (approved):

PR #1361, opened 2024-05-25
Fix Unaware

JakubHanko:
Right now, Unaware could still ignore speed stat changes (which is wrong).

Moreover, IgnoreOpponentStatChangesAbAttr was tightly coupled to this move so I changed its name to UnawareAbAttr and made it ignore speed stat.


JakubHanko:
Didn't think about that. The "fix" is then lacking and calling it a fix is an overstatement. It does make the situation better though in my opinion 😅

Closing the PR, I have bitten off more than I can chew. This needs a proper redesign to work well.

PR #1363, opened 2024-05-25
Fix Illuminate

JakubHanko:
After SV, the Illuminate ability was changed to also prevent other Pokémon from lowering the accuracy stat stage of the Pokémon with this Ability, and also ignores the target's evasion stages. 

The change is just adding of IgnoreOpponentEvasionAbAttr attr and update of misleading docs.

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -1377,11 +1377,11 @@ export class IgnoreOpponentEvasionAbAttr extends AbAttr {
     super(false);
   }
   /**
-   * Checks if enemy Pokemon is trapped by an Arena Trap-esque ability
+   * Changes the evasion stat to 0 during hit check computation.
    * @param pokemon N/A
    * @param passive N/A
    * @param cancelled N/A
-   * @param args [0] {@linkcode Utils.IntegerHolder} of BattleStat.EVA
+   * @param args [0] {@linkcode Utils.IntegerHolder} of {@linkcode BattleStat.EVA}
    * @returns if evasion level was successfully considered as 0
    */
   apply(pokemon: Pokemon, passive: boolean, cancelled: Utils.BooleanHolder, args: any[]) {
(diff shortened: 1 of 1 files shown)


bennybroseph:
I know this wasn't coded by you, but this attribute uses the exact same value. Can we test this to make sure it isn't resetting Evasion?

JakubHanko:
> I know this wasn't coded by you, but this attribute uses the exact same value. Can we test this to make sure it isn't resetting Evasion?

I am not sure I understand the question correctly. The newly added attribute is working with Utils.IntegerHolder and it is nowhere near any setting of battle stats so it shouldn't be able to do any permanent damage.

Or is your question alluding to a video demonstrating that the evasion is not reset?

src/data/ability.ts
```diff
@@ -1377,11 +1377,11 @@ export class IgnoreOpponentEvasionAbAttr extends AbAttr {
...
-   * Checks if enemy Pokemon is trapped by an Arena Trap-esque ability
+   * Changes the evasion stat to 0 during hit check computation.
    * @param pokemon N/A
    * @param passive N/A
    * @param cancelled N/A
-   * @param args [0] {@linkcode Utils.IntegerHolder} of BattleStat.EVA
+   * @param args [0] {@linkcode Utils.IntegerHolder} of {@linkcode BattleStat.EVA}
    * @returns if evasion level was successfully considered as 0
    */
   apply(pokemon: Pokemon, passive: boolean, cancelled: Utils.BooleanHolder, args: any[]) {
```
bennybroseph, line 1387: It's right there

image

bennybroseph (changes requested):

src/data/ability.ts
JakubHanko: I can see that it's right there but I still don't understand what am I supposed to do.

The attribute is applied only in MoveEffect#hitCheck and yes, the evasion is "reset" but only in the scope of that function (since we are not working directly with the battle stat but only with the IntegerHolder).

bennybroseph: We just had this conversation, what's the difference?

image

JakubHanko: None, in the Unaware PR the code was working as well (same reasons as I keep reiterating here). When I closed the PR, I was AFK and I got convinced that I was wrong.

Moreover, I didn't reopen the Unaware PR because I found the code unnecessarily bloated. In the end, my changes added nothing relevant because the speed was already being ignored with how the code is structured.

bennybroseph: Understood, closing this as well

JakubHanko: Not sure why. The "understood" is clearly misplaced.

PR #1365, opened 2024-05-25
Implement Spectral Thief

jaimefd:
Implemented the move Spectral Thief.

It steals positive stat boosts from the target before hitting them, as described [here](https://bulbapedia.bulbagarden.net/wiki/Spectral_Thief_(move)#Effect). Abilities that protect or alter stat changes only work on the attacker, not on the defender, even abilities that can't be ignored like Full Metal Body don't stop it, as described in its own [page](https://bulbapedia.bulbagarden.net/wiki/Full_Metal_Body_(Ability)#Effect)

Here's an example:

https://github.com/pagefaultgames/pokerogue/assets/22176793/f3685d48-7948-42c3-9427-a4d0674fbfe4

If the move fails, like the target being immune, stats are not stolen, as expected:

https://github.com/pagefaultgames/pokerogue/assets/22176793/341a728a-1f3b-4e5b-83be-18cbad6f53d9

And as mentioned above, abilities are still applied for the attacker, in this case we have both Simple and Contrary:

https://github.com/pagefaultgames/pokerogue/assets/22176793/e5449ba0-0c0e-4f48-874b-c95c2149282f

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -2903,6 +2903,11 @@ export class PreAttackStatChangeAttr extends MoveAttr {
 /**
  * Attribute used for moves that steal the enemy's positive stat boosts before dealing damage
  * @extends PreAttackStatChangeAttr
+ * @param user The {@linkcode Pokemon} that is stealing the boosts
+ * @param target The {@linkcode Pokemon} that is having their positive stat boosts stolen and reset to 0
+ * @param move N/A
+ * @param args [0] {@linkcode boolean} Whether the move could hit the target, if false (immune, dodged...) no stats are stolen
+ * @returns true if stats are stolen, false if move didn't hit or the target had no positive stats to take
  */
 export class StealStatBoostsAttr extends PreAttackStatChangeAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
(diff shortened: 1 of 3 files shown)


jaimefd:
Attack boosts were being applied after the move, when they should be applied before, the last commit fixed that:

https://github.com/pagefaultgames/pokerogue/assets/22176793/a996f89d-b5e2-4876-88bf-880facf7c945

src/data/move.ts
```diff
@@ -2801,6 +2801,51 @@ export class DefDefAttr extends VariableDefAttr {
...
+export class StealStatBoostsAttr extends PreAttackStatChangeAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    // Can't steal if the move doesn't land
+    const canHit = args[0] as boolean;
+    if (!user || !target || !canHit) {
+      return false;
+    }
+
+    // If the target has any stat boost (at least +1 in any stat)
+    if (target.summonData.battleStats.find((stat) => stat > 0)) {
```
XGSleepWalker, line 2827: Can you rework this logic to iterate only a single time for battleStats?

XGSleepWalker (changes requested):

src/data/move.ts
```diff
@@ -2801,6 +2801,51 @@ export class DefDefAttr extends VariableDefAttr {
...
+    return false;
+  }
+}
+
+/**
+ * Attribute used for moves that steal the enemy's positive stat boosts before dealing damage
+ * @extends PreAttackStatChangeAttr
+ */
+export class StealStatBoostsAttr extends PreAttackStatChangeAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
```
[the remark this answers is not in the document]
jaimefd: Done

src/data/move.ts
jaimefd: Done, I reduced it from 4 iterations to just 2. I believe we should still iterate at least twice, once to take the stats from the target, and the second time to apply them to the user.

This is because StatChangePhase can only increase stats by the same amount, so if the target has different levels of boost for different stats, we need multiple phases. And I don't think it's a good idea to start any boost phase before removing all stats, in case we ever have abilities that react to enemy stat boosts.

jaimefd:
Tested it and everything still works fine after the changes

flx-sta:

-----

PR #1370, opened 2024-05-25
Bugfix: Right Pokemon no longer hits self when left dies with no replacement when using spread move

td76099:
## What are the changes?
Adds a clause in `MoveEffectPhase` that skips user being its own target if the move selected is `MoveTarget.ALL_NEAR_OTHERS`. This encompasses moves like `Discharge`, `Earthquake`, etc. Tried to limit it to the main problem moves being reported in Discord to minimize impact just in case.

It's a bug - unintended behavior. At a bad time too because it means the user only has one Pokemon left 😬 

## What did change?
Right Pokemon no longer hits itself with spread move if it becomes the only Pokemon after left dies before spread move goes off. 

## Videos
Before change (includes left dying w/ right using spread move which is the problem and right dying w/ left using the spread move which works properly rn)

https://github.com/pagefaultgames/pokerogue/assets/85713900/073d0eb9-46c2-467c-aa80-d4d9d1cf6ac6

After change

https://github.com/pagefaultgames/pokerogue/assets/85713900/212e6345-0917-4e7b-9969-567d461b4b1d

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -2629,8 +2629,14 @@ export class MoveEffectPhase extends PokemonPhase {
 
   constructor(scene: BattleScene, battlerIndex: BattlerIndex, targets: BattlerIndex[], move: PokemonMove) {
     super(scene, battlerIndex);
-
     this.move = move;
+    // In double battles, if the right Pokemon selects a spread move and the left Pokemon dies
+    // with no party members available to switch in, then the right Pokemon takes the index
+    // of the left Pokemon and gets hit unless this is checked.
+    if (targets.includes(battlerIndex) && this.move.getMove().moveTarget === MoveTarget.ALL_NEAR_OTHERS) {
+      const i = targets.indexOf(battlerIndex);
+      targets.splice(i,i+1);
+    }
     this.targets = targets;
   }
 


bennybroseph:
This works, but this doesn't feel like the place where this should go. This is after an accuracy check right before a move shows animations and everything. Shouldn't you just remove yourself from the target list way back when those are being calculated?

td76099:
> This works, but this doesn't feel like the place where this should go. This is after an accuracy check right before a move shows animations and everything. Shouldn't you just remove yourself from the target list way back when those are being calculated? Even if not, I know for sure you can do it earlier than this implementation.

I put it there in case it got removed for any other reason - I can most certainly put it at the beginning tho if you think that makes the most sense.

bennybroseph:
Yes. There's no need to run an accuracy check on yourself for a move that shouldn't target yourself

td76099:
> Yes. There's no need to run an accuracy check on yourself for a move that shouldn't target yourself

Changed it so it as early as possible within this class (the constructor)

Another video for verification purposes

https://github.com/pagefaultgames/pokerogue/assets/85713900/3b0eab2a-7741-4ba1-8a5c-8de115d49d79

bennybroseph:
what about when there are no valid targets left, only the user who got swapped

td76099:
> what about when there are no valid targets left, only the user who got swapped

Reasonable question - when first passed to the constructor, it gives the battle indexes of all affected regardless of status. However, in `start()`, targets gets assigned to the output of `getTargets()` which checks if the battler at those indices is active. If not, they are not included in the returned array. This is when dead Pokemon are typically removed. Therefore, the `applyMoveAttrs` function has plenty of checks already in it for when targets is just `[]`. It's like when a Pokemon with `Rock Slide` goes at the very end after the enemy side has already been killed. The move tries to go off, but the `targets` array is empty so it doesn't do anything.

image


https://github.com/pagefaultgames/pokerogue/assets/85713900/1d0ea79f-f040-4385-b503-cbd2cbb09737

bennybroseph (approved):

PR #1375, opened 2024-05-25
Classic Turbo game mode

arperry:
This adds a shorter version of Classic that skips half of the levels but never any fixed fights or bosses.

While you now heal every five fights instead of ten, the game ends up more difficult, largely because you receive 44% of the usual number of shops (4 out of every 5 instead of 9 of 10). To counterbalance this, rerolls have been made much less expensive. You end up with much less money than usual, so that change isn't nearly enough to bring it to parity. Still, it's probably best that this game mode errs on the side of too difficult rather than too easy.

There is a single known issue with the game mode: trainers fought immediately after a biome change do not appear on screen. This is guaranteed to happen at wave 182. It is a cosmetic bug that has no gameplay effects.

### DIFF
--- a/public/battle-anims/psychic.json
+++ b/public/battle-anims/psychic.json
@@ -1,6 +1,6 @@
 {
   "id": 94,
-  "graphic": "PRAS- PsychicBG",
+  "graphic": "PRAS- Psychic BG",
   "frames": [
     [
       {
--- a/public/battle-anims/psychic.json
+++ b/public/battle-anims/psychic.json
@@ -1,6 +1,6 @@
 {
   "id": 94,
-  "graphic": "PRAS- Psychic BG",
+  "graphic": "PRAS- PsychicBG",
   "frames": [
     [
       {
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -863,16 +863,18 @@ export default class BattleScene extends SceneBase {
     if (newWaveIndex === undefined) {
       if (this.currentBattle?.waveIndex === undefined) {
         newWaveIndex = startingWave;
-      } else if (this.gameMode.isTurbo) {
+      } else {
         newWaveIndex = this.currentBattle?.waveIndex + 1;
-        if (!fixedBattles[newWaveIndex]) {
+
+        // Turbo mode: If we don't have a fixed battle next, skip a wave
+        if (this.gameMode.isTurbo && !fixedBattles[newWaveIndex]) {
           newWaveIndex++;
+          // If we're on an odd wave (because we just started the game or left a fixed battle), skip back to even
+          // Effectively, Turbo spends all its time on even-numbered waves except to hit fixed battles
           if (newWaveIndex % 2 && !fixedBattles[newWaveIndex]) {
             newWaveIndex++;
           }
         }
-      } else {
-        newWaveIndex = this.currentBattle?.waveIndex + 1;
       }
     }
     let newDouble: boolean;
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -859,7 +859,22 @@ export default class BattleScene extends SceneBase {
   }
 
   newBattle(waveIndex?: integer, battleType?: BattleType, trainerData?: TrainerData, double?: boolean): Battle {
-    const newWaveIndex = waveIndex || ((this.currentBattle?.waveIndex || (startingWave - 1)) + 1);
+    let newWaveIndex = waveIndex;
+    if (newWaveIndex === undefined) {
+      if (this.currentBattle?.waveIndex === undefined) {
+        newWaveIndex = startingWave;
+      } else if (this.gameMode.isTurbo) {
+        newWaveIndex = this.currentBattle?.waveIndex + 1;
+        if (!fixedBattles[newWaveIndex]) {
+          newWaveIndex++;
+          if (newWaveIndex % 2 && !fixedBattles[newWaveIndex]) {
+            newWaveIndex++;
+          }
+        }
+      } else {
+        newWaveIndex = this.currentBattle?.waveIndex + 1;
+      }
+    }
     let newDouble: boolean;
     let newBattleType: BattleType;
     let newTrainer: Trainer;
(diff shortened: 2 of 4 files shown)


public/battle-anims/psychic.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": 94,
-  "graphic": "PRAS- PsychicBG",
+  "graphic": "PRAS- Psychic BG",
```
bennybroseph, line 3: What's this for

src/phases.ts
```diff
@@ -4971,7 +4982,14 @@ export class SelectModifierPhase extends BattlePhase {
...
     }
-    return Math.min(Math.ceil(this.scene.currentBattle.waveIndex / 10) * baseValue * Math.pow(2, this.rerollCount), Number.MAX_SAFE_INTEGER);
+    const cost = Math.min(
+      Math.ceil(this.scene.currentBattle.waveIndex / 10) * baseValue * Math.pow(2, this.rerollCount),
+      Number.MAX_SAFE_INTEGER);
+    // Turbo has less than half the items offered; make rerolls cheaper
+    if (this.scene.gameMode.isTurbo) {
+      return Math.ceil(cost / 4);
+    }
+    return cost;
```
bennybroseph, line 4992: Should we be making everything cheaper, not just rerolls?

src/game-mode.ts
```diff
@@ -162,6 +165,8 @@ export class GameMode implements GameModeConfig {
     switch (modeId) {
     case GameModes.CLASSIC:
       return waveIndex === 200;
+    case GameModes.TURBO:
+      return waveIndex >= 200;
```
bennybroseph, line 169: shouldn't this just fallthrough to match classic?

src/battle-scene.ts
```diff
@@ -859,7 +859,22 @@ export default class BattleScene extends SceneBase {
...
+        if (!fixedBattles[newWaveIndex]) {
+          newWaveIndex++;
+          if (newWaveIndex % 2 && !fixedBattles[newWaveIndex]) {
+            newWaveIndex++;
+          }
+        }
+      } else {
+        newWaveIndex = this.currentBattle?.waveIndex + 1;
+      }
+    }
```
bennybroseph, line 877: can we get some comments in here? Needing this many nested conditionals doesn't seem right

bennybroseph (changes requested):

bennybroseph:
Could we get some screenshots or videos of what this looks like? Also, how was the reward system changed for this mode since it's way faster than Classic? Also, what are the unlock conditions for the mode?

public/battle-anims/psychic.json
arperry: There is no "PsychicBG" file; this fixes a typo.

src/game-mode.ts
arperry: If Classic is fine with the >= logic, then that's OK. Otherwise, as Turbo skips 191, the logic to detect the final biome fails. I figured this was less intrusive.

public/battle-anims/psychic.json
bennybroseph: let's make this a separate pull request

src/battle-scene.ts
arperry: Added.

src/game-mode.ts
bennybroseph: What floor does Turbo end on if not 200?

arperry:
> Could we get some screenshots or videos of what this looks like?

I'm afraid it's not visually much different from Classic mode. Here's the menu:
image

Here's a GIF of the first few waves (1, 4, 5, 8 - the next would be 10).

Recording 2024-05-25 at 15 44 03


> Also, how was the reward system changed for this mode since it's way faster than Classic?

Experience is increased by a factor of 1.5. Because this mode is trainer-heavy, it gives more experience overall, but some of that will usually be lost to fainted Pokémon (more than usual, I mean).

Rerolls cost 25% as much. Other item costs are unchanged. Because heals are more frequent, you don't need to purchase items much before the E4, but cash is noticeably tighter at that point. The aim is to roughly match Classic in difficulty, but it errs on the side of being a bit harder.

> Also, what are the unlock conditions for the mode?

Because you don't get a mode select until you unlock Endless, it currently shares the Endless unlock condition: clear Classic mode.

src/phases.ts
arperry: Possibly, but as a starting point for this mode, I didn't want to overcorrect and make it easier than Classic. The reroll discount allows you to see more items, and you shouldn't need to heal quite as frequently as in Classic, so if the shop were discounted, it would probably make sense to discount it about 20% instead of the 75% that we apply here.

bennybroseph:
We'd probably want some alternate unlock condition for this. Also, I should have been more clear, I meant the reward system of gacha tickets that you get for E4, Champion, and Eternatus

src/game-mode.ts
arperry: It ends on 200, but the logic to detect end biome is waveIndex + 9, and we skip 191.

public/battle-anims/psychic.json
arperry: Reverted.

arperry:
> We'd probably want some alternate unlock condition for this.

I'm indifferent to the unlock condition, and I don't have any good ideas for alternate conditions. Do you have any suggestions?

> Also, I should have been more clear, I meant the reward system of gacha tickets that you get for E4, Champion, and Eternatus

This is treated identically to Classic in that regard. If you've beaten a Gym Leader in Classic, you'll get the degraded reward in Turbo, and vice versa.

TempsRay:
I think I'm personally more in favor of a new game mode that's 100 floors long but have the trainer battles rebalanced for that mode, like gym leaders on X0 floors so there's exactly 8, then having the E4 and champ in floors 90-100

Xavion3:
I feel like this could fall under my challenges PR, rather than being another game mode.

Tempo-anon:
We really appreciate the time and effort you've spent in making this new game mode but this is something that can be bundled in Xavion's challenges in the future. You're welcome to submit something as a shorter faster version of Classic mode using that framework then.

arperry:
No worries. Thanks for following up.

PR #1377, opened 2024-05-25
Adds a Small Flyout Panel to the Battle Info Object

bennybroseph:
## What are the changes?
Adds a new Flyout panel to display an enemy Pokemon's moveset and PP usage.

This was drafted up weeks ago and is finally ready for implementation. 
This change allows the user to keep track of a Pokemon's moveset and PP usage even if it switches out. 

This is only step one as there is another panel that will show field effects. You can check progress here #1096 

## What did I change?
Other than the menu itself, the top right text was also moved to the lowest depth within its container. This does not affect the objects visually except when they interact with the flyout. This change allows them to interact properly where this text does not cover the flyout.

This also edits one of the images used by BattleInfo. The change only affect the new Flyout as the image edited fully overlaps the HP graphic anyway and never needed the additional piece which I've removed.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/13838608/514ac958-2396-4d08-aca6-17a19a3a039c

https://github.com/pagefaultgames/pokerogue/assets/13838608/41c56e4b-3bc0-4c09-9e2d-8e6871747b1d

image

## How to test the changes?
The most complex test would be to boot up the game on classic and test this change against a trainer who switches and uses their full movesets. 

## Checklist
- [x] Have I checked that there is no overlap with another PR?
- [x] Have I made sure the PR is self-contained and cannot be split into smaller PRs?
- [x] Are the changes visual?


bennybroseph:
Moved to a separate key:
- V on Keyboard
- Y on Xbox
- Triangle on PS
- V on Mobile

https://github.com/pagefaultgames/pokerogue/assets/13838608/ce645c23-9824-4334-b9a1-da22bf2935d4

bennybroseph:
Setting added

https://github.com/pagefaultgames/pokerogue/assets/13838608/8ae02767-b28b-4391-a5b4-d4a1b98ac50d

molker (approved):

TheSofas:
This seems to have issues when the enemy uses a Leppa Berry to restore points to a move. The flyout does not update with the correct amount of PP, and instead stays at 0. Observed this on a Gurdurr with Dynamic Punch (5PP). After using the Leppa Berry he continued to use Dynamic Punch but the Flyout still showed 0/5. Have not tried it on moves that the Leppa Berry can restore the full +10, instead of only 5 for a max 5PP move.

PR #1381, opened 2024-05-25
Tate Animation was 5 times slower as the one from Liza. So i just shr…

CodeTappert:
…unk down the file for it to have only every fifth frame

Before:

https://github.com/pagefaultgames/pokerogue/assets/38758606/ce9f7415-9fac-4ca4-8d5f-806317429096

After: 

https://github.com/pagefaultgames/pokerogue/assets/38758606/a8ec9c45-4293-4a53-aa35-e937f7e052cf

How did i do it?
With the following golang code:

```go

package main

import (
    "encoding/json"
    "fmt"
    "io/ioutil"
    "os"
)

type Frame struct {
    Filename         string `json:"filename"`
    Rotated          bool   `json:"rotated"`
    Trimmed          bool   `json:"trimmed"`
    SourceSize       Size   `json:"sourceSize"`
    SpriteSourceSize Size   `json:"spriteSourceSize"`
    Frame            Size   `json:"frame"`
}

type Size struct {
    W int `json:"w"`
    H int `json:"h"`
    X int `json:"x,omitempty"`
    Y int `json:"y,omitempty"`
}

type Texture struct {
    Image  string  `json:"image"`
    Format string  `json:"format"`
    Size   Size    `json:"size"`
    Scale  int     `json:"scale"`
    Frames []Frame `json:"frames"`
}

type JSONData struct {
    Textures []Texture `json:"textures"`
    Meta     MetaData  `json:"meta"`
}

type MetaData struct {
    App       string `json:"app"`
    Version   string `json:"version"`
    SmartUpdate string `json:"smartupdate"`
}

func main() {
    // Load the tate.json file
    inputFile := "tate.json"
    outputFile := "tate_reduced.json"
    file, err := ioutil.ReadFile(inputFile)
    if err != nil {
        fmt.Println("Error reading file:", err)
        os.Exit(1)
    }

    // Parse JSON data
    var jsonData JSONData
    err = json.Unmarshal(file, &jsonData)
    if err != nil {
        fmt.Println("Error parsing JSON:", err)
        os.Exit(1)
    }

    // Reduce the number of frames by taking every fifth frame
    reducedFrames := []Frame{}
    for i, frame := range jsonData.Textures[0].Frames {
        if i%5 == 0 {
            reducedFrames = append(reducedFrames, frame)
        }
    }
    jsonData.Textures[0].Frames = reducedFrames

    // Convert back to JSON
    outputJSON, err := json.MarshalIndent(jsonData, "", "    ")
    if err != nil {
        fmt.Println("Error marshaling JSON:", err)
        os.Exit(1)
    }

    // Write the reduced JSON to a new file
    err = ioutil.WriteFile(outputFile, outputJSON, 0644)
    if err != nil {
        fmt.Println("Error writing to file:", err)
        os.Exit(1)
    }

    fmt.Printf("Reduced tate.json saved as %s\n", outputFile)
}

```

### DIFF
--- a/public/images/trainer/tate.json
+++ b/public/images/trainer/tate.json
@@ -1,8084 +1,1637 @@
 {
-	"textures": [
-		{
-			"image": "tate.png",
-			"format": "RGBA8888",
-			"size": {
-				"w": 271,
-				"h": 271
-			},
-			"scale": 1,
-			"frames": [
-				{
-					"filename": "0120.png",
-					"rotated": false,
-					"trimmed": true,
-					"sourceSize": {
-						"w": 80,
-						"h": 80
-					},
-					"spriteSourceSize": {
-						"x": 15,
-						"y": 24,
-						"w": 50,
-						"h": 56
-					},
-					"frame": {
-						"x": 1,
-						"y": 1,
-						"w": 50,
-						"h": 56
-					}
-				},
-				{
-					"filename": "0121.png",
-					"rotated": false,
-					"trimmed": true,
-					"sourceSize": {
-						"w": 80,
-						"h": 80
-					},
-					"spriteSourceSize": {
-						"x": 15,
-						"y": 24,
-						"w": 50,
-						"h": 56
-					},
-					"frame": {
-						"x": 1,
-						"y": 1,
-						"w": 50,
-						"h": 56
-					}
-				},
-				{
-					"filename": "0122.png",
-					"rotated": false,
-					"trimmed": true,
-					"sourceSize": {
-						"w": 80,
-						"h": 80
-					},
-					"spriteSourceSize": {
-						"x": 15,
-						"y": 24,
-						"w": 50,
-						"h": 56
-					},
-					"frame": {
-						"x": 1,
-						"y": 1,
-						"w": 50,
-						"h": 56
-					}
-				},
-				{
-					"filename": "0123.png",
-					"rotated": false,
-					"trimmed": true,
-					"sourceSize": {
-						"w": 80,
-						"h": 80
-					},
-					"spriteSourceSize": {
-						"x": 15,
-						"y": 24,
-						"w": 50,
-						"h": 56
-					},
-					"frame": {
-						"x": 1,
-						"y": 1,
-						"w": 50,
-						"h": 56
-					}
-				},
-				{
-					"filename": "0124.png",
-					"rotated": false,
-					"trimmed": true,
-					"sourceSize": {
-						"w": 80,
-						"h": 80
-					},
-					"spriteSourceSize": {
-						"x": 15,
-						"y": 24,
-						"w": 50,
-						"h": 56
-					},
-					"frame": {
-						"x": 1,
-						"y": 1,
-						"w": 50,
-						"h": 56
-					}
-				},
-				{
-					"filename": "0125.png",
-					"rotated": false,
-					"trimmed": true,
-					"sourceSize": {
-						"w": 80,
-						"h": 80
-					},
-					"spriteSourceSize": {
-						"x": 15,
-						"y": 24,
-						"w": 50,
-						"h": 56
-					},
-					"frame": {
-						"x": 53,
-						"y": 1,
-						"w": 50,
-						"h": 56
-					}
-				},
-				{
-					"filename": "0126.png",
-					"rotated": false,
-					"trimmed": true,
-					"sourceSize": {
-						"w": 80,
-						"h": 80
-					},
-					"spriteSourceSize": {
-						"x": 15,
-						"y": 24,
-						"w": 50,
-						"h": 56
-					},
-					"frame": {
-						"x": 53,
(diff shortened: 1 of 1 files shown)


bennybroseph:
Looks like this deletes animation frames and not duplicate ones. This makes the animation desynced with her twin which seems to be the intention with the animation. Would need to rewrite the script to only delete a frame if it's the same as the previous one or something

CodeTappert:
> Looks like this deletes animation frames and not duplicate ones. This makes the animation desynced with her twin which seems to be the intention with the animation. Would need to rewrite the script to only delete a frame if it's the same as the previous one or something

Yes it deletes all but every fifth. 

Now it is as fast as the one of liza. And way closer to the original games.

See https://youtu.be/cV0l1mAnAA8?si=iBh8ldPSQPiUqwdJ&t=35

If you really think i did something wrong then i can try to delete only duplicate frames tommorow (which still would be the same amount as i do know since the json for tate on main is literally 5 times longer then the one for liza)

CodeTappert:
Would be something like this at the end (this was just with a quick code update that actually compares frames and not stores doubles).


https://github.com/pagefaultgames/pokerogue/assets/38758606/bb9ee694-becd-44a7-910c-5f011a3f405c

bennybroseph:
I think you misunderstand, you'd still delete every fifth frame but it would skip to the next index if it's new. It's very subtle, but it should make it sync better than it does currently. If I'm wrong we still have the old one

CodeTappert:
I think this is it:

https://github.com/pagefaultgames/pokerogue/assets/38758606/625d124b-5694-4cb9-96cf-4114c17d852b

Used code:

```go
package main

import (
    "encoding/json"
    "fmt"
    "io/ioutil"
    "os"
)

type Frame struct {
    Filename         string `json:"filename"`
    Rotated          bool   `json:"rotated"`
    Trimmed          bool   `json:"trimmed"`
    SourceSize       Size   `json:"sourceSize"`
    SpriteSourceSize Size   `json:"spriteSourceSize"`
    Frame            Size   `json:"frame"`
}

type Size struct {
    W int `json:"w"`
    H int `json:"h"`
    X int `json:"x,omitempty"`
    Y int `json:"y,omitempty"`
}

type Texture struct {
    Image  string  `json:"image"`
    Format string  `json:"format"`
    Size   Size    `json:"size"`
    Scale  int     `json:"scale"`
    Frames []Frame `json:"frames"`
}

type JSONData struct {
    Textures []Texture `json:"textures"`
    Meta     MetaData  `json:"meta"`
}

type MetaData struct {
    App        string `json:"app"`
    Version    string `json:"version"`
    SmartUpdate string `json:"smartupdate"`
}

func main() {
    // Load the tate.json file
    inputFile := "tate.json"
    outputFile := "tate_reduced.json"
    file, err := ioutil.ReadFile(inputFile)
    if err != nil {
        fmt.Println("Error reading file:", err)
        os.Exit(1)
    }

    // Parse JSON data
    var jsonData JSONData
    err = json.Unmarshal(file, &jsonData)
    if err != nil {
        fmt.Println("Error parsing JSON:", err)
        os.Exit(1)
    }

    // Reduce the number of frames by taking every fifth frame, restarting for new sequences
    var lastFrame Frame
    reducedFrames := []Frame{}
    count := 0
    for i, frame := range jsonData.Textures[0].Frames {
        if i == 0 || isDifferentSequence(lastFrame, frame) {
            count = 0 // Restart count for new sequence
        }
        if count%5 == 0 {
            reducedFrames = append(reducedFrames, frame)
        }
        count++
        lastFrame = frame
    }
    jsonData.Textures[0].Frames = reducedFrames

    // Convert back to JSON
    outputJSON, err := json.MarshalIndent(jsonData, "", "    ")
    if err != nil {
        fmt.Println("Error marshaling JSON:", err)
        os.Exit(1)
    }

    // Write the reduced JSON to a new file
    err = ioutil.WriteFile(outputFile, outputJSON, 0644)
    if err != nil {
        fmt.Println("Error writing to file:", err)
        os.Exit(1)
    }

    fmt.Printf("Reduced tate.json saved as %s\n", outputFile)
}

// isDifferentSequence checks if the current frame is part of a new sequence compared to the last frame
func isDifferentSequence(lastFrame, currentFrame Frame) bool {
    return lastFrame.Frame.X != currentFrame.Frame.X ||
        lastFrame.Frame.Y != currentFrame.Frame.Y ||
        lastFrame.Frame.W != currentFrame.Frame.W ||
        lastFrame.Frame.H != currentFrame.Frame.H
}

```

CodeTappert:
I dont know what else to do. SO either we use that in the original post or what i posted last

bennybroseph:
Hmm yeah let's use the new one. Very subtle differences. Feels like neither are correct, but Liza's animation seems weird too and we didn't change those. Wouldn't be surprised if all the animations are slightly messed up and that's why it feels off.

bennybroseph (approved):

PR #1382, opened 2024-05-25
Added Knock Off and Incinerate functionality

MPekelny:
## What are the changes?
Adding the move specific functionality to Knock Off and Incinerate.

These are two moves that currently do not have functionality, and they are ones where the extra functionality is relatively simple, so it is a good first thing for me to do to dip my toes into contributing.

## What did I change?
There was an existing RemoveHeldItemAttribute which seems to be what was intended to do the knock off effect, but it was stealing items. So I updated it to instead only remove an item, adding a tryRemoveHeldItemModifier that works similar to the transfer version. I did also see the current pull request for Incinerate functionality, but I think rather than having a specific attribute for berries, it is better to update the existing property to handle both cases.

**Screenshots/Videos**

https://github.com/pagefaultgames/pokerogue/assets/73674424/74b03cdd-2b7c-40c8-8b9a-d993fdd85677

## How to test the changes?
Add to overrides.ts to give your Pokemon Knock Off and Incinerate and the opponent some held item and try them in battle. I also changed the starting wave and gave the opponent only non-damaging moves so it can take several hits and not damage back to make testing easier.


src/data/move.ts
```diff
@@ -1487,29 +1487,50 @@ export class StealHeldItemChanceAttr extends MoveEffectAttr {
   }
 }
 
+/**
+ * An attribute that will cause an item held by the target do be dropped on hit.
+ */
 export class RemoveHeldItemAttr extends MoveEffectAttr {
   private chance: number;
+  private effectText: string;
+  private onlyBerries: boolean;
```
MPekelny, line 1496: It may be that something other than a berry/not berry boolean would work better, but I don't think there are any moves that remove anything other than all items or only berries, so something more special seems unnecessary.

MPekelny:
I realized I made a mistake and am pull requesting this from my main, rather than a branch. I am going to close this and then open a new one that is properly pull requesting from a branch.

PR #1384, opened 2024-05-25
Bugfix: Added check for move changing type before determining if defender is immune to it because of an ability

td76099:
## What are the changes?
The main change is checking if a move has the `VariableMoveTypeAttr` attached to it - if it does then the type being checked for immunity by `TypeImmunityAbAttr` is the move's updated type instead of default type. This change allows moves like Ivy Cudgel, Hidden Power, Terrain Blast, Weather Ball, etc. to all be checked against their variable type instead of default type.

ex: Fire Mask (idk name) Ogerpon using Ivy Cudgel vs a Flash Fire Pokemon should trigger Flash Fire because Ivy Cudgel is a Fire type move with the Fire Mask. However, it was not triggering it because the ability check for immunities was only checking the default type (grass)

It's a bug - unintended behavior.

## What did change?
In `TypeImmunityAbAttr`, there is now an extra variable called `moveType` which is defaulted to the move's default type. Afterwards, it tries to look up if the move has the `VariableMoveTypeAttr` and assign the result of that move's new type to `moveType`. That variable is then checked against `immuneType` just like how `move.getMove().type` was before.

### Before changes
Water Mask Ivy Cudgel vs Water Absorb - incorrect behavior:

https://github.com/pagefaultgames/pokerogue/assets/85713900/7f3ef205-a164-459a-a716-40c1a53e673f

### After changes
Water Mask Ivy Cudgel and Weather Ball vs Water Absorb:

https://github.com/pagefaultgames/pokerogue/assets/85713900/0725b1c4-3fbf-45b2-bfec-5799148ceada

Fire Mask Ivy Cudgel and Sunny Weather Ball vs Flash Fire:

https://github.com/pagefaultgames/pokerogue/assets/85713900/ee66e3f8-0997-4ffc-b426-9183d30d5973

Rock Mask Ivy Cudgel vs Sap Sipper:

https://github.com/pagefaultgames/pokerogue/assets/85713900/c00d4534-0bb0-4bb2-aca9-1e70d77565d6

Grass Mask Ivy Cudgel vs Sap Sipper:

https://github.com/pagefaultgames/pokerogue/assets/85713900/dee79d51-51d4-4003-8ff0-165582a9eaec

Default Mask Ivy Cudgel vs Sap Sipper:

https://github.com/pagefaultgames/pokerogue/assets/85713900/3efba2cd-21cb-4189-b9f9-5f7e1560ac2b

Grassy Terrain Terrain Pulse (and electric terrain) vs Sap Sipper:

https://github.com/pagefaultgames/pokerogue/assets/85713900/3cb418be-5141-4930-82d4-94cf7c61eb98

Electric Terrain Terrain Pulse (and grassy terrain) vs Volt Absorb:

https://github.com/pagefaultgames/pokerogue/assets/85713900/be411a6a-4b49-49ee-88ed-76074cc30453

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -9,7 +9,7 @@ import { BattlerTag } from "./battler-tags";
 import { BattlerTagType } from "./enums/battler-tag-type";
 import { StatusEffect, getStatusEffectDescriptor, getStatusEffectHealText } from "./status-effect";
 import { Gender } from "./gender";
-import Move, { AttackMove, MoveCategory, MoveFlags, MoveTarget, StatusMoveTypeImmunityAttr, FlinchAttr, OneHitKOAttr, HitHealAttr, StrengthSapHealAttr, allMoves, StatusMove, SelfStatusMove, VariablePowerAttr, applyMoveAttrs, IncrementMovePriorityAttr  } from "./move";
+import Move, { AttackMove, MoveCategory, MoveFlags, MoveTarget, StatusMoveTypeImmunityAttr, VariableMoveTypeAttr, FlinchAttr, OneHitKOAttr, HitHealAttr, StrengthSapHealAttr, allMoves, StatusMove, SelfStatusMove, VariablePowerAttr, applyMoveAttrs, IncrementMovePriorityAttr  } from "./move";
 import { ArenaTagSide, ArenaTrapTag } from "./arena-tag";
 import { ArenaTagType } from "./enums/arena-tag-type";
 import { Stat } from "./pokemon-stat";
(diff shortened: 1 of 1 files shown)


RNGezeus:
How about moves that change typing due to abilities?

td76099:
> How about moves that change typing due to abilities?

Did not specifically test those but considering those are because of an ability and not a move attribute I would bet those still don't get checked properly.

Upos1997:
tested the abilities, ability changing the type of the move is not taken into account.

Both the move type changing and ability type changing attributes are applied in pokemon/apply and the type is stored in variableType.
Would you be able to get this variableType to the TypeImmunityAbAttr?
That way both are checked at once, and we're not checking the same conditions twice.

td76099:
I'd have to look into it more to see where everything is getting called but that might be for the best

At the very least I could create a function that could get reused in both places so they use the same logic for checking type. Was actually thinking about that this morning because I cooked something up for checking move type because of ability modifiers and thought it would be nice to make sure they are both using the same logic

bennybroseph:
This definitely works, but it's undermining the whole attribute system to do so. A moves's type changing attribute clearly needs to be applied before determining if a pokemon is immune to it. This needs to be looked at in phase.ts where all of this happens, not in a single ability attribute. This will cause problems later even if it fixes a small problem now.

PR #1386, opened 2024-05-26
Update French biome.ts

Adri1:
After a debate with Sangara on Discord and a pool in  the French channel, we came up with these edits

### DIFF
--- a/src/locales/fr/biome.ts
+++ b/src/locales/fr/biome.ts
@@ -1,7 +1,7 @@
 import { SimpleTranslationEntries } from "#app/plugins/i18n";
 
 export const biome: SimpleTranslationEntries = {
-  "unknownLocation": "Dans un endroit inconnu",
+  "unknownLocation": "vous avez oublié où",
   "TOWN": "Ville",
   "PLAINS": "Plaines",
   "GRASS": "Herbes",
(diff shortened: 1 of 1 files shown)


src/locales/fr/biome.ts
```diff
@@ -1,7 +1,7 @@
 import { SimpleTranslationEntries } from "#app/plugins/i18n";
 
 export const biome: SimpleTranslationEntries = {
-  "unknownLocation": "Dans un endroit inconnu",
+  "unknownLocation": "vous avez oublié où",
```
XGSleepWalker, line 4: Shouldn't this be Vous avez oublié où? Capitalized V?

XGSleepWalker (dismissed):

src/locales/fr/biome.ts
Adri1: Unless if proven otherwise, this is only used in trainer memo as a complition of the sentence after a coma and it's not the name of a place.
So it doesn't have to be capitalized we took took rid of it.

Example from French Pokémon Emerald : 
echange grainipiot5
"obtenu dans un échange" (obtained from a trade) here follows the same logic and is not capitalized either

bennybroseph (approved):

PR #1391, opened 2024-05-26
Update inputs-controller.ts

tsutsurana:
Updating Button.Action in inputs-controller.ts in order to Add Y as a viable input method to confirm menus.

### DIFF
--- a/src/inputs-controller.ts
+++ b/src/inputs-controller.ts
@@ -342,7 +342,7 @@ export class InputsController {
       [Button.LEFT]: [keyCodes.LEFT, keyCodes.A],
       [Button.RIGHT]: [keyCodes.RIGHT, keyCodes.D],
       [Button.SUBMIT]: [keyCodes.ENTER],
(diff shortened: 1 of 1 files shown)


CodeTappert:
Yes Please. It is so annoying that it only works for Z atm

bennybroseph:
Are there other common layouts that we don't support properly?

tsutsurana:
QWERTZ is the main one that comes to mind for me personally, but basically any Layout that doesn't have Z besides X is either a reach or requires remapping / changing the layout to play with keyboard.
Optimally there should be a remapping system to map the actions to whatever a user wants, but I currently lack the knowledge to implement that.

CodeTappert:
Dont the french have something completly different?

bennybroseph:
Yeah, unfortunately that opens up a weird can of worms. AZERTY has W where z normally is and that obviously cannot be mapped here. I think we should just wait until our keybind remapping PR is finished instead of making an exception here.

PR #1393, opened 2024-05-26
Implement Treasures of Ruin abilities

innerthunder:
I've created a new set of arena tags to implement the abilities of the Treasures of Ruin (namely Vessel of Ruin, Sword of Ruin, Tablets of Ruin, and Beads of Ruin).

The basic idea of the implementation:

- When a Pokemon with a Treasure of Ruin ability is summoned, it will create a `TreasureOfRuinTag` with the corresponding `Stat` type to weaken.
- While the `TreasureOfRuinTag` is active, it will multiply its corresponding stat by 0.75 during battle stat calculation in `Pokemon.getBattleStat()`
- When the source Pokemon leaves battle, the `TreasureOfRuinTag` is removed without any notification.

I've written the required arena tags and modified `getBattleStat()`, so all I need to do now is incorporate the tag addition and removal in `data/ability`.

Also, a note about implementation: As far as I know, this is the first ability to use an arena tag in its implementation. The `ArenaTag` class requires a `sourceMove` field, which I've left as `Moves.NONE` for `TreasureOfRuinTag`. It might be a good idea to generalize the `ArenaTag` class in the future to include support for source abilities.


src/data/arena-tag.ts
```diff
@@ -269,6 +271,83 @@ class CraftyShieldTag extends ConditionalProtectTag {
...
+   * @param args[0] (Pokemon)            The Pokemon to which this tag is applied
+   * @param args[1] (Stat)               The type of the checked Stat value
+   * @param args[2] (Utils.NumberHolder) A container for the checked Stat value
+   * @returns true if the flag modifies the Stat value, false otherwise.
+   */
+  apply(arena: Arena, args: any[]): boolean {
+    const target = args[0] as Pokemon;
+    const stat = args[1] as Stat;
+
+    if (target.id !== this.sourceId && this.stat === stat) {
```
TempsRay, line 319: The stat penalty should apply to all pokemon that do not have the corresponding ruin ability

TempsRay:
The hard part about implementing this is removing the arena tag properly. There's quite a few scenarios (a pokemon fainting, a pokemon switching out, or when a move is used that either removes the pokemon or the ability). Even then, the arena tag should persist so long as at least one pokemon with the corresponding ability is still on the field.

innerthunder:
## Major Updates
- `data/arena-tag`: Added a new generic class `AbArenaTag` for arena tags generated from abilities. Some of the functionality from `TreasureOfRuinTag` was pulled to here, as well as additional checks for ability bypass.
- `data/ability`: Added a new ability attribute `PostSummonAddArenaTagAbAttr` that adds an arena tag on summon
- `field/arena`: Added a new function `Arena.refreshAbTags`, which removes all ability arena tags such that there is no Pokemon on the field that currently has the tag's source ability.

### Where is `Arena.refreshAbTags` called?
- `phases`: In `SwitchSummonPhase.onEnd` after a switch phase has resolved.
- `phases`: In `FaintPhase.doFaint` after a fainted Pokemon is removed from the field.
- `data/move`:  In `AbChangeAttr`, accounting for Worry Seed, Entrainment, etc.
- `data/move`: In `SuppressAbilitiesAttr`, accounting for Gastro Acid and Core Enforcer.
- `data/move`: In `TransformAttr`, accounting for Transform.
- `data/ability`: In `PostDefendAbilityGiveAbAttr`, accounting for Mummy and Lingering Aroma.

## Test Videos

### Sword of Ruin vs. Mummy

- The battle starts, and Chien-Pao's ability is shown.
- Chien-Pao attacks Minccino with Quick Attack, converting Chien-Pao's ability to Mummy.

https://github.com/pagefaultgames/pokerogue/assets/168692175/74f55a0a-0bc0-4839-a991-3f681e12d5bd

A closer look at the logs:

Screenshot 2024-05-27 171701

At the beginning of the encounter, the logs show that Minccino has 7 Defense. In the damage calculation for Chien-Pao's Quick Attack, the Defense value passed to `Pokemon.damage` is shown as 5. The final message in the picture, "Removed 285's effects," is a console log I added for testing within `AbArenaTag.onRemove()`; 285 is the ID for Sword of Ruin.

After the recording, I had Chien-Pao use Quick Attack on Minccino again:

Screenshot 2024-05-27 172837

Here, Minccino's Defense value when passed to `Pokemon.damage` is once again shown as 7, meaning the effects of Sword of Ruin have successfully been removed after Chien-Pao's ability changed.

### Sword of Ruin when switched out

- Chien-Pao switches out, while Charmander uses Quick Attack on Minccino.
- A closer look at the logs shows that Sword of Ruin's effects were removed before Charmander attacks.

https://github.com/pagefaultgames/pokerogue/assets/168692175/af852740-c376-482e-91f0-3cfa0ae80f6a


Screenshot 2024-05-27 180352

damocleas:
A (somewhat) niche interaction that can occur is two of the same ruin abilities fighting eachother, they cancel it out vs eachother. ex: Chien-Pao vs Chien-Pao, neither have the defense drop vs eachother, but in a double battle, the two non-Chien-Pao on the field will have the defense drop.
Would need to label as (P) if not planning on trying to get this interaction to work, but seems fine.

innerthunder:
> A (somewhat) niche interaction that can occur is two of the same ruin abilities fighting eachother, they cancel it out vs eachother. ex: Chien-Pao vs Chien-Pao, neither have the defense drop vs eachother, but in a double battle, the two non-Chien-Pao on the field will have the defense drop. Would need to label as (P) if not planning on trying to get this interaction to work, but seems fine.

```
This interaction should be covered already. `Arena.addTag()` only allows for one tag of a given type to be on the field at a time, and the `TreasureOfRuinTag` only applies its stat debuff to Pokemon who don't have its source ability.

**Comment by innerthunder** [comment]
## Fixes for Neutralizing Gas

- `data/ability`: The ruinous abilities now bypass suppression. The suppression of these abilities is instead carried out in their respective arena tags.
- `data/arena-tag`: `AbArenaTag` now has a `isSuppressed` function that checks if any active Pokemon on the field has a suppressing ability (i.e. Neutralizing Gas). `apply` now uses this to help check if the tag's effect can be applied.

### Test Video: Sword of Ruin vs. Neutralizing Gas

- The battle starts with Sword of Ruin Chien-Pao vs. Neutralizing Gas Minccino.
- Chien-Pao uses Quick Attack against Minccino, dealing damage. According to the logs, Minccino's Defense is read as 7 both when it is initialized and when it is hit by Quick Attack.

https://github.com/pagefaultgames/pokerogue/assets/168692175/60ca4254-f55d-41f3-9564-b23c700b4d0c

Screenshot 2024-05-27 210217

File: src/data/ability.ts [hunk 2]
```diff
@@ -1483,6 +1484,21 @@ export class PostSummonAddBattlerTagAbAttr extends PostSummonAbAttr {
   }
 }
 
+export class PostSummonAddArenaTagAbAttr extends PostSummonAbAttr {
```
Tempo-anon, line 1487: New classes need documentation

src/data/arena-tag.ts
```diff
@@ -632,6 +634,137 @@ class TailwindTag extends ArenaTag {
   }
 }
 
+export abstract class AbArenaTag extends ArenaTag {
```
Tempo-anon, line 637: Need documentation here too

innerthunder:
Closing this PR. A cleaner version of this feature using ability attributes is available at https://github.com/pagefaultgames/pokerogue/pull/1637.

PR #1394, opened 2024-05-26
Allow selection of item quantity when transferring items (fixes #723)

dielle000:
The content is the same as pull request #723, except this time I haven't messed up the commits.

The UI is changed as suggested:
screenshot


molker:
I like the addition to the text for changing quantity, can you make it so if the player scrolls away from the item they have changed the quantity for that the number goes back to the current amount?

dielle000:
Updated as suggested:

https://github.com/pagefaultgames/pokerogue/assets/168655502/b42a500e-c1e5-4a71-9805-e6e7811a371a

bennybroseph (changes requested):
Last thing, need some comments in and around this change. Lots of code, not seeing any docs

dielle000:
Documentation added

bennybroseph (approved):

PR #1412, opened 2024-05-26
Added Prisoner trainer class

Alekemon:
4th of the 4 trainer classes, uses Steel and Dark type Pokémon.

Credits for the sprite go for King_Waluigi


CodeTappert:
This trainer class does not seem to exist in the games?

Alekemon:
Closed because the trainer class isn't official

PR #1418, opened 2024-05-27
Functionality for Poison Heal

newmattock:
Added the healing mechanism for the Poison Heal ability and prevented Pokemon with this ability from taking poison damage.


src/phases.ts
```diff
@@ -3281,21 +3281,32 @@ export class PostTurnStatusEffectPhase extends PokemonPhase {
       applyAbAttrs(BlockNonDirectDamageAbAttr, pokemon, cancelled);
 
       if (!cancelled.value) {
-        this.scene.queueMessage(getPokemonMessage(pokemon, getStatusEffectActivationText(pokemon.status.effect)));
         let damage: integer = 0;
         switch (pokemon.status.effect) {
         case StatusEffect.POISON:
-          damage = Math.max(pokemon.getMaxHp() >> 3, 1);
+          if (pokemon.hasAbility(90)){
```
ReneGV, line 3287:
```suggestion
          if (pokemon.hasAbility(Abilities.POISON_HEAL)){
```

src/data/ability.ts
```diff
@@ -3503,7 +3503,8 @@ export function initAbilities() {
     new Ability(Abilities.IRON_FIST, 4)
       .attr(MovePowerBoostAbAttr, (user, target, move) => move.hasFlag(MoveFlags.PUNCHING_MOVE), 1.2),
     new Ability(Abilities.POISON_HEAL, 4)
-      .unimplemented(),
+      .conditionalAttr(pokemon => !!pokemon.status && pokemon.status.effect === StatusEffect.POISON, PostTurnHealAbAttr)
```
ReneGV, line 3506:
```suggestion
      .conditionalAttr(
          (user: Pokemon, target:Pokemon, move: Move) => pokemon.status?.effect === StatusEffect.POISON,
          PostTurnHealAbAttr
      )
```

ReneGV: I am not sure if you should check for StatusEffect.TOXIC

ReneGV: Also I believe PostTurnHealAbAttr restores 1/8 HP according to [bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Poison_Heal_(Ability)#:~:text=Restores%20HP%20if%20the%20Pok%C3%A9mon%20is%20poisoned%20instead%20of%20losing%20HP.&text=If%20poisoned%2C%20the%20Pok%C3%A9mon%20has,Ability%20of%20Shroomish%20and%20Breloom.)

While I think the PostTurnHealAbAttr heals 1/16

src/phases.ts
ReneGV: However a better solution would be something similar to line `3281`:
```
applyAbAttrs(BlockNonDirectDamageAbAttr, pokemon, cancelled);
```
You can create an attribute called `BlockPoisonDammageAttr`
and apply it under line 3281

ReneGV (commented):
It looks good, however I would do some minor adjustments.

bennybroseph:
This has been attempted several times before. What's different about this one? Can we get a video showing how this was tested?

Issue #1287, opened 2024-05-23
Triple axel


Diamnd08:
Triple axel on arctovish egg move , when used on flying types does not show as super effective

Tempo-anon:
Should be fixed with #2981

PR #783, opened 2024-05-12
Critical Hits, Super effective, multi-hit moves dialogue

DustinLin:
### Fixes issue #680 
---- original behavior on `main`, on faint doesn't display effectiveness dialogue, or multi hit result:

https://github.com/pagefaultgames/pokerogue/assets/39450497/919527d3-7a3a-4ea4-8c84-7f17b746be67

---- PR fix: critical hits, effectiveness, and multi-hit results (even if ended early due to faint) how properly show when fainting

https://github.com/pagefaultgames/pokerogue/assets/39450497/f392e989-6380-4e34-9b4f-2d68631979eb

---- some more bugs to do?

For moves like absorb that have subsequent dialogue, it runs into the same issue of not rendering on faint, but it was not fixed with this PR, perhaps this can be done in a separate issue, needs more investigating. More generally the `end()` function of the `MoveEffectPhase` class in `phases.ts` doesn't queue message properly on faint, it seems like those messages are lose

https://github.com/pagefaultgames/pokerogue/assets/39450497/0d346fdd-dc1a-40a4-8a67-5fa3563340b4

PR #1420, opened 2024-05-27
Fix: display every hit result of multi-hit moves

SnowCharmQ:
## What are the changes:

- Remove the condition that judges whether to display _HitResult_ information.

## Why am I making these changes:

- This PR aims to address #1287. After changing, the hit result will be shown even if the opponent's Pokemon faints before the final hit of the multi-hit move.
- Follow the official versions, every hit result for multi-hit moves needs to be displayed correctly and respectively.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,7 @@
 				"vitest-canvas-mock": "^0.3.3"
 			},
 			"engines": {
-				"node": ">=18.0.0"
+				"node": ">=20.0.0"
 			}
 		},
 		"node_modules/@ampproject/remapping": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,7 @@
 				"vitest-canvas-mock": "^0.3.3"
 			},
 			"engines": {
-				"node": ">=20.0.0"
+				"node": ">=18.0.0"
 			}
 		},
 		"node_modules/@ampproject/remapping": {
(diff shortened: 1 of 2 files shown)


torranx:
I think this is already being worked on, #783

SnowCharmQ:
> I think this is already being worked on, #783

In #783, the condition is changed to `if (source.turnData.hitsLeft === 1 || this.isFainted())`. In fact, simply removing the condition can solve this problem.

torranx:
> > I think this is already being worked on, #783
> 
> In #783, the condition is changed to `if (source.turnData.hitsLeft === 1 || this.isFainted())`. In fact, simply removing the condition can solve this problem.

have you tested other scenarios? pretty sure theres a reason why that condition is present

SnowCharmQ:
> > > I think this is already being worked on, #783
> > 
> > 
> > In #783, the condition is changed to `if (source.turnData.hitsLeft === 1 || this.isFainted())`. In fact, simply removing the condition can solve this problem.
> 
> have you tested other scenarios? pretty sure theres a reason why that condition is present

The condition in #783 just displays information when multi-hit moves have finished all hits or the opponent has fainted. In this case, the information can be displayed only once per move. But in the official Pokemon versions, every hit result information will be displayed correctly and respectively. (indeed I do not test all versions) But it is better to follow the official version.

torranx:
I meant have you tested your code change not other pokemon versions...

Tempo-anon:
I believe https://github.com/pagefaultgames/pokerogue/pull/2857 should address this so I will close this issue out for now

PR #1431, opened 2024-05-27
Added "Skip Seen Dialogues" option

FredeX:
## Why?
This seems to get requested a lot in the #feature-vote discord channel. It's more of a QoL for grinding players who basically click through every dialogue anyway and play at 5x speed. The dialogues can get annoying for some people especially when you F5/retry a lot at e.g. lvl195. The option only allows to skip **seen** dialogues! The dialogue key (e.g. `PGFdialogue:rival.encounter.1`) is used to mark a dialogue as seen, so the option is independent of the text of the dialogue as well as the language.

image

## Changes
- added skip seen dialogues option
- updated ui.showDialogue to skip a dialogues if it has been seen and when the option is active.
- updated game-data.ts to save/load the key of a seen dialogue to/from local storage
- added `shouldSkipDialogue(key:string): boolean` as a helper function
- only show character sprite when **not** skipping dialogue

image

## Videos & Screenshots

Here are 5 different working scenarios. Most of them showcase the first time a dialogue is seen and then the second time (this one should be skipped). The last one also shows the dialogue key working in action: changing the language doesn't effect the "has seen"-state.

**Note: This is on 5x Speed!!!**

**Scenario 1: Rival Stage 8 - Start Of Battle:**

https://github.com/pagefaultgames/pokerogue/assets/1838682/bd0ca171-d7d1-4fd8-bab1-0121fcca7824

**Scenario 2: Rival Stage 8 - End Of Battle:**

https://github.com/pagefaultgames/pokerogue/assets/1838682/8a190c11-c2b3-47d6-963d-407f9d9a5f72

**Scenario 3: Rival Stage 200 - Winning Classic Mode:**

https://github.com/pagefaultgames/pokerogue/assets/1838682/86c1ab71-942a-4e17-bf72-d63babbde4b9

**Scenario 4: Trainer Stage 5 - Start Of Battle + End Of Battle:**

https://github.com/pagefaultgames/pokerogue/assets/1838682/48aaf5b3-897a-4d68-b384-afbfa619fa66

**Scenario 5: Changing ingame language:**

https://github.com/pagefaultgames/pokerogue/assets/1838682/5fddfa37-276e-4fb6-86eb-25f5d20adef5

## How to test the changes?
Go to the game settings, find the option, activate it and go to any dialogue (you can manipulate your starting wave by changing `STARTING_WAVE_OVERRIDE` in `overrides.ts` to e.g. 5 (trainer encounter), 8 first rival encounter, 195 (the longest rival dialogue) or 201 to see the classic win dialogue). After the dialogue you can reload and check the same dialogue again. It should now get skipped. Reactivate the option and it will reappear. To check a end of battle dialogue just set `export const OPP_LEVEL_OVERRIDE: number = 1;` and `export const STARTING_LEVEL_OVERRIDE: integer = 100;` to make you OP and quickly end the fight.

## Checklist
- [x] Are the changes visual?


src/ui/settings-ui-handler.ts
```diff
@@ -220,6 +220,14 @@ export default class SettingsUiHandler extends UiHandler {
       return false;
     }
 
+    // Don't allow skipping dialogues if the player has not won a session yet
```
torranx, line 223: Instead of disallowing player and playing a sound, why not just hide the menu to avoid confusion- "why can't I turn this on?"

FredeX: Yeah, absolutly makes sense to me. Committed an updated version with no error sound and instead hiding the option when there are no classic wins yet. Way better. Thanks for your review

torranx: **This is a nice-to-have suggestion only and may not necessarily included in your PR:**
After the first classic win, it would be nice to queue a message of something like "You unlocked a new option in game settings to skip dialogues!"

FredeX: Good idea. I opened an issue for this so I can implement it the next days since im busy today https://github.com/pagefaultgames/pokerogue/issues/1442

FredeX: Hey, i just added the skip dialogues option to the Unlockables. So there is a unlock message like "Endless Mode" has been unlocked to let the user know the option is now available in the game settings :)

image

torranx (approved):
lgtm - did not test

flx-sta:
Maybe the user should be informed about unlocking this option after finishing classic for the first time ?

FredeX:
Maybe my descriptions wasnt clear enough: I added the option to the unlockables, so there is in fact a message that you unlocked the option :) You can see it in the thrid screenshot (the one of the victory screen). I will edit the description to make that more clear. Thank you

flx-sta:
> > Maybe the trainer should be informed about unlocking this option after finishing classic for the first time ?
> > “YOU can now use the `Skip Dialogs` option in the Game Settings”
> > something like that
> 
> Maybe my description wasnt clear enough: I added the option to the unlockables, so there is in fact a message that you unlocked the option :) You can see it in the thrid screenshot (the one of the victory screen). I will edit the description to make that more clear. Thank you

ah! perfect! Lazy me, sorry

flx-sta (approved):
Works as intended for me (yey, my first classic win)

FredeX:
(no changes, i only synched the PR to avoid any merge conflicts)

flx-sta (approved):
@FredeX there is no easy way to override the value for `gamesWon` right?
I've just disabled the if-checks in my code
It still works though
LGTM 🚀 

--------

### Video proof

https://github.com/pagefaultgames/pokerogue/assets/50131232/fc639b47-80c3-40d6-b126-f6ce28d91f06

OrangeRed:
Won't putting that setting at the bottom of `enum Setting` put it at the bottom of the settings in game? Is that a desirable / logical place for that setting to exist?

flx-sta:
> Won't putting that setting at the bottom of `enum Setting` put it at the bottom of the settings in game? Is that a desirable / logical place for that setting to exist?
> 
> I would prefer if the setting were next to Tutorials / Enable Retries. That feels like a more logical place for that setting to be located

I assume like this?
  Skip_Dialogues = "SKIP_DIALOGUES"

OrangeRed:
> > Won't putting that setting at the bottom of `enum Setting` put it at the bottom of the settings in game? Is that a desirable / logical place for that setting to exist?
> > I would prefer if the setting were next to Tutorials / Enable Retries. That feels like a more logical place for that setting to be located
> 
> # Like this?
> image

Yeah something like that but that is just my preference. Ask other people what they think about it.

bennybroseph:
After discussion, we'd like this feature to only skip **seen** dialogue. I can understand someone not wanting to see the same thing twice, but at least viewing it once seems perfectly reasonable.

FredeX:
Okay, I thought unlocking the option after winning for the first time would be sufficient, but yeah i feel you. New trainer encounters would be skipped as well. 

Would you be fine with storing the information about which dialogue was seen in the **SystemSaveData** or only localStorage?

FredeX:
and after these changes, should the option still need to be unlocked first?

flx-sta:
@FredeX how about a wrapper that easily allows to change to any storage system with minimal effort (I’d put it in local storage, just like tutorials)

FredeX:
Yeah, sounds good. One problem with that wrapper would be that local storage saves instantly when seen but systemSaveData only AFTER the fight. But many people F5 often in e.g. stage 195 even before retries would kick. Then you would always have to click through the rivals dialogue even when you did see it before.

I do have a workaround for this by combining those two but to be honest i think that's an absolute overkill. 

I think it's best, like you said, to make it similiar to tutorials, which are very comparable in a way.

FredeX:
Okay i pushed a new version:

### What did I change?
- removed the option from the Unlockables. No need to be unlocked first when you skip only seen dialogues. WDYT?
- positioned the option under "Enabled Retries" as @OrangeRed suggested
- saves seen dialogues in local storage under the key 'seenDialogues' `{ [key: string]: boolean }`. As a key i take the localization key (e.g.  `PGFdialogue:rival.encounter.1`). This key is perfect for this and i don't have to watch out for splitted dialogues BUT this will not work for unlocalized dialogues. 
- only allow skipping **seen** dialogues.

bennybroseph:
> * BUT this will not work for unlocalized dialogues.

Could you expand on what this means?

FredeX:
I don't know how your guidelines are regarding new dialogues added to pokerogue but let's say theres a new hardcoded dialogue  added to the game which is not contained in any locale. Since I'm using the key (e.g. `PGFdialogue:rival.encounter.1`) to mark the dialogue as seen, this new dialogue has no such key to easily identify it. I could just store the content/text of each dialogue to mark them as seen but then if you decide to update a dialogue or fix a typo or the user changes the ingamelanguage, the dialogue would show up again because the contents are different. Also, i am relucatant to string compare a long text rather than a key which is basically a ID.

Here is an example: It's much easier and convenient to check for "ending_female" rather than the whole dialogue. And it's independent of possible changes to the dialogue

image

That's why this universal key is perfect for this. So only a hardcoded text would not be marked as seen. (It's not important if the dialogue is localized in the language of the user yet)

flx-sta:
Would that mean, that if I change my language I have to look at the dialoge in the new language again?

FredeX:
No, the key is independent of the contents of lanuage.

> as far as I understand, using the universal i18n key covers that to NOT happen

Correct!

bennybroseph:
Okay, this should be fine then. We should be using localized dialogue where possible, so if this isn't working it's not this feature's fault.

FredeX:
Okay nice, I will update the description tomorrow accordingly and maybe then you can turn the draft back into a PR

FredeX:
Okay I updated the description. Time to leave draft? :)

flx-sta (approved):
Tested with level 8 rival dialog.
Turned the setting on and off and on and off ...
worked as expected

FredeX:
Added Videos & Screenshot to showcase the behaviour with the stage 8 rival

FredeX:
Updated the PR:
- don't show the charachter sprite when the dialogue is skipped
- small refactoring
- `shouldSkipDialogue(key: string): boolean` helper function added

Also changed the videos provided in the description of the PR to showcase mulitple working scenarios

XGSleepWalker (approved):

PR #918, opened 2024-05-15
Localization: Added biomes, summary-ui-handler, party-ui-handler and summary images

rnicar245:

PR #1436, opened 2024-05-27
zhTW Type Images are now correctly loaded.

CodeTappert:
Test with giving it the german images (for the test) to make sure it doesnt load the zn_CN images instead.

image

Of course it uses the english images for now

### DIFF
--- a/src/loading-scene.ts
+++ b/src/loading-scene.ts
@@ -216,6 +216,10 @@ export class LoadingScene extends SceneBase {
         break;
       }
       if (Utils.verifyLang(lang)) {
+        if (lang === "zh_CN") {
+          // Load also the traditional Chinese since it doesn't have a separate language code in supportedLngs
+          this.loadAtlas("types_zh_TW", "");
+        }
         this.loadAtlas(`types_${lang}`, "");
       }
     });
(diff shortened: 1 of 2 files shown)


bennybroseph (approved):

src/loading-scene.ts
```diff
@@ -216,6 +216,10 @@ export class LoadingScene extends SceneBase {
         break;
       }
       if (Utils.verifyLang(lang)) {
+        if (lang === "zh_CN") {
+          // Load also the traditional Chinese since it doesn't have a separate language code in supportedLngs
+          this.loadAtlas("types_zh_TW", "");
+        }
```
bennybroseph, line 222: Does this make every language load the Traditional Chinese ones in the background? If so, we gotta find another solution

bennybroseph (changes requested):

src/loading-scene.ts
CodeTappert: As far as i am understanding the code (most is from #918 anyways.

It loads every language types in every language anyways. 

Since there is a foreach. So if you select english it also loads french etc.


Yes we should find a better solution. But we need to make sure that changing the language ingame also reloads that images.

bennybroseph: I'll accept this PR, but we really should do something about loading all languages. That feels really wrong.

bennybroseph (approved):

PR #1438, opened 2024-05-27
When the language is changed the game will complety reload (same as F5).

CodeTappert:
Also now only the current language type image is loaded. If it doesnt exist for a language it will load the english one as a fallback

Tested with all lanugages

Before: (Does not reload titles, trainer classes and names correctly. Loads all type images always)

https://github.com/pagefaultgames/pokerogue/assets/38758606/e17cb595-ad45-4e8a-8088-d5cc22ac72cb

After: (Reloads the whole game so titles, trainer classes and names are correctly localized. Only loads the current type image. (If it doesnt exist it loads the english one)

https://github.com/pagefaultgames/pokerogue/assets/38758606/41bd542b-5137-4216-94ed-65dc1111595b

### DIFF
--- a/src/loading-scene.ts
+++ b/src/loading-scene.ts
@@ -206,23 +206,19 @@ export class LoadingScene extends SceneBase {
     this.loadAtlas("items", "");
     this.loadAtlas("types", "");
 
-    (i18next.options.supportedLngs as Array<any>).forEach(lang => {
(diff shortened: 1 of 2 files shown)


CodeTappert:
@bennybroseph I implemented what we have talked about

bennybroseph (approved):

PR #1439, opened 2024-05-27
Implement Steely Spirit

hayuna:
Description:
The power of Steel-type moves used by a Pokémon with Steely Spirit and its allies is increased by 50%. This effect stacks if multiple Pokémon with this Ability are on the field.

Scenario 1: Both pokemon have ability (attack should be multiply by 2.25
Scenario 2: Attacker has ability (attack should be multiply by 1.5)
Scenario 3: Ally of attacker has ability (attack should be multiply by 1.5)

[Link to demo](https://www.youtube.com/watch?v=N6fcroNh2S4)


hayuna:
Implementation:

- [x]  Both pokemon have ability (attack should be multiply by 2.25
- [x]  Attacker has ability (attack should be multiply by 1.5)
- [ ]  Ally of attacker has ability (attack should be multiply by 1.5)

src/data/ability.ts
```diff
@@ -1073,6 +1073,40 @@ export class MoveTypePowerBoostAbAttr extends MovePowerBoostAbAttr {
...
+   * @param args Utils.NumberHolder as move power
+   * @returns true if the function succeeds
+   */
+  applyPreAttack(pokemon: Pokemon, passive: boolean, defender: Pokemon, move: PokemonMove, args: any[]): boolean {
+    const isSteelMove = move.getMove().type === Type.STEEL;
+    const onePokemonWithAbility = pokemon.hasAbility(Abilities.STEELY_SPIRIT);
+    const twoPokemonWithAbility = pokemon.scene.currentBattle.double && pokemon.hasAbility(Abilities.STEELY_SPIRIT) && pokemon.getAlly().hasAbility(Abilities.STEELY_SPIRIT);
+    const power = (args[0] as Utils.NumberHolder);
+
+    if (isSteelMove) {
```
td76099, line 1092: You should look into using the .condition instead of placing it inside of `applyPreAttack`. You can see an example of how to implement it into your class [here](https://github.com/pagefaultgames/pokerogue/blob/795c50425fe8159a7cfa56ad898d883e3075184a/src/data/ability.ts#L1049) and an example of it being used [here](https://github.com/pagefaultgames/pokerogue/blob/795c50425fe8159a7cfa56ad898d883e3075184a/src/data/ability.ts#L3773). It's the standard most abilities use for conditions like this.

src/data/ability.ts
```diff
@@ -1073,6 +1073,40 @@ export class MoveTypePowerBoostAbAttr extends MovePowerBoostAbAttr {
...
+      if (twoPokemonWithAbility) {
```
td76099, line 1093: Can't you just (roughly) do:
```
power.value *= 1.5 // You only got here because you have this attribute and the move is Steel, no need to check ability
if (pokemon.getAlly().hasAbility(Abilities.STEELY_SPIRIT) power.value *= 1.5 // Is important to check ally tho
return true;
```
If self has it, it's 1.5. If ally has it, it's 1.5. If both have it, it's 1.5*1.5.

src/field/pokemon.ts
```diff
@@ -1801,6 +1801,15 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
           damage.value = 0;
         }
 
+        // If ally has ability with attribute SteelySpiritAbAttr, steel moves are boosted.
```
td76099, line 1804: There are very little abilities implemented that affect just your ally, so there isn't much of a precedent set, but I can imagine that every single one can get its own clause like this. I would recommend looking at something like `Lightning Rod` or `Storm Drain` because they do something similar (check someone else's ability on their own turn) but not by individually checking each AbAttr.

td76099 (changes requested):
Got a couple more things to figure out - mainly just how to cleanly check an ally's abilities for a damage boost and simplifying the logic in the new attribute itself by adding a conditional where it gets registered and simplifying logic in the attribute.

src/field/pokemon.ts
hayuna: Thanks for the feedback ❤️ Will take a look and try to be better contributor 😊

hayuna: @td76099 Im afraid I cannot base on Lightning Rod or Storm Drain, this ability is raising move power based on ally ability, I tied to move it into phases file but its increasing power forever, then tried to use some Battletags, but seems like its overthinking.

torranx:
@hayuna , do you have plans to update this PR? if not, ill create a new one instead

hayuna:
@torranx go ahead. Lastly I have less time to contribute

hayuna:
New implementation here

https://github.com/pagefaultgames/pokerogue/pull/2749

PR #1449, opened 2024-05-27
RAZOR_CLAW and RAZOR_FANG now available for their evolutions

CodeTappert:
…supposed to according to the comment in the code)

Here is it working for the Hisuian Form (i forced the time of day to "DAY" even when it doesnt look like it)
https://github.com/pagefaultgames/pokerogue/assets/38758606/9a95f130-61d8-4700-928d-10fc61b7fd4a

### DIFF
--- a/src/data/pokemon-evolutions.ts
+++ b/src/data/pokemon-evolutions.ts
@@ -41,6 +41,7 @@ export enum EvolutionItem {
   STRAWBERRY_SWEET,
   UNREMARKABLE_TEACUP,
   RAZOR_CLAW,
+  RAZOR_FANG,
 
   CHIPPED_POT = 51,
   BLACK_AUGURITE,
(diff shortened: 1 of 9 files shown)


TempsRay:
Is the razor fang also in game already?

Xavion3:
Yeah, I think the plan was to only do these after razor claw/fang was actually in the game.

CodeTappert:
So i only need to add it to the evolition item list and do localization. Like i did with razor claw

Xavion3:
Still think this is a really bad idea to add these requirements before adding razor claw and razor fang to the game.

CodeTappert:
> Still think this is a really bad idea to add these requirements before adding razor claw and razor fang to the game.

But they are both in the game (which the changes done in the Review).
Look in my Video in the main post. It shows razor claw working. After i added it

Xavion3:
And it increases the critical hit rate?

CodeTappert:
> And it increases the critical hit rate?
> 
> That's the key, these are not evolution items, they're normal held items which are also required for evolution.

Ohh. I am not deep enough into Pokemon i guess. Well yes then this pr can be Closed. They should be held items instead

PR #1451, opened 2024-05-27
Implement Power Spot

hayuna:
Description: The power of moves used by the allies of the Pokémon with Power Spot is increased by 30%.

[Link to demo](https://youtu.be/LL1kypKD9Dg)


Xavion3:
This should be an attribute, not a hardcoded check for the ability.

hayuna:
Implemented here https://github.com/pagefaultgames/pokerogue/pull/2268

PR #1453, opened 2024-05-27
Implemented "Triple Axel" and "Triple Kick" increasing damage functionality

asvivian:
Triple Axel and Triple Kick have a functionality where their damage is multiplied by 2 on the second hit and 3 on the third hit.

A comment indicates that this functionality would theoretically be added in the MultiHitAddr class for MultiHitType._3_INCR, but it seems that this only really serves to provide a "hitTimes" value.

It seemed more reasonable to add a VariablePowerAttr for this kind of move that increases the power based on number of hits left as VariablePowerAttr is applied for every hit.

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -2356,6 +2356,19 @@ export class BeatUpAttr extends VariablePowerAttr {
   }
 }
 
+export class MultiHitThreeIncAttr extends VariablePowerAttr {
+
+  /**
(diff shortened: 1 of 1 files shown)


asvivian:
Closing this pull request for now.

It seems that the interaction between this and the multi-hit scope cause some complications, going to look into it a bit more before making adjustments:
https://github.com/pagefaultgames/pokerogue/pull/1341

PR #1459, opened 2024-05-27
Add Challenges

Xavion3:
Implements basic challenges. This is still incomplete, but this is being created in draft mode for now so people can see it.


src/data/challenge.ts
```diff
@@ -0,0 +1,542 @@
...
+import Pokemon from "#app/field/pokemon.js";
+import { BattleType, FixedBattleConfig } from "#app/battle.js";
+import { TrainerType } from "./enums/trainer-type";
+import Trainer, { TrainerVariant } from "#app/field/trainer.js";
+import { GameMode } from "#app/game-mode.js";
+
+export enum ChallengeType {
+   /** Challenges which modify what starters you can choose
+    * @param args [0] {@link PokemonSpecies} The species to check
+    *             [1] {@link Utils.BooleanHolder} Sets to false if illegal, pass in true.
```
Tempo-anon, line 15: Enums can have params? I thought by default they were just 0, 1, 2, and so on

src/data/challenge.ts
```diff
@@ -0,0 +1,542 @@
...
+    throw new Error("Method not implemented! Use derived class");
+  }
+}
+
+type ChallengeCondition = (data: GameData) => boolean;
+
+/**
+ * Implements a mono generation challenge.
+ */
+export class SingleGenerationChallenge extends Challenge {
```
Tempo-anon, line 246: nit: Should all the `Challenge`s go in their own file instead of `challenge.ts`?

src/data/challenge.ts
```diff
@@ -0,0 +1,542 @@
...
+      if (species.generation !== this.value) {
+        isValidStarter.value = false;
+        return true;
+      }
+      break;
+    case ChallengeType.POKEMON_IN_BATTLE:
+      const pokemon = args[0] as Pokemon;
+      const isValidPokemon = args[1] as Utils.BooleanHolder;
+      if (pokemon.isPlayer() && ((pokemon.species.formIndex === 0 ? pokemon.species : getPokemonSpecies(pokemon.species.speciesId)).generation !== this.value || (pokemon.isFusion() && (pokemon.fusionSpecies.formIndex === 0 ? pokemon.fusionSpecies : getPokemonSpecies(pokemon.fusionSpecies.speciesId)).generation !== this.value))) {
+        isValidPokemon.value = false;
```
Tempo-anon, line 272: Please add a comment here describing the valid logic

src/data/challenge.ts
```diff
@@ -0,0 +1,542 @@
...
+  static loadChallenge(source: SingleTypeChallenge | any): SingleTypeChallenge {
+    const newChallenge = new SingleTypeChallenge();
+    newChallenge.value = source.value;
+    newChallenge.severity = source.severity;
+    return newChallenge;
+  }
+}
+
+/**
+ * Implements a mono type challenge.
```
Tempo-anon, line 374: Wrong comment?

src/locales/de/challenges.ts
Tempo-anon, line 1: Localization is good but really do bloat up PRs. Maybe something to discuss with the other devs on if localization for features should be in their own PR

src/locales/en/achv.ts
```diff
@@ -168,4 +168,99 @@ export const achv: AchievementTranslationEntries = {
     name: "Undefeated",
     description: "Beat the game in classic mode",
   },
+
+  "MONO_GEN_ONE": {
+    name: "The Original Rival",
```
Tempo-anon, line 173: I think the names should reference the gen/region rather than the final champion. For this one I'd either go with something like "Origins" for a more serious note and "Smell ya later" if going for more levity

src/locales/en/challenges.ts
```diff
@@ -0,0 +1,68 @@
+import { SimpleTranslationEntries } from "#app/plugins/i18n";
+
+export const challenges: SimpleTranslationEntries = {
+  "title": "Challenge Modifiers",
+  "points": "Bad Ideas",
```
Tempo-anon, line 5: Maybe `Challenge Points` to make it clearer for players?

src/phases.ts
```diff
@@ -1287,20 +1310,20 @@ export class SummonPhase extends PartyMemberPokemonPhase {
...
+    if (!partyMember.isAllowedInBattle()) {
+      console.warn("The Pokemon about to be sent out is fainted or illegal under a challenge. Attempting to resolve...");
       const party = this.getParty();
 
       // Find the first non-fainted Pokemon index above the current one
-      const nonFaintedIndex = party.findIndex((p, i) => i > this.partyMemberIndex && !p.isFainted());
-      if (nonFaintedIndex === -1) {
+      const legalIndex = party.findIndex((p, i) => i > this.partyMemberIndex && p.isAllowedInBattle());
+      if (legalIndex === -1) {
         console.error("Party Details:\n", party);
```
Tempo-anon, line 1321: This just freezes and doesn't send the player to a game over?

src/system/game-data.ts
```diff
@@ -124,6 +125,7 @@ export interface SessionSaveData {
   trainer: TrainerData;
   gameVersion: string;
   timestamp: integer;
+  challenges: ChallengeData[];
```
Tempo-anon, line 128: Will there need to be a change to the server so this data is actually saved and loaded?

src/ui/challenges-select-ui-handler.ts
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  setup() {
+    const ui = this.getUi();
+
+    this.challengesContainer = this.scene.add.container(1, -(this.scene.game.canvas.height / 6) + 1);
+
+    this.challengesContainer.setInteractive(new Phaser.Geom.Rectangle(0, 0, this.scene.game.canvas.width / 6, this.scene.game.canvas.height / 6), Phaser.Geom.Rectangle.Contains);
+
+    // TODO: Change this back to /9 when adding in difficulty
```
Tempo-anon, line 42: I assume all the commented out stuff is related to difficulty which is not implemented yet?

Xavion3: Correct, the difficulty code is related to the full challenges system. For now this is just trying to get the basics in and make monotype and monogen playable.

src/data/challenge.ts
Xavion3: I'd say no unless we also want to restructure things like moves and abilities to do the same as well.

src/data/challenge.ts
Xavion3: They can't, so the info here is to describe the parameters when using the apply function for the various challenge types.

src/data/challenge.ts
Tempo-anon: Honestly I wouldn't be opposed to doing that, but I don't know if there's a clear cut way to breaking up the giant abilities or moves file. Here though, each challenge should be distinct enough

src/phases.ts
Xavion3: I mean yes, but that's because the game would be in an error state. It's the same error that happens if you try to switch to a fainted pokemon we already have. Do we want to delete peoples saves when that happens?

src/locales/en/achv.ts
Xavion3: I don't mind some champion references, but I'd need alternative ideas for names. I've considered something like `151` for gen one.

src/locales/en/challenges.ts
Xavion3: Oh yeah, this was an early joke-y name.

src/data/challenge.ts
Xavion3: It's the mono generation challenge, it checks if the pokemon is from the correct generation. I'm not sure what you could assume `.generation !== this.value` to mean other than that.

src/data/challenge.ts
Xavion3: Fixed

src/data/challenge.ts
Xavion3: I feel like that'll just massively bloat file count to give them their own challenges too, remember while there are only really two here there's plans for dozens.

src/system/game-data.ts
XGSleepWalker: https://github.com/pagefaultgames/rogueserver/pull/34

Tempo-anon (approved):
Fill out the PR description and title please

patapancakes (approved):

Simplyfenton:
Would it be possible to merge the functionality of Transfer Items and Check Team? Currently the changes made in [this commit](https://github.com/pagefaultgames/pokerogue/pull/1459/commits/d9a65eff1356da70c4ed303dfed9d99581c9056b#diff-bdf7f0eba287c240a7aa81f9dbd07af7aa8579cb374585985f3e1c9c43b7fd74) conflict heavily with [my PR](https://github.com/pagefaultgames/pokerogue/pull/1093/) in such a way that that's likely the only solution that would make my PR viable without reverting back to the less ideal UI handling around the lower shop options. I also don't much want to make any changes to my PR without asking first, since the work needed to resolve the conflicts could be pointless if my PR isn't valuable/being considered in the first place.

PR #1464, opened 2024-05-28
Adjusted movesets to add more Dexited moves

Fontbane:
This PR adjusts movesets to add moves back to a few mon's movesets that they had in Gen 7 and lost after dexit. A few of them also had the levels they learn moves at adjusted to match USUM to accommodate the bigger learnset.

I also added real damaging moves to mons which lacked them at level 5, as it was insisted that no mon be useless at starting level.

The list of adjusted mons are:

- Rattata, Raticate
- Kadabra, Alakazam
- Cubone, Marowak, Alolan Marowak
- Smoochum, Jynx
- Makuhita, Hariyama
- Skitty
- Luvdisc
- Pachirisu
- Glameow, Purugly
- Purrloin, Liepard
- Woobat, Swoobat
- Audino
- Timburr, Gurdurr, Conkeldurr
- Maractus
- Yungoos, Gumshoos
- Wynaut, Wobbuffet (Confusion at level 1)
- Magikarp (Tackle at level 5)
- Feebas (Tackle at level 5)

### DIFF
--- a/src/data/pokemon-level-moves.ts
+++ b/src/data/pokemon-level-moves.ts
@@ -2223,8 +2223,7 @@ export const pokemonSpeciesLevelMoves: PokemonSpeciesLevelMoves = {
   ],
   [Species.MAGIKARP]: [
     [ 1, Moves.SPLASH ],
-    [ 5, Moves.WATER_GUN ], //Custom
-    [ 15, Moves.TACKLE ],
+    [ 5, Moves.TACKLE ], //Custom, moved from 15 to 5
     [ 25, Moves.FLAIL ],
   ],
   [Species.GYARADOS]: [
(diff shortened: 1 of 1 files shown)


bennybroseph:
Part of this just reverts a commit that was made intentionally to adjust the learnsets of Pokemon to S/V DLC

Fontbane:
> Part of this just reverts a commit that was made intentionally to adjust the learnsets of Pokemon to S/V DLC

Yes, this made it easier to slot in the moves that were removed. Is there a mon in particular this is a problem for?

bennybroseph:
Unless there's some dire balance change, I'm pretty sure we want to remove old moves from learnsets where possible. if they're actually good moves, they are candidates for egg moves.

Fontbane:
> Unless there's some dire balance change, I'm pretty sure we want to remove old moves from learnsets where possible. if they're actually good moves, they are candidates for egg moves.

I would not imagine many would agree with this. There's no reason to put an extra burden on egg moves just to give back moves the mons already had in their learnsets before the moves got removed. Frankly, some of the post dexit movesets are utterly sauceless, and this PR adds back some variety and flavor to movesets without taking any tools away from these mons. This PR was made in accordance with the balance team's wishes.

bennybroseph:
After discussion and a vote, we'd like learnsets to match their Gen 9+ (whatever the current gen is at any given time) counterparts as much as is possible. If a move is absolutely needed for a Pokémon's identity or balance, we can discuss adding a TM or Egg Move for that particular move. Preferably, we should add a TM for moves which are no longer in Gen 9+. This allows these Pokémon to benefit from an in game reward system without disrupting their intended learnsets.

PR #1467, opened 2024-05-28
Implemented Triple Kick and Triple Axel properly

asvivian:
Added a proposed implementation for cyclically incrementing power in the moves Triple Kick and Triple Axel.

This implementation is based on MultiHitIncrementerThree which extends VariablePowerAttr.

This should cause each hit of these abilities to do 1, 2, or 3 times its base power on each hit.

This is repeated for each multi lens the Pokémon is holding as well.

For example, a Pokémon holding 1 multi lens using Triple Axel would have each hit of Triple Axel have a power, respectively, of: 20, 40, 60, 20, 40 60.

This seems like the natural follow-up to https://github.com/pagefaultgames/pokerogue/pull/1341 which, to be honest, I am not sure that was not followed through on.

However, since this has nothing to do with the accuracy component of these moves and only their power, Population Bomb is not associated with this update at all despite its inclusion in 1341 (which **does** address accuracy).

Edit: Small adjustments for clarity about accuracy component's exclusion.

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -2356,6 +2356,19 @@ export class BeatUpAttr extends VariablePowerAttr {
   }
 }
 
+export class MultiHitIncrementerThree extends VariablePowerAttr {
+
+  /**
+  * Gets the current number of hits left and doubles or triples the power for the 2nd and 3rd hits respectively
+  */
+
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    const power = args[0] as Utils.NumberHolder;
+    power.value = move.power * (1+((3-(user.turnData.hitsLeft%3))%3));
+    return true;
+  }
+}
+
 const doublePowerChanceMessageFunc = (user: Pokemon, target: Pokemon, move: Move) => {
   let message: string = null;
   user.scene.executeWithSeedOffset(() => {
@@ -5409,6 +5422,7 @@ export function initMoves() {
       .ignoresVirtual(),
     new AttackMove(Moves.TRIPLE_KICK, Type.FIGHTING, MoveCategory.PHYSICAL, 10, 90, 10, -1, 0, 2)
       .attr(MultiHitAttr, MultiHitType._3_INCR)
+      .attr(MultiHitIncrementerThree)
       .attr(MissEffectAttr, (user: Pokemon, move: Move) => {
         user.turnData.hitsLeft = 1;
         return true;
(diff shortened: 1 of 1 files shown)


Blitz425:
Did you test the branch or look at the comments? The person already addressed it in a commit 2 days ago

asvivian:
> Did you test the branch or look at the comments? The person already addressed it in a commit 2 days ago

@Blitz425 
Yep, tested and it is working as intended.

I do see that they are actively making commits in https://github.com/pagefaultgames/pokerogue/pull/1341 but I think the solution I am proposing is significantly simpler.

xsn34kzx:
> I do see that they are actively making commits in #1341 but I think the solution I am proposing is significantly simpler.

Our implementations of the power aspect of the two moves pretty much only differ by [one line](https://github.com/xsn34kzx/pokerogue/blob/2e2c0d0d94d2cd271834bf31689b31d0eed61654/src/data/move.ts#L2761), and I'm unsure how mine is any more complex. Moreover, the commit doesn't address the current conversation in #1341 and the complexity of the moves' interactions, let alone the separate accuracy checks mentioned in a [previous PR](https://github.com/pagefaultgames/pokerogue/pull/365#issuecomment-2106103675). This would still make the moves partially implemented rather than a proper, full implementation.

asvivian:
> Our implementations of the power aspect of the two moves pretty much only differ by [one line](https://github.com/xsn34kzx/pokerogue/blob/2e2c0d0d94d2cd271834bf31689b31d0eed61654/src/data/move.ts#L2761), and I'm unsure how mine is any more complex. Moreover, the commit doesn't address the current conversation in #1341 and the complexity of the moves' interactions, let alone the separate accuracy checks mentioned in a [previous PR](https://github.com/pagefaultgames/pokerogue/pull/365#issuecomment-2106103675). This would still make the moves partially implemented rather than a proper, full implementation.

@xsn34kzx 
You're right! The implementation of the power aspect is very similar.

I'm getting a kind of hostile tone (?) from your comment but I think we are on the same team so let me just clear up what I am saying. When I say my solution is simpler, all I am saying is that my PR is focused solely on this power aspect and nothing more.

I agree that something needs to be implemented to address the specific accuracy element of Triple Axel, Triple Kick, and Population Bomb, but that seems to me like it would warrant its own separate PR since it doesn't really need to have any interaction with the incrementing power component. That is what I meant in my original comment explaining why I excluded Population Bomb from my changes.

> Moreover, the commit doesn't address the current conversation in #1341 and the complexity of the moves' interactions

I agree that this implementation is ignoring the accuracy component for the stated reasons, but what interactions do you think are not properly handled by this implementation?

In #1341 I see discussion of capping the power but that isn't really relevant with the modulating power approach (which you seem to have taken as well).

There is also discussion about Skill Link, but that (again) seems like something that would matter only for the implementation of the accuracy component and not the power.

All that being said, I think you are doing good work over on #1341, I just think the power component and accuracy component are two different matters and there is no reason one can't be added without the other.

Tempo-anon:
Closing since the other PR was merged in

PR #1468, opened 2024-05-28
Translation fixes ptBR

josericardo-fo:


TempsRay:
Since you made changes to the info container can you show some images of before/after?

Issue #1356, opened 2024-05-25
Enforce space around curly braces


JakubHanko:
In my opinion, we use the spaces around curly braces fairly inconsistently, see for example imports in `trainer.ts`, `battle-message-ui-handler.ts:122` (the definition of `levelUpStatsValuesContent` variable), just to name a few.

Do we want to add a new rule to ESLint?

https://eslint.style/rules/js/object-curly-spacing

If yes, I can just add one more commit to #1323 .

j-diefenbach:
Might aswell imo

PR #1469, opened 2024-05-28
Fix gender override

torranx:
This is just an import issue

### DIFF
--- a/src/overrides.ts
+++ b/src/overrides.ts
@@ -12,7 +12,7 @@ import { Stat } from "./data/pokemon-stat";
 import { PokeballCounts } from "./battle-scene";
 import { PokeballType } from "./data/pokeball";
 import {TimeOfDay} from "#app/data/enums/time-of-day";
-import {Gender} from "pokenode-ts";
+import { Gender } from "./data/gender";
 
 /**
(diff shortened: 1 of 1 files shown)


src/overrides.ts
```diff
@@ -12,7 +12,7 @@ import { Stat } from "./data/pokemon-stat";
 import { PokeballCounts } from "./battle-scene";
 import { PokeballType } from "./data/pokeball";
 import {TimeOfDay} from "#app/data/enums/time-of-day";
```
TempsRay, line 14: line 14 import also looks a bit off?

TempsRay (commented):
What happened here? What's `pokenode-ts`?

torranx:
> What happened here? What's `pokenode-ts`?

This is a node_module

src/overrides.ts
torranx: yeah idk linter didnt catch this, pretty sure there are also other occurrences of this

torranx: Issue raised here too: https://github.com/pagefaultgames/pokerogue/issues/1356

TempsRay (approved):

PR #1472, opened 2024-05-28
Wide Lens to Ultra item tier, Berry Count limits, and small Berry Pouch % chance reduction

damocleas:
- Wide Lens moved from Rogue -> Ultra Tier, same weight
- Changed Berry Pouch success rate %, from 33% per stack down to 30%, capped at 90% for 3 stacks.
- Reduced Berry Count limits, 2 for Lum, Leppa, Sitrus, Enigma - 3 for the rest


Blitz425:
Wide Lens had NO business being in Rogue Tier.

Xavion3:
There any explanation for some of these? Particularly making you get more than 10x as many vouchers from rewards, that seems like a pretty dramatic change to be just snuck in under an update which is otherwise focused on actual mechanical balance.

ben-lear:
These changes might have a (possibly unintended) significant impact on the functionality of Pickup mons, as you will no longer be able to go as many floors picking up berries before you are forced to re-distribute due to max stacks. It could get pretty annoying to micromanage berry stacks that small, especially with no ability to split stacks.

I really don't think the ability to stack berries to 10 is anywhere near game-breaking, especially with some of the other wacky tools at players' disposal.

damocleas:
> Is there any explanation for some of these? Particularly making you get more than 10x as many vouchers from rewards, that seems like a pretty dramatic change to be just snuck in under an update which is otherwise focused on actual mechanical balance.

Moving them to a different PR.

Also tested this, confirmed automatically reduces berry counts in current runs.

src/modifier/modifier.ts
```diff
@@ -951,7 +951,14 @@ export class BerryModifier extends PokemonHeldItemModifier {
...
   getMaxHeldItemCount(pokemon: Pokemon): integer {
-    return 10;
+    switch (this.berryType) {
+    case BerryType.LUM:
+    case BerryType.LEPPA:
+    case BerryType.SITRUS:
+    case BerryType.ENIGMA:
+      return 2;
+    }
+    return 3;
```
Xavion3, line 961: Can we not use switch statements as a janky includes?

Xavion3 (changes requested):

XGSleepWalker (changes requested):
Waiting on Discord announcement

Xavion3 (approved):

XGSleepWalker (approved):

PR #1473, opened 2024-05-28
Update biome.ts fix zh_CH biome translation based on chinese community poll

40chyan:
some ambigious translation has been changed
the names are based on the result of a poll from chinese player community
Pokerogue区域译名投票 (1)


CodeTappert:
Could you please change your title so it is Clear that you are talking about zn_CH?

So a Person looking to Chance the zn_TW locales doesnt think there is already a PR for that

TempsRay (approved):

PR #1477, opened 2024-05-28
Match Shiny Grafaiai animation frames with Non-Shiny counterparts

chaosgrimmon:
A user reported [here](https://discord.com/channels/1125469663833370665/1238339524778790922/1244817919985061978) that "Shiny grafaiai keeps fading in and out of existence".

The presumed culprit is frame 17. It has a width and height of 3 pixels. Also, no sprite exists at the indicated position, which would basically necessitate a third column [here](https://github.com/pagefaultgames/pokerogue/blob/main/public/images/pokemon/exp/shiny/945.png).
To fix this, the code from the non-shiny [.json](https://github.com/pagefaultgames/pokerogue/blob/main/public/images/pokemon/exp/945.json#L139) was used instead, which has dimensions of 66x55, somewhat more reasonable for the targeted slice of spritesheet.

Not tested, with due concern given to the abysmal odds of specifically encountering a shiny Grafaiai.
... but the change shouldn't break anything, I hope?


TempsRay:
You can manually test this by going in to `src/overrides.ts` and spawning in a shiny Grafaiai with the overrides

chaosgrimmon:
Before:
https://github.com/pagefaultgames/pokerogue/assets/31082757/9606af33-cdbd-488a-803b-7ffa2dc9cf14

After:
https://github.com/pagefaultgames/pokerogue/assets/31082757/aca5781f-857f-4af5-9aaa-b662aab40702

Confident that the sole cause of the issue was fully addressed, and is ready to merge.
ps. thanks for reminding me that overrides.ts exists

Tempo-anon (approved):

PR #1479, opened 2024-05-28
zh_CN translation Update pokemon-info-container.ts  [Localization]

40chyan:


40chyan:
@Tempo-anon I speak Chinese so basicly I think these terms are correct. But just in case, all translation of terms based on https://wiki.52poke.com/ , which is simply Chinese version of Bulbapedia.

SnowCharmQ:
@40chyan Typo again in your PR title, you can fix it.

40chyan:
@SnowCharmQ Yes, haven't fixed this since you said yesterday, sorry for that.

Tempo-anon (approved):
I'll take snowcharm's word that these terms are accepted

PR #1480, opened 2024-05-28
Starter select - Implement up and down offset navigation

zronn:
#### Changes

- Add conditions when moving up or down at the edges of the select starter ui

#### Purpose

- Make it easier to navigate in the select starter ui by moving the cursor to the opposing edge

#### Demo

https://github.com/pagefaultgames/pokerogue/assets/12510911/06b218b0-8028-431b-85cf-93d2780d642e


bennybroseph (changes requested):
would like some simple comments to mark each block

zronn:
@bennybroseph done 👍

bennybroseph (approved):

PR #1482, opened 2024-05-28
Fix - Removes useless '/' make it them doubles in somes URLs

Dakurei:
I was tired of seeing malformed URLs in my Nginx logs 🤣

> Before
image
image
image

> After
image
image
image


TempsRay:
Are the double slashes in everything? Or will this possibly break existing stuff that requires it that way?

PR #1485, opened 2024-05-28
Updated pkmnems Font

Adri1:
In French, when **pkmnems.ttf** is used, it sometimes looks like this because Œ and œ are missing:
font before

So I updated it :
font corrected


Adri1:
> how'd we get this font?

It's not in the branch used for this PR?

bennybroseph (approved):

PR #1501, opened 2024-05-28
[QOL] Removed King's Rock requirement for Politoed/Slowking evos

Fontbane:
Removes King's Rock as an evolution requirement for Politoed and Slowking.

Currently, Poliwhirl and Slowpoke evolve into Politoed and Slowking by using a Linking Cord while holding a King's Rock. This requires the player to roll a Rogue tier item and then an Ultra tier item. It's easier to get Archaludon than Politoed. 

This PR simply removes the King's Rock requirement and just requires the player to use a Linking Cord. This is in line with other evolutions that require an evolution item in the main series, such as Kingdra and Slurpuff.


bennybroseph:
Needs ESLint run

PR #1502, opened 2024-05-28
[Bug] Fix weather effects to work on the first turn of being cast

ben-lear:
## What are the changes?
This update changes the turn phase to always have a weather phase (even if no weather is set), and check for the state of the weather when the weather phase begins. If no weather is set, nothing happens, but otherwise the weather effects will propagate normally

Prior to this fix, if a weather effect was cast as part of a move in the middle of a turn, `WeatherPhase` was not updated properly to reflect the state change, and effects like damage from Sandstorm/Hail or healing from Ice Body would not proc on the first turn of the weather. 

## What did change?
Weather effects will now proc correctly on the first turn of weather being cast.

### Screenshots/Videos
Old version:
https://github.com/pagefaultgames/pokerogue/assets/110984302/14e6e695-035f-4d0a-b310-8df41bf24c1f

New version (shows swapping weather effects still updates properly, etc.):
https://github.com/pagefaultgames/pokerogue/assets/110984302/fa0eceb6-061c-4b30-a40e-a8a384bb8e21

## How to test the changes?
Session data used to test:
[session data.zip](https://github.com/pagefaultgames/pokerogue/files/15473543/session.data.zip)

## Checklist
- [x] Are the changes visual?


src/field/arena.ts
```diff
@@ -302,11 +302,12 @@ export class Arena {
     this.weather = weather ? new Weather(weather, hasPokemonSource ? 5 : 0) : null;
 
     if (this.weather) {
-      this.scene.tryReplacePhase(phase => phase instanceof WeatherEffectPhase && phase.weather.weatherType === oldWeatherType, new WeatherEffectPhase(this.scene, this.weather));
```
ben-lear, line 305: tryReplacePhase() is no longer necessary as `WeatherEffectPhase` will programmatically determine the weather state at the start of the phase

PR #1504, opened 2024-05-28
Implement starters overrides

Kiriox94:
## What are the changes?
Addition of 3 new overrides to manage available starters

I'm currently working on the localization of pokemon forms, so I've added these overrides to make it easier to check that the form name is displayed in the starters menu.

## How to test the changes?
Go to `src/overrides.ts` and change `STARTERS_OVERRIDE` or `STARTERS_FORMS_OVERRIDE` or `STARTERS_ALL_FORMS_OVERRIDE` to `true`.

## Checklist

### DIFF
--- a/src/overrides.ts
+++ b/src/overrides.ts
@@ -105,3 +105,19 @@ export const OPP_MODIFIER_OVERRIDE: Array<ModifierOverride> = [];
 
 export const STARTING_HELD_ITEMS_OVERRIDE: Array<ModifierOverride> = [];
 export const OPP_HELD_ITEMS_OVERRIDE: Array<ModifierOverride> = [];
+
+/**
+ * STARTERS OVERRIDE
+ * Unlock all starters
+ */
+export const STARTERS_OVERRIDE: boolean = false;
+/**
+ * STARTERS FORMS OVERRIDE
+ * Unlock all forms for starters
+ */
+export const STARTERS_FORMS_OVERRIDE: boolean = false;
+/**
+ * STARTERS OVERRIDE
+ * Unlock all form even those not usually available for starters
+ */
+export const STARTERS_ALL_FORMS_OVERRIDE: boolean = false;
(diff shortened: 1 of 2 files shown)


Tempo-anon (changes requested):
I don't know about this. This seems pretty dangerous to include as an override. Wouldn't it be better just to have a full psrv save file to be used for debugging? 

Also linter fails.

Kiriox94:
> I don't know about this. This seems pretty dangerous to include as an override. Wouldn't it be better just to have a full psrv save > file to be used for debugging?
>Also linter fails.

I don't see what's more dangerous in this than in the other overrides. If you're worried that it won't be deactivated before a pull request, you should know that `overrides.ts` is in .gitignore, so there's no risk there.

bennybroseph:
> > I don't know about this. This seems pretty dangerous to include as an override. Wouldn't it be better just to have a full psrv save > file to be used for debugging?
> > Also linter fails.
> 
> I don't see what's more dangerous in this than in the other overrides. If you're worried that it won't be deactivated before a pull request, you should know that `overrides.ts` is in .gitignore, so there's no risk there.

Unfortunately that's not how .gitignore works

bennybroseph:
I'm with Temp, this is way too dangerous. If somebody is not paying attention and merges this it completely corrupts every single user in the DB's save file irreversibly. We need some better debugging tools tbh. This is just asking for trouble.

Kiriox94:
> I'm with Temp, this is way too dangerous. If somebody is not paying attention and merges this it completely corrupts every single user in the DB's save file irreversibly. We need some better debugging tools tbh. This is just asking for trouble.

I don't understand how my code can corrupt saves or database since my code doesn't touch it, it just falsifies the conditions of `starters-ui-handler.ts`

Kiriox94:
> > > I don't know about this. This seems pretty dangerous to include as an override. Wouldn't it be better just to have a full psrv save > file to be used for debugging?
> > > Also linter fails.
> > 
> > I don't see what's more dangerous in this than in the other overrides. If you're worried that it won't be deactivated before a pull request, you should know that `overrides.ts` is in .gitignore, so there's no risk there.
> 
> Unfortunately that's not how .gitignoreHowever, it seemed to me that its operation was to make gît ignore the files that are in it, am I wrong?

PR #1511, opened 2024-05-28
[Localization] Updated pixel font for Simplified Chinese

Siyuanw:
## What are the changes?

Introduced pixel font for Chinese, to match the game style.

The font comes from an open-source project https://github.com/TakWolf/fusion-pixel-font

And you can find the licence here https://github.com/TakWolf/ark-pixel-font/blob/develop/LICENSE-OFL

The game currently uses a font that only contains basic latin characters.

All the CJK characters are fall back to a system defined font.

## What did change?

Added a pixel font in the static resource, and use the font when the language is set to Simplified Chinese.

There're too many places which are designed for "emerald" font.

After the change, the "emerald" font still take the precedence over the newly introduced pixel font.

### Screenshots/Videos

main

battle

## How to test the changes?

Please switch the language to Chinese Simplified to verify the change.

change

## Checklist
- [x] Are the changes visual?


Tempo-anon:
While this does fit the game aesthetics better I feel like it loses some readability

40chyan:
Actually I think this font is good, the original ch font not only looks weird, several pixels on the very top of each chinese character are not being displayed correctly. In terms of readability, I don't think there's any problem on it. Or you think I should make a poll in CH community to make sure? @Tempo-anon 
image

Tempo-anon:
> Actually I think this font is good, the original ch font not only looks weird, several pixels on the very top of each chinese character are not being displayed correctly. In terms of readability, I don't think there's any problem on it. Or you think I should make a poll in CH community to make sure? @Tempo-anon image

If you can get the CH player's community's thoughts on this that would be great since they would be the primary users of this text.

bennybroseph:
It seems like the community has a different solution for this. Feel free to join the discussion in the Discord

PR #1512, opened 2024-05-28
Update starter-select-ui-handler.ts

bennybroseph:
Fixes an error with shiny star placement

### Before

image

### After

image


Tempo-anon (commented):
Can you please add before/after images in the PR desc to match the requirements we have for other PRs?

Tempo-anon (approved):

PR #1514, opened 2024-05-28
Fixes setPositionRelative Errors

bennybroseph:
## What are the changes?

There are no user-facing changes

This fixes ~100 errors in VS Code due to a missing reference

## What did change?

The global function setPositionRelative is now defined properly as an extension method and is therefore recognized by the IDE

## How to test the changes?

Open the Summary Screen during an Endless Fused run. The Fusion Icon calls this function.

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -5,7 +5,6 @@ import { version } from "../package.json";
 import UIPlugin from "phaser3-rex-plugins/templates/ui/ui-plugin";
 import BBCodeTextPlugin from "phaser3-rex-plugins/plugins/bbcodetext-plugin";
 import InputTextPlugin from "phaser3-rex-plugins/plugins/inputtext-plugin.js";
-import BBCodeText from "phaser3-rex-plugins/plugins/bbcodetext";
 import TransitionImagePackPlugin from "phaser3-rex-plugins/templates/transitionimagepack/transitionimagepack-plugin.js";
 import { LoadingScene } from "./loading-scene";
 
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -5,7 +5,6 @@ import { version } from "../package.json";
 import UIPlugin from "phaser3-rex-plugins/templates/ui/ui-plugin";
 import BBCodeTextPlugin from "phaser3-rex-plugins/plugins/bbcodetext-plugin";
 import InputTextPlugin from "phaser3-rex-plugins/plugins/inputtext-plugin.js";
-import BBCodeText from "phaser3-rex-plugins/plugins/bbcodetext";
```
bennybroseph, line 8: This is on purpose, this never gets used

src/main.ts
```diff
@@ -72,23 +71,82 @@ const config: Phaser.Types.Core.GameConfig = {
...
+		}
+	}
+}
+
 Phaser.GameObjects.Container.prototype.setPositionRelative = setPositionRelative;
 Phaser.GameObjects.Sprite.prototype.setPositionRelative = setPositionRelative;
 Phaser.GameObjects.Image.prototype.setPositionRelative = setPositionRelative;
 Phaser.GameObjects.NineSlice.prototype.setPositionRelative = setPositionRelative;
 Phaser.GameObjects.Text.prototype.setPositionRelative = setPositionRelative;
-BBCodeText.prototype.setPositionRelative = setPositionRelative;
```
bennybroseph, line 91: This is on purpose, this never gets used

src/main.ts
```diff
@@ -72,23 +71,82 @@ const config: Phaser.Types.Core.GameConfig = {
...
 };
 
+/**
+ * Sets this object's position relative to another object with a given offset
+ * @param guideObject {@linkcode Phaser.GameObjects.GameObject} to base the position off of
+ * @param x The relative x position
+ * @param y The relative y position
+ */
 const setPositionRelative = function (guideObject: any, x: number, y: number) {
-  if (guideObject && guideObject instanceof Phaser.GameObjects.GameObject) {
```
molker, line 76: is there not a case where the guideobject is not an extension of game object we need to worry about here? I know it was giving errors with guideObject.x but those would go away if you did `(guideObject as any) instanceof Phaser.GameObjects.GameObject`

bennybroseph: Nah, that's from when `BBCodeText` also 'used' this function and therefore wasn't an instance of `GameObject`

everything else should extend `GameObject` now, even `Rectangle` which extends from `Shape` which extends from `GameObject`

`BBCodeText` extended the method but it never actually got called. If you know how to write the extension method for `BBCodeText` we can add it back, but it'd only be for future use since it wasn't ever called before.

molker: Alright fair, worst case scenario if we need it later we can add it back in but for now that works for me

molker (approved):

PR #1516, opened 2024-05-28
Voucher Item tier changes/addition

damocleas:
## What are the changes?
- Voucher moved from Ultra -> Great Tier, given a weight of 1 and disappears after first reroll
- Voucher Plus moved from Master -> Rogue Tier, with weight starting at 9 -> 5 and decreasing with each reroll with 3 -> 2
- Voucher Premium added to Master Tier with a lower weight of the previous spot of Voucher Plus, and disabled in Endless / Endless Spliced

Overall would be a 40-45% increase in total vouchers from rng in perfect conditions (luck, healthy team due to how great ball tier is weighted, etc.)

Vouchers should still appear more often with a healthy team than they did in ultra tier at all, gives more time where they'd appear and wouldn't feel bad picking them up.

Voucher Plus should appear just a small amount more often than before, and was moved down to make room for Voucher Premium while keeping them just around the same rarity.

Voucher Premiums should appear ~20% of the time with *perfect luck* in a whole average classic run, and gives a much better feeling than getting a random 5 if you end up getting that tier in Classic.

## How to test the changes?
For making sure Voucher Premiums didn't show up in endless, just set them with a huge weight value and ran through endless for a few rounds.
Easy to test with math and running a classic/starting through an endless run with it otherwise

## Checklist
- [✔] There is no overlap with another PR?
- [✔ ] The PR is self-contained and cannot be split into smaller PRs?
- [ ? ] Have I provided a clear explanation of the changes?
- [✔ ] Have I tested the changes (manually)?
    - [✔ ] Are all unit tests still passing? (`npm run test`)
- [ X ] Are the changes visual?
  - [ X ] Have I provided screenshots/videos of the changes?


Xavion3:
Can you provide how you reached those conclusions about those rates? Just for verification.

brain-frog (approved):

PR #1518, opened 2024-05-28
[Enhancement] Add Hidden Ability Gacha

aeloan:
## What are the changes?

Added a gacha with an x2 hidden ability chance up from the feature vote : https://trello.com/c/JBNmgfID/59-egg-gacha-for-higher-hidden-ability-chance
Can be seen at the right of the shiny gacha with an ability pill on top

These changes where asked on a feature vote, as a begginer on open-source project, I tried it

### Screenshots/Videos

image

## How to test the changes?

Change the UP_HIDDEN_ABILITY_CHANCE_OVERRIDE in the overrides.ts and test with multiple egg , you will know the presence of the hidden abilty during the hatch

## Checklist
- [ ] Are the changes visual?


public/images/egg/gacha_underlay_hidden_ability.json
```diff
@@ -0,0 +1,62 @@
+{
+	"textures": [
+		{
+			"image": "gacha_underlay_shiny.png",
```
Icehawk78, line 4: This looks like the wrong filename

src/egg-hatch-phase.ts
```diff
@@ -440,7 +441,8 @@ export class EggHatchPhase extends Phase {
 
         const pokemonSpecies = getPokemonSpecies(species);
 
-        ret = this.scene.addPlayerPokemon(pokemonSpecies, 1, undefined, undefined, undefined, false);
+        const hiddenAbilityChance = this.egg.gachaType === GachaType.HIDDEN_ABILITY ? BASE_HIDDEN_ABILITY_CHANCE_OVERRIDE : UP_HIDDEN_ABILITY_CHANCE_OVERRIDE;
+        ret = this.scene.addPlayerPokemon(pokemonSpecies, 1, undefined, undefined, undefined, false, undefined, undefined, undefined, undefined, hiddenAbilityChance);
```
Icehawk78, line 445: Could you create a new method similar to the `trySetShiny` for setting the hidden ability, to avoid having to modify the constructor signature?

src/overrides.ts
```diff
@@ -45,6 +45,8 @@ export const POKEBALL_OVERRIDE: { active: boolean, pokeballs: PokeballCounts } =
     [PokeballType.MASTER_BALL]: 0,
   }
 };
+export const BASE_HIDDEN_ABILITY_CHANCE_OVERRIDE: integer = 256;
+export const UP_HIDDEN_ABILITY_CHANCE_OVERRIDE: integer = 128;
```
Icehawk78, line 49: These seem more like global values than overrides. I don't hate the idea of extracting a bunch of these out into a more centralized location to make balancing changes easier, but this is definitely a different pattern than what the rest of the project uses. Probably worth bringing up in a larger discussion with the wider dev group on Discord?

Icehawk78 (changes requested):

src/egg-hatch-phase.ts
aeloan: I saw the method trySetShiny but I don't see the point of adding a method. Why would we change the presence of shiny on a pokemon after his initialization ? Where we use it for now is before the user knows it is shiny or not but unless we want to add an item that change the shinyness of a pokemon, it wouldn't happen on a real pokemon game

bennybroseph:
This is incredibly unprofessional for us and I really do apologize, but that Trello has been out of date for months. We use a different tool for this purpose now and it is not user-facing. We no longer want this feature implemented. I hope you do not see this as time wasted though, as everything is a learning experience. Everything done here just makes the next change easier.

Issue #1398, opened 2024-05-26
[Bug/Suggestion] Battle menu cursor should default to Attack after switching Pokemon/when starting a new battle


PsikoBlock:
Hi, loving the game so far :)
One small thing that threw me off a few times: my muscle memory tells me that I can instantly press A to attack after a switch-out, but that leads to opening the Pokemon menu. Also, when starting a new battle after catching a Pokemon, the Ball button is still selected. I think that in both cases defaulting to Attack matches the original games best – that's what you want to do anyways most of the time.

TobRojekt:
I think keeping the cursor on the last selected option is not that bad. In late phases of endless runs I mainly keep running away from encounters until a shiny or legendary shows up. I understand that this might be a specific case but maybe an option to toggle between defaulting to attack and keeping it where it was would be nice. :)

PsikoBlock:
Hmm, maybe the Run action could be an exception then – or making it configurable.

damocleas:
This is normal functionality as of https://github.com/pagefaultgames/pokerogue/pull/4666
Added Cursor Memory Setting in #5028

PR #1522, opened 2024-05-28
[QoL] Cursor memory setting

LeBulldoge:
## What are the changes?
This PR adds a new setting to control whether or not the game maintains the cursor at its last position between battles, inside the command and move menus. The possible values are:
* Off
* Commands
* Moves
* Both

The default value is `Both`, in line with current behavior.

I frequently find myself in a loop of sorts, as my mind is expecting the `Fight` button to be highlighted after each combat. This especially hurts if I managed to fumble the inputs last fight and hit `Run`.
However, I do find cursor memory for the move menu quite useful. This way players can pick and choose whichever setting feels best for them.
This should also close out #1398 

## What did change?
Added a new setting `Cursor memory`, which will let players control in which combat menus they want the last cursor position to be remembered.

### Screenshots/Videos
https://github.com/pagefaultgames/pokerogue/assets/13983982/a53d862c-bfce-4c58-a268-858cb682620d

## How to test the changes?
Change the setting to the desired value. While in combat, the cursor position should, or should not, be remembered on the command or move menus, according to the chosen value.

## Checklist
- [x] Are the changes visual?


flx-sta:
Does this also handle that the cursor memory is separate in tag-team fights?
If not, would be a good addition

LeBulldoge:
> Does this also handle that the cursor memory is separate in tag-team fights? If not, would be a good addition.
> 
> Right now the previous choice of Pokemon 2 can affect the next choice of Pokemon 1 in the next turn (and the other way around same turn)

This PR doesn't change that and it probably shouldn't, seems like a whole separate bug. I wouldn't mind looking into it after this though. Is there an issue open for it?

flx-sta:
> > Does this also handle that the cursor memory is separate in tag-team fights? If not, would be a good addition.
> > Right now the previous choice of Pokemon 2 can affect the next choice of Pokemon 1 in the next turn (and the other way around same turn)
> 
> This PR doesn't change that and it probably shouldn't, seems like a whole separate bug. I wouldn't mind looking into it after this though. Is there an issue open for it?

I don’t think there is. Try searching for it. Otherwise you can suggest it on their discord

Tempo-anon:
Cool idea but this PR is a bit out of date and has merge conflicts. You're welcome to make a new PR for this feature though

PR #1525, opened 2024-05-29
[Move] Added Item Removal functionality for Incinerate and Knock Off

MPekelny:
As a note, I had previously put up this pull request, but I had done it on my checkout's main. Now I am opening the pull request with the changes on a branch.

## What are the changes?
Adding the item removal functionality to Knock Off and Incinerate.

These are two moves that currently do not have functionality, and they are ones where the extra functionality is relatively simple, so it is a good first thing for me to do to dip my toes into contributing.

## What did change?
There was an existing RemoveHeldItemAttribute which seems to be what was intended to do the knock off effect, but it was stealing items. So I updated it to instead only remove an item, adding a tryRemoveHeldItemModifier that works similar to the transfer version. I did also see the current pull request for Incinerate functionality, but I think rather than having a specific attribute for berries, it is better to update the existing property to handle both cases.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/73674424/74b03cdd-2b7c-40c8-8b9a-d993fdd85677

## How to test the changes?
Add to overrides.ts to give your Pokemon Knock Off and Incinerate and the opponent some held item and try them in battle. I also changed the starting wave and gave the opponent only non-damaging moves so it can take several hits and not damage back to make testing easier.

## Checklist
    - There is one pull request that also does Incinerate, but it is basically the same functionality as Knock Off and I think this more general way of doing it is better.
- [x] Are the changes visual?

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1877,6 +1877,54 @@ export default class BattleScene extends SceneBase {
   }
 
   /**
+   * Tries to remove a given held item from the target pokemon.
+   * @param itemModifier The type of held item to try removing.
+   * @param target The Pokemon the held item should be removed from.
+   * @param removeStack If true, the full stack of the held item should be removed. If false, only one of that item should be removed.
+   * @param instant Whether the modifiers should be updated instantly.
+   * @returns A promise where the resolve boolean is if the remove action successfully removed the specified item.
+   */
+  tryRemoveHeldItemModifier(itemModifier: PokemonHeldItemModifier, target: Pokemon, removeStack: boolean, instant?: boolean): Promise<boolean> {
+    return new Promise(resolve => {
+      const source = itemModifier.pokemonId ? itemModifier.getPokemon(target.scene) : null;
+      const cancelled = new Utils.BooleanHolder(false);
+      Utils.executeIf(source && source.isPlayer() !== target.isPlayer(), () => applyAbAttrs(BlockItemTheftAbAttr, source, cancelled)).then(() => {
+        if (cancelled.value) {
+          return resolve(false);
+        }
+        const newItemModifier = itemModifier.clone() as PokemonHeldItemModifier;
+        newItemModifier.pokemonId = target.id;
+        const matchingModifier = target.scene.findModifier(m => m instanceof PokemonHeldItemModifier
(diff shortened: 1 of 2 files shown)


geefuoco:
This is a good start but there is some functionality missing
https://bulbapedia.bulbagarden.net/wiki/Knock_Off_(move)

- Knock off's damage is increased by 50% is an item is knocked off (not sure if this is implemented already). 
- Focus Band, Focus Sash, Colbur berry can all be used before being knocked off. 
- We also need to consider the interaction with Sticky Hold ability

Another consideration is that wild pokemon cannot knock off a player's items (only trainers). 

Incinerate looks ok

MPekelny:
Knock Off already had a KnockOffPowerAttr, which the code looks like it already does that, so I didn't bother with that, however it is probably a good idea to do a test and double check it is doing what it is supposed to.

MPekelny:
Actually, I think so that the item removal can get in rather than sit around for however long it takes to get the other functionality working, I will confirm if the power increase is working, then fix the conflicts, and make Knock Off partial again. Then after merge I will work on getting the other aspects of Knock Off implemented.

MPekelny:
So, something I have realized looking at the code again. The code for which item is stolen by Thief (which I largely used for Knock Off and Incinerate) is not choosing items randomly. Instead it is just getting the highest tier item that is held by a Pokemon and picking that (I'm not entirely sure the logic behind that). This means if a Pokemon has a Multi Lens and it gets hit by Knock Off or Thief, the item that gets taken will always be Multi Lens, regardless of what other items the Pokemon is holding.

I don't think this is the behavior we actually want (maybe it is fine for Thief since it is a less common move and only a 30% steal rate? Also maybe fine if it is doing that for the non player Pokemon?), I think this probably needs a lead dev input but which of these options should be the case:

1. Fine as is, for both Thief and Knock Off.
2. Fine as if for Thief, but Knock Off should pick randomly.
3. Both Thief and Knock Off should pick randomly.
4. Fine to pick highest tier if it is the enemy Pokemon being hit, but randomly if the player Pokemon being hit.
5. Something else?

MPekelny:
#1671 has a more complete implementation, so I will close this and let that one be merged.

PR #1526, opened 2024-05-29
[enhancement] Added overrides for Post Battle rewards

MPekelny:
## What are the changes?
Added the ability to override how many and what items appear during the post battle reward screen.

There are a couple thing I want to do at some point where having the ability to specify rewards will be very helpful and I can also imagine others might find this functionality useful. So, I figure it is probably preferable to get this change reviewed/committed first separately from the other stuff I will be doing.

## What did change?
I added a REWARD_AMOUNT_OVERRIDE value that can be used to override the amount of rewards that appears in the selection, and REWARD_OVERRIDES that can override some or all of the rewards that appear.

I also, instead of having the override code in the place that uses it, put it in the override.ts file. I think once this is merged, something I will look to do is move any overriding code to overrides so that it can all be in one place rather than scattered throughout the codebase.

### Screenshots/Videos
Example shot showing it is the first wave, but I have overridden it to show 5 rewards and the first two rewards is the Acrobatics TM and Lum Berry.

## How to test the changes?
Go into overrides.ts and change the REWARD_AMOUNT_OVERRIDE and/or REWARD_OVERRIDES. Then win a battle and see the rewards use the values you specify.

## Checklist
- [x] Are the changes visual?


src/overrides.ts
```diff
@@ -105,3 +130,56 @@ export const OPP_MODIFIER_OVERRIDE: Array<ModifierOverride> = [];
...
+        // A tier must be specified otherwise the pokeball will just show an open great ball.
+        // Since this is an override for testing purposes, just use common.
+        overriddenModifier.setTier(ModifierTier.COMMON);
+        options[i] = new ModifierTypeOption(overriddenModifier, 0);
+      }
+    }
+  }
+
+  return options;
+}
```
bennybroseph, line 185: did you add functions to overrides.ts?

bennybroseph (changes requested):

src/overrides.ts
MPekelny: Are we not supposed to? It seems like it would be better if instead of the code that applies overrides be all over the place, all that code was contained in overrides.

bennybroseph: yeah, we edit overrides to debug. i don't wanna look at functions in that file.

MPekelny: That is very fair. I have moved those functions to a new overrides-functions.ts and added a comment to overrides to help make it clear for anyone else who might think to do the same.

I assume I am not supposed to resolve this conversation, you are if my change works for you, correct?

flx-sta:

-----

Issue #1315, opened 2024-05-24
Bug bite not removing opponents berries


RNGezeus:
Bug bite only copies the effect of the berries it targets, they are not actually getting stolen.

Edit: Can note that the two primary relevant cases this impacts is
A: you can infinitely buff from eating a single berry
B: You can't reduce targets for grip claw / Black hole when trying to steal items.

PR #232, opened 2024-04-22
Implemented Bug Bite, Pluck, Teatime

lucfd:
- Implemented Bug Bite, Pluck, Stuff Cheeks, and Teatime using two new move effect attributes:
  - `EatBerryAttr`: Makes the target eat a specified berry. If none is specified, a random berry is chosen from the ones the Pokemon is holding.
  - `StealEatBerryAttr`: Attempts to steal & eat a berry from the target of your attack.
- Stealing berries with `StealEatBerryAttr` (Pluck & Bug Bite) is blocked by `BlockItemTheftAbAttr` (Sticky Hold)
- Moves using these attributes are unaffected by Unnerve
- `EatBerryAttr` and `StealEatBerryAttr` benefit from Berry Pouch

PR #1532, opened 2024-05-29
[Bug] Adjusts bug bite/pluck functionality with three berry pouches

schmidtc1:
#1315 

## What are the changes?
Changes to the StealEatBerryAttr class to account for the user having three berry pouches (100% chance to keep berry after using it).  
The berry is first transferred to the user's held items before being selected to eat using the super class (EatBerryAttr). 

To fix the possibility of having an infinite berry supply from the opposing pokemon, as previously the berry would be kept on the opposing pokemon's held items.

Preventing an infinite berry farming glitch.

## What did change?
Removes previous logic from #232 which would simply set this.chosenBerry to a berry from the opposing pokemon's held items before eating it.
With the new logic, the berry is transferred into the user's pokemon's held items before being eaten.
I use the tryTransferHeldItemModifier to transfer into the user's pokemon's held items. I had to provide Promise<boolean> as a return option for this to work in both the StealEatBerryAttr class and the EatBerryAttr class.
I then grab the latest berry from the user's berries (length - 1). 

## How to test the changes?
Tested using the `src/overrides.ts` file. 
Recommend testing with no berry pouches vs. three berry pouches to see the effect.
No automated tests were used.

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -1532,7 +1532,7 @@ export class EatBerryAttr extends MoveEffectAttr {
  * @param args Unused
  * @returns {boolean} true if the function succeeds
  */
-  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean | Promise<boolean> {
     if (!super.apply(user, target, move, args)) {
       return false;
     }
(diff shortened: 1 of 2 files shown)


Tempo-anon:
There's another PR that will change 3 berry pouches to 90% (30% each)

Smewkie:
How does this function when the user already has maximum stacks of every berry, out of curiosity?

schmidtc1:
> How does this function when the user already has maximum stacks of every berry, out of curiosity?

With max stacks, the berry is not added to the user's inventory, or removed from the opponent's inventory, but still provides the effects of the berry and decrements the stack of the user's inventory. 
Example: User has 10 sitrus berries, opponent has 5. User uses Bug Bite, fails to transfer berry from opponent, consumes user's 10th sitrus berry. User ends with 9 sitrus berries, oppenent has 5 still. 

I'll look into this case.

PR #1553, opened 2024-05-29
[zh_CN localization] fixing text size of zh_CN Update text.ts

40chyan:
## What are the changes?
some fontsize has been fixed so that dialogues can be showed correctly. Chinese characters in some interface also looks better

### Screenshots/Videos
**before**
image
image
image
image

**after**
image
image
image
image
image


40chyan:
I will merge this PR into a new PR including font changes

PR #1554, opened 2024-05-29
[Ability]Added Perish Body ability

newmattock:
## What are the changes?
Added functionality for the Perish Body ability, applying the "Perish Song" effect to both Pokemon after the user is hit with a physical attacking move.

I introduced these changes because I thought Perish Body was a strategically interesting ability. My changes do not come from other PRs or previous issues, however, I believe it could improve the player experience by offering another ability that users could base their runs on, adding more variability.

## What did change?
I added functionality for the Perish Body ability, which was unimplemented before I added these changes.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/152253273/13979d87-63fd-4d9d-8e3b-982fbe31904b

## Checklist
- [x] Are the changes visual?


src/data/ability.ts
```diff
@@ -820,6 +820,30 @@ export class PostDefendContactDamageAbAttr extends PostDefendAbAttr {
...
+  constructor(turns: integer) {
+    super();
+
+    this.turns = turns;
+  }
+
+  applyPostDefend(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: PokemonMove, hitResult: HitResult, args: any[]): boolean {
+    if (move.getMove().checkFlag(MoveFlags.MAKES_CONTACT, attacker, pokemon)) {
+      attacker.addTag(BattlerTagType.PERISH_SONG, this.turns);
+      pokemon.addTag(BattlerTagType.PERISH_SONG, this.turns);
```
Tempo-anon, line 835: Bulbapedia says
`When the Pokémon with Perish Body is hit with a [contact](https://bulbapedia.bulbagarden.net/wiki/Contact) move, both the attacker and the Pokémon with Perish Body will faint in three turns. If the attacker already has a perish count or is wearing [Protective Pads](https://bulbapedia.bulbagarden.net/wiki/Protective_Pads) or has [Long Reach](https://bulbapedia.bulbagarden.net/wiki/Long_Reach_(Ability)), the Ability does not affect anyone.`

Tempo-anon (changes requested):
1. Needs documentation
2. Fix linting issues
3. Functionality needs to check if one of the pokemon is already being affected by perish song

newmattock:
Going to try to add the documentation and the functionality for if the attacker has perish song already by today. Let me know if there need to be any more changes afterward!

Tempo-anon:
The ability should do nothing if either the attacker OR defender are already under the effects of perish song. Otherwise seems fine to me

src/data/ability.ts
```diff
@@ -819,6 +819,42 @@ export class PostDefendContactDamageAbAttr extends PostDefendAbAttr {
     return getPokemonMessage(pokemon, `'s ${abilityName}\nhurt its attacker!`);
   }
 }
+/**
+ * @description: This ability applies the Perish Song tag the attacking pokemon
```
Tempo-anon, line 823: nit: You're missing a word `to` here

src/data/ability.ts
```diff
@@ -819,6 +819,42 @@ export class PostDefendContactDamageAbAttr extends PostDefendAbAttr {
...
+    this.turns = turns;
+  }
+
+  applyPostDefend(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: PokemonMove, hitResult: HitResult, args: any[]): boolean {
+    // Checking for physical contact
+    if (move.getMove().checkFlag(MoveFlags.MAKES_CONTACT, attacker, pokemon)) {
+      // Checking for if either of the pokemon already has the Perish Song tag
+      if (pokemon.getTag(BattlerTagType.PERISH_SONG) || attacker.getTag(BattlerTagType.PERISH_SONG)) {
+        return false;
+      // Applying the Perish Song tag to both pokemon
```
Tempo-anon, line 844: nit: I think the code here is clear enough that there's no need for comments

src/data/ability.ts
```diff
@@ -819,6 +819,42 @@ export class PostDefendContactDamageAbAttr extends PostDefendAbAttr {
...
+        attacker.addTag(BattlerTagType.PERISH_SONG, this.turns);
+        pokemon.addTag(BattlerTagType.PERISH_SONG, this.turns);
+        return true;
+      }
+    }
+    return false;
+  }
+
+  getTriggerMessage(pokemon: Pokemon, abilityName: string, ...args: any[]): string {
+    return getPokemonMessage(pokemon, `'s ${abilityName}\n will faint both pokemon in ${this.turns - 1} turns!`);
```
Tempo-anon, line 855: Prob best practice to localize this https://github.com/pagefaultgames/pokerogue/blob/main/src/locales/en/ability-trigger.ts

Tempo-anon (commented):
Looks mostly good to me. Just some minor nitpicks

src/data/ability.ts
newmattock: working on it right now, thank you!

Issue #1284, opened 2024-05-23
[SUGGESTION] Keep PP of Fusion Pokemon


pedwards95:
Currently any PPUP or PPMax used on the 2nd pokemon's moves are wasted, but the first pokemon's moves are kept. This should be consistent between the two pokemon -- either they both lose their PPUPs or neither do.

PR #1556, opened 2024-05-29
[Enhancement] Preserve move PP when fusing

knispeja:
## What are the changes?

Move PP is now preserved when learning any moves that come from fusion. This is reflected on the move summary UI. Additionally, for the edge case when both members of the fusion have the same move known, the higher PP is used.

Issue: https://github.com/pagefaultgames/pokerogue/issues/1284
I thought this issue was a good idea -- fusion order and what is retained is already confusing, I think it's best to limit that confusion where possible. It's also just nice for the player not to lose PP for "no reason".

## What did change?

Added a PPUp override value that can be passed through into `LearnMovePhase` as well as `Pokemon.setMove()` that allows for moves to begin with non-zero values for `ppUp`. Provided this override value to the fusion move learning phase in `pokemon.ts`. I also moved PP calculation into a static method so PP could be calculated with PP Ups taken into account when a move hasn't been learned yet.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/9343160/d20ccfcb-bfca-4654-abf0-a7d259a8fba6

Note: to clarify the video above, Sand Attack has had a PP Max used on it, and that carries over during fusion.

## How to test the changes?

Increase the drop rate of PP Up, PP Max, and DNA Splicer. Try different combinations -- there's a particular edge case when both know the same move with different PP.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/enums/moves.ts
+++ b/src/data/enums/moves.ts
@@ -1,5 +1,5 @@
 export enum Moves {
-  /**{@link https://bulbapedia.bulbagarden.net/wiki/None_(move) | Source} */
+  /** Represents an empty slot in the moveset */
   NONE,
   /**{@link https://bulbapedia.bulbagarden.net/wiki/Pound_(move) | Source} */
   POUND,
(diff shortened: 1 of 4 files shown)


flx-sta:

-----

PR #1561, opened 2024-05-29
[Feature] Add possibility to override foe level

hayuna:
## What are the changes?
Add more flexibility in pokemon overrides

Adding possibility to override foe level to have more flexibility in testing different things

## What did change?
Added new option in Overrides to hardcoded foe level

### Screenshots/Videos
Screenshot 2024-05-29 at 15 41 38

High wave, low levels
Screenshot 2024-05-29 at 18 27 31

## How to test the changes?
You can just set number for variable OPP_LEVEL_OVERRIDE
`export const OPP_LEVEL_OVERRIDE: number | null = 100;`

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/overrides.ts
+++ b/src/overrides.ts
@@ -79,6 +79,7 @@ export const OPP_GENDER_OVERRIDE: Gender = null;
 export const OPP_MOVESET_OVERRIDE: Array<Moves> = [];
 export const OPP_SHINY_OVERRIDE: boolean = false;
 export const OPP_VARIANT_OVERRIDE: Variant = 0;
+export const OPP_LEVEL_OVERRIDE: number | null = null;
 
 /**
  * MODIFIER / ITEM OVERRIDES
--- a/src/overrides.ts
+++ b/src/overrides.ts
@@ -79,7 +79,7 @@ export const OPP_GENDER_OVERRIDE: Gender = null;
 export const OPP_MOVESET_OVERRIDE: Array<Moves> = [];
 export const OPP_SHINY_OVERRIDE: boolean = false;
 export const OPP_VARIANT_OVERRIDE: Variant = 0;
-export const OPP_LEVEL_OVERRIDE: number | null = null;
+export const OPP_LEVEL_OVERRIDE: number = 0;
 
 /**
  * MODIFIER / ITEM OVERRIDES
--- a/src/overrides.ts
+++ b/src/overrides.ts
@@ -73,13 +73,13 @@ export const VARIANT_OVERRIDE: Variant = 0;
  */
 
 export const OPP_SPECIES_OVERRIDE: Species | integer = 0;
+export const OPP_LEVEL_OVERRIDE: number = 0;
 export const OPP_ABILITY_OVERRIDE: Abilities = Abilities.NONE;
 export const OPP_PASSIVE_ABILITY_OVERRIDE = Abilities.NONE;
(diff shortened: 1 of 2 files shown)


src/overrides.ts
```diff
@@ -79,6 +79,7 @@ export const OPP_GENDER_OVERRIDE: Gender = null;
 export const OPP_MOVESET_OVERRIDE: Array<Moves> = [];
 export const OPP_SHINY_OVERRIDE: boolean = false;
 export const OPP_VARIANT_OVERRIDE: Variant = 0;
+export const OPP_LEVEL_OVERRIDE: number | null = null;
```
Tempo-anon, line 82: nit: I'd move this up to be in between OPP_SPECIES_OVERRIDE and OPP_ABILITY_OVERRIDE to match with how STARTING_LEVEL_OVERRIDE is

Tempo-anon: Also maybe an `integer: 0` because STARTING_LEVEL_OVERRIDE and OPP_LEVEL_OVERRIDE should match

Tempo-anon (commented):
Looks mostly good to me, just some tiny nitpicks. If you can also include a screenshot of a low level opponent at a very high wave that'd be great

flx-sta (commented):
Why is the species overwritten before calling `const pokemon = new EnemyPokemon...` but level & gender afterwards?

src/overrides.ts
hayuna: Replaced null with 0 as init value

hayuna:
@flx-sta 
Screenshot 2024-05-29 at 18 33 20

It would be totally the same

flx-sta:
> @flx-sta Screenshot 2024-05-29 at 18 33 20
> 
> It would be totally the same. There is some good practice to not override function parameters that why I used `pokemon.level`

I agree with the clean-code thing.
I'd just adapt the species to it too i think. To keep it the same

hayuna:
> > @flx-sta Screenshot 2024-05-29 at 18 33 20
> > It would be totally the same. There is some good practice to not override function parameters that why I used `pokemon.level`
> 
> I agree with the clean-code thing. I'd just adapt the species to it too i think. To keep it the same

Ok, could it be, but im not touching code which is not related to my changes

flx-sta (approved):

src/overrides.ts
```diff
@@ -73,6 +73,7 @@ export const VARIANT_OVERRIDE: Variant = 0;
  */
 
 export const OPP_SPECIES_OVERRIDE: Species | integer = 0;
+export const OPP_LEVEL_OVERRIDE: number = 0;
```
Tempo-anon, line 76: `number` instead of `integer`? I feel like it should match `export const STARTING_LEVEL_OVERRIDE: integer = 0;`

hayuna: There is no `integer` type in typescript, its number. To fix this incorrect typing typescript team created alias 

`type integer = number`



In TypeScript, there is no equivalent to integer or float, all numbers are floating-point values. TypeScript supports the number literals for decimal, hexadecimal, binary, and octal literals. The type of these values and literals ​​is simply a `number`.

flx-sta: @hayuna lol and I was wondering what the heck is going on with all these `integer` types.
I thought it's some fancy stuff.

PR #1564, opened 2024-05-29
Fix move.ts Errors

bennybroseph:
## What are the changes?

There are no user-facing changes

This change fixes ~4 errors in move.ts

## What did change?

Removes an unnecessary toString() call, an invalid return value, and a function call that was missing an unused argument

## How to test the changes?

Since this change deals with localization for the largest change, you can simply place a breakpoint in the changed line and observe on game start that each move still translates properly in the appropriate language. 

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -144,8 +144,8 @@ export default class Move implements Localizable {
   localize(): void {
     const i18nKey = Moves[this.id].split("_").filter(f => f).map((f, i) => i ? `${f[0]}${f.slice(1).toLowerCase()}` : f.toLowerCase()).join("") as unknown as string;
 
-    this.name = this.id ? `${i18next.t(`move:${i18nKey}.name`).toString()}${this.nameAppend}` : "";
-    this.effect = this.id ? `${i18next.t(`move:${i18nKey}.effect`).toString()}${this.nameAppend}` : "";
+    this.name = this.id ? `${i18next.t(`move:${i18nKey}.name`)}${this.nameAppend}` : "";
+    this.effect = this.id ? `${i18next.t(`move:${i18nKey}.effect`)}${this.nameAppend}` : "";
   }
 
   getAttrs(attrType: { new(...args: any[]): MoveAttr }): MoveAttr[] {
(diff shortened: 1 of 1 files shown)


Tempo-anon (approved):

src/data/move.ts
```diff
@@ -2011,10 +2011,10 @@ export class PostVictoryStatChangeAttr extends MoveAttr {
   }
   applyPostVictory(user: Pokemon, target: Pokemon, move: Move): void {
     if (this.condition && !this.condition(user, target, move)) {
-      return false;
+      return;
     }
     const statChangeAttr = new StatChangeAttr(this.stats, this.levels, this.showMessage);
-    statChangeAttr.apply(user, target, move);
+    statChangeAttr.apply(user, target, move, undefined);
```
flx-sta, line 2017: `apply()` expects an any-array for the `args` paramater thus it is not complaining about `undefined`.
I think it's cleaner to pass an empty array:
```js
statChangeAttr.apply(user, target, move, []);
```

flx-sta (changes requested):

src/data/move.ts
bennybroseph: does this make any sense as well? `<any>[]`

flx-sta: @bennybroseph I'm not sure what you are trying to say? Please elaborate

bennybroseph:
```ts
statChangeAttr.apply(user, target, move, <any[]>[]);
```

flx-sta: @bennybroseph no. you shouldn't need any type-casting. simply the way I suggested should work (or doesn't it?)

```js
statChangeAttr.apply(user, target, move, []);
```

bennybroseph: they all work, just asking which feels the best. none produce an error or cause an issue

flx-sta: Okay. the type casting doesn't make any sense then. using `any` as a type is usually not a permanent state/solution anyway as it just swallows whatever you throw at it. Stick to the empty array I'd say

bennybroseph: actually i just realized. an empty array creates memory for no reason. doing `[]` may look better, but I don't see any reason to do it since in both cases the variable is never referenced or used. it's better to just leave it as `undefined` and not create an array for nothing.

flx-sta: huh. you learn something new every day. Thanks for letting me know.

bennybroseph: yeah, it's something i always forget in these super relaxed languages since you hardly ever have to type `new` to create memory. I'm sure it's not a big deal to do so, but whenever we're thinking about it we should try to avoid stuff like that. I'm unsure how good of a garbage collector TypeScript has.

PR #1569, opened 2024-05-29
[Feature] Move touch controls configuration

TimPerdok:
## What are the changes?

I've added a setting that opens a configuration panel to move the touch controls around the screen. This setting is only visible when the client has a touch screen. This setting does not use option values like the other settings, but instead opens the configuration panel using the action button. Both landscape mode and portrait mode have their own control positions saved separately.

I implemented these changes to improve the user experience by offering greater flexibility in the placement of the touch controls. This way the game is more accessible and comfortable to play on mobile devices. 

## What did change?

Most changes are inside a new singleton class (MoveTouchControlsHandler). This class is responsible for maintaining the configuration panel for the touch controls. Aside from this the CSS styling and HTML structure of the touch controls had to be changed to allow for these configuration options. Finally a new setting is added that can be activated using the action button that opens up the configuration panel.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/43152351/9e4f2937-5c8a-45bd-a071-52a86272a10a

## How to test the changes?

Open the branch on a mobile device or a simulated mobile device using browser dev tools. Then navigate to the new setting on the bottom. This setting only appears when the client has a touch screen. By using the action button on the setting, a new panel opens up where you can drag the touch controls around the screen to reconfigure their position.


joop339:
Thank you! Works like a charm!

brain-frog:
Settings got reworked so this will need to be updated to work with that, but I like this idea so please let me know when the merge conflicts are resolved and I'll review

TimPerdok:
@brain-frog I have fixed the merge conflicts. Let me hear if you find anything.

index.css
```diff
@@ -31,145 +23,256 @@ body {
...
+  width: var(--control-group-extra-size);
+  height: var(--control-group-extra-size);
 }
+/* Hide buttons on specific UIs */
 
-#apad > * {
-  width: var(--controls-size);
-  height: var(--controls-size);
+/* Show #cycleForm and #cycleShiny only on STARTER_SELECT and SETTINGS */
+#touchControls:not([data-ui-mode='STARTER_SELECT']):not([data-ui-mode='SETTINGS']):not([data-ui-mode='SETTINGS_ACCESSIBILITY']):not([data-ui-mode='SETTINGS_GAMEPAD']):not([data-ui-mode='SETTINGS_KEYBOARD']) #apadCycleForm,
```
brain-frog, line 110: please update the second not to be
`:not([data-ui-mode^='SETTINGS'])` 
so it captures all current and future SETTINGS modes

index.html
```diff
@@ -50,55 +50,84 @@
...
-	<div id="touchControls">
-		<div id="dpad" class="unselectable">
-			<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 72 72">
-				<path id="dpadUp" data-key="UP" d="M48,5.8C48,2.5,45.4,0,42,0H29.9C26.6,0,24,2.4,24,5.8V24h24V5.8z" />
-				<path id="dpadRight" data-key="RIGHT" d="M66.2,24H48v24h18.2c3.3,0,5.8-2.7,5.8-6V29.9C72,26.5,69.5,24,66.2,24z" />
-				<path id="dpadDown" data-key="DOWN" d="M24,66.3c0,3.3,2.6,5.7,5.9,5.7H42c3.3,0,6-2.4,6-5.7V48H24V66.3z" />
-				<path id="dpadLeft" data-key="LEFT" d="M5.7,24C2.4,24,0,26.5,0,29.9V42c0,3.3,2.3,6,5.7,6H24V24H5.7z" />
-				<rect id="dpadCenter" x="24" y="24" width="24" height="24" />
-			</svg>
+	<div id="touchControls" class="touchControls">
```
brain-frog, line 53: what is the reasoning behind it being both an id and a class

src/ui/settings/settings-accessiblity-ui-handler.ts
```diff
@@ -14,7 +14,7 @@ export default class SettingsAccessibilityUiHandler extends AbstractSettingsUiHa
   constructor(scene: BattleScene, mode?: Mode) {
     super(scene, mode);
     this.title = "Accessibility";
-    this.settings = Setting.filter(s => s.type === SettingType.ACCESSIBILITY);
+    this.settings = Setting.filter(s => s.type === SettingType.ACCESSIBILITY && !s.hidden?.());
```
brain-frog, line 17: if it should be done every time it should probably be handled in the abstract class

src/ui/settings/touch/move-touch-controls-handler.ts
```diff
@@ -0,0 +1,347 @@
...
+  /**
+   * Sets a control element to the given position.
+   * The x values are either offsets to the left or right boundary of the screen, depending on the side of the element.
+   * E.g. For left elements, (0, 0) is the bottom left corner of the screen and
+   * for right elements, (0, 0) is the bottom right corner of the screen.
+   * @param controlElement
+   * @param x Either an offset to the left or right boundary of the screen.
+   * @param y An offset to the bottom boundary of the screen.
+   */
+  private setPosition(controlElement: HTMLElement, x: number, y: number) {
```
brain-frog, line 230: is there any check making sure users dont overlap buttons? I'd like to avoid buttons being pressed at the same time or lost under one another since this can lead to exploits easier

brain-frog (changes requested):

src/ui/settings/touch/move-touch-controls-handler.ts
TimPerdok: I don't think this would cause issues, as I expect that Phaser handles the inputs sequentially. This would be the same for users pressing keyboard keys simultaneously. Aside from this, the controls are all HTML, which means inputs can easily be modified by users. Obscuring that will have little advantage.

index.html
TimPerdok: None, changed.

index.css
TimPerdok: Good idea

src/system/settings/settings.ts
```diff
@@ -252,6 +257,23 @@ export const Setting: Array<Setting> = [
     options: AUTO_DISABLED,
     default: 0,
     type: SettingType.GENERAL
+  },
+  {
+    key: SettingKeys.Language,
```
Greenlamp2, line 262: look like a bad copy/paste here

src/touch-controls.ts
```diff
@@ -8,14 +8,35 @@ export const keysDown = new Map<string, string>();
...
+
 /**
  * Initialize touch controls by binding keys to buttons.
  *
  * @param events - The event emitter for handling input events.
  */
 export function initTouchControls(events: EventEmitter): void {
   preventElementZoom(document.querySelector("#dpad"));
-  preventElementZoom(document.querySelector("#apad"));
+  document.querySelectorAll(".apadBtn").forEach((element) => preventElementZoom(element as HTMLElement));
```
Greenlamp2, line 39: Did you test if this line of code crash when there is no Dom ? like in tests run for example

TimPerdok: All the tests succeed. querySelectorAll always returns a NodeList. So if an element does not exist, it would just skip the line.

Tempo-anon:
Closing this because of merge conflicts and it's been a while. If this is still needed someone can open a PR for this feature

PR #1574, opened 2024-05-29
[Item] Add Toxic Orb and Flame Orb

xsn34kzx:
## What are the changes?

Completely adds and implements the following items: [`Toxic Orb`](https://bulbapedia.bulbagarden.net/wiki/Toxic_Orb) and [`Flame Orb`](https://bulbapedia.bulbagarden.net/wiki/Flame_Orb).

There was no active issue or PR that brought up the incorporation of these items, but it was brought to my attention that, as the game currently sits, a large chunk of Pokemon cannot reliably use passive and regular abilities associated with retaining status conditions due to the map transitions forcing a full party heal. As a result, players end up being forced to use status effect moves in double battles on themselves or are left to the chance of being hit with a status effect in a wild or trainer encounter. As such, this aims to make those abilities have a larger impact in terms of viability and overall consistent usability.

## What did change?

The item internal functionality was achieved by creating a new class `TurnStatusEffectModifier` extending `PokemonHeldItemModifier` in `src/modifier/modifier.ts` and adding a line to `TurnEndPhase`'s `start` function in `src/phases.ts` to apply the new modifier towards the latter half of the function, effectively activating the modifier after berry and ability cures. 

Subject-to-change weight functions were included for the items to make them appear more often when abilities that benefit from status conditions are present within the party at the time of reward rolls. In particular, `Toxic Orb` gets a higher weighting when any of the following abilities are present in the party: [`Marvel Scale`](https://bulbapedia.bulbagarden.net/wiki/Marvel_Scale_(Ability)), [`Guts`](https://bulbapedia.bulbagarden.net/wiki/Guts_(Ability)), [`Quick Feet`](https://bulbapedia.bulbagarden.net/wiki/Quick_Feet_(Ability)), [`Toxic Boost`](https://bulbapedia.bulbagarden.net/wiki/Toxic_Boost_(Ability)), and [`Poison Heal`](https://bulbapedia.bulbagarden.net/wiki/Poison_Heal_(Ability)). Likewise, `Flame Orb` gets a higher weighting for the following abilities: [`Marvel Scale`](https://bulbapedia.bulbagarden.net/wiki/Marvel_Scale_(Ability)), [`Guts`](https://bulbapedia.bulbagarden.net/wiki/Guts_(Ability)), [`Quick Feet`](https://bulbapedia.bulbagarden.net/wiki/Quick_Feet_(Ability)), and [`Flare Boost`](https://bulbapedia.bulbagarden.net/wiki/Flare_Boost_(Ability)). This behavior and the decision to gauge these items as `Ultra` tier were suggested by balance team member @Blitz425, which is reflected in `src/modifier/modifier-types.ts`. 

The English name and description for both items were added to the corresponding localization file for modifiers. 

Sprites for both items were included as individual `.png` files and stitched to the current `items.png` texture atlas in `public/images` with values for the texture positions in the associated `.json` file. These sprites were sourced from [Black/White 2](https://www.spriters-resource.com/ds_dsi/pokemonblack2white2/sheet/48051/).

### Screenshots/Videos

Note that these items were not partially added prior to this PR. Likewise, note that the following demos temporarily modified the common reward pool to give `Toxic Orb` and `Flame Orb` more frequently for demonstration purposes; these modifications are not present in the PR.

#### Toxic Orb Video Demo

https://github.com/pagefaultgames/pokerogue/assets/109637146/074a012a-3ddd-4d0e-8d7f-2605a3fb28d6

Demonstrates the reward process of getting `Toxic Orb`, how it applies its status effect, and, as a bonus, how it interacts with `Shed Skin`, which mirrors how it behaves with curing berries as well.

#### Flame Orb Video Demo

https://github.com/pagefaultgames/pokerogue/assets/109637146/edd25946-0d38-4a6a-b19d-8700542ceafa

Demonstrates the same situations as the `Toxic Orb` demo and also shows that only one of the same orb can be held.

#### Different Orb Video Demo

For context, a concern for balance and usability was having more than one of two different types of orbs. Since modifiers are applied linearly, only one orb would be active at a time; that is, by the time the next orb would apply, a status condition would already be there, and it would fail. This isn't necessarily game-breaking or out of line with the intended order of operations, but it was discussed and concluded that stockpiling orbs should not be allowed. The solution to this was commented on in the `TurnStatusEffectModifier` class since it makes use of the `matchType` function.

https://github.com/pagefaultgames/pokerogue/assets/109637146/4aec9c18-93aa-48b8-97b3-84aec50d6fed

Demonstrates that, if one type of orb is held, another cannot be applied.

#### Wave Heal Video Demo

https://github.com/pagefaultgames/pokerogue/assets/109637146/aa6b6acb-dd11-4b32-a9db-be792396d781

Demonstrates that the orbs will activate the moment a turn ends after a full party heal is performed, showing that users can now carry these status conditions across the game as long as an orb is held.

## How to test the changes?

For general testing related to item functionality, the `src/overrides.ts` file was modified such that the `STARTING_HELD_ITEMS_OVERRIDE` array included an entry for either `Toxic Orb`, `{ name: "TOXIC_ORB" }` or `Flame Orb`, `{ name: "FLAME_ORB" }`. 

The `ABILITY_OVERRIDE` was set to `Abilities.SHED_SKIN` in particular for testing the interaction between cures and the items.

The aforementioned rewards pool testing was done by modifying the pools directly since there is not yet an override for post-battle loot. That being said, actual testing of the tentative weight was not frequent; that is, the weight function works, but re-rolling to actually get the items in the user context took way too long.

### Comments

It's worth noting again that balance is mostly out of scope for the PR solely, so any discussion to where these items should sit as far as reward pooling and weight is encouraged; more than willing to adjust this as necessary.

Lastly, I wasn't sure if the anonymous functions used for the weight functions were too long or didn't utilize a TypeScript shortcut to make them a tad more readable and compact. They're also very similar, but there wasn't precedence for defining weighting functions separately, so I pretty much followed the formatting of other items. 

## Checklist
- [X] Are the changes visual?


flx-sta:
Judging by the code I'm not sure if you covered the `if no other status is already present` part.
You aren't exactly checking if the Pokemon already has a status, you are just running `trySetStatus()` but from reading the code of that method it's not covering that either

xsn34kzx:
> Judging by the code I'm not sure if you covered the `if no other status is already present` part. You aren't exactly checking if the Pokemon already has a status, you are just running `trySetStatus()` but from reading the code of that method it's not covering that either

I actually thought the same thing initially (I was going to write a `shouldApply` override to call `canSetStatus` for validation), but `trySetStatus` already calls `canSetStatus` in the very beginning, so I could avoid the redundant call. I think if `trySetStatus` didn't catch it, the status effect applied message would run every turn, which doesn't happen in the demos.

flx-sta:
@xsn34kzx sooo what if the mon is burning alread, should it still get poisoned too?

xsn34kzx:
> @xsn34kzx sooo what if the mon is burning alread, should it still get poisoned too?

It should lead to [this line](https://github.com/xsn34kzx/pokerogue/blob/d59625e6612523b1ad4379ff434ff4899a27d9c9/src/field/pokemon.ts#L2288), which basically checks if `overrideStatus` is true or not. Since it's not in this case, it'll default to checking if `this.status` is set to a value or not, which, if it is, will return false and `canSetStatus` will effectively stop `trySetStatus` from doing anything. To double check though, I tested the scenario and recorded it.

https://github.com/pagefaultgames/pokerogue/assets/109637146/5490d5f5-881e-4b1e-8943-500dc609ae38

Here I basically gave Squirtle a status effect, gave it Toxic Orb to see if it would activate at the end of a turn, and then gave it back to Charmander to make sure it did activate on someone without a pre-existing status effect at the end of a turn.

src/modifier/modifier.ts
```diff
@@ -851,6 +851,55 @@ export class TurnHealModifier extends PokemonHeldItemModifier {
...
+   * @param args [0] {@linkcode Pokemon} that holds the held item
+   * @returns true if the status effect was applied successfully, false if
+   * otherwise
+   */
+  apply(args: any[]): boolean {
+    return (args[0] as Pokemon).trySetStatus(this.effect, true, undefined, undefined, this.type.name);
+  }
+
+  getMaxHeldItemCount(pokemon: Pokemon): integer {
+    return 1;
```
flx-sta, line 899: I’m curious if this one why it’s a getter method. Is it going to change any time soon ?

flx-sta (approved):

src/modifier/modifier.ts
xsn34kzx: Since it extends `PokemonHeldItemModifier`, it has to implement it, but I mostly just followed the format of `TurnHealModifier`, which does something similar by returning a constant.

xsn34kzx:
With some discussion, the weight functions for both items were adjusted to dynamically provide different weights for the items based on certain conditions of the party. The functions generally act as follows for both items:
1. Existing orb holders are filtered out of the party and no weight is provided (0). 
2. From the filtered list, each member is checked for the specific ability that benefits from that orb's status effect specifically. For Toxic Orb, this would be `Toxic Boost` and `Poison Heal`. For Flame Orb, this would be `Flare Boost`. If found for any party member, the maximum tentative weight is given (4).
3. If no member is found in the previous step, each member is checked against the list of abilities that benefit from status effects generally. The abilities in this list are `Marvel Scale`, `Guts`, and `Quick Feet`. If found for any party member, the next tentative weight is given (2).
4. If no member is found in the previous step, each member is checked against the list of moves that have more general utility with the either orb. The moves in this list are `Facade`, `Trick`, `Fling`, `Switcheroo`, and `Psycho Shift`. If found for any party member, the next tentative weight is given (1).
5. Whatever the value of the tentative weight is by the end, this value is multiplied by # of waves / 15 (increased multiplier for every 15 waves). The resulting value from this maxes out at a weight of 32 in the `Ultra` tier reward pool.

I definitely like the idea of weighing based on most utility to no utility and having the lower weighted "tiers" scaling based on how many waves there are. This means that, even if you never have a relevant ability in a run, having moves will still get a great chance at rolling the item during that run as you get to later waves. As always though, the tentative weights at each "tier" are flexible and so is the maximum weight, so balance can be done on both levels of the function past the PR if necessary.

bennybroseph:
If you can't swap this to an enemy somehow, this feels very incomplete. I feel like we need a system to mark an item as a temporary steal or something first before we add something like this. Having a new item for only one ability and move in the whole game feels unneeded to me. Who all uses this at the moment?

Blitz425:
> If you can't swap this to an enemy somehow, this feels very incomplete. I feel like we need a system to mark an item as a temporary steal or something first before we add something like this. Having a new item for only one ability category and move in the whole game feels unneeded to me. Who all uses this at the moment?

Saying this is unneeded is wholefully untrue. You can look up the distribution of these abilities listed in the PR yourself and see this affects a wide range of Pokemon who have completely dud abilities otherwise. The orb exists in the mainline games for a reason? This also affects passives since the balance team have given out those abilities for strategies and saying this is unnecessary just kills those strategies that we made. The weighting of the item is also already dependent on moves that do what you say in swapping them to an enemy, but as of right now there are so many pokemon who just do not have a proper ability and this item has been necessary and heavily requested for the longest time.

xsn34kzx:
>     1. Existing orb holders are filtered out of the party and no weight is provided (0).
This stage of the weighting functions for both orbs has been adjusted to also exclude any party members that have the abilities that benefit from the specific status effect of the opposite orb. This addresses the edge case where a member, specifically Breloom, has the passive ability `Guts` but has the ability `Poison Heal`. Prior to the latest commit, this would lead to a chance for both Toxic Orb and Flame Orb to appear, where the former would have a higher weight (and is generally more preferred) but the latter would still have a decently high weight in the roll. Now, in the Flame Orb weight calculation, the member is not counted; meaning that, if that is the only party member with the relevant abilities and/or moves, the weight of Flame Orb will be 0, putting roll preference on Toxic Orb entirely due to the presence of `Poison Heal`. Note that this doesn't change the weighting behavior overall; the presence of a general status effect ability and/or utility move and not a specific status effect ability will allow both Toxic Orb and Flame Orb to be rolled.

flx-sta (commented):
Instead of using 
```ts
!!filteredParty.filter(p => p.getMoveset().filter(m => moveList.includes(m.moveId))).length
```

try using
```ts
filteredParty.some(p => p.getMoveset().filter(m => moveList.includes(m.moveId)))
```

this has two benefits:
- some already returns a bool
- some stops checking if one conditions was found true

Sauce 🍝 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

xsn34kzx:
Suggested by @InnocentGameDev, each "tier" in talked about [here](https://github.com/pagefaultgames/pokerogue/pull/1574#issuecomment-2138695589) will now cap out at a maximum weight of 32, 16, and 8, keeping different conditions from eventually coalescing at a max weight of 32.

xsn34kzx:
> Instead of using
> 
> ```ts
> !!filteredParty.filter(p => p.getMoveset().filter(m => moveList.includes(m.moveId))).length
> ```
> 
> try using
> 
> ```ts
> filteredParty.some(p => p.getMoveset().filter(m => moveList.includes(m.moveId)))
> ```
> 
> this has two benefits:
> 
>     * some already returns a bool
> 
>     * some stops checking if one conditions was found true
> 
> 
> Sauce 🍝 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

Ooo, this is the exact use case, I'll make the change now 👍🏼

flx-sta (approved):

xsn34kzx:
Party filter for weighting has now been adjusted to also make sure that status effect applicability was checked for edge cases like Flareon with `Guts` being weighed for Flame Orb when it can't use it. Additionally, because `canSetStatus` has a case where having the same status effect being checked returns false, this was covered outside of the function to avoid the edge case where, for example, being burned or poisoned would filter members out of the party even if they had the listed abilities/moves, which would seemingly punish players for having the status effect, self-inflicted or otherwise.

xsn34kzx:
Addressing the latest commits: 
- For the first two, it was observed that, due to the constant application of status effects while having any orb, `Full Heal` and `Full Restore` would get weighted higher in reward pools since they both increase in weight when a party member has a status condition. As such, both weighting functions have been adjusted to check if the member also has either orb and if the status condition they currently have matches the intended status of the orb. The latter distinction was made to make sure that if the player had an unwanted status condition while still holding an orb, both curing items would count that member and increase weight accordingly. Otherwise, the member is ignored during the weighting process since players, by taking either orb, wouldn't want to cure the status condition at all.
- For the latest one, a bug was found when the game reloaded the modifiers in play during a session. Because of the way `PokemonHeldItemModifierType`s are handled, extra parameters in the constructor of a modifier are not properly set (see [`newModifier`](https://github.com/xsn34kzx/pokerogue/blob/2de077cd44bfb9b670b985f244ed1eb35707c600/src/modifier/modifier-type.ts#L221) and [`toModifier`](https://github.com/xsn34kzx/pokerogue/blob/2de077cd44bfb9b670b985f244ed1eb35707c600/src/system/modifier-data.ts#L47)). As such, changes were made to change intended status effect based on `ModifierType`'s `id`, which effectively does the same thing anyway by setting the appropriate status effect in the modifier's constructor. After this was changed, the appropriate status applied after reloading a run.

bennybroseph:
once this is ready for review, please let us know

bennybroseph (approved):

PR #1578, opened 2024-05-29
Move DamagePhase to separated file

hayuna:
Refactor to reduce lines of code in phases.ts

## What are the changes?
Move phases to separated files

Reducing code in one huge file

## What did change?
Moving DamagePhase to separated file

## Checklist
- [ ] Are the changes visual?


hayuna:
If this refactor will be acctepted and merged, I will move more files as well

flx-sta:
Since this is tech-debt fixes, would it make sense to add some test-coverage too?
_I don't have an idea how much test tooling is already there to get going. I do know there is some existing tests_

hayuna:
@flx-sta I dont know how to test this. If there would be at least example of testing that classes, would be easier to do this.

Tempo-anon (dismissed):
Looks good to me but I think this should get some opinions on the lead devs

XGSleepWalker (changes requested):
I would like this to stay on hold until we have the testing framework done.

flx-sta:
> I would like this to stay on hold until we have the testing framework done.

Is there a PR/issue and/or ETA of the testing framework?
The one I see is vitest. Is it being ditched?

hayuna:
Ok, we will see the progress of vitest and testing framework. But in my opinion is better to keep files more simpler, it would be even easier to future testing

flx-sta:
@hayuna nobody is disagreeing with that.
the reasons it's postponed is that this is a big code-refactoring and you  never know which accidental bugs you introduce that way. Thus some test-coverage necessary

src/phases/DamagePhase.ts
```diff
@@ -0,0 +1,108 @@
...
+import { battleSpecDialogue } from "#app/data/dialogue";
+import { SpeciesFormChangeManualTrigger } from "#app/data/pokemon-forms";
+import { BattleSpec } from "#app/enums/battle-spec";
+import { DamageResult, EnemyPokemon, HitResult } from "#app/field/pokemon";
+import { PersistentModifier } from "#app/modifier/modifier";
+import { getModifierType, modifierTypes } from "#app/modifier/modifier-type";
+import { PokemonPhase, SummonPhase, ToggleDoublePositionPhase } from "#app/phases";
+import * as Utils from "#app/utils";
+
+export class DamagePhase extends PokemonPhase {
```
ben-lear, line 12: Nit: Seeing as all other files in the project are in kebab case, can the file be renamed to “damage-phase.ts” instead of “DamagePhase.ts”

flx-sta: I disagree. This is a class file and it’s also the beginning of a big refactoring.

Tempo-anon:
Closing for now since it seems like it might be a while until we break apart phases even though it's planned

PR #780, opened 2024-05-12
Implement Parental Bond

jaimefd:
Implemented the ability Parental Bond, with a new ExtraHitMoveAbAttr that takes two values, the amount of extra hits and the damage modifier for them (in this case 1 and 0.25).

From my testing it seems to work like original Pokemon, with the exception that I intentionally allowed it to activate with Multi-Hit moves, so that it doesn't become redundant with Multi Lens and so that Maushold makes sense with Parental Bond as passive (otherwise it would be a useless passive). This can be changed easily if needed.

Tested that non-attacking moves don't activate it, misses also stop the second hit, and it procs secondary effects as expected.

PR #1579, opened 2024-05-29
Parental Bond Implementation v2

frutescens:
Implementing Parental Bond with the edge cases mentioned in the previous PR here: https://github.com/pagefaultgames/pokerogue/pull/780 

Pre-Hit cases will end up in ability.ts
https://bulbapedia.bulbagarden.net/wiki/Parental_Bond_(Ability)

During-Hit / Post-Hit cases will likely end up in phases.ts


bennybroseph:
Please create this using the template

PR #1399, opened 2024-05-26
Add new Double Battles

CodeTappert:
New Sprites by Brumirage whose Sprites we already use. (Source: https://play.pokemonshowdown.com/sprites/trainers/credits/)

This will stay as a draft until we merge https://github.com/pagefaultgames/pokerogue/pull/1254

since i dont want to do the dialogue twice.

I will add more doubles when they make sense. 

Currently:
Added Volo/Cynthia and Leon/Hop and 

- Koga & Janine
- Geeta & Nemona
- Opal & Bede
- Kieran & Carmine
- Lance (Champ) & Clair
- Hau & Kukui
- Diantha & Korrina (They meeet each others several times in the anime)
- - Sinnoh Gym Leader 1/Sinnoh Steel Gym Leader (they are father and son)

With those 4 all champs have a potential partner

Also Trainers now use
- main: 1,2,3 of their pool
- Partner -1,-2,-3 (from behind) of their pool
If they dont have enough they will get a random

PR #1582, opened 2024-05-29
[Bug] Double Battle Fixes

CodeTappert:
This takes things from https://github.com/pagefaultgames/pokerogue/pull/1399/files that are bug fixes that are needed and makes it their own pr

## What are the changes?

- Double Battle Trainers will now have Pokemon that scale in strengh (no first stage pokmeon at wave 150 anymore)
- If a double battle trainer is a derived type (like LANCE_CHAMPION), it does not correctly use the base TrainerType sprite (LANCE). This fix retrieves the derived trainer type (in this case, LANCE).

The first change is a bug found by players. The second one i saw in developing #1399

## What did change?

The double battle trainer now have pokemon that scale by level. 

### Screenshots/Videos

Before:
image
After:
image

## How to test the changes?

- Spawn Marnie & Piers at Like Wave 150 or so (with override.ts and battle.ts)

## Checklist
  - No, it was a part of #1399 but works on its own 
  - No. I could still split those two changes into seperate PRs
- [x] Are the changes visual?

### DIFF
--- a/src/field/trainer.ts
+++ b/src/field/trainer.ts
@@ -328,11 +328,16 @@ export default class Trainer extends Phaser.GameObjects.Container {
       }
 
       // If useNewSpeciesPool is true, we need to generate a new species from the new species pool, otherwise we generate a random species
-      const species = useNewSpeciesPool
+      let species = useNewSpeciesPool
         ? getPokemonSpecies(newSpeciesPool[Math.floor(Math.random() * newSpeciesPool.length)])
         : template.isSameSpecies(index) && index > offset
           ? getPokemonSpecies(battle.enemyParty[offset].species.getTrainerSpeciesForLevel(level, false, template.getStrength(offset)))
           : this.genNewPartyMemberSpecies(level, strength);
+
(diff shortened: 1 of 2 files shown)


src/field/trainer.ts
```diff
@@ -328,12 +328,17 @@ export default class Trainer extends Phaser.GameObjects.Container {
       }
 
       // If useNewSpeciesPool is true, we need to generate a new species from the new species pool, otherwise we generate a random species
-      const species = useNewSpeciesPool
+      let species = useNewSpeciesPool
         ? getPokemonSpecies(newSpeciesPool[Math.floor(Math.random() * newSpeciesPool.length)])
         : template.isSameSpecies(index) && index > offset
           ? getPokemonSpecies(battle.enemyParty[offset].species.getTrainerSpeciesForLevel(level, false, template.getStrength(offset)))
           : this.genNewPartyMemberSpecies(level, strength);
```
Tempo-anon, line 335: We really should move away from chained ternaries but that's out of scope for this PR

Tempo-anon (approved):

Issue #1568, opened 2024-05-29
[BUG] Future Sight does not behave like the mainline games


Smewkie:
Currently, when a pokemon uses Future Sight, the move only stays in effect if that user stays in battle, in the mainline games, the move would normally stay in effect and do damage 2 turns later even if the user is switched out, however the move should be cancelled if the user faints, which is happening correctly.

**To Reproduce**
1: Use Future Sight
2: swap out for another pokemon
3: Future Sight does not activate the next turn.

**Additional context**
Future Sight in general seems like it's going to be a pain to get working correctly, it has a lot of fail cases that just aren't currently checked for, for example you shouldn't be able to use multiple Future Sights in a row on the same pokemon, I could not tell you if the damage from Future Sight is currently calculated with the users stats and abilities on the turn it hits or the turn it's used.
And to top it all off, you have Doom Desire which works the same way, I do not know if these bugs extend to Doom Desire however, if whoever looks into this checks out Doom Desire as well, do note that Future Sight and Doom Desire should normally block each other from working.

GameOverFlowChart:
Are you sure that it shouldn't hit if the use faints? I don't think that was the case in gen 3, I don't know the newer gens that well.

Snailman11:
Link to Discord's bug report: (Player's Future Sight goes away if player user switches)
https://discord.com/channels/1125469663833370665/1245140616392867950

PigeonBar:
Could not reproduce Future Sight activating Enigma Berry on the predict turn, so hopefully that bug is fixed now:

https://github.com/user-attachments/assets/808473e1-196c-40c0-99e5-7dc40c8c5d2b

PR #1590, opened 2024-05-30
[Bug] Fix FutureSight

geefuoco:
## What are the changes?

- Enable Future Sight / Doom Desire to hit if mon switches out / dies
- Fail if Future Sight / Doom Desire if mon is already being targeted by one of these moves
- Display text when future sight cannot hit a target if it has fainted
- Enables Future Sight / Doom Desire to hit through protection moves.
    - Caveat here is that it shouldn't kill mons with Endure, Sturdy, Focus Sash, or Focus Band proc

Its a bug, Future sight should work if user is switched out. Addresses issue #1568 
## What did change?

- Add a check for delayed attack move in MoveEffectPhase 
### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/87877876/c2fe7b2d-c04f-49b6-832e-4e983609a057
## How to test the changes?

Give futuresight to mon your mon in overrides and use it, then switch out.

## Checklist
~~- [] Are the changes visual?~~


geefuoco:
https://github.com/pagefaultgames/pokerogue/assets/87877876/df0314bb-e53d-429c-aaa0-9546c2c312ce

---

Focus band and Sturdy works, but endure does not. 
Changing endure functionality is out of scope

geefuoco:
@Tempo-anon can you review ?

src/data/arena-tag.ts
```diff
@@ -466,13 +466,14 @@ class ToxicSpikesTag extends ArenaTrapTag {
   }
 }
 
-class DelayedAttackTag extends ArenaTag {
+export class DelayedAttackTag extends ArenaTag {
```
Tempo-anon, line 469: Can you add documentation to this class?

src/data/move.ts
```diff
@@ -1925,13 +1925,14 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
   }
 
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): Promise<boolean> {
```
Tempo-anon, line 1927: Also documentation on the `DelayedAttackAttr` class and this `apply`

src/data/move.ts
```diff
@@ -1925,13 +1925,14 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
...
+    const side = user.isPlayer() ? ArenaTagSide.ENEMY : ArenaTagSide.PLAYER;
```
Tempo-anon, line 1928: shouldn't this be `const side = target.isPlayer() ? ArenaTagSide.PLAYER : ArenaTagSide.ENEMY` ? For the edge case for when you use a delayed attack move on your ally

src/data/move.ts
```diff
@@ -5729,7 +5731,8 @@ export function initMoves() {
       .attr(StatChangeAttr, BattleStat.SPDEF, -1)
       .ballBombMove(),
     new AttackMove(Moves.FUTURE_SIGHT, Type.PSYCHIC, MoveCategory.SPECIAL, 120, 100, 10, -1, 0, 2)
-      .attr(DelayedAttackAttr, ArenaTagType.FUTURE_SIGHT, ChargeAnim.FUTURE_SIGHT_CHARGING, "foresaw\nan attack!"),
+      .attr(DelayedAttackAttr, ArenaTagType.FUTURE_SIGHT, ChargeAnim.FUTURE_SIGHT_CHARGING, "foresaw\nan attack!")
+      .ignoresProtect(),
```
Tempo-anon, line 5735: Should this also be on doom desire too?

src/phases.ts
```diff
@@ -2528,10 +2528,37 @@ export class MovePhase extends BattlePhase {
 
       this.scene.triggerPokemonFormChange(this.pokemon, SpeciesFormChangePreMoveTrigger);
 
+      const isDelayedAttack = (this.move.moveId === Moves.FUTURE_SIGHT || this.move.moveId === Moves.DOOM_DESIRE);
+      if (isDelayedAttack) {
+        // Check the player side arena if future sight is active
+        const futureSightTags = this.scene.arena.findTags(t => t.tagType ===  ArenaTagType.FUTURE_SIGHT);
+        const doomDesireTags = this.scene.arena.findTags(t => t.tagType === ArenaTagType.DOOM_DESIRE);
+        console.log("Future sight tag:", futureSightTags);
+        console.log("Doom Desire tag:", doomDesireTags);
```
Tempo-anon, line 2537: nit: remove `console.log`s

src/data/move.ts
geefuoco: yes good catch

geefuoco:
I've updated the code, note that tests are not passing, but its coming from the main branch and not this one
image

src/data/move.ts
```diff
@@ -2197,16 +2208,21 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
     this.chargeText = chargeText;
   }
 
+  /**
+   * @param {Pokemon} user - User of the move
+   * @param {Pokemon} target - Target of the move
+   * @param {Move} move - The move being used
+   * @param {any[]} args- Additional arguments to pass when applying a move
```
Tempo-anon, line 2215: You should describe what args[0] and args[1] are since they are used in the code

geefuoco: I'm not entirely sure what they are tbh. I didn't create the class from scratch, I just updated an existing class.

Tempo-anon: Can you figure it out and document it? Also add documentation to `DelayedAttackTag`

geefuoco: Not really since its implementing an abstraction. The interface itself has this definition 
image

Or here from the base class
image

so assuming its the same, I can just make note of that

Tempo-anon (dismissed):

Tempo-anon:
Merge conflicts?

src/data/move.ts
```diff
@@ -2303,15 +2314,22 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
...
 
+  /**
+   * @param {Pokemon} user - User of the move
+   * @param {Pokemon} target - Target of the move
+   * @param {Move} move - The move being used
+   * @param {any[]} args- Additional arguments to pass when applying a move. Typically this would be Utils.BooleanHolder
+   * for overridden, and a boolean for virtual
+   */
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): Promise<boolean> {
+    const side = target.isPlayer() ? ArenaTagSide.PLAYER: ArenaTagSide.ENEMY;
```
Tempo-anon, line 2325: This will cause a crash if future sight lands on the same turn you KO a trainer's last Pokemon

Tempo-anon (changes requested):

flx-sta:

-----

PR #1613, opened 2024-05-30
[Bug] Fix bug with variant rate

Tempo-anon:
## What are the changes?
Fixed variant rates

A previous PR accidentally changed variant rates

## What did change?
Variant rates should be 60/30/10 instead of 70/20/10

### Screenshots/Videos
n/a

## How to test the changes?
`npm run start:dev`

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1268,8 +1268,8 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
   /**
    * Generates a variant
    * Has a 10% of returning 2 (epic variant)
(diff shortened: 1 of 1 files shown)


XGSleepWalker (approved):

flx-sta (commented):
Alternatively you could use an array:
```ts
const odds: number[] = [
 2,
 1, 1, 1,
 0, 0, 0, 0, 0, 0
];

const rand = Utils.randSeedInt(odds.length);

return odds[rand];
```

might increase readability

PR #1618, opened 2024-05-30
[Bug] Fairy/Dark Aura Ability Correction + Aura Break Implementation

aeloan:
## What are the changes?

Implementation of zygarde first ability : aura break
Correction of yveltal and xerneas abilities : dark aura and fairy aura that boost only their dark/fairy move instead of all the field dark/fairy moves

After searching for an unimplemendant ability, I found that bug for the 2 others abilities so i patched them making the new one easier to implement, and making changes that could be used by others abilities

## What did change?
Added an ability attribute that can add an arena tag to target all the pokemon on the field
Replace the self-usage of the dark aura and fairy aura ability in order to work on other pokemon
Added the aura break ability that inverse the effects of dark aura and fairy aura
*The previous method are now non-used, shoud I delete them or let them here if needed in the future ?*

### Screenshots/Videos

Usage of 2 pokemon with feint attack that have a base power of 60, we can see on the logs the boost or weaken of the base power while the ability are on the field

Test of dark aura, boost of dark type move :
dark_aura_test

Test of aura break + dark aura, inversion of the dark type move boost
aura_break_test

###  On override.ts, start with a pokemon (attacker or defender) with the Ability Dark Aura
- Add dark type move to the one with dark aura, does it buff the value with and without the ability ?
- Add dark type move to the one without dark aura, does it buff the value with and without the ability ?
### On override.ts, start with a pokemon (attacker or defender) with the Ability Dark Aura and one with Aura Break
- Add dark type move to the one with dark aura, does it weaken the value with and without the ability ?
- Add dark type move to the one without dark aura, does it weaken the value with and without the ability ?

## Checklist
- [ ] Are the changes visual?


aeloan:
Saw some issues and improvement to do on my code with @innerthunder (ex: no account of ability being suppressed), I'm gonna change it soon on next commit

PR #1619, opened 2024-05-30
[Move] Implement Alluring Voice and Burning Jealousy

ElliottSimmonds:
## What are the changes?
Implementing the special functionality of Alluring Voice and Burning Jealousy, which will either confuse or burn target Pokemon if they have had their stats boosted previously on the same turn.

Both moves are partially implemented and I think they're cool.

## What did change?
* Added `STATS_BOOSTED` tag. This is added to a Pokemon until the end of the current turn when they recieve a stat boost.
* Added `ConfuseIfBoostedAttr` class for Alluring Voice, which will confuse the target if their stats have been boosted this turn.
* Added `BurnIfBoostedAttr` class for Burning Jealousy, which will burn the target(s) if their stats have been boosted this turn.

I had previously attempted to implement stat boost tracking using a `PokemonTurnData` property, but found problems when it came to boosts that occured on summon on turn 1. The `resetTurnData()` function call within `TurnInitPhase` would reset the attribute before move selection. There are other ways this could be done, but I have found using a battler tag instead provides a cleaner solution. I'm not entirely sure if it fits with the intended use case of battler tags, though.

Another thing to note: Stat boost copying, such as by Ditto's transform or psych up, do not trigger the move effects (verified on showdown).

### Screenshots/Videos
Example of the move effects triggering on turn 1:

[2024-05-29 19-44-16.webm](https://github.com/pagefaultgames/pokerogue/assets/33560264/be8f076c-2f8d-41af-876a-b64b2db00b38)

[2024-05-29 19-45-01.webm](https://github.com/pagefaultgames/pokerogue/assets/33560264/07df7d27-6196-4ce1-be17-24b5f20afd5b)

Example of the move effects triggering after the target has boosted their stats:

[2024-05-29 19-51-21.webm](https://github.com/pagefaultgames/pokerogue/assets/33560264/7024fff4-6ece-4b37-871d-71c1f175f2ea)

Example of move effects not triggering, for good measure:

[2024-05-29 19-52-43.webm](https://github.com/pagefaultgames/pokerogue/assets/33560264/f81348e7-ec4a-409b-840d-5470d1a3f93b)

## How to test the changes?

I just made use of overrides, using abilities that boost stats on summon and stat boosting moves.
The effects trigger if the moves hit after stats have been boosted, so your starter will need to be slower than the opponent in most situations.
I skipped to the Eternatus battle to make sure it works against the boosts bosses recieve when one of their health bars is knocked out.

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -4864,6 +4864,62 @@ export class DestinyBondAttr extends MoveEffectAttr {
...
 
+/**
+ * Attribute to apply {@linkcode BattlerTagType.CONFUSED} to the target if they have had their stats boosted this turn.
+ *
+ * @extends AddBattlerTagAttr
+ */
+export class ConfuseIfBoostedAttr extends AddBattlerTagAttr {
+  constructor() {
+    super(BattlerTagType.CONFUSED, false, false, 2, 5);
+  }
```
OrangeRed, line 4875: Can you make it so that this class takes a `BattlerTagType` as a constructor param instead of hard coding it to only use `BattlerTagType.CONFUSED`?

src/data/move.ts
```diff
@@ -4864,6 +4864,62 @@ export class DestinyBondAttr extends MoveEffectAttr {
...
+/**
+ * Attribute to apply {@linkcode StatusEffect.BURN} to the target if they have had their stats boosted this turn.
+ *
+ * @extends MoveEffectAttr
+ */
+export class BurnIfBoostedAttr extends MoveEffectAttr {
+  constructor() {
+    super(true, MoveEffectTrigger.HIT);
+  }
+
```
OrangeRed, line 4904: Can you make it so that this class takes a `StatusEffect` as a constructor param instead of hard coding it to only use `StatusEffect.BURN`?

src/phases.ts
```diff
@@ -3069,6 +3069,15 @@ export class StatChangePhase extends PokemonPhase {
       }
 
       for (const stat of filteredStats) {
+        if (levels.value > 0 && pokemon.summonData.battleStats[stat] + levels.value <= 6) {
+          if (pokemon.turnData) {
+            pokemon.turnData.statsBoosted = true;
+          }
+          if (pokemon.scene.currentBattle.turn === 1) {
+            pokemon.battleData.statsBoostedFirstTurn = true;
+          }
```
OrangeRed, line 3078: Do you need this if you're already checking in the Attrs whether the target has stat boosted this turn?
isn't `user.scene.currentBattle.turn === 1 && target.turnData.statBoosted` handle this variable and even still if you want to be super safe you can use `PokemonTurnData.acted` to check if the target acted in the turn prior to the move's user?

src/field/pokemon.ts
```diff
@@ -3732,6 +3733,7 @@ export class PokemonBattleSummonData {
 export class PokemonTurnData {
   public flinched: boolean;
   public acted: boolean;
+  public statsBoosted: boolean;
```
OrangeRed, line 3736: I would default these to false so it resets properly instead of resetting to undefined

OrangeRed (changes requested):

src/data/move.ts
ElliottSimmonds: Thanks, I've made these changes :)

src/phases.ts
OrangeRed: > The problem is during the first turn of wild battles pokemon.turnData is undefined during the StatsChangePhase, so statBoosted can't be set to true.

If this is the case I would explore trying to add `pokemon.resetTurnData()` during the EncounterPhase or the TurnStartPhase. This will require some experimentation to figure out the optimal location, you should also consult with some the devs about this.

ElliottSimmonds: Looking into it further, turnData being undefined isn't the real issue, but rather the turnData being reset in the `resetTurnData()` call in the `TurnInitPhase`.  Since on summon effects happen before, any changes to `statsBoosted` triggered by abilities such as download will be reset back to false. I'll defo need to give the devs a shout from the look of it

ElliottSimmonds: I have found using a battler tag instead provides a cleaner solution

OrangeRed (changes requested):
Can you some tests that verify this works correctly?

frutescens:
Work continued in #3508

PR #1620, opened 2024-05-30
[Ability] Added Partial and TODO to Compound Eyes and Super Luck

EvasiveAce:
Just removed partial from Super Luck. It makes the ability look better, and the actual in battle ability does work as intended. Compound eyes is the exact same situation but does not have .partial(). 

## Checklist
- [x] Are the changes visual?
  - (it just removes (P))


OrangeRed:
Can you add the + 1 luck attr to this PR and rename it to include you did that?

EvasiveAce:
> Can you add the + 1 luck attr to this PR and rename it to include you did that?

I will do a separate PR since it seems like a design type thing.

PR #1627, opened 2024-05-31
Add Arena Events

bennybroseph:

This adds events when Weather, Terrain, or Tags change in any way. This allows future UI elements to listen for these events and update without needing any update or refresh loop.

## What did change?

This change adds a new file to define all of the events and enums. It also adds events for when the arena's values change including during sessions being loaded.

## How to test the changes?

Since the changes are not visual, you'll need to add breakpoints when each event occurs and trigger them. This can be done by setting up weather/terrain/tags and then saving on the next floor and loading to completely test all events.

## Checklist

### DIFF
--- a/src/field/arena-events.ts
+++ b/src/field/arena-events.ts
@@ -1,0 +1,77 @@
+import { ArenaTagSide } from "#app/data/arena-tag.js";
+import { ArenaTagType } from "#app/data/enums/arena-tag-type.js";
+import { TerrainType } from "#app/data/terrain.js";
+import { WeatherType } from "#app/data/weather.js";
+
+/** Alias for all {@linkcode ArenaEvent} type strings */
+export enum ArenaEventType {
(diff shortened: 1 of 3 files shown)


src/field/arena-events.ts
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Container class for {@linkcode ArenaEventType.TERRAIN_CHANGED} events
+ * @extends ArenaEvent
+*/
+export class TerrainChangedEvent extends ArenaEvent {
+  /** The {@linkcode TerrainType} being overridden */
+  public oldTerrainType: TerrainType;
+  /** The {@linkcode TerrainType} being set */
+  public newTerrainType: TerrainType;
+  constructor(oldTerraintType: TerrainType, newTerrainType: TerrainType, duration: number) {
```
torranx, line 55: typo on oldTerraintType

XGSleepWalker (approved):

PR #1633, opened 2024-05-31
[QoL] Add type inference to getAttrs methods and refactor accordingly

OrangeRed:
## What are the changes?
Add type inference to `getAttrs` methods in `move.ts` and `ability.ts`

`as Type` assertions are dangerous and we should rely more on the typescript LSP

## Checklist
- [ ] Are the changes visual?


src/data/ability.ts
```diff
@@ -55,8 +55,22 @@ export class Ability implements Localizable {
...
   }
 
-  getAttrs(attrType: { new(...args: any[]): AbAttr }): AbAttr[] {
-    return this.attrs.filter(a => a instanceof attrType);
+  /**
+   * Get all ability attributes that match `attrType`
+   * @param attrType any attribute that extends {@linkcode AbAttr}
+   * @returns Array of attributes that match `attrType`, Empty Array if none match.
+   */
+  getAttrs<T extends AbAttr>(attrType: new(...args: any[]) => T ): T[] {
```
flx-sta, line 63: I know you didn't name this method but it could be confused with a getter for `this.atts`.
Maybe it should be renamed to `getAttrsOfType()` or the parameter should be made optional to return the whole `this.attrs` field (idk if that is needed)

src/data/move.ts
```diff
@@ -148,8 +148,22 @@ export default class Move implements Localizable {
...
   }
 
-  getAttrs(attrType: { new(...args: any[]): MoveAttr }): MoveAttr[] {
-    return this.attrs.filter(a => a instanceof attrType);
+  /**
+   * Get all move attributes that match `attrType`
+   * @param attrType any attribute that extends {@linkcode MoveAttr}
+   * @returns Array of attributes that match `attrType`, Empty Array if none match.
+   */
+  getAttrs<T extends MoveAttr>(attrType: new(...args: any[]) => T): T[] {
```
flx-sta, line 156: I know you didn't name this method but it could be confused with a getter for `this.atts`.
Maybe it should be renamed to `getAttrsOfType()` or the parameter should be made optional to return the whole `this.attrs` field (idk if that is needed)

flx-sta (changes requested):

flx-sta:
@OrangeRed https://github.com/OrangeRed/pokerogue/pull/1

hayuna:
@flx-sta If this test is for specific ability / move, could we start creating deeper structure?

src/tests/ability/Magician.test.ts
src/tests/move/RainDance.test.ts

Would be easier to future development to not store every ability tests in single one file.

flx-sta:
> @flx-sta If this test is for specific ability / move, could we start creating deeper structure?
> 
> src/tests/ability/Magician.test.ts src/tests/move/RainDance.test.ts
> 
> Would be easier to future development to not store every ability tests in single one file.

The abilities were chosen randomly. It is generally about testing the methods inside ability & move

OrangeRed:
To clear up any confusion I changed the tests to use `Ability.None` and `Move.None`. The id of the move doesn't really effect anything when it comes to testing its internal functionality so that should be a safe change to make.

flx-sta:
> To clear up any confusion I changed the tests to use `Ability.None` and `Move.None`. The id of the move doesn't really effect anything when it comes to testing its internal functionality so that should be a safe change to make.

Funny enough initially I went with these too.
But they I thought I should use some cooler Abilities. 
well... here we are :d

flx-sta (approved):

flx-sta:
@hayuna an approval (if you tested) is always appreciated ❤️

bennybroseph (changes requested):
I'm very glad we've included tests for this change, but please split those tests off into a different PR as we have critiques which will prevent us from merging the change itself.

flx-sta:
> I'm very glad we've included tests for this change, but please split those tests off into a different PR as we have critiques which will prevent us from merging the change itself.

Ruthless 🗡️ . 
They were meant to provide coverage/leverage :D

src/test/ability.test.ts
XGSleepWalker, line 1: This file structure will scale very poorly, I would rather bring a baseline that everyone can follow later on than building on top of this right now. Please remove the tests and we'll add them at a later stage.

XGSleepWalker (changes requested):

bennybroseph (approved):

PR #1635, opened 2024-05-31
[Feature] View modifier/Item bar tooltips via keyboard/controller

LeBulldoge:
## What are the changes?
This PR adds a way for the user to view the tooltips attached to items without having to reach for the mouse.
The keys are as follows:
    * 🎮`LB`, ⌨️`O` - hold to enter *player* item view mode, release to exit
    * 🎮`RB`, ⌨️`P` - hold to enter *enemy* item view mode, release to exit
Once in the item view mode, left and right keys can be used to move the cursor through the items.
The above, of course, can easily be changed in case there are other plans for those keys.

Players had to reach for their mouse to view the tooltips on items, regardless of their input method (other than touch).

## What did change?
Modified the controller mappings to allow multiple buttons per input, as to match the keyboard input's behavior, and also because all of the buttons that made sense for the feature were already used, but in a different mode.
Added a new UI handler to add the cursor and tooltip to the modifier bar, and control user input over it.
For this to work without making a whole lot of changes across the board, I have made the `BattleScene.modifierBar` and `BattleScene.enemyModifierBar` fields public.

### Screenshots/Videos
https://github.com/pagefaultgames/pokerogue/assets/13983982/b1b08ff1-5967-4869-8614-94e40669636a

## How to test the changes?
Enter combat with items equipped on either side (via overrides or otherwise), then press the respective key for your input method, which you can find above. Move between the items with left and right keys.

## Checklist
- [x] Are the changes visual?


bennybroseph (changes requested):
So this iterates through them in a list? Like, if i'm at the first item and i want to go down a row, i have to go through all of the ones on the current row first right? That seems very slow and annoying with how many items we have in the game. Can we make this swap input modes and use the D-Pad instead? Otherwise this is too clunky to be useful. 

Also need comments on this once those changes are made. There is a document in the `docs` folder on standards.

LeBulldoge:
> So this iterates through them in a list? Like, if i'm at the first item and i want to go down a row, i have to go through all of the ones on the current row first right? That seems very slow and annoying with how many items we have in the game. Can we make this swap input modes and use the D-Pad instead? Otherwise this is too clunky to be useful.
> 
> Also need comments on this once those changes are made. There is a document in the `docs` folder on standards.

Thank you for the reply. I modified it to allow up/down movement and uploaded a new video. Going to work on comments.

LeBulldoge:
I left comments in places that seemed like they needed them, I think it's good to review again.

bennybroseph:
Would like to get the conflicts on this fixed and then have it retested

bennybroseph:
@LeBulldoge Is this still being worked on?

LeBulldoge:
> @LeBulldoge Is this still being worked on?

Sorry, I was a bit preoccupied this week, I'll pick it up again over the weekend

bennybroseph:
Still hoping to get this added @LeBulldoge

flx-sta:
Closing due to inactivity.
If you want to re-open it let us know

PR #1641, opened 2024-05-31
[Bug] Fix status abilities not considering Comatose (fixes #1639)

### LINKED ISSUE #1639: [BUG/DISCUSSION] Komala's guts passive doesn't work.
Did I search my issue first?
I've searched Komala, Guts, Comatose and can't see this mentioned anywhere

**Describe the bug**
Currently unlocking Komala's passive Guts does nothing for it. This is likely same as mainline games, more at added context.

**To Reproduce**
I opened local server and tested tackle using lv 50 Komala VS  floor 81 Scrafty with all permutations of No ability/Comatose/Guts.

I consistently did 19-21 damage base, and 30 dmg crit.

Then I went only guts and got paralyzed on purpose. 28-31 dmg base didn't verify crit

**Expected behavior**
Given the the obvious attempt at giving Komala a cool synergy with the passive I expected a damage increase. Instead it was currently wasted 25 candy since you can't status Komala unless you lose the ability with skillswap/mummy etc.

**Additional context**
So given Comatose technically isn't a status this seems fully inline with mainline games, and we can't test in canon format since 2 abilities impossible. So either add comatose exception to guts to make synergy work or change Komala passive since it's currently useless.

jaimefd:
## What are the changes?

The abilities Guts, Marvel Scale and Quick Feet now properly boost the user's stats if it also has the ability Comatose.

The issue https://github.com/pagefaultgames/pokerogue/issues/1639 showed that the interaction between Comatose and other status abilities wasn't working, probably due to the fact that it's not a thing present in main series games, as you can't have more than 1 ability there. The solution seemed pretty easy and straight forward, so I did it.

## What did change?

Simply added an extra condition for the 3 mentioned abilities, instead of checking if `!!pokemon.status` is true, it now checks if `!!pokemon.status` OR `pokemon.hasAbility(Abilities.COMATOSE)` are true. This is the same logic the ability already uses in the [code](https://github.com/pagefaultgames/pokerogue/commit/bfe018ef65aaeb4621661eb35930d9ae1ebc6c21), so it is consistent.

I don't believe a tag or custom status is needed for now, as Comatose is the only existing ability with such functionality. But if more abilities like this were introduced in the future, it would be an idea to consider then.

## Screenshots/Videos

Tested it using Gastro Acid on the enemy, to cancel Guts mid battle and see the difference. Since Gastro Acid is currently coded to only remove main Ability (and not passive) I had to put Guts as ability and Comatose as passive.

Before: (Hit with Guts does 23, hit without Guts deals 22, showing no real difference)

https://github.com/pagefaultgames/pokerogue/assets/22176793/dc2973d4-617d-43a5-9506-922e01970849

After: (Hit with Guts does 33, hit without Guts deals 22, the 50% boost to Attack is working now!)

https://github.com/pagefaultgames/pokerogue/assets/22176793/6a1abb8e-26e1-4dc8-82ff-5a81d8886f26

## How to test the changes?

I simply used the overrides.ts file, gave the enemy a moveset of 4 Gastro Acid (to guarantee it doesn't use anything else) and I gave my pokemon the Guts ability and Comatose passive (the passive must be Comatose and not vice versa for us to see the difference with the Gastro Acid method). Of course, choose a tanky and slow enemy pokemon for easier testing, so that you can attack once before the ability is removed.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -3716,9 +3716,9 @@ export function initAbilities() {
       .conditionalAttr(pokemon => !Utils.randSeedInt(3), PostTurnResetStatusAbAttr),
     new Ability(Abilities.GUTS, 3)
       .attr(BypassBurnDamageReductionAbAttr)
-      .conditionalAttr(pokemon => !!pokemon.status, BattleStatMultiplierAbAttr, BattleStat.ATK, 1.5),
+      .conditionalAttr(pokemon => !!pokemon.status || pokemon.hasAbility(Abilities.COMATOSE), BattleStatMultiplierAbAttr, BattleStat.ATK, 1.5),
     new Ability(Abilities.MARVEL_SCALE, 3)
-      .conditionalAttr(pokemon => !!pokemon.status, BattleStatMultiplierAbAttr, BattleStat.DEF, 1.5)
+      .conditionalAttr(pokemon => !!pokemon.status || pokemon.hasAbility(Abilities.COMATOSE), BattleStatMultiplierAbAttr, BattleStat.DEF, 1.5)
       .ignorable(),
     new Ability(Abilities.LIQUID_OOZE, 3)
       .attr(ReverseDrainAbAttr),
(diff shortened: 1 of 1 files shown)


Tempo-anon:
What happens if Comatose is the main ability and it gets suppressed/negated while Guts is the passive?

jaimefd:
> What happens if Comatose is the main ability and it gets suppressed/negated while Guts is the passive?

That will never be the case since Comatose cannot be supressed, swapped, ignored, or any other effect that messes with abilities ([source](https://bulbapedia.bulbagarden.net/wiki/Comatose_(Ability)#Effect)).

This was already accounted for before my PR, with `.attr(UncopiableAbilityAbAttr).attr(UnswappableAbilityAbAttr).attr(UnsuppressableAbilityAbAttr)` in the ability.

There is one very very very very rare edge case, mentioned in the wiki, where if an already supressed Pokémon uses Transform, Comatose won't work. This is the only situation where it doesn't work like the main series games, Comatose will still work here.

However, it is the only ability in the game with this situation, and it seems like a bug in the original games. All other abilities that are also UnsuppressableAbilityAbAttr work even if a supressed mon transforms (or they don't allow you to transform at all). I do not believe pokerogue should stay consistent here, in terms of balance and consistency.

It even seems to be explicitly checked for in the code already:
https://github.com/pagefaultgames/pokerogue/blob/fc61489d59ef4be9e4ab9ce97e94c36774faa64f/src/field/pokemon.ts#L976-L978

"Fixing" this seems to give more problems than solutions, and either way should probably belong to other PR, as I'm just using the same logic Comatose was already using, such as for the Hex move, for example:
https://github.com/pagefaultgames/pokerogue/blob/99e17c94ebb206935a050ae06b55f3028f1ade05/src/data/move.ts#L6440-L6443

bennybroseph (changes requested):
Instead of checking explicitly for Comatose, can Comatose just count as a status condition in the pokemon.status property?

bennybroseph (approved):

PR #1654, opened 2024-05-31
[Qol]? Disables all double battles but tate and Liza. And those two will now…

CodeTappert:
… always have lunatone and the other one (solrock?) as their first pokemon

## What are the changes?

We discussed in contribuer chat about it. The double battles destroy the balance. We want to keep them for a possible future mode. 
Only tate&liza should have a chance to appear at a 1/3 chance. 

Because we came to a conclusion to do so. And i was the person who introduced the double battle thing in the first place

## What did change?

The champs and piers/marnie can not longer appear in the double battles. 
And Tate&Liza now always have a solrock and lunatone as their first pokemon.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/38758606/c1bd6479-cc4f-4b54-848c-3adf28604771

(Not quite sure how i would show the removal of the other double battles)

## How to test the changes?

Made Blue spawn as the champion (forced) for wave 190 several times. He didnt appear as a double even once.
Made Tate&Liza forced spawn as a double at wave 5 to see if the their first pokemon are always the ones we want

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/field/trainer.ts
+++ b/src/field/trainer.ts
@@ -20,6 +20,7 @@ import {trainerNamePools} from "../data/trainer-names";
 import {ArenaTagSide, ArenaTrapTag} from "#app/data/arena-tag";
 import {getIsInitialized, initI18n} from "#app/plugins/i18n";
 import i18next from "i18next";
+import {Species} from "#app/data/enums/species";
 
 export enum TrainerVariant {
     DEFAULT,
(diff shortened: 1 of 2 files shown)


src/field/trainer.ts
flx-sta, line 1: instead of hard-coding the trainers in the if and their pokemon I think it would make sense to create a matcher for it.
something along this
```ts
const variable_name: Record<TrainerType, Species[]> = {
  TrainerType.TATE: [ Species.SOLROCK ],
  TrainerType.LIZA: [ Species.LUNATONE ],
}
```

flx-sta (changes requested):

src/field/trainer.ts
CodeTappert: Why? This will never get extended. 

The feature will remain unchanged until we, maybe, at some point do a "double classic" mode or something. 

And then the logic for building the team gets reworked completly.

CodeTappert:
To test just set the wave to 5 and then this line:
https://github.com/pagefaultgames/pokerogue/blob/fe34051687dd6720167b2eca6feaf42baa7904dc/src/battle.ts#L331

to  

```
    .setGetTrainerFunc(scene => new Trainer(scene, TrainerType.TATE, Utils.randSeedInt(2) ? TrainerVariant.DOUBLE : TrainerVariant.DOUBLE)),

```

flx-sta (approved):
lgtm 🚀 

image

CodeTappert:
Maybe @bennybroseph can take a Look. Since you apporved the other PRs about that

XGSleepWalker (approved):

PR #1567, opened 2024-05-29
[Localization] modifier-select-ui-handler Locales

Adri1:
## What are the changes?
Locales for item selection menu (modifier-select-ui-handler) between battles

## Why am I doing these changes?
To localize this menu and for accessibility

## What did change?
If a translation is provided, this menu will be in the set language
Number format adapting to the language (ex : English "1,000" for a thousand and "1 000" in French for a thousand)

**Only one minor visual issue :**
The cursor position doesn't adjusts according to text length, it disappears under the text if it is longer than the original text, so if anybody has a solution to provide...
image

### Screenshots/Videos
**Before** (French)
image

**After** (French)
image
image

## How to test the changes?
Being in Classic or Endless with a Lock Capsule and at least one item held by any Pokémon of you team to unlock "Lock rarities" and "Transfer" options in the menu.

## Checklist
- [X] There is no overlap with another PR?
- [X] The PR is self-contained and cannot be split into smaller PRs?
- [X] Have I provided a clear explanation of the changes?
- [X] Have I tested the changes (manually)?
    - [X] Are all unit tests still passing? (`npm run test`)
- [X] Are the changes visual?
  - [X] Have I provided screenshots/videos of the changes?

PR #1656, opened 2024-05-31
[Localization] modifier-select-ui-handler Locales, with corrections

Adri1:
#1567 but with corrections

## What are the changes?
Locales for item selection menu (modifier-select-ui-handler) between battles

To localize this menu and for accessibility

## What did change?
If a translation is provided, this menu will be in the set language
Number format adapting to the language (ex : English "1,000" for a thousand and "1 000" in French for a thousand)

**Only one minor visual issue :**
The cursor position doesn't adjusts according to text length, it disappears under the text if it is longer than the original text, so if anybody has a solution to provide...
image

### Screenshots/Videos
**Before** (French)
image

**After** (French)
image
image

## How to test the changes?
Being in Classic or Endless with a Lock Capsule and at least one item held by any Pokémon of you team to unlock "Lock rarities" and "Transfer" options in the menu.

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/locales/ko/ability.ts
+++ b/src/locales/ko/ability.ts
@@ -1154,7 +1154,7 @@ export const ability: AbilityTranslationEntries = {
   },
   beadsOfRuin: {
     name: "재앙의구슬",
-    description: "재앙을 부르는 곡옥의 힘으 자신을 제외한 모든 포켓몬의 특수방어를 약하게 만든다."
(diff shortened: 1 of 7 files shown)


returntoice:
Does this one still on progress now? I think it is the urgent one for localization

Adri1:
> Does this one still on progress now? I think it is the urgent one for localization

I havn't touched it even since and many things changed, so if anyone wants to step in while I can't, they are welcome

Issue #1614, opened 2024-05-30
[BUG] Roar Self target and animation


DustinLin:
**Describe the bug**

Opponent can target itself with roar, which shouldn't be possible. Additionally sprites become clear

**To Reproduce**

`overrides.ts` allow doubles, and set all opponents moves to roar

**Expected behavior**

Both of the above shouldn't happen

**Screenshots / Videos**

https://github.com/pagefaultgames/pokerogue/assets/39450497/3847ec5f-9091-4566-8799-fee12ff2296d

**Device**

**Additional context**

DustinLin:
Just realized this is also referenced in issue #481

flx-sta:
Related to #481

PR #1660, opened 2024-05-31
fix Roar

cywk99:
## What are the changes?

Prevent enemy from using ROAR on ally. 
Fix the sprite render bug after enemy use ROAR.

#1614 

## What did change?

In moves.ts, for the ROAR move, removed `.hidesTarget()` and added `.target(MoveTarget.ALL_ENEMIES)`
### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/60039008/ace3269f-b5bc-4b20-bd49-e8a92b69b14e

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -5167,7 +5167,7 @@ export function initMoves() {
     new StatusMove(Moves.ROAR, Type.NORMAL, -1, 20, -1, -6, 1)
       .attr(ForceSwitchOutAttr)
       .soundBased()
-      .target(MoveTarget.ALL_ENEMIES),
+      .target(MoveTarget.NEAR_ENEMY),
     new StatusMove(Moves.SING, Type.NORMAL, 55, 15, -1, 0, 1)
       .attr(StatusEffectAttr, StatusEffect.SLEEP)
       .soundBased(),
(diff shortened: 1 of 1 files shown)


cywk99:
Hi @Tempo-anon Can I have someone review this PR? Thanks!

EvasiveAce:
I've been working on Roar, and the `hidesTarget()` is not the issue. As you can see in your video, as well as the one I am going to post, Roar moves the pokemon out, but after the animation, you can briefly see it return to the middle. I think this is entirely from `Roar.json` breaking `hidesTarget()`

As for `.target(Movetarget.ALL_ENEMIES)` I was under the impression that it would hit both, but it seems that is not the case with your video. If that works as a single targeting move, then it's fine.

Easier to see animation flash after Roar (OBS would not pick it up, seems to be a frame broken?):

[Video](https://github.com/pagefaultgames/pokerogue/assets/71776311/f4cdf040-8985-42da-92d1-75b1d94d1e4c)

cywk99:
Hi @EvasiveAce 

Not really sure about `hideTarget()`, but by removing it, roar seems to work fine?

I believe roar should be a single target move, and yeah I have updated to `.target(MoveTarget.NEAR_ENEMY)` , it makes the `multiple` variable false and the player's pokemon are able to select a single enemy as roar target now.

Tempo-anon:
Going to close this since this changes the functionality of roar while it looks like there are still some issues with it

PR #1667, opened 2024-05-31
Implement support for relearner only moves

Xavion3:
## What are the changes?
Implements support for relearner only moves, moves that aren't learned by level up or evolution, only by memory mushrooms.

This has been reported before, and is an outstanding bug.

## What did change?
This is mostly just engine changes to support level -1 moves as relearner exclusive moves. It also adds a few extra comments and type annotations to functions touched by this PR.

## How to test the changes?
Look at Zangoose in the starter screen and notice it no longer gets 15 moves at starter select.
You can also test that memory mushrooms still give those moves, and that level up/evolution moves still work.

## Checklist
- [ ] Are the changes visual?


Madmadness65 (approved):
lgtm, I didn't experience any major issues in my test of this, and Zangoose does indeed get those moves properly by memory mushroom now. Always good to have a second opinion before the merge, though.

PR #1670, opened 2024-06-01
[Balance] Simplify and fix flame/toxic orb weight

Xavion3:
## What are the changes?
Reworks the flame and toxic orb algorithms, fixing a variety of bugs and simplifying it to be more consistent with other item weights.

The new system gives it a high weight in line with the other highest weight combat items in tier if it has a relevant move or ability and doesn't already have one of the two items.

There were like five different bugs in the previous algorithm and it was literally the highest weight item in its tier.

## What did change?
The anonymous function determining the weight of the two mentioned items was rewritten.

### Screenshots/Videos
N/A

## How to test the changes?
Run hundreds of games with and without the pokemon?

## Checklist
- [N] Are the changes visual?

### DIFF
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -1338,13 +1338,13 @@ const modifierPool: ModifierPool = {
     new WeightedModifierType(modifierTypes.TOXIC_ORB, (party: Pokemon[]) => {
       const checkedAbilities = [Abilities.QUICK_FEET, Abilities.GUTS, Abilities.MARVEL_SCALE, Abilities.TOXIC_BOOST, Abilities.POISON_HEAL];
       const checkedMoves = [Moves.FACADE, Moves.TRICK, Moves.FLING, Moves.SWITCHEROO, Moves.PSYCHO_SHIFT];
-      // If a party member doesn't already have one of these two items and has one of the above moves or abilities
+      // If a party member doesn't already have one of these two orbs and has one of the above moves or abilities, the orb can appear
       return party.some(p => !p.getHeldItems().some(i => i instanceof Modifiers.TurnStatusEffectModifier) && (checkedAbilities.some(a => p.hasAbility(a, false, true)) || p.getMoveset(true).some(m => checkedMoves.includes(m.moveId)))) ? 10 : 0;
     }, 10),
     new WeightedModifierType(modifierTypes.FLAME_ORB, (party: Pokemon[]) => {
       const checkedAbilities = [Abilities.QUICK_FEET, Abilities.GUTS, Abilities.MARVEL_SCALE, Abilities.FLARE_BOOST];
       const checkedMoves = [Moves.FACADE, Moves.TRICK, Moves.FLING, Moves.SWITCHEROO, Moves.PSYCHO_SHIFT];
-      // If a party member doesn't already have one of these two items and has one of the above moves or abilities
+      // If a party member doesn't already have one of these two orbs and has one of the above moves or abilities, the orb can appear
       return party.some(p => !p.getHeldItems().some(i => i instanceof Modifiers.TurnStatusEffectModifier) && (checkedAbilities.some(a => p.hasAbility(a, false, true)) || p.getMoveset(true).some(m => checkedMoves.includes(m.moveId)))) ? 10 : 0;
     }, 10),
     new WeightedModifierType(modifierTypes.REVIVER_SEED, 4),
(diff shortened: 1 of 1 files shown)


src/modifier/modifier-type.ts
```diff
@@ -1336,39 +1336,15 @@ const modifierPool: ModifierPool = {
...
-      } else {
-        const moveList = [Moves.FACADE, Moves.TRICK, Moves.FLING, Moves.SWITCHEROO, Moves.PSYCHO_SHIFT];
-        if (filteredParty.some(p => p.getMoveset().some(m => moveList.includes(m.moveId)))) {
-          weight = 1;
-        }
-      }
-      return Math.min(Math.ceil(party[0].scene.currentBattle.waveIndex / 15) * weight, 8 * weight);
-    }, 32),
+      const checkedAbilities = [Abilities.QUICK_FEET, Abilities.GUTS, Abilities.MARVEL_SCALE, Abilities.TOXIC_BOOST, Abilities.POISON_HEAL];
+      const checkedMoves = [Moves.FACADE, Moves.TRICK, Moves.FLING, Moves.SWITCHEROO, Moves.PSYCHO_SHIFT];
```
Tempo-anon, line 1340: I think it'd make sense to put the (N) moves in a /**/ with a comment for later

src/modifier/modifier-type.ts
```diff
@@ -1336,39 +1336,15 @@ const modifierPool: ModifierPool = {
...
+      return party.some(p => !p.getHeldItems().some(i => i instanceof Modifiers.TurnStatusEffectModifier) && (checkedAbilities.some(a => p.hasAbility(a, false, true)) || p.getMoveset(true).some(m => checkedMoves.includes(m.moveId)))) ? 10 : 0;
```
Tempo-anon, line 1341: This is not easily understandable at a glance. I think it makes sense to include a comment describing what the calculation is in English

src/modifier/modifier-type.ts
```diff
@@ -1336,39 +1336,15 @@ const modifierPool: ModifierPool = {
...
-        const moveList = [Moves.FACADE, Moves.TRICK, Moves.FLING, Moves.SWITCHEROO, Moves.PSYCHO_SHIFT];
-        if (filteredParty.some(p => p.getMoveset().some(m => moveList.includes(m.moveId)))) {
-          weight = 1;
-        }
-      }
-      return Math.min(Math.ceil(party[0].scene.currentBattle.waveIndex / 15) * weight, 8 * weight);
-    }, 32),
+      const checkedAbilities = [Abilities.QUICK_FEET, Abilities.GUTS, Abilities.MARVEL_SCALE, Abilities.FLARE_BOOST];
+      const checkedMoves = [Moves.FACADE, Moves.TRICK, Moves.FLING, Moves.SWITCHEROO, Moves.PSYCHO_SHIFT];
+      return party.some(p => !p.getHeldItems().some(i => i instanceof Modifiers.TurnStatusEffectModifier) && (checkedAbilities.some(a => p.hasAbility(a, false, true)) || p.getMoveset(true).some(m => checkedMoves.includes(m.moveId)))) ? 10 : 0;
```
Tempo-anon, line 1346: Is this the same calc as toxic orb? If it's being repeated I think it'd make sense to be pulled into its own helper function

src/modifier/modifier-type.ts
Xavion3: Added

src/modifier/modifier-type.ts
Xavion3: I don't think it needs a helper here despite being similar, maybe once we get more items which are dependent on abilities/moves we can look into a central helper function.

src/modifier/modifier-type.ts
Xavion3: Since unimplemented moves are effectively unobtainable I don't think this is an issue, as the player and AI shouldn't ever be able to get them.

src/modifier/modifier-type.ts
```diff
@@ -1336,39 +1336,17 @@ const modifierPool: ModifierPool = {
...
-        const moveList = [Moves.FACADE, Moves.TRICK, Moves.FLING, Moves.SWITCHEROO, Moves.PSYCHO_SHIFT];
-        if (filteredParty.some(p => p.getMoveset().some(m => moveList.includes(m.moveId)))) {
-          weight = 1;
-        }
-      }
-      return Math.min(Math.ceil(party[0].scene.currentBattle.waveIndex / 15) * weight, 8 * weight);
-    }, 32),
+      const checkedAbilities = [Abilities.QUICK_FEET, Abilities.GUTS, Abilities.MARVEL_SCALE, Abilities.FLARE_BOOST];
+      const checkedMoves = [Moves.FACADE, Moves.TRICK, Moves.FLING, Moves.SWITCHEROO, Moves.PSYCHO_SHIFT];
+      // If a party member doesn't already have one of these two items and has one of the above moves or abilities
```
Tempo-anon, line 1347: nit: I think the comment should be explicit. 
`// If a party member doesn't already have a status orb and has one of the checked moves or abilities the weight is 10. Otherwise the weight is 0`

Xavion3: Comment changed.

Tempo-anon (approved):

PR #1671, opened 2024-06-01
[Move] Implemented Incinerate and Knock Off

Arxalc:
## What are the changes?

Incinerate incinerates berries. Knock off removes items and boosts power as expected.

This [PR](https://github.com/pagefaultgames/pokerogue/pull/1525) and [PR](https://github.com/pagefaultgames/pokerogue/pull/1253) cover the same thing, but I believe this implementation is more complete.

## What did change?

* Incinerate successfully burns up berries
* Knock off removes _removeable_ items 
* Knock off's power boost only applies when an item can be knocked off
* Items have an equal, random chance of being removed (unlike Thief and Covet which prioritize higher tier items - idk if that was intentional or not, but I'm not messing with that)

_Note: There are 2 unimplemented edge cases which apply to Thief as well. It's unlikely to occur, so I'm submitting this for now. These are commented in for the future.
"If the user faints due to the target's Ability (Rough Skin or Iron Barbs) or held Rocky Helmet, it cannot remove the target's held item."
"If Knock Off causes a Pokémon with the Sticky Hold Ability to faint, it can now remove that Pokémon's held item."_

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/63990624/a47cf2f5-0e20-4648-841c-2fb1a1b25ec8
_Incinerate only incinerates berries. Knock off power varies._
https://github.com/pagefaultgames/pokerogue/assets/63990624/a43266aa-d7d3-496a-8b98-ee45f021e463
_Yes, focus band can activate before being knocked off as intended._

## How to test the changes?

Override opponent held items to berries and other items. Override starter moves for Incinerate and Knock Off.

## Checklist
- [See Above] There is no overlap with another PR?
- [x] Are the changes visual?


MPekelny:
I agree, this is a more complete implementation than mine, so this is the PR that should get merged. I will close mine. 

The only thing is I believe because there are edge cases not handled, even though it is 99% implemented, it still needs to be marked as partial. A lead dev can confirm if this is the case.

src/data/move.ts
```diff
@@ -1474,37 +1478,53 @@ export class StealHeldItemChanceAttr extends MoveEffectAttr {
...
   }
 
-  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): Promise<boolean> {
-    return new Promise<boolean>(resolve => {
-      const rand = Phaser.Math.RND.realInRange(0, 1);
-      if (rand >= this.chance) {
-        return resolve(false);
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean { 
+
+      if (!this.berriesOnly && target.isPlayer()) { // Wild Pokemon cannot knock off Player Pokemon's held items.
```
innerthunder, line 1498: Not sure if disabling this entirely against player Pokemon is the way to go. Limiting this effect to only remove berries on player Pokemon seems like a good middle ground; it would prevent the player from losing key/rare items while staying more faithful to the move's functionality.

Arxalc: I was following the wiki ["Wild Pokémon cannot remove held items of the player's Pokémon."](https://bulbapedia.bulbagarden.net/wiki/Knock_Off_(move)) but I can change it if desired for PokeRogue's balancing.

src/data/move.ts
```diff
@@ -1474,37 +1478,53 @@ export class StealHeldItemChanceAttr extends MoveEffectAttr {
...
+ * Not Implemented Cases: (Same applies for Thief)
+ * "If the user faints due to the target's Ability (Rough Skin or Iron Barbs) or held Rocky Helmet, it cannot remove the target's held item."
+ * "If Knock Off causes a Pokémon with the Sticky Hold Ability to faint, it can now remove that Pokémon's held item."
+ */
 export class RemoveHeldItemAttr extends MoveEffectAttr {
-  private chance: number;
+  private berriesOnly: boolean; 
 
-  constructor(chance: number) {
+  constructor(berriesOnly: boolean) { // Optional restriction for item pool to berries only i.e. Differentiating Incinerate and Knock Off
```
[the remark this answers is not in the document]
Arxalc: Done. Should be ready to review again.
Kept opponents being unable to knock off player items as is. Unless specified by you.

src/data/move.ts
```diff
@@ -1474,37 +1478,57 @@ export class StealHeldItemChanceAttr extends MoveEffectAttr {
...
+      const cancelled = new Utils.BooleanHolder(false);
+      applyAbAttrs(BlockItemTheftAbAttr, target, cancelled); // Check for abilities that block item theft
+
+      if(cancelled.value == true)
+        return false;
+
+      // Considers entire transferrable item pool by default (Knock Off). Otherwise berries only if specified (Incinerate).
+      let heldItems = this.getTargetHeldItems(target).filter(i => i.getTransferrable(false));
+      if (this.berriesOnly) {
+        heldItems = heldItems.filter(m => m instanceof BerryModifier && (m as BerryModifier).pokemonId === target.id, target.isPlayer());
```
OrangeRed, line 1513: You don't need this `m as BerryModifier`. Because you did `m instanceof BerryModifier` right before this typescript is smart enough to infer the type of `m` without you telling it using `as`

OrangeRed (changes requested):

src/data/move.ts
```diff
@@ -1474,37 +1478,57 @@ export class StealHeldItemChanceAttr extends MoveEffectAttr {
...
       }
 
-      resolve(false);
-    });
+      if (heldItems.length) {
+        const removedItem = heldItems[user.randSeedInt(heldItems.length)];
+
+        // Decrease item amount and update icon
+        if (!--removedItem.stackCount) {
+          target.scene.removeModifier(removedItem, !target.isPlayer());
```
Fractaos, line 1521: I think it's "important" to check if the item has properly been removed (the return of the `removeModifier` method) to avoid displaying the message once the target pokemon is dead for instance.

I'm not 100% sure so it needs to be verified but if you don't do this check you might have the message that get displayed even if nothing got actually removed due to K.O before being applied.

Fractaos:
I also do agree that this is a more complete implemention of my own PR, which I'll close, thanks for the contribution 👍

src/data/move.ts
Arxalc: ...After reviewing the removeModifier method, I realized it's seeing whether it's removing a form-changing/Tera Shard item. Because we've already filtered out those items, this check is actually useless so I'm removing it.

I have double checked that items will be still be removed if the target faints.

Fractaos (approved):
LGTM

Tempo-anon (approved):

PR #1673, opened 2024-06-01
[Feature] Move to start button when you can't add party anymore

Ohmry:
## What are the changes?

Move to start button when you can't add party anymore in starter-select-ui.

I thought it is better that move to start button automatically when user can't select starter pokemon anymore.

## What did change?

Added some codes that decide if move to start button in tryUpdateValue() and if user can't select any starter pokemon, send cursor to start button.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/13056250/526ab3c3-f922-473f-8028-e32476f507ce

## How to test the changes?

check out my PR and you can test

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -216,6 +216,7 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
   private canCycleNature: boolean;
   private canCycleVariant: boolean;
   private value: integer = 0;
+  private canAddParty: boolean;
 
   private assetLoadCancelled: Utils.BooleanHolder;
   private cursorObj: Phaser.GameObjects.Image;
@@ -1030,6 +1031,12 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
                     this.tryStart();
                   }
                   this.updateInstructions();
+
+                  if (!this.canAddParty) {
+                    this.startCursorObj.setVisible(true);
+                    this.setGenMode(true);
+                  }
+
                   ui.playSelect();
                 } else {
                   ui.playError();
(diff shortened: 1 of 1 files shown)


flx-sta:
@Ohmry may I ask why you chose a field over a method ? Idk if there is any performance benefit but a field for that feels impractical

Ohmry:
> @Ohmry may I ask why you chose a field over a method ? Idk if there is any performance benefit but a field for that feels impractical

I created new method for calculate wheter can I select more pokemon at first time. but I found the code to set alpha value of all species. It used to for loop for explore all species. I think new method has to use for loop too. 

then for loop use twice in tryUpdateValue and new method. so I create the logic that calculates remained staterValue in tryUpdateValue() not create new method.

flx-sta:
Does it reset the field when restarting ?

Ohmry:
> Does it reset the field when restarting ?

tryUpdateValue function is called every select pokemon. and `canAddParty` field set false before exploring all species at line 2105.

RNGezeus:
Given the fact that a there is currently a popup to start when you max out the point score, as well as the ENTER hotkey shortcut that acts same as start, is this mostly a just for visuals or is it meant for early players who may not have enough dex entries that always add up to 10 points.

In my personal case I quite often enter last mon, then back out and adjust moveset order/ability/nature so personally it's a downgrade. But only my opinion doesn't matter much for the vibe check as a whole.

Ohmry:
> Given the fact that a there is currently a popup to start when you max out the point score, as well as the ENTER hotkey shortcut that acts same as start, is this mostly a just for visuals or is it meant for early players who may not have enough dex entries that always add up to 10 points.
> 
> In my personal case I quite often enter last mon, then back out and adjust moveset order/ability/nature so personally it's a downgrade. But only my opinion doesn't matter much for the vibe check as a whole.

I didn't know that ENTER hotkey. I thoguth so uncomportable that move to start button when I selected all but remain points. but if hotkey exists, this PR is not nessary I think. your optiion might be right.

bennybroseph:
We would like to add this feature, so re-opening this PR.

src/ui/starter-select-ui-handler.ts
```diff
@@ -2094,11 +2101,26 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
...
+        const speciesSprite = this.starterSelectGenIconContainers[g].getAt(s) as Phaser.GameObjects.Sprite;
+
+        if (remainValue >= speciesStarterValue) {
+          speciesSprite.setAlpha(1);
+          if (speciesStarterDexEntry?.caughtAttr) {
+            this.canAddParty = true;
+          }
+        } else {
+          speciesSprite.setAlpha(0.375);
+        }
```
bennybroseph, line 2120: Could we get a code comment or two in this section to explain?

bennybroseph (changes requested):

src/ui/starter-select-ui-handler.ts
Ohmry: Sure, 
- `speciesStartValue` is cost of each pokemon for selecting starter.
- `speciesStarterDexEntry` is used to detected whethere pokemon is able to select. 
- `speciesSprite` variable is each pokemon sprite for setting alpha value in starter-select-ui.

If `speciesStarterValue` is less than or equal to `remainValue`, the user can select this pokemon. so the alpha value of this pokemon sprite set to 1. otherwise, the user can't select this pokemon. so the alpha value of this pokemon sprite will set to .375.

in line 2115 means although the user can select a pokemon based on remain cost, if this pokemon is not registered as starter, the user can't selected it. so `canAddParty` variable is set `true` if this pokemon is the user has cost greater than pokemon's cost and this pokemon is registered as starter pokemon.

Ohmry: In addition, I needed a variable to dected whether tIn addition, I needed a variable to detect whether the user could select pokemon more. at first, I created new method for that. this method uses  loops for searching all pokemon. but I found 'tryUpdateValue()' method uses loops too. If I created new method, loops may works twice. so I didn't create new method then I added some code for detecting that to this line.he user colud select

flx-sta: I think what he meant was he wants you to add comments into the actual code.

Ohmry: I did 6c4fe716cdb7a47b871f4608aa023dbc8e026519

bennybroseph (approved):

PR #1674, opened 2024-06-01
[Localization] Add localization about message overwrite the data in selected slot

Ohmry:
## What are the changes?

Add localization about message "overwrite the data in selected slot"

There is a need to translate it into Korean.

## What did change?

- Added save-slot-select-ui-handler.ts into each localization folders
- Added "saveSlotSelectorUiHandler" resource into i18next.resources
- Added "saveSlotSelectUiHandler" into each localizatoin config files
- Translated "Overwrite the data in the selected slot?" to Korean
> Other languages remains with English

### Screenshots/Videos

image

## How to test the changes?

1. Check out this PR
2. Change language to Korean
3. You can see these changes when you save game data after selecting all of your starter Pokémon.

## Checklist
- [X] Are the changes visual?


src/locales/de/save-slot-select-ui-handler.ts
```diff
@@ -0,0 +1,5 @@
+import { SimpleTranslationEntries } from "#app/plugins/i18n";
+
+export const saveSlotSelectUiHandler: SimpleTranslationEntries = {
+  "Overwrite_the_data": "Overwrite the data in the selected slot?"
```
CodeTappert, line 4:
```suggestion
  "Overwrite_the_data": "Die Daten des ausgewählten Speicherstandes überschreiben?"
```

(Could you please test if this is too long?)

Ohmry: image

Sure!

CodeTappert: Hmm. Seems barely to long...

CodeTappert: Okay. I need to make it shorter. So 

```suggestion
  "Overwrite_the_data": "Den ausgewählten Speicherstand überschreiben?"
```

Ohmry: image

Much better I think

CodeTappert: Yes a lot better! Thank you

src/locales/ko/save-slot-select-ui-handler.ts
```diff
@@ -0,0 +1,5 @@
+import { SimpleTranslationEntries } from "#app/plugins/i18n";
+
+export const saveSlotSelectUiHandler: SimpleTranslationEntries = {
+  "Overwrite_the_data": "선택한 슬롯에 데이터를 덮어쓰시겠습니까?"
```
sodaMelon, line 4: 한국어로 의미를 정확히 살려서 번역하셨네요👍

sodaMelon (approved):
LGTM 👍  Korean users can enjoy games more clearly~~

Ohmry:
@Tempo-anon 
Hi, could you please check this PR? the point of PR is to change some code that show message about saving slot. 

Translate to Korean is done. and confirmed by @sodaMelon. and I created some files for localizing in each localize folders. I hope to help that other translators let know to exists things needs to localize.

If you are not in charge, can you tell who is? I want to feedback of this PR.

Thanks.

bennybroseph:
Does this really need to be in its own file? Can we add it to a menu one that we already have or add more translations to this?

Ohmry:
> Does this really need to be in its own file? Can we add it to a menu one that we already have or add more translations to this?

@bennybroseph 
I thought just it would be easy to find location of translation files by UI file's name. but if not, I think so it is good idea to add translation into 'menu.ts' too. I modified the code and pushed again. please check it.

Ohmry:
Already merged from #1728

PR #1675, opened 2024-06-01
[WIP] ESLint Import Sorting

bennybroseph:

Adds more ESLint formatting consistency. This one requires imports to be sorted a certain way.

## What did change?

All imports will be sorted according to the ESlint config.

## Checklist


JakubHanko:
Would it be possible to incorporate rule mentioned in #1356 ? I think it makes sense to add it to this PR since the scope of the changes is already huge.

bennybroseph:
This did not merge well at all

OrangeRed:
@bennybroseph 
I don't think adding this to ESlint is a good choice because it puts the onus on developers to sort imports by showing errors. If we instead employ an auto formatter then this becomes a lot easier to enforce using a plugin like https://github.com/trivago/prettier-plugin-sort-imports

bennybroseph:
> @bennybroseph I don't think adding this to ESlint is a good choice because it puts the onus on developers to sort imports by showing errors. If we instead employ an auto formatter then this becomes a lot easier to enforce using a plugin like https://github.com/trivago/prettier-plugin-sort-imports

If we go that route, what other things should be using prettier? I wouldn't want add it just for imports

OrangeRed:
I mean Prettier just formats your code so you can use it automatically enforce the eslint config. Instead of spitting out errors on pre-commit / pre-push, prettier will automatically format the code in such a way that there are no eslint errors.

Like auto format to add semi colons, automatically convert tabs into spaces, etc. (whatever is defined in the eslint config)

bennybroseph:
I thought ESLint's `--fix` already did that though. It just doesn't work on imports for some reason which is why I'm asking what else ESLint can't fix that prettier can fix, like imports

OrangeRed:
There is this github conversation chain with the Eslint team about --fix not working on imports by default/
https://github.com/eslint/eslint/issues/11542

The way around that is to use one of the two plugins recommended at the bottom of the conversation.
This one has more npm downloads so maybe consider using this over the other one. I haven't looked into how it works or if its desirable but its worth giving a read.
https://www.npmjs.com/package/eslint-plugin-simple-import-sort

PR #1677, opened 2024-06-01
[QoL] Don't offer ball rewards when you have max in Classic mode

Fontbane:
## What are the changes?

Balls are no longer offered in rewards if you have max of them in classic.

Because it's stupid if you have 99 Great Balls and the rewards gives you Great Balls and you can take it and not get anything because you already have max so you just waste your reward for that floor.

## What did change?

I added a check in the modifierPool that sets the weights of a ball to 0 if you have 99 (or more) of that ball and you're in Classic.

### Screenshots/Videos

How it currently works:
image
Wave 1, Red Ball in rewards
image
Wave 2, Too many balls
image
Wave 2, no Red Balls in rewards
image
Wave 2, no Red Balls after 14 rerolls with 99 in bag
image
I got into a double battle with 2 flying types after that so I had to restart but here it is still offering Red Ball when I have 98
image
Change doesn't apply in Endless:
image

## How to test the changes?

In src/modifier/modifier-type.ts, change the following line:
`export const modifierTypes = {
  POKEBALL: () => new AddPokeballModifierType("pb", PokeballType.POKEBALL, 5),`
The last parameter in AddPokeballModifierType, currently 5, is the amount of balls you get from one reward. Change it to 94 or more so you can hit 99 after the first wave.
In the same file, in modifierPool:
`const modifierPool: ModifierPool = {
  [ModifierTier.COMMON]: [
    new WeightedModifierType(modifierTypes.POKEBALL, (party: Pokemon[]) => (party[0].scene.gameMode.isClassic && party[0].scene.pokeballCounts[PokeballType.POKEBALL] >= 99) ? 0 : 6, 6),`
Change the two 6's at the end (which are the weight and the maximum weight, respectively) to something really high so you can get 99 ASAP and ensure that luck isn't a factor when you're testing if it was removed from the pool.

Then in phases.ts, at the end of SelectModifierPhase.getRerollCost:
`return Math.min(Math.ceil(this.scene.currentBattle.waveIndex / 10) * baseValue * Math.pow(2, this.rerollCount), Number.MAX_SAFE_INTEGER);`

Change Number.MAX_SAFE_INTEGER to 1, or just make the whole line `return 1`. This will set the reroll cost to 1 so you can test rerolling many times.

Then start a new run, get to 99 red balls and bear witness to its absence in the next set of rewards.

## Checklist
- [X] Are the changes visual?


Tempo-anon:
Can you please fill out the PR description properly? And also include some screenshots/videos of testing where you override a bunch of money at floors 190+ and reroll a bunch of times?

HighMans:
> This PR removes balls from the reward pool from wave 189 onwards in classic, as they're no longer usable after that.

Is that true though? You can capture paradox-mon that you've got as starters.

Fontbane:
> > This PR removes balls from the reward pool from wave 189 onwards in classic, as they're no longer usable after that.
> 
> Is that true though? You can capture paradox-mon that you've got as starters, no?

Only if they're shiny

HighMans:
> > > This PR removes balls from the reward pool from wave 189 onwards in classic, as they're no longer usable after that.
> > 
> > 
> > Is that true though? You can capture paradox-mon that you've got as starters, no?
> 
> Only if they're shiny, but that rare edge case isn't worth the hindrance on most other runs.

That's not true, I'm 99% sure I've caught Iron Bundle with a master ball during a classic run that wasn't shiny.

flx-sta (changes requested):
You should put the wave number into a descriptive variable.
Something like:
```ts
CLASSIC_POKEBALL_REWARD_MAX_WAVE=  184
```

maybe something shorter

RNGezeus:
> > > This PR removes balls from the reward pool from wave 189 onwards in classic, as they're no longer usable after that.
> > 
> > 
> > Is that true though? You can capture paradox-mon that you've got as starters, no?
> 
> Only if they're shiny, but that rare edge case isn't worth the hindrance on most other runs.

 Thats false, you can capture non-shiny paradoxes as well, provided  hatched them first.

Its one of the side reasons to run Classic very late in your collection, to scan paradoxes to improve their IVs

Tempo-anon (changes requested):
This makes changes to classic _and_ endless which seems like it'd be larger than the initial scope especially since endless lets players farm items/balls pretty easily and this will change pool weights.

Also, you still haven't filled out the proper PR template along with proof of testing

Fontbane:
> This makes changes to classic _and_ endless which seems like it'd be larger than the initial scope especially since endless lets players farm items/balls pretty easily and this will change pool weights.
> 
> Also, you still haven't filled out the proper PR template along with proof of testing

Should I make it classic-only then? Or should it just be classic-only for the higher tiers, and the same changes would apply to only red balls in endless?

src/modifier/modifier-type.ts
brain-frog, line 1: If the same arrow function is being used every time can this just be converted into a reusable function above and used as a parameter
also please create a descriptive const for 99 since its being used every time
`MAX_BALL_COUNT` or something similar

brain-frog (changes requested):

src/modifier/modifier-type.ts
```diff
@@ -1236,9 +1236,15 @@ interface ModifierPool {
   [tier: string]: WeightedModifierType[]
 }
 
+const MAX_BALLS = 99;
+
+function CheckForMaximumBalls(party: Pokemon[], ballType: PokeballType): boolean {
```
Tempo-anon, line 1241: 1. Should add documentation to new functions
2. Since this is a function the first letter should not be capitalized
3. A boolean function should make it clear it returns a boolean `hasMaximumBalls` for example

src/modifier/modifier-type.ts
```diff
@@ -1236,9 +1236,22 @@ interface ModifierPool {
   [tier: string]: WeightedModifierType[]
 }
 
+const MAX_BALLS = 99;
+
+
+/**
+ * Used to check if the player has max of a given ball type
```
brain-frog, line 1243: specify this is for classic in comment please

flx-sta:
@Fontbane Marks as resolved when done 
image

brain-frog (approved):

Tempo-anon (approved):

Issue #1669, opened 2024-05-31
[BUG] Switch moves that KO can softlock


Wairpre:
**Describe the bug**
Knocking out a wild pokemon with a switch move (i.e. U-Turn) can softlock the game after a new pokemon is switched in (specifically a shiny, but could include non-shinies). 

**To Reproduce**
1. Have a pokemon with a switch move (i.e. U-Turn, Volt Switch, etc.) and another pokemon (shiny in this case)
2. Knock out a wild pokemon with a switch move
3. Send out other pokemon
4. Softlock

**Expected behavior**
After knocking out the wild pokemon, the game should progress after switching in the player's next pokemon

**Screenshots / Videos**
Used team (note shiny Porygon):
![image](https://github.com/pagefaultgames/pokerogue/assets/56366921/0dc4453f-b6bf-4a16-9758-199895caf020)
Uses U-Turn:
![image](https://github.com/pagefaultgames/pokerogue/assets/56366921/5ecfc93b-a798-4c73-8755-431d14969afb)
Swaps in shiny Porygon:
![image](https://github.com/pagefaultgames/pokerogue/assets/56366921/2a50f736-165a-4f8b-82cc-291df1c5e5a4)
Softlock

**Device**
Opera GX, Windows

**Additional context**
Softlock occurs specifically after the next pokemon is swapped in, so the case of a shiny could be the cause of the softlock

PR #1679, opened 2024-06-01
[Bug] Fix ability Download switch in (1669)

Nohbody:
## What are the changes?
1. Changes the Download ability.

1. Fixes: https://github.com/pagefaultgames/pokerogue/issues/1669

## What did change?
1. Check if the opponent has fainted yet before trying to check the stats for Download ability.
  -- This will no longer soft-lock if a Pokemon with Download switches in when all opponents have fainted.

## How to test the changes?
1. Create save state using overrides.ts:
```
export const STARTING_LEVEL_OVERRIDE: integer = 100;

export const MOVESET_OVERRIDE: Array<Moves> = [Moves.U_TURN];
```

2. Start the game with a Porygon as your second Pokemon (that will be switched into).
3. Use the move U-Turn with your level 100 Pokemon to faint the first wave opponent.
4. Switch into Porygon. The ability Download should not trigger and the game should continue.

## Checklist


src/data/ability.ts
```diff
@@ -1662,6 +1662,7 @@ export class DownloadAbAttr extends PostSummonAbAttr {
   private enemySpDef: integer;
   private enemyCountTally: integer;
   private stats: BattleStat[];
+  private opponents: Pokemon[];
```
Tempo-anon, line 1665: Does this need to be defined as a class variable?

src/data/ability.ts
```diff
@@ -1677,8 +1678,10 @@ export class DownloadAbAttr extends PostSummonAbAttr {
     this.enemySpDef = 0;
     this.enemyCountTally = 0;
 
-    if (pokemon.getOpponents()[0].summonData !== undefined) {
-      for (const opponent of pokemon.getOpponents()) {
+    this.opponents = pokemon.getOpponents();
+
+    if (this.opponents[0].summonData !== undefined && this.opponents[0].summonData.battleStats[Stat.HP] > 0) {
+      for (const opponent of this.opponents) {
```
Xavion3, line 1684: This is not how you check this, and I'm kind of surprised it works if I'm honest.

Can you just remove the encasing if statement here entirely and just directly run the loop and rounding?

Nohbody: I'm noticing Download isn't triggering at all with these changes. oops
What am I supposed to check for the opponent's current hp

Nohbody:
@Tempo-anon @Xavion3 It should work now.

src/data/ability.ts
```diff
@@ -1677,15 +1677,13 @@ export class DownloadAbAttr extends PostSummonAbAttr {
     this.enemySpDef = 0;
     this.enemyCountTally = 0;
 
-    if (pokemon.getOpponents()[0].summonData !== undefined) {
```
Nohbody, line 1680: This caused the soft-lock when both opponents were fainted `pokemon.getOpponents()` = `[]` so `pokemon.getOpponents()[0]` = `undefined` which caused a `TypeError: Cannot read properties of undefined`.

Nohbody: Alternatively, it could be fixed by changing it to `pokemon.getOpponents()[0]?. !== undefined` but as far as I can tell the check isn't even necessary.

src/data/ability.ts
```diff
@@ -1677,15 +1677,13 @@ export class DownloadAbAttr extends PostSummonAbAttr {
...
-        this.enemyCountTally++;
-        this.enemyDef += opponent.getBattleStat(Stat.DEF);
-        this.enemySpDef += opponent.getBattleStat(Stat.SPDEF);
-      }
-      this.enemyDef = Math.round(this.enemyDef / this.enemyCountTally);
-      this.enemySpDef = Math.round(this.enemySpDef / this.enemyCountTally);
+    for (const opponent of pokemon.getOpponents()) {
+      this.enemyCountTally++;
+      this.enemyDef += opponent.getBattleStat(Stat.DEF);
+      this.enemySpDef += opponent.getBattleStat(Stat.SPDEF);
```
torranx, line 1683: question: is there an instance where `pokemon.getOpponents()` returns an empty array- meaning no enemies ex. you summoned a pokemon where an opponent havent yet? if so, won't this crash the game because youre accessing an element that isnt there?

Nohbody: When you use a for...of loop on an empty array the for loop conditional short-circuits and continues pass the loop. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

image

torranx: oh I see, I've been using js and this is quite new to me lol. thanks for clarifying

src/data/ability.ts
```diff
@@ -1693,15 +1693,13 @@ export class DownloadAbAttr extends PostSummonAbAttr {
...
-    if (pokemon.getOpponents()[0].summonData !== undefined) {
-      for (const opponent of pokemon.getOpponents()) {
-        this.enemyCountTally++;
-        this.enemyDef += opponent.getBattleStat(Stat.DEF);
-        this.enemySpDef += opponent.getBattleStat(Stat.SPDEF);
-      }
-      this.enemyDef = Math.round(this.enemyDef / this.enemyCountTally);
-      this.enemySpDef = Math.round(this.enemySpDef / this.enemyCountTally);
+    for (const opponent of pokemon.getOpponents()) {
+      this.enemyCountTally++;
```
Tempo-anon, line 1697: Is enemyCountTally even needed if both enemyDef and enemySpDef will be divided by it?

Tempo-anon (approved):
Get this in to fix download crashing players

Issue #1657, opened 2024-05-31
[BUG] Stealth rock not ignoring Delta stream


RNGezeus:
Searches for similar reports?
I cleared filter and searched "Delta stream" and "Stealth Rock" respectively, didn't see anything.

**Describe the bug**
Stealth rock is currently being affected by Strong winds from delta stream, making it do less damage.

**Expected behavior**
Would expected SR to do its intended damage,

**Screenshots / Videos**
https://github.com/pagefaultgames/pokerogue/assets/46976469/96f2ceec-b245-48a1-99da-2e54db69af57

https://github.com/pagefaultgames/pokerogue/assets/46976469/9bf9418d-5218-4442-a78c-3e6928328e79

**Additional context**
https://bulbapedia.bulbagarden.net/wiki/Stealth_Rock_(move)
"Stealth Rock is unaffected by strong winds"
https://bulbapedia.bulbagarden.net/wiki/Strong_winds
"This has no effect on Stealth Rock and Anticipation"

PR #1682, opened 2024-06-01
[Bug] Fix strong winds' interaction with Stealth Rock and Anticipation (#1657)

Corrade:
## What are the changes?
Made Stealth Rock and Anticipation ignore the effects of strong winds.

We currently fail to uphold that [Stealth Rock and Anticipation should not be affected by strong winds](https://bulbapedia.bulbagarden.net/wiki/Delta_Stream_(Ability)).

The Stealth Rock portion of the bug was reported in issue #1657.

## What did change?
We allowed an optional parameter for ignoring strong winds in getAttackTypeEffectiveness() and provided its argument in the Stealth Rock and Anticipation impls.

### Screenshots/Videos
The following GIFs show a single sequence that demos both the changes described below.
#### Before
strong_winds_stealth_rock_anticipation_before

#### After
strong_winds_stealth_rock_anticipation_after

### Anticipation
- Override starter to a flying type with ability Anticipation
- Override opponent with ability Delta Stream (for strong winds) and an ice, electric or rock move
- Observe the anticipation alert correctly displaying despite strong winds
### Stealth Rock
- Override starter with the move Stealth Rock
- Override opponent to a flying type with ability Delta Stream (for strong winds)
- Set Stealth Rock
- Defeat the first opponent
- Observe the switched-in flying type opponent take 2x Stealth Rock damage, correctly ignoring strong winds' resistance

## Checklist
- [X] The PR is self-contained and cannot be split into smaller PRs? **>> We could split this into two separate fixes for Stealth Rock and Anticipation respectively, but I feel the complexity is already trivial and there's a uniting strong winds theme**
- [X] Are the changes visual? **>> No**


src/field/pokemon.ts
```diff
@@ -1070,7 +1070,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     return (!cancelled.value ? typeMultiplier.value : 0) as TypeDamageMultiplier;
   }
 
-  getAttackTypeEffectiveness(moveType: Type, source?: Pokemon): TypeDamageMultiplier {
+  getAttackTypeEffectiveness(moveType: Type, source?: Pokemon, ignoreStrongWinds: boolean = false): TypeDamageMultiplier {
```
OrangeRed, line 1073: I'm personally against adding such a specific boolean into a function but the alternative will require a lot of changes so this is fine I think

OrangeRed (approved):

Tempo-anon (approved):

PR #1276, opened 2024-05-23
useMove + pokemon affix localization

laeticiapierre:
Issue #1274

EDIT 05/27: 
ENG
Capture d’écran 2024-05-27 à 17 13 48
Capture d’écran 2024-05-27 à 17 13 54

FR
Capture d’écran 2024-05-27 à 17 12 53
Capture d’écran 2024-05-27 à 17 13 01

PR #1684, opened 2024-06-01
[QoL] Show message when quick claw is triggered

torranx:
## What are the changes?
- Added message when quick claw is triggered: _"{{pokemonName}} used its quick claw to move faster!"_

- To avoid confusion when slower pokemon "suddenly" gets the first action, the same way it shows a message when grip claw is triggered.

## What did change?
- Message when quick claw is triggered

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/68144167/1388aab4-3b25-41ea-bd7b-2c2ccd9e0c5a

## How to test the changes?
- set `STARTER_SPECIES_OVERRIDE` to a slow pokemon like Shuckle
- set `STARTING_HELD_ITEMS_OVERRIDE: Array<ModifierOverride> = [{name: "QUICK_CLAW", count: 1}];`

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/modifier/modifier.ts
+++ b/src/modifier/modifier.ts
@@ -775,6 +775,9 @@ export class BypassSpeedChanceModifier extends PokemonHeldItemModifier {
 
     if (!bypassSpeed.value && pokemon.randSeedInt(10) < this.getStackCount()) {
       bypassSpeed.value = true;
+      if (this.type instanceof ModifierTypes.PokemonHeldItemModifierType && this.type.id === "QUICK_CLAW") {
+        pokemon.scene.queueMessage(getPokemonMessage(pokemon, " used its quick claw to move faster!"));
+      }
       return true;
     }
 


src/modifier/modifier.ts
```diff
@@ -775,6 +775,9 @@ export class BypassSpeedChanceModifier extends PokemonHeldItemModifier {
 
     if (!bypassSpeed.value && pokemon.randSeedInt(10) < this.getStackCount()) {
       bypassSpeed.value = true;
+      if (this.type instanceof ModifierTypes.PokemonHeldItemModifierType && this.type.id === "QUICK_CLAW") {
+        pokemon.scene.queueMessage(getPokemonMessage(pokemon, " used its quick claw to move faster!"));
```
Tempo-anon, line 779: Should probably be localized

torranx: unless we remove the pokemon prefix coming from the `getPokemonMessage()` and place the string to a regular i18n locale translation, I dont think this is possible to be localized. please see `src/messages.ts`

so we have 2 options. 
- keep the pokemon prefix but wont be localized
- or remove the pokemon prefix so localization would be possible

flx-sta: How about you abstract it to make it work?
```suggestion
        const pokemonPrefix = getPokemonPrefix(pokemon);
        pokemon.scene.queueMessage(i18n.t(:itemUse:quickclaw', {pokemon, pokemonPrefix}));
```

and then in your translations something like
```ts
const de: Record<string, any> = {
  "itemUse:quickclaw": "{{pokemonPrefix}} {{pokemon.name}} used its quick claw to move first!"
}
```

_(just an inspiration, not actual code)_

torranx: the `getPokemonPrefix()` returns an untranslated text too tho. the whole `src/messages.ts` does not have translation functionality. i think it needs a separate PR for it

flx-sta: I agree.
but you can still translate the part you are in control of

torranx: i cant translate only part of it tho, and i wouldnt wanna touch anything from `messages.ts` on this PR because its being used all over the codebase.
also for reference someone is already working on the translation https://github.com/pagefaultgames/pokerogue/pull/1276/files

flx-sta: @Tempo-anon it's now up to you

Tempo-anon (approved):
Got user complaints about this. I'm fine with not having it localized just yet and having the localization in a different PR.

flx-sta:
@Tempo-anon localization support can be found in this PR
https://github.com/pagefaultgames/pokerogue/pull/1276

PR #1694, opened 2024-06-01
[QoL] Add ALL option when transferring items to Pokemons 

francktrouillez:
## What are the changes?

This PR introduces the "ALL" option in the transfer selection dropdown when transferring items from one Pokemon to another.

I recently was playing in the Endless mode, and started to build a "mule", that would carry items until I find a good Pokemon to transfer them to. I realized that I had to click multiple times to transfer all items from one Pokemon to another, and thought that it would be a good idea to have an "ALL" option in the dropdown.

## What did change?

From a user experience, the dropdown now has an "ALL" option, that will transfer all items from the source Pokemon to the target Pokemon. This takes the limitations of the current transfer system into account, where you cannot transfer to a Pokemon that has already reached the maximum number of a specific item.

Code-wise, I've added a new attribute to the `PartyUiHandler` to track whether or not we need to transfer all items.

I also took the liberty to add a translation file for `PartyUiHandler`, currently only containing the translation for the "ALL" option for `en`.

### Screenshots/Videos

- Before

  https://github.com/pagefaultgames/pokerogue/assets/57403591/3b86db61-ecd5-4ad9-aa13-ba1352804566

- After
  - When we can transfer all items

    https://github.com/pagefaultgames/pokerogue/assets/57403591/0fdec769-7135-4f9d-8434-7f320e1d3a96

  - When the receive already has the maximum number of items

    https://github.com/pagefaultgames/pokerogue/assets/57403591/f8183ff0-a64b-4317-9bd6-fc451c531a67

## How to test the changes?

In order to test the changes, you can modify the `src/overrides.ts` file to include the following code:

```ts
export const STARTING_LEVEL_OVERRIDE: integer = 100; // Just to be quicker to beat the first Pokemon
// ...
export const STARTING_HELD_ITEMS_OVERRIDE: Array<ModifierOverride> = [
  { name: "LUCKY_EGG" },
  { name: "MULTI_LENS", count: 3 }
];
```

Then, you can start a new game with 2 Pokemon, and try to transfer items from one Pokemon to another. You should see the "ALL" option in the dropdown whenever there's more than one item to transfer.

To try the case where the target Pokemon already has the maximum number of items, you can modify the count of the `MULTI_LENS` item to be 5, and try to transfer items from the source Pokemon to the target Pokemon, since the maximum number of `MULTI_LENS` is 3 per Pokemon.

## Checklist
- [x] Are the changes visual?


Greenlamp2 (approved):

Greenlamp2:
Actually, when transfering all, it also tries to transfer the "ALL" label, spitting error in the console and transfer nothing.
savefile to reproduce:
https://file.io/CfkwBu4mC07H

Greenlamp2 (changes requested):
Fix needed to prevent the "ALL" label to be considered as a modifier.

francktrouillez:
@Greenlamp2 
Thanks for the feedback!

> Fix needed to prevent the "ALL" label to be considered as a modifier.

It wasn't really considered as a modifier, but the option cursor to detect whether we were in a `transferAll` state or not wasn't well defined.

Indeed, it wasn't working when the scroll appear in the selection. I wrongly assumed that `party-ui-handler#options` was giving all the options, and not just the displayable options.

I fixed the issue that basically use the "current" option cursor, instead of the "whole" option cursor. (019823efa97b69b636bea75d62f16928c40b28c6)

I'm attaching a video using your save showcasing the feature using your save:

https://github.com/pagefaultgames/pokerogue/assets/57403591/62991e08-5176-475c-8139-079612c19fab

Thanks for your help, let me know if there's anything else I can do!

Greenlamp2:
One last thing, could you exclude the stats modifiers items like protein, carbon etc ? because they are not transferable items.

francktrouillez:
@Greenlamp2 

Thanks for the quick response!

> One last thing, could you exclude the stats modifiers items like protein, carbon etc ? because they are not transferable items.

I wasn't able to reproduce the issue. Actually, I've been only using items that are transferable for this `All` option, as you can see here:
https://github.com/francktrouillez/pokerogue/blob/777193eea7f2372642d9f8a7b0354345970cbf2f/src/ui/party-ui-handler.ts#L321

which is using this function that aims at getting transferable items:
https://github.com/francktrouillez/pokerogue/blob/777193eea7f2372642d9f8a7b0354345970cbf2f/src/ui/party-ui-handler.ts#L300-L301

I wasn't able to reproduce the situation where stat modifiers are transferred.

https://github.com/pagefaultgames/pokerogue/assets/57403591/90779680-d407-4e60-9758-c5c797360913

Did I miss something? Let me know if there's anything I can do!

Greenlamp2:
> @Greenlamp2
> 
> Thanks for the quick response!
> 
> > One last thing, could you exclude the stats modifiers items like protein, carbon etc ? because they are not transferable items.
> 
> I wasn't able to reproduce the issue. Actually, I've been only using items that are transferable for this `All` option, as you can see here:
> 
> https://github.com/francktrouillez/pokerogue/blob/777193eea7f2372642d9f8a7b0354345970cbf2f/src/ui/party-ui-handler.ts#L321
> 
> which is using this function that aims at getting transferable items:
> 
> https://github.com/francktrouillez/pokerogue/blob/777193eea7f2372642d9f8a7b0354345970cbf2f/src/ui/party-ui-handler.ts#L300-L301
> 
> I wasn't able to reproduce the situation where stat modifiers are transferred.
> 
>  2024-06-15.01-07-46.mp4 
> Did I miss something? Let me know if there's anything I can do!

indeed, my bad, it was because each time i was doing a transfer, i saw that on the console:
image

but after further check, no stats modifier items were transfered nor any stats changed

Greenlamp2 (approved):

PR #1700, opened 2024-06-01
[Feature] Add buying specific species Egg with candy

jaimefd:
## What are the changes?

Added a new third way to spend starter candy, "Buy Egg", that can be used infinitely to exchange some candy for eggs that will always hatch that species. This gives players a use for their candies after unlocking passive and the 2 cost reductions, that isn't unbalanced since nothing is guaranteed: you may improve the IVs... or not, you may unlock an Egg Move... or not, etc.

Captura de pantalla 2024-06-01 204612

I am aware that implementing a feature doesn't mean it will be added. However I wanted to test myself and I believe this is a nice and interactive way to propose an idea to the community. I don't mind it being rejected, but I would love if it is considered a bit after seeing this example implementation and my arguments in favour of it, as it would really improve the game.

Candy progression is fun, as it allows players to have things to be excited for and to upgrade their favourites. However, once the passive and the two cost reductions are unlocked, continuing to use that Pokémon starts feeling like a waste. I am aware that the devs mentioned having plans to add more uses for candy.

However I still believe an option to buy eggs is good, as it feels like a natural way to progress with your starter. It preserves the randomness that roguelites are known for, keeping the players excited, which is better than other ideas that just allow you to straight up max an IV. This will make it balanced as it should not be strictly better than the other 2 ways of spending candy, while still allowing you to unlock many things like hidden ability, shinies, egg moves... if you are lucky, of course. And as a bonus, it also works as some sort of replacement for "breeding" for the time being.

## What did change?

- Added a new option in the candy exchange menu to spend candy to buy an egg, that will always hatch that starter
- I set the costs to 30 for 1-3 value starters, 20 for 4-5 value starters, 15 for 6-7 value and 10 for 8-9. No eggs for 10 cost Eternatus. These numbers are of course mostly for testing, and can easily be changed
- These eggs do not benefit from any extra rates, since they don't come from any gacha
- They have the same tier (and therefore, same waves to hatch) as if they came from gacha
- Eggs now have an EggSource attribute, currently only GACHA or CANDY
- Since the constructor and some functions in egg.ts have been changed, I have updated everywhere in the code where those are referenced accordingly, including EggData, to make sure nothing breaks
- The egg information menu will show if an egg was obtained through candy exchange and for what starter it is

## Screenshots/Videos

Here is a video that shows the process of buying the eggs, viewing the egg information, and hatching them:

https://github.com/pagefaultgames/pokerogue/assets/22176793/0a3d7ee9-4969-437b-93bd-e350960fc91d

https://github.com/pagefaultgames/pokerogue/assets/22176793/7e4d25a9-4deb-4d4b-bf68-a7c983e028a4

(had to split them in two because of the 10MB limit)

Also here's a quick test that gacha eggs are still working fine:

https://github.com/pagefaultgames/pokerogue/assets/22176793/77b4572d-1a6c-4c9c-bd2f-c1e7aaad182f

## How to test the changes?

Set the `export const IMMEDIATE_HATCH_EGGS_OVERRIDE: boolean = true;` to true in the overrides.ts file so that you just need to clear one wave to test the eggs. You will also probably need to go into Manage Data > Import Data in the game itself to import a data file with a bunch of candies so you can actually test it. I found my file in one of the pinned messages of the offiline-and-modding channel in the discord server.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/egg.ts
+++ b/src/data/egg.ts
@@ -24,21 +24,21 @@ export class Egg {
   public timestamp: integer;
   public source: EggSource;
   public gachaType: GachaType;
-  public species: Species;
+  public forcedSpecies: Species;
 
 
-  constructor(id: integer, hatchWaves: integer, timestamp: integer, source: EggSource, gachaType: GachaType = null, species: Species = null) {
+  constructor(id: integer, hatchWaves: integer, timestamp: integer, source: EggSource, gachaType: GachaType = null, forcedSpecies: Species = null) {
     this.id = id;
-    this.tier = species === null ? Math.floor(id / EGG_SEED) : getSpeciesTier(species);
+    this.tier = forcedSpecies === null ? Math.floor(id / EGG_SEED) : getSpeciesTier(forcedSpecies);
     this.hatchWaves = hatchWaves ?? getEggTierDefaultHatchWaves(this.tier);
     this.timestamp = timestamp;
     this.source = source;
     this.gachaType = gachaType;
-    this.species = species;
+    this.forcedSpecies = forcedSpecies;
   }
 
   isManaphyEgg(): boolean {
-    return (this.tier === EggTier.COMMON && !(this.id % 255)) || this.species	=== Species.PHIONE || this.species === Species.MANAPHY;
+    return (this.tier === EggTier.COMMON && !(this.id % 255)) || this.forcedSpecies	=== Species.PHIONE || this.forcedSpecies === Species.MANAPHY;
   }
 
   getKey(): string {
@@ -93,9 +93,6 @@ export function getEggHatchWavesMessage(hatchWaves: integer): string {
 export function getEggSourceDescriptor(scene: BattleScene, egg: Egg): string {
   switch (egg.source) {
   case EggSource.GACHA:
-    if (egg.gachaType === null) {
-      break;
-    }
     switch (egg.gachaType) {
     case GachaType.LEGENDARY:
       return `${i18next.t("egg:gachaTypeLegendary")} (${getPokemonSpecies(getLegendaryGachaSpeciesForTimestamp(scene, egg.timestamp)).getName()})`;
@@ -105,7 +102,7 @@ export function getEggSourceDescriptor(scene: BattleScene, egg: Egg): string {
       return i18next.t("egg:gachaTypeShiny");
     }
   case EggSource.CANDY:
-    return `${i18next.t("egg:sourceCandyExchange")} (${getPokemonSpecies(egg.species).getName()})`;
+    return `${i18next.t("egg:sourceCandyExchange")} (${getPokemonSpecies(egg.forcedSpecies).getName()})`;
   }
 }
 
@@ -131,16 +128,18 @@ export function getLegendaryGachaSpeciesForTimestamp(scene: BattleScene, timesta
 }
 
 /**
- * Obtains the egg tier of a starter species, using its starter value
+ * Obtains the egg tier of a species, using its starter value
  * @param species The {@linkcode Species} of which we are trying to obtain the egg tier
- * @returns the {@linkcode EggTier} from which this species usually hatches from, or {@linkcode EggTier.COMMON} if not a starter species
+ * @returns the {@linkcode EggTier} from which this species usually hatches from
  */
 export function getSpeciesTier(species: Species): EggTier {
   if (species === Species.PHIONE || species === Species.MANAPHY) {
     return EggTier.COMMON;
   }
 
-  const starterValue = speciesStarters[species] ?? 1;
+  const rootSpecies = getPokemonSpecies(species).getRootSpeciesId();
+  const starterValue = speciesStarters[rootSpecies];
+
   if (starterValue >= 8) {
     return EggTier.MASTER;
   }
--- a/src/data/egg.ts
+++ b/src/data/egg.ts
@@ -6,6 +6,11 @@ import i18next from "../plugins/i18n";
 
 export const EGG_SEED = 1073741824;
 
+export enum EggSource {
+  GACHA,
+  CANDY
+}
+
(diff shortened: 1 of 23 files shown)


src/data/egg.ts
```diff
@@ -6,6 +6,11 @@ import i18next from "../plugins/i18n";
 
 export const EGG_SEED = 1073741824;
 
+export enum EggSource {
```
Tempo-anon, line 9: Would it make sense to combine EggSource and GachaType into a single enum?

src/data/egg.ts
```diff
@@ -15,20 +20,25 @@ export enum GachaType {
 export class Egg {
   public id: integer;
   public tier: EggTier;
-  public gachaType: GachaType;
   public hatchWaves: integer;
   public timestamp: integer;
+  public source: EggSource;
+  public gachaType: GachaType;
+  public species: Species;
```
Tempo-anon, line 27: Since gacha eggs are determined on hatching I think this variable should be renamed because just `species` seems a little too generic to me

src/data/egg.ts
```diff
@@ -80,14 +90,22 @@ export function getEggHatchWavesMessage(hatchWaves: integer): string {
...
-    return `${i18next.t("egg:gachaTypeLegendary")} (${getPokemonSpecies(getLegendaryGachaSpeciesForTimestamp(scene, egg.timestamp)).getName()})`;
-  case GachaType.MOVE:
-    return i18next.t("egg:gachaTypeMove");
-  case GachaType.SHINY:
-    return i18next.t("egg:gachaTypeShiny");
+export function getEggSourceDescriptor(scene: BattleScene, egg: Egg): string {
+  switch (egg.source) {
+  case EggSource.GACHA:
+    if (egg.gachaType === null) {
+      break;
```
Tempo-anon, line 97: Is this needed? I don't think there should ever be a case where this gets reached

src/data/egg.ts
```diff
@@ -111,3 +129,26 @@ export function getLegendaryGachaSpeciesForTimestamp(scene: BattleScene, timesta
 
   return ret;
 }
+
+/**
+ * Obtains the egg tier of a starter species, using its starter value
+ * @param species The {@linkcode Species} of which we are trying to obtain the egg tier
+ * @returns the {@linkcode EggTier} from which this species usually hatches from, or {@linkcode EggTier.COMMON} if not a starter species
```
Tempo-anon, line 136: Which case will it return EggTier.COMMON if it's not a starter species?

src/data/egg.ts
```diff
@@ -111,3 +129,26 @@ export function getLegendaryGachaSpeciesForTimestamp(scene: BattleScene, timesta
...
+ */
+export function getSpeciesTier(species: Species): EggTier {
+  if (species === Species.PHIONE || species === Species.MANAPHY) {
+    return EggTier.COMMON;
+  }
+
+  const starterValue = speciesStarters[species] ?? 1;
```
Tempo-anon, line 143: Can you explain what this line is doing?

Tempo-anon (changes requested):
I believe one of the features of buying a species-egg with candy is that it should only be unlocked after the passive and both cost reductions were bought first

src/data/egg.ts
jaimefd: I attempted that originally, however, this causes some issues with the logic that already exists in egg-gacha-ui-handler.ts, when trying to obtain the types of gacha that exist with `Utils.getEnumValues(GachaType).forEach((gachaType, g)`. I could change that to only iterate over the specific indexes that correspond to gacha types, but I felt like it shouldn't be that specific

src/data/egg.ts
jaimefd: I was trying to protect it in the case someone passes the evolved species in this function, since `starterSpecies` doesn't have evolutions.

But after you made me think about it I found `getRootSpeciesId()` and used it to implement it in a better way

src/data/egg.ts
jaimefd: My idea was to make it so if someone passes, let's say Charizard in here instead of Charmander, we would get some sort of default value, in this case 1, so it would be COMMON (the comment you mentioned above).

I didn't know about the existance of `getRootSpeciesId`, now I do and I fixed it, the new code is
`  const rootSpecies = getPokemonSpecies(species).getRootSpeciesId();
  const starterValue = speciesStarters[rootSpecies];`

src/data/egg.ts
jaimefd: It is not, I removed it

src/data/egg.ts
jaimefd: Done

jaimefd:
Improved the logic a bit based on the feedback in the new commit https://github.com/pagefaultgames/pokerogue/pull/1700/commits/c7a2612a1a15794b7a446516757a991bae4ffa31


> I believe one of the features of buying a species-egg with candy is that it should only be unlocked after the passive and both cost reductions were bought first

I agree, I changed it to work like that in the last commit, the option now only shows up after all those 2 are fully unlocked:

https://github.com/pagefaultgames/pokerogue/assets/22176793/d4c28cbb-2f63-4982-8bdd-a92d9c7348ec

Let me know if this is what you meant @Tempo-anon

chacolah:
Tested this after rebasing and solving a merge conflict.

One major issue and one Minor issue came up:
Major: You can purchase eggs above the egg limit causing the UI to bug and have the eggs go off screen in egg list.
Minor: the ability to purchase eggs does not trigger the icon or animation for available candy options.

chacolah:
image

Tempo-anon:
Closing this since this PR is out of date and has merge conflicts. Buying eggs with candy has been implemented in https://github.com/pagefaultgames/pokerogue/pull/1833

PR #1701, opened 2024-06-01
Let Corrosion mons get badly poisoned by Toxic Orb

Fontbane:
## What are the changes?

Toxic and Flame Orb use the holder's abilities when trying to set status.
In the main series, Toxic Orb will activate on a Poison or Steel type with the ability Corrosion and inflict bad poison. Here, since there's a possibility of two abilities on the same mon, it also allows abilities to interact. For example, an Axew with Mold Breaker ability and Immunity passive would become badly poisoned by Toxic Orb. It would be as if an opponent with Mold Breaker used Toxic on it, essentially.

The way these orbs work in the official games considers the item holder as the "source" of the status condition. This is a very situational but also very funny edge case, and I'm sure other items have conditions similar to this we could look at later on. The Mold Breaker/Immunity interaction doesn't exist in the main games, but is likely how it would work if there ever were a way for one mon to have two abilities. It's consistent with the Corrosion interaction.

## What did change?

This PR simply passes the item holder as the "source" of the status condition for Toxic and Flame Orb. This would allow the proper abilities to apply, such as Corrosion on a Poison- or Steel-type, or Mold Breaker with a status immunity ability. 

### Screenshots/Videos

Bulbasaur, who does not have Corrosion, is not poisoned by Toxic Orb.
image
Salandit, who does have Corrosion, is.
image
Magnemite, who I gave Corrosion as a passive via override, is.
image
It doesn't get poisoned if it's immune due to ability such as Immunity (also override).
image
Salandit with Neutralizing Gas passive will still Poison itself...
image
But that's just because Neutralizing Gas doesn't work on Corrosion.
image
It otherwise seems to work as expected.
image
Something to note is that a mon with both Mold Breaker and Immunity (due to passives) will also poison itself from Toxic Orb.
image

## How to test the changes?

- Add `{"name":"TOXIC_ORB"}` to the starting held items overrides
- Start a game with Corrosion Salandit or Glimmet to see it work
- Start a game with another Poison- or Steel-type mon to see it not work
- Start a game with a Mold Breaker mon and set the passive ability override to Immunity to see that interaction

## Checklist
- [X] Are the changes visual?


HopsWas:
Can confirm, interaction with Mold breaker and Immunity does not work as described, Mon is not poisoned.

flx-sta:

-----

PR #1716, opened 2024-06-02
[QoL] Update pokemon.ts to allow ALL rare/epic shinies to be encountered, caught, and used with common shiny sprites if variant sprite isn't implemented

HitchEdits:
## What are the changes?
By simply removing the no variant data check, it allows for common shiny sprites to act as a placeholder for rare and epic shinies. This means that players can now encounter, catch, and use rare and epic shinies for all pokemon, including those that do not have the epic variant sprites yet.

This change is to allow players to take full advantage of the shiny luck system that has been put in place for the game.

## What did change?
Removed check for `!variantData.hasOwnProperty(this.species.speciesId)`

### Screenshots/Videos
Screenshot 2024-06-01 223548

## How to test the changes?
You can either use the `overrides.ts` to change enemy overrides so that OPP_SHINY_OVERRIDE = true, and update OPP_VARIANT_OVERRIDE as needed, or you can change line 1264 on `pokemon.ts` to be as follows:
`this.shiny = true;`
We did not use any automated tests. We tested both on new saves with variant override, and on an imported endless, and endless spliced, session with only shiny rates changed. We've tested around 50 rounds with no issues.

## Checklist
- [x] Are the changes visual?


sirzento:
Just to explain it a bit further why this change is wanted:

Currently, you can only catch/hatch shinys in variants that have an artwork submitted, and only ~41% of them have a custom artwork.
The intended rates for shinys are a common variant 60% of the time, a rare variant 30% of the time and an epic variant 10% of the time. But now if you hatch, for example, a charmander, the code checks if a variant artwork is available and if not, you will be locked to a common variant with 1 luck.

- That means the intended rates of variants is statistically wrong and is dependent on the total count of variant artworks.

- Players will get denied on chances to get a luck 2/3 variant. If the player hatches 20 shiny charmanders, all of them will have luck 1.

- Players who try to farm a specific high luck variant will never be able to catch it. I have heard about people that tried to get an luck 3 variant of a specific pokemon and grinded on endless for it with shiny charms and the map to only find out the variant artwork isn't in the game so they never even had the chance to encounter it.

- Players could have the feeling of being "scammed" by variant chances. If you get 20 shinys and all of them have 1 luck, with the 60% chance this would have been a "(0.6^20)*100=" 0.003% chance of happening. I have also read that it is a big letdown to gatcha for shiny eggs if you know that, if you pulled a shiny, that you didn't even had the chance to get something better than luck 1 in 59% of the times.

I intentionally talked about luck 1 and 2 variants here and not about rare and epic variants because the point here is not to limit the luck gameplay element to existing artwork but to implement it as it is intended for all pokemon.

Tempo-anon (changes requested):
Personally I like the idea but last time this was brought up it was shut down because each new batch of variant updates adds a lot of hype to the game

Tempo-anon:
Oh yeah, also adding for visibility/context. The other proposed solution at the time was to silently record shiny pokemon that were supposed to be rare/epic when they were caught and then award them to the player when the variants actually released. That idea was also shot down because "it would confuse players" and "cause a lot of bug reports"

sirzento:
@Tempo-anon In the feature request we talked about it with Lost in Thought and he said its fine to create a pull request if we are able to implement it and it might be considered. The "New shinys create hype" point was from the head of art but the final say does come from the division heads. [Discord Message Link](https://discord.com/channels/1125469663833370665/1245989717326041129/1246612299461038100)

HitchEdits:
>The other proposed solution at the time was to silently record shiny pokemon that were supposed to be rare/epic when they were caught and then award them to the player when the variants actually released. That idea was also shot down because "it would confuse players" and "cause a lot of bug reports"

Yeah, my opinion is that I'd rather not implement a completely silent version, because that would add even more confusion if we were to hide the shinies and lie to the players. This way it's still not *that* confusing since there still is a blue/red sparkle when they enter the field, and the rare/epic shiny icon next to the name. People get to catch them, use them and their luck, and they'll know what shiny variant sprites to look forward to when they release, rather than randomly getting them and wondering if they actually caught them or if it was a bug.

HitchEdits:
Regardless, this version *just works* (as tested by at least myself, @zabe, and @jrkong in discord) and is available now. As soon as the official testers are able to test it, it can go out immediately with minimal issues, which resolves half the conflict around this changes (that it was going to be too much work).

jrkong:
This fix also resolves issue #1745. This bug originates from the `|| !variantData.hasOwnProperty(this.species.speciesId)` portion of the if statement within `generateVariant()` as `this.species.speciesId` only contains the species ID and not the form. As this fix removes that condition completely this would resolve the bug.

I would like to also acknowledge that #1783 also resolves the issue by adding the form into the queried key however this would be redundant if this change is implemented as `generateVariant()` only gatekeeps the generation of a variant therefore it is agnostic to whether the species form is passed if the completed variant check is removed.

jrkong:
Silently tracking this is definitely something that can be considered. The real complexity of it would be, when would be the best time to check if the rarity can be restored after the pokemon's original generation. However the issue I can see arising from the silent tracking implementation is there is a reasonable justification for players to run away or kill shinies they already have in a run as catching them in shiny battles or whittling them down can be a bit of a hassle. This scenario could still result in lost variants which is not a great player experience.

Some general notes on how a silent system can be implemented, the block `!variantData.hasOwnProperty(this.species.speciesId` should be updated to account for species and form if it exists and should be used in places where variance is checked. Ideally, once this check is performed the pokemon's original variant value is restored if it is implemented.

I am cautious of this as this will increase the testing surface that needs to be tested as we are working with checks placed throughout the code rather then stepping in at the time of object generation.

jrkong:
To add some context for to this, I've been experimenting with a proof of concept to override the sparkle animation pipeline to load a `_wip` version of the sprite which implements the already familiar `N` into the sparkles to indicate that a variant isn't fully implemented. I am running into some difficulty getting the animation to load after it is loaded into the `textureManager` but this is a solution that can be considered as well.

HitchEdits:
Closing this since it's been made clear it won't be accepted. Might as well take one less PR off your hands =/

PR #1728, opened 2024-06-02
Localized game-mode and save-slot-select-ui-handler

josericardo-fo:
## What are the changes?

Localized game-mode and save-slot-select-ui-handler and added ptBR translations

Those pages weren't localized yet
## What did change?

Translations and text in game-mode and save-slot-select-ui-handler files
"Empty" locale was only used in the save-slot-select-ui-handler, so I moved it from menu.ts to save-slot-select-ui-handler.ts
### Before
image
### After
image

### Before
image
image
### After
image
image

### Before
image
### After
image

## How to test the changes?

Check the new game screen, load game screen and the text when overwriting a save slot.
## Checklist
Overlap with #1674, but he just localized the overwriteData text
- [X] Are the changes visual?


flx-sta:
There is a lot of “noise” in the PR because you added the English translation for every locale.
Did you try if it will fall-back to English by default ? Because this makes reviewing more of a chore and it is Is intact not a translation. Just noise

josericardo-fo:
What do you mean? When I localize something, I don't translate the localized files beacuse I'm not qualified to do that. It's up to the other devs to add PRs to my PR to translate or wait for it to get merged than create their own PRs.

flx-sta:
I mean something like the `src/locales/de/game-mode.ts` file. It now contains the English translations. But if English is the fallback language anyways that shouldn’t be necessary and those are not “translations” anyway.

josericardo-fo:
I agree. But, as I said, I localized it and now other people can translate it if they want to. If it is working in every language, than its fine.

flx-sta:
That’s why I’m saying: see what is being printed if you remove a language that is not translated. Technically it should show the English translation. This you don’t have to provide those inside the locale file of a different language

src/locales/de/game-mode.ts
```diff
@@ -0,0 +1,9 @@
+import { SimpleTranslationEntries } from "#app/plugins/i18n";
+
+export const gameMode: SimpleTranslationEntries = {
+  "classic": "Classic",
+  "endless": "Endless",
+  "endlessSpliced": "Endless (Spliced)",
+  "dailyRun": "Daily Run",
+  "unknown": "Unknown",
```
CodeTappert, line 8:
```suggestion
  "classic": "Klassik",
  "endless": "Endlos",
  "endlessSpliced": "Endlos (Fusion)",
  "dailyRun": "Täglicher Run",
  "unknown": "Unbekannt",
```

src/locales/de/save-slot-select-ui-handler.ts
```diff
@@ -0,0 +1,9 @@
+import { SimpleTranslationEntries } from "#app/plugins/i18n";
+
+export const saveSlotSelectUiHandler: SimpleTranslationEntries = {
+  "overwriteData": "Overwrite the data in the selected slot?",
+  "loading": "Loading...",
+  "wave": "Wave",
+  "lv": "Lv",
+  "empty": "Leer",
```
CodeTappert, line 8:
```suggestion
  "overwriteData": "Den ausgewählten Speicherstand überschreiben?",
  "loading": "Läd...",
  "wave": "Welle",
  "lv": "Lvl",
  "empty": "Leer",
```

src/locales/es/game-mode.ts
```diff
@@ -0,0 +1,9 @@
+import { SimpleTranslationEntries } from "#app/plugins/i18n";
+
+export const gameMode: SimpleTranslationEntries = {
```
GoldTra, line 3:
```suggestion
export const gameMode: SimpleTranslationEntries = {
```

GoldTra (changes requested):

src/locales/es/save-slot-select-ui-handler.ts
```diff
@@ -0,0 +1,9 @@
+import { SimpleTranslationEntries } from "#app/plugins/i18n";
+
+export const saveSlotSelectUiHandler: SimpleTranslationEntries = {
+  "overwriteData": "Overwrite the data in the selected slot?",
+  "loading": "Loading...",
+  "wave": "Wave",
+  "lv": "Lv",
```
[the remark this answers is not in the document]
CodeTappert: The override thing may be too long. Could someone check that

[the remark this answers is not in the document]
GoldTra: > The override thing may be too long. Could someone check that

Captura1
Captura3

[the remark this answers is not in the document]
CodeTappert: Okay then its all good. Just wanted to be sure since my german sentence was orignal to long

Adri1:
PR opened on your branch with French translations

Adri1:
Does this PR also convet numbers according to the language ?
image
Because I see this and for example in French, we use the comas before deciamals not dots.
It shoud be possible, as it is already the case for many numbers formatting when, for example money is displayed in some places.

josericardo-fo:
I will take a look on that

> Also, does this PR also converts dynamically numbers according to the set language, the navigator or system language? image Because I see this and for example in French, we use comas before deciamals not dots. (and just spaces where in English they use comas) It shoud be possible, as it is already the case for many number formatting when, for example, money is displayed in some places.

josericardo-fo:
@Adri1 
image
At the end, they return a dot by default after decimals. It can be changed, but the comma is kinda strange:
image
I didn't know that there was a way to localize that, looking for it becuse numbers here in Brazil are separated like this "$1.300,50" (one thousand, three hundred dollars and fifty cents) and I didn't insert that in, for example, the MoneyAchvs localization

Adri1:
For example, here is what I did in PR #1656 :
image
I always only encountered `toLocaleString("en-US")` when I happened to come across numbers. So for what you have here, I admit I dont have any more ideas.
But if what I did can give you any ideas to adapt it to your case...

josericardo-fo:
Created a PR for that @Adri1 #1762

GoldTra:
@josericardo-fo PR opened on your branch with Spanish translations

src/locales/ko/game-mode.ts
```diff
@@ -0,0 +1,10 @@
+import { SimpleTranslationEntries } from "#app/plugins/i18n";
+
+export const gameMode: SimpleTranslationEntries = {
+  "classic": "Classic",
+  "endless": "Endless",
+  "endlessSpliced": "Endless (Spliced)",
+  "dailyRun": "Daily Run",
+  "unknown": "Unknown",
+  "challenge": "Challenge",
```
returntoice, line 9:
```suggestion
  "classic": "클래식",
  "endless": "엔들리스",
  "endlessSpliced": "엔들리스(융합체)",
  "dailyRun": "데일리 런",
  "unknown": "언노운",
  "challenge": "챌린지",
```
Korean update

src/locales/ko/save-slot-select-ui-handler.ts
```diff
@@ -0,0 +1,9 @@
+import { SimpleTranslationEntries } from "#app/plugins/i18n";
+
+export const saveSlotSelectUiHandler: SimpleTranslationEntries = {
+  "overwriteData": "Overwrite the data in the selected slot?",
+  "loading": "Loading...",
+  "wave": "Wave",
```
returntoice, line 6:
```suggestion
  "overwriteData": "선택한 슬롯에 데이터를 덮어쓰시겠습니까?",
  "loading": "로딩 중...",
  "wave": "웨이브",
```

returntoice (changes requested):
Update for Korean locales

Tempo-anon (approved):
Going forward let's try having each separate localization go in their own PR since there indeed is a lot of noise to this one which makes it harder to read and review.

PR #1733, opened 2024-06-02
Balance endless tokens

Fontbane:
## What are the changes?

- Removed Sleep and Freeze tokens
- Rebalanced the others to be slightly less overbearing

Sleep and freeze tokens are widely agreed to make endless unfun. It’s honestly been a long time coming but the recent berry rebalancing made it much more urgent to finally rebalance these tokens.
These changes were made in accordance to discussions with the balance team and other devs. 
[This is where this change was initiated.](https://discord.com/channels/1125469663833370665/1229525892624154684/1246871204149858384)

## What did change?

- Sleep and freeze tokens have been erased from existence
- Reduced paralyze token proc chance from 10% to 2.5% and increased stack limit from 5 to 10
- For both burn and poison tokens, reduced proc chance from 10% to 5% and increased stack limit from 5 to 10
- Reduced endure token proc chance from 2.5% to 2%
- Reduced status heal token proc chance from 10% to 5%
- Reduced enemy heal token stack limit from 15 to 10
- Total max amount of tokens remains the same

### Screenshots/Videos

image

## How to test the changes?

Import a current endless run and try to get another token. Try to set status on the opponent and take some hits to check the chances. You could override the opponent moves to Brutal Swing so there’s 3 proc chances in one move.

## Checklist
- [ ] Are the changes visual?


bennybroseph:
> These changes were made in accordance to discussions with the balance team and other devs.

Let's link the start of these discussions now and in future PRs so that others can catch themselves up in the event that they were offline at the time.

src/system/game-data.ts
```diff
@@ -1052,6 +1054,9 @@ export class GameData {
           if (md?.className === "ExpBalanceModifier") { // Temporarily limit EXP Balance until it gets reworked
             md.stackCount = Math.min(md.stackCount, 4);
           }
```
bennybroseph, line 1056: This section can be removed now since most people should have already gotten the change saved

src/system/game-data.ts
```diff
@@ -1052,6 +1054,9 @@ export class GameData {
...
+          if (md?.className === "EnemyAttackStatusEffectChanceModifierType" && [StatusEffect.FREEZE, StatusEffect.SLEEP].includes((md as EnemyAttackStatusEffectChanceModifier)?.effect)) {
+            md.stackCount = Math.min(md.stackCount, 0);
```
bennybroseph, line 1058: Math.min here means it'll always be 0 or less than 0. Is that the intention? If so, we should just set it to 0 I think. We wouldn't want this to go negative if for some reason it is.

bennybroseph (changes requested):

src/system/game-data.ts
Fontbane: Most, but not all. Doesn’t hurt to account for older accounts.

bennybroseph (approved):

Fontbane:
The token code is so bad. It's unrecoverable. I'm refactoring this. It's so bad. It's so bad.

Tempo-anon (approved):
Gonna trust Damo and Fontbane that they have tested this

Madmadness65 (approved):
Changes look good to me. I will trust that these changes have been thoroughly tested and wont break existing saves in odd ways.

PR #1737, opened 2024-06-02
[Balance] Give Partner Pikachu its Signature Moves and change Cosplay's stats

Blitz425:
## What are the changes?
Partner Pikachu's Learnset
Cosplay Pikachu's (+Variants) Stats

Currently Partner Pikachu is able to be selected from the starter screen but with no access to egg moves it would never have gotten access to its signature moves otherwise, Raichu currently gets access to Zippy Zap and Pikapapow via custom learnset so this would be a fair change. Cosplay Pikachu has the same effect without access to egg moves or being able to evolve, so making it more in line with Starter Pikachu who has the same caveat by matching its stat total would equally be fair and enhance user experience for those wanting to use these Pikachu forms. This was discussed with fellow balance team members including Damo.

## What did change?
Added the moves Zippy Zap, Splishy Splash, and Floaty Fall to Partner Pikachu's learnset all at Level 25 and changes the Cosplay Pikachu's stats to match Partner Pikachu's.

### Screenshots/Videos
Before
ogpartner

After
partner

Partner for comparison
partner2

Partner Level Up
ezgif-5-8e99b96b9d

## How to test the changes?
Level a Partner Pikachu up to Level 25
Compare Cosplay Pikachu's stats to Partner Pikachu

## Checklist
- [x] Are the changes visual?


bennybroseph:
Can we have the title marked with something other than QoL?

PR #1744, opened 2024-06-02
[Bug] Fixed primal weather interaction.

### LINKED ISSUE #1715: [BUG] Primordial Sea and Desolate Land don't overwrite each other
**Describe the bug**
When either Heavy Rain or Harsh Sunlight are active and you switch in a Pokémon with any of Primordial Sea, Desolate Land, or Delta Stream, it does not overwrite the weather, it just keeps the current weather condition active

**To Reproduce**
1. Start a battle with Primal Groudon, so that Desolate Land activates
2. Switch to Primal Kyogre
3. Notice that Primordial Sea does not activate on switch in, and Harsh Sunlight remains active

**Expected behavior**
Primordial Sea should activate on switch in, and Heavy Rain should become active

**Screenshots / Videos**
https://github.com/pagefaultgames/pokerogue/assets/7389390/35220f71-945d-44cb-93c4-4c1de7bc9169

**Device**
PC, Google Chrome

**Additional context**
The weather does correctly change at the start of a new biome, but not when changing Pokémon at the start of a new battle in the same biome or when switching between the 2 Pokémon in the same battle

Arxalc:
## What are the changes?

Closes #1715
* Primal weather can overwrite each other
* Primal weather ends upon switching out or fainting

Interactions weren't working before.

## What did change?

If a primal weather was already in effect, it could not be overriden by another -> now it can
Primal weather would not revert upon Pokemon switching out or fainting -> now it does

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/63990624/68eb7021-59a7-48ea-802a-5692f72bb95c

## How to test the changes?

Override ability to Desolate Land, Primordial Sea, or Delta Stream.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -1870,22 +1870,35 @@ export class PreSwitchOutClearWeatherAbAttr extends PreSwitchOutAbAttr {
    * @returns {boolean} Returns true if the weather clears, otherwise false.
    */
   applyPreSwitchOut(pokemon: Pokemon, passive: boolean, args: any[]): boolean | Promise<boolean> {
-    const abilities: Abilities[] = [pokemon.getAbility().id, pokemon.getPassiveAbility().id];
     const weatherType = pokemon.scene.arena.weather.weatherType;
+    let turnOffWeather = false;
 
-    // Clear weather if ability matches weather and if no one else has the ability
-    abilities.forEach((ab) => {
-      if ((ab === Abilities.DESOLATE_LAND && weatherType === WeatherType.HARSH_SUN) ||
-      (ab === Abilities.PRIMORDIAL_SEA && weatherType === WeatherType.HEAVY_RAIN) ||
-      (ab === Abilities.DELTA_STREAM && weatherType === WeatherType.STRONG_WINDS)) {
-
-        // Besides the user, check if another has the ability on the field.
-        if (pokemon.scene.getField(true).filter(p => p !== pokemon).filter(p => p.hasAbility(ab)).length === 0) {
-          pokemon.scene.arena.trySetWeather(WeatherType.NONE, false);
-          return true;
-        }
+    // Clear weather only if user's ability matches the weather and no other pokemon has the ability.
+    switch (weatherType) {
+    case (WeatherType.HARSH_SUN):
+      if (pokemon.hasAbility(Abilities.DESOLATE_LAND)
+          && pokemon.scene.getField(true).filter(p => p !== pokemon).filter(p => p.hasAbility(Abilities.DESOLATE_LAND)).length === 0) {
+        turnOffWeather = true;
       }
-    });
+      break;
+    case (WeatherType.HEAVY_RAIN):
+      if (pokemon.hasAbility(Abilities.PRIMORDIAL_SEA)
+          && pokemon.scene.getField(true).filter(p => p !== pokemon).filter(p => p.hasAbility(Abilities.PRIMORDIAL_SEA)).length === 0) {
+        turnOffWeather = true;
+      }
+      break;
+    case (WeatherType.STRONG_WINDS):
+      if (pokemon.hasAbility(Abilities.DELTA_STREAM)
+          && pokemon.scene.getField(true).filter(p => p !== pokemon).filter(p => p.hasAbility(Abilities.DELTA_STREAM)).length === 0) {
+        turnOffWeather = true;
+      }
+      break;
+    }
+
+    if (turnOffWeather) {
+      pokemon.scene.arena.trySetWeather(WeatherType.NONE, false);
+      return true;
+    }
 
     return false;
   }
@@ -3032,22 +3045,35 @@ export class PostFaintClearWeatherAbAttr extends PostFaintAbAttr {
    * @returns {boolean} Returns true if the weather clears, otherwise false.
    */
   applyPostFaint(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: PokemonMove, hitResult: HitResult, args: any[]): boolean {
-    const abilities: Abilities[] = [pokemon.getAbility().id, pokemon.getPassiveAbility().id];
     const weatherType = pokemon.scene.arena.weather.weatherType;
+    let turnOffWeather = false;
 
-    // Clear weather if ability matches weather and if no one else has the ability
-    abilities.forEach((ab) => {
-      if ((ab === Abilities.DESOLATE_LAND && weatherType === WeatherType.HARSH_SUN) ||
-      (ab === Abilities.PRIMORDIAL_SEA && weatherType === WeatherType.HEAVY_RAIN) ||
-      (ab === Abilities.DELTA_STREAM && weatherType === WeatherType.STRONG_WINDS)) {
-
-        // Besides the user, check if another has the ability on the field.
-        if (pokemon.scene.getField(true).filter(p => p.hasAbility(ab)).length === 0) {
-          pokemon.scene.arena.trySetWeather(WeatherType.NONE, false);
-          return true;
-        }
+    // Clear weather only if user's ability matches the weather and no other pokemon has the ability.
+    switch (weatherType) {
+    case (WeatherType.HARSH_SUN):
+      if (pokemon.hasAbility(Abilities.DESOLATE_LAND)
+          && pokemon.scene.getField(true).filter(p => p.hasAbility(Abilities.DESOLATE_LAND)).length === 0) {
+        turnOffWeather = true;
       }
-    });
+      break;
+    case (WeatherType.HEAVY_RAIN):
+      if (pokemon.hasAbility(Abilities.PRIMORDIAL_SEA)
+          && pokemon.scene.getField(true).filter(p => p.hasAbility(Abilities.PRIMORDIAL_SEA)).length === 0) {
+        turnOffWeather = true;
+      }
+      break;
+    case (WeatherType.STRONG_WINDS):
+      if (pokemon.hasAbility(Abilities.DELTA_STREAM)
+          && pokemon.scene.getField(true).filter(p => p.hasAbility(Abilities.DELTA_STREAM)).length === 0) {
+        turnOffWeather = true;
+      }
+      break;
+    }
+
+    if (turnOffWeather) {
+      pokemon.scene.arena.trySetWeather(WeatherType.NONE, false);
+      return true;
+    }
 
     return false;
   }
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -1728,7 +1728,9 @@ export class PostSummonWeatherChangeAbAttr extends PostSummonAbAttr {
   }
 
   applyPostSummon(pokemon: Pokemon, passive: boolean, args: any[]): boolean {
-    if (!pokemon.scene.arena.weather?.isImmutable()) {
+    if ((this.weatherType as WeatherType === WeatherType.HEAVY_RAIN ||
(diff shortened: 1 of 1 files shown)


src/data/ability.ts
```diff
@@ -1728,7 +1728,9 @@ export class PostSummonWeatherChangeAbAttr extends PostSummonAbAttr {
   }
 
   applyPostSummon(pokemon: Pokemon, passive: boolean, args: any[]): boolean {
-    if (!pokemon.scene.arena.weather?.isImmutable()) {
+    if ((this.weatherType as WeatherType === WeatherType.HEAVY_RAIN || 
+      this.weatherType as WeatherType === WeatherType.HARSH_SUN || 
+      this.weatherType === WeatherType.STRONG_WINDS) || !pokemon.scene.arena.weather?.isImmutable()) {
```
Tempo-anon, line 1733: Wonder if it would be worth having a util function in weather.ts `isPrimalWeather`

src/data/ability.ts
```diff
@@ -1856,6 +1858,16 @@ export class PreSwitchOutResetStatusAbAttr extends PreSwitchOutAbAttr {
   }
 }
 
+/**
+ * Clears weather upon Primal Groudon/Kyogre/Rayquaza switch out
```
Tempo-anon, line 1862: nit: this ability attribute is not exclusive to those Pokemon, but rather on the abilites heavy rain, harsh sun, and strong winds

src/data/ability.ts
```diff
@@ -1856,6 +1858,16 @@ export class PreSwitchOutResetStatusAbAttr extends PreSwitchOutAbAttr {
...
+ */
+export class PreSwitchOutClearWeatherAttr extends PreSwitchOutAbAttr {
```
Tempo-anon, line 1864: nit: `PreSwitchOutClearWeatherAbAttr`

src/data/ability.ts
```diff
@@ -1856,6 +1858,16 @@ export class PreSwitchOutResetStatusAbAttr extends PreSwitchOutAbAttr {
...
+  applyPreSwitchOut(pokemon: Pokemon, passive: boolean, args: any[]): boolean | Promise<boolean> {
+    pokemon.scene.arena.trySetWeather(WeatherType.NONE, false);
```
Tempo-anon, line 1866: Won't this cause a bug if there are two Primal Groudons out and one of them switches out? The harsh sun should still be in effect

src/data/ability.ts
```diff
@@ -2982,6 +2994,20 @@ export class PostFaintAbAttr extends AbAttr {
...
+/**
+ * Clears weather upon Primal Groudon/Kyogre/Rayquaza fainting
+ */
+export class PostFaintClearWeatherAbAttr extends PostFaintAbAttr {
+  constructor() {
+    super ();
+  }
+
+  applyPostFaint(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: PokemonMove, hitResult: HitResult, args: any[]): boolean {
+    pokemon.scene.arena.trySetWeather(WeatherType.NONE, false);
```
Tempo-anon, line 3006: Same issue as above

Tempo-anon (changes requested):
There's some edge cases to consider here

src/data/ability.ts
Arxalc: You're right. Will try and fix.
Tested interactions on Showdown and found it works like this:
* Groudon vs. Kyogre. When Kyogre switches in, Desolate Land is overriden by Primordial Sea. If Kyogre dies, weather clears.
* Groudon 1 vs. Groudon 2. When Groudon 2 switches in, nothing special happens. If Groudon 1 dies, weather persists because Groudon 2 has the ability and is still on the field.

Tempo-anon: What happens if Kyogre switches out after overwriting Desolate Land with Primordial Sea? If the weather still clears up it just means you need to add a check so on faint/switchout it checks the field to make sure no other Pokemon has that ability which means this ability attribute also needs to know what ability it's checking

Arxalc:
I added checking abilities for both switch out/faint so now all cases should be covered.

Fixed 1st comment case: Harsh sun persists after Desolate Land user faints and another Desolate Land ability remains
https://github.com/pagefaultgames/pokerogue/assets/63990624/277bb5d6-2ff2-4261-8f8c-dc666f80f546

2nd comment case: According to Showdown after overwriting a primal weather, weather should clear if Kyogre switches out/faints. This is intentional. 
https://github.com/pagefaultgames/pokerogue/assets/63990624/b6859ec9-34f5-46ea-bb22-415691a3468e

src/data/ability.ts
```diff
@@ -1728,7 +1728,9 @@ export class PostSummonWeatherChangeAbAttr extends PostSummonAbAttr {
   }
 
   applyPostSummon(pokemon: Pokemon, passive: boolean, args: any[]): boolean {
-    if (!pokemon.scene.arena.weather?.isImmutable()) {
+    if ((this.weatherType as WeatherType === WeatherType.HEAVY_RAIN ||
+      this.weatherType as WeatherType === WeatherType.HARSH_SUN ||
+      this.weatherType === WeatherType.STRONG_WINDS) || !pokemon.scene.arena.weather?.isImmutable()) {
```
Tempo-anon, line 1733: nit: how come the strong winds check does not `as`?

src/data/ability.ts
```diff
@@ -1856,6 +1858,29 @@ export class PreSwitchOutResetStatusAbAttr extends PreSwitchOutAbAttr {
   }
 }
 
+/**
+ * Clears Desolate Land/Primordial Sea/Delta Stream upon the Pokemon switching out.
+ */
+export class PreSwitchOutClearWeatherAbAttr extends PreSwitchOutAbAttr {
+  applyPreSwitchOut(pokemon: Pokemon, passive: boolean, args: any[]): boolean | Promise<boolean> {
```
Tempo-anon, line 1865: Please add tsdoc to the `applyPreSwitchOut` function, especially since you're using `pokemon` in it

src/data/ability.ts
```diff
@@ -1856,6 +1858,29 @@ export class PreSwitchOutResetStatusAbAttr extends PreSwitchOutAbAttr {
...
+    const currentAbility = pokemon.getAbility().id;
```
Tempo-anon, line 1866: I believe `getAbility` does not take into account passives which might be worth considering

src/data/ability.ts
```diff
@@ -1856,6 +1858,29 @@ export class PreSwitchOutResetStatusAbAttr extends PreSwitchOutAbAttr {
...
+    const weatherType = pokemon.scene.arena.weather.weatherType;
+
+    // Clear weather if ability matches weather and if no one else has the ability
+    if ((currentAbility === Abilities.DESOLATE_LAND && weatherType === WeatherType.HARSH_SUN) ||
+    (currentAbility === Abilities.PRIMORDIAL_SEA && weatherType === WeatherType.HEAVY_RAIN) ||
+    (currentAbility === Abilities.DESOLATE_LAND && weatherType === WeatherType.HARSH_SUN)) {
```
Tempo-anon, line 1872: There's two desolate land/harsh sun checks here but no strong winds check

src/data/ability.ts
```diff
@@ -1856,6 +1858,29 @@ export class PreSwitchOutResetStatusAbAttr extends PreSwitchOutAbAttr {
...
+
+      // Besides the user, check if another has the ability on the field.
+      if (pokemon.scene.getField(true).filter(p => p !== pokemon).filter(p => p.hasAbility(currentAbility)).length === 0) {
```
Tempo-anon, line 1875: I wonder if there's a more elegant way of doing this

src/data/ability.ts
```diff
@@ -2982,6 +3007,29 @@ export class PostFaintAbAttr extends AbAttr {
...
+ */
+export class PostFaintClearWeatherAbAttr extends PostFaintAbAttr {
+  applyPostFaint(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: PokemonMove, hitResult: HitResult, args: any[]): boolean {
+    const currentAbility = pokemon.getAbility().id;
+    const weatherType = pokemon.scene.arena.weather.weatherType;
+
+    // Clear weather if ability matches weather and if no one else has the ability
+    if ((currentAbility === Abilities.DESOLATE_LAND && weatherType === WeatherType.HARSH_SUN) ||
+    (currentAbility === Abilities.PRIMORDIAL_SEA && weatherType === WeatherType.HEAVY_RAIN) ||
+    (currentAbility === Abilities.DESOLATE_LAND && weatherType === WeatherType.HARSH_SUN)) {
```
Tempo-anon, line 3021: two desolate land checks here, which I guess makes sense since I assume this is copy/pasted. All my other comments about the other class also apply here

Tempo-anon (changes requested):
See some errors but looking good!

Arxalc:
1. Removed the `as` part. (tldr: I only know Java. Tried `==`. Had errors. Read you needed `as`. Turns out I just needed to change to `===` instead.)
2. Added passive abilities into consideration. (Balance-wise, unlikely to ever happen
3. I completely missed strong winds 🤡
4. Lmk if there's a better way to do this. I don't think it's worth adding util functions for something that'll only happen twice ever.

src/data/ability.ts
```diff
@@ -2982,6 +3017,42 @@ export class PostFaintAbAttr extends AbAttr {
...
+  /**
+   * @param pokemon The {@linkcode Pokemon} with the ability
+   * @param passive N/A
+   * @param attacker N/A
+   * @param move N/A
+   * @param hitResult N/A
+   * @param args N/A
+   * @returns {boolean} Returns true if the weather clears, otherwise false.
+   */
+  applyPostFaint(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: PokemonMove, hitResult: HitResult, args: any[]): boolean {
```
Tempo-anon, line 3034: Since we can make the assumption that there's only ever one weather at a time, would something like this make sense?
```ts
    let turnOffWeather = false;
    switch (weatherType) {
      case (WeatherType.HARSH_SUN):
        if (pokemon.hasAbility(Abilities.DESOLATE_LAND) 
          && pokemon.scene.getField(true).filter(p => p.hasAbility(Abilities.DESOLATE_LAND).length === 0)) {
            turnOffWeather = true;
        }
        break;
      case (WeatherType.HEAVY_RAIN):
        if (pokemon.hasAbility(Abilities.PRIMORDIAL_SEA) 
          && pokemon.scene.getField(true).filter(p => p.hasAbility(Abilities.PRIMORDIAL_SEA).length === 0)) {
            turnOffWeather = true;
        }
        break;
      case (WeatherType.STRONG_WINDS):
        if (pokemon.hasAbility(Abilities.DELTA_STREAM) 
          && pokemon.scene.getField(true).filter(p => p.hasAbility(Abilities.DELTA_STREAM).length === 0)) {
            turnOffWeather = true;
        }
        break;
      default:
    }
    if (turnOffWeather) {
      pokemon.scene.arena.trySetWeather(WeatherType.NONE, false);
      return true;
    }
    return false;
  ```

Arxalc: Is there a benefit of this code over mine? It seems to just be changing the order of checking conditions in more lines.

Arxalc: Replaced 👍

Tempo-anon (approved):

PR #1747, opened 2024-06-03
[QoL] Modifiers type inference

OrangeRed:
## What are the changes?
Add type inference to all modifier classes

`as Type` assertions are dangerous and this make the user experience better by adding intellisense

## What did change?
1. Make applyModifier method accept generic modifier classes and get typeof args from inference
2. Refactor all places modifiers are used to follow the new type inference param scheme
3. Create a new utility type that maps to `new(...args: any[]) => SomeClass` 
4. Refactor all places that used `new(...args: any[]) => SomeClass` to now use `Utils.ClassType<T>`

## How to test the changes?
Try it out with intellisense.
Run the game in localhost to see the changes aren't breaking

## Checklist
  - The `Utils.ClassType<T>` alias can be its own separate PR 

~~- [ ] Are the changes visual?~~
~~- [ ] Have I provided screenshots/videos of the changes?~~


src/battle-scene.ts
```diff
@@ -1837,8 +1838,8 @@ export default class BattleScene extends SceneBase {
           return Promise.allSettled([this.party.map(p => p.updateInfo(instant)), ...modifierPromises]).then(() => resolve(success));
         } else {
           const args = [ this ];
-          if (modifier.shouldApply(args)) {
-            const result = modifier.apply(args);
+          if (modifier.shouldApply(...args)) {
```
flx-sta, line 1841: Wouldn’t the original here have the Pokémon included ? Is that relevant ?

OrangeRed: What do you mean? 
args gets defined on line 1840  `const args = [ this ];` right before being passed into `modifier.shouldApply` and `modifier.apply`

src/modifier/modifier.ts
```diff
@@ -451,10 +442,12 @@ export abstract class PokemonHeldItemModifier extends PersistentModifier {
     return [ this.pokemonId ];
   }
 
-  shouldApply(args: any[]): boolean {
-    return super.shouldApply(args) && args.length && args[0] instanceof Pokemon && (this.pokemonId === -1 || (args[0] as Pokemon).id === this.pokemonId);
+  shouldApply(...args: Parameters<this["apply"]>): boolean {
+    return super.shouldApply(...args) && (this.pokemonId === -1 || args[0].id === this.pokemonId);
```
flx-sta, line 446: i think you should still check if args[0] is actually defined (to match the original code)

src/utils.ts
```diff
@@ -318,6 +318,8 @@ export function apiPost(path: string, data?: any, contentType: string = "applica
   });
 }
 
+export type ClassType<T> = new(...args: any[]) => T;
```
flx-sta, line 321: Im not 100% happy with the name of the type but I can't come up with better ones.
Maybe `ClassContructorType`?

src/battle-scene.ts
flx-sta: Nvm, I slipped in the rows.
I was looking at the one initializting with `[ pokemon ]` but thats in a different if-case

src/modifier/modifier.ts
```diff
@@ -110,11 +110,11 @@ export abstract class Modifier {
     return false;
   }
 
-  shouldApply(_args: any[]): boolean {
+  shouldApply(...args: Parameters<this["apply"]>): boolean {
     return true;
   }
 
-  abstract apply(args: any[]): boolean | Promise<boolean>;
+  abstract apply(...args: any[]): boolean | Promise<boolean>;
```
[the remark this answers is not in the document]
flx-sta: actually might even go with

```suggestion
  abstract apply(...args: unknown[]): boolean | Promise<boolean>;
```

src/modifier/modifier.ts
OrangeRed: I agree that it might be worth checking but typescript won't let you deliberately pass undefined as args[0].
The only way for it get through is if a variable that is undefined is asserted to be a Pokemon. If that is worth guarding against I don't know.

src/utils.ts
OrangeRed: Yeah I just wrote something to write something without much thought. I think something like `OfTypeClass` or `ClassDef` is more descriptive

src/modifier/modifier.ts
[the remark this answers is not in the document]
OrangeRed: > Would this also be viable?

This is viable but I feel like that adds a little too much typescript magic.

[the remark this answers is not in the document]
OrangeRed: > actually might even go with

unknown is arguably better and we should switch over to it

src/modifier/modifier.ts
flx-sta: Unfortunately we have `strict = false` and `strictNullChecks = false`.
But I still don't see when this would happen unless deliberately...
I trust the devs so I'd say: drop it then

src/utils.ts
flx-sta: the good old "what variable name" discussions :D
I think i like `OfTypeClass`

src/modifier/modifier.ts
OrangeRed: I will add change it to `args[0]?.id` just in case

flx-sta (approved):
lgtm 🚀

flx-sta (approved):

OrangeRed:
In this gist I saw they called the same construct we're as Constructor so I switched to that name
```ts
// create constructor function type alias
type Ctor<U> = new ( ...args: any[] ) => U;

// factory function
function simpleFactory<T>(
    ctor: Ctor<T>,
    ...args: any[]
): T {
    return new ctor( ...args );
}
```

src/modifier/modifier.ts
```diff
@@ -138,11 +138,11 @@ export abstract class Modifier {
     return false;
   }
 
-  shouldApply(_args: any[]): boolean {
+  shouldApply(...args: Parameters<this["apply"]>): boolean {
     return true;
   }
 
-  abstract apply(args: any[]): boolean | Promise<boolean>;
+  abstract apply(...args: unknown[]): boolean | Promise<boolean>;
```
flx-sta, line 145: Something that just came to my mind. Why didn't we consider this too?
```suggestion
  abstract apply<T extends Array = unknown[]>y(...args: T): boolean | Promise<boolean>;
```
?

OrangeRed: This prevents spreading. By forcing `...args: T` the parameters of the apply function of all children must be an array.

flx-sta:
@xsn34kzx can you take a look at this?
This seems to have been quite scrambled by the stats refactor

flx-sta:
Mirrored in #4294

PR #1755, opened 2024-06-03
[Ability] Implement Ice Face

torranx:
## What are the changes?
- implement ice face
- **making this bold because it's easy to miss on files changed: renamed audio file to match Eiscue's no-ice form key**

- implement ice face

## What did change?
- new ability attr `IceFaceImmunityAbAttr`
- new BattlerTag `IceFaceTag`
- renamed Eiscue's no-ice audio file

### Screenshots/Videos
- Ice face is hit with physical move

https://github.com/pagefaultgames/pokerogue/assets/68144167/3623b71b-7e57-48b4-8c11-5d7b58c659c4

- Ice face is hit with special move

https://github.com/pagefaultgames/pokerogue/assets/68144167/97b5a7de-5088-4302-a0ad-98c3eb577b79

- No ice is hit with physical

https://github.com/pagefaultgames/pokerogue/assets/68144167/958002b5-5c84-459d-a194-71f8d9a7f634

- No ice used hail- reverts to ice face. if noice is on the field with already active hail/snow, does not revert to ice face

https://github.com/pagefaultgames/pokerogue/assets/68144167/096fc235-9ad0-4b49-ad3a-c3793ccc62e5

- No ice summoned under Hail weather

https://github.com/pagefaultgames/pokerogue/assets/68144167/678cd96f-a5f9-4ed4-b9c5-610da8d15510

- No ice will revert to ice face on _pokerogue's context of battle end_, regardless of weather

https://github.com/pagefaultgames/pokerogue/assets/68144167/e19309f8-7daf-4caf-9d38-2c09b4ccf372

https://github.com/pagefaultgames/pokerogue/assets/68144167/9497e0c2-6135-4c86-92dc-a458d6304f8d

## How to test the changes?
[ability description](https://bulbapedia.bulbagarden.net/wiki/Ice_Face_(Ability))
- in `overrides.ts`:
```
export const STARTING_BIOME_OVERRIDE: Biome = Biome.SNOWY_FOREST;
export const STARTER_SPECIES_OVERRIDE: Species | integer = Species.EISCUE; 
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.HAIL];
// set physical and special moves to check for the form change and immunity to physical
export const OPP_MOVESET_OVERRIDE: Array<Moves> = [Moves.DRAGON_PULSE, Moves.DRAGON_PULSE, Moves.TACKLE, Moves.TACKLE];
```

## Checklist
- ~[] Are the changes visual?~


src/battle-scene.ts
```diff
@@ -1010,6 +1011,12 @@ export default class BattleScene extends SceneBase {
       if (resetArenaState) {
         this.arena.removeAllTags();
         playerField.forEach((_, p) => this.unshiftPhase(new ReturnPhase(this, p)));
+
+        for (const pokemon of this.getParty()) {
+          if (pokemon.hasAbility(Abilities.ICE_FACE)) {
```
Tempo-anon, line 1016: Should probably also add a hardcoded check for Eiscue here for now, otherwise this will cause a crash if a fused Pokemon has ice face

src/data/ability.ts
```diff
@@ -3339,6 +3339,52 @@ export class PostSummonStatChangeOnArenaAbAttr extends PostSummonStatChangeAbAtt
...
+
+  /**
+   * Gets the message triggered when the Pokémon avoids damage using the Ice Face ability.
+   * @param {Pokemon} pokemon - The Pokémon with the Ice Face ability.
+   * @param {string} abilityName - The name of the ability.
+   * @param {...any} args - Additional arguments.
+   * @returns {string} - The trigger message.
+   */
+  getTriggerMessage(pokemon: Pokemon, abilityName: string, ...args: any[]): string {
+    return `${pokemon.name} avoided damage\nwith ${abilityName}!`;
```
Tempo-anon, line 3384: abilityTrigger should have localization files

Tempo-anon (commented):
Might want to have some hardcoded checks for if the species is just Eiscue for now. Functionality looks good to me but you might want to check in with Walker since he did zero to hero and also Xavion to see if maybe my idea of using battler-tags for this ability implementation is too hacky

src/battle-scene.ts
torranx: resolved in https://github.com/pagefaultgames/pokerogue/pull/1755/files#diff-a777336298ea2da66969083089194dd8eafb5f7b8a089e19999c0f1bbf93648aR1375

src/data/ability.ts
```diff
@@ -3339,6 +3339,52 @@ export class PostSummonStatChangeOnArenaAbAttr extends PostSummonStatChangeAbAtt
...
 }
 
+/**
+ * Applies immunity to physical moves.
+ * This is used in Ice Face ability.
+ */
+export class IceFaceMoveImmunityAbAttr extends MoveImmunityAbAttr {
+  constructor(immuneCondition: PreDefendAbAttrCondition) {
+    super(immuneCondition);
+  }
```
OrangeRed, line 3349: If this extends `MoveImmunityAbAttr` do you need to do this?
Won't instantiating this class without this constructor use the parent one instead?

src/data/ability.ts
```diff
@@ -4325,8 +4371,14 @@ export function initAbilities() {
...
-      .ignorable()
-      .unimplemented(),
+      // Add BattlerTagType.ICE_FACE if the pokemon is in ice face form
+      .conditionalAttr(pokemon => pokemon.formIndex === 0, PostSummonAddBattlerTagAbAttr, BattlerTagType.ICE_FACE, 0, false)
+      // When summoned with active HAIL or SNOW, add BattlerTagType.ICE_FACE
+      .conditionalAttr(getWeatherCondition(WeatherType.HAIL, WeatherType.SNOW), PostSummonAddBattlerTagAbAttr, BattlerTagType.ICE_FACE, 0)
+      // When weather changes to HAIL or SNOW while pokemon is fielded, add BattlerTagType.ICE_FACE
+      .attr(PostWeatherChangeAddBattlerTagAttr, BattlerTagType.ICE_FACE, 0, WeatherType.HAIL, WeatherType.SNOW)
+      .attr(IceFaceMoveImmunityAbAttr, (target, user, move) => move.getMove().category === MoveCategory.PHYSICAL && !!target.getTag(BattlerTagType.ICE_FACE))
+      .ignorable(),
```
OrangeRed, line 4381: > This form change persists even when Eiscue is switched out, but is undone at the end of the battle. 

Does Eiscue switch to Ice Face mode at the start of a trainer battle / biome transition with your current implementation? If not look at how other form change attrs work

> Gastro Acid, Worry Seed, and Simple Beam fail if the target has Ice Face. Role Play, Skill Swap, Entrainment, and Doodle fail if either Pokémon has Ice Face. Ice Face cannot be copied by Trace. Ice Face cannot be suppressed by Neutralizing Gas, traded by Wandering Spirit, or replaced by Lingering Aroma. 

I don't see any unswappable / untrasferable attrs here, do the above fail as well against Ice Face?

src/data/ability.ts
torranx: Screenshot 2024-06-06 at 4 43 04 AM

I'm not very well-versed with OOP yet but if I remove the constructor, won't the above `super(true)` not called anymore?

src/data/ability.ts
torranx: > Does Eiscue switch to Ice Face mode at the start of a trainer battle / biome transition with your current implementation? If not look at how other form change attrs work

Yes it does. Its implemented [here](https://github.com/pagefaultgames/pokerogue/pull/1755/files#diff-4b1a5f9160d1d8aecce0db85774f20e3c254f97c27cc0d0473911d5cbd565396R1025-R1029) 
Video can also be found on the PR description under bullet "No ice will revert to ice face on pokerogue's context of battle end, regardless of weather"

> I don't see any unswappable / untrasferable attrs here, do the above fail as well against Ice Face?

You need to expand it up top to see the other attributes.

Screenshot 2024-06-06 at 4 47 39 AM

src/data/ability.ts
OrangeRed: if you don't supply a constructor to a class it will use the constructor of the first parent constructor it can find up the hierarchy tree.
So for example in `a extends b extends c extends d` if neither a nor b nor c have constructors they will all invoke d's constructor

src/data/ability.ts
OrangeRed: I see I missed those

src/data/ability.ts
torranx: got it, thanks for clarifying

PR #1759, opened 2024-06-03
[Localization] Use proper postpositional particle for Korean

yurical:
## What are the changes?
This PR introduces ability to properly select and print the correct form of the Korean postpositional particle.

In Korean, postpositional particles vary depending on whether the preceding syllable ends in a consonant or a vowel. Currently there is no ability to differentiate between the two types of particles, so both forms are being used at the same time. This is a significant UI glitch for Korean users.

## What did change?
- Added i18next processor and i18next-korean-postposition-processor
- Updated the current Korean l10n to utilize added processor
- 번역문에는 받침이 없는 쪽에 사용하는 조사((을-***를***, 이-***가***, 은-***는***, 으로-***로***, 과-***와***)를 썼습니다.

### Screenshots/Videos
Before:
before

After:
after

'을' and '를' are postpositional particles.

## How to test the changes?
Set the language to Korean, start any session, and check if the postpositional particles are shown properly.

## Checklist
- [x] Are the changes visual?


Tempo-anon:
@MrWaterT and @returntoice need your opinions since you've done some Korean localization

returntoice:
> @MrWaterT and @returntoice need your opinions since you've done some Korean localization

I don't know much about coding, but based on @yurical's explanation, this seems like a very good localization update. Due to the characteristics of Korean postpositional particles, two particles are used at once with brackets. but if this patch is applied, legibility will be greatly improved. I believe there should be no problem with merging it.

returntoice (approved):

Tempo-anon (approved):
Looks good to me but this does add a new `"i18next-korean-postposition-processor": "^1.0.0",` dependency

yurical:
> Looks good to me but this does add a new `"i18next-korean-postposition-processor": "^1.0.0",` dependency

Yes, I couldn't find a neater way to do this since we're using i18next though.

MrWaterT:
> > @MrWaterT and @returntoice need your opinions since you've done some Korean localization
> 
> I don't know much about coding, but based on @yurical's explanation, this seems like a very good localization update. Due to the characteristics of Korean postpositional particles, two particles are currently used at once with brackets. but if this patch is applied, legibility will be greatly improved. I believe there should be no problem with merging it.

That's right.
If Eng locale has this problem, we can see the sentences like `{{Itemname}} is(are) added to inventory!` whenever.
99+% of Koreans should be happy with merging this PR.


> > Looks good to me but this does add a new `"i18next-korean-postposition-processor": "^1.0.0",` dependency
> 
> Yes, I couldn't find a neater way to do this since we're using i18next though.

I also thought plugin dependency and global usage in i18n would have lengthened the discussion time,
but that is *the only reason* I didn't implement it in the first Korean locale update.

Tempo-anon (approved):

PR #1769, opened 2024-06-03
[Balance] Give Delibird a new Hidden Ability

Blitz425:
## What are the changes?
Delibird's Hidden Ability

Currently Delibird has the ability Vital Spirit but the hidden ability Insomnia. Which both accomplish the exact same thing, considering Delibird received a completely custom moveset and certain Pokemon's abilities have changed in the past I think this would be a really neat and fair change to a Pokemon that requires all the help it needs and allow fans of the Pokemon to have a better experience with it without making it broken.

## What did change?
Delibird's Hidden Ability Insomnia -> Refrigerate

### Screenshots/Videos
Before
image

After
image

## How to test the changes?
Check Delibird's Hidden Ability

## Checklist
- [x] Are the changes visual?


RNGezeus:
> There is a small difference between Insomnia and Vital Spirit in the main serie.
> 
> `If a Pokémon with Vital Spirit is the leading party Pokémon, wild Pokémon are 50% more likely to be higher level.`
> 
> source: https://pokemondb.net/ability/vital-spirit
> 
> But I don't know if this is actually implemented in pokerogue. 🤔
> 
> Anyway I agree with the proposed change. 👍

The party lead function has seemingly been removed as of gen 8
https://bulbapedia.bulbagarden.net/wiki/Vital_Spirit_(Ability)

If we aim to make Delibird more usable I'm not sure how helpful it is giving it a offensive ability looking at its lackluster stats (I hadn't realized Delibird was equal to a first-form starter in stats until now), maybe something more supportive?

Tempo-anon:
Not a fan of this change tbh

PR #1772, opened 2024-06-03
[QoL] Added an indicator when money changes.

Mehauk:
## What are the changes?
There is now a short animation and color change occurs when money is added to or removed.

When the player uses money or gains money, an indicator to show that the money was changed would be nice. Mostly due to the golden punch not feeling like its doing anything unless you are looking at the top right corner of the screen.

#### New Animation
[recording-2024-06-06-10-13-57.webm](https://github.com/pagefaultgames/pokerogue/assets/65061866/2b09654b-4b6c-4a6c-886e-0f0da32889df)

#### Old Animation
[recording-2024-06-03-23-12-04.webm](https://github.com/pagefaultgames/pokerogue/assets/65061866/b786d502-32df-4a29-8804-3b51f9c1b43e)

## How to test the changes?
steps to test:
1. start/load a session
2. buy an item to test the animation for losing money
3. gain some money (trainer battler, nugget, golden punch) to trigger animation for gaining money

## Checklist
- [x] Are the changes visual?


Mehauk:
I also created an issue on the issue board. I am not sure how to link the two. If this change is merged should I close the issue as well?

Tempo-anon (changes requested):
Can you please fill out the PR template? And also provide a video of this in action?

ben-lear:
Can you add a video and screenshots of the change?

src/phases.ts
```diff
@@ -4900,6 +4900,7 @@ export class SelectModifierPhase extends BattlePhase {
             this.scene.ui.setMode(Mode.MESSAGE).then(() => super.end());
             this.scene.money -= rerollCost;
             this.scene.updateMoneyText();
+            this.scene.animateMoneyChanged(false);
```
ben-lear, line 4903: Why not put this inside the updateMoneyText() function so that the two are more tightly coupled? Having to keep track of calling both functions might become cumbersome.

Mehauk: That is what I did initially but the `updateMoneyText` was being called when starting/loading the game, as well as when the money was changed. Are we okay with showing the animation whenever someone starts playing? Another option was to add a couple of extra parameters, but whenever we would need to use it, we would have to provide the first optional parameter as well. Something like `updateMoneyText(true, true, false)`, which feels unreadable. Another option that comes to mind RN, is to use an object with the appropriate fields and only supply the fields that are required at the time;

```js
updateMoneyText({
    forceVisible: boolean,
    shouldAnimate: boolean,
    positiveChange: boolean,
}): void {}
```

bennybroseph (changes requested):
I like the animation alright. I think it should be vertically centered, slightly faster, and slightly larger though. 

Can we upload another video with those changes to see how they would potentially look?

bennybroseph:
Got some conflicts. Also, the code looks correct to center the text scaling, but the video is definitely not centered. Is that an old video or is it actually not working?

Mehauk:
> Got some conflicts. Also, the code looks correct to center the text scaling, but the video is definitely not centered. Is that an old video or is it actually not working?

I have handled the merge conflict. I had dev tools open in the new video so I replaced it as well.

bennybroseph (approved):

PR #1774, opened 2024-06-04
[Move] Implement Fusion Flare and Fusion Bolt

xsn34kzx:
## What are the changes?

Implemented the power boosting effects to the following moves: [``Fusion Flare``](https://bulbapedia.bulbagarden.net/wiki/Fusion_Flare_(move)) and [``Fusion Bolt``](https://bulbapedia.bulbagarden.net/wiki/Fusion_Bolt_(move)).

The two moves, prior to this PR, were left partially implemented and did not include their secondary effect, doubling their base power if the opposite move was used immediately before the other without any other moves in between. This PR aims to achieve this effect. Note that, on the moves' Bulbapedia pages, the phrase "except perhaps for failed moves" is taken to mean that all failed moves, if any have occurred that turn, are ignored when checking if the opposing move has happened "immediately" before.

## What did change?

For some preliminary context, move order within a turn is determined in `TurnStartPhase`. However, move order is not stored anywhere (as far as I could tell); individual move histories are stored in `Pokemon` with appropriate getters like `getMoveHistory()` and `getLastXMoves()`, but the context for when those moves happen relative to other `Pokemon` is not stored. In a single battle, it's easy to check if one acted before the other, but this becomes an issue in double battles and trying to determine the moves that have occurred in the turn so far in addition to the order they occurred. To work around this, a new `order` field was added to `PokemonTurnData` and set, starting from a `0` index, during `TurnStartPhase` when `MovePhase`s are being provided to each `Pokemon` in the order they occur.

To achieve the effect of both moves, `LastMoveDoublePowerAttr`, extending `VariablePowerAttr`, was written such that its `apply` function would get the `Pokemon` that have already acted during the turn, sort them in descending order by the new `order` field, and check in order if the last move they used had failed (moving on to the next acting `Pokemon`), was a successful use of the specified move (doubling the power of the move), or was not (doing nothing).

#### Move Behavior - Before

flareToBoltBefore

Note that the value to the right of the move name is the base power and that it doesn't change even if the opposing move is used.

#### Move Behavior - After

flareToBolt

boltToFlare

In the two scenarios where either opposing move is used one after another, the second move used always has its base power doubled.

flareToFlare

In the case that the same move is used by two different `Pokemon`, the second move does not have its base power doubled.

## How to test the changes?

For general testing of the moves functionality, the `src/overrides.ts` file was modified such that the `MOVESET_OVERRIDE` array included both `Moves.FUSION_FLARE` and `Moves.FUSION_BOLT`. `DOUBLE_BATTLE_OVERRIDE` was also set to test the moves in a double battle context since that was the most rigorous scenario for determining whether the moves' effect worked. 

~~Note that, due to the limits of the given overrides and the sparse scenarios for moves failing while also having opponents interleave the moves in a given turn, the case in which, for example, `Fusion Bolt` was used, moves failed in between, and `Fusion Flare` was used was not directly tested.~~ 

For further stress testing, unit tests are included to check both the individual properties of the moves alone and check the interactions between the usage of both in mock battles.

## Comments

A couple things worth noting:
- The Bulbapedia description of these moves mention an interaction with `Instruct`, where if one of the moves is used that turn and the target then uses the opposing move right after, the power doubling effect comes into play. So far, this isn't addressed since `Instruct` is not implemented yet; however, I have thought about how to account for this if needed now for future-proofing the PR.
- The current implementation, by checking who acted so far in that turn, guarantees that the last move used will be one that happened that turn; that is, it should be impossible for any empty move histories to appear to necessitate move history length checking. However, specifically related to expanding the implementation for `Instruct`, it does not take into account multiple moves from the same user in the same turn. This is fine for `Multi Lens`-modified moves since the same move can only be used multiple times, but this becomes an issue if different moves can happen from the same user in the same turn, as in the case with `Instruct`. Noting this down for the sake of discussion and bookkeeping later.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/test/moves/fusion_flare_bolt.test.ts
+++ b/src/test/moves/fusion_flare_bolt.test.ts
@@ -1,0 +1,323 @@
+import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+import Phaser from "phaser";
+import GameManager from "#app/test/utils/gameManager";
+import overrides from "#app/overrides";
+import { MoveEffectPhase, MovePhase, MoveEndPhase, TurnStartPhase, DamagePhase } from "#app/phases";
+import { getMovePosition } from "#app/test/utils/gameManagerUtils";
+import { Stat } from "#app/data/pokemon-stat";
+import { allMoves } from "#app/data/move";
+import { BattlerIndex } from "#app/battle";
+import { Species } from "#enums/species";
+import { Moves } from "#enums/moves";
+
+describe("Moves - Fusion Flare and Fusion Bolt", () => {
+  let phaserGame: Phaser.Game;
+  let game: GameManager;
+
+  const fusionFlare = allMoves[Moves.FUSION_FLARE];
+  const fusionBolt = allMoves[Moves.FUSION_BOLT];
+
+  beforeAll(() => {
+    phaserGame = new Phaser.Game({
+      type: Phaser.HEADLESS,
+    });
+  });
+
+  afterEach(() => {
+    game.phaseInterceptor.restoreOg();
+  });
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    vi.spyOn(overrides, "MOVESET_OVERRIDE", "get").mockReturnValue([ fusionFlare.id, fusionBolt.id ]);
+    vi.spyOn(overrides, "STARTING_LEVEL_OVERRIDE", "get").mockReturnValue(1);
+
+    vi.spyOn(overrides, "OPP_SPECIES_OVERRIDE", "get").mockReturnValue(Species.RESHIRAM);
+    vi.spyOn(overrides, "OPP_MOVESET_OVERRIDE", "get").mockReturnValue([ Moves.REST, Moves.REST, Moves.REST, Moves.REST ]);
+
+    vi.spyOn(overrides, "BATTLE_TYPE_OVERRIDE", "get").mockReturnValue("double");
+    vi.spyOn(overrides, "STARTING_WAVE_OVERRIDE", "get").mockReturnValue(97);
+    vi.spyOn(overrides, "NEVER_CRIT_OVERRIDE", "get").mockReturnValue(true);
+
+    vi.spyOn(fusionFlare, "calculateBattlePower");
+    vi.spyOn(fusionBolt, "calculateBattlePower");
+  });
+
+  it("FUSION_FLARE should double power of subsequent FUSION_BOLT", async() => {
+    await game.startBattle([
+      Species.ZEKROM,
+      Species.ZEKROM
+    ]);
+
+    game.doAttack(getMovePosition(game.scene, 0, fusionFlare.id));
+    game.doSelectTarget(BattlerIndex.ENEMY);
+
+    game.doAttack(getMovePosition(game.scene, 0, fusionBolt.id));
+    game.doSelectTarget(BattlerIndex.ENEMY);
+
+    await game.phaseInterceptor.to(TurnStartPhase, false);
+
+    // Force user party to act before enemy party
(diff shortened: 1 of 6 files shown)


xsn34kzx:
Unit tests are now bundled in this PR for both moves. Two move-specific tests were added to test their individual secondary properties (whether `Fusion Flare` thaws freeze or not and whether `Fusion Bolt` takes damage from contact-based abilities, specifically `Rough Skin`).

xsn34kzx:
New unit tests were added to verify that an original corner case / bug has now been addressed (the user targeting their own ally would not add enemies to `pokemonActed` list and therefore would not accurately determine the last `Pokemon` that acted that turn).

src/test/moves/fusion_flare_bolt.test.ts
```diff
@@ -94,18 +93,17 @@ describe("Moves - Fusion Flare and Fusion Bolt", () => {
...
     game.doAttack(getMovePosition(game.scene, 0, moveOne));
     game.doSelectTarget(BattlerIndex.ENEMY);
 
     game.doAttack(getMovePosition(game.scene, 0, moveTwo));
     game.doSelectTarget(BattlerIndex.ENEMY);
 
+    await game.phaseInterceptor.to(TurnStartPhase, false);
+
+    // Force user party to act before enemy party
+    (game.scene.getCurrentPhase() as TurnStartPhase).getOrder = vi.fn().mockReturnValue([ BattlerIndex.PLAYER, BattlerIndex.PLAYER_2, BattlerIndex.ENEMY, BattlerIndex.ENEMY_2 ]);
```
flx-sta, line 105: Too bad `vi.spyOn()` doesn’t work cause overwriting the actual variable might cause unexpected side effects due to not being reset manually properly at the end

xsn34kzx: Had a shower thought and dug around for an example on the [cheat sheet](https://vitest.dev/guide/mocking#cheat-sheet), all instances of `vi.fn()` are now converted to `vi.spyOn()` 🎉

flx-sta: Showers. The ultimate debugger

flx-sta (changes requested):

src/test/moves/fusion_flare_bolt.test.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    // Get rid of any modifiers that may alter power
+    game.scene.clearEnemyHeldItemModifiers();
+    game.scene.clearEnemyModifiers();
+
+    // Ensure survival of team by reducing enemy attack stat and boosting party defense stat
+    enemyParty[0].stats[Stat.SPATK] = 1;
+    enemyParty[1].stats[Stat.SPATK] = 1;
+    party[0].stats[Stat.SPDEF] = 250;
+    party[1].stats[Stat.SPDEF] = 250;
```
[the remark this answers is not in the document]
flx-sta: You cannot mock a specific array index but you can mock the whole array. Thus you have to buffer the original one (unless you don't care about the other values) and then edit only the value you want to change in the mock return

src/test/moves/fusion_flare_bolt.test.ts
```diff
@@ -0,0 +1,299 @@
...
```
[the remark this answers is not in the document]
flx-sta: You cannot mock a specific array index but you can mock the whole array. Thus you have to buffer the original one (unless you don't care about the other values) and then edit only the value you want to change in the mock return\

flx-sta (approved):
lgtm

Tempo-anon (approved):
I wonder if similar logic can be used for the pledge moves (nobody will ever use the pledge move interactions)

PR #1776, opened 2024-06-04
[Ability] Implement Unseen Fist

innerthunder:
## What are the changes?

This implements the effects of the ability [Unseen Fist](https://bulbapedia.bulbagarden.net/wiki/Unseen_Fist_(Ability)).

Unseen Fist isn't currently implemented, and with most (if not all) protection moves now implemented, it's possible to cover all edge cases and fully implement this ability.

## What did change?

- `data/ability`: A new attribute `IgnoreProtectOnContactAbAttr` is added, along with a global (`export`) function that checks if the attribute can be applied. Unseen Fist is now given this attribute.
- `field/pokemon`, `data/move`, and `phases`: All protection checks account for Unseen Fist using the function `data/ability#canApplyIgnoreProtectOnContact()`.

### Screenshots/Videos

For both test videos below, Pokemon and moves are preset using `src/overrides.ts`

- All battles are forced to be double battles.
- The player's Pokemon are Urshifu (with Unseen Fist) and Charmander (without Unseen Fist). Both Pokemon are given Breaking Swipe (a move that targets all nearby foes **and makes contact**) and Rock Slide (a move that targets all nearby foes **and does not make contact**).
- The enemy Pokemon are Ninjasks with Protect and Wide Guard.

#### Test 1: Unseen Fist vs. Protect

- Both opposing Ninjasks use Protect.
- Urshifu uses Breaking Swipe. Because of Unseen Fist, this bypasses Protect and hits both Ninjasks, causing one to faint.
- Charmander also uses Breaking Swipe on the surviving Ninjask, which is blocked by Protect.

https://github.com/pagefaultgames/pokerogue/assets/168692175/89aa8e2a-8f5c-4424-ad9f-0eed9a26bcd8

#### Test 2: Unseen Fist vs. Wide Guard

- Both opposing Ninjasks use Wide Guard
- Urshifu uses Breaking Swipe. Because of Unseen Fist, this bypasses Wide Guard and hits both Ninjasks.
- Charmander also uses Breaking Swipe, which is blocked by Wide Guard.
- The next turn, the surviving Ninjask uses Wide Guard again, and both Urshifu and Charmander use Rock Slide
- Since Rock Slide doesn't make contact, Urshifu and Charmander's attacks are both blocked by Ninjask's Wide Guard.

https://github.com/pagefaultgames/pokerogue/assets/168692175/2c1f48bc-013f-4000-87dd-4348c59a097e

## How to test the changes?

In `src/overrides.ts`:
- Set `DOUBLE_BATTLE_OVERRIDE` to `true`.
- Set `STARTER_SPECIES_OVERRIDE` to `Species.URSHIFU`.
- For `MOVESET_OVERRIDE`, include moves that make/do not make contact, and if testing conditional protect moves e.g. Quick Guard, include moves that would activate protection.
- Set `OPP_MOVESET_OVERRIDE`, fill the array with 4 copies of the protect move you want to test against. The opponent AI gives the conditional protect moves low priority, so it's important to overwrite their other moves.
- `npm run start:dev`, then test the effects of Unseen Fist as shown in the videos above

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -3393,6 +3393,9 @@ export class SuppressFieldAbilitiesAbAttr extends AbAttr {
 
 export class AlwaysHitAbAttr extends AbAttr { }
 
+/** Attribute for abilities that allow moves that make contact to ignore protection (i.e. Unseen Fist) */
+export class IgnoreProtectOnContactAbAttr extends AbAttr { }
+
 export class UncopiableAbilityAbAttr extends AbAttr {
   constructor() {
     super(false);
@@ -4641,7 +4644,7 @@ export function initAbilities() {
     new Ability(Abilities.QUICK_DRAW, 8)
       .unimplemented(),
     new Ability(Abilities.UNSEEN_FIST, 8)
-      .unimplemented(),
+      .attr(IgnoreProtectOnContactAbAttr),
     new Ability(Abilities.CURIOUS_MEDICINE, 8)
       .attr(PostSummonClearAllyStatsAbAttr),
     new Ability(Abilities.TRANSISTOR, 8)
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -9,7 +9,7 @@ import { Type } from "./type";
 import * as Utils from "../utils";
 import { WeatherType } from "./weather";
 import { ArenaTagSide, ArenaTrapTag } from "./arena-tag";
-import { UnswappableAbilityAbAttr, UncopiableAbilityAbAttr, UnsuppressableAbilityAbAttr, BlockRecoilDamageAttr, BlockOneHitKOAbAttr, IgnoreContactAbAttr, MaxMultiHitAbAttr, applyAbAttrs, BlockNonDirectDamageAbAttr, applyPreSwitchOutAbAttrs, PreSwitchOutAbAttr, applyPostDefendAbAttrs, PostDefendContactApplyStatusEffectAbAttr, MoveAbilityBypassAbAttr, ReverseDrainAbAttr, FieldPreventExplosiveMovesAbAttr, ForceSwitchOutImmunityAbAttr, BlockItemTheftAbAttr, applyPostAttackAbAttrs, ConfusionOnStatusEffectAbAttr, HealFromBerryUseAbAttr } from "./ability";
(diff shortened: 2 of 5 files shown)


OrangeRed:
I think it would be better to just create a new PreAttackAbAttr that simply calls `Move.ignoresProtect()` on the user's move if it has the `MoveFlags.MAKES_CONTACT` flag enabled. That way you eliminate having to make a new function and make changes to Pokemon.ts and Phases.ts

innerthunder:
## Updates
- `canApplyIgnoreProtectOnContact()` was removed. The logic to apply `IgnoreProtectOnContactAbAttr` is now in `Move.checkTag()` in `data/move.ts`, and is structured similarly to the logic of Long Reach and Mold Breaker's attributes.
- The protection checks in `field/pokemon.ts` and `phases.ts` now use `Move.checkTag()` instead of `Move.hasTag()` to factor in the effects of Unseen Fist.

Will upload more test video with these updates tomorrow (if needed).

src/data/ability.ts
```diff
@@ -3199,6 +3199,9 @@ export class SuppressFieldAbilitiesAbAttr extends AbAttr {
 
 export class AlwaysHitAbAttr extends AbAttr { }
 
+/** Attribute for abilities that allow moves that make contact to ignore protection (i.e. Unseen Fist) */
+export class IgnoreProtectOnContactAbAttr extends AbAttr { }
+
```
OrangeRed, line 3204: Why can't you implement all of this logic inside of this attribute and trigger it before a move checks if its protected against on line 2752 in phases.ts?
Is there something stopping you from doing something like this in the apply function?
```ts
export class IgnoreProtectClass extends PreAttackAbAttr {
  applyPreAttack(pokemon: Pokemon, passive: boolean, defender: Pokemon, move: PokemonMove, args: any[]): boolean | Promise<boolean> {
    if (move.getMove().hasFlag(MoveFlags.MAKES_CONTACT)) {
      move.getMove().makesContact(true)
    }

    return true
  }
}
```

innerthunder: `move.getMove().ignoresProtect()` changes the move globally. If I did this, another Pokemon without Unseen Fist could then use the same move and bypass Protect.

With the required checks now moved into `Move.checkTag()`, the code for this ability is logically similar to how Long Reach and Mold Breaker are applied.

innerthunder:
## Test Video with Updates

- The opposing Yungoos uses Wide Guard, then Urshifu uses Earthquake. Since Earthquake does not make contact and is thus unaffected by Unseen Fist, Yungoos is protected from the attack.
- Yungoos uses Wide Guard again the next turn, but this time Urshifu uses Breaking Swipe, which does make contact. Because of Unseen Fist, Urshifu's Breaking Swipe bypasses protection and KO's Yungoos.

https://github.com/pagefaultgames/pokerogue/assets/168692175/07c922dd-3900-461a-b023-ad192c444003

OrangeRed:
You might want to write some test cases, especially for the edge cases.

OrangeRed (approved):

Tempo-anon (approved):

PR #1780, opened 2024-06-04
[Move] Psycho Shift treats Comatose as Sleep

CarsonVB:
Checked other PRs, didn't see this before

## What are the changes?
This PR will allow a user with the ability Comatose to use Psycho Shift to put the enemy Pokemon to sleep

In the Pokemon games, no Pokemon that learns Psycho Shift has Comatose, so we have to assume a few things about how Comatose is supposed to work in this case.
According to Bulbapedia - while a Pokemon with the ability Comatose technically does not have the Sleep status condition, moves which check for the sleep status condition pass on a Pokemon with Comatose. Snore and Sleep Talk may be used, Hex and Wake-Up Slap deal increased damage and it can be dealt damage by Dream Eater, Nightmare, and Bad Dreams. Additionally, it cannot be afflicted with any status condition, even with Mold Breaker, Turboblaze, Teravolt, and Sunsteel Strike + Poison Touch.
With these facts, we can assume that it is treated as having the Sleep status condition, without it being visibly applied. Therefore, I believe it's a reasonable expectation that Psycho Shift used by a Pokemon with Comatose would induce Sleep on the target.
It can be a cool niche interaction for fusions with Komala(or Jirachi if you have the passive unlocked).

## What did change?
Simply, a Pokemon with Comatose using Psycho Shift will afflict the target with the Sleep status condition. Previously, this was not the case.

#### Before

https://github.com/pagefaultgames/pokerogue/assets/59876651/ec7cac46-040f-4b5e-b48b-5364a3c3ba87

https://github.com/pagefaultgames/pokerogue/assets/59876651/dab0a1ca-6428-4afb-928f-8cf7f7585da2

#### After

https://github.com/pagefaultgames/pokerogue/assets/59876651/f93133da-db78-402b-b13c-1d3f8fe6d833

https://github.com/pagefaultgames/pokerogue/assets/59876651/1834fa1b-d89c-4670-9cbf-8c2262cbfdcc

#### Checking to ensure base functionality still intact

https://github.com/pagefaultgames/pokerogue/assets/59876651/f4bae3bb-d112-4c3c-85e0-7ade9e18f7bc

Fail

https://github.com/pagefaultgames/pokerogue/assets/59876651/5b8e40ab-4279-410e-bd67-915005f54ec5

No Status

https://github.com/pagefaultgames/pokerogue/assets/59876651/53d1ead9-0552-4164-9b5e-47e17cbf1cae

Burn

https://github.com/pagefaultgames/pokerogue/assets/59876651/cfac1b2d-71b3-43b9-ba18-80b66bbd405a

Paralyzed

https://github.com/pagefaultgames/pokerogue/assets/59876651/cdba8247-3152-408b-b921-6076d7701a18

Poison

https://github.com/pagefaultgames/pokerogue/assets/59876651/5ad076a8-5618-4fde-9194-6f2d46856a6f

Sleep

https://github.com/pagefaultgames/pokerogue/assets/59876651/9ffb3d0d-961e-4f3c-bdc2-e3b1b3e37669

Toxic

## How to test the changes?
Override the starter as Komala, and give it the move Psycho Shift.

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -6164,8 +6166,9 @@ export function initMoves() {
         || user.status?.effect === StatusEffect.POISON
         || user.status?.effect === StatusEffect.TOXIC
         || user.status?.effect === StatusEffect.PARALYSIS
-        || user.status?.effect === StatusEffect.SLEEP)
-        && target.canSetStatus(user.status?.effect, false, false, user)
+        || user.status?.effect === StatusEffect.SLEEP
+        || user.hasAbility(Abilities.COMATOSE))
+        && target.canSetStatus(user.status?.effect ?? (user.hasAbility(Abilities.COMATOSE) ? StatusEffect.SLEEP : undefined), false, false, user)
```
OrangeRed, line 6171: A lot of conditionals like this is really hard to parse in my opinion. Can you instead clean up this function so it looks something like this
```ts
(user, target, move) => {
  let statusToApply = // Sleep if comatose undefined otherwise

  // Psycho Shift cures any non-volatile status condition the user has and afflicts the target with it.
  if (user.status?.effect && isNonVolatileStatusEffect(user.status.effect)) {
    statusToApply = user.status.effect
  }

  return  statusToApply && target.canSetStatus(statusToApply, ...)
}
```

OrangeRed (changes requested):

src/data/move.ts
CarsonVB: I agree, was just adding functionality to the existing code, but in future I will refactor/optimize as I see fit.

Issue #1745, opened 2024-06-02
[BUG] Shiny variants failing to generate for certain Pokemon with different forms


Tempo-anon:
**Describe the bug**
![](https://cdn.discordapp.com/attachments/1176874654015684739/1246965641341046884/image.png?ex=665e4ec0&is=665cfd40&hm=63719497f32c7910006132f3c111f700f03cfecf6e0d66296b1299fdba0bddb3&)

**To Reproduce**
Override a Wormadam battle
Modify the trySetShiny code so it tries to set a variant
log the relevant information

**Expected behavior**
Wormadam variants should generate properly since they exist in the game

**Screenshots / Videos**
n/a

**Device**
n/a

**Additional context**
Might also affect other Pokemon with forms

PR #1783, opened 2024-06-04
[Bug] Fix generateVariant to account for forms

c4vv:
## What are the changes?

Currently Pokemon with forms (with the exception of forms that had a formKey of "", an empty string) that have shiny variants cannot spawn as Rare or Epic shinies. This change will allow them to properly generate. 

Hoopa and Castform both have a base form that has a key of an empty string, which allowed the old code to evaluate correctly without a formKey and only their species ID, ex: 720 for Hoopa.

Because blue and red shiny good.

[Issue](https://github.com/pagefaultgames/pokerogue/issues/1745)

## What did change?

Adds formKey to the check to see if variantData has a variant for the given Pokemon and its given form.

### Screenshots/Videos

I just recorded about 12 different tests and all of the files are corrupted so I'm uploading this now and will add them in the morning.

image

Tested so far:
- Burmy
- Wormadam
- Shellos
- Gastrodon
- Vivillion
- Meloetta
- Shaymin
- Hoopa

## How to test the changes?

To test, you can set the OPP_SHINY_OVERRIDE to true as that will always cause the opposing Pokemon to be shiny but not guarantee its variant. Before the changes it will always pass just the species index in, and you can see that it will return that there is no data for Pokemon like Gastrodon or Wormdam. After the changes proposed you will variants for both of those examples.

This logging in the generate variant function might also prove useful to test:
```
console.log("old check");
console.log(this.species.speciesId);
console.log(variantData.hasOwnProperty(this.species.speciesId));
console.log("new check");
console.log(variantDataIndex);
console.log(variantData.hasOwnProperty(variantDataIndex));
```

## Checklist
- [x] Are the changes visual?
  
  
image

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1273,7 +1273,15 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
    * @returns the shiny variant
    */
   generateVariant(): Variant {
-    if (!this.shiny || !variantData.hasOwnProperty(this.species.speciesId)) {
+    const formIndex = this.formIndex;
+    let variantDataIndex = this.species.speciesId;
+    if (this.species.forms.length > 0) {
+      const formKey = this.species.forms[formIndex]?.formKey;
+      if (formKey) {
+        variantDataIndex = `${variantDataIndex}-${formKey}`;
+      }
+    }
+    if (!this.shiny || !variantData.hasOwnProperty(variantDataIndex)) {
       return 0;
     }
     const rand = Utils.randSeedInt(10);
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1281,7 +1281,8 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
         variantDataIndex = `${variantDataIndex}-${formKey}`;
       }
     }
-    if (!this.shiny || !variantData.hasOwnProperty(variantDataIndex)) {
+    // Checks if there is no variant data for both the index or index with form
+    if (!this.shiny || (!variantData.hasOwnProperty(variantDataIndex) && !variantData.hasOwnProperty(this.species.speciesId))) {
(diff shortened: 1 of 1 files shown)


c4vv:
## !this.shiny
I have no idea why this is included as generateVariant _should_ only be called when a shiny is generated, but I believe tampering with is outside the scope of this PR at any rate. Leaving that unchanged at the cost of a bit of a clunky if statement. We are just looking to see if there isn't variant data for both ID and ID + form key.

The if statement seems to be functioning correctly from my testing, but more testing would definitely be required. 

## Videos
All show the start of round due to the fact that generateVariant only runs then. Overridden species will take the session data variant, an important note for testing.

### Wormadam
https://github.com/pagefaultgames/pokerogue/assets/132619649/69d6fb66-cbc9-467c-ab17-1178775b66e8

### Shaymin
https://github.com/pagefaultgames/pokerogue/assets/132619649/d8076485-4c97-48ae-9593-30e1d847dbd9

### Gastrodon
https://github.com/pagefaultgames/pokerogue/assets/132619649/ec2da03c-45e4-4cdd-b903-563969567dc7

### Scatterbug
Works when there is no variant data for the forms, but there is data for the base ID
https://github.com/pagefaultgames/pokerogue/assets/132619649/0beff8e3-fe69-4ba9-9a2c-d3727d17bf7a

### Vivillion 
Works when there is no variant data for the base ID, but there is for the forms.
https://github.com/pagefaultgames/pokerogue/assets/132619649/0a5fffba-e022-457a-b0a3-df9dd7df49ed

c4vv:
I don't understand how to resolve the issue that Github Pages is taking as it is describing the variable variantDataIndex as a Species when it is not one:

` let variantDataIndex = this.species.speciesId;`

 How can I run this check myself to troubleshoot the issue?

Tempo-anon (approved):

PR #1784, opened 2024-06-04
[BUG] fix language auto detect

BubbleBear:
## What are the changes?
currently language auto detection is broken,
fixed according to [i18next documentation](https://www.i18next.com/how-to/faq#how-should-the-language-codes-be-formatted)

currently language auto detection is broken for **Simplified Chinese, Traditional Chinese, Portugese Brazilian**

## What did change?
i18next config resource key

#### BEFORE FIX
image

----

#### AFTER FIX
image

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/plugins/i18n.ts
+++ b/src/plugins/i18n.ts
@@ -174,13 +174,13 @@ export function initI18n(): void {
       de: {
         ...deConfig
       },
-      pt_BR: {
+      "pt-BR": {
         ...ptBrConfig
       },
-      zh_CN: {
+      "zh-CN": {
         ...zhCnConfig
       },
-      zh_TW: {
+      "zh-TW": {
         ...zhTwConfig
       },
       ko: {
--- a/src/locales/zh_CN/tutorial.ts
+++ b/src/locales/zh_CN/tutorial.ts
@@ -1,7 +1,7 @@
 import { SimpleTranslationEntries } from "#app/plugins/i18n";
 
 export const tutorial: SimpleTranslationEntries = {
-  "intro": `欢迎来到PokéRogue！这是一款以战斗为核心的融合了roguelite元素的宝可梦同人游戏。
+  "intro": `欢迎来到PokéRogue！这是一款以战斗为核心的\n融合了roguelite元素的宝可梦同人游戏。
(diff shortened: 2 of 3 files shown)


Tempo-anon:
Can you show some proof that this works?

src/plugins/i18n.ts
```diff
@@ -174,13 +174,13 @@ export function initI18n(): void {
...
-      pt_BR: {
+      "pt-BR": {
         ...ptBrConfig
       },
-      zh_CN: {
+      "zh-CN": {
         ...zhCnConfig
       },
-      zh_TW: {
+      "zh-TW": {
```
bennybroseph, line 183: assuming the rest is correct, this adds quotes here which cannot be correct

bennybroseph (changes requested):

src/plugins/i18n.ts
BubbleBear: this is the grammer, you cannot use `-` in object key unless wrap it in quotes

BubbleBear:
> Can you show some proof that this works?

sure thing sir

BubbleBear:
@Tempo-anon @bennybroseph 
Thanks for your review and patiences, I've added some screenshot proofs.

davidyan-hub:
I did not have envirement to do this...
Who can modify the text.ts to change the font-family and font-size for chinese.

Code as below:

fontFamily: localStorage.getItem('prLang')==="zh_CN"?'simhei':'emerald',

styleOptions.fontFamily = localStorage.getItem('prLang')==="zh_CN"?'MuZaiPixel':'pkmnems';

and same judgement logic for changing font-size, let the size is smaller.

bennybroseph (approved):

BubbleBear:
> I did not have envirement to do this... Who can modify the text.ts to change the font-family and font-size for chinese.
> 
> Code as below:
> 
> fontFamily: localStorage.getItem('prLang')==="zh_CN"?'simhei':'emerald',
> 
> styleOptions.fontFamily = localStorage.getItem('prLang')==="zh_CN"?'MuZaiPixel':'pkmnems';
> 
> and same judgement logic for changing font-size, let the size is smaller.

it's not like that easy, i've tried but the font size logic runs different from CSS

PR #1788, opened 2024-06-04
[Bug] Epic Drifblim using Drifloon sprite

chaosgrimmon:
## What are the changes?
Epic shiny Drifblim gets a perfectly functional animated backsprite, and stops using incorrectly-sized slices of epic Drifloon.

There have been so many reports of this I swear.
image

## What did change?
`2`, means to "use the `426_3.png` provided", which is the stray Drifloon sprite.
`1`, instead, tells it to literally just recolor the default Drifblim sprite, the way it does perfectly well for its rare variant. 

### Screenshots/Videos
Before:
image
After:
image

## How to test the changes?
overrides.ts can set a `SHINY=true, VARIANT=2` `Species.DRIFBLIM` as your first starter.

## Checklist
- [x] Are the changes visual?


chaosgrimmon:
Frankly I have no idea why [pokemon/variant/back/426_3.png](https://github.com/pagefaultgames/pokerogue/blob/main/public/images/pokemon/variant/back/426_3.png) exists. Obviously the numbering is off by one, but even then, the colours match, and it doesn't seem to add any detail that isn't already in Drifloon's epic variant.

It's not an inexplicably load-bearing TF2!coconut.jpg or anything either, because I deleted it without so much as a console warning.
The #artist backlog might have more details, but I don't have access to those, so :shrug:

Tempo-anon (approved):
426_3 was added by Sam

PR #1792, opened 2024-06-04
[QoL] Auto Size Text

FredeX:
Im working an a wrapper class for BBCodeText and Phaser.GameObjects.Text that allows auto sized text, meaning text that automatically is **downsized** to fit a certain width. When downsizing, the texts paddingtop  is used to center the text to make downsizing visually pleasing ;) This can be very useful for overflowing text issues (e.g. happens in german a lot) and eliminates manually adjusting font sizes all across the game for every localization.

The wrapper classes (AutoText & AutoBBCodeText) have a new **optional** parameter `limits: {right?: number, width?: number}`. Existing text in code can easily be limited and works independent of selected language.

Example for using this:
old label:
`this.label= addTextObject(this.scene, 50, 18, "test", TextStyle.WINDOW, { fontSize: 50});`
new label with max width limit added:
`this.label= addTextObject(this.scene, 50, 18, "test", TextStyle.WINDOW, { fontSize: 50}, { width: 30});`

Auto Text in action:

Here is the `limit.width` used. Works basically as a max width value. 

Screenshot 2024-06-09 095900

.
.

Here is the `limit.right` used. This is different to `width` because the ability-, passive- and nature-label can have different x positions per localization. So the values are limited by their right bound (text.x + text.width).

https://github.com/pagefaultgames/pokerogue/assets/1838682/5d85e63f-9331-4b3a-a3e5-42027fa6129a

.
.
.

Here is an example in portugese. For portugese the font size was downsized in general for the texts to fit the container. With my changes, portugese can keep the original font size and only adjust when needed.

Before:
Screenshot 2024-06-09 102805
After:

https://github.com/pagefaultgames/pokerogue/assets/1838682/fa682a8a-4fbb-4c84-8f46-652d58210fa8


CodeTappert:
Text seems to big for the box now.  Cant you just reduce the spez def, spez atk thing in size a little. Not the rest

CodeTappert:
Of course having it be set automatically (with a upper border) sounds better then hardcoding the size

FredeX:
What do you think of this as a solution? Espcially noticable are the last 3 natures in the video: "Forsch" und "Sacht" are slightly downscaled - still centered though - and "Kauzig" is the normal font size.

This could be used all over the project if needed.

https://github.com/pagefaultgames/pokerogue/assets/1838682/da686980-e413-455a-a9c0-38f75a66c18e

CodeTappert:
> What do you think of this as a solution? Especially noticable are the last 3 natures in the video: "Forsch" und "Sacht" are slightly downscaled - still centered though - and "Kauzig" is the normal font size.
> 
> This could be used all over the project if needed.
> 
>  autoresize.mp4

Looks good! Does this impact other languages?

FredeX:
Yes, it would, but there are 2 ways to limit a text, "text width" and "right" position (x+width). 

In the image you can see the green debug line. This is not a the "width" limit but the "right" limit. So it does not matter where the text starts or which font size it's suppose to be. The logic for "right" is simple: `while(x+width > limit.right) reduce font size`

image

CodeTappert:
Please look if it still readable at a phone. If not this wont work

FredeX:
Big update for this. Fixed some bugs and massivly improved the performance. Tested this in many scenarios and seems to work very nicely!
Regarding mobile: I set a minFontSize which makes sure no matter what the text is still readable (it's still bigger than the move font size).

FredeX:
Okay I think I'm ready to leave draft now.
Not sure about what should be included in this PR. I was thinking just the wrapper classes and methods but applying them should be seperate PRs

CodeTappert:
PLEASE fix your conflicts. I really want to see this in the game

flx-sta:
### Are you still working on this PR?
-----
No change since June 9 and open conflicts

flx-sta:
Closing due to inactivity.
If you want to re-open it let us know

PR #1795, opened 2024-06-04
[Bug] Handle if the browser gives a long form of a language (like "de-DE") …

CodeTappert:
…for cases where we only have the short form "de".

## What are the changes?

Used https://www.i18next.com/overview/api#resolvedlanguage instead of language

While testing my browser autodetected the language to "de-DE" (German - Germany). So the type icons were not loaded correctly in german but in the fallback english. Same would happen for "fr-FR" (French - France) and others.

## What did change?

Changes i18next.language to i18next.resolvedLanugage everywhere

So for example if the browser says "de-DE". It will search that, not find it. Then try "de" find that and set it as a language.

This doesnt break the fallback to english at all that we have for completly unsupported languages (like currently for example CZ or something)

### Screenshots/Videos

None

## How to test the changes?

Set your browser to "de-DE" or "fr-FR". It should still load the type images in the correct languages

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/loading-scene.ts
+++ b/src/loading-scene.ts
@@ -29,6 +29,28 @@ export class LoadingScene extends SceneBase {
   }
 
   preload() {
+
+    let languageToSet = "en";
+    // Check if the language is valid, if not, try to load the language without the region and if that fails, fall back to English
+    if (!Utils.verifyLang(i18next.language) && i18next.language.includes("-")) {
+      // Try without the region
+      const lang = i18next.language.split("-")[0];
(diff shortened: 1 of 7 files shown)


CodeTappert:
I was shown a way to do it with i18next itself. So i will correct this pr

flx-sta (approved):
Build ✅ 
Tests ✅ 
Launch & Play ✅ 
Code review ✅

bennybroseph (approved):

PR #1799, opened 2024-06-04
[Move] Implemented Court Change

MutenYoshii:
## What are the changes?

Added Court Change into the game. 

The move was unimplemented

## What did change?

 Added a swaptagsattr to move.ts. Also added a boolean value named quiet to ignore onAdd and onRemove tags when the move is used in arena-tag and arena.ts

## How to test the changes?

I tested the move through override and the enemy would use on field effects, move works with spikes, toxic spikes, stealth rock,
sticky web, tailwind, light screen, reflect, and mist

I did.

No

## Checklist
- [x] Are the changes visual?

Sticky Web:
https://github.com/pagefaultgames/pokerogue/assets/171395666/488d7831-cd10-4c2c-b327-8fab43232ebc
Mist: 
https://github.com/pagefaultgames/pokerogue/assets/171395666/b1a5fce9-d4e9-4156-b066-a026a4e4fd09
Light Screen:
https://github.com/pagefaultgames/pokerogue/assets/171395666/2d027324-7749-446f-9bff-0c898770c8bf
Toxic Spikes:
https://github.com/pagefaultgames/pokerogue/assets/171395666/0b9eb43e-012d-46a9-9e2a-d919ae2b3ced
Spikes:
https://github.com/pagefaultgames/pokerogue/assets/171395666/4081f822-707e-47a4-9c84-d453a2d0519a

### DIFF
--- a/src/data/arena-tag.ts
+++ b/src/data/arena-tag.ts
@@ -28,23 +28,25 @@ export abstract class ArenaTag {
   public sourceMove: Moves;
   public sourceId: integer;
   public side: ArenaTagSide;
+  public quiet: boolean;
 
-  constructor(tagType: ArenaTagType, turnCount: integer, sourceMove: Moves, sourceId?: integer, side: ArenaTagSide = ArenaTagSide.BOTH) {
+  constructor(tagType: ArenaTagType, turnCount: integer, sourceMove: Moves, sourceId?: integer, side: ArenaTagSide = ArenaTagSide.BOTH, quiet: boolean = false) {
     this.tagType = tagType;
     this.turnCount = turnCount;
     this.sourceMove = sourceMove;
     this.sourceId = sourceId;
     this.side = side;
+    this.quiet = quiet;
   }
 
   apply(arena: Arena, args: any[]): boolean {
     return true;
   }
 
-  onAdd(arena: Arena, quiet: boolean = false): void { }
+  onAdd(arena: Arena): void { }
 
-  onRemove(arena: Arena, quiet: boolean = false): void {
-    if (quiet === false) {
+  onRemove(arena: Arena): void {
+    if (!this.quiet) {
       arena.scene.queueMessage(`${this.getMoveName()}\'s effect wore off${this.side === ArenaTagSide.PLAYER ? "\non your side" : this.side === ArenaTagSide.ENEMY ? "\non the foe's side" : ""}.`);
     }
   }
@@ -67,11 +69,11 @@ export class MistTag extends ArenaTag {
     super(ArenaTagType.MIST, turnCount, Moves.MIST, sourceId, side);
   }
 
-  onAdd(arena: Arena, quiet: boolean): void {
-    super.onAdd(arena, quiet);
+  onAdd(arena: Arena): void {
+    super.onAdd(arena);
 
     const source = arena.scene.getPokemonById(this.sourceId);
-    if (quiet === false) {
+    if (!this.quiet) {
       arena.scene.queueMessage(getPokemonMessage(source, "'s team became\nshrouded in mist!"));
     }
   }
@@ -117,8 +119,8 @@ class ReflectTag extends WeakenMoveScreenTag {
     return false;
   }
 
-  onAdd(arena: Arena, quiet: boolean): void {
-    if (quiet === false) {
+  onAdd(arena: Arena): void {
+    if (!this.quiet) {
       arena.scene.queueMessage(`Reflect reduced the damage of physical moves${this.side === ArenaTagSide.PLAYER ? "\non your side" : this.side === ArenaTagSide.ENEMY ? "\non the foe's side" : ""}.`);
     }
   }
@@ -141,8 +143,8 @@ class LightScreenTag extends WeakenMoveScreenTag {
     return false;
   }
 
-  onAdd(arena: Arena, quiet: boolean): void {
-    if (quiet === false) {
+  onAdd(arena: Arena): void {
+    if (!this.quiet) {
       arena.scene.queueMessage(`Light Screen reduced the damage of special moves${this.side === ArenaTagSide.PLAYER ? "\non your side" : this.side === ArenaTagSide.ENEMY ? "\non the foe's side" : ""}.`);
     }
   }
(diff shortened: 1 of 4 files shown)


src/data/arena-tag.ts
```diff
@@ -65,11 +67,13 @@ export class MistTag extends ArenaTag {
     super(ArenaTagType.MIST, turnCount, Moves.MIST, sourceId, side);
   }
 
-  onAdd(arena: Arena): void {
-    super.onAdd(arena);
+  onAdd(arena: Arena, quiet: boolean): void {
+    super.onAdd(arena, quiet);
```
OrangeRed, line 71: if you're going to explicitly check `quiet === false` you should default this variable to false everywhere.

src/data/move.ts
```diff
@@ -4021,6 +4021,55 @@ export class RemoveScreensAttr extends MoveEffectAttr {
...
+    if (!super.apply(user, target, move, args)) {
+      return false;
+    }
+    const swapTags = [ArenaTagType.AURORA_VEIL, ArenaTagType.LIGHT_SCREEN, ArenaTagType.MIST, ArenaTagType.REFLECT, ArenaTagType.SPIKES, ArenaTagType.STEALTH_ROCK, ArenaTagType.STICKY_WEB, ArenaTagType.TAILWIND, ArenaTagType.TOXIC_SPIKES];
+
+    for (const swapTagType of swapTags) {
+
+      const tagPlayerTemp = user.scene.arena.getTagOnSide(swapTagType, ArenaTagSide.PLAYER);
+      const tagEnemyTemp = user.scene.arena.getTagOnSide(swapTagType, ArenaTagSide.ENEMY);
+
```
OrangeRed, line 4047: Is getting arena tags one by one like this really the best option?
You can use `Arena.findTagsOnSide(matchFn)` to get all of the tags that pass the match function and then do enemy swapping / player swapping separately using a shared method.

OrangeRed (changes requested):

src/data/move.ts
MutenYoshii: Sorry, I don't know how to do a Match Function, would it be something like a function with a switch statement that returns true if any of the valid tags are present? Would I put the shared method in arena or in move?

OrangeRed: this is the type defintion of the match function `(t: ArenaTag) => boolean`
So `Arena.getTagsOInSide()` will return all tags that pass (return true) from any function that has the same signature 
For example `(tag) => !!swapTags.includes(tag)` <- will filter out all tags that are inside of swapTags

MutenYoshii: So ```ts const tagPlayerTemp = user.scene.arena.findTagsOnSide((t => swapTags.includes(t.tagType)), ArenaTagSide.PLAYER);``` would work

bennybroseph (changes requested):
I'm not sure that I like this implementation. Does this work with the new Arena Flyout?

MutenYoshii:
Didn't know I could just check the arena effects 

https://github.com/pagefaultgames/pokerogue/assets/171395666/819f9fb1-451e-4c4c-b11c-46017e9eabd4


https://github.com/pagefaultgames/pokerogue/assets/171395666/e1bdb0dd-d9c3-4802-b271-371d77ebf5a5


https://github.com/pagefaultgames/pokerogue/assets/171395666/fa7f48d3-10bc-4c30-882b-7a9a96b453d5


https://github.com/pagefaultgames/pokerogue/assets/171395666/0a12df36-a23e-4043-892d-17c184bdbe8a

bennybroseph (approved):

src/data/arena-tag.ts
```diff
@@ -67,11 +69,11 @@ export class MistTag extends ArenaTag {
     super(ArenaTagType.MIST, turnCount, Moves.MIST, sourceId, side);
   }
 
-  onAdd(arena: Arena, quiet: boolean): void {
```
bennybroseph: I'd rather this than what you have now except `quiet` should be defaulted to false and only Court Change should remove/add it with `quiet` set to true. that way nothing changes for all other situations.

bennybroseph (changes requested):

src/field/arena.ts
```diff
@@ -539,14 +539,15 @@ export class Arena {
     this.applyTagsForSide(tagType, ArenaTagSide.BOTH, ...args);
   }
 
-  addTag(tagType: ArenaTagType, turnCount: integer, sourceMove: Moves, sourceId: integer, side: ArenaTagSide = ArenaTagSide.BOTH, targetIndex?: BattlerIndex): boolean {
+  addTag(tagType: ArenaTagType, turnCount: integer, sourceMove: Moves, sourceId: integer, side: ArenaTagSide = ArenaTagSide.BOTH, quiet: boolean = false, targetIndex?: BattlerIndex): boolean {
```
MutenYoshii, line 542: I added the ```quiet``` boolean and initialized it to false. would that be good or would you want me to just make the change on move.ts

bennybroseph: on arena-tag.ts where I commented

src/data/move.ts
```diff
@@ -4021,6 +4021,48 @@ export class RemoveScreensAttr extends MoveEffectAttr {
...
+  constructor() {
+    super(true, MoveEffectTrigger.POST_APPLY);
+  }
+
+  apply(user:Pokemon, target:Pokemon, move:Move, args: any[]): boolean {
+    if (!super.apply(user, target, move, args)) {
+      return false;
+    }
+    const swapTags = [ArenaTagType.AURORA_VEIL, ArenaTagType.LIGHT_SCREEN, ArenaTagType.MIST, ArenaTagType.REFLECT, ArenaTagType.SPIKES, ArenaTagType.STEALTH_ROCK, ArenaTagType.STICKY_WEB, ArenaTagType.TAILWIND, ArenaTagType.TOXIC_SPIKES];
```
bennybroseph, line 4041: I feel like this should go somewhere outside of this function so that if people implement new swappable tags they can find this easier

bennybroseph (changes requested):

bennybroseph (approved):

PR #1807, opened 2024-06-04
[Ability] Implementation of the Tera Shell ability

PhoenixFarfelu:
## What are the changes?
If everything is correct, we could get rid of the ".unimplemented()" of the tera shell ability, making it possible to use

  - I try to implement new abilities (that already exist in Pokémon)
  - This ability is a signature ability of the legendary Pokémon Terapagos and being unable to use it can be frustrating

## What did change?
  - Now, Terapagos, once tera shifted and at full health, will always take a "not very effective attack" if hit. To do this, I took the amount of damage he should take and, depending on the super effectiveness of the move, divide the amount of damage taken.

## How to test the changes?

  - take Terapagos, throw it to battle and, once tera shifted, see if all attacks results in a "not very effective" move

  - get rid of the ".unimplemented()" at the end of the "TERA_SHELL" ability

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -3272,6 +3272,34 @@ export class IgnoreTypeStatusEffectImmunityAbAttr extends AbAttr {
   }
 }
 
+/*
+ * Here is an idea for the TERA_SHELL ability
+ * I'm not a pro with typescript so there may be mistakes in the code below so please check it before doing anything <3
+ */
+export class AllMovesNotEffective extends AbAttr { // not sure if something should be put in front of "AbAttr"
+  apply(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: PokemonMove, cancelled: Utils.BooleanHolder, args: any[]): boolean {
+    if (pokemon.hp === pokemon.getMaxHp() {
+      if (move.getMove() instanceof AttackMove && pokemon.getAttackTypeEffectiveness(move.getMove().type, attacker) === 4) { // if twice super effective
+        cancelled.value = true; // don't know what it realy does but saw it in other parts of the code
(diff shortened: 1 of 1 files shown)


src/data/ability.ts
```diff
@@ -3272,6 +3272,34 @@ export class IgnoreTypeStatusEffectImmunityAbAttr extends AbAttr {
...
 }
 
+/*
+ * Here is an idea for the TERA_SHELL ability
+ * I'm not a pro with typescript so there may be mistakes in the code below so please check it before doing anything <3
+ */
+export class AllMovesNotEffective extends AbAttr { // not sure if something should be put in front of "AbAttr"
+  apply(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: PokemonMove, cancelled: Utils.BooleanHolder, args: any[]): boolean {
+    if (pokemon.hp === pokemon.getMaxHp() {
+      if (move.getMove() instanceof AttackMove && pokemon.getAttackTypeEffectiveness(move.getMove().type, attacker) === 4) { // if twice super effective
```
hayuna, line 3282: Could we merge these 3 conditions into one?

```javascript 
if (pokemon.hp === pokemon.getMaxHp() {
  const effectiveness = pokemon.getAttackTypeEffectiveness(move.getMove().type, attacker)
  if (move.getMove() instanceof AttackMove && effectiveness >= 1) { 
    cancelled.value = true; 
    const multiplier = 0.5 / effectiveness;
    (args[0] as Utils.NumberHolder).value *= multiplier;
    return true;
  }
}
```

flx-sta:
possible clone of https://github.com/pagefaultgames/pokerogue/pull/695

src/data/ability.ts
PhoenixFarfelu: That's a good idea. It would make the code shorter

PR #1808, opened 2024-06-04
[Feature] [Localization] Add GH action to generate missing translations

francktrouillez:
## What are the changes?

This PR introduces a new GH action to automatically update the translation files whenever a new PR is merged into the `main` branch.

I've noticed that translation are currently being updated manually. I believe this is really nice from the community.
However, I believe it also has some drawbacks:
- It happens that we don't have translations for all languages, and sometimes we forget to update the translations, leading to a fallback to English.
- It generates a lot of localization PRs

I believe that we can do better. I believe we can achieve that by using an automated translation system to keep the translations up-to-date. Of course, it doesn't ensure 100% accuracy, but I believe it's better than not trying at all. It would also help decreasing the number of PRs we have right now.

People can of course still contribute to the translations, but all the "dumb" translations can be handled by the automated system. The more complex or nuanced translations can be reported and fixed by the community.

## What did change?

I added a GH action that currently will run whenever a push to the `main` branch is made.

Currently, the action will make sure that the translations are up-to-date, and will update them if necessary.

To do so, I'm using [DeeplAPI](https://www.deepl.com/pro-api?cta=header-pro-api), that allows for 500k characters to be translated per month for free. There's no particular reason for going with this provider, other than it was easy to use, free, and that it allows to mark some words as "do not translate", which is useful for variables. To use the API, you need to provide an API key, which I've added as a secret in my fork, but you can do the same in the main repo. I allowed to use multiple keys, in case we want to distribute the load between multiple accounts.

Once we get the translations, we update the translation files for all languages. Currently, the translations live in TS files under `src/locales`.

After that, the action will commit and push the changes to the `main` branch.

**The goal of this action is to only ADD new translations, and not to overwrite existing ones.** This is to ensure that we don't overwrite any community contribution. This also means that **we don't want people to auto-fill other languages with EN translations**, as any new translation will not be handled by the action.

I also took the liberty to add a new checkbox in the PR template to remind people about translations.

### Screenshots/Videos

The change is not really visual, but I made a [branch](https://github.com/francktrouillez/pokerogue/tree/test/trigger-translate-action) that actually ran the [action](https://github.com/francktrouillez/pokerogue/actions/runs/9374676075/job/25811247749) on my fork, and the game seems to run fine.

- I tested in a very base game in French:

  https://github.com/pagefaultgames/pokerogue/assets/57403591/b3fcdc4d-4e4a-4b54-a46f-401763f71652

- I tested when meeting the rival at the end of the game, as these dialogues use facial expressions:

  https://github.com/pagefaultgames/pokerogue/assets/57403591/92d39994-8b2f-4138-9e64-60a73f456c74

## How to test the changes?

I tested the action on my fork, and it seems to work fine. You can find a run of the action [here](https://github.com/francktrouillez/pokerogue/actions/runs/9374676075/job/25811247749).

This made changes on my [test branch](https://github.com/francktrouillez/pokerogue/tree/test/trigger-translate-action), and it actually added a [commit](https://github.com/francktrouillez/pokerogue/commit/9c39227ec0e2882e4772bcc26eb5d4d1a32fd9bf) with all the new translations. It also applied the eslint changes, but I believe it's fine to have those changes, as it allows to have a unified code style across translations.

You can actually find new translations in the [commit](https://github.com/francktrouillez/pokerogue/commit/9c39227ec0e2882e4772bcc26eb5d4d1a32fd9bf), like `PGFdialogue:firebreather.encounter.1` in `src/locales/fr/dialogue.ts` that was detected and generated by the action to transform `My flames shall devour you!` into `Mes flammes vous dévoreront !` for instance.

I then tested that the game was working, and it seems to work fine as shown on the videos above.

## Checklist

## Considerations

If we decide to move forward with such action, I believe we have some things to consider.

### Secret ENV variables

If we decide to go with such approach and implementation, we need to not forget to add the secret `DEEPL_API_KEYS` to the main repository, that will contain a comma-separated list of API keys to use. It is well-explained in the [doc](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-a-repository).

### When should we trigger this action?

Right now, I've setup the job to run whenever a push is made to the `main` branch. This is to ensure that the translations are always up-to-date. However, we could also decide to run the job only once a day, or once a week, to reduce the load on both the DeeplAPI and the GH action.

Something else to consider is that we could also make it so that the action doesn't actually push the changes, but creates a PR instead. I believe it adds a bit more friction, but it also allows for a human to review the changes before they are merged.

### Manual changes to translations

If we decide to move forward, we now need to make sure we don't manually change the schema of the translation files, as it could break the action. We could also decide to add a check in the action to ensure that the schema is correct, but I believe it's a bit overkill.

Also, we need to make sure we don't add EN translations to other languages, as these would be ignored by the action. I believe we have such cases here and there, so we might want to regenerate them, but let's not add new ones.

## Thoughts?

I believe this can be a great improvement in terms of localization, and it would allow us to have a more up-to-date game in terms of translations, without exclusively relying on contributions for updating different languages. I also believe it would as a side-effect reduce the number of PRs we have right now.

What do you think? I'm open to any feedback, and I'm happy to make any changes if necessary!


francktrouillez:
FYI, I saw this [PR check](https://github.com/pagefaultgames/pokerogue/actions/runs/9375015279/job/25812312188?pr=1808) failed, but I'll dive into it if the idea described in this PR brings any interest!

CodeTappert:
I am very much against using AI to translate things. Most of the time it does get the context wrong or some specific wording is needed that the AI just can not generate.

And seeing an english string in my lanugage tells me "ok i need to still translate this". So i would say no to that.

Fontbane:
> I am very much against using AI to translate things. Most of the time it does get the context wrong or some specific wording is needed that the AI just can not generate.
> 
> And seeing an english string in my lanugage tells me "ok i need to still translate this". So i would say no to that.
> 
> The current way of adding the english value in all languages is the best way of doing it

I agree with this. Accuracy aside, the lack of a translation is the best way to know when one is needed. Using potentially low quality translations as a placeholder would just muddy things.

francktrouillez:
@CodeTappert 

Thanks for the response and feedback!

> Most of the time it does get the context wrong or some specific wording is needed that the AI just can not generate.

I'm not entirely convinced by that. I think it does a pretty decent job, specially with the current tools we have right now (I'm using DeepL, but it could be anything to be honest, and it could even be ChatGPT if we had an API token, so that we can feed him a context, etc). There are many companies that actually use (and pay for) such tools to automatically translate text to display to their clients/customers. They can always apply changes afterwards, but I have the feeling that it brings more value.

> And seeing an english string in my lanugage tells me "ok i need to still translate this". So i would say no to that.

I agree with you, from a developer perspective. However, not everyone is a developer, or wants to contribute to the project. Also, this assumes you know English. But for someone that actually just want to play the game, I think it's weird to have a mix of English and its native language when playing a game. And I believe this doesn't prevent you from detecting some wrongly translated strings, either in the code itself, or whenever you're playing.

What are your thoughts?

> The current way of adding the english value in all languages is the best way of doing it

I have to disagree on that. I don't think that's the universal best way of doing it. As said before, there are many companies/games that actually use and pay for tools to translate their content. I'm not saying that that's the way to go, but I think it's a strong assumption to say that what we're doing is the best solution.

I feel it is more a matter of preference here, but that's maybe just me. What do you think?

CodeTappert:
I dont want to speak for others. But it seemed pretty unanimous that this isnt wanted at all. But a dev will probably comment here at some point as well.

francktrouillez:
Sure! Thanks for the feedback once again!

In the end, it is just a proposition I made to improve a bit the project, but if the dev team feel it doesn't bring anything, then let's close this!

francktrouillez:
FYI, I'm not gonna spend time keeping this up-to-date with `main` unless the team finds the idea relevant.

I'll wait for a reaction from the core team before continuing to work on this feature.

Adri1:
I think this PR should be closed by now.
I'm totally against AI in translation, in professional / semi-pro works.
Not because I'm a main translator on Rogue. But because of the morallity and the quality. Above all else I can't caution its use when a good chunk of my friends works either in art or have real translation-related jobs.
They currently are all freaking about what they love and graduated for and becoming irrelevant because of all this normalization and of the unregulated use of AI threatening ti put them out of the job.

And atop of this, it's still crappy, "decent" only on the appearance but you can quicly see the fraud once you scratched a bit further. Useful at best for traveling, kinda.

So by respect for them, and because it's still not on point anyway, it can't caution it.

CodeTappert:
Well i agree. We dont want AI translations.

What we would want is a tool that shows us which keys are missing/are not translated in a certain language compared to english. 

So yeah. I will close this

Issue #1767, opened 2024-06-03
[BUG] Toxic Chain Activates on Status Moves


xdotsv:
**Describe the bug**
The Toxic Chain ability badly poisoned my Pokemon when a status move was used

**To Reproduce**
1. Encounter a member of the loyal three
2. Hope that the loyal three member uses a non-damaging move AND gets the 30% toxic chain activation
3. See error

**Expected behavior**
Toxic Chain should only activate on a damaging move.

**Screenshots / Videos**
N/A

**Device**
Android Phone

**Additional context**
In my specific case, it was Fezandipiti using the move Flatter.

DustinLin:
Can we get a source for the expected behavior? I don't see the requirement that it needs to be a damaging move anywhere (bulbapedia, pokemondb, smogon)

drew9696:
I’ve found 2 unofficial sources that reinforce this:
https://www.smogon.com/forums/threads/toxic-chain-update-description.3731624/


https://videogames.si.com/guides/pokemon-scarlet-violet-toxic-chain

Snailman11:
Link to Toxic Chain on Discord https://discord.com/channels/1125469663833370665/1247816303054422061

PR #1440, opened 2024-05-27
Bugfix: Abilities check final move type instead of default move type

td76099:
## What are the changes?
These changes make it so that `Move` object is passed to abilities instead of `PokemonMove` as well as simplifying some of the logic using for abilities that change move type.

## Why am I doing these changes?
Right now, abilities are searching and pulling Moves every single time because they are (mostly) being passed `PokemonMove` instead of `Move`. They usually call `move.getMove()` over and over again instead of just passing it along. This means that if a move has a type change associated with it because of how the move works or an ability, then other abilities do not take those changes into account because they are just looking up the default type. [This bug report](https://discord.com/channels/1125469663833370665/1244862160862052352) lists it pretty well. There's numerous reports in the `bug-report` channel of this exact issue. #1390 is an instance of this bug as well.

[Asked a lead dev](https://discord.com/channels/1125469663833370665/1176874654015684739/1245735588720541827) before I started testing and this was an approved implementation. A previous implementation was trying to implement it on a need-to-know basis using a new [getType() function](https://github.com/td76099/pokerogue/commit/58bca3687d32f7055688fd8fb007e50abb38cbd0#diff-62e9cf38a6ba2a2b69a102788cdfef2a33cdc173f79124df114636685a51e762R187), but it was recommended to only run those move/attribute applications once [based on this convo](https://discord.com/channels/1125469663833370665/1244059545328947300/1244638318734475335). That led to updating it to make sure they only get ran once by passing the `Move` object instead of looking up the move every time with `move.getMove()`.

[These moves](https://bulbapedia.bulbagarden.net/wiki/Category:Moves_that_change_type) and [these abilities](https://bulbapedia.bulbagarden.net/wiki/Category:Abilities_that_can_modify_move_types) will all properly interact with all of [these defending abilities](https://bulbapedia.bulbagarden.net/wiki/Category:Abilities_that_alter_damage_taken). There will be no more bug reports about Liquid Voice Torch Song not being absorbed by Water Absorb or being improperly affected by Heatproof and more.

## What did change?
[...]

PR #1814, opened 2024-06-05
[Bug] Fix Toxic Chain applying for non-damaging moves

drew9696:
## What are the changes?
This changes the behavior of Toxic Chain to only activate on damaging moves

This [bug report](https://github.com/pagefaultgames/pokerogue/issues/1767) outlines the functional behavior of `Toxic Chain`, even if the description is not fully accurate to its effects

Previously, status moves could trigger Toxic Chain which is unacceptable.
Additionally, the default behavior of `PostAttackAbAttr` actually wasn't checking that we were using a damaging move, it triggered if any move was used.

## What did change?
This modifies the base behavior of `PostAttackAbAttr` to check that the move used is a damaging attack and not a `STATUS` move.
* This behavior is applied by default to all extensions of this class, but only needs to be "turned off" in the case of the ability `Poison Puppeteer` which is defined by `ConfusionOnStatusEffectAbAttr`
* All other `PostAttackAbAttr` abilities have been checking explicitly to enforce this condition already in slightly different ways, so this change gives a simpler/more generic way ensuring any future extensions can leverage the "damaging move required" functionality
  * Classes that will apply this behavior and can remove their special handling:
    * `PostAttackStealHeldItemAbAttr`
    * `PostAttackApplyStatusEffectAbAttr`
    * `PostAttackContactApplyStatusEffectAbAttr`
    * `PostAttackApplyBattlerTagAbAttr`

### Screenshots/Videos
*Before my fix:*
https://github.com/pagefaultgames/pokerogue/assets/25591514/d26db051-9612-418e-a6b8-9c1c8f668ba1

*After fix:*
Not really much to show except that toxic chain doesn't occur in this scenario anymore

## How to test the changes?
Can test the fix with the `src/overrides.ts` file
```
export const ABILITY_OVERRIDE: Abilities = Abilities.TOXIC_CHAIN;
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.GROWL];
```

Can regression test by changing the ability to any of the following:
```
Abilities.MAGICIAN
Abilities.POISON_TOUCH
Abilities.STENCH
```

Can regression test poison puppeteer's confusion with the following:
```
export const STARTER_SPECIES_OVERRIDE: Species = Species.PECHARUNT;
export const ABILITY_OVERRIDE: Abilities = Abilities.POISON_PUPPETEER;
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.TOXIC];
```

## Checklist
   - #1440 changes the move type on `applyPostAttack` from `PokemonMove` to `Move`
    - N/A Are all unit tests still passing? (`npm run test`)
- [X] Are the changes visual?


src/data/ability.ts
```diff
@@ -1235,7 +1235,25 @@ export class BattleStatMultiplierAbAttr extends AbAttr {
...
 
 export class PostAttackAbAttr extends AbAttr {
+  private attackRequired: boolean;
+
+  constructor(attackRequired: boolean = true) {
+    super();
+
+    this.attackRequired = attackRequired;
+  }
+
```
OrangeRed, line 1245: I understand what you're doing here but I think passing in a `type PokemonAttackCondition = (user: Pokemon, target: Pokemon, move: Move) => boolean;` function would allow for more control / be more readable than a naked boolean like this.

OrangeRed (changes requested):

src/data/ability.ts
drew9696: Makes sense, but we do lose the `hitResult` check by doing this. I do think upon further investigation that we actually need to have hit before calling this function anyways so I think that was a frivolous check in the first place. I'll change it up, thanks!

src/data/ability.ts
```diff
@@ -1235,23 +1235,42 @@ export class BattleStatMultiplierAbAttr extends AbAttr {
...
+    // When attackRequired is true, we require the move to be an attack move and to deal damage before checking secondary requirements.
+    // If attackRequired is false, we always defer to the secondary requirements.
+    if (this.attackCondition(pokemon, defender, move.getMove())) {
+      return this.applyPostAttackAfterMoveTypeCheck(pokemon, passive, defender, move, hitResult, args);
+    } else {
+      return false;
+    }
+  }
+
+  applyPostAttackAfterMoveTypeCheck(pokemon: Pokemon, passive: boolean, defender: Pokemon, move: PokemonMove, hitResult: HitResult, args: any[]): boolean | Promise<boolean> {
```
OrangeRed, line 1257: Instead of doing all of these renaming shenanigans can you instead invoke `super.applyPostAttack()` in all of children classes that potentially need to check the default move condition?

OrangeRed (changes requested):

src/data/ability.ts
drew9696: I actually prefer the renaming stuff because invoking `super.applyPostAttack()` requires that any future class that extends this also calls that method. By having a second method that is the default one to be overridden, it guarantees that each class will have this behavior applied for free without having to remember to call it, and calling it should happen in every invocation of this IMO. If you feel strongly against this and prefer the explicit calling of a super method I can change it but I honestly find this way to be better where we apply default behavior by default rather than by remembering.

OrangeRed: I agree that its nice to get the functionality 'for free' via inheritance but in the future when someone who has never touched this code before tries to make a new class or extend a `PostAttack` child they might have a hard time understanding what is happening when something not explicitly invoked is being executed.

If you want to continue with that route at the very least add some jsdoc comments to the parent `applyPostAttack` and `applyPostAttackAfterMoveTypeCheck` functions that explain the above. Make sure to add a `/** {@linkcode applyPostAttack} */` so someone unfamiliar can jump to parent's implementation when they need.

drew9696: Thank you for your flexibility! Will do and I'll merge to resolve conflicts!

OrangeRed:
I think the changes here are good but its probably for the best that you try to create some tests that verify everything works.

src/data/ability.ts
```diff
@@ -1531,24 +1531,51 @@ export class BattleStatMultiplierAbAttr extends AbAttr {
...
+    this.stealCondition = stealCondition;
   }
 
-  applyPostAttack(pokemon: Pokemon, passive: boolean, defender: Pokemon, move: Move, hitResult: HitResult, args: any[]): Promise<boolean> {
+  applyPostAttackAfterMoveTypeCheck(pokemon: Pokemon, passive: boolean, defender: Pokemon, move: Move, hitResult: HitResult, args: any[]): Promise<boolean> {
     return new Promise<boolean>(resolve => {
-      if (hitResult < HitResult.NO_EFFECT && (!this.condition || this.condition(pokemon, defender, move))) {
-        const heldItems = this.getTargetHeldItems(defender).filter(i => i.isTransferrable);
+      if (hitResult < HitResult.NO_EFFECT && (!this.stealCondition || this.stealCondition(pokemon, defender, move))) {
+        const heldItems = this.getTargetHeldItems(defender).filter(i => i.getTransferrable(false));
```
Tempo-anon, line 1578:
```suggestion
        const heldItems = this.getTargetHeldItems(defender).filter(i => i.isTransferrable);
```

PR #1678, opened 2024-06-01
[Localization] Korean Font update

MrWaterT:
## What are the changes?
Korean font, `PokePT_Wansung.ttf`, has new version now.

## Why am I doing these changes?
First version of this font is something-hurry-up version.
There are some mistakes and bad shape glyphs, and I edited them.

(Full modified list is in the commit... [in my font repo](https://github.com/MrWaterT/PokePT_Wansung).
In 77 Korean glyph modifications, about 10 of them in below is used in pokerogue.)

## What did change?
At least 20 pokemon names look better, and more simillar to original game style.
(Of course moves, abilities and items have those glyphs)

Numbers have uniform width.

ASCII glyphs have more reasonable spacing.

### Screenshots/Videos
* Pokemon names in starter select ui

Left one is modified version name. Right one in old one.

image
푸호꼬 (Fennekin) : Separated ㅍ and ㅜ. No Skewered ㅎ

image
두두 (Doduo) : Separated ㄷ and ㅜ

image
테일로 (Taillow) : Separated ㅌ and ㅔ

image
동미러 (Bronzor) : Separated ㄹ and ㅓ

image
줄뮤마 (Blitzle) : Separated ㅈ and ㅜ

image
나룸퍼프 (Swirlix) : Separated ㅍ and ㅓ

image
꾸왁스 (Quaxly) : Distinguishable ㅇ and ㅘ

* Statistics ui

image
No more narrow 1.  
314 and 729 has same width.

왕(Blastoise), 즐 and 줄 can be in somewhere in the game (common glyphs).

## How to test the changes?
Set language to Korean, just play the game.
From daily ranking board, you can see more beautiful result (score is aligned properly).

## Checklist
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #1823, opened 2024-06-05
[Localization][Bug] Korean Font update and expand coverage

MrWaterT:
## What are the changes?
Large Korean texts (using `emerald`) become slightly smaller, like Alphabets.
Texts are more fit to UI.

Small Korean texts (using 'pkmnes') are no longer large-unreferenced-clipped glyphs.

It can be 2nd version of #1678.
This font file is generated for pokerogue, so it is also an unfinished product and has issues.

I want to make overall look better.

## What did change?
Update `PokePT_Wansung.ttf` to new version.
No change, only the glyph size shrinks (8/9).

In `i18n.ts`, one more *font-face* is loaded when locale is Korean.
Cuz of *unicodeRange*, ASCII range is not changed.
Only large texts are changed to pixel style font.

### Screenshots/Videos
Orange marks are drawn to help understanding.

* Smaller text
image
Left one is old, center is new. Right one is English version for comparison.
Text heights are same.

* Pixelated text
image
`pkmnems` font-family is now loaded. Left one is old.
You can also see ♀ and other word have same height.

* Clear pokemon names
image
From left to right, before loading font-face(old one), font-face loaded, and font file updated(got smaller).
In summary ui. Names are its right position, not just none-clipped.

* Modifier select UI
image
image
Upper one is old, Lower is new.
Item name is not overlapped by its price.

## How to test the changes?
`emerald` font is everywhere. ASAP turn on the game, result is shown.
`pkmnems` is more rare. Differences are in:
* `same` Daily run : "Score"
* center`diff`, corner`same` Reward after battle : All text except in the chatbox
* name`diff`, other`same` Pokemon Summary : Name, level, ♂♀, HP counter, cancel button in the corner.
* `diff` Pokemon Catching and starter start UI : Pokemon move list items.
* + `diff` starter select UI : Right side texts in the pokemon stat box.
* `same` Pokemon level in Save&Load UI
* `diff` Gacha UI : Result rarity text.

## Checklist
- [x] Are the changes visual?


MrWaterT:
9e2d305 seems something hacky... But gh-page does not show `alternativeFonts` in `i18n.ts` now and before.
Could this be a problem? Anyway assigning `size-adjust` after declaring works well.

MrWaterT:
567f66f fix werid text (cuz of `f`) display in Challenge select ui

(from)
image

(to)
image

returntoice (approved):

hibiyasleep:
I think original version of korean glyph looks more natural, when mixed with numbers.

image
Original version has same pixel sizes,

image
but it no longer.

how about keeping glyph sizes, and change interface font size instead?

MrWaterT:
2fcb20d: https://github.com/MrWaterT/PokePT_Wansung v2.51



> I think original version of korean glyph looks more natural, when mixed with numbers.
> 
> image Original version has same pixel sizes, (character heights may looks differ, but this mismatch is normal in east-asian typography)
> 
> image but it no longer, so '5' looks little bigger and bolder.
> 
> how about keeping glyph sizes, and change interface font size instead?

This problem has solved.

src/plugins/i18n.ts
```diff
@@ -89,8 +89,11 @@ export interface Localizable {
 }
 
 const fonts = [
-  new FontFace("emerald", "url(./fonts/PokePT_Wansung.ttf)"),
-  new FontFace("emerald", "url(./fonts/pokemon-emerald-pro.ttf"),
+  new FontFace("emerald", "url(./fonts/PokePT_Wansung.ttf)", { unicodeRange: "U+AC00-D7AC"}),
+  Object.assign(
```
flx-sta, line 93: You can use the spread operator instead

Tempo-anon (approved):

src/plugins/i18n.ts
MrWaterT: Can I use spread syntax with FontFace Interface?

PR #1828, opened 2024-06-05
[Documentation] documentation of move.ts

DustinLin:
## What are the changes?

Documentation

Documentation is good

## What did change?
Nothing, just comments

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -536,9 +536,9 @@ export default class Move implements Localizable {
 
   /**
    * Checks if the move flag applies to the pokemon(s) using/receiving the move
-   * @param flag {@linkcode MoveFlags}
-   * @param user {@linkcode Pokemon}
-   * @param target {@linkcode Pokemon}
+   * @param flag {@linkcode MoveFlags} MoveFlag to check on user and/or target
+   * @param user {@linkcode Pokemon} the Pokemon using the move
+   * @param target {@linkcode Pokemon} the Pokemon receiving the move
    * @returns boolean
    */
   checkFlag(flag: MoveFlags, user: Pokemon, target: Pokemon): boolean {
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -42,7 +42,7 @@ export enum MoveTarget {
   /** {@link https://bulbapedia.bulbagarden.net/wiki/Category:Moves_that_target_all_adjacent_Pok%C3%A9mon Moves that target all adjacent Pokemon} */
   ALL_NEAR_OTHERS,
   NEAR_ENEMY,
-  /** {@link https://bulbapedia.bulbagarden.net/wiki/Category:Moves_that_target_all_adjacent_foes Moves that taret all adjacent foes} */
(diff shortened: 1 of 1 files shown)


src/data/move.ts
```diff
@@ -166,10 +166,22 @@ export default class Move implements Localizable {
     return this.attrs.some((attr) => attr instanceof attrType);
   }
 
+  /**
+   * takes as input a boolean function and returns the first MoveAttr in attrs that matches true
```
EvasiveAce, line 170: nitpick: Starting the doc as capitalized since the rest seem to do as well

src/data/move.ts
```diff
@@ -334,42 +458,92 @@ export default class Move implements Localizable {
...
   checkAllHits(checkAllHits?: boolean): this {
     this.setFlag(MoveFlags.CHECK_ALL_HITS, checkAllHits);
     return this;
   }
 
+  /**
+   * Checks if the move flag applies to the pokemon(s) using/receiving the move
+   * @param flag a {@linkcode MoveFlags}
+   * @param user a {@linkcode Pokemon}
+   * @param target a {@linkcode Pokemon}
```
bennybroseph, line 542: For all of these `@param` comments, please start your sentence off with the `{@linkcode <TYPE>}` portion instead of having it randomly within the sentence. i.e.
```ts
/**
...
* @param flag {@linkcode MoveFlags} to check against
* @param user {@linkcode Pokemon} with this flag
* @param target {@linkcode Pokemon} target of the move
...
*/
```

Would need this done for all comments, not just this function

bennybroseph (changes requested):

src/data/move.ts
DustinLin: fixed!

src/data/move.ts
```diff
@@ -537,9 +536,9 @@ export default class Move implements Localizable {
 
   /**
    * Checks if the move flag applies to the pokemon(s) using/receiving the move
-   * @param flag a {@linkcode MoveFlags}
-   * @param user a {@linkcode Pokemon}
-   * @param target a {@linkcode Pokemon}
+   * @param flag {@linkcode MoveFlags}
+   * @param user {@linkcode Pokemon}
+   * @param target {@linkcode Pokemon}
```
bennybroseph, line 541: what's the point of these comments

bennybroseph (changes requested):

src/data/move.ts
DustinLin: Not sure if you mean the @param comments, or the whole:
- @param comments are for jsdocs? luckily the param names are pretty self explanatory given the description in the above comment (line 538). I can add a blurb about them, if you think they aren't needed I could remove them altogether.
- the whole comment block is for the function, figured `checkFlag` didn't have a jsdoc entry so wanted to add one for it

bennybroseph: there's no comments though

bennybroseph (approved):

PR #1833, opened 2024-06-05
[Feature] [Same species Egg] Egg class rewrite to enable fully parameterized eggs to generate same species eggs + Egg overrides

sirzento:
## What are the changes?

With these changes it is possible to parameterize eggs completely when creating them to make it possible to generate eggs with exact content such as the egg tier, the species or even whether it will be a shiny and which variant.

I have additionally added the option to buy an egg of the same species with candies. The exact costs would have to be revised again as I currently have them set to 10 for each level.

I have also created a few overrides to be able to test this better. These overrides could also be useful for other developers in the future.

At first, I wanted to make it possible to easily award shiny eggs in a certain variant rarity. This is useful for special occasions, events or even as a reward for future missions, gamemodes or other content.

But as I was already in the process of revising everything, I tidied everything up and added additional parameters. As I saw that it is now very easy to generate eggs of a certain species, I took over this part and added it to the candy menu selection. Since this is also an item on the roadmap, I think it's a good change.

Additionally, this could be used to increase the chances of getting rare or epic variants in the shiny gacha machine if desired. There is already a [thread about this in the feature-vote](https://discord.com/channels/1125469663833370665/1246442386050125865).

### Before:

The only properties in eggs where `id`, `tier`, `gachaType`, `hatchWaves` and `timestamp`. It was not possible to define anything else on egg creation. It is not possible to define a specific species to hatch from an egg.

### After:

It is now possible to parameterize eggs completely when creating them to make it possible to generate eggs with exact content such as the egg tier, the species or even whether it will be a shiny and which variant.

Following `eggOptions` are now available:

- `id` type `number`

   Sets the `id` of an egg.
- `timestamp` type `number`

   Sets the timestamp when the egg was created
- `pulled` type `boolean`

   Will define if the egg got pulled from a gacha or not. If true, egg pity and pull statistics will be applyed. The egg will also 
   be automaticly added to the game data. _Needs `scene`eggOption to work._
- `gachaType` type `GachaType`

   Defines the gacha machine type where this egg was pulled from. Used for shiny and species calculations.
- `tier` type `EggTier`

   Sets the tier of the egg. Only species of this tier can be hatched from this egg.
- `hatchWaves` type `number`

   Sets how many waves it will take till this egg hatches.
- `species` type `Species`

   Sets the exact species that will hatch from this egg. When a species is given, the egg tier will change accordingly. _Needs `scene`eggOption if **not** provided._
- `isShiny` type `boolean`

   Defines if the hatched pokemon will be a shiny.
- `variantTier` type `VariantTier`

  Defines the variant of the pokemon that will hatch from this egg. If no `variantTier` is given the normal variant rates will 
  apply. The `speciesPool` will be filtered when generating the species to only include pokemons with the given variant available to prevent wrong variants.

- `eggMoveIndex` type `number`

  Defines which egg move will be unlocked when this egg hatches. 3 = rare egg move.
- `overrideHiddenAbility` type `boolean`

  Defines if this egg will hatch with an hidden ability. If no hidden ability exist, a random one will get choosen.
- `scene` type  `BattleSccene`

   This need to be defined if `pulled` is defined or if no `species` or `isShiny` is defined since this will be needed to generate 
   those.

There is also a new menu option on the "candy shop" menu where it is now possible to buy an egg of the same species.

Override additions:

`EGG_TIER_OVERRIDE`: Overrides the tier of the egg **when the egg is created**.
`EGG_SHINY_OVERRIDE`: Overrides if the pokemon will be shiny **when the egg is created**.
`EGG_VARIANT_OVERRIDE`: Overrides the rarity of the shiny **when the egg is created**.
`EGG_FREE_GACHA_PULLS_OVERRIDE`: Overrides the gacha machiene cost. It's now free.
`EGG_GACHA_PULL_COUNT_OVERRIDE`: Overrides the amount of eggs pulled per use.
`EGG_IMMEDIATE_HATCH_OVERRIDE`: Renamed from `IMMEDIATE_HATCH_EGG_OVERRIDE` (Eggs hatch after 1 round).
`FREE_CANDY_UPGRADE_OVERRIDE`: Overrides the cost of purchase made with candy. It's now free.

The `egg-gacha-ui-handler.ts` is now free of egg generation logic since I rewrote and moved it into the `Egg` class where it should be. It's now even simpler to create eggs.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/16718523/e791da1d-be29-4e92-b168-f8e45d6107c3

## How to test the changes?

**`overrides.ts`:**
Set `EGG_IMMEDIATE_HATCH_OVERRIDE`, `EGG_FREE_GACHA_PULLS_OVERRIDE` and `FREE_CANDY_UPGRADE_OVERRIDE` to `true`.

Buy a "same species egg" and hatch it after one wave and verify if it was the same species.

**`ui/starter-select-ui-handler.ts`:**
Change anything in the `new Egg()` option parameter and buy an egg via candies. The egg will hatch after 1 wave and you can verify if the pokemon has the same properties as defined in the parameter. Don't forget to include the `scene` eggOption if you test for `isShiny` and `species`.
grafik

You can also pull some eggs from the gacha machine to verify that no eggs are broken.

I also added unit tests that you can run with `npm run test eggs`:

```
✓ should hatch an Arceus. Set from legendary gacha
✓ should hatch an Arceus. Set from species
✓ should return an common tier egg
✓ should return an rare tier egg
✓ should return an epic tier egg
✓ should return an legendary tier egg
✓ should return a manaphy egg set via species
✓ should return a manaphy egg set via id
✓ should return an egg with 1000 hatch waves
✓ should return an shiny pokemon
✓ should return a shiny common variant
✓ should return a shiny rare variant
✓ should return a shiny epic variant
✓ should return an egg with an egg move index of 0, 1, 2 or 3
✓ should return an egg with an rare egg move. Egg move index should be 3
✓ should return a hatched pokemon with a hidden ability
✓ should add the egg to the game data
✓ should override the egg tier to common
✓ should override the egg hatch waves
```

## How does it effect the current egg data?

When the game is loaded, the raw egg data is converted to the current egg class. When the legacy egg hatches, it will be filled with the missing properties. It is therefore fully compatible with the old egg data on the server.

## Points for which I still need feedback

- ~~I need your help with that. I tryed to do the translations with an translator but i am not sure if thats alright.~~
- ~~The cost of eggs for different levels of starters still needs to be adjusted. They are currently set to 10. What would be a good value?~~ Same as the passive alright!
- ~~Should the icon or animation also be displayed on the starter icon if you can afford an egg? Personally, I don't think it should do that.~~
- Should be shiny change be higher? The chance with normal eggs is 1/128 and I find that really low if you want to farm a shiny through this. Even 1/64 like the Shiny Up! gacha machine would be a bit too rare. I'm in favor of the chance being greater here. The only question is which chances would be a good balance. 1/32? 1/16?

**Any questions or do you want to discuss it? Just @sirzento me on discord in #dev-corner and we can talk :)**

## Checklist
- [x] Are the changes visual?


sirzento:
I think I will add onto this PR to make the egg even more flexible. My goal would be to clean up all the egg code such as removing the code responsible for generating the eggs from the gacha ui class because it just doesn't belong there. I would also rework the egg class to be able to fully parameterize the eggs.
This would make it possible to define all the properties of the eggs when they are created, such as which species they are. At the same time, I would also simply implement the roadmap point “Same species eggs buyable with candies” as this PR makes it trivial. I hope it's okay that I revise this and don't open a new PR as it builds on the changes here.

sirzento:
Should the icon or animation also be displayed on the starter icon if you can afford an egg? Personally, I don't think it should do that.

src/system/voucher.ts
```diff
@@ -81,6 +82,17 @@ export function getVoucherTypeIcon(voucherType: VoucherType): string {
   }
 }
 
+export function getGuaranteedEggTierFromPullCount(pullCount: number): EggTier {
```
brain-frog, line 85: why is this defined here if only used in the egg gacha ui handler

brain-frog:
> Should the icon or animation also be displayed on the starter icon if you can afford an egg? Personally, I don't think it should do that.

Probably not, we need to make sure the ui isnt getting too cluttered

> Should I just play the "buy" sounds when buying an egg or should I add an summary container to display it the same way as if you pulled a single egg from a gacha machine?

we can discuss the UI for this, buy sound is a good idea. I dont think we need a summary display

src/overrides.ts
```diff
@@ -35,9 +37,9 @@ export const STARTING_BIOME_OVERRIDE: Biome = Biome.TOWN;
 export const ARENA_TINT_OVERRIDE: TimeOfDay = null;
 // Multiplies XP gained by this value including 0. Set to null to ignore the override
 export const XP_MULTIPLIER_OVERRIDE: number = null;
-export const IMMEDIATE_HATCH_EGGS_OVERRIDE: boolean = false;
 // default 1000
 export const STARTING_MONEY_OVERRIDE: integer = 0;
+export const STARTER_CANDY_FREE_NO_COST_OVERRIDE: boolean = false;
```
brain-frog, line 42: can this be renamed to something clearer
`FREE_CANDY_UPGRADE_OVERRIDE` or something

src/system/voucher.ts
sirzento: I wasn't sure about this myself. I just thought that I don't want it inside the gacha ui handler since that function has nothing to do with gacha ui. Where would be a better place for this? I just thought it made sense in there because it will get the guaranteed egg from a specific voucher/pull count.

sirzento:
> Probably not, we need to make sure the ui isnt getting too cluttered

> we can discuss the UI for this, buy sound is a good idea. I dont think we need a summary display

Perfect. It's like that in the code.

We also need to handle the points I mention in "Points for which I still need feedback" at the bottom of the post but those are more a balancing aspect and since I never grinded in endless I have no clue what a good price would be and how much the shiny chance should be raised.

src/overrides.ts
sirzento: Ok yes that's way better. I will do that if I got time later

src/overrides.ts
```diff
@@ -98,6 +100,17 @@ export const OPP_SHINY_OVERRIDE: boolean = false;
...
 export const OPP_IVS_OVERRIDE: integer | integer[] = [];
 
+/**
+ * EGG OVERRIDES
+ */
+
+export const EGG_IMMEDIATE_HATCH_OVERRIDE: boolean = false;
+export const EGG_TIER_OVERRIDE: EggTier = EggTier.COMMON;
+export const EGG_SHINY_OVERRIDE: boolean = false;
+export const EGG_VARIANT_OVERRIDE: VariantTier = VariantTier.COMMON;
```
brain-frog, line 110: because of the nature of the enums it seems like its not possible to override egg tier and variant tier to be common since they will just read as 0 / false, which makes these more confusing for non-developers to look at
can you make these a union type of `Egg/VariantTier | boolean` and have the override typecheck before use? or use an active property like the pokeball override?

src/system/voucher.ts
brain-frog: a lot of the functionality needs to be removed from the ui handler but that doesnt need to be handled during this change, for now keeping it all in one place would be better for later refactors
you can move this into the egg gacha ui handler so it doesnt need to be needlessly exported/imported

src/ui/starter-select-ui-handler.ts
```diff
@@ -98,17 +100,17 @@ const languageSettings: { [key: string]: LanguageSetting } = {
...
+  { passive: 50, costReduction: [30, 75], egg: 50 }, // 1
+  { passive: 45, costReduction: [25, 60], egg: 45 }, // 2
+  { passive: 40, costReduction: [20, 50], egg: 40 }, // 3
+  { passive: 30, costReduction: [15, 40], egg: 30 }, // 4
+  { passive: 25, costReduction: [12, 35], egg: 25 }, // 5
+  { passive: 20, costReduction: [10, 30], egg: 20 }, // 6
+  { passive: 15, costReduction: [8, 20], egg: 15 },  // 7
+  { passive: 10, costReduction: [5, 15], egg: 10 },  // 8
+  { passive: 10, costReduction: [3, 10], egg: 10 },  // 9
+  { passive: 10, costReduction: [3, 10], egg: 0 },  // 10
```
brain-frog, line 113: lol 10 cost pokemon aren't getting free eggs
for now you can make it 10 like the passive

sirzento: Only eternatus is 10 cost pokemon and I have a check that it won't show the option to buy it if the egg cost is 0. But now that I think about it, it's actually better if I just check if the Pokemon is eternatus.
Or should it be possible to buy eternatus eggs?

brain-frog: it should be possible to buy eternatus eggs
if someone has unlocked eternatus, that's the only way to get its egg moves

src/system/voucher.ts
sirzento: Ok I moved the fuction back to the gacha ui handler

src/overrides.ts
sirzento: I just set the dafault value of those to `null` and changed some logic in the `egg` class so the override gets used if its not `null`

src/ui/starter-select-ui-handler.ts
sirzento: Ok it's buyable again and cost back up to 10.

brain-frog (approved):

PR #1280, opened 2024-05-23
Follow Me, Rage Powder, Spotlight [WIP]

mareksison:
Calling these 5 moves. Will be working on these 5 concurrently

PR #1834, opened 2024-06-05
[Move] Implement Follow Me, Rage Powder, and Spotlight

innerthunder:
## What are the changes?

This implements the [Center of Attention](https://bulbapedia.bulbagarden.net/wiki/Center_of_attention) volatile status effect as well as all of the moves that apply it.

The moves Follow Me, Rage Powder, and Spotlight are currently unimplemented. [Another PR](https://github.com/pagefaultgames/pokerogue/pull/1280) started work on these moves but now seems to be inactive.

## What did change?

- Follow Me, Rage Powder, and Spotlight add a new battler tag `CenterOfAttentionTag` to the target. Only one Pokemon on each side of the playfield can have this tag at a time, and the tag cannot be overwritten until it expires the next turn.
- `src/phases`: When a Pokemon targets an enemy with a move, it checks the opposing side of the field for the Center of Attention tag. If a tag is found, the move's target is changed to the index of the tag's owner. The original target may be restored if the user has an ability or move that ignores redirection.
- Rage Powder is partially implemented with this change since it is not ignored by Grass types or Pokemon with the ability Overcoat.

### Screenshots/Videos

Video includes:

- Using Follow Me, Rage Powder, and Spotlight in battle
- Interaction when both allies use "center of attention" moves

https://github.com/pagefaultgames/pokerogue/assets/168692175/21afc7a1-3d1e-4bc3-9f7f-fc7b54449825

## Checklist
- [x] Are the changes visual?


src/phases.ts
```diff
@@ -2568,6 +2568,11 @@ export class MovePhase extends BattlePhase {
     if (moveTarget) {
       const oldTarget = moveTarget.value;
       this.scene.getField(true).filter(p => p !== this.pokemon).forEach(p => applyAbAttrs(RedirectMoveAbAttr, p, null, this.move.moveId, moveTarget));
+      this.pokemon.getOpponents().forEach(p => {
```
Tempo-anon, line 2571: Grass types and pokemon with Overcoat will ignore this forced targeting if it was set with rage powder

Tempo-anon (commented):
Does snipe shot, propeller tail, and stalwart ignore the CenterOfAttentionTag?

innerthunder:
> Does snipe shot, propeller tail, and stalwart ignore the CenterOfAttentionTag?

Yes, in the same way as abilities like Storm Drain currently work. The original target is stored in a variable before possible redirection, and if the user has a move or ability that bypasses redirection, the original target is restored.

Here's the logic for redirection from `MovePhase` in `phases.ts`, for reference:
```ts
// Move redirection abilities (ie. Storm Drain) only support single target moves
  const moveTarget = this.targets.length === 1
    ? new Utils.IntegerHolder(this.targets[0])
    : null;
  if (moveTarget) {
    // Store the original target in case it needs to be restored
    const oldTarget = moveTarget.value;
    // Apply redirection from abilities
    this.scene.getField(true).filter(p => p !== this.pokemon).forEach(p => applyAbAttrs(RedirectMoveAbAttr, p, null, this.move.moveId, moveTarget));
    // Check target's side of field for Center of Attention tags
    this.pokemon.getOpponents().forEach(p => {
      const redirectTag = p.getTag(CenterOfAttentionTag) as CenterOfAttentionTag;
      if (redirectTag && (!redirectTag.powder || (!this.pokemon.isOfType(Type.GRASS) && !this.pokemon.hasAbility(Abilities.OVERCOAT)))) {
        moveTarget.value = p.getBattlerIndex();
      }
    });
    //Check if this move is immune to being redirected, and restore its target to the intended target if it is.
    if ((this.pokemon.hasAbilityWithAttr(BlockRedirectAbAttr) || this.move.getMove().hasAttr(BypassRedirectAttr))) {
      //If an ability prevented this move from being redirected, display its ability pop up.
      if ((this.pokemon.hasAbilityWithAttr(BlockRedirectAbAttr) && !this.move.getMove().hasAttr(BypassRedirectAttr)) && oldTarget !== moveTarget.value) {
        this.scene.unshiftPhase(new ShowAbilityPhase(this.scene, this.pokemon.getBattlerIndex(), this.pokemon.getPassiveAbility().hasAttr(BlockRedirectAbAttr)));
      }
      moveTarget.value = oldTarget;
	  }
    this.targets[0] = moveTarget.value;
  }
```

Tempo-anon (approved):

PR #1836, opened 2024-06-05
[Balance] Dark Void 80% Accurate

damocleas:
## What are the changes?
Dark Void from 50% -> 80% accuracy like in Gen 6.

Dark Void was nerfed in gen 7 due to pvp. This is looking to undo that.

## What did change?
A 50 to an 80 for accuracy.

### Screenshots/Videos
image
image

## How to test the changes?
Give something Dark Void, check accuracy.

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -6404,7 +6404,7 @@ export function initMoves() {
       .attr(OpponentHighHpPowerAttr),
     new AttackMove(Moves.MAGMA_STORM, Type.FIRE, MoveCategory.SPECIAL, 100, 75, 5, 100, 0, 4)
       .attr(TrapAttr, BattlerTagType.MAGMA_STORM),
-    new StatusMove(Moves.DARK_VOID, Type.DARK, 50, 10, -1, 0, 4)
+    new StatusMove(Moves.DARK_VOID, Type.DARK, 80, 10, -1, 0, 4)
       .attr(StatusEffectAttr, StatusEffect.SLEEP)
       .target(MoveTarget.ALL_NEAR_ENEMIES),
     new AttackMove(Moves.SEED_FLARE, Type.GRASS, MoveCategory.SPECIAL, 120, 85, 5, 40, 0, 4)
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -6404,7 +6404,7 @@ export function initMoves() {
       .attr(OpponentHighHpPowerAttr),
     new AttackMove(Moves.MAGMA_STORM, Type.FIRE, MoveCategory.SPECIAL, 100, 75, 5, 100, 0, 4)
       .attr(TrapAttr, BattlerTagType.MAGMA_STORM),
-    new StatusMove(Moves.DARK_VOID, Type.DARK, 80, 10, -1, 0, 4)
+    new StatusMove(Moves.DARK_VOID, Type.DARK, 80, 10, -1, 0, 4)  //Accuracy from Generations 4-6
       .attr(StatusEffectAttr, StatusEffect.SLEEP)
       .target(MoveTarget.ALL_NEAR_ENEMIES),
     new AttackMove(Moves.SEED_FLARE, Type.GRASS, MoveCategory.SPECIAL, 120, 85, 5, 40, 0, 4)


EvasiveAce:
I'm not on the balance team but personally I think if a change happens between games, the newest should be upheld.

Also: why not just do Legends Arceus?
In [Pokémon Legends: Arceus](https://bulbapedia.bulbagarden.net/wiki/Pok%C3%A9mon_Legends:_Arceus), Dark Void's accuracy was raised to 90%, and lowers the user's action priority. It makes the target drowsy and lowers its defensive stats. Like all other moves in this game, it targets only one Pokémon.

CodeTappert:
Since Gen 7 its 50% per Pokemon. So currently it is accurate. https://bulbapedia.bulbagarden.net/wiki/Dark_Void_(move)

damocleas:
> I'm not on the balance team but personally I think if a change happens between games, the newest should be upheld.
> 
> Also: why not just do Legends Arceus? In [Pokémon Legends: Arceus](https://bulbapedia.bulbagarden.net/wiki/Pok%C3%A9mon_Legends:_Arceus), Dark Void's accuracy was raised to 90%, and lowers the user's action priority. It makes the target drowsy and lowers its defensive stats. Like all other moves in this game, it targets only one Pokémon.

90% spread sleep even if on a limited pokemon is not very balanced even if it is the more recent version of a higher accuracy..

Fontbane:
I agree with this. Dark Void was nerfed due to Smeargle in the same generation Smeargle was made unable to use it.

CodeTappert:
> I agree with this change. Dark Void was nerfed due to Smeargle in the same generation Smeargle was made unable to use it.

And never unnerfed

Tempo-anon:
If this ever gets in can we save it for a doubles update (alongside other doubles implementations like commander and updated double champion/gym leaders?

damocleas:
> If this ever gets in can we save it for a doubles update (alongside other doubles implementations like commander and updated double champion/gym leaders?

why would we delay a simple change for something without a time table?

Mystery118:
> I'm not on the balance team but personally I think if a change happens between games, the newest should be upheld.
> 
> Also: why not just do Legends Arceus? In [Pokémon Legends: Arceus](https://bulbapedia.bulbagarden.net/wiki/Pok%C3%A9mon_Legends:_Arceus), Dark Void's accuracy was raised to 90%, and lowers the user's action priority. It makes the target drowsy and lowers its defensive stats. Like all other moves in this game, it targets only one Pokémon.

Newest does not equal better. in fact we shouldn't be attempted to uphold godawful changes made by game freak in regards to "balancing"

ben-lear:
Should we consider changing Darkrai’s Spore egg move to something else alongside these changes?

Mystery118:
> Should we consider changing Darkrai’s Spore egg move to something else alongside these changes?

We probably should, since Spore would be redundant, not to mention not making sense for darkrai to have. I would personally replace it with something like shadow force or some other similar move

flx-sta (approved):

src/data/move.ts
```diff
@@ -6404,7 +6404,7 @@ export function initMoves() {
       .attr(OpponentHighHpPowerAttr),
     new AttackMove(Moves.MAGMA_STORM, Type.FIRE, MoveCategory.SPECIAL, 100, 75, 5, 100, 0, 4)
       .attr(TrapAttr, BattlerTagType.MAGMA_STORM),
-    new StatusMove(Moves.DARK_VOID, Type.DARK, 50, 10, -1, 0, 4)
+    new StatusMove(Moves.DARK_VOID, Type.DARK, 80, 10, -1, 0, 4)
```
frutescens, line 6407:
```suggestion
    new StatusMove(Moves.DARK_VOID, Type.DARK, 80, 10, -1, 0, 4) //Accuracy from Gen. 6
```

frutescens (approved):
Using the comment is optional.

innerthunder (approved):
Code looks good to me. Is Darkrai currently the only Pokemon that can use this move? If so, should that be changed as well?

frutescens:
@innerthunder Looking at data/move.ts, there's no attribute that makes the move exclusive to Darkrai.

damocleas:
> Code looks good to me. Is Darkrai currently the only Pokemon that can use this move? If so, should that be changed as well?

Fusions make 'pokemon locked' moves just not work, and we've given it out as an egg move already, I don't see the issue with keeping it as is. I can add to the comment though, but we have 0 pokemon locked moves in this game as of right now thanks to fusion so I'm not looking to get that changed.

DayKev (approved):

Issue #1817, opened 2024-06-05
[Feature] Add Day/Night cycle icon


Odizinne:
**Is your feature request related to a problem? Please describe.**

It can be hard to determine day / night cycle in certain biome.

**Describe the Feature**

Having an indicator, probably right below money would be beneficial, and would not harm visibility

**Additional context**

Could be 4 different icons representing dusk, dawn, day, night

PR #1846, opened 2024-06-05
[QoL] Add Time of Day Widget to Arena Flyout

bennybroseph:
## What are the changes?

This change adds a time of day indicator in the Arena Flyout. This is an icon so no translation is needed. Settings are provided to enable/disable the feature as well as two different tween animations as seen below.

Time of day is an important variable for many evolutions and may involve more mechanics in the future. However, it is hard to tell exactly what time of day it currently is. This removes ambiguity is the player chooses.

## What did change?

I've added several images to be rendered in a layer which creates a neat little animation upon changing time of day. This animation will wait to play when the flyout is visible if needed. 

### Screenshots/Videos

Back All
Bounce All Slow

## How to test the changes?

Load into a game with a high level pokemon in endless and 0-KO Pokemon until a time of day change occurs. 

## Checklist
- [x] Are the changes visual?


Tempo-anon (commented):
Love the animation, especially the little bounce. For QOL is there a way to add how many floors until the next time? From the wiki:
> The time of day cycles with a loop of 40 waves: It is Day for 15 waves, Dusk for 5 waves, Night for 15 waves, then Dawn for 5 waves.
At the start of a run, the time of day is randomly set to a multiple of 5 in the loop (from 0 to 35, inclusive). Thus, the first time change will occur on wave 6, 11, or 16.
It is always Night in the Abyss.

bennybroseph:
I can for a v2 of this. Adding a tooltip to it which gets that specific sounds like a different feature than this was intended to be

Tempo-anon (approved):

Odizinne:
This is neat. You can then autoclose #1817

PR #1847, opened 2024-06-06
dont make api calls when no server is connected in local

brain-frog:
## What are the changes?

Api calls won't be attempted if a local server isn't connected

I'm tired of seeing errors for api calls when I'm well aware my local game is not connected to the server

## What did change?

checks during the loading sequence if a local server is connected
(also setting fusionLuck as an additional fix to the earlier issue) 

## How to test the changes?

Run the game `npm run start:dev` without running the rogueserver locally and see if errors come up 

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -211,8 +211,8 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
           this.generateFusionSpecies();
         }
       }
-
       this.luck = (this.shiny ? this.variant + 1 : 0) + (this.fusionShiny ? this.fusionVariant + 1 : 0);
+      this.fusionLuck = this.luck;
     }
 
     this.generateName();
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -285,8 +286,17 @@ export function getCookie(cName: string): string {
   return "";
 }
 
+export function localPing() {
```
Tempo-anon, line 289: is this the best name for this? Also if you could add some doc to new functions added it would be great even if it's a QOL

src/field/pokemon.ts
```diff
@@ -211,8 +211,8 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
           this.generateFusionSpecies();
         }
       }
-
       this.luck = (this.shiny ? this.variant + 1 : 0) + (this.fusionShiny ? this.fusionVariant + 1 : 0);
+      this.fusionLuck = this.luck;
```
Tempo-anon, line 215: Does this fix the issue with fused mons having undefined luck?

Tempo-anon (dismissed):

src/field/pokemon.ts
brain-frog: yup!

src/utils.ts
brain-frog: I'll add some documentation but to me it makes sense 
its pinging the server to see if its alive when you're in local

Tempo-anon (approved):
When merging, please make the commit title reflect that it also fixes the fusion luck bug

PR #1859, opened 2024-06-06
[QoL] Show currently bound controls on starter menu for last active device

SquillWall:
## What are the changes?
The starter screen will now show the appropriate keybinds based on the input device you are using, rather than a static string

I play mostly on mobile with controller, but also on PC with controller on occasion. Prior to these changes, there was no immediate way to know (from a UX perspective) which buttons on the controller corresponded to which toggles in the starter screen. This PR aims to fix that by polling input from the last device and dynamically updating the control icons to reflect it.

## What did change?
There was a significant change to starter-select-ui-handler::updateInstructions() in order to accomodate the new UI changes

### Screenshots/Videos
Prior to my changes: image

After my changes (keyboard example): 
image

After my changes (controller example):
image

## How to test the changes?
- Select new game
- ensure you have a starter with multiple genders/natures/abilities unlocked
- connect an external controller to your device OR simply use your keyboard
- navigate the menu with one or both options to see the icons dynamically update based on your preferred input
- rebinds are also supported and should update the icons

## Checklist
- [x] Are the changes visual?

image
All unit tests passed


src/ui-inputs.ts
```diff
@@ -29,8 +29,23 @@ export class UiInputs {
     this.listenInputs();
   }
 
+  detectInputMethod(evt): void {
```
Greenlamp2, line 32: Not needed, you can already know which is the last inputMethod with the "lastSource" property in input-controller

src/battle-scene.ts
```diff
@@ -208,6 +208,8 @@ export default class BattleScene extends SceneBase {
   public rngSeedOverride: string = "";
   public rngOffset: integer = 0;
 
+  public inputMethod: string;
```
Greenlamp2, line 211: not needed, same property as lastSource in input-controller

src/ui/starter-select-ui-handler.ts
```diff
@@ -1434,45 +1441,55 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
...
-    const cycleInstructionLines = [];
+    const currentLanguage = i18next.resolvedLanguage;
+    const langSettingKey = Object.keys(languageSettings).find(lang => currentLanguage.includes(lang));
+    const textSettings = languageSettings[langSettingKey];
+    const instructionTextSize = textSettings.instructionTextSize;
+    this.instructionRowX = 0;
+    this.instructionRowY = 0;
+    this.instructionsContainer.removeAll();
+    let gamepadType;
+    if (this.scene.inputMethod === "gamepad") {
```
Greenlamp2, line 1467: not needed, you just need to update the texture (.setTexture) to the lastSource to update correctly the icons displayed

Greenlamp2 (changes requested):

SquillWall:
@Greenlamp2 thanks for the feedback! I'll resolve those issues when i'm off work tonight

src/battle-scene.ts
SquillWall: Unfortunately as far as i can tell, both this and detectInputMethod are needed... lastSource only detects whether the last input was from a keyboard or a controller, and since there's no distinction between the touch controls and an actual keyboard, i had to make one myself. This becomes an issue when you rebind a key on the keyboard, as you would then have a discrepancy between the static touch controls and the new icon showing from the rebound key.

src/ui/starter-select-ui-handler.ts
SquillWall: This is also a bit more difficult since the elements are being dynamically populated into the container based on the types caught in the dex entry... If it would be more performant i could iterate over the list of sprites and update their textures but it will definitely add a lot more edge cases to account for than simply clearing the container and repopulating it with the appropriate objects. Let me know what you think would be best i guess

src/ui-inputs.ts
SquillWall: Same as below

src/battle-scene.ts
Greenlamp2: good point, but i was working on syncing keys displayed on touch with the keys bound on keyboard, (it's in one of my stash, but it has not been discussed with other dev yet if it's something we really want) until then, ok to leave it like that, but when/if synced keys are merged, i count on you to update your work

src/ui/starter-select-ui-handler.ts
Greenlamp2: Yeah, I understand. Indeed, the gain is minimal. If it can avoid adding unexpected issue, that's fine with me.

src/battle-scene.ts
SquillWall: Yeah definitely can do just give me a shout when/if those changes are merged in

SquillWall:
Merge conflicts are resolved so assuming this is ready for merge unless you had any other feedback for me to take care of greenlamp

Greenlamp2 (dismissed):

SquillWall:
NOTE BEFORE MERGE: i need to update locales other than english

SquillWall:
sorry to ping you again @Greenlamp2 nothing changed functionally here, was just updating the control strings for other locales as well

Greenlamp2 (approved):

Greenlamp2:
It's good for me, i ask the input from the other dev, this is the final step, if they are all okay, i'll merge it, otherwise, i'll let you know what need to be changed.

bennybroseph:
Does this hide text when that cycle key isn't available? For example since Sentret doesn't have other forms, the form button should be hidden.

SquillWall:
@bennybroseph it does! The buttons will hide and show dynamically the same way that they did before (based on what you've caught/unlocked). Here's an example of my spearow for example from my local save that only has gender and nature to toggle
image

Greenlamp2:
image

some elipsis might be needed there

PR #1889, opened 2024-06-06
[QoL] Add API mocking for local development and future testing

flx-sta:
> It's deactivated/off by default

## What are the changes?
I've added the Mock Service Worker library which allows to mock server requests

It allows users to "run" a server without having to launch the actual on locally. 
In the future it is also very useful to automated testing as you have full control what the response of a request is going to be.

## What did change?
The main.ts now checks if a mocked api is wanted and if yes launches the required worker of msw.

### Screenshots/Videos
here is a video (it will be downloaded):
https://cdn.discordapp.com/attachments/1176874654015684739/1248349468110229646/Screen_Recording_2024-06-06_at_11.54.00_AM.mov?ex=6663578a&is=6662060a&hm=9475d5d4ed6cea21d7d68bf4d36c4c6e07c497e7ce6318805342d7f4513a7bd0&

or the discord message:
https://discord.com/channels/1125469663833370665/1176874654015684739/1248349469230235818

## How to test the changes?
1. check out the branch
2. make sure to disable the login bypass in the `.env.development` file
3. Make sure to activate `API_MOCKING`
4. Launch in dev mode
5. play around with the login/register and you should also see a scoreboard and stuff.

## Checklist


src/system/game-data.ts
```diff
@@ -84,7 +84,7 @@ export function getDataTypeKey(dataType: GameDataType, slotId: integer = 0): str
 export function encrypt(data: string, bypassLogin: boolean): string {
   return (bypassLogin
     ? (data: string) => btoa(data)
-    : (data: string) => AES.encrypt(data, saveKey))(data) as unknown as string; // TODO: is this correct?
+    : (data: string) => AES.encrypt(data, saveKey))(data) as string as unknown as string;
```
DayKev, line 87: What does chaining `as` statements actually do?

flx-sta: This one is just nonsense I think 😂 but for e.g. the compiler sometimes only allows you to cast it to a type if you  cast it to “unknown” first. It’s basically a trick

PR #1193, opened 2024-05-21
Feature: Finish Magnet Pull and clauses for Arena Trap

td76099:
@Piefrenzy on Discord

As per Bulbapedia, [Arena Trap](https://bulbapedia.bulbagarden.net/wiki/Arena_Trap_(Ability)#In_battle) does not affect Pokemon that are not [grounded](https://bulbapedia.bulbagarden.net/wiki/Grounded), and [Magnet Pull](https://bulbapedia.bulbagarden.net/wiki/Magnet_Pull_(Ability)#In_battle) only affects Steel type Pokemon and none others. Clauses have been added to the `applyCheckTrapped` function to facilitate this.

I did see `.condition((pokemon: Pokemon) => pokemon.getOpponent()?.isOfType(Type.STEEL))` commented out under Magnet Pull. I was working on trying to implement something easy like that, but there didn't seem to be a way to do. Another way of thinking about it is these abilities are technically always active, but it requires opponent Pokemon to meet certain conditions to be affected by them.

Ghost type not blocked by Shadow Tag

https://github.com/pagefaultgames/pokerogue/assets/85713900/18cf582a-105d-43ee-aa16-bbb9795727af

Magnet Pull trapping a Steel type

https://github.com/pagefaultgames/pokerogue/assets/85713900/aa62c8bb-1143-4224-9bce-ac2587fe5d75

Pokemon with the Levitate ability not getting blocked by Arena Trap

https://github.com/pagefaultgames/pokerogue/assets/85713900/f8b88485-0a3a-4157-8256-0b5ab8757f7d

Flying type not getting blocked by Arena Trap

https://github.com/pagefaultgames/pokerogue/assets/85713900/a9632942-0e10-4669-9c2b-ab21dae10d10

PR #1890, opened 2024-06-06
[Bug] Fix run away ability

SeafoamQueen:
## What are the changes?
Pokemon with the ability Run Away are not able to flee from battles where an opposing pokemon has a trapping ability.

A Pokémon with the Run Away ability can always flee and use Teleport successfully, regardless of trapping moves and abilities.
- https://bulbapedia.bulbagarden.net/wiki/Run_Away_(Ability)#Effect

## What did change?
`ArenaTrapAbAttr` in `ability.ts` was modified to add a check for the ability `Run_Away`.

### Screenshots/Videos
The attached video shows a pokemon with the ability Run Away successfully running from pokemon with the ability Arena Trap and the passive Shadow Tag.

https://github.com/pagefaultgames/pokerogue/assets/167576411/c8c3981d-3c33-4cf1-a696-f2e66d8d9af9

## How to test the changes?
Start a new game with a pokemon with the ability Run Away.
Set the following overrides in `overrides.ts` :
- `OPP_ABILITY_OVERRIDE` to `Abilities.ARENA_TRAP` or `Abilities.SHADOW_TAG`

Using each pokemon attempt to run from battles.
You should successfully escape everytime, unless you are in a trainer battle in Classic or an x50 battle in Endless.

## Checklist
- [ ] Are the changes visual?


DecaP0d:
As I've said before, I'm heavily against the implementation of Ghost types being given built-in Run Away. This might be something that works for the main series games, but I feel is harmful for Pokérogue. The choice to use a Run Away pokémon should come at a cost due to its potential to bypass fights that could end your run, and most pokémon that have the ability are either weak or have to give up an ability slot for it. For future modes that pose any challenge, this change will stop any Run Away passives from being added and will create a bad dynamic where players will want to always have a ghost type on their team. Please reconsider this! The change to how Run Away interacts with Shadow Tag and Arena Trap is great though.

src/phases.ts
```diff
@@ -2199,7 +2199,16 @@ export class TurnStartPhase extends FieldPhase {
             const fasterPokemon = playerActivePokemon[0].getStat(Stat.SPD) > playerActivePokemon[1].getStat(Stat.SPD) ? playerActivePokemon[0] : playerActivePokemon[1];
             // check if either active pokemon has the ability "Run Away"
             const hasRunAway = playerActivePokemon.find(p => p.hasAbility(Abilities.RUN_AWAY));
-            runningPokemon = hasRunAway !== undefined ? hasRunAway : fasterPokemon;
+            // check if either active pokemon has the Ghost type
+            const hasGhostType = playerActivePokemon.find(p => p.getTypes(true).includes(Type.GHOST) || (p.getTypes(true).includes(Type.STELLAR) && p.getTypes().includes(Type.GHOST)));
```
brain-frog, line 2203: is the stellar and ghost case necessary? won't getting ghost on its own cover that?

SeafoamQueen: I thought the same thing, but did it this way based on the way this PR to finish Magnet Pull and Arena Trap handled ghost types.
https://github.com/pagefaultgames/pokerogue/pull/1193

SeafoamQueen:
@brain-frog The ghost type mechanics have been removed from the PR. It should be fine to merge now.

SeafoamQueen:
> What about `Magnet Pull` ?

Magnet Pull shouldn't trap pokemon with the ability Run Away either, and it uses the same `ArenaTrapAbAttr` property that the other trapping abilities use to determine if a pokemon is trapped by the ability.
I've modified that property to check for the ability Run Away, and not trap the pokemon if it has it.

brain-frog (approved):

PR #1893, opened 2024-06-06
[Bug] Show C-Button on Mobile during Modifier Selection

prime-dialga:
## What are the changes?
The C-Button is now shown on Mobule when selecting a modifier.

The C-Button is required to access additional information for the move a TM teaches.

## What did change?
Adjusted CSS Rule responsible for hiding the button during modifier selection to show the button.

### Screenshots/Videos
Before:
IMG_0247

Now a C Button appears in the bottom right corner, similar to the combat menu.

## How to test the changes?
Enter a fight, end the fight, the button should still be present.

Though, to be more specific:
Change the following override to get a TM as a reward.
`export const ITEM_REWARD_OVERRIDE: Array<String> = [];`
with
`export const ITEM_REWARD_OVERRIDE: Array<String> = ["TM_COMMON"];`
When hovering over the TM, before selecting it, you should be able to press C to view additional information.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/index.css
+++ b/index.css
@@ -152,7 +152,7 @@ body {
   display: none;
 }
 
-#touchControls:not([data-ui-mode='COMMAND']):not([data-ui-mode='FIGHT']):not([data-ui-mode='BALL']):not([data-ui-mode='TARGET_SELECT']) #apad .apadBattle {
+#touchControls:not([data-ui-mode='COMMAND']):not([data-ui-mode='FIGHT']):not([data-ui-mode='BALL']):not([data-ui-mode='TARGET_SELECT']):not([data-ui-mode='MODIFIER_SELECT']) #apad .apadBattle {
   display: none;
 }
 


flx-sta:
lgtm ✅ 
image

index.css
```diff
@@ -152,7 +152,7 @@ body {
   display: none;
 }
 
-#touchControls:not([data-ui-mode='COMMAND']):not([data-ui-mode='FIGHT']):not([data-ui-mode='BALL']):not([data-ui-mode='TARGET_SELECT']) #apad .apadBattle {
+#touchControls:not([data-ui-mode='COMMAND']):not([data-ui-mode='FIGHT']):not([data-ui-mode='BALL']):not([data-ui-mode='TARGET_SELECT']):not([data-ui-mode='MODIFIER_SELECT']) #apad .apadBattle {
```
flx-sta, line 155: since CSS4 not can be used with multiple params to chain them 

```suggestion
#touchControls:not([data-ui-mode='COMMAND'], [data-ui-mode='FIGHT'], [data-ui-mode='BALL'], [data-ui-mode='TARGET_SELECT'], data-ui-mode='MODIFIER_SELECT']) #apad .apadBattle {
```
**untested!**

source: https://developer.mozilla.org/en-US/docs/Web/CSS/:not?ref=serpapi

flx-sta (changes requested):

index.css
prime-dialga: Tested on FF and it works; however: Opera (Android) doesn't seem to support that function, according to your reference link.
So it might not be worth it to improve the code style...

flx-sta: that is true. Not worth it then. MB

bennybroseph (approved):

PR #1696, opened 2024-06-01
[Localization] Locales punctuation fix in descriptions in (en/es/de/it/fr/ko/pt/zh) modifier-type.ts, and other German locales punctuation fixes

InnocentGameDev:
## Why am I doing these changes?
Ok, so I'll start with this considering the nature of the change. I made some other localization changes in another PR and was getting suggestions that included a "." at the end of the description of items. 95% of English item descriptions didn't end with punctuation and almost all of the Spanish, Italian and German item descriptions didn't have punctuation either. French was at 50/50 split.

So two goals are achieved here:
1- Consistency across the board.
2- Said consistency replicates the S/V item descriptions.

## What are the changes?
For now, not all modifier-type.ts are changed, but I wanted to make the PR already in case anyone is working on this as well.

- [X] English punctuation fix in modifier-type.ts
- [X] Spanish punctuation fix in modifier-type.ts
- [x] German punctuation fix in modifier-type.ts and other places too, thanks @CodeTappert 
- [x] French punctuation fix in modifier-type.ts
- [x] Italian punctuation fix in modifier-type.ts
- [x] PT_BR punctuation fix in modifier-type.ts
- [x] Korean punctuation fix in modifier-type.ts and Berry Pouch locale fix, thanks @MrWaterT 
- [x] Chinese_CN punctuation fix
- [x] Chinese_TW punctuation fix

## What did change?
Descriptions in locales/lang/modifier-type.ts got proper punctuation when applicable.

The commit also contains toxic/flame orb description in English being fixed, done wrong by me and fixed here -- wasn't the one in the main games, and also a Wind Power translation in DE which was missing -- this last one is by @CodeTappert.

I can confirm that Traditional Chinese descriptions in the main games use "。", but it would be nice to have a native speaker review it just in case they also have some sort of suffix that needs to be added to the sentence as well. ChatGPT says the translations are still well written in 6-7 random examples I tested.

This PR https://github.com/pagefaultgames/pokerogue/pull/1472 missed the number change in the korean locale for newest Berry Pouch change, so that's added too by @MrWaterT 

## How to test the changes?
Sentences should now show ".", -- or  "。" for the Chinese translations --when showing an item's description.

## Screenshots

image

post

## Checklist
- [X] There is no overlap with another PR? There have been many merge conflicts throughout the approval of this which have been merged.

- [X] The PR is self-contained and cannot be split into smaller PRs? It bundles some minor locale changes, but it's 99% localisation.

[...]

PR #1899, opened 2024-06-07
[Item] Add Light Ball, Thick Club, Metal Powder, and Quick Powder

xsn34kzx:
## What are the changes?

Completely adds and implements the following items: [`Light Ball`](https://bulbapedia.bulbagarden.net/wiki/Light_Ball), [`Thick Club`](https://bulbapedia.bulbagarden.net/wiki/Thick_Club), [`Metal Powder`](https://bulbapedia.bulbagarden.net/wiki/Metal_Powder), and [`Quick Powder`](https://bulbapedia.bulbagarden.net/wiki/Quick_Powder).

There was no active issue or PR that brought up these items, but this PR primarily aims to continue the incorporation of useful items that are not yet included in the game from the mainline titles. As a consequence, however, these items will increase the meta-viability of Pikachu (including its forms), the Cubone line (including Alolan Marowak), and Ditto. 

## What did change?

The item internal functionality was achieved by creating a new class, `SpeciesStatBoosterModifier` extending `PokemonHeldItemModifier` in `src/modifier/modifier.ts` and adding a line to `Pokemon`'s `getBattleStat()` function in `src/field/pokemon.ts` to apply the new modifier to the stat value prior to the calculation of the battle stat. This effectively applies the stat boost as needed, regardless of context; as long as `getBattleStat()` is called, the stat boost should be reflected if the species condition is met.

Weight functions were included for the items to appear when the specific species that would benefit from the respective items. That is, if there is a Pikachu in the party, `Light Ball` would have a non-zero weight as long as it doesn't have a `Light Ball` already. Continuing with this example, note that `Light Ball` also has utility with [`Fling`](https://bulbapedia.bulbagarden.net/wiki/Fling_(move)#Trivia). As such, although it's not implemented yet, it is included for future-proofing purposes. Max weight, item tier, and weight function behavior are all subject-to-change and open to any opinions related to balance, simplicity, or any other concerns.

The English names and descriptions for all items were added to the corresponding localization file for modifiers. Descriptions were taken relatively verbatim from Scarlet/Violet as listed on the item's respective Bulbapedia pages to reflect the guidelines/discussion from #1696.

Sprites for all items were included as individual `.png` files, and TexturePacker was used along with the repo's `configuration.tps` file to generate a newly optimized texture atlas (`items.png`) along with an updated `items.json` file. Since the same settings were used to generate the original texture atlas, item appearance should have not changed, but I'd definitely like to know if there are any discrepancies. In the test runs that I've done for these items, nothing stuck out as broken or misaligned, but it's possible. These sprites were sourced from [Black/White 2](https://www.spriters-resource.com/ds_dsi/pokemonblack2white2/sheet/48051/).

### Screenshots/Videos

Note that these items were not in the game at all prior to this PR. Additionally, a modified reward pool was used to demonstrate the items' descriptions and appearances and is not in the PR itself.

#### Light Ball - Appearance and Description

lightBallDesc

Tests were done with Pikachu and his standard form, Cute Cosplay form, and G-Max form. Note that the first Object with a `value` field is `sourceAtk` with type `Util.NumberHolder`, and the second Object with a `value` field is `targetDef` with type `Util.NumberHolder`.

##### Light Ball - Behavior, Before

Using a physical attack,
lightBallNoneAtk

Using a special attack, 
lightBallNoneSpAtk

##### Light Ball - Behavior, After

Using a physical attack,
lightBallAtk

Using a special attack,
lightBallSpAtk

#### Thick Club - Appearance and Description

thickClubDesc

Tests were done with Cubone, Marowak, and Alolan Marowak.

##### Thick Club - Behavior, Before

Using a physical attack,
thickClubNoneAtkMaro

##### Thick Club - Behavior, After

Using a physical attack,
thickClubAtkMaro

#### Metal/Quick Powder - Appearance and Description

metalPowderdesc

quickPowderDesc

Tests were done with Ditto, both before and after using Transform.

##### Metal Powder - Behavior, Before

Being hit with a physical attack,
metalPowderNoneDef

##### Metal Powder - Behavior, After

Being hit with a physical attack,
metalPowderDef

##### Quick Powder - Behavior, Before

Using any move,
quickPowderNoneSpd

##### Quick Powder - Behavior, After

Using any move,
quickPowderSpd

#### General Behavior

Note that all items were tested on species that they do not apply to, and it was observed that the application messages in the console never appeared.

## How to test the changes?

For general testing related to item functionality, the `src/override.ts` file was modified such that the `STARTING_HELD_ITEMS_OVERRIDE` array included any of the item names: `LIGHT_BALL`, `THICK_CLUB`, `METAL_POWDER` and `QUICK_POWDER`. `STARTER_SPECIES_OVERRIDE` was also used to test the specific `Species` affected by the items: `Species.PIKACHU`, `Species.CUBONE`, `Species.MAROWAK`, `Species.ALOLA_MAROWAK`, and `Species.DITTO`. 

`STARTER_FORM_OVERRIDE` was additionally used to specifically observe how `Light Ball` interacted with Pikachu's different forms (indices 0 through 8).

## Comments

A couple things worth noting:
- `Metal Powder` and `Quick Powder` intentionally stat boost Ditto even after it transforms, which is not what it does in the mainline titles. This was done to give the items more utility since they typically don't have any viability when limited to pre-Transform Ditto. This is subject-to-change however since this is a balance-related quality than necessarily required for the overall usability/inclusion of the items.
- The weighting functions reflect the experience from working on #1574; however, there was a consideration to make these items behave like evolution items as far as using a `Generator` pattern instead of individual `ModifierType` entries. This is something I'm back and forth on since this would limit the items to one slot at most in a reward roll, may require another set of enums to streamline the generation process, etc. I'm not entirely sure if it would be the better solution, but I wanted to get some methodology for weighting in the PR for discussion and opinions. If it makes sense after review, I'm more than fine with changing the implementation to reflect that.
- For the above bullet, note that this change would need to be done before merging since generators and types save differently to data files and refactoring the implementation would break saves that, for example, have `Light Ball`s with `args` rather than `typePregenArgs`.

## Checklist
- [X] Are the changes visual?


hayuna:
[Question] What if I have 2 pikachu in the team? Should this item not show in the shop if only one of them has this item held?

src/locales/en/modifier-type.ts
```diff
@@ -182,6 +182,11 @@ export const modifierType: ModifierTypeTranslationEntries = {
 
     "SOOTHE_BELL": { name: "Soothe Bell" },
 
+    "LIGHT_BALL": { name: "Light Ball", description: "It's a mysterious orb that boosts Pikachu's Attack and Sp. Atk stats." },
+    "THICK_CLUB": { name: "Thick Club", description: "This hard bone of unknown origin boosts Cubone or Marowak's Attack stat." },
+    "QUICK_POWDER": { name: "Quick Powder", description: "Extremely fine yet hard, this odd powder boosts Ditto's Speed stat." },
+    "METAL_POWDER": { name: "Metal Powder", description: "Extremely find yet hard, this odd powder boosts Ditto's Defense stat." },
```
WingofaGriffin, line 188: Typo here on "fine" for metal powder.

xsn34kzx: Good catch, I'll fix it in the next commit!

xsn34kzx:
> Yeah time for Ditto Imposter to shine, Quick Powder is exactly what it needs for sweeping (in the main serie you would use Choice Scarf) and Metal Powder is a nice touch too so it can tank a bit. 👍
> 
> Edit: I realized, since you did those to work for Ditto after transform, did you make it also for Pikachu and Cubone-line? There is an edge case possible, if you fuse them with Ditto they can learn transform, then you unfuse them and they still have the move.
> 
> So to be consistent maybe their item should also apply after transform? That's a very niche scenario however. 🤔

@walt253 Yeah, it should be the case that Transform is ignored for every Pokemon regardless of if they learn it or not naturally, but I haven't tested it directly. If it works for Ditto though, I'd be surprised if it didn't work for everyone else.

xsn34kzx:
> [Question] What if I have 2 pikachu in the team? Should this item not show in the shop if only one of them has this item held?

@hayuna As of right now, the weight function pretty much scans the party one by one until one member satisfies the condition. So in this scenario, if only one Pikachu has a Light Ball, then the other Pikachu will get counted and it'll be considered in reward rolls.

xsn34kzx:
The items now apply if a member is fused with the relevant species for that specific item. Now, for example, you can have a Pikachu and Marowak fusion with both Light Ball and Thick Club applying their respective boosts.

WingofaGriffin:
Should we also consider adding in Leek https://bulbapedia.bulbagarden.net/wiki/Leek
I know we are already using the icon for Baton, but there are plenty of substitutes for that icon left.

Maybe also worth considering Deep Sea Scale/tooth and letting the effects work on the Clamperl evolutions?

damocleas:
> Should we also consider adding in Leek https://bulbapedia.bulbagarden.net/wiki/Leek I know we are already using the icon for Baton, but there are plenty of substitutes for that icon left.
> 
> Maybe also worth considering Deep Sea Scale/tooth and letting the effects work on the Clamperl evolutions?

this is being put in in a separate PR

xsn34kzx:
> Should we also consider adding in Leek https://bulbapedia.bulbagarden.net/wiki/Leek I know we are already using the icon for Baton, but there are plenty of substitutes for that icon left.
> 
> Maybe also worth considering Deep Sea Scale/tooth and letting the effects work on the Clamperl evolutions?

Echoing what damo said, definitely got a plan for Leek very soon and ideas for the Clamperl items. Those things will definitely have to be different PRs though, since Leek will have a different implementation from these and the Clamperl items are pretty unique as far as also making them act as evolutionary items.

xsn34kzx:
Summarizing the last batch of commits:
- The items are now grouped into a generator class `SpeciesStatBoosterModifierTypeGenerator`, which essentially combines the weighting functions of each item individually to a simple weighting algorithm that determines which items in the group are useful based on the party's members and randomly chooses one from the determined batch. A lot of the commits involve refactoring this to be more performant and clean but also to prepare in advance for #2161 and the Clamperl items (since I'm probably gonna get on those sometime after this).
- The localization entry has been shifted and renamed into its own group for the reasons mentioned in the generator point. Moreover, localization has been done for all languages except Portuguese from Scarlet/Violet directly. A few translations definitely go outside of text box bounds, but these aren't as easy to take an educated guess at adjusting or shortening, so I'll leave that to the localizers. 
Sources:
  - `Light Ball`: [name](https://abcboy101.github.io/poke-corpus/#id=sv.itemname.ITEMNAME_236), [description](https://abcboy101.github.io/poke-corpus/#id=sv.iteminfo.ITEMINFO_236)
  - `Thick Club`: [name](https://abcboy101.github.io/poke-corpus/#id=sv.itemname.ITEMNAME_258), [description](https://abcboy101.github.io/poke-corpus/#id=sv.iteminfo.ITEMINFO_258)
  - `Metal Powder`: [name](https://abcboy101.github.io/poke-corpus/#id=sv.itemname.ITEMNAME_257), [description](https://abcboy101.github.io/poke-corpus/#id=sv.iteminfo.ITEMINFO_257)
  - `Quick Powder`: [name](https://abcboy101.github.io/poke-corpus/#id=sv.itemname.ITEMNAME_274), [description](https://abcboy101.github.io/poke-corpus/#id=sv.iteminfo.ITEMINFO_274)

xsn34kzx:
Unit tests have been added for every item in the PR that effectively replicate the before/after testing from the initial post and images.

damocleas (approved):

src/locales/pt_BR/modifier-type.ts
```diff
@@ -241,6 +241,8 @@ export const modifierType: ModifierTypeTranslationEntries = {
     "ENEMY_ENDURE_CHANCE": { name: "Token de Persistência" },
     "ENEMY_FUSED_CHANCE": { name: "Token de Fusão", description: "Adiciona uma chance de 1% de que um Pokémon selvagem seja uma fusão." },
   },
+  SpeciesBoosterItem: {
+  },
```
Madmadness65, line 245: Is the modifier being left blank on the Portuguese localization intended? At the very least having a copy of the English entries here would probably be preferable.

xsn34kzx: Yeah, it was intentional since, if there's no localized entries present, we automatically default to English as a fallback language (only noticed this while running through the list of languages).

englishFallback

If you want, I can still add the English entries anyway though.

Madmadness65: Ah, I wasn't aware that was even the case when left blank like that, not just missing. It should be fine then.

Madmadness65 (approved):

PR #1900, opened 2024-06-07
Remove Reviver Seed from trainer modifier pool

Fontbane:
## What are the changes?

Removed the reviver seed from trainer modifier pool so Ivy's Rayquaza can't say tee hee I actually revived so you have to fight me again

As per head of balance:
image

## What did change?

The seed was removed from the trainer modifier pool

## How to test the changes?

- Fight a bunch of trainers and rejoice as none will have the seed of sorrow

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -1447,7 +1447,6 @@ const trainerModifierPool: ModifierPool = {
     m.setTier(ModifierTier.ULTRA); return m;
   }),
   [ModifierTier.ROGUE]: [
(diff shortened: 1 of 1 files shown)


Tempo-anon (approved):
Yes please. Reviver seed is way too swingy as a randomly generated item.

RNGezeus:
I don't disagree with anything. Seeds you don't see always jumpscares you. I always sigh audibly as I face it. If it is removed I won't be the one crying on its funeral.

In a roguelike fashion though, its one of the primary random elements that can interact with setup sweepers.

PR #1906, opened 2024-06-07
[Refactor] Move enum BattleStyle into separated file

hayuna:
## What are the changes?
I've moved battle-style enum into their own files inside src/data/enums

Moving enum to respective files
Avoiding hardcoded magic numbers and replace them with enums

## Checklist
- [ ] Are the changes visual?


src/enums/battle-style.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Determines the selected battle style.
+ * - 'Shift' - The option to switch the active pokemon at the start of a battle will be displayed.
+ * - 'Set' - The option to switch the active pokemon at the start of a battle will not display.
+*/
+export enum BattleStyle {
+    SHIFT,
+    SET
```
Tempo-anon, line 8: nit: According to `setting.ts` this should be switch/set instead of shift/set https://github.com/pagefaultgames/pokerogue/blob/1a6b4119875c633d2f5535f8f07110ad044bb42e/src/system/settings/settings.ts#L116

hayuna: Done

Tempo-anon (approved):

Tempo-anon (approved):

PR #1908, opened 2024-06-07
[Testing] Flexible Testing Wrapper for Phaser-Based Battle-Scenes

Greenlamp2:
## What are the changes?
This PR introduces a wrapper for Phaser.Game to allow testing any project feature without worrying about the "drawing" or "rendering" parts. While testing some graphical aspects of the game would be a plus, it's not the main focus at the moment. With this wrapper, we can now instantiate a Battle-Scene and test phase by phase what we want. 

These changes are introduced to facilitate testing by allowing manipulation of data between phases to test different scenarios. It also enables intercepting displayed messages, pausing the game between phases, executing code during a prompt, and sending Button-type inputs as if in the game. This improvement stems from the need to enhance the testing capabilities for the project, providing a more controlled and flexible testing environment.

## What did change?
- The introduction of a Phaser.Game wrapper allows for:
  - Instantiation of a Battle-Scene for phase-by-phase testing.
  - Data manipulation between phases.
  - Message interception.
  - Game pausing between phases.
  - Execution of code during prompts.
  - Simulation of Button-type inputs.
- Mocking fetches to retrieve assets from the public folder.
- Interception of console.log outputs.
- Mocking of all drawing classes (text, rectangle, sprite, container, polygon, image, nineslice, graphics) to ensure the proper functioning of certain game actions.
- A wrapper for local storage to simulate it in tests.
- A wrapper for Phaser.Time.clock to update the clock every 100ms instead of during pre-update and update of the scene, as the Phaser engine is simulated and doesn't actually exist.

## Why are not all tests passing?
Because they are detecting bugs that need to be resolved, but it is not the purpose of this PR to fix them.

### Screenshots/Videos
image

## How to test the changes?
To test these changes, reviewers can check out the branch and run the provided rudimentary tests which include:
- Ability tests (intimidate and moxie)
- HeldItems test (toxic orb)
- Achievement tests
- Input tests with a fake gamepad/keyboard/mobile
- Move tests (Tackle, Growth, Tail Whip)
- Starter selection screen tests
- Specific battle tests like execution order when multiple Pokémon have the same speed

Ensure that when testing or stopping at phases not yet intercepted, they are added to PhaseInterceptor.

## Checklist
- [ ] Are the changes visual?


src/inputs-controller.ts
```diff
@@ -548,7 +548,8 @@ export class InputsController {
     if (!this.isButtonLocked(button)) {
       return false;
     }
-    if (this.time.now - this.interactions[button].pressTime >= repeatInputDelayMillis) {
+    const duration = Date.now() - this.interactions[button].pressTime;
+    if (duration >= repeatInputDelayMillis) {
       return true;
     }
   }
```
OrangeRed, line 555: Should this input delay change be a part of this PR?

src/ui/game-stats-ui-handler.ts
```diff
@@ -1,3 +1,4 @@
+import Phaser from "phaser";
```
OrangeRed, line 1: Is this lone import necessary if nothing else in this file changed?

src/system/misc/parseSessionData.ts
OrangeRed, line 1: This function was moved to a separate file but its not being consumed in `src/system/game-data.ts`?

src/inputs-controller.ts
Greenlamp2: yes, otherwise inputs test are not possible

src/ui/game-stats-ui-handler.ts
Greenlamp2: oh yes, phaser is using black magic ! it this import is not at this specific spot, nothing works anymore ahah
image

src/system/misc/parseSessionData.ts
Greenlamp2: indeed it's not used anymore

OrangeRed: Wouldn't it better to have this a separate file still in case in the future we decide to use something like zod to validate and add typedefs to session data ?

Greenlamp2: i will not contradict you, but it will not be in this PR

brain-frog (approved):
finally. unit tests

PR #1909, opened 2024-06-07
[Refactor] Move enum ExpNotification into separated file

hayuna:
## What are the changes?
I've moved exp-notification enum into their own files inside src/data/enums

Moving enum to respective files
Avoiding hardcoded magic numbers and replace them with enums

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -4303,7 +4303,7 @@ export class ShowPartyExpBarPhase extends PlayerPartyMemberPokemonPhase {
         this.end();
       }
     } else if (this.scene.expGainsSpeed < 3) {
-      this.scene.partyExpBar.showPokemonExp(pokemon, exp.value, this.scene.expParty === ExpNotification.ONLY_LEVEL_UP, newLevel).then(() => {
+      this.scene.partyExpBar.showPokemonExp(pokemon, exp.value, false, newLevel).then(() => {
         setTimeout(() => this.end(), 500 / Math.pow(2, this.scene.expGainsSpeed));
       });
     } else {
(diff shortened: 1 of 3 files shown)


Tempo-anon (commented):
Looks good to me, but should battle-scene's `expParty` also be changed from an `integer` to the new enum?

src/phases.ts
```diff
@@ -4289,20 +4290,20 @@ export class ShowPartyExpBarPhase extends PlayerPartyMemberPokemonPhase {
...
-        this.scene.partyExpBar.showPokemonExp(pokemon, exp.value, this.scene.expParty === 1, newLevel).then(() => {
+        this.scene.partyExpBar.showPokemonExp(pokemon, exp.value, this.scene.expParty === ExpNotification.ONLY_LEVEL_UP, newLevel).then(() => {
           setTimeout(() => this.end(), 800 / Math.pow(2, this.scene.expGainsSpeed));
         });
       } else {
         this.end();
       }
     } else if (this.scene.expGainsSpeed < 3) {
-      this.scene.partyExpBar.showPokemonExp(pokemon, exp.value, this.scene.expParty === 1, newLevel).then(() => {
+      this.scene.partyExpBar.showPokemonExp(pokemon, exp.value, false, newLevel).then(() => {
```
hayuna, line 4306: This condition will be always false, because `this.scene.expParty === 1` condition was checked in line 4295

Tempo-anon:
Does the settings saving this selection still work? https://github.com/pagefaultgames/pokerogue/blob/20a3a4f60fe58a5fe929a51dff76a5db64080492/src/system/settings/settings.ts#L411

Just want to make sure so if you could test that then it should be all good to go

hayuna:
Uploading video in progress

PR #1924, opened 2024-06-07
[Ability] Implement forecast

Tempo-anon:
## What are the changes?
Adding Forecast

Make Castform's ability work

## What did change?
* Added all 12 of Castform's form change transformations
* Added a new `SpeciesFormChangeWeatherTrigger` class for species form changes based on weather
* Modified Arena.trySetWeather so it checks all active Pokemon if any of them have any weather based form changes
* Added a new `PostSummonFormChangeByWeatherAbAttr` ability attribute

## How to test the changes?
Override a Castform with weather moves
```ts
export const STARTER_SPECIES_OVERRIDE: Species | integer = Species.CASTFORM;
export const ABILITY_OVERRIDE: Abilities = Abilities.FORECAST;
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.SANDSTORM, Moves.SUNNY_DAY, Moves.RAIN_DANCE, Moves.HAIL];
```

## Checklist
- [x] Are the changes visual?


src/data/pokemon-forms.ts
```diff
@@ -231,6 +232,47 @@ export class SpeciesFormChangeTimeOfDayTrigger extends SpeciesFormChangeTrigger
   }
 }
 
+/**
+ * Class used for triggering form changes based on weather.
+ * Used for Castform and Cherrim
+ */
+export class SpeciesFormChangeWeatherTrigger extends SpeciesFormChangeTrigger {
```
flx-sta, line 239: Wouldn't it be `SpeciesFormWeatherChangeTrigger` ?

src/data/pokemon-forms.ts
```diff
@@ -231,6 +232,47 @@ export class SpeciesFormChangeTimeOfDayTrigger extends SpeciesFormChangeTrigger
...
+  canChange(pokemon: Pokemon): boolean {
+    const currentWeather = pokemon.scene.arena.weather?.weatherType ?? WeatherType.NONE;
+    let weatherSuppressedFlag = false;
+    if (this.clearWeather) {
+      if (currentWeather === WeatherType.NONE) {
+        return true;
+      }
+      weatherSuppressedFlag = pokemon.scene.arena.weather.isEffectSuppressed(pokemon.scene);
+    }
+    return weatherSuppressedFlag || (pokemon.hasAbility(this.ability)
```
flx-sta, line 270: when `weatherSuppresedFlag = true` it would return `true` thus saying `canChange => true`, right?

src/field/arena.ts
```diff
@@ -318,6 +320,13 @@ export class Arena {
       this.scene.queueMessage(getWeatherClearMessage(oldWeatherType));
     }
 
+    this.scene.getField(true).forEach( p => {
+      if ((p.hasAbility(Abilities.FORECAST) && p.species.speciesId === Species.CASTFORM)) {
```
flx-sta, line 324:
```suggestion
      if (p.hasAbility(Abilities.FORECAST) && p.species.speciesId === Species.CASTFORM) {
```

src/field/arena.ts
```diff
@@ -318,6 +320,13 @@ export class Arena {
...
+        this.scene.triggerPokemonFormChange(p, SpeciesFormChangeWeatherTrigger);
+      }
+    }
+    );
```
flx-sta, line 328:
```suggestion
    });
```

flx-sta (changes requested):

src/data/pokemon-forms.ts
Tempo-anon: I modeled it after `SpeciesFormChangeTimeOfDayTrigger`

src/field/arena.ts
Tempo-anon: Yeah I have a double pair of parantheses here because eventually Cherrim needs to be added here too

src/data/pokemon-forms.ts
Tempo-anon: Yes. If the weather is suppressed then it's treated as no weather so the Pokemon should trigger a form change to go back to formIndex 0 if it is in a different state

torranx:
arent you missing some interactions? https://bulbapedia.bulbagarden.net/wiki/Forecast_(Ability)

Screenshot 2024-06-08 at 8 25 25 PM

Tempo-anon:
> If a Pokémon [transforms](https://bulbapedia.bulbagarden.net/wiki/Transform_(move)) into Castform, the Pokémon will remain in the same form as the target Castform, regardless of the weather.

Proof Transform keeps you in the same form you transformed into. There's some undefined behavior if Castform transforms into another Castform but for now the Forecast check has a species lock on Castform.

> If Castform loses Forecast or Forecast is suppressed with [Gastro Acid](https://bulbapedia.bulbagarden.net/wiki/Gastro_Acid_(move)), it will revert to its normal form and not be able to change form. If Castform regains Forecast, it will change form to match the current weather, unless it would be in its normal form in the current weather.

We currently don't have the framework of abilities triggering on gain/loss. Same issue with intimidate not activating when the ability is gained.

> [Role Play](https://bulbapedia.bulbagarden.net/wiki/Role_Play_(move)) fails if the target has the Ability Forecast. [Entrainment](https://bulbapedia.bulbagarden.net/wiki/Entrainment_(move)) fails if the user has the Ability Forecast. Forecast cannot be copied by [Power of Alchemy](https://bulbapedia.bulbagarden.net/wiki/Power_of_Alchemy_(Ability)) or [Receiver](https://bulbapedia.bulbagarden.net/wiki/Receiver_(Ability)), nor can it be traded by [Wandering Spirit](https://bulbapedia.bulbagarden.net/wiki/Wandering_Spirit_(Ability)).

Gave Forecast the `UnswappableAbilityAbAttr` (That name is.... ability ability?!?)

Tempo-anon:
Oh there's an edge case about weather suppressors coming in/out of combat, as well as the primal weathers... I think I should also refactor the check everyone for form changes based on weather into its own function I can call

flx-sta:
Figure I've tested as much as I could.Couldn't see anything odd except the already mentioned edge cases

src/data/ability.ts
```diff
@@ -3069,6 +3139,27 @@ export class PostFaintClearWeatherAbAttr extends PostFaintAbAttr {
...
+   * @param hitResult n/a
+   * @param args n/a
+   * @returns true
+   */
+  applyPostFaint(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: Move, hitResult: HitResult, args: any[]): boolean {
+    // THIS DOES NOT TRIGGER ON SELF-KO MOVES LIKE MEMENTO
+    pokemon.scene.arena.triggerWeatherBasedFormChanges();
+    return true;
+  }
+}
```
OrangeRed, line 3161: I think this should do the same thing the other for change abilities do like schooling and zen mode in that they trigger on `PostBattleInitFormChangeAbAttr` instead of PostFaint. 
This will require you to change `PostBattleInitFormChangeAbAttr` to accept any trigger type instead of the current manual trigger

src/data/ability.ts
```diff
@@ -1752,6 +1752,57 @@ export class PostSummonFormChangeAbAttr extends PostSummonAbAttr {
...
+ * Used for Forecast and Flower Gift
+ */
+export class PostSummonFormChangeByWeatherAbAttr extends PostSummonAbAttr {
+  private ability: Abilities;
+
+  constructor(ability: Abilities) {
+    super(true);
+
+    this.ability = ability;
+  }
```
OrangeRed, line 1788: Does this need to be its own class? Can't you do similar with the already existing `PostSummonFormChange` class?

OrangeRed (changes requested):

src/data/pokemon-forms.ts
```diff
@@ -465,6 +532,20 @@ export const pokemonFormChanges: PokemonFormChanges = {
...
+    new SpeciesFormChange(Species.CASTFORM, "", "snowy", new SpeciesFormChangeWeatherTrigger(Abilities.FORECAST, [WeatherType.HAIL, WeatherType.SNOW])),
+    new SpeciesFormChange(Species.CASTFORM, "sunny", "snowy", new SpeciesFormChangeWeatherTrigger(Abilities.FORECAST, [WeatherType.HAIL, WeatherType.SNOW])),
+    new SpeciesFormChange(Species.CASTFORM, "rainy", "snowy", new SpeciesFormChangeWeatherTrigger(Abilities.FORECAST, [WeatherType.HAIL, WeatherType.SNOW])),
+    new SpeciesFormChange(Species.CASTFORM, "sunny", "", new SpeciesFormChangeWeatherOriginalFormTrigger(Abilities.FORECAST, [WeatherType.NONE, WeatherType.SANDSTORM, WeatherType.STRONG_WINDS])),
+    new SpeciesFormChange(Species.CASTFORM, "rainy", "", new SpeciesFormChangeWeatherOriginalFormTrigger(Abilities.FORECAST, [WeatherType.NONE, WeatherType.SANDSTORM, WeatherType.STRONG_WINDS])),
+    new SpeciesFormChange(Species.CASTFORM, "snowy", "", new SpeciesFormChangeWeatherOriginalFormTrigger(Abilities.FORECAST, [WeatherType.NONE, WeatherType.SANDSTORM, WeatherType.STRONG_WINDS])),
+  ],
```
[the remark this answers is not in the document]
Tempo-anon: I couldn't think of any hence this abomination. This is also seen in Ogerpon

src/data/ability.ts
```diff
@@ -1752,6 +1752,57 @@ export class PostSummonFormChangeAbAttr extends PostSummonAbAttr {
...
 }
 
+/**
+ * Ability attribute for all PostSummon abilities that affect the weather
+ * Used in cloud nine and air lock
+ */
+export class PostSummonTriggerWeatherSuppressedFormChangesForAllAbAttr extends PostSummonAbAttr {
+  constructor() {
+    super();
+  }
```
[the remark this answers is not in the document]
Tempo-anon: Neat!

src/data/pokemon-forms.ts
```diff
@@ -231,6 +232,72 @@ export class SpeciesFormChangeTimeOfDayTrigger extends SpeciesFormChangeTrigger
...
+/**
+ * Class used for reverting to the original form based on weather.
+ * Used in cases where the weather runs out or is suppressed
+ * Used for Castform and Cherrim
+ */
+export class SpeciesFormChangeWeatherOriginalFormTrigger extends SpeciesFormChangeTrigger {
+  /** The ability that  triggers the form change*/
+  public ability: Abilities;
+  /** The list of weathers that will also trigger a form change to original form */
+  public weathers: WeatherType[];
```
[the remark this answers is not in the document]
Tempo-anon: It's slightly different since weather suppressing abilities will always revert a Pokemon to their base form so the SpeciesFormChangeWeatherFormTrigger has different behavior when it comes to weather suppression. When a weather suppression ability is on the field SpeciesFormChangeWeatherFormTrigger will always return `false`. However, when a weather suppresor enters the field then it should always be `true` hence why I created two triggers

src/data/ability.ts
Tempo-anon: Did you mean this on a different AbAttr? This is added onto weather suppression abilities to trigger all the weather form changers when they faint

src/data/ability.ts
Tempo-anon: I believe PostSummonFormChange only accounts for a single form and only works with `SpeciesFormChangeManualTrigger`s

src/data/pokemon-forms.ts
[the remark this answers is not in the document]
chaosgrimmon: you'd have to introduce some kind of tag, basically a ANY_FORM_EXCEPT_EVO_FORM -> "evo_form_key", right? probably not worth the hassle provided the game only checks form change conditions provided a species with a form change has an existing Pokemon Object in play.

also, if you want these to not do the evolution-phase animation, you have to put a `true` at the end of them, to apply `quiet`.
```ts
 constructor(speciesId: Species, preFormKey: string, evoFormKey: string, trigger: SpeciesFormChangeTrigger, quiet: boolean = false, ...conditions: SpeciesFormChangeCondition[]) {
 ```
(see ogerpon for examples)

Tempo-anon:
Merge conflicts and edge cases... will leave this for swain who has a lot more experience with form changes like ice face and gulp missile

PR #1928, opened 2024-06-08
[QoL] Slight Resize and Reallocate to IV graph in new game screen

tomasrggm:
## What are the changes?
This commit slightly resizes the iv graph and repositions it so as to not overlap the moves segment. The text was also resized and repositioned so as to not interfere with the graph or the other components in the stats container using the custom styling options available.

Currently the IV chart and its labels needlessly clutter the stats container in a new playthrough screen for when you wish to see stats and moves at the same time. This change should make it less inconvenient if you wish to see both at the same time.

## What did change?
-add offsetValues for x and y coordinates for labels
-Change labels position, and labels size according to the showDiff boolean (using it as an indicator for post capture or ivtoggle)
-Offset labels so they don't interfere with chart or moves
-Reduce moves container to be in line with egg moves

### Screenshots/Videos
Before the change:
image
After the change:
image

## How to test the changes?
Start a new game and toggle IV in any pokemon

## Checklist
- [X] Are the changes visual?


sirzento:
Good idea but I feel like the text is a bit hard to read now. Imo just hiding the moveset would be enough.

tomasrggm:
I see your point. Although I prefer having both information presented to me at the same time. If there was a way to better outline the text or increase the shadow so it makes a contrast for the words to be more perceptible would be nice. I'm also fiddling with larger words to see how much i can squeeze out of the container (here's with 58 px as an example):
image

tomasrggm:
Another idea is maybe offsetting the already existing shadows in the font by the x coordinate? As I move the words around that see that the shadow changed. That could help

sirzento:
Yeah resizing the moveset to the same size as the egg moves would be a good idea if you still want to display them. This will enable the graph to be a bigger size but then the only change would be resizing the moveset because I think the original size would fit.

Also test other languages since the text could be longer.

tomasrggm:
Changed the PR according to feedback. The moveset container is smaller and reverted the changes done to the chart for the most part. Also tested with other languages and everything seems to be in order with the exception of some asian languages because for some reason their characters are absolutely massive for some reason (Don't think my changes had any impact in this regard).
Here is what is looks like right now:
image

tomasrggm:
Also, just now realized that doing fontStyle: 62 is not the same as fontStyle: "62px". The way I was doing it was messing up the shadows. Seems to be in order now:
image

bennybroseph:
It's better to create a new font style entirely rather than manipulating stuff to make one fit.

tomasrggm:
Isn't that excessive in this instance? Why even have the ability for custom properties if we're being so strict with it? All that was done was reduce the font size by 2px in comparison to the default. There's other places in the code in which this was done as well.

bennybroseph:
It fixes the issue you're having with shadows

bennybroseph:
Also, let's not forget to test this in legacy UI

tomasrggm:
Looks fine on legacy UI. Checked in all languages. There was never an issue with shadows, it was because I misinterpreted the fontSize property which led me to make a mistake and messed it up myself. 
image

brain-frog:
2 things 
1. can you please update the PR description with the most recent screenshots when visuals change
2. can you show examples of languages that show up differently or have a different font (Chinese, Korean)

tomasrggm:
1. Updated the pr as requested
2. Here's examples for chinese and korean. Western languages look the same as english visual wise:
image
image
image
The chinese fonts seem needlessly massive in my opinion. Though I think that is out of scope for this PR and should be checked at someplace else

bennybroseph:
> Looks fine on legacy UI. Checked in all languages. There was never an issue with shadows, it was because I misinterpreted the fontSize property which led me to make a mistake and messed it up myself.

Assuming this is the latest version, there is definitely some issues with shadows. In general the font is very blurry, but the shadow is barely offset at all and is making it more apparent. Just compare the icon sharpness (which needs some help anyway) to the font and you'll see what I mean

image
image

tomasrggm:
I see your point now. Yes fiddling with the font in a way its original scale isn't preserved is probably what is making it blurry. I've decided to settle with the original size which now that I've downscaled the moves container isn't a bad solution in itself. Here's an example in english and chinese (remaining western/asian languages are about the same):
image
image

tomasrggm:
My mistake, I misinterpreted the suggested changes hence why the eslint failed. Should be good to go as intended now.

tomasrggm:
Any updates on whether this can be merged or not?

flx-sta:
### Are you still working on this PR?
-----
No actions for over a month

tomasrggm:
I've addressed all feedback hence why it hasn't been updated since.

flx-sta:
> I've addressed all feedback hence why it hasn't been updated since.

All good. I’m just going through PRs to sort out outdated/stale ones. Thanks for the response

flx-sta:
@tomasrggm there is some merge-conflicts to resolve

tomasrggm:
@flx-sta should be good to go

flx-sta (approved):
lgtm

## Desktop (DE)
image

## Mobile (DE)
image
image

src/ui/stats-container.ts
```diff
@@ -55,12 +57,13 @@ export class StatsContainer extends Phaser.GameObjects.Container {
     this.ivStatValueTexts = [];
 
     for (const s of PERMANENT_STATS) {
-      const statLabel = addTextObject(this.scene, ivChartBg.x + (ivChartSize) * ivChartStatCoordMultipliers[s][0] * 1.325, ivChartBg.y + (ivChartSize) * ivChartStatCoordMultipliers[s][1] * 1.325 - 4 + ivLabelOffset[s], i18next.t(getStatKey(s)), TextStyle.TOOLTIP_CONTENT);
+      const statLabel = addTextObject(this.scene, ivChartBg.x + (ivChartSize) * ivChartStatCoordMultipliers[s][0] * 1.325 + (this.showDiff ? 0 : ivLabelOffset[s]), ivChartBg.y + (ivChartSize) * ivChartStatCoordMultipliers[s][1] * 1.325 - 4 + (this.showDiff ? 0 : ivChartLabelyOffset[s]), i18next.t(getStatKey(s)), TextStyle.TOOLTIP_CONTENT);
```
DayKev, line 60:
```suggestion
      const statLabel = addTextObject(
        this.scene,
        ivChartBg.x + (ivChartSize) * ivChartStatCoordMultipliers[s][0] * 1.325 + (this.showDiff ? 0 : ivLabelOffset[s]),
        ivChartBg.y + (ivChartSize) * ivChartStatCoordMultipliers[s][1] * 1.325 - 4 + (this.showDiff ? 0 : ivChartLabelyOffset[s]),
        i18next.t(getStatKey(s)),
        TextStyle.TOOLTIP_CONTENT
      );
```
This line is getting needlessly long.

tomasrggm:
@DayKev applied suggestion.

DayKev (approved):

flx-sta (approved):

innerthunder (approved):

flx-sta:
@f-fsantos as previously wanted I did a check on beta and I admit it’s fairly small now but one can still read it (iPhone 13 screen).
That’s my feedback

PR #1935, opened 2024-06-08
[Localization] Abbreviate long german translations in pokemon info boxes

kvngrnr:
As in many games, some german words tend to be a lot longer than their translations in other languages. Therefore they often clip out of frames or overlap. Pokerogue is no exception.
This PR abbreviates the words for Ability (=Fähigkeit) and Gender (=Geschlecht) on the starter screen and the pokemon info screen (for catches and hatches).

## What are the changes?
Abbreviates german translations on starter and pokemon info screen to avoid overlapping.

It slightly annoyed me while playing.

## What did change?
Fixes overlapping of german translations on starter and pokemon info screen by abbreviating the words.
As I understand, moving or changing the layout of these frames is not an option, so abbreviation seems to be the best solution.
There are no official dictionary abbreviations for these words, but I would say that it' still easy to understand given the context.

### Screenshots/Videos
Before the change:
Bildschirmfoto 2024-06-08 um 03 18 10
Bildschirmfoto 2024-06-08 um 03 20 18

After the change:
Bildschirmfoto 2024-06-08 um 03 22 15
Bildschirmfoto 2024-06-08 um 03 21 44

## How to test the changes?
1. Set the game to german.
2. Check a pokemon on starter screen.
3. Start a run and catch or hatch a pokemon.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/locales/de/pokemon-info-container.ts
+++ b/src/locales/de/pokemon-info-container.ts
@@ -2,8 +2,8 @@ import { SimpleTranslationEntries } from "#app/plugins/i18n";
 
 export const pokemonInfoContainer: SimpleTranslationEntries = {
   "moveset": "Attacken",
-  "gender": "Geschlecht:",
-  "ability": "Fähigkeit:",
+  "gender": "Geschl.:",
+  "ability": "Fähigk.:",
   "nature": "Wesen:",
   "epic": "Episch",
   "rare": "Selten",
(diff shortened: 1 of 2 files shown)


CodeTappert:
No dont do that. Rather go with;
https://github.com/pagefaultgames/pokerogue/pull/1792

I would rather have the text be a little smaller but full instead of abbreviated like here

kvngrnr:
@CodeTappert ah my bad, didn't see that PR. Definitely the better solution. Gonna close this

PR #1976, opened 2024-06-09
[Bug] Make getTypes ignore transform when ignoring overrides

Xavion3:
## What are the changes?
Type checking effects that were meant to check only the base types now ignore transform. This most notably means that transform no longer breaks challenge runs, though it also affects pancham's evolution and tera shard type selection.

Transform was violating monotype runs.

## What did change?
ignoreOverride is now passed into the getSpeciesForm and getFusionSpeciesForm functions in getTypes.

## How to test the changes?
Use transform on anything that would violate a monotype challenge.

## Checklist


brain-frog (approved):
tested and lgtm

PR #1978, opened 2024-06-09
[Test] Add Tests for Zen Mode Ability

Greenlamp2:
cc @OrangeRed 
## What are the changes?
- Previously, there were no tests to verify the behavior of the Zen Mode ability under specific conditions. With this PR, tests have been added to:
  - Check if Zen Mode is triggered when the Pokémon's life drops below 50%.
  - Verify if the Pokémon retains its Zen Mode form after fainting.

### Screenshots/Videos
- As these changes involve backend logic for battle mechanics, there are no direct visual changes to showcase.

## How to test the changes?
- To test these changes, follow these steps:
  1. Check out the branch containing these changes.
  2. Run the test suite using `npm run test`.
  3. Ensure that all tests, including the newly added ones for Zen Mode, pass successfully.
  - No special steps are required beyond running the standard test suite.

## Checklist
- [ ] Are the changes visual?


src/phases.ts
```diff
@@ -1493,6 +1493,10 @@ export class SwitchSummonPhase extends SummonPhase {
     this.batonPass = batonPass;
   }
 
+  start(): void {
```
ben-lear, line 1496: I might be missing something, what does this serve as opposed to leaving undefined?

Greenlamp2: this is for the phase interceptor, otherwise it doesn't know that it inherit from its parent and it's not called

PR #1982, opened 2024-06-09
[Bug] resolve issue with input text persisting in top left of game screen in certain cases

SquillWall:
## What are the changes?
these changes fix a bug where inputs would persist on the ui where they should not

a bug was found where the starter select input prompts would still be visible at the top left of the screen in some cases, this is to fix that
image

## What did change?
The ui objects are now being appropriately destroyed so that they don't get added the scene without a container 

### Screenshots/Videos
before my fix 
[pre-inputFix.webm](https://github.com/pagefaultgames/pokerogue/assets/170910983/b0ab378c-c5c4-47c8-b5c8-237b00b947c7)

after my fix 
[post-inputFix.webm](https://github.com/pagefaultgames/pokerogue/assets/170910983/b2e41115-7895-4ea5-843c-3cc145a84e4a)

## How to test the changes?
From my experience, reproducing was as simple as...
 - going to the start screen
 - selecting a pokemon with some options to cycle (gender, shiny, etc.) 
 - moving cursor to the start button 
 - clicking start

As stated above, the options that were previously shown in the bottom left of starter select would then move to the top left of the battle scene

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -191,8 +191,21 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
   private pokemonCaughtCountText: Phaser.GameObjects.Text;
   private pokemonHatchedCountText: Phaser.GameObjects.Text;
   private genOptionsText: Phaser.GameObjects.Text;
+
   private instructionsContainer: Phaser.GameObjects.Container;
-  private instructionsText: Phaser.GameObjects.Text;
+  private shinyIconElement: Phaser.GameObjects.Sprite;
+  private formIconElement: Phaser.GameObjects.Sprite;
+  private abilityIconElement: Phaser.GameObjects.Sprite;
+  private genderIconElement: Phaser.GameObjects.Sprite;
+  private natureIconElement: Phaser.GameObjects.Sprite;
+  private variantIconElement: Phaser.GameObjects.Sprite;
+  private shinyLabel: Phaser.GameObjects.Text;
+  private formLabel: Phaser.GameObjects.Text;
+  private genderLabel: Phaser.GameObjects.Text;
+  private abilityLabel: Phaser.GameObjects.Text;
+  private natureLabel: Phaser.GameObjects.Text;
+  private variantLabel: Phaser.GameObjects.Text;
+
   private starterSelectMessageBox: Phaser.GameObjects.NineSlice;
   private starterSelectMessageBoxContainer: Phaser.GameObjects.Container;
   private statsContainer: StatsContainer;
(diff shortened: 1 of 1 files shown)


EvasiveAce:
Videos for convenience:

[337944590-b0ab378c-c5c4-47c8-b5c8-237b00b947c7.webm](https://github.com/pagefaultgames/pokerogue/assets/71776311/47081d7a-7080-4443-8ade-a876719f8314)

[337944594-b2e41115-7895-4ea5-843c-3cc145a84e4a.webm](https://github.com/pagefaultgames/pokerogue/assets/71776311/99ea9c4d-e790-4fa6-9913-bb4ea3f36070)

src/ui/starter-select-ui-handler.ts
```diff
@@ -1542,7 +1542,7 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
     const instructionTextSize = textSettings.instructionTextSize;
     this.instructionRowX = 0;
     this.instructionRowY = 0;
-    this.instructionsContainer.removeAll();
+    this.instructionsContainer.removeAll(true);
```
EvasiveAce, line 1545: Looks good to me, but I'm not sure where else it could be interacted with.

EvasiveAce (approved):

bennybroseph (approved):

src/ui/starter-select-ui-handler.ts
bennybroseph, line 1545: couldn't you just hide that container? i wasn't the one that originally reviewed the other PR. are you destroying the UI elements instead of hiding them when a pokemon doesn't have a form/etc.?

bennybroseph (changes requested):

src/ui/starter-select-ui-handler.ts
SquillWall: There were some complications with simply hiding a row in the container if the pokemon doesn't have the form etc. because it's no longer just strings of text, it's a dynamically populated container, so hiding unnecessary elements could possibly leave gaps

bennybroseph: well, you should do both. removing it from the container just puts it into world space coordinates. and it doesn't hide it when the starter select screen is hidden. no need to destroy them

bennybroseph: .setVisible

SquillWall: ah gotcha, that's definitely cleaner, i'll do that

SquillWall: okay @bennybroseph i did the following if you could give it a quick test. Everything looked good on my end but always better to be safe...

- moved the sprites and their text to static elements so they're only initialized once and not created/destroyed on the fly
- just call setTexture on the sprites to update them if a new input device is detected
- hide and show them dynamically, same as before
- when we redirect from the starter select screen, hide all of the elements so they don't hang around in the world space coords as you said

PR #2006, opened 2024-06-09
[Bug] Fixed moves with type multiplier effects not being considered in type hints and enemy AI

Mewtwo2387:
## What are the changes?
Fixed a bug where move type multiplier effects (such as Freeze-Dry being super effective against water) is not considered in the type hints and enemy AI.

Currently, moves like Freeze-Dry against water are shown as orange (not very effective) in the type hint, and not prioritized by the enemy, despite being super effective.

## What did change?
Added a single line of `applyMoveAttrs` in `Pokemon.getAttackMoveEffectiveness`.

`getAttackMoveEffectiveness` is only used in
1. `Pokemon.getMoveEffectiveness`, which is used for the text and colour in type hints. 
2. `EnemyPokemon.getNextMove`, using it for assigning move scores.

So it should not affect anything else.

## Screenshots/Videos
Before:
image

After:
image

## How to test the changes?
1. Turn on type hints
2. Use a pokemon with moves that have special type multiplier effects, such as Freeze-Dry and Thousand Arrows.
3. Fight against an enemy of the boosted type.
4. It should now be shown as green with a x2. 

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1085,6 +1085,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     const typeless = move.hasAttr(TypelessAttr);
     const typeMultiplier = new Utils.NumberHolder(this.getAttackTypeEffectiveness(move.type, source));
     const cancelled = new Utils.BooleanHolder(false);
+    applyMoveAttrs(VariableMoveTypeMultiplierAttr, source, this, move, typeMultiplier);
     if (!typeless) {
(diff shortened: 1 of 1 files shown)


Tempo-anon (approved):

Tempo-anon (approved):

Tempo-anon:
This has the unintended side effect of revealing if the opponent has an ability that absorbs type like sap sipper or flash fire

PR #2012, opened 2024-06-09
[QoL] Remove Map item from being the shop on waves 180-200

EmberCM:
## What are the changes?
Remove the Map from being found on waves 190-200

This is because you cannot change the biome in those waves anyway

### Screenshots/Videos
Before vs After
before
after

## Checklist
- [x] Are the changes visual?


Fontbane:
Should extend to 180, once you hit 181 you’re staying in that biome till End biome

Tempo-anon:
On one hand, yes, I agree. On the other hand, seeing map show up in the end is always still pretty funny to me

Greenlamp2 (approved):

PR #2019, opened 2024-06-09
[Enhancement] Golden IVs Chart when all IVs are perfect

prime-dialga:
## What are the changes?
The IVs chart is now golden in case all IVs are perfect.

Dynamic colors for the IV chart was already part of a different mod so i thought i might as well as this effectively only requires one line of code to change.
Though the additional code for testing overrides added some more changes...

## What did change?
The IV chart shown for new captures and starters will now appear golden, in case all IVs are perfect (31).
This color change is also animated on the starter selection screen.

For testing purposes the following overrides have been introduced:
- `OPP_IVS_OVERRIDE : integer | integer[]`
    This may take an integer to replace all enemy IVs or an array with each element replacing the respective IV.
- `window.perfectIVsChartColor`
    This global variable can be set by the user and changes the default golden color for perfect IVs. This is intended temporarily to allow testers to modify the color and chose a better one than the current default setting. However, as it has no functional influence on the game itself, i'm inclined to leave this variable in the game.

### Screenshots/Videos
golden-chart

@ edit: alternative chart with `window.perfectIVsChartColor = 0xFFD700;`
golden-chart2
as suggested by DerTapp from discord.

## How to test the changes?
During the starter selection, when toggling IVs, the altered color can appear, assuming all IVs are perfect.

Please use the following overrides for the `src/overrides.ts` file to generate any required IVs:
- replace `[PokeballType.MASTER_BALL]: 0,` with `[PokeballType.MASTER_BALL]: 1,` to get a masterball for a new run.
- replace `export const OPP_IVS_OVERRIDE: integer | integer[] = [];` with `export const OPP_IVS_OVERRIDE: integer | integer[] = 31;` to force any enemy pokemon to spawn with perfect IVs

Assign any color as a number to `window.perfectIVsChartColor` in the dev console to adjust the color of the chart. E.g. 0xFF0000, 0xFFFF00, etc.
A restart of the game isn't necessary, however a pokemon with perfect IV has to be reselected as the variable is only checked each time a pokemon is selected.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/ui/stats-container.ts
+++ b/src/ui/stats-container.ts
@@ -8,12 +8,6 @@ const ivChartStatCoordMultipliers = [ [ 0, -1 ], [ 0.825, -0.5 ], [ 0.825, 0.5 ]
 const ivChartStatIndexes = [0,1,2,5,4,3]; // swap special attack and speed
 const defaultIvChartData = new Array(12).fill(null).map(() => 0);
 
-declare global {
-  interface Window {
-    perfectIVsChartColor: integer;
-  }
-}
-
 export class StatsContainer extends Phaser.GameObjects.Container {
   private showDiff: boolean;
   private statsIvsCache: integer[];
@@ -69,13 +63,10 @@ export class StatsContainer extends Phaser.GameObjects.Container {
   }
 
   updateIvs(ivs: integer[], originalIvs?: integer[]): void {
-    this.updateValues(ivs, 31, ivs.reduce((res, iv) => res && iv === 31, true) ? window.perfectIVsChartColor ?? 0xFFBE00 : 0x98d8a0, originalIvs);
-  }
-
-  updateValues(ivs: integer[], scale : integer, color : integer, originalIvs?: integer[]): void {
     if (ivs) {
-      const ivChartData = new Array(6).fill(null).map((_, i) => [ (ivs[ivChartStatIndexes[i]] / scale) * ivChartSize * ivChartStatCoordMultipliers[ivChartStatIndexes[i]][0], (ivs[ivChartStatIndexes[i]] / scale) * ivChartSize * ivChartStatCoordMultipliers[ivChartStatIndexes[i]][1] ] ).flat();
+      const ivChartData = new Array(6).fill(null).map((_, i) => [ (ivs[ivChartStatIndexes[i]] / 31) * ivChartSize * ivChartStatCoordMultipliers[ivChartStatIndexes[i]][0], (ivs[ivChartStatIndexes[i]] / 31) * ivChartSize * ivChartStatCoordMultipliers[ivChartStatIndexes[i]][1] ] ).flat();
       const lastIvChartData = this.statsIvsCache || defaultIvChartData;
+      const perfectIVColor: string = getTextColor(TextStyle.SUMMARY_GOLD, false, (this.scene as BattleScene).uiTheme);
       this.statsIvsCache = ivChartData.slice(0);
 
       this.ivStatValueTexts.map((t: BBCodeText, i: integer) => {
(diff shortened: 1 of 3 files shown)


src/ui/stats-container.ts
```diff
@@ -8,6 +8,12 @@ const ivChartStatCoordMultipliers = [ [ 0, -1 ], [ 0.825, -0.5 ], [ 0.825, 0.5 ]
 const ivChartStatIndexes = [0,1,2,5,4,3]; // swap special attack and speed
 const defaultIvChartData = new Array(12).fill(null).map(() => 0);
 
+declare global {
+  interface Window {
+    perfectIVsChartColor: integer;
+  }
+}
+
```
brain-frog, line 16: this is unnecessary, it doesnt get set then gets null checked

src/ui/stats-container.ts
```diff
@@ -63,8 +69,12 @@ export class StatsContainer extends Phaser.GameObjects.Container {
   }
 
   updateIvs(ivs: integer[], originalIvs?: integer[]): void {
+    this.updateValues(ivs, 31, ivs.reduce((res, iv) => res && iv === 31, true) ? window.perfectIVsChartColor ?? 0xFFBE00 : 0x98d8a0, originalIvs);
```
brain-frog, line 72: no need to make this a 1 line function that only exists to call the function below it
also please use the existing color used for text of max IVs

brain-frog (changes requested):
I think this is a good idea but overthinking it
besides the overrides, this can be done simply by putting this at stats-container line 87:
```js
if (ivs.every((i) => i === 31)) {
  this.ivChart.fillColor = 0xe8e8a8;
} else {
  this.ivChart.fillColor = 0x98d8a0;
}
```

src/ui/stats-container.ts
prime-dialga: This is unfortunately required to prevent typedocs from throwing errors when using the global variable.

bennybroseph: a global variable just for a color?

prime-dialga: It was mainly intended for testing purposes, but there were also a few people in the dev chat who suggested different colors and thought that the perfect IV yellow is a bit hard to notice.
So i added this to allow for an easy override while the game is running, so testers could offer suggestions for other colors.
I'm personally inclined to leave this variable as a way for moders to adjust the color of the chart and perfect IV text (keeping both linked as requested by brain-frog), though i can see arguments against this.

bennybroseph: as someone coming from C++/C# the word global makes my skin crawl. It's probably not great here either, but I'm not positive

brain-frog: its also not required at all 
if you see my overall comment on the PR

prime-dialga: I also think it's rather weird and the code would run without this (as tested), but typedocs would throw errors if we try to access a variable it doesn't know for an object it was never declared.
And to declare a variable for `window` that seems to be the "correct" way, even if it's weird.

Anyway, I'll remove the variable for now.
While i think we should allow the user to adjust stuff like text color to improve accessability, adding this functionality randomly with this feature and only for this feature is rather weird.

prime-dialga:
Changes have been implemented as noted.
Though:
- The color change animation still remains (it looks better than simply setting the color to a new one)
- I'm parsing the color obtained from getTextColor so potential future changes to the color don't require additional changes to this code.

golden-chart3

EvasiveAce:
> Changes have been implemented as noted. Though:
> 
>     * The color change animation still remains (it looks better than simply setting the color to a new one)
> 
>     * I'm parsing the color obtained from getTextColor so potential future changes to the color don't require additional changes to this code.
> 
> 
> golden-chart3

I like the color better. Nice work.

brain-frog (approved):

PR #2020, opened 2024-06-09
[QoL] Revert input delay default and add setting

c4vv:
## What are the changes?

Users will notice the speed of the game returning to normal when a key is pressed and held. Additionally the setting to change the repeat input delay to allow for customization in the case that a longer delay is required.

The vast, vast majority of users will have their QoL improved by having the repeat input delay set back to where it was previously at 250ms. Many noticed that holding any key down had become much slower without knowing why. A change was implemented to double that number to 500ms to fix a problem for a very small subset of users experiencing double keypresses, while making the gameplay experience worse for almost the entire playerbase.

While making the setting for it, I figured I'd add the option to speed up this delay for players that prefer this delay to be even shorter than the default.

## What did change?

- Added `repeatInputDelay` as a setting with three options: `["Short","Normal","Long"]` translating to 125ms, 250ms, and 500ms respectively.
- Added the switch statement to set `repeatInputDelayMillis` to the corrseponding milliseconds.

### Screenshots/Videos

UI:
image

In action:

https://github.com/pagefaultgames/pokerogue/assets/132619649/38149012-6b1a-4db7-8947-68846f879ad5

You can see in the video, the delay is tested by simply holding down on the arrow key. At the current default of 500ms, it is painfully slow. Once it is set to 250ms it is visibly more responsive. And then the final setting of 125ms gives a much quicker scrolling experience. 

## How to test the changes?

One can perform the test pictured in the video, or simple play through a few rounds holding key presses to feel the difference.

## Checklist
- [x] Are the changes visual?
  
## Unit Tests
No idea what is going on with `npm run test` right now but these are the failing tests on main:
  main:
image

This branch setting the repeat input delay default back to 250ms fixes some of these failed tests, but there are still failing tests.

input_delay:
image

¯\\\_(ツ)\_/¯

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -154,6 +154,13 @@ export default class BattleScene extends SceneBase {
    * - 1 = 'Set' - The option to switch the active pokemon at the start of a battle will not display.
    */
   public battleStyle: integer = 0;
+  /**
+   * Determines the repeat input delay.
+   * - 0 = 'Short' or 125ms delay
+   * - 1 = 'Normal' or 250ms delay being the default
+   * - 2 = 'Long' or 500ms delay
+   */
+  public repeatInputDelay: integer = 1;
 
   /**
   * Defines whether or not to show type effectiveness hints
--- a/src/inputs-controller.ts
+++ b/src/inputs-controller.ts
@@ -48,7 +48,7 @@ export interface InterfaceConfig {
     custom?: MappingLayout;
 }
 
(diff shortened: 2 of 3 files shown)


Greenlamp2:
No.
the input delay was just a temporary fix for double inputs while i was working on the solution.
When this PR is merged, it will get back to normal and there should be no issue anymore.
https://github.com/pagefaultgames/pokerogue/pull/1936

PR #2024, opened 2024-06-10
[Move] Implement Core Enforcer

zacharied:
## What are the changes?
Implement the move [Core Enforcer](https://bulbapedia.bulbagarden.net/wiki/Core_Enforcer_(move)).

Core Enforcer lacked its secondary effect of suppressing the target's ability when used after the target in a turn.

## What did change?
Added a new `MoveEffectAttr` called `SuppressAbilitiesIfActedAttr`. This basically just calls `SuppressAbilitiesAttr` under the hood; since this move succeeds regardless of whether the suppression happens, we need to make a separate effect with no extra condition.

Also fixed the text for ability suppression to include `'s` at the start.

### Screenshots/Videos

Case 1: Core Enforcer is used after the Slaking moves. Slaking's ability Truant is suppressed and it can use a move on the next turn.

https://github.com/pagefaultgames/pokerogue/assets/6374275/1fbb9c50-ee01-4a4e-a1a6-6dac5c553bb6

Case 2: Core Enforcer is used before the Slaking moves. Slaking's ability is not suppressed, and it undergoes Truant on the next turn.

https://github.com/pagefaultgames/pokerogue/assets/6374275/e612306f-9265-49c0-83f3-cf759c2d0aa6

## How to test the changes?
Use the following player and opponent overrides:

```ts
/**
 * PLAYER OVERRIDES
 */

// default 5 or 20 for Daily
export const STARTING_LEVEL_OVERRIDE: integer = 5;
/**
 * SPECIES OVERRIDE
 */
export const STARTER_SPECIES_OVERRIDE: Species | integer = Species.SKRELP;
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.CORE_ENFORCER, Moves.SPLASH, Moves.SCARY_FACE, Moves.NONE];

/**
 * OPPONENT / ENEMY OVERRIDES
 */

export const OPP_SPECIES_OVERRIDE: Species | integer = Species.SLAKING;
export const OPP_LEVEL_OVERRIDE: number = 10;
export const OPP_ABILITY_OVERRIDE: Abilities = Abilities.TRUANT;
export const OPP_STATUS_OVERRIDE: StatusEffect = null;
export const OPP_MOVESET_OVERRIDE: Array<Moves> = [Moves.SPLASH, Moves.SPLASH, Moves.SPLASH, Moves.SPLASH];
```

Observe that using Core Enforcer suppresses Slaking's ability because Skrelp moves after it. Then, reload and use Scary Face until Skrelp outspeeds Slaking. Observe that Core Enforcer does not suppress Slaking's ability because Skrelp moves before it. Repeat this in a doubles battle to observe the effects on multiple targets.

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -5115,12 +5115,18 @@ export class SuppressAbilitiesAttr extends MoveEffectAttr {
 }
 
 /**
- * Adds a secondary effect to a move which applies the effects of {@linkcode SuppressAbilitiesAttr} to the target
- * if it has already moved this turn.
+ * Applies the effects of {@linkcode SuppressAbilitiesAttr} if the target has already moved this turn.
  * @extends MoveEffectAttr
  * @see {@linkcode Moves.CORE_ENFORCER} (the move which uses this effect)
  */
 export class SuppressAbilitiesIfActedAttr extends MoveEffectAttr {
+  /**
+   * If the target has already acted this turn, apply a {@linkcode SuppressAbilitiesAttr} effect unless the
+   * abillity cannot be suppressed. This is a secondary effect and has no bearing on the success or failure of the move.
+   *
+   * @returns True if the move occurred, otherwise false. Note that true will be returned even if the target has not
+   * yet moved or if the target's abiilty is un-suppressable.
+   */
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
     if (!super.apply(user, target, move, args)) {
       return false;
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -5114,6 +5114,12 @@ export class SuppressAbilitiesAttr extends MoveEffectAttr {
   }
 }
 
+/**
+ * Adds a secondary effect to a move which applies the effects of {@linkcode SuppressAbilitiesAttr} to the target
+ * if it has already moved this turn.
+ * @extends MoveEffectAttr
+ * @see {@linkcode Moves.CORE_ENFORCER} (the move which uses this effect)
+ */
 export class SuppressAbilitiesIfActedAttr extends MoveEffectAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
     if (!super.apply(user, target, move, args)) {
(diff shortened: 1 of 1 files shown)


src/data/move.ts
```diff
@@ -5104,7 +5104,7 @@ export class SuppressAbilitiesAttr extends MoveEffectAttr {
 
     target.summonData.abilitySuppressed = true;
 
-    target.scene.queueMessage(getPokemonMessage(target, " ability\nwas suppressed!"));
+    target.scene.queueMessage(getPokemonMessage(target, "'s ability\nwas suppressed!"));
```
Tempo-anon, line 5107: Do you need a `\` in front of the `'`?

src/data/move.ts
```diff
@@ -5114,6 +5114,23 @@ export class SuppressAbilitiesAttr extends MoveEffectAttr {
   }
 }
 
+export class SuppressAbilitiesIfActedAttr extends MoveEffectAttr {
```
Tempo-anon, line 5117: Please add documentation to new classes/functions

src/data/move.ts
zacharied: Doesn't seem like it.

src/data/move.ts
zacharied: Done.

src/data/move.ts
```diff
@@ -5114,6 +5114,29 @@ export class SuppressAbilitiesAttr extends MoveEffectAttr {
...
 }
 
+/**
+ * Adds a secondary effect to a move which applies the effects of {@linkcode SuppressAbilitiesAttr} to the target
+ * if it has already moved this turn.
+ * @extends MoveEffectAttr
+ * @see {@linkcode Moves.CORE_ENFORCER} (the move which uses this effect)
+ */
+export class SuppressAbilitiesIfActedAttr extends MoveEffectAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
```
Tempo-anon, line 5124: Can you also add doc to the apply function? After that it's good to go

zacharied:
I've added some moves as a proof of concept to another branch. For example, here's Throat Chop and Taunt: https://github.com/pagefaultgames/pokerogue/commit/94cfb53c243ecf37bf389a2dcd5ccde1f80bbd67

zacharied:
All set.

Tempo-anon (approved):

PR #2027, opened 2024-06-10
[Bug] Fix Clear terrains upon Trainer Battle

### LINKED ISSUE #2018: [BUG] Psychic Surge terrain not removed and re-applied when entering trainer battle
**Describe the bug**

My Delphox's Psychic Surge passive creates psychic terrain during wild pokemon battles, and when I enter a trainer battle the psychic terrain is not removed so that Psychic Surge can trigger again.

**To Reproduce**

1. Have Delphox with Psychic Surge passive unlocked
2. Activate Psychic Surge during wild battle
3. Enter trainer battle before the Psychic Terrain dissipates.

**Expected behavior**

Trainer battles start from a "clean slate", removing all battlefield effects.

**Screenshots / Videos**

image
Turn 1 against a Battle Girl

**Device**

Chrome on PC

**Additional context**

Session data: [sessionData1_Terrorforge-psychicTerrain.zip](https://github.com/user-attachments/files/15754165/sessionData1_Terrorforge-psychicTerrain.zip)

Arxalc:
## What are the changes?

Terrains would persist over to trainer battles. This line would clear any terrains upon entering a Trainer Battle.

closes #2018
### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/63990624/1eb9cee7-bc56-49b2-b90e-fd834f33ca9d

## How to test the changes?

Override pokemon ability to Psychic Terrain. Activate it on wave 4. Observe it ends before going into the trainer battle on wave 5.
## Checklist
    - [Well... none of the errors have to do with this] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?

### DIFF
--- a/src/field/arena.ts
+++ b/src/field/arena.ts
@@ -621,6 +621,14 @@ export class Arena {
     }
   }
 
+  /**
+   * Clears terrain and arena tags when entering new biome or trainer battle.
+   */
+  resetArenaEffects(): void {
+    this.trySetTerrain(TerrainType.NONE, false, true);
+    this.removeAllTags();
+  }
+
   preloadBgm(): void {
     this.scene.loadBgm(this.bgm);
   }
--- a/src/field/arena.ts
+++ b/src/field/arena.ts
@@ -621,9 +621,7 @@ export class Arena {
     }
   }
 
-  /**
-   * Clears terrain and arena tags when entering new biome or trainer battle.
-   */
+  /** Clears terrain and arena tags when entering new biome or trainer battle. */
   resetArenaEffects(): void {
     this.trySetTerrain(TerrainType.NONE, false, true);
     this.removeAllTags();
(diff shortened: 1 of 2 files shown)


RNGezeus:
> This does not clear weather. I assume it intentionally doesn't because it seems odd if you were in the middle of a sandstorm and it just randomly stopped when you fight this dude.

Is terrain that much different in that regard? To me both sound like global area effects slightly bend reality and follow the user. What I can give is that the vibe of terrains feel they are bonded stronger to the user and thus more likely to stop if the connection breaks from recall. Though you can freely switch during combat while terrains up so eh idk.

They also both provide a substantial power buff if you are the user who sets them up so it's weird to make a distinction between Terrain users and Weather users. Even weirder since then it might be the final remaining arena effect that doesn't reset.

DustinLin:
Not sure if this is related, but PokeRogue does make the distinction that stat changes are preserved over battles. This could be something for the balance team to make a decision on

src/field/arena.ts
```diff
@@ -621,6 +621,11 @@ export class Arena {
     }
   }
 
+  resetArenaEffects(): void {
```
DustinLin, line 624: may be good to include comments/documentation on this new function for continuity

OrangeRed (approved):

PR #2044, opened 2024-06-10
[Bug] Make speed ties seed based on current turn

Xavion3:
## What are the changes?
Speed ties should no longer vary based on reload.

Speed ties currently resolve differently based on how long it's been since a reload.

## What did change?
Previously, speed ties used the default wave rng offset which can vary based on a reload. This changes them to instead seed based on the current turn, making it consistent across reloads.

## How to test the changes?
Turn the rng debugger on then start a battle in a new save, reload in a new tab/session and retry the same battle. Specifically look at the RNG call made after TurnStartPhase, this should be different without this patch and the same with this patch.

## Checklist

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -661,7 +661,14 @@ export abstract class FieldPhase extends BattlePhase {
     const enemyField = this.scene.getEnemyField().filter(p => p.isActive()) as Pokemon[];
 
     // We shuffle the list before sorting so speed ties produce random results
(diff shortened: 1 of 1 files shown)


Tempo-anon (approved):
Are there any other parts in the code where this needs to be done?

Xavion3:
Unfortunately tracking down things like this is difficult as we don't really have great static analysis tools, so I'm sorta just going one step at a time, patching an RNG inconsistency, waiting till we get a report of another inconsistency, and just repeating.

PR #2047, opened 2024-06-10
[Bug] Adjusts bug bite/pluck to remove berry pouch preservation on opponent berries

schmidtc1:
## What are the changes?

Currently, bug bite/pluck has a chance to preserve the berry in an opponents inventory if the user has berry pouches, potentially creating a nearly endless supply. With these changes, bug bite/pluck will no longer coincide with berry pouch functionality.

Bug bite/pluck has a chance to preserve the berry in an opponents inventory if the user has berry pouches, potentially creating a nearly endless supply.

#1315

Balances bug bite/pluck with berry pouch better to avoid infinite supplies of berries.

## What did change?

Removes the StealEatBerryAttr class extension of EatBerryAttr and removes berry pouch preservation when stealing an opponents berry.

Before:
StealEatBerryAttr extended EatBerryAttr and the super class (EatBerryAttr) was called to eat the berry, gain its effects, and check for berry pouch preservation.
After:
StealEatBerryAttr extends only MoveEffectAttr and no longer calls the super class to eat the berry and no longer checks for berry pouch preservation.

## How to test the changes?

Tested using the `src/overrides.ts` file. 
export const STARTING_HELD_ITEMS_OVERRIDE: Array<ModifierOverride> = [{name: "BERRY", count: 3, type: BerryType.SITRUS}, {name: "BERRY_POUCH", count: 3}];
export const OPP_HELD_ITEMS_OVERRIDE: Array<ModifierOverride> = [{name: "BERRY", count: 3, type: BerryType.SITRUS}];

Recommend testing with no berry pouches vs. three berry pouches to see the effect.

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -1813,7 +1813,7 @@ export class EatBerryAttr extends MoveEffectAttr {
  * @param args Unused
  * @returns {boolean} true if the function succeeds
  */
-  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean | Promise<boolean> {
```
flx-sta, line 1816: Changing the return type to be a non-promise or a Promise seems absolutely confusing and inconsistent.
why is the Promise necessary now?

src/data/move.ts
```diff
@@ -1880,17 +1882,37 @@ export class StealEatBerryAttr extends EatBerryAttr {
     if (heldBerries.length) { // if the target has berries, pick a random berry and steal it
       this.chosenBerry = heldBerries[user.randSeedInt(heldBerries.length)];
 
-      if (this.chosenBerry.stackCount === 1) { // remove modifier if its the last berry
+      if (this.chosenBerry === undefined) { // if no berry has been provided, pick a random berry from their inventory
```
flx-sta, line 1885:
```suggestion
      if (!this.chosenBerry) { // if no berry has been provided, pick a random berry from their inventory
```
don't check for `undefined` UNLESS you really want to exclude `null`

src/data/move.ts
```diff
@@ -1880,17 +1882,37 @@ export class StealEatBerryAttr extends EatBerryAttr {
...
+        const heldBerries = this.getTargetHeldBerries(target);
+        if (heldBerries.length <= 0) {
+          return false;
+        }
+        this.chosenBerry = heldBerries[user.randSeedInt(heldBerries.length)];
+      }
+
+      user.scene.queueMessage(getPokemonMessage(user, ` stole and ate\n${target.name}'s ${this.chosenBerry.type.name}!`));
+
+      if (!--this.chosenBerry.stackCount) {
```
flx-sta, line 1895: don't do decrementing inside of an if-clause.
This is an abomination for readability

src/data/move.ts
```diff
@@ -1880,17 +1882,37 @@ export class StealEatBerryAttr extends EatBerryAttr {
...
-      if (this.chosenBerry.stackCount === 1) { // remove modifier if its the last berry
+      if (this.chosenBerry === undefined) { // if no berry has been provided, pick a random berry from their inventory
+        const heldBerries = this.getTargetHeldBerries(target);
+        if (heldBerries.length <= 0) {
+          return false;
+        }
+        this.chosenBerry = heldBerries[user.randSeedInt(heldBerries.length)];
+      }
+
+      user.scene.queueMessage(getPokemonMessage(user, ` stole and ate\n${target.name}'s ${this.chosenBerry.type.name}!`));
```
flx-sta, line 1893: We should probably localize (i18n) this, not more hard-coded English texts please

flx-sta (changes requested):

src/data/move.ts
schmidtc1: Leftover code from previous testing on the EatBerryAttr class. Removed.

src/data/move.ts
schmidtc1: Adjusted this code for readability.

src/data/move.ts
schmidtc1: Attempted localization steps. Created translation entries under "stealEatBerry" in the battle.ts file of each locale.

src/data/move.ts
```diff
@@ -1880,17 +1882,38 @@ export class StealEatBerryAttr extends EatBerryAttr {
     if (heldBerries.length) { // if the target has berries, pick a random berry and steal it
       this.chosenBerry = heldBerries[user.randSeedInt(heldBerries.length)];
 
-      if (this.chosenBerry.stackCount === 1) { // remove modifier if its the last berry
+      if (!this.chosenBerry) { // if no berry has been provided, pick a random berry from their inventory
+        const heldBerries = this.getTargetHeldBerries(target);
+        if (heldBerries.length <= 0) {
+          return false;
+        }
+        this.chosenBerry = heldBerries[user.randSeedInt(heldBerries.length)];
```
flx-sta, line 1890: wouldn't this possibly cause an `index out of bounds` error?
Because `heldBerries[heldBerries.length]` is 1 index too high (0 index starts)

schmidtc1: Correct me if I am wrong: I believe randSeedInt is non-inclusive on the range, meaning it is up to, not including heldBerries.length. 
I have not seen any index out of bounds errors when testing so far.

schmidtc1:
Before:
https://github.com/pagefaultgames/pokerogue/assets/62030095/e4e7374e-38b5-49c6-83b8-3ddca29bf97a

After:
https://github.com/pagefaultgames/pokerogue/assets/62030095/a9c842d2-d124-4d71-8fa9-f20b9f5a5187

schmidtc1:
Sticky hold example:

https://github.com/pagefaultgames/pokerogue/assets/62030095/d0de7482-1f5b-4be2-9a42-6619661f67c5

schmidtc1:
Example with multi lens bug bite:

https://github.com/pagefaultgames/pokerogue/assets/62030095/3e19c7e7-8621-4437-bcc7-d2e5891d7fac

src/locales/fr/battle.ts
```diff
@@ -59,6 +59,7 @@ export const battle: SimpleTranslationEntries = {
   "wildPokemonWithAffix": "{{pokemonName}} sauvage",
   "foePokemonWithAffix": "{{pokemonName}} ennemi",
   "useMove": "{{pokemonNameWithAffix}} utilise\n{{moveName}} !",
+  "stealEatBerry": "{{pokemonName}} stole and ate\n{{targetName}}'s {{berryName}}!",
```
Adri1, line 62:
```suggestion
  "stealEatBerry": "{{pokemonName}} vole et mange\nla {{berryName}} de {{targetName}} !",
```

Adri1 (commented):
French translation

src/locales/es/battle.ts
```diff
@@ -61,6 +61,7 @@ export const battle: SimpleTranslationEntries = {
   "useMove": "¡{{pokemonNameWithAffix}} usó {{moveName}}!",
   "drainMessage": "¡{{pokemonName}} tuvo su\nenergía absorbida!",
   "regainHealth": "¡{{pokemonName}} recuperó\nPS!",
+  "stealEatBerry": "{{pokemonName}} stole and ate\n{{targetName}}'s {{berryName}}!",
```
InnocentGameDev, line 64: "¡{{pokemonName}} robó la {{berryName}}\n de {{targetName}} y se la comió!"

src/locales/es/battle.ts
```diff
@@ -61,7 +61,7 @@ export const battle: SimpleTranslationEntries = {
   "useMove": "¡{{pokemonNameWithAffix}} usó {{moveName}}!",
   "drainMessage": "¡{{pokemonName}} tuvo su\nenergía absorbida!",
   "regainHealth": "¡{{pokemonName}} recuperó\nPS!",
-  "stealEatBerry": "{{pokemonName}} stole and ate\n{{targetName}}'s {{berryName}}!",
+  "¡{{pokemonName}} robó la {{berryName}}\n de {{targetName}} y se la comió!"
```
Adri1, line 64:
```suggestion
  "¡{{pokemonName}} robó la {{berryName}}\nde {{targetName}} y se la comió!",
```

Adri1 (commented):
Missing "stealEatBerry" and coma here (and removed unnecessary space)

src/locales/es/battle.ts
schmidtc1: Should be resolved in latest commits

src/locales/ko/battle.ts
```diff
@@ -61,6 +61,7 @@ export const battle: SimpleTranslationEntries = {
   "useMove": "{{pokemonNameWithAffix}}의 {{moveName}}!",
   "drainMessage": "{{pokemonName}}[[로]]부터\n체력을 흡수했다!",
   "regainHealth": "{{pokemonName}}[[는]]\n기력을 회복했다!",
+  "stealEatBerry": "{{pokemonName}} stole and ate\n{{targetName}}'s {{berryName}}!",
```
EnochG1, line 64:
```suggestion
  "stealEatBerry": "{{pokemonName}}[[가]]\n{{targetName}}의 {{berryName}}[[를]] 빼앗아 먹었다!",
```

Tempo-anon (dismissed):

src/data/move.ts
```diff
@@ -1986,17 +1988,42 @@ export class StealEatBerryAttr extends EatBerryAttr {
     if (heldBerries.length) { // if the target has berries, pick a random berry and steal it
       this.chosenBerry = heldBerries[user.randSeedInt(heldBerries.length)];
 
-      if (this.chosenBerry.stackCount === 1) { // remove modifier if its the last berry
+      if (!this.chosenBerry) { // if no berry has been provided, pick a random berry from their inventory
```
Tempo-anon, line 1991: Will this ever be false?

src/data/move.ts
```diff
@@ -1986,17 +1988,42 @@ export class StealEatBerryAttr extends EatBerryAttr {
...
+          return false;
+        }
+        this.chosenBerry = heldBerries[user.randSeedInt(heldBerries.length)];
+      }
+
+      const message = i18next.t("battle:stealEatBerry", {pokemonName: user.name, targetName: target.name, berryName: this.chosenBerry.type.name});
+      user.scene.queueMessage(message);
+
+
+
```
Tempo-anon, line 2003: nit: excess lines

src/data/move.ts
```diff
@@ -1986,17 +1988,42 @@ export class StealEatBerryAttr extends EatBerryAttr {
...
+      } else {
+        this.chosenBerry.stackCount--;
       }
       target.scene.updateModifiers(target.isPlayer());
 
-      user.scene.queueMessage(getPokemonMessage(user, ` stole and ate\n${target.name}'s ${this.chosenBerry.type.name}!`));
-      return super.apply(user, user, move, args);
+      getBerryEffectFunc(this.chosenBerry.berryType)(user); // user eats the berry
+
+      this.chosenBerry = undefined;
```
Tempo-anon, line 2013: Does this need to be set to undefined after?

src/data/move.ts
```diff
@@ -1961,9 +1961,11 @@ export class EatBerryAttr extends MoveEffectAttr {
  *  Attribute used for moves that steal a random berry from the target. The user then eats the stolen berry.
  *  Used for Pluck & Bug Bite.
  */
-export class StealEatBerryAttr extends EatBerryAttr {
+export class StealEatBerryAttr extends MoveEffectAttr {
+  protected chosenBerry: BerryModifier;
   constructor() {
-    super();
+    super(true, MoveEffectTrigger.HIT);
+    this.chosenBerry = undefined;
```
OrangeRed, line 1968: Can you instead just add a method to EatBerryAttr that accepts a preserve boolean and handles the berry eating logic. I don't see a need to logically separate stealing a berry and eating berry (especially when some logic is reused) when instead you can make an inherited class method.

OrangeRed (changes requested):

src/data/move.ts
```diff
@@ -1986,17 +1988,42 @@ export class StealEatBerryAttr extends EatBerryAttr {
...
       }
       target.scene.updateModifiers(target.isPlayer());
 
-      user.scene.queueMessage(getPokemonMessage(user, ` stole and ate\n${target.name}'s ${this.chosenBerry.type.name}!`));
-      return super.apply(user, user, move, args);
+      getBerryEffectFunc(this.chosenBerry.berryType)(user); // user eats the berry
+
+      this.chosenBerry = undefined;
+
+      applyAbAttrs(HealFromBerryUseAbAttr, target, new Utils.BooleanHolder(false));
```
OrangeRed, line 2015: This logic could be moved into a shared class method and called on demand. There is no need to separate StealBerry and EatBerry if you do that.

src/data/move.ts
schmidtc1: StealEatBerry and EatBerry were separated on the original bug bite implementation, I can work on merging the logic if that is preferred

OrangeRed: > StealEatBerry and EatBerry were separated on the original bug bite implementation, I can work on merging the logic if that is preferred

They were not separated. The former inherits the latter and truly separating them like you do in this PR just increases how much identical code you have to write

schmidtc1: You're right I misremembered. I'll work on merging the logic.

src/data/move.ts
schmidtc1: Should be resolved in new commit that combines StealEatBerryAttr and EatBerryAttr

src/data/move.ts
schmidtc1: Resolved in latest commit

src/data/move.ts
schmidtc1: No, resolved.

src/data/move.ts
schmidtc1: Combined StealBerry and EatBerry into EatBerryAttr. Should be resolved.

src/data/move.ts
```diff
@@ -1912,51 +1912,52 @@ export class EatBerryAttr extends MoveEffectAttr {
...
+      this.eatBerry(target);
+      return true;
     }
-
-    this.chosenBerry = undefined;
-
-    applyAbAttrs(HealFromBerryUseAbAttr, target, new Utils.BooleanHolder(false));
-
-    return true;
+    return false;
```
OrangeRed, line 1938: I prefer early returns over nesting, so can you invert this to like it was before this change.
```ts
      if (heldBerries.length <= 0) {
        return false;
      }

     // ...

    return true;
```

OrangeRed (changes requested):
Overall this looks like what I was expecting. Can you just invert the conditional like I mentioned inside both EatBerryAbAttr and StealBerryAbAttr

src/data/move.ts
schmidtc1: Should be resolved in latest commit

OrangeRed (approved):

PR #2051, opened 2024-06-10
[Move] Use BattlerTag for move-disabling effects

zacharied:
## What are the changes?
This PR implements a new `BattlerTag` for move-disabling effects such as those done by Disable, Taunt, Throat Chop, etc.. It also replaces the implementation of the move Disable and the ability Cursed Body with a new implementation using this tag.

### The current approach

Disabling is a somewhat common effect of moves, but the current implementation of disabling has a few issues:

* Only one move can be disabled per Pokemon at a time, and very few disable effects actually work this way in-game
* No way to introspect on source of the disable
* Introduces more state management to the Pokemon object
* We have to modify `phases.ts` to implement new disabling effects

#751 is an example of a PR that attempts to solve this problem by extending the current approach, using Throat Chop as a baseline. It carries with it the same issues as the current implementation of Disable, namely in that it adds more state for the Pokemon object to deal with. Incorrect behavior can be observed in #751 if a pokemon were to learn a sound-based move after getting hit by Throat Chop; the sound-based move would execute successfully because the `disabledMoves` array does not reflect the actual disabled moves according to the rules. This would be incorrect, as all sound-based moves should be disabled regardless of if they were originally in the moveset.

### Related work-in-progress

Other PRs implement new Tags, which feels safer to me. #1121 implements Imprison correctly, but my concern is over tacking more conditions onto `PokemonMove.isUsable`. Every condition added to that function also means more code spread elsewhere to account for displaying the move-specific messages and other such effects. I am concerned that by the time we have all the moves implemented, the function will become incredibly long while simultaneously requiring its callers to take on the work of presenting the correct messages. With that said, a lot of work has gone into verifying the correctness of Imprison in #1121, so I don't want to overstep any boundaries of what seems to be the more mature PR.

Likewise, there is #1153, which is an overall agreeable solution, but I'm unsure if a PR with that many features will get merged anytime soon. An intermediate PR that lays the groundwork for those moves, like this PR, might be necessary to get that through.

To avoid invalidating the work of these existing commits, my idea was to introduce some system-level changes that will make move-disabling effects more consistent and easier to implement. I haven't checked the exact merges myself, but I think things like #1121 and #1153 could coexist with and benefit from these changes they were to adapt this tag system. But since those commits are further along the pipeline, I'd also be happy to wait for them to go mainline and then rebase my changes onto those implementations. I'm open to all discussion!

### This PR

In this PR, to implement new disable types, you only need to extend `DisablingBattlerTag` and implement its `isMoveDisabled` to match the filter applied by the move. Then, any Pokemon that has that tag will handle disabling the matching moves -- no need to edit `PokemonMove.isUsable` or `MovePhase.canMove` or anything like that.

## What did change?

Add abstract class `DisablingBattlerTag` that serves as the base for all tags that disable moves. The tag lapses in `PRE_MOVE` to cancel any disable move getting used. It also lapses at `TURN_END` to remove the tag. Also updated menus and AI command selection to prevent players and enemies from selecting disabled moves.

Add a battler tag `DisabledTag` that represents Disabling as applied by the move Disable and ability Cursed Body. Also remove `DisableMoveAttr` (the old implementation of the Disable move) and replace Disable's effect with adding a `DisabledTag`.

Add battler tag type `DISABLED` which maps to `DisabledTag`. This tag has been given the same `targetBenefit` as was previously bestowed by `DisableMoveAttr`.

`disabledMove` and `disabledTurns` fields removed from `PokemonSummonData`. Any time those fields would be set, instead an instance of tag `DisabledTag` is applied. Because this uses the tag system, all the code that had to manually manipulate/check `disabledTurns` has been removed as well.

Add new functions `Pokemon.getDisablingTag(Moves)` and `Pokemon.isMoveDisabled(Moves)` that find the tag that disables a given move (if it exists), and checks whether the Pokemon has a tag that disables the given move, respectively.

## How to test the changes?

I used the following overrides for testing beyond the unit tests.

```ts
/**
 * OVERALL OVERRIDES
 */

export const SINGLE_BATTLE_OVERRIDE: boolean = true;
export const STARTING_WAVE_OVERRIDE: integer = 1;
// default 1000
  active: false,
  pokeballs: {
    [PokeballType.POKEBALL]: 5,
    [PokeballType.GREAT_BALL]: 0,
    [PokeballType.ULTRA_BALL]: 0,
    [PokeballType.ROGUE_BALL]: 0,
  }
};

/**
 * PLAYER OVERRIDES
 */

export const STARTING_LEVEL_OVERRIDE: integer = 10;

export const STARTER_SPECIES_OVERRIDE: Species | integer = Species.BIDOOF;
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.DISABLE, Moves.SPLASH, Moves.GROWL, Moves.TAIL_WHIP];

export const OPP_SPECIES_OVERRIDE: Species | integer = Species.BIDOOF;
export const OPP_LEVEL_OVERRIDE: number = 1;
export const OPP_ABILITY_OVERRIDE: Abilities = Abilities.NONE;
export const OPP_STATUS_OVERRIDE: StatusEffect = StatusEffect.NONE;
export const OPP_MOVESET_OVERRIDE: Array<Moves> = [Moves.DISABLE, Moves.SPLASH, Moves.NONE, Moves.NONE];
```

## Checklist
- [ ] Are the changes visual?


zacharied:
I've added some moves as a proof of concept to another branch. For example, here's Throat Chop and Taunt: https://github.com/pagefaultgames/pokerogue/commit/94cfb53c243ecf37bf389a2dcd5ccde1f80bbd67

zacharied:
Rewrote OP to focus more on integrating the variety of disparate Disable effect commits.

zacharied:
Converted to draft due to significant overlap with other PRs.

DayKev:
Imo it'd be best if this PR was merged and then the other PRs were updated to use the new, non-spaghetti code.

audriddax:
I'm just going through this, but as the developer of #1121, I have no objection to cleaning this up. I kinda touched on it in some of my back and forth in that one, I wasn't thrilled with what it required to implement. Not only did I talk about how I was hesitant to update code on something that basically affected every move, but on my end I also was seeing exactly what you are trying to address based on your summary here.

I'm all for a change that creates a better framework for my code, I'll gladly update it rather than add extra code to an "unrelated" section just for my move. I'm actually glad I had to step away for a few weeks and haven't had my code committed if it means we're getting a better framework!

audriddax:
> Imo it'd be best if this PR was merged and then the other PRs were updated to use the new, non-spaghetti code.

I agree with this, don't try to update my move, I'll do that gladly!

zacharied:
WTF did i do to make those tests fail 😭

DayKev:
One of the overrides changed how it worked (instead of `SINGLE_BATTLE_OVERRIDE`/`DOUBLE_BATTLE_OVERRIDE` there's just `BATTLE_TYPE_OVERRIDE: "double" | "single" | null`).
Also there are helper functions to change overrides now (`game.override.x()`) that you can make use of.

torranx:
Are you still working on this @zacharied ?

zacharied:
> Hi, thank you for working on this. I was wondering if this was able to be extended to multiple categories / attrs of moves. I'm currently working on Heal Block right now and was hoping to try disabling moves based on attributes.

Yep, the Disable tag takes a filter function as argument, so you can disable based on any criteria you want.

zacharied:
Okay, just need to localize a couple strings and fix these test failures.

For the PURIFY failure, it seems like the mons are moving out of order. The Magikarp uses Splash when it's supposed to be the Pyukumuku using Purify. I can fix this by forcing Pyukumuku to be faster -- seems like an error with the test file?

Still investigating the HYPER BEAM failure, that's a weird one.

DayKev:
Fixed the test bugs in #3449

zacharied:
@DayKev looks like Hyper Beam is still RNG?

DayKev:
I wish Github would give the console log for the test... guess I'll just have to run the test a whole bunch.

DayKev:
Oh I guess you did break something..? Looking at the test in this branch after pulling in the latest beta changes, the Tackle never happens after the Hyper Beam. I guess it's something to do with the recharge part of hyper beam not being reset properly?
```
[ 'Magikarp used !' ]
[ '%cStart Phase MessagePhase', 'color:green;' ]
[ 'But it failed!' ]
```

zacharied:
> Oh I guess you did break something..? Looking at the test in this branch after pulling in the latest beta changes, the Tackle never happens after the Hyper Beam. I guess it's something to do with the recharge part of hyper beam not being reset properly?
> 
> ```
> [ 'Magikarp used !' ]
> [ '%cStart Phase MessagePhase', 'color:green;' ]
> [ 'But it failed!' ]
> ```

Oh, my bad then. Gonna dig into it.

src/data/battler-tags.ts
```diff
@@ -92,6 +92,110 @@ export interface TerrainBattlerTag {
...
+      phase.fail();
+    } else if (lapseType === BattlerTagLapseType.TURN_END) {
+      // On turn end, subtract from lifetime and remove this tag if 0
+      if (!super.lapse(pokemon, lapseType)) {
+        return false;
+      }
+    }
+
+    return true;
+  }
```
innerthunder, line 127:
```suggestion
  lapse(pokemon: Pokemon, lapseType: BattlerTagLapseType): boolean {
    if (lapseType === BattlerTagLapseType.PRE_MOVE) {
      // Cancel the affected pokemon's selected move
      const phase = pokemon.scene.getCurrentPhase() as MovePhase;
      const move = phase.move;

      if (this.moveIsDisabled(move.moveId)) {
        pokemon.scene.queueMessage(this.interruptedText(pokemon, move.moveId));
        phase.fail();
      }
      return true;
    }
    // On turn end, subtract from lifetime and remove this tag if 0
    return super.lapse(pokemon, lapseType);
  }
```
Some style changes, should still function the same.

src/data/battler-tags.ts
```diff
@@ -92,6 +92,110 @@ export interface TerrainBattlerTag {
...
+      // Cancel the affected pokemon's selected move
+      const phase = pokemon.scene.getCurrentPhase() as MovePhase;
+      const move = phase.move;
+
+      if (!this.moveIsDisabled(move.moveId)) {
+        return true;
+      }
+
+      pokemon.scene.queueMessage(this.interruptedText(pokemon, move.moveId));
+      phase.fail();
```
innerthunder, line 118: I think this should be `phase.cancel()` instead, so the move doesn't use PP when cancelled this way.

src/data/battler-tags.ts
```diff
@@ -92,6 +92,110 @@ export interface TerrainBattlerTag {
...
+  abstract interruptedText(pokemon: Pokemon, move: Moves): string;
+}
+
+/**
+ * Tag representing the "disabling" effect performed by {@linkcode Moves.DISABLE} and {@linkcode Abilities.CURSED_BODY}.
+ * When the tag is added, the last used move of the tag holder is set as the disabled move.
+ */
+export class DisabledTag extends DisablingBattlerTag {
+  /** The move being disabled. Gets set when {@linkcode onAdd} is called for this tag. */
+  public moveId: integer = 0;
```
innerthunder, line 143:
```suggestion
  public moveId: Moves = Moves.NONE;
```

src/data/battler-tags.ts
```diff
@@ -92,6 +92,110 @@ export interface TerrainBattlerTag {
...
+      return false;
+    }
+
+    if (this.moveId === 0) {
+      console.warn(`attempt to disable move ID 0 on ${pokemon}`);
+      return false;
+    }
+
+    return true;
+  }
```
innerthunder, line 164: Can be removed; we probably don't want to keep the `console.warn(...)`

src/data/battler-tags.ts
```diff
@@ -92,6 +92,110 @@ export interface TerrainBattlerTag {
...
+
+  /**
+   * Ensures that move history exists and has a valid move. If so, sets the {@link moveId} and shows a message.
+   * Otherwise, something has gone wrong, so the move ID will not get assigned and this tag will get removed next turn.
+   */
+  override onAdd(pokemon: Pokemon): void {
+    super.onAdd(pokemon);
+
+    const history = pokemon.getLastXMoves();
```
innerthunder, line 173: Can you filter out `virtual` moves here (so moves generated from Metronome, etc. aren't considered)?

innerthunder (changes requested):
Will check out the test code later 🔍

src/data/battler-tags.ts
DayKev: That'll be removed when #3380 gets merged.

innerthunder: So `Move.virtual` is getting removed? I checked the PR and all I saw that was removed was the `IGNORE_VIRTUAL` flag

src/test/moves/disable.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    game.phaseInterceptor.restoreOg();
+  });
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override.battleType("single");
+    game.override.ability(Abilities.NONE);
+    game.override.enemyAbility(Abilities.NONE);
+    game.override.moveset([Moves.DISABLE, Moves.SPLASH, Moves.SPLASH, Moves.SPLASH]);
+    game.override.enemyMoveset([Moves.SPLASH,Moves.SPLASH,Moves.SPLASH,Moves.SPLASH]);
```
innerthunder, line 48: Use one of the test utils instead:
```suggestion
    game.override.enemyMoveset(ALL_SPLASH);
```

src/test/moves/disable.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    game.override.starterSpecies(Species.PIKACHU);
+    game.override.enemySpecies(Species.SHUCKLE);
+  });
+
+  it("DISABLE fails if enemy has no move history", async() => {
+    // Player goes first
+    await game.startBattle();
+    expect(game.scene.getParty()[0].stats[Stat.SPD]).toBeGreaterThan(game.scene.getEnemyParty()[0].stats[Stat.SPD]);
```
innerthunder, line 56: Instead of using `game.scene.get<Enemy>Party()[0]` everywhere, get a reference to the Pokemon with `game.scene.getPlayerPokemon` or `game.scene.getEnemyPokemon`, e.g.
```ts
const pikachu = game.scene.getPlayerPokemon()!; // "!" to avoid strict-null issues
const shuckle = game.scene.getEnemyPokemon()!;

expect(pikachu.stats[Stat.SPD]).toBeGreaterThan(shuckle.stats[SPD])
/* ... */
```

src/test/moves/disable.test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    _useMove();
+
+    await game.phaseInterceptor.runFrom(EnemyCommandPhase).to(TurnInitPhase);
+    expect(game.scene.getParty()[0].getMoveHistory().at(0)!.result).toBe(MoveResult.FAIL);
```
innerthunder, line 61: Might want to add another `expect` to make sure `getMoveHistory` isn't empty

src/test/moves/disable.test.ts
innerthunder, line 1: Some more tests I'd like to see:
- Disable fails if the target's last move was Struggle (you can use overrides to change the enemy's moveset to `Array(4).fill(Moves.STRUGGLE)` for this)
- Disable disables Nature Power, not the move invoked by Nature Power
- Disable forces the target to use Struggle if it doesn't have any other usable moves ([Metapod](https://bulbapedia.bulbagarden.net/wiki/Metapod_(Pok%C3%A9mon)) can only learn Harden, so it might be helpful if overrides don't work for this case)

innerthunder (changes requested):

zacharied:
Whoops, sorry for the extra review request.

src/test/moves/disable.test.ts
DayKev: It's `SPLASH_ONLY` btw.

innerthunder (approved):
lgtm (code-wise, haven't tested)

torranx (approved):
looks good to me

src/field/pokemon.ts
```diff
@@ -2455,6 +2455,26 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     this.updateInfo();
   }
 
+  /**
+   * Gets whether a move is currently disabled for this Pokemon.
```
snoozbuster, line 2459:
```suggestion
   * Gets whether the given move is currently disabled for this Pokemon.
```

src/field/pokemon.ts
```diff
@@ -2455,6 +2455,26 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   */
+  isMoveDisabled(moveId: Moves): boolean {
+    return this.getDisablingTag(moveId) !== null;
+  }
+
+  /**
+   * Gets the {@link DisablingBattlerTag} that is disabling the given move, or null if that move is not disabled.
+   */
+  getDisablingTag(moveId: Moves): DisablingBattlerTag | null {
+    for (const tag of this.findTags(t => t instanceof DisablingBattlerTag)) {
```
snoozbuster, line 2470: Can't this be `return this.findTags(...).find(() =>...)`?

src/field/pokemon.ts
```diff
@@ -4312,11 +4330,24 @@ export class PokemonMove {
...
+   * @param ignorePp If true, skips the PP check
+   * @param ignoreDisableTags If true, skips the check for move-disabling tags
+   * @returns True if the move can be selected and used by the Pokemon, otherwise false.
+   */
+  isUsable(pokemon: Pokemon, ignorePp?: boolean, ignoreDisableTags?: boolean): boolean {
+    if (this.moveId && !ignoreDisableTags && pokemon.isMoveDisabled(this.moveId)) {
+      return false;
+    }
+
+    if (this.getMove().name.endsWith(" (N)")) {
```
snoozbuster, line 4346: This check for the move name is insane lol surely this fails in non-english locales? There is no better way to do this? (I realize this was here before but it's still so insane)

src/data/battler-tags.ts
```diff
@@ -92,6 +92,101 @@ export interface TerrainBattlerTag {
   terrainTypes: TerrainType[];
 }
 
+/**
+ * Base class for tags that disable moves. Descendants can override {@linkcode moveIsDisabled} to disable moves that
+ * match a condition. A disabled move gets cancelled before it is used. Players and enemies should not be allowed
+ * to select disabled moves.
+ */
+export abstract class DisablingBattlerTag extends BattlerTag {
```
snoozbuster, line 100: This is maybe a weird ask but thoughts on calling this `MoveRestrictionBattlerTag` or something along those lines? I think given it's intended genericism that's something of a better name (for example, choice effects might also be implentable through this base class) but also the existing name kinda gives off "handicapped Pokémon" vibes to me 😅

snoozbuster, line 100: Is this tag passed via baton pass? Is it supposed to be? (I didn't read the tests you added yet but since this tag is intended to be used across lots of moves it probably makes sense for it to be part of the baton pass test suite)

innerthunder: Baton pass logic should be rewritten soon (#3472 is the first step), so I wouldn't worry about this too much

src/field/pokemon.ts
innerthunder: I think " (N)" gets slapped onto the name of any move with `.unimplemented()`, regardless of language.

Obviously this check isn't ideal, but it shouldn't fail bc of localization issues.

snoozbuster: wild. insane. bizarre

DayKev: Unfortunately Sam decided the best way to check if a move is unimplemented is to slap " (N)" at the end and check for that, instead of properly adding some sort of `unimplemented` attribute. So we're stuck with it, and since 99% of moves are completed there's no real incentive to try and fix it.

src/data/battler-tags.ts
DayKev: The (new) default is for Baton Pass to not pass any `BattlerTag`s by default, requiring them to be specifically marked as transferrable. Since Baton Pass isn't supposed to pass Disable-type effects then it should automatically work properly.

zacharied:
Ah shit i forgor about this lmao

src/test/moves/disable.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    move ??= Moves.DISABLE;
+
+    game.onNextPrompt("CommandPhase", Mode.COMMAND, () => {
+      game.scene.ui.setMode(Mode.FIGHT, (game.scene.getCurrentPhase() as CommandPhase).getFieldIndex());
+    });
+    game.onNextPrompt("CommandPhase", Mode.FIGHT, () => {
+      const movePosition = getMovePosition(game.scene, 0, move);
+      (game.scene.getCurrentPhase() as CommandPhase).handleCommand(Command.FIGHT, movePosition, false);
+    });
+  };
```
innerthunder, line 30: Just use `game.move.select` instead of this; I think this is based on old test code

src/data/ability.ts
```diff
@@ -1085,29 +1085,20 @@ export class PostDefendMoveDisableAbAttr extends PostDefendAbAttr {
   }
 
   applyPostDefend(pokemon: Pokemon, passive: boolean, simulated: boolean, attacker: Pokemon, move: Move, hitResult: HitResult, args: any[]): boolean {
-    if (!attacker.summonData.disabledMove) {
+    if (attacker.getTag(BattlerTagType.DISABLED) === null) {
```
flx-sta, line 1088: Is it possible to be `undefined`?
E.g. due to `JSON.parse()` or something.

src/data/battler-tags.ts
```diff
@@ -98,6 +98,104 @@ export interface TerrainBattlerTag {
...
+  abstract interruptedText(pokemon: Pokemon, move: Moves): string;
+}
+
+/**
+ * Tag representing the "disabling" effect performed by {@linkcode Moves.DISABLE} and {@linkcode Abilities.CURSED_BODY}.
+ * When the tag is added, the last-used move of the tag holder is set as the disabled move.
+ */
+export class DisabledTag extends MoveRestrictionBattlerTag {
+  /** The move being disabled. Gets set when {@linkcode onAdd} is called for this tag. */
+  public moveId: Moves = Moves.NONE;
```
flx-sta, line 151: NIT:
wouldn't `protected` be enough?

src/field/pokemon.ts
```diff
@@ -2670,6 +2670,27 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     this.updateInfo();
   }
 
+  /**
+   * Gets whether the given move move is currently disabled for this Pokemon.
```
torranx, line 2674: nit:
```suggestion
   * Gets whether the given move is currently disabled for this Pokemon.
```

src/test/moves/disable.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    game.override.enemySpecies(Species.SHUCKLE);
+  });
+
+  it("restricts moves", async () => {
+    await game.classicMode.startBattle();
+    const playerMon = game.scene.getPlayerPokemon()!;
+    const enemyMon = game.scene.getEnemyPokemon()!;
+
+    playerMon.stats[Stat.SPD] = 1;
+    enemyMon.stats[Stat.SPD] = 2;
```
torranx, line 44: nit: these should be mocked (same with other occurrences here)

torranx (commented):
couple nits, some docs are missing `@param` and `@returns` too, otherwise looks good to me

src/test/moves/disable.test.ts
flx-sta: not a nit. We want to avoid spreading more of this `false` approach`

flx-sta: @torranx Is this really the recommended way though?

```ts
vi.spyOn(playerPokemon[0], "stats", "get").mockReturnValue([20, 20, 20, 20, 20, 100]);
```

torranx: aint pretty but it works. if you find better alternative, please inform us 👍

torranx (dismissed):
lgtm

DayKev (approved):

torranx (approved):

innerthunder (approved):

f-fsantos (approved):

PR #2055, opened 2024-06-10
[Feature] Add option to change canvas scale

devansh08:
## What are the changes?

This PR adds an option in the Display Settings to change the scale/size of the Phaser canvas.

On screens larger than mobiles (laptop/desktop) the scale of the UI elements (text, sprites etc.) starts looking too pixelated and hard to focus on, to me at the very least. With this option, the user can set the scale in terms of percentages, 100% being the current scale (100vh equivalent) down to 10%, per device as this is stored in local storage.

## What did change?

By default, there are no changes to the UI experience. Only on changing the Canvas Scale settings will there be a visible change in the size of the Phaser canvas.

### Screenshots/Videos

[pokerogue-canvas-scale-demo.webm](https://github.com/pagefaultgames/pokerogue/assets/21225957/ff9fe450-4d68-42c7-b25d-500515e71c83)

## How to test the changes?

Change the value of "Canvas Scale" option under Settings -> Display. No other code changes required to view the feature change.

## Checklist
    - [x] Are all unit tests still passing? (`npm run test`) - [Did not increase the number of failed cases]
- [x] Are the changes visual?

### DIFF
--- a/src/system/settings/settings.ts
+++ b/src/system/settings/settings.ts
@@ -68,7 +68,8 @@ export const SettingKeys = {
   Master_Volume: "MASTER_VOLUME",
   BGM_Volume: "BGM_VOLUME",
   SE_Volume: "SE_VOLUME",
-  Music_Preference: "MUSIC_PREFERENCE"
+  Music_Preference: "MUSIC_PREFERENCE",
+  Canvas_Scale: "CANVAS_SCALE"
 };
 
 /**
@@ -277,6 +278,13 @@ export const Setting: Array<Setting> = [
     type: SettingType.DISPLAY
   },
   {
+    key: SettingKeys.Canvas_Scale,
+    label: "Canvas Scale",
+    options: new Array(10).fill(null).map((_, i) => `${(i + 1) * 10}%`).reverse(),
+    default: 0,
+    type: SettingType.DISPLAY
+  },
+  {
     key: SettingKeys.Master_Volume,
     label: "Master Volume",
     options: VOLUME_OPTIONS,
@@ -457,6 +465,9 @@ export function setSetting(scene: BattleScene, setting: string, value: integer):
     break;
   case SettingKeys.Type_Hints:
     scene.typeHints = Setting[index].options[value] === "On";
+    break;
+  case SettingKeys.Canvas_Scale:
+    document.getElementById("app").style.height = Setting[index].options[value];
     break;
   case SettingKeys.Language:
     if (value) {
--- a/src/system/settings/settings.ts
+++ b/src/system/settings/settings.ts
@@ -467,7 +467,10 @@ export function setSetting(scene: BattleScene, setting: string, value: integer):
     scene.typeHints = Setting[index].options[value] === "On";
     break;
   case SettingKeys.Canvas_Scale:
-    document.getElementById("app").style.height = Setting[index].options[value];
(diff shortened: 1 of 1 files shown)


Dakurei:
If you can make the canvas stay centered horizontally AND vertically now that it can no longer take up the whole height, that might be nicer no ?

Perhaps other people's opinions might be useful 👍

https://github.com/pagefaultgames/pokerogue/assets/8146474/da66d671-c17e-44d0-94a7-f5030a3a8c28

torranx:
Possible duplicate: https://github.com/pagefaultgames/pokerogue/pull/1114

Also, there's lot to take into account when changing styles/scaling outside of phaser. You can read the PR conversation for reference.

Dakurei:
> Possible duplicate: #1114
> 
> There's lot to take into account when changing styles/scaling outside of phaser. You can read the PR conversation for reference. Also, I was told not to make an in-game setting for this, but should auto apply instead.


No placement worries if the scale is done in the right place 🤔


image
image

image

image


image
image
image


But yes, if it's already been done in another PR, it's not necessarily worth reinventing the wheel 👍

torranx:
Discussions happened on [discord](https://discord.com/channels/1125469663833370665/1249827339014115359). Also refer here:
Screenshot 2024-06-11 at 4 56 34 AM

devansh08:
> If you can make the canvas stay centered horizontally AND vertically now that it can no longer take up the whole height, that might be nicer no ?
> 
> Perhaps other people's opinions might be useful 👍
> 
> https://github.com/pagefaultgames/pokerogue/assets/8146474/da66d671-c17e-44d0-94a7-f5030a3a8c28

I did want to do this initially but could not get it to work with the usual "center a div via css" approaches without adding more flex to the parent, which may have interfered with the controls stuff.

Maybe I need to look into Phaser's APIs more for some options on this. Possibly having Phaser itself scale up/down would be better than CSS.

devansh08:
> Possible duplicate: https://github.com/pagefaultgames/pokerogue/pull/1114
> 
> There's lot to take into account when changing styles/scaling outside of phaser. You can read the PR conversation for reference. Also, I was told not to make an in-game setting for this, but should auto apply instead. 

The reason why I wanted it to be an option was so that every user can set the scale up for themselves, rather than there being a default. Maybe someone would like/need it to be larger/smaller ? 🤷

I'm not really sure how one would decide what the right size is for the screen, unless we standardize based on resolution or something. What do you think ? @torranx 

Thanks for linking the PR and discord convo, will go through those!

torranx:
> > Possible duplicate: #1114
> > There's lot to take into account when changing styles/scaling outside of phaser. You can read the PR conversation for reference. Also, I was told not to make an in-game setting for this, but should auto apply instead.
> 
> The reason why I wanted it to be an option was so that every user can set the scale up for themselves, rather than there being a default. Maybe someone would like/need it to be larger/smaller ? 🤷
> 
> I'm not really sure how one would decide what the right size is for the screen, unless we standardize based on resolution or something. What do you think ? @torranx
> 
> Thanks for linking the PR and discord convo, will go through those!

I personally think we don't need to have an option to manually scale the canvas smaller/larger. Instead we need to center it horizontally and vertically without losing the aspect ratio.

flx-sta:
Closed for now. No more feedback & activity (stale)

PR #2066, opened 2024-06-11
[Feature] add zh localization font

brain-frog:
## What are the changes?

Add a new font fallback for chinese localization

Chinese font size has been cut off and looking bad for TOO LONG

## What did change?

Added a font, updated i18n detection options, made some constants for the text (refactor coming soon I hope) 

### Screenshots/Videos

new-font

## How to test the changes?

set your language to one of the chinese ones and look around

## Checklist
- [x] Are the changes visual?


Tempo-anon (approved):

src/plugins/i18n.ts
```diff
@@ -94,44 +94,22 @@ const alternativeFonts = {
...
+function initFonts() {
+  Object.keys(alternativeFonts).forEach((language: string) => {
+    alternativeFonts[language].forEach((fontface: FontFace) => {
+      fontface.load().then(f => document.fonts.add(f)).catch(e => console.error(e));
     });
   });
 }
 
-export function initI18n(): void {
+export async function initI18n(): Promise<void> {
```
Greenlamp2, line 105: i think you never return a Promise

brain-frog: you have to set it to Promise<void> if its async
the point is to await the i18n init

Greenlamp2: you never await init18n anywhere though.
You don't have to set it to Promise if its async, you have to set it if you want to return an awaitable

Greenlamp2: "You don't have to set it to Promise if its async"
Or maybe this is a typescript rule i was not aware of in term of types

Greenlamp2 (approved):

PR #2077, opened 2024-06-11
Make starter selection account for evolutions and forms

Xavion3:
## What are the changes?
You will be able to choose pokemon that could evolve or change into a legal pokemon in challenges. It also correctly accounts for forms as a result, oricorio or paldean tauros now function correctly for monotype. Pokemon that could in the future be valid but currently can't can't be selected as the first pokemon on your team.

Forms were broken and evolutions weren't checked, preventing things like magikarp being selected for mono-flying or oricorio-pompom for mono-electric.

## What did change?
Added two arguments to the starter choice challenge to allow for forms and scaling with amount of pokemon.

## How to test the changes?
Try various combinations of challenges and look at what is allowed in the starter screen, and how it changes after you add your first pokemon to the team.

## Checklist


src/data/challenge.ts
```diff
@@ -277,8 +279,22 @@ export class SingleGenerationChallenge extends Challenge {
     case ChallengeType.STARTER_CHOICE:
       const species = args[0] as PokemonSpecies;
       const isValidStarter = args[1] as Utils.BooleanHolder;
+      const amountOfPokemon = args[3] as number;
```
Tempo-anon, line 282: Is there a place in the comments describing what args[] is and what each index represents?

src/data/challenge.ts
```diff
@@ -277,8 +279,22 @@ export class SingleGenerationChallenge extends Challenge {
...
       const starterGeneration = species.speciesId === Species.VICTINI ? 5 : species.generation;
-      if (starterGeneration !== this.value) {
+      const generations = [starterGeneration];
+      if (amountOfPokemon > 0) {
+        const speciesToCheck = [species.speciesId];
+        while (speciesToCheck.length) {
+          const checking = speciesToCheck.pop();
+          if (pokemonEvolutions.hasOwnProperty(checking)) {
+            pokemonEvolutions[checking].forEach(e => {
```
Tempo-anon, line 290: Isn't modifying a list you're iterating through bad practice?

src/data/challenge.ts
Xavion3: In the enum, this is kinda hacky for now yes, but this will be dramatically improved in the ongoing refactor.

src/data/challenge.ts
Tempo-anon: doesn't line 291 add to `speciesToCheck` and 288 remove from it while the whole thing is in a `while` on `speciesToCheck`?

Xavion3: Yes, but this is fine because we're not actually iterating over the list. This is a similar principle to stuff like search algorithms, there you'd keep track of a list of things to search and things you've searched, here since all evolution families are trees and there are no cycles we don't need the latter.

brain-frog (approved):

Issue #1505, opened 2024-05-28
[Feature] Implement Magnet RIse


Tempo-anon:
**Is your feature request related to a problem? Please describe.**
Magnet Rise (and the associated levitation mechanics) need to be implemented

**Describe the Feature**
We currently have a battlertag already for Magnet Rise but all the intricacies of being off the ground need to be implemented

**Additional context**
* Fails if user is ingrained (does ingrain remove the tag?)
* Removed if gravity is in play and fails if gravity is already in play
* Removed if user is smacked down or thousand arrow'd and cannot use it again

GrantMickle21:
I would love to have this implemented since I use Magnezone a lot. It was my first shiny.

abhinav-selvaraj:
im looking at the src code and im struggling to follow the hierarchy and see how its not implemented

Dakurei:
Implement in https://github.com/pagefaultgames/pokerogue/commit/89eba349c9f77985b10a0befadae7b5caf8edd0b

PR #2081, opened 2024-06-11
[Move] Implemented Magnet Rise

Frisk17:
## What are the changes?
Implemented the move [Magnet Rise](https://bulbapedia.bulbagarden.net/wiki/Magnet_Rise_(move))

The move was partially implemented with a `BattlerTag` but the effect of being immune from ground type moves wasn't.
[The issue for this move](https://github.com/pagefaultgames/pokerogue/issues/1505)

## What did change?
- Removed unimplemented from Magnet Rise in `move.ts`
- Implemented a message when the move is used in `battler-tag.ts`
- Implemented the immunity in `pokemon.ts`

### Screenshots/Videos
https://github.com/pagefaultgames/pokerogue/assets/38005728/4b9512df-914b-4d67-a195-0b311657ead9

https://github.com/pagefaultgames/pokerogue/assets/38005728/820f1d3b-994b-43f0-b7b6-1a488d014bf8

## How to test the changes?
`overrides.ts`
```ts
import { Species } from "./data/enums/species";
import { Abilities } from "./data/enums/abilities";
import { Biome } from "./data/enums/biome";
import { BerryType } from "./data/enums/berry-type";
import { StatusEffect } from "./data/status-effect";
import { modifierTypes } from "./modifier/modifier-type";

/**
 * Overrides for testing different in game situations
 * if an override name starts with "STARTING", it will apply when a new run begins
 */

/**
 * OVERALL OVERRIDES
 */

export const SINGLE_BATTLE_OVERRIDE: boolean = false;
export const STARTING_WAVE_OVERRIDE: integer = 0;
// default 1000
  active: false,
  pokeballs: {
    [PokeballType.POKEBALL]: 5,
    [PokeballType.GREAT_BALL]: 0,
    [PokeballType.ULTRA_BALL]: 0,
    [PokeballType.ROGUE_BALL]: 0,
  }
};

/**
 * PLAYER OVERRIDES
 */

// default 5 or 20 for Daily
export const STARTING_LEVEL_OVERRIDE: integer = 0;
/**
 * SPECIES OVERRIDE
 */
export const STARTER_SPECIES_OVERRIDE: Species | integer = Species.MAGNEZONE;
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.MAGNET_RISE, Moves.GRAVITY, Moves.BATON_PASS, Moves.SPLASH];

/**
 * OPPONENT / ENEMY OVERRIDES
 */

export const OPP_SPECIES_OVERRIDE: Species | integer = Species.GOLEM;
export const OPP_LEVEL_OVERRIDE: number = 0;
export const OPP_ABILITY_OVERRIDE: Abilities = Abilities.ARENA_TRAP;
export const OPP_STATUS_OVERRIDE: StatusEffect = StatusEffect.NONE;
export const OPP_MOVESET_OVERRIDE: Array<Moves> = [Moves.EARTHQUAKE, Moves.EARTHQUAKE, Moves.EARTHQUAKE, Moves.SPIKES];

/**
 * MODIFIER / ITEM OVERRIDES
 * if count is not provided, it will default to 1
 * @example Modifier Override [{name: "EXP_SHARE", count: 2}]
 * @example Held Item Override [{name: "LUCKY_EGG"}]
 *
 * Some items are generated based on a sub-type (i.e. berries), to override those:
 * @example [{name: "BERRY", count: 5, type: BerryType.SITRUS}]
 * types are listed in interface below
 * - TempBattleStat is for TEMP_STAT_BOOSTER / X Items (Dire hit is separate)
 * - Stat is for BASE_STAT_BOOSTER / Vitamin
 * - Nature is for MINT
 * - Type is for TERA_SHARD or ATTACK_TYPE_BOOSTER (type boosting items i.e Silk Scarf)
 * - BerryType is for BERRY
 */
interface ModifierOverride {
    name: keyof typeof modifierTypes & string,
    count?: integer
}
export const STARTING_MODIFIER_OVERRIDE: Array<ModifierOverride> = [];
export const OPP_MODIFIER_OVERRIDE: Array<ModifierOverride> = [];

export const STARTING_HELD_ITEMS_OVERRIDE: Array<ModifierOverride> = [];
export const OPP_HELD_ITEMS_OVERRIDE: Array<ModifierOverride> = [];
export const NEVER_CRIT_OVERRIDE: boolean = false;

/**
 * An array of items by keys as defined in the "modifierTypes" object in the "modifier/modifier-type.ts" file.
 * Items listed will replace the normal rolls.
 * If less items are listed than rolled, only some items will be replaced
 * If more items are listed than rolled, only the first X items will be shown, where X is the number of items rolled.
 */
export const ITEM_REWARD_OVERRIDE: Array<String> = [];
```

## Checklist
- [X] Are the changes visual?


src/field/pokemon.ts
```diff
@@ -1680,6 +1680,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     if (types.find(t => move.isTypeImmune(t))) {
       typeMultiplier.value = 0;
     }
+    if (this.getTags(BattlerTagType.MAGNET_RISEN).filter((tag) => tag instanceof TypeImmuneTag && tag.immuneType === move.type).length !== 0) {
```
zacharied, line 1695: any special reason we specify MAGNET_RISEN here? I'd imagine all TypeImmuneTags are relevant here

Frisk17: no, I think I did it because the only `TypeImmuneTag` was `MagnetRisen`
I've just changed it to handle every `TypeImmuneTag` now

Tempo-anon:
Does this account for all the edge cases and niche interactions? It's removed by smack down/thousand arrows and also fails if smacked down or thousand arrow'd. Gravity also disables magnet rise's effect during its duration. It's also transferable by baton pass.

Frisk17:
> Does this account for all the edge cases and niche interactions? It's removed by smack down/thousand arrows and also fails if smacked down or thousand arrow'd. Gravity also disables magnet rise's effect during its duration. It's also transferable by baton pass.

It already works with baton pass and ignores spikes, stealth rock etc. as expected. I just made changes that makes gravity and Smack Down / Thousand Arrows also work as expected. Everything other than Snatch (which is still unimplemented) should work fine.

Greenlamp2:
Would it be possible to write unit test to test these edge case ?

src/data/move.ts
```diff
@@ -6500,8 +6500,7 @@ export function initMoves() {
       .attr(AddBattlerTagAttr, BattlerTagType.MAGNET_RISEN, true, true)
       .condition((user, target, move) => !user.scene.arena.getTag(ArenaTagType.GRAVITY) &&
       !user.getTag(BattlerTagType.IGNORE_FLYING) && !user.getTag(BattlerTagType.INGRAIN) &&
-      !user.getTag(BattlerTagType.MAGNET_RISEN))
-      .unimplemented(),
+      !user.getTag(BattlerTagType.MAGNET_RISEN)),
```
OrangeRed, line 6503: Can you change this condition to use `[BattlerTagType].some(fn)` instead of three different conditional checks

src/field/pokemon.ts
```diff
@@ -1705,6 +1706,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     switch (moveCategory) {
     case MoveCategory.PHYSICAL:
     case MoveCategory.SPECIAL:
+      if (this.summonData.tags.filter((tag) => tag instanceof TypeImmuneTag && tag.immuneType === move.type).length !== 0) {
+        typeMultiplier.value = 0;
+      }
```
OrangeRed, line 1711: instead of `Array.filter()` can you use `Array.find()`

OrangeRed, line 1711: This check should be moved into `Pokemon.getAttackTypeEffectiveness()` because that function is also used by a pokemon ai to determine what move to use

src/data/battler-tags.ts
```diff
@@ -1187,8 +1187,11 @@ export class HideSpriteTag extends BattlerTag {
 
 export class TypeImmuneTag extends BattlerTag {
   public immuneType: Type;
+
   constructor(tagType: BattlerTagType, sourceMove: Moves, immuneType: Type, length: number) {
-    super(tagType, BattlerTagLapseType.TURN_END, 1, sourceMove);
+    super(tagType, BattlerTagLapseType.TURN_END, length, sourceMove);
+
+    this.immuneType = immuneType;
```
OrangeRed, line 1194: Can you default the length to 1 so it retains the previous functionality

src/data/arena-tag.ts
```diff
@@ -633,6 +633,11 @@ export class GravityTag extends ArenaTag {
 
   onAdd(arena: Arena): void {
     arena.scene.queueMessage("Gravity intensified!");
+    arena.scene.getField(true).forEach((pokemon) => {
+      if (pokemon !== null) {
+        pokemon.removeTag(BattlerTagType.MAGNET_RISEN);
+      }
+    });
```
OrangeRed, line 640: You might want to add a message to `onRemove()` to `MagnetRisenTag` to inform the player the effect has ended.

OrangeRed (changes requested):
In general can you also write some tests to verify the move is working correctly.

src/data/move.ts
Frisk17: I created a `MoveConditionFunc` separately for magnet rise but I'm unsure what you mean by `[BattlerTagType].some(fn)` since BattlerTagType is an enum

src/field/pokemon.ts
Frisk17: I added the check there too but I didn't remove this one since it doesn't look like that function is used for damage calculations (only the AI)

src/data/arena-tag.ts
Frisk17: Yes, I've added a message in `onRemove`

src/test/moves/magnet_rise.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    game.onNextPrompt("CommandPhase", Mode.COMMAND, () => {
+      game.scene.ui.setMode(Mode.FIGHT, (game.scene.getCurrentPhase() as CommandPhase).getFieldIndex());
+    });
+
+    game.onNextPrompt("CommandPhase", Mode.FIGHT, () => {
+      const movePosition = getMovePosition(game.scene, 0, moveToUse);
+      (game.scene.getCurrentPhase() as CommandPhase).handleCommand(Command.FIGHT, movePosition, false);
+    });
+    await game.phaseInterceptor.runFrom(EnemyCommandPhase).to(TurnEndPhase);
+    const finalHp = game.scene.getParty()[0].hp;
```
OrangeRed, line 52: Use the `game.doAttack()` utility instead of manually handling CommandPhase like this.

You also don't need to use `game.runFrom(Phase).to(Phase)` anymore, the `game.to(Phase)` should be sufficient in 90% of use cases

src/field/pokemon.ts
```diff
@@ -1705,6 +1709,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     switch (moveCategory) {
     case MoveCategory.PHYSICAL:
     case MoveCategory.SPECIAL:
+      if (!!this.summonData.tags.find((tag) => tag instanceof TypeImmuneTag && tag.immuneType === move.type)) {
+        typeMultiplier.value = 0;
+      }
```
OrangeRed, line 1714: after some closer inspection I think this type check shouldn't be here or inside of`Pokemon.getAttackMoveEffectiveness()` but instead should be moved into `Pokemon.getAttackTypeEffectiveness()` because that gets executed inside of both `Pokemon.getAttackMoveEffectiveness()` and `Pokemon.apply()`

OrangeRed (changes requested):

src/field/pokemon.ts
OrangeRed: I added another comment explaining.

src/data/move.ts
OrangeRed: I meant to use `Array.some(fn)` where the array is an array of BattlerTagTypes. It would look something like `[BattlerTagType.MAGNET_RISEN, ...].some(tag => user.getTag(tag))` 
You would have to figure out the boolean logic however so it still works the same way

src/field/pokemon.ts
Frisk17: I've made the changes

OrangeRed: > I've made the changes

Can you resolve conversations that you finish.

src/data/move.ts
```diff
@@ -5281,6 +5281,12 @@ const failIfDampCondition: MoveConditionFunc = (user, target, move) => {
   return !cancelled.value;
 };
 
+const failIfIngrainCondition: MoveConditionFunc = (user, target, move) => !user.getTag(BattlerTagType.INGRAIN);
+
+const magnetRiseCondition: MoveConditionFunc = (user, target, move) => {
+  return failOnGravityCondition(user, target, move) && failIfIngrainCondition(user, target, move) && !user.getTag(BattlerTagType.IGNORE_FLYING) && !user.getTag(BattlerTagType.MAGNET_RISEN);
+};
+
```
OrangeRed, line 5289: Can you instead do `[ BattlerTagType.INGRAIN, BattlerTagType.IGNORE_FLYING, BattlerTagType.MAGNET_RISE ].some((tag) => !user.getTag(tag))` instead of making a separate function.
Just double check that the logic is still the same.

OrangeRed (changes requested):

src/data/move.ts
Frisk17: I've made the requested changes

OrangeRed (approved):

OrangeRed:
Fix the conflicts and someone can merge thid

src/field/pokemon.ts
```diff
@@ -1163,7 +1163,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       multiplier /= 2;
     }
 
-    if (!!this.summonData.tags.find((tag) => tag instanceof TypeImmuneTag && tag.immuneType === moveType)) {
+    if (!!this.summonData && !!this.summonData.tags.find((tag) => tag instanceof TypeImmuneTag && tag.immuneType === moveType)) {
```
OrangeRed, line 1166: Can you just null coalesce via `this.summonData?.tags.find...`

Greenlamp2 (approved):

PR #2098, opened 2024-06-11
[Bug] Remove broken unit test

zacharied:
## What are the changes?
Remove the Intimidate unit test, which currently fails. Its creator should readd after fixing the current error.

PR guidelines request that `npm test` succeeds, but that is impossible currently due to this test. This means that all current PRs are failing the requirements.

## What did change?
Remove the Intimidate unit test.

## Checklist
- [ ] Are the changes visual?


zacharied:
Oh, looks like this has already been fixed.

PR #2101, opened 2024-06-11
[Feature] Add code formatter

### LINKED ISSUE #2069: [Feature] add a code formatter (please!)
**Is your feature request related to a problem? Please describe.**

It’s quite frustrating and tedious to stay on par with the existing code formatting as there is no standardized way to format the code

**Describe the Feature**

Can we please add an opinionated code formatter like prettier ?

I know there is people out there that might not be happy with the formatting it will dictate but thats always going to be the case with a project of this size. There is a lot of noise in PRs here and there because of formatting going back and forth.

**Additional context**

flx-sta:
# These two files represent a change done by prettier:

[battle.ts](https://github.com/pagefaultgames/pokerogue/pull/2101/files#diff-1408d7e8d1ec67295c6599addc08cf7ca6d009939c9813d0a7b6fe958d6ba328)
[utils.ts](https://github.com/pagefaultgames/pokerogue/pull/2101/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351)

## What are the changes?

As discussed in #2069 I've added the code formatter [prettier](https://prettier.io/) to the project.
Right now it's set up in a very basic way and I've just applied the formatting to the `utils.ts` file as an example.

Additional features we can add are: 
- Preferred configurations of the formatter
- Format on git-hook
- CI check for the formatting (to see if someone somehow skipped it)

closes #2069 

## What did change?

See `Files` tab

### Screenshots/Videos

n/a

## How to test the changes?

n/a 

## Checklist
  - ~~[] Have I provided screenshots/videos of the changes?~~


.prettierignore
```diff
@@ -0,0 +1,49 @@
...
+public/images/trainer/convert/*
+public/images/battle_anims/input/*.png
+public/images/pokemon/input/*.png
+public/images/pokemon/input/output/*
+public/images/pokemon/icons/input/*.png
+public/images/pokemon/icons/input/output/*
+public/images/character/*/
+src/data/battle-anim-raw-data*.ts
+src/data/battle-anim-data.ts
+src/overrides.ts
```
OrangeRed, line 34: I don't think this should be ignored

.prettierignore
```diff
@@ -0,0 +1,49 @@
...
+# Local Documentation
+/typedoc
+
+# misc
+package.json
+package-lock.json
+/public/audio
+**/*.md
+/.github
+eslint.config.js
```
OrangeRed, line 48: Same here I don't think this should be ignored

OrangeRed (commented):
You should also check how this handles the large blocks of code that we have where we init abilities and moves. If might be prudent to add // @prettier-ignore next to those so they don't format automatically

.prettierignore
flx-sta: good point

.prettierignore
flx-sta: yeah.
I also think package.json shouldn' be ignored by it

flx-sta:
> You should also check how this handles the large blocks of code that we have where we init abilities and moves. If might be prudent to add // @prettier-ignore next to those so they don't format automatically

I don't necessarily see why we shouldn't let them get formatted.
Otherwise we can also just ignore those files in `.prettierignore`

flx-sta:
Closing this for now. Can be reopened if needed/wanted

PR #2111, opened 2024-06-11
[Enhancement] Minor Search Engine Optimization

Mewtwo2387:
## What are the changes?
Added some meta tags that might help with SEO. 

Certain users have reported that in some regions, the other fake websites appears before pokerogue.net.
There aren't much we can do for a website that literally only have a canvas, but it doesn't hurt to add a few more meta tags.

## What did change?
Some twitter and open graph meta tags. 
Meta keywords, which probably doesn't do much in 2024.

I doubt this helps much if at all, but gotta try.

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,9 +4,13 @@
 <head>
 	<meta charset="UTF-8" />
 	<title>PokéRogue</title>
+	<meta name="title" content="PokéRogue" />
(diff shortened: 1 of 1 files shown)


OrangeRed (approved):

OrangeRed (commented):
Can you add the following meta tags as well
```
<meta name="distribution" content="Global">
<meta http-equiv="audience" content="General">
<meta name="title" content="...">
<meta name="news_keywords" content="...">
```

flx-sta (approved):
Awesome. This should definitely help.

brain-frog (approved):

PR #2142, opened 2024-06-12
[QoL] Reorganize Enum Locations

brain-frog:
## What are the changes?

Put all enum files into the enum folder, including moving ones out of data/enums since they are used in more than just the data folder

I want to organize the whole codebase better

## What did change?

Enum file placements and imports
Also fixed import session data overwrite message not displaying correctly

## How to test the changes?

Play the game

## Checklist
- [ ] Are the changes visual?


OrangeRed:
Can we also add an `index.ts` into `src/enums` that just re-exports all of the enums from the files in `src/enums`? If we do that then when importing inside other files we can use this syntax.
```ts
// Original
import { EnumA } from "#enums/EnumA"
import { EnumB } from "#enums/EnumB"
import { EnumC } from "#enums/EnumC"

// If there is an index that re-exports
import { EnumA, EnumB, EnumC } from "#enums"
```

bennybroseph (dismissed):

brain-frog:
> Can we also add an `index.ts` into `src/enums` that just re-exports all of the enums from the files in `src/enums`? If we do that then when importing inside other files we can use this syntax.
> 
> ```ts
> // Original
> import { EnumA } from "#enums/EnumA"
> import { EnumB } from "#enums/EnumB"
> import { EnumC } from "#enums/EnumC"
> 
> // If there is an index that re-exports
> import { EnumA, EnumB, EnumC } from "#enums"
> ```

good idea, I was trying to think of the best way to do this

Tempo-anon (dismissed):

Tempo-anon (approved):

PR #240, opened 2024-04-22
Fixed type immunity abilities not working against status moves

lucfd:
Currently type immunity abilities only work against instances of `AttackMove`, leading to unintentional behaviour such as moves like Leech Seed or Sleep Powder working against Pokemon with Sap Sipper.
- `TypeImmunityAbAttr` has been changed to work against all move categories (instances of `Move`)
- Added additional logic to prevent ability activation against moves that target the field (preventing Earth Eater from blocking spikes, Sap Sipper from blocking Grassy Terrain, etc)

PR #2145, opened 2024-06-12
[Move] Fix type immunities granted by abilities only applying to attacks

### LINKED ISSUE #278: Sap Sipper does not block Spore
Confirmed this behaviour shouldn't occur, Sap sipper should activate on grass type status moves.
Unconfirmed if this occurs with other status moves.

zacharied:
Revival of #240.

## What are the changes?
Type immunity abilities (such as Levitate and Sap Sipper) apply to non-field status moves, not just attacks.

Fixes #278. The current behavior of these abilities is incorrect.

## What did change?
In `TypeImmunityAbAttr.applyPreDefend()`, instead of just blocking attacks, block any move that does not target the field. Also remove an incorrect check there for `StatusMoveTypeImmunityAttr`, since that attribute applies to types of Pokemon, not types of Moves.

Add tests for Sap Sipper.

## How to test the changes?
The provided unit tests should be enough to verify the functionality. If you want to test it on the current build, consider these overrides:

```ts
export const STARTER_SPECIES_OVERRIDE: Species | integer = Species.BIDOOF;
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.SPORE];
export const OPP_ABILITY_OVERRIDE: Abilities = Abilities.SAP_SIPPER;
```

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -9,7 +9,7 @@ import { BattlerTag } from "./battler-tags";
 import { BattlerTagType } from "./enums/battler-tag-type";
 import { StatusEffect, getNonVolatileStatusEffects, getStatusEffectDescriptor, getStatusEffectHealText } from "./status-effect";
 import { Gender } from "./gender";
-import Move, { AttackMove, MoveCategory, MoveFlags, MoveTarget, StatusMoveTypeImmunityAttr, FlinchAttr, OneHitKOAttr, HitHealAttr, allMoves, StatusMove, SelfStatusMove, VariablePowerAttr, applyMoveAttrs, IncrementMovePriorityAttr  } from "./move";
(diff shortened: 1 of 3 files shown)


Tempo-anon (approved):

Tempo-anon:
Can you double check examples like rain dance vs storm drain should still work and charge vs volt absorb should still work?

zacharied:
> Can you double check examples like rain dance vs storm drain should still work and charge vs volt absorb should still work?

Confirmed, we're all good.

flx-sta:
@zacharied 
https://discord.com/channels/1125469663833370665/1250632973909229629 Please have a look at this

zacharied:
Oh yeah we're not good at all. I'm taking a look.

zacharied:
@flx-sta @Tempo-anon See #2168.

PR #2153, opened 2024-06-12
[Bug] Fix PostSummonPhase Logic to Support Single Pokémon Teams

Greenlamp2:
## What are the changes?
I've fixed the condition for `postSummonPhase` that checks how many Pokémon there were on the field for double battles. It was lacking a condition for when there is only 1 Pokémon available in our team, even if it's a double battle.

## What did change?
- **Before:** The `postSummonPhase` condition only checked for the number of Pokémon on the field for double battles, without considering the scenario of having only 1 Pokémon available in the team.
- **After:** Added a condition to handle the case where there is only 1 Pokémon available in our team during a double battle.

### Screenshots/Videos
webstorm64_qlcdP2nrjQ

## How to test the changes?
1. Check out the branch with the changes.
2. Run the application and navigate to a double battle scenario.
3. Ensure that there is only 1 Pokémon available in the team.
4. Observe the `postSummonPhase` behavior and verify that it now correctly handles the scenario.
5. Run all unit tests to ensure no other parts of the code are affected.
    - `npm run test`

## Checklist
- [ ] Are the changes visual?


src/phases.ts
```diff
@@ -1029,8 +1029,10 @@ export class EncounterPhase extends BattlePhase {
         const a = !!this.scene.getParty()?.length;
         // how many player pokemon are on the field ?
         const amountOnTheField = this.scene.getParty().filter(p => p.isOnField()).length;
+        // if it's a 2vs1, there will never be a 2nd pokemon on our field even
+        const minPossible = Math.min(this.scene.getParty().filter((p) => !p.isFainted()).length, 2);
         // if it's a double, there should be 2, otherwise 1
-        const b = this.scene.currentBattle.double ? amountOnTheField === 2 : amountOnTheField === 1;
+        const b = this.scene.currentBattle.double ? amountOnTheField === minPossible : amountOnTheField === 1;
         return a && b;
```
Tempo-anon, line 1036: can `a` and `b` be more readable variable names? also the line of setting b hurts me 🥲

Greenlamp2:
This will also fix the double battle crashes

brain-frog (approved):

PR #2163, opened 2024-06-13
Egg Common Grade Korean Expression Modification

esperar:
For Korean, the term "일반" would be more natural than "커먼" to describe a Common grade Egg

### DIFF
--- a/src/locales/ko/egg.ts
+++ b/src/locales/ko/egg.ts
@@ -5,7 +5,7 @@ export const egg: SimpleTranslationEntries = {
   "greatTier": "레어",
   "ultraTier": "에픽",
   "masterTier": "레전더리",
-  "defaultTier": "커먼",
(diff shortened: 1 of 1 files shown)


MrWaterT:
"Common" is read as "커먼" in Korean and its meaning is "일반" or "흔한", etc.
But "레어" "에픽" "레전더리" are written the same way-read English itself.
So I think converting all or doing nothing will be better.

Also in some Korean RPGs like Dungeon and Fighter use "커먼-레어-에픽-레전더리" for item rarity. No problem now.
("일반-고급-희귀-전설" is used, like in Diablo III, so converting all of 4 can be a discussion)

esperar:
I agree.

Tempo-anon:
Going to close this out since it seems like you've come to an agreement to keep it for now.

PR #2164, opened 2024-06-13
[QoL] Pity Toggle Setting

GladiusOmega:
## What are the changes?
Added a pity toggle setting that can be enabled or disabled, and will accordingly enable or disable the gacha's pity code via a new extra conditional for the appropriate sections. Gacha pity is on by default.

Some players do not like the pity system, as it makes the game feel more like a typical grind, and takes away from the satisfaction of getting something rare from the gacha. By adding an option to turn it off or back on, these players can enjoy the game without dealing with the pity, while the players that want it can still have it, and people can turn it on or off as they wish if they change their mind.

## What did change?
There is a new option in the options menu that sets a new option variable called gachaPity to true or false based on whether on or off is selected, respectively. An extra condition on each of the sections of code correlating to the pity system has been added, both for species and egg rarity, that requires the value of the gacha pity option to be on (true) for those sections of code to execute. There are also some small updates to the code comments to reflect this.

### Screenshots/Videos
temp2
temp

## How to test the changes?
Add console logging outputs to desired locations around the new code, set the instant hatching override to true, run the game locally, and use breakpoints/console commands to acquire vouchers. It should be possible to verify functionality cases by turning the new option on/off and checking the console outputs.

## Checklist
- [x] Are the changes visual?


Madmadness65:
I personally don't think this is a necessary addition, especially how the pity system is nothing but beneficial to the player, especially considering how many egg only epic and legendary egg Pokémon there are. I'll leave it open for now, but I don't think this will get accepted.

GladiusOmega:
The problem is that some people don't **want** the benefit, and it is actively disruptive to being able to enjoy the game for such people. That's the whole reason I made this. I don't think it would hurt to allow people the decision of how best to enjoy the game in this regard.

brain-frog:
This is a desired global feature of the game. If you would like to play without it, feel free to create an offline mod for your personal use.

PR #2167, opened 2024-06-13
[QoL] Post-Reroll UI Targeting

mKroogz:
## What are the changes?
This PR added a requesting QoL change for the OPTION of not jumping to the 1st item after rerolling.  It adds new reroll target options to the Display section of Game Settings.

Highly upvoted/bump/requested change in the feature-vote section of the discord.  Very nice convenience change for people who want to multi-reroll. 

## What did change?
* Added a setting to manage where the cursor will target after an item reroll
* Added logic in shop ui to handle cursor targeting following a reroll
* Tracks if user has rerolled at parent scene level with boolean, while handling that boolean at the child shop scene level

## How to test the changes?
* change in overrides.ts to give yourself a lot of money
* change in display settings where you would like to target after rerolling
* win a wave fight by any means and see while it starts on the first item, it should switch to your preferred option after reroll

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -106,7 +106,9 @@ export default class BattleScene extends SceneBase {
   public seVolume: number = 1;
   public gameSpeed: integer = 1;
   public damageNumbersMode: integer = 0;
+  public hasRerolled: boolean = false;
   public reroll: boolean = false;
+  public rerollTarget: integer = 1;
   public showMovesetFlyout: boolean = true;
   public showArenaFlyout: boolean = true;
   public showTimeOfDayWidget: boolean = true;
--- a/src/system/settings/settings.ts
+++ b/src/system/settings/settings.ts
@@ -54,6 +54,7 @@ export const SettingKeys = {
   Damage_Numbers: "DAMAGE_NUMBERS",
   Move_Animations: "MOVE_ANIMATIONS",
   Show_Stats_on_Level_Up: "SHOW_LEVEL_UP_STATS",
+  Reroll_Target: "REROLL_TARGET",
   Candy_Upgrade_Notification: "CANDY_UPGRADE_NOTIFICATION",
   Candy_Upgrade_Display: "CANDY_UPGRADE_DISPLAY",
   Move_Info: "MOVE_INFO",
@@ -193,6 +194,13 @@ export const Setting: Array<Setting> = [
     key: SettingKeys.Show_Stats_on_Level_Up,
     label: "Show Stats on Level Up",
     options: OFF_ON,
+    default: 1,
+    type: SettingType.DISPLAY
+  },
+  {
+    key: SettingKeys.Reroll_Target,
+    label: "Reroll Target",
+    options: ["Reroll", "Items", "Shop"],
     default: 1,
     type: SettingType.DISPLAY
   },
@@ -424,6 +432,8 @@ export function setSetting(scene: BattleScene, setting: string, value: integer):
   case SettingKeys.Show_Stats_on_Level_Up:
     scene.showLevelUpStats = Setting[index].options[value] === "On";
     break;
+  case SettingKeys.Reroll_Target:
+    scene.rerollTarget = value;
   case SettingKeys.EXP_Gains_Speed:
     scene.expGainsSpeed = value;
     break;
--- a/src/ui/modifier-select-ui-handler.ts
+++ b/src/ui/modifier-select-ui-handler.ts
@@ -231,7 +231,11 @@ export default class ModifierSelectUiHandler extends AwaitableUiHandler {
       });
 
       this.setCursor(0);
-      this.setRowCursor(1);
+      if(this.scene.hasRerolled){
+        this.setRowCursor(this.scene.rerollTarget);
+      } else {
+        this.setRowCursor(1);
+      }
 
       handleTutorial(this.scene, Tutorial.Select_Item).then(() => {
         this.setCursor(0);
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -4985,6 +4985,7 @@ export class SelectModifierPhase extends BattlePhase {
   constructor(scene: BattleScene, rerollCount: integer = 0, modifierTiers?: ModifierTier[]) {
     super(scene);
 
+    this.scene.hasRerolled = rerollCount > 0;
     this.rerollCount = rerollCount;
     this.modifierTiers = modifierTiers;
   }


src/battle-scene.ts
```diff
@@ -106,7 +106,9 @@ export default class BattleScene extends SceneBase {
   public seVolume: number = 1;
   public gameSpeed: integer = 1;
   public damageNumbersMode: integer = 0;
+  public hasRerolled: boolean = false;
```
flx-sta, line 109: Doesn’t this new field mean this would have to be kept up to date too ? I prefer having a single source of truth (reroll count)

src/battle-scene.ts
```diff
@@ -106,7 +106,9 @@ export default class BattleScene extends SceneBase {
...
   public reroll: boolean = false;
+  public rerollTarget: integer = 1;
```
flx-sta, line 111: Maybe match the name to the “cursor notation”?

src/battle-scene.ts
mKroogz: This field is kept up to date in the constructor for SelectModifierPhase with  `this.scene.hasRerolled = rerollCount > 0;`

I do get the want for a single source of truth though.  The reason I went with this method is that it seemed like the UI menu that manages the cursor doesn't have access to the SelectModifierPhase extension of scene.....but it does have access to the base scene?

What do we think on this?

flx-sta: Unfortunate I don’t have the necessary experience to answer that.

mKroogz: I'm trying to code around it but..... I don't think I can get rerollCount into the UI Cursor part :(

src/ui/modifier-select-ui-handler.ts
```diff
@@ -231,7 +231,11 @@ export default class ModifierSelectUiHandler extends AwaitableUiHandler {
       });
 
       this.setCursor(0);
-      this.setRowCursor(1);
+      if(this.scene.hasRerolled){
```
torranx, line 234: nit: space after `if` and ending `)`

mKroogz: You got it chief 🫡

src/battle-scene.ts
torranx: there's a notion we're following right now that UI should be event-based if possible. I'm not sure tho if this small change warrants an event-based approach. but then again these "small" things might accumulate. 

cc @bennybroseph

src/ui/modifier-select-ui-handler.ts
mKroogz: Updated :)

src/battle-scene.ts
mKroogz: Yeah I can absolutely try to make changes to meet design philosophy.  I figured this was "safe" in that the cursor targeting change happens right on the start/draw phase of the UI, and the rerollCount/hasRerolled logic is handled in the appropriate phase.

It's all small and pretty self-contained but if I need to make changes I'll do my best to meet the philosophy.

tejuRajbabu: Is there a design philosophy that dictates that we shouldn't keybind rerolls to a key? Like "R"?

Greenlamp2: you can access the ui handler for a specific mode by doing:
`this.scene.ui.handlers[Mode.XXX] as XXXUiHandler; `
and you can access the current phase by doing:
`this.scene.getCurrentPhase();`

`Is there a design philosophy that dictates that we shouldn't keybind rerolls to a key? Like "R"?`
you should always keep in mind that every bind you use for keyboard should also exists for gamepad & touch

bennybroseph: i don't know exactly why that boolean is needed, but if all you need is for the UI to something when something happens, just use an event.

torranx:
@mKroogz are you still working on this?

mKroogz:
> @mKroogz are you still working on this?

Super sorry for slackin on this definitely gonna finish this tonight/tomorrow

torranx:
bumping this, do you intend to finish this? @mKroogz

src/battle-scene.ts
mKroogz: Removed the boolean :)

mKroogz:
> bumping this, do you intend to finish this? @mKroogz

just updated it with the removed boolean as requested!  Hopefully this is good :)

src/ui/modifier-select-ui-handler.ts
```diff
@@ -231,7 +231,13 @@ export default class ModifierSelectUiHandler extends AwaitableUiHandler {
       });
 
       this.setCursor(0);
-      this.setRowCursor(1);
+      const phase = this.scene.getCurrentPhase() as any;
+      const rerollCount = phase.hasOwnProperty("rerollCount") ? phase.rerollCount : 0;
```
torranx, line 235: probably better to type this as `SelectModifierPhase` than `any` so you dont have to check "hasOwnProperty"

src/ui/modifier-select-ui-handler.ts
```diff
@@ -231,7 +231,13 @@ export default class ModifierSelectUiHandler extends AwaitableUiHandler {
...
+      if (rerollCount) {
+        this.setRowCursor(this.scene.rerollTarget);
+      } else {
+        this.setRowCursor(1);
+      }
```
torranx, line 240: nit: a ternary expression could shorten this w/o losing readability

mKroogz: That is so smart. On that right away haha.

src/ui/modifier-select-ui-handler.ts
torranx: typing it to `SelectModifierPhase` here essentially means youre asserting that the the current phase is a `SelectModifierPhase`, thus `rerollCount` _exists_

mKroogz: Yeah as long as the current phase here is always SelectModifierPhase that sounds good to me!  Added a getter for rerollCount too since it was private.

torranx:
lgtm, please fix merge conflicts as well

src/system/settings/settings.ts
```diff
@@ -360,6 +361,26 @@ export const Setting: Array<Setting> = [
...
+      },
+      {
+        value:"Items",
+        label: i18next.t("settings:Items")
+      },
+      {
+        value:"Shop",
+        label: i18next.t("settings:Shop")
+      }
+    ],
```
torranx, line 380: I don't think these keys exist. Please test your changes.

src/system/settings/settings.ts
```diff
@@ -360,6 +361,26 @@ export const Setting: Array<Setting> = [
     default: 1,
     type: SettingType.DISPLAY
   },
+  {
+    key: SettingKeys.Reroll_Target,
+    label: "Reroll Target",
```
torranx, line 366: Please allow translation here too

src/system/settings/settings.ts
mKroogz: ah....do you mean test in a different language?  I tested in english to make sure

src/system/settings/settings.ts
mKroogz: fixed this :D

src/system/settings/settings.ts
```diff
@@ -363,7 +363,7 @@ export const Setting: Array<Setting> = [
   },
   {
     key: SettingKeys.Reroll_Target,
-    label: "Reroll Target",
+    label: i18next.t("settings:Reroll Target"),
```
flx-sta, line 366: Normally we don't use "human writing" for keys but `camelCase`

```suggestion
    label: i18next.t("settings:rerollTarget"),
```

src/system/settings/settings.ts
torranx: The keys `Reroll`, `Items`, and `Shop` do not exist on any `settings` localization files. Please input them to allow for localization- and mind the format (camelCase).

src/system/settings/settings.ts
flx-sta: @mKroogz make sure to press the `resolve` button at the bottom

src/system/settings/settings.ts
flx-sta: @mKroogz make sure to press the `resolve` button at the bottom

torranx:
I'm once again following up on this @mKroogz

PR #2168, opened 2024-06-13
Fix type-immunity abilities activating on self-targeted status moves

zacharied:
## What are the changes?
Fix a regression introduced by #2145 where self-targeted status moves like CHARGE would activate the user's type-immunity ability like VOLT ABSORB.

Because I fucked up.

## What did change?
Check that the user is different than the attacker in `TypeImmunityAbAttr`.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/6374275/1304de4f-acfb-4500-9bd2-50d7a48bac3e

## How to test the changes?
Unit tests, or these overrides:

```ts
export const ABILITY_OVERRIDE: Abilities = Abilities.VOLT_ABSORB;
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.CHARGE];
```

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -9,7 +9,7 @@ import { BattlerTag } from "./battler-tags";
 import { BattlerTagType } from "./enums/battler-tag-type";
 import { StatusEffect, getNonVolatileStatusEffects, getStatusEffectDescriptor, getStatusEffectHealText } from "./status-effect";
 import { Gender } from "./gender";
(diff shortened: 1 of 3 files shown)


zacharied:
There's probably some stupid edge case I'm not thinking of where you can target yourself with an attack. That would break this solution. If anyone can think of an example of that please lmk.

Tempo-anon (approved):

PR #1499, opened 2024-05-28
[Bug] Added check for expGainsSpeed to updatePokemonExp to disable XP Bar animation on Skip setting

Teufelsstern:
As of now updatePokemonExp() does not respect the expGainsSpeed setting "Skip". This means that the exp bar animation still plays. This added check aims to fix that.

All necessary checks and status updates are performed before returning and the exp bar is instantly set to the appropriate value.

How to test: The xp bar animation will no longer trigger with setting "Exp Gains Speed" to "Skip".

## Checklist
- [X ] There is no overlap with another PR?
- [X ] The PR is self-contained and cannot be split into smaller PRs?
- [X ] Have I provided a clear explanation of the changes?
- [X ] Have I tested the changes (manually)?
    - [X ] Are all unit tests still passing? (`npm run test`)
- [ ] Are the changes visual?
  - [ ] Have I provided screenshots/videos of the changes?

PR #2169, opened 2024-06-13
[QoL] ExpGainsSpeed Setting works in expBar

MrWaterT:
## What are the changes?
Exp Gains Speed in General setting affects in expBar, not only party exp gaining.

Since there is 'Skip' in setting values, skipping expBar animation will be correct result.
Repeat from #1499 but it doesn't seem to be active.

## What did change?
Update `updatePokemonExp()`.
Replicate [`updatePokemonHp()`](https://github.com/MrWaterT/pokerogue/blob/a111c96adff83b932bdaa721e95437b753e98651/src/ui/battle-info.ts#L562) and `scene.hpBarSpeed`.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/87186129/28115dde-a4de-4932-8171-a34be4230e92

https://github.com/pagefaultgames/pokerogue/assets/87186129/ee71748d-313a-485b-868e-31433c5c08fa

https://github.com/pagefaultgames/pokerogue/assets/87186129/814e0142-b47f-444e-9166-1e9985752768

https://github.com/pagefaultgames/pokerogue/assets/87186129/d8d6be0c-6b16-4c92-af21-c5fe2c124689

## How to test the changes?
In varient condition, win battle and gain EXP.

## Checklist
- [x] Are the changes visual?


src/ui/battle-info.ts
```diff
@@ -661,7 +661,11 @@ export default class BattleInfo extends Phaser.GameObjects.Container {
         instant = true;
       }
       const durationMultiplier = Phaser.Tweens.Builders.GetEaseFunction("Sine.easeIn")(1 - (Math.max(this.lastLevel - 100, 0) / 150));
-      const duration = this.visible && !instant ? (((levelExp - this.lastLevelExp) / relLevelExp) * 1650) * durationMultiplier * levelDurationMultiplier : 0;
+      let duration = this.visible && !instant ? (((levelExp - this.lastLevelExp) / relLevelExp) * 1650) * durationMultiplier * levelDurationMultiplier : 0;
+      const speed = (this.scene as BattleScene).expGainsSpeed;
+      if (speed) {
+        duration = speed >= 3 ? 0 : duration / Math.pow(2, speed);
```
hayuna, line 667: In which case speed could be more than 3?

Could we avoiding magic numbers and move `ExpGainSpeed` into enums and then replace this 3 with ExpGainSpeed.SKIP?

Same example as ExpNotification here
https://github.com/pagefaultgames/pokerogue/pull/1909

MrWaterT: Surely better way, but It's looks like something bigger than this PR.
Is additional PR... better? for '*cannot be split into smaller*'?

Maybe I can modify `ExpGainsSpeed` and `HpBarSpeed` at once, with same shape `["Normal","Fast","Faster","Skip"]`,
But I don't know which one is better-using 1 enum or 2 enums.
I'll wait and follow the refactoring merge if there is. (Or make another PR after merged)

hayuna: 2 enums. What if in the future there will be decision to add new value to hpBarSpeed but not to the ExpGainSpeed?

MrWaterT: Aren't they in exactly the same category?
Setting about speedup options for skippable animations in battle, which are directly exposed to users.
Like volume settings, there are 3 vars but in same category, so declared from single Array `VOLUME_OPTIONS`.
Modify only one of them can make worse UI & UX. I can hardly agree...

20% agree with 2 enums and 40% with 1 enum. I won't refactor them myself, for the project.

src/ui/battle-info.ts
```diff
@@ -661,7 +661,11 @@ export default class BattleInfo extends Phaser.GameObjects.Container {
         instant = true;
       }
       const durationMultiplier = Phaser.Tweens.Builders.GetEaseFunction("Sine.easeIn")(1 - (Math.max(this.lastLevel - 100, 0) / 150));
-      const duration = this.visible && !instant ? (((levelExp - this.lastLevelExp) / relLevelExp) * 1650) * durationMultiplier * levelDurationMultiplier : 0;
+      let duration = this.visible && !instant ? (((levelExp - this.lastLevelExp) / relLevelExp) * 1650) * durationMultiplier * levelDurationMultiplier : 0;
+      const speed = (this.scene as BattleScene).expGainsSpeed;
+      if (speed) {
```
Tempo-anon, line 666: Is this checking if it's not null or of if it's greater than 0? Should be more explicit
`if (speed > ExpGains.DEFAULT)`

src/ui/battle-info.ts
Tempo-anon, line 667: `duration = (speed >= ExpGainsSpeed.SKIP) ? 0 : (duration / Math.pow(2, speed));`

flx-sta:
Closing for mirror instead of zombie

PR #2179, opened 2024-06-13
[QoL] Language option cursor resets after cancelling or switching language

Mewtwo2387:
## What are the changes?
After you switch language or cancel switching language, the option cursor now resets back to being on the language instead of on "change"
image

The way switching language works right now is, you have to move from your language to change for the options to pop up.

However, after switching languages, the cursor stayed on change, you have to move back to your language then back to change to switch languages again. 

It makes more sense for the cursor to be hovering over the new language instead of on change, and makes it a bit more convenient for people testing localization changes.

## What did change?
`cancelHandler` when setting language

## How to test the changes?
1. Switch languages
2. Go to settings again

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/system/settings/settings.ts
+++ b/src/system/settings/settings.ts
@@ -463,8 +463,7 @@ export function setSetting(scene: BattleScene, setting: string, value: integer):
       if (scene.ui) {
         const cancelHandler = () => {
           scene.ui.revertMode();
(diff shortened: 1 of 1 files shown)


torranx (approved):
lgtm. this has been bugging me too

Tempo-anon (approved):

PR #2182, opened 2024-06-13
[Bug] Fixes manaphy egg rate back to 1/256

sirzento:
## What are the changes?

Changes the manaphy egg rate back to 1/256. It wasn't 1/256 because the egg first needs to roll common and after that the 1/255 check was done.

To set the intended rate.

## What did change?

just one line for the manaphy egg generation.
grafik

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/egg.ts
+++ b/src/data/egg.ts
@@ -28,7 +28,7 @@ export class Egg {
   }
 
   isManaphyEgg(): boolean {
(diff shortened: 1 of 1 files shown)


sirzento:
I don't know. dev's needs to decide. I only talked about that with Temp and he said that this will most likely will get merged since the intended rates where 1/256

bennybroseph (approved):

PR #2189, opened 2024-06-13
Adds Delta Pokemon From Pokemon Insurgance

Meman82:
Add delta Pokemon from Insurgence to poke rogue as alternative forms for base Pokemon. 8/10 moves and 20/22 abilities for these delta Pokemon and there mega evolution have been added.

Updated Starter-select-ui-handler to handle cases where the sub-species does not have the same abilities as the main form,
Updated Battle-Scene to pick between the new forms.
untitle2
Untitled
Untitled3


Meman82:
> Very cool, maybe they should be their own separated specie tho?

That was my initial plan but I ran into some difficulties. The best solution I could think of was to add another region for them due to the number present breaking the 9x9 of the starter ui. However, I was never able to get any details to appear for the new Pokemon once an additional region was added which is what lead me to try this approach. This approach does have certain benefits such as not needing duplicate cries, nor needing to manually add the Pokemon into different zones/trainers. 

With that said if someone could point me to the correct way to add a new generation/implement as separate species I think I can convert it to that method fairly quickly.

returntoice:
I'm sorry, but I don't think we should add Delta Pokemons to Pokerogue. Although Pokerogue has many unique elements, it fundamentally retains the original types and appearances of the original Pokemon, even when using the DNA Splicer. The original Pokemon from custom ROMs have their own value and can fit well within their stories, but I think it doesn't mean it's appropriate to add them to Pokerogue. While Pokerogue does have passive abilities and various additional features, I think we should maintain the current state of the Pokemon pool.

auger1234:
뇌절이야

OrangeRed:
I like this change but maybe we can make the delta forms super rare so that the majority of pokemon for the average player are the base ones.

bennybroseph:
This has huge implications. Also, unless you're the artist/developer of that game we definitely don't have permission to just add that.

jaimefd:
> I'm sorry, but I don't think we should add Delta Pokemons to Pokerogue. Although Pokerogue has many unique elements, it fundamentally retains the original types and appearances of the original Pokemon, even when using the DNA Splicer. The original Pokemon from custom ROMs have their own value and can fit well within their stories, but I think it doesn't mean it's appropriate to add them to Pokerogue. While Pokerogue does have passive abilities and various additional features, I think we should maintain the current state of the Pokemon pool.

Why should PokeRogue be stuck with what Game Freak gave us? Is it really necessary to limit the game that much? This is a community made project from players to players, purely to have fun and enjoyment. It is a constantly growing and expanding project, we don't have the limitations Nintendo has when making their games.

I think as long as we ask the original artists of these fan forms for permission, this is a great idea. Just like there's lots of great and creative programmning minds in the community implementing cool features and improvements for the game, there's also great artistic minds designing really amazing forms for pokemon, that will just improve the experience overall and give you new cool ways to play with your fav pokemon.

As long as we stick to just new forms and megas (so no completely new pokemon) and make them very rare, this is awesome

Heppity:
I feel the main issue with this is the fact we still have loads of content we could go for mainline pokemon wise variant without even touching Delta as touching other fan games and such can make alot of things messy imo.

amatsune:
Honestly, seeing as the bulk of the work is already done, I'd set this as a hidden option, so it really shows up to people who've done "everything" the game currently has to offer.
If the player has reached the end of endless and has all the achievements, then an option in settings becomes available:
Toggle fan-game/romhack content (experimental)? Yes No

Then, and only then, new games will feature this content.
> I feel the main issue with this is the fact we still have loads of content we could go for mainline pokemon wise variant without even touching Delta as touching other fan games and such can make alot of things messy imo.
Give players who meet the requirements to unlock this no warning of the new setting and let them find it out for themselves, at least until the core of the game content is "finished"

OrangeRed:
Now that we have Challenges we could probably lock this feature behind a difficult challenge.

But realistically speaking this branch will probably be relegated to only be playable offline where modding is accepted.

Tempo-anon:
Closing due to inactivity and merge conflicts. This is quite a gigantic suggestion, but might I suggest this as an unofficial mod for the game rather than as an official feature? Getting permission for all this stuff sounds like a logistical nightmare.

amatsune:
It's actually much smaller than it seems at first, but it is a challenge to me as a newbie coder to cope with all the compatibility issues.

I'm in my exam season, so I can't really give this the time it needs to work for another week or so. I guess by then I can make a new pr working from a more recent fork

Filipe Sant'Anna.
________________________________

PR #2200, opened 2024-06-14
[Move] Implement Fling

AyushBarik:
## What are the changes?

The user can now utilize the move fling. 

It is an important move in the game.

Greater diversity of available moves, which enhances user experience.

## What did change?

The move fling is now fully functional.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/128145955/7281d3bf-87da-410e-8d2c-a5aac97c5b5c

## How to test the changes?

I made use of src/overrides.ts to test different items. I did not introduce automated tests. 
I carried out tests on various items and confirmed functionality. If there is an issue, it may be in 
determining which items are able to be flung and which are not. You may want to check my arrays
to see if the items are appropriately sorted. You can also go through the code and try to see
if the logic makes sense. I left behind some comm

My Discord username is bgjormungandr, I'm on the pokerogue server. Feel free to contact me there if needed.

## Checklist
- [ yes] There is no overlap with another PR?
- [ yes] The PR is self-contained and cannot be split into smaller PRs?
- [ yes] Have I provided a clear explanation of the changes?
- [ yes] Have I tested the changes (manually)?
    - [did not implement tests] Are all unit tests still passing? (`npm run test`)
- [ yes] Are the changes visual?
  - [yes ] Have I provided screenshots/videos of the changes?


hayuna:
Could you update title of this PR?

[Move] Implement Fling

src/data/move.ts
```diff
@@ -2615,6 +2616,139 @@ export class LessPPMorePowerAttr extends VariablePowerAttr {
...
+      power.value = 0;
+      user.scene.queueMessage("But it failed!");
+      return false;
+    }
+    const validHeldItems = heldItems.filter(item => allItems.includes(item.type.name.toLowerCase().replace(/_/g, " ")));
+    if (validHeldItems.length === 0) { //user has no flingable held item
+      power.value = 0;
+      user.scene.queueMessage("No valid item found!");
+      return false;
+    }
```
Arxalc, line 2680: You can put this as an else at the end of your else-if statements since it's already checking through all the arrays.

src/data/move.ts
```diff
@@ -26,6 +26,7 @@ import { BattlerTagType } from "#enums/battler-tag-type";
 import { Biome } from "#enums/biome";
 import { Moves } from "#enums/moves";
 import { Species } from "#enums/species";
+import { BerryType } from "#enums/berry-type";
```
Arxalc, line 29: Instead of importing BerryType, maybe you can check `if (this.randomItem instanceof BerryModifier || arr10.includes(this.randomItem)` ?

Arxalc:
I see why you included all the item names, but since most of them aren't currently in PokeRogue, I'd just comment them out and let someone else add them in once the item is implemented in the future.

You can completely get rid of evolutionary stones. Pokemon don't hold those.

src/data/move.ts
```diff
@@ -2615,6 +2616,139 @@ export class LessPPMorePowerAttr extends VariablePowerAttr {
...
+        user.scene.updateModifiers(user.isPlayer());
+        return true;
+      } else if (arr130.includes(this.randomItem)) {
+        power.value = 130;
+        user.scene.removeModifier(randomHeldItemModifier, !user.isPlayer());
+        user.scene.updateModifiers(user.isPlayer());
+        return true;
+      }
+    }
+
```
Arxalc, line 2739:
```suggestion
  } else if (arr130.includes(this.randomItem)) {
          power.value = 130;
  } else {
        power.value = 0;
        user.scene.queueMessage("No valid item found!");
        return false;
  } // else-if
      user.scene.removeModifier(randomHeldItemModifier, !user.isPlayer());
      user.scene.updateModifiers(user.isPlayer());
      return true;
} // if
```

src/data/move.ts
AyushBarik: That code works when I tried it however,

In line 2675 **const validHeldItems = heldItems.filter(item => allItems.includes(item.type.name.toLowerCase().replace(/_/g, " ")));**

I filter out all the non-flingable items. By not importing berries and not adding them in my array, the berries would get filtered out as non-flingable items. How do I deal with this? 

I could manually type in the different berry types into the array and we can avoid having to import berry-type.ts. Is that something I should do?

AyushBarik:
Changes made as per the Suggestions:
- Got rid of evolutionary items as flingable items
- heldItems check moved to the beginning of the function
- Dealt with berries without importing berry-type.ts

Changes not made: 
- I was suggested to move the validHeldItems check at the end of the else-if statements but I felt that may cause bugs so I have not implemented it for now

Additionally, as for commenting out unimplemented items from the list of flingable items - I am not sure which items have or have not been implemented. Perhaps someone more familiar with the game can change the arrays accordingly.

src/data/move.ts
```diff
@@ -2653,6 +2654,137 @@ export class LessPPMorePowerAttr extends VariablePowerAttr {
   }
 }
 
+export class FlingAttr extends VariablePowerAttr {
+  protected randomItem;
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
```
Tempo-anon, line 2659: class and apply need tsdoc

src/data/move.ts
```diff
@@ -2653,6 +2654,137 @@ export class LessPPMorePowerAttr extends VariablePowerAttr {
...
+        return true;
+      }
+    }
+
+    power.value = 0;
+    user.scene.queueMessage("But it failed!");
+    return false;
+  }
+
+  getMyHeldItems(user: Pokemon): PokemonHeldItemModifier[] {
```
Tempo-anon, line 2781: nit: not sure if this needs to be its own function if it'll only be used for this. You can probably just inline this, though I'm surprised there isn't one already

AyushBarik (commented):
Implemented all feedback from Tempo-anon:
- tsdoc created for class and apply
- unnecessary comments removed
- getMyHeldItems removed as a function and made inline instead.

src/data/move.ts
Arxalc: Oh I didn't think clearly on this. The way you have it is fine.

src/data/move.ts
```diff
@@ -2653,6 +2654,154 @@ export class LessPPMorePowerAttr extends VariablePowerAttr {
...
+    const arr70i = ["Dragon Fang", "Poison Barb", "Power Anklet", "Power Band", "Power Belt", "Power Bracer",
+      "Power Lens", "Power Weight"];
+    const arr80i = [
+      "Assault Vest", "Quick Claw", "Razor Claw",
+      "Sachet", "Safety Goggles", "Sticky Barb",
+      "Weakness Policy"
+    ];
+    const arr90i = ["Grip Claw"];
+    const arr100i = ["Hard Stone"];
+    const arr130i = ["Iron Ball"];
```
OrangeRed, line 2717: Instead of making huge arrays of item names like this wouldn't it be better to just add the flingPower / weight of the item onto that item's class (ModifierType)?

src/data/move.ts
```diff
@@ -2653,6 +2654,154 @@ export class LessPPMorePowerAttr extends VariablePowerAttr {
...
+   * @returns A boolean indicating whether the move was successfully applied.
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    const power = args[0] as Utils.NumberHolder;
+
+    // Inline logic to retrieve the held items for the given Pokémon
+    const heldItems = user.scene.findModifiers(
+      (m) => m instanceof PokemonHeldItemModifier && (m as PokemonHeldItemModifier).pokemonId === user.id,
+      user.isPlayer()
+    ) as PokemonHeldItemModifier[];
```
OrangeRed, line 2682: `(m as PokemonHeldItemModifier)` isn't necessary because typescript infers the type from `m instanceof PokemonHeldItemModifier`

OrangeRed (changes requested):

AyushBarik:
https://github.com/pagefaultgames/pokerogue/assets/128145955/b37ce86b-6420-460a-8fbc-d031567b7017

Test that displays the functionality of fling as well as the case if none of the held items possessed by the Pokemon are flingable. (For testing purposes flame orb was made unflingable temporarily).

AyushBarik:
https://github.com/pagefaultgames/pokerogue/assets/128145955/7467c694-9ad9-4eb9-a1b2-e37ab9f9e569

Displays what happens if the user has no held items

src/data/move.ts
```diff
@@ -2689,6 +2690,148 @@ export class LessPPMorePowerAttr extends VariablePowerAttr {
...
+    if (heldItems.length === 0) {
+      return "But it failed!";
+    }
+
+    const allItems = Object.values(this.flingPowerMap).flat();
+    const validHeldItems = heldItems.filter(item => allItems.includes(item.type.name.toLowerCase().replace(/_/g, " ")));
+
+    if (validHeldItems.length === 0) {
+      return "No valid held items!";
+    }
```
OrangeRed, line 2777: Is this necessary? Doesn't the game just show "But it failed" if you don't have a valid item?

OrangeRed (commented):
Since you're validating against the names of items does this work correctly in different locales?

src/data/move.ts
AyushBarik: Not necessary, I just thought it would be useful. The user might have held items, but might be wondering why fling failed. this message tells them that their items are not flingable.

AyushBarik: "Since you're validating against the names of items does this work correctly in different locales?"

I tried running the game in other languages and fling does not seem to work... How do i fix this issue?

Tempo-anon (changes requested):
This does not work since it's checking for hardcoded english names. Could we not update `modifier-type.ts` to include the relevant type and damage info while just having the special effects in the FlingAttr?

PR #2243, opened 2024-06-15
[Bug] Call preSwitchOutAb when Transitioning Double > Single Battle

Arxalc:
## What are the changes?

Weather clears upon the pokemon returning.

Didn't apply before. Example: Double -> Single battle, if Primal Groudon returned, its weather would remain.
## What did change?

Added a preSwitchOutAb call on the Return phase.
This makes sure other abilities that extend PreSwitchOutAbAttr will activate correctly in the same scenario as well.
### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/63990624/8044a328-9782-4558-a963-b4402e6ed3bd

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -1959,10 +1959,7 @@ export class PreSwitchOutClearWeatherAbAttr extends PreSwitchOutAbAttr {
    */
   applyPreSwitchOut(pokemon: Pokemon, passive: boolean, args: any[]): boolean | Promise<boolean> {
     let turnOffWeather = false;
-    let weatherType = WeatherType.NONE;
-    if (pokemon.scene.arena.weather.weatherType) {
-      weatherType = pokemon.scene.arena.weather.weatherType;
-    }
+    const weatherType = pokemon.scene.arena?.weather?.weatherType ?? WeatherType.NONE;
 
     // Clear weather only if user's ability matches the weather and no other pokemon has the ability.
     switch (weatherType) {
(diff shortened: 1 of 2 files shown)


src/data/ability.ts
```diff
@@ -1958,8 +1958,11 @@ export class PreSwitchOutClearWeatherAbAttr extends PreSwitchOutAbAttr {
    * @returns {boolean} Returns true if the weather clears, otherwise false.
    */
   applyPreSwitchOut(pokemon: Pokemon, passive: boolean, args: any[]): boolean | Promise<boolean> {
-    const weatherType = pokemon.scene.arena.weather.weatherType;
     let turnOffWeather = false;
+    let weatherType = WeatherType.NONE;
+    if (pokemon.scene.arena.weather.weatherType) {
+      weatherType = pokemon.scene.arena.weather.weatherType;
+    }
```
OrangeRed, line 1965: What is the point of this?

src/phases.ts
```diff
@@ -1679,6 +1679,11 @@ export class ReturnPhase extends SwitchSummonPhase {
 
     this.scene.updateFieldScale();
 
+    const preSwitchAbList = pokemon.getAbilityAttrs(PreSwitchOutAbAttr);
+    if (preSwitchAbList.length > 0) {
+      applyPreSwitchOutAbAttrs(PreSwitchOutAbAttr, pokemon);
+    }
+
```
OrangeRed, line 1686: You don't need to do this conditional because `applyPreSwitchOutAbAttrs()` already checks that.

src/data/ability.ts
Arxalc: There's a console warning sometimes that scene weather is undefined. This was my attempt to correct it, but it doesn't work and I forgot to update it. Going to do truthy to check this

OrangeRed: if weatherType is sometimes undefined then use the null coalescing operator to default it to WeatherType.NONE
Something along the lines of
`const weatherType = pokemon.scene.arena?.weather?.weatherType ?? WeatherType.NONE;`

OrangeRed (approved):

OrangeRed:
@Arxalc Are you sure that all PreSwitchOut abilities should be activate in this scenario? Like for example should Finizen turn into Palafin? Similarly should Natural Cure and Regenerator activate?

DayKev:
Closing in favor of #3493

PR #2244, opened 2024-06-15
Fix bug for Spite and fully implement Eerie Spell

schmidtc1:
## What are the changes?

Fixes a visual bug with Spite where a move with 0pp would display no message and 'skip' the turn. Fully implements Eerie Spell using Spite attribute class, and removes the hard coded value of the pp reduction from the apply function. 

I wanted to be able to use Eerie Spell in my runs with full functionality. While working on this, I discovered a 'visual bug' with Spite, where if a user selects a move with <4pp and the opponent uses spite on that move (to reduce it to 0pp), the turn would 'skip' with no text or move message displayed. I added move failed-text for this scenario to improve clarity.

## What did change?

Fully implements Eerie Spell by creating a child class of Spite's ReducePpMoveAttr class. Fixes visual bug with Spite/0pp.

### Screenshots/Videos

0pp visual bug before

https://github.com/pagefaultgames/pokerogue/assets/62030095/75993a47-c6fa-4c75-8787-7d6bd52b459f

0pp visual bug fixed (properly displays 'But it failed' text)

https://github.com/pagefaultgames/pokerogue/assets/62030095/7470e751-f580-4eee-b0bd-32289f62d7c4

Eerie spell before (no pp reduction)

https://github.com/pagefaultgames/pokerogue/assets/62030095/ef04b739-3093-4972-a797-6e883cda70a4

Eerie spell after 

https://github.com/pagefaultgames/pokerogue/assets/62030095/347ad4e7-547a-4c76-bfb8-2eff28aaacc1

## How to test the changes?
Use the src/overrides.ts file to give the opponent Spite/Eerie Spell, use moves and check the pp to make sure they are properly reducing. Also, test the 0pp visual bug by getting a move to <4pp and using it while the opponent spite's it before your turn.

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -4959,6 +4981,42 @@ export class ReducePpMoveAttr extends MoveEffectAttr {
...
+   * Checks if the target has used a move prior to the attack. PP-reduction is applied through the super class if so.
+   *
+   * @param user {@linkcode Pokemon} that used the attack
+   * @param target {@linkcode Pokemon} targeted by the attack
+   * @param move {@linkcode Move} being used
+   * @param args N/A
+   * @returns {boolean} true
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    const lastMove = target.getLastXMoves().find(() => true);
```
Tempo-anon, line 5003: Does this guarantee it gets the actual last move? find returns the first element

schmidtc1: getLastXMoves slices the move history by default from length-1 to length with no input, which obtains the last element. So there would only be one element in the array for .find to retrieve I believe.

Also, this logic comes from the original Spite implementation which correctly obtains the last move in the live version.

Tempo-anon: The fact that getLastXMoves returns an array of size 1 makes me upset

Tempo-anon (approved):

PR #2246, opened 2024-06-15
[Bug]: fix dark aura and fairy aura not applying to moves

OrangeRed:
## What are the changes?
Fix Dark Aura and Fairy Aura not applying a damage bonus

Dark Aura and Fairy Aura don't apply a damage bonus when they should

## What did change?
Find a pokemon with an aura on the field and apply the ability to the move

## How to test the changes?
Add overrides for `Abilities.DARK_AURA` and without and see the damage difference
I can write e2e tests if necessary

## Checklist

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -22,7 +22,7 @@ import { BattlerTag, BattlerTagLapseType, EncoreTag, HelpingHandTag, HighestStat
 import { WeatherType } from "../data/weather";
 import { TempBattleStat } from "../data/temp-battle-stat";
 import { ArenaTagSide, WeakenMoveScreenTag, WeakenMoveTypeTag } from "../data/arena-tag";
-import { Ability, AbAttr, BattleStatMultiplierAbAttr, BlockCritAbAttr, BonusCritAbAttr, BypassBurnDamageReductionAbAttr, FieldPriorityMoveImmunityAbAttr, FieldVariableMovePowerAbAttr, IgnoreOpponentStatChangesAbAttr, MoveImmunityAbAttr, MoveTypeChangeAttr, PreApplyBattlerTagAbAttr, PreDefendFullHpEndureAbAttr, ReceivedMoveDamageMultiplierAbAttr, ReduceStatusEffectDurationAbAttr, StabBoostAbAttr, StatusEffectImmunityAbAttr, TypeImmunityAbAttr, VariableMovePowerAbAttr, WeightMultiplierAbAttr, allAbilities, applyAbAttrs, applyBattleStatMultiplierAbAttrs, applyPreApplyBattlerTagAbAttrs, applyPreAttackAbAttrs, applyPreDefendAbAttrs, applyPreSetStatusAbAttrs, UnsuppressableAbilityAbAttr, SuppressFieldAbilitiesAbAttr, NoFusionAbilityAbAttr, MultCritAbAttr, IgnoreTypeImmunityAbAttr, DamageBoostAbAttr, IgnoreTypeStatusEffectImmunityAbAttr, ConditionalCritAbAttr, applyFieldBattleStatMultiplierAbAttrs, FieldMultiplyBattleStatAbAttr } from "../data/ability";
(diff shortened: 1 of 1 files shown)


src/field/pokemon.ts
```diff
@@ -1747,7 +1747,10 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
         power.value = 60;
       }
       applyPreAttackAbAttrs(VariableMovePowerAbAttr, source, this, move, power);
-      this.scene.getField(true).map(p => applyPreAttackAbAttrs(FieldVariableMovePowerAbAttr, this, source, move, power));
+      const fieldAuraPokemon = this.scene.getField(true).find(p => p.hasAbilityWithAttr(FieldVariableMovePowerAbAttr));
+      if (fieldAuraPokemon) {
+        applyPreAttackAbAttrs(FieldVariableMovePowerAbAttr, fieldAuraPokemon, fieldAuraPokemon, move, power);
+      }
```
torranx, line 1753: wont this be enough? 

```suggestion
      this.scene.getField(true).map(p => applyPreAttackAbAttrs(FieldVariableMovePowerAbAttr, p, this, move, power));
```

OrangeRed: Doing this would make the damage bonus from the ability apply for every pokemon on the field that has the aura ability.
If two pokemon have dark_aura then the damage bonus would apply twice. I didn't think that was desirable behavior so I stop after the first is found.

torranx: > Doing this would make the damage bonus from the ability apply for every pokemon on the field that has the aura ability. If two pokemon on the field have dark_aura then the damage bonus would apply twice. I didn't think that was desirable behavior so I stop after the first is found.

i see, i did not think of that. nice catch! but why are you passing the same pokemon on the `applyPreAttackAbAttrs`?

OrangeRed: it doesn't matter which you pass in as the second 'defender' pokemon since `FieldVariableMovePowerAbAttr` doesn't use the defender at all. Ideally in the future we would get rid of this unnecessary variable passing but there is a PR blocking me from doing that

torranx: gotcha, consider this resolved, thanks!

torranx (approved):
lgtm

src/field/pokemon.ts
```diff
@@ -1747,7 +1747,10 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
         power.value = 60;
       }
       applyPreAttackAbAttrs(VariableMovePowerAbAttr, source, this, move, power);
-      this.scene.getField(true).map(p => applyPreAttackAbAttrs(FieldVariableMovePowerAbAttr, this, source, move, power));
+      const fieldAuraPokemon = this.scene.getField(true).find(p => p.hasAbilityWithAttr(FieldVariableMovePowerAbAttr));
+      if (fieldAuraPokemon) {
```
XGSleepWalker, line 1751: What happens if there's a Dark Aura and Fairy Aura in the field at the same time? Won't just one of them apply?

XGSleepWalker (changes requested):

src/field/pokemon.ts
OrangeRed: if there are two different auras there is no way for them to apply to the same move?
since they don't have overlapping conditions.
```ts
// boostedType is Type.DARK for dark_aura and is Type.FAIRY for fairy_aura
super((pokemon, defender, move) => move.type === boostedType, powerMultiplier || 1.5);
```

OrangeRed:
I was having issues where using the `applyPreAttackAbAttr()` method would activate both a dark aura and a fairy aura if a pokemon had both even if a different pokemon had already activated one of the auras.

Using `applyPreAttackAbAttr()` this would happen:
PokemonA: Dark Aura only
PokemonB: Dark Aura + Fairy Aura
Result -> the move's power would get boosted twice by Dark Aura

So I had to be more granular and use the AbAttr's `applyPreAttack()` directly

src/field/pokemon.ts
```diff
@@ -1747,7 +1747,16 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
         power.value = 60;
       }
       applyPreAttackAbAttrs(VariableMovePowerAbAttr, source, this, move, power);
-      this.scene.getField(true).map(p => applyPreAttackAbAttrs(FieldVariableMovePowerAbAttr, this, source, move, power));
+
+      const fieldAuras = new Set(
+        this.scene.getField(true)
+          .map((p) => p.getAbilityAttrs(FieldVariableMovePowerAbAttr) as FieldVariableMovePowerAbAttr[])
```
torranx, line 1753: can we change this to `FieldMoveTypePowerBoostAbAttr` instead? since only `FieldMoveTypePowerBoostAbAttr` extends `FieldVariableMovePowerAbAttr`. this would also "isolate" Dark/Fairy Aura

changed it on my PR https://github.com/pagefaultgames/pokerogue/pull/2268/files#diff-1867e67cee517f3348c17ed44e0f02a45a23e59575f38b6e15aacf194768d6dbR1755

OrangeRed: done

torranx (approved):

XGSleepWalker (approved):

PR #2255, opened 2024-06-15
[Sprite] Replace Togedemaru Animation Datafiles

chaosgrimmon:
## What are the changes?
Both the front and back of Togedemaru(777)'s animated sprites are currently affected because their spritesheets were significantly updated without the animation frame data changing to match. This PR gives them animations that better suit the current spritesheets.

The following has been repeatedly reported.
image
image
image

## What did change?
Since the non-shiny backsprite was updated to match the shiny backsprite, that .json was trivial to replace.
Although this could have been also done for the non-shiny frontsprite, it was requested [here](https://discord.com/channels/1125469663833370665/1229465263532019712/1251422382745325609) that the animation be stretched slightly by adding more frames. As a result, this change was also applied to the shiny frontsprite, necessitating its modification.

The wholesale replacement was caused by a need to correctly re-number a significant majority of its frames, and so this was done programmatically, instead of manually editing the file. This caused intimidating diff blocks, but very little actual difference, save for the extra blocks corresponding to newly-inserted frames.

#### Before
image
image

image
#### After
image
image
image

## How to test the changes?
Anywhere a Togedemaru can be found, though variants may benefit from using overrides.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/public/images/pokemon/exp/777.json
+++ b/public/images/pokemon/exp/777.json
@@ -4,639 +4,30 @@
 			"image": "777.png",
 			"format": "RGBA8888",
 			"size": {
-				"w": 128,
-				"h": 128
+				"w": 163,
+				"h": 163
 			},
 			"scale": 1,
 			"frames": [
-				{
-					"filename": "0045.png",
-					"rotated": false,
-					"trimmed": true,
-					"sourceSize": {
-						"w": 45,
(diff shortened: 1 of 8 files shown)


Tempo-anon (approved):
Thanks!

PR #2268, opened 2024-06-15
[Ability] Implement Power Spot & Battery

torranx:
## What are the changes?
- implements Power Spot and Battery

- implements Power Spot and Battery

## What did change?
- removed `FieldVariableMovePowerAbAttr` and it seems more logical to call `FieldMoveTypePowerBoostAbAttr` in https://github.com/pagefaultgames/pokerogue/pull/2268/files#diff-1867e67cee517f3348c17ed44e0f02a45a23e59575f38b6e15aacf194768d6dbR1755 as this is only being used by Dark/Fairy Aura 
- new class `AllyMoveCategoryPowerBoostAbAttr` for Battery and Power Spot

### Screenshots/Videos
Refer to console logs for before and after power value

- [Power Spot](https://bulbapedia.bulbagarden.net/wiki/Power_Spot_(Ability))

https://github.com/pagefaultgames/pokerogue/assets/68144167/15572c22-28de-4789-aefa-8e132f6147ac

- [Battery](https://bulbapedia.bulbagarden.net/wiki/Battery_(Ability))

https://github.com/pagefaultgames/pokerogue/assets/68144167/c2502c53-76c7-4a5e-b480-cd28b57b51bc

## Checklist


src/data/ability.ts
```diff
@@ -3426,6 +3426,18 @@ export class NoFusionAbilityAbAttr extends AbAttr {
...
+    super(false);
+  }
+}
+
+export class PowerSpotAbAttr extends AbAttr {
+  constructor() {
+    super(false);
+  }
+}
+
```
OrangeRed, line 3440: Rather than make two different classes for this wouldn't it be better to extend `FieldVariableMovePowerAbAttr` and update the move's power via a condition instead?

Tempo-anon: I agree with that. The only different thing is the conditions

torranx: I think I've tried this but the problem was it _should_ apply to the ally only, hence why I resorted to placing the condition on `src/field/pokemon.ts`. It gets real messy when extending `FieldVariableMovePowerAbAttr` but I'll check again.

OrangeRed: Don't plus and minus do a similar thing to these abilities? In the AbAttr that those use there is a condition that checks for a double battle and if there is an ally

torranx: Plus and Minus targets the user of the ability itself and not its ally, and also targets battlestat instead of move. I believe we dont have an implemented ability similar to Battery/Power Spot where it targets the ally's move.

hayuna: I had a similar ability implementation like Steely Spirit. Didn't find a way to not include my code in pokemon.ts as well

https://github.com/pagefaultgames/pokerogue/pull/1439

torranx: After deliberately finding a way to make this happen, I'm now convinced there is no way to implement it the way we normally implement abilities, for the ff reasons:
- the target of these abilities is the **ally's move power**
- we can only get the current move from the pokemon which has the active turn
- because of 2nd bullet, we can't apply multiplier to the move's power dynamically inside the ability attribute

If I'm missing something or there is a mistake I'm having, please tell me. Otherwise, I'll clean this PR, add unit tests and move it from draft.

OrangeRed: > the target of these abilities is the ally's move power
> we can only get the current move from the pokemon which has the active turn
> because of 2nd bullet, we can't apply multiplier to the move's power dynamically inside the ability attribute

Why not? Currently Dark Aura and Fairy Aura do this although they're a little bugged. 
Prior to a move executing we check if any pokemon on the field has one of those abilities and activate it while passing in the power of the move currently being used as `args[0]`. These abilities then mutate the value of that passed power if the condition returns true (the move is either dark or fairy)

torranx: I've rechecked again and I think I've finally nailed the code right. I was overcomplicating it on my head when trying to implement it using ability attributes like Dark/Fairy Aura. Thanks for holding up with your idea!

torranx: I've pushed an update that use the same logic as the Dark/Fairy Aura. The parent class `UserFieldVariableMovePowerAbAttr` I created could also be used by other moves that function similar like Steely Spirit cc. @hayuna . 

If you're wondering why I did not extend `FieldVariableMovePowerAbAttr` instead, I needed to separate this from where the `FieldVariableMovePowerAbAttr` is called- it's because of the condition that the move power boost should only apply if the pokemon's ally has Battery/Power Spot. LMK if I missed anything @OrangeRed

src/test/abilities/battery.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    game.doAttack(getMovePosition(game.scene, 0, moveToBeUsed));
+    game.doAttack(getMovePosition(game.scene, 1, Moves.SPLASH));
+
+    const appliedPower = getAppliedMovePower(game.scene.getEnemyField()[0], game.scene.getPlayerField()[0], allMoves[moveToBeUsed]);
+
+    await game.phaseInterceptor.to(TurnEndPhase);
+
+    expect(appliedPower).not.toBe(undefined);
+    expect(appliedPower).toBe(basePower);
+    expect(appliedPower).not.toBe(basePower * batteryMultiplier);
```
OrangeRed, line 87: Instead of defining a constant `batteryMultiplier` can you instead `spyOn` the value from inside the AbAttr class (this will require a small change to the class to add powerMultipiler` as a property`

src/data/ability.ts
```diff
@@ -1310,12 +1311,34 @@ export class FieldMovePowerBoostAbAttr extends FieldVariableMovePowerAbAttr {
...
   constructor(boostedType: Type, powerMultiplier?: number) {
     super((pokemon, defender, move) => move.type === boostedType, powerMultiplier || 1.5);
   }
 }
 
+/**
+ * Boosts the power of moves in specified categories.
+ * @extends FieldMovePowerBoostAbAttr
+ */
+export class AllyMoveCategoryPowerBoostAbAttr extends FieldMovePowerBoostAbAttr {
```
OrangeRed, line 1332: Is there a point in explicitly calling this `Ally...` when the ally check happens in `pokemon.ts` rather than this AbAttr?

torranx: I wanted to specify the name as happening on the ally's side because im extending the `FieldMovePowerBoostAbAttr`, and if ever there is a need to create a class of `MoveCategoryPowerBoostAbAttr` but would extend `MoveTypePowerBoostAbAttr` in this case.

OrangeRed (approved):
Looks good just wait for #2246 to be merged in so you can add that to this PR

Tempo-anon (approved):

PR #2275, opened 2024-06-15
fix import eslint module not working in new flat eslint config

meepen:
## What are the changes?

Fixes eslint rules not working for the eslint-plugin-import module

To fix eslint in preparation of cleanup of circular dependencies

## What did change?

Fixes the lines that conflict with the rules that now work, and fixes the eslint config

## How to test the changes?

You can pull only the eslint config and package changes to see the eslint rules working on the broken files that were updated

## Checklist


brain-frog (dismissed):

flx-sta:
@meepen do you think you can also get it to work with eslint v9?

Cause we are going to update to that soon following #3160.
I've checked the `eslint-plugin-import` github and it seems they haven't caught up with eslint 9 yet so I was wondering if you might know more about this

DayKev:
@flx-sta is this PR still relevant?

flx-sta:
Not unless it’s updated for eslint 9

Tempo-anon:
Looks like this is not updated for eslint 9 and has a lot of conflicting files. Will close out this PR for now.

PR #2287, opened 2024-06-16
[Ability] Implement Quick Draw

ysh4296:
## What are the changes?

- implementing quick draw
- now slowbro(galarian from) has quick draw

- add ability

## What did change?

- at turn order determine step the BypassSpeedChanceAbAttr is applied to set  the pokemon has bypassSpeed (like quick claw)
- it applied before quick claw modifier effect
- it activate only when turn command was Fight and move has power (doing damage)

### Screenshots/Videos

- when quick draw activated

https://github.com/pagefaultgames/pokerogue/assets/29995264/db79a18f-eb20-41fa-8e65-71ec199227ef

- quick claw still works fine

https://github.com/pagefaultgames/pokerogue/assets/29995264/5ab8a83f-e6ae-47b5-aa77-19c364fd1892

## How to test the changes?

- i checked it with manually and unit test.
- manual
  - i used override.ts to test ability with slowbro
  - i gave 3 quick claws to slowbro to test Interaction between quick draw and quick claw
- unit test (npm run test quick_draw)
  - makes pokemon going first in its priority bracket
  - does not triggered by non damage moves
  - does not increase priority
 
image

## Checklist
- [X] Are the changes visual?


torranx:
unit tests would be nice to have

ysh4296:
agree i'll reopen pr with unit test thank you!

src/test/abilities/quick_draw.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      Moves.TACKLE,
+      Moves.TACKLE,
+      Moves.TACKLE,
+      Moves.TACKLE,
+    ]);
+
+    allAbilities[Abilities.QUICK_DRAW] = new Ability(
+      Abilities.QUICK_DRAW,
+      8
+    ).attr(BypassSpeedChanceAbAttr, 100);
```
ysh4296, line 54: i tested attribute with this way (to ensure ability has 100% chance to active)

any other suggestions?

ysh4296:
reopened with unit test!

src/data/ability.ts
```diff
@@ -3598,6 +3598,50 @@ export class IceFaceMoveImmunityAbAttr extends MoveImmunityAbAttr {
...
+      const hasPower = move?.power !== -1;
+
+      if (isCommandFight && hasPower) {
+        bypassSpeed.value = true;
+        pokemon.scene.queueMessage(
+          getPokemonMessage(
+            pokemon,
+            " can act faster than normal, thanks to its Quick Draw!"
+          )
+        );
```
OrangeRed, line 3636: Can you move this to `this.getTriggerMessage()`. This AbAttr property is invoked when `this.apply()` returns true.

src/test/abilities/quick_draw.test.ts
OrangeRed: Rather than globally setting it can you instead `spyOn` it?

src/test/abilities/quick_draw.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    game.doAttack(getMovePosition(game.scene, 0, Moves.TACKLE));
+
+    await game.phaseInterceptor.run(EnemyCommandPhase);
+    await game.phaseInterceptor.run(TurnStartPhase);
+
+    await game.phaseInterceptor.to(TurnEndPhase);
+
+    expect(game.phaseInterceptor.log).toContain("ShowAbilityPhase");
+    expect(game.phaseInterceptor.log).toContain("VictoryPhase");
```
OrangeRed, line 72: Rather than checking the phaseInterceptor log can't you instead check the `Pokemon.battleData.abilitiesApplied` or `Pokemon.summonData.abilitiesApplied` array?

src/test/abilities/quick_draw.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    game.doAttack(getMovePosition(game.scene, 0, Moves.TACKLE));
+
+    await game.phaseInterceptor.run(EnemyCommandPhase);
+    await game.phaseInterceptor.run(TurnStartPhase);
+
+    await game.phaseInterceptor.to(TitlePhase);
+
+    expect(game.phaseInterceptor.log).toContain("ShowAbilityPhase");
+    expect(game.phaseInterceptor.log).not.toContain("VictoryPhase");
```
OrangeRed, line 121: Instead of verifying which pokemon went first by checking for the victory phase can't you also check `scene.currentBattle.turnCommands` to determine move priority?

ysh4296: I think `turnCommands` is not suitable for determining move priority.

`turnCommands` only shows which field index selects which move, but it does not provide the actual move order.

There is an internal variable `const moveOrder` in `TurnStartPhase`, which I believe represents the actual move order.

How about saving `moveOrder` in `pokemon.battleData` at `MovePhase` and use it to check the actual move order?

Please let me know if I'm wrong.

src/data/ability.ts
ysh4296: modefied thank you!

src/test/abilities/quick_draw.test.ts
ysh4296: modefied thank you!

src/test/abilities/quick_draw.test.ts
ysh4296:
``` typescript
vi.spyOn(
  allAbilities[Abilities.QUICK_DRAW].getAttrs(BypassSpeedChanceAbAttr)[0],
  "getCondition"
).mockReturnValue(null);
```

- i changed .attr() to .conditionalAttr()
- now i could use `spyOn` to getCondition function to null
- bypass canApplyAttr() function & ability always activated.

any other suggestions?

ysh4296: i changed `private chance` to `public` and spyOn the chance to 100.

``` typescript
vi.spyOn(
  allAbilities[Abilities.QUICK_DRAW].getAttrs(BypassSpeedChanceAbAttr)[0],
  "chance","get"
).mockReturnValue(100);
```

i hope this would be better solution.

ysh4296:
i also added localizations

OrangeRed (approved):

OrangeRed (approved):

src/data/ability.ts
```diff
@@ -3732,6 +3732,54 @@ export class IceFaceBlockPhysicalAbAttr extends ReceivedMoveDamageMultiplierAbAt
...
+ */
+export class BypassSpeedChanceAbAttr extends AbAttr {
+  public chance: integer;
+
+  constructor(chance: integer) {
+    super(true);
+    this.chance = chance;
+  }
+
+  apply(
```
Tempo-anon, line 3748: Add tsdoc to the `apply` functions

src/data/ability.ts
```diff
@@ -3732,6 +3732,54 @@ export class IceFaceBlockPhysicalAbAttr extends ReceivedMoveDamageMultiplierAbAt
...
+      if (isCommandFight && hasPower) {
+        bypassSpeed.value = true;
+        return true;
+      }
+    }
+
+    return false;
+  }
+
+  getTriggerMessage(
```
Tempo-anon, line 3772: not sure how I feel about the formatting of these functions. On one hand it's nice seeing all the variables in their own lines, but that this is inconsistent with the rest of the codebase and tsdoc also lists variables on their own line

src/data/ability.ts
```diff
@@ -3732,6 +3732,54 @@ export class IceFaceBlockPhysicalAbAttr extends ReceivedMoveDamageMultiplierAbAt
...
+
+  apply(
+    pokemon: Pokemon,
+    passive: boolean,
+    cancelled: Utils.BooleanHolder,
+    args: any[]
+  ): boolean {
+    const bypassSpeed = args[0] as Utils.BooleanHolder;
+
+    if (!bypassSpeed.value && pokemon.randSeedInt(100) < this.chance) {
```
Tempo-anon, line 3756: I assume the `!bypassSpeed.value` is the quick claw check?

src/data/ability.ts
```diff
@@ -3732,6 +3732,54 @@ export class IceFaceBlockPhysicalAbAttr extends ReceivedMoveDamageMultiplierAbAt
...
+      const turnCommand =
+        pokemon.scene.currentBattle.turnCommands[pokemon.getBattlerIndex()];
+      const isCommandFight = turnCommand?.command === Command.FIGHT;
+      const move = allMoves[turnCommand.move?.move];
+      const hasPower = move?.power !== -1;
```
Tempo-anon, line 3761: Does this work correctly with how some damaging moves have -1 listed power?

src/data/ability.ts
OrangeRed: They were probably using prettier and which auto formatted the file for them.

src/data/ability.ts
```diff
@@ -4815,7 +4863,7 @@ export function initAbilities() {
       .attr(NoFusionAbilityAbAttr)
       .condition((pokemon) => !pokemon.isTerastallized()),
     new Ability(Abilities.QUICK_DRAW, 8)
-      .unimplemented(),
+      .attr(BypassSpeedChanceAbAttr,30),
```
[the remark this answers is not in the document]
ysh4296: do you mean space?

src/data/ability.ts
ysh4296: no. quick draw works before quick claw
i added `!bypassSpeed.value`  just for safty
but since it has no practical effect i'll remove it

src/data/ability.ts
ysh4296: yes my prettier automatically formatted it! sorry.
i'll modify it!

src/data/ability.ts
ysh4296: oh i made a mistake...

currently some damaging move with power -1 (power is not specified by itself, but is assigned after move active) not working properly.

cheking damaging move has changed to `check move.category` (category is `PHYSICAL` or `SPECIAL`)

i hope this would be better solution.

src/data/ability.ts
ysh4296: added tsdoc to `apply` and `constructor`

src/data/ability.ts
OrangeRed: > but since it has no practical effect i'll remove it

Don't remove it. we don't have strict null checking so `args[0]` could be null or undefined and we want to guard against that.

ysh4296: Understood, thank you.

Tempo-anon (approved):

PR #2291, opened 2024-06-16
[Bug] Friendship-based moves have Base Power -1 when Pokemon is wild

chaosgrimmon:
## What are the changes?
Friendship-based moves, which do a percentage of their maximum damage based on the user's percentage of friendship level, were previously only friendship-dependent for `PlayerPokemon` 'mons.
Pokemon in wild encounters would instead find their move to have a base power of minus one.

This PR fixes that, by scaling it to the base friendship for that Species. Essentially, the Move will be exactly as powerful during a wild encounter, as it will be immediately after you've caught them.

This bug affects RETURN, FRUSTRATION, PIKA_POW, and VEEVEE_VOLLEY.
The full effects of the bug are not immediately apparent in the level range of the typical Classic run. At those levels, friendship-based moves *will* be insultingly weaker than other moves, but believably so for moves that mention the bond between Trainer and Pokemon in their descriptions.
image
image

At higher levels, the resulting damage, from `-1` base power, will also be negative, causing the target to gain HP.
image
image
[The attacking Pokemon was Lv.2000.]

Although the HP does not persist between encounters, it's still not intended or canonically-consistent behaviour.

## What did change?
Instead of leaving the base power unchanged when the user is *not* a PlayerPokemon, defaulting to the `-1` specified in the declaration of each `new AttackMove()` in `moves.ts`, the section of the formula that requires the user's friendship level is now a ternary, `user.friendship` if the Pokemon is a PlayerPokemon and `user.species.baseFriendship` to get the default value for the wild Pokemon's Species. 

#### Before
image
image
#### After (at Lv.2000)
image
image

## How to test the changes?
Overriding opponent moveset guarantees that relevant Moves are used by the opponent. Overriding opponent level is not necessary to ensure functionality, but was made use of.
```ts
export const OPP_MOVESET_OVERRIDE: Array<Moves> = [Moves.VEEVEE_VOLLEY, Moves.PIKA_PAPOW, Moves.RETURN, Moves.FRUSTRATION];
```

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -3034,10 +3034,8 @@ export class FriendshipPowerAttr extends VariablePowerAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
     const power = args[0] as Utils.NumberHolder;
 
-    if (user instanceof PlayerPokemon) {
-      const friendshipPower = Math.floor(Math.min(user.friendship, 255) / 2.5);
-      power.value = Math.max(!this.invert ? friendshipPower : 102 - friendshipPower, 1);
-    }
+    const friendshipPower = Math.floor(Math.min(user instanceof PlayerPokemon ? user.friendship : user.species.baseFriendship, 255) / 2.5);
+    power.value = Math.max(!this.invert ? friendshipPower : 102 - friendshipPower, 1);
 
     return true;
   }


chaosgrimmon:
Additional reading:
https://discord.com/channels/1125469663833370665/1251763288870617108
https://discord.com/channels/1125469663833370665/1176874654015684739/1251792831006117990

Tempo-anon (approved):

PR #2303, opened 2024-06-16
[Move] Sappy Seed shouldn't be Making Contact

chaosgrimmon:
## What are the changes?
Previously, [Sappy Seed](https://bulbapedia.bulbagarden.net/wiki/Sappy_Seed_(move)), being a Physical Attack, was assumed to be a contacting attack in the absence of a flag indicating otherwise. This meant users might be accidentally placed in harm's way, whether by [Flame Body](https://bulbapedia.bulbagarden.net/wiki/Flame_Body_(Ability)), [Iron Barbs](https://bulbapedia.bulbagarden.net/wiki/Iron_Barbs_(Ability)), etc.
This PR ensures that Sappy Seed is not a contacting attack, and so doesn't trigger those effects.

image

Players expecting not to receive contact effects will be surprised if Sappy Seed does not behave as it does in other games, particularly given the other projectile-based Seed moves.
It was reported [here](https://discord.com/channels/1125469663833370665/1251886856673558528).

## What did change?
The attribute `makesContact` is set to `false`.

#### Before (note Sappy Seed being the most recent move in the console logs)
image
#### After
https://github.com/pagefaultgames/pokerogue/assets/31082757/c94b3d12-fb5a-4522-a5ca-a3bdf3ce9967

## How to test the changes?
Overriding the opponent to have the abiiity Iron Barbs helps ensure that all contacting moves are clearly indicated.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -7578,6 +7578,7 @@ export function initMoves() {
     new AttackMove(Moves.BADDY_BAD, Type.DARK, MoveCategory.SPECIAL, 80, 95, 15, -1, 0, 7)
       .attr(AddArenaTagAttr, ArenaTagType.REFLECT, 5, false, true),
     new AttackMove(Moves.SAPPY_SEED, Type.GRASS, MoveCategory.PHYSICAL, 100, 90, 10, 100, 0, 7)
+      .makesContact(false)
       .attr(AddBattlerTagAttr, BattlerTagType.SEEDED),
     new AttackMove(Moves.FREEZY_FROST, Type.ICE, MoveCategory.SPECIAL, 100, 90, 10, -1, 0, 7)
       .attr(ResetStatsAttr),


CodeTappert:
Is this a broader problem? Also for other moves?

Tempo-anon (approved):

chaosgrimmon:
> Is this a broader problem? Also for other moves?

I can't immediately be sure. The generally accepted method would likely entail referencing of the chart on https://bulbapedia.bulbagarden.net/wiki/Contact.
![image] (https://github.com/pagefaultgames/pokerogue/assets/31082757/203e0e5a-d551-4e1c-81ab-32988a64bb29)

However, that chart doesn't have Sappy Seed, and, upon cursory inspection, Astral Barrage. As a result, Moves.ASTRAL_BARRAGE is similarly not `.makesContact(false)` in the code, despite obviously involving ranged stellar projectiles.

Ideally, someone with a greater familiarity with the global moveset should be doing this, in order to catch anomalies, but yeah I'll probably take a glance and make another PR with things that immediately stand out.

chaosgrimmon:
Sidenote: Like the initial PR mentions, this fix was caused by a helpful reporter specifically regarding Sappy Seed, and at the time I assumed it was a one-off event, with an expected factor likely being that the LG!P/E moves had been added at a later point.

PR #1754, opened 2024-06-03
[QOL] Updating pokemon-info-container to show what is new when you catch/hatch a pokemon

Opaque02:
## What are the changes?
When catching/hatching a pokemon, the pokemon info container now shows what is new in the container. The gender icon has been moved to the top left and has been replaced with the form (if the pokemon has one). This PR shows if any of the following are new: gender, form (if applicable), ability, nature and shiny/variant. 

The shiny/variant and gender will have a blue (+) next to it if it's new, whereas the form, ability and nature will have blue text. Also since the form names can be quite long, this also adds the ability to hover over the form area to see the form name with a hover indicator like the shiny icon

## Why am I doing these changes?
I'm doing these changes because I was playing endless and kept finding pokemon that had a tinted pokeball, but when I caught it, I didn't know why it was tinted. So I made these changes to help me see what's new with pokemon

## What did change?
All these changes are within the pokemon info container
1. The gender text has been replaced with a form section, which includes a "Form:" label and a form text section. If a pokemon has a form, this section will show up, and if it's a new form, the "Form:" label will be blue, otherwise white. Hovering over the form name will show the name of the form in a new popup. Also if the form name is too long to fit on the screen, the text will be truncated and replaced with a ... at the end
2. The gender icon has been moved to the top left, with a new blue (+) indicator if it's a new gender you haven't previously caught
3. The shiny icon also has a (+) indicator if it's a new variant/shiny version
4. The ability and nature labels will now show as blue if either of them are new
5. The y location of the form, ability and nature has been moved down to make more space for the text
6. The stat labels in the iv chart have been moved around to make more space for the text. Specifically the speed label has been moved up, and then the Sp. Def and Def have been moved up slightly to bring them closer to the middle of the IV hexagon, and the Sp. Atk and Atk have been moved down slightly to match

### Screenshots/Videos
I did a lot of testing using scatterbug as it's got lots of forms to test:
This image shows me catching a scatterbug with a new nature, new shiny and new form
image

This shows me catching a new gender, form, ability and nature:
image

This shows a new form and that hovering over the form shows the form name, AND that the form name is truncated and "..." has been added to the end:
image

This shows a new nature and existing form, and showing that hovering over the form shows the form name:
image

This shows a new form and a new variant - there's a (+) in the bottom left next to the shiny icon to show it's a variant we haven't gotten before - in this case because it's a blue shiny:
image

This shows a capture of nothing new:
image

[...]

PR #2314, opened 2024-06-17
[QoL] Updated form name logic to not show on certain pokemon

Opaque02:
## What are the changes?
For the pokemon info container changes from #1754, forms were shown on the pokemon info box if the pokemon caught had a form name. This PR changes it so that the form label doesn't show up as often

I'm doing these changes to make it clearer for players when it comes to form names, and to hide forms that shouldn't necessarily be shown.

## What did change?
Currently on live, there's a check to see if the caught/hatched pokemon has a form name at all. If so, it will show the form label. This works well for some pokemon like Scatterbug, Deerling or Unown, where the form is cosmetic only. However, it would also pick up form names for forms that weren't purely cosmetic. For example, Gyarados can mega, which is a separate form. This means for pokemon that can mega, gmax, primal etc, they have a base form - or "Normal" form - and their mega/gmax etc form(s). While for Scatterbug and Deerling, every form can be caught from the wild or hatched, but you can't catch/hatch megas like mega gyarados. This means that for pokemon that can mega/gmax/primal/etc, the only version of them you'll see in the wild/eggs are their base "Normal" forms. 

Since it doesn't make sense to only show a form label of "Normal" for these pokemon (since that's the _only_ form you can catch/hatch), this PR changes the logic for the form label accordingly:
1. If the pokemon species doesn't have any forms, don't show the form label (this is unchanged)
2. Check the pokemon's species' forms against the [noStarterFormKeys](https://github.com/pagefaultgames/pokerogue/blob/c3c6b96487a9cc92ba885baa4784521bd7457867/src/data/pokemon-species.ts#L3245) list from the pokemon-species.ts file, and remove all forms that have a formKey on the noStarterFormKey list
3. If the remaining list has a length of 1, this means that there's only one form that pokemon species has that's not a mega/primal etc; this is usually "Normal" form. This means if the length of the remaining list is 1, to not show the form label as it's the only form of the pokemon that isn't a mega/primal etc

**NOTE:** I've applied the above logic changes to all the pokemon in the game and have generated a [text file here](https://github.com/user-attachments/files/15860379/Pokemon.with.different.forms.txt)
with a list of all the pokemon and whether or not they should or shouldn't have their form labels shown. Now, I'm not great at pokemon and the only mainline games I've played are all gen 1 games where there haven't been forms, so I didn't even know forms existed until I started playing Pokerogue. This means I'm not an expert regarding forms, and I don't know for sure if this is the best method to find whether or not a form should be shown or not. I'm attaching the .txt file here so others can look through this and see if it might need adjusting. There's also an argument to be made for using the [pokemon-forms.ts](https://github.com/pagefaultgames/pokerogue/blob/main/src/data/pokemon-forms.ts) file instead of the pokemon-species.ts file I used, but I used the latter due to talking to people in the discord who mentioned doing it this way; again, I'm not an expert so I'm happy to change the logic if there's a better way.

### Screenshots/Videos
<details>
  <summary>This is an example of catching a mew previously (no form)</summary>

image

</details>
<details>
  <summary>This is an example of catching a mew with the update (no form). As you can see, this is unchanged</summary>

image

</details>
<details>
  <summary>This is an example of catching a scatterbug currently (has a cosmetic form only)</summary>

image

</details>
<details>
  <summary>This is an example of catching a scatterbug after the logic change. As you can see, this is unchanged</summary>

image

</details>
<details>
  <summary>This is an example of catching a gyarados in live. As you can see, it comes up with the form "Normal"</summary>

image

</details>
<details>
  <summary>This is an example of catching a gyarados after the changes</summary>
As you can see, there's now no more form label for gyarados, since the only two forms are mega and normal, and since you can't catch/hatch mega gyarados, the only form available is normal, which hides the label:

image

</details>

## How to test the changes?
There's a few things in overrides file you can change:

Lines 42-49 - giving yourself 99 masterballs to start with:

```
  active: true,
  pokeballs: {
    [PokeballType.POKEBALL]: 5,
    [PokeballType.GREAT_BALL]: 0,
    [PokeballType.ULTRA_BALL]: 0,
    [PokeballType.ROGUE_BALL]: 0,
    [PokeballType.MASTER_BALL]: 99,
  }
```

Line 79 - changing the enemy species to gyarados for testing forms:

`export const OPP_SPECIES_OVERRIDE: Species | integer = Species.GYARADOS;`

**NOTE:** You can change the enemy pokemon to anything using the Species.{PokemonName} in line 79 of the overrides (above). For example, if you want to change it to Scatterbug or Charizard, you could do Species.SCATTERBUG or Species.CHARIZARD respectively. However, if you wanted to, you could also make an array of multiple pokemon, and the game will choose one at random. For example, the following override will make it so that only gyarados, scatterbug, charmander, mew and bulbasaur can spawn:

`export const OPP_SPECIES_OVERRIDE: Species | integer = [Species.GYARADOS, Speices.SCATTERBUG, Species.CHARMANDER, Species.MEW, Species.BULBASAUR];`

You can add as many pokemon as you'd like to this array and you'll get a chance to spawn them randomly as you play (though I believe rarities may still apply?)

## Checklist
- [x] Are the changes visual?


src/ui/pokemon-info-container.ts
```diff
@@ -226,7 +227,19 @@ export default class PokemonInfoContainer extends Phaser.GameObjects.Container {
         this.pokemonGenderText.setVisible(false);
       }
 
-      if (pokemon.species.forms?.[pokemon.formIndex]?.formName) {
+      let isFormVisible = true;
```
Tempo-anon, line 230: Should probably add a comment here explaining what this chunk of code is doing

chaosgrimmon:
shortlist of all affected mons
```py
25	Pikachu: true
133	Eevee: true
172	Pichu: true
201	Unown: true
351	Castform: true
386	Deoxys: true
412	Burmy: true
413	Wormadam: true
421	Cherrim: true
422	Shellos: true
423	Gastrodon: true
479	Rotom: true
492	Shaymin: true
493	Arceus: true
550	Basculin: true
555	Darmanitan: true
585	Deerling: true
586	Sawsbuck: true
646	Kyurem: true
647	Keldeo: true
648	Meloetta: true
649	Genesect: true
656	Froakie: true
657	Frogadier: true
658	Greninja: true
664	Scatterbug: true
665	Spewpa: true
666	Vivillon: true
669	Flabébé: true
670	Floette: true
671	Florges: true
676	Furfrou: true
678	Meowstic: true
681	Aegislash: true
710	Pumpkaboo: true
711	Gourgeist: true
716	Xerneas: true
718	Zygarde: true
720	Hoopa: true
741	Oricorio: true
744	Rockruff: true
745	Lycanroc: true
746	Wishiwashi: true
773	Silvally: true
774	Minior: true
778	Mimikyu: true
800	Necrozma: true
801	Magearna: true
802	Marshadow: true
845	Cramorant: true
849	Toxtricity: true
854	Sinistea: true
855	Polteageist: true
869	Alcremie: true
875	Eiscue: true
876	Indeedee: true
877	Morpeko: true
888	Zacian: true
889	Zamazenta: true
892	Urshifu: true
893	Zarude: true
898	Calyrex: true
902	Basculegion: true
916	Oinkologne: true
925	Maushold: true
931	Squawkabilly: true
964	Palafin: true
978	Tatsugiri: true
982	Dudunsparce: true
999	Gimmighoul: true
1007	Koraidon: true
1008	Miraidon: true
1012	Poltchageist: true
1013	Sinistcha: true
1017	Ogerpon: true
1024	Terapagos: true
1061	Darmanitan: true
1080	Tauros: true
```

src/ui/pokemon-info-container.ts
Opaque02: Good point - it's been fixed and commented and pushed :)

chaosgrimmon:
See https://bulbapedia.bulbagarden.net/wiki/List_of_Pok%C3%A9mon_with_form_differences for more information.

What I'll note is that there are a few Item-dependent form(e)s in there.
I'm pretty sure that you can't start with Arceus in a Plate-having form without a Plate, nor either Rider Calyrex without that set of Reins.
There's a few of those here, including Deoxys, Genesect, Silvally, Hoopa, Zacian, Zazamenta, Urshifu, Ogerpon if I recall correctly.


Pretty sure you need to Fuse Kyurem for its forms, so that one's out too.

Same for Necrozma.

Morpeko's form is Ability-based, defaults to Full Belly, and tbh shouldn't even show up in Starter Select, much less here.

Eiscue is also because of its Ability, probably shouldn't be selectable in No-Ice Face mode, though tbh if the Balance Team gave them more abilities provided you don't take Ice Face I'd be cool with that. However, at least it doesn't proc CycleForm in Starter Select.

Mimikyu's Disguise is literally the same.

Cramorant's Gulp Missile is combat dependent, based on using Surf/Dive.

Wishiwashi has Schooling.


Terpagos form changes by Terastallising, and so it does not show up in Starter Select.

Zenith Marshadow is a combat effect, so actually that also needs to be reflected in Starter Select, where currently CycleForm is visible.

Xerneas uses Neutral mode when not in combat/Contests, Active when it is, and also needs to be fixed for Starter Select.

Aegislash formes are caused by its Ability, Stance Change.

Both Darmanitan regionals can have Zen Mode.

Palafin's from is, again, Ability-dependent, and does not affect Finizen, and so does not have any Starter implications.

What *do* the Miraidon/Koraidon forms do? Roguedex says that only the base form is available (in Lab/Ruins, respectively) so I don't know where they're coming from. They don't seem to offer mechanical advantage (no Move, Ability, Stat differences), unless you consider "literally none of the alt Forms have sprites, shinys, or animations at all" a mechanical advantage. (You shouldn't. Unless seeing more of the Field is good somehow?)
Like, I know in Gen9 that matters because they're unlocked over time kind of like how HMs used to be, but here... ???

Dudunsparce (Three-Segment) is the evolution counterpart of a shiny? So I'd have to look into if that's actually a thing that can be chosen in Starter Select, or if that's calculated in the Evolution Phase to know if that's relevant here.

Same for Tandemaus -> Maushold (Family of Three/Four).

For sure though that's the case for Alcremie. Milcery's evo in PRogue is dependent on biome, and cannot be chosen at Select.

I *think*? Meowstic's form is because they've got major gender differences and that's how Ability/Learnset differences would be implemented? (Though `level-moves` seems to just use the male learnset so ???) (Hidden Ability diffs are in tho)
Point is, that's handled by Espurr's CycleGender, and it doesn't get a CycleForm anyways.
I could see keeping this one though because it does let people know that gender affects your Meowstic results, so this one could be an edge case.
On the other hand, Ralts (Gardevoir/Gallade) doesn't, so it's whatver.

On the other other hand, Indeedee is a starter, and on Starter Select, both CycleGender and CycleForm cycle both at the same time.

Same for Lechonk -> Oinkologne (M/F)
Same for Basculin (White-Striped) -> Basculegion (M/F)
        (note to self, Basculin's Ability_1 isn't changing as expected in Starter Select. It does appear to show up correctly in team summary though.)

Same for Toxtricity, though its based on Nature, not Gender.

This is kind of like how Lycanroc works. Though Rockruff gets the "", "Own Tempo" options, as is ideal, non-OWN_TEMPO Lycanroc also depends on Time of Day, so it wouldn't make sense for Midday/Midnight Lycanroc to show up as unlocked forms.

Opaque02:
> See https://bulbapedia.bulbagarden.net/wiki/List_of_Pok%C3%A9mon_with_form_differences for more information.
> 
> What I'll note is that there are a few Item-dependent form(e)s in there. I'm pretty sure that you can't start with Arceus in a Plate-having form without a Plate, nor either Rider Calyrex without that set of Reins. There's a few of those here, including Deoxys, Genesect, Silvally, Hoopa, Zacian, Zazamenta, Urshifu, Ogerpon if I recall correctly.
> 
> ~Pretty sure you need to Fuse Kyurem for its forms, so that one's out too.~ ~Same for Necrozma.~ Kyurem and Necrozma are also Item-based form changes, sorry about that.
> 
> Morpeko's form is Ability-based, defaults to Full Belly, and tbh shouldn't even show up in Starter Select, much less here.
> 
> Eiscue is also because of its Ability, probably shouldn't be selectable in No-Ice Face mode, though tbh if the Balance Team gave them more abilities provided you don't take Ice Face I'd be cool with that. However, at least it doesn't proc CycleForm in Starter Select.
> 
> Mimikyu's Disguise is literally the same.
> 
> Cramorant's Gulp Missile is combat dependent, based on using Surf/Dive.
> 
> Wishiwashi has the ability Schooling, which only triggers in combat.
> 
> Terpagos form changes by Terastallising, and so it does not show up in Starter Select.
> 
> Zenith Marshadow is a combat effect, so actually that also needs to be reflected in Starter Select, where currently CycleForm is visible.
> 
> Xerneas uses Neutral mode when not in combat/Contests, Active when it is, and also needs to be fixed for Starter Select.
> 
> Aegislash formes are caused by its Ability, Stance Change.
> 
> Both Darmanitan regionals can have Zen Mode.
> 
> Palafin's from is, again, Ability-dependent, and does not affect Finizen, and so does not have any Starter implications.
> 
> What _do_ the Miraidon/Koraidon forms do? Roguedex says that only the base form is available (in Lab/Ruins, respectively) so I don't know where they're coming from. They don't seem to offer mechanical advantage (no Move, Ability, Stat differences), unless you consider "literally none of the alt Forms have sprites, shinys, or animations at all" a mechanical advantage. (You shouldn't. Unless seeing more of the Field is good somehow?) Like, I know in Gen9 they matter because they're unlocked over time kind of like how HMs used to be, but here... ???
> 
> Dudunsparce (Three-Segment) is the evolution counterpart of a shiny? So I'd have to look into if that's actually a thing that can be chosen in Starter Select, or if that's calculated in the Evolution Phase to know if that's relevant here.
> 
> Same for Tandemaus -> Maushold (Family of Three/Four).
> 
> For sure though that's the case for Alcremie. Milcery's evo in PRogue is dependent on biome, and cannot be chosen at Select.
> 
> I _think_? Meowstic's form is because they've got major gender differences and that's how Ability/Learnset differences would be implemented? (Though `level-moves` seems to just use the male learnset so ???) (Hidden Ability diffs are in tho) Point is, that's handled by Espurr's CycleGender, and it doesn't get a CycleForm anyways. I could see keeping this one though because it does let people know that gender affects your Meowstic results, so this one could be an edge case. On the other hand, Ralts (Gardevoir/Gallade) doesn't, so it's whatver.
> 
> On the other other hand, Indeedee is a starter, and on Starter Select, both CycleGender and CycleForm cycle both at the same time.
> 
> Same for Lechonk -> Oinkologne (M/F) Same for Basculin (White-Striped) -> Basculegion (M/F) (note to self, Basculin's Ability_1 isn't changing as expected in Starter Select. It does appear to show up correctly in team summary though.)
> 
> Same for Toxtricity, though its based on Nature, not Gender.
> 
> This is kind of like how Lycanroc works. Though Rockruff gets the "", "Own Tempo" options, as is ideal, non-OWN_TEMPO Lycanroc also depends on Time of Day, so it wouldn't make sense for Midday/Midnight Lycanroc to show up as unlocked forms.

This is amazing, thank you so much! I'll need some time to go through this and check it all out. There's some pokemon that can be solved easily but some can't, so I'll need to do more digging and see what we can apply to make it more consistent. The main issue is trying to figure out how it all works together so we can get something that works all the time with every pokemon we see. Out of curiosity, there's a few things on your list that weren't on my original list - I'm assuming if they're not on there they're correct on my list? So for example are things like Zygarde are correct in that you can find them in the wild with the different forms? 

The tricky bit though is that it doesn't have to necessarily be whether it can be chosen in the starter select or not. For example, if I'm reading this right, Alcremie's starter doesn't have any forms to select in the starter select, but it might be possible to encounter different forms of Alcremie in the wild? If so, even though the starter select screen doesn't allow form changing, we should still show the different forms of Alcremie, which makes it tricky?

chaosgrimmon:
tbh i'm not just sure about zygarde. i thought they were collectable/selectable as a Starter? i'm aware that it's also an Ability-based transformation into 100%, but i had heard that there was some other precondition or something in a prior version of the game.
tldr, i didn't have precise actionable knowledge aside from that it's not Zygarde-Cell based like in canon. they should be Lab bosses tho
image

yeah i completely loss the plot wrt kind of feedback you were looking for.
i assumed the point was, since the only place where you can remotely see what you've collected is Starter Select, that it wouldn't be relevant if you 'collected' something that you wouldn't actually get to keep. in retrospect, i totally agree that it should be visible on the Capture Infobox in the sense that "Different forms for this Pokemon exist, and knowing which one it is may be relevant to your build or collection," but it'd suck to be excited for a new unlock and it turns out to not actually *be* an unlock.
I certainly wouldn't want it to highlight in blue about it in that case, since that ideally should be reserved for actual confirmed additions.

i'm leaving that list up because imo ideally those wouldn't trigger highlights, and expect a list in a separate message to follow more along the lines of "This won't have mid-run implications and is also not a collectible"

chaosgrimmon:
See https://bulbapedia.bulbagarden.net/wiki/List_of_Pok%C3%A9mon_with_form_differences for more information.

Item-based Evos:
 - Urshifu - Useful to know, not collectable

Item-based Forme Changes:
> Arceus, Calyrex, Deoxys, Genesect, Silvally, Hoopa, Keldeo, Zacian, Zazamenta, Ogerpon, Kyurem, Necrozma.
 - Not convinced these naturally occur wildly, not collectable

Ability-based Combat Effect:
 - Morpeko - Full Belly, not persistent, not collectable
 - Eiscue - Ice Face, not persistent, not collectable
 - Mimikyu - Disguise, not persistent, not collectable
 - Cramorant - Gulp Missile, not persistent, not collectable
 - Terpagos - Tera Shell:
 > Form data
Outside of [battle](https://bulbapedia.bulbagarden.net/wiki/Pok%C3%A9mon_battle), Terapagos is always in its Normal Form. When it enters a battle for the first time, it changes from its Normal Form into its Terastal Form through its [signature Ability](https://bulbapedia.bulbagarden.net/wiki/Signature_Ability) [Tera Shift](https://bulbapedia.bulbagarden.net/wiki/Tera_Shift_(Ability)) and will never change back to Normal Form until the battle ends, even if it [faints](https://bulbapedia.bulbagarden.net/wiki/Faint). As a result, its Normal Form cannot legitimately be used in a battle. Terastal Form Terapagos also has Tera Shift replaced by its other signature Ability, [Tera Shell](https://bulbapedia.bulbagarden.net/wiki/Tera_Shell_(Ability)). If Terapagos [Terastallizes](https://bulbapedia.bulbagarden.net/wiki/Terastal_phenomenon) during a battle, it transforms into its Stellar Form, where Tera Shell is replaced by its third signature Ability, [Teraform Zero](https://bulbapedia.bulbagarden.net/wiki/Teraform_Zero_(Ability)).

If I'm understanding correctly, it's should basically a Item-based forme change, depending on if a Tera Shard is equipped. Not sure if that's the case in practice yet though, since I've seen Normal Terapagos in the wild (overridden in, but still). Not collectable.
 - Aegislash - Stance Change, not persistent, not collectable
 - (Galarian) Darmanitan - Zen Mode, not persistent, not collectable
 - Palafin - Zero to Hero, not persistent, not collectable
 
Non-Ability-based innate combat effect:
 - Marshadow/Zenith Marshadow
 - Neutral Xerneas/Active Xerneas

I'm not convinced these actually exist but if they did they'd be collectible but probably not useful gameplaywise:
 - Miraidon
 - Koraidon

Rare Evolution:
 - Dudunsparce - Three-Segment, weight difference, not collectable
 - Maushold - Family of Three, weight difference, not collectable

(but like if you're a gamer that's planning around that you know to just look at them )

 - Alcremie - assorted Cream types, no gameplay difference, not collectable

Gender-dependent Evolution:
 - Meowstic - Ability and Learnset differences, not collectable
 - Oinkolgne - Ability/Learnset/Stat differences, not collectable
 - Basculegion - Stat differences, not collectable
 - Indeedee - Ability/Learnset/Stat differences, technically collectable but handled by new Gender marker

Nature-dependent Evolution:
 - Toxtricity - Ability and Learnset differences, technically collectable but handled by new Nature marker
 (tbh there's a world where it makes sense to add a form change marker to *Toxel*, to make starter select more straightforward)

Lycanroc:
 - Lycanroc - Yes

src/ui/pokemon-info-container.ts
```diff
@@ -12,6 +12,7 @@ import ConfirmUiHandler from "./confirm-ui-handler";
 import { StatsContainer } from "./stats-container";
 import { TextStyle, addBBCodeTextObject, addTextObject, getTextColor } from "./text";
 import { addWindow } from "./ui-theme";
+import * as pokemonSpecies from "../data/pokemon-species";
```
Tempo-anon, line 15: import * :pikastare:

Tempo-anon (commented):
Is this still wanted/relevant? I like the changes, I'm just a bit concerned about the `import *`

flx-sta:
@Opaque02 you still working on this?

flx-sta:
> I think I'll update it and put it through
> If there's a better way, we can figure it out later
> \- @Opaque02 | Discord

PR #2315, opened 2024-06-17
[Bug] Unauthorized 401 error fix

XGSleepWalker:
## What are the changes?

The server is now timing out the session cookie properly, so the client needs to handle that by kicking the user back to the login screen.

The users are currently stuck and have to manually delete their cookies in order to login back once more to regain a valid cookie.

## What did change?

Added validations around updateUserInfo to handle 401 cases.

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -7,6 +7,7 @@ export interface UserInfo {
 }
 
 export let loggedInUser: UserInfo = null;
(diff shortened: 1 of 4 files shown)


Tempo-anon (approved):
Should document what `success` and `response` are since them being arrays seems weird

PR #2320, opened 2024-06-17
[Refactor/Bug] Overhaul & Document Item/Modifier Overrides

xsn34kzx:
## What are the changes?

Refactors `overrideHeldItems`, `overrideModifiers`, and a portion of `getPlayerModifierTypeOptions` dealing with the `ITEM_REWARD_OVERRIDE` to fix an issue where overridden reward items would not persist on reload. Additionally, `ITEM_REWARD_OVERRIDE` had its type changed for more specificity with generators.

The primary goal was fixing the override portion of `getPlayerModifierTypeOptions` such that any reward items picked up during a run would continue to exist on reload. Likewise, `ITEM_REWARD_OVERRIDE` was limited in that it could not specify pre-generator arguments, meaning that generator items were subject to RNG or party weighting. While I was working on those two things, I noticed that this portion of code and the override functions dealing with items and modifiers behave very similarly if not exactly the same at some points, so I figured it would be best to have parity between all the functions. On top of this, because the documentation isn't too in-depth or JSDocs-friendly in `overrides.ts`, I decided to tackle that as well. 

## What did change?

`overrideHeldItems` and `overrideModifiers` are essentially rewritten to be a little easier to parse through, a tad more performant, and paralleling one another. A small addition is that there is now an option to forgo the use of `type` for generators that do not use `party` as an argument (like `TEMP_STAT_BOOSTER` or `BERRY`). 

For `ITEM_REWARD_OVERRIDE`, the array type was changed to `ModifierOverride`. Inconsequential changes at this stage included going from `Array<ModifierOverride>` to `ModifierOverride[]` and making `ModifierOverride` a type rather than an interface (this change was really only done because we have no intention to extend it, which is the only difference as stated [here](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces)). By using `ModifierOverride` instead of `string`, `type` can now be used to specify pre-generator arguments; likewise, `getPlayerModifierTypeOptions` was rewritten such that it matched the other override functions and called `withIdFromFunc` to set its `id` correctly, which was needed to save and reload any overridden item rewards.

Documentation for the override variables was written, and comments were adjusted to be more informative since the difference between the overrides isn't obvious at first glance for contributors, let alone testers. A lot of the work I've done with items has made more savvy to them, so I figure that it would be good to write the things I had to work through initially to save time and provide more accessibility for manual testers and people less familiar with the code base. JSDoc `linkcode` annotations were used in addition to some silent imports to allow people to quickly jump to those relevant classes.

### Screenshots/Videos

Note that most of the changes are effectively indistinguishable in-game; however, the bug related to saving `ITEM_REWARD_OVERRIDE` modifiers has a clear before and after.

#### `ITEM_REWARD_OVERRIDE` - Before, Generator-based Item

https://github.com/pagefaultgames/pokerogue/assets/109637146/a8eda429-55fe-49be-acf9-454620c698cd

#### `ITEM_REWARD_OVERRIDE` - After, Generator-based Item

https://github.com/pagefaultgames/pokerogue/assets/109637146/0e78ced9-e746-4a69-aa24-66e1b23b7092

#### `ITEM_REWARD_OVERRIDE` - Before, Held Item

https://github.com/pagefaultgames/pokerogue/assets/109637146/d9b4e1a9-29f5-434e-8daa-0c2de4fccb53

#### `ITEM_REWARD_OVERRIDE` - After, Held Item

https://github.com/pagefaultgames/pokerogue/assets/109637146/09a1c343-f437-4704-a927-434e5dfbfc07

## How to test the changes?

Testing can be done by just using any of the `MODIFIER_OVERRIDE`s, `HELD_ITEM_OVERRIDE`s, and/or `ITEM_REWARD_OVERRIDE` and making sure they apply correctly and persist on reload. It's worth mentioning that, because these are developer overrides, there is no strict checking beyond making sure the name of the entry exists in `modifierTypes`. This means that, if you provide the wrong `type` pre-generator argument, it'll crash, which matches the behavior from before this PR regardless.

## Checklist
- [x] Are the changes visual?


src/overrides.ts
```diff
@@ -99,36 +100,60 @@ export const OPP_VARIANT_OVERRIDE: Variant = 0;
...
+     * - {@linkcode Stat}, for {@linkcode modifierTypes.BASE_STAT_BOOSTER} / Vitamins
+     * - {@linkcode Nature}, for {@linkcode modifierTypes.MINT}
+     * - {@linkcode Type}, for {@linkcode modifierTypes.TERA_SHARD} or {@linkcode modifierTypes.ATTACK_TYPE_BOOSTER} / Type-boosting items
+     * - {@linkcode BerryType}, for {@linkcode modifierTypes.BERRY}
+     */
     type?: TempBattleStat|Stat|Nature|Type|BerryType
-}
-export const STARTING_MODIFIER_OVERRIDE: Array<ModifierOverride> = [];
-export const OPP_MODIFIER_OVERRIDE: Array<ModifierOverride> = [];
+};
```
OrangeRed, line 137: Can you use this piece of typescript instead to force the `ModifierOverride` Type to have a subtype when necessary. Also I would like for this typedef to be in `modifier-types.ts` since that feels like a more logical place to put it rather than in `overrides.ts`
```ts
type BaseModifierTypes = {
  name: Exclude<keyof typeof modifierTypes, GeneratorModifierTypes["name"]>;
  count: number;
};

type GeneratorModifierTypes = { count: number } & (
  | {
      name: "TEMP_STAT_BOOSTER";
      type: TempBattleStat;
    }
  | {
      name: "BASE_STAT_BOOSTER";
      type: Stat;
    }
  | {
      name: "MINT";
      type: Nature;
    }
  | {
      name: "TERA_SHARD" | "ATTACK_TYPE_BOOSTER";
      type: Type;
    }
  | {
      name: "BERRY";
      type: BerryType;
    }
);

type ModifierTypes = GeneratorModifierTypes | BaseModifierTypes;
```

src/overrides.ts
```diff
@@ -99,36 +100,60 @@ export const OPP_VARIANT_OVERRIDE: Variant = 0;
...
+ * member to hold it (typically is, extends, or generates {@linkcode ModifierType}s),
+ * like `EXP_SHARE`, `CANDY_JAR`, etc.
+ *
+ * Note that, if count is not provided, it will default to 1. Additionally, note that some
+ * held items and modifiers are grouped together via a {@linkcode ModifierTypeGenerator} and
+ * require pre-generation arguments to get a specific item.
+ *
+ * @example STARTING_MODIFIER_OVERRIDE = [{name: "EXP_SHARE", count: 2}] // will have a quantity of 2 in-game
+ * @example STARTING_HELD_ITEM_OVERRIDE = [{name: "LUCKY_EGG"}] // will have a quantity of 1 in-game
+ * @example {name: "BERRY", count: 5, type: BerryType.SITRUS} // type must be given to get a specific berry
```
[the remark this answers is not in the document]
OrangeRed: I wrote the typescript for the first point in a later comment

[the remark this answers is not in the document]
xsn34kzx: I gotchu, those single-line comments were largely legacy comments, so I didn't want to necessarily shelve them without second opinion.

xsn34kzx:
Type changes were introduced as suggested by @OrangeRed with some adjustment to allow for generator output to be random or specified.

### Example
With the following overrides:
- `STARTING_MODIFIER_OVERRIDE = [{ name: "EXP_SHARE", count: 3}, { name: "DYNAMAX_BAND" }];`
- `ITEM_REWARD_OVERRIDE = [{ name: "TEMP_STAT_BOOSTER" }, { name: "TEMP_STAT_BOOSTER", type: TempBattleStat.ATK }];`

generatorRoll1

generatorRoll2

As a result, the first `TEMP_STAT_BOOSTER` is random per roll while the second `TEMP_STAT_BOOSTER` is always X Attack.

Additionally, `EvolutionItem` and `FormChangeItem` were added to the types that can be used and specified.

### Example
With the following overrides:
- `STARTING_MODIFIER_OVERRIDE = [{ name: "MEGA_BRACELET" } ];`
- `ITEM_REWARD_OVERRIDE = [{ name: "EVOLUTION_ITEM", type: EvolutionItem.CRACKED_POT }, { name: "FORM_CHANGE_ITEM" }];`

evolutionItem

evolutionFormItems

The first slot is always `CRACKED_POT`, but without Audino, no `FORM_CHANGE_ITEM` replaces the second roll slot. However, with an Audino and a Mega Bracelet, it appears.

src/overrides.ts
```diff
@@ -1,17 +1,21 @@
...
 import { Variant } from "./data/variant";
+/* eslint-disable @typescript-eslint/no-unused-vars */
 import { TempBattleStat } from "./data/temp-battle-stat";
 import { Nature } from "./data/nature";
 import { Type } from "./data/type";
 import { Stat } from "./data/pokemon-stat";
+import { EvolutionItem } from "./data/pokemon-evolutions";
+import { FormChangeItem } from "./data/pokemon-forms";
+import { BerryType } from "#enums/berry-type";
+/* eslint-enable @typescript-eslint/no-unused-vars */
```
OrangeRed, line 11: instead of disabling eslint why not remove the things that aren't being used anymore?

The only reason allSpecies has the eslint disable is because its being used in a comment but not in javascript.

src/modifier/modifier-type.ts
```diff
@@ -1914,3 +1916,70 @@ export function getLuckTextTint(luckValue: integer): integer {
...
+ * held items and modifiers are grouped together via a {@linkcode ModifierTypeGenerator} and
+ * require pre-generation arguments to get a specific item from that group. If a type is
+ * not set, the generator will either use the party to weight item choice or randomly
+ * pick an item.
+ *
+ * @example STARTING_MODIFIER_OVERRIDE = [{name: "EXP_SHARE", count: 2}] // will have a quantity of 2 in-game
+ * @example STARTING_HELD_ITEM_OVERRIDE = [{name: "LUCKY_EGG"}] // will have a quantity of 1 in-game
+ * @example {name: "BERRY", type: BerryType.SITRUS} // type must be given to get a specific berry
+ * @example {name: "BERRY"} // a random berry will be generated at runtime
+ */
```
OrangeRed, line 1984: instead of 4 @example statements can you just make one @example that has all four examples inside back ticks.

OrangeRed (approved):

src/modifier/modifier-type.ts
```diff
@@ -1903,3 +1916,72 @@ export function getLuckTextTint(luckValue: integer): integer {
...
+      name: "EVOLUTION_ITEM" | "RARE_EVOLUTION_ITEM";
+      type?: EvolutionItem;
+    }
+  | {
+      name: "FORM_CHANGE_ITEM";
+      type?: FormChangeItem;
+    }
+  | {
+      name: "TM_COMMON" | "TM_GREAT" | "TM_ULTRA";
+    }
```
OrangeRed, line 1961: Does this need to be added here if you're not giving a type property?

src/modifier/modifier-type.ts
```diff
@@ -1903,3 +1916,72 @@ export function getLuckTextTint(luckValue: integer): integer {
...
+
+type BaseModifierOverride = {
+  /** Key for any given modifier, held item, or generator in {@linkcode modifierTypes} */
+  name: Exclude<keyof typeof modifierTypes, GeneratorModifierOverride["name"]>;
+  /** Quantity of the held item or modifier desired */
+  count?: number;
+};
+
+type GeneratorModifierOverride = {
+    /** Quantity of the held item or modifier desired */
```
OrangeRed, line 1928: This comment isn't needed. The combined type will inherit the jsdoc from the Base type

src/modifier/modifier-type.ts
xsn34kzx: Added them to group them together in the "item generated from a generator" category and because I'd also want to give a type to TMs to eventually specify them in an item roll. Wasn't sure if that was out of scope or not for this PR, but that was the main intention.

OrangeRed: Just add them when it becomes relevant not now preemptively

xsn34kzx: Gotcha, in that case I'll implement that too to round out the overhaul.

xsn34kzx:
In addition to form-changing items and evolution items, TMs can now be specified and/or randomly generated, which covers the last of the generator-based items (at least until #1899).

### Example
With the following override:
```ts
export const ITEM_REWARD_OVERRIDE: ModifierOverride[] = [ { name: "TM_ULTRA" }, { name: "TM_COMMON", type: Moves.FLAMETHROWER } ];
```

generatorRoll3

generatorRoll4

Similar to the previous items, not specifying the move of the TM leaves it up to the generator to decide the TM that is eventually rolled, which uses RNG and compatibility with the party (which, in these examples, give Flare Blitz and Fire Blast). If the move is specified, the desired TM will always roll. 

Also worth noting that any move can be used, even if it isn't an existing TM.

### Example

With the following override:
```ts
export const ITEM_REWARD_OVERRIDE: ModifierOverride[] = [ { name: "TM_ULTRA" }, { name: "TM_COMMON", type: Moves.GIGATON_HAMMER } ];
```

customTm

OrangeRed:
What happens when you try to create a TM_ULTRA with a move that is normally TM_COMMON?

xsn34kzx:
> What happens when you try to create a TM_ULTRA with a move that is normally TM_COMMON?

@OrangeRed Because `TM_COMMON`, `TM_GREAT`, and `TM_ULTRA` use the same generator class, they should all get caught in the `pregenArgs` condition before tier matters, but I did a double check with a few differently tiered moves just to make sure.

Using Bulk Up (Common), Brine (Great), and Close Combat (Ultra) with `TM_ULTRA` name,
```ts
export const ITEM_REWARD_OVERRIDE: ModifierOverride[] = [ { name: "TM_ULTRA", type: Moves.BULK_UP }, { name: "TM_ULTRA", type: Moves.BRINE }, { name: "TM_ULTRA", type: Moves.CLOSE_COMBAT } ];
```
rollTmUltra

Using the same set of moves with `TM_GREAT` name,
rollTmGreat

Using the same set of moves with `TM_COMMON` name,
rollTmCommon

From these examples, I pretty much double checked that the following scenarios work and don't error on roll:
- A Common tier TM is used under Ultra/Great
- A Great tier TM is used under Common/Ultra
- An Ultra tier TM is used under Common/Great

OrangeRed (approved):

src/overrides.ts
```diff
@@ -1,19 +1,14 @@
...
-import { TempBattleStat } from "./data/temp-battle-stat";
-import { Nature } from "./data/nature";
-import { Type } from "./data/type";
-import { Stat } from "./data/pokemon-stat";
 import { PokeballCounts } from "./battle-scene";
 import { PokeballType } from "./data/pokeball";
 import { Gender } from "./data/gender";
 import { StatusEffect } from "./data/status-effect";
-import { modifierTypes } from "./modifier/modifier-type";
+import { ModifierOverride } from "./modifier/modifier-type";
```
OrangeRed, line 7: Can you add `type` here so its clear. `import { type ... } from "..."`

OrangeRed (approved):

xsn34kzx:
> In addition to form-changing items and evolution items, TMs can now be specified and/or randomly generated, which covers the last of the generator-based items (at least until #1899).

Now that #1899 has been merged, `SPECIES_STAT_BOOSTER` is now included among the generator-based items and can also now be specified.

src/modifier/modifier-type.ts
```diff
@@ -2041,3 +2055,76 @@ export function getLuckTextTint(luckValue: integer): integer {
   const modifierTier = luckValue ? luckValue > 2 ? luckValue > 5 ? luckValue > 9 ? luckValue > 11 ? ModifierTier.LUXURY : ModifierTier.MASTER : ModifierTier.ROGUE : ModifierTier.ULTRA : ModifierTier.GREAT : ModifierTier.COMMON;
```
flx-sta, line 2055: ik, ik. Not your PR but could you clean those up in your PR?

src/modifier/modifier-type.ts
```diff
@@ -2041,3 +2055,76 @@ export function getLuckTextTint(luckValue: integer): integer {
...
+ * pick an item.
+ *
+ * @example ```ts
+ * STARTING_MODIFIER_OVERRIDE = [{name: "EXP_SHARE", count: 2}] // will have a quantity of 2 in-game
+ * STARTING_HELD_ITEM_OVERRIDE = [{name: "LUCKY_EGG"}] // will have a quantity of 1 in-game
+ * {name: "BERRY", type: BerryType.SITRUS} // type must be given to get a specific berry
+ * {name: "BERRY"} // a random berry will be generated at runtime
+ * ```
+ */
+export type ModifierOverride = GeneratorModifierOverride | BaseModifierOverride;
```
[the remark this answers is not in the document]
flx-sta: @OrangeRed what do you think?

OrangeRed (changes requested):

src/modifier/modifier-type.ts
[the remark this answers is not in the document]
OrangeRed: @flx-sta I reviewed this again today and I think this needs to be separated somehow because as it is, its very possible to pass a held item as a party modifier and crash the game

src/modifier/modifier-type.ts
```diff
@@ -1878,13 +1879,25 @@ export function getPlayerModifierTypeOptions(count: integer, party: PlayerPokemo
...
+      let modifierType = modifierFunc();
+
+      if (modifierType instanceof ModifierTypeGenerator) {
+        modifierType = modifierType.generateType(party, ("type" in override) && (override.type !== null) ? [override.type] : null);
+      }
+
+      if (modifierType) {
+        options[i].type = modifierType.withIdFromFunc(modifierFunc);
+      }
+    }
```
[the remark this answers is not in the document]
xsn34kzx: The only issue with how its done in that PR is that the game crashes when there's more entries in `ITEM_REWARD_OVERRIDE`, which I tested really quick before merging:

```ts
const overrides = {
  ITEM_REWARD_OVERRIDE: [ "AMULET_COIN", "AMULET_COIN", "AMULET_COIN", "AMULET_COIN", "AMULET_COIN", "AMULET_COIN"]
```

rewardError

Sticking with this implementation since `minLength` covers this and the other scenario where there's less entries in `ITEM_REWARD_OVERRIDE`.

src/modifier/modifier.ts
```diff
@@ -2614,69 +2614,71 @@ export class EnemyFusionChanceModifier extends EnemyPersistentModifier {
...
-    // we create the item
-    itemModifier.pokemonId = pokemon.id; // we assign the created item to the pokemon
-    itemModifier.stackCount = qty; // we say how many items we want
-    if (player) {
-      scene.addModifier(itemModifier, true, false, false, true);
-    } else {
-      scene.addEnemyModifier(itemModifier, true, true);
+    if (item.name in modifierTypes) {
+      const modifierFunc = modifierTypes[item.name];
+      let modifierType = modifierFunc();
```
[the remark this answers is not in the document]
xsn34kzx: `modifierFunc` is used again a little later in the function when calling `withIdFromFunc`.

src/modifier/modifier-type.ts
```diff
@@ -968,7 +968,10 @@ class SpeciesStatBoosterModifierTypeGenerator extends ModifierTypeGenerator {
 
 class TmModifierTypeGenerator extends ModifierTypeGenerator {
   constructor(tier: ModifierTier) {
-    super((party: Pokemon[]) => {
+    super((party: Pokemon[], pregenArgs?: any[]) => {
+      if (pregenArgs) {
+        return new TmModifierType(pregenArgs[0] as Moves);
+      }
```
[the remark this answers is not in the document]
xsn34kzx: Ended up going one by one to every instance of this in the generator classes and incorporating type checking. The only caveat is that `instanceof` specifically can't be used for enums, where most if not all of the `pregenArgs` themselves are enums, so I did `pregenArgs[0] in ...` instead for the most part. Another random thing to note is that pretty much all of the `pregenArgs` should just have one element of that type, probably a refactor waiting to happen that may or may not be annoying/break how generator items are regenerated during save loading.

xsn34kzx:
Entries in `ITEM_REWARD_OVERRIDE` now correctly reflect their tier using a new function provided by @ben-lear 🎉 

### Example

With the following override:
```ts
ITEM_REWARD_OVERRIDE: [ { name: "FORM_CHANGE_ITEM", type: FormChangeItem.RED_ORB }, { name: "SCOPE_LENS" }, { name: "LEEK" } ]
```

tieredRewards

flx-sta (approved):
lgtm 🚀 
and works like a charm (play-test)a

DayKev (approved):

PR #2328, opened 2024-06-17
[QoL] Added setting for alternative canvas rendering resolutions

7koi:
## What are the changes?
I added an entry to the display settings that allows for alternative canvas rendering resolutions.

I wanted to be able to play PokeRogue on my Raspberry Pi 4. The hard coded canvas rendering resolution is 1920x1080, which does not run at a smooth framerate on the Raspberry Pi 4. My changes are different from #2055 because my changes change the actual canvas rendering resolution, and #2055 just changes the scale of the canvas so it does not have to take up the entire browser window.

## What did change?
In addition to the display settings entry, I also had to introduce an additional variable in the base scene class that stored the resolution scaling factor. Since this scaling factor also has to be available when the canvas is initialized using the Phaser game config, I also (regrettably) had to make an entry in the local storage for the scaling factor. Since many of the assets were originally designed for a 320x180 display, many places in the UI code had a hard coded value of 6 for a scaling factor. These hard coded scaling factors were replaced by the resolution scaling factor stored in the scene object.

### Screenshots/Videos
If I did everything correctly, the user experience should be identical regardless of the chosen resolution. Some text ends up being pixelated if the lower resolutions are chosen.

## How to test the changes?
I tested the changes by playing the game at 640x360 on my Raspberry Pi 4. The framerate is much improved, and if I saw any UI element out of place, I corrected it and re-tested it. A reviewer can test the changes on any device by changing the resolution in the settings, reloading the game, and navigating through all the menus.

## Checklist
- [ ] Are the changes visual?


flx-sta:
@7koi   closed for being stale (no updates for over a month)

----

PR #2331, opened 2024-06-17
[Refactor] Remove inconsistent partial() usage in healing abilities

3ae3ae:
## Why am I doing these changes?

#2330 

## What did change?
Removed partial() usage in healing abilities

## Checklist
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [ ] Are the changes visual?
  - [ ] Have I provided screenshots/videos of the changes?

Issue #2330, opened 2024-06-17
[Docs] Inconsistencies in healing moves and abilities


3ae3ae:
**Is your feature request related to a problem? Please describe.**

Healing-related abilities like Ice Body and Hospitality are annotated with "// Healing not blocked by Heal Block" and are implemented with partial().
However, healing moves like Megadrain are not implemented with partial(), even though they don't interact with Heal Block.
This creates an inconsistency in the code.

**Describe the Feature**

I suggest applying partial() consistently to healing moves and abilities.

**Additional context**

Since Heal Block, which is not implemented at all yet, is a move that consistently blocks all healing, I think it would be better to remove partial() from abilities.

Bertie690:
Somehow between last year and now the partial notices got removed, so Imma just close this...

PR #2332, opened 2024-06-17
[Move] Add 'partial' to heal-related moves.

3ae3ae:
#2330 
#2331 

## What did change?
Add 'partial' to heal-related moves due to incomplete interaction with heal block

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -5683,9 +5683,11 @@ export function initMoves() {
     new AttackMove(Moves.STRENGTH, Type.NORMAL, MoveCategory.PHYSICAL, 80, 100, 15, -1, 0, 1),
     new AttackMove(Moves.ABSORB, Type.GRASS, MoveCategory.SPECIAL, 20, 100, 25, -1, 0, 1)
       .attr(HitHealAttr)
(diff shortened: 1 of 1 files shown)


RNGezeus:
> I'm not sure if this is needed. Since heal block is still (N), it should be up to the implementation of heal block that takes care of all of this rather than slapping a bunch of (P)s on moves that function fine

I fully agree that really should hang on Heal block, but currently notation can feel inconsistent since moves list as complete but abilites use (P) despite it being healblocks fault.

OrangeRed:
> > I'm not sure if this is needed. Since heal block is still (N), it should be up to the implementation of heal block that takes care of all of this rather than slapping a bunch of (P)s on moves that function fine
> 
> I fully agree that really should hang on Heal block, but currently notation can feel inconsistent since moves list as complete but abilites use (P) despite it being healblocks fault.

@Tempo-anon 
Maybe instead you can remove the (P) tag from moves that currently only have it because of the move.

Tempo-anon:
Closing due to merge conflicts. I am of the belief we should remove (P) from all the healing moves/abilities.

PR #2333, opened 2024-06-17
[Move] Implement Natural Gift

AyushBarik:
## What are the changes?

User will be able to utilize the move Natural Gift.

Enhances User Experience.

## What did change?

Natural Gift is now a usable move.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/128145955/2d87acf0-4f3b-4496-8079-24de921c842d

## How to test the changes?

Made use of overrides to test the move. One can test the changes by testing the move Natural Gift. 

The move has been tested in English and Spanish with a variety of different berries and it functions as intended. Tsdoc has also been made. Much of the implementation is similar to fling.

## Checklist
- [yes ] There is no overlap with another PR?
- [ yes] The PR is self-contained and cannot be split into smaller PRs?
- [ yes] Have I provided a clear explanation of the changes?
- [ yes] Have I tested the changes (manually)?
    - [yes ] Are all unit tests still passing? (`npm run test`)
- [ yes] Are the changes visual?
  - [ yes] Have I provided screenshots/videos of the changes?


Tempo-anon:
Should be pretty similar to Fling implementation. I assume it will just choose a random berry if the user has more than 1 applicable berry.

src/data/move.ts
```diff
@@ -2709,6 +2709,83 @@ export class MovePowerMultiplierAttr extends VariablePowerAttr {
...
+ * This class handles the logic for the Natural Gift move in the game.
+ */
+export class NaturalGiftAttr extends VariablePowerAttr {
+  protected randomItem: string;
+
+  /**
+   * Determines if the move can be used based on the user's held berries.
+   * @returns A function that checks the condition for using the move.
+   */
+  getCondition(): MoveConditionFunc {
```
Tempo-anon, line 2723: Doesn't this just return a boolean of whether or not the user has berries?

src/data/move.ts
```diff
@@ -2709,6 +2709,83 @@ export class MovePowerMultiplierAttr extends VariablePowerAttr {
...
+      ) as BerryModifier[];
+      return userBerries.length > 0;
+    };
+  }
+
+  /**
+   * Retrieves the failure text for the Natural Gift move, indicating why it failed.
+   * @returns The failure message or null if the move does not fail.
+   */
+  getFailedText(user: Pokemon, target: Pokemon, move: Move, cancelled: Utils.BooleanHolder): string | null {
```
Tempo-anon, line 2737: This is also checking for `cancelled` value too right? Is returning `null` the correct response for failure text when the user has berries?

src/data/move.ts
```diff
@@ -2709,6 +2709,83 @@ export class MovePowerMultiplierAttr extends VariablePowerAttr {
   }
 }
 
+/**
+ * Represents the Natural Gift attribute which extends VariablePowerAttr.
+ * This class handles the logic for the Natural Gift move in the game.
+ */
+export class NaturalGiftAttr extends VariablePowerAttr {
```
Tempo-anon, line 2716: Shouldn't this also change the type of the attack based on the berry too?

Tempo-anon (changes requested):

src/data/move.ts
AyushBarik: yes. there was an issue that "return false" wasn't working the way it was supposed to in fling so i was suggested to use this method by orangeRed

src/data/move.ts
AyushBarik: what do you mean by cancelled value?

src/data/move.ts
AyushBarik: ah it does. let me fix that

src/data/move.ts
Tempo-anon: Oh I see on 2768 and 2783 you're passing true in. Should update the function tsdoc to reflect all the variables

src/data/move.ts
AyushBarik: in order to pass information from one class to another i set up a global variable. is tht ok? maybe i can make setter and getter functions so its safer

src/modifier/modifier.ts
```diff
@@ -1200,6 +1200,34 @@ export class LevelIncrementBoosterModifier extends PersistentModifier {
...
+  [BerryType.APICOT]: 100,
+  [BerryType.SALAC]: 100,
+  [BerryType.LANSAT]: 100,
+  [BerryType.STARF]: 100,
+  [BerryType.LEPPA]: 80,
+};
+
+const berryNaturalGiftTypeMap: { [key in BerryType]: number } = {
+  [BerryType.SITRUS]: 6,  // Type.PSYCHIC
+  [BerryType.LUM]: 1,     // Type.FLYING
```
Tempo-anon, line 1219: Isn't Type.FLYING actually 2? https://github.com/pagefaultgames/pokerogue/blob/42304070a0e59193ebcc4b66c01b5e0e25d209c6/src/data/type.ts#L5

There's no need to use actual numbers here when the enums map to numbers. It'll also prevent errors like this where all your numbers seem to be off by 1

Tempo-anon (changes requested):

src/modifier/modifier.ts
AyushBarik: You are right. I removed the array and stopped passing in an index for the array and directly pass in the move type instead.

src/battle-scene.ts
```diff
@@ -228,6 +228,8 @@ export default class BattleScene extends SceneBase {
   public uiContainer: Phaser.GameObjects.Container;
   public ui: UI;
 
+  public naturalGiftBerryModifier: BerryModifier;
```
innerthunder, line 231: Can this be moved to a class closer in scope to Nature Power's attributes? `Pokemon` or even `TurnData` should be able to support this.

innerthunder (changes requested):

src/battle-scene.ts
AyushBarik: Both of those classes are in pokemon.ts which does not import BerryModifier. Should I import it in? Or stick with battle-scene.ts?

AyushBarik: moved it to TurnData

src/data/move.ts
```diff
@@ -3661,6 +3661,100 @@ export class HiddenPowerTypeAttr extends VariableMoveTypeAttr {
...
+  /**
+   * Applies the Natural Gift power to the move.
+   * @param user - The Pokémon using the move.
+   * @param target - The target Pokémon.
+   * @param move - The move being used.
+   * @param args - Additional arguments for the move. The first element is expected to be a Utils.NumberHolder which holds the power of the move.
+   * @returns A boolean indicating whether the power was successfully applied.
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    const power = args[0] as Utils.NumberHolder;
```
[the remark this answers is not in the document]
AyushBarik: ok

src/data/move.ts
```diff
@@ -3641,6 +3641,100 @@ export class HiddenPowerTypeAttr extends VariableMoveTypeAttr {
...
+    return null;
+  }
+
+  /**
+   * Applies the Natural Gift move logic to the given user and target Pokémon.
+   * Determines the berry to be used, sets the move type, and removes the berry.
+   * @param user - The Pokémon using the move.
+   * @param target - The target Pokémon.
+   * @param move - The move being used.
+   * @param args - Additional arguments for the move.
```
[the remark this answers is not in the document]
AyushBarik: ok

src/data/move.ts
```diff
@@ -3641,6 +3641,100 @@ export class HiddenPowerTypeAttr extends VariableMoveTypeAttr {
...
+
+  /**
+   * Retrieves the failure text for the Natural Gift move, indicating why it failed.
+   * @param user - The Pokémon using the move.
+   * @param target - The target Pokémon.
+   * @param move - The move being used.
+   * @param cancelled - A boolean holder indicating if the move was cancelled.
+   * @returns The failure message or null if the move does not fail.
+   */
+  getFailedText(user: Pokemon, target: Pokemon, move: Move, cancelled: Utils.BooleanHolder): string | null {
```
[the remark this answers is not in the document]
AyushBarik: ok

src/field/pokemon.ts
```diff
@@ -46,7 +46,7 @@ import { Abilities } from "#enums/abilities";
 import { ArenaTagType } from "#enums/arena-tag-type";
 import { BattleSpec } from "#enums/battle-spec";
 import { BattlerTagType } from "#enums/battler-tag-type";
-import { BerryType } from "#enums/berry-type";
+import { yyType } from "#enums/berry-type";
```
Tempo-anon, line 49: Is this a typo?

AyushBarik: took care of that. made a new PR since this one got infiltrated with a bunch of commits from other people.

flx-sta:
closing in favor of #3247

Issue #2283, opened 2024-06-16
[BUG] Can move starter selection while managing moves for a selected Pokémon


travbp:
**Description**

Directional inputs right after

1. clicking Manage Moves
2. clicking a specific move to swap
3. Or, picking a move to swap in

will move the selected starter. Essentially, providing a directional input while clicking through the Manage Moves menu or any submenu of it will change your selected starter.

**To Reproduce**

1. Select a starter
2. Click on manage moves and immediately provide a directional input

**Expected behavior**
Starter selection movement locked down while in the menus

**Device**
Steam Deck offline with Xbox controller

travbp:
Duplicate of https://github.com/pagefaultgames/pokerogue/issues/1879

Issue #1879, opened 2024-06-06
[BUG] Selecting other pokémon is possible while Manage Moves


hibiyasleep:
**Describe the bug**

**To Reproduce**
1. New Game
2. Focus at any pokémon
3. Manage Moves
4. Select any move
5. then instantly press any arrow key, BEFORE the second list appear
6. 
    1. if current cursor is on uncaught, then freeze
    2. else, showing other pokémon's moves.

by the way, this does not allow selection of impossible moves.

**Expected behavior**

Arrow key should be ignored between selections.

**Screenshots / Videos**

<img width="1194" alt="image" src="https://github.com/pagefaultgames/pokerogue/assets/5380174/ca750c7f-6629-4cec-bb33-bcf19143a950">

Frozen screenshot as 6. i.

<img width="1194" alt="image" src="https://github.com/pagefaultgames/pokerogue/assets/5380174/3a9ad6f9-9386-4ab4-b5a9-b7f881198ede">

Showing other pokémon's moves as 6. ii.

**Device**
macOS Chrome (PWA)

**Additional context**

flx-sta:
Starter select ui is being redone with filters anyway rn.

PR #2342, opened 2024-06-17
[Bug] block input while waiting for transition to prevent changing starter during move select

scottarrr:
## What are the changes?

This uses the existing blockInput flag to prevent inputs from changing the starter when opening the manage moves menu.

addresses these copies of the same bug
https://github.com/pagefaultgames/pokerogue/issues/1879
https://github.com/pagefaultgames/pokerogue/issues/2283

## What did change?

Previously inputs that happened in between the triggering of swap to options mode and before it completed would be handled by the starter select, changing the starter selection. 

### Screenshots/Videos

a video wouldn't really prove that i'm hitting the buttons as described in the bug ;/

## How to test the changes?

Pull this down, open starter select, pick bulbasaur and very quickly input a right movement action.
Previously bulbasaur's moveset would open with charmander selected
With this branch bulbasaur's moves are shown and bulbasaur remains the selection

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -1149,6 +1149,9 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
           ];
           if (this.speciesStarterMoves.length > 1) {
             const showSwapOptions = (moveset: StarterMoveset) => {
+
+              this.blockInput = true;
+
               ui.setMode(Mode.STARTER_SELECT).then(() => {
                 ui.showText(i18next.t("starterSelectUiHandler:selectMoveSwapOut"), null, () => {
                   this.moveInfoOverlay.show(allMoves[moveset[0]]);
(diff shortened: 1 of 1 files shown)


src/ui/starter-select-ui-handler.ts
```diff
@@ -1149,6 +1149,9 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
           ];
           if (this.speciesStarterMoves.length > 1) {
             const showSwapOptions = (moveset: StarterMoveset) => {
+
+              this.blockInput = true;
```
brain-frog, line 1153: can you add this to the manage nature menu handler as well so the bug doesnt happen there too

brain-frog (dismissed):

src/ui/starter-select-ui-handler.ts
scottarrr: Absolutely. Good call!

scottarrr:
@brain-frog @f-fsantos Sorry for the ping, I see the base branch was swapped to beta but I'm not all that familiar with the process here. Is there anything waiting on me?

scottarrr:
@Tempo-anon  Is this still needed? I see the bugs were closed so lmk if I should close this out too.

PR #2343, opened 2024-06-17
[Move] Partially implement heal block.

3ae3ae:
## What are the changes?

I have partially implemented the Heal block.

It was not implemented.

## What did change?

I added a Heal block to the Battler tag, so that if the user uses a heal move on themselves, it will fail, otherwise the heal amount will be 0.
I added a getCondition() method to Heal Attr and its child classes so that the move fails if the healing is sealed.
In addition, I modified the apply() method to either return false or adjust the amount of healing to zero.

In the original game, if a Pokémon is under a healing seal, it can't select a healing move. However, we didn't have a rule in the project to block moves selection yet, so I couldn't apply it.

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -1220,8 +1220,15 @@ export class HealAttr extends MoveEffectAttr {
   }
 
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    if (target.getTag(BattlerTagType.HEAL_BLOCK)) {
+      return false;
+    }
     this.addHealPhase(this.selfTarget ? user : target, this.healRatio);
     return true;
+  }
+
+  getCondition(): MoveConditionFunc {
+    return (user, target, move) => user !== target || !target.getTag(BattlerTagType.HEAL_BLOCK);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


Tempo-anon:
Does it work to stop healing for all the following?

- [ ] Moves that heal (also cancels the move)
- [ ] Heal pulse/wish/lunardance/healingwish/floralhealing/pollen puff healing
- [ ] present/ingrain/aquaring/grassyterrain healing
- [ ] Leech seed still works but provides no healing (but does deal damage if the seeded Pokemon has liquid ooze)
- [ ] Stops leftovers/shell bell healing
- [ ] Stops sitrus berries being eaten
- [ ] If a sitrus is eaten with bug bite/pluck it doesn't heal
- [ ] Should stop reviver seed(?)
- [ ] Stops type immune abilities like dry skin or water absorb from healing (though it still does no damage)
- [ ] Stops cheek pouch from healing
- [ ] Regenerator still works
- [ ] ~~Z moves that heal ignore heal block~~ Z-moves currently not planned at this time

zacharied:
Overlaps with #1153

3ae3ae:
Sorry, some conditions were not met.

PR #2351, opened 2024-06-17
[Balance] Gym Leaders/E4 Terastalize off-type mons

Fontbane:
## What are the changes?

Gym Leaders and E4 members will Terastalize any mons that don't fit their specialty type. For example, many of the Paldea leader's aces, or Lorelei's Slowbro.

I'm going to redo gym leader teams and this will allow Paldea Gym Leaders to have teams reflecting their canon teams.
image
As you can see this proposal was met with no objections by the balance team.

## What did change?

Modified the getRandomTeraModifiers to set teras on off type mons. If they're not meant to have that many teras... they'll get them anyway.

### Screenshots/Videos

I gave Katy a Fidough:
image

## How to test the changes?

To do a quick test to show it works:
- Change a gym leader's first signature species (in src/data/trainer-config.ts) to an off-type mon
- Set the level 5 battle to be that gym leader
For a more thorough test:
- Change about a quarter of gym leader/e4 in src/data/trainer-config.ts to include off-type mons in their signature species
- Do a couple runs of classic like this

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/data/trainer-config.ts
+++ b/src/data/trainer-config.ts
@@ -610,7 +610,7 @@ export class TrainerConfig {
     this.setVictoryBgm("victory_gym");
     this.setGenModifiersFunc(party => {
       const waveIndex = party[0].scene.currentBattle.waveIndex;
-      return getRandomTeraModifiers(party, waveIndex >= 100 ? 1 : 0, specialtyTypes.length ? specialtyTypes : null);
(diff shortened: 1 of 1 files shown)


Tempo-anon:
Closing since it's still in draft and I'm not personally convinced this is the best implementation. I think it can be thematic and neat if a Gym Leader or Elite 4 has an off type Pokemon so you can't just sweep through with a single type.

PR #2358, opened 2024-06-18
[Move] Implement Happy Hour

Madmadness65:
## What are the changes?

Implements the move [Happy Hour](https://bulbapedia.bulbagarden.net/wiki/Happy_Hour_(move)). Since it's normally an event-only move, nothing gets it naturally currently, even as an egg move.

The move was not yet implemented.

## What did change?

Like how it functions in canon, Happy Hour doubles the prize money from trainers and from money moves like Pay Day and Make it Rain, but does _not_ affect money items like Nuggets. It also expires when the Pokémon are recalled like all other arena tags.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/59298170/eb96877b-2ae1-4d10-9c92-4fe55f4d3f9a

https://github.com/pagefaultgames/pokerogue/assets/59298170/e22bd645-bcfe-401e-b5bd-c0f73dbc6afd

## How to test the changes?

Set the move override to have the move Happy Hour and any other attacking move. Complete a battle with a trainer or having used a money attack with and without Happy Hour in effect, and observe the difference in money earned.

## Checklist
- [x] Are the changes visual?


SpookyJelly:
I saw this in passing, and it's a great idea.  lgtm too

src/data/arena-tag.ts
```diff
@@ -679,6 +679,20 @@ class TailwindTag extends ArenaTag {
   }
 }
 
+class HappyHourTag extends ArenaTag {
```
Tempo-anon, line 682: New classes should have tsdoc

src/data/arena-tag.ts
```diff
@@ -679,6 +679,20 @@ class TailwindTag extends ArenaTag {
...
+  constructor(turnCount: integer, sourceId: integer, side: ArenaTagSide) {
+    super(ArenaTagType.HAPPY_HOUR, turnCount, Moves.HAPPY_HOUR, sourceId, side);
+  }
+
+  onAdd(arena: Arena): void {
+    arena.scene.queueMessage("Everyone is caught up in the happy atmosphere!");
+  }
+
+  onRemove(arena: Arena): void {
+    arena.scene.queueMessage("The atmosphere returned to normal.");
```
Tempo-anon, line 692: oh we don't have any arena localizations lol

Tempo-anon (approved):

src/data/arena-tag.ts
Madmadness65: I can do it in a follow-up PR, as it appears that more than this class is still missing documentation in the arena tags file, and it'd be best to do them all in one fell swoop.

Tempo-anon: All right, but I'll hold you to your word for documenting arena tags then

PR #1933, opened 2024-06-08
[enhacement] Show Egg tier icon on starter UI

ReneGV:
## What are the changes?
Update the starter select UI so, instead of displaying the default egg icon, display the species egg tier.

### Before
Zapdos show default egg
image

### After
Zapdos show ultra tier egg
image

## Why am I doing these changes?
Helping new players to identify how egg tiers work.

## What did change?
- Added a new function to calculate the egg tier based on species.
- Modified starter selection screen
### Screenshots/Videos

UI video

https://github.com/pagefaultgames/pokerogue/assets/31145813/84a255c1-db76-4d30-8cdc-1f6acd24e16f

### COMMON GREAT ULTRA MASTER & MANAPHY
image

### Pokemon with prevolutions & eternatus
They dont display a egg icon
image

### Species with cost reduction
Staryu costs 4 but cost was reduced to 3 with candy. Display EggTier.GREAT, because the species can only be hatched from these eggs.
image

### Uncaught and unseen species

image

## How to test the changes?

Load your session and oper starter selection UI

## Checklist
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #2359, opened 2024-06-18
[enhancement] Move up form text when candies and hatched eggs are hidden

ReneGV:
## What are the changes?

Follow up of  one of my previous PRs
- https://github.com/pagefaultgames/pokerogue/pull/1933

When a Pokemon does not need to display the  candy count the form text remains on the same place, creating a big empty space between the pokeball icon and the text.

This PR just moves the text up to avoid empty space

I forgot this UI detail on my last PR

## What did change?

Starter select UI form text location for pikachu (no other species has forms and cross gen prevolutions)

## **BEFORE / AFTER**
image

**Starters for whose we display eggs and candy are unaffected.**

image

## How to test the changes?
- Open starter select UI and change Pikachu form

### DIFF
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -1975,6 +1975,7 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
             this.pokemonHatchedIcon,
             this.pokemonHatchedCountText
           ].map(c => c.setVisible(false));
+          this.pokemonFormText.setY(25);
         } else if (species.speciesId === Species.ETERNATUS) {
           this.pokemonHatchedIcon.setVisible(false);
           this.pokemonHatchedCountText.setVisible(false);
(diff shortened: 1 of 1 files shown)


src/ui/starter-select-ui-handler.ts
```diff
@@ -1975,6 +1975,7 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
             this.pokemonHatchedIcon,
             this.pokemonHatchedCountText
           ].map(c => c.setVisible(false));
+          this.pokemonFormText.setY(25);
         } else if (species.speciesId === Species.ETERNATUS) {
           this.pokemonHatchedIcon.setVisible(false);
```
Tempo-anon, line 1980: Sort of unrelated to this PR, but is this why users report a bug where Eternatus's candies aren't displaying correctly?

Tempo-anon (approved):

src/ui/starter-select-ui-handler.ts
ReneGV: 🤔 It could be the case Eternatus is the only hardcoded species on `starter-select-ui-handler.ts`.
Do they report wrong numbers on the candy count, or is it UI display error ?

chaosgrimmon: Eternatus is 10-cost, so it doesn't actually show up in any eggs, yeah. (Box Legendaries, etc. that come in Legendary (pink) eggs are only ever 8 to 9 starter cost.)

Having the egg icon not show up at all might be a point of confusion, sure, but otherwise, we'd have to decide on which sprite to use, since the colour of the egg changes to match the egg rarity it comes from.
I'm not sure that's worth it, and frankly, if someone has played long enough to *have* an Eternatus, they've hatched all the egg-only mons, and probably should have figured out by now that Eternatus does not come from eggs.

PR #2380, opened 2024-06-18
[Ability] Implemented Unburden 

MutenYoshii:
## What are the changes?

Added the ability unburden into the game by adding a boolean tag in summon data for when 
an item is lost

Unburden is a popular ability for some pokemon (Drifloon Line, Hawlucha)

## What did change?

Set the unburden ability to the speed multiplierattr with a new Item Lost Condition that returns the boolean
value from the summon data. Added lines that in the berry and moves file that set the value to true upon losing or 
using the item. 

### Screenshots/Videos

Unburden working upon eating a berry
https://github.com/pagefaultgames/pokerogue/assets/171395666/a3cab2b7-8c66-4b1a-9f57-d18f2e165d32

Unburden working when an item is stolen via Pluck
https://github.com/pagefaultgames/pokerogue/assets/171395666/895f9f03-314f-41d1-ac69-de5dfa39007d

Unburden working when an item is stolen via Thief
https://github.com/pagefaultgames/pokerogue/assets/171395666/00ddf005-c253-4302-89e0-7476818a40a6

Unburden being lost when switched out
https://github.com/pagefaultgames/pokerogue/assets/171395666/f1ccefa9-0999-4db2-b32e-e82fb5bed92d

Unburden coming back after ability is disabled
https://github.com/pagefaultgames/pokerogue/assets/171395666/2f631293-8415-441b-8bf0-f11c48b08aca

## How to test the changes?

In the overrides, create a situation where the pokemon with the ability will always go second due to speed
have the pokemon lose item either by use or theft
check if the pokemon now goes first in the interaction.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2723,11 +2723,6 @@ function getAnticipationCondition(): AbAttrCondition {
   };
 }
 
-function getItemLostCondition(): AbAttrCondition {
-  return (pokemon:Pokemon) => {
-    return pokemon.summonData.itemLost;
-  };
-}
 /**
  * Creates an ability condition that causes the ability to fail if that ability
  * has already been used by that pokemon that battle. It requires an ability to
@@ -4449,7 +4444,9 @@ export function initAbilities() {
       .attr(PostDefendCritStatChangeAbAttr, BattleStat.ATK, 6),
     new Ability(Abilities.UNBURDEN, 4)
       .attr(BattleStatMultiplierAbAttr, BattleStat.SPD, 2)
-      .condition(getItemLostCondition()),
+      .condition(function(user: Pokemon) {
+        return user.summonData.itemLost;
+      }),
     new Ability(Abilities.HEATPROOF, 4)
       .attr(ReceivedTypeDamageMultiplierAbAttr, Type.FIRE, 0.5)
       .ignorable(),
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2575,6 +2575,7 @@ export class BlockOneHitKOAbAttr extends AbAttr {
   }
 }
 
+
 export class IncrementMovePriorityAbAttr extends AbAttr {
   private moveIncrementFunc: (pokemon: Pokemon, move: Move) => boolean;
   private increaseAmount: integer;
@@ -2722,6 +2723,11 @@ function getAnticipationCondition(): AbAttrCondition {
   };
 }
 
+function getItemLostCondition(): AbAttrCondition {
+  return (pokemon:Pokemon) => {
+    return pokemon.summonData.itemLost;
+  };
+}
 /**
  * Creates an ability condition that causes the ability to fail if that ability
  * has already been used by that pokemon that battle. It requires an ability to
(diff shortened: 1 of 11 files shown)


hayuna:
[Ability] Implement Unburden

Please update title of PR

src/data/ability.ts
```diff
@@ -2528,6 +2529,11 @@ function getAnticipationCondition(): AbAttrCondition {
   };
 }
 
+function getItemLostCondition(): AbAttrCondition {
+  return (pokemon:Pokemon) => {
+    return pokemon.summonData.itemLost;
+  };
+}
```
OrangeRed, line 2536: This function is only used in one place, can you instead just add it as an anonymous fn directly on the ability

src/data/berry.ts
```diff
@@ -73,6 +73,7 @@ export function getBerryEffectFunc(berryType: BerryType): BerryEffectFunc {
       applyAbAttrs(DoubleBerryEffectAbAttr, pokemon, null, hpHealed);
       pokemon.scene.unshiftPhase(new PokemonHealPhase(pokemon.scene, pokemon.getBattlerIndex(),
         hpHealed.value, getPokemonMessage(pokemon, `'s ${getBerryName(berryType)}\nrestored its HP!`), true));
+      pokemon.summonData.itemLost = true;
```
OrangeRed, line 76: Instead of setting this variable to true inside of each of the individual berry functions wouldn't it better to just toggle it after the the `getBerryEffectFunc()` call?

MutenYoshii: I moved the toggle to modifier.ts when a berry is consumed

src/data/ability.ts
MutenYoshii: Removed this condition and moved it to an anonynmous fn in condition

OrangeRed:
Does this current implementation work in all circumstances? From what I can tell this currently doesn't work for moves like incinerate / knock off.

MutenYoshii:
Added the ability effect to include both Incinerate and Knock Off
Incinerate

https://github.com/pagefaultgames/pokerogue/assets/171395666/56522dbc-0583-4ec0-ba1e-d92242f3aa1d
Knock Off

https://github.com/pagefaultgames/pokerogue/assets/171395666/ba2bb3d3-efa8-49a5-a814-d379aa673157

OrangeRed (approved):

Dynamixus:
Hi! What about these:

> If Unburden is suppressed with [Neutralizing Gas](https://bulbapedia.bulbagarden.net/wiki/Neutralizing_Gas_(Ability)), a new boost cannot be gained, but a previous boost will return once [Neutralizing Gas](https://bulbapedia.bulbagarden.net/wiki/Neutralizing_Gas_(Ability)) leaves the field.

And also this:

> This boost is lost if the Pokémon gains a held item.

MutenYoshii:
> Hi! What about these:
> 
> > If Unburden is suppressed with [Neutralizing Gas](https://bulbapedia.bulbagarden.net/wiki/Neutralizing_Gas_(Ability)), a new boost cannot be gained, but a previous boost will return once [Neutralizing Gas](https://bulbapedia.bulbagarden.net/wiki/Neutralizing_Gas_(Ability)) leaves the field.
> 
> And also this:
> 
> > This boost is lost if the Pokémon gains a held item.

Hello, This is a video showing that it unburden works after neutralizing gas leaves the field 

https://github.com/user-attachments/assets/9c3f6da1-bea7-4fc9-b8f2-a8fd299cc5af

and for the second description, Since Pokemon can hold multiple items, I think it would be best that unburden activate only upon losing an item and losing the buff when switched out. I wouldn't mind changing it to the official way if needed.

flx-sta:
Hi @MutenYoshii.
Thank you for your contribution.
Are you still working on this?
If so please resolve/address the merge issue

flx-sta:
@MutenYoshii  closed for being stale (no updates for over a month)

----

PR #2384, opened 2024-06-18
[Ability] Implement Parental Bond

innerthunder:
## What are the changes?

This implements the ability [Parental Bond](https://bulbapedia.bulbagarden.net/wiki/Parental_Bond_(Ability)) with a new pre-attack ability attribute.

Parental Bond is currently unimplemented, and there isn't any other PR currently working on the implementation.

## What did change?

- `data/ability`: A new `PreAttackAbAttr`, `AddTwoStrikeAbAttr`, adds a second hit to most single-hit, single-target moves and applies a damage multiplier to this second hit. Parental Bond uses only this attribute in its implementation.
- `phases`: `AddTwoStrikeAbAttr` is now applied within the multi-hit logic in `MoveEffectPhase`.
- `field/pokemon`: `AddTwoStrikeAbAttr` now applies to a damage multiplier in the damage calculation in `Pokemon.apply()`

### Screenshots/Videos

The video below shows Parental Bond's interaction with Tackle and Power-Up Punch.

https://github.com/pagefaultgames/pokerogue/assets/168692175/29d36f15-db1a-4ea4-8028-3e110ac8f78b

The console logs show a second strike is added with a 0.25x damage multiplier.
image

## How to test the changes?

Unit tests (run via `npm run test`) are WIP.

## Checklist
- [x] Are the changes visual?


src/data/ability.ts
```diff
@@ -1229,6 +1229,78 @@ export class PokemonTypeChangeAbAttr extends PreAttackAbAttr {
   }
 }
 
+/**
+ * Class for abilities that convert single-strike moves to two-strike moves (i.e. Parental Bond).
+ * @param damageMultiplier the damage multiplier for the second strike, relative to the first.
+ */
+export class AddTwoStrikeAbAttr extends PreAttackAbAttr {
```
Tempo-anon, line 1236: nit: this should probably be named `AddSecondStrikeAbAttr` since AddTwo makes it sound like it's adding two rather than just another one

src/data/ability.ts
```diff
@@ -1229,6 +1229,78 @@ export class PokemonTypeChangeAbAttr extends PreAttackAbAttr {
...
+
+  constructor(damageMultiplier: number) {
+    super(false);
+
+    this.damageMultiplier = damageMultiplier;
+  }
+
+  canApplyPreAttack(move: Move, numTargets: integer): boolean {
+    const exceptAttrs: { new(...args: any[]): MoveAttr }[] = [
+      MultiHitAttr,
```
Tempo-anon, line 1250: This is actually a bit more complicated since if it can only hit 1 target (singles or everyone else is fainted) the second hit will be applied

src/data/ability.ts
```diff
@@ -1229,6 +1229,78 @@ export class PokemonTypeChangeAbAttr extends PreAttackAbAttr {
...
+  // This toggles within applyPreAttack() to determine whether the damage multiplier should be applied.
+  private firstStrike: boolean = true;
+
+  constructor(damageMultiplier: number) {
+    super(false);
+
+    this.damageMultiplier = damageMultiplier;
+  }
+
+  canApplyPreAttack(move: Move, numTargets: integer): boolean {
```
Tempo-anon, line 1248: Could use a doc on this function to describe in English what times it can apply

src/data/ability.ts
```diff
@@ -1229,6 +1229,78 @@ export class PokemonTypeChangeAbAttr extends PreAttackAbAttr {
...
+      && !exceptMoves.some(id => move.id === id)
+      && move.category !== MoveCategory.STATUS;
+  }
+
+  /**
+   * If conditions are met, this doubles the move's hit count (via args[1])
+   * or multiplies the damage of secondary strikes (via args[2])
+   * @param pokemon unused
+   * @param passive unused
+   * @param defender unused
```
Tempo-anon, line 1275: nit: I believe we use `n/a` for unused

Tempo-anon (commented):
How does this interact with fixed damage moves (secondHit should not get reduced damage) or multi-lens (for simplicity's sake you can just disable one or the other since Parental Bond already doesn't work on multi-hit moves)?

src/data/ability.ts
innerthunder: I think this case should be covered with `numTargets === 1` in the return statement. In `MoveEffectPhase`, the number that is passed to this attribute is basically `getTargets().length`.

I'll write unit tests for this case to make sure.

innerthunder:
## Interaction with Multi-Lens:

Like any multi-hit attack with Multi-Lens, Parental Bond-boosted attacks stack multiplicatively. For example, if a Pokemon with Parental Bond and 1 stack of Multi-Lens uses a single-strike attack, the attack will strike 4 times, with the strike order:
1. Strong Hit ($DamageMultiplier = 1 \times 0.4=0.4$)
2. Weak Hit ($DamageMultiplier = 0.25 \times 0.4 = 0.1$)
3. Strong Hit
4. Weak Hit

### Video of Interaction

Using the same test cases as the other video, except that the player Pokemon now has a Multi-Lens equipped:

https://github.com/pagefaultgames/pokerogue/assets/168692175/c85cd228-2d90-48ad-92f7-4225050b3d2c

Damage values when using Tackle during the test:

image

src/data/ability.ts
innerthunder: Made unit tests to verify that this interaction works correctly.

src/data/ability.ts
```diff
@@ -1230,6 +1230,86 @@ export class PokemonTypeChangeAbAttr extends PreAttackAbAttr {
...
+    const multiplier = args[2] as Utils.NumberHolder;
+
+    if (this.canApplyPreAttack(move, numTargets)) {
+      if (!!hitCount.value) {
+        hitCount.value *= 2;
+      }
+
+      if (!!multiplier.value && pokemon.turnData.hitsLeft % 2 === 1) {
+        multiplier.value *= this.damageMultiplier;
+      }
```
OrangeRed, line 1306: Can you do a pre-emptive null check here. If for some reason args[1..2] are somehow undefined you'd get a runtime error here.

src/data/move.ts
```diff
@@ -4582,7 +4582,7 @@ export class RemoveTypeAttr extends MoveEffectAttr {
   private messageCallback: ((user: Pokemon) => void) | undefined;
 
   constructor(removedType: Type, messageCallback?: (user: Pokemon) => void) {
-    super(true, MoveEffectTrigger.POST_APPLY);
+    super(true, MoveEffectTrigger.POST_TARGET);
     this.removedType = removedType;
     this.messageCallback = messageCallback;
```
OrangeRed, line 4587: Can you explain why this had to change to POST_TARGET?

src/data/ability.ts
```diff
@@ -1230,6 +1230,86 @@ export class PokemonTypeChangeAbAttr extends PreAttackAbAttr {
...
+      MultiHitAttr,
+      ChargeAttr,
+      SacrificialAttr,
+      SacrificialAttrOnHit
+    ];
+
+    /**
+     * Parental Bond cannot apply to these specific moves
+     */
+    const exceptMoves: Moves[] = [
```
OrangeRed, line 1267: small nitpick but one line comments like this I would write as `/** comment */` so they don't take up as much space

src/data/move.ts
innerthunder: POST_TARGET effects are only triggered once, on the last strike of the move, and after all other effects of the move are applied. Since Burn Up and Double Shock isn't meant to remove the user's type until after the last hit when they are boosted by Parental Bond, so this change fixes that interaction.

OrangeRed (approved):

innerthunder:
## Updates to multi-strike interactions

- Added a new `lastHitOnly` option to `MoveEffectAttr` which allows effects to trigger only on the last hit of a multi-hit move. This fixes the effect timing for the following moves:
    - Moves that ground the target (Smack Down, Thousand Arrows)
    - Moves that trap the target (e.g. Anchor Shot. **Note:** there's an unrelated issue where the `TRAPPED` tag incorrectly lapses at the end of the turn it is applied)
    - Moves that force the user or target to switch out (e.g. U-turn, Dragon Tail)
    - Moves that require the user to recharge on the next turn (e.g. Hyper Beam)
    - Moves that remove status effects from the target (e.g. Wake-Up Slap, Scald's thawing effect)
- Added a new function `Pokemon.stopMultiHit()` which, when called while applying a hit from a multi-hit move, cancels all subsequent hits. This is used to fix the following interactions:
    - When a hit has no effect, all subsequent hits are cancelled (i.e. effects from Storm Drain, etc. only proc once)
    - Multi-hit moves should hit into protected targets only once (i.e. effects from King's Shield, etc. only proc once)
    - If Present is used as a multi-hit move and rolls a heal on the first strike, subsequent strikes are cancelled. Also, if it rolls damage on the first strike, subsequent strikes cannot heal the target.
- Multi-hit moves are now only registered once in the user's `moveHistory` queue, fixing interactions with Fury Cutter and Echoed Voice
- Recoil is now a `lastHitOnly` effect, and recoil damage is now based on the user's cumulative damage dealt during the turn.
- Meloetta's transformation after using Relic Song only triggers after Relic Song's last strike.

## Testing new multi-hit interactions
Most effects that now use `lastHitOnly` were tested via unit tests. You can run `npm run test parental_bond` in terminal to check these interactions.

### Test Videos

Using Present with Parental Bond:

https://github.com/pagefaultgames/pokerogue/assets/168692175/92ca1ad4-f99e-4b91-8d65-de050dd0d7e8

Using a Recoil move:

https://github.com/pagefaultgames/pokerogue/assets/168692175/4dd8f083-60fc-4f4d-9286-4f2cff10076e

Using Pay Day:

https://github.com/pagefaultgames/pokerogue/assets/168692175/f8f0d87c-c302-4412-afac-4d08f7871d37

Using Relic Song as Meloetta:

https://github.com/pagefaultgames/pokerogue/assets/168692175/a61a332b-2902-4195-964b-d994dd3da39c

Tempo-anon (approved):

PR #2399, opened 2024-06-19
[QoL] Add Quick Ball and Timer Ball

bennybroseph:
## What are the changes?

Adds a Quick Ball and Timer Ball modifier which mimics mainline series behavior

This is some pre-commits for adding the Achievement Shop

## What did change?

Timer and Quick Ball modifiers have been added. Quick Ball only activates on the first turn of battle. Timer Ball ramps up starting on the second turn of battle. 

However, these modifiers are not new Poke Balls. They instead modify existing ones. When a ball is thrown, the modifier creates an overlay which fades out as the Ball shakes. If the catch would have failed but instead succeeds due to the increased multiplier, the overlay animates again.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/13838608/aad37f5c-74bd-4d4a-a4d7-51160821226c

## How to test the changes?

Since this modifier is not enabled in the item pool, you will need to add the modifier manually using overrides. 

## Checklist
- [x] Are the changes visual?


CodeTappert:
Why not making it a new item? 

The way you have done it is kinda confusing no? Or is this meant as "We will add them as seperate items later, but currently they arent"-kind of thing?

bennybroseph:
I think it's just a misunderstanding. They are new items, but I'm not adding them to the item pool. You can only get them in the Achievement Shop. Since that isn't in Main yet and I've been asked to add the items separately, this PR exists.

Heppity:
> I think it's just a misunderstanding. They are new modifiers, but not new Poke Balls, and I'm not adding them to the item pool. You can only get them in the Achievement Shop. Since that isn't in Main yet and I've been asked to add the items separately, this PR exists.

Could they work better as X items instead where they're active for a certain amount of battles? Could be a insetting choice in choosing one active poke ball effect at a time and you decide which one the best

bennybroseph:
> > I think it's just a misunderstanding. They are new modifiers, but not new Poke Balls, and I'm not adding them to the item pool. You can only get them in the Achievement Shop. Since that isn't in Main yet and I've been asked to add the items separately, this PR exists.
> 
> Could they work better as X items instead where they're active for a certain amount of battles? Could be a insetting choice in choosing one active poke ball effect at a time and you decide which one the best

No, they don't show up in the shop at all

flx-sta:
Work will continue elsewhere

PR #2407, opened 2024-06-19
[Feature] Synchronise

BartTerpstra:
## What are the changes?
Implements the ability Synchronise.

I think Synchronise is neat, especially as it existed in early generations.

## What did change?
-  [ ] synchronise now gives status to opposing pokémon
   -  [ ] shows itself if it would get a status or could give a status, even if they don't happen
   -  [ ] burn
   -  [ ] paralysed
   -  [ ] poison
   -  [ ] heavy poison is given as poison
   -  [ ] only when caused by the opponent
   -  [ ] ignore toxic spikes
   -  [ ] ignore substitute (not relevant, not implemented)
   -  [ ] works correctly when mirrored
   -  [ ] correct interaction with psychoshift
   -  [ ] ability triggers before user eats berry
 -  [ ] synchronise now has a 0%/50%/100% chance to force the nature of a wild pokémon to match the owners if the fight started with the owner in play
 -  [ ] synchronise does not affect eggs

### Screenshots/Videos

Not relevant?

## How to test the changes?
How can a reviewer test your changes once they check out on your branch?
I hope to make some test, but I'm still learning how to make these.

Did you just make use of the `src/overrides.ts` file?
No

Did you introduce any automated tests?
Intend to

Do the reviewer need to do something special in order to test your change?
No

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -3491,6 +3491,24 @@ export class WeightMultiplierAbAttr extends AbAttr {
   }
 }
 
+/** When ability owner would be burned, paralysed or poisoned, show ability.
+ *  if user becomes burned, paralysed or any kind of poisoned by the opposing pokemon,
+ *  inflict it on the opposing pokemon immediately if they can have this status.
+ *  all poisons are inflicted on opponent as regular poisoned.
+ *  "by" the opposing pokemon means ability or attack.
+ *  ability triggers before user eats berry.
+ *  triggers in between psycho shift transferring the status and psychoshift curing the user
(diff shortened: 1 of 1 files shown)


BartTerpstra:
The first issue with implementing Synchronise is whether or not to implement it's passive that changes wild pokémon natures.
It was 50% in early generations, then later changed to 100% and then (seemingly) removed in Scarlet and Violet.

BartTerpstra:
nvm, there already is a PR i missed.

PR #2421, opened 2024-06-19
[Enhancement][Challenge] Nuzlocke challenge

sirzento:
## What are the changes?

This PR adds the challenge mod "Nuzlocke" to the game. As a result, various new challenge checks will be added that can also be used for other challenges in the future.

I have seen some feature request for a nuzlocke challenge. Also, the old roadmap even had a nuzlocke mode as goal.

## What did change?

The Nuzlocke challenge aswell as the achievement for it was added. I also changed the challenge UI and made the description field larger to be able to display more text.

5 new `challengeType`'s where added to check for various rules in the nuzlocke challenge. 
Those types are:

- `NO_HEAL_PHASE`
Checks if the heal phase after every 10th round should be activated.

- `SHOP_ITEM_BLACKLIST`
Filters the shop for specific items.

- `RANDOM_ITEM_BLACKLIST`
Filters the random item selection for specific items.

- `ADD_POKEMON_TO_PARTY`
Checks if the current catched pokemon can be added to the party.

- `MOVE_BLACKLIST`
Checks if move the user wants to use is illegal for this challenge.

### Screenshots/Videos

Nuzlocke Regular:
grafik

Nuzlocke Hardcore:
grafik

## How to test the changes?

Set `POKEBALL_OVERRIDE` to have some masterballs and `MOVESET_OVERRIDE` to include `Moves.REVIVAL_BLESSING`.
You can also set `OPP_LEVEL_OVERRIDE` to 1.

Start a new challenge with Nuzlocke challenge enabled and check:
- if you are healed after the 10th round
- if you can buy revive
- if you can role for a reviver seed
- you can use revival blessing
- if you can only add one pokemon every 10 floors to the party

## Checklist
- [x] Are the changes visual?


hayuna:
@sirzento What about this scenario?

Opponent has REVIVAL_SEED
Player is using thief move

Player will take REVIVAL_SEED from opponent. Should this be illegal in nuzlocke mode?

sirzento:
> @sirzento What about this scenario?
> 
> Opponent has REVIVER_SEED Player is using thief move
> 
> Player will take REVIVER_SEED from opponent. Should this be illegal in nuzlocke mode?

@hayuna Yeah I thought about it and just thought it will be alright because it's just so rare to get it like that. I could disable it somehow if that's wanted but I think it will be alright like that.

src/locales/de/challenges.ts
```diff
@@ -23,4 +24,13 @@ export const challenges: TranslationEntries = {
     "desc_default": "Du kannst nur Pokémon des gewählten Typs verwenden."
     // types in pokemon-info
   },
+  "nuzlocke": {
+    "name": "Nuzlocke",
+    "desc": "Die Nuzlocke Challenge vordert verschiedene Einschränkungen um eine besondere Challenge zu gestallten.",
```
CodeTappert, line 29:
```suggestion
"desc": "Die Nuzlocke-Challenge beinhaltet verschiedene Einschränkungen, um eine besondere Herausforderung zu gestalten."
```

src/locales/de/challenges.ts
```diff
@@ -23,4 +24,13 @@ export const challenges: TranslationEntries = {
...
+    "desc.1": "Nuzlocke Regeln:\n- Keine Beleber.\n- Nur ein Pokémon pro 10 Wellen kann zum Team hinzugefügt werden.\n",
```
CodeTappert, line 30:
```suggestion
"desc.1": "Nuzlocke-Regeln:\n- Keine Beleber.\n- Pro 10 Wellen darf nur ein Pokémon zum Team hinzugefügt werden.\n"
```

src/locales/de/achv.ts
```diff
@@ -264,6 +264,14 @@ export const PGMachv: AchievementTranslationEntries = {
   "MONO_FAIRY": {
     name: "Ein ewiges Abenteuer!",
   },
+  "NUZLOCKE_REGULAR": {
+    name: "Unfaintable",
+    description: "Complete the regular nuzlocke challenge.",
```
CodeTappert, line 269:
```suggestion
  "NUZLOCKE_REGULAR": {
    "name": "Unbezwingbar",
    "description": "Schließe die normale Nuzlocke-Herausforderung ab.",
```

src/locales/de/achv.ts
```diff
@@ -264,6 +264,14 @@ export const PGMachv: AchievementTranslationEntries = {
...
+  },
+  "NUZLOCKE_HARDCORE": {
+    name: "How?",
+    description: "Complete the hardcore nuzlocke challenge.",
```
CodeTappert, line 273:
```suggestion
  "NUZLOCKE_HARDCORE": {
    "name": "Wie?",
    "description": "Schließe die Hardcore-Nuzlocke-Herausforderung ab.",
```

src/locales/de/challenges.ts
sirzento: Why did you remove the ","? This is needed ^^Is it possible to change this suggested code here?

src/locales/de/challenges.ts
```diff
@@ -23,4 +24,13 @@ export const challenges: TranslationEntries = {
     "desc_default": "Du kannst nur Pokémon des gewählten Typs verwenden."
     // types in pokemon-info
   },
+  "nuzlocke": {
+    "name": "Nuzlocke",
+"desc": "Die Nuzlocke-Challenge beinhaltet verschiedene Einschränkungen, um eine besondere Herausforderung zu gestalten."
```
CodeTappert, line 29:
```suggestion
"desc": "Die Nuzlocke-Herausforderung beinhaltet verschiedene Einschränkungen, um eine besondere Herausforderung zu gestalten."
```

src/locales/de/challenges.ts
CodeTappert: Okay sorry i didnt mean to remove the ,

CodeTappert (approved):

src/data/challenge.ts
```diff
@@ -733,6 +811,97 @@ export class LowerStarterPointsChallenge extends Challenge {
...
+      overrideValue = this.value;
+    }
+    let desc: string = i18next.t("challenges:nuzlocke.desc");
+    switch (overrideValue) {
+    case 1:
+      desc = i18next.t("challenges:nuzlocke.desc.1");
+      break;
+    case 2:
+      desc = i18next.t("challenges:nuzlocke.desc.1");
+      desc += i18next.t("challenges:nuzlocke.desc.2");
```
Xavion3, line 889: Can we just have two localisation entries for levels 1 and 2 instead of concatenating them? We try to avoid just concatenating multiple localisation when possible. You'd be able to remove this entire function override if you did that too.

src/data/challenge.ts
```diff
@@ -104,6 +130,8 @@ export abstract class Challenge {
 
   public conditions: ChallengeCondition[];
 
+  public additionalData: {[x: string]: any};
```
Xavion3, line 133: Don't do this, you can include additional values in the specific challenge instances when necessary.

src/data/challenge.ts
```diff
@@ -733,6 +811,97 @@ export class LowerStarterPointsChallenge extends Challenge {
...
+    const moveBlacklist = [Moves.REVIVAL_BLESSING];
+    moveCanBeUsed.value = !moveBlacklist.includes(move.moveId);
+    return true;
+  }
+
+  applyStarterChoice(pokemon: PokemonSpecies, valid: Utils.BooleanHolder, dexAttr: DexAttrProps, soft?: boolean): boolean {
+    if (this.value === 1) {
+      return false;
+    }
+    valid.value = !pokemon.legendary;
```
Xavion3, line 861: Why restrict only legends and not say, paradox, minor legends, mythicals, etc.? It seems very arbitrary.

src/system/challenge-data.ts
```diff
@@ -4,11 +4,13 @@ export default class ChallengeData {
   public id: integer;
   public value: integer;
   public severity: integer;
+  public additionalData: {[x: string]: any};
 
   constructor(source: Challenge | any) {
     this.id = source.id;
     this.value = source.value;
     this.severity = source.severity;
+    this.additionalData = source.additionalData;
```
Xavion3, line 13: This will not work as is and will require server changes.

src/ui/challenges-select-ui-handler.ts
```diff
@@ -52,7 +52,7 @@ export default class GameChallengesUiHandler extends UiHandler {
     this.challengesContainer.add(bgOverlay);
 
     // TODO: Change this back to /9 when adding in difficulty
-    const headerBg = addWindow(this.scene, 0, 0, (this.scene.game.canvas.width / 6), 24);
+    const headerBg = addWindow(this.scene, 0, 0, (this.scene.game.canvas.width / 6) - 2, 24);
```
Xavion3, line 55: For this, and all the changes below, why are you changing the size of the UI elements? What will the challenge selection screen look like after you do so?

src/phases.ts
```diff
@@ -5008,6 +5020,13 @@ export class AttemptCapturePhase extends PokemonPhase {
         });
       };
       Promise.all([pokemon.hideInfo(), this.scene.gameData.setPokemonCaught(pokemon)]).then(() => {
+        const challengeCanAddToParty = new Utils.BooleanHolder(true);
+        applyChallenges(this.scene.gameMode, ChallengeType.ADD_POKEMON_TO_PARTY, pokemon, this.scene.currentBattle.waveIndex, challengeCanAddToParty);
+        if (!challengeCanAddToParty.value) {
+          removePokemon();
+          end();
+          return;
+        }
```
Xavion3, line 5029: What does this actually look like in game? It looks like it only allows you to add the first pokemon caught in each section to your party, instead of only allowing you to add one pokemon caught in each section to your party?

src/modifier/modifier-type.ts
```diff
@@ -1807,14 +1809,18 @@ export function getModifierTypeFuncById(id: string): ModifierTypeFunc {
   return modifierTypes[id];
 }
 
-export function getPlayerModifierTypeOptions(count: integer, party: PlayerPokemon[], modifierTiers?: ModifierTier[]): ModifierTypeOption[] {
+export function getPlayerModifierTypeOptions(count: integer, party: PlayerPokemon[], gameMode: GameMode, modifierTiers?: ModifierTier[]): ModifierTypeOption[] {
```
Xavion3, line 1812: You don't need to add an argument here since the scene object is already passed in.

Xavion3 (changes requested):
For the new signatures, can we be clearer whether the apply functions are expected to set it to true or false for each? It looks like they all assume true, and set to false if unexpected, but just having some kind of note in the function documentation about that would be good.

src/data/challenge.ts
sirzento: I did that because the level 2 description is exactly as the first one but with additional text. But I can change that if you want.

src/data/challenge.ts
sirzento: Yeah that's something to discuss what should be used there for balancing. I can also restrict those if that would be a better overall challenge.

src/system/challenge-data.ts
sirzento: Mhh are you 100% sure about that? I tested it with an running server and it did work fine since the challenges have no defined table schema inside the db. (Not 100% sure since I have just the knowledge from looking at the server code and nothing more)

src/data/challenge.ts
sirzento: Do those values persist after reloading etc? With the current setup I can save data in there that can be loaded back up again after reloading

src/ui/challenges-select-ui-handler.ts
sirzento: It will look like the screenshots I posted. Without those changes the challenge text box was to small for displaying all the text of the challenge. Is it a proplem that I modified it?

src/phases.ts
sirzento: If you catch a pokemon and this challenge check returns false, it just wont get added to the party. So ingame it looks the same as every other catch with the exception that you wont have it in your team after that.

src/modifier/modifier-type.ts
sirzento: Am I missing something or how can I access the gameMode object there via scene?

src/data/challenge.ts
Xavion3: Similar text is fine, it's expected that some localization entries will be similar.

src/data/challenge.ts
Xavion3: I'm just thinking, it feels weird to restrict say, Lugia but not Arceus is all. I think realistically this restriction could just not exist baked in, it has no bearing on the rules of a nuzlocke after all.

src/modifier/modifier-type.ts
Xavion3: Pokémon always have access to the scene.

src/phases.ts
sirzento: And yes, the first pokemon you catch each biome/10 waves will be added to the team and all others after that won't be added but you can still catch them if you want.

src/data/challenge.ts
sirzento: We could also restrict the other ones. That could actually be better than just legendaries. And yes, restricting legendaries is not part of the basic nuzlocke ruleset but it is listed as optional rule that people use to increase the difficulty. Same with the restriction of the healing phase. That would be the same as  a "no pokemon center" optional rule.

src/system/challenge-data.ts
sirzento: After testing it again and looking more into the backend server code I have to say that you are right. I must have mixed up the ChallengeData field with another one and thought it should work like this. So the `ChallengeData` struct would only need this field to be added for it to work. Other challenges would not be affected by the change as they do not use this data.

But is it a problem that I have defined this field? I think it's not so bad to have a way to store challenge specific data like the information on which wave a Pokemon was last caught. Or what is your opinion on this?

sirzento: I think I'll change it so that only the FIRST Pokemon in each biome can be caught. Just like in a classic Nuzlocke Challenge. Then I would no longer need this field.

sirzento:
> For the new signatures, can we be clearer whether the apply functions are expected to set it to true or false for each? It looks like they all assume true, and set to false if unexpected, but just having some kind of note in the function documentation about that would be good.

I added some information to the comment above the apply functions that describes better what the boolean holder does. Thats what you meant right?

DayKev:
Can you fix the PR and then split this into multiple discrete challenges (ie: permadeath challenge, no auto-heal challenge, etc)?

flx-sta:
@sirzento I'm putting this into draft while you resolve those conflicts and work on the open comments.

---

Are you actually still actively working on this?

DayKev:
Ugh, trying to rebase this on current beta and it's just completely fucked. We really should have reverted the first beta merge and redid it correctly with rebasing...

DayKev:
image

Now to check if I missed anything when manually copying the changes...

DayKev:
Draft update at #4311

flx-sta:
@DayKev so should this one be closed in favor of yours?

DayKev:
> @DayKev so should this one be closed in favor of yours?

Probably? I guess since the history is broken in this one it can't be used regardless.

PR #2443, opened 2024-06-20
[Fix] Fix randomness in Gastro Acid unit test

zacharied:
## What are the changes?

Prevent Gastro Acid tests from randomly failing due to type-immune opponents.

If a Fairy-type opponent was randomly chosen for the test "fails if used on an enemy with an already-suppressed ability," the test would fail since the first attack (Core Enforcer) would be immuned by the Fairy-type. Thus the fairy-type would not get its ability suppressed, so Gastro Acid would be successful on the next turn.

## What did change?

Add opponent species override as Bidoof in Gastro Acid unit test helper `beforeEach`.

## Checklist
 
[x] There is no overlap with another PR?
[x] The PR is self-contained and cannot be split into smaller PRs?
[x] Have I provided a clear explanation of the changes?
[x] Have I tested the changes (manually)?
  [x] Are all unit tests still passing? (npm run test)
[] Are the changes visual?
  [] Have I provided screenshots/videos of the changes?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -5290,7 +5290,16 @@ export class SwitchAbilitiesAttr extends MoveEffectAttr {
   }
 }
 
+/**
+ * Attribute used for moves that suppress abilities like {@linkcode Moves.GASTRO_ACID}.
+ * A suppressed ability cannot be activated.
+ *
+ * @extends MoveEffectAttr
+ * @see {@linkcode apply}
+ * @see {@linkcode getCondition}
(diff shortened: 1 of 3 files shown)


zacharied:
@Tempo-anon

CodeTappert:
I would love from you if you could describe a little more how and what you fixed :)

zacharied:
Sorry, got lazy bc temp and I had already talked in discord. I've updated the description.

Issue #2438, opened 2024-06-20
[BUG] Sometimes npm run test:silent failed irregularly.


EnochG1:
**Describe the bug**
When test workflow runs at Pull Request, src/test/abilities/screen_cleaner.test.ts makes error sometimes.
https://github.com/pagefaultgames/pokerogue/actions/runs/9591456882/job/26448400826?pr=2395

I was confused because it has no connection to the modified code.
And when I modified simple things (like modify single string) to run workflow again, tests work successful.
https://github.com/pagefaultgames/pokerogue/actions/runs/9591522884/job/26448574603?pr=2395

**To Reproduce**
It's hard to reproduce but occurs when PR runs 'npm run test:silent'.

**Expected behavior**

**Screenshots / Videos**
![image](https://github.com/pagefaultgames/pokerogue/assets/19601207/19ef7749-f8be-4351-833e-6e24c8df65fc)

**Device**
Chrome, but occurs at github.

**Additional context**

zacharied:
This happens in the rare chance that the test generates an encounter with a Pokemon with Poison Heal. This can be replicated with the following override:

`vi.spyOn(overrides, "OPP_ABILITY_OVERRIDE", "get").mockReturnValue(Abilities.POISON_HEAL);`

DayKev:
> This happens in the rare chance that the test generates an encounter with a Pokemon with Poison Heal. This can be replicated with the following override:
> 
> `vi.spyOn(overrides, "OPP_ABILITY_OVERRIDE", "get").mockReturnValue(Abilities.POISON_HEAL);`

But the test that is failing seems to explicitly use Magikarps for the test, so how is that possible?

At the start of each test (L36, 53, 70) in `src/test/abilities/screen_cleaner.test.ts` is:  `await game.startBattle([Species.MAGIKARP, Species.MAGIKARP]);`

torranx:
This is a bug on the poison heal, all screen cleaner passed the test but the error still shows.
![image](https://github.com/pagefaultgames/pokerogue/assets/68144167/fbe6dd33-49a8-4d5b-b143-59f604e51dd2)

PR #2444, opened 2024-06-20
[Enhancement] Optimize Unit tests

innerthunder:
## What are the changes?

Unit tests for the following abilities and moves have been optimized to finish much faster:
- Unseen Fist
- Follow Me
- Rage Powder
- Spotlight
- Battery

Because of issues with how I originally wrote them, these tests take a much longer time than needed.

## What did change?

- `unseen_fist`, `follow_me`, `rage_powder`, and `spotlight`: Fixed an issue where some tests would unnecessarily trigger the `FaintPhase`.
- `battery`: Fixed an issue where tests unnecessarily run through the `TurnEndPhase`

### Screenshots/Videos

N/A

## How to test the changes?

`npm run test`. Compare the runtime between this branch and `upstream/main`.

## Checklist
- [N/A] Are the changes visual?
  - [N/A] Have I provided screenshots/videos of the changes?

### DIFF
--- a/src/test/abilities/battery.test.ts
+++ b/src/test/abilities/battery.test.ts
@@ -3,7 +3,6 @@ import Phaser from "phaser";
 import GameManager from "#app/test/utils/gameManager";
 import * as overrides from "#app/overrides";
 import { Species } from "#enums/species";
-import { TurnEndPhase, } from "#app/phases";
 import { Moves } from "#enums/moves";
 import { getMovePosition } from "#app/test/utils/gameManagerUtils";
 import Move, { allMoves, MoveCategory } from "#app/data/move.js";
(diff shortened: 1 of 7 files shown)


src/test/abilities/battery.test.ts
```diff
@@ -44,8 +43,6 @@ describe("Abilities - Battery", () => {
     const multiplier = getAttrPowerMultiplier(game.scene.getPlayerField()[1]);
     const appliedPower = getAppliedMovePower(game.scene.getEnemyField()[0], game.scene.getPlayerField()[0], allMoves[moveToBeUsed]);
 
-    await game.phaseInterceptor.to(TurnEndPhase);
```
torranx, line 47: I've checked this and this still only works because it's not apparently checking the actual damage dealt, hence the entire actual battle is unnecessary.

innerthunder:
### More Updated Tests:
- `power_spot`: Same optimization as `battery`.
- `screen_cleaner`: Fixed an issue where random opponent selection would sometimes cause the Aurora Veil test to fail. https://github.com/pagefaultgames/pokerogue/issues/2438

torranx (approved):
lgtm
note to future reader: battery and power spot tests right now only [mocks](https://github.com/pagefaultgames/pokerogue/blob/main/src/test/abilities/battery.test.ts#L101) the applied values and not compared in the actual battle- hence the turn events are unnecessary.

Ideally we want to compare the mocked multiplied power value vs the applied multiplied power during battle but that seems not possible right now

Greenlamp2 (approved):

PR #2451, opened 2024-06-20
[QoL] IV scanner update to hide the prompt and let you see the stats mid battle on the stats screen

Opaque02:
## What are the changes?
The IV scanner has been updated in two ways - the first is a setting that will allow you to disable the IV scanner prompt and all associated parts (animation, message etc), and the second is to change the stat container labels to show green/gold for the IV mid battle to show if the stats of the pokemon are better than what you've currently got.

A lot of people on discord have mentioned this is a feature they wanted, so I decided to make it

## What did change?
The first option is a setting that enables or disables the prompt about scanning a pokemon's IV. This is under the "general" tab of the settings.

The second option is to change the labels of the stat container to be green or gold depending on the stat. The stats container will show you the same amount of IV updates that you would from the IV scanner - so if you have no scanners you won't see any updates, all the way up to seeing all 6 stats for 3 scanners. If the stat is better than what you have, it'll show green, whereas if it's perfect IV it'll show gold; this is so if you do have the prompts being skipped, you can still see if the stat is perfect. Lastly, if the stat is unknown (not enough scanners), not scanned (ACC/EVA), or equal/worse than what you have, the label is white as usual.

The last change is to do with the stat labels themselves. Now if you have at least one IV scanner, the enemy stats will show a label for HP. There is no +/- number associated with it, as it's just there purely for IV purposes. This label does not show for player pokemon, and only shows when an IV scan would happen - so this label doesn't show for trainer battles

### Screenshots/Videos
Here is a screenshot of the settings screen showing the new setting:

image

Here is a screenshot of the stats screen in battle showing some stats and the IV scanner on screen:

image

As you can see, the only stats that are better than what we have are HP, Def and Sp. Def. Since Def is "Best" (i.e. 31), it shows up gold on the stats container. However, HP and Sp. Def are not "Best", so it shows up green. The Atk for this pokemon isn't better than what we have, so it's white like usual. The other stats are unknown/unscanned, so they also show up as white. This is with 2 IV scanners, so only 4 stats are shown.

Below is an example of a fight where pokemon have changed their stats with various stat changing moves: 

image

Below is an example with 2 IV scanners in a run showing the difference with the IV scanner setting on (current behaviour) and the IV scanner setting off:

https://github.com/pagefaultgames/pokerogue/assets/66582645/47438007-3edb-4a4e-a0f3-74605671ad12

Here's an example of a run where there's no IV scanners. You'll notice that the HP stat isn't shown, but everything has been moved over a bit. This is so that when the HP stat does get shown, the ui elements don't shift because of it:

https://github.com/pagefaultgames/pokerogue/assets/66582645/4e590d8e-56d3-4c78-bf69-a7d767bd1499

Here's an example with 3 IV scanners:

https://github.com/pagefaultgames/pokerogue/assets/66582645/77a66662-862d-4d34-8aec-2dfc2387d274

Here's a double battle with 1 IV scanner while also going through different waves:

https://github.com/pagefaultgames/pokerogue/assets/66582645/7696524a-6c99-45ee-855f-1be7cec2a94f

Below is an example of fighting a trainer and not seeing the IV scanner prompt:

https://github.com/pagefaultgames/pokerogue/assets/66582645/c2900bdf-7013-4303-87dc-b7e8a188b3cc

Lastly, below is an example where you start with no IV scanners and then pick up on in the store:

https://github.com/pagefaultgames/pokerogue/assets/66582645/4dab40d5-9ed9-47c5-a1f2-ab89ed8c2de8

Finally, here's a screenshot of the IV scanner in action for the legacy UI:

image

## How to test the changes?
Download the branch to play with the changes. When you're at the battle screen, you can open the stats container (default Shift/C) to see the changes. Also in the settings, there should be an option to turn the setting on/off to hide the IV prompts

You can also use any of the following overrides:

Line 32 - this will force a double battle:

`export const DOUBLE_BATTLE_OVERRIDE: boolean = true;`

Line 90 - this will let you choose a specific pokemon to fight against. For example, this is set to spawn a charizard, but you can change that to anything else to fight other things:

`export const OPP_SPECIES_OVERRIDE: Species | integer = Species.CHARIZARD;`

Line 96 - this will let you set the moves the enemy opponent has; for example this could be used to set stat changing moves for enemy pokemon to see the stat numbers appear. In this example, we have given them the moves no retreat, quiver dance, and shell smash twice. This can be set to anything:

`export const OPP_MOVESET_OVERRIDE: Array<Moves> = [Moves.NO_RETREAT, Moves.QUIVER_DANCE, Moves.SHELL_SMASH, Moves.SHELL_SMASH];`

Line 99 - this will let you set the IVs of the enemy pokemon. It takes an array of 6 numbers, which, in order, are HP, Atk, Def, Sp. Atk, Sp. Def, Spd:

`export const OPP_IVS_OVERRIDE: integer | integer[] = [31, 24, 5, 3, 19, 16];`

Line 121 - this will give you 3 IV scanners. To change it, change the "count: 3" to "count: 1" or "count: 2":

`export const STARTING_MODIFIER_OVERRIDE: Array<ModifierOverride> = [{name: "IV_SCANNER", count: 3}];`

**NOTE:** If you want to get rid of your IV scanners, you have to revert this; changing the count to 0 will NOT work, so to have 0 IV scanners, you need to use:

`export const STARTING_MODIFIER_OVERRIDE: Array<ModifierOverride> = [];`

Also note that for the IV scanners, you'll need to start a new game to see the effects. At first it'll look like it didn't work, but you'll be given the choice to see the pokemon's IVs before the fight as usual.

## Checklist
- [x] Are the changes visual?


Kelvez:
Do you think it is possible to add the possibility to see if the pokemon has better stats without having to press a key.
Maybe an icon next to the pokeball icon or a different color on the pokemon name or something else.
Be sure to not miss pokemon with better stats because we forgot to press the button.
And a big big thanks for implementing this feature 🙏. You dropped this 👑

Opaque02:
@Kelvez - yay, I'm glad you like it so far! So, ui stuff usually has to be approved by devs, so I'd have to see if they're happy with another icon near the pokeball, but based on the conversations we had around this, I can't promise that they'd go for it - I think they want to try and keep it pretty contained with the stat container for this feature

src/locales/en/settings.ts
```diff
@@ -15,6 +15,7 @@ export const settings: SimpleTranslationEntries = {
   "skipSeenDialogues": "Skip Seen Dialogues",
   "battleStyle": "Battle Style",
   "enableRetries": "Enable Retries",
+  "hideIvs": "Hide IV scanner",
```
f-fsantos, line 18: Please add the keys in all the languages. You can reuse the english version, but make sure all the keys are there.

src/phases.ts
```diff
@@ -5407,19 +5407,45 @@ export class ScanIvsPhase extends PokemonPhase {
...
+    const uiTheme = (this.scene as BattleScene).uiTheme; // Assuming uiTheme is accessible
+    for (let e = 0; e < enemyField.length; e++) {
+      enemyIvs = enemyField[e].ivs;
+      const currentIvs = this.scene.gameData.dexData[enemyField[e].species.speciesId].ivs;
+      const ivsToShow = this.scene.ui.getMessageHandler().topIvs(enemyIvs, this.shownIvs);
+      statsContainer = enemyField[e].getBattleInfo().getStatsValueContainer().list;
+      statsContainerLabels = statsContainer.filter(m => m.name.indexOf("icon_stat_label") >= 0);
+      for (let s = 0; s < statsContainerLabels.length; s++) {
+        const ivStat = Stat[statsContainerLabels[s].frame.name];
+        if (enemyIvs[ivStat] > currentIvs[ivStat] && ivsToShow.indexOf(Number(ivStat)) >= 0) {
```
f-fsantos, line 5423: This behavior is fundamentally different from what exists today, you shouldn't be using your currentIv as a comparison point to change the UI, it should only be around the scanned values themselves. Try to keep the same logic that exists today to determine the color of IV (1,10,20,30,31).

f-fsantos (changes requested):

src/phases.ts
Opaque02: @f-fsantos This was actually discussed in the discord with Temp and other members of the community. Currently I think the numbers you've got there are related to the wording - things like Good, Best, Fantastic etc - but the colour is just that if it's better than what you've currently got, it's green. [Here's](https://github.com/pagefaultgames/pokerogue/blob/01f6ec2ef301dee9a6bfddd8f2cee4f91bd11864/src/phases.ts#L5470) a link for the ScanIvsPhase where we go to the promptIvs method, which is [here](https://github.com/pagefaultgames/pokerogue/blob/01f6ec2ef301dee9a6bfddd8f2cee4f91bd11864/src/ui/battle-message-ui-handler.ts#L192). Within that, for each shown stat, we call [getIvDescriptor](https://github.com/pagefaultgames/pokerogue/blob/01f6ec2ef301dee9a6bfddd8f2cee4f91bd11864/src/ui/battle-message-ui-handler.ts#L216C74-L216C89), which gets the text and the colour for the IV. Finally, that [getIvDescriptor](https://github.com/pagefaultgames/pokerogue/blob/01f6ec2ef301dee9a6bfddd8f2cee4f91bd11864/src/ui/battle-message-ui-handler.ts#L230) method checks to see if the [IV stat of the enemy is better than what your pokemon's IVs are](https://github.com/pagefaultgames/pokerogue/blob/01f6ec2ef301dee9a6bfddd8f2cee4f91bd11864/src/ui/battle-message-ui-handler.ts#L237), and if so, it shows as TextStyle.SUMMARY_GREEN, otherwise it shows as the default TextStyle.SUMMARY colour. The numbers [1, 10, 20, 30, 31] are only used [to get the actual wording](https://github.com/pagefaultgames/pokerogue/blob/01f6ec2ef301dee9a6bfddd8f2cee4f91bd11864/src/ui/battle-message-ui-handler.ts#L242) of the stat, not the colour.

For this PR, we decided to make it so that if an IV can be shown and is better than what you've got, it'll show up as green, which is the same as what we've got in live, but if it's Best (31), it shows in gold as the feedback people were having was that they'd like to know a way to tell if an IV was Best (31) for a stat that people didn't have so that people can choose to catch it if they choose. Again these only show if it's better than what you've got, so if you've already got a 31 IV stat, it won't show up on an enemy

src/locales/en/settings.ts
Opaque02: This should be fixed in the latest commit

CodeTappert:
Honestly i find this very confusing. Why not just show the same dialogue as when the scanner is activated the first time?

The colors are hard to distinct for people with color problems

Opaque02:
> Honestly i find this very confusing. Why not just show the same dialogue as when the scanner is activated the first time?
> 
> The colors are hard to distinct for people with color problems

@CodeTappert do you mean the "Use IV scanner on {pokemon name}" prompt? The idea was to not show it so that you didn't have to wait through the animations and wait for it to scan, yet to still be able to see if an enemy pokemon has a better IV in case you want to catch it. For example, you may not care about catching most pokemon so you can turn the prompt off, but there may be some you still care about, so you can see it when you open the stats container.

When we were thinking of the colour, we didn't think about people with difficulties with colours. Do you have suggestions on some good colours we could use instead? Temp originally said green for a stat that's better than what you've got and gold for something at 31 IV so you can still tell when an IV is at max (since lots of people still want that, and without the prompt, we need two colours), but we can change it if there's more suitable colour choices

CodeTappert:
> > Honestly i find this very confusing. Why not just show the same dialogue as when the scanner is activated the first time?
> > 
> > The colors are hard to distinct for people with color problems
> 
> @CodeTappert do you mean the "Use IV scanner on {pokemon name}" prompt? The idea was to not show it so that you didn't have to wait through the animations and wait for it to scan, yet to still be able to see if an enemy pokemon has a better IV in case you want to catch it. For example, you may not care about catching most pokemon so you can turn the prompt off, but there may be some you still care about, so you can see it when you open the stats container.
> 
> When we were thinking of the colour, we didn't think about people with difficulties with colours. Do you have suggestions on some good colours we could use instead? Temp originally said green for a stat that's better than what you've got and gold for something at 31 IV so you can still tell when an IV is at max (since lots of people still want that, and without the prompt, we need two colours), but we can change it if there's more suitable colour choices

No i mean the result screen that is shown after the animation. Which has actual text

CodeTappert:
image

Opaque02:
> > > Honestly i find this very confusing. Why not just show the same dialogue as when the scanner is activated the first time?
> > > The colors are hard to distinct for people with color problems
> > 
> > 
> > @CodeTappert do you mean the "Use IV scanner on {pokemon name}" prompt? The idea was to not show it so that you didn't have to wait through the animations and wait for it to scan, yet to still be able to see if an enemy pokemon has a better IV in case you want to catch it. For example, you may not care about catching most pokemon so you can turn the prompt off, but there may be some you still care about, so you can see it when you open the stats container.
> > When we were thinking of the colour, we didn't think about people with difficulties with colours. Do you have suggestions on some good colours we could use instead? Temp originally said green for a stat that's better than what you've got and gold for something at 31 IV so you can still tell when an IV is at max (since lots of people still want that, and without the prompt, we need two colours), but we can change it if there's more suitable colour choices
> 
> No i mean the result screen that is shown after the animation. Which has actual text

The problem is showing that on double battles, there's no easy way to show which is which. Also people wanted something unrelated to the current IV prompt so we ended up with this design

CodeTappert:
The feature works fine...But as i said. I dont like that is only colours an no text. But since this is only a problem for people like me i will approve it.

CodeTappert (approved):

torranx:
Closing because of the remade PR linked, if this is a mistake, please ping me.

PR #2460, opened 2024-06-20
[Bug] Fix Minimize AI and add NaN AI score safeguard

Xavion3:
## What are the changes?
Different browsers will no longer treat Minimize differently for the AI.

Minimize was being chosen inconsistently across browsers and causing other moves to also be used inconsistently.

## What did change?
The minimized battler tag now has a score of 0. In addition if a target's score is NaN it'll now be logged in the console and treated as 0 instead.

## How to test the changes?
Give an AI minimize and some other moves, try in firefox vs chrome and notice that the list of moves sorted by score was different without this patch and consistent with this patch.

## Checklist

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -3964,6 +3964,8 @@ export class AddBattlerTagAttr extends MoveEffectAttr {
       return -3;
     case BattlerTagType.ENCORE:
       return -2;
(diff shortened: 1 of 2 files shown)


Tempo-anon (approved):
~~Is the change needed in both places?~~ didn't see minimize didn't have a score to begin with

Issue #1816, opened 2024-06-05
[BUG] Metal Burst fails 


hightowernl:
**Describe the bug**

Metal Burst calculates the damage the user took and then double it to the opponent, if the user moves last. I faced an Electabuzz that was faster and used Volt Switch. It switched into Electrode and my Metal Burst failed. It should do damage even if the opponent used a move like Volt Switch or U-Turn.
**To Reproduce**

1. Face a faster opponent with Volt Switch
2. Use Metal Burst while the opponent uses Volt Switch
3. Opponent switches into another Pokemon

**Expected behavior**

Metal burst should double the damage even when the opponent used a move that switched it out.
**Screenshots / Videos**

**Device**

PC
**Additional context**

schmidtc1:
This bug also affects Mirror Coat and Counter, since the target is set to the attacker but the attacker switches out, the move fails instead of redirecting the target to the new opponent.

PR #2462, opened 2024-06-20
[Bug] Adjust how counter attacks target to account for uturn/voltswitch and double battles

schmidtc1:
#1816

## What are the changes?

Fixes counterattack (Counter, Mirror Coat, Metal Burst, Comeuppance) targeting when the attacker switches out with uturn/volt switch/etc.
Also, Metal Burst and Comeuppance have additional target switching for when the attacker faints in a double battle:
"However, if the last Pokémon to hit the user faints before the user moves, the remaining enemy Pokémon will be hit instead." 
This does not apply to Counter and Mirror Coat
https://bulbapedia.bulbagarden.net/wiki/Metal_Burst_(move)
https://bulbapedia.bulbagarden.net/wiki/Comeuppance_(move)

The moves will fail if hit by U-Turn or switch out attacks, when they should still hit the switched-in opponent.
Additionally, Metal Burst and Comeuppance have their own target redirection that hasn't been implemented prior to this PR.

## What did change?

All counterattack moves will now deal the counter damage to the switched in target from U-Turn/Volt Switch/switch out attacks. 
Metal Burst and Comeuppance will now redirect damage to remaining opponents if the original attacker faints before the counterattack goes off.

### Screenshots/Videos

Before (single battle, fails)

https://github.com/pagefaultgames/pokerogue/assets/62030095/6e6798ee-19f0-406a-9c16-5b7c61a750ac

After (single battle, no longer fails)

https://github.com/pagefaultgames/pokerogue/assets/62030095/dcc250d8-742f-49e3-b5d8-1e116f3dbdda

Before redirect in double battles (fails)

https://github.com/pagefaultgames/pokerogue/assets/62030095/cfb6c943-8553-4a43-b1e3-242e025fa73e

Metal Burst/Comeuppance redirect in double battles

https://github.com/pagefaultgames/pokerogue/assets/62030095/15fba225-2747-4464-99ac-0cc87ea4251c

## How to test the changes?

Use `src/overrides.ts` to give the opponent only U-Turn/Volt Switch
Give your starting pokemon the four counter moves: Counter, Mirror Coat, Metal Burst, Comeuppance
Make sure that you are accounting for Counter/Mirror Coat only countering physical/special damage, respectively.
To test the double battle redirection with Metal Burst and Comeuppance, give the opponent a faster pokemon, use metal burst and have your ally faint the attacker before the counter goes off.

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -549,6 +553,17 @@ export default class Move implements Localizable {
     return this;
   }
 
+  /**
+   * Sets the {@linkcode MoveFlags.REDIRECT_COUNTER} flag for the calling Move
+   * @param redirectCounter The value (boolean) to set the flag to
+   * example: @see {@linkcode Moves.METAL_BURST}
+   * @returns The {@linkcode Move} that called this function
+   */
+  redirectCounter(redirectCounter?: boolean): this {
```
Tempo-anon, line 562: These are not set for counter and mirror coat?

src/phases.ts
```diff
@@ -2623,11 +2623,18 @@ export class MovePhase extends BattlePhase {
       this.targets[0] = moveTarget.value;
     }
 
+    // Check for counterattack moves to switch target
     if (this.targets.length === 1 && this.targets[0] === BattlerIndex.ATTACKER) {
       if (this.pokemon.turnData.attacksReceived.length) {
-        const attacker = this.pokemon.turnData.attacksReceived.length ? this.pokemon.scene.getPokemonById(this.pokemon.turnData.attacksReceived[0].sourceId) : null;
-        if (attacker?.isActive(true)) {
-          this.targets[0] = attacker.getBattlerIndex();
+        const attack = this.pokemon.turnData.attacksReceived[0];
```
OrangeRed, line 2629: Are you sure that counter is triggered by the 0th index move? The wiki states that "Counter will hit the last opponent that dealt physical damage to the [move user]" which to me sounds like we should be checking for index -1

src/field/pokemon.ts
```diff
@@ -3825,6 +3825,7 @@ export interface AttackMoveResult {
   damage: integer;
   critical: boolean;
   sourceId: integer;
+  attackingPosition: BattlerIndex;
```
OrangeRed, line 3828: Can you rename this to something better like "sourcePosition" or "sourceBattlerIndex"

src/phases.ts
```diff
@@ -2623,11 +2623,18 @@ export class MovePhase extends BattlePhase {
...
+        this.targets[0] = attack.attackingPosition;
+
+        // account for metal burst and comeuppance hitting remaining targets in double battles
+        // counterattack will redirect to remaining ally if original attacker faints
+        if (this.scene.currentBattle.double && this.move.getMove().hasFlag(MoveFlags.REDIRECT_COUNTER)) {
+          if (!this.scene.getEnemyField()[this.targets[0]]) {
+            this.targets[0] = this.scene.getEnemyField().find(p => p.isActive(true)).getBattlerIndex();
+          }
```
OrangeRed, line 2637: Do we need to do this? Can't we instead just redirect to position 0 if the source attack position is 1 and if the pokemon at that position is fainted.

src/data/move.ts
schmidtc1: Counter and mirror coat do not have the same functionality that metal burst and comeuppance do. Metal burst an comeuppance specifically mention the redirection for double battles in the wiki.
https://m.bulbapedia.bulbagarden.net/wiki/Counter_(move)
https://m.bulbapedia.bulbagarden.net/wiki/Mirror_Coat_(move)

src/phases.ts
schmidtc1: Line 3706 in phases.ts has this line: const lastAttack = pokemon.turnData.attacksReceived[0];
I am pretty sure it index 0 is the last attack, but I agree it should intuitively be index - 1

schmidtc1: Also, the original counterattack code used attacksReceived[0] for the last attack, and that aspect of the move has been functioning correctly I believe.

src/field/pokemon.ts
schmidtc1: I will make this change when I get back to my computer this Monday.

src/phases.ts
schmidtc1: I thought this would be easier than having to write two cases for switching attack position 1 to 0 and another case for position 0 to 1. It also takes into account only active pokemon in case both of them faint.

src/field/pokemon.ts
schmidtc1: Change made in latest commit

schmidtc1:
This PR was nuked accidentally due to lack of experience with git. I created a PR into Beta off of the same branch my original PR was based off. Merged conflicts from beta and messed up this PR that pointed to main.

PR #2475, opened 2024-06-21
[Balance] Removes most NFE from starter select

damocleas:
## What are the changes?
Evaporated the NFE mons (except Pikachu, because of its forms) from starter select.

A pain to work around for balancing, one form on the select ends up always being better before and after egg moves, with or without meta-progression tagged on them as well.
No, giving two vouchers for finishing with any of them is not an argument to keep them in giving them artificial value.

## What did change?
Deleted-
- Clefairy
- Jigglypuff
- Hitmonlee
- Hitmonchan
- Chansey
- Mr. Mime
- Jynx
- Electabuzz
- Magmar
- Snorlax
- Marill
- Sudowoodo
- Wobbuffet
- Mantine
- Roselia
- Chimecho
- Galarian Mr. Mime
From the starter select screen

### Screenshots/Videos
image

## How to test the changes?
Open the starter select. Do you see the mons? No you don't.

## Checklist
- [X] Are the changes visual?


Heppity:
I be honest this seems really silly change that doesn't really justify the reason why we can't have both as it just leaves people confused and just doesn't work NFE-wise as stuff like Gagarin Mr.Mime a good thing since you can choose your form instead on relying on RNG, this just seems change for sake of change without a good reason.

damocleas:
> I be honest this seems really silly change that doesn't really justify the reason why we can't have both as it just leaves people confused and just doesn't work NFE-wise as stuff like Gagarin Mr.Mime a good thing since you can choose your form instead on relying on RNG, this just seems change for sake of change without a good reason.

If a form is rng, it should be changed to have a different evolution method. Simple as, but not in the scope of this PR.

Heppity:
Wondering about some things mainly how shines are treated, did both baby/evol get the shiny or it was separate?

damocleas:
> Wondering about some things mainly how shines are treated, did both baby/evol get the shiny or it was separate?

They were bundled together / applied to all forms, this is unaffected :)

Tempo-anon:
Removing options for no tangible benefit is never something I'd want to support. Is Jynx worse than Smoochum 90% of the time? Yes, but how does it improve the game to remove it as an option? Going by this logic you might as well remove all the shitmons that nobody ever picks as starters like Castform too.

RNGezeus:
I stand on the side of keeping them in for a more accurate view of the contents per respective gen. 

The bonus vouchers were negligible before since it's less than +20 vouchers across 500 starters, and even more so as of the voucher tier reduction so you just grab them mid run instead. I think user wouldn't feel any impact if these vouchers got removed.

What part of balancing do you mean? Point costs? I feel they are all over as is already both with and without babies in mind. What justifies Smeargle as a 3 cost here? 

Is it candy costs given their lines have more than 1 entry point?  Like a one-time cost to enable it to adult form as well. Potentially wider gaps between baby and adult form (I.E. Magby 3, Magmar 5)

Honestly I'm not a fan of being able to pick baby-starters if their adult form does exist in the given generation for a mono-gen challenge. Relatedly this is also weird for generations that introduced evolutions for previous generations, most extreme case for Rogue being gen 6 since the gen that introduced megas actually can't make use of them. Ignore Diancie given its inaccesibility unless starter.

on a somewhat related tangent: I think starter-stage friendship evolutions are too easy to achieve.

CodeTappert:
I dont really get why this change is needed?
Also taking things away isnt a good option that players will like.

I think less people want to use the baby forms for example to get to the not-baby stages like snorlax. And i fully get that.

As previously discussed. Not having the ability to use a specific evolution kinda disencouraged to ever take like Tyrogue for example.

And at least in your post i dont see a good reason to remove them besided "A pain to work around for balancing". Like how would the balancing change with them removed? Would they get more stats to balance the hassle to use them?

Tempo-anon:
Cleaning up the PR mines. Going to close this one out for now because there was no justification given about balancing for NFEs and we should never be removing options from players without very good justification.

damocleas:
Closing because I can just open a new one that's faster.

Issue #1337, opened 2024-05-24
Unlocked Egg Moves should show up with Memory Mushroom


jaimefd:
Remved Leaf Blade (egg move) on my Zacian because I thought it was useless but later caught a Soak pokemon, sadly the move doesn't show up again with a Memory Mushroom, it probably should. (and I'd do the same with unlearned TMs)

DayKev:
Added with #2482

PR #2374, opened 2024-06-18
[Feature] Add run info

Corrade:
image
(This screenshot shows the latest UI)

## What are the changes?
Per-run stats!
- Knockouts, assists, faints, damage dealt, damage taken (blocked)
- The menu can be pulled up mid-run and is also shown after a run ends
- Persistence via session data

See the "[#2374] Run info" thread in the Discord's #dev-corner channel for more discussion.

## Why am I doing these changes?
- Provides a much-needed recap at the end of a run
- Low-risk, innocuous feature that's easily ignored

## What did change?
- Added runData class to Pokemon and PokemonData classes
- Modified Pokemon::damage() and its callers to maintain runData. This was a top-down approach that benefited from the fact that Pokemon::damage() is consistently used for all damage-related actions.
- Added sourceId field to Weather and Status to support damage attribution
- Added sourceId field to ConfusedTag, NightmareTag and PerishSongTag to support damage attribution
- Added runInfoUiHandler to the menuUiHandler
- Hooked up runInfoUiHandler fields to localisation
- Added EndRunInfoPhase to show the run info UI at the end of a run

### What's changed since the GIFs below were recorded
- Renamed from 'run stats' to 'run info' in anticipation of future non-stat extensions
- Elevated menu option from below 'stats' to below 'game settings' as I presume this UI would be more frequently opened
- Renamed stats, e.g. from kills (K) to knockouts (K.O.), as shown in the topmost screenshot here
- Flipped rows and columns in the UI to support longer stat labels
- Removed Pokemon names from the UI

### Screenshots/Videos
- Basic demo
run_stats_basic
- Run info UI appears at the end of the game
run_stats_end

#### Damage attribution
- Works with OHKO moves
run_stats_ohko
- Works with weather (credits the weather setter)
run_stats_weather
- Works with non-volatile status effects
run_stats_non_volatile_status
- Works with abilities (e.g. rough skin)
run_stats_abilities
- Works with arena tags (e.g. stealth rock)
run_stats_arena_tags
- Works with battler tags (e.g. confusion)
run_stats_battler_tags
- Works with pain split
run_stats_pain_split

#### Assists
- A Pokemon gets an assist if it was present in the fight or dealt damage (e.g. via stealth rock)
- Exception: at the start of an encounter, a lead that gets immediately swapped out does not get assist credit
run_stats_immediate_swaps_are_uncredited
- Exception: in a double battle, a Pokemon switching in does not get credit for an opponent who dies on the same turn of the switch
[...]

PR #2482, opened 2024-06-21
[Enhancement] Allow Starters to Remember Egg Moves

xsn34kzx:
## What are the changes?

Fixes pagefaultgames/pokerogue#1337; allows original party members picked at the start of a run to relearn their egg moves, if applicable, using a Memory Mushroom regardless of whether they've evolved or not.

Currently, egg moves only stay in a party member's moveset for as long as they're not replaced, which makes having egg moves hard to coordinate with how dynamic runs are and puts a lot of weight on the decision to replace those moves in situations where its unavoidable (move-related evolutions like Steenee) or accidental (misinputs or lapses in judgement). A good example of this comes from issue #1337, where you find uses for egg moves later when previously there may not have been a reason to keep it over another move or TM but now you're effectively locked out of it and any other egg move for the rest of the run.

To alleviate the fear of losing those egg moves and the opportunity to make use of those egg moves, it would only make sense to add them to the list of learnable moves provided by Memory Mushroom but only for party members you started with, maintaining the intention of egg moves being exclusive to starters while removing the need to potentially hoard egg moves.

## What did change?

`getLearnableLevelMoves` was changed such that it still gets a list of learnable level moves, but, in the case that the Pokemon being used to call the function is a starter (if `metBiome` is `-1`), then all applicable egg moves are appended to the list before filtering out any already known moves within the Pokemon's moveset. Getting the unlocked egg moves is done through a new function `getUnlockedEggMoves`, which mostly uses the logic from `starter-select-ui-handler.ts` that determines what egg moves are unlocked. Within this function, however, a new field was added for `Pokemon`, `metSpecies`.

The context for this field is due to an edge case where several Pokemon in a single evolution line can be selected at a given time (i.e., Pikachu and Pichu). In all other cases, all that's needed is to call `getRootSpeciesId` to get, for example, Charmander from Charizard. However, because Pikachu does not have any egg moves and Pichu does, it would curb the trade-off of picking Pichu for his egg moves if Pikachu can just inherit those egg moves without having to start as Pichu. `getRootSpeciesId` is still used, but it's limited in that it, when `forStarter` is set to `true`, it'll stop if the Pokemon can be used as a starter. This means that, if a starting Pichu evolves, it would then get locked out of its own egg moves since `getRootSpeciesId` can't tell the difference between a starting Pikachu and an evolved Pichu. 

This leads to the necessity of a `metSpecies` field, which essentially notes down what `Species` the Pokemon was at the time it was captured/put into the party. To accommodate the addition of this field, lines were added to the `Pokemon` constructor to either make an educated guess for loading `Pokemon` from an existing save (for example, if your run has a Charizard you started with, it'll use `getRootSpeciesId` to fill in the field with `Species.CHARMANDER`) but otherwise store the current species in `metSpecies` since it would be impossible to know if, for example, you caught a Fearow or evolved a Spearow in a run that's already progressed. For new saves however, `metSpecies` is set correctly for both caught Pokemon and starters. The application of this field is most readily shown in the Pikachu example in the Screenshots section.

`PokemonData` was changed to include the field and save it properly.

#### General Example

For a starter Charizard with `Dragon Dance` already in its moveset,
| rememberEggMoveBefore | rememberEggMoveStarter |
| --- | --- |
| Before | After |

Note that the coloring comes from the logic already used for making inherited fusion moves stand out from standard level moves.

#### Edge Cases - Differentiating Starters and Catches

Using the same species,
| rememberEggMovesExStarter | rememberEggMovesExNonStarter |
| --- | --- |
| Starter | Caught |

Note that the second Charmander's move list does not include any of the egg moves present in the first one's move list.

#### Edge Cases - Differentiating Multiple Starters in an Evolution Line

Using `metSpecies` between two starters of the same species,

| rememberEggMovesEvolvedPichu | rememberEggMovesStarterPikachu |
| --- | --- |
| Evolved Pichu | Pikachu |

Note that `metSpecies` allows Pichu to retain the ability to relearn his egg moves but keeps Pikachu from having access to those egg moves despite being a starter as well.

## How to test the changes?

To test if egg moves are visible in Memory Mushroom's list, the following override can be used to force a Memory Mushroom roll so that it can be used immediately:
```ts
export const ITEM_REWARD_OVERRIDE: Array<String> = [ "MEMORY_MUSHROOM" ];
```

`STARTER_SPECIES_OVERRIDE` can be changed to any `Species` that has egg moves already or evolves from a starter that does. 

Additionally, to test the absence of egg moves in Memory Mushroom's list, catch any wild Pokemon (which can be combined with `OPP_SPECIES_OVERRIDE` to force a `Species` you know has egg moves) and use the Memory Mushroom on them.

## Comments

A few things worth noting:
- Because `metSpecies` is a new field and will be `null` for old saves, the Pichu-Pikachu edge case will affect all evolution lines with selectable baby Pokemon starters. This means that old saves won't be able to take advantage of this change for those specific Pokemon (i.e., an existing Raichu evolved from a Pichu will not be able to relearn its egg moves). However, the use of `getRootSpeciesId` works well for standard evolutionary lines, so this change would allow, for example, existing Charizards that were in the starting party to access their egg moves.
- `metSpecies` was purely introduced to address the last edge case. Without it, balance would definitely be hurt (the aforementioned case where Pikachu is just always the better pick over Pichu if he can use Pichu's egg moves). This being its only use is a bit cumbersome, especially since it adds a data point, but I can see applications for knowing the species you met a Pokemon (#2374 immediately comes to mind) in the future.
- For simplicity, fusions do not add their own egg moves into the learnable move pool as of right now, even if the fused Pokemon is also a starter. This *can* be added if desired, but I figured I would aim to get the base Pokemon's egg moves only first before trying to include more.

## Checklist
  - ~#1143~
- [X] Are the changes visual?


Fontbane:
I actually made a PR to do this a month ago (#1143). I didn’t consider the baby edge case though, that’s an annoying one.

xsn34kzx:
> I actually made a PR to do this a month ago (#1143). I didn’t consider the baby edge case though, that’s an annoying one.

My bad, I searched the issues sections right before writing things but forgot to double check the PRs section. That edge case though is definitely throws a wrench in how much cleaner and clear cut this would've been.

damocleas (approved):

damocleas:
> This has been discussed before and as far as I know the current plan is to not though, although the possibility of making a distinct item used for regaining egg moves has been raised. This will need review by head devs at least because it is potentially a significant change to game balance.

This is simply a feature to allow starter pokemon to remember their Pokerogue* 4 egg moves, as they are the only ones that use the specified met location this uses.

PR #2494, opened 2024-06-21
[Sprite] Fixed shinies for Zubat, Golett, female Garchomp

Fontbane:
Converted to draft while awaiting further instruction from the Head Artists.
## What are the changes?

- Fixed shiny/rare Zubat, Golbat, Crobat
- Fixed inconsistencies in Garchomp's shiny
- Fixed Golett's shiny front sprite still using base replacement

- Zubat was meant to have its base replacement removed. This was not done correctly, resulting in its shiny and rare variant colors being mixed up, and also different between male and female.
- Shiny Garchomp was also very slightly different based on gender which was unintentional.
- Golurk was meant to have its base replacement removed a month ago
image
image

## What did change?

- Zubat line entirely restored to original shiny color, male rare variants fixed
- Female Garchomp now matches male's colors
- Golett shiny front sprite fixed
- Removed entries in _masterlist.json for the genies and Palossand, whose base replacements were reverted (having their entries allowed their nonexistent variants to be caught)

### Screenshots/Videos

Before:
image
Screenshot_766
Screenshot_768
Screenshot_770
Screenshot_771
Screenshot_772
Screenshot_773
Screenshot_774
After:

Screenshot_775
Screenshot_776
Screenshot_777
Screenshot_778
Screenshot_779
Screenshot_780

## How to test the changes?

- Use a save file with every variant and shiny.
- Check every affected mon in game to see if their shinies/variants are correctly colored.

## Checklist
- [X] Are the changes visual?


chaosgrimmon:
did you ensure that all icons still match?

chaosgrimmon:
also, i'm not sure why Crobat's paletteswap was replaced when it's not a BSR?
image
image

i don't believe any mobile user has complained about that one either, even.

Fontbane:
> also, i'm not sure why Crobat's paletteswap was replaced when it's not a BSR? image image
> 
> i don't believe any mobile user has complained about that one either, even.

It wasn't replaced, just deleted. It's not needed since it has the sprite.

chaosgrimmon:
yeah, but paletteswaps should be favoured since they impact the server load less. 
if that weren't the case, they wouldn't exist, since none of the artists go out of their way to make them, they just submit sprites in their submissions. sprite fixes wouldn't be necessary either, since you could just upload a variant sprite with whatever patterns you wanted

Fontbane:
> yeah, but paletteswaps should be favoured since they impact the server load less. if that weren't the case, they wouldn't exist, since none of the artists go out of their way to make them, they just submit sprites in their submissions. sprite fixes wouldn't be necessary either, since you could just upload a variant sprite with whatever patterns you wanted

I see. Would it be kept as 1 in the master list then?

chaosgrimmon:
yeah, so long as the json exists

Fontbane:
> yeah, but paletteswaps should be favoured since they impact the server load less. if that weren't the case, they wouldn't exist, since none of the artists go out of their way to make them, they just submit sprites in their submissions. sprite fixes wouldn't be necessary either, since you could just upload a variant sprite with whatever patterns you wanted

Actually I'm not sure that's a good idea because palette swaps are broken on mobile in many cases

Issue #2439, opened 2024-06-20
[BUG] Sheer Force flyout shows up with improper timing


DayKev:
**Describe the bug**
If the opposing Pokémon has Sheer Force and uses a move that would activate it, the flyout shows up when you use your move and when the opposing Pokémon uses their move. The issue might (partially?) be with this line: https://github.com/pagefaultgames/pokerogue/blob/aba3e2abc1932ec359bb7fa317568b9f9397d8eb/src/data/ability.ts#L1058 (switching this to false fixes the bug).

**To Reproduce**
To test, I started with Squirtle with Ice Beam and set the opposing Pokémon to be Magikarp with Ice Beam and Sheer Force via `overrides.ts`. When attacking with Ice Beam, Magikarp's ability flyout activates briefly before disappearing, and then once the opposing Magikarp attacks with Ice Beam it displays again. The bug only occurs if the opposing Pokémon uses a move that activates Sheer Force, thus requiring giving it a move with a secondary effect to display the bug.

**Expected behavior**
The flyout should only show up when the opposing Pokémon uses their move.

**Screenshots / Videos**
Current behavior (bugged):

https://github.com/pagefaultgames/pokerogue/assets/34855794/5db1f16d-d7d0-4eb8-9564-cca00e92e6b6

Behavior after switching line 1058 of `ability.ts` to false:

https://github.com/pagefaultgames/pokerogue/assets/34855794/1a32bb97-75d4-4779-9380-74ccccc72e10

**Device**
Chrome

DayKev:
The bug probably also happens with Serene Grace (given the shared codepaths), but I didn't test that.

okimin:
Sheer Force flyout still shows is because it also has MovePowerBoostAbAttr which also defaults to true. After testing, Serene Grace won't show at all if it is defaulted to false. But thank you. Will work with this knowledge!!

PR #2496, opened 2024-06-21
Sheer Force/Serene Grace Flyout Bugfix

okimin:
## What are the changes?
Added a new parameter to getMoveChance for showingAbility.
- Disabled showAbility in MoveEffectChanceMultiplierAbAttr during EnemyCommandPhase as it was showing inappropiately at the start of the turn due to getMoveChance being called when calculating the benefit scores.

To fix the bug in [this issue](https://github.com/pagefaultgames/pokerogue/issues/2439) .

## What did change?
Sheer Force and Serene ability flyout only happen for the opponent when after the move is hit during the MoveEffectPhase.

### Screenshots/Videos
Can refer to the bugfix video for the before video for sheer force. Serene Grace had the same behavior.

After changes
https://discord.com/channels/1125469663833370665/1252426197389021256/1253842581302022175

## How to test the changes?
Override with sheer force or serene grace on enemy side and observe when the ability flyout.

## Checklist
- [x] Are the changes visual?


src/data/ability.ts
```diff
@@ -1063,7 +1063,7 @@ export class MoveEffectChanceMultiplierAbAttr extends AbAttr {
    *             [1]: {@linkcode Moves } Move used by the ability user.
```
Tempo-anon, line 1063: Should update this for what `args[4]` is

okimin: Fixed!!

src/data/move.ts
```diff
@@ -847,9 +847,9 @@ export class MoveEffectAttr extends MoveAttr {
    * @param selfEffect {@linkcode Boolean} if move targets user.
    * @returns Move chance value.
```
DustinLin, line 848: would be good to add the new added param to the jsdoc comment above for the function

PR #2497, opened 2024-06-21
[QOL] Add yellow background glow behind starters with unlocked passives in starter select

Fontbane:
## What are the changes?

- Adds a light yellow spotlight behind starters in the starter selection menu who have their passives unlocked

As passives are more expensive than initial cost reductions, you end up having a lot of mons with enough for a cost reduction but not the passive. In passive only mode for the candy upgrade indicators, mons with their passives unlocked aren't marked, even if they have a candy upgrade available. In "All" mode, you have everything that can get a cost reduction or passive regardless of status, so it's not helpful for keeping track what can be upgrading and what you should keep saving for. An indicator like this unintrusive background would help keep track of which passives you've unlocked and give the player some satisfaction watching the board fill up with these indicators. This way you can set it to "All" mode to see if cost reductions are available without having to go and check if you're still saving for the passive.

## What did change?

- Light yellow background behind starters that had passives unlocked in the starter selection screen

### Screenshots/Videos

image
In this image, Chikorita has all upgrades completed, including passive and full cost reduction; Cyndaquil has its passive unlocked but still has cost reduction available; Totodile has undergone no upgrades.
image
This is what it looks like on a more cluttered starter
image
Large sprites such as Legendaries still have the glow clearly visible
image
What it looks like mid-hop using the Animation option for upgrade notification

## How to test the changes?

- Load a save file with at least one passive unlocked, or enough candy to unlock one
- Start a new run and look at that starter on the selection screen

## Checklist
- [X] Are the changes visual?


KimJeongSun (approved):
LGTM
If this is applied, it will be much easier to find Pokémon with unlocked passives in the future. Thank you.

Naothe:
I really need this. It'll make things a lot easier when trying to check all the passives that you have unlocked

PR #2501, opened 2024-06-22
[Enhancement] improve typing

2wheeh:
## What are the changes?

It's minor type improvement.

It's better for type safety.

## What did change?

- makes some implicit any to have proper types 
- removes some unnecessary assertions.

## Checklist
- ~~[ ] Are the changes visual?~~
  - ~~[ ] Have I provided screenshots/videos of the changes?~~

### DIFF
--- a/src/data/pokemon-evolutions.ts
+++ b/src/data/pokemon-evolutions.ts
@@ -54,6 +54,13 @@ export enum EvolutionItem {
   SYRUPY_APPLE
 }
 
+/**
+ * Pokemon Evolution tuple type consisting of:
+ * @property 0 {@linkcode Species} The species of the Pokemon.
+ * @property 1 {@linkcode integer} The level at which the Pokemon evolves.
+ */
+export type EvolutionLevel = [species: Species, level: integer];
+
 export type EvolutionConditionPredicate = (p: Pokemon) => boolean;
 export type EvolutionConditionEnforceFunc = (p: Pokemon) => void;
 
--- a/src/data/pokemon-species.ts
+++ b/src/data/pokemon-species.ts
@@ -3,7 +3,7 @@ import BattleScene, { AnySound } from "../battle-scene";
 import { Variant, variantColorCache } from "./variant";
 import { variantData } from "./variant";
 import { GrowthRate } from "./exp";
-import { SpeciesWildEvolutionDelay, pokemonEvolutions, pokemonPrevolutions } from "./pokemon-evolutions";
+import { EvolutionLevel,SpeciesWildEvolutionDelay, pokemonEvolutions, pokemonPrevolutions } from "./pokemon-evolutions";
 import { Type } from "./type";
 import { LevelMoves, pokemonFormLevelMoves, pokemonFormLevelMoves as pokemonSpeciesFormLevelMoves, pokemonSpeciesLevelMoves } from "./pokemon-level-moves";
 import { uncatchableSpecies } from "./biomes";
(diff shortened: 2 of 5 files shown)


src/data/pokemon-species.ts
```diff
@@ -528,7 +528,7 @@ export abstract class PokemonSpeciesForm {
       }
     }
 
-    const pixelColors = [];
+    const pixelColors: integer[] = [];
```
OrangeRed, line 531: Typescript doesn't have an `integer` type can you use `number` instead?

src/field/pokemon.ts
```diff
@@ -1226,7 +1226,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       const evolutionChain = this.species.getSimulatedEvolutionChain(this.level, this.hasTrainer(), this.isBoss(), this.isPlayer());
       for (let e = 0; e < evolutionChain.length; e++) {
         // TODO: Might need to pass specific form index in simulated evolution chain
-        const speciesLevelMoves = getPokemonSpeciesForm(evolutionChain[e][0] as Species, this.formIndex).getLevelMoves();
+        const speciesLevelMoves = getPokemonSpeciesForm(evolutionChain[e][0], this.formIndex).getLevelMoves();
         levelMoves.push(...speciesLevelMoves.filter(lm => (includeEvolutionMoves && !lm[0]) || ((!e || lm[0] > 1) && (e === evolutionChain.length - 1 || lm[0] <= evolutionChain[e + 1][1]))));
```
OrangeRed, line 1230: Can you destructure here so these loops are easier to read.
1) `const [ species, level ] = evolutionChain[e]`
2) `speciesLevelMoves.filter(([ level, move ]) => { ... }`

src/data/pokemon-evolutions.ts
```diff
@@ -54,6 +54,8 @@ export enum EvolutionItem {
   SYRUPY_APPLE
 }
 
+export type EvolutionLevel = [Species, integer];
+
```
OrangeRed, line 58: Can you add jsdoc comments that explain what this type means. For example
```ts
/**
 * Pokemon evolution tuple type
 * @property 0 - Pokemon Species
 * @property 1 - The level at which it evolves
 */
```

src/field/pokemon.ts
OrangeRed: This might be worth doing inside of a separate PR instead of this one.

OrangeRed (changes requested):

src/data/pokemon-species.ts
2wheeh: I found there are many `integer` in the codebase:
https://github.com/pagefaultgames/pokerogue/blob/edb4c7e4a27679d0a9716546eadc5bdf59e8ced0/src/data/pokemon-species.ts#L533

which is type alias of `number` from `phaser` package.
```ts
// nodue_modules/phaser/types/phaser.d.ts
declare type integer = number;
```

So I thought it would be better for consistency if convention is to use `integer` over `number` when the value is integer case.

WDYT?

src/field/pokemon.ts
2wheeh: will do in next PR !

src/data/pokemon-evolutions.ts
2wheeh: I added jsdoc in the way you suggest and changed it to a named tuple:

> image

I wonder what is the best way to document tuple elements with jsdoc though. It doesn't seem well supported...

OrangeRed: Yeah I tried looking it up but I couldn't find anything either. This is the best that I could come up with but if you find better by all means try it.

src/data/pokemon-species.ts
OrangeRed: Sure for the phaser stuff we can keep it consistent and use integers

OrangeRed:
@2wheeh can you resolve the conflicts so this can be merged

2wheeh:
@OrangeRed done ! sorry for the late iterate.

torranx:
@2wheeh please fix merge conflicts (again)

2wheeh:
@torranx fixed.

torranx (approved):

innerthunder (approved):
🚀

xsn34kzx (approved):

PR #2502, opened 2024-06-22
[Sprite] Reposition Glameow Animation Frames

chaosgrimmon:
## What are the changes?
Parts of Glameow's sprites, front and back, shiny and default, were being cut off, and the change that caused that was reverted.

image
image

## What did change?
As this was introduced in a prior change to the game, it was easy enough to revert by taking assets from the original commit.

#### Before
image
#### After
image
image
image

## How to test the changes?
As Glameow is a starter, any sufficiently-filled .prsv will do, including either of the test saves.

## Checklist
- [x] Are the changes visual?


Tempo-anon (approved):
🐱

PR #2503, opened 2024-06-22
[QoL] Making the transfer button show which pokemon can have items be transferred to them

Opaque02:
## What are the changes?
This PR makes it so when you go to transfer items off a pokemon, once you choose an item it will show you whether the other pokemon in your party can accept the item or if they're at max stacks.

At the moment it's a bit of a pain to try and figure out what pokemon can take what items, especially on endless with a carry mon. This change aims to make it easier to tell straight away what items your pokemon can take.

## What did change?
I changed the party-ui-handler to do a check for every pokemon when an item is selected. It sees if the item exists on that pokemon, and if so, if it's at max stacks. If so, it hides the hp related fields and shows the TM field (which has been renamed to a description field), and shows "Able" or "Not able" depending on whether the item can be transferred to a pokemon. If a pokemon is doing the transferring, their description field is blank. Lastly, if you cancel or complete a transfer the description fields go away and the health bars come back

This should be mostly good, but I'm not sure if this is all the functionality we want or if we want more? I'm open to suggestions. Also, I've put this up but may have missed some comments or console logs around - I think I got them all, but not 100% sure

### Screenshots/Videos
Below is a screenshot of me trying to transfer a lum berry from my carry mon (Velursby) to my other mons:

image

As you can see, Groudon, Garganacl and Klefki have "Not able" next to their names - this indicates that I can't transfer from Velursby to those pokemon, as they are full on lum berries. 

Here's an example of me trying to transfer black glasses from my carry mon to my other party members:

image

As you can see, all of the pokemon are able to take the black glasses

Below is a video of this in action:

https://github.com/pagefaultgames/pokerogue/assets/66582645/af16edd4-8e41-4c7a-a695-15168f0bcdc6

As you can see, I go to transfer lum berries from velursby, and groundon, garganacl and klefki are unable to accept it. Going into groudon's items you can see it's got 2 lum berries already. Trying to transfer dragon fangs shows that anyone can take it, and transferring silver powder from conkeados also shows that anyone can take it. Lastly transferring reviver seeds shows that groudon, garganacl and velursby once again can't take it, and trying to transfer to groudon shows the usual error. For all of these transfers, the description label of the pokemon doing the transfer is blank.

Below is a screenshot of the changes for double battles: 

image

Below is a screenshot of transferring to a pokemon that has fainted:

image

And here's a screenshot of trying to transfer from a pokemon that's fainted:

image

## How to test the changes?
Download this PR and go to transfer items between pokemon. Try to transfer items to pokemon when they're full (i.e. full on berries, reviver seeds, grip claws etc). You should see the description labels of the pokemon say "able" or "not able" in the transfer screen if they can't fit any more of the transfer item

## Checklist
- [x] Are the changes visual?


CodeTappert:
image
Please fix this error

Opaque02:
@CodeTappert - Sorry, missed that one in my rush to push and get this before I ran out of time. It should be fixed as of the latest commit

src/ui/party-ui-handler.ts
```diff
@@ -278,6 +278,27 @@ export default class PartyUiHandler extends MessageUiHandler {
         const pokemon = this.scene.getParty()[this.cursor];
         if (this.partyUiMode === PartyUiMode.MODIFIER_TRANSFER && !this.transferMode && option !== PartyOption.CANCEL) {
           this.startTransfer();
+
+          let newPokemon: PlayerPokemon;
+          for (let p = 0; p < this.scene.getParty().length; p++) {
+            if (p !== this.transferCursor) {
+              newPokemon = this.scene.getParty()[p];
```
torranx, line 285: since youre not reassigning a value to `let newPokemon: PlayerPokemon;`, cant this just be a `const` instead ?

src/ui/party-ui-handler.ts
```diff
@@ -278,6 +278,27 @@ export default class PartyUiHandler extends MessageUiHandler {
         const pokemon = this.scene.getParty()[this.cursor];
         if (this.partyUiMode === PartyUiMode.MODIFIER_TRANSFER && !this.transferMode && option !== PartyOption.CANCEL) {
           this.startTransfer();
+
+          let newPokemon: PlayerPokemon;
+          for (let p = 0; p < this.scene.getParty().length; p++) {
```
Tempo-anon, line 283: This giant chunk of code deserves some comments

Tempo-anon:
I do like the functionality aspect of this but I don't know if I really like the idea of just turning the name red. Would it be possible to tint the entire box red?

Opaque02:
> I do like the functionality aspect of this but I don't know if I really like the idea of just turning the name red. Would it be possible to tint the entire box red? 

@Tempo-anon - so I just did some digging into it - turns out there's a few different options for that party slot box. They're shown [here](https://github.com/pagefaultgames/pokerogue/blob/main/public/images/ui/party_slot.png), and there's another version for the [main party slot too](https://github.com/pagefaultgames/pokerogue/blob/main/public/images/ui/party_slot_main.png)

So the different frames here correlate to different situations, but in order there's the main background, the selected frame, the fainted frame (so the background when a pokemon has fainted), the selected fainted frame (which is what happens when you hover over a fainted pokemon), the swap frame (which is what happens when you go to choose something to swap off), and swap frame select (which, honestly, I have no idea when this is used?)

The issue here is that the fainted frame and fainted frame select both have a red background, like so:

image

In practice, this is what it looks like (using fainted and fainted select respectively for the backgrounds):
image
image

I feel like adding a different red colour might be confusing - perhaps we could do a different colour that we haven't used, though adding more colours may make it even more confusing?

For reference, this is what my changes of the red text look like on a fainted pokemon with the fainted frame and the fainted selected frame; I think we'll definitely have to rethink how this is shown perhaps? Feel free to @ me here or on discord to talk more about it:
image
image

src/ui/party-ui-handler.ts
Opaque02: Good point, you're right - all fixed in latest commit

Opaque02:
@Tempo-anon - here's the updated view of having the description, and a video of it in action - let me know what you think of the concept/wording/etc :)

DayKev (approved):

DayKev:
Well I didn't mean to leave a blank review like that lol, how do I remove it?

Issue #2491, opened 2024-06-21
[Feature] Accessibility: option to turn off flash from OHKO moves and perish song 


Tempo-anon:
**Is your feature request related to a problem? Please describe.**
The OHKO screen flash is unpleasant to people prone to seizures or epilepsy

**Describe the Feature**
An option to disable this feature

**Additional context**
This should be part of move animations but the effect isn't

PR #2516, opened 2024-06-22
[Accessbility] Remove screen inversion if move animations are off

Tempo-anon:
## What are the changes?
Made the screen inversion part of OHKO moves tied to move animations

closes https://github.com/pagefaultgames/pokerogue/issues/2491
The OHKO screen flash is unpleasant to people prone to seizures or epilepsy

## What did change?
Added an `if (this.scene.moveAnimations)` to check that move animations are on before inverting the scene

### Screenshots/Videos
[](https://github.com/pagefaultgames/pokerogue/assets/163687446/311c602d-a6f9-4374-a245-5d7919b179fa)

## How to test the changes?
Override no guard and a OHKO move then turn move animations on and off

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -3580,7 +3580,9 @@ export class DamagePhase extends PokemonPhase {
     super.start();
 
     if (this.damageResult === HitResult.ONE_HIT_KO) {
-      this.scene.toggleInvert(true);
+      if (this.scene.moveAnimations) {
+        this.scene.toggleInvert(true);
+      }
       this.scene.time.delayedCall(Utils.fixedInt(1000), () => {
         this.scene.toggleInvert(false);
         this.applyDamage();


torranx (approved):

src/phases.ts
```diff
@@ -3580,7 +3580,9 @@ export class DamagePhase extends PokemonPhase {
     super.start();
 
     if (this.damageResult === HitResult.ONE_HIT_KO) {
-      this.scene.toggleInvert(true);
+      if (this.scene.moveAnimations) {
+        this.scene.toggleInvert(true);
+      }
       this.scene.time.delayedCall(Utils.fixedInt(1000), () => {
         this.scene.toggleInvert(false);
```
flx-sta, line 3587: Shouldn’t this be made optional too then ? I know it might now be relevant but calling it seems rather dirty never the less

flx-sta: I didn't mean the delay itself. I mean the fact that
```ts
this.scene.toggleInvert(false)
```
is called even though it wasn't toggled to true to begin with. It basically unnecessarily called

Tempo-anon: I figured it wouldn't hurt to leave it if it doesn't affect anything. Having another `if (this.scene.moveAnimations)` check _might_ help improve performance but I don't think an unnecessary call would be worth adding in the if check

flx-sta: your call. 
✅  resolved

flx-sta (approved):

f-fsantos (approved):

PR #2526, opened 2024-06-22
[Bug] Remove grounding moves' arbitrary turn limits

innerthunder:
## What are the changes?

This change refactors the battler tag used for Smack Down and Thousand Arrows to not have arbitrary turn limits.

The grounding effects of Smack Down and Thousand Arrows currently have arbitrary turn limits of 5 and 20 turns, respectively, when they are meant to persist until the target is removed from the field.

## What did change?

- `data/battler-tags`: A new `GroundedTag` class is used for `IGNORE_FLYING` and `GROUNDED` tag types. `IGNORE_FLYING` tags (from Smack Down/Thousand Arrows) now use the `CUSTOM` lapse type, while `GROUNDED` tags still use the `TURN_END` lapse type (with a turn count set to 1).
- `data/move`: Smack Down, Thousand Arrows, and Roost's `AddBattlerTagAttr` now have no turn limit (Roost's turn limit is coded into `GroundedTag`).
- `test/moves/roost.test`: New test to check if `GROUNDED` tag lapses correctly for Roost.

### Screenshots/Videos

N/A

## How to test the changes?

`npm run test thousand_arrows`
`npm run test roost`

## Checklist


src/data/battler-tags.ts
```diff
@@ -1467,6 +1467,12 @@ export class CursedTag extends BattlerTag {
   }
 }
 
+export class GroundedTag extends BattlerTag {
```
Tempo-anon, line 1470: Can you add documentation to this `GroundedTag` and also what exactly `IGNORE_FLYING` and `GROUNDED` do and where they come from?

src/data/battler-tags.ts
```diff
@@ -1618,9 +1629,9 @@ export function getBattlerTag(tagType: BattlerTagType, turnCount: integer, sourc
   case BattlerTagType.BYPASS_SLEEP:
     return new BattlerTag(BattlerTagType.BYPASS_SLEEP, BattlerTagLapseType.TURN_END, turnCount, sourceMove);
   case BattlerTagType.IGNORE_FLYING:
-    return new BattlerTag(tagType, BattlerTagLapseType.TURN_END, turnCount, sourceMove);
+    return new GroundedTag(tagType, BattlerTagLapseType.CUSTOM, sourceMove);
   case BattlerTagType.GROUNDED:
```
Tempo-anon, line 1633: Is GROUNDED only used for roost and for one turn? I feel like the name should be something like ROOSTED or ONE_TURN_IGNORE_FLYING then.

Tempo-anon: Actually I decided that I hate `BattlerTagType.GROUNDED` because it's misleading. Can you rename it something else please?

Tempo-anon (commented):
Did you also remember to update pokemon.isGrounded()? https://github.com/pagefaultgames/pokerogue/blob/main/src/field/pokemon.ts#L1099

Tempo-anon (approved):

PR #2536, opened 2024-06-23
[Bug] Add charge status to focus punch

WingofaGriffin:
## What are the changes?

Adds the "<Pokemon> is tightening its focus!" charge status to Focus Punch instead of it behaving only as a lowered priority move.

Example in Scarlet/Violet: https://youtu.be/Lh_DQwsOiSM?t=75

Feature parity to real in game behavior.

## What did change?

Focus Punch has changed in priority to 5 with a charge status of same turn with priority -3. It now has charging text reflecting the game.

Charging animation is temporarily using the focus energy animation as a placeholder. Please let me know if this needs to be changed.

Note that the failed text right now is the same as the generic, as replacing the failed text with the "lost its focus" text seems to require a bigger code refactor than what this change currently does.

### Screenshots/Videos

Old:  focuspunchold

New: focuspunch

## How to test the changes?

Add `Moves.FOCUS_PUNCH` to `export const MOVESET_OVERRIDE: Array<Moves> = [];` and test with any starter pokemon.

## Checklist
- [x] Are the changes visual?


torranx:
the message "lost its focus" can probably be achieved by creating a new class that extends the `ChargeAttr` and overriding `getFailedText()` inside

src/data/move.ts
```diff
@@ -6365,7 +6365,8 @@ export function initMoves() {
       .attr(MovePowerMultiplierAttr, (user, target, move) => user.status
         && (user.status.effect === StatusEffect.BURN || user.status.effect === StatusEffect.POISON || user.status.effect === StatusEffect.TOXIC || user.status.effect === StatusEffect.PARALYSIS) ? 2 : 1)
       .attr(BypassBurnDamageReductionAttr),
-    new AttackMove(Moves.FOCUS_PUNCH, Type.FIGHTING, MoveCategory.PHYSICAL, 150, 100, 20, -1, -3, 3)
+    new AttackMove(Moves.FOCUS_PUNCH, Type.FIGHTING, MoveCategory.PHYSICAL, 150, 100, 20, -1, 5, 3)
+      .attr(ChargeAttr, ChargeAnim.FOCUS_PUNCH_CHARGING, "is tightening its focus!", undefined, false, true, -3)
```
torranx, line 6369: might want the "is tightening its focus!" to be localizable

torranx:
unit tests would be good to have here

src/data/move.ts
WingofaGriffin: I just followed the same format as all other charge moves have- can you elaborate?

WingofaGriffin:
> the message "lost its focus" can probably be achieved by creating a new class that extends the `ChargeAttr` and overriding `getFailedText()` inside

Agreed, though I think this is a small enough thing that it can be saved for a different PR.

> unit tests would be good to have here

Any examples you can point to for reference?

torranx:
> Agreed, though I think this is a small enough thing that it can be saved for a different PR.
> > i mean youre already passing the `ChargeAttr` so...
> > unit tests would be good to have here
> 
> Any examples you can point to for reference?
> > unit tests for moves can be found in `src/test/moves`

src/data/move.ts
torranx: using `i18next.t()` makes the text translatable. not everything in the codebase has it right now but we want all new text to be translatable-ready

torranx: ok so ive checked the code for the logic behind that charge text and it seems this would require another PR, can probably skip localizing it for now in this PR

Tempo-anon:
Closing due to adding in move headers: https://github.com/pagefaultgames/pokerogue/pull/2716

PR #2543, opened 2024-06-23
[Move] Implement Flower Shield

torranx:
## What are the changes?
- implemented [flower shield](https://bulbapedia.bulbagarden.net/wiki/Flower_Shield_(move))

- unimplemented move

## What did change?
- implemented flower shield

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/68144167/7de33be2-509b-4bcc-a005-78c2e00be661

Screenshot 2024-06-23 at 11 46 16 PM

### How to test the changes?
- give your mon Flower Shield
- have a grass type mon in field
- `$ npm run test flower_shield`

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -7238,7 +7238,7 @@ export function initMoves() {
       .attr(AddArenaTagAttr, ArenaTagType.CRAFTY_SHIELD, 1, true, true),
     new StatusMove(Moves.FLOWER_SHIELD, Type.FAIRY, -1, 10, -1, 0, 6)
       .target(MoveTarget.ALL)
-      .attr(StatChangeAttr, BattleStat.DEF, 1, false, (user, target, move) => target.getTypes().includes(Type.GRASS) && !target.getTag(BattlerTagType.FLYING) && !target.getTag(BattlerTagType.UNDERGROUND)),
+      .attr(StatChangeAttr, BattleStat.DEF, 1, false, (user, target, move) => target.getTypes().includes(Type.GRASS) && !target.isInSemiVulnerableState()),
     new StatusMove(Moves.GRASSY_TERRAIN, Type.GRASS, -1, 10, -1, 0, 6)
       .attr(TerrainChangeAttr, TerrainType.GRASSY)
       .target(MoveTarget.BOTH_SIDES),
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -1,7 +1,7 @@
 import { ChargeAnim, MoveChargeAnim, initMoveAnim, loadMoveAnimAssets } from "./battle-anims";
 import { BattleEndPhase, MovePhase, NewBattlePhase, PartyStatusCurePhase, PokemonHealPhase, StatChangePhase, SwitchSummonPhase } from "../phases";
 import { BattleStat, getBattleStatName } from "./battle-stat";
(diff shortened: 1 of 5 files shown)


src/data/move.ts
```diff
@@ -7238,7 +7238,7 @@ export function initMoves() {
       .attr(AddArenaTagAttr, ArenaTagType.CRAFTY_SHIELD, 1, true, true),
     new StatusMove(Moves.FLOWER_SHIELD, Type.FAIRY, -1, 10, -1, 0, 6)
       .target(MoveTarget.ALL)
-      .unimplemented(),
+      .attr(StatChangeAttr, BattleStat.DEF, 1, false, (user, target, move) => target.getTypes().includes(Type.GRASS) && !target.getTag(BattlerTagType.FLYING) && !target.getTag(BattlerTagType.UNDERGROUND)),
```
Tempo-anon, line 7241: I think pokemon using dive or in the shadow realm should also be excluded

src/field/pokemon.ts
```diff
@@ -286,6 +286,26 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     return this.isAllowedInBattle() && !!this.scene && (!onField || this.isOnField());
   }
 
+  /**
+   * Checks if the Pokémon is in a semi-invulnerable state due to certain tags.
+   *
+   * @returns `true` if the Pokémon is in a semi-invulnerable state, otherwise `false`.
+   */
+  isInSemiVulnerableState(): boolean {
```
innerthunder, line 294: A potentially cleaner alternative to this function:

```ts
!!this.getTag(HideSpriteTag);
```

It doesn't include the check that `this.scene` is defined, but `getTag()` has a null check for `this.summonData` that might accomplish the same thing.

torranx: i agree this is cleaner, but would this be intuitive to use if it is not wrapped inside a function?

another option would be to use named import instead just to be more descriptive ie` import HideSpriteTag as SemiInvulnerableTag`

innerthunder: I think with a named import it would be clear enough.

torranx: done, thanks!

innerthunder (approved):

src/data/move.ts
```diff
@@ -1,7 +1,7 @@
 import { ChargeAnim, MoveChargeAnim, initMoveAnim, loadMoveAnimAssets } from "./battle-anims";
 import { BattleEndPhase, MovePhase, NewBattlePhase, PartyStatusCurePhase, PokemonHealPhase, StatChangePhase, SwitchSummonPhase } from "../phases";
 import { BattleStat, getBattleStatName } from "./battle-stat";
-import { EncoreTag } from "./battler-tags";
+import { EncoreTag, HideSpriteTag as SemiInvulnerableTag } from "./battler-tags";
```
Tempo-anon, line 4: Man... we really need to have some consistency in naming. `HideSpriteTag` is also imported in `phases.ts` but there it's been renamed as `HiddenTag`

torranx: I've renamed all instances of `HideSpriteTag` to `SemiInvulnerableTag` instead which makes a lot more sense

Tempo-anon (approved):

PR #2544, opened 2024-06-23
[Localization] Fixed missing punctuation when being freed from trapping moves

Fontbane:
## What are the changes?

- Adds punctuation to the message that shows after being freed from wrap

- It was missing punctuation and that looked weird

## What did change?

- Added an exclamation point to the end of battlerTagsTrappedOnRemove where it was missing

### Screenshots/Videos

image

## How to test the changes?

- Use a trapping move like Wrap or Whirlpool
- Wait for the opponent to be freed

## Checklist
- [X] Are the changes visual?


CodeTappert:
Change the . in german to a ! too please

Tempo-anon (approved):

PR #2545, opened 2024-06-23
[Feature] Fusion Names Overhaul

amatsune:
Following up this suggestion:
https://github.com/pagefaultgames/pokerogue/issues/2383
This is an early attempt to make it work

## What are the changes?
Fusion names should change a little.

Following up this suggestion I made.
https://github.com/pagefaultgames/pokerogue/issues/2383
 Now it will be possible to know the pokémon fused by the fusion name alone.

## What did change?

Created new localization files with affixes for each pokémon.
Edited the /src/interface/locales.ts to accomodate the new files
Edited the /src/data/pokemon-species.ts to make use of the new affixes replacing the original fusion name function

## How to test the changes?
Some tweaking of the pokemon-species.ts needs to happen before its functional. Was suggested to make a PR so it could be looked over.

## Checklist
- [x] Are the changes visual?


src/locales/de/pokemon-fusion-affixes.ts
```diff
@@ -0,0 +1,4332 @@
...
+    fusionPrefix: "Schig",
+    fusionSuffix: "iggy",
+  },
+  wartortle: {
+    fusionPrefix: "Schill",
+    fusionSuffix: "illok",
+  },
+  blastoise: {
+    fusionPrefix: "Tur",
+    fusionSuffix: "urtok",
```
CodeTappert, line 38:
```suggestion
    fusionSuffix: "tok",
```

src/locales/de/pokemon-fusion-affixes.ts
```diff
@@ -0,0 +1,4332 @@
...
+    fusionPrefix: "Raup",
+    fusionSuffix: "upy",
+  },
+  metapod: {
+    fusionPrefix: "Saf",
+    fusionSuffix: "afcon",
+  },
+  butterfree: {
+    fusionPrefix: "Smet",
+    fusionSuffix: "ettbo",
```
CodeTappert, line 50:
```suggestion
    fusionSuffix: "bo",
```

src/locales/zh_TW/pokemon-fusion-affixes.ts
Mewtwo2387, line 1: Looks rather inconsistant and most of these doesn't really work, might be better to keep zh_CN and zh_TW as English for now and wait till our localization team discuss the correct prefix and suffix.

Mewtwo2387 (changes requested):

src/locales/zh_TW/pokemon-fusion-affixes.ts
amatsune: What do you mean by inconsistent?

If it's about the code working: it doesn't…

If it's about the naming convention, can you point out what you mean by it?

Mewtwo2387: Take some 3 word names for example - There are randomly some prefixes that uses the entire 3 words while some uses 1 like the rest, despite being the same evolution line with the same way of naming.

Also, this PR is supposed to make fusion names more clear, but this set of prefix/suffix names have a completely opposite effect as a lot of names only took 1 word out of 4.

Mewtwo2387: Take the squirtle line for example:
```
  squirtle: {
    fusionPrefix: "傑尼龜",
    fusionSuffix: "龜",
  },
  wartortle: {
    fusionPrefix: "卡",
    fusionSuffix: "龜",
  },
  blastoise: {
    fusionPrefix: "水",
    fusionSuffix: "龜",
  },
  ```
  This is pretty much like
  ```
    squirtle: {
    fusionPrefix: "squirtle",
    fusionSuffix: "tle",
  },
  wartortle: {
    fusionPrefix: "wa",
    fusionSuffix: "tle",
  },
  blastoise: {
    fusionPrefix: "bla",
    fusionSuffix: "ise",
  },
```
Out of these 3, squirtle is the only one that uses the full name as the prefix, while the rest didn't even use the entire  name for prefix+suffix

Same problem with most of the other pokemon.

amatsune: Oh, yes for ideographic languages I honestly don't have a good rule for the division. I don't have the knowledge to make a rule that attends to the intricacies of a language I don't know the first thing about. I don't even know how it works currently with those languages, I assume it splits them in half?

In any case, if you have a proposed rule for tackling these localizations, I'm all ears. Otherwise, I'll take your suggestion and leave the English set for the time being.

> Take some 3 word names for example - There are randomly some prefixes that uses the entire 3 words while some uses 1 like the rest, despite being the same evolution line with the same way of naming.

This was something I tried to address, but my script kept coming up with blanks due to how the comparison worked. If the word shared characters with another word, then it would copy give the "prefix" and nothing else
If it didn't it would just leave it blank, for those I set it to output the full word instead.
I believe a solution that keeps to the principles I've set for Latin languages just isn't suited to non-syllabic structures.

Despite this, I think the possibility to manually tweak the Prefixes and Suffixes proposed by this PR should be a step forward in making sure we have the power to fix this manually on a case-by-case basis where necessary, and should there be a good proposed algorithm for a general rule, I'd love to use it.

amatsune: > Take the squirtle line for example:
> 
> ```
>   squirtle: {
>     fusionPrefix: "傑尼龜",
>     fusionSuffix: "龜",
>   },
>   wartortle: {
>     fusionPrefix: "卡",
>     fusionSuffix: "龜",
>   },
>   blastoise: {
>     fusionPrefix: "水",
>     fusionSuffix: "龜",
>   },
> ```
> 
> This is pretty much like
> 
> ```
>   squirtle: {
>   fusionPrefix: "squirtle",
>   fusionSuffix: "tle",
> },
> wartortle: {
>   fusionPrefix: "wa",
>   fusionSuffix: "tle",
> },
> blastoise: {
>   fusionPrefix: "bla",
>   fusionSuffix: "ise",
> },
> ```
> 
> Out of these 3, squirtle is the only one that uses the full name as the prefix, while the rest didn't even use the entire name for prefix+suffix
> 
> Same problem with most of the other pokemon.

As to using the entire name:


That's not the goal. Prefix and suffix are meant to include just the part of the name that isn't shared with any other 'mon (and in latin languages extend that to the next consonant/vowel, for phonetic reasons)
This way if you fuse bulbasaur with itself:

  bulbasaur: {
    fusionPrefix: "Bul",
    fusionSuffix: "asaur",
  },

you'll get Bulasaur.

That is on purpose. When playing with fusions, this indicates immediately that something isn't quite like it should, even though for all intents and purposes it will look like a normal bulbasaur.
By following these rules it should always be possible to tell the pokémon involved in the fusion.

Mewtwo2387: All these 3 uses 龜 as the suffix. In chinese, the last word of a lot of evolution lines remains the same. So you'll never be able to figure out which pokemon is involved if you just use the last world - hence why I mentioned having to use more

src/field/pokemon.ts
```diff
@@ -300,7 +300,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       this.name = this.species.getName(this.formIndex);
       return;
     }
-    this.name = getFusedSpeciesName(this.species.getName(this.formIndex), this.fusionSpecies.getName(this.fusionFormIndex));
+    this.name = getFusedSpeciesName(this.species.getName(this.species), this.fusionSpecies.getName(Species));
```
CodeTappert, line 303:
```suggestion
    this.name = getFusedSpeciesName(this.species, this.fusionSpecies);
```

src/data/pokemon-species.ts
```diff
@@ -51,71 +51,8 @@ export function getPokemonSpeciesForm(species: Species, formIndex: integer): Pok
 }
 
 export function getFusedSpeciesName(speciesAName: string, speciesBName: string): string {
```
CodeTappert, line 53:
```suggestion
export function getFusedSpeciesName(speciesA: PokemonSpecies, speciesB: PokemonSpecies): string {
```

src/data/pokemon-species.ts
```diff
@@ -51,71 +51,8 @@ export function getPokemonSpeciesForm(species: Species, formIndex: integer): Pok
...
-
-  if (splitNameA.length > 1) {
-    fragA = `${splitNameA.slice(0, splitNameA.length - 1).join(" ")} ${fragA}`;
-  }
-
-  fragB = `${fragB.slice(0, 1).toLowerCase()}${fragB.slice(1)}`;
-
-  return `${speciesAPrefix || speciesBPrefix}${fragA}${fragB}${speciesBSuffix || speciesASuffix}`;
+//I WAS HERE, DON'T KNOW HOW TO MAKE THIS WORK
+  return `${speciesAPrefix}${speciesBSuffix}`;
```
CodeTappert, line 55:
```suggestion
// Basically the only thing missing is now to get the non local species name in lowercase
return i18next.t(fusionAffixes:${speciesA}.fusionPrefix)+i18next.t(fusionAffixes:${speciesB}.fusionSuffix)

```

src/data/pokemon-species.ts
```diff
@@ -50,72 +50,11 @@ export function getPokemonSpeciesForm(species: Species, formIndex: integer): Pok
   return retSpecies;
 }
 
-export function getFusedSpeciesName(speciesAName: string, speciesBName: string): string {
-  const fragAPattern = /([a-z]{2}.*?[aeiou(?:y$)\-\']+)(.*?)$/i;
-  const fragBPattern = /([a-z]{2}.*?[aeiou(?:y$)\-\'])(.*?)$/i;
+export function getFusedSpeciesName(speciesA: PokemonSpecies, speciesB: PokemonSpecies): string {
+//I WAS HERE, DON'T KNOW HOW TO MAKE THIS WORK
+// Basically the only thing missing is now to get the non local species name in lowercase
+return i18next.t(fusionAffixes:${speciesA}.fusionPrefix)+i18next.t(fusionAffixes:${speciesB}.fusionSuffix)
```
CodeTappert, line 56:
```suggestion
  const prefix = i18next.t(`fusionAffixes:${Species[speciesA.speciesId].toLowerCase()}.fusionPrefix`);
  const suffix = i18next.t(`fusionAffixes:${Species[speciesB.speciesId].toLowerCase()}.fusionSuffix`);

  return `${prefix}${suffix}`;
}
```

amatsune:
Screenshot 2024-06-24 185519
Screenshot 2024-06-24 191024

jaimefd:
I know this is a lot of work and by no means I want to disrespect it but... is this really needed? Short names have their charm and I feel like this method creates very long and unnatural looking names. Also, we do not need to be able to tell the fusion immediately with the name, you can scroll over it.

I may sound immature here but the laughs and funny moments I got with all of these fusion names will really be missed, this new ones seem too serious. Maybe a toggle for players to choose which fusion naming method to use? I really love the current one

amatsune:
Besides the reasons stated in my initial suggestion, which goes over in length over many points, it turns out the current system doesn't handle other languages quite well, causing some names just appear broken.

Filipe Sant'Anna.
________________________________

amatsune:
image
Now handles CN names in a semantically correct order

amatsune:
> You seem to already solved it, but i was kinda curious.
> 
> I tryed some things in python and for symbolic languages found that `jieba` has a good solution.
> 
> Maybe you can analyze jiebe for that matter if some problems a rise down the line.

I included the python code I used in the pr, if you want to take a look. It's the long a commented out portion at the end that I used to deal with symbolic languages. It's not perfect by any means, so most of the CN affixes were adjusted manually to optimize meaning. That means I'm not sure if they are optimally following the original rule of keeping the species name always unique, but: the "double inversion" thing I mentioned makes sure that the suffixes (which is the part that tends to repeat more often through evolution lines) should at least be unique to evolution lines.

With that and the sprite it should be possible to know the fused species.

40chyan (approved):

src/data/pokemon-species.ts
```diff
@@ -50,72 +50,37 @@ export function getPokemonSpeciesForm(species: Species, formIndex: integer): Pok
...
-  const fragAPattern = /([a-z]{2}.*?[aeiou(?:y$)\-\']+)(.*?)$/i;
-  const fragBPattern = /([a-z]{2}.*?[aeiou(?:y$)\-\'])(.*?)$/i;
+export function getFusedSpeciesName(speciesA: PokemonSpecies, speciesB: PokemonSpecies): string {
+  let firstKey = "fusionPrefix";
+  let secondKey = "fusionSuffix";
 
-  const [ speciesAPrefixMatch, speciesBPrefixMatch ] = [ speciesAName, speciesBName ].map(n => /^(?:[^ ]+) /.exec(n));
-  const [ speciesAPrefix, speciesBPrefix ] = [ speciesAPrefixMatch, speciesBPrefixMatch ].map(m => m ? m[0] : "");
+  const shouldReverse = i18next.t("fusionAffixes:shouldReverse");
+  console.log(`shouldReverse: ${shouldReverse}`); // logs the value of shouldReverse
```
CodeTappert, line 58: Cant this be removed before merging to main?

src/data/pokemon-species.ts
```diff
@@ -50,72 +50,37 @@ export function getPokemonSpeciesForm(species: Species, formIndex: integer): Pok
...
-    fragB = splitNameB[splitNameB.length - 1];
+  if (i18next.t("fusionAffixes:shouldReverse") === "true") {
+    prefix = desKey;
+    suffix = radKey;
   }
 
-  if (splitNameA.length > 1) {
-    fragA = `${splitNameA.slice(0, splitNameA.length - 1).join(" ")} ${fragA}`;
-  }
+  console.log(`prefix: ${prefix}, suffix: ${suffix}`); // logs the values of prefix and suffix
```
CodeTappert, line 79: This also

src/locales/fusion-affixes-generator.py
```diff
@@ -0,0 +1,230 @@
+#This script produces the list of affixes (prefix and suffix) used to create the fusion names.
```
CodeTappert, line 1: I dont think this should be here. 
Maybe create a "tools" Folder in the locales folder or something

src/data/pokemon-species.ts
```diff
@@ -54,32 +54,19 @@ export function getFusedSpeciesName(speciesA: PokemonSpecies, speciesB: PokemonS
   let firstKey = "fusionPrefix";
   let secondKey = "fusionSuffix";
 
-  const shouldReverse = i18next.t("fusionAffixes:shouldReverse");
```
CodeTappert: Why did you also remove this line?

CodeTappert (approved):

40chyan:
can you solve the conflicts so we can ask a dev to approve this?

40chyan (approved):

innerthunder:
closing in favor of https://github.com/pagefaultgames/pokerogue/pull/3262

Issue #2523, opened 2024-06-22
[BUG] Lures when stacked


C-ater:
**Describe the bug**
If you stack lures, the stacked lure will not double your odds of getting a double battle

**To Reproduce**

1. Grab a Super Lure
2. Grab any other lure before the SL gets to 5 waves remaining
3. Grab another regular lure when the SL has 5 waves remaining
4. Sometimes see that you will get a single battle

**Expected behavior**
The stack of lures should act as two lures, and along with the other lure, guarantee double battles

**Screenshots / Videos**
![Screenshot_20240622_160725_Chrome 1](https://github.com/pagefaultgames/pokerogue/assets/173323483/31fa0f4f-809b-45d6-8c88-27f3a4443f0a)
There is one lure with 4 waves remaining, and 2 lures with 1 wave remaining (A super lure from 9 waves ago, and a regular lure 4 waves ago)

**Device**
Phone (Samsung S22)

**Additional context**
![image](https://github.com/pagefaultgames/pokerogue/assets/173323483/7e705e5b-298a-488e-a4cb-617c11c75af1)
Usually, 3 lures should be guaranteeing a double battle, but when stacked a single battle occurs signifying that the odds are affected in some way.

Drednaw's ability is Strong Jaw

When you grab a lure that has the same starting value as a remaining value of a lure you currently have (Grabbing a Super Lure when your Max Lure has 10 Waves left or a Lure when the Super Lure has 5 waves remaining), it will stack with the prexisting Lure, setting the bottom left number as 2 signifiying two of that type of lure

Issue #1912, opened 2024-06-07
[BUG] Multiple lures with same remaining duration do not stack


anbecker:
**Describe the bug**
If you have multiple lures with the same remaining duration, their icons will stack, and they will not both be counted toward increased double battle encounter rate.

**To Reproduce**
Example: I have four super/max lures running. One of them has 5 battles remaining on it. I collect a regular lure from the end of a battle. Instead of adding a regular lure icon, the super lure with a '5' next to it increments to a counter of quantity 2 (the white number). I enter a boss battle. Despite having 5 lures running, it is only counted as four, so it is possible for it to be a single battle at a 50% chance.

**Expected behavior**
The new double battle lure should be listed separately from the pending ones, and should be counted toward increased doble battle encounter rate.

**Screenshots / Videos**
<img width="652" alt="Screenshot 2024-06-07 at 8 35 12 AM" src="https://github.com/pagefaultgames/pokerogue/assets/16811393/cf53b410-5ccd-4c5a-a674-c308816d50e3">

**Device**
Arc Browser (Chromium based), macOS. This info is irrelevant, but hey.

**Additional context**
5 lures should guarantee a double battle for bosses - if one is not counted, it hurts people's chances at a legendary boss encounter.

kleenxfeu:
I second this, same issue on windows/edge

CodeTappert:
l

CodeTappert:
Closed as solved by the linked pr

PR #2550, opened 2024-06-23
[Bug] Prevent lures from stacking different tiers

EmberCM:
## What are the changes?
Checks the ModifierType#id when attempting to stack Lures

Fix multiple issues #1912 #2523

## What did change?
Added a check for ModifierType#id to the match(Modifier) method inside the DoubleBattleChanceBoosterModifier class

### Screenshots/Videos
Before vs After
before
after

## How to test the changes?
```ts
export const ITEM_REWARD_OVERRIDE: Array<String> = ["LURE", "SUPER_LURE", "MAX_LURE"];
```

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/modifier/modifier.ts
+++ b/src/modifier/modifier.ts
@@ -328,7 +328,8 @@ export class DoubleBattleChanceBoosterModifier extends LapsingPersistentModifier
 
   match(modifier: Modifier): boolean {
     if (modifier instanceof DoubleBattleChanceBoosterModifier) {
-      return (modifier as DoubleBattleChanceBoosterModifier).battlesLeft === this.battlesLeft;
+      // Check type id to not match different tiers of lures
+      return modifier.type.id === this.type.id && modifier.battlesLeft === this.battlesLeft;
     }
     return false;
   }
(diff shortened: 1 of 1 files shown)


Tempo-anon (approved):

bennybroseph:
What about existing saves?

EmberCM:
> What about existing saves?

Fixed in the new commit. The chance will be divided by the amount of lures in the stack

src/modifier/modifier.ts
```diff
@@ -341,9 +342,19 @@ export class DoubleBattleChanceBoosterModifier extends LapsingPersistentModifier
...
   }
 
+  /**
+   * Modifies the chance of a double battle occurring
+   * @param args A single element array containing the double battle chance as a NumberHolder
+   * @returns {boolean} Returns true if the modifier was applied
+   */
   apply(args: any[]): boolean {
+    // Use Math.max to ensure no dividing by 0
+    const chanceMultiplier = 2 * Math.max(1, this.getStackCount());
```
Tempo-anon, line 352: This will be unused legacy code in the future right? Since line 332 will stop lures from stacking in the first place. Personally I'm okay with just leaving legacy lures stacked and just fix new lures going forward

Tempo-anon (approved):

src/modifier/modifier.ts
EmberCM: Sounds good to me, let them suffer
elmo

Tempo-anon (approved):
Could have also just modified the apply so stacked lures will get counted multiple times but w/e this is probably cleaner

Issue #2190, opened 2024-06-13
[BUG] Soft lock after roar on double battle. Video & console logs provided.


HighMans:
**Describe the bug**
Soft lock after roar on double battle. Video & console logs provided.

**To Reproduce**
Session and data can be provided to a dev on discord. View video for steps.

**Expected behavior**
No soft locking.

**Screenshots / Videos**

https://github.com/pagefaultgames/pokerogue/assets/42877729/11c82db2-d443-4f1c-9ddf-b51ca6d1799f

**Device**
PC

**Additional context**

Graymare:
This also happens with Dragon Tail, I would surmise it also affects any other "forced removal" attack.

HighMans:
> Can't play the video, maybe reupload

Nothing wrong on my end, try downloading it?

DustinLin:
https://github.com/pagefaultgames/pokerogue/assets/39450497/3d2e3836-e38d-4989-a892-aace07172055

able to reproduce:
bug is trying to run after roar-ing an opponent in a doubles battle

DustinLin:
looking into this...

Graymare:
To clarify, this bug also happens in double battles whenever Dragon Tail is used and does **not** KO, without any attempt to Run from the battle.  If Dragon Tail does KO, the game works fine.

To Reproduce:
Have a double battle with two pokemon on each side.
Dragon Tail one opponent such that it does not faint from Dragon Tail's damage.
Game softlocks.

DustinLin:
is this still an issue with the changes from #2553 ?

innerthunder:
fixed by https://github.com/pagefaultgames/pokerogue/pull/3346

PR #2553, opened 2024-06-23
[Bug] fix softlock/runtime error after roar in doubles

DustinLin:
## What are the changes?

Added undefined/null checks to 2 places where `this.scene` is traced back to from the error

Solves issue #2190 

## What did change?

The bug as discussed in issue #2190 is running away from a battle when one of the opponent pokemon's fainted. This bug doesn't happen anymore and transitions to the next battle.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/39450497/8b08c31b-a315-473f-9cc3-0009466932bb

## How to test the changes?

Reproduce using roar on one pokemon, and trying to run the next turn.

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1070,7 +1070,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
    * in effect, and both passive and non-passive. This is one of the two primary ways to check
    * whether a pokemon has a particular ability.
    * @param {AbAttr} attrType The ability attribute to check for
-   * @param {boolean} canApply If false, it doesn't check whether the abiltiy is currently active
+   * @param {boolean} canApply If false, it doesn't check whether the ability is currently active
    * @param {boolean} ignoreOverride If true, it ignores ability changing effects
    * @returns {boolean} Whether an ability with that attribute is present and active
    */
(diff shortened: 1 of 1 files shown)


src/field/pokemon.ts
```diff
@@ -1082,8 +1091,14 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
 
+  /**
+   * Gets the tera-formed type of the pokemon, or UNKNOWN if not present
+   * @returns the {@linkcode Type}
+   */
   getTeraType(): Type {
-    const teraModifier = this.scene.findModifier(m => m instanceof TerastallizeModifier
+    // this.scene can be undefined for a fainted mon in doubles
+    // use nullish coalescing to auto-set to false to avoid runtime errors
+    const teraModifier = false ?? this.scene.findModifier(m => m instanceof TerastallizeModifier
```
Tempo-anon, line 1101: Why `false ??`

DustinLin: Thanks for pointing this out, my code logic was messed up and fix it

src/field/pokemon.ts
```diff
@@ -1082,13 +1091,21 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   */
   getTeraType(): Type {
-    const teraModifier = this.scene.findModifier(m => m instanceof TerastallizeModifier
-      && m.pokemonId === this.id && !!m.getBattlesLeft(), this.isPlayer()) as TerastallizeModifier;
-    if (teraModifier) {
-      return teraModifier.teraType;
+    // this.scene can be undefined for a fainted mon in doubles
+    if (this.scene !== undefined) {
+      const teraModifier = false ?? this.scene.findModifier(m => m instanceof TerastallizeModifier
+        && m.pokemonId === this.id && !!m.getBattlesLeft(), this.isPlayer()) as TerastallizeModifier;
```
Tempo-anon, line 1102: `const x = false ?? y` is equivalent to just doing `const x = y` since the `??` operator on anything not null or undefined will just result in the thing to the right of it

src/field/pokemon.ts
```diff
@@ -1082,13 +1091,21 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      // return teraType
+      if (teraModifier) {
+        return teraModifier.teraType;
+      }
     }
-
+    // if scene is undefined, or if teraModifier is considered false, then return unknown type
```
Tempo-anon, line 1108: Seems weird that unknown type is the default if tera modifier is false

src/field/pokemon.ts
DustinLin: yeah I forgot to remove the operator, thanks

src/field/pokemon.ts
DustinLin: I'm guessing the motivation is that if there isn't any new additional type to add, then the logic just returns `UNKNOWN` (to conform with the return type of the function?).


This PR just moves the code around so it's introducing anything new

Tempo-anon (approved):

PR #2559, opened 2024-06-24
[Move] Implement Substitute

innerthunder:
## What are the changes?

This will implement the move Substitute

Substitute is (N)

### Current Changes:
- `data/battler-tags`: Added new `SubstituteTag` which represents the [Substitute Doll effect](https://bulbapedia.bulbagarden.net/wiki/Substitute_(doll)#Effect) and stores the doll's HP. This tag uses the new type `SUBSTITUTE` and several lapse types (including a new `HIT` lapse type) to trigger messages and animations.
- `data/move`: 
    - Added a new attribute `AddSubstituteAttr` for Substitute's effect.
    - Added a function that checks if a move can hit targets behind Substitutes as well as a new move flag that enables certain moves to bypass Substitute.
    - Added a new attribute `RemoveAllSubstitutesAttr` to Tidy Up that removes all Substitutes from the field.
- `field/pokemon`:
    -  `Pokemon.apply()` is updated so that if the `Pokemon` has a `SubstituteTag`:
        - Attack moves ([unless unaffected](https://bulbapedia.bulbagarden.net/wiki/Substitute_(move)#Unaffected_moves)) have their damage redirected to the `SubstituteTag`, removing the tag if it brings the tag's HP to 0 (or less)
        - Status moves (unless unaffected or self-targeting) will fail.
    - Added methods to manage the Pokemon's sprite and Substitute sprite (if it exists)
- `pokemon-anim-phase` and `enums/pokemon-anim-type`: Implemented new animations for Substitute, including on-add, pre- and post-move, and on-remove animations.
- `phases`: Various changes to help manage the sprite of a Pokemon's Substitute when the Pokemon switches, faints, or uses Baton Pass.
- `test/battlerTags/substitute.test`: New unit test suite to test the functionality of `SubstituteTag`.
- `test/moves/substitute.test`: New integration test suite to test the effects of Substitute in battle.
- `test/moves/tidy_up.test`: Updated "clears substitutes" test to use Substitute's new implementation

### Animations

With Substitute comes a new suite of in-battle animations (and a new phase to implement them)! The videos below show all of the new animations surrounding the Substitute doll in action during a double battle.

#### Videos

**Adding Substitutes:**

https://github.com/user-attachments/assets/7394e1ad-e066-4a79-a215-00c04f30ec9d

When a Pokemon uses Substitute, a new sprite is generated (with its reference attached to the respective `SubstituteTag`), and the Pokemon is moved "out of focus," becoming semi-transparent.

**Switching Pokemon with Substitutes:**

https://github.com/user-attachments/assets/edd2be80-a024-44f4-8538-bd96a7c2c2d5

- When a Pokemon with an active Substitute switches normally, the Substitute's sprite despawns (after a short animation) before the switched-in Pokemon is summoned.
- When a Pokemon with an active Substitute uses Baton Pass, the Substitute's sprite remains on the field and becomes visible after the move resolves.

**Hitting and Removing Substitutes:**

https://github.com/user-attachments/assets/884f6d08-0172-4815-9ac8-c9fb520546be

- When a Pokemon with a Substitute uses a move, it moves into focus, uses the move, then moves back out of focus.
- When a Pokemon's Substitute takes more than its remaining HP in damage, a fading animation despawns the Substitute's sprite and brings the Pokemon back "into focus."

**KOing Pokemon with Substitutes:**

https://github.com/user-attachments/assets/602954cd-b456-43c0-8594-dd3fb51ea2af

When a Pokemon faints while it has a Substitute active, the Substitute's sprite is also removed from the field.

**Loading in Pokemon with Substitutes:**

https://github.com/user-attachments/assets/2ab94298-7ac5-47c8-86d5-180899268e3c

When the player loads into a wave where a Pokemon started with a Substitute, the on-add animation is replayed to regenerate the Substitute sprite.

## How to test the changes?

`npm run test substitute`
`npm run test tidy_up` (see the "clears substitutes" test)

## Checklist
- [x] Are the changes visual?


src/data/battler-tags.ts
```diff
@@ -1529,6 +1530,38 @@ export class IceFaceTag extends BattlerTag {
   }
 }
 
+export class SubstituteTag extends BattlerTag {
+  public substituteHp: integer;
```
Tempo-anon, line 1534: nit: we should be moving away from `integer` and just use `number` since they're functionally identical

src/data/battler-tags.ts
```diff
@@ -26,7 +26,8 @@ export enum BattlerTagLapseType {
   AFTER_MOVE,
   MOVE_EFFECT,
   TURN_END,
-  CUSTOM
+  CUSTOM,
+  HIT
```
Tempo-anon, line 30: Hmm, would this also work for something like disguise, ice face, or gulp missile?

src/data/battler-tags.ts
```diff
@@ -1529,6 +1530,38 @@ export class IceFaceTag extends BattlerTag {
...
+
+  onRemove(pokemon: Pokemon): void {
+    // Should revert Pokemon's form to its original
+    if (this.substituteHp <= 0) {
+      pokemon.scene.queueMessage(getPokemonMessage(pokemon, "'s substitute faded!"));
+    }
+  }
+
+  lapse(pokemon: Pokemon, lapseType: BattlerTagLapseType): boolean {
+    const ret = lapseType !== BattlerTagLapseType.CUSTOM || super.lapse(pokemon, lapseType);
```
Tempo-anon, line 1555: nit: `const ret = (lapseType !== BattlerTagLapseType.CUSTOM || super.lapse(pokemon, lapseType));`

src/data/move.ts
```diff
@@ -1222,6 +1229,53 @@ export class HalfSacrificialAttr extends MoveEffectAttr {
...
+   * @returns true if the attribute successfully applies, false otherwise
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    if (!super.apply(user, target, move, args) || !!user.getTag(BattlerTagType.SUBSTITUTE)) {
+      return false;
+    }
+
+    const hpCost = Math.ceil(user.getMaxHp() / 4);
+    if (user.hp > hpCost) {
+      user.damageAndUpdate(hpCost, HitResult.OTHER, false, true, true);
```
Tempo-anon, line 1258: Just want to double check, does this `damageAndUpdate` work through things like magic guard?

src/data/move.ts
```diff
@@ -1222,6 +1229,53 @@ export class HalfSacrificialAttr extends MoveEffectAttr {
...
+      return true;
+    }
+    return false;
+  }
+
+  getUserBenefitScore(user: Pokemon, target: Pokemon, move: Move): number {
+    if (user.isBoss()) {
+      return -10;
+    }
+    return Math.ceil((user.getHpRatio() - 0.25) * 4/3 * 10 - 5);
```
Tempo-anon, line 1269: Did you get these numbers from somewhere or did you choose it arbitrarily? Some doc on this function to describe the calculation for now would be good

src/field/pokemon.ts
```diff
@@ -1948,6 +1948,13 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
 
         const oneHitKo = result === HitResult.ONE_HIT_KO;
         if (damage.value) {
+          this.lapseTags(BattlerTagLapseType.HIT);
+          const subTag = this.getTag(SubstituteTag) as SubstituteTag;
+
+          if (subTag && !move.bypassesSubstitute(source)) { // TODO: Make this also check if the source has Infiltrator
+            subTag.substituteHp -= damage.value;
+            damage.value = 0;
```
Tempo-anon, line 1956: Is there a message for when a substitute gets hit normally? Also, how does this work with this game's version of spaghetti multi-hit moves?

Tempo-anon (commented):
Going to just assume that you got all the bypassSubstitute moves right based off the table on bulbapedia

src/data/battler-tags.ts
innerthunder: I think it would! My only concern would be that the tags themselves can't interact with the `damage` values directly, which might make things a bit messy.

src/field/pokemon.ts
innerthunder: It is meant to send a message when the substitute gets hit (and this is covered in the battler tag), but only for the first hit for multi-hit moves, as far as I know. I can easily check this with `hitCount` and `hitsLeft` from `source.turnData`, though.

src/data/move.ts
innerthunder: I simplified the calculation a bit. The idea is that the AI shouldn't use Substitute if they are a boss or don't have enough HP for it to succeed, and otherwise the score increases based on the user's HP ratio (so they're more inclined to use it at higher HP percentages)

src/data/move.ts
innerthunder: It does now, just updated `AddSubstituteAttr` to account for magic guard in the `hpCost` calculation.

src/data/battler-tags.ts
innerthunder: The other issue would be determining the order in which these tags would lapse. If a Pokemon with Disguise used Substitute, it'd be expected to have the damage redirected to the Substitute doll first.

src/locales/de/battle.ts
```diff
@@ -132,5 +132,8 @@ export const battle: SimpleTranslationEntries = {
   "battlerTagsSaltCuredOnAdd": "{{pokemonNameWithAffix}} wurde eingepökelt!",
   "battlerTagsSaltCuredLapse": "{{pokemonNameWithAffix}} wurde durch {{moveName}} verletzt!",
   "battlerTagsCursedOnAdd": "{{pokemonNameWithAffix}} nimmt einen Teil seiner KP und legt einen Fluch auf {{pokemonName}}!",
-  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} wurde durch den Fluch verletzt!"
+  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} wurde durch den Fluch verletzt!",
+  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}} put in a substitute!",
+  "battlerTagsSubstituteOnHit": "The substitute took damage for {{pokemonName}}!",
+  "battlerTagsSubstituteOnRemove": "{{pokemonNameWithAffix}}'s substitute faded!"
```
CodeTappert, line 138:
```suggestion
  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} wurde durch den Fluch verletzt!",
  "battlerTagsSubstituteOnAdd": "Ein Delegator von {{pokemonNameWithAffix}} ist erschienen!",
  "battlerTagsSubstituteOnHit": "Der Delegator steckt den Schlag für {{pokemonName}} ein!",
  "battlerTagsSubstituteOnRemove": "Der Delegator von {{pokemonNameWithAffix}} hört auf zu wirken!"
```

src/data/battler-tags.ts
```diff
@@ -1535,6 +1536,38 @@ export class IceFaceTag extends BattlerTag {
...
+    if (this.substituteHp <= 0) {
+      pokemon.scene.queueMessage(i18next.t("battle:battlerTagsSubstituteOnRemove", { pokemonNameWithAffix: getPokemonNameWithAffix(pokemon) }));
+    }
+  }
+
+  lapse(pokemon: Pokemon, lapseType: BattlerTagLapseType): boolean {
+    const ret = (lapseType !== BattlerTagLapseType.CUSTOM || super.lapse(pokemon, lapseType));
+
+    if (ret) {
+      pokemon.scene.queueMessage(i18next.t("battle:battlerTagsSubstituteOnHit", { pokemonName: pokemon.name }));
```
CodeTappert, line 1564: Why arent you using the pokemonNameWithAffix: getPokemonNameWithAffix(pokemon) here?

src/locales/de/battle.ts
```diff
@@ -132,5 +132,8 @@ export const battle: SimpleTranslationEntries = {
   "battlerTagsSaltCuredOnAdd": "{{pokemonNameWithAffix}} wurde eingepökelt!",
   "battlerTagsSaltCuredLapse": "{{pokemonNameWithAffix}} wurde durch {{moveName}} verletzt!",
   "battlerTagsCursedOnAdd": "{{pokemonNameWithAffix}} nimmt einen Teil seiner KP und legt einen Fluch auf {{pokemonName}}!",
-  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} wurde durch den Fluch verletzt!"
+  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} wurde durch den Fluch verletzt!",
+  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}} put in a substitute!",
+  "battlerTagsSubstituteOnHit": "The substitute took damage for {{pokemonNameWithAffix}}!",
+  "battlerTagsSubstituteOnRemove": "{{pokemonNameWithAffix}}'s substitute faded!"
```
CodeTappert, line 138:
```suggestion
  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} wurde durch den Fluch verletzt!",
  "battlerTagsSubstituteOnAdd": "Ein Delegator von {{pokemonNameWithAffix}} ist erschienen!",
  "battlerTagsSubstituteOnHit": "Der Delegator steckt den Schlag für {{pokemonNameWithAffix}} ein!",
  "battlerTagsSubstituteOnRemove": "Der Delegator von {{pokemonNameWithAffix}} hört auf zu wirken!"
```

src/locales/zh_CN/battle.ts
```diff
@@ -132,5 +132,8 @@ export const battle: SimpleTranslationEntries = {
   "battlerTagsSaltCuredOnAdd": "{{pokemonNameWithAffix}}\n陷入了盐腌状态！",
   "battlerTagsSaltCuredLapse": "{{pokemonNameWithAffix}}\n受到了{{moveName}}的伤害！",
   "battlerTagsCursedOnAdd": "{{pokemonNameWithAffix}}削减了自己的体力，\n并诅咒了{{pokemonName}}！",
-  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}}\n正受到诅咒！"
+  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}}\n正受到诅咒！",
+  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}} put in a substitute!",
+  "battlerTagsSubstituteOnHit": "The substitute took damage for {{pokemonNameWithAffix}}!",
+  "battlerTagsSubstituteOnRemove": "{{pokemonNameWithAffix}}'s substitute faded!"
```
RimKnight, line 138:
```suggestion
  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}}的\n替身出现了！",
  "battlerTagsSubstituteOnHit": "替身代替{{pokemonNameWithAffix}}\n承受了攻击！",
  "battlerTagsSubstituteOnRemove": "{{pokemonNameWithAffix}}的\n替身消失了……"
```

src/locales/zh_TW/battle.ts
```diff
@@ -130,4 +130,7 @@ export const battle: SimpleTranslationEntries = {
   "battlerTagsSaltCuredLapse": "{{pokemonNameWithAffix}} 受到了{{moveName}}的傷害！",
   "battlerTagsCursedOnAdd": "{{pokemonNameWithAffix}}削減了自己的體力，並詛咒了{{pokemonName}}！",
   "battlerTagsCursedLapse": "{{pokemonNameWithAffix}}正受到詛咒！",
+  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}} put in a substitute!",
+  "battlerTagsSubstituteOnHit": "The substitute took damage for {{pokemonNameWithAffix}}!",
+  "battlerTagsSubstituteOnRemove": "{{pokemonNameWithAffix}}'s substitute faded!"
```
RimKnight, line 135:
```suggestion
  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}}的\n替身出現了！",
  "battlerTagsSubstituteOnHit": "替身代替{{pokemonNameWithAffix}}承受了攻擊！",
  "battlerTagsSubstituteOnRemove": "{{pokemonNameWithAffix}}的\n替身消失了……"
```

RimKnight (commented):
Add zh-cn and zh-tw translation.

src/locales/ko/battle.ts
```diff
@@ -133,4 +133,7 @@ export const battle: SimpleTranslationEntries = {
   "battlerTagsSaltCuredLapse": "{{pokemonNameWithAffix}}[[는]] 소금절이의\n데미지를 입고 있다.",
   "battlerTagsCursedOnAdd": "{{pokemonNameWithAffix}}[[는]]\n자신의 체력을 깎아서\n{{pokemonName}}에게 저주를 걸었다!",
   "battlerTagsCursedLapse": "{{pokemonNameWithAffix}}[[는]]\n저주받고 있다!",
+  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}} put in a substitute!",
+  "battlerTagsSubstituteOnHit": "The substitute took damage for {{pokemonNameWithAffix}}!",
+  "battlerTagsSubstituteOnRemove": "{{pokemonNameWithAffix}}'s substitute faded!"
```
EnochG1, line 138:
```suggestion
  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}}의\n대타가 나타났다!",
  "battlerTagsSubstituteOnHit": "{{pokemonNameWithAffix}}[[를]] 대신하여\n대타가 공격을 받았다!",
  "battlerTagsSubstituteOnRemove": "{{pokemonNameWithAffix}}의\n대타는 사라져 버렸다..."
```

EnochG1 (commented):
Korean translation

EnochG1:
@innerthunder 
Just out of curiosity, what happens if pokemon don't have enough HP?
(In the original game, it fails with the message "But it does not have enough HP left to make a substitute!")

innerthunder:
> @innerthunder Just out of curiosity, what happens if pokemon don't have enough HP? (In the original game, it fails with the message "But it does not have enough HP left to make a substitute!")

Right now it should just fail with the generic "But it failed!" message.

src/locales/fr/battle.ts
```diff
@@ -132,5 +132,8 @@ export const battle: SimpleTranslationEntries = {
   "battlerTagsSaltCuredOnAdd": "{{pokemonNameWithAffix}}\nest couvert de sel !",
   "battlerTagsSaltCuredLapse": "{{pokemonNameWithAffix}} est blessé\npar la capacité {{moveName}} !",
   "battlerTagsCursedOnAdd": "{{pokemonNameWithAffix}} sacrifie des PV\net lance une malédiction sur {{pokemonName}} !",
-  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} est touché par la malédiction !"
+  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} est touché par la malédiction !",
+  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}} put in a substitute!",
+  "battlerTagsSubstituteOnHit": "The substitute took damage for {{pokemonNameWithAffix}}!",
+  "battlerTagsSubstituteOnRemove": "{{pokemonNameWithAffix}}'s substitute faded!"
```
Adri1, line 138:
```suggestion
  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}}\ncrée un clone !",
  "battlerTagsSubstituteOnHit": "Le clone subit les dégâts à la place\nde{{pokemonNameWithAffix}} !",
  "battlerTagsSubstituteOnRemove": "Le clone de {{pokemonNameWithAffix}}\ndisparait…"
```

Adri1 (commented):
Added French

src/locales/pt_BR/battle.ts
```diff
@@ -132,5 +132,8 @@ export const battle: SimpleTranslationEntries = {
   "battlerTagsSaltCuredOnAdd": "{{pokemonNameWithAffix}} está sendo curado com sal!",
   "battlerTagsSaltCuredLapse": "{{pokemonNameWithAffix}} foi ferido pelo {{moveName}}!",
   "battlerTagsCursedOnAdd": "{{pokemonNameWithAffix}} cortou seus PS pela metade e amaldiçoou {{pokemonName}}!",
-  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} foi ferido pelo Curse!"
+  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} foi ferido pelo Curse!",
+  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}} put in a substitute!",
+  "battlerTagsSubstituteOnHit": "The substitute took damage for {{pokemonNameWithAffix}}!",
+  "battlerTagsSubstituteOnRemove": "{{pokemonNameWithAffix}}'s substitute faded!"
```
josericardo-fo, line 138:
```suggestion
  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} foi ferido pelo Curse!",
  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}} colocou um substituto!",
  "battlerTagsSubstituteOnHit": "O substituto tomou o dano pelo {{pokemonNameWithAffix}}!",
  "battlerTagsSubstituteOnRemove": "O substituto de {{pokemonNameWithAffix}} desbotou!"
```

NicusPulcis:
hey @innerthunder i've added the italian translation

innerthunder:
## Animations

With Substitute comes a new suite of in-battle animations (and a new phase to implement them)! The videos below show all of the new animations surrounding the Substitute doll in action during a double battle.

### Videos

**Adding Substitutes:**

https://github.com/user-attachments/assets/7394e1ad-e066-4a79-a215-00c04f30ec9d

When a Pokemon uses Substitute, a new sprite is generated (with its reference attached to the respective `SubstituteTag`), and the Pokemon is moved "out of focus," becoming semi-transparent.

**Switching Pokemon with Substitutes:**

https://github.com/user-attachments/assets/edd2be80-a024-44f4-8538-bd96a7c2c2d5

- When a Pokemon with an active Substitute switches normally, the Substitute's sprite despawns (after a short animation) before the switched-in Pokemon is summoned.
- When a Pokemon with an active Substitute uses Baton Pass, the Substitute's sprite remains on the field and becomes visible after the move resolves.

**Hitting and Removing Substitutes:**

https://github.com/user-attachments/assets/884f6d08-0172-4815-9ac8-c9fb520546be

- When a Pokemon with a Substitute uses a move, it moves into focus, uses the move, then moves back out of focus.
- When a Pokemon's Substitute takes more than its remaining HP in damage, a fading animation despawns the Substitute's sprite and brings the Pokemon back "into focus."

**KOing Pokemon with Substitutes:**

https://github.com/user-attachments/assets/602954cd-b456-43c0-8594-dd3fb51ea2af

When a Pokemon faints while it has a Substitute active, the Substitute's sprite is also removed from the field.

**Loading in Pokemon with Substitutes:**

https://github.com/user-attachments/assets/2ab94298-7ac5-47c8-86d5-180899268e3c

When the player loads into a wave where a Pokemon started with a Substitute, the on-add animation is replayed to regenerate the Substitute sprite.

src/data/move.ts
```diff
@@ -541,14 +554,25 @@ export default class Move implements Localizable {
   /**
    * Sets the {@linkcode MoveFlags.CHECK_ALL_HITS} flag for the calling Move
    * @param checkAllHits The value (boolean) to set the flag to
-   * example: @see {@linkcode Moves.TRIPLE_AXEL}
+   * example: @see {@linkcode Moves.WHIRLWIND}
```
Xavion3, line 557: Why is this being changed?

src/data/move.ts
```diff
@@ -1292,6 +1316,49 @@ export class HalfSacrificialAttr extends MoveEffectAttr {
...
+   * @param args n/a
+   * @returns true if the attribute successfully applies, false otherwise
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    if (!super.apply(user, target, move, args)) {
+      return false;
+    }
+
+    const blockSelfDamage = new Utils.BooleanHolder(false);
+    applyAbAttrs(BlockNonDirectDamageAbAttr, user, blockSelfDamage);
```
Xavion3, line 1344: Magic guard doesn't prevent you damaging yourself with substitute.

src/data/move.ts
```diff
@@ -6218,14 +6349,15 @@ export function initMoves() {
       .attr(HighCritAttr)
       .slicingMove(),
     new SelfStatusMove(Moves.SUBSTITUTE, Type.NORMAL, -1, 10, -1, 0, 1)
-      .attr(RecoilAttr)
-      .unimplemented(),
+      .attr(AddSubstituteAttr)
+      .condition((user, target, move) => !user.getSubstitute() && user.getHpRatio() > 0.25 && !user.hasAbility(Abilities.WONDER_GUARD, false, true)),
```
Xavion3, line 6353: Just check if max hp is > 1 probably, as that is what we're actually looking for.

src/field/pokemon.ts
```diff
@@ -595,6 +613,60 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * @see {@linkcode SubstituteTag}
+   */
+  getSubstitute(): SubstituteTag {
+    const subTag = this.getTag(BattlerTagType.SUBSTITUTE);
+    if (!(subTag instanceof SubstituteTag)) {
+      return null;
+    } else {
+      return subTag;
+    }
+  }
```
Xavion3, line 627: Why does this exist when its just a wrapper on the already existing getTag?

Xavion3 (changes requested):
Started reviewing this, but its a lot.

src/field/pokemon.ts
innerthunder: I made it because `getTag` would return the substitute tag as a `BattlerTag`. Since the code to manage the Substitute's sprite (outside of the tag itself) needs to access the tag's `sprite` field, it needs to be interpreted as a `SubstituteTag`.

I.e. instead of getting the tag's sprite like this:
```ts
const substitute = pokemon.getSubstitute();
if (!!substitute) {
  // do stuff with substitute.sprite
}
```
I'd either have to add a `instanceof` check wherever I need to access the Substitute's sprite
```ts
const substitute = pokemon.getTag(BattlerTagType.SUBSTITUTE);
if (!!substitute && substitute instanceof SubstituteTag) {
  // do stuff with substitute.sprite
}
```
or use `as` to change the type, which isn't as safe as using `instanceof`.
```ts
const substitute = pokemon.getTag(BattlerTagType.SUBSTITUTE) as SubstituteTag;
if (!!substitute) {
  // do stuff with substitute.sprite
}
```

src/data/move.ts
innerthunder: It was a typo 😶, just fixed it

src/locales/es/battle.ts
```diff
@@ -154,5 +154,8 @@ export const battle: SimpleTranslationEntries = {
   "battlerTagsSaltCuredOnAdd": "{{pokemonNameWithAffix}} is being salt cured!",
   "battlerTagsSaltCuredLapse": "{{pokemonNameWithAffix}} is hurt by {{moveName}}!",
   "battlerTagsCursedOnAdd": "{{pokemonNameWithAffix}} cut its own HP and put a curse on the {{pokemonName}}!",
-  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} is afflicted by the Curse!"
+  "battlerTagsCursedLapse": "{{pokemonNameWithAffix}} is afflicted by the Curse!",
+  "battlerTagsSubstituteOnAdd": "{{pokemonNameWithAffix}} put in a substitute!",
+  "battlerTagsSubstituteOnHit": "The substitute took damage for {{pokemonNameWithAffix}}!",
+  "battlerTagsSubstituteOnRemove": "{{pokemonNameWithAffix}}'s substitute faded!"
```
AsdarDevelops, line 160:
```suggestion
  "battlerTagsSubstituteOnAdd": "¡{{pokemonNameWithAffix}} creó un sustituto!",
  "battlerTagsSubstituteOnHit": "¡El sustituto recibe daño en lugar del {{pokemonNameWithAffix}}!",
  "battlerTagsSubstituteOnRemove": "¡El sustituto del {{pokemonNameWithAffix}} se debilitó!"
```

AsdarDevelops (changes requested):
Spanish localization

innerthunder:
## Updates since opening this PR

- ES locale updated. All of Substitute's messages should be fully translated now!
- `data/battle-anims`: Added an option in `BattleAnim.play` to center target-based animations on the target Pokemon's Substitute instead of the target Pokemon itself. Calls to `play` in `phases.ts` and other files were also edited to account for the new option.
- `phases`: Target Pokemon no longer flash in the `DamagePhase` when the damage taken for that phase is 0. This means a Pokemon will not flash when it redirects damage with a Substitute.
- `BattleAnimPhase` and `BattleAnimType` have been renamed to `PokemonAnimPhase` and `PokemonAnimType` respectively to avoid confusion with the `BattleAnim` class in `data/battle-anims.ts`. The files containing these classes have also been renamed accordingly.

src/field/pokemon.ts
Xavion3: `as` is implicitly safe here due to the structure of getTag as only being able to return the given tag or a falsy value. Like, we could teach typescript this if we really wanted, but we haven't bothered because it'd add dozens of lines of code just to make this slightly less aggressive about throwing warnings.

innerthunder: Just removed this function and added overloads to `getTag` from https://github.com/pagefaultgames/pokerogue/pull/2960 as discussed in the Discord:

```ts
/** @overload */
getTag(tagType: BattlerTagType): BattlerTag;

/** @overload */
getTag<T extends BattlerTag>(tagType: Constructor<T>): T;
```

src/test/moves/substitute.test.ts
```diff
@@ -0,0 +1,442 @@
...
+  });
+
+  afterEach(() => {
+    game.phaseInterceptor.restoreOg();
+  });
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+
+    vi.spyOn(overrides, "SINGLE_BATTLE_OVERRIDE", "get").mockReturnValue(true);
```
DayKev, line 36: There's [an overrides refactor](https://github.com/pagefaultgames/pokerogue/pull/3104) coming soon as requested by flx so you'll probably need to change this to `vi.spyOn(overrides, "BATTLE_TYPE_OVERRIDE", "get").mockReturnValue("single");`.

DayKev:
Oops I almost forgot to mention, you should probably add a test for Baton Pass.

>A substitute can be passed by [Baton Pass](https://bulbapedia.bulbagarden.net/wiki/Baton_Pass_(move)) with the HP it has remaining.

- https://bulbapedia.bulbagarden.net/wiki/Substitute_(doll)#Effect

snoozbuster:
Are we able to implement shed tail by combining AddSubstituteAttr and ForceSwitchOutAttr? I looked through move.ts in this PR and it didn't seem like it had been done

src/data/ability.ts
```diff
@@ -1965,6 +1969,10 @@ export class PostSummonStatChangeAbAttr extends PostSummonAbAttr {
       if (this.intimidate) {
         applyAbAttrs(IntimidateImmunityAbAttr, opponent, cancelled);
         applyAbAttrs(PostIntimidateStatChangeAbAttr, opponent, cancelled);
+
+        if (!!opponent.getTag(BattlerTagType.SUBSTITUTE)) {
```
torranx, line 1973: I think placing the `opponent.getTag()` inside an if condition already converts it to bool?

src/data/move.ts
```diff
@@ -4451,6 +4553,22 @@ export class FaintCountdownAttr extends AddBattlerTagAttr {
   }
 }
 
+export class RemoveAllSubstitutesAttr extends MoveEffectAttr {
```
torranx, line 4556: add docs

src/test/moves/substitute.test.ts
```diff
@@ -0,0 +1,403 @@
...
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+
+    game.override
+      .battleType("single")
+      .moveset([Moves.SUBSTITUTE, Moves.SWORDS_DANCE, Moves.TACKLE, Moves.SPLASH])
+      .enemySpecies(Species.SNORLAX)
+      .enemyAbility(Abilities.INSOMNIA)
+      .enemyMoveset(Array(4).fill(Moves.SPLASH))
```
torranx, line 40:
```suggestion
      .enemyMoveset(SPLASH_ONLY)
```

src/test/moves/substitute.test.ts
```diff
@@ -0,0 +1,403 @@
...
+    async () => {
+      await game.startBattle([Species.MAGIKARP]);
+
+      const leadPokemon = game.scene.getPlayerPokemon();
+
+      game.doAttack(getMovePosition(game.scene, 0, Moves.SUBSTITUTE));
+
+      await game.phaseInterceptor.to(MoveEndPhase, false);
+
+      expect(leadPokemon.hp).toBe(Math.ceil(leadPokemon.getMaxHp() * 3/4));
```
torranx, line 56: since the `enemyLevel` is high, there's a tendency that it will heal through item and this could fail

DayKev: The player Pokémon shouldn't spawn with any held items, and the enemy Pokémon is already set to only have Splash. In this case I don't think any held items can interfere.

torranx: oh right, I thought it was checking for enemy

src/battle-scene.ts
```diff
@@ -990,10 +992,12 @@ export default class BattleScene extends SceneBase {
     this.enemyModifierBar.removeAll(true);
 
     for (const p of this.getParty()) {
+      p.destroySubstitute();
```
snoozbuster, line 995: is there some reason why `p.destroy()` shouldn't call destroySubstitute? is there some weird edge case where we want to keep the sub even though the pkmn is destroyed?

src/data/ability.ts
```diff
@@ -1605,6 +1605,10 @@ export class PostAttackApplyStatusEffectAbAttr extends PostAttackAbAttr {
   }
 
   applyPostAttackAfterMoveTypeCheck(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: Move, hitResult: HitResult, args: any[]): boolean {
+    if (pokemon !== attacker && !!pokemon.getTag(BattlerTagType.SUBSTITUTE) && !move.canIgnoreSubstitute(attacker)) {
```
snoozbuster, line 1608: `!!` on `getTag` shouldn't be necessary, all types are coerced to boolean inside an if statement (as far as I'm aware, even TS shouldn't complain on strict settings). it's only necessary if you want to return the value or store it in a boolean variable

src/data/battle-anims.ts
```diff
@@ -711,12 +712,15 @@ export abstract class BattleAnim {
       const user = !isOppAnim ? this.user : this.target;
       const target = !isOppAnim ? this.target : this.user;
 
+      const targetSubstitute = (!!onSubstitute && user !== target) ? target!.getTag(SubstituteTag) : null;
```
snoozbuster, line 715: this is another place where `!!` shouldn't be necessary

src/data/battler-tags.ts
```diff
@@ -1828,6 +1829,94 @@ export class ExposedTag extends BattlerTag {
...
+  }
+
+  /**
+  * When given a battler tag or json representing one, load the data for it.
+  * @param {BattlerTag | any} source A battler tag
+  */
+  loadTag(source: BattlerTag | any): void {
+    super.loadTag(source);
+    this.hp = source.hp;
+    // TODO: load this tag's sprite (or generate a new one upon loading a game)
```
snoozbuster, line 1916: still relevant?

src/data/move.ts
```diff
@@ -99,11 +99,12 @@ export enum MoveFlags {
   /**
    * Enables all hits of a multi-hit move to be accuracy checked individually
    */
-  CHECK_ALL_HITS   = 1 << 17,
+  CHECK_ALL_HITS    = 1 << 17,
+  IGNORE_SUBSTITUTE = 1 << 18,
```
snoozbuster, line 103: can you add a docstring? I understand that this particular flag is relatively obvious, but it's still a good pattern to have.

src/data/move.ts
```diff
@@ -318,6 +319,18 @@ export default class Move implements Localizable {
...
   }
 
+  /**
+   * Checks if the move can bypass Substitute to directly hit its target
+   * @param user The {@linkcode Pokemon} using this move
+   * @returns `true` if the move can bypass the target's Substitute; `false` otherwise.
+   */
+  canIgnoreSubstitute(user: Pokemon): boolean {
+    return this.moveTarget === MoveTarget.USER
+        || user?.hasAbility(Abilities.INFILTRATOR)
```
snoozbuster, line 329: pursuit needs to conditionally bypass substitute - it should only bypass it if the special effect triggers. instead of explicitly checking for an ability do you think that we could use a battler tag or something like that, which then infiltrator can apply? do you know if there's a better solution for this?

src/data/move.ts
```diff
@@ -583,7 +607,8 @@ export default class Move implements Localizable {
     // special cases below, eg: if the move flag is MAKES_CONTACT, and the user pokemon has an ability that ignores contact (like "Long Reach"), then overrides and move does not make contact
     switch (flag) {
     case MoveFlags.MAKES_CONTACT:
-      if (user.hasAbilityWithAttr(IgnoreContactAbAttr)) {
+      if (user.hasAbilityWithAttr(IgnoreContactAbAttr) ||
+          (target?.getTag(BattlerTagType.SUBSTITUTE) && !this.canIgnoreSubstitute(user))) {
```
snoozbuster, line 611: I feel like I've seen the pattern of `pkmn.getTag(sub) && canIgnoreSub(user)` in a number of places already. do you think these could be combined somehow? for example, perhaps `pkmn.getTag(sub)?.canIgnoreSub(...)` would work well - it automatically evaluates to a falsy value if the sub tag is missing. I think there's also some value in putting the logic for what moves can ignore sub within the main part of the sub code (the tag) - keeps related code in more or less the same place. 

maybe there's a better place for this to live - perhaps the Pokemon class could have the `canIgnoreSubstitute` method which internally first looks for the tag and then does the rest of the logic (although in that case, based on the implementation and params it would need to be called something like `isBypassingSubstitute`). but it seems like it ought to be one fluid call somehow instead of two distinct checks that always need to happen together.

src/data/move.ts
```diff
@@ -596,8 +621,8 @@ export default class Move implements Localizable {
         }
       }
     case MoveFlags.IGNORE_PROTECT:
-      if (user.hasAbilityWithAttr(IgnoreProtectOnContactAbAttr) &&
-          this.checkFlag(MoveFlags.MAKES_CONTACT, user, target)) {
+      if (user.hasAbilityWithAttr(IgnoreProtectOnContactAbAttr)
+          && this.checkFlag(MoveFlags.MAKES_CONTACT, user, null)) {
```
snoozbuster, line 625: why the change to null?

src/data/move.ts
```diff
@@ -1941,6 +2004,10 @@ export class StatusEffectAttr extends MoveEffectAttr {
   }
 
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    if (!this.selfTarget && !!target.getTag(SubstituteTag) && !move.canIgnoreSubstitute(user)) {
```
snoozbuster, line 2007: another unnecessary `!!`, plus another instance of the `getTag() && canIgnoreSub()` pattern

src/data/move.ts
```diff
@@ -2033,6 +2100,9 @@ export class StealHeldItemChanceAttr extends MoveEffectAttr {
 
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): Promise<boolean> {
     return new Promise<boolean>(resolve => {
+      if (!!target.getTag(SubstituteTag) && !move.canIgnoreSubstitute(user)) {
```
snoozbuster, line 2103: same pattern

src/data/move.ts
```diff
@@ -2102,6 +2172,10 @@ export class RemoveHeldItemAttr extends MoveEffectAttr {
       return false;
     }
 
+    if (!!target.getTag(SubstituteTag) && !move.canIgnoreSubstitute(user)) {
```
snoozbuster, line 2175: etc and so on (I'm not commenting on any more but I imagine you can find them all by searching for `canIgnoreSubstitute`

src/data/move.ts
```diff
@@ -6628,14 +6769,15 @@ export function initMoves() {
       .attr(HighCritAttr)
       .slicingMove(),
     new SelfStatusMove(Moves.SUBSTITUTE, Type.NORMAL, -1, 10, -1, 0, 1)
-      .attr(RecoilAttr)
-      .unimplemented(),
+      .attr(AddSubstituteAttr)
+      .condition((user, target, move) => !user.getTag(SubstituteTag) && user.getHpRatio() > 0.25 && user.getMaxHp() > 1),
```
snoozbuster, line 6773: I feel like you should encode this condition into `AddSubstituteAttr`'s `apply()` method instead. actually, is that how that works? I guess the move doesn't fail if that function returns false. isn't there another hook that attrs have to force-fail the move though?

given how heavily dependent the attr's `apply()` method is on this check being present and correct, it seems like the condition that guarantees it should be somewhere nearby.

src/data/move.ts
```diff
@@ -7600,6 +7762,7 @@ export function initMoves() {
     new AttackMove(Moves.SKY_DROP, Type.FLYING, MoveCategory.PHYSICAL, 60, 100, 10, -1, 0, 5)
       .attr(ChargeAttr, ChargeAnim.SKY_DROP_CHARGING, i18next.t("moveTriggers:tookTargetIntoSky", {pokemonName: "{USER}", targetName: "{TARGET}"}), BattlerTagType.FLYING) // TODO: Add 2nd turn message
       .condition(failOnGravityCondition)
+      .condition((user, target, move) => !target.getTag(BattlerTagType.SUBSTITUTE))
```
snoozbuster, line 7765: don't you need the canIgnoreSub check here? what if a sky drop user has infiltrator?

src/data/battler-tags.ts
```diff
@@ -213,8 +215,9 @@ export class TrappedTag extends BattlerTag {
   canAdd(pokemon: Pokemon): boolean {
     const isGhost = pokemon.isOfType(Type.GHOST);
     const isTrapped = pokemon.getTag(BattlerTagType.TRAPPED);
+    const hasSubstitute = pokemon.getTag(BattlerTagType.SUBSTITUTE);
```
snoozbuster, line 218: I guess sub always blocks trapping moves even if the user has inflitrator?

src/data/move.ts
innerthunder: From [Bulba](https://bulbapedia.bulbagarden.net/wiki/Substitute_(doll)#Unaffected_moves):

> Since [Generation VI](https://bulbapedia.bulbagarden.net/wiki/Generation_VI), a Pokémon with [Infiltrator](https://bulbapedia.bulbagarden.net/wiki/Infiltrator_(Ability)) can ignore a substitute with any move except [Transform](https://bulbapedia.bulbagarden.net/wiki/Transform_(move)) or [Sky Drop](https://bulbapedia.bulbagarden.net/wiki/Sky_Drop_(move)).

src/data/move.ts
innerthunder: Done

src/data/move.ts
innerthunder: Unseen Fist is meant to ignore Protect with every move that's allowed to make contact, even if the move would not make contact against a specific target.

Consider this scenario:
- Player is a Pokemon with Unseen Fist, Enemy is a Pokemon with Flame Body and a Substitute from a previous turn.
- Enemy uses Protect, then Player attacks into Enemy with Tackle (a move that makes contact in normal situations).

Because Player has Unseen Fist, it would ignore Enemy's Protect and deal damage to Enemy's Substitute. However, because the attack was deflected onto the Substitute, Player technically didn't make contact with Enemy, so Enemy's Flame Body will never activate.

Oh, and if you're wondering "Why not just use `hasTag`?", Unseen Fist still respects Long Reach if the user were to have both abilities (as far as I know, this is currently impossible with the passives available, but if a similar effect to Long Reach were implemented via something like Punching Gloves, then this would be relevant).

src/data/move.ts
innerthunder: I ended up reducing this (and similar checks) to `move.hitsSubstitute(user, target)` (or the inverse). This cleaned things up pretty well, but I still feel like there are some redundant calls to it that I can reduce.

src/data/battler-tags.ts
innerthunder: nope 🍝

src/data/move.ts
innerthunder: Done

src/data/move.ts
innerthunder: I figure the easiest way would be to add a field to `turnData` (say, `isPursuing`). Not really sure if there's an easy way to check if a specific move attribute has applied via the move itself atm.

src/data/move.ts
innerthunder: It probably would be better off in a `getCondition()` method for the attribute (putting it in `apply()` wouldn't generate a fail message for that condition), but yeah.

EnochG1 (dismissed):

src/data/battler-tags.ts
innerthunder: Not totally sure on this. Bulba says gaining a Substitute removes all binding effects from the user, but whether or not blocked effects are still applicable when the attacker has Infiltrator isn't made clear.

I ended up changing this condition to `!move.hitsSubstitute(attacker)`, so trap effects will go through if the attacker has Infiltrator for the time being.

Chapybara-jp (approved):

src/data/move.ts
```diff
@@ -102,11 +102,15 @@ export enum MoveFlags {
...
+  CHECK_ALL_HITS    = 1 << 17,
+  /**
+   * Indicates a move is able to bypass its target's Substitute (if the target has one)
+   */
+  IGNORE_SUBSTITUTE = 1 << 18,
   /**
    * Indicates a move is able to be redirected to allies in a double battle if the attacker faints
    */
-  REDIRECT_COUNTER = 1 << 18,
+  REDIRECT_COUNTER  = 1 << 19,
```
DayKev, line 113:
```suggestion
  REDIRECT_COUNTER = 1 << 19,
```

src/data/move.ts
```diff
@@ -1439,6 +1470,58 @@ export class HalfSacrificialAttr extends MoveEffectAttr {
...
+   * @param move the {@linkcode Move} with this attribute.
+   * @param args n/a
+   * @returns true if the attribute successfully applies, false otherwise
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    if (!super.apply(user, target, move, args)) {
+      return false;
+    }
+
+    const hpCost = Math.floor(user.getMaxHp() / 4);
```
DayKev, line 1497: This should be a parameter for the move to set, because Shed Tail deals 50% damage to the user (though the sub still only has HP equal to 25% of the user's max HP).

src/field/pokemon.ts
```diff
@@ -640,6 +658,47 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  /** If this Pokemon has a Substitute on the field, removes its sprite from the field. */
+  destroySubstitute(): void {
+    const substitute = this.getTag(SubstituteTag);
+    if (!!substitute && !!substitute.sprite) {
```
DayKev, line 697:
```suggestion
    if (substitute && substitute.sprite) {
```

src/field/pokemon.ts
```diff
@@ -640,6 +658,47 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+  }
+
+  /**
+   * Returns whether or not the Pokemon's position on the field is offset because
+   * the Pokemon has a Substitute active.
+   * @see {@linkcode SubstituteTag}
+   */
+  isOffsetBySubstitute(): boolean {
+    const substitute = this.getTag(SubstituteTag);
+    if (!!substitute) {
```
DayKev, line 678:
```suggestion
    if (substitute) {
```

src/field/pokemon.ts
```diff
@@ -3124,6 +3200,11 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
           this.summonData[k] = this.summonDataPrimer[k];
         }
       }
+      // If this Pokemon has a Substitute when loading in, play an animation to add its sprite
+      if (!!this.getTag(SubstituteTag)) {
```
DayKev, line 3204:
```suggestion
      if (this.getTag(SubstituteTag)) {
```

src/phases/switch-summon-phase.ts
```diff
@@ -11,6 +11,7 @@ import { Command } from "#app/ui/command-ui-handler";
 import i18next from "i18next";
 import { PostSummonPhase } from "./post-summon-phase";
 import { SummonPhase } from "./summon-phase";
+import { SubstituteTag } from "#app/data/battler-tags.js";
```
DayKev, line 14:
```suggestion
import { SubstituteTag } from "#app/data/battler-tags";
```

src/phases/switch-summon-phase.ts
```diff
@@ -65,6 +66,16 @@ export class SwitchSummonPhase extends SummonPhase {
 
     if (!this.batonPass) {
       (this.player ? this.scene.getEnemyField() : this.scene.getPlayerField()).forEach(enemyPokemon => enemyPokemon.removeTagsBySourceId(pokemon.id));
+      const substitute = pokemon.getTag(SubstituteTag);
+      if (!!substitute) {
```
DayKev, line 70:
```suggestion
      if (substitute) {
```

src/phases/switch-summon-phase.ts
```diff
@@ -115,8 +126,18 @@ export class SwitchSummonPhase extends SummonPhase {
...
         );
-        // Ensure improperly persisted summon data (such as tags) is cleared upon switching
-        if (!this.batonPass) {
+        /**
+         * If this switch is passing a Substitute, make the switched Pokemon match the returned Pokemon's state as it left.
+         * Otherwise, clear any persisting tags on the returned Pokemon.
+         */
+        if (this.batonPass) {
+          const substitute = this.lastPokemon.getTag(SubstituteTag);
+          if (!!substitute) {
```
DayKev, line 135:
```suggestion
          if (substitute) {
```

src/ui/target-select-ui-handler.ts
```diff
@@ -8,6 +8,7 @@ import {Button} from "#enums/buttons";
 import { Moves } from "#enums/moves";
 import Pokemon from "#app/field/pokemon";
 import { ModifierBar } from "#app/modifier/modifier";
+import { SubstituteTag } from "#app/data/battler-tags.js";
```
DayKev, line 11:
```suggestion
import { SubstituteTag } from "#app/data/battler-tags";
```

src/data/move.ts
innerthunder: I don't think Shed Tail will be able to make use of this attribute easily either way. The sequencing for adding the substitute doll fundamentally changes since Shed Tail has to generate the substitute onto the switched-in Pokemon (kinda like Baton Pass's implementation, I imagine).

DayKev: Well the HP of the sub needs to be preserved when using Shed Tail or Baton Pass, are you saying they'd have to generate a new sub and manually set the HP value of it or something?

innerthunder: Nvm, I just read Smogon's description of Shed Tail. I assumed the substitute was generated on the switched-in Pokemon after the switch resolved, but it's actually generated beforehand on the move user and then transferred over.

innerthunder: Actually, I just noticed that the HP costs for Substitute and Shed Tail are rounded differently???

Substitute:
>The user takes 1/4 of its maximum HP, rounded down, and puts it into a substitute to take its place in battle.

Shed Tail:
>The user takes 1/2 of its maximum HP, rounded up, and creates a substitute that has 1/4 of the user's maximum HP, rounded down.

innerthunder: Eh, I think Shed Tail needs a dedicated attribute anyway. The order of adding the substitute, then switching and transferring the substitute to the party Pokemon is something that should be enforced by a single attribute imo.

SangaraSorama (approved):

src/data/move.ts
DayKev: The sub itself has the same rounding, at least.

DayKev:
Other than sprite issues, it seems good.

Sub sprite offset from where it should be:
image
Gastly still offset after transitioning to a trainer battle:
image

DayKev (approved):

PR #2568, opened 2024-06-24
[Bug] Fix noice eiscue crashing on arena reset

torranx:
## What are the changes?
- fixes crash caused by noice eiscue on arena reset
- unit test did not catch this because the error came from the pokemon's cry 

- when the arena resets, the game loads the noice form audio file (875-no-ice.m4a) instead of the 875.m4a
- my best guess is that the game loads the audio first before the `pokemon.formIndex = 0` was called, hence why it cant find the audio and the game crashes

## What did change?
- change manually setting `pokemon.formIndex = 0` to `triggerPokemonFormChange()`

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/68144167/d55f0c16-57d8-4f46-9bb6-ed6ecc3f3f19

## How to test the changes?
- use noice eiscue on wave 4, play to wave 5

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -37,7 +37,7 @@ import UIPlugin from "phaser3-rex-plugins/templates/ui/ui-plugin";
 import { addUiThemeOverrides } from "./ui/ui-theme";
 import PokemonData from "./system/pokemon-data";
 import { Nature } from "./data/nature";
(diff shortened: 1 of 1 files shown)


Tempo-anon (approved):
Can you update your PR desc to have more details? Why was setting the formIndex manually causing the crash? How does using triggerPokemonFormChange prevent the crash?

PR #2576, opened 2024-06-24
[QoL] Allow swapping keyboard bindings

### LINKED ISSUE #2511: [BUG] No warning when rebinding keyboard keys that are already in use
**Describe the bug**

When binding a function to a key that's already in use, the Gamepad settings will warn you that "[Button1] will swap with [Button2]" and ask you to "Confirm swap". If you do, the two buttons trade places.

When doing the same thing in the Keyboard menu, you receive no warning up front and instead of swapping, the key is just silently removed from the old function, leaving it unbound. This can obviously lead to real problems with accidentally unbinding keys.

**To Reproduce**

1. Go into Game Settings
2. Change a function in the Gamepad menu to another key that's already in use
3. Repeat in the Keyboard menu

**Expected behavior**

The game warns you when you're about to change a bind in a way that affects other existing binds.

**Screenshots / Videos**

image
Gamepad menu

image
Keyboard menu

**Device**

Chrome on PC.

**Additional context**

I assume the "Delete" prompt is for unbinding a function entirely, but pressing Del on my keyboard doesn't seem to do anything.

EmberCM:
## What are the changes?
Allow users to swap keyboard bindings instead of just removing the old one without warning

Fixes #2511 

## What did change?
The keyboard keybind change screen now shows what you're changing from and to.

### Screenshots/Videos
Before vs After
before
new

## How to test the changes?
1. Go to Game Settings > Keyboard
2. Setting Up (Alt) to S will set it to S and change Down (Alt) to W, swapping the keybinds

## Checklist
- [x] Are the changes visual?

tests

### DIFF
--- a/src/configs/inputs/configHandler.ts
+++ b/src/configs/inputs/configHandler.ts
@@ -118,17 +118,25 @@ export function assign(config, settingNameTarget, keycode): boolean {
   if (!canIAssignThisKey(config, getKeyWithKeycode(config, keycode)) || !canIOverrideThisSetting(config, settingNameTarget)) {
     return false;
   }
+
+  // Check if the new key pressed already exists for a keybind
   const previousSettingName = getSettingNameWithKeycode(config, keycode);
-  // if it was already bound, we delete the bind
-  if (previousSettingName) {
-    const previousKey = getKeyWithSettingName(config, previousSettingName);
-    config.custom[previousKey] = -1;
-  }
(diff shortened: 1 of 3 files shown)


Greenlamp2:
Unfortunately, this is not that easy. this was how i did first but there was too many unknown issue with it, as you could have guess with commented code you uncommented to do this PR.
I chose to do it that way, so i don't want to revert it to this, thanks though for your work.

Issue #2558, opened 2024-06-24
[BUG] Disguise doesn't work?


C-ater:
**Describe the bug**
Disguise does not protect from damage?

**To Reproduce**

1. Hit Mimikyu 
2. Mimikyu dies 
3. Disguise activates afterwards...?

**Expected behavior**
Disguise makes you take portion of damage (that I forgot) and protects Mimikyu from further damage

**Screenshots / Videos**

https://github.com/pagefaultgames/pokerogue/assets/173323483/e2aeb508-75cd-4bf3-8cc2-42abcc4e8064

**Device**
Phone (Samsung S22)

**Additional context**
I've had this happen multiple times where I'll OHKO a Mimikyu through Disguise, and this same sequence of events occurs. 

Everytime this has happened, Mimikyu was OHKO'd by the move it was supposed to block

I've only ever had this happen to the AI

torranx:
This should be fixed by now, please close upon confirmation.

C-ater:
Mimikyu now is a pain  on Ivy's team and not a free kill 💀

PR #2592, opened 2024-06-25
[Bug] Fix Dry Skin and ReceivedMoveDamageMultiplierAbAttr abilities

Corrade:
## What are the changes?
- Fixed Dry Skin's damage during sunlight not applying to Pokemon at maximum health
- Fixed ReceivedMoveDamageMultiplierAbAttr not working for all values != 0 (i.e. it was broken for everything except Ice Face), which fixes the damage modification part of the following abilities: Filter, Solid Rock, Multiscale, Fur Coat, Fluffy, Shadow Shield, Prism Armor, Punk Rock, Ice Scales, Thick Fat, Heatproof, Dry Skin, Water Bubble, Purifying Salt, Disguise
- Added tests for Dry Skin

- Bug fixes

## What did change?
- ReceivedMoveDamageMultiplierAbAttr changes damage instead of power in pokemon.ts
- PostWeatherLapseDamageAbAttr is fixed to damage Pokemon at max HP in ability.ts
- Renamed "power" to "damage" in ReceivedMoveDamageMultiplierAbAttr and friends (checking Bulbapedia, it's indeed appropriate to scale damage rather than (or) power for all the abilities above)

### Screenshots/Videos
- N/A

## How to test the changes?
- Run the "dry_skin" test suite. Reverse either change in ability.ts or pokemon.ts and observe respective failures in the "during sunlight, lose 1/8 of maximum health at the end of each turn" and "opposing fire attacks do 25% more damage" test cases.
- The bugs in the other abilities can be observed using overrides.

## Checklist
- [No] Are the changes visual?
  - [N/A] Have I provided screenshots/videos of the changes?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2726,7 +2726,7 @@ export class PostWeatherLapseDamageAbAttr extends PostWeatherLapseAbAttr {
   }
 
   applyPostWeatherLapse(pokemon: Pokemon, passive: boolean, weather: Weather, args: any[]): boolean {
-    if (pokemon.getHpRatio() < 1) {
+    if (pokemon.getHpRatio() <= 1) {
       const scene = pokemon.scene;
       const abilityName = (!passive ? pokemon.getAbility() : pokemon.getPassiveAbility()).name;
       scene.queueMessage(getPokemonMessage(pokemon, ` is hurt\nby its ${abilityName}!`));
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2726,15 +2726,11 @@ export class PostWeatherLapseDamageAbAttr extends PostWeatherLapseAbAttr {
   }
 
   applyPostWeatherLapse(pokemon: Pokemon, passive: boolean, weather: Weather, args: any[]): boolean {
-    if (pokemon.getHpRatio() <= 1) {
-      const scene = pokemon.scene;
-      const abilityName = (!passive ? pokemon.getAbility() : pokemon.getPassiveAbility()).name;
-      scene.queueMessage(getPokemonMessage(pokemon, ` is hurt\nby its ${abilityName}!`));
-      pokemon.damageAndUpdate(Math.ceil(pokemon.getMaxHp() / (16 / this.damageFactor)), HitResult.OTHER);
-      return true;
-    }
-
-    return false;
+    const scene = pokemon.scene;
+    const abilityName = (!passive ? pokemon.getAbility() : pokemon.getPassiveAbility()).name;
+    scene.queueMessage(getPokemonMessage(pokemon, ` is hurt\nby its ${abilityName}!`));
+    pokemon.damageAndUpdate(Math.ceil(pokemon.getMaxHp() / (16 / this.damageFactor)), HitResult.OTHER);
+    return true;
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/field/pokemon.ts
```diff
@@ -1954,11 +1954,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
         }
 
         applyMoveAttrs(ModifiedDamageAttr, source, this, move, damage);
-        applyPreDefendAbAttrs(ReceivedMoveDamageMultiplierAbAttr, this, source, move, cancelled, power);
-
```
Corrade, line 1958: Before, we were modifying "power", but at this point power no longer impacts the damage calculation. It was factored in on line 1889. The "damage" variable is the source of truth hereafter as it's passed into damageAndUpdate() on line 1980. I've accordingly changed this call to modify "damage".

Corrade:
For the record, ReceivedMoveDamageMultiplierAbAttr was moved in this PR a week ago: https://github.com/pagefaultgames/pokerogue/pull/2416

I think this fix is the best of both worlds?

src/data/ability.ts
```diff
@@ -326,7 +326,7 @@ export class ReceivedTypeDamageMultiplierAbAttr extends ReceivedMoveDamageMultip
   }
 }
 
-export class PreDefendMovePowerToOneAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
+export class PreDefendMoveDamageToOneAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
```
Corrade, line 329: Cosmetic renaming, but fyi the underlying fix to ReceivedMoveDamageMultiplierAbAttr in pokemon.ts likely explains and fixes issue https://github.com/pagefaultgames/pokerogue/issues/2558.

Aside: I feel like this piece of code should set the damage to 0 rather than 1 since Disguise is supposed to prevent all initial damage right..? Idk the details of how Disguise was implemented though.

src/data/ability.ts
```diff
@@ -2726,15 +2726,11 @@ export class PostWeatherLapseDamageAbAttr extends PostWeatherLapseAbAttr {
   }
 
   applyPostWeatherLapse(pokemon: Pokemon, passive: boolean, weather: Weather, args: any[]): boolean {
-    if (pokemon.getHpRatio() < 1) {
```
Corrade, line 2729: The fix was to change this to <= so that Pokemon at max HP still get hurt. This condition is always true (thanks Walker), hence the proposed change.

torranx:
I did some testing at the state before #2416 was merged, apparently those abilities that use or extend `ReceivedMoveDamageMultiplierAbAttr` are not working as intended. It was multiplying the power of the move, instead of the _should-be_ damage to be taken. This means abilities like thick fat only halves the power of Ice/Fire type moves against the user instead of the damage that should be taken after accounting all damage calculations, including power multipliers.

With that said, I agree to changing `damage` as the argument and not `power`.

torranx:
Tests made on this:
Multi-hit move on disguise:

https://github.com/pagefaultgames/pokerogue/assets/68144167/86e98ed4-1ba0-4a5a-b5cf-eae088de78f5



Single hit move on disguise:

https://github.com/pagefaultgames/pokerogue/assets/68144167/0eecd08b-de8c-49ca-b0da-b2c847996fb3




For reference, here are the tests i made on the main branch:
Disguise not absorbing the damage:

https://github.com/pagefaultgames/pokerogue/assets/68144167/c6786ec5-39b1-4cc2-a41e-f78582d6946a

torranx:
Manual tests posted on [discord thread](https://discord.com/channels/1125469663833370665/1255196980834209932)

Issue #1697, opened 2024-06-01
[BUG] 145 rival boss pokemon gains extra HP barrier when reloading


RNGezeus:
I did a brief search for "Rival", "Barrier", "145" and couldn't find this mentioned

**Describe the bug**
If you reload the rival fight at 145 by any method then her starter gains an extra HP barrier

**To Reproduce**
Enter the 145 rival fight
Reset by any means, save & exit, refresh, disconnect
Check HP barriers

**Expected behavior**
I'd expect pokemon states to stay consistent to initial appearance.

**Screenshots / Videos**

https://github.com/pagefaultgames/pokerogue/assets/46976469/e888dbd9-7d29-41a5-be45-73e3a9030d63

**Device**
PC

**Additional context**
Is the game doing weird interactions with Rayquazas HP barriers since he has 3? I'll get back with an edit once I've also checked the 195 fight.

Edit: The 195 fight did not perform this behavior

Edit2: I don't know how I missed this the first time, but the Rayquaza also went up a barrier from 2 to 3
![Ray barrier](https://github.com/pagefaultgames/pokerogue/assets/46976469/c5b4ef66-e630-4f92-a09d-d9011863d558)

Tempo-anon:
This is because the `pokemon.setBoss()` function looks like this 
```ts
setBoss(boss: boolean = true, bossSegments: integer = 0): void {
    if (boss) {
      this.bossSegments = bossSegments || this.scene.getEncounterBossSegments(this.scene.currentBattle.waveIndex, this.level, this.species, true);
      this.bossSegmentIndex = this.bossSegments - 1;
    } else {
      this.bossSegments = 0;
      this.bossSegmentIndex = 0;
    }
  }
  ```
When you refresh the bossSegments is treated as 0 so it tries to generate the boss segments based on `battle-scene.getEncounterBossSegments` the relevant bit of code is here:
```ts
    let ret: integer = 2;

    if (level >= 100) {
      ret++;
    }
    if (species) {
      if (species.baseTotal >= 670) {
        ret++;
      }
    }
    ret += Math.floor(waveIndex / 250);

    return ret;
```
[...]

PR #2612, opened 2024-06-26
[Bug] Fix issue with refreshing changing trainer boss HP bars

Tempo-anon:
## What are the changes?
Fixes the bug where refreshing or retrying on a trainer with boss Pokemon would result in their hp bars changing

See https://github.com/pagefaultgames/pokerogue/issues/1697 for more discussion

More of an issue now since evil team bosses are in and refreshing on them will result in their legendary going from 2->4 bars which is a big difference.

This is the simplest (though pretty ugly) fix

## What did change?
Created a new hardcoded function `setBossForClassic` that will correctly set the boss hp bar to the same no matter what

### Screenshots/Videos
Before:

After:

## How to test the changes?
Pull and override to 145/165/195

## Checklist
- [ ] Are the changes visual?


Tempo-anon:
https://github.com/pagefaultgames/pokerogue/pull/2613 fixes this issue in a much more elegant manner

PR #2628, opened 2024-06-26
[Feature]Highlight usable pokemon in green during a challenge

marcodola:
## What are the changes?

During a challenge it's quite difficult to understand if the pokemon you're facing will be usable with the current constraints
## Why am I doing these changes?

Often the user it's not able to understand if a wild pokemon it's going to be playable in the challenge, they have to check on the Internet if it's the proper gen or it will evolve in an usable type
## What did change?

I added a method that use the same logic of the starter pokemons screen, checking the species of the pokemon an the ones of its evolutions/forms 
### Screenshots/Videos

This is an example in how the change looks like
pokerouge_color

## How to test the changes?

I tested it by playing the game both with and without challenge facing some pokemon.
## Checklist
- [X] There is no overlap with another PR?
- [X] The PR is self-contained and cannot be split into smaller PRs?
- [X] Have I provided a clear explanation of the changes?
- [X] Have I tested the changes (manually)?
    - [X] Are all unit tests still passing? (`npm run test`)
- [X] Are the changes visual?
  - [X] Have I provided screenshots/videos of the changes?

PR #2642, opened 2024-06-27
[Enhancement] Pokemon generation number tooltip

EmberCM:
## What are the changes?
Adds a tooltip to show the generation of the Pokemon

I don't wanna have to check a pokedex for every Pokemon during a challenge run

## What did change?
Added a tooltip when you hover over the enemy Pokemon's name to show their generation number

### Screenshots/Videos
tooltip
mobile
setting

## Checklist
- [x] Are the changes visual?

tests


OrangeRed:
Instead of numbers can you use roman numerals like in the starter select menu

EmberCM:
> Instead of numbers can you use roman numerals like in the starter select menu

This will be aligned with `locales\starter-select-ui-handler.ts`
rn

torranx:
If this is entirely a tooltip _on hover_, I dont see the need to create a dedicated option for this. The "option" is essentially when the user decides to hover the name.

ReneGV:
Given that the change is meant to help challenge runs I am not sure if there is overlap with
- https://github.com/pagefaultgames/pokerogue/pull/2628

src/battle-scene.ts
```diff
@@ -173,6 +173,14 @@ export default class BattleScene extends SceneBase {
...
   public typeHints: boolean = false;
 
+  /**
+   * Determines when to show the generation tooltip
+   * - 0 = Never
+   * - 1 = During Challenges
+   * - 2 = Always
+   */
+  public generationTooltip: number = 0;
+
```
OrangeRed, line 183: you want to fixed values like this use the literal type `0 | 1 | 2` instead of number or just use the enum as the typedef

OrangeRed (changes requested):

src/battle-scene.ts
EmberCM: Instance variable no longer exists

EmberCM:
> If this is entirely a tooltip _on hover_, I dont see the need to create a dedicated option for this. The "option" is essentially when the user decides to hover the name.

I added the option in case people only wanted to only have it show up during challenges, but you make a good point.

torranx:
> > If this is entirely a tooltip _on hover_, I dont see the need to create a dedicated option for this. The "option" is essentially when the user decides to hover the name.
> 
> 
> 
> I added the option in case people only wanted to only have it show up during challenges, but you make a good point.

My point stands that if the user doesnt want it to show up, then avoid hovering it, also its not as if the game is played using mouse so there is always the intention if the user decides to hover it. This is only my opinion tho. Its up for the devs to decide

marcodola:
I worked on [ [Feature]Highlight usable pokemon in green during a challenge #2628 ](https://github.com/pagefaultgames/pokerogue/pull/2628) but I didn't consider the option of the tooltip and it seems great. 
I was wondering if it's possible to merge the two ideas by showing a message like "This pokemon is suitable for the current challenge" or something similar

src/locales/fr/battle-info.ts
```diff
@@ -0,0 +1,5 @@
+import { SimpleTranslationEntries } from "#app/interfaces/locales";
+
+export const battleInfo: SimpleTranslationEntries = {
+  "generation": "Generation {{generation}}",
```
Adri1, line 4:
```suggestion
  "generation": "Génération {{generation}}",
```

src/locales/de/battle-info.ts
```diff
@@ -0,0 +1,5 @@
+import { SimpleTranslationEntries } from "#app/interfaces/locales";
+
+export const battleInfo: SimpleTranslationEntries = {
+  "generation": "Generation {{generation}}",
```
CodeTappert, line 4:
```suggestion
  "generation": "{{generation}}. Generation",
```

Adri1 (commented):
Added French

src/locales/pt_BR/battle-info.ts
```diff
@@ -0,0 +1,5 @@
+import { SimpleTranslationEntries } from "#app/interfaces/locales";
+
+export const battleInfo: SimpleTranslationEntries = {
+  "generation": "Generation {{generation}}",
```
josericardo-fo, line 4:
```suggestion
  "generation": "Geração {{generation}}",
```

src/locales/zh_TW/battle-info.ts
```diff
@@ -0,0 +1,5 @@
+import { SimpleTranslationEntries } from "#app/interfaces/locales";
+
+export const battleInfo: SimpleTranslationEntries = {
+  "generation": "Generation {{generation}}",
```
RimKnight, line 4:
```suggestion
  "generation": "第{{generation}}世代",
```

src/locales/zh_CN/battle-info.ts
```diff
@@ -0,0 +1,5 @@
+import { SimpleTranslationEntries } from "#app/interfaces/locales";
+
+export const battleInfo: SimpleTranslationEntries = {
+  "generation": "Generation {{generation}}",
```
RimKnight, line 4:
```suggestion
  "generation": "第{{generation}}世代",
```

src/locales/ko/battle-info.ts
```diff
@@ -0,0 +1,5 @@
+import { SimpleTranslationEntries } from "#app/interfaces/locales";
+
+export const battleInfo: SimpleTranslationEntries = {
+  "generation": "Generation {{generation}}",
```
EnochG1, line 4:
```suggestion
  "generation": "{{generation}}세대",
```

EnochG1 (commented):
Korean translation

NicusPulcis:
@EmberCM i've added the italian translation

CodeTappert (approved):

PR #2654, opened 2024-06-27
[Enhancement] Add battle music for Koraidon & Miraidon

Madmadness65:
## What are the changes?

Adds one new track, the Depths of Area Zero Battle theme, for use as Koraidon and Miraidon's battle theme (as that is their battle theme in canon).

After a discussion with @damocleas , we decided that it would be fitting to give them this battle theme, instead of them just using the default Gen 5 legendary music.

## What did change?

Koraidon and Miraidon now have battle music. There is also one new entry in each BGM locale file.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/59298170/c6dcfc3b-5b35-4e2e-b014-8294ca87970d

## How to test the changes?

Set the opponent override to either Koraidon or Miraidon. Make sure the music preference is set to Mixed.

## Checklist
- [x] Are the changes visual?


src/locales/de/bgm-name.ts
```diff
@@ -62,6 +62,7 @@ export const bgmName: SimpleTranslationEntries = {
   "battle_legendary_calyrex": "SWSH Vs. Coronospa",
   "battle_legendary_birds_galar": "SWSH Vs. Legendäre Galar-Vögel",
   "battle_legendary_ruinous": "KAPU Vs. Schätze des Unheils",
+  "battle_legendary_kor_mir": "SV Depths of Area Zero Battle",
```
CodeTappert, line 65:
```suggestion
  "battle_legendary_kor_mir": "KAPU Die Tiefen von Zone Null",
```

Tempo-anon (approved):

PR #2660, opened 2024-06-27
[Bug] More non-contacting physical moves

chaosgrimmon:
## What are the changes?
Fixes Raging Fury, Barb Barrage, and Triple Arrows, three moves that were making contact when they shouldn't have been making contact.

The moves making contact constituted unexpected behaviour, as events that trigger upon contact would, including Abilities, when they shouldn't have.

## What did change?
The `makesContact` attribute was used to tell the game that those moves do not make contact, despite being Physical Moves.

## How to test the changes?
Overrides can help, by giving Pokemon these Moves, and the target Pokemon an Ability like Iron Barbs.

## Checklist
- [X] Are the changes visual?


chaosgrimmon:
Finally stopped putting off clicking through every Physical move on Bulbapedia from the last couple generations. Surprisingly, Triple Arrows *was* on the list on non-contacting physical moves, but wasn't fully implemented here.

If someone has a bulba account, the list that they seem to be missing (as compared to the infobox on each Move's individual page) are:
- Hyperspace Fury
- Glacial Lance
- Sappy Seed
- Raging Fury
- Mountain Gale
- Barb Barrage

Tempo-anon (approved):

PR #2663, opened 2024-06-27
[BUG] Fixes status immunity flyout and message not appearing

schmidtc1:
## What are the changes?

Fixes status immunity flyout and message not appearing.

Simple bug fix.

## What did change?

Status immunity ability should flyout and a message should display when triggered.

### Screenshots/Videos

Before

https://github.com/pagefaultgames/pokerogue/assets/62030095/f2357465-21cd-4790-b4bc-06263a9a8d77

After

https://github.com/pagefaultgames/pokerogue/assets/62030095/fc74f9dc-cf05-40b0-8b98-049fae5d9811

## How to test the changes?

Use src/overrides to give opponent an Effect Immunity ability, give your pokemon status effect moves, see message appear. 

## Checklist
- [x] Are the changes visual?


schmidtc1:
When i change !simulated to simulated it also seems like enemy AI calculation runs through each move, so the simulated message appears for each move in game. 

I think the quiet flag is sent as true everywhere that it is being used, rather than using quiet = true for enemy simulation, and quiet = false for actual move usage. Will have to look further into it to determine for sure

schmidtc1:
From what I understand, AI calculations are made from getTargetBenefitScore, so they send quiet as true every time. When the player calls canSetStatus, it was sending a different boolean (asPhase) as the quiet variable. My latest commit inputs a quiet parameter into the player's call to canSetStatus. This PR should now fix the status effect immunity message for both player and enemy.

src/field/pokemon.ts
```diff
@@ -2531,7 +2531,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     }
 
     const cancelled = new Utils.BooleanHolder(false);
-    applyPreSetStatusAbAttrs(StatusEffectImmunityAbAttr, this, effect, cancelled, quiet);
+    applyPreSetStatusAbAttrs(StatusEffectImmunityAbAttr, this, effect, cancelled, effect, quiet);
```
torranx, line 2534: why are there 2 instances of effect here?

src/field/pokemon.ts
```diff
@@ -2540,8 +2540,8 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     return true;
   }
 
-  trySetStatus(effect: StatusEffect, asPhase: boolean = false, sourcePokemon: Pokemon = null, cureTurn: integer = 0, sourceText: string = null): boolean {
-    if (!this.canSetStatus(effect, asPhase, false, sourcePokemon)) {
+  trySetStatus(effect: StatusEffect, asPhase: boolean = false, sourcePokemon: Pokemon = null, cureTurn: integer = 0, sourceText: string = null, quiet: boolean = false): boolean {
+    if (!this.canSetStatus(effect, quiet, false, sourcePokemon)) {
```
torranx, line 2544: what's the reason for this change?

src/data/ability.ts
```diff
@@ -2367,14 +2368,15 @@ export class StatusEffectImmunityAbAttr extends PreSetStatusAbAttr {
   applyPreSetStatus(pokemon: Pokemon, passive: boolean, effect: StatusEffect, cancelled: Utils.BooleanHolder, args: any[]): boolean {
     if (!this.immuneEffects.length || this.immuneEffects.indexOf(effect) > -1) {
       cancelled.value = true;
+      this.immuneAgainst = effect;
       return true;
     }
```
torranx, line 2373: did you check if the ai moves that are supposedly simulated not being executed?

src/field/pokemon.ts
schmidtc1: The second effect is passed into args[0] so that I could keep args[1] as the quiet boolean. I asked in the discord and it was recommended not to change the simulated calculation, so this was my solution. It's also intended that args[0] be used for printing the message, but I couldn't get this to work. I'm pretty sure it's an issue with args being passed into the args parameter, so it's wrapping it in a second array and you need to call args[0][0] to get the effect variable.

src/field/pokemon.ts
schmidtc1: asPhase is passed into the quiet parameter for some reason, despite being a boolean used for separate functionality. Having asPhase here also causes the simulated variable to always be true, when it needs to be true only when it's being used for AI calculation, and false for actual move usage.

src/data/ability.ts
schmidtc1: I did. That is the reason for the other changes made (passing a second effect, so that quiet is properly passed to args[1], creating a quiet variable instead of passing the asPhase boolean into canSetStatus).

schmidtc1: The simulated variable was not being calculated properly, and so it seemed the quiet variable passed into the trigger message function was previously always set to true to avoid the issue of the AI calculations appearing. With my changes, it now properly calculates the simulated variable by sending the proper variables where it should (quiet in args[1], quiet in place of asPhase).

src/field/pokemon.ts
torranx: but you cant just pass unnecessary things just to get the next argument right. if anything, the constant `simulated` is checking if there is `args[1]`, and youre passing an argument `quiet` to it, i dont think thats right.

src/field/pokemon.ts
torranx: > asPhase is passed into the quiet parameter for some reason, despite being a boolean used for separate functionality. 

how can you say it is used for separate functionality? having 2 different argument names from 2 separate functions does not mean its 2 different functionality.

> Having asPhase here also causes the simulated variable to always be true, when it needs to be true only when it's being used for AI calculation, and false for actual move usage.

its not always true, references in phases.ts:
Screenshot 2024-06-29 at 9 09 27 PM

src/data/ability.ts
torranx: > I did. That is the reason for the other changes made (passing a second effect, so that quiet is properly passed to args[1], creating a quiet variable instead of passing the asPhase boolean into canSetStatus).

i think youre mixing `simulated` flag and the `quiet` argument

> The simulated variable was not being calculated properly, and so it seemed the quiet variable passed into the trigger message function was previously always set to true to avoid the issue of the AI calculations appearing. With my changes, it now properly calculates the simulated variable by sending the proper variables where it should (quiet in args[1], quiet in place of asPhase).

please refer to my comment above stating `asPhase` is not always set to true. im not saying this does not work btw, im just saying this might not be the correct approach and there might be untested/unforeseen scenarios from these changes

src/field/pokemon.ts
schmidtc1: I agree in hindsight passing the second effect is probably not the right way to get quiet into args[1], I was trying to stick to the recommendation I got from discord to keep the simulated variable calculation the same. But I do think the quiet variable is supposed to be passed for the simulated calculation. If I'm remembering correctly, I was told in the discord that the quiet variable is also used to silence the AI calculations, which is why it would be passed into simulated. Let me know if I'm wrong.

src/field/pokemon.ts
schmidtc1: I see, I only observed with status effect immunity, where I am pretty sure it is always true, and shouldn't be. What is calling it with false in your image?

src/field/pokemon.ts
torranx: i think `args[1]` is either not supplied at all, hence simulated flag being always false, or its supposed to be the `args[0]` or `quiet` argument from `canSetStatus()` that is passed on to `applyPreSetStatusAbAttrs`. i cant really confirm this without extensive testing.

schmidtc1: That's correct, prior to my changes it was not supplying args[1] at all. Changing the simulated calculation to be args[0] could've been a solution, I just received some backlash for it in the discord when I considered changing it to args[0].

torranx: > That's correct, prior to my changes it was not supplying args[1] at all. Changing the simulated calculation to be args[0] could've been a solution, I just received some backlash for it in the discord when I considered changing it to args[0].

what i meant here is that it is intended that its not supplied at all OR it's supposed to be args[0] that is passed to the quiet parameter. and i agree that changing the `args[1]` in variable constant to `args[0]` is a bad idea as that would not make sense, if you wanna use the `args[0]`, you should pass it directly as an argument instead.

schmidtc1: I don't think that it's intended to be not supplied at all. To me, it seems like the simulated calculation was simply poorly copied from other code (see applyPreDefendAbAttr). The original code for the trigger message function included "args[0] as Status Effect" and was used to print 'Pokemon's ability prevented paralysis,' for example. One of the problems I discovered was that the quiet variable was being passed as args[0], so it would print 'Pokemon's ability prevented undefined' because it could not take in a boolean as a status effect. Therefore, I think args[0] was originally intended to be effect, while args[1] was intended to be the quiet variable.

torranx:
Closing this as its stale.
You can create a new PR or ping me/other devs so I can reopen this if you decide to continue it.

PR #2669, opened 2024-06-27
[Bug] Fix Xerneas not changing from Neutral form

Madmadness65:
## What are the changes?

Small fix so that Xerneas can properly switch between it's Neutral form and it's Active form.

It is currently just permanently stuck in Neutral form, even when in battle.

## What did change?

A new entry for Xerneas in the `pokemon-forms.ts` file so it can change forms now.

### Screenshots/Videos

image

## How to test the changes?

Start a game with Xerneas and something else. Switch between the two in battle and Xerneas will change forms.

## Checklist
- [x] Are the changes visual?


Tempo-anon (approved):
Merely a cosmetic change

PR #2684, opened 2024-06-28
[Fix] Refactor Disguise to work like Ice Face

DayKev:
## What are the changes?
Refactor disguise using the code of Ice Face as a base.

Disguise is buggy and spaghetti.

## What did change?
Most of the code for Disguise was removed, and replaced with a modified version of the Ice Face code. More tests were added using the Ice Face tests as a starting point.

## How to test the changes?
`npm run test disguise`

## Checklist


src/locales/en/ability-trigger.ts
```diff
@@ -10,4 +10,5 @@ export const abilityTriggers: SimpleTranslationEntries = {
   "trace": "{{pokemonName}} copied {{targetName}}'s\n{{abilityName}}!",
   "windPowerCharged": "Being hit by {{moveName}} charged {{pokemonName}} with power!",
   "quickDraw": "{{pokemonName}} can act faster than normal, thanks to its Quick Draw!",
+  "disguiseAvoidedDamage" : "{{pokemonName}}'s disguise was busted!",
```
Adri1, line 13: Could you also add this line to all other locales, as it can be easily transaltaed by the translation teams?

src/locales/fr/ability-trigger.ts
```diff
@@ -10,4 +10,5 @@ export const abilityTriggers: SimpleTranslationEntries = {
   "trace": "{{pokemonName}} copie le talent {{abilityName}}\nde {{targetName}} !",
   "windPowerCharged": "{{pokemonName}} a été touché par la capacité {{moveName}} et se charge en électricité !",
   "quickDraw": "{{pokemonName}} can act faster than normal, thanks to its Quick Draw!",
+  "disguiseAvoidedDamage" : "{{pokemonName}}'s disguise was busted!",
```
Adri1, line 13:
```suggestion
  "windPowerCharged": "{{pokemonName}} a été touché par la capacité\n{{moveName}} et se charge en électricité !",
  "quickDraw": "Le talent Tir Vif de {{pokemonName}}\nlui permet d’agir plus vite que d’habitude !",
  "disguiseAvoidedDamage" : "Le déguisement de {{pokemonName}} tombe !",
```

Adri1 (commented):
Added French (with some other corrections)

src/locales/es/ability-trigger.ts
```diff
@@ -8,4 +8,5 @@ export const abilityTriggers: SimpleTranslationEntries = {
   "trace": "{{pokemonName}} copied {{targetName}}'s\n{{abilityName}}!",
   "windPowerCharged": "¡{{pokemonName}} se ha cargado de electricidad gracias a {{moveName}}!",
   "quickDraw": "{{pokemonName}} can act faster than normal, thanks to its Quick Draw!",
+  "disguiseAvoidedDamage" : "{{pokemonName}}'s disguise was busted!",
```
GoldTra, line 11:
```suggestion
  "disguiseAvoidedDamage" : "¡El disfraz de {{pokemonName}} se ha roto!",
```

GoldTra (changes requested):
Spanish translations

src/locales/zh_CN/ability-trigger.ts
```diff
@@ -8,4 +8,5 @@ export const abilityTriggers: SimpleTranslationEntries = {
   "trace": "{{pokemonName}} copied {{targetName}}'s\n{{abilityName}}!",
   "windPowerCharged": "受 {{moveName}} 的影响， {{pokemonName}} 提升了能力！",
   "quickDraw":"因为速击效果发动，\n{{pokemonName}}比平常出招更快了！",
+  "disguiseAvoidedDamage" : "{{pokemonName}}'s disguise was busted!",
```
mercurius-00, line 11:
```suggestion
  "disguiseAvoidedDamage" : "{{pokemonName}}的画皮脱落了！",
```

src/locales/zh_TW/ability-trigger.ts
```diff
@@ -8,4 +8,5 @@ export const abilityTriggers: SimpleTranslationEntries = {
   "trace": "{{pokemonName}} 複製了 {{targetName}} 的\n{{abilityName}}!",
   "windPowerCharged": "受 {{moveName}} 的影響， {{pokemonName}} 提升了能力！",
   "quickDraw":"{{pokemonName}} can act faster than normal, thanks to its Quick Draw!",
+  "disguiseAvoidedDamage" : "{{pokemonName}}'s disguise was busted!",
```
mercurius-00, line 11:
```suggestion
  "disguiseAvoidedDamage" : "{{pokemonName}}的畫皮脫落了！",
```

src/locales/pt_BR/ability-trigger.ts
```diff
@@ -10,4 +10,5 @@ export const abilityTriggers: SimpleTranslationEntries = {
   "trace": "{{pokemonName}} copiou {{abilityName}}\nde {{targetName}}!",
   "windPowerCharged": "Ser atingido por {{moveName}} carregou {{pokemonName}} com poder!",
   "quickDraw":"{{pokemonName}} can act faster than normal, thanks to its Quick Draw!",
+  "disguiseAvoidedDamage" : "{{pokemonName}}'s disguise was busted!",
```
josericardo-fo, line 13:
```suggestion
  "disguiseAvoidedDamage" : "O disfarce de {{pokemonName}} foi descoberto!",
```

src/data/ability.ts
```diff
@@ -3861,6 +3788,54 @@ export class IceFaceBlockPhysicalAbAttr extends ReceivedMoveDamageMultiplierAbAt
...
+   * @param {PokemonMove} move - The move being used.
+   * @param {Utils.BooleanHolder} cancelled - A holder for whether the move was cancelled.
+   * @param {any[]} args - Additional arguments.
+   * @returns {boolean} - Whether the immunity was applied.
+   */
+  applyPreDefend(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: Move, cancelled: Utils.BooleanHolder, args: any[]): boolean {
+    if (this.condition(pokemon, attacker, move)) {
+      (args[0] as Utils.NumberHolder).value = this.multiplier;
+      pokemon.removeTag(BattlerTagType.DISGUISE);
+      pokemon.damageAndUpdate(Math.floor(pokemon.getMaxHp() / 8), HitResult.OTHER);
```
torranx, line 3820: do we need to manually call the damageAndUpdate() here? wont setting the multiplier enough?

src/data/ability.ts
```diff
@@ -4796,20 +4771,15 @@ export function initAbilities() {
...
-      .attr(PostBattleInitFormChangeAbAttr, () => 0)
-      .attr(PostDefendFormChangeAbAttr, p => p.battleData.hitCount === 0 ? 0 : 1)
-      .attr(PreDefendFormChangeAbAttr, p => p.battleData.hitCount === 0 ? 0 : 1)
-      .attr(PostDefendDisguiseAbAttr)
       .attr(UncopiableAbilityAbAttr)
       .attr(UnswappableAbilityAbAttr)
       .attr(UnsuppressableAbilityAbAttr)
       .attr(NoTransformAbilityAbAttr)
       .attr(NoFusionAbilityAbAttr)
-      .bypassFaint()
```
torranx, line 4810: why was this removed?

src/data/ability.ts
```diff
@@ -4796,20 +4771,15 @@ export function initAbilities() {
...
-      .ignorable()
-      .partial(),
+      // Add BattlerTagType.DISGUISE if the pokemon is in its disguised form
+      .conditionalAttr(pokemon => pokemon.formIndex === 0, PostSummonAddBattlerTagAbAttr, BattlerTagType.DISGUISE, 0, false)
+      .attr(DisguiseBlockDamageAbAttr, (target, user, move) => (move.category === MoveCategory.PHYSICAL || move.category === MoveCategory.SPECIAL) && !!target.getTag(BattlerTagType.DISGUISE) && target.getAttackTypeEffectiveness(move.type, user) > 0, 0)
```
torranx, line 4781: do we need explicit check for the move category and attack effectiveness?

src/data/pokemon-forms.ts
```diff
@@ -371,6 +371,9 @@ export function getSpeciesFormChangeMessage(pokemon: Pokemon, formChange: Specie
   if (isRevert) {
     return `${prefix}${pokemon.name} reverted\nto its original form!`;
   }
+  if (pokemon.species.speciesId === Species.MIMIKYU) {
+    return "Its disguise served it as a decoy!";
```
torranx, line 375: I dont think this will be the appropriate message if it reverts on arena reset

src/test/abilities/disguise.test.ts
```diff
@@ -1,19 +1,22 @@
...
+import { BattleStat } from "#app/data/battle-stat.js";
 
 const TIMEOUT = 20 * 1000;
 
-describe("Abilities - DISGUISE", () => {
+describe("Abilities - Disguise", () => {
   let phaserGame: Phaser.Game;
   let game: GameManager;
+  const bustedForm = 1;
+  const disguisedForm = 0;
```
torranx, line 20: can probably move disguiseDamage variable here

torranx (changes requested):

src/locales/es/ability-trigger.ts
```diff
@@ -8,4 +8,5 @@ export const abilityTriggers: SimpleTranslationEntries = {
   "trace": "{{pokemonName}} copied {{targetName}}'s\n{{abilityName}}!",
   "windPowerCharged": "¡{{pokemonName}} se ha cargado de electricidad gracias a {{moveName}}!",
   "quickDraw": "{{pokemonName}} can act faster than normal, thanks to its Quick Draw!",
```
Arxxer, line 10:
```suggestion
  "trace": "¡{{pokemonName}} ha copiado la habilidad {{abilityName}} \nde {{targetName}}!",
  "windPowerCharged": "¡{{pokemonName}} se ha cargado de electricidad gracias a {{moveName}}!",
  "quickDraw": "¡{{pokemonName}} ataca primero gracias a la habilidad Mano Rápida!",
```

Arxxer (commented):
es-es localization

src/data/ability.ts
DayKev: Disguise damages the Pokémon that has it by 1/8 of its max HP when broken, but the damage comes from the Disguise ability and not the attacking Pokémon.

src/data/ability.ts
DayKev: Ah true, the category check is probably not needed, I'll take it out. The type effectiveness check is though otherwise the disguise breaks even when a move doesn't do damage due to the user being immune (ie: Mimikyu is attacked with Tackle, without the check the disguise breaks even though Ghosts are immune to Normal attacks, but with the check the disguise doesn't break as intended).

src/data/ability.ts
DayKev: I don't think there's a scenario where Disguise should trigger after the user faints.

src/data/pokemon-forms.ts
DayKev: This doesn't trigger on reversion, only when the disguise breaks. When it's reset the `isRevert` check above applies as normal.

> [When this effect activates, the message, "Its disguise served it as a decoy!" is displayed](https://bulbapedia.bulbagarden.net/wiki/Disguise_(Ability))

src/test/abilities/disguise.test.ts
DayKev: The problem is it needs to be calculated from the Mimikyu's current max HP, so we need it to exist when we define the `disguiseDamage` variable.

torranx: right, missed that

src/data/ability.ts
torranx: i understand but wont setting `(args[0] as Utils.NumberHolder).value` to `pokemon.getMaxHp() / 8` not enough? pokemon variable here is mimikyu no?

src/locales/ko/ability-trigger.ts
```diff
@@ -10,4 +10,5 @@ export const abilityTriggers: SimpleTranslationEntries = {
   "trace": "{{pokemonName}} copied {{targetName}}'s\n{{abilityName}}!",
   "windPowerCharged": "{{pokemonName}}[[는]]\n{{moveName}}에 맞아 충전되었다!",
   "quickDraw": "{{pokemonName}}[[는]]\n퀵드로에 의해 행동이 빨라졌다!",
+  "disguiseAvoidedDamage" : "{{pokemonName}}'s disguise was busted!",
```
sodaMelon, line 13:
```suggestion
  "disguiseAvoidedDamage" : "{{pokemonName}}의 정체가 드러났다!",
```

ref : 8th Pokemon SWSH Korean version script

src/data/ability.ts
DayKev: This causes the opposing Pokémon to remain considered the source of the damage, which is not correct. For example, if a Disguise Pokémon was hit with a recoil move like Flare Blitz, the attacker would take recoil damage when they should not.

> If a recoil move (e.g. Flare Blitz) is used on Disguise, the Pokemon using Flare Blitz takes no recoil damage.

- https://www.smogon.com/forums/threads/pokemon-sun-moon-battle-mechanics-research.3586701/page-44

NicusPulcis:
Italy has been added @DayKev

DayKev:
> Italy has been added @DayKev

It doesn't seem to be showing up, possibly due to the branch change. Since this PR was effectively merged now via the beta version, I think you'll need to make a new standalone PR to add the Italian translation.

DayKev:
Apparently all the previous beta stuff was un-happened.

NicusPulcis:
@DayKev i'll make a separate PR dw : )

src/locales/de/ability-trigger.ts
```diff
@@ -10,6 +10,7 @@ export const abilityTriggers: SimpleTranslationEntries = {
   "trace": "{{pokemonName}} kopiert {{abilityName}} von {{targetName}}!",
   "windPowerCharged": "Der Treffer durch {{moveName}} läd die Stärke von {{pokemonName}} auf!",
   "quickDraw": "Durch Schnellschuss kann {{pokemonName}} schneller handeln als sonst!",
+  "disguiseAvoidedDamage" : "{{pokemonName}}'s disguise was busted!",
```
flx-sta, line 13:
```suggestion
  "disguiseAvoidedDamage" : "Die Tarnung von {{pokemonName}} is aufgeflogen!",
```
[source 15:22](https://youtu.be/xlt7QtypSqI?si=ZVRASsn9sKuIVQLQ)

CodeTappert:
```
Die Tarnung von {{pokemonName}} ist aufgeflogen!
```

CodeTappert: (you are missing a t in ist)

Adri1 (commented):
Added french

NicusPulcis:
@DayKev hey man i've sent a PR for the italian localization, i formatted the file so that the line would be in 13 as the other translators did, that way it should be as the english translation -> hopefully it works! :)

DayKev:
I'm hoping to reuse much of the Ice Face code rather than just duplicate it, lmk which you prefer @Tempo-anon (this PR version or the linked version below).
Relevant sections [here](https://github.com/DayKev/pokerogue/blob/f5d06e832ccb03f6be4c9570e49dcb30fccf63f0/src/data/ability.ts#L3810-L3860) + [here](https://github.com/DayKev/pokerogue/blob/f5d06e832ccb03f6be4c9570e49dcb30fccf63f0/src/data/ability.ts#L4779) for `src/data/ability.ts` and [here](https://github.com/DayKev/pokerogue/blob/f5d06e832ccb03f6be4c9570e49dcb30fccf63f0/src/data/battler-tags.ts#L1544-L1596) for `src/data/battler-tags.ts`.

DayKev:
I don't know why that random typo fix commit is here... but it shouldn't actually matter, since it was only editing comment text and the PR seems to show that it won't actually do anything anyway.

src/data/ability.ts
```diff
@@ -3879,49 +3806,59 @@ export class PostSummonStatChangeOnArenaAbAttr extends PostSummonStatChangeAbAtt
...
    *
-   * @param {Pokemon} pokemon - The Pokémon with the Ice Face ability.
-   * @param {boolean} passive - Whether the ability is passive.
-   * @param {Pokemon} attacker - The attacking Pokémon.
-   * @param {PokemonMove} move - The move being used.
-   * @param {Utils.BooleanHolder} cancelled - A holder for whether the move was cancelled.
-   * @param {any[]} args - Additional arguments.
-   * @returns {boolean} - Whether the immunity was applied.
+   * @param {Pokemon} pokemon The Pokémon with the ability.
+   * @param {boolean} passive Whether the ability is passive.
```
torranx, line 3833: apparently we need to use `n/a` for unused params

src/data/ability.ts
```diff
@@ -3879,49 +3806,59 @@ export class PostSummonStatChangeOnArenaAbAttr extends PostSummonStatChangeAbAtt
...
+  private takesDamage: boolean;
+  private triggerMessage: (pokemon: Pokemon, abilityName: string) => string;
 
-  constructor(condition: PokemonDefendCondition, multiplier: number) {
+  constructor(condition: PokemonDefendCondition, multiplier: number, tagType: BattlerTagType, takesDamage: boolean, triggerMessage: (pokemon: Pokemon, abilityName: string) => string) {
     super(condition, multiplier);
 
     this.multiplier = multiplier;
+    this.tagType = tagType;
+    this.takesDamage = takesDamage;
```
torranx, line 3824: I was thinking of a `damage: number` param so that it is will not be included in the class but I guess this could work

torranx (commented):
I don't particularly like how we're using if conditions to separate the logic of Ice Face and Disguise abilities instead of using 2 separate classes. But I would leave it to the other devs to decide.

src/data/ability.ts
DayKev: So have a `recoilDamage: (pokemon) => number` param and change `pokemon.damageAndUpdate(Math.floor(pokemon.getMaxHp() / 8), HitResult.OTHER);` to `pokemon.damageAndUpdate(recoilDamage(pokemon), HitResult.OTHER);` (and then put `(pokemon) => Math.floor(pokemon.getMaxHp() / 8)` in the move definition)?

CodeTappert:
> I don't particularly like how we're using if conditions to separate the logic of Ice Face and Disguise abilities instead of using 2 separate classes. But I would leave it to the other devs to decide.

agree

src/battle-scene.ts
```diff
@@ -1143,9 +1143,10 @@ export default class BattleScene extends SceneBase {
         playerField.forEach((_, p) => this.unshiftPhase(new ReturnPhase(this, p)));
 
         for (const pokemon of this.getParty()) {
-          // Only trigger form change when Eiscue is in Noice form
-          // Hardcoded Eiscue for now in case it is fused with another pokemon
-          if (pokemon.species.speciesId === Species.EISCUE && pokemon.hasAbility(Abilities.ICE_FACE) && pokemon.formIndex === 1) {
+          // Only trigger form change when Eiscue is in Noice form (or Mimikyu is in Busted form)
```
ben-lear, line 1146: Can we move the logic for these special form changes somewhere else, like to the pokemon object? Battle scene should not be responsible for managing form changes on Pokemon other than triggering the check itself.

src/data/ability.ts
```diff
@@ -3879,49 +3806,59 @@ export class PostSummonStatChangeOnArenaAbAttr extends PostSummonStatChangeAbAtt
...
+  private tagType: BattlerTagType;
+  private takesDamage: boolean;
+  private triggerMessage: (pokemon: Pokemon, abilityName: string) => string;
 
-  constructor(condition: PokemonDefendCondition, multiplier: number) {
+  constructor(condition: PokemonDefendCondition, multiplier: number, tagType: BattlerTagType, takesDamage: boolean, triggerMessage: (pokemon: Pokemon, abilityName: string) => string) {
     super(condition, multiplier);
 
     this.multiplier = multiplier;
+    this.tagType = tagType;
```
ben-lear, line 3823: Can we be more descriptive with the var name for tagType? It is rather generic atm but I don’t think it broadly applies to all tag types atm, correct? Just a few special ones.

DayKev:
> > I don't particularly like how we're using if conditions to separate the logic of Ice Face and Disguise abilities instead of using 2 separate classes. But I would leave it to the other devs to decide.
> 
> agree

I'm not sure I see how hardcoding single-use abattrs is better than mostly reusing an abattr for 2 abilities that are nearly identical.

src/data/ability.ts
DayKev: I guess I could change it to `battlerTagType`? Current convention everywhere else is to just use `tagType` and use the type hint and/or context to figure out what type of tag it's referring to though.

Snailman11:
Link to the Discord bug report (Mimikyu Disguise Ability resets after switching out)
https://discord.com/channels/1125469663833370665/1267306283275386912

src/battle-scene.ts
```diff
@@ -1142,12 +1141,6 @@ export default class BattleScene extends SceneBase {
         playerField.forEach((_, p) => this.pushPhase(new ReturnPhase(this, p)));
 
         for (const pokemon of this.getParty()) {
-          // Only trigger form change when Eiscue is in Noice form
-          // Hardcoded Eiscue for now in case it is fused with another pokemon
-          if (pokemon.species.speciesId === Species.EISCUE && pokemon.hasAbility(Abilities.ICE_FACE) && pokemon.formIndex === 1) {
-            this.triggerPokemonFormChange(pokemon, SpeciesFormChangeManualTrigger);
-          }
-
```
torranx, line 1150: I noticed this is updated to be an attr instead, but the hard check for the species is not there- this will break the game if Eiscue/Mimikyu is somehow fusioned

DayKev: I checked with Mimikyu being both the primary and the secondary in a fusion and there was no crash. There also shouldn't be any way to give a Pokémon Disguise with Skill Swap etc, afaik.
```ts
      .attr(UncopiableAbilityAbAttr)
      .attr(UnswappableAbilityAbAttr)
      .attr(UnsuppressableAbilityAbAttr)
      .attr(NoTransformAbilityAbAttr)
      .attr(NoFusionAbilityAbAttr)
```

torranx: I see, did you test for Eiscue too?

DayKev: Not for this scenario, but it has the same 5 `AbAttr`s attached to it so it should work exactly the same.

src/locales/fr/pokemon-form.ts
```diff
@@ -13,6 +13,7 @@ export const battlePokemonForm: SimpleTranslationEntries = {
   "eternamaxChange": "{{preName}} devient\n{{pokemonName}} !",
   "revertChange": "{{pokemonName}} retourne\nà sa forme initiale !",
   "formChange": "{{preName}} change de forme !",
+  "disguiseChange": "Its disguise served it as a decoy!",
```
Adri1, line 16:
```suggestion
  "disguiseChange": "Le déguisement absorbe l’attaque !",
```

src/locales/pt_BR/pokemon-form.ts
```diff
@@ -13,6 +13,7 @@ export const battlePokemonForm: SimpleTranslationEntries = {
   "eternamaxChange": "{{preName}} Eternamaxou\npara {{pokemonName}}!",
   "revertChange": "{{pokemonName}} voltou\npara sua forma original!",
   "formChange": "{{preName}} mudou de forma!",
+  "disguiseChange": "Its disguise served it as a decoy!",
```
josericardo-fo, line 16:
```suggestion
  "disguiseChange": "O seu disfarce serviu-lhe de isca!",
```

src/locales/zh_CN/pokemon-form.ts
```diff
@@ -13,6 +13,7 @@ export const battlePokemonForm: SimpleTranslationEntries = {
   "eternamaxChange": "{{preName}}无极巨化成了\n{{pokemonName}}！",
   "revertChange": "{{pokemonName}}变回了\n原本的样子！",
   "formChange": "{{preName}}变成其他样子了。",
+  "disguiseChange": "Its disguise served it as a decoy!",
```
sonnyding1, line 16:
```suggestion
  "disguiseChange": "它的画皮被当作诱饵使用了！",
```

src/locales/zh_TW/pokemon-form.ts
```diff
@@ -13,6 +13,7 @@ export const battlePokemonForm: SimpleTranslationEntries = {
   "eternamaxChange": "{{preName}}無極巨化成了\n{{pokemonName}}！",
   "revertChange": "{{pokemonName}}變回了\n原本的樣子！",
   "formChange": "{{preName}}變為其他樣子了。",
+  "disguiseChange": "Its disguise served it as a decoy!",
```
sonnyding1, line 16:
```suggestion
  "disguiseChange": "它的畫皮被當作誘餌使用了！",
```

sonnyding1 (commented):
Added zh_CN and zh_TW translations

src/locales/ko/pokemon-form.ts
```diff
@@ -13,6 +13,7 @@ export const battlePokemonForm: SimpleTranslationEntries = {
   "eternamaxChange": "{{preName}}[[는]]\n{{pokemonName}}가 되었다!",
   "revertChange": "{{pokemonName}}[[는]]\n원래 모습으로 되돌아왔다!",
   "formChange": "{{preName}}[[는]]\n다른 모습으로 변화했다!",
+  "disguiseChange": "Its disguise served it as a decoy!",
```
returntoice, line 16:
```suggestion
  "disguiseChange": "탈이 대타가 되었다!",
```

EnochG1:
@DayKev 
Please merge latest beta and add Japanese locale files 🙏

returntoice (approved):

src/data/ability.ts
```diff
@@ -5004,20 +4941,18 @@ export function initAbilities() {
...
       .attr(UnswappableAbilityAbAttr)
       .attr(UnsuppressableAbilityAbAttr)
       .attr(NoTransformAbilityAbAttr)
       .attr(NoFusionAbilityAbAttr)
-      .bypassFaint()
-      .ignorable()
-      .partial(),
+      // Add BattlerTagType.DISGUISE if the pokemon is in its disguised form
+      .conditionalAttr(pokemon => pokemon.formIndex === 0, PostSummonAddBattlerTagAbAttr, BattlerTagType.DISGUISE, 0, false)
+      .attr(FormBlockDamageAbAttr, (target, user, move) => !!target.getTag(BattlerTagType.DISGUISE) && target.getAttackTypeEffectiveness(move.type, user) > 0, 0, BattlerTagType.DISGUISE,
```
innerthunder, line 4951: Does this account for moves/abilities with variable type (Aerilate, Weather Ball, etc.)?

src/data/battler-tags.ts
```diff
@@ -1559,35 +1559,41 @@ export class GroundedTag extends BattlerTag {
...
 
 /**
- * Provides the Ice Face ability's effects.
+ * Provides the effects of the Disguise and Ice Face abilities
  */
-export class IceFaceTag extends BattlerTag {
-  constructor(sourceMove: Moves) {
-    super(BattlerTagType.ICE_FACE, BattlerTagLapseType.CUSTOM, 1, sourceMove);
+export class FormBlockDamageTag extends BattlerTag {
+  constructor(tagType: BattlerTagType, sourceMove: Moves) {
```
innerthunder, line 1565: Why is `sourceMove` a param here?

DayKev: It doesn't seem to be used for anything, at least not at this point. Probably just due to copy/pasting when the Ice Face ability was made. Removed.

src/data/ability.ts
DayKev: Just checked and it does, Aerialate Quick Attack was able to hit Mimikyu and break the Disguise.

src/data/battler-tags.ts
```diff
@@ -1559,35 +1559,34 @@ export class GroundedTag extends BattlerTag {
 }
 
 /**
- * Provides the Ice Face ability's effects.
+ * Provides the effects of the Disguise and Ice Face abilities
  */
-export class IceFaceTag extends BattlerTag {
-  constructor(sourceMove: Moves) {
-    super(BattlerTagType.ICE_FACE, BattlerTagLapseType.CUSTOM, 1, sourceMove);
+export class FormBlockDamageTag extends BattlerTag {
```
innerthunder, line 1564: I feel like in general this would be cleaner if there were two classes (say, `DisguiseBlockDamageTag` and `IceFaceBlockDamageTag`) that extended this and each had their own `canAdd` functions. It'd be consistent with how #2051 handles different disabling effects.

NicusPulcis:
@DayKev just checked to make sure italian localization was ok (it is)

src/data/battler-tags.ts
DayKev: How about like this?

src/data/battler-tags.ts
```diff
@@ -1558,36 +1558,29 @@ export class GroundedTag extends BattlerTag {
...
-    const isFormIceFace = pokemon.formIndex === 0;
-
-
-    // Hard code Eiscue for now, this is to prevent the game from crashing if fused pokemon has Ice Face
-    if ((pokemon.species.speciesId === Species.EISCUE && isFormIceFace) ||  isWeatherSnowOrHail) {
+    if (pokemon.formIndex === 0) {
       return true;
     }
+
     return false;
```
innerthunder, line 1577:
```suggestion
    return pokemon.formIndex === 0;
```

src/data/battler-tags.ts
```diff
@@ -1609,6 +1602,28 @@ export class IceFaceTag extends BattlerTag {
...
+   */
+  canAdd(pokemon: Pokemon): boolean {
+    const weatherType = pokemon.scene.arena.weather?.weatherType;
+    const isWeatherSnowOrHail = weatherType === WeatherType.HAIL || weatherType === WeatherType.SNOW;
+
+    if (super.canAdd(pokemon) || isWeatherSnowOrHail) {
+      return true;
+    }
+
+    return false;
```
innerthunder, line 1624:
```suggestion
    return super.canAdd(pokemon) || isWeatherSnowOrHail
```

innerthunder (approved):
lgtm 🚀

torranx (approved):
thanks for the update, lgtm

PR #2687, opened 2024-06-28
[Bug] Fix the water shuriken did not hit 3 guaranteed hits in battle bond form.

KimJeongSun:
## What are the changes?
This PR is for fixing bug of water shuriken did not hit 3 guaranteed hits in battle bond form.

it looks already implemented before, but in the previous code, assigning multi hit condition is missing.
so, water shuriken always hit 2-5 times even if in battle bond form.
this will make water shuriken 3 guaranteed hits in battle bond form.

## What did change?

add code for assigning multiHitType with hitType.value after ChangeMultiHitTypeAttr function.
```
      applyMoveAttrs(ChangeMultiHitTypeAttr, user, target, move, hitType);
      this.multiHitType = hitType.value;
      hitTimes = this.getHitCount(user, target);
```

### Screenshots/Videos

to debugging, I added temporary log message and changing power of water shuriken when battle bond 20-> 1 when i recording following videos.

- before fix ( + debug codes )
344268792-fa30eb44-27aa-4807-970e-f6ee3c2f5708
water shuriken not hit 3 times ( multiHitType is 1 which is multiHitType._2_TO_5 )

- after fix
344269542-c6237830-8eae-4fbb-b5db-17ac5212a972
water shuriken hit 3 times ( multiHitType is 2 which is MultiHitType._3 )

## How to test the changes?

[sessionData___SUN__ (1).prsv.zip](https://github.com/user-attachments/files/16034372/sessionData___SUN__.1.prsv.zip)
you can test with this session file.
this is session of trainer battle situation.
After the first battle with Electrode, Greninja changed into Battle Bond form and took the video. (This is due to the 10M video size limit. I can attach the entire video if you need.)

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -1614,6 +1614,7 @@ export class MultiHitAttr extends MoveAttr {
     } else {
       const hitType = new Utils.IntegerHolder(this.multiHitType);
       applyMoveAttrs(ChangeMultiHitTypeAttr, user, target, move, hitType);
+      this.multiHitType = hitType.value;
       hitTimes = this.getHitCount(user, target);
     }
 


Tempo-anon:
Waiting on dev consensus on whether or not water shuriken should have this interaction as the most recent gen (9) removed this feature.

Tempo-anon (approved):
Going to merge this in for now because
1. This maintains current functionality with `WaterShurikenPowerAttr` which is a battle-bond interaction
2. It's a very very common user request/report

If we want to decouple battle bond completely from water shuriken we can do so in the future.

PR #2624, opened 2024-06-26
[Sprite] Fix variant icons from Gen7

chaosgrimmon:
## What are the changes?
Some variant icons that should've been on `pokemon_icons_7v.png` were not. This PR fixes that.

## Why am I doing these changes?
The changes allow the game to show the correctly-coloured icon as needed without errors.

## What did change?
Icons were added to `pokemon_icons_7v.png`, and `pokemon_icons_7v.json` was modified to reflect that. Part of the process included repositioning other icons in order to make efficient use of space.
Icons were also removed from `pokemon_icons_6v.png`, corresponding to the bytes removed from `pokemon_icons_6v.json`. Although that spritesheet was not adjusted to optimally use the freed-up space, hopefully having done so will come in useful the next time icons are added.
The concern would be that were this intermediary step not performed at this juncture, those icons would be assumed to be relevant, constituting a perpetual, extraneous burden.

Also, some Mimikyu nomenclature changes may have occurred, which could temporarily cause issues, but are necessary as currently Busted Mimikyu variants are not properly implemented.

#### Before
image

image

#### After
image
image

## How to test the changes?
Although Overrides can be useful, the easiest way would be to use `src/test/utils/saves/everything.prsv`

## Checklist
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #2690, opened 2024-06-28
[Sprite] Fixes Busted Mimikyu variants and icons; Fixes #2683

### LINKED ISSUE #2683: Mimikyu Common Shiny Sprite Inconsistency [BUG]
**Describe the bug**
In Mimikyu's common shiny, its sprite is slightly different in the area where the right ear connects to the head. This does not apply to the rare or epic shinies.

**To Reproduce**
1. Capture/hatch a shiny Mimikyu
2. Compare the sprites between the non-shiny and the common shiny variants
3. See error
Alternatively:
1. Open the [PokéRogue PokéDex](https://ydarissep.github.io/PokeRogue-Pokedex/?species=SPECIES_MIMIKYU&table=speciesTable&)
2. Compare the normal and common shiny sprites
3. See error

**Expected behavior**
Sprites should be consistent.

**Screenshots / Videos**
image
image

**Device**
Firefox, Windows 11

**Additional context**
I wanted to label this as Misc, but this is my first GitHub issue, and I couldn't figure out how to change the label. My apologies.

chaosgrimmon:
## What are the changes?
Previously, variant Mimikyu would use the shiny palette when becoming Busted. This PR ensures they have the same colours as while Disguised, including in their icon sprites.

They're visible discrepancies that will remain in the game until fixed. Also I basically promised to do all of this in #2624 so here we are.

## What did change?
Previously, in `pokemon-species.ts`, Disguised Mimikyu had a `formSpriteKey` of `""`. This means that the game expects any file with a name ending in `778` to be relating to Disguised form, and as Busted form did not have an explicitly initialised `formSpriteKey`, it defaulted to having files named `778-busted`.
Unfortunately, as a `""` `formSpriteKey` is the default for when a Pokemon has no relevant forms, Busted Mimikyu variant icons were not being properly located and used.
Fixing this involved giving all of the Disguised-form files a new suffix, and there really was nothing wrong with what the default would have been. The list of experimental animations, as well as the list of variants, were both updated to match the new name.

Furthermore, the Busted form simply didn't have the necessary files to perform the desired palette swaps. Adding those `.json`s makes it actually able to apply the correct variant colourations. Some non-shiny front sprites were invisibly modified in order to achieve the intended effects, as previously it was carried out using an extra set of files, labelled `/variant/exp/778_3`. Again, the list of variants was updated to reflect the difference in methodology.

Also, changes made to the static shiny front sprite to make it similar to the non-shiny sprite, which resolves #2683.

#### Before
image
image

#### After
image
image

## How to test the changes?
Overrides or `everything.prsv` can help to collect the necessary Mimikyu for testing.

## Checklist
- [x] Are the changes visual?


flx-sta:
Does this also fix this one ?
https://github.com/pagefaultgames/pokerogue/issues/2683

If so please link the issue to the pr

Tempo-anon (dismissed):

chaosgrimmon:
> Does this also fix this one ? #2683
> 
> If so please link the issue to the pr

It does now!
The linework on the ear wasn't the best, sure, but in fact there was a second difference in that the eyeholes were shaded separately, instead of both on the same side like the non-shiny sprite does.
Interestingly enough, that's always the case on the experimental animation sprites, though none of *those* have the exact ear pattern...
image

Tempo-anon (approved):

PR #2692, opened 2024-06-28
[Feature][ENV] make i18n debugging an optional env setting

flx-sta:
## What are the changes?

I've made i18n debugging an optional setting that can be changed via `.env` files

It's spamming the console but more importantly adding noise to the test-result output (a huge object in every test)

## What did change?

i18n is no longer always in debug mode but can be forced into it by changing `VITE_I18N_DEBUG` to the value `1`

## How to test the changes?

1. Launch the local game as usual
2. Check the console in the browser: no i18n output
3. Open `.env.development`
4. Change `VITE_I18N_DEBUG` to the value `1`
5. Restart local client
6. Check console: it should print i18n stufff

## Checklist


torranx (approved):
lgtm

flx-sta:
@Tempo-anon  can thus be considered ready for merge?

CodeTappert:
This should be default ON.

So we can easier see which strings did not load correctly in the console

Tempo-anon (approved):

PR #2695, opened 2024-06-28
[Bug] Making FormChangeItems Untransferrable

### LINKED ISSUE #208: Transferring Max Mushroom keeps the original Pokemon G-Maxed
After a Pokemon has evolved with the Max Mushroom you can transfer it with out reverting back to normal. The new pokemon holding the item will not evolved until the game is reloaded. This seems like unintended behaviour. Seems like it would also work with mega stones but I have not been able to test this.
G-max
image

Arxalc:
## What are the changes?

Makes form change items untransferrable.

closes #208 by preventing this scenario.
## What did change?

getTransferrable() really should have been an instance variable instead. The parameter was obselete and was used incorrectly.
### Screenshots/Videos

This G-Max Pikachu is holding a Max Mushroom, but it's no longer giving you the option to transfer it.
Screenshot (7)

## How to test the changes?

Give yourself a form change item. Observe how you can no longer transfer it to other pokemon.
## Checklist
- [x] Are the changes visual?


src/phases.ts
```diff
@@ -5098,7 +5098,7 @@ export class SelectModifierPhase extends BattlePhase {
           this.scene.ui.setModeWithoutClear(Mode.PARTY, PartyUiMode.MODIFIER_TRANSFER, -1, (fromSlotIndex: integer, itemIndex: integer, itemQuantity: integer, toSlotIndex: integer) => {
             if (toSlotIndex !== undefined && fromSlotIndex < 6 && toSlotIndex < 6 && fromSlotIndex !== toSlotIndex && itemIndex > -1) {
               const itemModifiers = this.scene.findModifiers(m => m instanceof PokemonHeldItemModifier
-                    && (m as PokemonHeldItemModifier).getTransferrable(true) && (m as PokemonHeldItemModifier).pokemonId === party[fromSlotIndex].id) as PokemonHeldItemModifier[];
+                    && (m as PokemonHeldItemModifier).isTransferrable && (m as PokemonHeldItemModifier).pokemonId === party[fromSlotIndex].id) as PokemonHeldItemModifier[];
```
OrangeRed, line 5101: Can you remove the inner `as PokemonHeldItemModifier` everwhere its not needed? 
typescript will infer the type from the `instanceof` that came right before so there is no reason to explicitly type cast with `as`

src/data/move.ts
```diff
@@ -1981,7 +1981,7 @@ export class StealEatBerryAttr extends EatBerryAttr {
       return false;
     }
 
-    const heldBerries = this.getTargetHeldBerries(target).filter(i => i.getTransferrable(false));
+    const heldBerries = this.getTargetHeldBerries(target);
```
OrangeRed, line 1984: Why did you remove this filter?

src/modifier/modifier.ts
```diff
@@ -461,6 +461,7 @@ export class TerastallizeAccessModifier extends PersistentModifier {
 
 export abstract class PokemonHeldItemModifier extends PersistentModifier {
   public pokemonId: integer;
+  public isTransferrable: boolean = true;
```
OrangeRed, line 464: Is being able to change this class property desirable? If not then change it to be readonly or the following if you want to make it readonly only outside the class
```ts
private _isTransferrable

get isTransferrable() { return this._isTransferrable }
```

OrangeRed (changes requested):

src/phases.ts
OrangeRed: Can you do this everywhere not just in this one specific place

OrangeRed:
> Can you remove the inner `as PokemonHeldItemModifier` everwhere its not needed? typescript will infer the type from the `instanceof` that came right before so there is no reason to explicitly type cast with `as`

Can you do this everywhere you made changes not just in the one place that I mentioned.

Arxalc:
> > Can you remove the inner `as PokemonHeldItemModifier` everwhere its not needed? typescript will infer the type from the `instanceof` that came right before so there is no reason to explicitly type cast with `as`
> 
> Can you do this everywhere you made changes not just in the one place that I mentioned.

Would that not be better as a separate PR?

OrangeRed:
I would say a small typescript change like this, that simply removes unnecessary type casts in places that you've touched, doesn't necessarily need to be its own PR.

OrangeRed (approved):

PR #2696, opened 2024-06-28
[Bug] Fix Make It Rain applying stat change twice after KO'ing both opponents

innerthunder:
## What are the changes?

This fixes a bug where if Make It Rain KO's both opponents in a double battle, the user's Sp. Atk. is lowered twice.

Make It Rain is (still) not working as intended.

## What did change?

- `data/move`: Added a `firstTargetOnly` parameter to `MoveEffectAttr`. Make It Rain now uses this parameter for its stat change effect.
- `phases`: Added logic in `MoveEffectPhase` to support the `firstTargetOnly` option. This is basically the old logic for `firstHitOnly` from before general changes to multi-hit logic, so Make It Rain's stat change should now function as it used to before then.

### Screenshots/Videos

The video below shows how Make It Rain behaves after the changes in two edge cases that cause problems upstream:
1. Player uses Make It Rain, and it KO's both opponents (on main, this sharply reduces the user's Sp. Atk.)
2. In a double battle, Player uses Make It Rain to hit and KO a single Pokemon (on main, behavior varies based on the Make It Rain target's field index)

https://github.com/pagefaultgames/pokerogue/assets/168692175/4846a841-4d38-4afe-83ac-640c645ecdf2

## How to test the changes?

1. `npm run test make_it_rain`
2. See the video above.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -813,12 +813,15 @@ export class MoveEffectAttr extends MoveAttr {
   public firstHitOnly: boolean;
   /** Should this effect only apply on the last hit? */
   public lastHitOnly: boolean;
+  /** Should this effect only apply on the first target hit? */
+  public firstTargetOnly: boolean;
 
-  constructor(selfTarget?: boolean, trigger?: MoveEffectTrigger, firstHitOnly: boolean = false, lastHitOnly: boolean = false) {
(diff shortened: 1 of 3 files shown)


flx-sta:
You sure you don't wanna try to write some tests for it now? image

Tempo-anon (approved):

PR #2698, opened 2024-06-29
[miscellaneous] Create single variable for max pokeballs

ReneGV:
## What are the changes?
There are two places where we check the user shouldn't have more than 99 pokeballs. 

- The Pokeball modifier that adds 5 Pokeballs to your bag.
- The item roll modifier that checks if player has the max ammount of balls in order to avoid rolling this item

This change just makes both of them to use the same variable.

Avoiding hardcoded constants.

## What did change?
  NA

### Screenshots/Videos
NA

## How to test the changes?
NA

### DIFF
--- a/src/data/pokeball.ts
+++ b/src/data/pokeball.ts
@@ -10,7 +10,7 @@ export enum PokeballType {
   LUXURY_BALL
 }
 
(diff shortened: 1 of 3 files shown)


flx-sta (approved):
Code review only

Tempo-anon (approved):

PR #2703, opened 2024-06-29
[Bug] Prevent fixed-damage and OHKO moves from being modified by damage-reducing abilities

Corrade:
## What are the changes?
- Fixed-damage moves (e.g. Dragon Rage) and OHKO moves (e.g. Fissure) are no longer modified by damage-reducing abilities (e.g. Fur Coat)
- OHKO moves (e.g. Fissure) are no longer modified by enemy damage boost/reduction wonder launcher items

- Bug fix: see "Fur coat persian survived fissure" thread in #bug-reports on Discord

## What did change?
- Refactored core damage calculation code to be more readable
- Prevent fixed damage moves and OHKO moves from being affected by ReceivedMoveDamageMultiplierAbAttr, EnemyDamageBoosterModifier and EnemyDamageReducerModifier
- Added comprehensive unit test for Dragon Rage
- Added one test case for Fissure. I was hoping to add all the Fissure tests here, but they opened up another can of worms. The PR for that is here (and depends on this one): https://github.com/pagefaultgames/pokerogue/pull/2705

### Screenshots/Videos
Before (from the Discord thread mentioned above):
image

After:
fissure_ohkos_furfrou

## How to test the changes?
- Run tests dragon_rage and fissure
- The dragon_rage tests can be sanity-checked by replacing Dragon Rage with Dragon Breath, which will cause all cases to fail as expected

## Checklist
- [N] Are the changes visual?


src/field/pokemon.ts
```diff
@@ -1953,7 +1963,11 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
           }
         }
 
-        if (!fixedDamage.value) {
+        const isOneHitKo = result === HitResult.ONE_HIT_KO;
+
+        if (!fixedDamage.value && !isOneHitKo) {
+          applyPreDefendAbAttrs(ReceivedMoveDamageMultiplierAbAttr, this, source, move, cancelled, damage);
```
Corrade, line 1969: Temp and I previously proposed fixing this problem by moving ReceivedMoveDamageMultiplierAbAttr above the fixed damage and OHKO sections. That'd fix the damage modification, but the ability would still be triggered for those move types. Keeping the existing ordering but enclosing it in these conditions makes more sense.

src/test/moves/fissure.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    partyPokemon = game.scene.getParty()[0];
+    partyPokemon.addTag(BattlerTagType.NO_CRIT, 99);
+
+    enemyPokemon = game.scene.getEnemyPokemon();
+    enemyPokemon.addTag(BattlerTagType.NO_CRIT, 99);
+  });
+
+  it("ignores damage modification from abilities such as fur coat", async () => {
+    vi.spyOn(overrides, "ABILITY_OVERRIDE", "get").mockReturnValue(Abilities.NO_GUARD);
+    vi.spyOn(overrides, "OPP_ABILITY_OVERRIDE", "get").mockReturnValue(Abilities.FUR_COAT);
```
torranx, line 56: is there a reason why you have 2 different overrides on a single test?

src/test/moves/fissure.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    game.doAttack(getMovePosition(game.scene, 0, Moves.FISSURE));
+    await game.phaseInterceptor.to(DamagePhase, true);
+
+    expect(enemyPokemon.isFainted()).toBe(true);
```
torranx, line 61: wont this fail if the fissure missed? i believe fissure hits based on chance

src/test/moves/dragon_rage.test.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    enemyPokemon = game.scene.getEnemyPokemon();
+    enemyPokemon.addTag(BattlerTagType.NO_CRIT, 99);
+  });
+
+  it("ignores weaknesses", async () => {
+    vi.spyOn(enemyPokemon, "getTypes").mockReturnValue([Type.DRAGON]);
+
+    game.doAttack(getMovePosition(game.scene, 0, Moves.DRAGON_RAGE));
+    await game.phaseInterceptor.to(TurnEndPhase);
```
torranx, line 63: wont `MoveEndPhase` suffice? (same with other tests)

torranx (commented):
code looks good, tests could be improved

src/test/moves/fissure.test.ts
Corrade: Yep, No Guard ensures the Fissure always hits so that we control for accuracy, and Fur Coat is the subject of the test.

src/test/moves/fissure.test.ts
Corrade: See above comment regarding No Guard. Note that there's a test case planned for Fissure + No Guard in particular, which should serve to document this understandably unintuitive interaction.

torranx: i see thank you for clarifying

src/test/moves/fissure.test.ts
torranx: what i mean here is that you have a spyons for `ABILITY_OVERRIDE` and `OPP_ABILITY_OVERRIDE` inside `beforeEach()`, and overriding it inside this single test

src/test/moves/dragon_rage.test.ts
Corrade: It could, although some extra work would be needed.

Awaiting MoveEndPhase takes you to the end of the next move from either the player or the opponent. These particular tests don't care about who goes first, so awaiting MoveEndPhase means we could either be inspecting the results of a Dragon Rage or the opposing Splash. We would get non-deterministic test output.

Though I was curious about the potential speed improvement. Replacing MoveEndPhase with TurnEndPhase subtracted about 0.5s per test case in this Dragon Rage suite.

---

Looking beyond this PR regarding this topic's relevance to all unit tests, I think it's worth highlighting that MoveEndPhase does some computation of its own:
```
    if (pokemon.isActive(true)) {
      pokemon.lapseTags(BattlerTagLapseType.AFTER_MOVE);
    }

    this.scene.arena.setIgnoreAbilities(false);
```

If we await MoveEndPhase instead of TurnEndPhase as a standard, I'm sure it'll cause bugs in future tests that copy existing MoveEndPhase-skipping code without knowing this computation is also skipped.

---

Overall, I'm cautiously against the change...

src/test/moves/fissure.test.ts
Corrade: Oh I see, thanks for the spot. The vast majority of the upcoming Fissure test cases won't override the stuff in `beforeEach()` so I kept standard things there to avoid duplication. I'll fix it for this PR!

src/test/moves/dragon_rage.test.ts
torranx: > Awaiting MoveEndPhase takes you to the end of the next move from either the player or the opponent. These particular tests don't care about who goes first, so awaiting MoveEndPhase means we could either be inspecting the results of a Dragon Rage or the opposing Splash. We would get non-deterministic test output.

you can always override the pokemon for a faster/slower one, ie pikachu vs shuckle

> Looking beyond this PR regarding this topic's relevance to all unit tests, I think it's worth highlighting that MoveEndPhase does some computation of its own:
> If we await MoveEndPhase instead of TurnEndPhase as a standard, I'm sure it'll cause bugs in future tests that copy existing MoveEndPhase-skipping code without knowing this computation is also skipped.

but before ending in `TurnEndPhase`, it would go through `MoveEndPhase` first no? as long as we know what to expect in that specific phase, i dont think it will cause bugs. also 1 reason i mentioned using `MoveEndPhase` instead is a pokemon can hold sitrus berry and procs that on `TurnEndPhase`, resulting for this test to fail.

Corrade: > you can always override the pokemon for a faster/slower one, ie pikachu vs shuckle

Yeah that's true, and I can the see the argument for doing this all the time as good practice anyway.

> but before ending in TurnEndPhase, it would still go through MoveEndPhase first no?

However, if the player's Pokemon is configured to be slower, the test would have to await the second MoveEndPhase.

> as long as we know what to expect in that specific phase, i dont think it will cause bugs.

But I think, at times, we won't know what to expect because the name of the phase isn't (perhaps by necessity) descriptive enough.

> also 1 reason i mentioned using MoveEndPhase instead is a pokemon can hold sitrus berry and would proc that on TurnEndPhase, resulting for this test to fail (as have happened on other tests similar to this before).

Thanks, this is a good point. I still think it's better solved by overriding random berries out altogether, but in any case I'll address berries in this PR's tests.

Corrade:
Seeing the failed test. It's one of the tests introduced in this PR. Will fix, although not sure why it's broken...

Smewkie:
Looks like this would also fix Protection Tokens in endless messing with OHKO moves, as they also use EnemyDamageReducerModifier, so this is great, will squash quite a few bugs with one PR!

src/field/pokemon.ts
```diff
@@ -1977,14 +1991,12 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
         }
 
         applyMoveAttrs(ModifiedDamageAttr, source, this, move, damage);
-        applyPreDefendAbAttrs(ReceivedMoveDamageMultiplierAbAttr, this, source, move, cancelled, damage);
```
torranx, line 1980: Should'nt this be the last attribute to apply to get all damage updates first?

Corrade: I took a closer look and I think ModifiedDamage should indeed be at the end. It's used exclusively by SurviveDamageAttr for False Swipe and Hold Back, which I believe should be applied after damage reduction abilities like Fur Coat.

I'm actually not sure what comment to put here. I think the most helpful thing for a dev to read here would be something like "this is used for stuff like False Swipe", but that comment would create a dependency. In an ideal world I feel like ModifiedDamage should be renamed / we should use SurviveDamageAttr directly.

I've accordingly neglected to leave this part as-is in this PR...

torranx: I see, thanks for confirming. 

Something simple like "this should be the last attribute to be applied" sounds reasonable.

Another thing, is your branch updated with beta? I notice you're still using `power.value`

Corrade: Done!

torranx (approved):
looks good 🚀

PR #2705, opened 2024-06-29
[Bug] Added Fissure tests, accordingly fixed Lock-On and Mind Reader not working on the first turn of a battle, fixed OHKO moves being affected by accuracy and evasion battle stats

Corrade:
**Note: depends on this PR, which is yet to be merged as of writing**: https://github.com/pagefaultgames/pokerogue/pull/2703

## What are the changes?
- Lock-On and Mind Reader now work when used on the first turn of a battle
- OHKO moves are no longer affected by accuracy and evasion, which includes abilities like Compound Eyes and items like Wide Lens

- Bug fixes
- Follow-up on incomplete Fissure tests from https://github.com/pagefaultgames/pokerogue/pull/2703

## What did change?
- Noted above
- Also divided src/phases.ts:hitCheck() into two functions. getHitAccuracy() is new and calculates an accuracy value. Then the old hitCheck() simply rolls an RNG value against getHitAccuracy() to determine whether or not the hit should succeed. I did this in order to isolate the accuracy value for testing. See this PR's Fissure tests for an example!
  - (Note: without this refactor, we could still experimentally determine an accuracy value by running hitCheck() thousands of times and counting the ratio of hits to misses. This isn't too slow either, provided we call hitCheck() directly. However, I think this refactor is much cleaner.)

### Screenshots/Videos
Before (Lock-On):
lock_on_fails_on_first_turn

After (Lock-On):
lock_on_succeeds_on_first_turn

## How to test the changes?
- Run _fissure_ tests

## Checklist
- [N] The PR is self-contained and cannot be split into smaller PRs?
  - But it all came from initially wanting to write Fissure tests and the subsequent fixes are small enough?
- [N] Are the changes visual?

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1924,6 +1924,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
         if (this.scene.arena.terrain?.terrainType === TerrainType.MISTY && this.isGrounded() && move.type === Type.DRAGON) {
           damage.value = Math.floor(damage.value / 2);
         }
+        applyPreDefendAbAttrs(ReceivedMoveDamageMultiplierAbAttr, this, source, move, cancelled, damage);
 
         const fixedDamage = new Utils.IntegerHolder(0);
         applyMoveAttrs(FixedDamageAttr, source, this, move, fixedDamage);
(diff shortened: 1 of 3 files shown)


torranx:
`The PR is self-contained and cannot be split into smaller PRs?`
please dont mix 3 separate things in 1 PR. its confusing

Corrade:
@torranx Will do, although I think in this particular case the expected confusion is smaller than the effort required to split this PR up. Anyway, I'll close this PR temporarily while its dependent PR remains in the backlog.

torranx:
> @torranx Will do, although I think in this particular case the expected confusion is smaller than the effort required to split this PR up. Anyway, I'll close this PR temporarily while its dependent PR remains in the backlog.

splitting 3 separate things into its own separate PRs makes it a lot easier to get reviewed and merged. also its easier to track the changes made as it is not convoluted 👍

PR #2706, opened 2024-06-29
[Balance] Allows the user to swap pokemon at the start of a trainer battle

c0nsy:
## What are the changes?
At the beginning of a pokemon trainer battle, the system will ask the user if they want to switch their current pokemon

Keeps the game consistent, we can do this with other encounters, like WILD and DOUBLE.

## What did change?
Logic change

### Screenshots/Videos
trainer
swap prompt

## How to test the changes?

Start a trainer battle and see if you get the prompt to switch your pokemon at the start of the battle 

## Checklist

- [x] Are the changes visual?

### DIFF
--- a/src/locales/en/party-ui-handler.ts
+++ b/src/locales/en/party-ui-handler.ts
@@ -4,6 +4,7 @@ export const partyUiHandler: SimpleTranslationEntries = {
   "ALL": "All",
   "SEND_OUT": "Send Out",
   "SUMMARY": "Summary",
+  "SWITCH": "Switch",
   "CANCEL": "Cancel",
   "RELEASE": "Release",
   "APPLY": "Apply",
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -416,7 +416,7 @@ export class TitlePhase extends Phase {
         this.scene.pushPhase(new SummonPhase(this.scene, 1, true, true));
       }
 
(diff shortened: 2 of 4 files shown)


torranx:
wasnt this an intended game mechanic?

c0nsy:
> wasnt this an intended game mechanic?

@torranx maybe, I just don't like/agree with it.

Blitz425:
There were talks about including a feature to allow you to swap the order of your team before a battle starts, having the option after seeing a Pokemon / the type of trainer that's in the next battle cheapens the experience and defeats the purpose/challenge of the randomness in Rogue (i.e. seeing the next trainer is a Grass type Gym Leader allowing you to counterpick a fire type for free). If you could change it to where the option to swap the order of your team only appears in the "Check Team" option after a battle ends, that would be much appreciated!

Opaque02:
Hey, as per discord chats, just leaving this here to let you know to change the name of the "check team" option to "manage team", and to change the description text when hovering over the check team button to be more in line with the team management side of things

sodaMelon:
hi i read all comment. (because i thought that is an intended game mechanic too)
The game will be less difficult, but it seems convenient. (some people like hard games... that's the reason i surprised)

i will expect your next commit! (if you change tutorial sentences, i will translate in to Korean. )

damocleas:
I don't really want people to be able to switch their mons on lead to trainer battles once both are sent out, but they SHOULD be able to rearrange their party between battles in the 'check team' menu or such. Makes more sense.

flx-sta:
Closing as only “re-arrange” team is wanted

PR #2713, opened 2024-06-29
[Balance] Fix N-Solarizer etc appearing without secondary mon registered in dex

Fontbane:
## What are the changes?

- Changes the Dark Stone, Light Stone, N-Solarizer, N-Lunarizer, and the Reins of Unity- all fusion items
- Previously these would show up like any other form change item by simply having Kyurem, Necrozma or Calyrex in your party
- Instead, you will now need to have the appropriate mon (Zekrom, Reshiram etc) registered in the Dex for this item to show up

- It was unbalanced and unthematic that these forms could be accessed without even having encountered the second half
- Frankly it makes no sense for the player to be able to materialize the horses out of thin air
- Adds an extra element of progression in the game by making filling the dex more important

## What did change?

- New function to make a form change condition based on whether a specific species is registered as caught by the player's dex data
- Applied this to Dark Stone (Zekrom), Light Stone (Reshiram), N-Solarizer (Solgaleo), N-Lunarizer (Lunala), Icy Reins of Unity (Glastrier) and Shadow Reins of Unity (Spectrier)
- Kyurem, Necrozma, and Calyrex will only have their form changing items show up if the second mon in their fusion is caught

## How to test the changes?

- Test if Icy Reins of Unity et al show up with and without Glastrier or the appropriate mon registered in the dex
- Test that other, unrelated form changes work as expected

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/pokemon-forms.ts
+++ b/src/data/pokemon-forms.ts
@@ -374,6 +374,12 @@ export function getSpeciesFormChangeMessage(pokemon: Pokemon, formChange: Specie
   return `${prefix}${preName} changed form!`;
 }
 
+/**
+ * Gives a condition for form changing checking if a species is registered as caught in the player's dex data.
+ * Used for fusion forms such as Kyurem and Necrozma.
+ * @param species
+ * @returns A {@linkcode SpeciesFormChangeCondition} checking if that species is registered as caught
+ */
 function getSpeciesDependentFormChangeCondition(species: Species): SpeciesFormChangeCondition {
   return new SpeciesFormChangeCondition(p => !!p.scene.gameData.dexData[species].caughtAttr);
 }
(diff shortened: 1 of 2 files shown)


damocleas (approved):

Tempo-anon:
1. On one hand I really like that we're putting more meta-progression features in since I think that's really cool, but I also think that it might suck for players in that their Kyurem is permanently stuck in its base form because they haven't ever caught a Reshiram or Zekrom. 
2. Do you have any tests of this? Like jacking up the form change item weights very high or trying to force them to drop as a reward with/without the specific Pokemon being registered in the dex?

Fontbane:
> 1. On one hand I really like that we're putting more meta-progression features in since I think that's really cool, but I also think that it might suck for players in that their Kyurem is permanently stuck in its base form because they haven't ever caught a Reshiram or Zekrom.
> 2. Do you have any tests of this? Like jacking up the form change item weights very high or trying to force them to drop as a reward with/without the specific Pokemon being registered in the dex?

Yeah I did those tests 👍

PR #2716, opened 2024-06-29
[Enhancement] Add Move Header phase and attributes

### LINKED ISSUE #898: Focus Punch does not show failure message when used
I'll try to be as precise as possible with the details this time.
I opened a fight against a Battle Girl by switching my Incineroar for my Primarina, and the opposing Conkeldurr used Focus Punch. Which hit immediately. No focus turn. Focus Punch may just be implemented differently than I thought, as me hitting Aqua Jets before its use canceled it.
Instafocus

EDIT: I was mistaken about Focus Punch. I thought it had to focus over the next turn, but it starts focusing at the start of the turn. The real bug here is that it doesn't show the "tightening its focus" message. At all, I'm pretty sure.

innerthunder:
## What are the changes?

This adds a new phase, `MoveHeaderPhase`, which takes place before all moves in a turn and applies all the `MoveHeaderAttr` attributes of a move.

At the moment, this acts as a partial fix to Focus Punch missing its on-cart messages (closes #898), but the overall intent is to create a framework so that Beak Blast and Shell Trap can be implemented (or reimplemented) in a more structured and intuitive way.

## What did change?

- `phases`: A new 'MoveHeaderPhase' is added and integrated into the game's turn sequence. With this new phase, the generic turn sequence (after all commands are selected) is as follows:
  1. If the player chooses to throw a ball, switch a Pokemon, or run away, those actions are resolved first.
  2. For each Pokemon that uses a move with a `MoveHeaderAttr`, a `MoveHeaderPhase` is added to the turn sequence. This phase applies all of the move's `MoveHeaderAttrs`, as long as the move itself can be used. If the user has a status that may block the use of the move, the status check is done here instead of in the `MovePhase`. `MoveHeaderPhases` are resolved in the same order as the turn's move order.
  3. `MovePhases` are added to the turn sequence for each move. If the user's status was previously checked in a `MoveHeaderPhase`, the status check here is ignored. Otherwise, this phase and all subsequent phases function in the same way as on upstream.
- `data/move`: `MoveHeaderAttr` is a new subclass of `MoveAttr` that encompasses all attributes that are applied in the `MoveHeaderPhase`. For now, the only `MoveHeaderAttr` implemented is `MessageHeaderAttr`, which queues a message, but more subclasses of `MoveHeaderAttr` can be created in the future to apply more effects, e.g. adding a battler tag to the user.
  - Focus Punch now uses a `MessageHeaderAttr` to show its initial message.
- `test/moves/focus_punch.test`: New unit test to ensure Focus Punch's effect timing is correct throughout the turn sequence

### Screenshots/Videos

Basic Example showing Focus Punch's new pre-move message:

https://github.com/pagefaultgames/pokerogue/assets/168692175/ca49eda0-341d-4269-9c5d-f7baf689fb26

## How to test the changes?

1. `npm run test focus_punch`
2. Use `src/overrides` to give the player Pokemon Focus Punch, then use `npm run start:dev` and observe the turn sequence in battle when using the move:
    1. What happens when using Focus Punch compared to other moves?
    2. What happens when one player Pokemon switches, and the other uses Focus Punch?
    3. What happens when one player Pokemon throws a ball or runs, and the other uses Focus Punch?

## Checklist
- [x] Are the changes visual?


innerthunder:
### Update + Test Video

When a Pokemon's turn command is skipped (e.g. after the player chooses to run or throw a ball), and Focus Punch was selected for that Pokemon, that Pokemon's `MoveHeaderPhase` is now also removed from the turn sequence. This means that if a player Pokemon using Focus Punch is skipped, the player should not see any effects from the move throughout the turn.

**Test Video:** Shows two other possible move sequences:
- One player Pokemon tries to run (and fails), while the other uses Focus Punch. In this case, Focus Punch is skipped, and there's no visible indication that the move was used.
- One player Pokemon uses Focus Punch, then the other player Pokemon attacks its ally with Tackle. This is meant to show the sequence of Focus Punch's effects if the move fails.

https://github.com/pagefaultgames/pokerogue/assets/168692175/d7524fdf-c990-43fc-836e-372fc4bfc022

src/phases.ts
```diff
@@ -2556,6 +2575,96 @@ export class CommonAnimPhase extends PokemonPhase {
...
+        } else {
+          if (healed) {
+            this.scene.queueMessage(getStatusEffectHealText(this.pokemon.status.effect, getPokemonNameWithAffix(this.pokemon)));
+            this.pokemon.resetStatus();
+            this.pokemon.updateInfo();
+          }
+          const movePhase = this.scene.findPhase(phase => phase instanceof MovePhase && phase.pokemon === this.pokemon);
+          if (movePhase instanceof MovePhase) {
+            movePhase.setIgnoreStatusCheck();
+          }
```
mcmontag, line 2639: [Beak Blast](https://bulbapedia.bulbagarden.net/wiki/Beak_Blast_(move))
>The charging phase occurs whether or not the user's status would prevent it from executing Beak Blast. For example, if the user is confused, the confusion check is performed before the strike, not the initial charge.

I know this is for Focus Punch, but I'm not sure that resolving non-volatile statuses during the charging phase is correct in either case.

innerthunder: I wasn't sure how it worked in mainline; it just seemed weird for the added messages to pop up even when statuses prevent the move from being used.

If that's how it's meant to work, though, that definitely simplifies things.

innerthunder: Just tested on Showdown, and I can confirm that non-volatile statuses only resolve when Focus Punch is actually used, and the pre-move message is shown regardless of the user's status. No preemptive status check needed!

mcmontag: Solid!  I really wasn't sure, haven't played a mainline game since gen 4 haha.

I think your battler tag idea is a good one, and I definitely think charge moves need reworked.

My instinct is to have a `ChargedAttackMove` class subclassing `AttackMove`, with a `charge` method (delegating to a constructor param) and a `chargePriority`.  `charge` would add a volatile status and display a message, and `apply` would check for it before execution as a condition (or in Focus Punch's case, it could flinch the holder immediately if they're hit with an attack).

`TurnStartPhase` would treat `ChargedAttackMove` as a special case, and handle it as two separate actions to order.

Totally spitballing...thoughts?

src/phases.ts
```diff
@@ -2301,12 +2301,14 @@ export class TurnStartPhase extends FieldPhase {
       return aIndex < bIndex ? -1 : aIndex > bIndex ? 1 : 0;
     });
 
+    this.queueMoveHeaders(moveOrder);
```
mcmontag, line 2304: Why not filter `moveOrder` to exclude skipped commands first?  Would save you from having to remove any move headers this queues.

innerthunder: Fixed with https://github.com/pagefaultgames/pokerogue/pull/2716/commits/c3dd186938ba245cdae8fb081282920316669afd

src/data/move.ts
```diff
@@ -6564,6 +6605,7 @@ export function initMoves() {
         && (user.status.effect === StatusEffect.BURN || user.status.effect === StatusEffect.POISON || user.status.effect === StatusEffect.TOXIC || user.status.effect === StatusEffect.PARALYSIS) ? 2 : 1)
       .attr(BypassBurnDamageReductionAttr),
     new AttackMove(Moves.FOCUS_PUNCH, Type.FIGHTING, MoveCategory.PHYSICAL, 150, 100, 20, -1, -3, 3)
+      .attr(MessageHeaderAttr, (user, move) => getPokemonMessage(user, " is tightening its focus!"))
```
EnochG1, line 6608: I think this message should be localize.

EnochG1 (changes requested):

src/data/move.ts
innerthunder: There isn't a file for move trigger messages like this one in the locales atm. Should localization for this wait until those files are added?

EnochG1: IMO, if `getPokemonMessage(user, " is tightening its focus!")` can be displayed by queueMessage, it should be localized.
You can refer https://github.com/pagefaultgames/pokerogue/pull/3034/files#diff-62e9cf38a6ba2a2b69a102788cdfef2a33cdc173f79124df114636685a51e762R5808

innerthunder: Just added a key for this message

EnochG1: Thanks!

EnochG1 (approved):

snoozbuster:
This is also useful for pursuit, I think - or at least, it's useful for the implementation of pursuit I have been working on. love this

src/locales/pt_BR/move-trigger.ts
```diff
@@ -21,6 +21,7 @@ export const moveTriggers: SimpleTranslationEntries = {
   "isGlowing": "{{pokemonName}} ficou envolto em uma luz forte!",
   "bellChimed": "Um sino tocou!",
   "foresawAnAttack": "{{pokemonName}} previu/num ataque!",
+  "isTighteningFocus": "{{pokemonName}} is\ntightening its focus!",
```
josericardo-fo, line 24:
```suggestion
  "isTighteningFocus": "{{pokemonName}} está\naumentando seu foco!",
```

src/locales/fr/move-trigger.ts
```diff
@@ -21,6 +21,7 @@ export const moveTriggers: SimpleTranslationEntries = {
   "isGlowing": "{{pokemonName}} est entouré\nd’une lumière intense !",
   "bellChimed": "Un grelot sonne !",
   "foresawAnAttack": "{{pokemonName}}\nprévoit une attaque !",
+  "isTighteningFocus": "{{pokemonName}} is\ntightening its focus!",
```
Adri1, line 24:
```suggestion
  "isTighteningFocus": "{{pokemonName}} se concentre\nau maximum !",
```

Adri1 (commented):
Added French

src/locales/ko/move-trigger.ts
```diff
@@ -21,6 +21,7 @@ export const moveTriggers: SimpleTranslationEntries = {
   "isGlowing": "{{pokemonName}}를(을)\n강렬한 빛이 감쌌다!",
   "bellChimed": "방울소리가 울려 퍼졌다!",
   "foresawAnAttack": "{{pokemonName}}는(은)\n미래의 공격을 예지했다!",
+  "isTighteningFocus": "{{pokemonName}} is\ntightening its focus!",
```
EnochG1, line 24:
```suggestion
  "isTighteningFocus": "{{pokemonName}}[[는]]\n집중력을 높이고 있다!",
```

EnochG1 (commented):
Korean Translation

src/locales/zh_CN/move-trigger.ts
```diff
@@ -21,6 +21,7 @@ export const moveTriggers: SimpleTranslationEntries = {
   "isGlowing": "强光包围了{{pokemonName}}\n！",
   "bellChimed": "铃声响彻四周！",
   "foresawAnAttack": "{{pokemonName}}\n预知了未来的攻击！",
+  "isTighteningFocus": "{{pokemonName}} is\ntightening its focus!",
```
sonnyding1, line 24:
```suggestion
  "isTighteningFocus": "{{pokemonName}}正在集中注意力！",
```

src/locales/zh_TW/move-trigger.ts
```diff
@@ -21,6 +21,7 @@ export const moveTriggers: SimpleTranslationEntries = {
   "isGlowing": "強光包圍了\n{{pokemonName}}！",
   "bellChimed": "鈴聲響徹四周！",
   "foresawAnAttack": "{{pokemonName}}\n預知了未來的攻擊！",
+  "isTighteningFocus": "{{pokemonName}} is\ntightening its focus!",
```
sonnyding1, line 24:
```suggestion
  "isTighteningFocus": "{{pokemonName}}正在集中注意力！",
```

sonnyding1 (commented):
Added zh_CN and zh_TW translations

src/locales/de/move-trigger.ts
```diff
@@ -21,6 +21,7 @@ export const moveTriggers: SimpleTranslationEntries = {
   "isGlowing": "{{pokemonName}} leuchtet grell!",
   "bellChimed": "Eine Glocke läutet!",
   "foresawAnAttack": "{{pokemonName}} sieht einen Angriff voraus!",
+  "isTighteningFocus": "{{pokemonName}} is\ntightening its focus!",
```
CodeTappert, line 24:
```suggestion
  "isTighteningFocus": "{{pokemonName}} konzentriert sich!",
```

src/locales/it/move-trigger.ts
```diff
@@ -21,6 +21,7 @@ export const moveTriggers: SimpleTranslationEntries = {
   "isGlowing": "{{pokemonName}} è avvolto da una luce intensa!",
   "bellChimed": "	Si sente suonare una campanella!",
   "foresawAnAttack": "{{pokemonName}} presagisce\nl’attacco imminente!",
+  "isTighteningFocus": "{{pokemonName}} is\ntightening its focus!",
```
EnochG1, line 24:
```suggestion
  "isTighteningFocus": "{{pokemonName}} si concentra al massimo!",
```

EnochG1 (commented):
Italian translation by @NicusPulcis

src/data/move.ts
```diff
@@ -983,6 +983,47 @@ export class MoveEffectAttr extends MoveAttr {
...
+ * Move Header effects apply at the beginning of a turn before any moves are resolved.
+ * They can be used to apply effects to the field (e.g. queueing a message) or to the user
+ * (e.g. adding a battler tag).
+ */
+export class MoveHeaderAttr extends MoveAttr {
+  constructor() {
+    super(true);
+  }
+
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean | Promise<boolean> {
```
snoozbuster, line 997: technically I don't think you really need to override this. the implementation is exactly the same as in the base class. could be some reason to that I'm not aware of though

src/data/move.ts
```diff
@@ -983,6 +983,47 @@ export class MoveEffectAttr extends MoveAttr {
...
+  private message: string | ((user: Pokemon, move: Move) => string);
+
+  constructor(message: string | ((user: Pokemon, move: Move) => string)) {
+    super();
+    this.message = message;
+  }
+
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    const message = typeof this.message === "string"
+      ? this.message as string
```
snoozbuster, line 1016: does typescript get angry without the cast? I would have thought that the typeof would be enough to convince it that this branch is a string, without the need to cast.

src/phases.ts
```diff
@@ -2594,6 +2597,32 @@ export class CommonAnimPhase extends PokemonPhase {
...
+
+  constructor(scene: BattleScene, pokemon: Pokemon, move: PokemonMove) {
+    super(scene);
+
+    this.pokemon = pokemon;
+    this.move = move;
+  }
+
+  canMove(): boolean {
+    return this.pokemon.isActive(true) && this.move.isUsable(this.pokemon, true);
```
snoozbuster, line 2612: strictly speaking - is it correct to ignore PP here on `this.move.isUsable`? I don't know how exactly you would manage to reduce PP between the time that the move was queued and the time that the move header phase runs - but suppose that you were out of PP somehow when the MoveHeaderPhase ran, it would still show the "tightening focus" message even though the move itself will fail, right? is there any gotcha I'm not thinking of with setting this to `false`?

src/test/moves/focus_punch.test.ts
```diff
@@ -0,0 +1,145 @@
...
+import { Species } from "#enums/species";
+import { Abilities } from "#enums/abilities";
+import { Moves } from "#enums/moves";
+import { getMovePosition } from "../utils/gameManagerUtils";
+import { BerryPhase, MessagePhase } from "#app/phases";
+import { StatusEffect } from "#app/data/status-effect";
+
+const TIMEOUT = 20 * 1000;
+
+describe("Moves - Focus Punch", () => {
```
snoozbuster, line 14: some other tests that might be worth adding, if you can figure out a good way to assert them:

- if the enemy queues a switch while you focus punch, the correct order for things to happen in is switch -> focus message -> focus punch
- maybe assert that if you use focus punch on both pokemon in doubles, the messages happen for each pokemon in speed order? (no idea if that's how it is on cart, but I assume so)
- is it possible to disable focus punch between tightening focus and the attack? is that special in some way that's worth testing?
- and if you want, a test for the weird PP edge case I mentioned above (especially if you can come up with a good way to have it actually happen in practice)

snoozbuster (approved):
minor cleanliness comments, nothing I would consider a blocker. I've been using this in a PR I've been working on and it seems to be working great for my use case there. tests are well-written (in my inexperienced-with-this-repo-opinion, anyway - they at least seem to succeed in accurately testing the behaviors they claim to test)

src/phases.ts
innerthunder: Good catch. I've been planning on adding an option to delay move headers (i.e. push them instead of unshift them in `TurnStartPhase`) for Chilly Reception's pre-move message. I don't think it matters atm, but having this check ignore PP in that case wouldn't be good.

src/data/move.ts
innerthunder: Yeah, the cast was redundant. Changing it with the next commit

src/data/move.ts
innerthunder: Removed on next commit

frutescens:
lgtm! 

Do you think this can be applied in the case of the status move Quash and other priority breaking moves? (The affected Pokemon's move is moved to the end of the turn order)  https://archives.bulbagarden.net/media/upload/7/7a/Quash_V.png

innerthunder:
> lgtm!
> 
> Do you think this can be applied in the case of the status move Quash and other priority breaking moves? (The affected Pokemon's move is moved to the end of the turn order) https://archives.bulbagarden.net/media/upload/7/7a/Quash_V.png

You'd have to manipulate the phase queue some other way, I think. The main use case for these is for effects whose order in the turn sequence can be resolved at the start of the turn.

frutescens:
Video of Test: 
https://github.com/user-attachments/assets/3336f626-3de1-456b-93f7-67a5223d51c0

src/test/moves/focus_punch.test.ts
innerthunder: - Added a test for the first case
- The second case was too difficult to implement under the current framework. I kept running into issues when trying to await the `MoveHeaderPhase` with the phase interceptor. I think since `MoveHeaderPhase` has such a short runtime, it completes before `game.phaseInterceptor` has a chance to catch it.
- The third case here is too similar to the "mid-turn sleep" test, imo.
- I also removed `ignorePP` from the call to `isUsable` in `MoveHeaderPhase`, so the last case no longer applies.

src/test/moves/focus_punch.test.ts
```diff
@@ -118,28 +122,19 @@ describe("Moves - Focus Punch", () => {
...
-      await game.phaseInterceptor.to(MessagePhase); // Header message
+      await game.phaseInterceptor.to(TurnStartPhase);
 
-      expect(leadPokemon.getMoveHistory().length).toBe(0);
-
-      await game.phaseInterceptor.to(BerryPhase, false);
-
-      expect(leadPokemon.getMoveHistory().length).toBe(1);
-      expect(enemyPokemon.hp).toBe(enemyPokemon.getMaxHp());
+      expect(game.scene.getCurrentPhase() instanceof SwitchSummonPhase).toBeTruthy();
```
snoozbuster, line 137: you might also want to assert that there are no instances of `"MoveHeaderPhase"` in `game.phaseInterceptor.log`, just to be double sure that the switch happened before the move header phase. as it stands, I don't think this test would have failed in my branch where I temporarily accidentally broke this behavior and swapped the header and switch phases.

innerthunder: I made it check that a `MoveHeaderPhase` exists in `game.scene.phaseQueue` instead. `game.phaseInterceptor.log` is challenging to use since TS reads it as `any`.

snoozbuster: I think I fixed the type in a different PR that hasn't merged yet, but the type of it is just `string[]` - just a list of phase names

snoozbuster: but yeah - that assertion also works fine

snoozbuster (approved):

snoozbuster (approved):

snoozbuster (approved):

PR #2718, opened 2024-06-30
[QoL] Allow learning of new level 0 moves on form change

Fontbane:
## What are the changes?

- If a mon changes into a different form with a different level 0 move from its previous form, it will learn that move
- Main example would be Zacian and Zamazenta learning Behemoth Blade/Bash on form change
- Also fixed Ash Greninja learning Water Shuriken twice

- QoL and thematic purposes so the moves are learned on form change as intended
- Save the player a memory mushroom

## What did change?

- Added an extra bit to FormChangePhase to compare the form's moveset to the previous one and trigger a LearnMovePhase if a new move is found at level 0

## How to test the changes?

- Move FORM_CHANGE_ITEM to Common tier and set its weight really high
- Start a run with Zacian, take a Rusted Sword, you will be prompted to learn Behemoth Blade

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/pokemon-level-moves.ts
+++ b/src/data/pokemon-level-moves.ts
@@ -18973,7 +18973,6 @@ export const pokemonFormLevelMoves: PokemonSpeciesFormLevelMoves = {
   [Species.GRENINJA]: {
     1: [
       [ 0, Moves.WATER_SHURIKEN ],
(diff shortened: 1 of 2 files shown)


flx-sta:
image

PR #2721, opened 2024-06-30
[QoL] EBU R 128 filter applied previously too strongly for some tracks

### LINKED ISSUE #2717: [BUG] BGM audio is not consistent in volume anymore
**Describe the bug**
Commit  #2691 messed up the volume of tracks in the game. Some tracks are way too loud now, others are quiet.

**To Reproduce**
1. Play the game and just notice the swinging audio levels of some biomes, sound effects, etc.

Some examples include:
- The evil team theme being too quiet.
- When evolving the Pokemon cries are extremely loud vs the evolution congratulations theme.
- Entering a trainer battle, the trainer music is pretty quiet. Once you win the battle, the start of the victory theme is extremely loud.

**Expected behavior**
Audio before the commit sounded consistent in volume, but now it's all over the place no matter what levels your BGM and SE audio sliders are at.

**Screenshots / Videos**

**Device**
Windows 10 22H2, Firefox.

Dakurei:
## What are the changes?
EBU R 128 filter applied previously (#2691) too strongly for some tracks, instead of using the recommended LUFS of -23, use the LUFS of -14 used by Youtube
And locking in a maximum peak value of -1.0dB to avoid the saturation still present at -14 LUFS for some tracks

If, with these changes, the volume still isn't right, we'll probably have to manually change the volume for each track that saturates, rather than normalizing all tracks.

Let's wait until everyone's cache expires to see if this solves the problem.

Fix #2717 (_in theory 👀_)

## How to test the changes?
Same as #2691 

## Checklist
    > Not on the PR, but not my bad
    > image


Greenlamp2:
make sense (g r1 compri)

HighMans:
Why not apply the filter to all audio tracks in general?

Dakurei:
> Why not apply the filter to all audio tracks in general?
That's already what's been done in #2691, and that's what people have been complaining about 🤔

So the filter was applied to all the tracks, but with a lower standard than the previous one
And if, in this case too, there are still problems, we'll normalize the saturated tracks on a case-by-case basis

HighMans:
> > Why not apply the filter to all audio tracks in general?
> 
> That's already what's been done in #2691, and that's what people have been complaining about 🤔
> 
> So the filter was applied to all the tracks, but with a lower standard than the previous one And if, in this case too, there are still problems, we'll normalize the saturated tracks on a case-by-case basis

You're doing it to the background music only though, no? Why not everything under /public/audio/*?

Dakurei:
> > > Why not apply the filter to all audio tracks in general?
> > 
> > 
> > That's already what's been done in #2691, and that's what people have been complaining about 🤔
> > So the filter was applied to all the tracks, but with a lower standard than the previous one And if, in this case too, there are still problems, we'll normalize the saturated tracks on a case-by-case basis
> 
> You're doing it to the background music only though, no? Why not everything under /public/audio/*?

SEs and Pokémon cries have nothing to do with BGMs
This is absolutely not the purpose of this PR, or the previous one

Tempo-anon (approved):

OreoCupcakes:
> > > > Why not apply the filter to all audio tracks in general?
> > > 
> > > 
> > > That's already what's been done in #2691, and that's what people have been complaining about 🤔
> > > So the filter was applied to all the tracks, but with a lower standard than the previous one And if, in this case too, there are still problems, we'll normalize the saturated tracks on a case-by-case basis
> > 
> > 
> > You're doing it to the background music only though, no? Why not everything under /public/audio/*?
> 
> SEs and Pokémon cries have nothing to do with BGMs This is absolutely not the purpose of this PR, or the previous one

It's not just the BGMs though. It's the discrepancy between each individual BGM and then also the SE.

Maybe this PR fixes the discrepancy between the BGM tracks, but it doesn't fix the discrepancy between the BGM tracks and SE tracks. #2691 basically broke the volume between the SEs and BGMs and BGM track X and BGM track Y.

If you're trying to fix the saturation of audio, then target those individual tracks of apply the audio supression algorithm on every audio file in the game.

Dakurei:
> > > > > Why not apply the filter to all audio tracks in general?
> > > > 
> > > > 
> > > > That's already what's been done in #2691, and that's what people have been complaining about 🤔
> > > > So the filter was applied to all the tracks, but with a lower standard than the previous one And if, in this case too, there are still problems, we'll normalize the saturated tracks on a case-by-case basis
> > > 
> > > 
> > > You're doing it to the background music only though, no? Why not everything under /public/audio/*?
> > 
> > 
> > SEs and Pokémon cries have nothing to do with BGMs This is absolutely not the purpose of this PR, or the previous one
> 
> It's not just the BGMs though. It's the discrepancy between each individual BGM and then also the SE.
> 
> Maybe this PR fixes the discrepancy between the BGM tracks, but it doesn't fix the discrepancy between the BGM tracks and SE tracks. #2691 basically broke the volume between the SEs and BGMs and BGM track X and BGM track Y.
> 
> If you're trying to fix the saturation of audio, then target those individual tracks or apply the audio supression algorithm on every audio file in the game. As it is on live, it feels like the SE volume slider is on 150 when the BGM slider fluctuates between 50 and 100 as it wants.

A discrepancy appeared between BGMs and SEs in the previous PR, because the BGMs had been turned down too much.
The aim of this PR is to raise the volume of most of them by setting the standard lower.

If, and only if, with the lower standard used, the disparity is still too great, then at that point the old tracks will be taken over as they were, and only the tracks identified with saturation will be corrected.

The LUFS filter used by EBU R 128 is not designed to handle sound effects, and applying it will only amplify their volume. It calculates the overall noise of the track in order to judge the appropriate sound volume, and a noise that is too short will be systematically increased unless it was already too loud, thereby making the difference even greater.

Madmadness65 (approved):

Issue #2671, opened 2024-06-27
[BUG] Ceaseless Edge does not set up multiple layers of spikes with Multi Lens


Terrorforge:
**Describe the bug**

Gave my Toxapex with Ceaseless Edge Multi Lens thinking it would be cool to set up two layers of spikes at a time. However, while the attack now goes off twice, only one layer of spikes is set.

**To Reproduce**

1. Give a pokemon Ceaseless Edge and a Multi Lens
2. Use Ceaseless Edge

**Expected behavior**

When a move is executed twice, its secondary effects should also happen twice, as happens with e.g. Power-Up Punch.

**Screenshots / Videos**

![image](https://github.com/pagefaultgames/pokerogue/assets/5107449/c2c43cf3-8bb1-40be-a602-5bd3feb77c34)
![image](https://github.com/pagefaultgames/pokerogue/assets/5107449/3ef85e9d-0241-4cab-abe2-85b57b29b250)
It should not even be possible for my Toxapex to only have one layer of spikes up.

**Device**

Chrome on PC

schmidtc1:
I believe this is due to the parent class having firstHitOnly as true when calling super:
https://github.com/pagefaultgames/pokerogue/blob/67a21b8a9496af739e14ca60954244891fc7fffe/src/data/move.ts#L4203

schmidtc1:
> I believe this is due to the parent class having firstHitOnly as true when calling super:
> 
> https://github.com/pagefaultgames/pokerogue/blob/67a21b8a9496af739e14ca60954244891fc7fffe/src/data/move.ts#L4203

Testing with it as false doesn't seem to affect Reflect/Stealth Rock/other trap moves

torranx:
is this confirmed to be not a UI issue only? i mean the flyout is maybe only showing inaccurate number

Terrorforge:
Not confirmed either way, no. I did only get the "spikes were scattered" message once, after the first hit, but that could also be a UI issuel.

schmidtc1:
Seeing as setting 'firstHitOnly' to false solves the issue, I feel as though it's not a UI issue.

PR #2727, opened 2024-06-30
[Bug] Removes firstHitOnly from AddArenaTagAttr to allow multi hit set up with Ceaseless Edge

schmidtc1:
#2671 

## What are the changes?

Removes firstHitOnly boolean from AddArenaTagAttr to allow for multi-hit set up (spikes) when using Ceaseless Edge.

firstHitOnly is a redundant variable (that is always set true in the super call) that only interferes with Ceaseless Edge. When using multi lens with Ceaseless Edge, only 1 layer of spikes will be set up instead of 2. There is also already a failOnOverlap boolean that takes care of having more than 1 layer of trap. The only reason to keep firstHitOnly could be for avoiding empty function calls/making functions more efficient.

## What did change?

Removed firstHitOnly from the super call. Allows one layer of spikes per hit of Ceaseless Edge with multi lens (still max 3). Note that other moves that failOnOverlap will still only add one layer if possible: Stone Axe, Glitzy Glow, etc.

### Screenshots/Videos

See #2671 for the before.
After:

https://github.com/pagefaultgames/pokerogue/assets/62030095/41cd5170-82cc-4c1d-9f37-5e42159c7486

## How to test the changes?

Give your pokemon Ceaseless Edge, Stone Axe, other AddArenaTagAttr moves in src/overrides.ts.
Check arena flyout/damage on battle transition. 

## Checklist
- [x] Are the changes visual?


flx-sta:
I think some tests for this would be a good idea. Arena traps have been a delicate thing in the past

schmidtc1:
> I think some tests for this would be a good idea. Arena traps have been a delicate thing in the past
Advice on writing tests? Is there a how-to/procedure I can look at or do I just have to emulate other tests?

src/test/moves/ceaseless_edge.test.ts
```diff
@@ -80,38 +82,53 @@ describe("Moves - Ceaseless Edge", () => {
...
       const enemyPokemon = game.scene.getEnemyPokemon();
       expect(enemyPokemon).toBeDefined();
 
-      const enemyStartingHp = enemyPokemon.hp;
-
       game.doAttack(getMovePosition(game.scene, 0, Moves.CEASELESS_EDGE));
+      await game.phaseInterceptor.to(MoveEffectPhase, false);
+      // Spikes should not have any layers before move effect is applied
+      const tagBefore = game.scene.arena.getTagOnSide(ArenaTagType.SPIKES, ArenaTagSide.ENEMY) as ArenaTrapTag;
+      expect(tagBefore).toBeUndefined();
```
flx-sta, line 118: shouldn't you first do an expect on if the tag is actually the expect tag?

schmidtc1: What do you mean by tag-class? Like `expect(tag).toBe(ArenaTagType.SPIKES)`?

flx-sta: more like 
```ts
expect(instanceof tagBefore).toBe(ArenaTrapTag);
```

schmidtc1: Do you mean to do this for tagAfter? tagBefore is supposed to be undefined prior to the MoveEffectPhase because the effect has not been applied yet, i.e. the spikes tag has not been set up.
Also, instanceof causes an error since its an object, not a class or function. Is there another way I should be writing this expect

flx-sta: probably this
```ts
expect(tagBefore instanceof ArenaTrapTag).toBeTruthy() //*
```
<sub>* untested code</sub>

schmidtc1: I see, that still fails with tagBefore since tagBefore is supposed to be undefined. Do you want me to have the expects like so?
```typescript
expect(tagBefore instanceof ArenaTrapTag).toBeFalsy();
// perform ceaseless edge to lay down ArenaTrap-Spikes
expect(tagAfter instanceof ArenaTrapTag).toBeTruthy();
```

flx-sta: Perfect. Yeah. I didn't know the whole logic but this is exactly what I wanted

schmidtc1: Before I make the commit, do I still need both of these expects, or should I remove toBeUndefined?
```typescript
expect(tagBefore instanceof ArenaTrapTag).toBeFalsy();
expect(tagBefore).toBeUndefined();
``` 
They seem like functionally the same test, but this is my first unit test attempt, so I could be mistaken

flx-sta: No loss to keep both but I'd stick with only the first since it's more descriptive of what we are actually expecting

Tempo-anon (approved):

PR #2728, opened 2024-06-30
[Bug] Let eternatus keep stat changes when changing form

Xavion3:
## What are the changes?
Eternatus was resetting stat stages on changing form.

Per discussion on discord, this is a bug. The reason why it has two buffing moves is its meant to keep the stat boosts.

## What did change?
Remove the stat stage reset on changing form.

### Screenshots/Videos
phase1
phase2

## How to test the changes?
KO eternatus at wave 200, without this PR it'll lose the stat boosts its acquired on changing form, with it it'll keep them.

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/form-change-phase.ts
+++ b/src/form-change-phase.ts
@@ -280,7 +280,6 @@ export class QuietFormChangePhase extends BattlePhase {
   end(): void {
     if (this.pokemon.scene?.currentBattle.battleSpec === BattleSpec.FINAL_BOSS && this.pokemon instanceof EnemyPokemon) {
       this.scene.playBgm();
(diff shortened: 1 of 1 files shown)


InnocentGameDev:
But then I can just super drop his ACC and ATK stats, no? I thought it was intended that it resets stats for that reason.

Madmadness65:
> But then I can just super drop his ACC and ATK stats, no? I thought it was intended that it resets stats for that reason.

It's a double edged sword, because yes, you could tank its stats in the first phase, but it could also use the two setup moves it has to straight sweep in the second phase if not prepared for.

damocleas (approved):

Madmadness65 (approved):

Issue #2689, opened 2024-06-28
[BUG] Some POST_APPLY moves are applying effects through protect


schmidtc1:
**Describe the bug**

Some moves that have the MoveEffectTrigger.POST_APPLY tag are applying their effects despite not hitting through protect. Changing to MoveEffectTrigger.HIT fixes the issue, but I am not sure if this causes other bugs.

**To Reproduce**

1. Give opponent protect move(s)
2. Use a post apply move with additional effects (ceaseless edge, jaw lock, etc.)
3. See effect applied after damage is blocked.

**Expected behavior**
The effect should not be going through after being blocked.

**Screenshots / Videos**

https://github.com/pagefaultgames/pokerogue/assets/62030095/54749570-b615-4ff4-93aa-0855945ae2f6

**Device**
Chrome

**Additional context**

See code below for where the attributes are being applied (L2947-L2948)
https://github.com/pagefaultgames/pokerogue/blob/15671a6922dadbfe905502fa6df0e6fc8f6b29a1/src/phases.ts#L2947-L2950

innerthunder:
These interactions should be fixed with https://github.com/pagefaultgames/pokerogue/pull/2731

PR #2731, opened 2024-06-30
[Move] Refactor Conditional Protection and Protect interactions

innerthunder:
## What are the changes?

This change refactors the processing of protection moves in the `MoveEffectPhase` such that Conditional protection moves (e.g. Quick Guard) are processed and applied in the same way as Protect. This PR also fixes the interactions of some moves when hitting into a protected Pokemon.

The checks to conditional protection moves are currently awkwardly placed in `Pokemon.apply()`, and as a result they cause issues mostly having to do with performance and readability:
- With the way they are currently implemented, conditional protection effects need a redundant check on whether the opponent ignores protection before they're applied.
- Because they are not resolved within `MoveEffectPhase`, a Pokemon protected by one of these conditional protection moves isn't recognized as being protected in the same way as Protect (or King's Shield, etc.).
- Each subclass (e.g. `QuickGuardTag`) is currently applied separately when it would be more concise to apply all effects under their superclass (i.e. `ConditionalProtectTag`).

In turn, this PR aims to make protection checks less fragmented and easier to read. This PR also fixes interactions that were reported to be incorrect in https://github.com/pagefaultgames/pokerogue/issues/2689

## What did change?

- `data/arena-tag`:
  - The condition functions for `ConditionalProtectTags` now have a uniform set of parameters (the Arena and the ID of the move used).
  - New `ignoresBypass` field in `ConditionalProtectTag` determines whether or not the conditional protection effect also blocks moves that ignore other forms of protection (e.g. Curse). This field is currently only `true` for Crafty Shield.
  - The `ConditionalProtectTag`s' condition functions have been reorganized and documented.
- `phases.ts`: In `MoveEffectPhase`, `ConditionalProtectTags` are now factored into the logic for `isProtected` instead of within `Pokemon.apply()`.
- `data/move`: Fixed the effects of the following moves incorrectly applying when hitting a protected Pokemon:
  - Recharging moves (e.g. Hyper Beam)
  - Ceaseless Edge
  - Stone Axe
  - Jaw Lock
  - Glaive Rush
- `test/moves`: New integration tests to check the functionality of Protect and conditional protection moves:
  - `protect.test.ts`
  - `quick_guard.test.ts`
  - `crafty_shield.test.ts`
  - `mat_block.test.ts`
  - `wide_guard.test.ts`
- `test/moves/glaive_rush.test`: New test to check the interaction between Glaive Rush and Protect.

### Screenshots/Videos

N/A

## How to test the changes?

`npm run test protect`
`npm run test quick_guard`
`npm run test crafty_shield`
`npm run test mat_block`
`npm run test wide_guard`
`npm run test glaive_rush`

## Checklist


torranx (approved):
looks good

frutescens (approved):
Looks good. If it's alright with you, I think it would be nice to add a Bulbapedia link / other references used to the move / ability / whatever in the comments of the test for easier future reference - but again, that's optional

DayKev (approved):

Issue #2736, opened 2024-07-01
[BUG] Game Crashes If Wild Tries to Steal a Thick Club


OreoCupcakes:
**Describe the bug**
If a wild (maybe enemy trainer also) Pokemon uses Thief (and I'm going to assume Covet and any other thieving move) on a Cubone/Marowak holding a Thick Club (as its only item?) the game will freeze up.

**To Reproduce**
1. Have a Cubone with a Thick Club.
2. Encounter a wild Pokemon with Thief. 
3. Have the wild Thief the Cubone and trigger the item steal effect.
4. Game freezes up right after.

**Expected behavior**
The wild Pokemon should steal the Thick Club and hold onto it.

**Screenshots / Videos**
https://puu.sh/Ka0ql/4f2509cf3c.mp4

**Device**
Windows 10 22H2, Firefox

PR #2747, opened 2024-07-01
[Bug] Thief Causing Crash with Species Stat Boosters

xsn34kzx:
## What are the changes?

Fixes an issue where using `Thief` on any item from the `SPECIES_STAT_BOOSTER` generator and successfully stealing the held item would cause a crash.

As brought up by #2736, this bug would cause the game to hang. Even though `Thick Club` was the only item to be reported in the issue, investigation showed that this would happen with any of the items, which makes the chance to experience stalling higher than if it was just one item and increases the necessity to address the bug.

## What did change?

As it turns out, the issue stemmed on the items writing their `SpeciesStatBoosterItem` key to their `id`, which, upon calling `generateType`, was set to their `generatorId` with `id` being replaced by `SPECIES_STAT_BOOSTER`, the identifier for the generator in the `modifierTypes` array. This would cause an issue where the `generatorId` was used instead of the `id` to infer the tier of the necessary modifier type, causing the `stolenItem` to be `null`. As a result, the fix was simply not assigning the `key` to the `id`.

However, I had already previously noticed that other generator items like `TEMP_STAT_BOOSTER` never set `generatorId` at all, making it perpetually `null` in saves and never used on reload. As such, I also took the opportunity to remove `generatorId` from game logic and saving/loading logic, since it doesn't do anything besides defer back to `id` in all cases.

### Screenshots/Videos

Compared to #2736, `Thief` now works as intended without causing the game to hang.

https://github.com/pagefaultgames/pokerogue/assets/109637146/583e824a-1c65-44d5-9034-0ec2378f93f6

## How to test the changes?

To test a similar scenario to the issue, a new run was started with Cubone as the only starter along with the following overrides:
- To force `Thick Club` to be held as the only held item,
```ts
export const STARTING_HELD_ITEMS_OVERRIDE: Array<ModifierOverride> = [{ name: "SPECIES_STAT_BOOSTER", type: "THICK_CLUB" }]; 
```
- To force a wild Pokemon to use `Thief`,
```ts
export const OPP_MOVESET_OVERRIDE: Array<Moves> = [ Moves.THIEF, Moves.THIEF, Moves.THIEF, Moves.THIEF ];
```

Because `Thief` has a 30% chance to steal an item, multiple tries might be needed before the interaction involving the item stealing logic occurs.

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -45,7 +45,6 @@ type NewModifierFunc = (type: ModifierType, args: any[]) => Modifier;
 
 export class ModifierType {
   public id: string;
-  public generatorId: string;
   public localeKey: string;
   public iconImage: string;
   public group: string;
@@ -102,7 +101,7 @@ export class ModifierType {
         if (!pool.hasOwnProperty(tier)) {
           continue;
         }
-        if (pool[tier].find(m => (m as WeightedModifierType).modifierType.id === (this.generatorId || this.id))) {
+        if (pool[tier].find(m => (m as WeightedModifierType).modifierType.id === this.id)) {
           return (this.tier = tier);
         }
       }
@@ -133,7 +132,6 @@ export class ModifierTypeGenerator extends ModifierType {
   generateType(party: Pokemon[], pregenArgs?: any[]) {
     const ret = this.genTypeFunc(party, pregenArgs);
     if (ret) {
-      ret.generatorId = ret.id;
       ret.id = this.id;
       ret.setTier(this.tier);
     }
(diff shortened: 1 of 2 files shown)


flx-sta:
@xsn34kzx please [link the issue](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) that would be closed with this properly:

xsn34kzx:
> @xsn34kzx please [link the issue](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) that would be closed with this properly!

@flx-sta Done 👍🏼

Tempo-anon (approved):

PR #2754, opened 2024-07-02
Add New Challenges

Xavion3:
This adds several new challenges. It is currently very much still a work in progress, but is created here as a draft PR for visibility and early testing.


brain-frog:
Can each challenge be a separate PR please

Issue #2756, opened 2024-07-02
[BUG] Incorrect displayed type effectiveness for Sheer Cold vs. an Ice type


AndrioCelos:
**Describe the bug**

The type hint for Sheer Cold on an Ice-type target shows `falsex`.

I'm not sure whether this applies to other one-hit KO moves.

**To Reproduce**
1. Have a Pokémon that knows Sheer Cold.
2. Enter a battle against an Ice-type Pokémon.
3. Ensure type hints are on in the display settings.
4. Move the cursor to Sheer Cold in the Fight menu.

**Expected behavior**

It should display 0× (because Ice-type Pokémon are immune to Sheer Cold). Unsure what multiplier should be displayed for non-boss targets that aren't immune to the move, since type effectiveness doesn't apply there either.

**Screenshots / Videos**
![image](https://github.com/pagefaultgames/pokerogue/assets/6897624/ca628bc3-97fd-448c-9b2a-8cbeab1474ad)

**Device**
Firefox v127.0.2 on Windows 11

hayuna:
FYI to someone who will be fixing this issue

class: Pokemon
method: getAttackMoveEffectiveness
function inside method: applyMoveAttrs returns boolean instead of number

PR #2757, opened 2024-07-02
[Bug] Fix incorrect move effectiveness

hayuna:
## What are the changes?
Correct type for move effectiveness

Fix https://github.com/pagefaultgames/pokerogue/issues/2756

## What did change?
Convert incorrect type (Boolean into Number)

### Screenshots/Videos
Before:
345063523-ca628bc3-97fd-448c-9b2a-8cbeab1474ad

After:
Screenshot 2024-07-02 at 17 34 40

## How to test the changes?
```javascript
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.SHEER_COLD];
export const OPP_SPECIES_OVERRIDE: Species | integer = Species.JYNX;
```

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1155,10 +1155,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       applyPreDefendAbAttrs(MoveImmunityAbAttr, this, source, move, cancelled, typeMultiplier, true);
     }
 
-    if (typeof typeMultiplier.value === "boolean") {
-      typeMultiplier.value = Number(typeMultiplier.value);
-    }
-    return (!cancelled.value ? typeMultiplier.value : 0) as TypeDamageMultiplier;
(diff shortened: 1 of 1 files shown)


src/field/pokemon.ts
```diff
@@ -1155,6 +1155,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       applyPreDefendAbAttrs(MoveImmunityAbAttr, this, source, move, cancelled, typeMultiplier, true);
     }
 
+    if (typeof typeMultiplier.value === "boolean") {
+      typeMultiplier.value = Number(typeMultiplier.value);
+    }
```
flx-sta, line 1160: that's an odd approach. why the specific boolean check which then leads to a number cast? why not always a number cast and check for `isNaN()`?

hayuna: You are right, just casting variable into number should be enough. `FALSE -> 0`

Issue #2758, opened 2024-07-02
[BUG] Sketch copies move from previous turn even if Smeargle moves second


dorox1:
**Describe the bug**

When Smeargle uses Sketch, it will copy the move used by the target on the previous turn, even if the target has used another move since then.

**To Reproduce**

1. Have Smeargle on your team and any non-Smeargle Pokemon as your lead.
2. Find an opposing Pokemon **with a higher speed than Smeargle**
3. Start the battle with the non-Smeargle Pokemon and immediately switch to Smeargle
4. Opponent uses **Move 1**
5. Select Sketch as your move for the turn
6. Opposing Pokemon uses **Move 2**
7. If **Move 1** and **Move 2** are different, Smeargle will Sketch the earlier **Move 1**

**Expected behavior**

Sketch should copy the most recently used move by the opposing Pokemon, even if it was used the same turn.

**Device**

Chrome Version 126.0.6478.116

**Additional context**

This problem may be difficult to replicate, as it relies on both facing a sufficiently fast opposing Pokemon and the AI's move selection being different between turns.

schmidtc1:
Seems like SketchAttr currently pulls the target's move history and copies the first move in the history, instead of the last move. So, it's not related to switching actually, it will just always get the first move used in the opponent's history. I'll open a PR to fix

https://github.com/pagefaultgames/pokerogue/blob/13b23289d1203470271b56deeab35e9f07deeca1/src/data/move.ts#L5151-L5156

Tempo-anon:
This has been fixed

PR #2759, opened 2024-07-02
[BUG] Fixes Sketch copying the first move used by the opponent instead of the last

schmidtc1:
## What are the changes?

Sketch currently copies the first move used in a turn instead of the most recent.
#2758 exposed this issue, but it is not due to switching out like the issue states. Instead, it just always copies the first move used in the opponent's history.

Fixes a bug with Sketch.

## What did change?

SketchAttr now gets the last move used by the opponent, instead of the entire move history.

### Screenshots/Videos

Before fix (copies first move used - Gust)

https://github.com/pagefaultgames/pokerogue/assets/62030095/7059caba-cf6c-40ac-bccc-a7b0378f811f

After fix (copies last move used - Absorb)

https://github.com/pagefaultgames/pokerogue/assets/62030095/71cd14ec-b748-4532-9f4e-010ab1d97ff6

## How to test the changes?

Give opponent different moves, select Smeargle as your starter (preferably with other moves if Smeargle is faster than opponent). Let opponent use different moves. Smeargle uses Sketch on opponent. Smeargle should copy latest move used.

## Checklist
- [x] Are the changes visual?


OrangeRed (approved):

src/data/move.ts
```diff
@@ -5148,12 +5148,12 @@ export class SketchAttr extends MoveEffectAttr {
       return false;
     }
 
-    const targetMoves = target.getMoveHistory().filter(m => !m.virtual);
-    if (!targetMoves.length) {
+    const [targetMove] = target.getLastXMoves(1);
+    if (!targetMove || targetMove.virtual) {
       return false;
     }
```
OrangeRed, line 5154: Some further discussion with the team and `TurnMove.virtual` actually means moves that were spawned by other moves like Metronome. With that being said can you revert back to the previous implementation and just access the last move via `Array.at(-1)`

OrangeRed (changes requested):

src/data/move.ts
schmidtc1: Change to this then? That would allow it to copy Metronome by skipping the virtual moves [which seems to be intended](https://bulbapedia.bulbagarden.net/wiki/Sketch_(move)#:~:text=Sketch%20can%20now%20be%20used%20to%20copy%20Metronome%2C%20Mimic%2C%20and%20Transform%20like%20any%20other%20move%20(without%20the%20need%20for%20the%20move%20to%20fail%20due%20to%20sleep).)
```typescript
    const targetMoves = target.getMoveHistory().filter(m => !m.virtual);
    if (!targetMoves.length) {
      return false;
    }

    const sketchedMove = allMoves[targetMoves.at(-1).move];
```

OrangeRed: just do this to get the move instead of handling the array of moves
`const targetMoves = target.getMoveHistory().filter(m => !m.virtual).at(-1);`

OrangeRed (approved):

OrangeRed:
@schmidtc1 can you also as a separate PR add a tag or check so that sketch fails when its used against unsketchable moves

schmidtc1:
> @schmidtc1 can you also make a separate PR add a tag or check so that sketch fails when its used against unsketchable moves

Like a move flag applied to all the unsketchable moves, and a check if the move has the flag? Or do you suggest a better way to do this?

OrangeRed:
@schmidtc1 something along those lines. Similar to how we have unswappable abilities

schmidtc1:
> @schmidtc1 something along those lines. Similar to how we have unswappable abilities

Started here: #3013. Will work on it further.

PR #2777, opened 2024-07-03
[Bug] Fixed Walking Wake and Iron Leaves Starter Moves

damocleas:
Walking Wake and Iron Leaves level 1 moves were incorrectly labelled '0' instead of '1' resulting in no starter moves after relearner only support added.


flx-sta:
closed in favor of https://github.com/pagefaultgames/pokerogue/pull/2776

PR #2779, opened 2024-07-03
[QoL][Refactor] replace `-1` with `RELEARN_MOVE` and `0` with `EVOLVE_MOVE`

flx-sta:
## What are the changes?

replaced magic numbers with descriptive constansts in `pokemon-level-moves.ts`

makes it more clear what the numbers are user for
image

## How to test the changes?

if it builds & runs all good

## Checklist

### DIFF
--- a/src/data/pokemon-level-moves.ts
+++ b/src/data/pokemon-level-moves.ts
@@ -15,6 +15,9 @@ interface PokemonSpeciesFormLevelMoves {
   [key: integer]: PokemonFormLevelMoves
 }
 
+const RELEARN_MOVE = -1;
+const EVOLVE_MOVE = 0;
+
 export const pokemonSpeciesLevelMoves: PokemonSpeciesLevelMoves = {
   [Species.BULBASAUR]: [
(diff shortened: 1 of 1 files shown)


Tempo-anon (dismissed):
Should add a comment to these `const`s

flx-sta:
> Should add a comment to these `const`s

Will do

flx-sta:
@Tempo-anon Done

src/data/pokemon-level-moves.ts
```diff
@@ -15,6 +15,11 @@ interface PokemonSpeciesFormLevelMoves {
   [key: integer]: PokemonFormLevelMoves
 }
 
+/** Moves that can only be learned with a memory-🍄 */
```
Tempo-anon, line 18: It's funny but let's not start adding emojis into files

flx-sta:
```suggestion
/** Moves that can only be learned with a memory-mushroom */
```

Tempo-anon (approved):

PR #2782, opened 2024-07-04
[QoL] Allow Own Tempo Rockruff to spawn in the wild.

Smewkie:
## What are the changes?
Adds Rockruff to battle-scene.ts in the section for wild pokemon form spawns, assume it's an oversight that it's not there at all due to Own Tempo's original implementation(it didn't work and wasn't a form previously).

This will give any Rockruff that spawns a 50/50 chance to have Own Tempo form, I don't see this as being much of a buff, as Froakie and it's line have the same logic for Battle Bond, which is arguably a much better ability and pokemon combo.

Arguably, for "balance", this could be changed to be a 1/3 chance instead, if the balance team would rather have that, that's easily doable I think?

Currently, own tempo form was locked to finding a lycanroc that had the form, which is a little obtuse, since a dusk form lycanroc can currently spawn without having the form.

## What did change?
Added one line to battle-scene.ts, line 1225.

### Screenshots/Videos
image

## How to test the changes?
Set your enemy's pokemon species in overrides to be rockruff
catch some rockruffs
some of them have own tempo! yay!

## Checklist
- [Y] There is no overlap with another PR?
- [Y] The PR is self-contained and cannot be split into smaller PRs?
- [Y] Have I provided a clear explanation of the changes?
- [Y] Have I tested the changes (manually)?
- [Y] Are the changes visual?
  - [Y] Have I provided screenshots/videos of the changes?


killeroid356:
Just asking if this is planned to be merged anytime soon? really want to use dusk lycanroc :)

Smewkie:
> Just asking if this is planned to be merged anytime soon? really want to use dusk lycanroc :)

I haven't bothered to move it to the beta yet, there were concerns from the dev team that this would shoehorn too many wild rockruffs into being dusk form, so I need to learn how to lower the chance of own tempo form

Issue #2790, opened 2024-07-04
[BUG] Metronome calling Beak Blast hardlocks the game


EJS2K:
**A pokemon using metronome, calling beak blast, hard locks the game (complete freeze, need to reload page).** I tried doing it more than once, seems to happen every time (since the same move keeps getting called).

**To Reproduce**

1. Get a pokemon with metronome (i used munchlax).
2. Use metronome.
3. Get beak blast.
4. Crash.

**Expected behavior**
Move working and doing damage

**Device**
This was done on mobile, not tested on PC. Using google chrome.

**Additional context**
To note the exact battle flow:
Lead with hydreigon against a donphan on endless mode
Swapped to munchlax when given the prompt
Chose metronome
Donphan went first, used scary face
Metronome went off, beak blast got selected.
Beak blast "charging" animation goes off, and then it freezes.

schmidtc1:
The issue is coming from this line of code which is called from ChargeAttr:

PR #2819, opened 2024-07-05
[BUG] Fixes bug with Metronome freezing the game

schmidtc1:
## What are the changes?

#2790
Metronome calling a ChargeAttr move (e.g. Beak Blast) will freeze the game.

Fixes a bug that freezes the game.

## What did change?

I implemented a check to see if the user has the move in their moveset, if not create a dummy PokemonMove for the case of Metronome and pass it into pushMovePhase.

### Screenshots/Videos

Before fix (game freezes from getMove)

https://github.com/pagefaultgames/pokerogue/assets/62030095/05761910-0c3d-4c66-8d1a-0ac40c5c226e

After fix 

https://github.com/pagefaultgames/pokerogue/assets/62030095/34db36ab-a938-4110-9664-d9e39ca102b0

## How to test the changes?

Give your pokemon Metronome and manipulate the moveId variable in RandomMoveAttr to always use Beak Blast 
``` const moveId = moveIds[585];```
Use Metronome with Beak Blast, see that game continues properly and Beak Blast goes off.

## Checklist
- [x] Are the changes visual?

### CODE REVIEW

File: src/data/move.ts [hunk 1]
```diff
@@ -2215,7 +2215,11 @@ export class ChargeAttr extends OverrideMoveEffectAttr {
           user.pushMoveHistory({ move: move.id, targets: [ target.getBattlerIndex() ], result: MoveResult.OTHER });
           user.getMoveQueue().push({ move: move.id, targets: [ target.getBattlerIndex() ], ignorePP: true });
           if (this.sameTurn) {
-            user.scene.pushMovePhase(new MovePhase(user.scene, user, [ target.getBattlerIndex() ], user.moveset.find(m => m.moveId === move.id), true), this.followUpPriority);
+            let movesetMove = user.moveset.find(m => m.moveId === move.id);
+            if (!movesetMove) { // account for Metronome usage, move does not exist in moveset
```
Tempo-anon, line 2219: This isn't exclusive for metronome right? This can also happen for moves like assist and natural power

schmidtc1: I'm not sure nature power uses any ChargeAttr moves, but it would work for assist. Although testing now with assist, it seems to only take the user's moveset rather than the entire party's movesets, which I don't think is correct. 
https://bulbapedia.bulbagarden.net/wiki/Assist_(move)

PR #2743, opened 2024-07-01
[BUG] Mutlihit Moves + Messages not properly rendering on faint

DustinLin:
## What are the changes?

This is a revision of pr #783 and address issues from #680 that are still present.

There are 2 existing bugs that are still present which have to do with rendering dialogue when fainting:
1. Multi hit moves messages: when a multi-hit move ends early from fainting a pokemon, the message is displayed out of order:
2. "Move messages": for example the absorb dialogue is displayed out of order when fainting a pokemon

https://github.com/pagefaultgames/pokerogue/assets/39450497/1cff9ec8-c893-46ab-af26-e85464c17cdd

https://github.com/pagefaultgames/pokerogue/assets/39450497/f66aa5dc-58df-436a-a215-4be4cfc8043a

## Why am I doing these changes?

Unresolved stuff from #680 and revising pr #783

## What did change?

Added extra conditionals to render the dialogue correctly, as well as documentation of `phase.ts` functions

### Screenshots/Videos

Absorb dialogue properly fixing:

https://github.com/pagefaultgames/pokerogue/assets/39450497/dee73788-9579-4a05-afe5-94390982806f

Multi-hit moves properly fixing:

https://github.com/pagefaultgames/pokerogue/assets/39450497/d58f183b-4e28-481a-9ac2-9005273af582

https://github.com/pagefaultgames/pokerogue/assets/39450497/5ed163a1-a452-4898-a7be-638f731789dc

## How to test the changes?

Override with multi-hit moves and check if *when the move faints* the dialogue is in the proper order.

## Checklist
- [ ] There is no overlap with another PR?
There is overlap with #783, but I wrote that and am closing it
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #2820, opened 2024-07-05
[Bug] fixing game breaking u-turn faint switch interaction

### LINKED ISSUE #2680: [BUG] U-turn incorrectly applies some effects to the switched in Pokemon
**Describe the bug**

When a Pokemon uses U-turn, some strange (and potentially game-breaking) interactions can take place:
- If the user attacks into a target with Rough Skin or Iron Barbs and faints on U-turn's last hit, the "Faint" status effect is applied to the switched in Pokemon instead of the user, causing the switched in Pokemon to also "faint" (see the video below).
- If the user attacks into a target with Flame Body, and the user would get burned on U-turn's last hit, the switched in Pokemon is burned instead of the user.

**To Reproduce**

1. Have a Pokemon with U-turn and 3x Multi-Lens (can be done via overrides)
2. Give the enemy Pokemon Rough Skin and Iron Barbs as its ability and passive (via overrides)
3. Attack into the enemy Pokemon with U-turn

**Expected behavior**

The effects listed above should be applied to U-turn's user, not its switch target.

**Screenshots / Videos**

Here's a video of U-turn's behavior when the user faints while using it:

https://github.com/pagefaultgames/pokerogue/assets/168692175/40fb9862-2c07-4eee-93d7-c4cdae02c594

**Device**

PC (Chrome)

**Additional context**

I already tried to fix this by adding another `MoveEffectTrigger` type for U-turn's attribute, but the same problems as before were observed. I think the problem is related to `FaintPhase` and `ObtainStatusEffectPhase` using battler indexes instead of the Pokemon themselves to apply effects, but I'm not sure what a clean fix to that would look like.

DustinLin:
## What are the changes?

The proper interaction now occurs (using pokemon now faints, and a new one is swapped in).

The dialogue boxes are still out of order from fainting, but hopefully #2743 addresses the problem.

Closes #2680 

## What did change?

Currently the `SwitchSummonPhase` is being added in the phase queues' in the wrong order for multi hit moves. It is being added too early (before the `DamagePhase`  and `FaintPhase` of the last multi-hit), which is causing the switch-in faint issue of #2680. I've written a call to add the phase right before the `MoveEndPhase` (which ensures that it wont interfere with anything of the multi-hit move, and the swapped in 'mon is present for the end-of-turn effects such as weather, abilities, field hazards).

I tried playing around with the existing functions, namely playing around with `phaseQueueSpliceIndex` to manipulate the order of the phases being added. The problem with this is that queuing ability-effects related phases also calls `setPhaseQueueSplice()`, which messes up the idea of trying to add all the phases before `SwitchSummonPhase`.

The solution I've come up with relies on the fact that `PhaseQueue` contains a `MoveEndPhase` (since it essentially does a linear search for it). I believe this should always be true, since these force-switches/damages occur during Moves. However this should be tested/confirmed.

### Screenshots/Videos

(can look at the `main` branch behavior from the linked issue)

- pr fix:

https://github.com/pagefaultgames/pokerogue/assets/39450497/ba14269c-6395-43ca-8953-de43b007accb

- interaction with weather, abilities:

https://github.com/pagefaultgames/pokerogue/assets/39450497/39bf64a4-0cfb-4968-b14f-5b5b9787d20f

- interaction with hazards:

https://github.com/pagefaultgames/pokerogue/assets/39450497/836ad2a8-83df-4a7a-89dd-f8dc15f66270

- enemy fixes as well

https://github.com/pagefaultgames/pokerogue/assets/39450497/3acd5c45-83fe-4e4c-9174-d902aba2309e

## How to test the changes?

3x Multi-Lens, and with the abilities Rough Skin and Iron Barbs gives the fainting from move effect (i used level 5 vs 15 to ensure it faints/survives)

I've only tested all the above with U-turn, we should make sure the other moves which use the ForceSwitch attribute still behave as normal

## Checklist
-[x] Have I tested the changes (manually)?
- [x] Are the changes visual?

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -67,7 +67,7 @@ import { Species } from "#enums/species";
 import { UiTheme } from "#enums/ui-theme";
 import { TimedEventManager } from "#app/timed-event-manager.js";
 
-export const bypassLogin = import.meta.env.VITE_BYPASS_LOGIN === "1";
+export const bypassLogin = true;
 
 const DEBUG_RNG = false;
 
@@ -180,8 +180,10 @@ export default class BattleScene extends SceneBase {
 
   public phaseQueue: Phase[];
   public conditionalQueue: Array<[() => boolean, Phase]>;
-  private phaseQueuePrepend: Phase[];
-  private phaseQueuePrependSpliceIndex: integer;
+  public phaseQueuePrepend: Phase[];
+
+  public phaseQueuePrependSpliceIndex: integer;
+
   private nextCommandPhaseQueue: Phase[];
   private currentPhase: Phase;
   private standbyPhase: Phase;
@@ -2062,6 +2064,29 @@ export default class BattleScene extends SceneBase {
     } else {
       this.pushPhase(movePhase);
     }
+  }
+
+  /**
+   * Tries to add the input phase to right before the MoveEndPhase, useful for resolving moves before the end-of-turn phases start
+   * @param phase {@linkcode Phase} the phase to be added
+   * @returns boolean if a MoveEndPhase was found
+   */
+  prependToMoveEndPhase(phase: Phase): boolean {
+    let flag = false;
+    // want to add the pahse after MoveEnd of U-turn? so it can ge the effects of whatever
+    for (let i = 0; i < this.phaseQueue.length; i++) {
+      if (this.phaseQueue[i].constructor.name === "MoveEndPhase") {
+        this.phaseQueue.splice(i, 0, phase);
+        flag = true;
+        break;
+      }
+    }
+    if (flag === false) {
+      // what to do here if there is none?
+      // it should be the case that during combat there is always a MoveEndPhase when a move is being used, extensive testing/reasoning needed
+      this.unshiftPhase(phase);
+    }
+    return flag;
   }
 
   queueMessage(message: string, callbackDelay?: integer, prompt?: boolean, promptDelay?: integer, defer?: boolean) {
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -67,7 +67,7 @@ import { Species } from "#enums/species";
 import { UiTheme } from "#enums/ui-theme";
 import { TimedEventManager } from "#app/timed-event-manager.js";
 
-export const bypassLogin = true;
+export const bypassLogin = import.meta.env.VITE_BYPASS_LOGIN === "1";
 
 const DEBUG_RNG = false;
 
@@ -180,10 +180,8 @@ export default class BattleScene extends SceneBase {
 
   public phaseQueue: Phase[];
   public conditionalQueue: Array<[() => boolean, Phase]>;
-  public phaseQueuePrepend: Phase[];
-
-  public phaseQueuePrependSpliceIndex: integer;
-
+  private phaseQueuePrepend: Phase[];
+  private phaseQueuePrependSpliceIndex: integer;
   private nextCommandPhaseQueue: Phase[];
   private currentPhase: Phase;
   private standbyPhase: Phase;
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -2065,26 +2065,23 @@ export default class BattleScene extends SceneBase {
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/battle-scene.ts
```diff
@@ -2064,6 +2064,29 @@ export default class BattleScene extends SceneBase {
...
+        break;
+      }
+    }
+    if (flag === false) {
+      // what to do here if there is none?
+      // it should be the case that during combat there is always a MoveEndPhase when a move is being used, extensive testing/reasoning needed
+      this.unshiftPhase(phase);
+    }
+    return flag;
+  }
```
innerthunder, line 2088:
```suggestion
    const targetIndex = this.phaseQueue.findIndex(ph => ph instanceof targetPhase);
    if (targetIndex !== -1) {
        this.phaseQueue.splice(targetIndex, 0, phaseToAdd);
        return true;
    } else {
        this.unshiftPhase(phaseToAdd);
        return false;
    }
  }
```

To go along with the change to the function definition (also using `findIndex` which performs better than a for loop).

src/data/move.ts
```diff
@@ -4512,7 +4513,8 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
 	  	user.scene.triggerPokemonFormChange(switchOutTarget, SpeciesFormChangeActiveTrigger, true);
 
 	  	if (switchOutTarget.hp) {
-          user.scene.unshiftPhase(new SwitchSummonPhase(user.scene, switchOutTarget.getFieldIndex(), user.scene.currentBattle.trainer.getNextSummonIndex((switchOutTarget as EnemyPokemon).trainerSlot), false, this.batonPass, false));
+        // for opponent switching out
+          user.scene.prependToMoveEndPhase(new SwitchSummonPhase(user.scene, switchOutTarget.getFieldIndex(), user.scene.currentBattle.trainer.getNextSummonIndex((switchOutTarget as EnemyPokemon).trainerSlot), false, this.batonPass, false));
```
innerthunder, line 4517:
```suggestion
          user.scene.prependToPhase(new SwitchSummonPhase(user.scene, switchOutTarget.getFieldIndex(), user.scene.currentBattle.trainer.getNextSummonIndex((switchOutTarget as EnemyPokemon).trainerSlot), false, this.batonPass, false), MoveEndPhase);
```
To go along with the change to the function definition.

src/field/pokemon.ts
```diff
@@ -3094,7 +3094,8 @@ export class PlayerPokemon extends Pokemon {
 
       this.scene.ui.setMode(Mode.PARTY, PartyUiMode.FAINT_SWITCH, this.getFieldIndex(), (slotIndex: integer, option: PartyOption) => {
         if (slotIndex >= this.scene.currentBattle.getBattlerCount() && slotIndex < 6) {
-          this.scene.unshiftPhase(new SwitchSummonPhase(this.scene, this.getFieldIndex(), slotIndex, false, batonPass));
+          // add a SwitchSummonPhase
+          this.scene.prependToMoveEndPhase(new SwitchSummonPhase(this.scene, this.getFieldIndex(), slotIndex, false, batonPass));
```
innerthunder, line 3098:
```suggestion
          this.scene.prependToPhase(new SwitchSummonPhase(this.scene, this.getFieldIndex(), slotIndex, false, batonPass), MoveEndPhase);
```
To go along with the change to the function definition

src/battle-scene.ts
```diff
@@ -2064,6 +2064,29 @@ export default class BattleScene extends SceneBase {
     }
   }
 
+  /**
+   * Tries to add the input phase to right before the MoveEndPhase, useful for resolving moves before the end-of-turn phases start
+   * @param phase {@linkcode Phase} the phase to be added
+   * @returns boolean if a MoveEndPhase was found
+   */
+  prependToMoveEndPhase(phase: Phase): boolean {
```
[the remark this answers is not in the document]
DustinLin: Thanks, great suggestion

innerthunder (approved):

DustinLin:
closing because PR is now pointing to `beta` and that was merged in with #2860

PR #2822, opened 2024-07-06
[Bugfix] Fix a bug during bgm-bar initialization

Dakurei:
## What are the changes?
Found during my previous PR #2603, and pointed out by a user on the Discord [here](https://discord.com/channels/1125469663833370665/1256331738318376960/1256331738318376960)
When the bgm-bar is first initialized, it takes on its maximum size, and remains empty. Most likely a residual behavior from its former mode of operation (when it was supposed to be displayed not only in the pause menu)

## What did change?
I've managed to solve this problem.
I've also made sure that displaying/hiding the bgm-bar is no longer dependent on a reboot.
As well as enabling the bgm-bar by default, as it is now displayed in the pause menu, contrary to what was originally intended when it was disabled by default.

bgm-bar_before

> After
bgm-bar_after

## How to test the changes?
Activate/deactivate the bgm-bar in the options, and check that the behavior is normal.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/system/settings/settings.ts
+++ b/src/system/settings/settings.ts
@@ -493,7 +493,7 @@ export const Setting: Array<Setting> = [
     key: SettingKeys.Show_BGM_Bar,
     label: i18next.t("settings:showBgmBar"),
     options: OFF_ON,
-    default: 0,
+    default: 1,
     type: SettingType.DISPLAY
   },
   {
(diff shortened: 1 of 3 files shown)


Dakurei:
By the way, the preview for the Legacy, even if its problem was solved in the previous PR, and this bugfix remains consistent regardless of the theme chosen, it's always better to display the 2 modes

> Before
bgm-bar_before_legacy

> After
bgm-bar_after_legacy

Tempo-anon (approved):

PR #1690, opened 2024-06-01
[Feature][Help/Discussion Wanted] Add Mystery Encounter Framework with MVP Encounters

ben-lear:
## What is a Mystery Encounter (ME)?
A random event that has a chance to take the place of a **wild** Pokemon encounter. These events could be anything like: a non-combat shop, a trade event with another trainer, a forced custom battle, a Faustian bargain event, or anything else you can dream of/code. They have more dialogue, choices, and non-battle player interaction than any standard floor outside of perhaps the final boss.

We have a beta site to test on! 
https://pokerogue-encounters-beta.netlify.app/

## How do Mystery Encounters Happen?
Encounters can only occur between floors 10 and 180 of Classic at time of writing. They spawn through a rand calculation to completely replace a **wild** Pokemon encounter. Biome bosses, rival battles, fixed battles, and standard trainer battles will never be replaced. The calculation for determining a spawn chance is a bit complicated ([see spawning logic here](https://github.com/pagefaultgames/pokerogue/pull/1690/files#diff-4b1a5f9160d1d8aecce0db85774f20e3c254f97c27cc0d0473911d5cbd565396R2768-R2786)), but you can expect that about 95% of Classic runs will have between 6-10 Mystery Encounter spawns throughout a full run. Also, there is an inverted weight system for encounter rarities, where common/uncommon encounters are more likely at the beginning of a run and rare/super rare encounters are more likely at the end of a run ([see here](https://github.com/pagefaultgames/pokerogue/pull/1690/files#diff-4b1a5f9160d1d8aecce0db85774f20e3c254f97c27cc0d0473911d5cbd565396R2768-R2786)).

Here is the current distribution of expected encounters over a Classic Run:
image

The spawn rate is implemented a bit conservatively, in the ME team's opinion. We think that it may be more beneficial to actually have a higher overall encounter number eventually (perhaps 12-15/run), but we also didn't want to greatly upset the current balance of replacing wild encounters too drastically at first.

## Why are Mystery Encounters a Good Thing^TM?
These events, by their random nature, and often times rare occurrence, are a great way to serve as a foundation for keeping every single run fresh, unique, and fun. From a design standpoint, this is essentially copying beat-for-beat another titular roguelike's encounter design, Slay the Spire. As proven there and elsewhere, feeding the player continuous choices makes for much more rewarding/engaging experiences.

## The Meat of the PR
[...]

PR #2824, opened 2024-07-06
Mystery Encounters

ben-lear:
See Main PR for details: https://github.com/pagefaultgames/pokerogue/pull/1690


returntoice:
Would you please add other languages' locales folder for translating into the other languages? If you make a new PR for that, then the translation team might request translations on the PR

PR #2825, opened 2024-07-06
[Bug] Disable Egg Gacha option in shop

Nohbody:
## What are the changes?
1. This prevents the "Egg Gacha" from being opened in the Shop screen.

1. There's a Candy exploit where a user can repeatedly catch a Pokemon, open the Egg Gacha, pull an egg to have the game force save the session (with the caught Pokemon and Candy), and then reload to have the scene reset to before they caught the Pokemon (but keeping the caught Pokemon and Candy).

## What did change?
1. The exploit is no longer possible because the Egg Gacha save won't be accessible until after the session has been dealt with otherwise (e.g. exited or saved by the next wave starting).

## How to test the changes?
1. You should no longer be able to open the "Egg Gacha" while the Shop Overlay is open.

https://github.com/pagefaultgames/pokerogue/assets/6127618/7eeef858-2a8b-4922-a15f-256e5fa8296d

## Checklist
- [ ] Are the changes visual?


Opaque02:
One question I have about this is if people want to do things unrelated to egg gatcha in the shop - say changing some settings or looking at achievements/stats? Maybe a better option would be to allow the menu to be opened but the egg gatcha to not be enabled while within the shop?

Nohbody:
> One question I have about this is if people want to do things unrelated to egg gatcha in the shop - say changing some settings or looking at achievements/stats? Maybe a better option would be to allow the menu to be opened but the egg gatcha to not be enabled while within the shop?

They can just access the gacha and settings at the start of the next battle. I don't see why they'd have to do those things while in the shop.

Opaque02:
> > One question I have about this is if people want to do things unrelated to egg gatcha in the shop - say changing some settings or looking at achievements/stats? Maybe a better option would be to allow the menu to be opened but the egg gatcha to not be enabled while within the shop?
> 
> They can just access the gacha and settings at the start of the next battle. I don't see why they'd have to do those things while in the shop.

Another option could be if someone wants to quit to restart - say they want to savescum and instead of reloading, they would normally quit to menu from within the shop menu? Or if you accidentally release a mon by accident and don't want to reload?

I bring this up because of a scenario I had earlier was that there was a bug that I found and I wanted to get my session data so I could raise a bug report. I went to open the menu from within the shop (where the bug appeared) and downloaded the session data. With this PR as is, I'd have to reload the page and get the session data from the start of the battle, whereas if you could disable the egg gacha from the menu during the shop, you'd still be able to do all this stuff, just without accessing the gacha?

Nohbody:
> > > One question I have about this is if people want to do things unrelated to egg gatcha in the shop - say changing some settings or looking at achievements/stats? Maybe a better option would be to allow the menu to be opened but the egg gatcha to not be enabled while within the shop?
> > 
> > 
> > They can just access the gacha and settings at the start of the next battle. I don't see why they'd have to do those things while in the shop.
> 
> Another option could be if someone wants to quit to restart - say they want to savescum and instead of reloading, they would normally quit to menu from within the shop menu? Or if you accidentally release a mon by accident and don't want to reload?
> 
> I bring this up because of a scenario I had earlier was that there was a bug that I found and I wanted to get my session data so I could raise a bug report. I went to open the menu from within the shop (where the bug appeared) and downloaded the session data. With this PR as is, I'd have to reload the page and get the session data from the start of the battle, whereas if you could disable the egg gacha from the menu during the shop, you'd still be able to do all this stuff, just without accessing the gacha?

I feel like it'd be a more confusing UX if the only time this menu didn't let you access the Gacha was from within the Shop. I always reset my save state by just reloading the game so it is interesting to hear it might be a legitimate issue for some people's experience.

Opaque02:
> > > > One question I have about this is if people want to do things unrelated to egg gatcha in the shop - say changing some settings or looking at achievements/stats? Maybe a better option would be to allow the menu to be opened but the egg gatcha to not be enabled while within the shop?
> > > 
> > > 
> > > They can just access the gacha and settings at the start of the next battle. I don't see why they'd have to do those things while in the shop.
> > 
> > 
> > Another option could be if someone wants to quit to restart - say they want to savescum and instead of reloading, they would normally quit to menu from within the shop menu? Or if you accidentally release a mon by accident and don't want to reload?
> > I bring this up because of a scenario I had earlier was that there was a bug that I found and I wanted to get my session data so I could raise a bug report. I went to open the menu from within the shop (where the bug appeared) and downloaded the session data. With this PR as is, I'd have to reload the page and get the session data from the start of the battle, whereas if you could disable the egg gacha from the menu during the shop, you'd still be able to do all this stuff, just without accessing the gacha?
> 
> I feel like it'd be a more confusing UX if the only time this menu didn't let you access the Gacha was from within the Shop. It absolutely will be perceived/reported as a bug. I always reset my save state by just reloading the game so it is interesting to hear it might be a legitimate issue for some people's experience.

I started out by getting into making daily guides - there's quite a lot of us that do it and are involved with it, but one of the things is being able to reset quickly if need be; when writing guides you need to make sure you've got good steps, and sometimes it happens that rng or skill goes badly and you need to reset. Most of the time this happens from within the battle itself, but for other times you might accidentally kill a mon you meant to catch, or you might try for some rerolls only to get nothing and need to reset within the shop. I know for a fact that the daily guide making community uses the quit to menu button from the shop every day (no pun intended!), so I know there's definitely people that do use it

Nohbody:
> > > > > One question I have about this is if people want to do things unrelated to egg gatcha in the shop - say changing some settings or looking at achievements/stats? Maybe a better option would be to allow the menu to be opened but the egg gatcha to not be enabled while within the shop?
> > > > 
> > > > 
> > > > They can just access the gacha and settings at the start of the next battle. I don't see why they'd have to do those things while in the shop.
> > > 
> > > 
> > > Another option could be if someone wants to quit to restart - say they want to savescum and instead of reloading, they would normally quit to menu from within the shop menu? Or if you accidentally release a mon by accident and don't want to reload?
> > > I bring this up because of a scenario I had earlier was that there was a bug that I found and I wanted to get my session data so I could raise a bug report. I went to open the menu from within the shop (where the bug appeared) and downloaded the session data. With this PR as is, I'd have to reload the page and get the session data from the start of the battle, whereas if you could disable the egg gacha from the menu during the shop, you'd still be able to do all this stuff, just without accessing the gacha?
> > 
> > 
> > I feel like it'd be a more confusing UX if the only time this menu didn't let you access the Gacha was from within the Shop. It absolutely will be perceived/reported as a bug. I always reset my save state by just reloading the game so it is interesting to hear it might be a legitimate issue for some people's experience.
> 
> I started out by getting into making daily guides - there's quite a lot of us that do it and are involved with it, but one of the things is being able to reset quickly if need be; when writing guides you need to make sure you've got good steps, and sometimes it happens that rng or skill goes badly and you need to reset. Most of the time this happens from within the battle itself, but for other times you might accidentally kill a mon you meant to catch, or you might try for some rerolls only to get nothing and need to reset within the shop. I know for a fact that the daily guide making community uses the quit to menu button from the shop every day (no pun intended!), so I know there's definitely people that do use it

Thanks for the insight. I will rework the pr to just grey out/disable the gacha while in the Shop instead.

Nohbody:
@Opaque02 Changes have been made. Does this work better for you?

Nohbody:
@flx-sta Can this be looked at again?

Nohbody:
@f-fsantos Merge conflict fixed.

src/ui/menu-ui-handler.ts
```diff
@@ -85,17 +85,29 @@ export default class MenuUiHandler extends MessageUiHandler {
     menuMessageText.setWordWrapWidth(1224);
     menuMessageText.setOrigin(0, 0);
 
-    this.optionSelectText = addTextObject(this.scene, 0, 0, this.menuOptions.map(o => `${i18next.t(`menuUiHandler:${MenuOptions[o]}`)}`).join("\n"), TextStyle.WINDOW, { maxLines: this.menuOptions.length });
-    this.optionSelectText.setLineSpacing(12);
+    this.optionContainer = this.scene.add.container();
+    let maxTextWidth = 0;
+    this.optionContainer.add(
+      this.menuOptions.map((o,i) => {
+        const nextOption = addTextObject(this.scene, 0, i*16, i18next.t(`menuUiHandler:${MenuOptions[o]}`), TextStyle.WINDOW);
```
flx-sta, line 92: What does the `*16` stand for ?

src/ui/ui.ts
```diff
@@ -524,6 +524,14 @@ export default class UI extends Phaser.GameObjects.Container {
     });
   }
 
+  public getPrevMode(): Mode {
+    if (this.modeChain.length) {
```
flx-sta, line 528: Is this a not 0 check ? If so please clarify the code.

is mode chain guranteed set ? Or can it be undefined ? If so add a null check

flx-sta (changes requested):

src/ui/ui.ts
Nohbody: What do you mean clarify the code? It's a length 0 check. modeChain is initialized as an empty array as part of the TypeScript declaration.

src/ui/menu-ui-handler.ts
Nohbody: I replaced both with a const named OPTION_HEIGHT. However, it's a bit of a misnomer because the actual text height is controlled by the Text Style, but this will control the size of the placement of the option and where the cursor points anyways (effective the option's rectangle).

flx-sta: fine by me. resolved

src/ui/ui.ts
Nohbody: This is less performant because Javascript is not a compiler-optimized language.

f-fsantos: We value readability over performance.

Nohbody: This is readable though? This is a pretty standard way to do this.

Nohbody: Other people have already done it this way in this codebase.

https://github.com/search?q=repo%3Apagefaultgames%2Fpokerogue%20.length)&type=code

Madmadness65 (approved):
lgtm, disabling the egg gacha option was definitely a better call than disabling the menu all together.

src/ui/ui.ts
f-fsantos: I'm still adamant on this change. 
Length is not a boolean value and we shouldn't do things around the silliness that javascript allows us to do. I understand there are more examples on the whole repo and even in this PR there's another one that wasn't explicitly called out, but this is something we should improve rather than sustain. All the other cases should also change and I hope to tackle them at some point.

But for now, please change both if ...length) to be explicit about what you are creating the condition around.

f-fsantos (changes requested):
Please change both if ...length) to be explicit about what you are creating the condition around.

Nohbody:
> Please change both if ...length) to be explicit about what you are creating the condition around.

I'm not making my code explicitly worse. There's nothing "less readable" about using a length check like I wrote it.

src/ui/ui.ts
Nohbody: This isn't "silliness that javascript allows us do". This is how conditional checks work in the vast majority of programming languages. 0 === false, everything else === true.

Nohbody: I literally did not commit `if (this.scene.gameData.eggs.length)`. Someone else who knows how a conditional check works wrote that.

CodeTappert, line 528: While you are correct that your apporach is correct in javascript. Not everybody knows this concept. For example i work mainly with golang which hasnt those kinds of checks.

Please make the change requested. It does not change the functionality of your code and help other, non javascript experts, to better understand. 

Also a lot of languages (go, c++, c# etc) dont allow the implicit conversion of non boolean types to boolean types. This is why your code, while correct in a javascript world, may confuse other developers that arent experts.

CodeTappert: While you are correct that your apporach is correct in javascript. Not everybody knows this concept. For example i work mainly with golang which hasnt those kinds of checks.

Please make the change requested. It does not change the functionality of your code and help other, non javascript experts, to better understand. 

Also a lot of languages (go, c++, c# etc) dont allow the implicit conversion of non boolean types to boolean types. This is why your code, while correct in a javascript world, may confuse other developers that arent experts.

Nohbody: This isn't a repo written in go, c++, or c#. If you're contributing Javascript code you should know basic Javascript syntax. I don't know why we'd even attempt to try to do what you're suggesting. It's like writing a comment on every line explaining what everything does. Pointless fluff.

flx-sta (changes requested):
Play-Test

The alpha is not updated properly

in-shop
next-wave after shop

Nohbody:
@flx-sta Fixed.

ben-lear:
PR needs to be rebased

f-fsantos:
Closed by #3304

Issue #2844, opened 2024-07-06
[BUG] Managing 2-Move Starter's Move Crashes the Game


Endermage77:
**Describe the bug**
When starting a new game, if you try to change the move of a pokemon that only has one move using the Manage Moves menu, it causes the game to freeze, forcing you to reset. This is most noticeable with Smeargle, who only has Sketch if you haven't yet unlocked any of its egg moves.

**To Reproduce**
1. Begin a **New Game**
2. Select a starter with **exactly one move** available (e.g. Smeargle)
3. Select "**Manage Moves**"
4. Attempt to **swap out the starter's move**
5. Crash occurs

**Expected behavior**
I was personally expecting nothing to happen or a message saying something to the tune of "No other moves available." to display.

**Screenshots / Videos**
The game frozen at the moment of the crash:
![image](https://github.com/pagefaultgames/pokerogue/assets/25313980/2a69181d-d081-4146-86fa-b038162750b3)

When the crash occurs, this error gets thrown in the console: 
![image](https://github.com/pagefaultgames/pokerogue/assets/25313980/20e5cfed-7709-4840-acb1-6022aa26d70b)

**Device**
Playing on a Google Chrome browser on Windows 10 PC.

**Additional context**

lnuvy:
@Endermage77 Hi, I found out that it was a bug and opened a PR for it. Thanks for reporting issue!

PR #2864, opened 2024-07-06
[Bug] handling when only one move in Manage Moves (#2844)

lnuvy:
## What are the changes?
For Pokémon with only one move(skill), the error is due to `move-info-overlay`.

Issue with #2844 

QoL

## What did change?
If a Pokemon with one move chooses Manage Moves, Only the selected move is visible.

### Screenshots/Videos

**Before**

https://github.com/pagefaultgames/pokerogue/assets/92131474/9b3ee649-3cb8-4d25-b090-ba9dc8643ccb

**After**

https://github.com/pagefaultgames/pokerogue/assets/92131474/3ebaf49e-fd04-499b-aa2f-2f3c21a37563

## Checklist
- [ ] Are the changes visual?


lnuvy:
Why do my videos look like links without previews? 😂

EmberCM:
> Why do my videos look like links without previews? 😂

You can embed them like this `name-of-media`

lnuvy:
> > Why do my videos look like links without previews? 😂
> 
> You can embed them like this `name-of-media`

Thanks, but still preview not displaying..

EmberCM:
> > > Why do my videos look like links without previews? 😂
> > 
> > 
> > You can embed them like this `name-of-media`
> 
> Thanks, but still preview not displaying..

You have an extra [ before the text. `[![Before]`
Make sure it's `media` exactly

lnuvy:
> > > > Why do my videos look like links without previews? 😂
> > > 
> > > 
> > > You can embed them like this `name-of-media`
> > 
> > 
> > Thanks, but still preview not displaying..
> 
> You have an extra [ before the text. `[![Before]` Make sure it's `media` exactly

I checked and I don't think there should be any text near the media file apart from the markdown grammar. Thanks for letting me know!

KimJeongSun (changes requested):
in my opinion, it seems like the above method can be used to work around this issue.

https://github.com/pagefaultgames/pokerogue/blob/2ab6e8631deb4f945e579bd4f96f0eac18ddfb6f/src/ui/starter-select-ui-handler.ts#L1190

but It looks like the purpose of this `if` condition is to disable the swap menu when there is only one starter move. However, it seems that this `if` condition is not working as intended.
Could you please check this code and fix it?

lnuvy:
@KimJeongSun Thanks for feedback!
Ditto, Blipbug, etc... meet the conditions mentioned above, but Smeargle has an error due to the following code:

```this.speciesStarterMoves.push(...levelMoves.filter(lm => lm[0] > 0 && lm[0] <= 5).map(lm => lm[1]));```

I checked the pokemonSpeciesLevelMoves object and found that, unlike other Pokémon with only one move, Smeargle had four instances of the same move, Sketch, registered. So, I fixed that part and rolled back the other code.


As a side note) As far as I know, there are five Pokémon with only one move.(Ditto, Unown, Smeargle, Beldum, and Blipbug) Some users might be interested in seeing which move these Pokémon have. I will study this part further and submit another PR for it in the future 😀

KimJeongSun:
First of all, thank you for the quick update. The updated modification method looks very reasonable and clean. 
However, I am unsure why there are 4 instances of `sketch` in the starter move, and I am a bit concerned about potential side effects if we reduce it to just one since I'm not aware of the origin of this configuration. 
If possible, could you track the GitHub log from the time when there were 4 `sketch` instances to understand the reason behind it?
If having 4 sketches is absolutely necessary, then we should not reduce the starter move.

lnuvy:
@KimJeongSun I checked the commit(Hash b054992ffa5c9b9a857bcd35e45b05d995db8a0f) that added the move management function including code
` if (this.speciesStarterMoves.length > 1) {  ... }` to the file `star-select-ui-handler.tsx`, and the commit that Smeargle's move list changed 1 -> 4 (hash 38a518a9e14b3211b3c31b1c1871ab75597592a6) and looked at commits before and after.

In my opinion, this commit(38a518a9e14b3211b3c31b1c1871ab75597592a6) explicitly highlights the file-change about Smeargle in the `pokemon-level-moves.ts` file, so it seems to have been intentionally inserted.

The implementation of the Sketch move itself was completed 15 months ago, and it seems there have been no changes to the implementation since then.

image

It is still a mystery. I don't have much time during weekdays due to my work, but I will share any anomalies I discover.


image

When I tested localhost, seems like fine. but  I'm still not sure why four moves were added to Smeargle's move list. 🫠

DayKev:
Is this obsoleted by #2986 being merged?

PR #2892, opened 2024-07-07
[Bug] Fix #2821 to show correct type hints for abilities like Pixilate

### LINKED ISSUE #2821: [BUG] pokemon.getMoveEffectiveness will not consider your own abilities that affect type effectiveness
**Describe the bug**
BR #2025 and PR #2042 made it so that `pokemon.getMoveEffectiveness` will ignore abilities when using the effectiveness hint function to show how effective are moves against an opponent. But it also made it so that it will ignore your own abilities such as Pixilate.

**To Reproduce**
1. Enter battle with a pokemon with Pixilate ability against, e.g., Sableye
2. move cursor to a normal-type attacking move

**Expected behavior**
The effectiveness hint should show 2x effectiveness with the move highlighted green because it knows the source Pokemon has Pixilate

**Actual behavior**
The move is greyed out and the hint shows 0x effectiveness because `pokemon.getMoveEffectiveness` ignores all abilities and thinks a normal move won't have any effect on the ghost type.

morriskurz:
## What are the changes?
The user now gets correct type hints when their mon has abilities like [Pixilate](https://bulbapedia.bulbagarden.net/wiki/Pixilate_(Ability)). For a full list of abilities that may change a move's type, see [here](https://bulbapedia.bulbagarden.net/wiki/Category:Abilities_that_can_modify_move_types

As it is my first PR for this repo, I would be grateful for better and faster ways to test. In my mind, the test I introduced would not need to set up a game scene, but I could not find a way to instantiate a Pokemon outside of a Scene/Game.

See #2821 for the original issue. It helps players with type hints to correctly assess the effectiveness of their attack.

## What did change?
Add test for the Pixilate ability, tackling a Sableye.
Add a check during `getAttackMoveEffectiveness` for the abilities of the attacker mon.

### Screenshots/Videos
Before, Tackle was type hinted incorrectly when the pokemon had Pixilate (should be super effective against dark, should hit against ghost):
grafik
After: 
grafik

## How to test the changes?
Get a pokemon with Pixilate as starting ability with Tackle move against Sableye. 
Tackle should now be super effective as a Fairy move against Dark.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1149,6 +1149,11 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
   getAttackMoveEffectiveness(source: Pokemon, pokemonMove: PokemonMove, ignoreAbility: boolean = false): TypeDamageMultiplier {
     const move = pokemonMove.getMove();
     const typeless = move.hasAttr(TypelessAttr);
+    // Apply abilities that change the move type
+    const typeChangeMovePowerMultiplier = new Utils.NumberHolder(1);
+    applyMoveAttrs(VariableMoveTypeAttr, source, this, move);
+    applyPreAttackAbAttrs(MoveTypeChangeAttr, source, this, move, typeChangeMovePowerMultiplier);
+
     const typeMultiplier = new Utils.NumberHolder(this.getAttackTypeEffectiveness(move.type, source));
     const cancelled = new Utils.BooleanHolder(false);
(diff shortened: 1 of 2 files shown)


Tempo-anon:
If the calculation hint gets changed shouldn't the type icon also be changed so Tackle should display as FAIRY instead of NORMAL? Just for consistency.

Also I believe this will also let the AI instantly know the player has type changing abilities when it comes to making its calculations

morriskurz:
Closing this PR as I can't work on it anymore.

PR #2899, opened 2024-07-07
[Enhancement] Decouple move accuracy and accuracy multiplier calculation from phases.ts

torranx:
## What are the changes?
- see title
- updated Wonder Skin unit tests to use the new method `calculateBattleAccuracy()`
- added unit tests involving the use of `calculateBattleAccuracy()` and `getAccuracyMultiplier()`

- cleaner phases file
- easier unit tests on move accuracy / accuracy multiplier
- more maintainable

## What did change?
- see title

### Screenshots/Videos
n/a

## How to test the changes?
play the game/run unit tests

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -10,9 +10,9 @@ import { Constructor } from "#app/utils";
 import * as Utils from "../utils";
 import { WeatherType } from "./weather";
 import { ArenaTagSide, ArenaTrapTag, WeakenMoveTypeTag } from "./arena-tag";
-import { UnswappableAbilityAbAttr, UncopiableAbilityAbAttr, UnsuppressableAbilityAbAttr, BlockRecoilDamageAttr, BlockOneHitKOAbAttr, IgnoreContactAbAttr, MaxMultiHitAbAttr, applyAbAttrs, BlockNonDirectDamageAbAttr, applyPreSwitchOutAbAttrs, PreSwitchOutAbAttr, applyPostDefendAbAttrs, PostDefendContactApplyStatusEffectAbAttr, MoveAbilityBypassAbAttr, ReverseDrainAbAttr, FieldPreventExplosiveMovesAbAttr, ForceSwitchOutImmunityAbAttr, BlockItemTheftAbAttr, applyPostAttackAbAttrs, ConfusionOnStatusEffectAbAttr, HealFromBerryUseAbAttr, IgnoreProtectOnContactAbAttr, IgnoreMoveEffectsAbAttr, applyPreDefendAbAttrs, MoveEffectChanceMultiplierAbAttr, applyPreAttackAbAttrs, MoveTypeChangeAttr, UserFieldMoveTypePowerBoostAbAttr, FieldMoveTypePowerBoostAbAttr, AllyMoveCategoryPowerBoostAbAttr, VariableMovePowerAbAttr } from "./ability";
+import { UnswappableAbilityAbAttr, UncopiableAbilityAbAttr, UnsuppressableAbilityAbAttr, BlockRecoilDamageAttr, BlockOneHitKOAbAttr, IgnoreContactAbAttr, MaxMultiHitAbAttr, applyAbAttrs, BlockNonDirectDamageAbAttr, applyPreSwitchOutAbAttrs, PreSwitchOutAbAttr, applyPostDefendAbAttrs, PostDefendContactApplyStatusEffectAbAttr, MoveAbilityBypassAbAttr, ReverseDrainAbAttr, FieldPreventExplosiveMovesAbAttr, ForceSwitchOutImmunityAbAttr, BlockItemTheftAbAttr, applyPostAttackAbAttrs, ConfusionOnStatusEffectAbAttr, HealFromBerryUseAbAttr, IgnoreProtectOnContactAbAttr, IgnoreMoveEffectsAbAttr, applyPreDefendAbAttrs, MoveEffectChanceMultiplierAbAttr, WonderSkinAbAttr, applyPreAttackAbAttrs, MoveTypeChangeAttr, UserFieldMoveTypePowerBoostAbAttr, FieldMoveTypePowerBoostAbAttr, AllyMoveCategoryPowerBoostAbAttr, VariableMovePowerAbAttr } from "./ability";
 import { allAbilities } from "./ability";
-import { PokemonHeldItemModifier, BerryModifier, PreserveBerryModifier, AttackTypeBoosterModifier, PokemonMultiHitModifier } from "../modifier/modifier";
+import { PokemonHeldItemModifier, BerryModifier, PreserveBerryModifier, PokemonMoveAccuracyBoosterModifier, AttackTypeBoosterModifier, PokemonMultiHitModifier } from "../modifier/modifier";
 import { BattlerIndex } from "../battle";
 import { Stat } from "./pokemon-stat";
 import { TerrainType } from "./terrain";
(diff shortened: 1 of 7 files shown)


src/phases.ts
```diff
@@ -3052,29 +3052,12 @@ export class MoveEffectPhase extends PokemonPhase {
       return false;
     }
 
-    const moveAccuracy = new Utils.NumberHolder(this.move.getMove().accuracy);
+    const moveAccuracy = this.move.getMove().calculateMoveAccuracy(user, target);
```
Corrade, line 3055: Could we move even more stuff from hitCheck() to calculateMoveAccuracy()?

For example, there happens to be a bug on line 3046 related to Lock-On and similar moves. I'd like to write a unit test to verify a fix. However, there's no easy way of doing this. I can't just verify that hitCheck() returns true since it could also return true by chance. I also can't use calculateMoveAcccuracy() because it doesn't involve the above line.

I'd suggest moving everything from hitCheck() into calculateMoveAccuracy() except the randomness roll, e.g. leaving hitCheck() to be something like:

```
  hitCheck(target: Pokemon): boolean {
    const moveAccuracy = this.calculateMoveAccuracy(target);
    return (moveAccuracy === -1)
      ? true
      : this.getUserPokemon().randSeedInt(100, 1) <= moveAccuracy;
  }
```

torranx: I'll check more on this when I have the time

torranx: done 😄 @Corrade

f-fsantos (changes requested):
Is this being covered by existing unit tests? Can we add more on top of these changes?

src/phases.ts
Corrade: Tyvm! I agree with your point, although there's a use-case to extract the remaining stuff as well.

To be concrete, I'm writing unit tests for OHKO moves. I want to ensure they work with Lock-On, which your recent update would've permitted. But I also want to check that they're unaffected by the accuracy and evasion battle stats. To achieve that, we'd also need to extract that calculation, if not to calculateMoveAccuracy() then maybe to a different function?

torranx: > But I also want to check that they're unaffected by the accuracy and evasion battle stats.

won't the update I recently made already cover this?

torranx:
> Is this being covered by existing unit tests? Can we add more on top of these changes?

Wonder Skin unit test is one but I could add some more

src/phases.ts
Corrade: Oh sorry true. I think the Lock-On stuff remains unextracted though, i.e. line 3046.

torranx: I don't plan to extract those code in this PR because the context of this is move/pokemon accuracy. I also don't get why you can't test Lock-On with the way it is now, if you can enlighten me, please ping me on discord.

src/test/moves/double_team.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it("increases the user's evasion by one stage.", async () => {
+    const moveToCheck = allMoves[Moves.TACKLE];
+    await game.startBattle([Species.MAGIKARP]);
+
+    const ally = game.scene.getPlayerPokemon();
+    const enemy = game.scene.getEnemyPokemon();
+    const preMoveAccMultiplier = enemy.getAccuracyMultiplier(ally, moveToCheck);
+
+    expect(ally.summonData.battleStats[BattleStat.EVA]).toBe(0);
```
Corrade, line 44: Might want to override ally and enemy abilities to be irrelevant (e.g. Ball Fetch) just in case there's something that can modify accuracy/evasion.

torranx: that's already why I'm using magikarp. but I guess additional checks wont hurt

src/test/arena/weather_fog.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    vi.spyOn(overrides, "OPP_SPECIES_OVERRIDE", "get").mockReturnValue(Species.MAGIKARP);
+    vi.spyOn(overrides, "MOVESET_OVERRIDE", "get").mockReturnValue([Moves.TACKLE]);
+    vi.spyOn(overrides, "OPP_MOVESET_OVERRIDE", "get").mockReturnValue(new Array(4).fill(Moves.SPLASH));
+  });
+
+  it("move accuracy is multiplied by 90%", async () => {
+    const moveToCheck = allMoves[Moves.TACKLE];
+    await game.startBattle([Species.MAGIKARP]);
+
+    game.doAttack(getMovePosition(game.scene, 0, Moves.TACKLE));
```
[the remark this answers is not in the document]
torranx: it will hang to `CommandPhase` until timeout if `doAttack()` is missing

src/phases.ts
```diff
@@ -3052,54 +3051,16 @@ export class MoveEffectPhase extends PokemonPhase {
       return false;
     }
```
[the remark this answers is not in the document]
torranx: Yeah I don't want to include that in this PR, feel free to open one for that.

src/data/move.ts
```diff
@@ -655,6 +655,44 @@ export default class Move implements Localizable {
...
+      return moveAccuracy.value;
+    }
+
+    const isOhko = this.hasAttr(OneHitKOAccuracyAttr);
+
+    if (!isOhko) {
+      user.scene.applyModifiers(PokemonMoveAccuracyBoosterModifier, user.isPlayer(), user, moveAccuracy);
+    }
+
+    if (user.scene.arena.weather?.weatherType === WeatherType.FOG) {
```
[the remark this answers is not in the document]
torranx: Please do so.

src/data/move.ts
```diff
@@ -655,6 +655,44 @@ export default class Move implements Localizable {
 
     return score;
   }
+
+  /**
+   * Calculates the accuracy of a move based on various conditions and attributes..
+   *
+   * @param user {@linkcode Pokemon} The Pokémon using the move.
+   * @param target {@linkcode Pokemon} The Pokémon being targeted by the move.
+   * @returns The calculated accuracy of the move.
```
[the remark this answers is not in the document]
torranx: While I agree that `-1` should be noted, I don't think this is the right place for that. `-1` is well all over the `move.ts` file, an enum would probably describe it better

Corrade (approved):
Pog

f-fsantos (approved):

src/test/arena/arena_gravity.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    expect(game.scene.arena.getTag(ArenaTagType.GRAVITY)).toBeDefined();
+
+    // Use non-OHKO move on second turn
+    await game.toNextTurn();
+    game.doAttack(getMovePosition(game.scene, 0, Moves.TACKLE));
+    await game.phaseInterceptor.to(MoveEffectPhase);
+
+    expect(moveToCheck.calculateBattleAccuracy).toHaveReturnedWith(100 * 1.67);
+  });
```
flx-sta, line 58: shouldn't the invert also be checked?
That it doesn't apply the 1.67 multiplier on non-OHKO moves?

torranx: will add OHKO move test later

src/test/arena/weather_fog.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    const moveToCheck = allMoves[Moves.TACKLE];
+
+    vi.spyOn(moveToCheck, "calculateBattleAccuracy");
+
+    await game.startBattle([Species.MAGIKARP]);
+    game.doAttack(getMovePosition(game.scene, 0, Moves.TACKLE));
+    await game.phaseInterceptor.to(MoveEffectPhase);
+
+    expect(moveToCheck.calculateBattleAccuracy).toHaveReturnedWith(100 * 0.9);
+  });
```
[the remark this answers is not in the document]
torranx: that shouldn't be part of fog test but a regular battle test imo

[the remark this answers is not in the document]
flx-sta: True.. true resolved ✅

src/test/arena/arena_gravity.test.ts
torranx: done @flx-sta

flx-sta (approved):
lgtm

@torranx FYI
Alternatively to this:
```ts
   vi.spyOn(moveToCheck, "calculateBattleAccuracy");
   //...
    expect(moveToCheck.calculateBattleAccuracy).toHaveReturnedWith(30);
```

You can also do this:
```ts
   const calculateBattleAccuracySpy = vi.spyOn(moveToCheck, "calculateBattleAccuracy");
   //...
    expect(calculateBattleAccuracySpy).toHaveReturnedWith(30);
```

I honestly didn't know the way you did it works but it makes sense. might actually prefer it as I'm not checking some odd spy var but rather the "original" one (at least by reading)

Issue #1695, opened 2024-06-01
[BUG] Fell Stinger activating after opponent kills itself


hightowernl:
**Describe the bug**

After using Fell Stinger and the opponent is able to live, then if the opponent uses a move with recoil (or possibly explosion etc) and dies, your Pokemon will get +3 attack while technically your Pokemon didn't kill the opponent.
**To Reproduce**

1. Have Fell Stinger
2. Face a Pokemon with a recoil move
3. Use Fell Stinger, the opposing Pokemon should live the attack
4. Opponent uses a move with recoil and dies
5. See error

**Expected behavior**

Your Pokemon should not get +3 attack after the opponent dies from recoil.
**Screenshots / Videos**

**Device**

PC
**Additional context**

Snailman11:
Link to Discord (Fell stinger late trigger) https://discord.com/channels/1125469663833370665/1262522451246448670

Snailman11:
https://github.com/user-attachments/assets/296b857e-7db7-4a77-91e2-78a85180fde6

PR #2909, opened 2024-07-08
[Bug] Fix Fell Stinger applying on a self-death

EmberCM:
## What are the changes?
Fixed a bug where Fell Stinger would activate if the enemy died on the same turn as using the move even if the move didn't do the final hit. Issue was found due to confusion self hit

Someone mentioned it in [Discord](https://discord.com/channels/1125469663833370665/1242236477807525940)

## What did change?
Added an entry to `pokemon.turnData.attacksReceived` when the Pokemon hits itself with confusion
Everywhere the latest attack received is used is for Feint abilities and moves, so this will fix any unintended behavior with those too

### Screenshots/Videos
Before

https://github.com/pagefaultgames/pokerogue/assets/17890775/e68d6110-6989-49f4-af1a-0c8802046d80

After

https://github.com/pagefaultgames/pokerogue/assets/17890775/fb8b23e7-b31a-42f8-b5eb-a2178dc73646

## How to test the changes?
In `overrides.ts`
```ts
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.CONFUSE_RAY];
```
In `ConfusedTag` class. This will make the enemy's confusion always hit itself, and force it to do half it's hp
```ts
  pokemon.scene.queueMessage(i18next.t("battle:battlerTagsConfusedLapse", { pokemonNameWithAffix: getPokemonNameWithAffix(pokemon) }));
  pokemon.scene.unshiftPhase(new CommonAnimPhase(pokemon.scene, pokemon.getBattlerIndex(), undefined, CommonAnim.CONFUSION));

  const damage = pokemon.stats[Stat.HP] / 2;
  pokemon.scene.queueMessage(i18next.t("battle:battlerTagsConfusedLapseHurtItself"));
  pokemon.damageAndUpdate(damage);
  const attackResult = { move: Moves.NONE, result: HitResult.OTHER as DamageResult, damage, critical: false, sourceId: pokemon.id };
  pokemon.turnData?.attacksReceived.unshift(attackResult);
  pokemon.battleData.hitCount++;
  (pokemon.scene.getCurrentPhase() as MovePhase).cancel();

  return true;
}
```

## Checklist

tests


innerthunder (changes requested):
The problem with Fell Stinger isn't just with confusion. To give another example:
1. A player Pokemon with the ability Rough Skin uses Fell Stinger on the enemy
2. The enemy attacks into the player Pokemon with a contact move, then faints from the damage caused by Rough Skin.
3. The player incorrectly gets a stat boost after the enemy faints

It's better to add a condition that checks if the damage source that caused the Pokemon to faint is indirect than to add anything to `turnData.attacksReceived`

EmberCM:
> The problem with Fell Stinger isn't just with confusion. To give another example:
> 
>     1. A player Pokemon with the ability Rough Skin uses Fell Stinger on the enemy
> 
>     2. The enemy attacks into the player Pokemon with a contact move, then faints from the damage caused by Rough Skin.
> 
>     3. The player incorrectly gets a stat boost after the enemy faints
> 
> 
> It's better to add a condition that checks if the damage source that caused the Pokemon to faint is indirect than to add anything to `turnData.attacksReceived`

Those are good points for fixing Fell Stinger specifically. However, I feel like this patch could potentially fix other issues as well when looking at self-KO due to confusion. The code changed doesn't change anything to do with Fell Stinger, just the confusion code.

innerthunder:
> Those are good points for fixing Fell Stinger specifically. However, I feel like this patch could potentially fix other issues as well when looking at self-KO due to confusion. The code changed doesn't change anything to do with Fell Stinger, just the confusion code. I can adjust the title and description of the PR accordingly if needed

I'd say you can go even broader than that. Post-victory effects from moves (e.g. Fell Stinger) and abilities (e.g. Moxie, Beast Boost) generally just check two conditions before triggering atm:

1. The user (or source of the ability) is still active on the field.
2. The last attack received by the fainting Pokemon on that turn is from the user or source.

The problem is that this doesn't account for the fainted Pokemon being KO'd by *any* source of indirect damage, not just confusion. However, this could be addressed by adding a third condition that just checks if the last source of damage on the fainted Pokemon is indirect (see `HitResult`/`DamageResult` in `phases.ts`, another field in `turnData` could track this).

EmberCM:
> I'd say you can go even broader than that. Post-victory effects from moves (e.g. Fell Stinger) and abilities (e.g. Moxie, Beast Boost) generally just check two conditions before triggering atm:
> 
>     1. The user (or source of the ability) is still active on the field.
> 
>     2. The last attack received by the fainting Pokemon on that turn is from the user or source.
> 
> 
> The problem is that this doesn't account for the fainted Pokemon being KO'd by _any_ source of indirect damage, not just confusion. However, this could be addressed by adding a third condition that just checks if the last source of damage on the fainted Pokemon is indirect (see `HitResult`/`DamageResult` in `phases.ts`, another field in `turnData` could track this).

Both my method and your method would require a refactor wherever indirect damage is done to a Pokemon. Mine changes `pokemon.turnData.attacksReceived` while yours would change the structure of `pokemon.turnData` itself. I feel like adding an attack to `attacksReceived` would give more info about the damage rather than an `indirect` flag even though most of the attack data is unused at the moment.

Snailman11:
Likely related to 
https://github.com/pagefaultgames/pokerogue/issues/1695

PR #2937, opened 2024-07-09
[Beta] Updating the manage data->unlock all option to work with all abilities, natures & forms

Opaque02:
## What are the changes?
**Up** until now, pressing the "unlock all" option from "manage data" menu does not unlock forms, abilities and natures
**Up** until now, people have demanded justice!
**Down** in the streets, beta testers want ways to have access to _all_ starter forms.
**Down** with the sadness of not having all forms, abilities and natures unlocked
**Left** to my own devices, I decided to fix this
**Right** now, I can't figure out how to save the changes (this is talked about later)
**Left** out forms can now rejoice we can use them as a starter
**Right** this wrong and merge this PR to see the changes for yourself
**B** happy that we can now play with any starter in the beta that we want
**A** new day awaits in this PR that contains absolutely NO OTHER changes whatsoever :)

People need various things to test beta related stuff - which includes a full set of unlocks. The current beta code doesn't include a fully unlocked save, which this now fixes

## What did change?
I changed the order of the "unlock all" option in the menu to be above the cancel button.
I also updated the code for unlocking pokemon - previously it was just unlocking everything in DexAttr, but now it also checks for forms, abilities and natures and unlocks all forms, abilities and natures AND passives for each pokemon

**NOTE:** at the moment, when you make the changes, they aren't immediately saved. THIS IS THE BEHAVIOUR ON THE LIVE VERSION OF BETA AT THE MOMENT TOO. This means refreshing the page will put you back at where you were before you used the unlock all option. I haven't figured out how to make it save - I tried using the same code the save and quit does, but it fails; I think since you're expected to be a wave when you use it, it tries to get session data, which may not exist, and so it crashes and fails. I haven't figured out how to get around it yet, but I've left my (non-working) code on line 68 of menu-ui-handler for it. However! In order you save, there are a few ways. The easiest is way is to force the game to save for you once you've used the unlock all option. For example, once you've used the unlock all button, you could start a new game, as starting a new game will force a save. I haven't tested it yet, but I think another option is to continue a run and go to the next wave. 

Until a way to save is figured out, this is the best I can figure out at the moment. 

### Screenshots/Videos
Below is a screenshot of the unlock all button in beta at the moment:

image

Below is an updated image (note the import session and data is only from the local dev side of things):

image

Below is a video of it in action:

https://github.com/pagefaultgames/pokerogue/assets/66582645/a637c60a-2ba7-4c64-b8bb-e3a65d3589b9

## How to test the changes?
Download the PR and use the unlock all option from the manage data menu. Whatever you do though, **DO NOT** press ↑↑↓↓←→←→BA from the main menu. Trust me on this, it won't do anything :)

## Checklist
- [x] Are the changes visual?


flx-sta:
@Opaque02 you should add an env-variable that needs to be activated to control if this feature is available (also the konami code).

Maybe something like 
`VITE_ALLOW_CHEATS=0`

Opaque02:
> @Opaque02 you should add an env-variable that needs to be activated to control if this feature is available (also the konami code).
> 
> Maybe something like `VITE_ALLOW_CHEATS=0`

@flx-sta I was trying to figure this out but didn't know what I was looking for! Do you have any suggestions on where to look to get it added? Would it be in overrides or somewhere else?

Also what Konami code? I don't know what you're talking about :-)

hayuna:
@Opaque02 You can add this variable to .env files and in the deployment process it would be processed to the game. In your code your can use `process.env.VITE_ALLOW_CHEATS` to allow cheats or not

Opaque02:
> @Opaque02 You can add this variable to .env files and in the deployment process it would be processed to the game. In your code your can use `import.meta.env.VITE_ALLOW_CHEATS` to allow cheats or not
> 
> example of use is inside src/battle-scene.ts
> 
> ```js
> export const bypassLogin = import.meta.env.VITE_BYPASS_LOGIN === "1";
> ```

@hayuna amazing, that sounds doable! I'll check it out in the next couple of days and see if I can figure it out :-). Thanks!

flx-sta:
@Opaque02 please make sure to also define that variable inside `vite.env.d.ts`.
It should also be `string` even though you are passing a number. That's a limitation we have to deal with in the code

PR #2940, opened 2024-07-09
[BUG] Fixes bug that prevents pokemon with froms from hatching as rare/epic shiny variant and preventing illegal variants from legendary gacha

sirzento:
## What are the changes?

Changed PokemonSpecies hasVariants() function to also check for pokemon with forms.
This function is used when creating an egg to check if the pokemon has variants or not.

I also added an check that prevents rare/epic variants from happening if the egg was pulled from a legendary gacha and rolled as that legendary if this pokemon has no variants.

To fix bugs. It isn't possible to pull a rare/epic miraidon (or other pokemon with forms) from eggs.
https://discord.com/channels/1125469663833370665/1259323885191692419
https://discord.com/channels/1125469663833370665/1259359879429361684

## What did change?

The `hasVariants()` function. It uses the same logic as the function that changes the variant of wild pokemon.
And a small check that sets the variant tier to common if the banner legendary from the gacha got pulled and no variants where available

## How to test the changes?

Buy a thousand miraidon eggs and see if you get an rare/epic shiny. Or just run the test script and see if the newly added test fails.

## Checklist

### DIFF
--- a/src/data/egg.ts
+++ b/src/data/egg.ts
@@ -166,12 +166,12 @@ export class Egg {
     if (eggOptions.species) {
       this._tier = this.getEggTierFromSpeciesStarterValue();
       this._hatchWaves = eggOptions.hatchWaves ?? this.getEggTierDefaultHatchWaves();
-      // If species has no variant, set variantTier to common. This needs to
-      // be done because species with no variants get filtered at rollSpecies but since the
-      // species is set the check never happens
-      if (!getPokemonSpecies(this.species).hasVariants()) {
-        this._variantTier = VariantTier.COMMON;
-      }
+    }
+    // If species has no variant, set variantTier to common. This needs to
+    // be done because species with no variants get filtered at rollSpecies but if the
+    // species is set via options or the legendary gacha pokemon gets choosen the check never happens
+    if (!getPokemonSpecies(this.species).hasVariants()) {
+      this._variantTier = VariantTier.COMMON;
     }
     // Needs this._tier so it needs to be generated afer the tier override if bought from same species
     this._eggMoveIndex = eggOptions.eggMoveIndex ?? this.rollEggMoveIndex();
@@ -337,12 +337,7 @@ export class Egg {
     } else if (this.tier === EggTier.MASTER
       && this._sourceType === EggSourceType.GACHA_LEGENDARY) {
       if (!Utils.randSeedInt(2)) {
-        const legendarySpecies = getLegendaryGachaSpeciesForTimestamp(scene, this.timestamp);
-        // Checks if the species has variants since this check gets skipped in this function
-        if (!getPokemonSpecies(legendarySpecies).hasVariants()) {
-          this._variantTier = VariantTier.COMMON;
-        }
-        return legendarySpecies;
+        return getLegendaryGachaSpeciesForTimestamp(scene, this.timestamp);
       }
     }
 
--- a/src/data/egg.ts
+++ b/src/data/egg.ts
@@ -170,7 +170,7 @@ export class Egg {
     // If species has no variant, set variantTier to common. This needs to
     // be done because species with no variants get filtered at rollSpecies but if the
     // species is set via options or the legendary gacha pokemon gets choosen the check never happens
-    if (!getPokemonSpecies(this.species).hasVariants()) {
+    if (this._species && !getPokemonSpecies(this._species).hasVariants()) {
       this._variantTier = VariantTier.COMMON;
     }
     // Needs this._tier so it needs to be generated afer the tier override if bought from same species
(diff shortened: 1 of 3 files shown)


src/test/eggs/egg.test.ts
```diff
@@ -303,4 +303,16 @@ describe("Egg Generation Tests", () => {
     expect(result1).toBe(expectedTier1);
     expect(result2).toBe(expectedTier2);
   });
+  it("should generate rare/epic shinys from pokemon with a differend form", () => {
```
brain-frog, line 306: different*

src/data/egg.ts
```diff
@@ -337,7 +337,12 @@ export class Egg {
     } else if (this.tier === EggTier.MASTER
       && this._sourceType === EggSourceType.GACHA_LEGENDARY) {
       if (!Utils.randSeedInt(2)) {
-        return getLegendaryGachaSpeciesForTimestamp(scene, this.timestamp);
+        const legendarySpecies = getLegendaryGachaSpeciesForTimestamp(scene, this.timestamp);
+        // Checks if the species has variants since this check gets skipped in this function
+        if (!getPokemonSpecies(legendarySpecies).hasVariants()) {
```
brain-frog, line 342: this same if statement is also in the constructor after rollSpecies, is that not covering this case?

sirzento: You'r right. I could just call that every time and not just if the species has ben set.

src/data/pokemon-species.ts
```diff
@@ -821,7 +821,15 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
   }
 
   hasVariants() {
-    return variantData.hasOwnProperty(this.speciesId);
+    const formIndex: number = this.formIndex;
+    let variantDataIndex: string | number = this.speciesId;
+    if (this.forms.length > 0) {
+      const formKey = this.forms[formIndex]?.formKey;
```
chaosgrimmon, line 827: Is `formIndex` used anywhere else? If not, is there a need to explicitly set it earlier instead of just grabbing `this.formIndex` here?

sirzento: No it's just there. I could change that

src/data/pokemon-species.ts
```diff
@@ -821,7 +821,14 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
   }
 
   hasVariants() {
-    return variantData.hasOwnProperty(this.speciesId);
+    let variantDataIndex: string | number = this.speciesId;
+    if (this.forms.length > 0) {
+      const formKey = this.forms[this.formIndex]?.formKey;
+      if (formKey) {
```
Tempo-anon, line 827: I don't like `if (variable)` checks. We should be explicit to make sure what we're doing since this is not clear at a glance if it's checking if formKey exists or if formKey is a nonzero number. If we're doing both I'd rather see `if (variable && variable > 0)` even if the first part would catch the second one because it's logically more consistent

src/test/eggs/egg.test.ts
```diff
@@ -303,4 +303,16 @@ describe("Egg Generation Tests", () => {
...
+    const scene = game.scene;
+    const eggs: Egg[] = [];
+
+    for (let i = 0; i < 1000; i++) {
+      eggs.push(new Egg({scene, isShiny: true, species: Species.MIRAIDON}));
+    }
+
+    const result = eggs.filter(x => x.variantTier === VariantTier.RARE || x.variantTier === VariantTier.EPIC).length;
+
+    expect(result).greaterThan(0);
```
Tempo-anon, line 316: Someone else can comment on the unit tests but I don't usually like probabilistic unit tests. Can't you spy on and mock some return variables to guarantee a rare or epic?

sirzento: Makes sense but I don't know about this because I would have to mock the return value of the function with the actuall changes. It's not the perfect way to test this but I don't know a better one right now.

src/data/pokemon-species.ts
sirzento: `formKey` is a string and can only be `undefined | "" | "something"` and that check would prevent the first two and an additional `.length` check would be a bit reduntant.

src/test/eggs/egg.test.ts
sirzento: Alright, fixed.

src/test/eggs/egg.test.ts
```diff
@@ -303,4 +303,10 @@ describe("Egg Generation Tests", () => {
     expect(result1).toBe(expectedTier1);
     expect(result2).toBe(expectedTier2);
   });
+  it("should generate an epic shiny from pokemon with a different form", () => {
```
flx-sta, line 306:
```suggestion

  it("should generate an epic shiny from pokemon with a different form", () => {
```
NIT

torranx (approved):
unit test looks good to me

PR #2941, opened 2024-07-09
[Bug] Hardcoded Pokemon should have proper names

Tempo-anon:
## What are the changes?
Opposing trainer Mega/Gmax Pokemon now display names correctly without needing to reload

Names were not accurate

## What did change?
Added `p.generateName()` to all hardcoded Pokemon with special formIndexes

### Screenshots/Videos
Before:

After:

## How to test the changes?
Override a champion or evil team boss or rival 6 fight

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -295,6 +295,11 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     return ret;
   }
 
+  /**
+   * Sets the Pokemon's name. Only called when loading a Pokemon so this function needs to be called when
+   * initializing hardcoded Pokemon or else it will not display the form index name properly.
+   * @returns n/a
+   */
   generateName(): void {
     if (!this.fusionSpecies) {
       this.name = this.species.getName(this.formIndex);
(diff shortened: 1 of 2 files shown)


src/field/pokemon.ts
```diff
@@ -295,6 +295,11 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     return ret;
   }
 
+  /**
+   * Sets the Pokemon's name. Only called when loading a Pokemon so this function needs to be called when
+   * initializing hardcoded Pokemon or else it will not display the form index name properly.
+   * @returns n/a
+   */
   generateName(): void {
```
flx-sta, line 303: Looking at the other method it should be `generateAndPopulateName()`, right?

Tempo-anon: Seems they are slightly different since there is a `generateAndPopulateMoveset()` as well as a `tryPopulateMoveset()`. 

`generateName()` right now is only used when the game data loads in the Pokemon hence why as of right now Pokemon with different forms will have the incorrect name at first but will have the correct name on reload. I'm just using the existing method to set the correct name since `generateName()` will also get called when the game is loaded.

CodeTappert:
Would this also work if the name is manually set? 
Like what i tried todo with XD001 if you remeber

src/field/pokemon.ts
flx-sta: resolved then

Tempo-anon:
> Would this also work if the name is manually set? Like what i tried todo with XD001 if you remeber

You'd have to edit the `generateName` method or game loading code so that if something already exists you take that value instead.

PR #2942, opened 2024-07-09
[Feature] add logger (for debugging)

flx-sta:
## What are the changes?

- logger stores all log messages in parallel to printing them
- logger controls which levels will be printed to the console (all will be stored always)
- add option to download pokerogue log

It might make debugging easier as it offers an easy option for ppl to download the log.
_Problem is though that often times ppls game is frozen when they would have to download it. Something to tackle later on_

## What did change?

Replaced all `console` calls with `logger`.
Printed log-level can now be controlled through `VITE_LOG_LEVEL`. A documentation of the levels can be found in the `logger.ts` file.

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/50131232/6634431a-0d1e-4c29-b0e9-d5f88c77212a

## How to test the changes?

1. Launch game
2. play a little bit around to generate some logs
3. Open Menu -> Manage Data -> Download log
4. View the log file (if you want)

## Checklist
- [x] Are the changes visual?


torranx:
i love the idea but i think the existing logs would be hard to go through. I would suggest to add a required key of `title` or `label` or similar describing what the log is (this would also allow easier finding of specific logs)

flx-sta:
> i love the idea but i think the existing logs would be hard to go through. I would suggest to add a required key of `title` or `label` or similar describing what the log is (this would also allow easier finding of specific logs)

an example of that and what those keys should contain?
I assume you mean the file it was called in?
I wanted to go step by step and get some MVP in first but sure if that's wanted we can do that. Takes a lot more effort then though :D

torranx:
> an example of that and what those keys should contain? I assume you mean the file it was called in? 

no, i mean the context of the log.
ie in [this line](https://github.com/pagefaultgames/pokerogue/pull/2942/files#diff-ae151444b91b7e5e0a6aa753024f7900f0ff0447afaa2d18ad6619d5c3e12318R52) , this could be `logger.error("Fetch user data failed", err);`

flx-sta:
> > an example of that and what those keys should contain? I assume you mean the file it was called in?
> 
> no, i mean the context of the log. ie in [this line](https://github.com/pagefaultgames/pokerogue/pull/2942/files#diff-ae151444b91b7e5e0a6aa753024f7900f0ff0447afaa2d18ad6619d5c3e12318R52) , this could be `logger.error("Fetch user data failed", err);` otherwise we would just be logging and storing things that we do not have an idea what that even is, unlike in web console where we can view the logs in real-time

Oh so basically make the `message` always a string. I mean yeah.. I see where you are coming from.. a bit of effort though but doable

torranx:
> Oh so basically make the `message` always a string. I mean yeah.. I see where you are coming from.. a bit of effort though but doable

not really. i meant to enforce it on the logs, part of the interface `LogEntry`

flx-sta:
> > Oh so basically make the `message` always a string. I mean yeah.. I see where you are coming from.. a bit of effort though but doable
> 
> not really. i meant to enforce it on the logs, part of the interface `LogEntry` (i know this is a lot of work but i think the benefit outweighs the effort 😄 , this could also be in another PR if you wanted to implement it)

So like this
```ts
  public error(message: string, ...optionalParams: any[]): void {
    this.addEntry("ERROR", message, ...optionalParams);
    if (this.logLevel <= 5) {
      console.error(message, ...optionalParams);
    }
  }
  ```
  this will make `message` always have to be a string and always have to be defined

torranx:
> ```ts
>   public error(message: string, ...optionalParams: any[]): void {
>     this.addEntry("ERROR", message, ...optionalParams);
>     if (this.logLevel <= 5) {
>       console.error(message, ...optionalParams);
>     }
>   }
> ```

i was thinking like this:

```ts
  public error(context: string, message: any, ...optionalParams: any[]): void {
    this.addEntry("ERROR", context, message, ...optionalParams);
    if (this.logLevel <= 5) {
      console.error(context, message, ...optionalParams);
    }
  }```

flx-sta:
Well that serves the same purpose with the difference that `message` then becomes a simple `optionalParameter` but being mandatory. so .. yeah

flx-sta:
@torranx it's done

torranx (approved):
lgtm

OrangeRed (approved):
lgtm :ring_buoy: :ring_buoy:

PR #2943, opened 2024-07-09
[Feature] Implements Conversion 2

schmidtc1:
## What are the changes?

Implements the move Conversion 2

To be able to use Conversion 2

## What did change?

Created a function to getTypeResistances based on an input type. Implemented logic for Conversion 2 in ResistLastMoveTypeAttr based on wiki: https://bulbapedia.bulbagarden.net/wiki/Conversion_2_(move)

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/62030095/efeb7db1-5ad7-446d-bad4-3f8cb99dc145

## How to test the changes?

Give the opponent various moves with different types. Give your pokemon Conversion 2. Test with moves and make sure your pokemon's new type is a resistance to the opponent's most recent move.

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -5592,6 +5590,52 @@ export class hitsSameTypeAttr extends VariableMoveTypeMultiplierAttr {
   }
 }
 
+/**
+ *  Attribute used for Conversion 2, to convert the user's type to a random type that resists the target's last used move.
```
Tempo-anon, line 5594: Can add all the other details in the doc like how it will fail if the user is already one of those types or how if a move has its type changed it will check with the new type

src/data/move.ts
```diff
@@ -5592,6 +5590,52 @@ export class hitsSameTypeAttr extends VariableMoveTypeMultiplierAttr {
...
+   * @param {Move} move Move being used
+   * @param {any[]} args Unused
+   * @returns {boolean} true if the function succeeds
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    if (!super.apply(user, target, move, args)) {
+      return false;
+    }
+
+    const [targetMove] = target.getLastXMoves(1); // target's most recent move
```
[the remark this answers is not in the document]
schmidtc1: It is array destructuring: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#array_destructuring. 
It takes the first element of the array and stores it in a variable itself, rather than having to call something like ```arr[0]```

src/data/type.ts
```diff
@@ -501,6 +501,55 @@ export function getTypeDamageMultiplier(attackType: integer, defType: integer):
   }
 }
 
+/**
+ * Retrieve the types resisting a given type
+ * @returns An array populated with Types, or an empty array if no resistances exist (Unknown or Stellar type)
+ */
+export function getTypeResistances(type: integer): Type[] {
```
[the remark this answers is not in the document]
schmidtc1: There probably is a solution where you can pass into ```getTypeDamageMultiplier``` the type of the opponent's move into ```attackType```, and pass in each possible defending type into ```defType```, then store any returns with multipliers <1 in an array. 
The problem with that is you would be checking 19 types every time the move is called, so it becomes a performance issue vs a memory issue. I'm not sure what's preferred.

src/data/move.ts
```diff
@@ -5592,6 +5590,52 @@ export class hitsSameTypeAttr extends VariableMoveTypeMultiplierAttr {
...
+    if (!targetMove) {
```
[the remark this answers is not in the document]
schmidtc1: Added early return for stellar and unknown types.

src/data/move.ts
schmidtc1: Added details, but just realized that move history does not use the Move class but rather the TurnMove interface. TurnMove does not store the Move used, only the ID of the move used. So type change information is not stored, meaning Hidden Power and other moves are accessed as their base types and not their type-changes.

src/data/move.ts
```diff
@@ -5592,6 +5590,62 @@ export class hitsSameTypeAttr extends VariableMoveTypeMultiplierAttr {
   }
 }
 
+/**
+ * Attribute used for Conversion 2, to convert the user's type to a random type that resists the target's last used move.
+ * Fails if both of the user's types are already resistant to the move.
```
Tempo-anon, line 5595: Is this statement true? Like if you're ground/dragon+grass (trickortreat) and the opponent uses an electric move if you conversion2 it should turn you electric since that's the only type that resists electric but isn't one of the user's current types

src/data/move.ts
```diff
@@ -5592,6 +5590,62 @@ export class hitsSameTypeAttr extends VariableMoveTypeMultiplierAttr {
...
+ * Fails if the opponent has not used a move yet
+ * Fails if the type is unknown or stellar
+ *
+ * TODO:
+ * If a move has its type changed (e.g. {@linkcode Moves.HIDDEN_POWER}), it will check the new type.
```
Tempo-anon, line 5600: Slap a `partial` on this move until this is added then? This also applies to things like Electrify/Normalize/Aerilate

src/data/move.ts
schmidtc1: You're right I phrased that incorrectly. It should say that the move fails if the user already has all types resisting the move, resolved in latest commit.

src/data/move.ts
schmidtc1: Added partial in latest commit

src/locales/it/battle.ts
```diff
@@ -69,6 +69,14 @@ export const battle: SimpleTranslationEntries = {
...
   "regainHealth": "{{pokemonName}} ha rigenerato\npunti salute!",
   "fainted": "{{pokemonNameWithAffix}} non è più in\ngrado di combattere!",
+  "statRose": "{{pokemonNameWithAffix}}'s {{stats}} è aumentato/a!",
+  "statSharplyRose": "{{pokemonNameWithAffix}}'s {{stats}} è aumentato/a molto!",
+  "statRoseDrastically": "{{pokemonNameWithAffix}}'s {{stats}} è aumentato/a drasticamente!",
+  "statWontGoAnyHigher": "{{pokemonNameWithAffix}}'s {{stats}} non può aumentare più di così!",
+  "statFell": "{{pokemonNameWithAffix}}'s {{stats}} è diminuito/a!",
+  "statHarshlyFell": "{{pokemonNameWithAffix}}'s {{stats}} è diminuito/a molto!",
+  "statSeverelyFell": "{{pokemonNameWithAffix}}'s {{stats}} è diminuito/a drasticamente!",
+  "statWontGoAnyLower": "{{pokemonNameWithAffix}}'s {{stats}} non può diminuire più di così!",
```
Tempo-anon, line 79: Should this be part of the PR?

src/data/move.ts
```diff
@@ -6139,7 +6193,8 @@ export function initMoves() {
     new AttackMove(Moves.FLAIL, Type.NORMAL, MoveCategory.PHYSICAL, -1, 100, 15, -1, 0, 2)
       .attr(LowHpPowerAttr),
     new StatusMove(Moves.CONVERSION_2, Type.NORMAL, -1, 30, -1, 0, 2)
-      .unimplemented(),
+      .attr(ResistLastMoveTypeAttr)
+      .partial(),
```
Tempo-anon, line 6197: Should also add a comment here explaining what about it is partial
`.partial(), // Checks the move's original typing and not if its type is changed through some other means`

src/locales/it/battle.ts
schmidtc1: It was a merge conflict that I asked about in the discord and was told "as long as there are no duplicates it will be fine."

src/data/move.ts
schmidtc1: Resolved.

src/locales/it/battle.ts
schmidtc1: Removed in latest commit, as other files do not have these lines.

PR #2944, opened 2024-07-09
[Bug] Fix battler tags lapsing at incorrect times

innerthunder:
## What are the changes?

This fixes a bug were some battler tags that don't use the `TURN_END` lapse type would incorrectly get removed at the end of the turn they are applied.

`Pokemon.lapseTags(BattlerTagLapseType.TURN_END)` currently checks all of the Pokemon's battler tags and removes them if their turn count is below 1. This is necessary check for the current implementation of the "flinching" mechanic, but it also creates some nasty side effects for other tag types. Many tags that use a lapse type where `turnCount` is irrelevant set their `turnCount` to 0, 1, or `null`, so when `Pokemon.lapseTags` is called at the end of a turn, it wrongfully detects that those tags have expired and removes them. Spirit Shackle and Thousand Arrows are some examples of moves with effects that incorrectly lapse out.

## What did change?

- `field/pokemon`: Removed the additional `TURN_END` condition in `Pokemon.lapseTags()`
- `data/battler-tags`:  Battler tags can now support multiple lapse types. A battler tag's `lapse` function is now called whenever `Pokemon.lapseTags` is called with a lapse type that matches one of the battler tag's lapse types. `FlinchedTag` is also rewritten to use both `PRE_MOVE` and `TURN_END` lapse types; it should be functionally identical to upstream.
- `test/abilities/parental_bond.test`: Enabled a test that failed because of this bug.
- `test/moves/astonish.test`: New test for basic flinch mechanics. The test covers the `FlinchedTag`'s behavior on the turn it is applied and on the turn after.

### Screenshots/Videos

N/A

## How to test the changes?

`npm run test astonish`
`npm run test parental_bond` (see the Anchor Shot test)

## Checklist

### DIFF
--- a/src/data/battler-tags.ts
+++ b/src/data/battler-tags.ts
@@ -154,6 +154,9 @@ export class TrappedTag extends BattlerTag {
   }
 }
 
+/**
+ * BattlerTag that represents the {@link https://bulbapedia.bulbagarden.net/wiki/Flinch Flinch} status condition
+ */
 export class FlinchedTag extends BattlerTag {
   constructor(sourceMove: Moves) {
     super(BattlerTagType.FLINCHED, [ BattlerTagLapseType.PRE_MOVE, BattlerTagLapseType.TURN_END ], 0, sourceMove);
@@ -169,6 +172,12 @@ export class FlinchedTag extends BattlerTag {
     return !pokemon.isMax();
   }
 
+  /**
+   * Cancels the Pokemon's next Move on the turn this tag is applied
+   * @param pokemon The {@linkcode Pokemon} with this tag
+   * @param lapseType The {@linkcode BattlerTagLapseType lapse type} used for this function call
+   * @returns `false` (This tag is always removed after applying its effects)
+   */
   lapse(pokemon: Pokemon, lapseType: BattlerTagLapseType): boolean {
     if (lapseType === BattlerTagLapseType.PRE_MOVE) {
       (pokemon.scene.getCurrentPhase() as MovePhase).cancel();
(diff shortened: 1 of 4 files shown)


src/data/battler-tags.ts
```diff
@@ -31,14 +31,14 @@ export enum BattlerTagLapseType {
...
+  public lapseType: BattlerTagLapseType[];
   public turnCount: integer;
   public sourceMove: Moves;
   public sourceId?: integer;
 
-  constructor(tagType: BattlerTagType, lapseType: BattlerTagLapseType, turnCount: integer, sourceMove: Moves, sourceId?: integer) {
+  constructor(tagType: BattlerTagType, lapseType: BattlerTagLapseType | BattlerTagLapseType[], turnCount: integer, sourceMove: Moves, sourceId?: integer) {
     this.tagType = tagType;
-    this.lapseType = lapseType;
+    this.lapseType = typeof lapseType === "number" ? [ lapseType ] : lapseType;
```
Tempo-anon, line 41: is there a better way of checking for type than `"number"`? Especially since there's a lot of `integer` also floating around (though I believe it should be fine since integer is just another name for number

innerthunder: `lapseType instanceof BattlerTagLapseType` doesn't work since `BattlerTagLapseType` is an enum, so I landed on this check instead. I could see if it would work the other way with `lapseType instanceof BattlerTagLapseType[]`, though, if that's preferred.

(Is there a preference between `number` and `integer` for this codebase, btw? I've seen both floating around)

Tempo-anon: `integer` was because of Sam but we should really be using `number` everywhere.

innerthunder: `lapseType instanceof BattlerTagLapseType[]` doesn't work either... As far as I know checking for `"number"` is the only way that works 😬

PR #2954, opened 2024-07-10
[Feature] Added partial implementation for Stockpile, Spit Up, Swallow

ElizaAlex:
## What are the changes?
Added partial implementation for Stockpile, and its related moves, Spit Up and Swallow. They are almost fully functional, except in rare situations where Stockpile does not succeed in raising your stats (such as because they are already at +6). In these instances,  Spit Up/Swallow will cause stat drops that would not occur on cartridge, as on cartridge the game tracks whether or not the stat boosts were successful, while the current implementation here simply checks how many Stockpile stacks you have to determine how much to lower your stats by. 

Additionally, I was unable to find the exact status text used for these moves, so that may need revision. 

I got a Shiny Cramorant and this is 3/4 of its starting moveset. It also just seemed like a doable but interesting first thing to try to contribute. 

## What did change?
I implemented a new BattlerTag:
  StockpileTag - Used to track and increment the number of Stockpile stacks, as well as apply the stat raises/drops that come along with them. 
  
and three new Move Attributes:
  VariableHealAttr - No current functionality, parallels VariablePowerAttr, Extends HealAttr
  SpitUpPowerAttr - Dictates Stockpile-dependent Spit Up power
  SwallowHealAttr - Dictates Stockpile-dependent Swallow Heal ratio, extends VariableHealAttr
  

### Screenshots/Videos

https://github.com/pagefaultgames/pokerogue/assets/132116525/5732ed69-eebe-454c-96ab-921445d09dbd

## How to test the changes?
Take any pokemon with Stockpile, Spit Up, and Swallow, and make sure that Stockpile fails after 3, that Stockpile stacks persist through reloads, that Stockpile's appropriately boosts/lowers by 1 stage per stack, upon gaining or losing stacks. Spit Up either fails or has 100/200/300 base power based on stacks, and Swallow either fails, or heals 25/50/100% based on stacks. 

## Checklist
- ✓ There is no overlap with another PR?
- ✓ The PR is self-contained and cannot be split into smaller PRs?
- ✓ Have I provided a clear explanation of the changes?
- ✓ Have I tested the changes (manually)?
    - ✓ Are all unit tests still passing? (`npm run test`)
- ✓ Are the changes visual?
  - ✓ Have I provided screenshots/videos of the changes?


src/data/move.ts
```diff
@@ -6365,12 +6411,15 @@ export function initMoves() {
...
+      .condition((user, target, move) => new StockpileTag(user.id).canAdd(target)),
     new AttackMove(Moves.SPIT_UP, Type.NORMAL, MoveCategory.SPECIAL, -1, 100, 10, -1, 0, 3)
-      .unimplemented(),
+      .attr(SpitUpPowerAttr)
+      .condition(new HasStockpileCondition()),
     new SelfStatusMove(Moves.SWALLOW, Type.NORMAL, -1, 10, -1, 0, 3)
       .triageMove()
-      .unimplemented(),
+      .attr(SwallowHealAttr)
+      .condition(new HasStockpileCondition()),
```
CodeTappert, line 6422: If you do partial (which i dont even know if we allow anymore) you need to add .partial() here

ElizaAlex:
Closing, as #2960 implements the edge-case functionality that is currently bugged in mine.

PR #2967, opened 2024-07-11
[Beta] Updating the manage data->unlock all option to work with all abilities, natures, forms and passives

Opaque02:
## What are the changes?
**Up** until now, pressing the "unlock all" option from "manage data" menu does not unlock forms, abilities and natures
**Up** until now, people have demanded justice!
**Down** in the streets, beta testers want ways to have access to _all_ starter forms.
**Down** with the sadness of not having all forms, abilities and natures unlocked
**Left** to my own devices, I decided to fix this
**Right** now, I can't figure out how to save the changes (this is talked about later)
**Left** out forms can now rejoice we can use them as a starter
**Right** this wrong and merge this PR to see the changes for yourself
**B** happy that we can now play with any starter in the beta that we want
**A** new day awaits in this PR that contains absolutely NO OTHER changes whatsoever :)

People need various things to test beta related stuff - which includes a full set of unlocks. The current beta code doesn't include a fully unlocked save, which this now fixes. Also note that this is a reopening of #2937, which I guess can be closed since it was built on the old version of the beta? There is a slight change from the previous code in that the old version (2937) would check enable passives for pokemon if they were on the starter screen, however this was wrong as babies are the ones that get passives, not evolutions, so having pikachu, chansey, jynx etc get a passive would break those on the starter screen. Now the passive code compares against the list of starter passives from the pokemon-species file instead, which only has pokemon that have passives.

## What did change?
I changed the order of the "unlock all" option in the menu to be above the cancel button.
I also updated the code for unlocking pokemon - previously it was just unlocking everything in DexAttr, but now it also checks for forms, abilities and natures and unlocks all forms, abilities and natures AND passives for each pokemon

**NOTE:** at the moment, when you make the changes, they aren't immediately saved. THIS IS THE BEHAVIOUR ON THE LIVE VERSION OF BETA AT THE MOMENT TOO. This means refreshing the page will put you back at where you were before you used the unlock all option. I haven't figured out how to make it save - I tried using the same code the save and quit does, but it fails; I think since you're expected to be a wave when you use it, it tries to get session data, which may not exist, and so it crashes and fails. I haven't figured out how to get around it yet, but I've left my (non-working) code on line 68 of menu-ui-handler for it. However! In order you save, there are a few ways. The easiest is way is to force the game to save for you once you've used the unlock all option. For example, once you've used the unlock all button, you could start a new game, as starting a new game will force a save. I haven't tested it yet, but I think another option is to continue a run and go to the next wave. 

Until a way to save is figured out, this is the best I can figure out at the moment. 

### Screenshots/Videos
Below is a screenshot of the unlock all button in beta at the moment:

image

Below is an updated image (note the import session and data is only from the local dev side of things):

image

Below is a video of it in action:

https://github.com/pagefaultgames/pokerogue/assets/66582645/a637c60a-2ba7-4c64-b8bb-e3a65d3589b9

## How to test the changes?
Download the PR and use the unlock all option from the manage data menu. Whatever you do though, **DO NOT** press ↑↑↓↓←→←→BA from the main menu. Trust me on this, it won't do anything :)

## Checklist
- [x] Are the changes visual?


CodeTappert:
So why do we need to enter the code at all? 
Why not just add that option to beta always

CodeTappert:
> So why do we need to enter the code at all? Why not just add that option to beta always

Nevermind i missunderstood it...Its another way of doing the unlock all. The option is always there xD

CodeTappert (approved):

src/ui/abstact-option-select-ui-handler.ts
```diff
@@ -42,6 +43,9 @@ export default abstract class AbstractOptionSelectUiHandler extends UiHandler {
 
   protected scrollCursor: integer = 0;
 
+  private konamiIndex: integer = 0;
+  private konamiCode: Button[] = [Button.UP, Button.UP, Button.DOWN, Button.DOWN, Button.LEFT, Button.RIGHT, Button.LEFT, Button.RIGHT, Button.CANCEL, Button.ACTION];
```
flx-sta, line 47: This should be a constant or static and read only as its value never changes and should never change. No need for an instanced attribute

src/ui/abstact-option-select-ui-handler.ts
```diff
@@ -159,6 +163,20 @@ export default abstract class AbstractOptionSelectUiHandler extends UiHandler {
 
     let playSound = true;
 
+    if (Utils.isLocal || Utils.isBeta) {
+      if (button === this.konamiCode[this.konamiIndex] && ui.getMode() === Mode.TITLE) {
```
flx-sta, line 167: NIT: you should check if in title screen first for optimized clause resolution

src/ui/menu-ui-handler.ts
```diff
@@ -30,6 +35,45 @@ const discordUrl = "https://discord.gg/uWpTfdKG49";
 const githubUrl = "https://github.com/pagefaultgames/pokerogue";
 const redditUrl = "https://www.reddit.com/r/pokerogue";
 
+export function unlockAll(scene: BattleScene) {
+  if (Utils.isLocal || Utils.isBeta) {
+    for (const species of Object.keys(Species).filter(s => !isNaN(Number(s)))) {
```
flx-sta, line 40: Mit: please move the Object.keys().filter() call into its own variable definition before the for…on

src/ui/menu-ui-handler.ts
```diff
@@ -30,6 +35,45 @@ const discordUrl = "https://discord.gg/uWpTfdKG49";
...
+      const pokemonSpecies = Number(species) > 2000 ? allSpecies.find(s => s.speciesId === Number(species)) : allSpecies[Number(species) - 1]; // thie converts the species to a pokemon from allSpecies by checking regional variants and returning the normal species index
```
flx-sta, line 41: Where is the `2000` coming from? Magic number ?

src/ui/menu-ui-handler.ts
```diff
@@ -30,6 +35,45 @@ const discordUrl = "https://discord.gg/uWpTfdKG49";
...
+        abilityAttr = AbilityAttr.ABILITY_1;
+        break;
+      case 2: // if it's one ability and the hidden ability, return ability 1 and the hidden ability
+        abilityAttr = AbilityAttr.ABILITY_1 + AbilityAttr.ABILITY_HIDDEN;
+        break;
+      case 3: // if it's 3 abilities, return all three
+        abilityAttr = AbilityAttr.ABILITY_1 + AbilityAttr.ABILITY_2 + AbilityAttr.ABILITY_HIDDEN;
+        break;
+      }
+      (scene.gameData.dexData[species] as DexEntry).seenAttr = BigInt(Math.pow(2, dexAttrLength) - 1); // we can set these values as 2^n - 1 if n is one more than the total number of total bits compared to what we need
```
flx-sta, line 59: Is there a way to avoid calling `as DexEnrtry` here ? E.g. by doing a type check instead ?

flx-sta: Same for all the following calls

src/utils.ts
```diff
@@ -292,6 +292,8 @@ export const apiUrl = localServerUrl ?? "https://api.pokerogue.net";
 // used to disable api calls when isLocal is true and a server is not found
 export let isLocalServerConnected = true;
 
+export const isBeta = import.meta.env.MODE === "development"; // this checks to see if the env mode is development. Technically this will be the same for beta AND for dev, so for now, isLocal and isBeta are technically the same thing
```
flx-sta, line 295: Use `import.meta.env.DEV` it already returns a boolean

flx-sta (changes requested):

src/ui/abstact-option-select-ui-handler.ts
Opaque02: The `ui.getMode() === Mode.TITLE` part should already check to make sure you're on the title screen - is that what you mean to check for?

flx-sta: I meant to put that check first. No need to check the other conditions if not on the title screen

src/ui/menu-ui-handler.ts
f-fsantos: When I developed the initial code for this, the IDE was just seeing this as a dictionary and providing no intellisense for it.

Opaque02:
Fixed reviewed items

flx-sta:
> Fixed reviewed items

can you please press the `resolved` button on the resolved discussions?

PR #1877, opened 2024-06-06
Heal by leftovers if hp is max_hp - 1

hayuna:
## What are the changes?
Leftovers is not healing if pokemon has almost full hp

## Why am I doing these changes?
Fix healing functionality

## How to test the changes?
Have Leftovers on a Pokemon
Have 441/442 HP, or possibly any HP with -1 of total

## example
this.hp = 441
this.getMaxHP() = 442

if result of calculation is less than 1 => heal
### Before change
Math.round((this.hp / this.getMaxHp()) * 100) / 100;
Math.round((441 / 442) * 100) / 100 => 1
### After change
Math.floor((this.hp / this.getMaxHp()) * 100) / 100;
Math.floor((441 / 442) * 100) / 100 => 0.9977

```javascript
import { describe, expect, test } from "vitest"

function getHpRatio(currentHp, maxHp) {
    const ratio = currentHp / maxHp;
    const multiplier = Math.pow(10, maxHp.toString().length)
    return Math.round(ratio * multiplier) / multiplier;
}

describe('getHpRatio', () => {
    test('should return 0 if currentHp is 0', () => {
        const ratio = getHpRatio(0, 433)
        expect(ratio).toBe(0)
    })

    test('should return 1 if currentHp is maxHp', () => {
        const ratio = getHpRatio(433, 433)
        expect(ratio).toBe(1)
    })

    test('should return number between 0 and 1', () => {
        const ratio = getHpRatio(1, 9)
        expect(ratio).toBeGreaterThan(0)
    })

    test('should return number between 0 and 1', () => {
        const ratio = getHpRatio(8, 9)
        expect(ratio).toBeGreaterThan(0)
    })
    test('should return number between 0 and 1', () => {
        const ratio = getHpRatio(1, 99)
        expect(ratio).toBeGreaterThan(0)
    })

    test('should return number between 0 and 1', () => {
        const ratio = getHpRatio(98, 99)
        expect(ratio).toBeGreaterThan(0)
    })

    test('should return number between 0 and 1', () => {
        const ratio = getHpRatio(432, 433)
        expect(ratio).toBeGreaterThan(0)
    })

    test('should return number between 0 and 1', () => {
        const ratio = getHpRatio(1, 433)
        expect(ratio).toBeGreaterThan(0)
    })


    test('should return number between 0 and 1', () => {
        const ratio = getHpRatio(1, 10000)
        expect(ratio).toBeGreaterThan(0)
    })

    test('should return number between 0 and 1', () => {
[...]

PR #2968, opened 2024-07-11
[QoL] Create leftovers test and isFullHp method for Pokemon class

### LINKED ISSUE #1626: [BUG] HP Recovery seems to be bugged at higher HP values
As seen in multiple bug reports on the discord:

https://discord.com/channels/1125469663833370665/1245880756987887696

https://discord.com/channels/1125469663833370665/1242294328613671073

https://discord.com/channels/1125469663833370665/1241732423788789833

https://discord.com/channels/1125469663833370665/1240787947855548418

From gathering data it seems to be specifically a leftovers problem, but I'm hesitant to say it's -only- leftovers, they would just be the most likely place to start.

**To Reproduce**
1: have a high HP value, I believe I've seen this happen as low as 50k
2: Get knocked down to 1 HP as usually happens in late game endless
3: Heal to almost full with a combination of enigma, sitrus berries, shell bell, and leftovers, with max healing charms.
4: should have a few hundred HP short of full health.
5: Subsequent turns with Protect will not heal those few hundred HP.

**Additional context**
This bug got introduced semi-recently, I believe in the last few weeks.

### LINKED ISSUE #1875: [BUG] Leftovers not working properly
**Describe the bug**

Leftovers doesn't work if remaining HP is -1 of total HP. My Marshadow is lvl 139, HP is 441/442 and Leftovers didn't activate after the next 3 Pokemon I beat from a Gym Leader. It was working until HP reached 441/442.
**To Reproduce**

1. Have Leftovers on a Pokemon
2. Have 441/442 HP, or possibly any HP with -1 of total
3. See error

**Expected behavior**

Leftovers should work to get that +1 HP.
**Screenshots / Videos**

**Device**

PC
**Additional context**

### LINKED ISSUE #2234: [BUG] Sturdy does not reset itself if you heal back to full hp after tanking a hit
**Describe the bug**

When a pokemon with sturdy tanks a hit, and then healed back to full HP in battle using healing moves, the next hit will still be able to OHKO them even though sturdy should normally reset itself after being healed to full hp.

**To Reproduce**

1. Use a pokemon with sturdy, in my case it was a gargancl, get hit and survive with sturdy.
2. Heal back to full HP using recover and other means (in my case the enemy was a lurantis which was slower than me and kept using Solar Blade, allowing me to protect and heal back to full)
3. Get hit with a move that would OHKO your pokemon while having healed back to full HP
4. Your pokemon still dies in one hit even though Sturdy should have let you survive.

**Expected behavior**

When healed back to full HP, sturdy should reset and allow you to tank another OHKO move.

**Screenshots / Videos**

- In this image my gargancl was knocked out despite healing back to full HP where sturdy should have protected me.
image

**Device**

PC, Firefox

**Additional context**

EmberCM:
## What are the changes?
Created a `isFullHp` method to use instead of `pokemon.getHpRatio() === 1`

The code was inconsistent in many places and caused issues such as with leftovers
Fixes #1875 and fixes #1626 and fixes #2234 

## How to test the changes?
```ts
export const STARTING_HELD_ITEMS_OVERRIDE: Array<ModifierOverride> = [{ name: "LEFTOVERS", count: 1 }];
```

## Checklist
 > Sort of? It might conflict with #1877 
- [x] Are the changes visual?

tests

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -773,6 +773,10 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     this.setNature(nature);
   }
 
+  isFullHp(): boolean {
+    return this.hp >= this.getMaxHp();
+  }
+
   getMaxHp(): integer {
     return this.getStat(Stat.HP);
   }
@@ -1981,7 +1985,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
 
         const oneHitKo = result === HitResult.ONE_HIT_KO;
         if (damage.value) {
-          if (this.getHpRatio() === 1) {
+          if (this.isFullHp()) {
             applyPreDefendAbAttrs(PreDefendFullHpEndureAbAttr, this, source, move, cancelled, damage);
           } else if (!this.isPlayer() && damage.value >= this.hp) {
             this.scene.applyModifiers(EnemyEndureChanceModifier, false, this);
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -266,7 +266,7 @@ export class PreDefendFormChangeAbAttr extends PreDefendAbAttr {
 }
 export class PreDefendFullHpEndureAbAttr extends PreDefendAbAttr {
   applyPreDefend(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: Move, cancelled: Utils.BooleanHolder, args: any[]): boolean {
-    if (pokemon.hp === pokemon.getMaxHp() &&
+    if (pokemon.isFullHp() &&
         pokemon.getMaxHp() > 1 && //Checks if pokemon has wonder_guard (which forces 1hp)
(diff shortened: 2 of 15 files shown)


src/field/pokemon.ts
```diff
@@ -773,6 +773,10 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     this.setNature(nature);
   }
 
+  isFullHp(): boolean {
+    return this.hp >= this.getMaxHp();
```
Tempo-anon, line 777: Is the `>=` check necessary or can you just use `===`?

Tempo-anon (commented):
How does this fix those bugs? It looks like you just did some renaming of variables

src/field/pokemon.ts
EmberCM: Probably isn't necessary, but I saw it used somewhere else in the code so wasn't 100% sure or not. Decided to be safe than sorry

PR #2969, opened 2024-07-11
[Bug] Grounded on Terrain fixes

### LINKED ISSUE #2926: [BUG] Terrain affecting ungrounded Pokemon
**Describe the bug**
Terrain effects should only apply to Pokemon on the ground. Currently, they apply to Pokemon that are ungrounded, whether Levitating, FLYING-type, or Magnet Rising, and to Pokemon in a semi-invulnerable turn, ie. offscreen due to Phantom Force, Dig, Dive, or Fly.

**To Reproduce**
1. Find or override a Pokemon to have a relevant Abiilty/Typing, or relevant Moves.
2. Use a Terrain.
3. Perform an action that would allow the Terrain to activate. (Priority move for Psychic, status effect for Misty, damage for Grassy, Sleep for Electric)
4. Notice that the Terrain effect applies, even though the target Pokemon is under an effect that should prevent that (or in the case of Grassy Terrain, the Pokemon heals when it shouldn't).

**Expected behavior**
image
The effect relevant for the Terrain used in step 3 should not apply.

**Screenshots / Videos**
image
(Bulbasaur, using a priority move, which Psychic Terrain blocks even though Bronzor is semi-invulnerable and thus not on the Terrain.)

**Additional context**
Psychic Terrain cancelling priority moves is in `terrain.ts`, and presumably the other effects are scattered all over the codebase, including the Same-Type Power Multiplier.
Making sure that opponents targeted did not have `SemiInvulnerableTag` (imported from `battler-tags.ts`) allowed the following, (note Shadow Sneak is the most recent move used) but obviously this is a systemic issue and should be addressed all at once.
image

chaosgrimmon:
## What are the changes?
Terrain Pulse now does not change type if its user is not grounded.
Psychic Terrain only blocks a priority move it at least one target is grounded on the Terrain.
The effects of Terrain only apply when a Pokemon is not in the semi-invulnerable turn of a two-turn move that takes the Pokemon offscreen.

These changes adjust behaviour to match the official games more closely.
image
image

Resolves #2926, as well as [this](https://discord.com/channels/1125469663833370665/1250110586050580480), [this](https://discord.com/channels/1125469663833370665/1240242817469775935), and [this](https://discord.com/channels/1125469663833370665/1125894949020381285/1259896446338076672).

## What did change?
`Pokemon.isGrounded` doesn't exist, and as the function is actually `Pokemon.isGrounded()`, Terrain Pulse was not properly checking that the user was grounded on a Terrain before taking on the Terrain's Typing. Changing that solves the issue immediately.

Psychic Terrain was not ensuring that one of the affected targets of a priority move was grounded on the Psychic Terrain before cancelling the priority move. Now, a Pokemon must be both targeted and grounded in order to be counted.

`Pokemon.isGrounded()` was returning `true` even if that Pokemon was not grounded due to being in a semi-invulnerable state. Checking for the `SemiInvulnerableTag` ensures that the Pokemon is on-screen and on the Terrain. Also, `SemiInvulnerableTag` had to be imported from `battler=tags`.

#### Before = Terrain Pulse (FAIRY-type, 100 power)
image
image
image
#### After - Terrain Pulse (NORMAL-type, 50 power)
image
image

#### Before - Psychic Terrain Cancelling
image
image
#### After - Psychic Terrain (not) Cancelling
image
image
image

#### Before - Grassy Terrain heals
image
#### After - Grassy Terrain doesn't heal
image

## How to test the changes?
I used overrides to get all the moves I wanted at the same time, as well as ensuring specific target Pokemon.

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -3617,7 +3617,7 @@ export class TerrainPulseTypeAttr extends VariableMoveTypeAttr {
    * @returns true if the function succeeds
    */
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    if (!user.isGrounded) {
+    if (!user.isGrounded()) {
```
Tempo-anon, line 3620: This was not even calling the function lol

Tempo-anon (approved):

PR #2971, opened 2024-07-11
[Test] Update tests to enable no-crits override

DayKev:
## What are the changes?
Two tests are updated to enable the override that prevents crits.

The tests used some unnecessary RNG workarounds when the problem was crits all along.

## What did change?
`NEVER_CRIT_OVERRIDE` is set to true for the Dry Skin and Parental Bond tests, and the hacky damage roll workaround is removed since it's not necessary.

## How to test the changes?
`npm run test dry_skin` and `npm run test parental_bond`

## Checklist

### DIFF
--- a/src/test/abilities/dry_skin.test.ts
+++ b/src/test/abilities/dry_skin.test.ts
@@ -25,6 +25,7 @@ describe("Abilities - Dry Skin", () => {
   beforeEach(() => {
     game = new GameManager(phaserGame);
     vi.spyOn(overrides, "SINGLE_BATTLE_OVERRIDE", "get").mockReturnValue(true);
+    vi.spyOn(overrides, "NEVER_CRIT_OVERRIDE", "get").mockReturnValue(true);
     vi.spyOn(overrides, "OPP_ABILITY_OVERRIDE", "get").mockReturnValue(Abilities.DRY_SKIN);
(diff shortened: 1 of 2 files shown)


src/test/abilities/dry_skin.test.ts
```diff
@@ -77,30 +78,28 @@ describe("Abilities - Dry Skin", () => {
...
-    // ensure the enemy doesn't die to this
-    vi.spyOn(overrides, "OPP_LEVEL_OVERRIDE", "get").mockReturnValue(30);
+    vi.spyOn(overrides, "MOVESET_OVERRIDE", "get").mockReturnValue([Moves.FLAMETHROWER]);
 
     await game.startBattle();
 
     const enemy = game.scene.getEnemyPokemon();
-    expect(enemy).not.toBe(undefined);
+    expect(enemy).toBeDefined();
+    const maxHP = 1000;
```
Tempo-anon, line 87: nit: `maxHP` name is a bit misleading. Wouldn't this be better as `initialHP` or something?

DayKev: Hm, true; I named it that because it's replacing the `enemy.getMaxHp()` calls, but changing it seems fine.

PR #2979, opened 2024-07-11
[BUG] Fixing move Chilly Reception

DustinLin:
## What are the changes?

Fixing the currently implemented move Chilly Reception. Prior to these fixes the move basically didn't work at all.

Noticed from the bug-reports channel on discord

## What did change?

Given Chilly Reception's unique make up of being a STATUS move, a SWITCHOUT move, and a WEATHER effecting move, many special circumstances had to made in the code for it to work.

For example, when checking the ForceSwitchOutAttr, the logic specifically checks that the move isn't a STATUS move. I'm guessing this is so that attacking moves such as U-turn still go through despite not having a switchout target, but this breaks Chilly Reception since it is a STATUS move with no attack.

According the [bulba](https://bulbapedia.bulbagarden.net/wiki/Chilly_Reception_(move)), the move still allows a user to switch out even if the weather changing portion of the move fails. This again requires a special circumstance, as normally weather changing moves auto-fail in this circumstance

### Screenshots/Videos

On `beta`, basically nothing works (game freezes on switch out, move fails when used solo)

https://github.com/user-attachments/assets/3baff715-b9f6-4582-acfd-425dcac5c124

https://github.com/user-attachments/assets/d1f5e6ec-e67d-4ba5-9962-f01a6eea208f

On the PR these issues are fixed: when used in doubles, it still allows switching even if the weather doesn't change. In single battles, it just changes the weather (and can fail if both the weather and switch out fails). Also added dialogue as per the move description

https://github.com/user-attachments/assets/425e5dcd-db9b-4fe4-bd94-57856ba6974d

https://github.com/user-attachments/assets/022fc0ac-7d4e-46cd-86fa-924648e684a4

## How to test the changes?

Try for single and double battles via overrides with Chilly Reception

## Checklist
- [x] Are the changes visual?


flx-sta (approved):
Looks good to me
I've tried multiple double battle scenarios and single battle scenarios.

One question: Do we really want to have hard-coded `CHILLY_RECEPTION` code? I understand that this ability is quite unique and thus it's acceptable. I can't think of any better solution EXCEPT allowing for parameter inside those Attribute that allow for custom `getDefinition` overrides. But I'm scared that ppl might abuse it for wrong stuff in the future.

KInd of like this
```ts
export class WeatherChangeAttr extends MoveEffectAttr {
  //...
  constructor(weatherType: WeatherType, getConditionOverride) {
    //..
  }

new SelfStatusMove(Moves.CHILLY_RECEPTION, Type.ICE, -1, 10, -1, 0, 9)
      .attr(WeatherChangeAttr, WeatherType.SNOW, (user, target, move) => {
         if ( move.getAttrs(ForceSwitchOutAttr)[0].getSwitchOutCondition()(user, target, move) === false &&  user.scene.arena.weather?.weatherType === WeatherType.SNOW ) {
          return false;
        }

        return user.scene.arena.trySetWeather(this.weatherType, true) || move.getAttrs(ForceSwitchOutAttr)[0].getSwitchOutCondition()(user, target, move);
      })

**Comment by DustinLin** [comment]
yeah the solution i have is awkward, as the existing "move infrastructure" doesn't easily support a move like this. I can already think of another move that will probably need to be solved in a similar way ([parting shot](https://bulbapedia.bulbagarden.net/wiki/Parting_Shot_(move)).

I like your proposal, as it moves all the special logic outside of the attribute code and groups them with the definition of the move but yes I can see it getting abused (as you are essentially ignoring the pre-exisiting implementation). But then again that's what my solution is doing anyways...

File: src/data/move.ts [hunk 1]
```diff
@@ -2097,7 +2097,27 @@ export class WeatherChangeAttr extends MoveEffectAttr {
   }
 
   getCondition(): MoveConditionFunc {
-    return (user, target, move) => !user.scene.arena.weather || (user.scene.arena.weather.weatherType !== this.weatherType && !user.scene.arena.weather.isImmutable());
+    // adding clause CHILLY_RECEPTION, so move doesn't fail when it's already snowing: since the move still allows a switch out
+    return (user, target, move) => {
+
+      if (move.id === Moves.CHILLY_RECEPTION) {
```
mcmontag, line 2103: Chilly Reception should just have its own subclasses of these two attributes, instead of hardcoding its logic (including weather type!) in the general-purpose versions.  Having this many comments about one move in an otherwise move-agnostic attribute is a big red flag.

DustinLin:
@flx-sta check out the ALT PR i linked, I think it helps consolidate the changes, and creates something that can be better used in the future. Think it would also address some of mcmontag's concerns

DustinLin:
`beta` got messed up so reopening #3198

PR #2983, opened 2024-07-11
[QoL] Remove unneeded PP Up and PP Max

EmberCM:
## What are the changes?
Removes PP Up and PP Max from the shop if all Party Pokemon have maxed PP Up

Someone asked in [Discord](https://discord.com/channels/1125469663833370665/1239793884553547838)

## What did change?
PP Up and PP Max no longer appear in shop if all Party Pokemon have 3/3 PP Up on all their moves

## How to test the changes?
Change the `WeightedModifierType` for `PP_UP` and `PP_MAX` to `1000`

## Checklist
- [x] Are the changes visual?

tests

### DIFF
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -1399,7 +1399,11 @@ const modifierPool: ModifierPool = {
   }),
   [ModifierTier.GREAT]: [
     new WeightedModifierType(modifierTypes.GREAT_BALL, (party: Pokemon[]) => (hasMaximumBalls(party, PokeballType.GREAT_BALL)) ? 0 : 6, 6),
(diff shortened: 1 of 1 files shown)


Tempo-anon:
This seems like it'll only affect current Endless and will increase vouchers. Sounds like a balance decision

damocleas:
Inclined to agree with Temp's comment, I personally will not be approving this and do not wish for it to be added off of the fact it could be used to exploit to get more vouchers or such. Even added as a classic only change it would be easily exploitable by running 1-2 hard carries and sweeping through with those for more.

f-fsantos:
Rejected per balance team stance.

PR #2847, opened 2024-07-06
[Enhancement] Decouple move power calculation from Pokemon.apply(), Fixes Power Spot & Battery not boosting ally's move

torranx:
## What are the changes?
- **included fix for Power Spot and Battery not boosting the ally's move - this is discovered when replacing the `getMockedMovePower()`, hence more reason this needs to be added 😄**
- moved move power calculation to its own method
- updated tests for power value/multipliers to use the new method instead of mimicking its implementation
- **kept the original order of power calculation**

## Why am I doing these changes?
- avoid convuloted code
- easier and cleaner unit tests for move power

## What did change?
- refactor
- fix Power Spot and Battery

### Screenshots/Videos
n/a

## How to test the changes?
run unit tests or play the game

## Checklist
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [ ] Are the changes visual?
  - [ ] Have I provided screenshots/videos of the changes?

PR #2984, opened 2024-07-11
[Enhancement] Decouple move power calculation from Pokemon.apply(), Fixes Power Spot & Battery not boosting ally's move

torranx:
merged my branch from #2847 to here with beta as base branch

## What are the changes?
- **included fix for Power Spot and Battery not boosting the ally's move - this was discovered when replacing the `getMockedMovePower()`, hence more reason this needs to be added 😄**
- moved move power calculation to its own method
- updated tests for power value/multipliers to use the new method instead of mimicking its implementation
- **kept the original order of power calculation**

- avoid convuloted code
- easier and cleaner unit tests for move power

## What did change?
- refactor
- fix Power Spot and Battery

### Screenshots/Videos
n/a

## How to test the changes?
run unit tests or play the game

## Checklist
- [ ] Are the changes visual?


src/data/move.ts
```diff
@@ -655,6 +655,73 @@ export default class Move implements Localizable {
...
+
+    const sourceTeraType = source.getTeraType();
+    if (sourceTeraType !== Type.UNKNOWN && sourceTeraType === this.type && power.value < 60 && this.priority <= 0 && !this.hasAttr(MultiHitAttr) && !source.scene.findModifier(m => m instanceof PokemonMultiHitModifier && m.pokemonId === source.id)) {
+      power.value = 60;
+    }
+
+    applyPreAttackAbAttrs(VariableMovePowerAbAttr, source, target, this, power);
+
+    if (source.getAlly()?.hasAbilityWithAttr(AllyMoveCategoryPowerBoostAbAttr)) {
+      applyPreAttackAbAttrs(AllyMoveCategoryPowerBoostAbAttr, source.getAlly(), target, this, power);
```
torranx, line 680: `source` was the previous argument here, which broke the ability attribute - should be `source.getAlly()`

src/test/abilities/battery.test.ts
```diff
@@ -41,11 +42,9 @@ describe("Abilities - Battery", () => {
     game.doAttack(getMovePosition(game.scene, 1, Moves.SPLASH));
 
     const multiplier = getAttrPowerMultiplier(game.scene.getPlayerField()[1]);
-    const mockedPower = getMockedMovePower(game.scene.getEnemyField()[0], game.scene.getPlayerField()[0], allMoves[moveToBeUsed]);
+    const movePower = allMoves[moveToBeUsed].calculatePower(game.scene.getPlayerField()[0], game.scene.getEnemyField()[0]);
```
OrangeRed, line 45: `scene.getPlayerPokemon()` returns the 0th index of the player field and the same goies for `scene.getEnemyPokemon()`

src/data/move.ts
```diff
@@ -655,6 +655,73 @@ export default class Move implements Localizable {
...
+
+  /**
+   * Calculates the power of a move based on various conditions and attributes.
+   *
+   * @param source {@linkcode Pokemon} The Pokémon using the move.
+   * @param target {@linkcode Pokemon} The Pokémon being targeted by the move.
+   * @returns The calculated power of the move.
+   */
+  calculatePower(source: Pokemon, target: Pokemon): number {
+    const power = new Utils.NumberHolder(this.power);
```
OrangeRed, line 667: I don't like how `calculatePower()` and `this.power` have somewhat colliding names. Can you change `this.power` to be something like `this.basePower`

OrangeRed (changes requested):

src/test/abilities/battery.test.ts
torranx: yes I know that, I've used this instead to be more explicit and easier to read because `getPlayerPokemon` is not apparent to return the index 0

src/data/move.ts
torranx: I would be changing a lot of things if we decide to change the property name, and potentially breaking some things in the process if ever I miss something.

src/test/abilities/battery.test.ts
OrangeRed: I am of the contrary opinion in that `scene.getPlayerField[0]` / `scene.getPlayerParty[0]` is harder to parse

src/data/move.ts
OrangeRed: It's not used in that many places what could go wrong with a simple F10 change all symbols :grimacing: 

image

torranx: Problem is we can's just find and replace `move.power`, there are other usages like move being stored in a variable or something like `move.getMove().power`. If you can suggest a better name for the method instead, that'll be safer 👌

src/test/abilities/battery.test.ts
torranx: I guess we agree to disagree 😆 , if other people too will tell me otherwise, I'd be happy to update this

mcmontag:
Getting the move's max base power from its attribute using a helper method is totally unnecessary.  These are integration tests, and the whole point of factoring out power calculation is to make it so we can test a move's behavior without needing to know about its attributes.

We know it's supposed to have a max power of 100, simply because the move is Hard Press.

src/test/moves/hard_press.test.ts
```diff
@@ -50,98 +31,71 @@ describe("Moves - Hard Press", () => {
...
+    vi.spyOn(enemy, "getHpRatio").mockReturnValue(randInt(99, 1));
 
-    game.doAttack(getMovePosition(game.scene, 0, moveToUse));
-    await game.phaseInterceptor.to(MoveEffectPhase, false);
+    game.doAttack(getMovePosition(game.scene, 0, Moves.HARD_PRESS));
+    await game.phaseInterceptor.to(MoveEffectPhase);
 
-    checkBasePowerChanges(game.scene.getCurrentPhase() as MoveEffectPhase, moveToUse, -1, 100);
+    expect(moveToCheck.calculateBattlePower).toHaveReturnedWith(Math.max(Math.floor(100 * enemy.getHpRatio()), 1));
   });
```
mcmontag, line 52: nitpick - I don't think this test is necessary.  If someone somehow managed to change the attributed used by this move to work for every enemy HP ratio except for 2, there's a very slim chance that this test would catch it.  On the other end, testing all 100 possible HP ratios is mega overkill.

If it were me, I'd test HP ratios of -1, 0, 1, 50, 100, and 101.  Key inputs within the bounds, and one input out of the expected bounds on each end.  1, 50, and 100 is probably fine, though.

torranx:
> Getting the move's max base power from its attribute using a helper method is totally unnecessary. These are integration tests, and the whole point of factoring out power calculation is to make it so we can test a move's behavior without needing to know about its attributes.
> 
> We know it's supposed to have a max power of 100, simply because the move is Hard Press.

Updated the code, I'm ~not~ ready for you to tear it apart 😆

src/test/abilities/aura_break.test.ts
```diff
@@ -7,15 +7,18 @@ import { MoveEffectPhase } from "#app/phases";
...
   let phaserGame: Phaser.Game;
   let game: GameManager;
-  const multiplier = 9 / 16;
+  const auraBreakMultiplier = 9 / 16;
+  const auraMultiplier = 4 / 3;
+  /**
+   * Apparently, the auraMultiplier is being multiplied first to the move's power then multiplied again to
+   * the auraBreakMultiplier. This means we can't net the multiplier like so:
+   * power * (auraMultiplier * auraBreakMultiplier). Doing so will make the result off by a decimal value.
+   */
```
mcmontag, line 21: Not sure I understand this comment.  Does using `.toBeCloseTo` instead of `.toBe` solve `This means we can't net the multiplier`?

mcmontag (approved):
Test changes LGTM - deferring to other reviewer(s) for power calculation refactor.

src/test/abilities/aura_break.test.ts
mcmontag: (noted in discord that `.toHaveReturnedWith(expect.closeTo(...))` is a more applicable suggestion; resolving as this is a nitpick anyway)

src/data/move.ts
```diff
@@ -655,6 +655,73 @@ export default class Move implements Localizable {
...
+    const sourceTeraType = source.getTeraType();
+    if (sourceTeraType !== Type.UNKNOWN && sourceTeraType === this.type && power.value < 60 && this.priority <= 0 && !this.hasAttr(MultiHitAttr) && !source.scene.findModifier(m => m instanceof PokemonMultiHitModifier && m.pokemonId === source.id)) {
+      power.value = 60;
+    }
+
+    applyPreAttackAbAttrs(VariableMovePowerAbAttr, source, target, this, power);
+
+    if (source.getAlly()?.hasAbilityWithAttr(AllyMoveCategoryPowerBoostAbAttr)) {
+      applyPreAttackAbAttrs(AllyMoveCategoryPowerBoostAbAttr, source.getAlly(), target, this, power);
+    }
```
OrangeRed, line 681: Checking `Pokemon.hasAbilityWithAttr()` here isn't necessary since that gets invoked inside the `applyAbilityInternals` function.

src/data/move.ts
```diff
@@ -655,6 +655,73 @@ export default class Move implements Localizable {
...
+      }
+    }
+
+    if (source.scene.arena.getTerrainType() === TerrainType.GRASSY && target.isGrounded() && this.type === Type.GROUND && this.moveTarget === MoveTarget.ALL_NEAR_OTHERS) {
+      power.value /= 2;
+    }
+
+    applyMoveAttrs(VariablePowerAttr, source, target, this, power);
+
+    source.scene.applyModifiers(PokemonMultiHitModifier, source.isPlayer(), source, new Utils.IntegerHolder(0), power);
```
OrangeRed, line 712: Can you check whether its safe to pass in null instead of a `new Utils.IntegerHolder`

src/test/abilities/aura_break.test.ts
```diff
@@ -33,63 +36,34 @@ describe("Abilities - Aura Break", () => {
...
-
-    const appliedPower = getMockedMovePower(game.scene.getEnemyField()[0], game.scene.getPlayerField()[0], allMoves[Moves.MOONBLAST]);
-
     await game.phaseInterceptor.to(MoveEffectPhase);
 
-    expect(appliedPower).not.toBe(undefined);
-    expect(appliedPower).not.toBe(basePower);
-    expect(appliedPower).toBe(basePower * multiplier);
-
+    expect(moveToCheck.calculateBattlePower).toHaveReturnedWith(basePower * auraMultiplier * auraBreakMultiplier);
```
OrangeRed, line 53: since you're using fractions here can you use `expect.closeTo()` to prevent any floating point errors

src/field/pokemon.ts
```diff
@@ -1817,29 +1792,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
         source.stopMultiHit(this);
         result = HitResult.NO_EFFECT;
       } else {
-        const typeBoost = source.findTag(t => t instanceof TypeBoostTag && t.boostedType === move.type) as TypeBoostTag;
-        if (typeBoost) {
-          power.value *= typeBoost.boostValue;
-          if (typeBoost.oneUse) {
-            source.removeTag(typeBoost.tagType);
-          }
-        }
```
OrangeRed, line 1826: Are you handling this change correctly? By moving this inside of the calculate function `typeBoost.oneUse` tags will be consumed even if the move fails / is cancelled which is not what happens currently

OrangeRed (changes requested):

src/field/pokemon.ts
torranx: All power calculation inside that `else` statement can be moved out safely because it's the end value that is only being consumed (part of damage calculation).

src/data/move.ts
torranx: Tried testing this but the game breaks when setting overrides for `ITEM_REWARD_OVERRIDE` or `STARTING_MODIFIER_OVERRIDE`. We can probably move this into another PR.

src/field/pokemon.ts
OrangeRed: Doesn't this particular block only trigger if the move has not been cancelled?
By moving this inside of calculatePower which happens before the is cancelled check you will sometimes run this particular block when it otherwise wouldn't be/

torranx: It wouldn't really matter because again it's only being consumed [here](https://github.com/pagefaultgames/pokerogue/blob/beta/src/field/pokemon.ts#L1913). Please correct me if I'm missing something.

PR #2985, opened 2024-07-11
[Feature] Fully implement Octolock

mcmontag:
## What are the changes?
Octolock is only partially implemented in its current state (it simply traps the target).

This PR finishes its implementation (both traps the target, and reduces its DEF and SPDEF by 1 stage at the end of each turn).

To finish Octolock's implementation.

## What did change?
Created a new battler tag, `OctolockTag`, which extends `TrappedTag` and reduces the affected pokemon's DEF and SPDEF by 1 stage each at the end of each turn.

In addition to testing the move's behavior via integration tests (which seems to be the status quo throughout the codebase), I've also added the `/test/battlerTags` directory, with _unit_ tests for the new tag (`/test/battlerTags/octolock.test.ts`).

Unit testing tags is made slightly more complicated by the fact that `Pokemon` is an abstract class, so I'm just manually creating bare-bones mocks to suit each test's need (as objects, casting them to `Pokemon`).

### Screenshots/Videos

https://github.com/user-attachments/assets/971c90b5-926f-4423-b41b-7a9dbfa6736e

## How to test the changes?
Use overrides to give yourself and/or your opponent Octolock. Verify that it works as expected.

## Checklist
- [x] Are the changes visual?


src/data/battler-tags.ts
```diff
@@ -626,6 +626,32 @@ export class IngrainTag extends TrappedTag {
...
+ */
+export class OctolockTag extends TrappedTag {
+  constructor(sourceId: number) {
+    super(BattlerTagType.OCTOLOCK, BattlerTagLapseType.TURN_END, 1, Moves.OCTOLOCK, sourceId);
+  }
+
+  canAdd(pokemon: Pokemon): boolean {
+    const isOctolocked = pokemon.getTag(BattlerTagType.OCTOLOCK);
+    return !isOctolocked;
+  }
```
mcmontag, line 641: Bulbapedia/Serebii/Smogon/etc don't mention anything about failing if the target is already trapped.  I'm only noting this because doing it that way (as I have) differs from the way similar moves are currently implemented (see: Ingrain's `canAdd` method).

Tempo-anon: You can test by making two accounts on Pokemon Showdown and challenging yourself to see what happens, that or testing on a cartridge.

flx-sta:
Pull the latest beta branch to fix the failing tests

src/data/battler-tags.ts
mcmontag: Tested and confirmed - Octolock applies successfully on a target that is already trapped.

PR #2990, opened 2024-07-11
[Localization][QoL] i18 changes to split Spanish into es-ES and es-MX like the original games are doing

AsdarDevelops:
## What are the changes? Why am I doing these changes?

I created a copy of "es" locale folder and named it "es_Mx". Old "es" is now "es_Es"
This has been discussed among translators for a while, but today we have some official LATAM translator, so it seems like a good idea to start splitting off Spanish from Spain and Mexico so that new move names and such can be updated.

## What did change?
The changes are only in the i18 file, and in the config.ts of "es" (now "es_Es") and the newly created directory "es_Mx". The rest of the changes are simply from the locales directory being renamed and a copy of it being added.

### Screenshots/Videos
No changes were made beyond the organizational stuff.

## Checklist
- [X] Are the changes visual?


flx-sta:
We should consider setting up a fallback from es-max to es-ES

```ts
// fallback depending on user language
i18next.init({
    fallbackLng: { 
        'de-CH': ['fr', 'it'], //French and Italian are also spoken in Switzerland
        'zh-Hant': ['zh-Hans', 'en'],
        'es': ['fr'],
        'default': ['en']
    }
});
```

AsdarDevelops:
> Note for later: We should consider setting up a fallback from es-mx to es-ES
> 
> ```ts
> // fallback depending on user language
> i18next.init({
>     fallbackLng: { 
>         'de-CH': ['fr', 'it'], //French and Italian are also spoken in Switzerland
>         'zh-Hant': ['zh-Hans', 'en'],
>         'es': ['fr'],
>         'default': ['en']
>     }
> });
> ```

Are you assuming we Spanish understand French better than English? We sure as hell don't 🤣 

This seems like a good idea to make "es-mx" default to "es-es" before going into "en"

However, I don't think it's a good idea to have this be merged to this PR, self-containment and all of that.

PR #2992, opened 2024-07-12
[Test] Fix/Extend Unit Tests for Hard Press

xsn34kzx:
## What are the changes?

Fixed `Hard Press`'s unit test such that it no longer succeeds or fails non-deterministically.

`Hard Press`'s unit test would cause PRs to "fail" during the `test` check. 

## What did change?

`hard_press.test.ts` was re-written such that it no longer is indeterminate or errors due to `undefined` enemy `Pokemon`. Additionally, the unit test was separated into several tests within the file to have both set scenarios and a wildcard scenario to effectively test the intended behavior of the move.

### Screenshots/Videos

N/A

## How to test the changes?

Enter `npm run test hard_press.test.ts` into a terminal sourced at the `pokerogue` folder to run the test file.

## Checklist
  - https://github.com/pagefaultgames/pokerogue/pull/2984
- [ ] Are the changes visual?

### DIFF
--- a/src/test/moves/hard_press.test.ts
+++ b/src/test/moves/hard_press.test.ts
@@ -9,10 +9,29 @@ import {
 import { Moves } from "#enums/moves";
 import { getMovePosition } from "#app/test/utils/gameManagerUtils";
 import { Abilities } from "#enums/abilities";
-import { NumberHolder } from "#app/utils.js";
-import Move from "#app/data/move.js";
-import Pokemon from "#app/field/pokemon.js";
-import { allMoves, OpponentHighHpPowerAttr } from "#app/data/move.js";
+import { applyMoveAttrs, VariablePowerAttr } from "#app/data/move";
+import * as Utils from "#app/utils";
+import { Stat } from "#enums/stat";
+
+/**
+   * Checks the base power of the {@linkcode intendedMove} before and after any
+   * {@linkcode VariablePowerAttr}s have been applied.
+   * @param phase current {@linkcode MoveEffectPhase}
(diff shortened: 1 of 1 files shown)


src/test/moves/hard_press.test.ts
```diff
@@ -9,10 +9,29 @@ import {
...
+   * @param before Expected base power before any base power changes
+   * @param after Expected base power after any base power changes
+   */
+const checkBasePowerChanges = (phase: MoveEffectPhase, intendedMove: Moves, before: number, after: number) => {
+  // Double check if the intended move was used and verify its initial base power
+  const move = phase.move.getMove();
+  expect(move.id).toBe(intendedMove);
+  expect(move.power).toBe(before);
+
+  /** Mocking application of {@linkcode VariablePowerAttr} */
```
mcmontag, line 30: Small nitpick - this isn't mocking the application of VariablePowerAttr, it's just straight up applying it.

IMO it may be worth running the MoveEffectPhase right before this, and expecting `applyMoveAttrs` to have been called with the same args exactly once.  That'll ensure that these tests break if `Pokemon.apply` stops doing what this method assumes it does.

mcmontag: nvm, it sounds like power calculation is going to be factored out of `Pokemon.apply` soon(?), so this'll end up being obsolete.

mcmontag (approved):
LGTM, noting that steps are being taken in another PR to clean up power calculation tests.

Tempo-anon (approved):
lol I didn't write this test

yellow caught emoji

PR #2995, opened 2024-07-12
[Bug] Evolution ability fix

DayKev:
## What are the changes?
If a Pokémon with their second ability evolves, it will swap to its first ability if the evolution has no second ability, instead of improperly "upgrading" to its hidden ability.

I noticed a bug report about the issue
https://discord.com/channels/1125469663833370665/1254274565627187211

cf [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Ability#Mechanics)
> In technical terms, a species' Abilities may be thought of as having separate slots, with an individual Pokémon's non-Hidden slot determined by its [personality value](https://bulbapedia.bulbagarden.net/wiki/Personality_value). For example, an [Eevee](https://bulbapedia.bulbagarden.net/wiki/Eevee_(Pok%C3%A9mon)) — with two non-Hidden Abilities — has [Run Away](https://bulbapedia.bulbagarden.net/wiki/Run_Away_(Ability)) for its first non-Hidden slot, [Adaptability](https://bulbapedia.bulbagarden.net/wiki/Adaptability_(Ability)) for its second, and [Anticipation](https://bulbapedia.bulbagarden.net/wiki/Anticipation_(Ability)) for its Hidden Ability slot. [Vaporeon](https://bulbapedia.bulbagarden.net/wiki/Vaporeon_(Pok%C3%A9mon)) — with only one non-Hidden Ability — can be considered to have [Water Absorb](https://bulbapedia.bulbagarden.net/wiki/Water_Absorb_(Ability)) for both non-Hidden slots. When a Pokémon evolves, its Ability slot remains the same.

## What did change?
A check is added in the `evolve()` function in `src/field/pokemon.ts` to ensure a Pokémon whose second ability is "NONE" (aka it doesn't have a second ability) is swapped to its first ability.

### Screenshots/Videos
Incorrect behavior (current on beta):

https://github.com/user-attachments/assets/3f7d2a15-92b8-4098-b06a-663794accbfe

Demonstration of correct behavior after the fix is applied:

https://github.com/user-attachments/assets/1a81b59c-33a8-473d-ae35-d79d2917141c

https://github.com/user-attachments/assets/a79d8c54-e137-4fe5-b494-03e474c94e10

## How to test the changes?
Start a run with an Eevee with Adaptability (its second ability) and evolve it with a Fire/Water/Thunder stone. It should now have its first ability (Flash Fire/Water Absorb/Volt Absorb) rather than its hidden ability (Guts/Hydration/Quick Feet).

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -3211,13 +3211,22 @@ export class PlayerPokemon extends Pokemon {
       this.generateName();
       if (!isFusion) {
         const abilityCount = this.getSpeciesForm().getAbilityCount();
-        if (this.abilityIndex >= abilityCount) { // Shouldn't happen
-          this.abilityIndex = abilityCount - 1;
+        // If a pokemon has its second ability and it evolves into a pokemon that doesn't have a second ability, switch to its first ability instead of its hidden ability
+        if (this.getSpeciesForm().ability2 === Abilities.NONE && this.abilityIndex === 1) {
+          this.abilityIndex = 0;
         }
-      } else {
+        // Prevent pokemon with an illegal ability value from breaking things too badly
+        if (this.abilityIndex >= abilityCount) {
+          console.warn("this.abilityIndex is somehow an illegal value, please report this");
+          console.warn(this.abilityIndex);
+          this.abilityIndex = 0;
+        }
+      } else { // Do the same as above, but for fusions
         const abilityCount = this.getFusionSpeciesForm().getAbilityCount();
-        if (this.fusionAbilityIndex >= abilityCount) {// Shouldn't happen
-          this.fusionAbilityIndex = abilityCount - 1;
+        if (this.fusionAbilityIndex >= abilityCount) {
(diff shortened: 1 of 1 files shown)


CodeTappert (approved):

Tempo-anon:
What if we just made it so instead of a Pokemon having Ability 1, NONE, Hidden Ability we just replaced all the NONEs with Ability 1 so that the ability index would stay the same?

DayKev:
I did that at first, but that requires changing the logic in the starter select screen so that you don't have to switch past a duplicate ability to reach a Pokémon's hidden ability (and also changing UI code is scary). I was worried there would be other issues due to similar code elsewhere as well, but if you're willing to accept that then I'm fine with implementing it that way instead.

DayKev:
Looking further, this would possibly mess with a lot of code and I really don't want to refactor large swathes of the code for a simple bug fix, so I prefer to keep what I've got here.

PR #2999, opened 2024-07-12
[QoL] Create default overrides class and export that with custom overrides

OrangeRed:
## What are the changes?
Separate Default Overrides from custom ones used for testing.

As per our discussion regarding how annoying it would be to maintain the default override values inside pre.tests.ts

## What did change?
Separate Default Overrides from custom ones used for testing.
Move Default Overrides into overrides.ts
Check if default export of overrides.ts is deeply equal to the defaults.
Add better type annotation for the modifier overrides

### Screenshots/Videos
N/A

## How to test the changes?
`tsc --noEmit --project tsconfig.json`
`npm run test:silent`
`npm rum docs`

## Checklist
~~- [ ] Are the changes visual?~~
  ~~- [ ] Have I provided screenshots/videos of the changes?~~

### DIFF
--- a/src/modifier/modifier.ts
+++ b/src/modifier/modifier.ts
@@ -2665,7 +2665,8 @@ export function overrideHeldItems(scene: BattleScene, pokemon: Pokemon, player: 
     const modifierType: ModifierType = modifierTypes[itemName](); // we retrieve the item in the list
     let itemModifier: PokemonHeldItemModifier;
     if (modifierType instanceof ModifierTypes.ModifierTypeGenerator) {
-      itemModifier = modifierType.generateType(null, [(item as ModifierTypes.GeneratorModifierOverride).type]).withIdFromFunc(modifierTypes[itemName]).newModifier(pokemon) as PokemonHeldItemModifier;
+      const pregenArgs = "type" in item ? [item.type] : null;
+      itemModifier = modifierType.generateType(null, pregenArgs).withIdFromFunc(modifierTypes[itemName]).newModifier(pokemon) as PokemonHeldItemModifier;
     } else {
       itemModifier = modifierType.withIdFromFunc(modifierTypes[itemName]).newModifier(pokemon) as PokemonHeldItemModifier;
     }
--- a/src/modifier/modifier.ts
+++ b/src/modifier/modifier.ts
@@ -20,7 +20,7 @@ import { achvs } from "../system/achv";
 import { VoucherType } from "../system/voucher";
 import { FormChangeItem, SpeciesFormChangeItemTrigger } from "../data/pokemon-forms";
 import { Nature } from "#app/data/nature";
-import * as Overrides from "../overrides";
+import Overrides from "../overrides";
 import { ModifierType, modifierTypes } from "./modifier-type";
 import { Command } from "#app/ui/command-ui-handler.js";
 import { Species } from "#enums/species";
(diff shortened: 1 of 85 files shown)


src/modifier/modifier.ts
```diff
@@ -2665,7 +2665,7 @@ export function overrideHeldItems(scene: BattleScene, pokemon: Pokemon, player:
     const modifierType: ModifierType = modifierTypes[itemName](); // we retrieve the item in the list
     let itemModifier: PokemonHeldItemModifier;
     if (modifierType instanceof ModifierTypes.ModifierTypeGenerator) {
-      itemModifier = modifierType.generateType(null, [item.type]).withIdFromFunc(modifierTypes[itemName]).newModifier(pokemon) as PokemonHeldItemModifier;
+      itemModifier = modifierType.generateType(null, [(item as ModifierTypes.GeneratorModifierOverride).type]).withIdFromFunc(modifierTypes[itemName]).newModifier(pokemon) as PokemonHeldItemModifier;
```
flx-sta, line 2668: would type-check instead of `as ...` work here too?

src/test/vitest.setup.ts
```diff
@@ -13,7 +13,16 @@ import { initAchievements } from "#app/system/achv.js";
...
+import { beforeAll, vi } from "vitest";
+
+vi.mock("#app/overrides", async (importOriginal) => {
+  const { defaultOverrides } = await importOriginal<typeof import("#app/overrides")>();
+
+  return {
+    default: defaultOverrides,
+    defaultOverrides
+  } satisfies typeof import("#app/overrides");
+});
```
flx-sta, line 25:
```suggestion
/**
 * Takes care of resetting any changes made to the overrides to always return the defaults
 */
vi.mock("#app/overrides", async (importOriginal) => {
  const { defaultOverrides } = await importOriginal<typeof import("#app/overrides")>();

  return {
    default: defaultOverrides,
    defaultOverrides
  } satisfies typeof import("#app/overrides");
});
```

flx-sta (changes requested):
Some little NIT

It's gonna be fun to explain the new way to use overrides to everyone 👍 👍 👍 👍

src/modifier/modifier.ts
OrangeRed: Type checking isn't possible because types can't be used as values so something like `typeof` or `instanceof` will cause an error.

The type of `item` here is `ModifierOverride` which is the same as `GeneratorModifierOverride | BaseModifierOverride`. But `BaseModifierOverride` doesn't have a property of `type` because those types of items don't need that property to be created. Since `class ModifierTypeGenerator` and `type GeneratorModifierOverride` are disconnected from each other there is no way to narrow the `ModifierOverride` type to just be `GeneratorModifierOverride` for typescript to understand that item should have a property of `type` without using a typecast (at least I could not figure out a way to do that so if you know how let me know).

I could do something like the following but I don't know if that is better than a typecast. 
```ts
generateType(null, ["type" in item ? item.type : undefined]) 
```

flx-sta: hell nah. That looks way more confusing.
Resolved ✅

OrangeRed: I did it anyway because its arguably safer to actually check for "type" instead of typecasting and assuming its there.

PR #3001, opened 2024-07-12
[Bug] Fixes bug with Lock-On and Zap Cannon where Zap Cannon misses if Lock-On is the first move on entering battle.

schmidtc1:
## What are the changes?

Fixes bug with Lock-On and Zap Cannon where Zap Cannon misses if Lock-On is the first move on entering battle.

Lock-On should guarantee a hit regardless of when it is used. 

## What did change?

Removed the ```.slice(1)``` call on ```getLastXMoves()``` so that the condition no longer skips over the first turn made.
```getLastXMoves().find(() => true)``` gets the most recent move used in the pokemon's move history.

### Screenshots/Videos

Before fix

https://github.com/user-attachments/assets/61604fa9-aa54-432a-a6ae-31e14b66b72e

After fix

https://github.com/user-attachments/assets/ac76a790-4e9e-44a6-9f81-2ae97fea80d9

## How to test the changes?

Give your mon Lock-On and Zap Cannon. Go to wave 5 and use Lock-On followed by Zap Cannon. See the result.
Be sure to test multiple times. Lock On should never fail with Zap Cannon. 

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -3058,7 +3058,7 @@ export class MoveEffectPhase extends PokemonPhase {
     }
 
     // If the user should ignore accuracy on a target, check who the user targeted last turn and see if they match
-    if (user.getTag(BattlerTagType.IGNORE_ACCURACY) && (user.getLastXMoves().slice(1).find(() => true)?.targets || []).indexOf(target.getBattlerIndex()) !== -1) {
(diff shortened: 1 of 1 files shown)


Tempo-anon:
I don't think Zap Cannon has anything to do with this. This is just for Lock On and Mind Reader on the first turn

schmidtc1:
> I don't think Zap Cannon has anything to do with this. This is just for Lock On and Mind Reader on the first turn

You're right, the issue was noticed from a bug report on discord that used lock on and zap cannon, but it is just a fix for lock on/mind reader

PR #3005, opened 2024-07-13
[Balance] Added candy bonus for Fresh Start challenge

Fontbane:
## What are the changes?

- Multiplies candy friendship gains by 1.5x for starter mons in the Fresh Start challenge

- The candy grind has been noted as needing improvement
- Challenges are a great avenue for rewarding the player with faster candy gains
- Fresh start is a good fit to test this on, as it tests the player's ability to use the starters effectively without relying on passives etc, and also creates an incentive to keep starter mons in this run. All these starters are the same cost and pretty standardized so a flat 1.5 bonus can be used. Other challenges might change their bonus based on cost or other factors.

## What did change?

- Added a new ChallengeType.FRIENDSHIP_BONUS
- Also made functions for it
- Called it before applying classic friendship boost to add it as a factor
- Applies appropriate friendship bonus for challenges- namely, 1.5x for starters in Fresh Start

## How to test the changes?

- Start Fresh Start challenge
- Defeat a mon or use a rare candy to increase friendship
- Insert a console.log for starterAmount.value in src/field/pokemon.ts on line 3136
- See that the friendship is multiplied
- Do this with a non-starter to see it not applied

## Checklist
- [ ] Are the changes visual?


Tempo-anon:
Why not have the bonus applied to all Pokemon in the fresh start challenge rather than just the 27 original starters?

Tempo-anon:
Too many merge conflicts because of beta so I will close it for now. I'm personally in favor of a candy bonus for all Pokemon in this challenge rather than just the initial starters but maybe you can get more opinions from others on what it should be if there are any

PR #3013, opened 2024-07-13
[Bug] Fix sketch being able to copy some unsketchable moves

schmidtc1:
## What are the changes?

Prevents Sketch from copying moves that are supposed to be [unsketchable](https://bulbapedia.bulbagarden.net/wiki/Sketch_(move))

To prevent these moves from being sketchable. 

## What did change?

With this change, you can no longer copy the following moves:
- Chatter
- Mirror Move
- Sleep Talk
- Struggle
- Sketch
- Revival Blessing
- Tera Starstorm
- Breakneck Blitz

NOTE: There are other moves that are technically unsketchable. I kept these as sketchable as other pokemon are given some of these moves as egg moves. These include:
- Dark Void 
- Hyperspace Fury 
- Wicked/Blazing/Noxious/Magical/Combat Torque 

### Screenshots/Videos

Before changes

https://github.com/user-attachments/assets/47d0bfe6-0763-4cfe-8b47-9dc7b51ce56a

After changes

https://github.com/user-attachments/assets/dda8dead-4886-4c9a-880f-6b5b30dea0fc

## How to test the changes?

Use Smeargle with Sketch. Do not give a mon Sketch with the overrides file as it does not properly replace the move. 
Give the opponent unsketchable moves, see that Sketch fails. 

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -92,6 +92,10 @@ export enum MoveFlags {
    * Enables all hits of a multi-hit move to be accuracy checked individually
    */
   CHECK_ALL_HITS   = 1 << 17,
+  /**
+   * Moves that cannot be copied by {@linkcode Moves.SKETCH}
+   */
+  UNSKETCHABLE     = 1 << 18,
```
Tempo-anon, line 98: Should we be adding another bit to every single move or should we just update the sketch function to account for the unsketchable moves?

schmidtc1: It seemed wrong to just hard code that many unsketchable moves into the function, but I can go that route if it makes more sense.

Tempo-anon: My thoughts:
* There are 900+ moves so that seems like it'll be more data than just hardcoding it into the sketch attribute
* This flag only interacts with a single move rather than with multiple moves/abilities (like wind or contact moves) so that makes me more inclined to put it on the attribute rather than as a move flag

Anyone else is welcome to give their feedback. As for moves that cannot be sketched:
* Chatter
* Mirror Move
* Sleep Talk
* Struggle
* Sketch
* Dark Void (added gen 9)
* Hypersapce Fury (added gen 9)
* Revival Blessing
* Tera Starstorm
* Breakneck Blitz (irrelevant since Z-moves are not implemented)
* Wicked/Blazing/Noxious/Magical/Combat Torque (balance team question since those moves have been handed out more commonly)

schmidtc1: Fair enough, I'll isolate it to the Metronome getCondition function. 

As for the moves that cannot be sketched, I think since a lot of these moves are given as egg moves to other mons, it makes sense to allow some of them to be Sketchable. My unsketchable move list would look like: 
- Chatter
- Mirror Move
- Sleep Talk
- Struggle
- Sketch
- Revival Blessing
- Tera Starstorm
- Breakneck Blitz

PR #2913, opened 2024-07-08
[Localization]Localized move-trigger text

mercurius-00:
## What are the changes?

Localized move-trigger text like:
- _"was damaged by the recoil!"_
- _" cut its own HP to power up its move!"_
- _"took {{targetName}}\ninto the sky!"_
- _"incinerated\n{{targetName}}'s {{itemName}}!"_
- _......_

## Why am I doing these changes?

It should be localized

## What did change?

- Extract the text from [data/move.ts](https://github.com/mercurius-00/pokerogue/blob/main/src/data/move.ts) to [locale/[language]/move-trigger.ts](https://github.com/mercurius-00/pokerogue/blob/main/src/locales/en/move-trigger.ts)
- Change some English text to match the original game.
    | original                                      | changed                                         |
    | ------------------------------------------- | ----------------------------------------------- |
    | " is hit with recoil!"                      | " was damaged by the recoil!"                   |
    | "took in sunlight!"                         | "absorbed light!"                               |
    | "dug a hole!"                               | "burrowed its way under the ground!"            |
    | "lowered\nits head!"                        | "tucked in its head!"                           |
    | "is glowing!"                               | "became cloaked in a harsh light!"              |
    | "is charging its power!"                    | "is absorbing power!"                           |
    | " is trying\nto take its foe down with it!" | " is hoping to take its attacker down with it!" |
- Fixed type name localization

### Screenshots/Videos

<details>

<summary>✅En</summary>

image
image
image
image

</details>

<details>
<summary>⬜Es</summary>

Missing

</details>

<details>
<summary>⬜It</summary>

Missing

</details>

<details>
<summary>✅Fr</summary>

image
image
image
image

</details>

<details>
<summary>✅De</summary>

image
image
image
image

</details>

<details>
<summary>✅Pt_BR</summary>

image
image
image
image

</details>

<details>
<summary>✅Zh-CN</summary>

image
image
image
image

</details>

<details>
<summary>✅Zh-TW</summary>

image
image
image
image

</details>

<details>
<summary>✅Ko</summary>

image
image
image
image

</details>

## How to test the changes?

1. set overrides.ts
`export const MOVESET_OVERRIDE: Array<Moves> = [Moves.SKY_DROP, Moves.BEAK_BLAST, Moves.MIND_BLOWN, Moves.STEEL_BEAM];`
2. run `npm run  start:dev`
3. enter a battle and use moves

## Checklist
- [x] There is no overlap with another PR?
[...]

PR #3014, opened 2024-07-13
[Bug] Fix Text for Future Sight and Doom Desire

CodeTappert:
## What are the changes?

It now correctly uses pokemonNameWithAffix and also doesnt show {{user}} randomly in the middle

It was a bug introduced in https://github.com/pagefaultgames/pokerogue/pull/2913 so i went through all changed moves there and found those two were wrong

## What did change?

Changed the attribute they were using to replace the placeholder correctly and also dont use "getPokemonMessage" anymore which we want to phase out anyways

### Screenshots/Videos

Before:

https://github.com/user-attachments/assets/620eb4fe-5768-4ca2-877c-fbdefd8958df

After:

https://github.com/user-attachments/assets/0d0c48e7-b86d-4510-998c-b99daebb3f52

## How to test the changes?

Override your pokemon to have those two moves and use them

## Checklist
- [x] Are the changes visual?


CodeTappert:
image

PR #3017, opened 2024-07-13
[Bug] Fixing inaccuracies with handling Struggle's recoil

frutescens:
## What are the changes?
According to [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Struggle_(move)), Struggle's recoil bypasses the abilities Magic Guard. This PR adds a check for Struggle so that Struggle's recoil damage is applied to Pokemon with Magic Guard

I was working on my Magic Guard test and noticed that Struggle's recoil bypassed Magic Guard. 

## What did change?
data/moves.ts --> Checks for Struggle before applying ability 
 
### Screenshots/Videos
Bulbasaur (through overrides.ts) has been given Magic Guard
Uploading Screen Recording 2024-07-13 at 4.34.05 PM.mov…

## How to test the changes? 
Run out of all PP with a Pokemon with Magic Guard --> check for recoil damage

## Checklist


src/data/ability.ts
```diff
@@ -4524,7 +4524,7 @@ export function initAbilities() {
     new Ability(Abilities.FRISK, 4)
       .attr(FriskAbAttr),
     new Ability(Abilities.RECKLESS, 4)
-      .attr(MovePowerBoostAbAttr, (user, target, move) => move.hasFlag(MoveFlags.RECKLESS_MOVE), 1.2),
+      .attr(MovePowerBoostAbAttr, (user, target, move) => move.hasFlag(MoveFlags.RECKLESS_MOVE) && move.id !== Moves.STRUGGLE, 1.2),
```
innerthunder, line 4527: Struggle already isn't boosted by Reckless (it doesn't have `.recklessMove()` in its instantiation)

src/data/move.ts
```diff
@@ -1083,7 +1083,7 @@ export class RecoilAttr extends MoveEffectAttr {
     }
 
     const cancelled = new Utils.BooleanHolder(false);
-    if (!this.unblockable) {
+    if (!this.unblockable && move.id !== Moves.STRUGGLE) {
```
innerthunder, line 1086: `unblockable` is `true` for Struggle (and only Struggle), so this check is already accounted for. I think the real issue is that `unblockable` isn't accounted for when Magic Guard is applied.

frutescens:
https://github.com/user-attachments/assets/380e032e-1a94-4df2-883f-465a2ed75d34

PR #3018, opened 2024-07-13
[Bug] Fix and replacement for getSheerForceHitDisableAbCondition() loop

NxKarim:
## What are the changes?
SheerForceHitDisabled for abilities that can get disabled by Sheer Force.
SheerForceHitDisableAbAttr for Sheer Force to check the opponent's ability.
A few format and info changes also were made to a previous Sheer Force related commit.

Both implemented to fix an infinite loop between abilities with the old getSheerForceHitDisableAbCondition() condition.
Abilities tried to check for Sheer Force between each other, causing the loop.
Affected abilities: Color Change, Pickpocket, Wimp Out, Emergency Exit, Berserk, Anger Shell.

## What did change?
Instead of using a condition, this Sheer Force edge case now uses attributes.

## Important Note 
**This implementation still has one main issue**, since it depends on the ability suppressed status, Sheer Force can cause previous suppressed from other sources to be altered (ability enabled) but this is only against the affected abilities. 

- This PR was requested by @demi.glace.sauce.

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -1513,6 +1513,39 @@ export class PostAttackAbAttr extends AbAttr {
   }
 }
 
+/**
+ * Attribute for Sheer Force interaction with certain abilities.
+ * Checks if last move used against target was affected by a Sheer Force user and:
+ * Supresses: Color Change, Pickpocket, Wimp Out, Emergency Exit, Berserk, Anger Shell
+ * @returns {AbAttrCondition} If false disables the ability which the condition is applied to.
+ */
+export class SheerForceHitDisableAbAttr extends PreAttackAbAttr {
+
+  constructor() {
+    super(false);
+  }
+
+  applyPreAttack(pokemon: Pokemon, passive: boolean, defender: Pokemon, move: Move, args: any[]): boolean {
+
+    /**if the last move chance is greater than or equal to cero, and the attacker's ability is sheer force*/
+    const SheerForceAffected = allMoves[move.id].chance >= 0 && defender.hasAbilityWithAttr(SheerForceHitDisabled);
+
+    defender.summonData.abilitySuppressed = SheerForceAffected;
+
+    return true;
+  }
+}
+/**
+ * Attribute applied to abilities related to Sheer Force.
+ * Allows for suppressing: Color Change, Pickpocket, Wimp Out, Emergency Exit, Berserk, Anger Shell
+ * @returns {AbAttr}
+ */
+export class SheerForceHitDisabled extends AbAttr {
+  constructor() {
+    super(false);
+  }
+}
+
 export class PostAttackStealHeldItemAbAttr extends PostAttackAbAttr {
   private condition: PokemonAttackCondition;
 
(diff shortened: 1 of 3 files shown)


Tempo-anon:
Tests are failing

src/data/ability.ts
```diff
@@ -1513,6 +1513,39 @@ export class PostAttackAbAttr extends AbAttr {
...
+  }
+
+  applyPreAttack(pokemon: Pokemon, passive: boolean, defender: Pokemon, move: Move, args: any[]): boolean {
+
+    /**if the last move chance is greater than or equal to cero, and the attacker's ability is sheer force*/
+    const SheerForceAffected = allMoves[move.id].chance >= 0 && defender.hasAbilityWithAttr(SheerForceHitDisabled);
+
+    defender.summonData.abilitySuppressed = SheerForceAffected;
+
+    return true;
```
mcmontag, line 1535: This un-suppresses abilities that are currently suppressed.  Is that the point?

src/data/ability.ts
```diff
@@ -1513,6 +1513,39 @@ export class PostAttackAbAttr extends AbAttr {
...
+export class SheerForceHitDisableAbAttr extends PreAttackAbAttr {
+
+  constructor() {
+    super(false);
+  }
+
+  applyPreAttack(pokemon: Pokemon, passive: boolean, defender: Pokemon, move: Move, args: any[]): boolean {
+
+    /**if the last move chance is greater than or equal to cero, and the attacker's ability is sheer force*/
+    const SheerForceAffected = allMoves[move.id].chance >= 0 && defender.hasAbilityWithAttr(SheerForceHitDisabled);
```
mcmontag, line 1531: this `const` should be camelCased

src/data/ability.ts
mcmontag: I'm also not seeing where `abilitySuppressed` is reversed.  So what happens if:
- Sheer Force activates -> disables Pickpocket
- Attacker loses Sheer Force, or switches out, or the Pickpocket target is hit by someone else -> ???

Wouldn't Pickpocket still be disabled in all of these cases?

src/data/ability.ts
```diff
@@ -1513,6 +1513,39 @@ export class PostAttackAbAttr extends AbAttr {
   }
 }
 
+/**
+ * Attribute for Sheer Force interaction with certain abilities.
+ * Checks if last move used against target was affected by a Sheer Force user and:
+ * Supresses: Color Change, Pickpocket, Wimp Out, Emergency Exit, Berserk, Anger Shell
+ * @returns {AbAttrCondition} If false disables the ability which the condition is applied to.
```
DustinLin, line 1520: you need a `{@linkcode AbAttrConidtion}` to get the jsDocs to show the referenced class

src/data/ability.ts
```diff
@@ -1513,6 +1513,39 @@ export class PostAttackAbAttr extends AbAttr {
...
+ */
+export class SheerForceHitDisableAbAttr extends PreAttackAbAttr {
+
+  constructor() {
+    super(false);
+  }
+
+  applyPreAttack(pokemon: Pokemon, passive: boolean, defender: Pokemon, move: Move, args: any[]): boolean {
+
+    /**if the last move chance is greater than or equal to cero, and the attacker's ability is sheer force*/
```
DustinLin, line 1530: this comment is just repeating the logic below, should be removed ["Rule 1: Comments should not duplicate the code"](https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/#h2-b7593c698a940)

src/data/ability.ts
```diff
@@ -1513,6 +1513,39 @@ export class PostAttackAbAttr extends AbAttr {
...
+
+    defender.summonData.abilitySuppressed = SheerForceAffected;
+
+    return true;
+  }
+}
+/**
+ * Attribute applied to abilities related to Sheer Force.
+ * Allows for suppressing: Color Change, Pickpocket, Wimp Out, Emergency Exit, Berserk, Anger Shell
+ * @returns {AbAttr}
```
DustinLin, line 1541: is this comment needed? this attr is just a constructor that calls its parent?

src/data/ability.ts
```diff
@@ -1513,6 +1513,39 @@ export class PostAttackAbAttr extends AbAttr {
...
+ */
+export class SheerForceHitDisabled extends AbAttr {
```
DayKev, line 1543: The name of this class is inconsistent with when it is checked for later on, one or the other needs to be renamed.

flx-sta:
@NxKarim  closed for being stale (no updates for over a month)

----

PR #3019, opened 2024-07-14
Update dialogue.ts

truisseaux:
finished translating all the dialogue, corrected and organized everything (normal trainers, team grunts, team leaders, gym leaders, elite 4, champions) both for the male and female protag.
tested this on my side with the localhost and the override file thing.

terminé de traduire tout les dialogues, tout corrigé et organisé (dresseurs normaux, sbires, leaders, champions, conseil 4, maîtres) pour le protag garçon et fille.
testé de mon côté avec le localhost et le fichier override machin

### DIFF
--- a/src/locales/fr/dialogue.ts
+++ b/src/locales/fr/dialogue.ts
@@ -5,40 +5,40 @@ export const PGMdialogue: DialogueTranslationEntries = {
   "youngster": {
     "encounter": {
       1: "Hé ! Combat ?",
-      2: "Toi aussi tu débutes ?",
+      2: "Toi aussi, tu débutes ?",
       3: "Hé, j’me souviens pas de ta tête. Combat !",
-	  4: "J’ai perdu, alors j’essaye de capturer d’autres Pokémon.\nHé, t’as l’air faible toi ! Allez, combat !",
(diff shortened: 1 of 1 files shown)


truisseaux:
also yeah, did make this pull request like a week ago, but "if i want to translate everything, might as well do everything at once"

Dakurei:
A translation that adds more errors than it corrects, notably by replacing `…` with `...` and deleting non-breaking spaces.

Reorganization of entries only on this file, making future modifications complicated when looking at the differences with the English reference file. (Including to check the differences of this modification)

Non-compliance with the PR template, and non-application of the modification to the beta branch instead of the main branch.

A total no go

Tempo-anon:
Please make the PR with the proper PR formatting, aimed at beta, and fix all the mistakes. Will close this PR for now.

truisseaux:
I'm sorry, I didn't know. I usually just type `...` instead of `…`, and I wanted everything to be consistent, so I replaced every `…` with `...`. I didn't think it would be a problem.
If I provide a dialogue.ts file where everything is translated, without sorting, and with non-breaking spaces where they should be, would that be ok ?

40chyan:
@Adri1 so the author definitely needs help from translation team to create a proper PR, idk if you want to get translation texts from this guy so you can upload that with a new PR?

PR #3035, opened 2024-07-16
[Balance] Fix Baton Pass TM learnset

EmberCM:
## What are the changes?
Added and removed some Pokemon from [Baton Pass TM](https://bulbapedia.bulbagarden.net/wiki/Baton_Pass_(move)) learnset

Someone asked for it in [Discord](https://discord.com/channels/1125469663833370665/1262618712674926704)

## Checklist


src/data/tms.ts
```diff
@@ -29332,7 +29332,11 @@ export const tmSpecies: TmSpecies = {
     Species.POLIWRATH,
     Species.RAPIDASH,
```
Tempo-anon, line 29333: Just a quick check reveals Rapidash also lost access to baton pass via tm in gen 9, as well as farefetchd

Tempo-anon (changes requested):
We really need to agree on how to deal with TM levelsets and what to do in cases where there's Pokemon that lose learning moves in between generations

EmberCM:
> We really need to agree on how to deal with TM levelsets and what to do in cases where there's Pokemon that lose learning moves in between generations

The way I implemented it currently was giving it to all the Pokemon in the list no matter the generation or if they lost it.

Fontbane:
> > We really need to agree on how to deal with TM levelsets and what to do in cases where there's Pokemon that lose learning moves in between generations
> 
> The way I implemented it currently was giving it to all the Pokemon in the list no matter the generation or if they lost it.

Yeah this is how we've been doing things so far, should be fine. Also maybe change from [Bug] to [Balance].

Tempo-anon:
Too many merge conflicts... closing this PR for now

PR #3036, opened 2024-07-16
[Move] GMAX moves weather implemented

DTZK:
## What are the changes?

I added weather conditions added from Gmax moves

I found out that although gmax moves were in the code, weather conditions were not set.

no

The gmax moves effects were not added in so I decided to add it in.

## What did change?

Added hail to MAX_HAILSTORM, added sandstorm in MAX_ROCKFALL, sunny in max flare, and rain in max gyser

### Screenshots/Videos

proof

## How to test the changes?

no

no

no

## Checklist
- [ x] There is no overlap with another PR?
- [ x] The PR is self-contained and cannot be split into smaller PRs?
- [ x] Have I provided a clear explanation of the changes?
- [x ] Are the changes visual?
  - [ x] Have I provided screenshots/videos of the changes?


CodeTappert:
How is this localization?

Tempo-anon:
This should also point to pagefaultgames:beta

src/enums/arena-tag-type.ts
```diff
@@ -21,5 +21,7 @@ export enum ArenaTagType {
   MAT_BLOCK = "MAT_BLOCK",
   CRAFTY_SHIELD = "CRAFTY_SHIELD",
   TAILWIND = "TAILWIND",
+  // SAFEGUARD = "SAFEGUARD",
```
Tempo-anon, line 24: Why does this PR also comment out all safeguard stuff

DTZK: I made the safeguard code before this but then exams came so I commented it out from my repo before this

CodeTappert: Then remove the change from this PR

DTZK: > Then remove the change from this PR

I have done so

Tempo-anon:
Not sure if this is needed for now since GMAX moves are not implemented. If we ever do decide to implement them we can do the change then.

PR #3034, opened 2024-07-16
[Localization]Localized move-trigger text fix

mercurius-00:
## What are the changes?

Localized move-trigger text like:
- _"was damaged by the recoil!"_
- _" cut its own HP to power up its move!"_
- _"took {{targetName}}\ninto the sky!"_
- _"incinerated\n{{targetName}}'s {{itemName}}!"_
- _......_

## Why am I doing these changes?

It should be localized

## What did change?

- Extract the text from [data/move.ts](https://github.com/mercurius-00/pokerogue/blob/main/src/data/move.ts) to [locale/[language]/move-trigger.ts](https://github.com/mercurius-00/pokerogue/blob/main/src/locales/en/move-trigger.ts)
- Change some English text to match the original game.
    | original                                      | changed                                         |
    | ------------------------------------------- | ----------------------------------------------- |
    | " is hit with recoil!"                      | " was damaged by the recoil!"                   |
    | "took in sunlight!"                         | "absorbed light!"                               |
    | "dug a hole!"                               | "burrowed its way under the ground!"            |
    | "lowered\nits head!"                        | "tucked in its head!"                           |
    | "is glowing!"                               | "became cloaked in a harsh light!"              |
    | "is charging its power!"                    | "is absorbing power!"                           |
    | " is trying\nto take its foe down with it!" | " is hoping to take its attacker down with it!" |
- Fixed type name localization

### Screenshots/Videos

<details>

<summary>✅En (Fully tested)</summary>

1. STRUGGLE
image
2. STEEL_BEAM
image
3. COVET
image
4. INCINERATE
image
5. KNOCK_OFF
image
6. ELECTRO_SHOT
image
7. FUTURE_SIGHT
image
8. DOOM_DESIRE
image
9. BELLY_DRUM
image
10. PSYCH_UP
image
12. HEART_SWAP
image
13. FICKLE_BEAM
image
14. MAGNITUDE
image
15. PRESENT
image
16. JUMP_KICK
image
17. LOCK_ON
image
18. WHIRLWIND
image
19. CAMOUFLAGE
image
20. SOAK
image
21. MIMIC
image
22. SKETCH
image
23. WORRY_SEED
image
24. ROLE_PLAY
image
25. DOODLE
image
26. ENTRAINMENT
image
27. SKILL_SWAP
image
28. TRANSFORM
image
29. PAY_DAY
image
30. DESTINY_BOND
image
31. POLTERGEIST
image
32. RAZOR_WIND
33. FLY
image
[...]

PR #2972, opened 2024-07-11
[Localization] Localize ability messages in ability.ts (trigger, apply...)

EnochG1:
## What are the changes?
Add the option to localize message text of ability events.

## Why am I doing these changes?
All text should be localizable.

## What did change?
 - Change the code for localize > src/data/ability.ts
 - Added localization keys in each language folders. (ability-trigger.ts) > missed key included.

### Screenshots/Videos
**Note : I provide just essential test result (to check parameters)
Please comment if you want to check specific case.**
Some case can't be test (message event not triggered). See https://github.com/pagefaultgames/pokerogue/issues/3006

<details>
<summary>Click to expand/collapse</summary>

| Ability | En | Ko |
| --- | --- | --- |
| VOLT_ABSORB | image | image |
| DISGUISE | image | image |
| LIQUID_OOZE | image | image |
| COLOR_CHANGE | image | image |
| ROUGH_SKIN | image | image |
| WANDERING_SPIRIT | image | image |
| MUMMY | image | image |
| CURSED_BODY | image | image |
| LIBERO | image | image |
| MAGICIAN | image | image |
| PICKPOCKET | image | image |
| OWN_TEMPO | image | image |
| HOSPITALITY | image | image |
| CURIOUS_MEDICINE | image | image |
| IMPOSTER | image | image |
| CLEAR_BODY | image | image |
| FOREWARN | image | image |
| FRISK | image | image |
| ICE_BODY | image | image |
| SOLAR_POWER | image | image |
| HARVEST | image | image |
| BALL_FETCH | image | image |
| CHEEK_POUCH | image | image |
| ARENA_TRAP | image | image |
| PICKUP | image | image |
| AFTERMATH | image | image |
| INNARDS_OUT | image | image |
| PRESSURE | image | image |
| MOLD_BREAKER | image | image |
| ANTICIPATION | image | image |
| TURBOBLAZE | image | image |
| TERAVOLT | image | image |
| DARK_AURA | image | image |
| FAIRY_AURA | image | image |
| NEUTRALIZING_GAS | image | image |
| AS_ONE_GLASTRIER | image | image |
| AS_ONE_SPECTRIER | image | image |
| VESSEL_OF_RUIN | image | image |
| SWORD_OF_RUIN | image | image |
| TABLETS_OF_RUIN | image | image |
| BEADS_OF_RUIN | image | image |

</details>

## How to test the changes?
Manually (Using override.ts)

### Note : message key and possible abilities matching list

<details>
[...]

PR #3039, opened 2024-07-16
[Misc] Merged implementation of CutHpStatBoostAttr moves

frutescens:
## What are the changes?
Before, Belly Drum was implemented separately from other moves that exchanged HP for stat boosts like Fillet Away and Clangorous Soul. In addition, the implementation of non-Belly Drum moves did not have a fail condition when all affected stats were already maxed out. This merges the two implementations into one and adds a check.

I thought it would be nice to reduce redundant code + I noticed this while working on some other stuff. 

## What did change?
- Deleted HalfHpStatMaxAttr
- Added to CutHpStatBoostAttr
   - Checks for fully-maxed stats
   - Additional check for unique single-stat max out message (for Belly Drum specifically)

### Screenshots/Videos

Stats @ Max Failure Message 1
https://github.com/user-attachments/assets/ecab240c-aa86-40d6-80cb-d2abdc27d2f6

Stats @ Max Failure Message 2
https://github.com/user-attachments/assets/383361cb-ef71-449e-aebc-268b6095e3e4

2/3 Stats @ Max Success Message
https://github.com/user-attachments/assets/ded648f8-d876-4cb0-ad11-85714c3c80d6

Belly Drum functionality
https://github.com/user-attachments/assets/35958fb3-4e4c-4ef5-8a3e-f3808cd775d7

## How to test the changes?
Would recommend use of overrides.ts

## Checklist


src/data/move.ts
```diff
@@ -6281,7 +6261,7 @@ export function initMoves() {
     new StatusMove(Moves.SWEET_KISS, Type.FAIRY, 75, 10, -1, 0, 2)
       .attr(ConfuseAttr),
     new SelfStatusMove(Moves.BELLY_DRUM, Type.NORMAL, -1, 10, -1, 0, 2)
-      .attr(HalfHpStatMaxAttr, BattleStat.ATK),
+      .attr(CutHpStatBoostAttr, [BattleStat.ATK], 6, 2),
```
Tempo-anon, line 6264: Belly drum doesn't raise attack by 6, it raises it by 12

src/data/move.ts
```diff
@@ -2562,13 +2539,16 @@ export class CutHpStatBoostAttr extends StatChangeAttr {
       user.damageAndUpdate(Math.floor(user.getMaxHp() / this.cutRatio), HitResult.OTHER, false, true);
       user.updateInfo().then(() => {
         const ret = super.apply(user, target, move, args);
+        if (this.levels === 6 && this.stats.length === 1) {
+          user.scene.queueMessage(getPokemonMessage(user, ` cut its own HP\nand maximized its ${getBattleStatName(this.stats[0])}!`));
+        }
```
Tempo-anon, line 2544: the levels check should be 12 instead of 6 and this should also have a comment explaining this is specifically for belly drum

frutescens:
Fixed!

src/data/move.ts
```diff
@@ -2562,13 +2539,17 @@ export class CutHpStatBoostAttr extends StatChangeAttr {
       user.damageAndUpdate(Math.floor(user.getMaxHp() / this.cutRatio), HitResult.OTHER, false, true);
       user.updateInfo().then(() => {
         const ret = super.apply(user, target, move, args);
+        if (this.levels === 12 && this.stats.length === 1) {
+          //This message is specifically for Belly Drum
+          user.scene.queueMessage(getPokemonMessage(user, ` cut its own HP\nand maximized its ${getBattleStatName(this.stats[0])}!`));
```
schmidtc1, line 2544: It would be a good idea to localize this message with i18next

frutescens: Yes, definitely. I think there is an effort to localize these messages in move.ts / ability.ts - and I'm a bit linguistically challenged so I will ask around

src/data/move.ts
```diff
@@ -2562,13 +2539,17 @@ export class CutHpStatBoostAttr extends StatChangeAttr {
       user.damageAndUpdate(Math.floor(user.getMaxHp() / this.cutRatio), HitResult.OTHER, false, true);
       user.updateInfo().then(() => {
         const ret = super.apply(user, target, move, args);
+        if (this.levels === 12 && this.stats.length === 1) {
+          //This message is specifically for Belly Drum
```
Tempo-anon, line 2543: I'd put this comment above the `if` since this entire if block is a check only done for belly drum and the only difference is that it displays a message exclusively for the move

src/data/move.ts
```diff
@@ -2562,13 +2539,17 @@ export class CutHpStatBoostAttr extends StatChangeAttr {
...
+          user.scene.queueMessage(getPokemonMessage(user, ` cut its own HP\nand maximized its ${getBattleStatName(this.stats[0])}!`));
+        }
         resolve(ret);
       });
     });
   }
 
   getCondition(): MoveConditionFunc {
-    return (user, target, move) => user.getHpRatio() > 1 / this.cutRatio;
+    return (user, target, move) => user.getHpRatio() > 1 / this.cutRatio && this.stats.some(s => user.summonData.battleStats[s] < 6);
```
Tempo-anon, line 2552: What does the `this.stats.some(s => user.summonData.battleStats[s] < 6)` check for? Doesn't this just check that at least one of the user's stats aren't maxed out? Wouldn't this be true for most cases?

frutescens: Yes, but all moves - Belly Drum, Clangorous Soul, and Fillet Away - fail if all involved stats are maxed out per Bulbapedia.

torranx, line 2552: I think @Tempo-anon is right here, the code will always be true if at least one of the user stat is not at max.

Belly Drum

Belly Drum fails if the user's current HP is less than half its maximum, or if the user's Attack is already at +6 (even if the user has [Contrary](https://bulbapedia.bulbagarden.net/wiki/Contrary_(Ability))).

torranx:
I think this warrants creation of integration tests for involved moves

src/data/move.ts
frutescens: If you look at the second and third video, it's not true. this.stats =/= the pokemon's stats - it's the stats used when declaring the class
+ the length of this.stats is also used for the conditional checking for belly drum - if it's all the pokemon's stats, it wouldn't work

torranx: The existing belly drum implementation uses this condition `user.summonData.battleStats[this.stats[BattleStat.ATK]] < 6` - it's explicitly checking `BattleStat.ATK`. 

Now in your change, what happens if `battleStats` changed say for `SPD`?

frutescens: The user's battleStats? It'll depend on how the move is initialized / its parameters

Let's look at how Fillet Away is implemented. 
`new SelfStatusMove(Moves.FILLET_AWAY, Type.NORMAL, -1, 10, -1, 0, 9)
      .attr(CutHpStatBoostAttr, [ BattleStat.ATK, BattleStat.SPATK, BattleStat.SPD ], 2, 2),`
Say that ATK:6, SPATK:4, and SPD:6

So... here we know that -
this.stats = [ BattleStat.ATK, BattleStat.SPATK, BattleStat.SPD ]

When we put this into
`this.stats.some(s => user.summonData.battleStats[s] < 6`
user.summonData.battleStats[s[0]] = user.summonData.battleStats[BattleStat.ATK] = 6 -> F
user.summonData.battleStats[s[1]] = 4 -> T
user.summonData.battleStats[s[2]] = 6 -> F
`
Index values inserted for clarity. 

Per Bulbapedia -> Fillet Away fails if the user's current HP is less than or equal to half its maximum, or if the user's aforementioned stats are all at +6 stages

In your question, there would be a move implemented like this : 
`new SelfStatusMove(Moves.SOME_MOVE, Type.NORMAL, -1, 10, -1, 0, 9)
      .attr(CutHpStatBoostAttr, [ BattleStat.SPD ], 12, 2),

Then it will go through the same checks as above. Fail if the user's SPD battlestat is at 6. When it succeeds, it will print out "User cut its HP to maximize its **Speed**!"

src/data/move.ts
```diff
@@ -2562,13 +2539,17 @@ export class CutHpStatBoostAttr extends StatChangeAttr {
       user.damageAndUpdate(Math.floor(user.getMaxHp() / this.cutRatio), HitResult.OTHER, false, true);
       user.updateInfo().then(() => {
         const ret = super.apply(user, target, move, args);
+        //This if-block is specifically for Belly Drum and its unique message
+        if (this.levels === 12 && this.stats.length === 1) {
+          user.scene.queueMessage(getPokemonMessage(user, ` cut its own HP\nand maximized its ${getBattleStatName(this.stats[0])}!`));
```
EnochG1, line 2544: @frutescens 
Hi,
move.ts localization is progressing at #3034. You can refer it.
(And ability.ts messages are localized and merged in beta #2972

frutescens: Thank you! I updated my code! and I'll leave a note in the comments

frutescens:
Waiting for implementation of this PR to finish --> https://github.com/pagefaultgames/pokerogue/pull/3034

PR #3041, opened 2024-07-16
[Bug] Fixes bug with frenzy moves keeping the confusion counter despite disruption

schmidtc1:
## What are the changes?

Frenzy moves (Outrage, Thrash, Petal Dance, Raging Fury) will maintain the confusion counter when disrupted, causing confusion immediately when calling the move again.
E.g. Use outrage on opponent and 1HKO that opponent, enemy switches in fairy type (outrage disrupted). Use outrage again later on and you will be immediately confused after one use.

This is a bug with all frenzy moves

## What did change?

Added the NoEffectAttr with frenzyMissFunc to each of the frenzy moves to lapse the Frenzy tag when a move has no effect. 
Refactored the FrenzyAttr apply function to actually lapse the tag each turn. 
Account for disruption by removing the tag in frenzyMissFunc
Removes ```tag.turnCount = 0``` in removeTag, since other functions do not rely on turnCount when being removed, like Frenzy does.

### Screenshots/Videos

Before (see [conversation](https://github.com/pagefaultgames/pokerogue/pull/3041#issuecomment-2231768795))
After (3 turn Outrage - disrupted with no confusion, followed by 2 turn Outrage - disrupted with confusion)

https://github.com/user-attachments/assets/80a45caf-f2fa-43ed-a77c-a198d505b576

## How to test the changes?

Give opponent Outrage, give yourself a non-Fairy mon and a Fairy mon. Let the opponent use outrage on you once, switch into Fairy type, switch back to original mon and let opponent use Outrage again. See change.

## Checklist
- [x] Are the changes visual?


flx-sta:
If I remember correctly the current behavior is the same as in the original Pokemon games. isn't it?

schmidtc1:
> If I remember correctly the current behavior is the same as in the original Pokemon games. isn't it?

I don't think it is. Outrage can be disrupted by type immunity, which should get rid of the Frenzy battle tag and stop it from executing. See below video for example of the bug. 

- Outrage is not disrupted from fairy type immunity.
- Charmander will get confused after 1 turn of outrage at the end.

https://github.com/user-attachments/assets/20ea57fb-ea81-4e36-966f-9bc84d4a0030

src/data/battler-tags.ts
```diff
@@ -480,14 +480,16 @@ export class NightmareTag extends BattlerTag {
 }
 
 export class FrenzyTag extends BattlerTag {
-  constructor(sourceMove: Moves, sourceId: integer) {
-    super(BattlerTagType.FRENZY, BattlerTagLapseType.CUSTOM, 1, sourceMove, sourceId);
+  constructor(turnCount: integer, sourceMove: Moves, sourceId: integer) {
```
Tempo-anon, line 483: we should be using `number` instead of `integer`

src/data/move.ts
```diff
@@ -3902,27 +3902,24 @@ export class FrenzyAttr extends MoveEffectAttr {
...
+    return true;
   }
 }
 
 export const frenzyMissFunc: UserMoveConditionFunc = (user: Pokemon, move: Move) => {
   while (user.getMoveQueue().length && user.getMoveQueue()[0].move === move.id) {
     user.getMoveQueue().shift();
   }
-  user.lapseTag(BattlerTagType.FRENZY);
+  user.removeTag(BattlerTagType.FRENZY); // FRENZY tag should be disrupted on miss/no effect
```
Tempo-anon, line 3922: Does this mean if your second Thrash misses or is trying to attack a ghost you just break out of the frenzy? Do you still get confused? Bulbapedia says that if it gets disrupted on the last turn the user is still confused

src/data/battler-tags.ts
schmidtc1: Is sourceId meant to be number as well?

src/data/move.ts
schmidtc1: If the second thrash misses or tries to attack a ghost, you do break out of the frenzy. Bulbapedia says that if it gets [disrupted on the final, confusion-inducing turn](https://bulbapedia.bulbagarden.net/wiki/Thrash_(move)#:~:text=If%20a%20disruption%20occurs%20on%20what%20would%20have%20been%20the%20final%2C%20confusion%2Dinducing%20turn%20of%20Thrash%2C%20the%20user%20will%20still%20become%20confused.), then it will confuse the user.
As an example, if the Frenzy count is for 2 turns, the user will be confused only on the 2nd turn (not the first).
If the Frenzy count is for 3 turns, the user will be confused only on the 3rd turn (not the first or second).

src/data/battler-tags.ts
schmidtc1: Changed every `integer` usage in battler-tags.ts to `number` in latest commit.

CodeTappert:
I can attest that this doesnt cause the enemy pokemon to be confused directly again when hitting the non.fairy pokemon

CodeTappert (approved):

PR #3047, opened 2024-07-17
[Move] Heal Block

frutescens:
## What are the changes?
This implements Heal Block and provides a test for it. 

I think it would be nice for it to be implemented. 

## What did change?
- HealBlockTag (Duration: 5 Turns)
- PokemonHealPhase now accounts for HealBlockTag 
- Not implemented yet: moves restricted by heal block 
- Not implemented yet: Tests 
- Lines for Localization 

### Screenshots/Videos
https://github.com/user-attachments/assets/08b71f5d-3325-437d-b4d4-975b7f803973

## How to test the changes?
Use Heal Block during a Pokemon battle and see if the affected pokemon can heal itself. 

## Checklist
- [x] Are the changes visual?


src/phases.ts
```diff
@@ -4734,10 +4734,16 @@ export class PokemonHealPhase extends CommonAnimPhase {
     const fullHp = pokemon.getHpRatio() >= 1;
 
     const hasMessage = !!this.message;
+
+    const healBlock = pokemon.getTag(BattlerTagType.HEAL_BLOCK);
```
Tempo-anon, line 4738: nit: if this is going to be a boolean then this should be `isHealBlocked` to make it clear it's a bool

src/test/moves/heal_block.test.ts
```diff
@@ -0,0 +1 @@
+
```
Tempo-anon, line 1: I assume this needs to be filled out. There's a lot of edge cases for heal block though, like items (leftovers, shell bell), abilities (dry skin and volt absorb will absorb the move but not heal, ice body and poison heal don't heal). Just follow what's on Bulbapedia for now: https://bulbapedia.bulbagarden.net/wiki/Heal_Block_(move)

src/phases.ts
frutescens: it's not a boolean; it's a tag - so it's a check for tag or null

src/test/moves/heal_block.test.ts
frutescens: yep, yep - lots to do

src/phases.ts
Tempo-anon: Ah I see. Do we have a `hasTag` or can we also just use `!!pokemon.getTag(BattlerTagType.HEAL_BLOCK)`?

frutescens: well, I'm using the tag instance to call a function so that a failure message is provided.

src/phases.ts
```diff
@@ -4734,10 +4734,16 @@ export class PokemonHealPhase extends CommonAnimPhase {
...
     const healOrDamage = (!fullHp || this.hpHealed < 0);
     let lastStatusEffect = StatusEffect.NONE;
 
-    if (healOrDamage) {
+    if (healBlock) {
```
innerthunder, line 4742:
```suggestion
    if (!!healBlock && healBlock instanceof HealBlockTag) {
```
Should fix the gh pages issue

frutescens:
TBC Tests that involve moves disabled by heal block + misc. interactions
image

PR #3054, opened 2024-07-17
[Misc] Allow Necrozma forms to play their special music

Madmadness65:
## What are the changes?

Necrozma's unique battle music can now play for the Dusk Mane, Dawn Wings, and Ultra forms (These forms still cannot be found in the wild normally, but these will play if forced into an encounter).

I previously added their music for some future use, such as if someone wanted to use them as unique boss fights. Since they now play their unique battle themes naturally with this change, it makes the music "used" now.

## What did change?

Necrozma no longer just plays the Solgaleo & Lunala battle theme no matter what. It plays that theme in its base form, plays the Dusk Mane & Dawn Wings theme in those respective forms, and plays the Ultra Necrozma theme in that form.

### Screenshots/Videos

https://github.com/user-attachments/assets/85cfbb9b-b6b4-4f35-9249-8633a740e537

## How to test the changes?

There isn't a way to force the form index of an opponent at present, so the easiest way to test would be to temporarily add Necrozma to the `getSpeciesFormIndex` in `battle-scene.ts` and encounter the forms randomly to see the music changes.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -294,8 +294,19 @@ export default class Battle {
           if (pokemon.species.speciesId === Species.TAPU_KOKO || pokemon.species.speciesId === Species.TAPU_LELE || pokemon.species.speciesId === Species.TAPU_BULU || pokemon.species.speciesId === Species.TAPU_FINI) {
             return "battle_legendary_tapu";
           }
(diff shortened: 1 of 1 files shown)


Tempo-anon (approved):
> There isn't a way to force the form index of an opponent at present

We don't have form index in overrides? I guess an alternative would be to force a trainer encounter who uses a specific form

PR #3061, opened 2024-07-18
[BUG] ALT Fixing move Chilly Reception

DustinLin:
## What are the changes?

An alternative solution/implementation of PR #2979 , probably  better? this abstracts the problem to a separate class, which may be of use for future move implementations.

See PR #2979 

### Screenshots/Videos

See PR #2979 , results are exactly the same

## Checklist
- [ ] There is no overlap with another PR? - There is overlap with PR #2979
- [x] Are the changes visual?


mcmontag:
Maybe I wasn't super clear in the other PR, sorry!  I think this is still overcomplicating the move. I don't know if it's worth generalizing everything (e.g. having `statusCondFunc` and `statusApplyFunc`, naming it `StatusForceSwitchOutAttr`, etc.) when Chilly Reception is literally the only thing that needs it.
```
export class ChillyReceptionAttr extends ForceSwitchOutAttr {
  apply(...args) {
    trySetWeather(...);
    return super.apply(...args);
  }
}
```

I definitely see what you're going for, but there's really nothing wrong with having a one-off attribute IMO.  If another use case comes along it'd be very easy to generalize it then.

mcmontag:
FWIW this is just my perspective, certainly not trying to claim that your approach isn't valid 🙂

PR #3062, opened 2024-07-18
[Optimization] Use inline assets for images/atlases

flx-sta:
## What are the changes?

UI Assets that are always loaded inside `LoadingScene` are now loaded as so called `inline-assets`.

[Static Asset Handling | Vite](https://vitejs.dev/guide/assets) 

Using inline assets increases the build-size but improves performance. Since these files are always loaded anyway they can be part of the build-file

## What did change?

I've added a new `loadImageAsset()` and `loadAtlasAsset()` method to load these inline-assets. Then I created some `ts` files that import images & jsons and export them as js object which is then looped through to load the files into phaser.

#### **With** inline-assets
image

#### **Without** inline assets
image

## How to test the changes?

1. Build the project => `npm run build`
2. Start in `preview` mode => `npm run preview`
3. Open in browser with dev-tools in `Network` tab.
    - Check the amount of requests sent vs. on live/main branch
    - To compare loading times you have to run both locally to have equal chances 
5. Check that everything is still running properly.

## Checklist


flx-sta:
Not as much of an improvement as I was hoping for it to make
memory-comparison
<sub>top: `beta` branch.</sub>

PR #3064, opened 2024-07-18
[QoL] Improve readability in MovePhase

mcmontag:
## What are the changes?
Nothing functional, at least not intentionally.  Added documentation and cleaned up `MovePhase`.

It's extremely difficult to follow in its current state.  Adding documentation and cleaning up the code here (and soon `MoveEffectPhase`) will pay off in the short term, and will ultimately make it easier to game-plan a larger refactor.

## What did change?
Large sections of code were split out into methods and variables were renamed for readability; quite a few comments were added in places where I felt they'd be helpful.  A few things were shuffled around, but only in non-functional ways (e.g. moving a variable declaration closer to where it ends up being used).

### Screenshots/Videos
n/a

## How to test the changes?
If you find a section of shuffled code that seems like it may affect a move, test it.

All tests are passing, without any need to update them.

## Checklist
- [ ] Are the changes visual?


src/phases.ts
```diff
@@ -2565,6 +2565,10 @@ export class MovePhase extends BattlePhase {
   protected failed: boolean;
   protected cancelled: boolean;
 
+  /**
+   * @param followUp Indicates that the move being uses is a "follow-up" - for example, a move being used by Metronome or Dancer.
+   *                 Follow-ups bypass a few failure conditions, including flinches, sleep/paralysis/freeze checks, and
+   */
   constructor(scene: BattleScene, pokemon: Pokemon, targets: BattlerIndex[], move: PokemonMove, followUp?: boolean, ignorePp?: boolean) {
```
mcmontag, line 2572: `followUp` is the only param that isn't completely obvious, so it's the only one I called out.

flx-sta:
@mcmontag since this is a refactoring I have to ask you to make sure that tests are written to cover that the results are the same as before.

mcmontag:
@flx-sta makes sense. I'll have to take some time to think about how I want to go about getting full coverage.

I'll mark this as a draft for now!

DayKev:
> @mcmontag since this is a refactoring I have to ask you to make sure that tests are written to cover that the results are the same as before.

Tbh I think the existing tests should be sufficient; it's a refactor of `MovePhase` so if he breaks something then at least one of the existing move tests should break as well.

PR #2986, opened 2024-07-11
[Feature] Change Smeargle/Sketch to pre-gen9 implementation

ElizaAlex:
## What are the changes?
Changed Smeargle to relearn Sketch at levels 11, 21, 31, etc, up to level 91 so that it is in line with its learnset in previous generations, as opposed to the current gen 9 implementation. 

## Why am I doing these changes? 
Currently, Smeargle requires a high amount of investment in the form of Memory Mushrooms. This makes Smeargle difficult to use, particularly when obtained early on due to the increased rarity of said Mushrooms at lower levels. This change would make Smeargle more useful early on as a Starter, without impacting its strength or flexibility beyond level 91. 

## What did change?
Smeargle's learn-set has been updated to learn Sketch at levels 11, 21, 31, 41, 51, 61, 71, 81, and 91. 

Pokemon.getLevelMoves() has been updated to narrow the Species level moves by level first, and then filter out any duplicates after, as the other way around was causing any higher level Sketch learns to be missed. 

## How to test the changes?
Level a Smeargle up to level 11, and have it relearn Sketch. 
NOTE: You must have actually used sketch in order to relearn it. 

## Checklist
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [ ] Are the changes visual?
  - [ ] Have I provided screenshots/videos of the changes?

PR #3068, opened 2024-07-19
[Bug] Fix Pokémon unable to learn level up moves that are also relearner moves

DayKev:
## What are the changes?
Relearn-only moves no longer prevent Pokémon from learning moves on level up.

Some Pokémon could not learn some of their level up moves properly because the move was also in their moveset as a relearner move.

## What did change?
`includeStarterMoves` added to `getLevelMoves()` as a parameter (default `true` to not impact other uses of the function elsewhere). The call to `getLevelMoves()` upon level up was modified to pass this parameter as `false` (which causes the modified behavior of this PR). Also added `const STARTER_MOVE = 1` in `src/data/pokemon-level-moves.ts` as an indicator, based on the existence of `RELEARN_MOVE` and `EVOLVE_MOVE`.

### Screenshots/Videos
Old behavior:

https://github.com/user-attachments/assets/ecdcf867-4cd6-4fb0-987c-a79744995763

New behavior:

https://github.com/user-attachments/assets/51960f44-cffc-4475-9d2b-53166b2f621d

## How to test the changes?
Use a Pokémon with a relearner-move that's also in its regular level up move pool (such as Wormadam-Trash) and level it up such that it should learn the level up move.

## Checklist

## Additional Comments
It might be better to change all the relearner-moves from level 1 to -1 for Pokémon that can't be starters, but I'm not sure how you'd go about doing that automatically and I sure don't feel like going through 1000+ Pokémon manually to change them (and that would potentially risk some errors anyway).


Tempo-anon (approved):
the changes for pokemon-level-moves.ts is so large it doesn't render so I'll just trust it's good lol

Gonna also give a heads up to Ydarissep if we tweak that file

DayKev:
Oh I didn't realize #2986 existed, lemme check if it fixes this issue as well; no need to have 2 PRs doing the same thing after all.

DayKev:
Okay it does indeed also fix the issue, and it also includes an extra fix so I'll defer to that PR and close mine.

PR #3070, opened 2024-07-19
[Optimization] Destroy LoadingScene when done

flx-sta:
# #3215


src/test/sprites/pokemonSprite.test.ts
```diff
@@ -34,11 +35,11 @@ describe("check if every variant's sprite are correctly set", () => {
...
   });
 
-  function getMissingMasterlist(mlist, dirpath, excludes = []) {
-    const errors = [];
+  function getMissingMasterlist(mlist: any, dirpath: string, excludes: string[] = []): string[] {
+    const errors: string[] = [];
     const trimmedDirpath = `variant${path.sep}${dirpath.split(rootDir)[1]}`;
     if (fs.existsSync(dirpath)) {
-      const files = fs.readdirSync(dirpath);
+      const files = fs.readdirSync(dirpath).filter((filename) => !/^\..*/.test(filename));
```
flx-sta, line 42: This takes care of excluding hidden file like `.DS-store` from the test. Cause those might make the tests fail locally

src/loading-scene.ts
```diff
@@ -491,6 +496,17 @@ export class LoadingScene extends SceneBase {
   }
 
   async create() {
+    this.events.once(Phaser.Scenes.Events.DESTROY, () => this.handleDestroy());
     this.scene.start("battle");
   }
+
+  handleDestroy() {
+    console.debug(`Destroying ${LoadingScene.KEY} scene`);
```
DustinLin, line 504: want to make sure this is intentionally kept, it's prob good in case something breaks in `beta`, but wouldn't `LoadingScene.KEY` always be equal to "loading" ?

flx-sta: @DustinLin it will be. But the key is used at multiple locations and this way it can be controlled from a single locations to update all.

flx-sta:
closed in favor of #3215

PR #3071, opened 2024-07-19
[Bug] Fix Beak Blast

mcmontag:
## What are the changes?
Same-turn charge moves (e.g. Beak Blast) are currently being pushed to the end of the phase queue, meaning they'd end up occurring after `TurnEndPhase`.  The timeline looks like:
- `TurnStartPhase` starts
  -> unshifts all the command/`MovePhase`s
  -> pushes `WeatherEffectPhase`, `BerryPhase`, and `TurnEndPhase`
- TurnStartPhase ends
- Command/`MovePhase`s were unshifted, so they start.
  - Beak Blast (+ other same-turn charge moves) uses `pushMovePhase` to queue the `MovePhase` that resolves its attack
    - If a lower-priority move (< -3) was unshifted during `TurnStartPhase`, BB's phase will get spliced in before it
    - If not, it'll just get pushed to the end -- *behind* the phases that `TurnStartPhase` had pushed.

This PR updates `pushMovePhase` to accept a `forceSameTurn` param.  If `forceSameTurn` is true, the method will look for a lower priority move OR the first post-move phase pushed by `TurnStartPhase` (currently `WeatherEffectPhase`).

`ChargeAttr` will now forward its existing `sameTurn` param into `pushMovePhase`.

Beak Blast be broken as hell.  It hits after Protect has expired.  It hits after the weather.  It hits after berries are eaten.

## What did change?
See above.

### Screenshots/Videos

https://github.com/user-attachments/assets/720ea595-facc-442a-8eb0-c8d6c6275036

Notice how:
- Beak Blast is blocked by Protect, instead of hitting after it's expired.
- Fly (a two-turn charge move) still works like normal

## How to test the changes?
Try to replicate Beak Blast's current bugs

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1,6 +1,6 @@
 import Phaser from "phaser";
 import UI from "./ui/ui";
-import { NextEncounterPhase, NewBiomeEncounterPhase, SelectBiomePhase, MessagePhase, TurnInitPhase, ReturnPhase, LevelCapPhase, ShowTrainerPhase, LoginPhase, MovePhase, TitlePhase, SwitchPhase } from "./phases";
+import { NextEncounterPhase, NewBiomeEncounterPhase, SelectBiomePhase, MessagePhase, TurnInitPhase, ReturnPhase, LevelCapPhase, ShowTrainerPhase, LoginPhase, MovePhase, TitlePhase, SwitchPhase, WeatherEffectPhase } from "./phases";
 import Pokemon, { PlayerPokemon, EnemyPokemon } from "./field/pokemon";
 import PokemonSpecies, { PokemonSpeciesFilter, allSpecies, getPokemonSpecies } from "./data/pokemon-species";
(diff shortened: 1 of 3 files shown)


mcmontag:
cc @innerthunder because I know you've been thinking about Beak Blast

mcmontag:
@innerthunder I've addressed the priority issue -- I'm not sure about speed, but can look into that.  I disagree that multi-lens will be harder to fix; it's pretty damn broken as-is.  Like you said, this is more of a temporary fix until a rework can be completed. 

@DustinLin `pushMovePhase` is a function that's specifically used for `ChargeAttr` - that's the one this PR is correcting.  `unshiftPhase` would make the follow-up resolve immediately after the charge.  `standbyPhase` is internal state, and doesn't seem like something that should be manipulated directly.

mcmontag:
Closing since @innerthunder is eventually going to resolve this

PR #3075, opened 2024-07-19
[Bug] Fix Assist bug of using only the user's moveset instead of whole team's, refactor Assist/Sleep Talk/Metronome implementation

schmidtc1:
## What are the changes?

Properly implements Assist. 
Refactors RandomMovesetMoveAttr to extend RandomMoveAttr. 
Creates arrays of invalid moves for each move: Assist, Sleep Talk, Metronome, since they each have different sets of uncallable moves.

Assist did not call ally moves, had useless code, and was given no distinction in logic from Sleep Talk. 
Metronome calls random moves in a similar manner to the two other moves, so it makes sense to have Assist/Sleep Talk reuse the code for calling moves.
None of Assist/Sleep Talk/Metronome had properly implemented lists of uncallable moves.

## What did change?

- Assist now uses the entire team's party
- Assist/Sleep Talk's attribute now extends `RandomMoveAttr` and uses its callMoves function (which has been modified slightly to accommodate this) to call the move
- Sleep Talk targeting is changed to `NEAR_ENEMY`
- Assist/Sleep Talk/Metronome now have individual arrays listing uncallable moves. These arrays were created referencing Bulbapedia's list of uncallable moves.
- Uncallable moves for [Assist](https://bulbapedia.bulbagarden.net/wiki/Assist_(move)#:~:text=in%20a%20battle.-,Uncallable%20moves,-This%20move%20effect)
- Uncallable moves for [Sleep Talk](https://bulbapedia.bulbagarden.net/wiki/Sleep_Talk_(move)#:~:text=Z%2DPower%20effect.-,Unselectable%20moves,-This%20section%20is)
- Uncallable moves for [Metronome](https://bulbapedia.bulbagarden.net/wiki/Metronome_(move)#:~:text=Z%2DPower%20effect.-,Unselectable%20moves,-This%20move%20effect)

Note: Sleep Talk specifies that it targets the [opposite opponent](https://bulbapedia.bulbagarden.net/wiki/Sleep_Talk_(move)#:~:text=In%20Double%20and%20Triple%20Battles%2C%20because%20of%20the%20inability%20to%20select%20a%20target%2C%20the%20chosen%20move%20will%20always%20hit%20the%20opponent%20directly%20across%20from%20the%20user%2C%20if%20the%20move%20called%20normally%20hits%20only%20one%20Pok%C3%A9mon.). Since there is no `MoveTarget` specifically for opposite opponents, I chose to make it `NEAR_ENEMY` for now.

### Screenshots/Videos

Metronome

https://github.com/user-attachments/assets/c8861801-acd0-4ae0-b817-181cffa549d7

Assist

https://github.com/user-attachments/assets/9146fc45-9294-4d59-a726-ffcdf31b5777

Sleep Talk

https://github.com/user-attachments/assets/afd243b0-b2e9-4d84-9ba9-13a924c4b439

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -4765,7 +4765,10 @@ export class RandomMovesetMoveAttr extends OverrideMoveEffectAttr {
 
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
     const moveset = (!this.enemyMoveset ? user : target).getMoveset();
-    const moves = moveset.filter(m => !m.getMove().hasFlag(MoveFlags.IGNORE_VIRTUAL));
+    const allies = user.scene.getParty().filter(p => p !== user);
+    const partyMoveset = moveset.concat(allies.map(p => p.moveset).flat());
```
Tempo-anon, line 4769: Question for how assist works, but is there any deduplication going on or do multiple Pokemon with the same move include the chances of it being called?

schmidtc1: Per [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Assist_(move)#:~:text=If%20the%20same%20move%20is%20known%20by%20multiple%20Pok%C3%A9mon%2C%20each%20of%20them%20counts%20and%20therefore%20the%20move%20is%20more%20likely%20to%20be%20called.), 

> If the same move is known by multiple Pokémon, each of them counts and therefore the move is more likely to be called.

schmidtc1: Looking at this move, it seems like it was just given Sleep Talk's attribute with no second thought. I'm gonna try to  refactor this attribute to account for this.

schmidtc1:
Notes about this refactor:

- Sleep Talk specifies that it [hits the pokemon across from it](https://bulbapedia.bulbagarden.net/wiki/Sleep_Talk_(move)#:~:text=In%20Double%20and%20Triple%20Battles%2C%20because%20of%20the%20inability%20to%20select%20a%20target%2C%20the%20chosen%20move%20will%20always%20hit%20the%20opponent%20directly%20across%20from%20the%20user%2C%20if%20the%20move%20called%20normally%20hits%20only%20one%20Pok%C3%A9mon.), but I am not sure if this currently has a `MoveTarget` enum entry (the closest seems to be `NEAR_ENEMY`). 
- The `IGNORE_VIRTUAL` flag seems very inconsistent for which moves are unselectable
-  Assist, Sleep Talk, and Metronome each have their own very different lists of moves that are not selectable, but this is not currently reflected in their implementation.

schmidtc1:
> I would like feedback on the following:
> 
> * Assist/Sleep talk should be kept in RandomMovesetMoveAttr together?
> * Assist/Sleep talk should be split into two separate attributes (still extending RandomMoveAttr - Metronome's attribute)
> * Assist/Sleep talk should be merged into Metronome's attribute
> * Whether passing in separate invalid-move arrays for assist/sleep talk into RandomMovesetMoveAttr is a bad idea

schmidtc1:
Misclicked the close with comment button

DayKev:
> I would like feedback on the following:
> 
> * If the `IGNORE_VIRTUAL` flag can be removed since it is only used for Assist/Sleep Talk/Metronome

Oh, the only things in the codebase that use `IGNORE_VIRTUAL` are these moves? Removing it is probably fine then, I'm not entirely sure what it was supposed to do (since it seems like it's trying to do multiple things at once but not actually doing any of them correctly lmao).

As for using the same MoveAttr for multiple moves here, do whatever makes it easier to read and maintain. If reusing the same attr for all of them makes the code better, do that; if it makes the code start becoming overly complex/convoluted, then go ahead and separate the attrs.

src/data/move.ts
```diff
@@ -5007,6 +4974,286 @@ export class RandomMoveAttr extends OverrideMoveEffectAttr {
...
+      return true;
+    };
+  }
+}
+
+const invalidMetronomeMoves: Moves[] = [
+  Moves.AFTER_YOU,
+  Moves.APPLE_ACID,
+  Moves.ARMOR_CANNON,
+  // Moves.ASSIST,
```
chaosgrimmon, line 5030: It should probably continue to invalidate Moves that were previously not valid, and are only left empty on Bulbapedia because that specific move isn't available in Gen 9.

DayKev: Yeah you should use the last known state of the move (ie: uncallable in Gen 7 but not added back in Gen8 or Gen9 -> still mark the move as uncallable).

Also Assist can't call moves that switch out the opposing Pokémon nor 2-turn moves, and Sleep Talk can't call 2-turn moves.

schmidtc1: Made a new PR for this due to beta issue: #3380. I'll correct any invalid moves that were left empty, and add invalid Max moves/Z moves to Metronome's list. 
Assist's invalid moves do already include switch out moves and 2-turn moves, and Sleep Talk's invalid moves include 2-turn moves.

PR #3078, opened 2024-07-20
Localization(pt): Updated move.ts

josericardo-fo:
## What are the changes?

Keys, double quotes and `’`

## Checklist
- [X] Are the changes visual?


Tempo-anon:
There's quite a lot of moves that weren't formatted correctly before... do you have some screenshots of after the fix?

josericardo-fo:
> There's quite a lot of moves that weren't formatted correctly before... do you have some screenshots of after the fix?

They were working as intended before. It doesn't change anything using the key with or without the double quotes. The only ones that weren't working were the ones with `'s`

PR #3082, opened 2024-07-21
[Move] Update HitTagAttr attributes

chaosgrimmon:
## What are the changes?
Particular moves had outdated effects, or had effects removed due to not being in the most recent games.
It adjusts Whirlwind affecting Fly, Steamroller affecting Minimize, and Malicious Moonsault affecting Minimize.

These cause the game to behave in a way that those familiar with canonical mechanics might not expect.

Whirlwind was only supposed to be able to hit Pokemon in the semi-invulnerable turn of Fly in Gen II, and so that behaviour is getting removed. In fact, it is currently also capable of hitting Pokemon in the semi-invulnerable turn of Bounce and Sky Drop, since those all share the FLYING BattlerTag.
image
Steamroller and Malicious Moonsault were effective against Minimized Pokemon as of the last time they were canonically implemented, and so that functionality is being restored.
image

## What did change?
Whirlwind lost the attribute allowing it to hit Pokemon in the FLYING turn of the two-turn moves Fly, Bounce, and Sky Drop.
Steamroller and Malicious Moonsault gained the attribute that bypasses accuracy checks on Minmized targets, as well as the attribute that does double damage against them.

## Checklist
- [ ] Are the changes visual?


flx-sta:
I know is still a draft but:

please make sure to write automated test for your changes. To cover expected outcomes

Issue #3048, opened 2024-07-17
[BUG] `Save & Quit` can be spammed if "delayed"


flx-sta:
**Describe the bug**

When opening the menu the `Save & Quit` option can be spammed. This creates a queue of POST requests towards the server. As soon as the first request is resolved you are brought back to the Title screen and have this "reload" animation. For every extra request you did it will throw you back and "reload" for every resolved request (obviously)

**To Reproduce**

1. Try it on the [beta](https://beta.pokerogue.net) (since it's notoriously slower)
2. Open Dev-Tools -> `Network`
3. Login
4. Start/Load Session
5. Play 1 wave
6. Open Menu
7. Spam `Save and Quit` and observe the Spam

**Expected behavior**

It should only send it once (and maybe show some loading animation)

**Screenshots / Videos**

https://github.com/user-attachments/assets/8f48e17e-9bf6-4b8d-a9da-36a56415bdd7

**Device**

n/a

**Additional context**

n/a

PR #3084, opened 2024-07-21
[BUG][UI/UX] Save & quit crash for spam

Vassiat:
## What are the changes?
That the page does not crash when spamming Save and exit

It hardly affects if the internet is not slow.

Resolves https://github.com/pagefaultgames/pokerogue/issues/3048 by @flx-sta

## What did change?
add a "loading" variable so that when the promise returns
"resolve" in the "saveall" function stopped loading.

## Other solutions/Alternatives

image

1. #### In the file ```menu-ui-handler.ts``` remove conditional if(currentcurrentBattle.turn > 1) {} and else {}, since it will always ask for confirmation and therefore take longer to spam

2. #### Disable inputs of device after Save & quit, and enable after promise resolved

## Before:
[Before](https://github.com/user-attachments/assets/f817bcac-b7ee-4275-896c-88f5ea5c7aec)

_only five spams, more and it crashes_

## After:
[After](https://github.com/user-attachments/assets/07c947b7-5789-4342-a3cc-625ed90e4302)

_infinite spams since the promise wait does not pass_

## How to test the changes?
Using database (from rogueserver) and "Slowed 3G" (Presets from Network) for simulation.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -3031,7 +3031,11 @@ export class PostTurnLootAbAttr extends PostTurnAbAttr {
     ) as BerryModifier | undefined;
 
     if (!berryModifier) {
-      pokemon.scene.addModifier(new BerryModifier(chosenBerry, pokemon.id, chosenBerryType, 1));
+      if (pokemon.isPlayer()) {
+        pokemon.scene.addModifier(new BerryModifier(chosenBerry, pokemon.id, chosenBerryType, 1));
+      } else {
+        pokemon.scene.addEnemyModifier(new BerryModifier(chosenBerry, pokemon.id, chosenBerryType, 1));
+      }
     } else if (berryModifier.stackCount < berryModifier.getMaxHeldItemCount(pokemon)) {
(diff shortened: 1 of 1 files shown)


flx-sta:
Please link the issue properly by using [keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue)

flx-sta:
@Vassiat if you want you can add solving the same issue for `Log Out`.

flx-sta (changes requested):
❌ Not the correct solution

`GameData` is a class that is also being used to save the data to local-storage with. It shouldn't contain a `loading` property (as it would be saved to the local-storage accidentally which could soft-lock a players account).
It also shouldn't contain any UI code, which loading is.

Instead the same/a similar logic should be moved into the UI classes instead. `saveAll()` is returning it's promise anyway thus you can listen outside of the method for it.

PR #3092, opened 2024-07-22
[Documentation] Document `EnemyCommandPhase` and Enemy AI functions

innerthunder:
## What are the changes?

This adds a new doc `enemy-ai.md` which includes details on the algorithms used to determine the enemy AI's actions within the `EnemyCommandPhase`. This also includes comments for the relevant functions within `field/pokemon.ts` and `phases.ts`, and fixes an issue in `Pokemon.getMatchupScore` where Speed comparisons were having the opposite of the intended effect on matchup scores.

In its current implementation, the Enemy AI decision-making algorithm is fragmented and difficult to understand. Because of this, it can be difficult for new contributors to get a complete frame of reference for the benefit score functions when adding new move attributes. The new documentation should help make the AI's functionality easier to understand and analyze.

## What did change?

-`docs/enemy-ai.md`: New document covering the enemy AI's decision-making algorithms and formulas
-`phases.ts` and `field/pokemon.ts`: Added comments to enemy AI functions.
-`field/pokemon.ts`: Fixed a bug in `Pokemon.getMatchupScore()` where the `outspeed` logic had the opposite effect of what was intended.

### Screenshots/Videos

N/A

## How to test the changes?

N/A

## Checklist
- [n/a] Have I tested the changes (manually)?
    - [n/a] Are all unit tests still passing? (`npm run test`)

### DIFF
--- a/docs/enemy-ai.md
+++ b/docs/enemy-ai.md
@@ -21,6 +21,15 @@ where
 - $\text{defScore}$ is the inverse of the opposing Pokémon's $\text{atkScore}$ against the source Pokémon's defensive typing, or $(\prod_{\text{types}} \text{typeEffectiveness}(\text{type}, \text{sourcePokemon}))^{-1}$. Unlike $\text{atkScore}$, $\text{defScore}$ is capped at a maximum score of 4.
 - $\text{hpDiffRatio}= \text{sourceHpRatio}-\text{oppHpRatio}+1$. This is further multiplied by 1.5 if the source Pokémon has a higher Speed stat than the opposing Pokémon; however, $\text{hpDiffRatio}$ cannot be higher than 1.
 
+The maximum possible matchup score a Pokémon could have against a single opponent is $(16+16)\times 2=64$, which occurs when
+- the Pokémon hits its opponent for 4x super effective damage with both of its types.
+- the Pokémon is immune to or resists both of the opponent's types by 4x.
+- the Pokémon is at max HP while the opponent's HP ratio is near zero.
+
+In most situations, though, a Pokémon's matchup score against an opponent will be at most 16, which is equivalent to having two super effective types and resisting both of the opponent's types with the same HP ratios as before.
+
+The minimum possible matchup score a Pokémon could have against a single opponent is near zero, which occurs when the Pokémon's HP ratio is near zero while the opponent is at max HP. However, a Pokémon's matchup score can also be very low when its type(s) are 4x weak to and/or resisted by its opponent's types.
+
 ### Determining Switches in EnemyCommandPhase
 
 The `EnemyCommandPhase` follows this process to determine whether or not an enemy Pokémon should switch on each turn during a Trainer battle.
--- a/docs/enemy-ai.md
+++ b/docs/enemy-ai.md
@@ -61,9 +61,9 @@ In `getNextMove()`, the enemy Pokémon chooses a move to use in the following st
 
 ### Calculating Move and Target Scores
 
-As part of the move selection process, the enemy Pokémon must compute a **target score (TS)** for each legal target for each move in its move pool. The base target score for all moves is a combination of the move's **user benefit score (UBS)** and **target benefit score (TBS)**, and is multiplied by -1 if the target is the user's ally.
+As part of the move selection process, the enemy Pokémon must compute a **target score (TS)** for each legal target for each move in its move pool. The base target score for all moves is a combination of the move's **user benefit score (UBS)** and **target benefit score (TBS)**.
 
-$$\text{TS(move, target)}=(\text{UBS}+\text{TBS}) \times \begin{cases} -1 && \text{if target is an ally}\\ 1 && \text{otherwise} \end{cases}$$
+$$\text{TS(move, target)}=\text{UBS}+\text{TBS} \times \begin{cases} -1 && \text{if target is an opponent}\\ 1 && \text{otherwise} \end{cases}$$
 
 A move's UBS and TBS are computed with the respective functions in the `Move` class:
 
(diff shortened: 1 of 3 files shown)


src/phases.ts
```diff
@@ -2184,6 +2193,15 @@ export class EnemyCommandPhase extends FieldPhase {
 
     const trainer = battle.trainer;
 
+    /**
+     * If the enemy has a trainer, decide whether or not the enemy should switch
+     * to another member in its party.
+     *
+     * This block compares the active enemy Pokemon's {@linkcode Pokemon.getMatchupScore | matchup score}
+     * against the active player Pokemon with the enemy party's other non-fainted Pokemon. If a party
+     * member's matchup score is 3x the active enemy's score (or 2x for "boss" trainers),
```
Tempo-anon, line 2202: Do we have a definition for boss trainers? Is it just gym leaders or E4 and champion?

src/field/pokemon.ts
```diff
@@ -1210,18 +1210,37 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     return multiplier;
   }
 
+  /**
+   * Computes the given Pokemon's matchup score against this Pokemon
+   * @param pokemon The Pokemon to compare this Pokemon against
+   * @returns A score value based on how favorable this Pokemon is when fighting the given Pokemon
```
Tempo-anon, line 1216: Is there a known range the return value should be within?

src/field/pokemon.ts
```diff
@@ -1210,18 +1210,37 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   */
   getMatchupScore(pokemon: Pokemon): number {
     const types = this.getTypes(true);
     const enemyTypes = pokemon.getTypes(true, true);
-    const outspeed = (this.isActive(true) ? this.getBattleStat(Stat.SPD, pokemon) : this.getStat(Stat.SPD)) <= pokemon.getBattleStat(Stat.SPD, this);
+    /** Is this Pokemon faster than the given Pokemon? */
```
Tempo-anon, line 1221: nit: `pokemon` here is ambiguous. I believe `pokemon` refers to the opposing Pokemon in this instance so the naming scheme should reflec that

src/field/pokemon.ts
```diff
@@ -1210,18 +1210,37 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    const outspeed = (this.isActive(true) ? this.getBattleStat(Stat.SPD, pokemon) : this.getStat(Stat.SPD)) >= pokemon.getBattleStat(Stat.SPD, this);
```
Tempo-anon, line 1222: I notice that `<=` was switched to `>=`. Was the outspeed function wrong this whole time?

docs/enemy-ai.md
```diff
@@ -0,0 +1,137 @@
...
+   $\text{typeMult}=\begin{cases} 
+     2  && \text{if move is super effective (or better)}\\
+     -2 && \text{otherwise}
+   \end{cases}$
+2. Compute a multiplier based on the move's category and the user's offensive stats:
+   1. Compute 
+      $\text{statRatio}=\begin{cases}
+        \frac{\text{userSpAtk}}{\text{userAtk}} && \text{if move is physical}\\
+        \frac{\text{userAtk}}{\text{userSpAtk}} && \text{otherwise}
+      \end{cases}$
```
Tempo-anon, line 96: I don't think these are rendering properly in the markdown

src/field/pokemon.ts
innerthunder: Seems like it. `outspeed` is used in two places during the matchup score calculation:
1. When calculating `atkScore`

   ```ts
   let atkScore = pokemon.getAttackTypeEffectiveness(types[0], this) * (outspeed ? 1.25 : 1);
   ```
2. As a bonus multiplier when calculating `hpDiffRatio`

   ```ts
   let hpDiffRatio = this.getHpRatio() + (1 - pokemon.getHpRatio());
   if (outspeed) {
     hpDiffRatio = Math.min(hpDiffRatio * 1.5, 1);
   }
   ```

With `outspeed` using `<=` instead of `>=`, it was rewarding the source Pokemon for being slower than its opponent, rather than faster.

(Side note: I also think the line `hpDiffRatio = Math.min(hpDiffRatio * 1.5, 1);` could be causing problems. Before that line `hpDiffRatio` is set within the interval (0, 2), so `outspeed` capping it at 1 doesn't make much sense to me)

docs/enemy-ai.md
innerthunder: How are you checking it? They were rendering just fine for me in VSCode's preview, but I couldn't find a PDF converter that rendered them.

src/phases.ts
innerthunder: The full list of "boss" trainers atm is
- All gym leaders, Elite 4 members and Champions
- All Evil Team leaders
- The last 3 Rival fights (i.e. on waves 95, 145, and 195)

src/field/pokemon.ts
innerthunder: changing to `opponent` in the next commit

src/phases.ts
innerthunder: Just added this list to the matchup score section on the main doc

src/field/pokemon.ts
innerthunder: Added some info on the range of return values in the main doc

docs/enemy-ai.md
Tempo-anon: You can check by just selecting the md file in github and choosing "view file" https://github.com/pagefaultgames/pokerogue/blob/2c13ca237525f700c375348de9169a6ac3b98b8a/docs/enemy-ai.md

src/field/pokemon.ts
```diff
@@ -1210,19 +1210,39 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     return multiplier;
   }
 
-  getMatchupScore(pokemon: Pokemon): number {
+  /**
+   * Computes the given Pokemon's matchup score against this Pokemon.
+   * In most cases, this score ranges from near-zero to 16, but the maximum possible matchup score is 64.
+   * @param opponent The Pokemon to compare this Pokemon against
```
DustinLin, line 1216: would be nice to add the types, eg: `@param {@linkcode Pokemon} - the Pokemon to comapre this Pokemon against`

docs/enemy-ai.md
innerthunder: I changed all the equations with "cases" into images (generated from [this equation editor](https://editor.codecogs.com/)). Not the prettiest fix, but at least the equations are readable now.

image

flx-sta (approved):
lgtm

Issue #2040, opened 2024-06-10
[BUG] Lucky Egg and Golden Egg stack multiplicatively instead of additively


DayKev:
**Describe the bug**
The EXP bonus from Lucky Egg items will stack multiplicatively with the EXP bonus from Golden Egg items. This means for example, a Pokémon with 2 Golden Eggs (+200%) and 5 Lucky Eggs (+200%) will gain 300% * 300% (900%) the base EXP received instead of 500% the base EXP received.

**To Reproduce**
Fight a lv5 Gyarados with a lv5 Mewtwo, and defeat it with Thunderbolt.

Base exp gain: 190 exp
Expected exp gain with 1 Lucky Egg + 1 Golden Egg: 456 exp (190 * 2.4)
Bugged exp gain with 1 Lucky Egg + 1 Golden Egg:   532 exp (190 * 2.8)

**Expected behavior**
Golden Eggs and Lucky Eggs would additively boost EXP gain when used together.

**Screenshots / Videos**
Edit: Forgot to save the video and the link expired, oops.

**Device**
Observed on both Firefox and Chrome

**Additional context**
https://discord.com/channels/1125469663833370665/1125894949020381285/1249670003112935476

PR #3094, opened 2024-07-22
[Bug] Make EXP boosting items stack additively and not multiplicatively

DayKev:
## What are the changes?
EXP boosting items are changed to stack additively instead of multiplicatively.

#2040

## What did change?
The modifier increments a `modifierBonusExp` variable instead of acting on the gained EXP variable directly, and then after all the bonus modifiers are added to the bonus multiplier variable it's applied to the gained EXP.

## How to test the changes?
`npm run test exp_booster`
Alternatively, defeat a lv5 Gyarados with a lv5 Mewtwo that's holding 1 Lucky Egg and 1 Golden Egg and observe the gained EXP being 456 (2.4x) and not 532 (2.8x) as it is now on main.

## Checklist

### DIFF
--- a/src/modifier/modifier.ts
+++ b/src/modifier/modifier.ts
@@ -1774,7 +1774,7 @@ export class PokemonExpBoosterModifier extends PokemonHeldItemModifier {
   }
 
   apply(args: any[]): boolean {
-    (args[1] as Utils.NumberHolder).value = Math.floor((args[1] as Utils.NumberHolder).value * (1 + (this.getStackCount() * this.boostMultiplier)));
+    (args[1] as Utils.NumberHolder).value += (this.getStackCount() * this.boostMultiplier);
 
     return true;
   }
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -3927,7 +3927,9 @@ export class VictoryPhase extends PokemonPhase {
           expMultiplier = Overrides.XP_MULTIPLIER_OVERRIDE;
         }
         const pokemonExp = new Utils.NumberHolder(expValue * expMultiplier);
-        this.scene.applyModifiers(PokemonExpBoosterModifier, true, partyMember, pokemonExp);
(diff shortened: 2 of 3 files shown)


Tempo-anon:
Gonna wait to see if this is a bug or a feature

Tempo-anon:
Let's test this in beta

star-krieg:
Changing how EXP boosting items work would great impact the level progression in Endless. Compared here is the same save across the live and beta versions:

50 exp charms (25%), 17 super exp charms (60%), 14 golden egg (100%), 85 lucky eggs (40%)

Live version: 
image
Beta version:
image

EXP in live goes from 3 billion EXP and drops to 300 million EXP in beta for the same loadout.

Difference in levels become noticeable even just from doing a few waves.
image

Lowered EXP because of this change would probably even more compounded for a longer run (I've only checked for difference for a few waves)
Live:
image
Beta:
image

damocleas:
I'm going to list my thoughts in the order I had them
- This needs to be tested
- IF this is a bug, it is a pretty major one (likely is, but no one has noticed/didn't affect much having it, removing it affects more)
- Testing current runs just for a few waves won't work as well with how it's designed

I'm going to try to get decently far into an endless run with this on and see if it affects anything major, average classic run shouldn't have affected, but I'd rather be safe than sorry and test that as well since this would affect both modes and you're given exp charms automatically.

DayKev:
If it turns out to be a problem in endless then waiting to revisit the issue until the endless rework seems fine.

> I'm going to try to get decently far into an endless run with this on and see if it affects anything major, average classic run shouldn't have affected, but I'd rather be safe than sorry and test that as well since this would affect both modes and you're given exp charms automatically.

It should only change the behavior of specifically the interaction of stacking Lucky Eggs and Golden Eggs together on the same pokemon, the EXP Charm items use a different modifier class that I didn't touch.

PR #3100, opened 2024-07-23
[Bug] Fix Thousand Arrows not hitting targets under the effects of Magnet Rise

innerthunder:
## What are the changes?

This fixes a bug where Thousand Arrows cannot hit targets who are under the effects of Magnet Rise.

This bug was reported [in the Discord](https://discord.com/channels/1125469663833370665/1258239366682247180)

## What did change?

- `field/pokemon`: `Pokemon.getAttackTypeEffectiveness` now supports specific moves as arguments (in addition to just the move's type). If a move is passed, and the move's target has a battler tag that grants it immunity to the move's type, the move is checked for a `HitsTagAttr` which has a matching tag type.
- `data/move`: Added a `HitsTagAttr` to Thousand Arrows to allow it to hit Pokemon with the `MAGNET_RISEN` tag.
- `test/moves/thousand_arrows.test`: Added a test for Thousand Arrows hitting into a Pokemon with Magnet Rise.

### Screenshots/Videos

N/A

## How to test the changes?

`npm run test thousand_arrows`

## Checklist

### DIFF
--- a/src/test/moves/thousand_arrows.test.ts
+++ b/src/test/moves/thousand_arrows.test.ts
@@ -43,13 +43,7 @@ describe("Moves - Thousand Arrows", () => {
     async () => {
       await game.startBattle([ Species.ILLUMISE ]);
 
-      const leadPokemon = game.scene.getPlayerPokemon();
-      expect(leadPokemon).toBeDefined();
-
       const enemyPokemon = game.scene.getEnemyPokemon();
-      expect(enemyPokemon).toBeDefined();
-
-      const enemyStartingHp = enemyPokemon.hp;
 
       game.doAttack(getMovePosition(game.scene, 0, Moves.THOUSAND_ARROWS));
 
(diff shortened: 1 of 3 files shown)


src/test/moves/thousand_arrows.test.ts
```diff
@@ -86,10 +86,35 @@ describe("Moves - Thousand Arrows", () => {
...
+
+  it(
+    "move should hit and ground targets under the effects of Magnet Rise",
+    async () => {
+      vi.spyOn(overrides, "OPP_SPECIES_OVERRIDE", "get").mockReturnValue(Species.SNORLAX);
+
+      await game.startBattle([ Species.ILLUMISE ]);
+
+      const leadPokemon = game.scene.getPlayerPokemon();
+      expect(leadPokemon).toBeDefined();
```
DayKev, line 104: Imo don't put `expect()`s for things like this that you're not testing for. That should always be defined unless the codebase and/or testing utils are fundamentally broken somehow, and if they are this isn't the test that should be checking it anyway.

PR #3131, opened 2024-07-25
[Move] Implement Flame Burst

torranx:
## What are the changes?
- implement flame burst

- unimplemented move

## What did change?
- new class `FlameBurstAttr`

## Checklist


src/data/move.ts
```diff
@@ -1429,6 +1429,34 @@ export class PartyStatusCureAttr extends MoveEffectAttr {
...
+   * @returns A boolean indicating whether the effect was successfully applied.
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean | Promise<boolean> {
+    const targetAlly = target.getAlly();
+
+    if (!targetAlly) {
+      return false;
+    }
+
+    targetAlly.damageAndUpdate(Math.floor(1/16 * targetAlly.getMaxHp()), HitResult.OTHER);
```
schmidtc1, line 1451: Very niche case, but how does this work against Innards Out.
E.g. Enemy 2 has Innards Out, Flame Burst on Enemy 1 -> Splash damage on Enemy 2 to faint it. Should Innards Out trigger on the Flame Burst user?

torranx: I don't think Innards Out would take effect from effect damage- bulba also says:
> ...knocked out by an opponent's attack.
The damage is not directly from an opponent's attack.

src/data/move.ts
```diff
@@ -1429,6 +1429,40 @@ export class PartyStatusCureAttr extends MoveEffectAttr {
...
+    if (cancelled.value || !targetAlly) {
+      return false;
+    }
+
+    targetAlly.damageAndUpdate(Math.max(1, Math.floor(1/16 * targetAlly.getMaxHp())), HitResult.OTHER);
+    return true;
+  }
+
+  getTargetBenefitScore(user: Pokemon, target: Pokemon, move: Move): integer {
+    return target.getAlly() ? -5 : 0;
```
Tempo-anon, line 1462: It's -5 if the target has an ally but 0 if they don't?

torranx: ikr this is quite confusing: (`getTargetBenefitScore`) negative target scores = more incentive to use it against the enemy

discussed here: https://discord.com/channels/1125469663833370665/1176874654015684739/1265796251044155587

Tempo-anon: That should probably be documented somewhere because that's sort of confusing

Tempo-anon (approved):

PR #3135, opened 2024-07-25
[BUG] - Parting shot debuff applies even if move fails

### LINKED ISSUE #3112: [Bug] Prankster Parting Shot used on Dark types shows "it doesn't affect [target]" but still shows stats lowering
### Describe the bug

As it says in the title. Saw this when my egg move Cottonee was switched into by a trainer with a Poochyena. The switch from Parting Shot doesn't happen, but the stat drops appear to.

### User data export file

_No response_

### Session export file

_No response_

### Expected behavior

All moves that activate Prankster should completely fail when used on Dark types.

### Screenshots / Videos

_No response_

PC/Windows

### If other please specify

_No response_

### Which browser do you use?

Firefox

### If other please specify

_No response_

### Additional context

Occurred on locally hosted instance but latest version pulled from Git.

DustinLin:
## What are the changes?

When parting shot fails, the debuffs no longer apply

Closes #3112 

## What did change?

Added check in move workflow that doesn't apply a move's attributes when a move fails

### Screenshots/Videos

- `beta`

https://github.com/user-attachments/assets/d4922730-e09f-4654-9d3e-7ab213b9a73f

https://github.com/user-attachments/assets/837e7ae6-c58f-402c-a2aa-0ed25b18578e

- `pr`

https://github.com/user-attachments/assets/05395f98-8365-47b5-81a3-a699db4c8d63

https://github.com/user-attachments/assets/5dc9342b-2747-47b3-a9a6-305e7e6d4518

## How to test the changes?

Using overrides, the 2 cases were:
- user of parting shot has ability Prankster and the opponent is a dark type
- opponent has the ability Good as Gold

## Checklist
- [x] Are the changes visual?


DustinLin:
https://github.com/pagefaultgames/pokerogue/blob/28c5dc6b59cd363c1dff55550bcaa620e7e29dda/src/phases.ts#L2945C14-L2950C44

While the PR above was a simple fix, i'm wondering if there are pokemon cases where a move "fails" but some attribute of the move still goes through. The current fix just adds a clause that the `PRE_APPLY` attributes don't apply if the move has no effect. But I'm wondering if adding another clause for `FAIL` would be good as well

fishsicles:
> https://github.com/pagefaultgames/pokerogue/blob/28c5dc6b59cd363c1dff55550bcaa620e7e29dda/src/phases.ts#L2945C14-L2950C44
> 
> While the PR above was a simple fix, i'm wondering if there are pokemon cases where a move "fails" but some attribute of the move still goes through. The current fix just adds a clause that the `PRE_APPLY` attributes don't apply if the move has no effect. But I'm wondering if adding another clause for `FAIL` would be good as well

The main thing I can think of is Thrash and the Thrash-likes, which will still confuse the user if the last turn hits into Protect or the like. Order Up used to work like this as well, but that was patched out of mainline.

DustinLin:
closing b/c `beta` got messed up, reopened at Pr #3195

PR #3145, opened 2024-07-25
[Bug] Caught Pokemon Summary-Modifier Display Fix

frutescens:
## What are the changes?
Modifiers were not displaying on the caught pokemon's summary because of how the game stores and determines possession of modifiers. summary-ui-handler.ts now uses the pokemon's ownership boolean (pokemon.battleInfo.player) to determine where to look for modifiers. 

As a player, I would like to see the modifiers of my caught pokemon + my PR, my responsibility 

## What did change?
- Temporary Pokemon object has tempPokemon.battleInfo.player set to false in phases.ts
- summary-ui-handler.ts uses pokemon.battleInfo.player as an argument in findModifiers()

### Screenshots/Videos

https://github.com/user-attachments/assets/e12af32d-a687-4d91-954b-2f7bdc1f642f

## How to test the changes?
Catch a wild Pokemon with modifiers with a full party and press the Summary Option

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/ui/summary-ui-handler.ts
+++ b/src/ui/summary-ui-handler.ts
@@ -93,6 +93,7 @@ export default class SummaryUiHandler extends UiHandler {
   private moveCursorBlinkTimer: Phaser.Time.TimerEvent;
 
   private pokemon: PlayerPokemon;
+  private playerParty: boolean;
   private newMove: Move;
   private moveSelectFunction: Function;
   private transitioning: boolean;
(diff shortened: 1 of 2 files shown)


src/ui/summary-ui-handler.ts
```diff
@@ -93,6 +93,7 @@ export default class SummaryUiHandler extends UiHandler {
   private moveCursorBlinkTimer: Phaser.Time.TimerEvent;
 
   private pokemon: PlayerPokemon;
+  private playerParty: boolean;
```
Tempo-anon, line 96: Can you add a comment for this? Something like `// This is false when checking the summary of a freshly caught Pokemon as it is not part of a player's party yet but still needs to display its items`

src/ui/summary-ui-handler.ts
```diff
@@ -272,7 +273,7 @@ export default class SummaryUiHandler extends UiHandler {
 
     this.pokemon = args[0] as PlayerPokemon;
     this.summaryUiMode = args.length > 1 ? args[1] as SummaryUiMode : SummaryUiMode.DEFAULT;
-
+    this.playerParty = args[4] ?? true;
```
Tempo-anon, line 276: Can we get some documentation on the args so we know what args 0, 1, and 4 are?

PR #3149, opened 2024-07-25
[Tests] Overrides helper utility

flx-sta:
> ⚠️ It might look like a lot of changes **BUT** it's not that many. Most of it adding the `overridesHelper.ts` file and the rest is just replacing all the `vi.spyOn(Overrides, […]` with `game.override.[…]`

special thanks to @ben-lear to help me refine this feature/class

## What are the changes?

I've added a util that makes using overrides in tests easier.

To hopefully make writing tests just a. tad easier

## What did change?

I've added a `OverridesHelper` class and attached an instance to the `GameManager` instance (game.override).

Now instead of calling `vi.spyOn(Overrides, "...", "get").mockReturnValue(..)` you can just write e.g. 
```ts
game.override.startingWave(11);
```
You can also chain those
```ts
game.override.startingWave(11).moveset([Moves.SPLASH]);
```

## How to test the changes?

Run `npm run test:silent`

## Checklist
- ~~[] Have I considered writing automated tests for the issue?~~


ben-lear (commented):
having written numerous new test suites with this, it's a very helpful utility for writing tests. Lgtm.

src/test/utils/overridesHelper.ts
```diff
@@ -0,0 +1,271 @@
...
+  /**
+   * Override the enemy (pokemon) held items
+   * @param items the items to hold
+   * @returns this
+   */
+  enemyHeldItems(items: ModifierOverride[]) {
+    vi.spyOn(Overrides, "OPP_HELD_ITEMS_OVERRIDE", "get").mockReturnValue(items);
+    this.log("Enemy Pokemon held items set to:", items);
+    return this;
+  }
```
torranx, line 266: is it possible that by default, the enemy does not have held items on all tests?

flx-sta: That would be in a next stage. For now I just wanted to introduce that feature as it touches a chunk of files.

Tempo-anon (approved):
94 files changed huh. Shouldn't have any gameplay impact. We should get this in to make unit tests easier.

flx-sta:
Closed for the sake of #3212

PR #3167, opened 2024-07-27
[Bug] Self effect moves fix for Shield Dust

NxKarim:
Moves like Meteor Mash had their self effects disabled by Shield Dust.

## What are the changes?
Added a condition to the application of IgnoreMoveEffectsAbAttr.

[Bug report at Discord](https://discord.com/channels/1125469663833370665/1265299187378028586)

## What did change?
Moves with self effects no longer affected by Shield Dust.

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -974,7 +974,7 @@ export class MoveEffectAttr extends MoveAttr {
     const moveChance = new Utils.NumberHolder(move.chance);
     applyAbAttrs(MoveEffectChanceMultiplierAbAttr, user, null, moveChance, move, target, selfEffect, showAbility);
     if (!selfEffect) {
-      applyPreDefendAbAttrs(IgnoreMoveEffectsAbAttr,target,user,null,null, moveChance);
+      applyPreDefendAbAttrs(IgnoreMoveEffectsAbAttr, target, user, null, null, moveChance);
     }
     return moveChance.value;
   }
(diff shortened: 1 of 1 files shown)


src/data/move.ts
```diff
@@ -973,7 +973,9 @@ export class MoveEffectAttr extends MoveAttr {
   getMoveChance(user: Pokemon, target: Pokemon, move: Move, selfEffect?: Boolean, showAbility?: Boolean): integer {
     const moveChance = new Utils.NumberHolder(move.chance);
     applyAbAttrs(MoveEffectChanceMultiplierAbAttr, user, null, moveChance, move, target, selfEffect, showAbility);
-    applyPreDefendAbAttrs(IgnoreMoveEffectsAbAttr,target,user,null,null, moveChance);
+    if (!selfEffect) {
+      applyPreDefendAbAttrs(IgnoreMoveEffectsAbAttr,target,user,null,null, moveChance);
```
torranx, line 977:
```suggestion
      applyPreDefendAbAttrs(IgnoreMoveEffectsAbAttr, target, user, null, null, moveChance);
```

Snailman11:
https://github.com/pagefaultgames/pokerogue/issues/3347 (Copycat on Flame Charge does not give speed boost if enemy is attacked) 
Where a Wooloo used Flame Charge on a Dustox, and Shield Dust is a possible ability.

PR #3179, opened 2024-07-28
[Hotfix] Use `getPokemonSpecies()` instead of `allSpecies[]`

DayKev:
Fix save bug again, finally, for real this time. Alolan(/etc) Pokémon should work now.


Tempo-anon (approved):
Let's remove all this excess code used to hotfix an issue in the next beta->main merge. We shouldn't be responsible for started runs that are over 2 weeks old

PR #3182, opened 2024-07-28
[Bug] [Fix] Metal Burst always hitting left target.

NxKarim:
## What are the changes?
Changed the condition of Metal Burst redirection when the target faints before the move can hit during a double battle.

Bug report.

## What did change?
Fixed Metal Burst always hitting left target, and the related freeze/crash.

## How to test the changes?
During a double battle make the last opponent to hit the Metal Burst user faint before it, move should redirect to the other non fainted enemy.

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -2665,7 +2665,7 @@ export class MovePhase extends BattlePhase {
     if (this.targets.length === 1 && this.targets[0] === BattlerIndex.ATTACKER) {
       if (this.pokemon.turnData.attacksReceived.length) {
         const attack = this.pokemon.turnData.attacksReceived[0];
-        this.targets[0] = attack.attackingPosition;
+        this.targets[0] = attack.sourceBattlerIndex;
 
         // account for metal burst and comeuppance hitting remaining targets in double battles
         // counterattack will redirect to remaining ally if original attacker faints
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -2670,8 +2670,8 @@ export class MovePhase extends BattlePhase {
         // account for metal burst and comeuppance hitting remaining targets in double battles
         // counterattack will redirect to remaining ally if original attacker faints
         if (this.scene.currentBattle.double && this.move.getMove().hasFlag(MoveFlags.REDIRECT_COUNTER)) {
-          if (!this.scene.getEnemyField()[this.targets[0]]) {
-            this.targets[0] = this.scene.getEnemyField().find(p => p.isActive(true)).getBattlerIndex();
+          if (this.scene.getField()[this.targets[0]].hp === 0) {
+            this.targets[0] = this.scene.getEnemyField().find(p => p.hp > 0)?.getBattlerIndex();
           }
         }
       }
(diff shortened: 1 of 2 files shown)


DayKev:
Tested with both opposing Pokémon using Self-Destruct and then my Pokémon using Metal Burst after them and there was no crash. So 👍 I guess!

src/phases.ts
```diff
@@ -2670,8 +2670,8 @@ export class MovePhase extends BattlePhase {
         // account for metal burst and comeuppance hitting remaining targets in double battles
         // counterattack will redirect to remaining ally if original attacker faints
         if (this.scene.currentBattle.double && this.move.getMove().hasFlag(MoveFlags.REDIRECT_COUNTER)) {
-          if (!this.scene.getEnemyField()[this.targets[0]]) {
-            this.targets[0] = this.scene.getEnemyField().find(p => p.isActive(true)).getBattlerIndex();
+          if (this.scene.getField()[this.targets[0]].hp === 0) {
+            this.targets[0] = this.scene.getEnemyField().find(p => p.hp > 0)?.getBattlerIndex();
```
innerthunder, line 2674:
```suggestion
            const opposingField = pokemon.isPlayer() ? this.scene.getEnemyField() : this.scene.getPlayerField()
            this.targets[0] = opposingField.find(p => p.hp > 0)?.getBattlerIndex();
```
If an enemy used Metal Burst, and the last attacker fainted, the enemy's Metal Burst would be redirected to their ally without this change

schmidtc1: Metal Burst is supposed to [redirect to the ally if the last attacker faints](https://bulbapedia.bulbagarden.net/wiki/Metal_Burst_(move)#:~:text=However%2C%20if%20the%20last%20Pok%C3%A9mon%20to%20hit%20the%20user%20faints%20before%20the%20user%20moves%2C%20the%20remaining%20enemy%20Pok%C3%A9mon%20will%20be%20hit%20instead%2C%20like%20with%20a%20regular%20single%2Dtarget%20damaging%20move%20(using%20the%20damage%20taken%20from%20the%20move%20of%20the%20now%2Dfainted%20Pok%C3%A9mon%20that%20had%20hit%20the%20user).)

src/phases.ts
```diff
@@ -2670,8 +2670,9 @@ export class MovePhase extends BattlePhase {
         // account for metal burst and comeuppance hitting remaining targets in double battles
```
schmidtc1, line 2670: If possible, could you rename `attackingPosition` to `sourceBattlerIndex`

The variable also appears here: https://github.com/pagefaultgames/pokerogue/blob/f25d4e1ee2ec0a04d1631fd1e1bcb74a90295e2a/src/field/pokemon.ts#L4021

And here:
https://github.com/pagefaultgames/pokerogue/blob/f25d4e1ee2ec0a04d1631fd1e1bcb74a90295e2a/src/field/pokemon.ts#L2084

(I thought I had this in my original PR for counter moves, but it must've gotten reverted.)

schmidtc1 (changes requested):

schmidtc1:
Unit tests and/or video evidence would be nice to confirm the moves working properly in a couple cases:

- enemy uses uturn on Metal Burster
- enemy uses self-destruct on Metal Burster 
- reverse cases: enemy uses metal burst, player uses uturn, etc. 
- any other simple cases: metal burst hits the proper attacker, metal burst hits the most recent attacker when attacked by both foes, etc.

PR #3198, opened 2024-07-29
[Bug] Fix move Chilly Reception

DustinLin:
## What are the changes?

Fixing the currently implemented move Chilly Reception. Prior to these fixes the move basically didn't work at all.

Noticed from the bug-reports channel on discord

## What did change?

Given Chilly Reception's unique make up of being a STATUS move, a SWITCHOUT move, and a WEATHER effecting move, many special circumstances had to made in the code for it to work.

For example, when checking the ForceSwitchOutAttr, the logic specifically checks that the move isn't a STATUS move. I'm guessing this is so that attacking moves such as U-turn still go through despite not having a switchout target, but this breaks Chilly Reception since it is a STATUS move with no attack.

According the [bulba](https://bulbapedia.bulbagarden.net/wiki/Chilly_Reception_(move)), the move still allows a user to switch out even if the weather changing portion of the move fails. This again requires a special circumstance, as normally weather changing moves auto-fail in this circumstance

Currently the move has 2 attributes, the ForceSwitchOut and WeatherChange. If either of these fail then the whole move fails. The change creates a new attributes that checks both to ensure the move still goes through despite one of them failing

### Screenshots/Videos

On `beta`, basically nothing works (game freezes on switch out, move fails when used solo)

https://github.com/user-attachments/assets/3baff715-b9f6-4582-acfd-425dcac5c124

https://github.com/user-attachments/assets/d1f5e6ec-e67d-4ba5-9962-f01a6eea208f

On the PR these issues are fixed: when used in doubles, it still allows switching even if the weather doesn't change. In single battles, it just changes the weather (and can fail if both the weather and switch out fails). Also added dialogue as per the move description

https://github.com/user-attachments/assets/425e5dcd-db9b-4fe4-bd94-57856ba6974d

https://github.com/user-attachments/assets/022fc0ac-7d4e-46cd-86fa-924648e684a4

edit: adding demos of enemy trainer and wild pokemon using the move: works as expected:

- trainer battle
https://github.com/user-attachments/assets/a8d950d7-efba-491e-bf60-e75e5d847893

- wild battle
https://github.com/user-attachments/assets/37e0663d-ae42-4907-a7ba-71bbff447083

## How to test the changes?

Try for single and double battles via overrides with Chilly Reception

## Checklist
- [x] Are the changes visual?
- [ ] Localization
  - [x] de
  - [x] en
  - [ ] es
  - [x] fr
  - [x] it
  - [x] ko
  - [x] pt_BR
  - [x] zh_CN
  - [x] zh_TW
  - [x] ja


src/data/move.ts
```diff
@@ -4789,6 +4790,40 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
   }
 }
 
+export class StatusForceSwitchOutAttr extends ForceSwitchOutAttr {
```
mcmontag, line 4793: Not to sound like a broken record, but there are a lot of comments in this class trying to justify a super niche behavior that's ultimately just for one move.  The fact that it's limited to `||`ing the conditions with fixed precedence makes it even less likely that a hypothetical move would be able to reuse this.

There are other moves that have their own attributes.  Why shouldn't this one?

mcmontag: Well let's back up a bit.  This is essentially a mixin.  An attribute that can be injected with additional `apply` and `condition` callbacks is effectively *two* attributes.  What justifies doing that, but only when one of the two is `ForceSwitchOutAttr`?

If mixins are the goal, there is actual language support for them.

But with that said -- all you're trying to do is continue using the move if the weather effect fails, right?  Why does it necessitate an entire new attribute in the first place?

DustinLin: yeah my bad I didn't really explain in the original solution, but basically the way the Moves work is that (when going through the checks to see if the move will fail, before any actual hit/accuracy/move damage is calculated) we loop over each `attr` attached to the move, and call it's `getCondition()`. If all the `attr`s satisfy then the we go on and do the rest of the turn.

The original bug/problem with the move is that the switch-out and weather-apply attrs were separated, meaning that if one of them fails then the entire move will fail. Based on how the move works this is problematic. The fix in PR #2979 (which you can view [here](https://github.com/pagefaultgames/pokerogue/pull/3198/commits/1ae0d3380291871df9ff2da3b0fe18f4bb0fb4aa) ) is to add special checks *separately* in each of the 2 attributes to ensure that it doesn't fail for chilly reception. imo this is messy, and intertwines the 2 classes.

Another solution (which this PR does) is to just make a new attribute that combines both the weather change and force switchout, and has a separate `getCondition()` logic.  Now how we want to design this attr class is still up for discussion ig, but that's the reasoning.

Maybe there is some other cleaner solution i haven't thought of, it would be good to think about

mcmontag: Right.  So the justification is that `MoveEffectAttr`s are typically "all or nothing" -- they all happen, or none of them happen.  This particular move has an effect that _happens_ to fit the description of two existing attributes, but differs in the fact that those two parts of the effect can occur independently of each other.  I get why you might want to create an abstraction for that kind of behavior, but I'm not seeing how it isn't just unnecessary complexity when it only applies to a single move.  Moreover, building the abstraction specifically for `ForceSwitchOutAttr` seems pointless.

> This class extends ForceSwitchOutAttr, we can think of this class as ForceSwitchOut with "something extra" in it's apply() and getConditions() functions

...this just restates one of the core purposes of inheritence.  If some new move comes along that needs to do "something extra" on top of `ForceSwitchOut`, subclassing `ForceSwitchOut` instead of `StatusForceSwitchOut` would take virtually the same amount of code, wouldn't duplicate any meaningful logic at all, and would be much easier to read and reason about.

If you're absolutely sure you want to abstract this, I'd suggest a few things.  For one, the name `StatusForceSwitchOutAttr` makes it seem like status moves which force a switch out would be able (or even *expected*) to leverage it, and that just isn't the case.  On top of that:
- It assumes that `statusApplyFunc` checks `statusCondFunc` internally (`trySetWeather` does that, but it's certainly not safe to assume that all hypothetical `statusApplyFunc`s would)
- The switch-out always occurs after the additional effect -- along the same lines, the return value of `apply` is always the result of the switch-out `apply`.
- Both `statusApplyFunc` and `statusCondFunc` are marked as nullable (also, they're missing types...), but neither is null-checked before being called (`this.statusApplyFunc(...)` instead of `this.statusApplyFunc?.(...)`) -- this isn't a problem because the only move that uses it _happens_ to provide both args.
- `statusCondFunc` has to be a `MoveConditionFunc` -- if there's a concrete instance of `MoveCondition`, why shouldn't it be valid to pass in as the condition?

flx-sta:
Please copy the description from https://github.com/pagefaultgames/pokerogue/pull/3061 into here with a note to the original PR

flx-sta:
Didn't mean to close it

src/data/move.ts
DustinLin: hmm so from a design perspective you would rather the new attribute be explicitly made for chilly reception, with the extra conditions "hardcoded" (not in a bad way) into the attribute's `getCondition()` and `apply()` functions?

I can see your reasoning, and wouldn't mind making the change. But perhaps we can get more opinions about this.

Snailman11:
Link to Discord's bug report (Chilly reception freezes the game)
https://discord.com/channels/1125469663833370665/1256731008078712912

src/locales/fr/move-trigger.ts
```diff
@@ -59,4 +59,5 @@ export const moveTriggers: SimpleTranslationEntries = {
   "copyType": "{{pokemonName}} prend le type\nde {{targetPokemonName}} !",
   "suppressAbilities": "Le talent de {{pokemonName}}\na été rendu inactif !",
   "swapArenaTags": "Les effets affectant chaque côté du terrain\nont été échangés par {{pokemonName}} !",
+  "chillyReception": "{{pokemonName}} is preparing to tell a chillingly bad joke!",
```
Adri1, line 62:
```suggestion
  "chillyReception": "{{pokemonName}} s’apprête\nà faire un mauvais jeu de mots…",
```

src/locales/de/move-trigger.ts
```diff
@@ -59,4 +59,5 @@ export const moveTriggers: SimpleTranslationEntries = {
   "copyType": "{{pokemonName}} hat den Typ von {{targetPokemonName}} angenommen!",
   "suppressAbilities": "Die Fähigkeit von {{pokemonName}} wirkt nicht mehr!",
   "swapArenaTags": "{{pokemonName}} hat die Effekte, die auf den beiden Seiten des Kampffeldes wirken, miteinander getauscht!",
+  "chillyReception": "{{pokemonName}} is preparing to tell a chillingly bad joke!",
```
CodeTappert, line 62:
```suggestion
  "chillyReception": "{{pokemonName}} erzählt einen schlechten Witz,
der nicht besonders gut ankommt...",
```

src/locales/pt_BR/move-trigger.ts
```diff
@@ -59,4 +59,5 @@ export const moveTriggers: SimpleTranslationEntries = {
   "copyType": "O tipo de {{pokemonName}}\nmudou para combinar com {{targetPokemonName}}!",
   "suppressAbilities": "A habilidade de {{pokemonName}}\nfoi suprimida!",
   "swapArenaTags": "{{pokemonName}} trocou os efeitos de batalha que afetam cada lado do campo!",
+  "chillyReception": "{{pokemonName}} is preparing to tell a chillingly bad joke!",
```
josericardo-fo, line 62:
```suggestion
  "chillyReception": "{{pokemonName}} está prestes a contar uma piada gelada!",
```

src/locales/zh_CN/move-trigger.ts
```diff
@@ -59,4 +59,5 @@ export const moveTriggers: SimpleTranslationEntries = {
   "copyType": "{{pokemonName}}\n变成了{{targetPokemonName}}的属性！",
   "suppressAbilities": "{{pokemonName}}的特性\n变得无效了！",
   "swapArenaTags": "{{pokemonName}}\n交换了双方的场地效果！",
+  "chillyReception": "{{pokemonName}} is preparing to tell a chillingly bad joke!",
```
mercurius-00, line 62:
```suggestion
  "chillyReception": "{{pokemonName}}\n说出了冷笑话！",
```

src/locales/zh_TW/move-trigger.ts
```diff
@@ -59,4 +59,5 @@ export const moveTriggers: SimpleTranslationEntries = {
   "copyType": "{{pokemonName}}變成了{{targetPokemonName}}的屬性！",
   "suppressAbilities": "{{pokemonName}}的特性\n變得無效了！",
   "swapArenaTags": "{{pokemonName}}\n交換了雙方的場地效果！",
+  "chillyReception": "{{pokemonName}} is preparing to tell a chillingly bad joke!",
```
mercurius-00, line 62:
```suggestion
  "chillyReception": "{{pokemonName}}\n說了冷笑話！",
```

mercurius-00 (commented):
zh

src/locales/ko/move-trigger.ts
```diff
@@ -59,4 +59,5 @@ export const moveTriggers: SimpleTranslationEntries = {
   "copyType": "{{pokemonName}}[[는]]\n{{targetPokemonName}}[[와]] 같은 타입이 되었다!",
   "suppressAbilities": "{{pokemonName}}의\n특성이 효과를 발휘하지 못하게 되었다!",
   "swapArenaTags": "{{pokemonName}}[[는]]\n서로의 필드 효과를 교체했다!",
+  "chillyReception": "{{pokemonName}} is preparing to tell a chillingly bad joke!",
```
returntoice, line 62:
```suggestion
  "chillyReception": "{{pokemonName}}[[는]] 썰렁한 개그를 선보였다!",
```

EnochG1:
@DustinLin 
Could you merge latest beta and add Japanese locale files? 🙏

DustinLin:
done, added localization tracker too, seems to be missing spanish, italian, and now japanese

src/test/moves/chilly_reception.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    game.phaseInterceptor.restoreOg();
+  });
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override.battleType("single");
+    game.override.moveset([Moves.CHILLY_RECEPTION, Moves.SPLASH, Moves.SNOWSCAPE]);
+    game.override.enemyMoveset(SPLASH_ONLY);
+    game.override.startingLevel(5);
+    game.override.enemyLevel(5);
```
flx-sta, line 33: you can chain those if you want to
```suggestion
    game.override.battleType("single")
      .moveset([Moves.CHILLY_RECEPTION, Moves.SPLASH, Moves.SNOWSCAPE])
      .enemyMoveset(SPLASH_ONLY)
      .startingLevel(5)
      .enemyLevel(5);
```

NicusPulcis (commented):
italian localization (also added another translation for something that still wasn't localized)

src/data/move.ts
DayKev: I think I ran into an issue with move `Attr`s being "all-or-nothing" before with a different move, so maybe that's what needs to be changed (so it's in-line with how abilities can have multiple `AbAttr`s with differing conditions).

DustinLin: yeah that's similar to the problem this move faces, is there a particular `Attr` or move that you remember?

DayKev: Unfortunately I don't remember which one specifically.

src/locales/ja/move-trigger.ts
```diff
@@ -59,5 +59,6 @@ export const moveTriggers: SimpleTranslationEntries = {
   "copyType": "{{pokemonName}}は　{{targetPokemonName}}と\n同じタイプに　なった！",
   "suppressAbilities": "{{pokemonName}}の　とくせいが　きかなくなった！",
   "swapArenaTags": "{{pokemonName}}は\nおたがいの　ばのこうかを　いれかえた！",
+  "chillyReception": "{{pokemonName}} is preparing to tell a chillingly bad joke!",
```
protimita, line 62:
```suggestion
  "chillyReception": "{{pokemonName}}は\nさむい　ギャグを　かました！",
```

src/data/move.ts
DustinLin: i looked around `move.ts` and couldn't find anything... I can either leave this implementation as is, or refactor it to it's own `ChillyRecepAttr`

Snailman11:
Discord Bug report Post (Chilly Reception bug(Beta))
https://discord.com/channels/1125469663833370665/1280455029744209950

-Please take a look at this - wild pokemon using Chilly Reception and the game opening up a second reality. I noticed that you didn't include videos of enemy pokemon using the move so I'd like to know if you're aware of this issue.

src/data/move.ts
DayKev: As much as I hate single-use `Attr`s/etc, the current system kinda forces it (for now). Ditch the genericization and have the new Attr just "hardcode" the Chilly Reception extras.
The "better" solution would be to make it so move attrs can apply independently if specified and then otherwise keep the original move implementation, but I don't know how much work that would be.

DayKev:
Whoops, should have marked it as draft before updating it.

src/locales/ja/move-trigger.json
```diff
@@ -60,5 +60,6 @@
   "copyType": "{{pokemonName}}は　{{targetPokemonName}}と\n同じタイプに　なった！",
   "suppressAbilities": "{{pokemonName}}の　とくせいが　きかなくなった！",
   "revivalBlessing": "{{pokemonName}}は\n復活して　戦えるようになった！",
-  "swapArenaTags": "{{pokemonName}}は\nおたがいの　ばのこうかを　いれかえた！"
+  "swapArenaTags": "{{pokemonName}}は\nおたがいの　ばのこうかを　いれかえた！",
+  "chillyReception": "{{pokemonName}}は\nさむい　ギャグを　かました！"
```
Chapybara-jp, line 64:
```suggestion
  "chillyReception": "{{pokemonName}}は\n寒い　ギャグを　かました！"
```

src/locales/ja/move-trigger.json
```diff
@@ -60,5 +60,6 @@
   "copyType": "{{pokemonName}}は　{{targetPokemonName}}と\n同じタイプに　なった！",
   "suppressAbilities": "{{pokemonName}}の　とくせいが　きかなくなった！",
   "revivalBlessing": "{{pokemonName}}は\n復活して　戦えるようになった！",
-  "swapArenaTags": "{{pokemonName}}は\nおたがいの　ばのこうかを　いれかえた！"
+  "swapArenaTags": "{{pokemonName}}は\nおたがいの　ばのこうかを　いれかえた！",
```
Chapybara-jp, line 63:
```suggestion
  "swapArenaTags": "{{pokemonName}}は\nお互いの　場の効果を　入れ替えた！",
```

Chapybara-jp (approved):

NicusPulcis (dismissed):
italian is approved

src/data/move.ts
DustinLin: thanks all, refactored in most recent change

src/test/moves/chilly_reception.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    expect(game.scene.arena.weather?.weatherType).toBe(WeatherType.SNOW);
+
+    await game.phaseInterceptor.to("TurnInitPhase", false);
+    game.move.select(Moves.CHILLY_RECEPTION);
+    game.doSelectPartyPokemon(1);
+
+    await game.phaseInterceptor.to("BerryPhase", false);
+    expect(game.scene.arena.weather?.weatherType).toBe(WeatherType.SNOW);
+    expect(game.scene.getPlayerField()[0].species.speciesId).toBe(Species.MEOWTH);
+  }, TIMEOUT);
```
ben-lear, line 59: nit: can we have a happy path test case for setting weather + switch out?

ben-lear (changes requested):

src/test/moves/chilly_reception.test.ts
DustinLin: added

DayKev (approved):

DayKev:
Another move that needs the `conditionalAttr()` system that abilities have

Chapybara-jp (approved):

flx-sta (approved):
Technically the tests shouldn't have a manual `timeout` set anymore but there is some other that sneaked in so I'm going to fix that in a new PR

SangaraSorama (approved):

PR #3259, opened 2024-07-30
[Refactor] use typescript `strict-null`

flx-sta:
## Todo:

- [x] fix Egg gacha pulls

## What are the changes?

Activate TS strict-null

Better code quality which is less error-prone due to null/undefined being more transparent

## What did change?

Too much 💀 

## Checklist

-----

- 🚀  Thanks to @DayKev for helping


src/account.ts
```diff
@@ -31,10 +31,10 @@ export function updateUserInfo(): Promise<[boolean, integer]> {
       // Migrate old data from before the username was appended
       [ "data", "sessionData", "sessionData1", "sessionData2", "sessionData3", "sessionData4" ].map(d => {
         if (localStorage.hasOwnProperty(d)) {
```
ben-lear, line 33:
```suggestion
        if (localStorage.hasOwnProperty(d) && !!loggedInUser?.username) {
          if (localStorage.hasOwnProperty(`${d}_${loggedInUser.username}`)) {
            localStorage.setItem(`${d}_${loggedInUser.username}_bak`, localStorage.getItem(`${d}_${loggedInUser.username}`));
          }
          localStorage.setItem(`${d}_${loggedInUser.username}`, localStorage.getItem(d));
```

src/battle-scene.ts
```diff
@@ -700,17 +700,17 @@ export default class BattleScene extends SceneBase {
 
   hasExpSprite(key: string): boolean {
     const keyMatch = /^pkmn__?(back__)?(shiny__)?(female__)?(\d+)(\-.*?)?(?:_[1-3])?$/g.exec(key);
-    let k = keyMatch[4];
-    if (keyMatch[2]) {
+    let k = keyMatch?.[4]!; // TODO: is this bang correct?
```
ben-lear, line 703: could we not just do an early return if `keyMatch` is nullish here?

src/battle-scene.ts
```diff
@@ -784,10 +784,10 @@ export default class BattleScene extends SceneBase {
 
   getPokemonById(pokemonId: integer): Pokemon {
     const findInParty = (party: Pokemon[]) => party.find(p => p.id === pokemonId);
-    return findInParty(this.getParty()) || findInParty(this.getEnemyParty());
+    return findInParty(this.getParty()) || findInParty(this.getEnemyParty())!; // TODO: is this bang correct?
```
ben-lear, line 787: I think return result of this function should be changed to match that of `find()`, so `Pokemon | undefined`

src/battle-scene.ts
```diff
@@ -2046,27 +2047,37 @@ export default class BattleScene extends SceneBase {
...
     if (!this.phaseQueue.length) {
       this.populatePhaseQueue();
       // Clear the conditionalQueue if there are no phases left in the phaseQueue
       this.conditionalQueue = [];
     }
-    this.currentPhase = this.phaseQueue.shift();
+
+    const shiftedPhase = this.phaseQueue.shift();
+    if (shiftedPhase) {
+      this.currentPhase = shiftedPhase;
```
ben-lear, line 2064: should `currentPhase` be reassigned to `null` if `shiftedPhase` is nullish? Seems odd that there could be a path where an "old" phase stays assigned to `currentPhase`

src/data/ability.ts
```diff
@@ -1089,7 +1089,7 @@ export class MoveEffectChanceMultiplierAbAttr extends AbAttr {
       return false;
     }
 
-    (args[0] as Utils.NumberHolder).value *= this.chanceMultiplier;
+    (args[0] as Utils.NumberHolder).value *= this.chanceMultiplier!; // TODO: is this bang correct?
```
ben-lear, line 1092: if `chanceMultiplier` can be undefined, I don't think a null assertion will do anything at runtime when it's actually undefined. Think a check for nullish is required before multiplication, otherwise you'll get NaN

src/data/ability.ts
```diff
@@ -2206,7 +2206,7 @@ export class PostSummonUserFieldRemoveStatusEffectAbAttr extends PostSummonAbAtt
     }
 
     for (const pokemon of allowedParty) {
-      if (this.statusEffect.includes(pokemon.status?.effect)) {
+      if (pokemon.status && this.statusEffect.includes(pokemon.status.effect)) {
```
ben-lear, line 2209: not sure what this changes other than removing the shorthand?

src/data/ability.ts
```diff
@@ -2262,14 +2262,15 @@ export class PostSummonTransformAbAttr extends PostSummonAbAttr {
...
+    target = target!; // compiler doesn't know its guranteed to be defined
     pokemon.summonData.speciesForm = target.getSpeciesForm();
     pokemon.summonData.fusionSpeciesForm = target.getFusionSpeciesForm();
     pokemon.summonData.ability = target.getAbility().id;
     pokemon.summonData.gender = target.getGender();
     pokemon.summonData.fusionGender = target.getFusionGender();
     pokemon.summonData.stats = [ pokemon.stats[Stat.HP] ].concat(target.stats.slice(1));
     pokemon.summonData.battleStats = target.summonData.battleStats.slice(0);
-    pokemon.summonData.moveset = target.getMoveset().map(m => new PokemonMove(m.moveId, m.ppUsed, m.ppUp));
+    pokemon.summonData.moveset = target.getMoveset().map(m => new PokemonMove(m!.moveId, m!.ppUsed, m!.ppUp)); // TODO: are those bangs correct?
```
ben-lear, line 2273:
```suggestion
    pokemon.summonData.moveset = target.getMoveset().map(m => new PokemonMove(m?.moveId, m?.ppUsed, m?.ppUp));
```

I think in this case we should be checking for `null` based on `pokemon-data` moveset definition
```
public moveset: (PokemonMove | null)[];
```

src/system/pokemon-data.ts
```diff
@@ -33,8 +33,8 @@ export default class PokemonData {
   public ivs: integer[];
   public nature: Nature;
   public natureOverride: Nature | -1;
-  public moveset: PokemonMove[];
-  public status: Status;
+  public moveset: (PokemonMove | null)[];
```
ben-lear, line 36: should this be `PokemonMove[] | null;`? Just checking

src/data/ability.ts
```diff
@@ -2805,15 +2806,15 @@ function getAnticipationCondition(): AbAttrCondition {
     for (const opponent of pokemon.getOpponents()) {
       for (const move of opponent.moveset) {
         // move is super effective
-        if (move.getMove() instanceof AttackMove && pokemon.getAttackTypeEffectiveness(move.getMove().type, opponent, true) >= 2) {
+        if (move!.getMove() instanceof AttackMove && pokemon.getAttackTypeEffectiveness(move!.getMove().type, opponent, true) >= 2) { // TODO: is this bang correct?
```
ben-lear, line 2809: can move ever be null? Feel like we should change typedef of `moveset`. See earlier comment.

src/data/ability.ts
```diff
@@ -3051,7 +3052,7 @@ export class PostTurnStatusHealAbAttr extends PostTurnAbAttr {
    * @returns Returns true if healed from status, false if not
    */
   applyPostTurn(pokemon: Pokemon, passive: boolean, args: any[]): boolean | Promise<boolean> {
-    if (this.effects.includes(pokemon.status?.effect)) {
+    if (pokemon.status && this.effects.includes(pokemon.status.effect)) {
```
ben-lear, line 3055: do we want to remove the shorthand here or is compiler yelling at something that required change?

src/data/battler-tags.ts
```diff
@@ -580,7 +580,7 @@ export class HelpingHandTag extends BattlerTag {
   onAdd(pokemon: Pokemon): void {
     pokemon.scene.queueMessage(
       i18next.t("battle:battlerTagsHelpingHandOnAdd", {
-        pokemonNameWithAffix: getPokemonNameWithAffix(pokemon.scene.getPokemonById(this.sourceId)),
+        pokemonNameWithAffix: getPokemonNameWithAffix(pokemon.scene.getPokemonById(this.sourceId!)), // TODO: is that bang correct?
```
ben-lear, line 583: feel like if we're having to make this many null assertions the typedef is wrong

flx-sta: @ben-lear there is definitely something wrong with it but getting to the bottom of that would require much more work and that's not the goal of my PR here unfortunately.
I just want to create a baseline of being able to use strict-null so we can get things cleaned up in the long run

src/account.ts
flx-sta: this doesn't resolve the issue with `localStorage.getItem(d)` potentially returning `undefined`.

src/battle-scene.ts
flx-sta: I was thinking the same but I'm not experienced enough with what this code does to be confident of that change

src/battle-scene.ts
flx-sta: `Found 10 errors in 5 files.` After changing that.
Working on it. Hopefully it's not another rabbit hole...

flx-sta: it was a bit of a rabbit hole. But not too crazy

src/battle-scene.ts
flx-sta: I honestly don't know

src/data/ability.ts
flx-sta: `Array.includes()` does not allow `undefined` or `null` as a value to look for.

src/data/ability.ts
flx-sta: unfortunately yes.

```ts
  setMove(moveIndex: integer, moveId: Moves): void {
    const move = moveId ? new PokemonMove(moveId) : null;
    this.moveset[moveIndex] = move;
    if (this.summonData?.moveset) {
      this.summonData.moveset[moveIndex] = move;
    }
  }
  ```

File: src/data/ability.ts (hunk 5 above)
Comment (flx-sta) [reply src/data/ability.ts:1092]: I'm aware. But idk which default values I should use as fallback. I was hoping for clearing ups from ppl with knowledge

File: src/data/ability.ts (hunk 10 above)
Comment (flx-sta) [reply src/data/ability.ts:3055]: `Array.includes()` does not allow `undefined` or `null` as a param

File: src/data/ability.ts (hunk 7 above)
Comment (flx-sta) [reply src/data/ability.ts:2273]: PokemonMove() constructor does not allow `undefined` for the move-id (ironic, isn't it?)

File: src/system/pokemon-data.ts (hunk 8 above)
Comment (flx-sta) [reply src/system/pokemon-data.ts:36]: no this is the correct definition. It's an array.
see 
```ts
  setMove(moveIndex: integer, moveId: Moves): void {
    const move = moveId ? new PokemonMove(moveId) : null;
    this.moveset[moveIndex] = move;
    if (this.summonData?.moveset) {
      this.summonData.moveset[moveIndex] = move;
    }
  }
  ```

File: src/field/pokemon.ts [hunk 12]
```diff
@@ -2591,7 +2595,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     return this.gender !== Gender.GENDERLESS && pokemon.gender === (this.gender === Gender.MALE ? Gender.FEMALE : Gender.MALE);
   }
 
-  canSetStatus(effect: StatusEffect, quiet: boolean = false, overrideStatus: boolean = false, sourcePokemon: Pokemon = null): boolean {
+  canSetStatus(effect: StatusEffect | undefined, quiet: boolean = false, overrideStatus: boolean = false, sourcePokemon: Pokemon | null = null): boolean {
```
DayKev, line 2598: Note for the future: `trySetStatus()` crashes if passed `null | undefined`, add a check in `canSetStatus()`.

flx-sta: That doesn't seem to be an issue.
I've added a unit-test to support this claim

src/data/ability.ts
DayKev: I don't think `changeMultiplier` should ever be undefined, it probably shouldn't be marked as an optional parameter in the first place because it doesn't make sense; I think it was just a minor mistake in implementation. This AbAttr is for Serene Grace and Sheer Force for reference.

Made a PR to change it https://github.com/flx-sta/pokerogue/pull/8

src/ui/dropdown.ts
MokaStitcher, line 1: While working to improve the starter UI filtering I've started refactoring the `DropDownOption` class because the way it's been implemented is way too error prone. Basically this whole class expects it, the DropDown and the FilterBar to be are initialized in exactly the correct way in regard to one another for it all to work.

Some of the updated fields here like `offStateLabel` & co have already been completely removed on my end, so i wouldn't worry too much about whether the bangs on them are 100% correct for this PR

The biggest problem in this file is that I think the `text` field of `DropDownOption` should also be marked as optional since it's not always initialized in the constructor. And there are currently no safeguards in the rest of code in case it's undefined.

This should also get updated on my end though, so maybe this PR can just ignore all that for now? I've double checked that the changes made here and in `starter-select-ui` relating to all this don't break anything so to me it's good to go as is, knowing that the issues mentioned above will be fixed soon

flx-sta: Soo... could you put some comments to the specific sections of code you mean? Its unfortunately a lot of changes I'm dealing with so I don't remember every exact change I did

src/battle-scene.ts
flx-sta: agree and adjusted 2b3f8c7a3b2fe1f1046e77f35bc4d25e7d0bdd5a

src/battle-scene.ts
flx-sta: done

src/account.ts
flx-sta: Using this solution
```ts
        const lsItem = localStorage.getItem(d);
        if (lsItem && !!loggedInUser?.username) {
          const lsUserItem = localStorage.getItem(`${d}_${loggedInUser.username}`);
          if (lsUserItem) {  
            localStorage.setItem(`${d}_${loggedInUser.username}_bak`, lsUserItem);  
          }  
          localStorage.setItem(`${d}_${loggedInUser.username}`, lsItem);
          localStorage.removeItem(d);
        }
```

src/test/moves/baton_pass.test.ts
```diff
@@ -81,7 +81,7 @@ describe("Moves - Baton Pass", () => {
     // assert
     const playerPkm = game.scene.getPlayerPokemon()!;
     // check buffs are still there
-    expect(playerPkm.summonData.battleStats[BattleStat.SPATK]).toEqual(2);
+    expect(game.scene.getEnemyPokemon()!.summonData.battleStats[BattleStat.SPATK]).toEqual(2);
```
DayKev, line 84: I think you messed up the merge here or something, you removed the const that was just declared right above this but left the declaration.

Issue #2524, opened 2024-06-22
[BUG] Arena Trap dialog prompt doesn't disappear


EmberCM:
**Describe the bug**
After getting the "Arena Trap prevents switching" dialog, the prompt arrow stays when switching dialogs

**To Reproduce**
1. Have an opponent with Arena Trap
2. Attempt to Run without a special run ability
3. Click Fight after the dialog appears

**Expected behavior**
The dialog arrow disappears

**Screenshots / Videos**
![run](https://i.ember.zone/u9QF7GsC7UmI9AVK.png)
![fight](https://i.ember.zone/a3YhBzC4U4CBhsc4.png)
![back](https://i.ember.zone/2ZWKx0gGnwWJn1EJ.png)

**Device**
Firefox Browser on PC

**Additional context**
Dialog is created here: https://github.com/pagefaultgames/pokerogue/blob/main/src/data/ability.ts#L3841
Arena Trap trigger message: https://github.com/pagefaultgames/pokerogue/blob/main/src/data/ability.ts#L3289

innerthunder:
fixed by https://github.com/pagefaultgames/pokerogue/pull/3271

PR #3271, opened 2024-07-31
[Bug] Show arena trap/shadow tag dialog immediately w/o locking instead of as a buffered message

### LINKED ISSUE #3217: [Bug] Can't switch because of Arena Trap doesn't show on switch screen
### Describe the bug

I was in a trainer battle, and I tried to switch my Pokemone. I was on the switch screen and I kept selecting the monster and hitting Send Out. Nothing happened. When I stopped and continued with the fight, I got all the "Can't switch because of Arena Trap" messages in the battle box. So the logic worked correctly, but the message didn't show at the right time.

### User data export file

_No response_

### Session export file

_No response_

### Expected behavior

The "can't switch out" message should appear on the Pokemon screen, not the battle screen.

### Screenshots / Videos

_No response_

Mac/OSX

### If other please specify

_No response_

### Which browser do you use?

Chrome

### If other please specify

_No response_

### Additional context

_No response_

chansyu:
## What are the changes?
The user will receive dialog that their Pokemon cannot switch out/run as their enemy pokemon has a arena trap-related ability (Arena Trap, Magnet Pull, Shadow Tag). This feedback will now not soft-lock the user as well. 
In the past, the user would receive feedback in an inappropriate time or get soft-locked by these abilities.

Closes #3217, #2524

## What did change?

The condition-checking stayed the same, but the side effect was modified. Instead of calling queueMessage(), the code will immediately show a dialog in the CommandPhase. 

### Screenshots/Videos

Before (currently):
All arena traps don't show anything until the user selects a command
[magnet-pull.webm](https://github.com/user-attachments/assets/af971038-ba59-462b-b248-a64958fec670)
[shadow-tag.webm](https://github.com/user-attachments/assets/fe4af576-d7f0-4e14-ad1a-976f38884820)
[arena-trap.webm](https://github.com/user-attachments/assets/0827cbd9-a9c3-4b75-8c58-b49ed06ebdb7)

Before (a month ago / tested from commit e2507a2):
User gets soft-locked
[soft-lock.webm](https://github.com/user-attachments/assets/b411aba3-2d82-4063-b3cf-b1a41fa156a6)

Changes:
[magnet-pull-2.webm](https://github.com/user-attachments/assets/3d629e01-0d8a-440e-8ba2-f55c0b406ce0)
[shadow-tag-2.webm](https://github.com/user-attachments/assets/58ac9051-66b2-45f0-9ff2-c72519d79525)
[arena-trap-2.webm](https://github.com/user-attachments/assets/94c56fa4-e7d8-4059-bffa-84b2e32bdf3b)

if pokemon can escape successfully
[escape.webm](https://github.com/user-attachments/assets/5b667460-7430-4a1f-8dc2-7cc4cc4a90f1)

## How to test the changes?

Edit the src/overrides.ts file to test it
I did not introduce any automated tests, but not sure if it should be part of this PR or a new PR.

## Checklist
   - Yes, would like to have feedback if this is necessary.
- [X] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -4094,6 +4094,7 @@ async function applyAbAttrsInternal<TAttr extends AbAttr>(
   args: any[],
   showAbilityInstant: boolean = false,
   quiet: boolean = false,
+  messages: string[] = undefined,
 ) {
   for (const passive of [false, true]) {
     if (!pokemon.canApplyAbility(passive)) {
@@ -4131,10 +4132,13 @@ async function applyAbAttrsInternal<TAttr extends AbAttr>(
           }
         }
 
-        if (!quiet) {
-          const message = attr.getTriggerMessage(pokemon, ability.name, args);
-          if (message) {
+        const message = attr.getTriggerMessage(pokemon, ability.name, args);
+        if (message) {
+          if (!quiet) {
             pokemon.scene.queueMessage(message);
+          }
+          if (messages !== undefined) {
+            messages.push(message);
           }
         }
       }
@@ -4266,8 +4270,8 @@ export function applyPostTerrainChangeAbAttrs(attrType: Constructor<PostTerrainC
 }
 
 export function applyCheckTrappedAbAttrs(attrType: Constructor<CheckTrappedAbAttr>,
-  pokemon: Pokemon, trapped: Utils.BooleanHolder, otherPokemon: Pokemon, ...args: any[]): Promise<void> {
-  return applyAbAttrsInternal<CheckTrappedAbAttr>(attrType, pokemon, (attr, passive) => attr.applyCheckTrapped(pokemon, passive, trapped, otherPokemon, args), args);
+  pokemon: Pokemon, trapped: Utils.BooleanHolder, otherPokemon: Pokemon, quiet: boolean, messages: string[], ...args: any[]): Promise<void> {
+  return applyAbAttrsInternal<CheckTrappedAbAttr>(attrType, pokemon, (attr, passive) => attr.applyCheckTrapped(pokemon, passive, trapped, otherPokemon, args), args, false, quiet, messages);
 }
 
 export function applyPostBattleAbAttrs(attrType: Constructor<PostBattleAbAttr>,
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -2079,8 +2079,9 @@ export class CommandPhase extends FieldPhase {
         const trapTag = playerPokemon.findTag(t => t instanceof TrappedTag) as TrappedTag;
         const trapped = new Utils.BooleanHolder(false);
         const batonPass = isSwitch && args[0] as boolean;
+        const trappedAbMessages: string[] = [];
         if (!batonPass) {
-          enemyField.forEach(enemyPokemon => applyCheckTrappedAbAttrs(CheckTrappedAbAttr, enemyPokemon, trapped, playerPokemon));
+          enemyField.forEach(enemyPokemon => applyCheckTrappedAbAttrs(CheckTrappedAbAttr, enemyPokemon, trapped, playerPokemon, true, trappedAbMessages));
         }
         if (batonPass || (!trapTag && !trapped.value)) {
           this.scene.currentBattle.turnCommands[this.fieldIndex] = isSwitch
@@ -2115,6 +2116,16 @@ export class CommandPhase extends FieldPhase {
                 this.scene.ui.setMode(Mode.COMMAND, this.fieldIndex);
               }
             }, null, true);
+        } else if (trapped.value && trappedAbMessages.length > 0) {
+          if (!isSwitch) {
+            this.scene.ui.setMode(Mode.MESSAGE);
+          }
+          this.scene.ui.showText(trappedAbMessages[0], null, () => {
+            this.scene.ui.showText(null, 0);
+            if (!isSwitch) {
+              this.scene.ui.setMode(Mode.COMMAND, this.fieldIndex);
+            }
+          }, null, true);
         }
       }
       break;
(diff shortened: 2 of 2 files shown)


chansyu:
How do I show embedded videos :sob:

Snailman11:
Link to the Discord Bug Report Shadow Tag Pops Up Inappropriately https://discord.com/channels/1125469663833370665/1267175681398145054

src/data/ability.ts
```diff
@@ -4094,6 +4094,7 @@ async function applyAbAttrsInternal<TAttr extends AbAttr>(
   args: any[],
   showAbilityInstant: boolean = false,
   quiet: boolean = false,
+  messages: string[] = undefined,
```
flx-sta, line 4097: messages has to allow `undefined` as type too or should be default empty array `[]` (Id prefer the latter)

src/data/ability.ts
```diff
@@ -4131,11 +4132,14 @@ async function applyAbAttrsInternal<TAttr extends AbAttr>(
...
 
-        if (!quiet) {
-          const message = attr.getTriggerMessage(pokemon, ability.name, args);
-          if (message) {
+        const message = attr.getTriggerMessage(pokemon, ability.name, args);
+        if (message) {
+          if (!quiet) {
             pokemon.scene.queueMessage(message);
           }
+          if (messages !== undefined) {
```
flx-sta, line 4140: If you are using default empty array just check the length for not being 0 `messages.length !== 0`

flx-sta (changes requested):

flx-sta:
You should make sure to merge the most recent `beta` branch as there might be compiler errors due to `strict-null` adaption

chansyu:
I missed some null / undefined cases. My bad.

flx-sta:
> I missed some null / undefined cases. My bad.

missed a few

chansyu:
@flx-sta I have some problems fixing the first issue mentioned.

I need to use `null` for src/phases.ts/PartyUIHandler/showText (line 655)
```
showText(text: string, delay?: integer, callback?: Function, callbackDelay?: integer, prompt?: boolean, promptDelay?: integer) {
if (text === null) {
      text = defaultMessage;
}
```

There are multiple references of the method that use `null` as an argument.

flx-sta:
> @flx-sta I have some problems fixing the first issue mentioned.
> 
> I need to use `null` as an argument for `this.scene.ui.showText(null, 0);` for this method src/phases.ts/PartyUIHandler/showText (line 655)
> 
> ```
> showText(text: string, delay?: integer, callback?: Function, callbackDelay?: integer, prompt?: boolean, promptDelay?: integer) {
> if (text === null) {
>       text = defaultMessage;
> }
> ```
> 
> There are multiple references of the method that use `null` as an argument. It seems to be used before returning back for user input.

You have to actually adjust the function/method then to allow `null` as a type for `text: string`

so
```ts
showText(text: string | null, delay?: integer, callback?: Function, callbackDelay?: integer, prompt?: boolean, promptDelay?: integer) {
if (text === null) {
      text = defaultMessage;
}
```

chansyu:
merge conflict :sleeping:

flx-sta:
@chansyu all part of the process..


Can you also make sure to check any calls of the changed “showText()” method that uses `showText(“”, …)` instead of `showText(null, …)`

chansyu:
It seems that the changes from `showText(null, ...)` to `showText("", ...)` broke the traptags. It shows an empty message after an user tries to switch while the pokemon is trapped.

Visual aid:

[empty-message-from-ingrain](https://github.com/user-attachments/assets/299b92f4-0037-4470-9163-e396387823a8)


I made a commit that should fix this problem as well as my issue right now by checking if the message is zero-length.

flx-sta:
Yeah I told you to leave it at null after you showed me the details. That’s why u allowed null as a type there. But for some reasons you allowed null but changed it to “” 👀

src/data/ability.ts
```diff
@@ -4318,8 +4320,8 @@ export function applyPostTerrainChangeAbAttrs(attrType: Constructor<PostTerrainC
 }
 
 export function applyCheckTrappedAbAttrs(attrType: Constructor<CheckTrappedAbAttr>,
-  pokemon: Pokemon, trapped: Utils.BooleanHolder, otherPokemon: Pokemon, ...args: any[]): Promise<void> {
-  return applyAbAttrsInternal<CheckTrappedAbAttr>(attrType, pokemon, (attr, passive) => attr.applyCheckTrapped(pokemon, passive, trapped, otherPokemon, args), args);
+  pokemon: Pokemon, trapped: Utils.BooleanHolder, otherPokemon: Pokemon, quiet: boolean, messages: string[], ...args: any[]): Promise<void> {
```
frutescens, line 4323: Would maybe suggest `isQuiet` as a variable name. I will do some more testing, but it looks good so far

frutescens (changes requested):

frutescens:
Briefly tested on my machine + looks good to me. Please put in flx's changes + let me know what you think of changing the variable name to isQuiet

chansyu:
I realized that if I had to change `class UI method showText()` to accept null strings, I would need to change almost every `showText()` function call to accept null strings and also write new conditionals.
Also, the code that checks for trapped tags was transformed to `this.scene.ui.showText("", 0);` during the merge anyways.
I thought that the new commits I made was a simpler and better solution.

flx-sta:
> I realized that if I had to change `class UI method showText()` to accept null strings, I would need to change almost every similar `showText()` function to accept null strings and also write new conditionals. Also, the code that checks for trapped tags was transformed to `this.scene.ui.showText("", 0);` during the merge anyways. I thought that the new commits I made was a simpler and better solution.

I changed it to that cause I wasn’t aware of that null check in the method. Since the method body can expect a “null” I think it’s correct to allow a null type in that method. Obv. All of them have to be adjusted, yeah but that shouldn’t be too big a deal I think

flx-sta (approved):

f-fsantos (approved):

PR #2217, opened 2024-06-14
[Ability] Implement ability illusion 

PyGaVS:
## What are the changes?

Adding illusion

## Why am I doing these changes?

Make ability illusion works so we can play Zoroark and make fusion with illusion

## What did change?

Added PreSummonAbAttr class as ability attribute.
A pokemon with illusion create an illusion from the last pokemon of the trainer's party.
If we encounter a wild Zoroark he will use the illusion of Entei, Suicune or Raiku like in the original game. We can change that in the method generateIllusion() in src/field/pokemon.ts.
The class pokemon has a new attribute illusion. 
A shiny Zoroark will create a shiny illusion.
A non shiny Zoroark will create a non-shiny illusion.
This doesn't affect the luck of course, it's just visual.
A pokemon with illusion can use illusion one time per battle like in the original game.
After a battle, the pokemon can use illusion again.
The AI move choice will be affected by the illusion.

### Screenshots/Videos

https://discord.com/channels/1125469663833370665/1176874654015684739/1251234015760089159
https://discord.com/channels/1125469663833370665/1176874654015684739/1251234045204103238
https://discord.com/channels/1125469663833370665/1176874654015684739/1251234047104127039

## How to test the changes?

I played Zoroark and override the enemy pokemon with Zoroark so we can see him using illusion of Raikou Suicune Entei
I used illusion on a fusion and with a fusion
I tried to switch with the illusion

## Checklist
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #3273, opened 2024-07-31
[Ability] Implement Illusion

PyGaVS:
## What are the changes?

Adding illusion ability, works with fusion also

It's missing

## What did change?

-Added PreSummonAbAttr class as ability attribute.
-Illusion uses the appearance of the last conscious and available Pokémon in the party (ignoring itself), including :
Species, Form, Shininess, Pokeball, Gender
-If We encounter a wild Zorua/Zoroak he will use the illusion of a pokemon in the current biome. Depending on Zorua or Zoroark level the illusion will be evolved or not.
-If we encounter a boss Zoroark he will use the illusion of Entei, Suicune or Raiku like in the original game.
-The AI move choice will be affected by the illusion.
-The type will always be displayed as the Pokémon it is disguised as, even when affected by Soak etc.
-The illusion will disappear if the Pokemon has its ability changed or suppressed.
-Pokemon cannot transform into a Pokémon that is disguised by Illusion.
-Pokemon that are disguised by Illusion cannot Transform.
-Illusion will not activate if the last conscious Pokémon is a Terastallized Ogerpon or Terapagos.

### Screenshots/Videos

https://discord.com/channels/1125469663833370665/1176874654015684739/1257835032647831682
https://discord.com/channels/1125469663833370665/1176874654015684739/1257864381937619045
https://discord.com/channels/1125469663833370665/1176874654015684739/1257865512801538059
https://discord.com/channels/1125469663833370665/1176874654015684739/1258418548074745958

## How to test the changes?

I played Zoroark and override the enemy pokemon with Zoroark so we can see him using correct illusion
I used illusion on a fusion and with a fusion
I tried to switch with the illusion
We can also run `npm run test illusion`

Old conversations here : 
https://github.com/pagefaultgames/pokerogue/pull/2217

The link for the localization pr :
https://github.com/pagefaultgames/pokerogue-locales/pull/26

## Checklist
- [x] Are the changes visual?


PyGaVS:
I don't know why but Github is considering that there are changes in package-lock.json and items.json files but there is not.

src/field/pokemon.ts
```diff
@@ -835,34 +984,77 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    }
   }
 
   getLuck(): integer {
     return this.luck + (this.isFusion() ? this.fusionLuck : 0);
   }
 
-  isFusion(): boolean {
-    return !!this.fusionSpecies;
+  isFusion(useIllusion: boolean = false): boolean {
```
xsn34kzx, line 1048: Using this example for a larger idea: not necessarily a fan of attaching `useIllusion` to every function, feels a bit too specific; however, it would make sense for this to be `ignoreOverride` (meaning that Illusion and things like Transform are treated generically as "overrides" in name) like other functions (see `getFusionGender`, `getSpeciesForm`, etc.) that act similarly. 

As an aside though, I wonder if Illusion data and state could've been put in `summonData` instead of being directly attached to the `Pokemon` class. This isn't a necessary change, just a thought.

src/test/abilities/illusion.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    await game.phaseInterceptor.to(TurnEndPhase);
+
+    const zoroark = game.scene.getPlayerPokemon();
+    const zorua = game.scene.getEnemyPokemon();
+
+    expect(zorua.illusion.active).equals(false);
+    expect(zoroark.illusion.active).equals(false);
+  });
+
+  it("disappear if the ability is suppressed", async () => {
```
xsn34kzx, line 66: NIT: `disappear` across all test names should be changed to `break` for consistency (because of `breakIllusion()`)

src/test/abilities/illusion.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    vi.spyOn(overrides, "ABILITY_OVERRIDE", "get").mockReturnValue(Abilities.NEUTRALIZING_GAS);
+    await game.startBattle([Species.KOFFING]);
+
+    const zorua = game.scene.getEnemyPokemon();
+
+    expect(zorua.illusion.active).equals(false);
+  });
+
+  it("trick the enemy AI", async () => {
```
xsn34kzx, line 75: I get what this test does, but it should have a more descriptive name (e.g., `move effectiveness uses ILLUSION type instead of original type`).

src/test/abilities/illusion.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    vi.spyOn(overrides, "OPP_MOVESET_OVERRIDE", "get").mockReturnValue([Moves.FLAMETHROWER, Moves.PSYCHIC, Moves.TACKLE, Moves.TACKLE]);
+    await game.startBattle([Species.ZOROARK, Species.AXEW]);
+
+    const enemy = game.scene.getEnemyPokemon();
+    const zoroark = game.scene.getPlayerPokemon();
+
+    const flameThwowerEffectiveness = zoroark.getAttackMoveEffectiveness(enemy, enemy.getMoveset()[0], false, true);
+    const psychicEffectiveness = zoroark.getAttackMoveEffectiveness(enemy, enemy.getMoveset()[1], false, true);
+
+    expect(psychicEffectiveness).above(flameThwowerEffectiveness);
```
xsn34kzx, line 85:
```suggestion
    const flameThrowerEffectiveness = zoroark.getAttackMoveEffectiveness(enemy, enemy.getMoveset()[0], false, true);
    const psychicEffectiveness = zoroark.getAttackMoveEffectiveness(enemy, enemy.getMoveset()[1], false, true);

    expect(psychicEffectiveness).above(flameThrowerEffectiveness);
```
NIT: Misspelled flamethrower

xsn34kzx (commented):
Mostly NITs and minor adjustments. Brought up the idea of making `illusion` a property of `summonData` rather than to `Pokemon` directly, may be something to look into but ultimately not a deal breaker.

src/data/ability.ts
```diff
@@ -1898,6 +1897,11 @@ export class PostSummonMessageAbAttr extends PostSummonAbAttr {
   applyPostSummon(pokemon: Pokemon, passive: boolean, args: any[]): boolean {
     pokemon.scene.queueMessage(this.messageFunc(pokemon));
 
+    pokemon.scene.getField(true).map(pokemon => {
+      if (pokemon.breakIllusion()) {
+        pokemon.scene.queueMessage(i18next.t("abilityTriggers:illusionBreak", { pokemonName: getPokemonNameWithAffix(pokemon) }));
+      }
+    });
```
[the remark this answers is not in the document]
torranx: \+ on this and curious why its on a loop?

src/data/ability.ts
```diff
@@ -4041,6 +4049,97 @@ export class IceFaceBlockPhysicalAbAttr extends ReceivedMoveDamageMultiplierAbAt
...
+   * @returns {boolean} - Whether the illusion was destroyed.
+   */
+  applyPostDefend(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: Move, hitResult: HitResult, args: any[]): boolean {
+
+    const breakIllusion: HitResult[] = [HitResult.EFFECTIVE, HitResult.SUPER_EFFECTIVE, HitResult.NOT_VERY_EFFECTIVE, HitResult.ONE_HIT_KO];
+    if (!breakIllusion.includes(hitResult)) {
+      return false;
+    }
+    pokemon.breakIllusion();
+    pokemon.scene.queueMessage(i18next.t("abilityTriggers:illusionBreak", { pokemonName: getPokemonNameWithAffix(pokemon) }));
```
torranx, line 4105: this 2 seem to be always present together, would it not be ideal to include the `queueMessage` in `breakIllusion()`?

src/data/ability.ts
```diff
@@ -4041,6 +4049,97 @@ export class IceFaceBlockPhysicalAbAttr extends ReceivedMoveDamageMultiplierAbAt
...
+    return true;
+  }
+}
+
+export class IllusionAfterBattle extends PostBattleAbAttr {
```
torranx, line 4110: shouldn't Illusion follow Ice Face and Disguise implementation where "after battle" means when the arena resets?

src/field/pokemon.ts
```diff
@@ -126,6 +127,8 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     const randAbilityIndex = Utils.randSeedInt(2);
 
     this.species = species;
+    this.illusion = {active: false, available: true};
```
torranx, line 130: nit
```suggestion
    this.illusion = { active: false, available: true };
```

src/test/abilities/illusion.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    vi.spyOn(overrides, "BATTLE_TYPE_OVERRIDE", "get").mockReturnValue("single");
+    vi.spyOn(overrides, "OPP_SPECIES_OVERRIDE", "get").mockReturnValue(Species.ZORUA);
+    vi.spyOn(overrides, "OPP_ABILITY_OVERRIDE", "get").mockReturnValue(Abilities.ILLUSION);
+    vi.spyOn(overrides, "OPP_MOVESET_OVERRIDE", "get").mockReturnValue([Moves.TACKLE, Moves.TACKLE, Moves.TACKLE, Moves.TACKLE]);
+    vi.spyOn(overrides, "OPP_HELD_ITEMS_OVERRIDE", "get").mockReturnValue([{name: "WIDE_LENS", count: 3}]);
+
+    vi.spyOn(overrides, "MOVESET_OVERRIDE", "get").mockReturnValue([Moves.WORRY_SEED, Moves.SOAK, Moves.TACKLE, Moves.TACKLE]);
+    vi.spyOn(overrides, "STARTING_HELD_ITEMS_OVERRIDE", "get").mockReturnValue([{name: "WIDE_LENS", count: 3}]);
```
torranx, line 38: please use the new test helpers

src/test/abilities/illusion.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  });
+
+  it("create illusion at the start", async () => {
```
torranx, line 41: nit
```suggestion
  it("creates illusion at the start", async () => {
```

src/test/abilities/illusion.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    await game.startBattle([Species.ZOROARK, Species.AXEW]);
+
+    const zoroark = game.scene.getPlayerPokemon();
+    const zorua = game.scene.getEnemyPokemon();
+
+    expect(zoroark.illusion.active).equals(true);
+    expect(zorua.illusion.active).equals(true);
+    expect(zoroark.illusion.available).equals(false);
+
```
torranx, line 50: nit
```suggestion
```

src/test/abilities/illusion.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  });
+
+  it("disappear after receiving damaging move and changing ability move", async () => {
```
torranx, line 53: this should be in 2 different test cases (1 for damaging move and 1 for ability changing move)

src/test/abilities/illusion.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    const enemy = game.scene.getEnemyPokemon();
+    const zoroark = game.scene.getPlayerPokemon();
+
+    const flameThwowerEffectiveness = zoroark.getAttackMoveEffectiveness(enemy, enemy.getMoveset()[0], false, true);
+    const psychicEffectiveness = zoroark.getAttackMoveEffectiveness(enemy, enemy.getMoveset()[1], false, true);
+
+    expect(psychicEffectiveness).above(flameThwowerEffectiveness);
+  });
+
+  it("do not disappear if the pokemon takes indirect damages", async () => {
```
torranx, line 88: please also test for moves like Flame Burst

torranx (changes requested):

torranx:
Please add spaces before and after the video links so it embeds correctly, and add label on what the video shows.

src/locales/pt_BR/config.ts
```diff
@@ -1,5 +1,6 @@
 import { ability } from "./ability";
 import { abilityTriggers } from "./ability-trigger";
+import { arenaFlyout } from "./arena-flyout";
```
Madmadness65, line 3: This is a duplicate import, it can be removed

PyGaVS: ok

src/test/abilities/illusion.test.ts
PyGaVS: ok

src/data/ability.ts
PyGaVS: yes

src/data/ability.ts
```diff
@@ -4041,6 +4049,97 @@ export class IceFaceBlockPhysicalAbAttr extends ReceivedMoveDamageMultiplierAbAt
   }
 }
 
+export class PreSummonAbAttr extends AbAttr {
```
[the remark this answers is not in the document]
PyGaVS: ok

src/data/ability.ts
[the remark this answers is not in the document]
PyGaVS: > + on this and curious why its on a loop?

It's on loop for checking every pokemons on the field

src/data/ability.ts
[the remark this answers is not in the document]
PyGaVS: ok

src/data/ability.ts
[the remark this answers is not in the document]
PyGaVS: > Does this have to be in the generic `PostSummonMessageAbAttr` or can it be in its own extended class?

so i will do a `PostSummonIllusionDisableAbAttr` class

src/field/pokemon.ts
PyGaVS: ignoreOverride and useIllusion or not the same parameter they are not synced, sometime we want to ignoreOverride and still useIllision i can't use ignoreOverride for illusion. Same for summonData we don't want to change the type of the Pokemon but only the displayed type and that's not the same thing

src/test/abilities/illusion.test.ts
PyGaVS: ok

src/test/abilities/illusion.test.ts
PyGaVS: yes good idea

src/test/abilities/illusion.test.ts
PyGaVS: my bad

src/data/ability.ts
PyGaVS: No because it break when it's the SelectModifierPhase, that's force switch and do not give immunity

src/test/abilities/illusion.test.ts
PyGaVS: ok

src/test/abilities/illusion.test.ts
PyGaVS: ok

torranx (commented):
I still don't like how it's adding `useIllusion`- a very specific parameter just for illusion to a lot of functions

src/data/ability.ts
```diff
@@ -4091,6 +4102,101 @@ export class IceFaceBlockPhysicalAbAttr extends ReceivedMoveDamageMultiplierAbAt
...
+    return false;
+  }
+}
+
+export class IllusionPreSummonAbAttr extends PreSummonAbAttr {
+  /**
+   * Apply a new illusion when summoning Zoroark if the illusion is available
+   *
+   * @param {Pokemon} pokemon - The Pokémon with the Illusion ability.
+   * @param {boolean} passive - N/A
```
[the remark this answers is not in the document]
PyGaVS: they asked me to do that but np i will remove lol

src/data/ability.ts
```diff
@@ -4091,6 +4102,101 @@ export class IceFaceBlockPhysicalAbAttr extends ReceivedMoveDamageMultiplierAbAt
   }
 }
 
+/**
+ * Base class for defining {@linkcode Ability} attributes before summon
+ * (should use {@linkcode PostSummonAbAttr} for most ability)
```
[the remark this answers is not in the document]
PyGaVS: same here

src/data/ability.ts
```diff
@@ -1945,7 +1944,15 @@ export class PostSummonMessageAbAttr extends PostSummonAbAttr {
 
   applyPostSummon(pokemon: Pokemon, passive: boolean, args: any[]): boolean {
     pokemon.scene.queueMessage(this.messageFunc(pokemon));
+    return true;
+  }
+}
 
+export class PostSummonRemoveIllusionAbAttr extends PostSummonAbAttr {
```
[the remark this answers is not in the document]
PyGaVS: It's for pokemon who has ability like neutralizing gas that breaks illusion if a pokemon has an illusion on the field. Some dev asked me to create that class too

src/field/pokemon.ts
```diff
@@ -71,6 +71,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
   public name: string;
   public nickname: string;
   public species: PokemonSpecies;
+  public illusion: Illusion;
```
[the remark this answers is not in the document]
PyGaVS: ok

src/data/ability.ts
```diff
@@ -4080,6 +4091,101 @@ export class FormBlockDamageAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
...
+  applyPreSummon(pokemon: Pokemon, passive: boolean, args: any[]): boolean {
+    let suppressed = false;
+    pokemon.scene.getField(true).filter(p => p !== pokemon).map(p => {
+      if (p.getAbility().hasAttr(SuppressFieldAbilitiesAbAttr) && p.canApplyAbility()) {
+        suppressed = true;
+      }
+      if (p.getPassiveAbility().hasAttr(SuppressFieldAbilitiesAbAttr) && p.canApplyAbility(true)) {
+        suppressed = true;
+      }
+    });
```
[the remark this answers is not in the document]
PyGaVS: ok

src/data/ability.ts
```diff
@@ -2249,14 +2256,18 @@ export class PostSummonTransformAbAttr extends PostSummonAbAttr {
...
-    let target: Pokemon;
+    let target: Pokemon = targets[0];
     if (targets.length > 1) {
       pokemon.scene.executeWithSeedOffset(() => target = Utils.randSeedItem(targets), pokemon.scene.currentBattle.waveIndex);
-    } else {
+    } else if (targets.length === 1) {
       target = targets[0];
+    } else {
+      return false;
     }
```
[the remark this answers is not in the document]
PyGaVS: i will change that

src/data/ability.ts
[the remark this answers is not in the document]
PyGaVS: the pokemon is not on field so i still need to check if there is SuppressFieldAbilitiesAbAttr, but i can shorten it

CodeTappert:
Please fix the problems before requesting a review

flx-sta:
@PyGaVS are you still working on this?

PyGaVS:
Okay i forgot that Zoroark can update his illusion if it's available after a switchout so i need to update that.
Otherwise i can't use BattleData or SummonData because i need to store a data (if the illusion is available or not) after a switchout if you know a way to use Illusion in SummonData or BattleData tell me.

torranx:
> The useIllusion parameter that's now fed to a lot of methods in Pokemon needs to change to account for future implementations that may want a similar effect (e.g. the enemy AI adjusting for form changes when evaluating type matchups). It would probably be more beneficial to adjust it to the inverse (say, useBaseType) where, if true, all effects that may change the Pokemon's observed type are ignored.

Did you consider updating your PR with this suggestion? Or have you got any idea on an alternative that would eliminate the use of the specific `useIllusion` param?

PyGaVS:
> > The useIllusion parameter that's now fed to a lot of methods in Pokemon needs to change to account for future implementations that may want a similar effect (e.g. the enemy AI adjusting for form changes when evaluating type matchups). It would probably be more beneficial to adjust it to the inverse (say, useBaseType) where, if true, all effects that may change the Pokemon's observed type are ignored.
> 
> Did you consider updating your PR with this suggestion? Or have you got any idea on an alternative that would eliminate the use of the specific `useIllusion` param?

@torranx i have got an idea tell me if this is better : we can add a param named "..._args: any[]" where we put additional arguments for specific params do you prefer this idea or should i rename useIllusion param ?

innerthunder:
> @torranx i have got an idea tell me if this is better : we can add a param named "..._args: any[]" just like in ability.ts where we put additional arguments for specific params do you prefer this idea or should i rename useIllusion param ?

Not sure how well this would work (typically the "...args" functions should be overridden by subclasses for stricter typing), but generally speaking it's better practice to use `unknown` over `any` to enforce stricter type checks.

src/data/move.ts
```diff
@@ -6644,7 +6645,8 @@ export class SuppressAbilitiesIfActedAttr extends MoveEffectAttr {
 
 export class TransformAttr extends MoveEffectAttr {
   async apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): Promise<boolean> {
-    if (!super.apply(user, target, move, args)) {
+    if (!super.apply(user, target, move, args) || target.battleData.illusion.active || user.battleData.illusion.active) {
+      user.scene.queueMessage(i18next.t("battle:attackFailed"));
```
innerthunder, line 6649: Do this through `getCondition` instead. Otherwise, it won't actually register using a failed move

src/field/pokemon.ts
```diff
@@ -149,6 +149,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     }
 
     this.species = species;
+
```
innerthunder, line 152:
```suggestion
```

src/data/move.ts
PyGaVS: ahhh yes

damocleas (commented):
Should try to check for Sub Legendaries / Mythicals / (and last resort) Legendaries (maybe if none of any of these can be found, another mon in the biome) in equal or higher boss tiers than a set 'Entei/Raikou/Suicune', as they aren't in every biome, but every biome does have a sub legend in it. Hope this makes sense, discussed on discord.

flx-sta (dismissed):

src/field/pokemon.ts
```diff
@@ -1216,9 +1391,10 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
    * @param includeTeraType - `true` to include tera-formed type; Default: `false`
    * @param forDefend - `true` if the pokemon is defending from an attack; Default: `false`
    * @param ignoreOverride - If `true`, ignore ability changing effects; Default: `false`
+   * @param useIllusion - `true` to return the types of the illusion instead of the actual types; Default: "AUTO"
```
innerthunder, line 1394: Needs docs on the meaning behind the "AUTO" setting

innerthunder (changes requested):
The enemy AI currently uses damage calculations to search for KOs, and will likely make more use of damage calculations when evaluating moves in the future. This means that the `Illusion` class should supply effective stats (HP, Attack, Defense, etc.) in support of bidirectional damage calculations. It might be easier to use a direct `Pokemon` reference in illusion and construct a new `Pokemon` object in cases where the Illusion isn't based on party Pokemon (i.e. Wild battles).

torranx (dismissed):
this is still using the very specific param for illusion `useIllusion`. if other devs are ok with it, feel free to dismiss my `requested changes`

src/field/pokemon.ts
PyGaVS: ok

PyGaVS:
> this is still using the very specific param for illusion `useIllusion`. if other devs are ok with it, feel free to dismiss my `requested changes`

@torranx i didn't found a better alt because we can't rename the params to a less specific name for 2 reasons :
- this param will always be specific because the illusion mechanic itself is specific
- if we rename this params other devs in the future will not understand the utility of this param

Maybe we can create a "purpose" param where purpose can either be "attack", "battleInfo", etc but only the illusion mechanic will use it today, so idk

torranx:
> @torranx i didn't found a better alt because we can't rename the params to a less specific name for 2 reasons :
> 
> * this param will always be specific because the illusion mechanic itself is specific
> * if we rename this params other devs in the future will not understand the utility of this param
> 
> Maybe we can create a "purpose" param where purpose can either be "forDefend", "battleInfo", etc but only the illusion mechanic will use it today, so idk

First of all thank you for being patient with us.
Regarding the param, it doesn't need to be specific, see @innerthunder 's [comment](https://github.com/pagefaultgames/pokerogue/pull/3273#pullrequestreview-2246210211)

PyGaVS:
> > @torranx i didn't found a better alt because we can't rename the params to a less specific name for 2 reasons :
> > 
> > * this param will always be specific because the illusion mechanic itself is specific
> > * if we rename this params other devs in the future will not understand the utility of this param
> > 
> > Maybe we can create a "purpose" param where purpose can either be "forDefend", "battleInfo", etc but only the illusion mechanic will use it today, so idk
> 
> First of all thank you for being patient with us. Regarding the param, it doesn't need to be specific, see @innerthunder 's [comment](https://github.com/pagefaultgames/pokerogue/pull/3273#pullrequestreview-2246210211)

I will do it later

PyGaVS:
> The enemy AI currently uses damage calculations to search for KOs, and will likely make more use of damage calculations when evaluating moves in the future. This means that the `Illusion` class should supply different effective stats (HP, Attack, Defense, etc.) in support of bidirectional damage calculations. It might be easier to use a direct `Pokemon` reference in illusion and construct a new `Pokemon` object in cases where the Illusion isn't based on party Pokemon (i.e. Wild battles).

So i used a direct Pokemon reference for base Pokemon in the interface Illusion this is now way better.
But for the specific params illusion i tried to remove it by adding other func for illusion but when i do that i'm forced to add if statement with illusion everywhere in the ui files

Mourouh:
https://github.com/user-attachments/assets/9efb92f4-8f6f-4282-8356-71b0317519f8

Illusion seems to be getting revealed in the battle end phase? Happens after both wild and trainer battles. Refreshing fixes it back. Recalling the Pokémon or arena change puts the illusion back.

It's not a visual error because move generation adapts to it. Tested this by doing this override and selecting Zorua and a Normal Type like Lillipup. On wave 2 the move AI will favour Hex on what's supposed to be a Normal type illusion Zorua. 

`OPP_MOVESET_OVERRIDE: [Moves.HEX, Moves.PSYBEAM]`

PyGaVS:
> Video.sin.titulo.Hecho.con.Clipchamp.1.mp4 
> Illusion seems to be getting revealed in the battle end phase? Happens after both wild and trainer battles. Refreshing fixes it back. Recalling the Pokémon or arena change puts the illusion back.
> 
> It's not a visual error because move generation adapts to it. Tested this by doing this override and selecting Zorua and a Normal Type like Lillipup. On wave 2 the move AI will favour Hex on what's supposed to be a Normal type illusion Zorua.
> 
> `OPP_MOVESET_OVERRIDE: [Moves.HEX, Moves.PSYBEAM]`



> Video.sin.titulo.Hecho.con.Clipchamp.1.mp4 
> Illusion seems to be getting revealed in the battle end phase? Happens after both wild and trainer battles. Refreshing fixes it back. Recalling the Pokémon or arena change puts the illusion back.
> 
> It's not a visual error because move generation adapts to it. Tested this by doing this override and selecting Zorua and a Normal Type like Lillipup. On wave 2 the move AI will favour Hex on what's supposed to be a Normal type illusion Zorua.
> 
> `OPP_MOVESET_OVERRIDE: [Moves.HEX, Moves.PSYBEAM]`

it was intentionnal but while correcting it i found another bug when we save and quit so i will update this pr this week end

PyGaVS:
> > Video.sin.titulo.Hecho.con.Clipchamp.1.mp4
> > Illusion seems to be getting revealed in the battle end phase? Happens after both wild and trainer battles. Refreshing fixes it back. Recalling the Pokémon or arena change puts the illusion back.
> > It's not a visual error because move generation adapts to it. Tested this by doing this override and selecting Zorua and a Normal Type like Lillipup. On wave 2 the move AI will favour Hex on what's supposed to be a Normal type illusion Zorua.
> > `OPP_MOVESET_OVERRIDE: [Moves.HEX, Moves.PSYBEAM]`
> 
> > Video.sin.titulo.Hecho.con.Clipchamp.1.mp4
> > Illusion seems to be getting revealed in the battle end phase? Happens after both wild and trainer battles. Refreshing fixes it back. Recalling the Pokémon or arena change puts the illusion back.
> > It's not a visual error because move generation adapts to it. Tested this by doing this override and selecting Zorua and a Normal Type like Lillipup. On wave 2 the move AI will favour Hex on what's supposed to be a Normal type illusion Zorua.
> > `OPP_MOVESET_OVERRIDE: [Moves.HEX, Moves.PSYBEAM]`
> 
> it was intentionnal but i found a bug when it's corrected so i will update this pr when i will have time again

Update : bug solved !

src/data/ability.ts
```diff
@@ -5183,6 +5202,77 @@ export class FormBlockDamageAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
...
+   */
+  applyPostDefend(pokemon: Pokemon, passive: boolean, simulated: boolean, attacker: Pokemon, move: Move, hitResult: HitResult, args: any[]): boolean {
+
+    const breakIllusion: HitResult[] = [ HitResult.EFFECTIVE, HitResult.SUPER_EFFECTIVE, HitResult.NOT_VERY_EFFECTIVE, HitResult.ONE_HIT_KO ];
+    if (!breakIllusion.includes(hitResult)) {
+      return false;
+    }
+    pokemon.breakIllusion();
+    return true;
+  }
```
SirzBenjie, line 5256: I believe this needs to be changed into `canApplyPostDefend` and `applyPostDefend` from #5267

src/data/ability.ts
```diff
@@ -5183,6 +5202,77 @@ export class FormBlockDamageAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
...
+   *
+   * @param {Pokemon} pokemon - The Pokémon with the Illusion ability.
+   * @param {boolean} passive - N/A
+   * @param {...any} args - N/A
+   * @returns {boolean} - Whether the illusion was applied.
+   */
+  applyPostBattle(pokemon: Pokemon, passive: boolean, simulated:boolean, args: any[]): boolean {
+    pokemon.breakIllusion();
+    pokemon.battleData.illusion.available = true;
+    return true;
```
SirzBenjie, line 5271: Split in accordance with #5267

src/field/pokemon.ts
```diff
@@ -686,6 +691,94 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      if (this.shiny) {
+        this.initShinySparkle();
+      }
+      this.loadAssets(false, true).then(() => this.playAnim());
+      this.updateInfo();
+    } else {
+      let availables: Species[] = [];
+      let randomIllusion: PokemonSpecies = globalScene.arena.randomSpecies(globalScene.currentBattle.waveIndex, this.level);
+      if (this.isBoss()) {
+        availables = [ Species.ENTEI, Species.RAIKOU, Species.SUICUNE ];
```
SirzBenjie, line 739: The choices for which legendary to use as boss need to be re-evaluated. Probably do this based on biome.

SirzBenjie (changes requested):
This PR is a large endeavor, and it is admirable that you have been continually updating it as we release new updates.

I believe the in-game behavior for illusion is mostly right (with the exception of the boss pokemon chosen for the illusion), as the AI is properly considering the type, etc.
We may want to discuss with the balance team whether the pokemon's level should also be faked.

I am a bit perturbed by the `breakIllusion` method. First, it's using the `!` operator to avoid null checks. We should avoid that when possible.
Unless I'm missing something, clearing the illusion should be as simple as setting `this.battleData.illusion` to be undefined. Then, in each of the methods that checks against illusion, all it needs to do is access the property it is looking for and do nullish coalescing.
This will free up all of that data when it doesn't need to be there anymore. (I assume that once an illusion is broken, it can never re-activate?)


(e.g. as you have done in `getPlayerParty()`

```ts
pokemon.fusionVariant = pokemon.battleData?.illusion.basePokemon?.fusionVariant ?? pokemon.fusionVariant;
pokemon.fusionShiny = pokemon.battleData?.illusion.basePokemon?.fusionShiny ?? pokemon.fusionShiny
```

src/field/pokemon.ts
```diff
@@ -694,54 +787,61 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+              fusionFormIndex, 
+              fusionShiny, 
+              fusionVariant
+            );
+            globalScene.loadPokemonAtlas(
+              this.getFusionBattleSpriteKey(true, ignoreOverride), 
+              this.getFusionBattleSpriteAtlasPath(true, ignoreOverride)
+            );
+          }
+          globalScene.load.once(Phaser.Loader.Events.COMPLETE, () => {
```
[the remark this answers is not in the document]
emdeann: The missing frame comment is unrelated to this pr, looks like it was just moved

src/field/pokemon.ts
```diff
@@ -1698,36 +1822,94 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    } else if (!ignoreOverride && this.summonData?.fusionGender !== undefined) {
       return this.summonData.fusionGender;
     }
     return this.fusionGender;
   }
 
-  isShiny(): boolean {
-    return this.shiny || (this.isFusion() && this.fusionShiny);
+  /**
+   * @param {boolean} fakeShininess - Whether we want the fake or real shininess (illusion ability).
```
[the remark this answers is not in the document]
PyGaVS: Some dev wanted me to rename these parameters to be usable for similar effect

PyGaVS:
> Unless I'm missing something, clearing the illusion should be as simple as setting this.battleData.illusion to be undefined. Then, in each of the methods that checks against illusion, all it needs to do is access the property it is looking for and do nullish coalescing.

This is not that simple : if Illusion isn't active we need to know if it's broken or available and if the illusion is active e also want the base pokemon properties to be accessible depending on the use, that's why there is so many parameters everywhere

src/messages.ts
```diff
@@ -8,7 +8,7 @@ import i18next from "i18next";
  * @param pokemon {@linkcode Pokemon} name and battle context will be retrieved from this instance
```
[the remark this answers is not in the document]
PyGaVS: yes

src/field/pokemon.ts
```diff
@@ -1698,36 +1822,94 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   */
+  isShiny(fakeShininess: boolean = false): boolean {
+    if (!fakeShininess && this.battleData?.illusion.active) {
+      return this.battleData?.illusion.basePokemon?.shiny || (!!this.battleData?.illusion.fusionSpecies && this.battleData?.illusion.basePokemon?.fusionShiny) || false;
+    } else {
+      return this.shiny || (this.isFusion(fakeShininess) && this.fusionShiny);
+    }
+  }
+
+  isDoubleShiny(fakeShininess: boolean = false): boolean {
```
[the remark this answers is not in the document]
PyGaVS: yes

src/data/ability.ts
PyGaVS: done

src/field/pokemon.ts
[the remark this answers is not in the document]
PyGaVS: the code was just moved

PyGaVS:
> > This is not that simple : if Illusion isn't active we need to know if it's broken or available and if the illusion is active e also want the base pokemon properties to be accessible depending on the use, that's why there is so many parameters everywhere
> 
> What I'm specifically asking is, why do we need to know whether it's broken or active? Is there an interaction that can turn the illusion back on other than switching? Can't the on summon attribute just activate it, and then set these properties? And then when it's broken, it goes away completely.

@SirzBenjie the corrections are done

src/data/ability.ts
```diff
@@ -5182,6 +5201,89 @@ export class FormBlockDamageAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
...
+    const lastPokemon: Pokemon = party.filter(p => p !==pokemon).at(-1) || pokemon;
+    pokemon.setIllusion(lastPokemon);
+  }
+
+  override canApplyPreSummon(pokemon: Pokemon, passive: boolean, args: any[]): boolean {
+    if(pokemon.hasTrainer()){
+      const party: Pokemon[] = (pokemon.isPlayer() ? globalScene.getPlayerParty() : globalScene.getEnemyParty()).filter(p => p.isAllowedInBattle());
+      const lastPokemon: Pokemon = party.filter(p => p !==pokemon).at(-1) || pokemon;
+      const speciesId = lastPokemon.species.speciesId;
+      if ( lastPokemon === pokemon || !!pokemon.summonData?.illusion ||
```
SirzBenjie:
```suggestion
      // Terapagos and Ogergpon cannot be an illusion if either they or the pokemon with illusion is terastallized.
      // See https://bulbapedia.bulbagarden.net/wiki/Illusion_(Ability)#Generation_IX
      if ( lastPokemon === pokemon || !!pokemon.summonData?.illusion ||
```

src/data/ability.ts
```diff
@@ -5182,6 +5201,89 @@ export class FormBlockDamageAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
...
+}
+
+export class IllusionPostBattleAbAttr extends PostBattleAbAttr {
+  /**
+   * Illusion will be available again after a battle and apply the illusion of the pokemon is already on field
+   *
+   * @param {Pokemon} pokemon - The Pokémon with the Illusion ability.
+   * @param {boolean} passive - N/A
+   * @param {...any} args - N/A
+   * @returns {boolean} - Whether the illusion was applied.
```
SirzBenjie, line 5280:
```suggestion
   * @param pokemon - The Pokémon with the Illusion ability.
   * @param passive - Unused
   * @param args - Unused
   * @returns - Whether the illusion was applied.
```

src/data/ability.ts
```diff
@@ -5182,6 +5201,89 @@ export class FormBlockDamageAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
...
+        return false;
+      }
+    }
+    return pokemon.canApplyAbility();
+  }
+}
+
+export class IllusionBreakAbAttr extends PostDefendAbAttr {
+  /**
+   * Destroy illusion if attack move deals damage to zoroark
```
SirzBenjie, line 5253:
```suggestion
   * Destroy the illusion upon taking damage
```

src/data/ability.ts
```diff
@@ -5182,6 +5201,89 @@ export class FormBlockDamageAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
...
+   * @param {Pokemon} pokemon - The Pokémon with the Illusion ability.
+   * @param {boolean} passive - N/A
+   * @param {...any} args - N/A
+   * @returns {boolean} - Whether the illusion was applied.
+   */
+  override applyPostBattle(pokemon: Pokemon, passive: boolean, simulated:boolean, args: any[]): void {
+    /*const illusion = pokemon.breakIllusion();
+    if (illusion) {
+      pokemon.summonData?.illusion.available = true;
+    }*/
```
SirzBenjie, line 5286:
```suggestion
```

src/data/ability.ts
```diff
@@ -5182,6 +5201,89 @@ export class FormBlockDamageAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
...
+
+  override canApplyPostDefend(pokemon: Pokemon, passive: boolean, simulated: boolean, attacker: Pokemon, move: Move, hitResult: HitResult, args: any[]): boolean {
+    const breakIllusion: HitResult[] = [ HitResult.EFFECTIVE, HitResult.SUPER_EFFECTIVE, HitResult.NOT_VERY_EFFECTIVE, HitResult.ONE_HIT_KO ];
+    return breakIllusion.includes(hitResult) && !!pokemon.summonData?.illusion
+  }
+}
+
+export class IllusionPostBattleAbAttr extends PostBattleAbAttr {
+  /**
+   * Illusion will be available again after a battle and apply the illusion of the pokemon is already on field
```
SirzBenjie, line 5275:
```suggestion
   * Break the illusion once the battle ends
```

src/field/pokemon.ts
```diff
@@ -687,6 +692,82 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      this.loadAssets(false, true).then(() => this.playAnim());
+      this.updateInfo();
+    } else {
+      const randomIllusion: PokemonSpecies = globalScene.arena.randomSpecies(globalScene.currentBattle.waveIndex, this.level);
+      /*
+      if (this.isBoss()) {
+        const availables = [ Species.ENTEI, Species.RAIKOU, Species.SUICUNE ];
+        randomIllusion = getPokemonSpecies(availables[this.randSeedInt(availables.length)]);
+      }
+      */
```
SirzBenjie, line 731:
```suggestion
```

src/battle-scene.ts
```diff
@@ -887,8 +887,20 @@ export default class BattleScene extends SceneBase {
     return true;
   }
 
-  public getPlayerParty(): PlayerPokemon[] {
-    return this.party;
+  public getPlayerParty(useIllusion = true): PlayerPokemon[] {
+    const party = this.party;
+    if (!useIllusion) {
```
SirzBenjie, line 892: This looks to be a bug, should be `useIllusion`, right? (or should the argument name be inverted?)

PyGaVS: By default we want to take the illusion in consideration. In case of saving (when we want to save the party), useIllusion parameter is set to false so the Zorua is saved properly without his properties being changed by illusion.

SirzBenjie:
Just to hash out what we discussed on discord.

We want to remove the `basePokemon` reference inside of the `Illusion` interface.
The illusion interface should contain _only_ the information about the pokemon featured in the illusion.
So it should look something like this:
```ts
/**
 * Illusion property
 */
interface Illusion {
  /**
   * Whether the illusion is active or not.
   * @type {boolean}
   */
  //active: boolean;
  /**
   * Whether the pokemon can generate an illusion or not.
   * @type {boolean}
   */
  //available: boolean;
  /** The name of the pokemon in the illusion*/
  name: string,
  /** The nickname of the pokemon in the illusion*/
  nickname: string
  /** The speciesId of the pokemon in the illusion*/
  speciesId: Species;
  /** The formIndex of the illusion */
  formIndex: number;
  /** The gender of the pokemon in the illusion
  gender: Gender;
  // And so on...
}
```

Then, you make each of these properties from the pokemon `private` (actually, likely `protected` so that the subclasses can acccess them).
All of the getter methods for these properties would then take a `useIllusion` parameter (which likely defaults to `true`). For example:

```ts
getGender(useIllusion = true) {
  if (useIllusion && this.summonData.illusion) {
    return this.summonData.illusion.gender;
  }
  return this.gender;
}
```

SirzBenjie (approved):
I think this is in a sufficient state that it's time to get it into the codebase.

Madmadness65 (approved):
At long last, it seems its finally ready to be merged 🎉
Everything seems to function as intended from a player's perspective from my in-game testing, so this lgtm

src/field/pokemon.ts
```diff
@@ -1660,36 +1794,99 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    if (!useIllusion && !!this.summonData?.illusion) {
+      return !this.isFusion(false) 
+      ? this.summonData?.illusion.basePokemon!.variant 
+      : Math.max(this.variant, this.fusionVariant) as Variant;
+    } else {
+      return !this.isFusion(true) 
+      ? this.variant 
+      : Math.max(this.variant, this.fusionVariant) as Variant;
+    }
+
```
DayKev, line 1858:
```suggestion
```

src/field/pokemon.ts
```diff
@@ -1823,17 +2022,19 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      const doIllusion: boolean = (useIllusion === "AUTO") ? !forDefend : useIllusion;
       if (
-        !ignoreOverride &&
-        this.summonData?.types &&
-        this.summonData.types.length > 0
+        !ignoreOverride && 
+        this.summonData?.types && 
+        this.summonData.types.length > 0 && 
+        (!!!this.summonData?.illusion || !doIllusion)
```
DayKev, line 2032:
```suggestion
        (!this.summonData?.illusion || !doIllusion)
```

src/data/ability.ts
```diff
@@ -5260,6 +5279,92 @@ export class FormBlockDamageAbAttr extends ReceivedMoveDamageMultiplierAbAttr {
...
+}
+
+export class IllusionPreSummonAbAttr extends PreSummonAbAttr {
+  /**
+   * Apply a new illusion when summoning Zoroark if the illusion is available
+   *
+   * @param {Pokemon} pokemon - The Pokémon with the Illusion ability.
+   * @param {boolean} passive - N/A
+   * @param {...any} args - N/A
+   * @returns {boolean} - Whether the illusion was applied.
```
SirzBenjie, line 5302:
```suggestion
   * @param pokemon - The Pokémon with the Illusion ability.
   * @param passive - N/A
   * @param args - N/A
   * @returns Whether the illusion was applied.
```

src/field/pokemon.ts
```diff
@@ -2386,15 +2587,17 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
    * @param ignoreAbility Whether to ignore abilities that might affect type effectiveness or immunity (defaults to `false`).
    * @param simulated Whether to apply abilities via simulated calls (defaults to `true`)
    * @param cancelled {@linkcode Utils.BooleanHolder} Stores whether the move was cancelled by a non-type-based immunity.
+   * @param {boolean} useIllusion - Whether we want the attack move effectiveness on the illusion or not
```
SirzBenjie, line 2590:
```suggestion
   * @param useIllusion - Whether we want the attack move effectiveness on the illusion or not
```

src/field/pokemon.ts
```diff
@@ -2520,19 +2723,21 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
    * @param ignoreStrongWinds whether or not this ignores strong winds (anticipation, forewarn, stealth rocks)
    * @param simulated tag to only apply the strong winds effect message when the move is used
    * @param move (optional) the move whose type effectiveness is to be checked. Used for applying {@linkcode VariableMoveTypeChartAttr}
+   * @param {boolean} useIllusion - Whether we want the attack type effectiveness on the illusion or not
```
SirzBenjie, line 2726:
```suggestion
   * @param useIllusion - Whether we want the attack type effectiveness on the illusion or not
```

src/field/pokemon.ts
```diff
@@ -7551,6 +7753,83 @@ export class EnemyPokemon extends Pokemon {
...
+  /**
+   * The fusionGender of the illusion if it's a fusion
+   * @type {Gender}
+   */
+  fusionGender?: Gender;
+  /**
+   * The level of the illusion (not used currently)
+   * @type {integer}
+   */
+  level?: number
```
SirzBenjie, line 7830:
```suggestion
  basePokemon: {
    /** The actual name of the Pokemon. */
    name: string;
    /** The actual nickname of the Pokemon */
    nickname: string;
    /** Whether the base pokemon is shiny or not */
    shiny: boolean;
    /** The shiny variant of the base pokemon */
    variant: Variant;
    /** Whether the fusion species of the base pokemon is shiny or not */
    fusionShiny: boolean;
    /** The variant of the fusion species of the base pokemon */
    fusionVariant: Variant;
  };
  /** The species of the illusion. */
  species: Species;
  /** The formIndex of the illusion */
  formIndex: number;
  /** The gender of the illusion*/
  gender: Gender;
  /** The pokeball of the illusion */
  pokeball: PokeballType;
  /** The fusion species of the illusion if it's a fusion */
  fusionSpecies?: PokemonSpecies;
  /** The fusionFormIndex of the illusion */
  fusionFormIndex?: number;
  /** The fusionGender of the illusion if it's a fusion */
  fusionGender?: Gender;
  /** The level of the illusion (not used currently) */
  level?: number
```

SirzBenjie (dismissed):

Madmadness65 (dismissed):

DayKev (approved):

DayKev (approved):

Madmadness65 (approved):

xsn34kzx (approved):

PR #3278, opened 2024-08-01
[Refactor/Bug] Use Bit Shifting & Masking

xsn34kzx:
## What are the changes?

May have the effect of improving randomly-generated IV distributions and shiny Pokemon since they were not generated correctly before, but realistically it should have no effect from the user perspective.

A majority of the motivation comes from seeing 2 raised to different powers in order to simulate left bit shifting instead of just using bit shifting directly, adding performance complexity with no benefit. While doing that, more instances were found where bit manipulation would be more performant (see `decToBin` and `binToDec` in `src/utils.ts`), so I aimed to address those as well.

## What did change?

- Refactored `getIvsFromId` such that it uses bit manipulation to extract IVs directly.
  - This will be explained a bit more in the Screenshots section.
- Change constants in `trySetShiny` to use bit manipulation to extract the lower and higher 16 bits directly.
- Remove `decToBin` and `binToDec` since they were only used in two specific scenarios addressed above.
- Changed most instances of `Math.pow(2, ...)` to `1 << ...` where it made sense or implied the usage of bit flags rather than trying to raise a value to the second power literally.

### Screenshots/Videos

In regards to `getIvsFromId`, it was first directly refactored to use bit manipulation, demonstrating that there was a mismatch between the input `id` and the output array with the extracted IVs.
generateIVs
Note that the above array are the expected values and the bottom array was the behavior before this PR. To verify this, the `id` was manually deconstructed: 
- The `id`, 2326818241, gives `10001010101100000110110111000001`
  - `10001` gives 17
  - `01010` gives 10
  - `11000` gives 24
  - `00110` gives 6
  - `11011` gives 27
  - `10000` gives 16
  - Last 2 bits are unused/ignored
 
As such, the original function seemingly did not generate IVs correctly.

This would carry over into `trySetShiny`, where bits were not correctly extracted:
generateShinyValues
Note that the first two values are the expected values and the last two are the values extracted before this PR. Again, to verify this, the `id` was manually deconstructed:
- The `id`, 349173278, gives `00010100110011111111011000011110`
  - `0001010011001111` gives 5327
  - `1111011000011110` gives 63006

Back to `getIvsFromId`, I also realized that, because [bit shifting converts operands to 32-bit integers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Left_shift) and the `randInt` used to generate `id` values ranges from `0` to the largest positive 32-bit value, the most significant bit may not ever be `1` by definition of positive numbers in two's complement. I didn't test it, but I figured the correct change would be the use the bits starting from the least significant bits since this is [what the mainline games do](https://bulbapedia.bulbagarden.net/wiki/Individual_values#Storage) anyway. As a result, the function was refactored and changed entirely to now generate IVs from `id`, ignoring the two most significant bits.

Lastly, note that all `console.log` calls were purely just for testing and aren't included in the PR.

## How to test the changes?

These changes are a bit scattered around, but the best way would most likely be at the starter select screen since that's where data reloading (unlocked egg moves, abilities, natures, etc.) can be verified as correct.

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/modifier/modifier.ts
+++ b/src/modifier/modifier.ts
@@ -1576,11 +1576,11 @@ export class PokemonNatureChangeModifier extends ConsumablePokemonModifier {
     const pokemon = args[0] as Pokemon;
     pokemon.natureOverride = this.nature;
     let speciesId = pokemon.species.speciesId;
-    pokemon.scene.gameData.dexData[speciesId].natureAttr |= Math.pow(2, this.nature + 1);
+    pokemon.scene.gameData.dexData[speciesId].natureAttr |= 1 << (this.nature + 1);
 
     while (pokemonPrevolutions.hasOwnProperty(speciesId)) {
       speciesId = pokemonPrevolutions[speciesId];
-      pokemon.scene.gameData.dexData[speciesId].natureAttr |= Math.pow(2, this.nature + 1);
+      pokemon.scene.gameData.dexData[speciesId].natureAttr |= 1 << (this.nature + 1);
     }
 
     return true;
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -165,40 +165,20 @@ export function getPlayTimeString(totalSeconds: integer): string {
   return `${days.padStart(2, "0")}:${hours.padStart(2, "0")}:${minutes.padStart(2, "0")}:${seconds.padStart(2, "0")}`;
 }
 
-export function binToDec(input: string): integer {
-  const place: integer[] = [];
-  const binary: string[] = [];
-
-  let decimalNum = 0;
-
-  for (let i = 0; i < input.length; i++) {
-    binary.push(input[i]);
-    place.push(Math.pow(2, i));
-    decimalNum += place[i] * parseInt(binary[i]);
-  }
-
-  return decimalNum;
-}
-
-export function decToBin(input: integer): string {
-  let bin = "";
-  let intNum = input;
-  while (intNum > 0) {
-    bin = intNum % 2 ? `1${bin}` : `0${bin}`;
-    intNum = Math.floor(intNum * 0.5);
-  }
-
-  return bin;
-}
-
-export function getIvsFromId(id: integer): integer[] {
+/**
+ * Generates IVs from a given {@linkcode id} by extracting 5 bits at a time
+ * starting from the least significant bit up to the 30th most significant bit.
+ * @param id 32-bit number
+ * @returns An array of six numbers corresponding to 5-bit chunks from {@linkcode id}
+ */
+export function getIvsFromId(id: number): number[] {
   return [
-    binToDec(decToBin(id).substring(0, 5)),
-    binToDec(decToBin(id).substring(5, 10)),
-    binToDec(decToBin(id).substring(10, 15)),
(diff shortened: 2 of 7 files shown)


src/modifier/modifier.ts
```diff
@@ -1576,11 +1576,11 @@ export class PokemonNatureChangeModifier extends ConsumablePokemonModifier {
     const pokemon = args[0] as Pokemon;
     pokemon.natureOverride = this.nature;
     let speciesId = pokemon.species.speciesId;
-    pokemon.scene.gameData.dexData[speciesId].natureAttr |= Math.pow(2, this.nature + 1);
+    pokemon.scene.gameData.dexData[speciesId].natureAttr |= 1 << (this.nature + 1);
 
     while (pokemonPrevolutions.hasOwnProperty(speciesId)) {
       speciesId = pokemonPrevolutions[speciesId];
-      pokemon.scene.gameData.dexData[speciesId].natureAttr |= Math.pow(2, this.nature + 1);
+      pokemon.scene.gameData.dexData[speciesId].natureAttr |= 1 << (this.nature + 1);
```
Tempo-anon, line 1583: nit: for consistency, can you wrap `1 << (this.nature + 1)` in parentheses too

src/utils.ts
```diff
@@ -165,40 +165,20 @@ export function getPlayTimeString(totalSeconds: integer): string {
...
-    binToDec(decToBin(id).substring(10, 15)),
-    binToDec(decToBin(id).substring(15, 20)),
-    binToDec(decToBin(id).substring(20, 25)),
-    binToDec(decToBin(id).substring(25, 30))
+    (id & 0x3E000000) >>> 25,
+    (id & 0x01F00000) >>> 20,
+    (id & 0x000F8000) >>> 15,
+    (id & 0x00007C00) >>> 10,
+    (id & 0x000003E0) >>> 5,
+    (id & 0x0000001F)
```
Tempo-anon, line 181: Isn't this backwards? The last 5 bits used to be the first element in the array but now it's the last element

Tempo-anon (commented):
Why would this change functionality if all this does is refactoring?

xsn34kzx:
> Why would this change functionality if all this does is refactoring?

@Tempo-anon Technically it's a refactor since it does the same thing and doesn't affect existing IVs anyway, but I wanted to also make the distinction since technically it does change how IVs are calculated. Basically, for documentation purposes, it's more of a revamp, but for practical purposes, nothing functionally changed.

src/utils.ts
xsn34kzx: The substring operations are deceiving because `(0, 5)` corresponds to the upper bits (from left to right), which is what line 176 is doing (masking out the bits and then shifting them all the way to the right). So it's in the same "order" technically, but I'm using bits 30 to 0 from left to right, while the old function used substrings for bits 31 to 2.

xsn34kzx: Following up so I remember, I could probably change the order to be the same as mainline instead of what was done before anyway to keep the decision to be more similar to how its done there consistent overall.

flx-sta (approved):
lgtm

PR #2808, opened 2024-07-05
[QOL] Candy Upgrade Notification Rework

josericardo-fo:
## What are the changes?

Reworked the Candy Upgrade Notification logic to take account of the new "Buy Egg" option.

## Why am I doing these changes?

A new mechanic was implemented (Buying Eggs).

## What did change?

The logic of the 'On' setting 

### Before
image

### After
image

## How to test the changes?

Test if the candy icon appears whenever you can buy eggs.

## Checklist
- [X] There is no overlap with another PR?
- [X] The PR is self-contained and cannot be split into smaller PRs?
- [X] Have I provided a clear explanation of the changes?
- [X] Have I tested the changes (manually)?
    - [X] Are all unit tests still passing? (`npm run test`)
- [X] Are the changes visual?
  - [X] Have I provided screenshots/videos of the changes?

PR #3279, opened 2024-08-01
[QOL] Candy Upgrade Notification Rework

josericardo-fo:
## What are the changes?

Remake of #2808 
Reworked the Candy Upgrade Notification logic to take account of the new "Buy Egg" option.

A new mechanic was implemented (Buying Eggs).

## What did change?

The logic of the 'On' setting.

### Before (Look at Spearow)
image

### After
image

## How to test the changes?

Test if the candy icon appears whenever you can buy eggs.

## Checklist
- [X] Are the changes visual?


Fontbane:
Would it be possible to only show this notification after both cost reductions were bought? It would be mildly annoying imo to be grinding for a cost reduction and see the notification, only to find out that you still don't have enough for the cost reduction, just enough to buy an egg. This is kind of the reasoning behind the "Passives Only" option for notifications.

src/ui/starter-select-ui-handler.ts
```diff
@@ -119,6 +119,8 @@ const starterCandyCosts: { passive: integer, costReduction: [integer, integer],
   { passive: 10, costReduction: [3, 10], egg: 10 },  // 10
 ];
 
+export const VALUE_REDUCTION_MAX = 2;
```
torranx, line 122: does this need to be an export?

josericardo-fo: you're right

josericardo-fo:
> Would it be possible to only show this notification after both cost reductions were bought? It would be mildly annoying imo to be grinding for a cost reduction and see the notification, only to find out that you still don't have enough for the cost reduction, just enough to buy an egg. This is kind of the reasoning behind the "Passives Only" option for notifications.

I don't really know how should I implement that. It would make more sense to add another option, but some languages don't even have enough space to fit it. Any suggestions?

josericardo-fo:
@Fontbane There's a new filter in the starter screen now. No need for these type of filters in the setting anymore.

josericardo-fo:
Even with the addition of the new filters, this should be a thing changed in the settings

torranx:
This PR needs some fixing - it carries past commits from other people

josericardo-fo:
> This PR needs some fixing - it carries past commits from other people

Moved to #3720

PR #3284, opened 2024-08-01
[Bug] Use `.fusionSpeciesForm` in `getFusionSpeciesForm()` if statement

DayKev:
## What are the changes?
There shouldn't be any user-facing changes.

Seems like a bug to me.

## What did change?
Change the `if` check in `Pokemon.getFusionSpeciesForm()` from `.speciesForm` to `.fusionSpeciesForm`.

## How to test the changes?
No idea, feels like this would only come up in some strange edge-case scenario.

## Checklist


torranx (changes requested):
The code looks like it's checking for a different property but that could also be intended. Unless you can provide a concrete proof that this is in fact a bug, I don't think it's a good idea to merge this.

PR #3285, opened 2024-08-01
[QoL] IV scanner update to hide the prompt and let you see the stats mid battle on the stats screen

Opaque02:
This is a remake of #2451 in the current beta branch since the commit merge thingy. Also I'd like to thank DayKev because without their help I'd be writing this code from scratch, so thanks!

## What are the changes?
The IV scanner has been updated in two ways - the first is a setting that will allow you to disable the IV scanner prompt and all associated parts (animation, message etc), and the second is to change the stat container labels to show green/gold for the IV mid battle to show if the stats of the pokemon are better than what you've currently got.

A lot of people on discord have mentioned this is a feature they wanted, so I decided to make it

## What did change?
The first option is a setting that enables or disables the prompt about scanning a pokemon's IV. This is under the "general" tab of the settings.

The second option is to change the labels of the stat container to be green or gold depending on the stat. The stats container will show you the same amount of IV updates that you would from the IV scanner - so if you have no scanners you won't see any updates, all the way up to seeing all 6 stats for 3 scanners. If the stat is better than what you have, it'll show green, whereas if it's perfect IV it'll show gold; this is so if you do have the prompts being skipped, you can still see if the stat is perfect. Lastly, if the stat is unknown (not enough scanners), not scanned (ACC/EVA), or equal/worse than what you have, the label is white as usual.

The next change is to do with the stat labels themselves. Now if you have at least one IV scanner, the enemy stats will show a label for HP. There is no +/- number associated with it, as it's just there purely for IV purposes. This label does not show for player pokemon, and only shows when an IV scan would happen - so this label doesn't show for trainer battles

Another small UI change is that the IV scanner prompt now shows as gold for "Best" (31 IV) instead of green as another way to show that it's at max to match the info from the container.

The last change is with how the IV scanner (and thus my code) and the pokemon info container (the container that shows when you catch/hatch a pokemon with the IV spread of the pokemon) work. The pokemon info container shows you increases against the baby form of a pokemon if it's available. For example, catching wigglytuff will show you how much of an increase igglybuff's IVs will take, whereas the IV scanner compares it against the non-baby form - in this case, jigglypuff. This means if jigglypuff and igglybuff have different IV spreads, they will show differently through the IV scanner and the pokemon info container. This last change makes it consistent. See pictures in the below section to see a before/after of this.

### Screenshots/Videos
Here is a screenshot of the settings screen showing the new setting:

image

Here is a screenshot of the stats screen in battle showing some stats and the IV scanner on screen:

image

As you can see, the only stats that are better than what we have are HP, Sp. Atk, Def and Sp. Def. Since Sp. Atk is "Best" (i.e. 31), it shows up gold on the stats container AND within the IV prompt. However, HP, Def and Sp. Def are not "Best", so they shows up green. The Atk and SPD for this pokemon aren't better than what we have, so it's white like usual. If there were any unscanned stats, they would also show up as white, but this is with 3 IV scanners, so all stats are shown.

Below is an example of a fight where pokemon have changed their stats with various stat changing moves: 

image

Below is an example with 2 IV scanners in a run showing the difference with the IV scanner setting on (current behaviour) and the IV scanner setting off:

https://github.com/pagefaultgames/pokerogue/assets/66582645/47438007-3edb-4a4e-a0f3-74605671ad12

Here's an example of a run where there's no IV scanners. You'll notice that the HP stat isn't shown, but everything has been moved over a bit. This is so that when the HP stat does get shown, the ui elements don't shift because of it:

https://github.com/pagefaultgames/pokerogue/assets/66582645/4e590d8e-56d3-4c78-bf69-a7d767bd1499

Here's an example with 3 IV scanners:

https://github.com/pagefaultgames/pokerogue/assets/66582645/77a66662-862d-4d34-8aec-2dfc2387d274

Here's a double battle with 1 IV scanner while also going through different waves:

https://github.com/pagefaultgames/pokerogue/assets/66582645/7696524a-6c99-45ee-855f-1be7cec2a94f

Below is an example of fighting a trainer and not seeing the IV scanner prompt:

https://github.com/pagefaultgames/pokerogue/assets/66582645/c2900bdf-7013-4303-87dc-b7e8a188b3cc

Lastly, below is an example where you start with no IV scanners and then pick up on in the store:

https://github.com/pagefaultgames/pokerogue/assets/66582645/4dab40d5-9ed9-47c5-a1f2-ab89ed8c2de8

Finally, here's a screenshot of the IV scanner in action for the legacy UI:

image

Regarding the change to the IV scanner and pokemon info container, below is a picture of the current behaviour:

image

You can see above that the IV scanner (which uses the same code as the display on the stat labels) shows that the HP, ATK, Sp DEF and SPD should be higher than what we've got, but in reality it's only the HP. Below is the update to make the IV scanner (and thus, my changes to the stat labels) be in line with the pokemon info container:

image

image

As you can see above, now according to the IV scanner (and the stat labels), the only thing better should be HP, and when we catch the pokemon and see the pokemon info container, we see that's the case

## How to test the changes?
Download the branch to play with the changes. When you're at the battle screen, you can open the stats container (default Shift/C) to see the changes. Also in the settings, there should be an option to turn the setting on/off to hide the IV prompts

You can also use any of the following overrides:

Line 32 - this will force a double battle:

`export const DOUBLE_BATTLE_OVERRIDE: boolean = true;`

Line 90 - this will let you choose a specific pokemon to fight against. For example, this is set to spawn a charizard, but you can change that to anything else to fight other things:

`export const OPP_SPECIES_OVERRIDE: Species | integer = Species.CHARIZARD;`

Line 96 - this will let you set the moves the enemy opponent has; for example this could be used to set stat changing moves for enemy pokemon to see the stat numbers appear. In this example, we have given them the moves no retreat, quiver dance, and shell smash twice. This can be set to anything:

`export const OPP_MOVESET_OVERRIDE: Array<Moves> = [Moves.NO_RETREAT, Moves.QUIVER_DANCE, Moves.SHELL_SMASH, Moves.SHELL_SMASH];`

Line 99 - this will let you set the IVs of the enemy pokemon. It takes an array of 6 numbers, which, in order, are HP, Atk, Def, Sp. Atk, Sp. Def, Spd:

`export const OPP_IVS_OVERRIDE: integer | integer[] = [31, 24, 5, 3, 19, 16];`

Line 121 - this will give you 3 IV scanners. To change it, change the "count: 3" to "count: 1" or "count: 2":

`export const STARTING_MODIFIER_OVERRIDE: Array<ModifierOverride> = [{name: "IV_SCANNER", count: 3}];`

**NOTE:** If you want to get rid of your IV scanners, you have to revert this; changing the count to 0 will NOT work, so to have 0 IV scanners, you need to use:

`export const STARTING_MODIFIER_OVERRIDE: Array<ModifierOverride> = [];`

Also note that for the IV scanners, you'll need to start a new game to see the effects. At first it'll look like it didn't work, but you'll be given the choice to see the pokemon's IVs before the fight as usual.

## Checklist
- [x] Are the changes visual?


src/data/battle-stat.ts
```diff
@@ -8,7 +8,8 @@ export enum BattleStat {
   SPD,
   ACC,
   EVA,
-  RAND
+  RAND,
+  HP
```
Tempo-anon, line 12: Right now this is only used for the IV scanner right? Might be worth adding a comment here. Also make sure adding a new enum to `BattleStat` won't break anything

src/phases.ts
```diff
@@ -5492,19 +5492,45 @@ export class ScanIvsPhase extends PokemonPhase {
...
-        new CommonBattleAnim(CommonAnim.LOCK_ON, pokemon, pokemon).play(this.scene, () => {
-          this.scene.ui.getMessageHandler().promptIvs(pokemon.id, pokemon.ivs, this.shownIvs).then(() => this.end());
+    let enemyIvs = [];
+    let statsContainer = [];
+    let statsContainerLabels = [];
+    const enemyField = this.scene.getEnemyField();
+    const uiTheme = (this.scene as BattleScene).uiTheme; // Assuming uiTheme is accessible
+    for (let e = 0; e < enemyField.length; e++) {
+      enemyIvs = enemyField[e].ivs;
+      const currentIvs = this.scene.gameData.dexData[enemyField[e].species.getRootSpeciesId(false)].ivs;  // we are using getRootSpeciesId(false) here because we want to check against the baby form, not the mid form if it exists
```
Tempo-anon, line 5502: nit: I believe `false` is the default so you can just do `getRootSpeciesId()`

src/ui/battle-info.ts
```diff
@@ -12,7 +12,9 @@ import BattleFlyout from "./battle-flyout";
 import { WindowVariant, addWindow } from "./ui-theme";
 import i18next from "i18next";
 
-const battleStatOrder = [ BattleStat.ATK, BattleStat.DEF, BattleStat.SPATK, BattleStat.SPDEF, BattleStat.ACC, BattleStat.EVA, BattleStat.SPD ];
+let battleStatOrder = [];
+const battleStatOrderPlayer = [BattleStat.ATK, BattleStat.DEF, BattleStat.SPATK, BattleStat.SPDEF, BattleStat.ACC, BattleStat.EVA, BattleStat.SPD];
+const battleStatOrderEnemy = [BattleStat.HP, BattleStat.ATK, BattleStat.DEF, BattleStat.SPATK, BattleStat.SPDEF, BattleStat.ACC, BattleStat.EVA, BattleStat.SPD];
```
torranx, line 17:
```suggestion
const battleStatOrderEnemy = [ BattleStat.HP, BattleStat.ATK, BattleStat.DEF, BattleStat.SPATK, BattleStat.SPDEF, BattleStat.ACC, BattleStat.EVA, BattleStat.SPD ];
```

src/ui/battle-info.ts
```diff
@@ -12,7 +12,9 @@ import BattleFlyout from "./battle-flyout";
 import { WindowVariant, addWindow } from "./ui-theme";
 import i18next from "i18next";
 
-const battleStatOrder = [ BattleStat.ATK, BattleStat.DEF, BattleStat.SPATK, BattleStat.SPDEF, BattleStat.ACC, BattleStat.EVA, BattleStat.SPD ];
+let battleStatOrder = [];
+const battleStatOrderPlayer = [BattleStat.ATK, BattleStat.DEF, BattleStat.SPATK, BattleStat.SPDEF, BattleStat.ACC, BattleStat.EVA, BattleStat.SPD];
```
torranx, line 16:
```suggestion
const battleStatOrderPlayer = [ BattleStat.ATK, BattleStat.DEF, BattleStat.SPATK, BattleStat.SPDEF, BattleStat.ACC, BattleStat.EVA, BattleStat.SPD ];
```

src/phases.ts
Opaque02: Fixed in next commit

src/ui/battle-info.ts
```diff
@@ -222,20 +224,44 @@ export default class BattleInfo extends Phaser.GameObjects.Container {
     this.statValuesContainer = this.scene.add.container(0, 0);
     this.statsContainer.add(this.statValuesContainer);
 
+    // this gives us a different starting location from the left of the label and padding between stats for a player vs enemy
+    // since the player won't have HP to show, it doesn't need to change from the current version
+    const startingX = this.player ? -this.statsBox.width + 8 : -this.statsBox.width + 5;
+    const paddingX = this.player ? 4 : 2;
+    const statOverflow = this.player ? 1 : 0;
+    battleStatOrder = this.player ? battleStatOrderPlayer : battleStatOrderEnemy; // this tells us whether or not to use the player or enemy battle stat order
```
torranx, line 232: I see a lot of occurences of this, might consider making `battleStatOrder` a class property instead.

src/ui/battle-message-ui-handler.ts
```diff
@@ -227,8 +210,31 @@ export default class BattleMessageUiHandler extends MessageUiHandler {
     });
   }
 
+  topIvs(ivs: integer[], shownIvsCount: integer): Stat[] {
```
torranx, line 213: I'm assuming this is a function to get the top IVs
```suggestion
  getTopIvs(ivs: integer[], shownIvsCount: integer): Stat[] {
```

src/ui/battle-info.ts
Opaque02: Just so I get my head around it properly, what would be an example of a class property that I can do with this instead? I don't quite know what one would be?

torranx: class properties are those you can access with `this` inside the class ie `private baseY`

src/ui/battle-message-ui-handler.ts
Opaque02: Fixed in next commit

src/ui/battle-info.ts
Opaque02: Fixed in next commit

src/ui/battle-info.ts
Opaque02: Fixed in next commit

src/ui/battle-info.ts
Opaque02: Fixed in next commit

src/data/battle-stat.ts
Opaque02: As far as I can tell, I can't see anything breaking from adding the new battle stat. The places we access the battle stats themselves are all based on set numbers assuming it starts at ATK (for example casting battle stat to regular stat), so adding HP at the end doesn't make a difference. There are a few places where it generates a random battle stat, but it checks the enum === BattleStat.RAND, then manually gets the value of the random stat using [this code](https://github.com/pagefaultgames/pokerogue/blob/main/src/phases.ts#L3395-L3398), which only goes up to speed and nothing further. There's a few other cases like [here](https://github.com/pagefaultgames/pokerogue/blob/main/src/data/move.ts#L2558-L2568) where it manually puts in the allowed BattleStats when generating a random one, but nothing I can see what uses _all_ BattleStats without some kind of limiter (i.e. specifically choosing what stats to use)

src/ui/battle-info.ts
```diff
@@ -12,7 +12,8 @@ import BattleFlyout from "./battle-flyout";
 import { WindowVariant, addWindow } from "./ui-theme";
 import i18next from "i18next";
 
-const battleStatOrder = [ BattleStat.ATK, BattleStat.DEF, BattleStat.SPATK, BattleStat.SPDEF, BattleStat.ACC, BattleStat.EVA, BattleStat.SPD ];
+const battleStatOrderPlayer = [ BattleStat.ATK, BattleStat.DEF, BattleStat.SPATK, BattleStat.SPDEF, BattleStat.ACC, BattleStat.EVA, BattleStat.SPD ];
+const battleStatOrderEnemy = [ BattleStat.HP, BattleStat.ATK, BattleStat.DEF, BattleStat.SPATK, BattleStat.SPDEF, BattleStat.ACC, BattleStat.EVA, BattleStat.SPD ];
```
torranx, line 16: is there any reason why these are defined outside the class? if none, please encapsulate them inside the class.

src/ui/battle-info.ts
```diff
@@ -70,6 +71,8 @@ export default class BattleInfo extends Phaser.GameObjects.Container {
 
   public flyoutMenu?: BattleFlyout;
 
+  public battleStatOrder: BattleStat[];
```
torranx, line 74: if this is not being used by other classes, make it private

src/ui/battle-info.ts
```diff
@@ -222,20 +225,44 @@ export default class BattleInfo extends Phaser.GameObjects.Container {
...
 
-    battleStatOrder.map((s, i) => {
-      const statX = i > 1 ? this.statNumbers[i - 2].x + this.statNumbers[i - 2].width + 4 : -this.statsBox.width + 8;
-      const statY = -this.statsBox.height / 2 + 4 + (i < battleStatOrder.length - 1 ? (i % 2 ? 10 : 0) : 5);
+    // this gives us a different starting location from the left of the label and padding between stats for a player vs enemy
+    // since the player won't have HP to show, it doesn't need to change from the current version
+    const startingX = this.player ? -this.statsBox.width + 8 : -this.statsBox.width + 5;
+    const paddingX = this.player ? 4 : 2;
+    const statOverflow = this.player ? 1 : 0;
+    this.battleStatOrder = this.player ? battleStatOrderPlayer : battleStatOrderEnemy; // this tells us whether or not to use the player or enemy battle stat order
```
torranx, line 233: There are 3 occurences of this line, if this is _always_ the case, maybe we can make this as default?

src/ui/battle-info.ts
Opaque02: Good point - a leftover from the old style. This has been fixed to be privates within the class in the next commit

src/ui/battle-info.ts
Opaque02: I've updated this to only be around in the constructor now as of next commit

src/ui/battle-info.ts
Opaque02: Fixed in next commit

src/locales/fr/settings.ts
```diff
@@ -15,6 +15,7 @@ export const settings: SimpleTranslationEntries = {
   "skipSeenDialogues": "Passer dialogues connus",
   "battleStyle": "Style de combat",
   "enableRetries": "Activer les réessais",
+  "hideIvs": "Hide IV scanner",
```
Adri1, line 18:
```suggestion
  "hideIvs": "Masquer Scanner d’IV",
```

src/locales/de/settings.ts
```diff
@@ -15,6 +15,7 @@ export const settings: SimpleTranslationEntries = {
   "skipSeenDialogues": "Gesehenen Dialog überspringen",
   "battleStyle": "Kampfstil",
   "enableRetries": "Erneut versuchen aktivieren",
+  "hideIvs": "Hide IV scanner",
```
CodeTappert, line 18:
```suggestion
  "hideIvs": "IS-Scanner verstecken",
```

src/locales/pt_BR/settings.ts
```diff
@@ -15,6 +15,7 @@ export const settings: SimpleTranslationEntries = {
   "skipSeenDialogues": "Pular Diálogos Vistos",
   "battleStyle": "Estilo de Batalha",
   "enableRetries": "Habilitar Novas Tentativas",
+  "hideIvs": "Hide IV scanner",
```
josericardo-fo, line 18:
```suggestion
  "hideIvs": "Esconder scanner de IV",
```

src/locales/ko/settings.ts
```diff
@@ -15,6 +15,7 @@ export const settings: SimpleTranslationEntries = {
   "skipSeenDialogues": "본 대화 생략",
   "battleStyle": "시합 룰",
   "enableRetries": "재도전 허용",
+  "hideIvs": "Hide IV scanner",
```
EnochG1, line 18:
```suggestion
  "hideIvs": "개체값탐지기 메시지 숨김",
```

sodaMelon, line 18:
```suggestion
  "hideIvs": "개체값 탐지기 숨기기",
```

sodaMelon: @EnochG1 how about "개체값 탐지기 숨기기"?  (This is shorter)

sodaMelon: Looking at the second option, I think we need to be a bit more comprehensive. (not only message off)

src/locales/zh_CN/settings.ts
```diff
@@ -15,6 +15,7 @@ export const settings: SimpleTranslationEntries = {
   "skipSeenDialogues": "跳过已读对话",
   "battleStyle": "对战模式",
   "enableRetries": "允许重试",
+  "hideIvs": "Hide IV scanner",
```
mercurius-00, line 18:
```suggestion
  "hideIvs": "禁用个体值探测器",
```

src/locales/zh_TW/settings.ts
```diff
@@ -15,6 +15,7 @@ export const settings: SimpleTranslationEntries = {
   "skipSeenDialogues": "跳過已讀對話",
   "battleStyle": "對戰模式",
   "enableRetries": "允許重試",
+  "hideIvs": "Hide IV scanner",
```
mercurius-00, line 18: 禁用個體值探測器

src/locales/ko/settings.ts
EnochG1: I modified it

frutescens (changes requested):
Mild nit - I noticed that a Best IV would be gold, but I feel like the colors should match up 
image

Opaque02:
> Mild nit - I noticed that a Best IV would be gold, but I feel like the colors should match up image

@frutescens this is a very good point but it was an intentional design choice to have "Best" be different because in the early days of this, people were giving feedback that if they had the IV scanner setting off they wouldn't know what IVs were best and wanted an indicator to know, since that might sway people to catch a pokemon if it's got a Best stat that you don't already have

frutescens:
I meant that the 'Best' used in IV scanner results on the right should be yellow, my bad!

Opaque02:
> I meant that the 'Best' used in IV scanner results on the right should be yellow too, my bad!

Ooooh, sorry, I'm with you! Maybe! I'll bring it up in the discord and see if we can get other ideas for it! Good thinking!

frutescens (approved):
thank you for putting in my suggestion! i know you worked hard on finding the right look :)

EnochG1:
@Opaque02 
Could you merge latest beta and add Japanese locale files? 🙏

Opaque02:
> @Opaque02 From now Japanese is added in beta. Could you merge latest beta and add Japanese locale files? 🙏

Done with the latest commit

PR #3289, opened 2024-08-01
[Balance][QoL] Prevent challenge-ineligible custom evolutions

Fontbane:
## What are the changes?

- Overrides custom evolution methods for split evolutions so they don't render mons ineligible for the current challenge
- For example, a male Kirlia in a mono Fairy run will evolve into Gardevoir at level 30, as the player would expect given main series mechanics, rather than Gallade as it usually would in Pokerogue
- I have another, bigger PR in the works for evolutions, so the evolutions that would be changed there (for example Alola Forms) are left alone for now (their custom methods are harder to trigger anyway)

- Pokerogue has a handful of evolution methods changed from the main series to be more convenient or accessible, specifically split evolutions
- This includes a few Hisuian forms, as well as some cross-gen split evos
- This results in some cases where a player may unknowingly bring a Pokemon or allow an evolution that would render that mon ineligible
- For example, if a player brings a Rufflet in a Gen 5 or mono Normal run and it evolves at night, it will evolve into Hisuian Braviary, which would then be ineligible for the challenge, which the player would not expect to happen. If the player does know about this, they have to delay its evolution until daytime.
- Many of these changes were made for convenience/QoL but in practice it creates a big inconvenience for challenges and could ruin a player's run. It's one thing if a player brings a male Combee by mistake, but we can't expect the player to be aware of these custom evolutions.

## What did change?

- In each of these custom evolutions, the evolution the player would expect to occur due to mainline mechanics (the "normal" one) will check for a specific challenge if it doesn't already meet the custom conditions we set, taking precedence over the custom, potentially ineligible evolution
- The following mons have their evo methods changed by this, listed with the problematic condition/challenge combo:
  - Quilava -> Typhlosion (night, gen 2)
  - Wurmple -> Silcoon (night, mono Flying)
  - Wurmple -> Cascoon (day, mono Poison)
  - Kirlia -> Gardevoir (male, gen 3, mono Fairy)
  - Snorunt -> Glalie (female, gen 3)
  - Dewott -> Samurott (night, gen 5)
  - Rufflet -> Braviary (night, gen 5, mono Normal)
  - Petilil -> Lilligant (night, gen 5)
  - Goomy -> Sliggoo (night, gen 6)
  - Bergmite -> Avalugg (night, gen 6)
  - Dartrix -> Decidueye (night, gen 7, mono Ghost)

## How to test the changes?

- Start a relevant challenge run with an effected mon

## Checklist
- [] Have I considered writing automated tests for the issue?
- [ ] Are the changes visual?

### DIFF
--- a/src/data/pokemon-evolutions.ts
+++ b/src/data/pokemon-evolutions.ts
@@ -11,6 +11,7 @@ import { Biome } from "#enums/biome";
 import { Moves } from "#enums/moves";
 import { Species } from "#enums/species";
 import { TimeOfDay } from "#enums/time-of-day";
+import { SingleGenerationChallenge, SingleTypeChallenge } from "./challenge";
 
 export enum SpeciesWildEvolutionDelay {
   NONE,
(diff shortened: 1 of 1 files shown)


Tempo-anon:
Will this apply the other way? Like a gen 8 challenge with Quilava forcing HTyphlosion?

Fontbane:
This is poorly implemented, I will close.

PR #3291, opened 2024-08-02
[Balance] [Beta] Give Dipplin Dragon Cheer as a relearn move

Blitz425:
## What are the changes?
Dipplin gets access to Dragon Cheer via Memory Mushroom
Applin's Dragon Cheer Egg Move is changed to Matcha Gotcha

As of right now, players are reliant on the Egg Move to evolve their Applin into the very commitment-heavy Hydrapple (Syrupy Apple is a different rarity from the other Apple items for some strange reason, then to evolve into Hydrapple you would have to luck out and get the TM within the same run which requires you to EVOLVE Applin first since Applin doesn't learn it OR the player must have the Egg Move unlocked. All while Appletun / Flapple can get their evolutions relatively early on and have access to their G-Max Forms.) Stantler currently has Psyshield Bash as a custom level up move when in mainline it is an egg move normally, so initially I would have wanted to make Dragon Cheer a level up move but after discussion with Damo and the other balance members we decided it would be better to have it via Memory Mushroom.

## What did change?
- Changed Applin's Learnset
- Changed Applin's Egg Move

### Screenshots/Videos
image
image

## How to test the changes?
Use a Memory Mushroom on Dipplin

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/pokemon-level-moves.ts
+++ b/src/data/pokemon-level-moves.ts
@@ -17287,7 +17287,7 @@ export const pokemonSpeciesLevelMoves: PokemonSpeciesLevelMoves = {
   ],
   [Species.DIPPLIN]: [
     [ EVOLVE_MOVE, Moves.DOUBLE_HIT ],
-    [ RELEARN_MOVE, Moves.INFESTATION ],
+    [ RELEARN_MOVE, Moves.DRAGON_CHEER ], // Custom
     [ 1, Moves.WITHDRAW ],
     [ 1, Moves.SWEET_SCENT ],
     [ 1, Moves.RECYCLE ],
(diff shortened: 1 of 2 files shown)


damocleas (approved):

src/data/pokemon-level-moves.ts
```diff
@@ -17288,6 +17288,7 @@ export const pokemonSpeciesLevelMoves: PokemonSpeciesLevelMoves = {
   [Species.DIPPLIN]: [
     [ EVOLVE_MOVE, Moves.DOUBLE_HIT ],
     [ RELEARN_MOVE, Moves.INFESTATION ],
+    [ RELEARN_MOVE, Moves.DRAGON_CHEER ],
```
Tempo-anon, line 17291: Add a comment here to indicate that this is because relying on a TM for an evolution is impractical

Tempo-anon (commented):
If this isn't a major hotfix this should be aimed at beta

PR #3299, opened 2024-08-02
[Bug] Rockruff, lycanroc wild spawn fix

Smewkie:
## What are the changes?
There's a few different changes here.

Change 1: Allows Own Tempo Rockruff to spawn in the wild, at a rate of 20%, so 20% of all rockruff's you might find can have Own Tempo Form, this allows for somewhat easier access to a pokemon that has the same base stats as its other forms, just put together differently.

Change 2: I changed Dusk Form Lycanroc from being form index 2, to form index 1, this makes it so if you run into a wild Dusk Form Lycanroc and haven't yet unlocked Own Tempo Rockruff, catching this Lycanroc will unlock Own Tempo Rockruff.

Change 3: I changed Lycanroc's forms from spawning based on Time of Day(including any Lycanroc owned by a enemy trainer), and instead just made it based on a % chance, the rates are 40% for Day form, 40% for Midnight form, and 20% for Dusk form.
I personally think it's silly to have the forms locked to a certain time of day in the first place, as Lycanroc itself isn't exactly a super strong pokemon to begin with.

Before, Own Tempo Rockruff couldn't spawn in the wild at all, and in order to get access to it you had to catch a midnight form Lycanroc.

### Screenshots/Videos
Midnight:
image

Dusk:
image

Day:
image

Own Tempo in the wild:
image

## How to test the changes?
Change your OPP_SPECIES_OVERRIDE to either 744 for Rockruff or 745 for Lycanroc
If on a fresh save file, you can see that catching a Dusk Lycanroc does indeed unlock Own Tempo Rockruff, while catching Midnight or Day Form only unlocks normal Rockruff.
And you can see that Lycanroc's forms can all spawn no matter the time of day, just based on random chance.

also sorry I really cannot figure out the whole cherry picking thing D:

## Checklist
- [x] Are the changes visual?


index.css
Tempo-anon, line 1: Pretty sure this file is unrelated to your fix so you should make sure your branch is up to date with beta

src/battle-scene.ts
```diff
@@ -1258,6 +1258,26 @@ export default class BattleScene extends SceneBase {
...
     case Species.INDEEDEE:
     case Species.BASCULEGION:
+    case Species.ROCKRUFF:
+      const funnynumber1 = Utils.randSeedInt(100, 1); // roll between 1 - 100
+      if (funnynumber1 <= 80) {
+        return 0;  // Normal form - 80% chance
+      } else if (funnynumber1 <= 100) {
+        return 1; // Own Tempo form - 20% chance
+      } else {
+      return "Blame Yda!" // just incase the random number goes haywire
```
Tempo-anon, line 1268: This is an error since it's trying to return a `string` when it's expecting a `number`

I imagine you want something like this
```ts
console.log("Failed to generate random number for Rockruff");
return 0 // Default form
```

src/battle-scene.ts
```diff
@@ -1258,6 +1258,26 @@ export default class BattleScene extends SceneBase {
     case Species.MEOWSTIC:
     case Species.INDEEDEE:
     case Species.BASCULEGION:
+    case Species.ROCKRUFF:
+      const funnynumber1 = Utils.randSeedInt(100, 1); // roll between 1 - 100
```
Tempo-anon, line 1262: randSeedInt(range, min) is not inclusive of range so this would be 1-99.

src/battle-scene.ts
```diff
@@ -1258,6 +1258,26 @@ export default class BattleScene extends SceneBase {
...
+      if (funnynumber1 <= 80) {
+        return 0;  // Normal form - 80% chance
+      } else if (funnynumber1 <= 100) {
+        return 1; // Own Tempo form - 20% chance
+      } else {
```
Tempo-anon, line 1267: This can be inline'd
```ts
return (Utils.randSeedInt(100) < 80) ? 1 : 0;
```

Tempo-anon (changes requested):
1. Code doesn't compile
2. Your branch is not up to date
3. Solution for the forms can be more elegant

src/battle-scene.ts
```diff
@@ -1258,6 +1258,26 @@ export default class BattleScene extends SceneBase {
...
+      return "Blame Yda!" // just incase the random number goes haywire
+      }
+    case Species.LYCANROC:
+      const funnynumber2 = Utils.randSeedInt(100, 1); // roll between 1 - 100
```
DayKev, line 1271: Variables should probably have normal/descriptive names.

src/battle-scene.ts
```diff
@@ -1258,6 +1258,26 @@ export default class BattleScene extends SceneBase {
...
+      if (funnynumber2 <= 40) {
+        return 0;  // Day Form - 40% chance
+      } else if (funnynumber2 <= 80) {
+        return 2; // Midnight Form - 40% chance
+      } else if (funnynumber2 <= 100) {
+        return 1; // Dusk Form - 20% chance
+      } else {
+      return "Blame flx!" // just incase the random number goes haywire
```
DayKev, line 1279: Do we need a condition here for if `randSeedInt()` fails? It seems like rSI should be catching any errors itself.

Smewkie: gonna be honest, I just copy / pasted all this in after flx wrote it all, but since it's still not right, I don't really have the knowledge to actually make it better, so I'll go ahead and close this pr

PR #3306, opened 2024-08-02
[QoL] Move victini to gen 5

KimJeongSun:
## What are the changes?

Made sure that Victini, a Gen 5 Pokémon temporarily placed in Gen 4, is correctly filtered as a Gen 5 Pokémon.
Victini, a Generation 5 Pokémon that was being filtered under Generation 4, is now correctly filtered under Generation 5.

Since the addition of scroll functionality to the starter UI, 
there's no longer a reason to incorrectly place Victini in Gen 4. 
This can also cause confusion among users.

## What did change?

- Updated Victini's species Gen information from 4 to 5.
- Removed the no longer necessary workaround code for Victini in the challenge checks.
- However, since Victini's icon sprite is still included in Gen 4, I've added a temporary workaround to map it correctly. (This code should be removed once the Victini icon sprite is moved to Gen 5. )

### Screenshots/Videos

https://github.com/user-attachments/assets/09d9361b-4bcc-4c3a-9785-83c004a7fcee

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/pokemon-species.ts
+++ b/src/data/pokemon-species.ts
@@ -312,7 +312,13 @@ export abstract class PokemonSpeciesForm {
   getIconAtlasKey(formIndex?: integer, shiny?: boolean, variant?: integer): string {
     const variantDataIndex = this.getVariantDataIndex(formIndex);
     const isVariant = shiny && variantData[variantDataIndex] && variantData[variantDataIndex][variant];
-    return `pokemon_icons_${this.generation}${isVariant ? "v" : ""}`;
+    const species = getPokemonSpecies(this.speciesId);
+    /**
+     * Currently, Victini's icon sprite is located in Gen 4, so we will temporarily bypass this issue with the following code.
+     * In the future, when Victini's icon sprite is properly moved to Gen 5, the temporary code below should be removed.
+     */
+    const generation = species.speciesId === Species.VICTINI ? 4 : this.generation;
+    return `pokemon_icons_${generation}${isVariant ? "v" : ""}`;
   }
 
   getIconId(female: boolean, formIndex?: integer, shiny?: boolean, variant?: integer): string {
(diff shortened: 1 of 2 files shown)


src/data/pokemon-species.ts
```diff
@@ -312,7 +312,13 @@ export abstract class PokemonSpeciesForm {
   getIconAtlasKey(formIndex?: integer, shiny?: boolean, variant?: integer): string {
     const variantDataIndex = this.getVariantDataIndex(formIndex);
     const isVariant = shiny && variantData[variantDataIndex] && variantData[variantDataIndex][variant];
-    return `pokemon_icons_${this.generation}${isVariant ? "v" : ""}`;
+    const species = getPokemonSpecies(this.speciesId);
+    /**
+     * Currently, Victini's icon sprite is located in Gen 4, so we will temporarily bypass this issue with the following code.
+     * In the future, when Victini's icon sprite is properly moved to Gen 5, the temporary code below should be removed.
+     */
```
Tempo-anon, line 319: How difficult is it to implement this? Is it just moving `public/images/pokemon/icons/4/494.png` and `public/images/pokemon/icons/4/494s.png` to the 5 directory?

KimJeongSun: Ah, I hadn't seen this part before and I looked into it here and there. 
If it was about moving assets, I would have used that method as well. 

The key point here is that the icon sprites for each generation are contained within a single PNG file for that generation's Pokémon.

To modify this, We would need to create new PNG files for generations 4 and 5 and generate new frame information. 

I did consider this for about a second, but it was impossible for me to do so.

KimJeongSun: I was thought it looks related with this [file](https://github.com/pagefaultgames/pokerogue/blob/main/public%2Fimages%2Fpokemon_icons_4.png) which is included every pokemon in this Gen 4.

but I will double check for this.

Tempo-anon: Ah you might have to also do something there too. Sam really leaving us with a lot of tech debt huh

KimJeongSun: Hmm, if possible, it would be great to get some help from the art team for this part in the future PR.
What I can do is update the frame information once I receive the newly updated icons image, and then remove the temporary bypass code.

Honestly, my image processing skills are limited to playing around with OpenCV or basic tools like MS Paint. 😓

KimJeongSun: https://discord.com/channels/1125469663833370665/1267339517371875361/1269139698203230270

i think this PR's approach is easy solution currently

PR #3310, opened 2024-08-03
Refactor duplicated code into a common Pokemon#leaveField() function

snoozbuster:
## What are the changes?
In a number of places, including `SummonPhase`, `SwitchSummonPhase`, `ForceSwitchOutAttr`, and `TurnInitPhase`, extremely similar logic is performed manually to cause a mon to leave the field. Some of this logic was re-used for player-pokemon switches via `PlayerPokemon#switchOut`, but I refactored the parts relevant to both player and enemy mons into a single function in the base Pokemon class, called `leaveField()`. 

DRY. Also, I think it might partially fix baton pass when used by the AI - notes inline below.

## What did change?

Ideally, very little.

### Screenshots/Videos

N/A

## How to test the changes?

Run the tests

## Checklist
- [ ] ~Are the changes visual?~

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -13,10 +13,9 @@ import { ArenaTagSide, ArenaTrapTag, WeakenMoveTypeTag } from "./arena-tag";
 import { UnswappableAbilityAbAttr, UncopiableAbilityAbAttr, UnsuppressableAbilityAbAttr, BlockRecoilDamageAttr, BlockOneHitKOAbAttr, IgnoreContactAbAttr, MaxMultiHitAbAttr, applyAbAttrs, BlockNonDirectDamageAbAttr, applyPreSwitchOutAbAttrs, PreSwitchOutAbAttr, applyPostDefendAbAttrs, PostDefendContactApplyStatusEffectAbAttr, MoveAbilityBypassAbAttr, ReverseDrainAbAttr, FieldPreventExplosiveMovesAbAttr, ForceSwitchOutImmunityAbAttr, BlockItemTheftAbAttr, applyPostAttackAbAttrs, ConfusionOnStatusEffectAbAttr, HealFromBerryUseAbAttr, IgnoreProtectOnContactAbAttr, IgnoreMoveEffectsAbAttr, applyPreDefendAbAttrs, MoveEffectChanceMultiplierAbAttr, WonderSkinAbAttr, applyPreAttackAbAttrs, MoveTypeChangeAttr, UserFieldMoveTypePowerBoostAbAttr, FieldMoveTypePowerBoostAbAttr, AllyMoveCategoryPowerBoostAbAttr, VariableMovePowerAbAttr } from "./ability";
 import { allAbilities } from "./ability";
 import { PokemonHeldItemModifier, BerryModifier, PreserveBerryModifier, PokemonMoveAccuracyBoosterModifier, AttackTypeBoosterModifier, PokemonMultiHitModifier } from "../modifier/modifier";
-import { BattlerIndex } from "../battle";
+import { BattlerIndex, BattleType } from "../battle";
 import { Stat } from "./pokemon-stat";
 import { TerrainType } from "./terrain";
-import { SpeciesFormChangeActiveTrigger } from "./pokemon-forms";
 import { ModifierPoolType } from "#app/modifier/modifier-type";
 import { Command } from "../ui/command-ui-handler";
 import i18next from "i18next";
@@ -1817,13 +1816,10 @@ export class MultiHitAttr extends MoveAttr {
     }
     case MultiHitType._2:
       return 2;
-      break;
     case MultiHitType._3:
       return 3;
-      break;
     case MultiHitType._10:
       return 10;
-      break;
     case MultiHitType.BEAT_UP:
       const party = user.isPlayer() ? user.scene.getParty() : user.scene.getEnemyParty();
       // No status means the ally pokemon can contribute to Beat Up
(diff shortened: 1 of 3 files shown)


src/data/move.ts
```diff
@@ -1817,13 +1816,10 @@ export class MultiHitAttr extends MoveAttr {
     }
     case MultiHitType._2:
       return 2;
-      break;
```
snoozbuster, line 1820: autofix by IDE

src/field/pokemon.ts
```diff
@@ -3272,25 +3289,21 @@ export class PlayerPokemon extends Pokemon {
     return true;
   }
 
-  switchOut(batonPass: boolean, removeFromField: boolean = false): Promise<void> {
```
snoozbuster, line 3275: I removed the second parameter from this function - it seemed unused.

src/data/move.ts
```diff
@@ -4780,19 +4776,14 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
...
-	  	// Switch out logic for the battle type
-	  	switchOutTarget.resetTurnData();
-	  	switchOutTarget.resetSummonData();
-	  	switchOutTarget.hideInfo();
-	  	switchOutTarget.setVisible(false);
-	  	switchOutTarget.scene.field.remove(switchOutTarget);
-	  	user.scene.triggerPokemonFormChange(switchOutTarget, SpeciesFormChangeActiveTrigger, true);
+	  } else if (user.scene.currentBattle.battleType !== BattleType.WILD) {
+	  	// Switch out logic for trainer battles
+        switchOutTarget.leaveField(!this.batonPass);
```
snoozbuster, line 4786: there is some weird spaces/tabs issue in this file I'm not sure how to fix, which causes this weird indentation here in GH. it doesn't seem to be a new problem.

snoozbuster: Additionally, this line effects a minor change - the `batonPass` flag will be respected here for enemy pokemon now (causing summon/battle data not to be reset), which was not the case before.

src/field/pokemon.ts
```diff
@@ -3161,6 +3161,23 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+  /**
+   * Causes a Pokemon to leave the field (such as in preparation for a switch out/escape).
+   * @param clearEffects Indicates if effects should be cleared (true) or passed
+   * to the next pokemon, such as during a baton pass (false)
+   */
+  leaveField(clearEffects: boolean = true) {
+    this.resetTurnData();
+    if (clearEffects) {
+      this.resetSummonData();
+      this.resetBattleData();
```
snoozbuster, line 3173: I'm not sure if it was correct, but I added clearing battle data to this function. Previously, I saw it happen in SwitchSummonPhase inside `summon`, and based on the data in the structure it didn't seem like this would have any negative effects.

src/field/pokemon.ts
```diff
@@ -3272,25 +3289,21 @@ export class PlayerPokemon extends Pokemon {
...
-      this.setVisible(false);
+      this.leaveField(!batonPass);
 
       this.scene.ui.setMode(Mode.PARTY, PartyUiMode.FAINT_SWITCH, this.getFieldIndex(), (slotIndex: integer, option: PartyOption) => {
         if (slotIndex >= this.scene.currentBattle.getBattlerCount() && slotIndex < 6) {
           this.scene.prependToPhase(new SwitchSummonPhase(this.scene, this.getFieldIndex(), slotIndex, false, batonPass), MoveEndPhase);
         }
-        if (removeFromField) {
-          this.setVisible(false);
-          this.scene.field.remove(this);
```
snoozbuster, line 3290: technically, these two lines now happen earlier than before (pre-party menu, instead of after). but since `setVisible` runs before any of this anyway, I don't think this should be an issue. I didn't observe any issues with switching, although I didn't check for issues with switching with a form change active.

src/phases.ts
```diff
@@ -1653,11 +1650,6 @@ export class SwitchSummonPhase extends SummonPhase {
             pokemonName: this.getPokemon().getNameToRender()
           })
         );
-        // Ensure improperly persisted summon data (such as tags) is cleared upon switching
-        if (!this.batonPass) {
```
snoozbuster, line 1657: moved this into leaveField(). didn't see a reason why it needed to happen in `switchAndSummon()` instead of in `preSummon()` - at least, didn't see any way to cancel a summon or anything.

DayKev: That may be a better place for it, I don't recall the reason for putting it here specifically. If the issue fixed in #3119 is still fixed then it should be fine.

snoozbuster: it seems like the bug from that PR did come back without this code. `party[this.fieldIndex]` refers to the pokemon entering the field, not the one leaving the field - different thing. I put it back.

I do sort of wonder why this code is necessary here, on the switch in, after saving and reloading though. mystery for another day, I guess

DayKev: Yeah it was really weird; at least part of the problem is most likely that saving is happening at the wrong time (ie: before all the data being saved is fully up-to-date with the current game state). Messing with the saving code is not something I'm confident in though.

src/field/pokemon.ts
DayKev: Tested with Eiscue (with and without Ice Face active) and it seemed to work fine. I didn't notice any abnormal behavior.

DayKev (approved):

DustinLin:
Tested doubles, as well as fainting u turn bugfix and all still seems to work fine

snoozbuster:
@DustinLin can you tell me what the fainting u turn case was? I saw the code that was intended to prevent it, but I wasn't sure what scenarios it was triggered in

PR #3315, opened 2024-08-03
[Bug] Orichalcum Pulse and Hadron Engine should be unswappable

DayKev:
## What are the changes?
Skill Swap now fails when used if either Pokémon has Hadron Engine or Orichalcum Pulse.

cf https://bulbapedia.bulbagarden.net/wiki/Orichalcum_Pulse_(Ability) and https://bulbapedia.bulbagarden.net/wiki/Hadron_Engine_(Ability)
Also tested manually in SV to make sure, and it's correct.

## What did change?
`UnswappableAbilityAbAttr` was added to Hadron Engine and Orichalcum Pulse.

## How to test the changes?
Use Skill Swap against Koraidon or Miraidon.

## Checklist

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -5259,11 +5259,13 @@ export function initAbilities() {
     new Ability(Abilities.ORICHALCUM_PULSE, 9)
       .attr(PostSummonWeatherChangeAbAttr, WeatherType.SUNNY)
       .attr(PostBiomeChangeWeatherChangeAbAttr, WeatherType.SUNNY)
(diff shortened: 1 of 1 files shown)


torranx:
Is it supposed to only block Skill Swap? How about Role play/Entrainment?

DayKev:
Confirmed via testing, they're also uncopiable.

DayKev:
According to someone who has a more up-to-date version of SV than I do, Role Play et al work on Hadron Engine and Orichalcum Pulse. For whatever reason GF seems to have changed it at some point. 🤷 
So... nevermind I guess??

PR #3312, opened 2024-08-03
[QoL] Auto filtering on challenge mode

KimJeongSun:
## What are the changes?

You no longer need to spend endless time searching for valid Pokémon in the starter select UI during challenge mode.

## Why am I doing these changes?

I think this approach is way better. Does anyone disagree?

## What did change?

In challenge mode, valid Pokémon will be pre-filtered according to the game mode.
( + The variable name `starterContainer` could be confusing, so I've corrected it to `starterContainers`. )

### Screenshots/Videos

https://github.com/user-attachments/assets/92442c3d-66de-4d78-bf67-5169601d8f83

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [ ] Have I considered writing automated tests for the issue?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #2916, opened 2024-07-08
[QoL] New Starter Select UI with Filter

KimJeongSun:
## What are the changes?

Add filter in starter select UI.
Most updates were cherry-picked from the previous [PR](https://github.com/pagefaultgames/pokerogue/pull/2036) with @FredeX's consent. 
This includes some bug fixes, test code updates, and minor code cleanups.

## Why am I doing these changes?

current starter UI is too difficult to find or check the over 500 pokemon.

## What did change?

added basic filters and merged the existing generation UI into the filter.

### Screenshots/Videos

https://github.com/user-attachments/assets/c5dbfedd-13f9-4318-a588-4f6cf913e232

This video has been updated based on this commit version. [8b64de5](https://github.com/pagefaultgames/pokerogue/pull/2916/commits/8b64de5119c7b54ba516ee701205ba697185387d)(latest)

## How to test the changes?

can test using the updated test code matching the updated UI.
`npm run test ui/starter-select`

## Checklist
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #3325, opened 2024-08-03
[Enhancement] Improvements to starter selection and filtering user experience

MokaStitcher:
## What are the changes?
Change the generation filter to an "hybrid" filter that allows both to pick any combination of gens like currently and to switch quickly from one gen to the other like before filters were introduced.
A good amount of refactoring, adding documentation and improvements on the "backend" filter code was involved as well.

Following PR #2916  there was a lot of negative feedback about how navigating through all Pokemon with the new UI took way too many inputs compared to previously.
By adding this hybrid behavior we hopefully have the best of both worlds: fast navigation between gens while still being able to filter specific gens in any way we want to.
A good amount of refactoring was pretty much needed to do the changes as the previous implementation for the filters left a lot of margin for error when creating filters. The changes should allow us to add more filters in the future way more easily.

### New hybrid type of filtering
* When all options in this filter are unselected, the option that is currently in focus will be used to filter instead of nothing. While in this state, moving from an option to the next with up/down will allow to quickly go through all options without any extra input.
* When selecting any option or more the behavior is the same as before
* The generation, type and shiny filters now all use this filtering
* For the generation filter, by default all is unselected and the cursor is on generation 1
* The generation filter has been moved to the left of the filter bar so that when going through the generations in this way the Pokemon are not hidden by the filters, effectively allowing to go through all generations quickly, as before the filters were introduced
* For the type and shiny filters, this behavior is not enabled by default

### UI updates
* Make the Filter Bar fill all available horizontal space
* Make the Starter Team fill up all available vertical space
* Update arrow navigation between filter bar, team and possible starters to work with these changes
* Automatically move cursor to the first starter when closing the filters
* Fix the coloring of modified filters for legacy theme
* Set the color of enabled filter options to blue instead of green to avoid green/red combination for color blindness
* Set the size of the multi state filters from PR #3252 to the widest of options in the filters, instead of having the container resize itself when toggling the filter
* Rename the Filter categories based on future plans for filters and add missing localization strings

### Other QoL stuff
* Reset all filters to default when creating a new game. Before the previous filters were kept for the whole session
* For challenges, all generations are selected by default instead of just generation 1. Along with PR #3312 all valid Pokemon are now properly shown when creating a new challenge

### "Backend" filter improvements that will help for future updates
* Helper functions to select or unselect all elements in a filter category
* Update the toggle filters introduced in PR #3252 to allow for any number of options instead of just 3
* Filter options can now display both a sprite and text, while before it was only one or the other

## Screenshots/Videos

(all before videos and screenshots are taken on current beta)

### UI update

Before (default theme):
_ui-default-before

After (default theme):
_ui-default-after

Before (legacy theme):
_ui_legacy_before

After (legacy theme):
_ui-legacy-after

### New filtering for Generation, Type and  Shiny

Before: trying to go through one gen at a time

https://github.com/user-attachments/assets/8050f0f0-cd60-4917-b232-9f9bb50de493

After: going through one gen at a time

https://github.com/user-attachments/assets/5f913d61-ebe0-4db7-895d-81218db6f211

After: showcasing the hybrid behavior of the new filter type

https://github.com/user-attachments/assets/b51c89d0-10c3-4c78-9203-05e5049fa50a

### Filter reset when creating new game

Before:

https://github.com/user-attachments/assets/0be4907b-ea9f-4649-b2b7-3cf99dd9257a

After

https://github.com/user-attachments/assets/a61259a2-1205-4875-9b88-9711f81c2e90

### Multi toggle filters no longer change size

Before:

https://github.com/user-attachments/assets/9d5faca9-093f-408a-9c0a-d45eef23be13

After:

https://github.com/user-attachments/assets/31f5610c-60df-4cdf-8145-cc6346d0cdf6

## How to test the changes?

- Create a new game in any mode and play around with the filters
- Try moving around between the UI elements in various states of filtering and team composition
- Exit and go back to creating a new game to make sure filters are properly reset

I haven't added unit tests for the DropDown and FilterBar classes, but it's definitely something that would be beneficial and that I'd like to add hopefully in the near future. But for now I'd rather have all the QoL improvements live regardless of those tests being in, if possible. Even moreso because some of those changes are blocking further updates to the filters mentioned

## Future Work

Putting here for reference, as this PR is already way too big for its own good

- Add more filters for Pokerus, hidden ability, and other types of unlocks [(see what's planned here)](https://discord.com/channels/1125469663833370665/1267339517371875361/1268014543511224360)
- Add shortcut to open the filter menu from anywhere
- When in filter menu, add shortcuts to select all and unselect all elements
- Update the look of hybrid filter to make the current state clearer
- Probably more that I forget and will add when I remember :') 

## Checklist
- [X] The PR is self-contained and cannot be split into smaller PRs? (not really due to heavy refactoring)
- [ ] Have I considered writing automated tests for the issue? (considered? yes. done? no) 
- [X] Are the changes visual?


torranx (commented):
code looks good to me but needs thorough testing. thanks for adding docs 👍

Adri1:
Why not calling directly "Unlocks" "Passives", and "Misc" "Ribbons" as they are the only things represented in each of their respective categories?
And it won't lie, it would make some translations less cranky, because for example "Débloqu." for French as a translation for "Unlocks" is really not great

src/locales/ko/filter-bar.ts
```diff
@@ -3,16 +3,19 @@ import { SimpleTranslationEntries } from "#app/interfaces/locales";
...
   "normal": "기본",
   "uncaught": "미포획",
+  "passive": "Passive",
   "passiveUnlocked": "패시브 해금",
   "passiveLocked": "패시브 잠김",
-  "hasWon": "완료",
-  "hasNotWon": "미완료",
+  "ribbon": "Ribbon",
+  "hasWon": "Ribbon - Yes",
+  "hasNotwon": "Ribbon - No",
```
EnochG1, line 18:
```suggestion
  "ribbon": "클리어 여부",
  "hasWon": "클리어 함",
  "hasNotwon": "클리어 안함",
```

src/locales/ko/filter-bar.ts
```diff
@@ -3,16 +3,19 @@ import { SimpleTranslationEntries } from "#app/interfaces/locales";
 export const filterBar: SimpleTranslationEntries = {
   "genFilter": "세대",
   "typeFilter": "타입",
+  "dexFilter": "Dex",
   "unlocksFilter": "등록",
```
EnochG1, line 7:
```suggestion
  "dexFilter": "도감",
  "unlocksFilter": "해금",
```

src/locales/ko/filter-bar.ts
```diff
@@ -3,16 +3,19 @@ import { SimpleTranslationEntries } from "#app/interfaces/locales";
...
-  "winFilter": "클리어",
+  "miscFilter": "Misc",
   "sortFilter": "정렬",
   "all": "전체",
   "normal": "기본",
   "uncaught": "미포획",
+  "passive": "Passive",
```
EnochG1, line 13:
```suggestion
  "miscFilter": "기타",
  "sortFilter": "정렬",
  "all": "전체",
  "normal": "기본",
  "uncaught": "미포획",
  "passive": "패시브",
```

MokaStitcher:
@Adri1 I've been calling them médaille myself but I could totally be wrong on that one. I should've have double checked how they are called in the achievements, sorry. Feel free to make a suggestion to replace these if ruban is indeed used elsewhere in the game.
As for the other issue, I agree that unlocks is really hard to translate, I welcome any idea for a better idea. But as described in the PR there are plans to add more filters in all these categories, so the titles and categories are set in regards to those and won't be merged unless those plans change

src/locales/ko/filter-bar.ts
EnochG1: It's not final version but for prevent mixed language.

Adri1:
@MokaStitcher Yeah I translated the "before" version of this UI and I never managed to find something satisfying for "Unlocks", so I used to call it "Autres" by default as my brazilian colleague also did, after struggling to find a translation that is acceptable. So I think it's gonna be difficult for this one.
Maybe we'll come up with someting better as new filters will be added, depending on their natures.
But if these new filters are not too imminent, I think it would be better to call these catergories by the only option they have as long as they are the only ones, but that's just my opinion

MokaStitcher:
@Adri1 yeah I totally understand why it was translated to Autres, but it gets too confusing when adding "Divers" in 😭
Plans for the other filters are linked in the pr under "future work", maybe it can help to find a better alternative.
The new filters should hopefully be added soon after this is merged but I'll probably take you up on that idea to change the category title if there's only one filter in it, thank you

Adri1:
@MokaStitcher Oh no, don't worry about the "Autres", I totally unstestand that you also had to make choices! 🫡
Okay thanks, I'm gonna dive into it

MokaStitcher:
@Adri1 actually, maybe using "upgrades" instead of "unlocks" could work and mostly mean the same thing. It would definitely be easier to translate. That's a problem for tomorrow me though, thanks for getting me thinking on this again!

KimJeongSun:
I fixed the bug can't open menu when 6 starter selected.

if you're ok, can you cherry pick this into your PR? ( or I can make other PR for it )
https://github.com/KimJeongSun/pokerogue/commit/ff2c233c18f5140df0578f6e797e60439f23556c

KimJeongSun:
I have some opinions about the default values of the hybrid filters:

1) Currently, it seems that both `ALL OFF (hovering on ALL)` and `ALL ON` are treated as defaults. If this was intended, it's fine.
2) I think it would be better to change the init setting of the type filter to `ALL OFF (hovering on ALL)` for the following reasons:
   a. The type filter is usually used to view some specific types or a combination of some types. Therefore, starting from `ALL OFF (hovering on ALL)` is more appropriate.
   b. When the type filter is set to ALL ON like now, it might be suitable for the UX where users want to exclude some specific types. But given the current nature of the type filter, such needs might not be significant.

To give an example of the current initial value for gen and type filters:
gen's init value:  `ALL OFF (hovering on GEN 1)`
type's init value: `ALL ON (hovering on ALL)` ( default )
This is how it is currently.

My suggestion of 2):
gen's init value: `ALL OFF (hovering on Gen 1)`
type's init value: `ALL OFF (hovering on ALL)` (default)
This would be more appropriate.

src/locales/zh_TW/filter-bar.ts
```diff
@@ -3,16 +3,19 @@ import { SimpleTranslationEntries } from "#app/interfaces/locales";
 export const filterBar: SimpleTranslationEntries = {
   "genFilter": "世代",
   "typeFilter": "屬性",
+  "dexFilter": "Dex",
   "unlocksFilter": "解鎖",
-  "winFilter": "通關",
+  "miscFilter": "Misc",
```
40chyan, line 8:
```suggestion
  "miscFilter": "混合",
```

src/locales/zh_TW/filter-bar.ts
```diff
@@ -3,16 +3,19 @@ import { SimpleTranslationEntries } from "#app/interfaces/locales";
...
   "sortFilter": "排序",
   "all": "全部",
   "normal": "通常",
   "uncaught": "未捕獲",
+  "passive": "Passive",
```
40chyan, line 13:
```suggestion
  "passive": "被動",
```

src/locales/zh_TW/filter-bar.ts
```diff
@@ -3,16 +3,19 @@ import { SimpleTranslationEntries } from "#app/interfaces/locales";
...
   "passiveUnlocked": "被動解鎖",
   "passiveLocked": "被動未解鎖",
-  "hasWon": "已通關",
-  "hasNotWon": "未通關",
+  "ribbon": "Ribbon",
+  "hasWon": "Ribbon - Yes",
+  "hasNotWon": "Ribbon - No",
```
40chyan, line 18:
```suggestion
  "ribbon": "緞帶",
  "hasWon": "有緞帶",
  "hasNotWon": "無緞帶",
```

src/locales/zh_CN/filter-bar.ts
```diff
@@ -3,16 +3,19 @@ import { SimpleTranslationEntries } from "#app/interfaces/locales";
 export const filterBar: SimpleTranslationEntries = {
   "genFilter": "世代",
   "typeFilter": "属性",
+  "dexFilter": "Dex",
   "unlocksFilter": "解锁",
-  "winFilter": "通关",
+  "miscFilter": "Misc",
```
40chyan, line 8:
```suggestion
  "miscFilter": "混合",
```

src/locales/zh_CN/filter-bar.ts
```diff
@@ -3,16 +3,19 @@ import { SimpleTranslationEntries } from "#app/interfaces/locales";
...
   "sortFilter": "排序",
   "all": "全部",
   "normal": "无闪光",
   "uncaught": "未捕获",
+  "passive": "Passive",
```
40chyan, line 13:
```suggestion
  "passive": "被动",
```

src/locales/zh_CN/filter-bar.ts
```diff
@@ -3,16 +3,19 @@ import { SimpleTranslationEntries } from "#app/interfaces/locales";
...
   "passiveUnlocked": "被动解锁",
   "passiveLocked": "被动未解锁",
-  "hasWon": "已通关",
-  "hasNotWon": "未通关",
+  "ribbon": "Ribbon",
+  "hasWon": "Ribbon - Yes",
+  "hasNotWon": "Ribbon - No",
```
40chyan, line 18:
```suggestion
  "ribbon": "缎带",
  "hasWon": "有缎带",
  "hasNotWon": "无缎带",
```

40chyan (commented):
zh_cn & zh_tw

src/locales/de/filter-bar.ts
```diff
@@ -3,14 +3,17 @@ import { SimpleTranslationEntries } from "#app/interfaces/locales";
 export const filterBar: SimpleTranslationEntries = {
   "genFilter": "Gen.",
   "typeFilter": "Typ",
+  "dexFilter": "Dex",
```
CodeTappert, line 6:
```suggestion
  "dexFilter": "Dex.",
```

src/locales/de/filter-bar.ts
```diff
@@ -3,14 +3,17 @@ import { SimpleTranslationEntries } from "#app/interfaces/locales";
...
   "unlocksFilter": "Freisch.",
-  "winFilter": "Abschluss",
+  "miscFilter": "Misc",
```
CodeTappert, line 8:
```suggestion
  "miscFilter": "Sonst.",
```

src/locales/de/filter-bar.ts
```diff
@@ -3,14 +3,17 @@ import { SimpleTranslationEntries } from "#app/interfaces/locales";
...
   "sortFilter": "Sort.",
   "all": "Alle",
   "normal": "Normal",
   "uncaught": "Nicht gefangen",
+  "passive": "Passive",
   "passiveUnlocked": "Passive freigeschaltet",
   "passiveLocked": "Passive gesperrt",
+  "ribbon": "Ribbon",
```
CodeTappert, line 16:
```suggestion
  "ribbon": "Band",
```

CodeTappert:
Please once you have added my suggestions have a look if any of the german text is too long and let me know

PR #3343, opened 2024-08-04
[Enhancement] add filters for pokerus, HA, cost reduction

KimJeongSun:
## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -431,8 +431,16 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
       new DropDownLabel(i18next.t("filterBar:passiveUnlocked"), undefined, DropDownState.ON),
       new DropDownLabel(i18next.t("filterBar:passiveLocked"), undefined, DropDownState.EXCLUDE),
     ];
+
+    const costReductionLabels = [
+      new DropDownLabel(i18next.t("filterBar:costReduction"), undefined, DropDownState.OFF),
+      new DropDownLabel(i18next.t("filterBar:costReductionUnlocked"), undefined, DropDownState.ON),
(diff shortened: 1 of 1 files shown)


torranx:
Is the video in the description updated? I see some keys there

KimJeongSun:
actually, it is before adding default english translation, if you want, i will update it

KimJeongSun:
video updated with initial english translation entries updates

src/ui/starter-select-ui-handler.ts
```diff
@@ -2132,13 +2154,25 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
       const fitsPassive = this.filterBar.getVals(DropDownColumn.UNLOCKS).some(unlocks => {
         if (unlocks.val === "PASSIVE" && unlocks.state === DropDownState.ON) {
           return isPassiveUnlocked;
+        } else if (unlocks.val === "PASSIVE" && unlocks.state === DropDownState.ON) {
```
MokaStitcher, line 2157:
```suggestion
```

src/ui/starter-select-ui-handler.ts
```diff
@@ -2132,13 +2154,25 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
...
+
```
MokaStitcher, line 2158:
```suggestion
```

MokaStitcher (changes requested):
Just a couple of lines you forgot to delete but other than that everything seems good!

src/ui/starter-select-ui-handler.ts
KimJeongSun: is it okay?

MokaStitcher: unless I'm missing something this else if has the same exact condition as the base if so it will never be reached. So it should just be removed?

PR #3321, opened 2024-08-03
[Bug] Fix a couple small issues with uturn and friends

snoozbuster:
## What are the changes?
Fixes three main issues with u-turn and company:
- U-turn would trigger regenerator twice on the user
- U-turn would trigger abilities like static twice on the attacker
- The party UI for u-turn and other switch moves would appear in the middle of the move, before effects and abilities were done applying

Note that this _does not_ fix issues with wilds fleeing using uturn, particularly when they flee and activate an ability such as poison point or poison touch at the same time. but it doesn't make it any worse, either - I want to do that in a separate PR

## Why am I doing these changes?
Pre-work for a pursuit implementation

## What did change?
- Removed the manual application of two ability attr types by ForceSwitchOutAttr
- Changed ForceSwitchOutAttr to use SwitchPhase for choosing and sending out a mon, rather than doing it inline in the MoveEffectPhase/apply() function
  - this should drastically improve the stability of the move in edge case scenarios, since the current MoveEffectPhase will complete entirely before the new pokemon is sent out. if additional phases are injected by other effects, though - the phase ordering between moves/abilities may need to be closely analyzed.
- Added UTs for everything that was broken, or that I noticed I had accidentally broke while working on these fixes
  - Added some new helpers/fixes for the UT harnesses to help test switch moves

#### Pre-change: U-turn damage and effects (Stamina) happen _after_ party menu
[screen-capture (4).webm](https://github.com/user-attachments/assets/5a9fed96-40c8-4e17-bee3-17b134580326)

#### Pre-change: U-turn triggers regenerator twice
[screen-capture (5).webm](https://github.com/user-attachments/assets/b0569f2f-b487-40f1-af2a-a4f97221eb4a)

#### Pre-change: U-turn triggers static twice (only with no party members to switch to, because... yeah)
[screen-capture (6).webm](https://github.com/user-attachments/assets/79a7a9c7-735b-43db-88dd-ec632e74f664)

#### Pre-change: Demonstration of U-turn in doubles
[screen-capture (9).webm](https://github.com/user-attachments/assets/78ce009d-a077-4f8b-83d4-e4662666ae07)

#### Post-change: U-turn damage and effects (Stamina) happen _before_ party menu
[screen-capture (1).webm](https://github.com/user-attachments/assets/f11fd89c-903d-4a40-acb5-f5ff4c0938d4)

#### Post-change: Baton pass still passes Work Up
[...]

PR #3346, opened 2024-08-04
[BUG] - dragon tail switchout ability in wild battles proc crashes game

DustinLin:
## What are the changes?

Fixed game freezing bug when force-out moves (commonly dragon tail) in doubles and single wild battles.

Bug often report on discord

## What did change?

Previous code was immediately destroying the forced-out pokemon object, causing any further reference to it (eg: showing its ability) to reference undefined and crash the game.

To remedy this I made a few changes:
- delay the destroying of wild pokemon objects until the end of the battle
- added an attribute to track wild pokemon that fled, this is to ensure that (since the fled pokemon's object is still around the `BattleScene` but not marked as `fainted`, it isn't selected by `getEnemyPokemon()` to move animations. Without this the game would sometimes crash, as Phaser would try to animate moves hitting a non-visible pokemon.
- added move auto-redirection (moves targeting a fled pokemon would redirect to the other in doubles), this is the same behavior as when a pokemon faints, so refactored code.

### Screenshots/Videos

- `beta`

https://github.com/user-attachments/assets/39c8cf24-141a-417c-a4f6-813348bf6c83
original report

- `pr`
doubles battles
https://github.com/user-attachments/assets/a8fc1963-6798-43b9-986e-dacfb8799fdc

https://github.com/user-attachments/assets/c002c4d9-867a-44e6-9f3e-6a6852dfa917

https://github.com/user-attachments/assets/240ac138-5593-4662-b9a1-e298a23c1b53

single battles
https://github.com/user-attachments/assets/bb1a9e8f-72ca-4f62-82e5-c06e69664e85

https://github.com/user-attachments/assets/7a4d1099-49f0-4359-8fbf-994565dee97b

https://github.com/user-attachments/assets/8ff4f70f-fa38-4ac2-84b2-def552db48b8

## How to test the changes?

Commonly reported abilities are "mummy" and "cotton down". There are probably other abilities, but this fix should work for all of them.

This bug ~~should only affect doubles wild battles~~ correction it should only affect *wild* battles

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -4772,13 +4772,18 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
           user.scene.prependToPhase(new SwitchSummonPhase(user.scene, switchOutTarget.getFieldIndex(), user.scene.currentBattle.trainer.getNextSummonIndex((switchOutTarget as EnemyPokemon).trainerSlot), false, this.batonPass, false), MoveEndPhase);
         }
 	  } else {
-	    // Switch out logic for everything else
-	  	switchOutTarget.setVisible(false);
+	    // Switch out logic for everything else (eg: WILD battles)
+	  	switchOutTarget.leaveField(false);
 
 	  	if (switchOutTarget.hp) {
-	  	  switchOutTarget.hideInfo().then(() => switchOutTarget.destroy());
-	  	  switchOutTarget.scene.field.remove(switchOutTarget);
+          switchOutTarget.setWildFlee(true);
 	  	  user.scene.queueMessage(i18next.t("moveTriggers:fled", {pokemonName: getPokemonNameWithAffix(switchOutTarget)}), null, true, 500);
+
+          // in double battles redirect potential moves off fled pokemon
+          if (switchOutTarget.scene.currentBattle.double) {
+            const allyPokemon = switchOutTarget.getAlly();
+            switchOutTarget.scene.redirectPokemonMoves(switchOutTarget, allyPokemon);
+          }
 	  	}
 
 	  	if (!switchOutTarget.getAlly()?.isActive(true)) {
@@ -7407,7 +7412,8 @@ export function initMoves() {
     new AttackMove(Moves.FROST_BREATH, Type.ICE, MoveCategory.SPECIAL, 60, 90, 10, 100, 0, 5)
       .attr(CritOnlyAttr),
     new AttackMove(Moves.DRAGON_TAIL, Type.DRAGON, MoveCategory.PHYSICAL, 60, 90, 10, -1, -6, 5)
-      .attr(ForceSwitchOutAttr),
+      .attr(ForceSwitchOutAttr)
+      .hidesTarget(),
     new SelfStatusMove(Moves.WORK_UP, Type.NORMAL, -1, 30, -1, 0, 5)
       .attr(StatChangeAttr, [ BattleStat.ATK, BattleStat.SPATK ], 1, true),
     new AttackMove(Moves.ELECTROWEB, Type.ELECTRIC, MoveCategory.SPECIAL, 55, 95, 15, 100, 0, 5)
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -88,6 +88,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
   public luck: integer;
   public pauseEvolutions: boolean;
   public pokerus: boolean;
+  public wildFlee: boolean;
 
   public fusionSpecies: PokemonSpecies;
   public fusionFormIndex: integer;
@@ -129,6 +130,8 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     this.species = species;
     this.pokeball = dataSource?.pokeball || PokeballType.POKEBALL;
     this.level = level;
+    this.wildFlee = false;
+
     // Determine the ability index
     if (abilityIndex !== undefined) {
       this.abilityIndex = abilityIndex; // Use the provided ability index if it is defined
@@ -298,14 +301,14 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
   }
 
   /**
-   * Check if this pokemon is both not fainted and allowed to be in battle.
+   * Check if this pokemon is both not fainted (or a fled wild pokemon) and allowed to be in battle.
    * This is frequently a better alternative to {@link isFainted}
    * @returns {boolean} True if pokemon is allowed in battle
    */
   isAllowedInBattle(): boolean {
     const challengeAllowed = new Utils.BooleanHolder(true);
     applyChallenges(this.scene.gameMode, ChallengeType.POKEMON_IN_BATTLE, this, challengeAllowed);
-    return !this.isFainted() && challengeAllowed.value;
+    return !this.isFainted() && !this.wildFlee && challengeAllowed.value;
   }
 
   isActive(onField?: boolean): boolean {
@@ -1771,6 +1774,14 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
         resolve();
       }
     });
+  }
+
+  /**
+   * sets if the pokemon has fled (implies it's a wild pokemon)
+   * @param status - boolean
+   */
+  setWildFlee(status: boolean): void {
+    this.wildFlee = status;
   }
 
   updateInfo(instant?: boolean): Promise<void> {
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -2556,6 +2556,15 @@ export class BattleEndPhase extends BattlePhase {
 
     this.scene.updateModifiers().then(() => this.end());
   }
+
+  end() {
+    // removing pokemon at the end of a battle
+    for (const p of this.scene.getEnemyParty()) {
+      p.destroy();
+    }
+
+    super.end();
+  }
 }
 
 export class NewBattlePhase extends BattlePhase {
@@ -3767,6 +3776,7 @@ export class FaintPhase extends PokemonPhase {
   doFaint(): void {
     const pokemon = this.getPokemon();
 
+
     // Track total times pokemon have been KO'd for supreme overlord/last respects
     if (pokemon.isPlayer()) {
       this.scene.currentBattle.playerFaints += 1;
@@ -3817,17 +3827,10 @@ export class FaintPhase extends PokemonPhase {
       }
     }
 
+    // in double battles redirect potential moves off fainted pokemon
     if (this.scene.currentBattle.double) {
       const allyPokemon = pokemon.getAlly();
-      if (allyPokemon?.isActive(true)) {
-        let targetingMovePhase: MovePhase;
-        do {
-          targetingMovePhase = this.scene.findPhase(mp => mp instanceof MovePhase && mp.targets.length === 1 && mp.targets[0] === pokemon.getBattlerIndex() && mp.pokemon.isPlayer() !== allyPokemon.isPlayer()) as MovePhase;
-          if (targetingMovePhase && targetingMovePhase.targets[0] !== allyPokemon.getBattlerIndex()) {
-            targetingMovePhase.targets[0] = allyPokemon.getBattlerIndex();
-          }
-        } while (targetingMovePhase);
-      }
+      this.scene.redirectPokemonMoves(pokemon, allyPokemon);
     }
 
     pokemon.lapseTags(BattlerTagLapseType.FAINT);
--- a/src/test/moves/dragon_tail.test.ts
+++ b/src/test/moves/dragon_tail.test.ts
@@ -43,15 +43,15 @@ describe("Moves - Dragon Tail", () => {
     async () => {
       await game.startBattle([Species.DRATINI]);
 
-      const enemyPokemon = game.scene.getEnemyPokemon();
+      const enemyPokemon = game.scene.getEnemyPokemon()!;
       expect(enemyPokemon).toBeDefined();
 
       game.doAttack(getMovePosition(game.scene, 0, Moves.DRAGON_TAIL));
 
       await game.phaseInterceptor.to(BerryPhase);
 
-      const isVisible = enemyPokemon?.visible;
-      const hasFled = enemyPokemon?.wildFlee;
+      const isVisible = enemyPokemon.visible;
+      const hasFled = enemyPokemon.wildFlee;
       expect(!isVisible && hasFled ).toBe(true);
 
       // simply want to test that the game makes it this far without crashing
@@ -66,22 +66,20 @@ describe("Moves - Dragon Tail", () => {
       await game.startBattle([Species.DRATINI]);
       game.override.enemyAbility(Abilities.ROUGH_SKIN);
 
-      const leadPokemon = game.scene.getPlayerPokemon();
+      const leadPokemon = game.scene.getPlayerPokemon()!;
       expect(leadPokemon).toBeDefined();
 
-      const enemyPokemon = game.scene.getEnemyPokemon();
+      const enemyPokemon = game.scene.getEnemyPokemon()!;
       expect(enemyPokemon).toBeDefined();
 
       game.doAttack(getMovePosition(game.scene, 0, Moves.DRAGON_TAIL));
 
       await game.phaseInterceptor.to(BerryPhase);
 
-      const isVisible = enemyPokemon?.visible;
-      const hasFled = enemyPokemon?.wildFlee;
+      const isVisible = enemyPokemon.visible;
+      const hasFled = enemyPokemon.wildFlee;
       expect(!isVisible && hasFled ).toBe(true);
       expect(leadPokemon?.hp < leadPokemon?.getMaxHp());
-
-      await game.phaseInterceptor.to(BattleEndPhase);
     }, TIMEOUT
   );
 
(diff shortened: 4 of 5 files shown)


src/data/move.ts
```diff
@@ -4772,13 +4772,18 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
...
 	  } else {
-	    // Switch out logic for everything else
-	  	switchOutTarget.setVisible(false);
+	    // Switch out logic for everything else (eg: WILD battles)
+	  	switchOutTarget.leaveField(false);
 
 	  	if (switchOutTarget.hp) {
-	  	  switchOutTarget.hideInfo().then(() => switchOutTarget.destroy());
-	  	  switchOutTarget.scene.field.remove(switchOutTarget);
+          switchOutTarget.setWildFlee(true);
```
snoozbuster, line 4779: instead of trying to do the flee inline in the phase and then having to handle redirecting moves - I think if we made a new EnemyFleePhase and added that to the phase order, it would allow the remaining parts of MoveEffectPhase to run normally (so for example, post-contact effects like static would work correctly and have all the references they need), then the flee phase could come after and handling removing the mon (probably similar to how FaintPhase does it, although I would need to look at that code). I think that MovePhase already handles cases where the move tries to run but the target is missing and needs to be re-directed (although again, I would need to confirm). 

that's how I was thinking about fixing this when I was looking into it, anyway. it seemed like all the relevant code existed already, it just needed to be sequenced properly - my experience with this attr so far is that using phases for its effects generally seems to be a lot more stable than trying to do things inline in the same MoveEffectPhase.

snoozbuster (changes requested):
I would really like to see at least one unit test added for not crashing when a fleeing uturn user activates poison point or another ability - it should be an easy scenario to setup and test. there's some prior art in https://github.com/pagefaultgames/pokerogue/pull/3321 which could be referenced.

src/data/move.ts
```diff
@@ -7407,7 +7412,8 @@ export function initMoves() {
     new AttackMove(Moves.FROST_BREATH, Type.ICE, MoveCategory.SPECIAL, 60, 90, 10, 100, 0, 5)
       .attr(CritOnlyAttr),
     new AttackMove(Moves.DRAGON_TAIL, Type.DRAGON, MoveCategory.PHYSICAL, 60, 90, 10, -1, -6, 5)
-      .attr(ForceSwitchOutAttr),
+      .attr(ForceSwitchOutAttr)
+      .hidesTarget(),
```
snoozbuster, line 7416: what does the hides target flag do? I see it's already on whirlwind and roar but there's no description for what it does in the code

src/field/pokemon.ts
```diff
@@ -88,6 +88,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
   public luck: integer;
   public pauseEvolutions: boolean;
   public pokerus: boolean;
+  public wildFlee: boolean;
```
snoozbuster, line 91: could call this `hasFled`

src/phases.ts
```diff
@@ -2556,6 +2556,15 @@ export class BattleEndPhase extends BattlePhase {
 
     this.scene.updateModifiers().then(() => this.end());
   }
+
+  end() {
+    // removing pokemon at the end of a battle
+    for (const p of this.scene.getEnemyParty()) {
```
snoozbuster, line 2562: do you need to do any checks here? is there any harm in double-destroying a mon?

src/phases.ts
```diff
@@ -3817,17 +3827,10 @@ export class FaintPhase extends PokemonPhase {
       }
     }
 
+    // in double battles redirect potential moves off fainted pokemon
```
snoozbuster, line 3830: personally I don't think this comment is super needed anymore given that the function is now named accurately and has a good description/jsdoc - but that's something of a personal preference

snoozbuster:
>This bug should only affect doubles wild battles

there's a similar issue with singles and uturn - eg, a uturn wild pokemon uses uturn and gets afflicted by poison point or rough skin as it flees, or a uturn user has something like poison touch and it activates when they make contact with the player. can you check if these have been fixed as well (and ideally write a test for them)?

src/phases.ts
DustinLin: don't think so, `destroy()`should only be called once at the end of the battle phase. I tested it real quick by calling `destroy()` multiple times and it wasn't game breaking.

src/data/move.ts
DustinLin: the flee logic is a bit different from fainting/`FaintPhase`. The latter is more complex (you gotta do some xp tracking, play the fainting cry, update some internal variables for if the 'mon and update scores, etc.) additionally fainting adds new phases like picking item rewards and all that. I thought of something similar to your solution at first. But I think this simpler in that we don't have to mess with adding more phases/code

i guess there is a philosophical discussion to be had with introducing new phases. For moves like dragon tail, the flee/switch out is *part of* the MoveEffect, so can make an argument for why it would make sense to include it in the same phase, rather than splitting up the move resolution to multiple phases.

snoozbuster: the main reason to put it in a separate phase, IMO, is MoveEffectPhase pretty much exclusively operates on fieldIndexes (at least, as far as I'm aware) so any changes to the field during the phase's resolution can cause pretty serious issues - from what I've been able to surmise during debugging, a lot of the issues with moves that use this attribute are basically due to the fact that it modifies the field during the phase. in the case of the other PR I'm working on, changing the player branch to use SwitchPhase instead of `pokemon.switchOut` dramatically improved the stability of the move (and actually improved the order that the player sees things happen in, too) without actually changing the effect of it. that to me is the main reason to consider putting it in a separate phase. 

but tbh, I'm pretty sure the solution in this PR works fine too, and I'm not 100% convinced a new phase for this is a great idea either (honestly, I think it's probably not worth it unless FaintPhase can be made to use the same phase for part of its logic). as long as it's got UTs so we know if/when it breaks, I am fine with this or basically any implementation, since it means if someone does need to change it later it will be easy to verify.

src/test/moves/dragon_tail.test.ts
```diff
@@ -0,0 +1,173 @@
...
+      expect(enemyPokemon).toBeDefined();
+
+      game.doAttack(getMovePosition(game.scene, 0, Moves.DRAGON_TAIL));
+
+      await game.phaseInterceptor.to(BerryPhase);
+
+      const isVisible = enemyPokemon?.visible;
+      const hasFled = enemyPokemon?.wildFlee;
+      expect(!isVisible && hasFled ).toBe(true);
+      expect(leadPokemon?.hp < leadPokemon?.getMaxHp());
```
snoozbuster, line 82: given that you assert earlier that leadPokemon is defined, you probably don't need to use optional chaining here (plus it might fix the typedocs issues you're having)

DustinLin: yeah i think that was the issue, from discord @ NightKev: 

"basically whenenever you do `enemy = game.scene.getEnemyPokemon();` and `player = game.scene.getPlayerPokemon(); `add a `!` after the `()`, for now anyway; maybe in the future it won't be nececssary "

adding the bang will tell the compiler that the variable can't be undefined (a bypass for writing tests)

src/test/moves/dragon_tail.test.ts
```diff
@@ -0,0 +1,171 @@
...
+  });
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override.battleType("single");
+    game.override.moveset([Moves.DRAGON_TAIL, Moves.SPLASH]);
+    game.override.enemySpecies(Species.WAILORD);
+    game.override.enemyMoveset(SPLASH_ONLY);
+    game.override.startingLevel(5);
+    game.override.enemyLevel(5);
```
snoozbuster, line 36: I personally don't mind it either way but some folks prefer these to be chained together (`game.overrides.m1().m2()...`)

src/test/moves/dragon_tail.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      const enemyPokemon = game.scene.getEnemyPokemon()!;
+      expect(enemyPokemon).toBeDefined();
+
+      game.doAttack(getMovePosition(game.scene, 0, Moves.DRAGON_TAIL));
+
+      await game.phaseInterceptor.to(BerryPhase);
+
+      const isVisible = enemyPokemon.visible;
+      const hasFled = enemyPokemon.wildFlee;
+      expect(!isVisible && hasFled ).toBe(true);
```
snoozbuster, line 55:
```suggestion
      expect(!isVisible && hasFled).toBe(true);
```

src/test/moves/dragon_tail.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+      // simply want to test that the game makes it this far without crashing
+      await game.phaseInterceptor.to(BattleEndPhase);
+      expect(true).toBe(true);
```
snoozbuster, line 59: there's not really much of a point to this assertion imo, the comment does a fine job of explaining what's happening and this assertion never fails, so it's not really useful to have it here

src/test/moves/dragon_tail.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      const enemyPokemon = game.scene.getEnemyPokemon()!;
+      expect(enemyPokemon).toBeDefined();
+
+      game.doAttack(getMovePosition(game.scene, 0, Moves.DRAGON_TAIL));
+
+      await game.phaseInterceptor.to(BerryPhase);
+
+      const isVisible = enemyPokemon.visible;
+      const hasFled = enemyPokemon.wildFlee;
+      expect(!isVisible && hasFled ).toBe(true);
```
snoozbuster, line 81:
```suggestion
      expect(!isVisible && hasFled).toBe(true);
```

src/test/moves/dragon_tail.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+      game.doAttack(getMovePosition(game.scene, 1, Moves.SPLASH));
+
+      await game.phaseInterceptor.to(TurnEndPhase);
+
+      const isVisibleLead = enemyLeadPokemon.visible;
+      const hasFledLead = enemyLeadPokemon.wildFlee;
+      const isVisibleSec = enemySecPokemon.visible;
+      const hasFledSec = enemySecPokemon.wildFlee;
+      expect(!isVisibleLead && hasFledLead && isVisibleSec && !hasFledSec ).toBe(true);
```
snoozbuster, line 115:
```suggestion
      expect(!isVisibleLead && hasFledLead && isVisibleSec && !hasFledSec).toBe(true);
```

src/test/moves/dragon_tail.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      expect(leadPokemon.hp < leadPokemon.getMaxHp());
+
+      // second turn
+
+      game.doAttack(getMovePosition(game.scene, 0, Moves.FLAMETHROWER));
+      game.doSelectTarget(BattlerIndex.ENEMY_2);
+      game.doAttack(getMovePosition(game.scene, 1, Moves.SPLASH));
+
+      await game.phaseInterceptor.to(BerryPhase);
+      expect(enemySecPokemon.hp < enemySecPokemon.getMaxHp() ).toBe(true);
```
snoozbuster, line 126: I'm not entirely sure what this second turn is testing about dragon tail - is it just that you can take a second turn and nothing is crashing? just running the phases back around to TurnInitPhase is probably sufficient to say that a second turn would happen - a good way to "test the test" is to run the test on beta without your changes and remove as much as possible from the test such that it still fails without your change, and then passes with your change

src/test/moves/dragon_tail.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      const hasFledLead = enemyLeadPokemon.wildFlee;
+      const isVisibleSec = enemySecPokemon.visible;
+      const hasFledSec = enemySecPokemon.wildFlee;
+      expect(!isVisibleLead && hasFledLead && !isVisibleSec && hasFledSec ).toBe(true);
+
+      expect(leadPokemon.hp < leadPokemon.getMaxHp());
+      expect(secPokemon.hp < secPokemon.getMaxHp());
+      expect(enemyLeadPokemon.hp < enemyLeadPokemon.getMaxHp());
+      expect(enemySecPokemon.hp < enemySecPokemon.getMaxHp());
+
```
snoozbuster, line 168: you may also want to run this test to the BattleEndPhase and make sure the game makes it there - I'm not 100% sure this test would actually fail pre-fix (excepting the `hasFled` checks) because I think the rough skin damage and everything might go through before the crash. not 100% sure on that though

src/data/move.ts
DustinLin: `.hidesTarget()` sets moves flags, which I think is used in the animations (`battle-anims.ts`). TBH I don't know more than that, but I put this because whirlwind and roar both do similar things so I figured it makes sense too

src/field/pokemon.ts
DustinLin: I think `wildFlee` is a better name: since this variable is only updated/messed around with for wild battles, i feel like including "wild" in the variable name is more descriptive. `hasFled` was something that crossed my mind, but that requires one step to infer that fleeing can only happen for wild battles

src/test/moves/dragon_tail.test.ts
DustinLin: Yes, one of the bugs that I ran into while fixing this was that on the turn after the pokemon flees, certain move animations would crash the game (one such move was flamethrower). There were also problems with the CPU deciding enemy pokemon's moves after one has left the field so the simulation should at least go beyond `TurnInitPhase`

I was going to include these as comments in the test, but it seems like a lot of my comments get marked as unnecessary so I figured not. I could include this though if you think it's helpful

src/data/move.ts
snoozbuster: Assuming is a dangerous game (but it does probably make sense here, I would hope that this flag causes a sprite to be hidden after the animation plays)

src/field/pokemon.ts
snoozbuster: What about wildHasFled or hasWildFled? Could be best of both worlds

src/test/moves/dragon_tail.test.ts
DustinLin: Running to `BerryPhase` should be enough, the abilities triggering (showAbilityPhase, the damage, messagePhase, etc.) all happen before Berry so we should be good here

src/test/moves/dragon_tail.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      const isVisibleLead = enemyLeadPokemon.visible;
+      const hasFledLead = enemyLeadPokemon.wildFlee;
+      const isVisibleSec = enemySecPokemon.visible;
+      const hasFledSec = enemySecPokemon.wildFlee;
+      expect(!isVisibleLead && hasFledLead && !isVisibleSec && hasFledSec ).toBe(true);
+
+      expect(leadPokemon.hp < leadPokemon.getMaxHp());
+      expect(secPokemon.hp < secPokemon.getMaxHp());
+      expect(enemyLeadPokemon.hp < enemyLeadPokemon.getMaxHp());
+      expect(enemySecPokemon.hp < enemySecPokemon.getMaxHp());
```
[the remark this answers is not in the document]
DustinLin: yeah that's a much better comparator/assert to use

flx-sta:
This PR broke advancing past any 50th wave in endless mode by causing `pokemon.scene` to be `null/undefined`

src/data/move.ts
DayKev: Given that Roar and Whirlwind have a bug where they inappropriately hide the enemy Pokémon...

flx-sta:
Also found the reason
```ts
  end() {
    // removing pokemon at the end of a battle
    for (const p of this.scene.getEnemyParty()) {
      p.destroy(); // <--- ⚠️⚠️⚠️
    }

    super.end();
  }
  ```
  
  this is causing `pokemon.scene` to be `null/undefined` at the end of the fight even though there is still some calculation necessary. SO if that is just some "nice addition" we could just remove it

**Comment by DayKev** [comment]
It only breaks endless past wave 50? That's... weird.

**Comment by flx-sta** [comment]
> It only breaks endless past wave 50? That's... weird.

No it break any `%50` wave in endless. (or just some!?) idk.
Here is some session to work with (just change file-extension to `.prsv`)
[sessionData3_Guest.txt](https://github.com/user-attachments/files/16555427/sessionData3_Guest.txt)

What happens is
```
Start Phase BerryPhase
Start Phase TurnEndPhase
Start Phase BattleEndPhase | THIS calls p.destroy() on all enemies
...
Start Phase EggLapsePhase
Start Phase ModifierRewardPhase
Start Phase AddEnemyBuffModifierPhase
```

the last phase calls `regenerateModifierPoolThresholds` with the enemyParty() which has just been destroyed.

```ts
start() {
    super.start();

    const waveIndex = this.scene.currentBattle.waveIndex;
    const tier = !(waveIndex % 1000) ? ModifierTier.ULTRA : !(waveIndex % 250) ? ModifierTier.GREAT : ModifierTier.COMMON;

    regenerateModifierPoolThresholds(this.scene.getEnemyParty(), ModifierPoolType.ENEMY_BUFF);

    const count = Math.ceil(waveIndex / 250);
    for (let i = 0; i < count; i++) {
      this.scene.addEnemyModifier(getEnemyBuffModifierForWave(tier, this.scene.findModifiers(m => m instanceof EnemyPersistentModifier, false), this.scene), true, true);
    }
    this.scene.updateModifiers(false, true).then(() => this.end());
  }
}
```

and inside `regenerateModifierPoolThresholds()` it's trying to call
```ts
 const existingModifiers = party[0].scene.findModifiers
 ``` 
 but `scene` is null/undefined because all pokemons were destroyed previously and are not there anymore

PR #3349, opened 2024-08-04
[Balance] Adjust Gallade/Froslass evos to require move of their type

Fontbane:
## What are the changes?

- Adjust Snorunt and Kirlia evolution methods:
  - Snorunt -> Glalie: Level 42 ~~(male)~~ _(either gender, unless you fulfill Froslass requirements)_
  - Snorunt -> Froslass: Level 42 (female) **while knowing a Ghost-type move**
  - Kirlia -> Gallade: Level 30 (male) **while knowing a Fighting-type move**
  - Kirlia -> Gardevoir: Level 30 ~~(female)~~ _(either gender, unless you fulfill Gallade requirements)_
  - To accommodate Kirlia's new evolution method, it now learns **Quick Guard** at level **25**.

- Froslass and Gallade's evolutions are custom in Rogue to just require reaching the level as the appropriate gender
- This is for convenience so these mons aren't nerfed by having to search for a Dawn Stone for players to evolve them
- This meant that only male Snorunt evolved into Glalie and only female Kirlia evolved into Gardevoir
- This created issues in cases of certain challenge runs where this custom evolution method, meant for convenience, could make these mons unusable if the player had one of the wrong gender
- The new solution follows more expected behavior in line with the main series, where both genders can evolve into Glalie or Gardevoir, while evolving into Froslass or Gallade requires a player choice (in this case, teaching a Ghost or Fighting move)
- Snorunt learns **Astonish** naturally at level 1. Kirlia was given Quick Guard at level 25 as it's a fitting move which Gallade learns by level up.

## What did change?

- Kirlia evolving into Gallade requires it being male and knowing a Fighting-type move
- Kirlia now learns Quick Guard at level 25
- Kirlia into Gardevoir now has no conditions other than level, though Gallade's evolution will take precedent if the conditions are met
- Snorunt evolving into Froslass requires it being female and knowing a Ghost-type move
- Snorunt into Glalie now has no conditions other than level, though Froslass's evolution will take precedent if the conditions are met

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/pokemon-evolutions.ts
+++ b/src/data/pokemon-evolutions.ts
@@ -425,8 +425,8 @@ export const pokemonEvolutions: PokemonEvolutions = {
     new SpeciesEvolution(Species.KIRLIA, 20, null, null)
   ],
   [Species.KIRLIA]: [
-    new SpeciesEvolution(Species.GARDEVOIR, 30, null, new SpeciesEvolutionCondition(p => p.gender === Gender.FEMALE, p => p.gender = Gender.FEMALE)),
-    new SpeciesEvolution(Species.GALLADE, 30, null, new SpeciesEvolutionCondition(p => p.gender === Gender.MALE, p => p.gender = Gender.MALE))
+    new SpeciesEvolution(Species.GALLADE, 30, null, new SpeciesEvolutionCondition(p => p.gender === Gender.MALE && !!p.getMoveset().find(m => m.getMove().type === Type.FIGHTING), p => p.gender = Gender.MALE)),
+    new SpeciesEvolution(Species.GARDEVOIR, 30, null, null)
   ],
   [Species.SURSKIT]: [
     new SpeciesEvolution(Species.MASQUERAIN, 22, null, null)
@@ -514,8 +514,8 @@ export const pokemonEvolutions: PokemonEvolutions = {
     new SpeciesEvolution(Species.DUSCLOPS, 37, null, null)
   ],
   [Species.SNORUNT]: [
-    new SpeciesEvolution(Species.GLALIE, 42, null, new SpeciesEvolutionCondition(p => p.gender === Gender.MALE, p => p.gender = Gender.MALE)),
-    new SpeciesEvolution(Species.FROSLASS, 42, null, new SpeciesEvolutionCondition(p => p.gender === Gender.FEMALE, p => p.gender = Gender.FEMALE))
+    new SpeciesEvolution(Species.FROSLASS, 42, null, new SpeciesEvolutionCondition(p => p.gender === Gender.FEMALE && !!p.getMoveset().find(m => m.getMove().type === Type.GHOST), p => p.gender = Gender.FEMALE)),
+    new SpeciesEvolution(Species.GLALIE, 42, null, null)
   ],
   [Species.SPHEAL]: [
     new SpeciesEvolution(Species.SEALEO, 32, null, null)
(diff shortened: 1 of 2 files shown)


damocleas (changes requested):
Could you not put it at the same level as evolve, since you can learn moves before you evolve that would dictate it?

damocleas (approved):

src/data/pokemon-level-moves.ts
```diff
@@ -4879,6 +4879,7 @@ export const pokemonSpeciesLevelMoves: PokemonSpeciesLevelMoves = {
     [ 15, Moves.TELEPORT ],
     [ 18, Moves.PSYBEAM ],
     [ 23, Moves.LIFE_DEW ],
+    [ 25, Moves.QUICK_GUARD ],
```
Tempo-anon, line 4882: grrrr more custom moves (also quick guard is pretty bad)

src/data/pokemon-evolutions.ts
```diff
@@ -425,8 +425,8 @@ export const pokemonEvolutions: PokemonEvolutions = {
     new SpeciesEvolution(Species.KIRLIA, 20, null, null)
   ],
   [Species.KIRLIA]: [
-    new SpeciesEvolution(Species.GARDEVOIR, 30, null, new SpeciesEvolutionCondition(p => p.gender === Gender.FEMALE, p => p.gender = Gender.FEMALE)),
-    new SpeciesEvolution(Species.GALLADE, 30, null, new SpeciesEvolutionCondition(p => p.gender === Gender.MALE, p => p.gender = Gender.MALE))
+    new SpeciesEvolution(Species.GALLADE, 30, null, new SpeciesEvolutionCondition(p => p.gender === Gender.MALE && !!p.getMoveset().find(m => m.getMove().type === Type.FIGHTING), p => p.gender = Gender.MALE)),
```
Tempo-anon, line 428: Needing to know a move of a specific type feels pretty weird to me. Personally I'd also just make Onix/Scyther be link cable only

Tempo-anon (commented):
I personally think that "Male Kirlia turns into Gallade and Female Snorunt turns into Froslass" intuitively makes more sense and is easier to grasp than "They need to be the right gender and have a specific type move."

In terms of moves, this means adding a custom mediocre move to Kirlia and forcing Snorunt to carry a very bad move to level 42 and as seen from players reactions to Pokemon like Yanma needing ancientpower, that's not a very popular thing.

DayKev:
If you're adding the move requirement why do Gallade and Froslass still have the gender requirement then?

Fontbane:
> If you're adding the move requirement why do Gallade and Froslass still have the gender requirement then?

Gallade and Froslass are male/female only

src/data/pokemon-level-moves.ts
Fontbane: If you're gonna evolve into Gallade it's not like you're keeping any of Kirlia's level up moves anyway

src/data/pokemon-evolutions.ts
Fontbane: Sylveon also does this

Fontbane:
The woke mob got me folks

damocleas (changes requested):
I feel like this can be gone about a different way for Froslass/Glalie, and if the same for Gallade/Gardevoir I don't see reason to gender lock Snorunt.

Fontbane:
> I feel like this can be gone about a different way for Froslass/Glalie, and if the same for Gallade/Gardevoir I don't see reason to gender lock them.

The move solution isn't as clean for Froslass/Glalie, true. I don't see why we would un-gender lock Gallade though.

Fontbane:
Not happening

PR #3355, opened 2024-08-05
[Bug] Fix Make It Rain and Clanging Scales stat drop trigger logic

innerthunder:
## What are the changes?

This fixes an issue where if Make It Rain only hits its second target in a double battle, the Sp. Atk drop would not trigger. This also changes the parameters to Clanging Scales' `StatChangeAttr` to be identical to Make It Rain's attribute, so both moves' stat changes should now behave in the same way.

This bug was reported [in the Discord](https://discord.com/channels/1125469663833370665/1269843880543785055)

## What did change?

- `phases`: Changed `MoveEffectPhase`'s `firstTarget` logic from "`true` for the first target of the move only" to "`true` for the move's first successful hit in a strike only".
- `data/move`: Changed Clanging Scales' `StatChangeAttr` params to match Make It Rain's `StatChangeAttr` params.
- `test/moves/make_it_rain.test`: Added test covering a case where Make It Rain misses it's first target and hits its second.

### Screenshots/Videos

N/A

## How to test the changes?

`npm run test make_it_rain`

## Checklist

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -2966,6 +2966,8 @@ export class MoveEffectPhase extends PokemonPhase {
 
       // Move animation only needs one target
       new MoveAnim(move.id as Moves, user, this.getTarget()?.getBattlerIndex()).play(this.scene, () => {
+        /** Has the move successfully hit a target (for damage) yet? */
+        let hasHit: boolean = false;
         for (const target of targets) {
           if (!targetHitChecks[target.getBattlerIndex()]) {
             this.stopMultiHit(target);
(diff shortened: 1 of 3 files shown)


src/phases.ts
```diff
@@ -2991,6 +2992,11 @@ export class MoveEffectPhase extends PokemonPhase {
 
           const hitResult = !isProtected ? target.apply(user, move) : HitResult.NO_EFFECT;
 
+          const firstTarget = (hitResult < HitResult.NO_EFFECT) && !hasHit;
```
Tempo-anon, line 2995: I'm not sure if I like using a `<` to compare for an enum. I know this is checking for 
```ts
  EFFECTIVE = 1,
  SUPER_EFFECTIVE,
  NOT_VERY_EFFECTIVE,
  ONE_HIT_KO,
  ```
but this is a bit hard to understand. I think it'd be easier to read if you put all those enum's into a list of acceptable results

PR #3369, opened 2024-08-06
[P2] Fixes party status cure moves only curing the player's pokemon, even when used by enemy pokemon

schmidtc1:
Rebranch of #3158

## TODO

- [x] Write tests for 
  - [x] Sparkly Swirl
  - [x] Aromatherapy
  - [x] Heal Bell

## What are the changes?

Fixes issue with enemy pokemon curing the player's pokemon when using party status cure moves due to using `getParty()`.
Also allows user to cure its own status regardless of ability, per [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Heal_Bell_(move)#:~:text=Heal%20Bell%20now%20always%20affects%20the%20user%2C%20even%20if%20it%20has%20Soundproof%3B%20however%2C%20it%20still%20cannot%20affect%20other%20active%20Pok%C3%A9mon%20with%20Soundproof.)

This is a bug. Also, the `PartyStatusCureAttr` has its own Phase created specifically to host its functionality. This functionality can be moved into the attribute and the Phase can be deleted.

## What did change?

Uses the proper function for player pokemon or enemy pokemon depending on which type of pokemon uses the move.
Refactors the `PartyStatusCurePhase` functionality entirely into the `PartyStatusCureAttr` instead of having it separated.
Allows user to cure its own status regardless of ability.

### Screenshots/Videos

Before

https://github.com/user-attachments/assets/3a1e8720-1c09-401d-88c5-4ae091511725

After

https://github.com/user-attachments/assets/fcf66135-82e6-4dd5-949f-995af232677a

## How to test the changes?

Give opponent status effect + status effect curing move, see that it no longer affects only the player's pokemon. 

## Checklist
- [x] Are the changes visual?


Snailman11:
Link to Discord's bug report (Sparkly Swirl cures statuses for player team when used by an opponent)
https://discord.com/channels/1125469663833370665/1266403858670817381

frutescens (dismissed):
Hi, thank you for working on this bug. Could you resolve merge conflicts?

src/data/move.ts
```diff
@@ -1514,12 +1514,26 @@ export class PartyStatusCureAttr extends MoveEffectAttr {
...
-    user.scene.unshiftPhase(new PartyStatusCurePhase(user.scene, user, this.message, this.abilityCondition));
+  cureStatus(pokemon: Pokemon, userId: number) {
+    if (!pokemon.isOnField() || pokemon.id === userId) { // user always cures its own status, regardless of ability
+      pokemon.resetStatus(false);
+      pokemon.updateInfo();
+    } else if (!pokemon.hasAbility(this.abilityCondition)) {
+      pokemon.resetStatus();
+      pokemon.updateInfo();
+    } else {
+      pokemon.scene.unshiftPhase(new ShowAbilityPhase(pokemon.scene, pokemon.id, pokemon.getPassiveAbility()?.id === this.abilityCondition));
```
schmidtc1, line 1535: Kept this from the party-status-cure-phase file, but it seems like an unorthodox method of showing the ability flyout. Any suggestions for this?

DayKev: I guess you can check what other abilities do? Not sure.

innerthunder: This is for Heal Bell not affecting party members with Soundproof, I think. I guess you could try to apply `MoveImmunityAttr` and check the `cancelled` argument afterwards instead, but the current code seems like a more lightweight way of handling it.

schmidtc1: Found out the reason this is required is because Heal Bell and Aromatherapy use `MoveTarget.PARTY`, but that targetting only affects the user so it doesn't actually hit its party members to trigger their ability flyouts

src/data/move.ts
```diff
@@ -1585,12 +1585,26 @@ export class PartyStatusCureAttr extends MoveEffectAttr {
...
+    if (this.message) {
+      user.scene.queueMessage(this.message);
+    }
+
     return true;
   }
 
-  addPartyCurePhase(user: Pokemon) {
-    user.scene.unshiftPhase(new PartyStatusCurePhase(user.scene, user, this.message, this.abilityCondition));
+  cureStatus(pokemon: Pokemon, userId: number) {
```
flx-sta, line 1598: NIT: missing scope and docs

flx-sta (dismissed):
⚠️ Code (minor NIT)
✅ Playtest

src/data/move.ts
flx-sta:
```suggestion
  /**
   * Tries to cure the status of the given {@linkcode Pokemon}
   * @param pokemon The {@linkcode Pokemon} to cure.
   * @param userId The ID of the (move) {@linkcode Pokemon | user}.
   */
  public cureStatus(pokemon: Pokemon, userId: number) {
```

flx-sta (approved):

PR #3373, opened 2024-08-06
[Bug] Starter UI challenge form fix 

Opaque02:
## What are the changes?
There's currently a bug with the starter select UI as discussed [here](https://discord.com/channels/1125469663833370665/1269144093938815098)

There's a few underlying bugs that needed to be fixed to fix this bug, so this PR addresses all of them

See above

## What did change?
1. Firstly, the starter UI update from #2830 changed some rules around challenge validity. While the changed files didn't make a different to the logic, they were redundant, so it has been removed.
2. Regarding the bug above from the discord channel, it partly came about because a lot of placed within the starter select screen use a function to get the default dex properties, which is different to the _currently_ chosen dex properties. I added a new function to generate dex properties based on the starter preferences of a pokemon. The other part was that some forms weren't getting checked for challenge validity (i.e. tauros fire form not being checked, so tauros and oricorio would not show on some challenges)
3. Due to the above function, I've made a change where different forms can be selected for challenges, and the logic applies to them. If you are not on a valid form, it will let you know it's not a valid pokemon for the challenge like usual
4. In live at the moment, when going through different pokemon on the starter select screen, a) the icon does not stay if you toggle shiny status and b) if you change variants, the pokemon icon isn't guaranteed to stay either. This has been fixed mainly with the help of using the above function instead of using default props
5. The persistent information for shiny status was not being kept, so this PR adds that functionality. 
6. Currently when choosing a pokemon, the variant star will be at max on the left panel when you move in to update the pokemon's details, regardless of whether you previously chose to have it shiny/an earlier tier of variant. This has now been updated using 
7. Party icons don't update with form/shiny/variant etc. This is now fixed with a new function that updates a party icon to use the current dex properties
8. The behaviour on the starter screen has changed. Previously when choosing a starter, it would choose the highest rank of variant that you have. Now if you don't have a selected choice, it chooses the base non-shiny version of the pokemon for you until you make changes and have them save
9. When reloading, the state of your favourites is not saved (i.e. shiny, form etc) as mentioned in the comment [here](https://github.com/pagefaultgames/pokerogue/pull/3373#pullrequestreview-2221703882). This has now been fixed by adding a section to the updateStarters method that updates the icons of all valid starters when you start a new game. I don't know how performance intensive it is, but it only happens once per new game at the beginning, so hopefully not too bad. See video below on it working

### Screenshots/Videos

This is an example of the behaviour in live. You can see that as I go through pokemon, the shiny and variant status doesn't necessarily stay the same as what I set. Also while I don't point it out, the shiny icon is always the max (in my case, tier 3 shiny):

https://github.com/user-attachments/assets/58b787d8-f0f4-46ee-982d-2368e4597ebf

This is my PR. Apologies for the breakpoint being hit, I hadn't removed it from testing :). As you can see, the shiny and variant status stay the same as what you set, and the shiny icon is the correct variant:

https://github.com/user-attachments/assets/de875d81-2df9-4d33-a3b1-1524ef1c7e7f

Here's the discord bug in beta. As you can see, tauros isn't shown in the fire challenge, nor is oricorio in the electric challenge. Also in the fire challenge, you can change oricorio's form to a non-fire one and it be valid:

https://github.com/user-attachments/assets/36a595c8-8ebc-4954-b657-a0171e7dc36d

Here's my PR fix. Tauros now shows up for a fire challenge, but is only valid to select when you've got the fire form chosen. This is the same for oricorio:

https://github.com/user-attachments/assets/86a0de9f-2bb0-44a7-9f9e-f3690d4c8a3e

Lastly, this is the party icons in live at the moment when you go to change shiny/variant/form:

https://github.com/user-attachments/assets/d84fd77e-2880-4b38-8f61-58bf96d75d88

Finally, this is the same thing in this PR:

https://github.com/user-attachments/assets/0170e354-fcb0-431a-bba7-30a46e988f6d

This is what happens when changing forms/shiny/variants of pokemon and then reloading the page with this PR:

https://github.com/user-attachments/assets/ead9f8c6-4360-4165-978e-8d775798e618

(sorry the video is a little blurry, I had to compress the video to make it fit here) As you can see, all forms/shiny/variant data is now saved when you reload, so all should be good

## How to test the changes?
Download this PR and have a play around with the starter ui. It helps if also have the live/beta version side by side to play around with it and see what's different

## Checklist
- [x] Are the changes visual?


src/ui/starter-select-ui-handler.ts
```diff
@@ -1296,11 +1296,13 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
...
-          } else {
-            Challenge.applyChallenges(this.scene.gameMode, Challenge.ChallengeType.STARTER_CHOICE, species, isValidForChallenge, this.scene.gameData.getSpeciesDexAttrProps(species, this.scene.gameData.getSpeciesDefaultDexAttr(species, false, true)), this.starterSpecies.length !== 0, false, false);
-          }
+          //if (isPartyValid) {
+          //  Challenge.applyChallenges(this.scene.gameMode, Challenge.ChallengeType.STARTER_CHOICE, species, isValidForChallenge, this.scene.gameData.getSpeciesDexAttrProps(species, this.scene.gameData.getSpeciesDefaultDexAttr(species, false, true)), this.starterSpecies.length !== 0);
+          //} else {
+          //  Challenge.applyChallenges(this.scene.gameMode, Challenge.ChallengeType.STARTER_CHOICE, species, isValidForChallenge, this.scene.gameData.getSpeciesDexAttrProps(species, this.scene.gameData.getSpeciesDefaultDexAttr(species, false, true)), this.starterSpecies.length !== 0, false, false);
+          //}
```
DayKev, line 1303: You can just remove the old code, that's what git is for lol.

src/ui/starter-select-ui-handler.ts
```diff
@@ -2446,7 +2475,8 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
     }
 
     if (this.lastSpecies) {
-      const dexAttr = this.scene.gameData.getSpeciesDefaultDexAttr(this.lastSpecies, false, true);
+      const dexAttr = this.getCurrentDexProps(this.lastSpecies.speciesId);
+      //const dexAttr = this.scene.gameData.getSpeciesDefaultDexAttr(this.lastSpecies, false, true);
```
DayKev, line 2479: Same here.

src/ui/starter-select-ui-handler.ts
```diff
@@ -3002,12 +3038,14 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
...
+      //if (isPartyValid) { // we have two checks here - one for the party being valid and one for not. This comes from mono type challenges - if the party is valid it will check pokemon's evolutions and forms, and if it's not valid it won't check their evolutions and forms
+      //  Challenge.applyChallenges(this.scene.gameMode, Challenge.ChallengeType.STARTER_CHOICE, this.allSpecies[s], isValidForChallenge, this.scene.gameData.getSpeciesDexAttrProps(this.allSpecies[s], this.scene.gameData.getSpeciesDefaultDexAttr(this.allSpecies[s], false, true)), !!(this.starterSpecies.length + (add ? 1 : 0)));
+      //} else {
+      //  Challenge.applyChallenges(this.scene.gameMode, Challenge.ChallengeType.STARTER_CHOICE, this.allSpecies[s], isValidForChallenge, this.scene.gameData.getSpeciesDexAttrProps(this.allSpecies[s], this.scene.gameData.getSpeciesDefaultDexAttr(this.allSpecies[s], false, true)), !!(this.starterSpecies.length + (add ? 1 : 0)), false, false);
+      //}
```
DayKev, line 3048: And here.

src/ui/starter-select-ui-handler.ts
```diff
@@ -3002,12 +3038,14 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
        * If speciesStarterDexEntry?.caughtAttr is true, this species registered in stater.
        * we change to can AddParty value to true since the user has enough cost to choose this pokemon and this pokemon registered too.
        */
+      //const isPartyValid = this.isPartyValid();
```
DayKev, line 3041: +1

src/ui/starter-select-ui-handler.ts
```diff
@@ -3122,12 +3160,53 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
...
+      if (this.starterPreferences[speciesId]?.variant) {
+        props += BigInt(Math.pow(2, this.starterPreferences[speciesId]?.variant)) * DexAttr.DEFAULT_VARIANT;
+      } else {
+        props += DexAttr.DEFAULT_VARIANT;
+      }
+    } else {
+      props += DexAttr.NON_SHINY;
+      //if (this.starterPreferences[speciesId]?.variant) {
+      //  this.starterPreferences[speciesId].variant = 0;
+      //}
```
DayKev, line 3195: And finally, here.

KimJeongSun (commented):
I've discovered an visual bug in here

It seems that the current logic was designed to work as follows, and was perhaps modified to ensure the following normal operation:
Change Sandshrew to shiny.
Return to the main menu.
Re-enter the starter select UI.
Sandshrew appears shiny.

However, if you refresh the page and restart instead of directly re-entering the starter select UI in step 3, it works as follows:
Change Sandshrew to shiny.
Return to the main menu.
Refresh the page.
Enter the starter select UI.
Sandshrew appears not shiny. But if you hover the cursor over Sandshrew, it seems to change back to shiny again. (Yes, it's a visual bug.)

MokaStitcher (commented):
Issues I found so far:
- Burmy is not able to be selected for Steel, Grass and Ground challenges

- For type challenge, when selecting as first Pokemon a Pokemon with default form not having that type it will not immediately enable the starters which need evolution to become valid Pokemon

For example in this video, selecting Heat Rotom as first starter does not show Type:null & others as available until I move the cursor to another Pokemon 

https://github.com/user-attachments/assets/796fd5c2-080f-4e1d-8fdd-33af65d587f6

If the same is done in a Ghost challenge with Rotom this issue is not here, I'm assume because Electric/Ghost is Rotom's default typing

- When removing a Pokemon from the team, the sprite of the other team members is incorrectly updated:

https://github.com/user-attachments/assets/5d26b82f-636e-4be0-b2e3-a3064105e1a3


- Issue with sprite getting wrongly updated when moving to Pokemon from the team:
Full details in [this discord message](https://discord.com/channels/1125469663833370665/1267339517371875361/1270443302482743347)

https://github.com/user-attachments/assets/63df6f5a-c487-4333-a656-5f166e26b522

src/ui/starter-select-ui-handler.ts
```diff
@@ -1664,15 +1661,20 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
         case Button.CYCLE_SHINY:
           if (this.canCycleShiny) {
             const newVariant = props.variant;
+            starterAttributes.shiny = starterAttributes.shiny ? !starterAttributes.shiny : true;
```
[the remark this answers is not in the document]
Opaque02: This one was a bit tricky and uses something that typescript has. Basically, starterAttributes.shiny can be undefined. What we're doing in the start of the ternary (the starterAttributes.shiny ? ) bit is checking if the starterAttributes for this pokemon already have a value for shiny, since the starterAttributes are actually empty until we populate it. This means if there's a pokemon we haven't changed the value of the starterAttributes for, it won't have a shiny value at all. Since we already know we can cycle the shiny (since we're within the check), it means we have both shiny and not shiny unlocked. So, the way it works is like so:

If the pokemon has the shiny and not shiny unlocked and it doesn't already have an attribute for starterAttributes.shiny, then that must mean it's currently not shiny and we're trying to make it shiny. So for that, starterAttributes.shiny doesn't exist, so the ternary returns false, and starterAttributes.shiny = true;
If a pokemon has both shiny and not shiny and it does already have a value for starterAttributes.shiny, then we want to toggle that to the opposite state - for example if it's already shiny, we want to toggle to not shiny. This means in our case the ternary returns true, and starterAttributes.shiny = !starterAttributes.shiny

Hopefully that makes sense - I can go into details about particular bits more if you want me to

Opaque02:
> Issues I found so far:
> 
>     * Burmy is not able to be selected for Steel, Grass and Ground challenges
> 

@MokaStitcher So I asked Temp in discord - burmy is a separate issue so would need to be part of a separate PR, though I believe this lays the groundwork for it at least :)

Opaque02:
> Issues I found so far:
> 
>     * For type challenge, when selecting as first Pokemon a Pokemon with default form not having that type it will not immediately enable the starters which need evolution to become valid Pokemon
> 
> 
> For example in this video, for a Fire type challenge. Selecting Heat Rotom as first starter does not show Type:null & others as available until I move the cursor to another Pokemon
> rotom-bug.mov
> 
> If the same is done in a Ghost challenge with Rotom this issue is not here, I assume because Electric/Ghost is Rotom's "default" typing
> 
>     * When removing a Pokemon from the team, the sprite of the other team members is incorrectly updated:
> 
> 
> remove-pokemon-bug.mov
> 
>     * Issue with sprite getting wrongly updated when moving to Pokemon from the team:
>       Full details in [this discord message](https://discord.com/channels/1125469663833370665/1267339517371875361/1270443302482743347)
> 
> 
> kang.mov

@MokaStitcher - This issue with rotom should now be fixed with the latest commit

Opaque02:
> I've discovered an visual bug in here
> 
> It seems that the current logic was designed to work as follows, and was perhaps modified to ensure the following normal operation:
> 
>     1. Change Sandshrew to shiny.
> 
>     2. Return to the main menu.
> 
>     3. Re-enter the starter select UI.
> 
>     4. Sandshrew appears shiny.
> 
> 
> However, if you refresh the page and restart instead of directly re-entering the starter select UI in step 3, it works as follows:
> 
>     1. Change Sandshrew to shiny.
> 
>     2. Return to the main menu.
> 
>     3. Refresh the page.
> 
>     4. Enter the starter select UI.
> 
>     5. Sandshrew appears not shiny. But if you hover the cursor over Sandshrew, it seems to change back to shiny again. (Yes, it's a visual bug.)

@KimJeongSun - this should now be fixed

KimJeongSun (approved):
LGTM!

PR #3395, opened 2024-08-07
[Enhancement] Achievements Page is now Scrollable

frutescens:
## What are the changes?
The achievements page is now scrollable, allowing for more room for more achievements. 

I need to learn how to implement scroll cursors for other things and this seems like the easiest way to put things into practice. 

## What did change?
achv-ui-handler.ts now has functions determining scroll position. 

### Screenshots/Videos
https://github.com/user-attachments/assets/89cb9ee1-cb49-469a-acbd-967bcd7b52e4

## How to test the changes?
Add a fake achievement or pull system/achv.ts from a PR that adds more achievements. Then open Menu > Achievements and scroll.

## Checklist
- [x] Are the changes visual?


src/ui/achvs-ui-handler.ts
```diff
@@ -19,10 +22,16 @@ export default class AchvsUiHandler extends MessageUiHandler {
   private scoreText: Phaser.GameObjects.Text;
   private unlockText: Phaser.GameObjects.Text;
 
+  private achvsTotal: integer;
+  private scrollCursor: integer;
```
torranx, line 26:
```suggestion
  private achvsTotal: number;
  private scrollCursor: number;
```

src/ui/achvs-ui-handler.ts
```diff
@@ -9,6 +9,9 @@ import { addWindow } from "./ui-theme";
 import { ParseKeys } from "i18next";
 import { PlayerGender } from "#enums/player-gender";
 
+const achvRows = 4;
+const achvCols = 17;
```
torranx, line 13: is there a reason why this is outside of the class?

frutescens: I was mirroring ui/vouchers-ui-handler.ts implementation and it was done like that.

xsn34kzx, line 13: My only NIT here would be that constants should be all caps with underscores in between words, i.e. `ACHV_ROWS`.

src/ui/achvs-ui-handler.ts
```diff
@@ -19,10 +22,16 @@ export default class AchvsUiHandler extends MessageUiHandler {
...
 
+  private achvsTotal: number;
+  private scrollCursor: number;
+
   private cursorObj: Phaser.GameObjects.NineSlice;
 
   constructor(scene: BattleScene, mode?: Mode) {
     super(scene, mode);
+
+    this.achvsTotal = Object.keys(achvs).length;
```
xsn34kzx, line 33: What I think would be a bit more ergonomic is having something like `this.achvs` be the keys themselves and then get the length as needed for spots like lines `204` and `233` where `Object.keys(...)` is called multiple times.

xsn34kzx (commented):
Minor comments, normally I'd be against single-use variables, but I think it helps with the legibility here a lot 👍🏼

src/ui/achvs-ui-handler.ts
torranx: I don't see the point of declaring the variables outside of the class, and ironically it is only being used by the class. imo the class should encapsulate these like so: `private readonly ACHV_ROWS = 4`

torranx: Also moving forward please don't immediately resolve comments esp if consensus was not made

frutescens: Ok, should I update vouchers-ui-handler.ts to match the new format as well?

src/ui/achvs-ui-handler.ts
frutescens: I think I will replace the use of Object.keys(achvs).length in lines 204 with this.achvTotal instead since this.achvTotal = Object.keys(achvs).length. I would argue for keeping the Object.keys(...) usage in line 233 since it's important for cursor positioning though.

src/ui/achvs-ui-handler.ts
frutescens: Will implement your solution depending on whether voucher-ui-handler.ts should be changed as well to match the new and improved formatting.

src/ui/achvs-ui-handler.ts
xsn34kzx: Oh no, I agree, I think the usage is fine; I more so mean that you can do something like `this.achvs = Object.keys(achvs)` so you can reduce that line (`233`) and then for everything that needs the length, you can still do `this.achvs.length` (like `204`, etc.).

frutescens: Oh, I think Object.key(achvs) on its own is only used in one line, while this.achvTotal / Object.key(achvs).length is used multiple times in the code. That's why I feel like it's unnecessary.

src/ui/achvs-ui-handler.ts
```diff
@@ -19,10 +22,16 @@ export default class AchvsUiHandler extends MessageUiHandler {
   private scoreText: Phaser.GameObjects.Text;
   private unlockText: Phaser.GameObjects.Text;
 
+  private achvsTotal: number;
+  private scrollCursor: number;
+
   private cursorObj: Phaser.GameObjects.NineSlice | null;
 
   constructor(scene: BattleScene, mode?: Mode) {
     super(scene, mode!); // TODO: is this bang correct?
```
flx-sta, line 31: @frutescens if you know how to get rid of the bang behind `mode!` I'd appreciate that fix :D

frutescens: Not at the moment. But could the fix be in another PR since UI stuff is delicate?

flx-sta: sure. I'm booting a PR for that RN

src/ui/achvs-ui-handler.ts
```diff
@@ -219,12 +230,54 @@ export default class AchvsUiHandler extends MessageUiHandler {
...
 
     if (updateAchv) {
-      this.showAchv(achvs[Object.keys(achvs)[cursor]]);
+      this.showAchv(achvs[Object.keys(achvs)[cursor + this.scrollCursor * this.ACHV_COLS]]);
     }
 
     return ret;
   }
 
+  setScrollCursor(scrollCursor: integer): boolean {
```
torranx, line 239: can we get docs here? considering the fact that we also have a method named `setCursor`

src/ui/achvs-ui-handler.ts
```diff
@@ -219,12 +230,54 @@ export default class AchvsUiHandler extends MessageUiHandler {
...
+    this.scrollCursor = scrollCursor;
+
+    this.updateAchvIcons();
+
+    this.showAchv(achvs[Object.keys(achvs)[Math.min(this.cursor + this.scrollCursor * this.ACHV_COLS, Object.values(achvs).length - 1)]]);
+
+    return true;
+  }
+
+  updateAchvIcons(): void {
```
torranx, line 253: here too please

src/ui/achvs-ui-handler.ts
frutescens: do you mean comments?

PR #3406, opened 2024-08-07
[Bug] Fix healing modifying the HP stat instead of the Pokemon's HP

DayKev:
## What are the changes?
Damaging moves that heal you are fixed.

Damaging moves that heal you were broken.

## What did change?
A variable is changed from `undefined` to `null` to pass a null check properly.

## How to test the changes?
Use an attacking move that heals you based on the damage dealt.

## Checklist

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -1650,7 +1650,7 @@ export class HitHealAttr extends MoveEffectAttr {
     super(true, MoveEffectTrigger.HIT);
 
     this.healRatio = healRatio!; // TODO: is this bang correct?
-    this.healStat = healStat!; // TODO: is this bang correct?
+    this.healStat = healStat ?? null;
   }
   /**
    * Heals the user the determined amount and possibly displays a message about regaining health.


src/data/move.ts
```diff
@@ -1650,7 +1650,7 @@ export class HitHealAttr extends MoveEffectAttr {
     super(true, MoveEffectTrigger.HIT);
 
     this.healRatio = healRatio!; // TODO: is this bang correct?
-    this.healStat = healStat!; // TODO: is this bang correct?
+    this.healStat = healStat ?? null;
```
snoozbuster, line 1653: tragic lol 

if it's not too much trouble it might be more worthwhile to go and find the place that explicitly cares about this being null and change it, but maybe that can happen in a follow-up PR to fix the issue

snoozbuster: I also can't believe there isn't a single healing move test 😱

snoozbuster (approved):

src/data/move.ts
DayKev: Maybe; feels like explicitly setting it to and checking for `null` is better though, as it "declares intent" in a way. 🤷

snoozbuster: if js didn't have undefined I would agree with you but because it has both and you can't really avoid undefined entirely, I've never found a good reason to use null or assign different meanings to null/undefined 🤷. in this case it's probably done because `healStat` is an enum with a `0` value, so you can't actually just check `if(this.healStat)`. we should probably have a simple `isNil()` utility somewhere which is just `x: any => return x === null || x === undefined` for cases like this, so that we don't write code that breaks if it gets something slightly unexpected

PR #3409, opened 2024-08-07
[Dev] Add typecheck command

snoozbuster:
## What are the changes?

Add package.json command to run typechecking

Github pages check will fail if there are type errors. running `npm run docs` locally doesn't surface those errors, and even if it did it does extra stuff anyway. having a dedicated command to check for type errors is useful for development, especially in a post strict-null world

## What did change?

added a package.json command to run `tsc`. I explicitly specified `--noEmit` even though it's in the tsconfig.json, to clearly indicate that the intent is to typecheck and not to actually build code

### Screenshots/Videos

####command with type error:
image

#### no type error:
image

## Checklist
- [ ] ~Are the changes visual?~

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
 		"test:cov": "vitest run --project pre && vitest run --project main --coverage",
 		"test:watch": "vitest run --project pre && vitest watch --project main --coverage",
 		"test:silent": "vitest run --project pre && vitest run --project main --silent",
+		"test:types": "tsc --noEmit",
 		"eslint": "eslint --fix .",
 		"eslint-ci": "eslint .",
 		"docs": "typedoc",


package.json
```diff
@@ -13,6 +13,7 @@
 		"test:cov": "vitest run --project pre && vitest run --project main --coverage",
 		"test:watch": "vitest run --project pre && vitest watch --project main --coverage",
 		"test:silent": "vitest run --project pre && vitest run --project main --silent",
+		"test:types": "tsc --noEmit",
```
flx-sta, line 16: this is not a test. the `docs` script is technically covering it.

A common name for it is `compile` or just `tsc` but please not `test:types`

snoozbuster: it also specifically doesn't compile, even `tsc` by itself doesn't compile because the tsconfig has `noEmit` as well.

it's a command that does nothing except produce an error if there are type errors and nothing if there aren't - so I think in the most general sense, it is a test. but if you want me to rename it `typecheck`, that's another name I've used for this sort of script.

snoozbuster, line 16:
```suggestion
		"typecheck": "tsc --noEmit",
```

flx-sta:
lgtm

flx-sta (approved):

torranx (approved):

PR #3411, opened 2024-08-07
[Move] fully implement pursuit

snoozbuster:
**Opening draft PR to ensure branch looks good after rebasing on beta - move does work in most cases but need to double-check niche scenarios and do full testing**

## What are the changes?

Adding the special effects of Pursuit: https://bulbapedia.bulbagarden.net/wiki/Pursuit_(move)

I tried to use it in a run and got sad when it didn't work

## What did change?

- [ ] TODO - writeup of how the move works, technically speaking

### Screenshots/Videos

- [ ] TODO - ingame screenshots/videos of various scenarios

## How to test the changes?

- [ ] TODO - this section

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -4902,26 +4902,65 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
         return resolve(false);
       }
 
-  	// Move the switch out logic inside the conditional block
-  	// This ensures that the switch out only happens when the conditions are met
 	  const switchOutTarget = this.user ? user : target;
+
+      let willBePursued = false;
+      if (switchOutTarget.hp > 0 && !this.batonPass && this.user && move.id !== Moves.TELEPORT) {
+        switchOutTarget.addTag(BattlerTagType.ESCAPING);
+
+        const opposingField = user.isPlayer() ? user.scene.getEnemyField() : user.scene.getPlayerField();
+        const opposingPursuitUsers = opposingField
+          .filter((op: Pokemon) => op.getTag(BattlerTagType.ANTICIPATING_ACTION)?.sourceMove === Moves.PURSUIT)
+          .sort((a, b) => b.turnData.order - a.turnData.order);
+        if (opposingPursuitUsers.length) {
+          willBePursued = true;
+          opposingPursuitUsers.forEach(pursuiter => {
+            if (user.scene.tryRemovePhase(p => p instanceof MovePhase && p.pokemon.id === pursuiter.id)) {
+              user.scene.prependToPhase(new MovePhase(user.scene, pursuiter, [switchOutTarget.getBattlerIndex()], pursuiter.getMoveset().find(m => m.moveId === Moves.PURSUIT) || new PokemonMove(Moves.PURSUIT), false, false), MoveEndPhase);
+            }
+          });
+        }
+      }
+
 	  if (switchOutTarget instanceof PlayerPokemon) {
-        switchOutTarget.leaveField(!this.batonPass);
+        if (!willBePursued) {
+          switchOutTarget.leaveField(!this.batonPass);
+        }
 
         if (switchOutTarget.hp > 0) {
-          user.scene.prependToPhase(new SwitchPhase(user.scene, switchOutTarget.getFieldIndex(), true, true), MoveEndPhase);
+          user.scene.prependToPhase(
+            new SwitchPhase(user.scene, switchOutTarget.getFieldIndex(), true, willBePursued),
+            MoveEndPhase
+          );
           resolve(true);
         } else {
           resolve(false);
         }
+
 	  	return;
 	  } else if (user.scene.currentBattle.battleType !== BattleType.WILD) {
+        if (!user.scene.currentBattle.trainer) {
+          return resolve(false); // what are we even doing here
+        }
+
 	  	// Switch out logic for trainer battles
-        switchOutTarget.leaveField(!this.batonPass);
+        if (!willBePursued) {
+          switchOutTarget.leaveField(!this.batonPass);
+        }
 
 	  	if (switchOutTarget.hp > 0) {
         // for opponent switching out
-          user.scene.prependToPhase(new SwitchSummonPhase(user.scene, switchOutTarget.getFieldIndex(), (user.scene.currentBattle.trainer ? user.scene.currentBattle.trainer.getNextSummonIndex((switchOutTarget as EnemyPokemon).trainerSlot) : 0), false, this.batonPass, false), MoveEndPhase);
+          user.scene.prependToPhase(
+            new SwitchSummonPhase(
+              user.scene,
+              switchOutTarget.getFieldIndex(),
+              user.scene.currentBattle.trainer.getNextSummonIndex((switchOutTarget as EnemyPokemon).trainerSlot),
+              willBePursued,
+              this.batonPass,
+              false
+            ),
+            MoveEndPhase
+          );
         }
 	  } else {
 	    // Switch out logic for everything else
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -4843,6 +4843,11 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
   private user: boolean;
   private batonPass: boolean;
 
+  /**
+   * @param user Indicates if this move effect will switch out the user of the
+   * move (true) or the target of the move (false).
+   * @param batonPass Indicates if this move is a usage of Baton Pass.
+   */
   constructor(user?: boolean, batonPass?: boolean) {
     super(false, MoveEffectTrigger.POST_APPLY, false, true);
     this.user = !!user;
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -1021,6 +1021,27 @@ export class MessageHeaderAttr extends MoveHeaderAttr {
   }
 }
 
+/**
+ * Applies a tag to a battler when a move is selected, before any other actions
+ * are taken. The tag is applied only for a single turn.
+ */
+export class AddBattlerTagOnMoveReadyAttr extends MoveHeaderAttr {
+  public tagType: BattlerTagType;
+
+  constructor(tagType: BattlerTagType) {
+    super();
+    this.tagType = tagType;
+  }
+
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    if (!super.apply(user, target, move, args)) {
+      return false;
+    }
+
+    return (this.selfTarget ? user : target).addTag(this.tagType, 1, move.id, user.id);
+  }
+}
+
 export class PreMoveMessageAttr extends MoveAttr {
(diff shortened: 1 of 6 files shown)


src/data/move.ts
```diff
@@ -1021,19 +1021,39 @@ export class MessageHeaderAttr extends MoveHeaderAttr {
...
+}
+
 /**
  * Header attribute to implement the "charge phase" of Beak Blast at the
  * beginning of a turn.
  * @see {@link https://bulbapedia.bulbagarden.net/wiki/Beak_Blast_(move) | Beak Blast}
  * @see {@linkcode BeakBlastChargingTag}
  */
-export class BeakBlastHeaderAttr extends MoveHeaderAttr {
+export class BeakBlastHeaderAttr extends AddBattlerTagOnMoveReadyAttr {
```
snoozbuster, line 1051: minor refactor to use the general-purpose move attr I added for applying battler tags

src/enums/battler-tag-type.ts
```diff
@@ -67,5 +67,17 @@ export enum BattlerTagType {
   IGNORE_DARK = "IGNORE_DARK",
   GULP_MISSILE_ARROKUDA = "GULP_MISSILE_ARROKUDA",
   GULP_MISSILE_PIKACHU = "GULP_MISSILE_PIKACHU",
+  /**
+   * Indicates the battler is waiting for their opponent to make some
+   * sort of action (switch out, use a type of move, make contact, etc)
+   */
+  ANTICIPATING_ACTION = "ANTICIPATING_ACTION",
```
snoozbuster, line 74: I thought this might also be useful for moves like shell trap and snatch, so that we don't keep adding special purpose tags for every move (especially since the move that applied the tag is available from `getTag()`).

src/data/move.ts
```diff
@@ -4867,26 +4907,76 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
...
+
 	  	// Switch out logic for trainer battles
-        switchOutTarget.leaveField(!this.batonPass);
+        if (!willBePursued) {
+          switchOutTarget.leaveField(!this.batonPass);
+        }
 
 	  	if (switchOutTarget.hp > 0) {
         // for opponent switching out
-          user.scene.prependToPhase(new SwitchSummonPhase(user.scene, switchOutTarget.getFieldIndex(), (user.scene.currentBattle.trainer ? user.scene.currentBattle.trainer.getNextSummonIndex((switchOutTarget as EnemyPokemon).trainerSlot) : 0), false, this.batonPass, false), MoveEndPhase);
```
snoozbuster, line 4889: only actual change here is the change of the `doReturn` parameter to be based on `willBePursued`, the rest is just breaking it up for readability

src/data/move.ts
```diff
@@ -4867,26 +4907,76 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
...
+          );
           resolve(true);
         } else {
           resolve(false);
         }
+
 	  	return;
 	  } else if (user.scene.currentBattle.battleType !== BattleType.WILD) {
+        if (!user.scene.currentBattle.trainer) {
+          return resolve(false); // what are we even doing here
```
snoozbuster, line 4959: this is mainly to make TS happy. if the current battleType is not wild but there's no trainer, I expect this will not be the only thing that's unhappy, but at least we won't crash here.

src/data/move.ts
```diff
@@ -4867,26 +4907,76 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
...
+                MoveEndPhase
+              );
+            }
+          });
+        }
+      }
+
 	  if (switchOutTarget instanceof PlayerPokemon) {
-        switchOutTarget.leaveField(!this.batonPass);
+        if (!willBePursued) {
```
snoozbuster, line 4942: A strange quirk of the way that pursuit interacts with switch out moves is that the pokemon does the animation for uturn or whatever, but then can't actually leave the field because we need to do the pursuit animation and effects before the switch. `SwitchSummonPhase` is able to do the `leaveField` itself (that's the `doReturn` parameter) so we can shift the logic to there when the switch move would be intercepted by a pursuit. it causes the animation to look a little funny, but I'm pretty sure it works the same in the few official games that have both pursuit and uturn implemented

src/data/move.ts
```diff
@@ -4867,26 +4907,76 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
         return resolve(false);
       }
 
-  	// Move the switch out logic inside the conditional block
-  	// This ensures that the switch out only happens when the conditions are met
 	  const switchOutTarget = this.user ? user : target;
+
+      let willBePursued = false;
+      if (switchOutTarget.hp > 0 && !this.batonPass && this.user && move.id !== Moves.TELEPORT) {
```
snoozbuster, line 4913: teleport and baton pass are the only self-switch moves which don't trigger pursuit. (technically speaking, we have no idea if parting shot would or would not, since they never existed at the same time - but my suspicion based on the patterns I see is that if they did, pursuit would hit a parting shot user.)

no moves that force an opponent out interact with pursuit (no pursuit + dragon tail or pursuit + roar, for example).

src/data/move.ts
```diff
@@ -4867,26 +4907,76 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
         return resolve(false);
       }
 
-  	// Move the switch out logic inside the conditional block
-  	// This ensures that the switch out only happens when the conditions are met
```
snoozbuster, line 4871: honestly, I have spent hours looking at this function over the last couple weeks, and I still have no idea what this comment means which the code itself does not clearly indicate, so I removed it.

src/data/move.ts
```diff
@@ -3714,6 +3734,21 @@ export class BlizzardAccuracyAttr extends VariableAccuracyAttr {
   }
 }
 
+const isPursuingFunc = (user: Pokemon, target: Pokemon) =>
+  user.getTag(BattlerTagType.ANTICIPATING_ACTION) && target.getTag(BattlerTagType.ESCAPING);
+
+export class PursuitAccuracyAttr extends VariableAccuracyAttr {
```
snoozbuster, line 3740: I didn't see any good way to implement the "pursuit guaranteed hits on escaping targets" other than to use a custom attr - let me know if I missed something like `MovePowerMultiplierAttr` that allows accuracy to be adjusted based on a function and I'll happily switch to it

src/overrides.ts
```diff
@@ -115,6 +115,11 @@ class DefaultOverrides {
   readonly OPP_VARIANT_OVERRIDE: Variant = 0;
   readonly OPP_IVS_OVERRIDE: integer | integer[] = [];
 
+  // --------------------------
+  // TRAINER/AI OVERRIDES
+  // --------------------------
+  readonly TRAINER_ALWAYS_SWITCHES: boolean = false;
```
snoozbuster, line 121: added an override for modifying the enemy AI, based on [`gameManager#forceOpponentToSwitch`](https://github.com/pagefaultgames/pokerogue/blob/a07d2c57a44361f4ad48e024067c9b7638c0d84d/src/test/utils/gameManager.ts#L218). I added this in a new section with a new prefix since there aren't any overrides right now that specifically affect enemy trainers/AI, and I thought that sort of thing might be useful to have in the future.

src/phases.ts
```diff
@@ -4641,52 +4677,56 @@ export class PostGameOverPhase extends Phase {
...
+   * `'faint'` (party member fainted), `'moveEffect'` (uturn, baton pass, dragon
+   * tail, etc), and `'switchMode'` (start-of-battle optional switch). This
+   * helps the phase determine both if the switch should be cancellable by the
+   * user, as well as determine if the party UI should be shown at all.
+   * @param doReturn Indicates if this switch should call back the pokemon at
+   * the {@linkcode fieldIndex} (true), or if the mon has already been recalled
+   * (false).
    */
-  constructor(scene: BattleScene, fieldIndex: integer, isModal: boolean, doReturn: boolean) {
+  constructor(scene: BattleScene, fieldIndex: integer, switchReason: "faint" | "moveEffect" | "switchMode", doReturn: boolean) {
```
snoozbuster, line 4696: I don't love this string union approach, but I couldn't figure out a way to implement the correct logic for all three usages of SwitchPhase without specifically indicating to the phase both whether the switch is due to a faint, and if the switch should be forced. instead of adding a third boolean parameter I thought this would be more obvious; and I didn't want to add yet another single-purpose enum for this when typescript will give you perfect type-safety as well as autocomplete on this type of union. 

if anyone has alternative suggestions, let me know, but the phase order here seems to constrain the possibilities. I would rather put some of this logic in the calling code, but I don't think that's possible because the game state can and often does change between the time this phase is queued and the time it actually runs.

src/phases.ts
```diff
@@ -2307,14 +2307,22 @@ export class TurnStartPhase extends FieldPhase {
...
+            return -1;
+          } else {
+            return 1;
+          }
+        } else if (bCommand.command === Command.FIGHT) {
+          if (bCommand.move?.move === Moves.PURSUIT && aCommand.command === Command.POKEMON) {
+            return 1;
+          } else {
+            return -1;
+          }
```
snoozbuster, line 2325: sorts pursuit commands before Pokemon commands. need to take another look at this after writing tests, I think this logic can be implemented with less nesting/less if-else.

snoozbuster:
Current PR state:

- Fainting a pokemon with a pursuit causes massive issues to the game state (has failing tests)
- There are a couple niche scenarios in doubles where the `.sort()` in `TurnStartPhase` produces the wrong ordering and ultimately causes pursuit to either run before it should or after it should. I kinda want to wait for #3547 to merge before I work on these because it will be much easier to assert the correct results from the `getCommandOrder` function in that PR for a variety of scenarios, and then just have a couple tests where I actually run through the phases just to make sure that even if the command order is correct, the phases don't get mixed up. as it stands the ways to observe this issue are pretty weird or far-removed from the actual problem. 
- there is a set of research I need to do to figure out some really niche interactions with what happens when there are multiple pursuits/switches/uturns + a faint from pursuit on the same turn in doubles, and then I need to write tests for those and see if they pass. specifically the cases I can think of are:
  - two pokemon pursue a single hard-switching target, which faints. what happens to the second pursuit? (fail, hits switch-in, hits original target)
  - two pokemon pursue a single uturning target, which faints. what happens to the second pursuit?
  - two pokemon pursue against two hard-switches, and the first one causes a faint. what happens to the second pursuit? (fail, hits other switching pokemon, etc)
  - two pokemon pursue against a hard-switch and a uturn, hard-switcher faints. what happens to the second pursuit?
  - two pokemon pursue agains two uturns and the first faints. what happens to the second pursuit?

frutescens:
Hi! Hope you're doing well. The turnStartPhase refactor got pushed into Beta

SirzBenjie:
This is really old and will need to be revisited. I am closing this for now. It can be reopened if there is interest in updaing it

PR #3195, opened 2024-07-29
[BUG] - Re: Parting shot debuff applies even if move fails

DustinLin:
duplicate of Pr #3135 
---

## What are the changes?

When parting shot fails, the debuffs no longer apply

## Why am I doing these changes?

Closes #3112 

## What did change?

Added check in move workflow that doesn't apply a move's attributes when a move fails

### Screenshots/Videos

- `beta`

https://github.com/user-attachments/assets/d4922730-e09f-4654-9d3e-7ab213b9a73f

https://github.com/user-attachments/assets/837e7ae6-c58f-402c-a2aa-0ed25b18578e

- `pr`

https://github.com/user-attachments/assets/05395f98-8365-47b5-81a3-a699db4c8d63

https://github.com/user-attachments/assets/5dc9342b-2747-47b3-a9a6-305e7e6d4518

## How to test the changes?

Using overrides, the 2 cases were:
- user of parting shot has ability Prankster and the opponent is a dark type
- opponent has the ability Good as Gold

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I considered writing automated tests for the issue?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #3412, opened 2024-08-07
[TEST] - adding parting shot move test

DustinLin:

Adding tests that cover PR #3195 (for parting shot), which originally fix issue #3112: that parting shot still debuffs when the move is supposed to fail

## What did change?

More tests
## How to test the changes?

`npm test src/test/moves/parting_shot.test.ts`
## Checklist

### DIFF
--- a/src/test/moves/parting_shot.test.ts
+++ b/src/test/moves/parting_shot.test.ts
@@ -36,34 +36,29 @@ describe("Moves - Parting Shot", () => {
   });
 
   test(
-    "Parting shot when buffed by prankster should fail against dark types",
+    "Parting Shot when buffed by prankster should fail against dark types",
     async () => {
-      await game.startBattle([Species.MURKROW]);
-      game.override.ability(Abilities.PRANKSTER);
-      game.override.enemySpecies(Species.POOCHYENA);
-
-      const enemyPokemon = game.scene.getEnemyPokemon()!;
-      expect(enemyPokemon).toBeDefined();
-
+      await game.startBattle([Species.MURKROW, Species.MEOWTH]);
+      game.override
+        .enemySpecies(Species.POOCHYENA)
+        .ability(Abilities.PRANKSTER);
       game.doAttack(getMovePosition(game.scene, 0, Moves.PARTING_SHOT));
 
       await game.phaseInterceptor.to(BerryPhase, false);
       const battleStatsOpponent = game.scene.currentBattle.enemyParty[0].summonData.battleStats;
       expect(battleStatsOpponent[BattleStat.ATK]).toBe(0);
       expect(battleStatsOpponent[BattleStat.SPATK]).toBe(0);
-
-
+      expect(game.scene.getPlayerField()[0].species.speciesId).toBe(Species.MURKROW);
     }, TIMEOUT
   );
 
   test(
-    "Parting shot when buffed by prankster should fail against good for gold ability",
+    "Parting shot when buffed by prankster should fail against good as gold ability",
     async () => {
-      await game.startBattle([Species.MURKROW]);
-      game.override.ability(Abilities.PRANKSTER);
-      game.override.enemySpecies(Species.GHOLDENGO);
-      game.override.enemyAbility(Abilities.GOOD_AS_GOLD);
-
+      await game.startBattle([Species.MURKROW, Species.MEOWTH]);
+      game.override
+        .enemySpecies(Species.GHOLDENGO)
+        .enemyAbility(Abilities.GOOD_AS_GOLD);
       const enemyPokemon = game.scene.getEnemyPokemon()!;
       expect(enemyPokemon).toBeDefined();
 
@@ -73,6 +68,7 @@ describe("Moves - Parting Shot", () => {
       const battleStatsOpponent = game.scene.currentBattle.enemyParty[0].summonData.battleStats;
       expect(battleStatsOpponent[BattleStat.ATK]).toBe(0);
       expect(battleStatsOpponent[BattleStat.SPATK]).toBe(0);
+      expect(game.scene.getPlayerField()[0].species.speciesId).toBe(Species.MURKROW);
     }, TIMEOUT
   );
 });
--- a/src/test/moves/parting_shot.test.ts
+++ b/src/test/moves/parting_shot.test.ts
@@ -38,10 +38,11 @@ describe("Moves - Parting Shot", () => {
   test(
     "Parting Shot when buffed by prankster should fail against dark types",
     async () => {
-      await game.startBattle([Species.MURKROW, Species.MEOWTH]);
       game.override
         .enemySpecies(Species.POOCHYENA)
         .ability(Abilities.PRANKSTER);
+      await game.startBattle([Species.MURKROW, Species.MEOWTH]);
+
       game.doAttack(getMovePosition(game.scene, 0, Moves.PARTING_SHOT));
 
       await game.phaseInterceptor.to(BerryPhase, false);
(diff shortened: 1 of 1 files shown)


src/test/moves/parting_shot.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      expect(battleStatsOpponent[BattleStat.SPATK]).toBe(0);
+
+
+    }, TIMEOUT
+  );
+
+  test(
+    "Parting shot when buffed by prankster should fail against good for gold ability",
+    async () => {
+      await game.startBattle([Species.MURKROW]);
```
Tempo-anon, line 62: Shouldn't this be a non dark type to make sure this test doesn't intersect with the above test

src/test/moves/parting_shot.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      game.override.ability(Abilities.PRANKSTER);
+
+      const enemyPokemon = game.scene.getEnemyPokemon()!;
+      expect(enemyPokemon).toBeDefined();
+
+      game.doAttack(getMovePosition(game.scene, 0, Moves.PARTING_SHOT));
+
+      await game.phaseInterceptor.to(BerryPhase);
```
innerthunder, line 50:
```suggestion
      await game.phaseInterceptor.to(BerryPhase, false);
```
Less room for items to affect the test outcome

src/test/moves/parting_shot.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      await game.startBattle([Species.MURKROW]);
+      game.override.ability(Abilities.PRANKSTER);
+      game.override.enemyAbility(Abilities.GOOD_AS_GOLD);
+
+      const enemyPokemon = game.scene.getEnemyPokemon()!;
+      expect(enemyPokemon).toBeDefined();
+
+      game.doAttack(getMovePosition(game.scene, 0, Moves.PARTING_SHOT));
+
+      await game.phaseInterceptor.to(BerryPhase);
```
innerthunder, line 71:
```suggestion
      await game.phaseInterceptor.to(BerryPhase, false);
```
Ditto

src/test/moves/parting_shot.test.ts
DustinLin: o yeah you are right, thanks for the catch

DustinLin:
ty for comments, updated

src/test/moves/parting_shot.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      const battleStatsOpponent = game.scene.currentBattle.enemyParty[0].summonData.battleStats;
+      expect(battleStatsOpponent[BattleStat.ATK]).toBe(0);
+      expect(battleStatsOpponent[BattleStat.SPATK]).toBe(0);
+
+
+    }, TIMEOUT
+  );
+
+  test(
+    "Parting shot when buffed by prankster should fail against good for gold ability",
```
snoozbuster, line 60:
```suggestion
    "Parting shot when buffed by prankster should fail against good as gold ability",
```

minor

src/test/moves/parting_shot.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    game.override.battleType("single");
+    game.override.moveset([Moves.PARTING_SHOT, Moves.SPLASH]);
+    game.override.enemyMoveset(SPLASH_ONLY);
+    game.override.startingLevel(5);
+    game.override.enemyLevel(5);
+
+  });
+
+  test(
+    "Parting shot when buffed by prankster should fail against dark types",
```
snoozbuster, line 39: I'm not familiar with how parting shot works - if the move fails, it probably also shouldn't switch out the user, right? it might be worth setting up the test with an extra mon in the back and asserting that the speciesId of the active pokemon is still the parting shot user to verify that neither the switch nor the stat drop happens (if that's the correct behavior, anyway)

src/test/moves/parting_shot.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+
+    }, TIMEOUT
+  );
+
+  test(
+    "Parting shot when buffed by prankster should fail against good for gold ability",
+    async () => {
+      await game.startBattle([Species.MURKROW]);
+      game.override.ability(Abilities.PRANKSTER);
```
snoozbuster, line 63: is prankster relevant here? I thought good as gold just generally prevented status moves, whether or not they were prankster-buffed or not. shouldn't parting shot always fail against good as gold?

src/test/moves/parting_shot.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      expect(enemyPokemon).toBeDefined();
+
+      game.doAttack(getMovePosition(game.scene, 0, Moves.PARTING_SHOT));
+
+      await game.phaseInterceptor.to(BerryPhase, false);
+      const battleStatsOpponent = game.scene.currentBattle.enemyParty[0].summonData.battleStats;
+      expect(battleStatsOpponent[BattleStat.ATK]).toBe(0);
+      expect(battleStatsOpponent[BattleStat.SPATK]).toBe(0);
+
+
```
snoozbuster, line 55:
```suggestion
```

probably don't need these newlines

src/test/moves/parting_shot.test.ts
DustinLin: yea it's best to separate them

src/test/moves/parting_shot.test.ts
DustinLin: good suggestion

torranx:
I feel like we can add more tests from [Parting Shot](https://bulbapedia.bulbagarden.net/wiki/Parting_Shot_(move))

DustinLin:
uh oh why is the test RNG failing....

snoozbuster (approved):
seems like a great start to me

innerthunder:
Some other tests I'd like to see:
- should do nothing against a target with Clear Body
- should do nothing while Mist is in effect
- should do nothing against a target with -6 Atk / -6 Sp. Atk (optional: also against a target with +6 / +6 and Contrary)
- shouldn't prompt a switch when the user is the last non-fainted player Pokemon

If any of these don't work, you can just skip them (`it.skip(...);`) and leave the debugging for another PR (open an issue for any bugs you encounter, though).

src/test/moves/parting_shot.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      await game.startBattle([Species.MURKROW]);
+
+      const enemyPokemon = game.scene.getEnemyPokemon()!;
+      expect(enemyPokemon).toBeDefined();
+
+      game.doAttack(getMovePosition(game.scene, 0, Moves.PARTING_SHOT));
+
+      await game.phaseInterceptor.to(BerryPhase, false);
+      const battleStatsOpponent = game.scene.currentBattle.enemyParty[0].summonData.battleStats;
+      expect(battleStatsOpponent[BattleStat.ATK]).toBe(0);
```
innerthunder, line 89: Only the "switch-out" portion of the move should fail in this case; Parting Shot should still lower the opponent's stats.

Source https://www.smogon.com/dex/sv/moves/parting-shot/

DustinLin: you are right (tested on showdown as well), seems like this move was incorrectly implemented. I'll fix the test but mark it so it won't run? Fixing the move can come in a different PR

DustinLin: Issue #3471

I could fix this up after I finish these test cases

DustinLin:
Added more tests based on innerthunder's checklist (same as the ones on [bulba](https://bulbapedia.bulbagarden.net/wiki/Parting_Shot_(move))

many of them fail currently, so are skipped

innerthunder (approved):

PR #3420, opened 2024-08-07
[Bug] Some fresh start fixes

Tempo-anon:
## What are the changes?
* Starters in fresh start are not hardcoded as 3 cost
* Mini Black Hole disabled in fresh start
* No catching allowed in the last biome in fresh start

Makes fresh start more inline to a true fresh start. Note that there are still some edge cases involving dexAttr which are not easily solved.

## What did change?
* Created 2 new helper functions in game-mode.ts
* Added a fresh start check for mini black hole and eviolite weight
* Added a fresh start check for catching in the last biome

## How to test the changes?
* Change starter costs from 3 and see if it's correctly reflected in fresh start
* Change MBH weight super high and see if it will spawn in fresh start
* Try catching in the last biome in fresh start

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/challenge.ts
+++ b/src/data/challenge.ts
@@ -636,7 +636,7 @@ export class FreshStartChallenge extends Challenge {
 
   applyStarterCost(species: Species, cost: Utils.NumberHolder): boolean {
     if (defaultStarterSpecies.includes(species) && cost.value !== 3) {
-      cost.value = 3;
+      cost.value = speciesStarters[species];
       return true;
     }
     return false;
--- a/src/data/challenge.ts
+++ b/src/data/challenge.ts
@@ -635,7 +635,7 @@ export class FreshStartChallenge extends Challenge {
   }
 
   applyStarterCost(species: Species, cost: Utils.NumberHolder): boolean {
-    if (defaultStarterSpecies.includes(species) && cost.value !== 3) {
+    if (defaultStarterSpecies.includes(species)) {
       cost.value = speciesStarters[species];
       return true;
     }
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -1618,7 +1618,7 @@ const modifierPool: ModifierPool = {
     new WeightedModifierType(modifierTypes.MULTI_LENS, 18),
     new WeightedModifierType(modifierTypes.VOUCHER_PREMIUM, (party: Pokemon[], rerollCount: integer) => !party[0].scene.gameMode.isDaily && !party[0].scene.gameMode.isEndless && !party[0].scene.gameMode.isSplicedOnly ? Math.max(5 - rerollCount * 2, 0) : 0, 5),
     new WeightedModifierType(modifierTypes.DNA_SPLICERS, (party: Pokemon[]) => !party[0].scene.gameMode.isSplicedOnly && party.filter(p => !p.fusionSpecies).length > 1 ? 24 : 0, 24),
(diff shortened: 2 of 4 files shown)


Xavion3:
I'd really prefer if we just had new challenge types, because challenges wanting to modify whether you can catch a pokemon or whether you can get a particular item are both useful for other challenges and also it means we don't need to be special casing things.

damocleas:
> I'd really prefer if we just had new challenge types, because challenges wanting to modify whether you can catch a pokemon or whether you can get a particular item are both useful for other challenges and also it means we don't need to be special casing things.

I have no idea what you mean by this at all.

Xavion3:
It's technical implementation things, I'm just saying I'd prefer it to be coded a different way so it's in line with how challenges do things instead of a hacky short term fix we'll want to change later anyway.

It has no impact on whether or not I think these features are warranted.

src/data/challenge.ts
```diff
@@ -636,7 +636,7 @@ export class FreshStartChallenge extends Challenge {
 
   applyStarterCost(species: Species, cost: Utils.NumberHolder): boolean {
     if (defaultStarterSpecies.includes(species) && cost.value !== 3) {
-      cost.value = 3;
+      cost.value = speciesStarters[species];
```
DayKev, line 639: Is the `if` even needed anymore?

src/modifier/modifier-type.ts
```diff
@@ -1623,7 +1623,7 @@ const modifierPool: ModifierPool = {
     new WeightedModifierType(modifierTypes.MULTI_LENS, 18),
     new WeightedModifierType(modifierTypes.VOUCHER_PREMIUM, (party: Pokemon[], rerollCount: integer) => !party[0].scene.gameMode.isDaily && !party[0].scene.gameMode.isEndless && !party[0].scene.gameMode.isSplicedOnly ? Math.max(5 - rerollCount * 2, 0) : 0, 5),
     new WeightedModifierType(modifierTypes.DNA_SPLICERS, (party: Pokemon[]) => !party[0].scene.gameMode.isSplicedOnly && party.filter(p => !p.fusionSpecies).length > 1 ? 24 : 0, 24),
-    new WeightedModifierType(modifierTypes.MINI_BLACK_HOLE, (party: Pokemon[]) => party[0].scene.gameData.unlocks[Unlockables.MINI_BLACK_HOLE] ? 1 : 0, 1),
+    new WeightedModifierType(modifierTypes.MINI_BLACK_HOLE, (party: Pokemon[]) => (!party[0].scene.gameMode.isFreshStartChallenge() && party[0].scene.gameData.unlocks[Unlockables.MINI_BLACK_HOLE]) ? 1 : 0, 1),
```
xsn34kzx, line 1626:
```suggestion
    new WeightedModifierType(modifierTypes.MINI_BLACK_HOLE, (party: Pokemon[]) => {
      const { scene } = party[0];
      return scene.gameMode.isFreshStartChallenge() && scene.gameData.unlocks[Unlockables.MINI_BLACK_HOLE]) ? 1 : 0
    }, 1),
```
+1 to this comment (https://github.com/pagefaultgames/pokerogue/pull/3420#discussion_r1711932591), we can apply the same syntax to this entry to be a bit cleaner.

xsn34kzx (commented):
Pretty much cosigning flx's syntax comments. Since this is more or less a band-aid for now, I'll probably look into making disabling unlocks, etc. apart of Fresh Start in general like Xavion was saying.

PR #3327, opened 2024-08-04
[Ability] Partially implement Gulp Missile

torranx:
## What are the changes?
partially implements gulp missile:
**Does not transform when Surf/Dive misses/is protected**

## Why am I doing these changes?
unimplemented ability

## What did change?
- added new abattr class `PostDefendGulpMissileAbAttr` 
- added new battlertag `GulpMissileTag`
- added new move attr `GulpMissileTagAttr`

### Screenshots/Videos
- Gulping form

https://github.com/user-attachments/assets/2723ffe4-1770-413f-a611-7a7e2f5e1f03

- Gorging form

https://github.com/user-attachments/assets/a3821217-24ea-426e-a763-aa24f509a0ec

## How to test the changes?
overrides used: 
```
{
  BATTLE_TYPE_OVERRIDE: "single",
  STARTER_SPECIES_OVERRIDE: Species.CRAMORANT,
  ABILITY_OVERRIDE: Abilities.GULP_MISSILE,
  MOVESET_OVERRIDE: [Moves.DIVE, Moves.SURF, Moves.SPLASH, Moves.QUICK_ATTACK],
  OPP_LEVEL_OVERRIDE: 5,
  OPP_MOVESET_OVERRIDE: Array(4).fill(Moves.TACKLE)
}
```
`$ npm run test gulp_missile`

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I considered writing automated tests for the issue?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [ ] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #3438, opened 2024-08-08
[Move] Fully Implement Gulp Missile

innerthunder:
## What are the changes?

This change fully implements Gulp Missile by fixing an interaction with Dive (Cramorant now changes form during Dive's first turn).

Gulp Missile was partially implemented in https://github.com/pagefaultgames/pokerogue/pull/3327. This finishes the implementation.

## What did change?

- `data/move`: Dive's `ChargeAttr` now uses the `chargeEffect` parameter to trigger Gulp Missile's form change on the move's first turn of use.
- `data/ability`:
  - Added a check for `SemiInvulnerableTag` so Gulp Missile no longer activates while Cramorant is underwater.
  - Removed `.partial()` from Gulp Missile's attributes.
- `form-change-phase`: Added a check for `SemiInvulnerableTag` so quiet form changes don't play animations while the target Pokemon is hidden from view
- `test/abilities/gulp_missile.test`: Added a test to check that Cramorant changes form on the first turn of using Dive.

### Screenshots/Videos

Video of Cramorant changing forms when using Dive:

https://github.com/user-attachments/assets/957ba842-168b-4deb-b5da-19787e1086fd

## How to test the changes?

`npm run test gulp_missile`

## Checklist
- [x] Are the changes visual?


src/data/ability.ts
```diff
@@ -517,7 +517,7 @@ export class PostDefendGulpMissileAbAttr extends PostDefendAbAttr {
    */
   applyPostDefend(pokemon: Pokemon, passive: boolean, attacker: Pokemon, move: Move, hitResult: HitResult, args: any[]): boolean | Promise<boolean> {
     const battlerTag = pokemon.getTag(GulpMissileTag);
-    if (!battlerTag || move.category === MoveCategory.STATUS) {
+    if (!battlerTag || move.category === MoveCategory.STATUS || pokemon.getTag(SemiInvulnerableTag)) {
```
Tempo-anon, line 520: Does gulp missile just not fire off if the user is semi invulnerable or just for dive? If Cramorant has gulp missile active and is flying and gets hit it should not fire right? Can never be too sure after seeing the weird surf/dive interactions

src/form-change-phase.ts
```diff
@@ -194,7 +195,7 @@ export class QuietFormChangePhase extends BattlePhase {
 
     const preName = getPokemonNameWithAffix(this.pokemon);
 
-    if (!this.pokemon.isOnField()) {
+    if (!this.pokemon.isOnField() || this.pokemon.getTag(SemiInvulnerableTag)) {
```
Tempo-anon, line 198: Ah so it's still changing form even when semi invulnerable, it's just silent?

innerthunder: Yeah, you can see the form change message in the video

src/data/ability.ts
innerthunder: Checked on Showdown:

[Gen9BalancedHackmons-2024-08-08-inthunder-innerthunder.zip](https://github.com/user-attachments/files/16552023/Gen9BalancedHackmons-2024-08-08-inthunder-innerthunder.zip)

1 Cram has Gulp Missile, the other has No Guard:
1. Gulp Missile Cram uses Surf to change form
2. GM Cram uses Fly, then No Guard Cram hits the flying Cram with Surf
3. GM Cram's Fly is interrupted, but GM doesn't fire

So yeah, this should apply to all forms of semi-invulnerability.

PR #2702, opened 2024-06-29
[Bug] Fix Jaw Lock not removing TRAPPED tag after enemy faints

EmberCM:
## What are the changes?
Fix Jaw Lock functionality

## Why am I doing these changes?
Fixes #170 

## What did change?
Add `BattlerTypeTag.JAW_LOCK`
Create `JawLockTag` which extends the `TrappedTag` BattlerTag
Save the fainted Pokemon's BattlerTags in the `FaintPhase`

### Screenshots/Videos
Before

https://github.com/pagefaultgames/pokerogue/assets/17890775/7466c9f7-52c6-4de1-a581-26ac11f82f69

After

https://github.com/pagefaultgames/pokerogue/assets/17890775/b0638b83-f8d8-44fb-97ca-71ab2588d07a

## How to test the changes?
```ts
export const MOVESET_OVERRIDE: Array<Moves> = [Moves.JAW_LOCK];
```

## Checklist
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

tests

PR #3450, opened 2024-08-08
[Bug] Fix Jaw Lock leaving the user trapped after the target faints

### LINKED ISSUE #170: Killing a wild Pokémon with Jaw lock still leaves the user trapped in the next battle
What the title says. I haven't used Jaw Lock again to further investigate because it's too expensive to watch Dreadnaw be chipped away to death lol. I figure this only occurs during waves of wild Pokémon and should be working properly during trainer battles.

innerthunder:
## What are the changes?

This fixes an issue where if a Pokemon uses Jaw Lock on a target, then KOs the target on a following turn, the user would remain trapped.

This is a mirror of #2702 , which closes #170 

## What did change?

- `data/battler-tags`: New `JawLockTag` to implement Jaw Lock's trapping effect on a Pokemon
- `data/move`: Jaw Lock now uses a new attribute `AddLinkedBattlerTagAttr` that adds a tag to the user and target, with each tag's `sourceId` being of the other Pokemon.
- `test/moves/jaw_lock.test`: New integration test suite checking how Jaw Lock's battler tags are maintained in different battle states.

### Screenshots/Videos

See #2702 for test videos.

## How to test the changes?

`npm run test jaw_lock`

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -4390,6 +4390,30 @@ export class GulpMissileTagAttr extends MoveEffectAttr {
...
+ * Attribute used for moves that add a tag to the user and target, such as Jaw Lock
+ * @extends AddBattlerTagAttr
+ */
+export class AddLinkedBattlerTagAttr extends AddBattlerTagAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    if (!super.canApply(user, target, move, args)) {
+      return false;
+    }
+
+    const moveChance = this.getMoveChance(user,target,move,this.selfTarget);
```
torranx, line 4403:
```suggestion
    const moveChance = this.getMoveChance(user, target, move, this.selfTarget);
```

src/test/moves/jaw_lock.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+      const leadPokemon = game.scene.getPlayerPokemon()!;
+      const enemyPokemon = game.scene.getEnemyPokemon()!;
+
+      game.doAttack(getMovePosition(game.scene, 0, Moves.JAW_LOCK));
+
+      await game.phaseInterceptor.to(MoveEffectPhase);
+
+      expect(leadPokemon.getTag(BattlerTagType.JAW_LOCK)).toBeDefined();
+      expect(enemyPokemon.getTag(BattlerTagType.JAW_LOCK)).toBeDefined();
```
torranx, line 104: why is the tag defined here on `MoveEffectPhase` but on other tests, its not?

innerthunder: the other tests use
```ts
await game.phaseInterceptor.to(MoveEffectPhase, false)
```
so they check for the tag before `MoveEffectPhase` gets to run.

torranx (approved):
code's lgtm (did not test)

src/test/moves/jaw_lock.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+      await game.phaseInterceptor.to(FaintPhase);
+
+      expect(leadPokemon.getTag(BattlerTagType.JAW_LOCK)).toBeUndefined();
+      expect(enemyPokemon.getTag(BattlerTagType.JAW_LOCK)).toBeUndefined();
+    }, TIMEOUT
+  );
+
+  it(
+    "should only trap the user until the target faints",
```
Tempo-anon, line 92: Shouldn't there be an `expect` in here to see the user and target get the jaw lock tag when the move is used but before either of the target is KO'd?

src/data/move.ts
```diff
@@ -4390,6 +4390,30 @@ export class GulpMissileTagAttr extends MoveEffectAttr {
   }
 }
 
+/**
+ * Attribute used for moves that add a tag to the user and target, such as Jaw Lock
+ * @extends AddBattlerTagAttr
+ */
+export class AddLinkedBattlerTagAttr extends AddBattlerTagAttr {
```
Tempo-anon, line 4397: Is this tag ever removed? I don't think I see that part

src/test/moves/jaw_lock.test.ts
Tempo-anon, line 1: nit: Should probably also include some edge case tests for like double battles since if your Pokemon jaw locks opponent A, then jaw locks opponent B, then KOs opponent A they should still be trapped until opponent B is also KO'd

src/data/move.ts
innerthunder: Only on faint (where `lapseTags(BattlerTagType.FAINT)` is called) and whenever `Pokemon.resetSummonData()` is called

src/test/moves/jaw_lock.test.ts
innerthunder: [Bulba](https://bulbapedia.bulbagarden.net/wiki/Jaw_Lock_(move)) says the trap effect shouldn't apply to opponent B at all in this case, since the user is already trapped.

Seems like I need to add some logic to account for overlapping effects either way, though.

frutescens (approved):

src/data/move.ts
```diff
@@ -4403,6 +4403,35 @@ export class GulpMissileTagAttr extends MoveEffectAttr {
...
```
f-fsantos, line 4410: Should `Destiny Bond` also apply to this?

frutescens: Would moves like octolock, thousand waves, spirit shackle, and anchor shot also benefit from this attr as well?

frutescens: Possible Reference: 
https://bulbapedia.bulbagarden.net/wiki/Escape_prevention
Future Sight and Doom Desire too maybe?

innerthunder: As discussed earlier in the Discord, this only applies to Jaw Lock

flx-sta (approved):

xsn34kzx (approved):

Snailman11:
Link to Discord's (Jaw Lock Still Prevents User Switch Out after making opponent faint)
https://discord.com/channels/1125469663833370665/1271997915556089936

PR #3456, opened 2024-08-09
[Enhancement] Allow nickname assignment in the starter select UI.

KimJeongSun:
## What are the changes?

In the starter select UI, it is now possible to set nicknames for Pokémon!

Many users have requested this feature (myself included!).

## What did change?

1. Added a rename menu for setting nicknames.
2. Included a nickname attribute in the starter settings, which will be used during the `tryStart` process.
3. Assigned this feature to integrate with the current nickname functionality.
( so, The rename settings in the start UI are only used for initializing the starter party in a new run. )

### Screenshots/Videos

https://github.com/user-attachments/assets/0c6ebf96-b9a2-4bda-844e-cff7d8dc5c44

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/ui/rename-form-ui-handler.ts
+++ b/src/ui/rename-form-ui-handler.ts
@@ -36,7 +36,7 @@ export default class RenameFormUiHandler extends FormModalUiHandler {
   show(args: any[]): boolean {
     if (super.show(args)) {
       const config = args[0] as ModalConfig;
-      if (args[1] && typeof (args[1] as any).getNameToRender === "function") {
+      if (args[1] && typeof (args[1] as PlayerPokemon).getNameToRender === "function") {
         this.inputs[0].text = (args[1] as PlayerPokemon).getNameToRender();
       } else {
         this.inputs[0].text = args[1];
--- a/src/ui/rename-form-ui-handler.ts
+++ b/src/ui/rename-form-ui-handler.ts
@@ -36,17 +36,13 @@ export default class RenameFormUiHandler extends FormModalUiHandler {
   show(args: any[]): boolean {
     if (super.show(args)) {
       const config = args[0] as ModalConfig;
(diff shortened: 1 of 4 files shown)


src/ui/rename-form-ui-handler.ts
```diff
@@ -36,17 +36,13 @@ export default class RenameFormUiHandler extends FormModalUiHandler {
   show(args: any[]): boolean {
     if (super.show(args)) {
       const config = args[0] as ModalConfig;
-      this.inputs[0].text = (args[1] as PlayerPokemon).getNameToRender();
-
+      if (args[1] && typeof (args[1] as any).getNameToRender === "function") {
```
torranx, line 39: why is this casted as `any`? shouldn't it be `PlayerPokemon` too?

src/ui/starter-select-ui-handler.ts
```diff
@@ -2573,7 +2601,13 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
 
     if (species && (this.speciesStarterDexEntry?.seenAttr || this.speciesStarterDexEntry?.caughtAttr)) {
       this.pokemonNumberText.setText(Utils.padInt(species.speciesId, 4));
-      this.pokemonNameText.setText(species.name);
+      // if starterAttributes nickname is set, use it, otherwise use species name
```
torranx, line 2604: this can be removed imo, the code is pretty straightforward and readable

KimJeongSun: OK

damocleas:
Does this stay at the start of a new session? If I nickname a Pikachu "Bolt" and I leave the menu does it stick?

KimJeongSun:
> Does this stay at the start of a new session? If I nickname a Pikachu "Bolt" and I leave the menu does it stick?

Sure, it gets saved along with the other starter settings. Even if you refresh or return to the main menu, the nickname will be retained!

PR #3458, opened 2024-08-09
[Bug][Test] Fix minor bug of ability Flash Fire, added test file

allen925:
## What are the changes?
baton pass no longer pass it. 
ability still activated if user is frozen. 
(and a test file for it. )

according to : https://bulbapedia.bulbagarden.net/wiki/Flash_Fire_(Ability)
image
image

## What did change?
As stated above, I mean changes in code are straight forward. 
Delete some code, move some code. 

### Screenshots/Videos

https://github.com/user-attachments/assets/3cc7b083-3c0a-41f9-9476-1536285fb74c

## How to test the changes?
test locally. 

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -4408,7 +4408,7 @@ export function initAbilities() {
       .attr(StatusEffectImmunityAbAttr, StatusEffect.POISON, StatusEffect.TOXIC)
       .ignorable(),
     new Ability(Abilities.FLASH_FIRE, 3)
-      .attr(TypeImmunityAddBattlerTagAbAttr, Type.FIRE, BattlerTagType.FIRE_BOOST, 1, (pokemon: Pokemon) => !pokemon.status || pokemon.status.effect !== StatusEffect.FREEZE)
+      .attr(TypeImmunityAddBattlerTagAbAttr, Type.FIRE, BattlerTagType.FIRE_BOOST, 1)
       .ignorable(),
     new Ability(Abilities.SHIELD_DUST, 3)
       .attr(IgnoreMoveEffectsAbAttr)
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -2436,10 +2436,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     for (const tag of source.summonData.tags) {
 
       // bypass those can not be passed via Baton Pass
-      const excludeSourceMoves = new Set([Moves.YAWN]);
-      const excludeTagTypes = new Set([BattlerTagType.INFATUATED, BattlerTagType.FIRE_BOOST]);
+      const excludeTagTypes = new Set([BattlerTagType.DROWSY, BattlerTagType.INFATUATED, BattlerTagType.FIRE_BOOST]);
 
-      if (excludeSourceMoves.has(tag.sourceMove) || excludeTagTypes.has(tag.tagType)) {
+      if (excludeTagTypes.has(tag.tagType)) {
         continue;
       }
 
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -2435,8 +2435,11 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     }
     for (const tag of source.summonData.tags) {
 
-      // bypass yawn, and infatuation as those can not be passed via Baton Pass
-      if (tag.sourceMove === Moves.YAWN || tag.tagType === BattlerTagType.INFATUATED) {
+      // bypass those can not be passed via Baton Pass
+      const excludeSourceMoves = new Set([Moves.YAWN]);
+      const excludeTagTypes = new Set([BattlerTagType.INFATUATED, BattlerTagType.FIRE_BOOST]);
+
+      if (excludeSourceMoves.has(tag.sourceMove) || excludeTagTypes.has(tag.tagType)) {
         continue;
       }
 
--- a/src/test/abilities/flash_fire.test.ts
+++ b/src/test/abilities/flash_fire.test.ts
@@ -5,12 +5,11 @@ import { Moves } from "#enums/moves";
 import Phaser from "phaser";
 import { afterEach, beforeAll, beforeEach, describe, expect, it } from "vitest";
 import { SPLASH_ONLY } from "#test/utils/testUtils";
-import { BattleStat } from "#app/data/battle-stat";
-import { Stat } from "#app/data/pokemon-stat";
-import { TurnEndPhase } from "#app/phases";
+import { MovePhase, TurnEndPhase } from "#app/phases";
 import { getMovePosition } from "#test/utils/gameManagerUtils";
 import { Status, StatusEffect } from "#app/data/status-effect.js";
 import { BattlerTagType } from "#app/enums/battler-tag-type.js";
+import { BattlerIndex } from "#app/battle.js";
 
 describe("Abilities - Flash Fire", () => {
   let phaserGame: Phaser.Game;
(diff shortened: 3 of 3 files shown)


src/field/pokemon.ts
```diff
@@ -2429,8 +2429,11 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     }
     for (const tag of source.summonData.tags) {
 
-      // bypass yawn, and infatuation as those can not be passed via Baton Pass
-      if (tag.sourceMove === Moves.YAWN || tag.tagType === BattlerTagType.INFATUATED) {
+      // bypass those can not be passed via Baton Pass
+      const excludeSourceMoves = new Set([Moves.YAWN]);
```
Tempo-anon, line 2433: Does excludeSourceMoves really need to check for Yawn specifically? Since Yawn is the only thing in the game to give Drowsy why not just exclude passing on the Drowsy battlertag?

src/field/pokemon.ts
```diff
@@ -2429,8 +2429,11 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      const excludeTagTypes = new Set([BattlerTagType.INFATUATED, BattlerTagType.FIRE_BOOST]);
```
Tempo-anon, line 2434: Could we maybe update the BattlerTagTypes so that they have a new variable for whether or not they can be baton passed?

src/test/abilities/flash_fire.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      .enemyLevel(20);
+  });
+
+  it("status move (will-o-wisp)", async() => {
+    game.override.enemyMoveset(Array(4).fill(Moves.WILL_O_WISP)).moveset(SPLASH_ONLY);
+    await game.startBattle([Species.BLISSEY]);
+
+    const blissey = game.scene.getPlayerPokemon()!;
+    expect(blissey).toBeDefined();
+    blissey.summonData.battleStats[BattleStat.EVA] = -6;
```
torranx, line 44: is this a 100% guarantee that the move will hit? I would suggest to spyOn hitCheck instead

src/test/abilities/flash_fire.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      .startingLevel(20)
+      .enemyLevel(20);
+  });
+
+  it("status move (will-o-wisp)", async() => {
+    game.override.enemyMoveset(Array(4).fill(Moves.WILL_O_WISP)).moveset(SPLASH_ONLY);
+    await game.startBattle([Species.BLISSEY]);
+
+    const blissey = game.scene.getPlayerPokemon()!;
+    expect(blissey).toBeDefined();
```
torranx, line 43: you can remove this expects, if we wanted to test this, it should be inside the battle test

src/test/abilities/flash_fire.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    blissey.summonData.battleStats[BattleStat.EVA] = -6;
+
+    game.doAttack(getMovePosition(game.scene, 0, Moves.SPLASH));
+
+    await game.phaseInterceptor.to(TurnEndPhase);
+    expect(blissey!.getTag(BattlerTagType.FIRE_BOOST)).toBeDefined();
```
torranx, line 49: please add expects for the expected ability's behavior/effect

src/test/abilities/flash_fire.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override
+      .battleType("single")
+      .ability(Abilities.FLASH_FIRE)
+      .startingLevel(20)
+      .enemyLevel(20);
+  });
+
+  it("status move (will-o-wisp)", async() => {
```
torranx, line 38: please make the titles better by describing what it tests ie `could be activated by Will-O-Wisp `

src/test/abilities/flash_fire.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  }, 20000);
+
+  it("after frozen", async() => {
+    game.override.enemyMoveset(Array(4).fill(Moves.EMBER)).moveset(SPLASH_ONLY);
+    await game.startBattle([Species.BLISSEY]);
+
+    const blissey = game.scene.getPlayerPokemon()!;
+    expect(blissey).toBeDefined();
+
+    blissey!.status = new Status(StatusEffect.FREEZE);
```
torranx, line 59: can we add an expect for blissey's freeze status?

src/test/abilities/flash_fire.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    game.doAttack(getMovePosition(game.scene, 0, Moves.SPLASH));
+
+    await game.phaseInterceptor.to(TurnEndPhase);
+    expect(blissey!.getTag(BattlerTagType.FIRE_BOOST)).toBeDefined();
+  }, 20000);
+
+  it("not passing with baton pass", async() => {
+    game.override.enemyMoveset(Array(4).fill(Moves.EMBER)).moveset([Moves.BATON_PASS, Moves.SPLASH, Moves.SPLASH, Moves.SPLASH]);
```
torranx, line 67: you dont need to fill the moveset with exactly 4 moves

src/test/abilities/flash_fire.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    await game.startBattle([Species.BLISSEY, Species.CHANSEY]);
+
+    const blissey = game.scene.getPlayerPokemon()!;
+    expect(blissey).toBeDefined();
+    // ensure use baton pass after enemy moved
+    blissey.stats[Stat.SPD] = 1;
```
torranx, line 73: updating `stats` is not ideal, you can spyOn `getOrder` instead

src/test/abilities/flash_fire.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    game.scene.currentBattle.enemyParty[0].stats[Stat.SPD] = 150;
```
torranx, line 74: updating stats is not ideal, you can spyOn getOrder instead

src/test/abilities/flash_fire.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override
+      .battleType("single")
+      .ability(Abilities.FLASH_FIRE)
+      .startingLevel(20)
+      .enemyLevel(20);
+  });
+
```
torranx, line 37: please add tests for the ff:
- `Flash Fire makes the Pokémon immune to Fire-type moves`
- `Flash Fire will not activate if the Pokémon is protected from the Fire-type move`
- `Flash Fire now technically boosts the Pokémon's Attack or Special Attack by 50% during damage calculation if a Fire-type move is being used`

reference: [Flash Fire](https://bulbapedia.bulbagarden.net/wiki/Flash_Fire_(Ability))

src/test/abilities/flash_fire.test.ts
DayKev: I believe current beta should have `game.setTurnOrder()`.

torranx: true that, just recently merged

src/field/pokemon.ts
allen925: updated

src/field/pokemon.ts
allen925: well i guess i can do something like add one extra field in BattlerTag and default to false, but I think it shouldn't be me or at least in this PR to do that. (as i am not that familiar with tags. )

src/test/abilities/flash_fire.test.ts
allen925: done. have to do two forceHit to apply one to enemy's move.

src/test/abilities/flash_fire.test.ts
allen925: updated.

src/test/abilities/flash_fire.test.ts
allen925: updated.

src/test/abilities/flash_fire.test.ts
allen925: updated.

src/test/abilities/flash_fire.test.ts
allen925: updated.

src/test/abilities/flash_fire.test.ts
allen925: updated.

src/test/abilities/flash_fire.test.ts
allen925: updated. note the last test of testing dmg, I'm referencing the dry skin test one that just test if the dmg is greater than original.

src/test/abilities/flash_fire.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    game.doSelectPartyPokemon(1);
+
+    await game.phaseInterceptor.to(TurnEndPhase);
+    const chansey = game.scene.getPlayerPokemon()!;
+    expect(game.scene.getPlayerPokemon()!.species.speciesId).toBe(Species.CHANSEY);
+    expect(chansey!.getTag(BattlerTagType.FIRE_BOOST)).toBeUndefined();
+  }, 20000);
+
+  it("damage test", async() => {
```
torranx, line 107: nit:
```suggestion
  it("boosts Fire-type move when the ability is activated", async() => {
```

src/data/ability.ts
```diff
@@ -4408,7 +4408,7 @@ export function initAbilities() {
       .attr(StatusEffectImmunityAbAttr, StatusEffect.POISON, StatusEffect.TOXIC)
       .ignorable(),
     new Ability(Abilities.FLASH_FIRE, 3)
-      .attr(TypeImmunityAddBattlerTagAbAttr, Type.FIRE, BattlerTagType.FIRE_BOOST, 1, (pokemon: Pokemon) => !pokemon.status || pokemon.status.effect !== StatusEffect.FREEZE)
```
torranx, line 4411: what was the purpose of `!pokemon.status` and why is it removed?

torranx:
possible overlap on implementation https://github.com/pagefaultgames/pokerogue/pull/3472

src/data/ability.ts
allen925: well i beliebe it is equivalent to "pokemon.status && pokemon.status.effect !== StatusEffect.FREEZE" but if you really want to know the reason why it is being written like that, time to ask flashFyre.

allen925:
> possible overlap on implementation #3472

That is intended, i do think the change of baton pass should not be in this pr under temp's discussion, not to say that the pr is just giving a framework right now. (which means one more pr is needed in future to modify this portion of code using the framework to replace what we currently have in the transferSummon() )

torranx (approved):

src/data/ability.ts
innerthunder: Flash Fire used to not work if the source was frozen, but that hasn't been the case since gen 5.

innerthunder (approved):
lgtm. The Baton Pass bit is a fine temporary fix until Baton Pass is fixed with `isTransferrable` from #3472

DayKev (approved):
lgtm

Issue #3453, opened 2024-08-09
[Bug] Endless `%50` waves soft-lock


flx-sta:
## Describe the bug

When fighting a `%50` wave e.g. `4950` there is a soft-lock with a console error.
PR #3346 caused it by calling `pokemon.destroy()` in `BattleEndPhase`

## Session export file

[⬇️ sessionData3_Guest.txt](https://github.com/user-attachments/files/16555476/sessionData3_Guest.txt)
<sub>Make sure to change to `prsv` file-extension</sub>

## Expected behavior

It shouldn't soft-lock and continue to next wave

## Screenshots / Videos

![discord image from starkrieg](https://media.discordapp.net/attachments/1244839350919888897/1271283091922751539/image.png?ex=66b6c624&is=66b574a4&hm=7781b3664990020ba34ae751e1940243620bf370b94175b547101e4e26fd6d23&=&format=webp&quality=lossless&width=1100&height=302)

## Additional context

- First reported [discord message from Starkrieg](https://discord.com/channels/1125469663833370665/1245857678991822879/1271283224831594507) (might be not visible/locked to some)

DustinLin:
PR #3346 made that call because it assumes that after `BattleEndPhase` the enemy pokemon aren't referenced anymore. Seems that in endless (and maybe other situations) this may not be the case?

torranx:
> PR #3346 made that call because it assumes that after `BattleEndPhase` the enemy pokemon aren't referenced anymore. Seems that in endless (and maybe other situations) this may not be the case?

Why are we even calling something based on assumption?

DustinLin:
I'm not sure if the original code was calling `.destroy()` for optimization reasons, if so then I think it's reasonable to make assumptions about when objects won't be needed anymore. Pre- pr #3346 the assumption was that pokemon won't be needed right after the `MoveEffectPhase` (which turned out to be wrong).

DustinLin:
From flx's comment in #3346 , the phases towards the end of battle go in this order:
```
Start Phase BerryPhase
Start Phase TurnEndPhase
Start Phase BattleEndPhase | THIS calls p.destroy() on all enemies
...
Start Phase EggLapsePhase
Start Phase ModifierRewardPhase
Start Phase AddEnemyBuffModifierPhase
```

[...]

PR #3461, opened 2024-08-09
[Bug][Hotfix] - Calling destroy() on enemy pokemon after BattleEndPhase breaks endless

DustinLin:
## What are the changes?

Removing the `.destroy()` calls, more testing needed to make sure no adverse affects

Addressing #3453 which was originally caused by #3346 
### Screenshots/Videos

After fix, doubles dragon tail fix still seems ok

https://github.com/user-attachments/assets/4b298cc5-8329-48cc-bb8c-ecdf806f1745

https://github.com/user-attachments/assets/b64f0a27-29d3-47f6-9cd1-424ddb35fdf3

https://github.com/user-attachments/assets/f49cb55c-9068-477f-9961-9cbb5c475337

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -2576,15 +2576,6 @@ export class BattleEndPhase extends BattlePhase {
 
     this.scene.updateModifiers().then(() => this.end());
   }
-
-  end() {
-    // removing pokemon at the end of a battle
-    for (const p of this.scene.getEnemyParty()) {
-      p.destroy();
-    }
-
-    super.end();
-  }
 }
 
 export class NewBattlePhase extends BattlePhase {


flx-sta (approved):
Well I basically tested that as I reported the bug and tried the same bugfix.
lgtm 🚀

DustinLin:
looking around the codebase, it seems like the only other time `destroy()` is being called for wild pokemon is `attemptRunPhase` and for the `GameOverPhase` + `PostGameOverPhase`. Notably something else happens for fainted pokemon, or going between levels.

I guess this begs the question for why `destroy()` was originally called pre-PR #3346 but if this PR works right now we can think more about this. There may be optimization concerns but it seems that fainted-pokemon wave-to-wave aren't being `destroy()`ed so maybe it's fine (look at `removeFaintedParticipant()`

PR #3466, opened 2024-08-09
[Bug] Prevent Relearnable Egg Moves in Fresh Start

xsn34kzx:
## What are the changes?

Fixed the unintended ability to allow starters to remember unlocked egg moves in the `Fresh Start` challenge with the `Memory Mushroom`.

By definition, the `Fresh Start` challenge is supposed remove all advantages gained from previous runs (abilities, egg moves, etc.). As such, having the ability to remember unlocked egg moves, even as a starter, goes against the intended restriction of not having access to egg moves to begin with. 

## What did change?

An additional check was added to the `if` condition in `getLearnableLevelMoves()` to check if `Fresh Start` is active within the `challenge` list.

### Screenshots/Videos

| Before | After |
| --- | --- |
|  mushroomBefore | mushroomAfter |

## How to test the changes?

With the `Challenge` game mode unlocked, use the following override and complete one wave:
```ts
ITEM_REWARD_OVERRIDE: [ "MEMORY_MUSHROOM" ]
```

Only traditionally learnable moves should be accessible when using the `Memory Mushroom` (i.e., no egg moves should be accessible).

## Checklist
- [x] Are the changes visual?


ben-lear:
Could we add 1-2 unit tests to get some coverage for this function? If you need assistance with navigating the Party menu in tests I can help out a bit

torranx (approved):

src/field/pokemon.ts
```diff
@@ -922,7 +923,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
    */
   getLearnableLevelMoves(): Moves[] {
     let levelMoves = this.getLevelMoves(1, true).map(lm => lm[1]);
-    if (this.metBiome === -1) {
+    if (this.metBiome === -1 && !this.scene.gameMode.challenges.some(c => c.id === Challenges.FRESH_START)) {
```
frutescens, line 926: It would be nice to have a comment above this conditional, but it isn't necessary.

frutescens (commented):
LMK your thoughts on adding a comment there, and I'll approve.

src/field/pokemon.ts
xsn34kzx: I was actually thinking about doing this before I got caught up doing stuff, so I'll probably do this and then bump 🫡

frutescens (approved):

src/field/pokemon.ts
Tempo-anon: Can use `isFreshStartChallenge()` that I define in https://github.com/pagefaultgames/pokerogue/pull/3420/files ?

PR #3469, opened 2024-08-09
[MOVE] Some more implementation for Tera Blast

cgyu7:
Tera blast now changes the type of damage dealt as the same type of the tera type applied to the mon. Tera blast also accounts for if the user is terastalized into Stellar type with increased base damage and dropping ATK/SPATK by one stage.

Tera blast is a commonly offered TM and loses a lot of utility due to the fact it's currently only possible to do Normal type damage. Also gives more incentive to pick up the tera orb, tera shards, and teach tera blast more often.

Added a TeraBlastTypeAttr class to check if the Pokemon is terastalized and if so, change tera blast's dmg to that of the tera type.
Added a TeraBlastPowerAttr class to check for if the user is tera'd into the Stellar type, and if so make the base power of tera blast 200 instead of 100 to account for the missing 2x dmg by Stellar type moves on all types.

Added to the move a ATK/SPATK drop by one stage after if the user is tera'd into Stellar type. 

Kept the move as .partial() due to the lack of Stellar offense typing implementation in the typings

### Screenshots/Videos
terablaststellar

## How to test the changes?
Use tera blast while terastalized

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -8749,7 +8749,7 @@ export function initMoves() {
       .attr(TeraBlastCategoryAttr)
       .attr(TeraBlastTypeAttr)
       .attr(TeraBlastPowerAttr)
-      .attr(StatChangeAttr, [ BattleStat.ATK, BattleStat.SPATK, ], -1, true, (user, target, move) => user.isTerastallized() && user.isOfType(Type.STELLAR))
+      .attr(StatChangeAttr, [ BattleStat.ATK, BattleStat.SPATK ], -1, true, (user, target, move) => user.isTerastallized() && user.isOfType(Type.STELLAR))
       .partial(),
     new SelfStatusMove(Moves.SILK_TRAP, Type.BUG, -1, 10, -1, 4, 9)
       .attr(ProtectAttr, BattlerTagType.SILK_TRAP),
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -3760,6 +3760,7 @@ export class TeraBlastCategoryAttr extends VariableMoveCategoryAttr {
 }
 
 /**
+ * Increases the power of Tera Blast if the user is Terastallized into Stellar type
  * @extends VariablePowerAttr
  */
 export class TeraBlastPowerAttr extends VariablePowerAttr {
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -8735,7 +8735,6 @@ export function initMoves() {
     new AttackMove(Moves.TERA_BLAST, Type.NORMAL, MoveCategory.SPECIAL, 80, 100, 10, -1, 0, 9)
       .attr(TeraBlastCategoryAttr)
       .attr(StatChangeAttr, [ BattleStat.ATK, BattleStat.SPATK, ], -1, true, (user, target, move) => user.isTerastallized() && user.isOfType(Type.STELLAR))
-      .makesContact(false)
       .partial(),
     new SelfStatusMove(Moves.SILK_TRAP, Type.BUG, -1, 10, -1, 4, 9)
       .attr(ProtectAttr, BattlerTagType.SILK_TRAP),
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -3748,16 +3748,15 @@ export class PhotonGeyserCategoryAttr extends VariableMoveCategoryAttr {
 
 export class TeraBlastCategoryAttr extends VariableMoveCategoryAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    const category = (args[0] as Utils.IntegerHolder);
-	  
+    const category = (args[0] as Utils.IntegerHolder);  
     move.power = 80;
 
     if (user.isTerastallized()) {
-      move.type = user.getTeraType(); 
+      move.type = user.getTeraType();
       //changes type to tera type
       if (move.type === Type.STELLAR) {
         if (move.id === Moves.TERA_BLAST) {
-          move.power = 200; 
+          move.power = 200;
           //200 instead of 100 to reflect lack of stellar being 2x dmg on any typew
         }
       }
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -3762,11 +3762,10 @@ export class TeraBlastCategoryAttr extends VariableMoveCategoryAttr {
 export class TeraBlastPowerAttr extends VariablePowerAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
     const power = args[0] as Utils.NumberHolder;
-    
     if (user.isTerastallized() && move.type === Type.STELLAR) {
       power.value = 200;
       //200 instead of 100 to reflect lack of stellar being 2x dmg on any type
-      return true
+      return true;
     }
 
     return false;
@@ -4029,8 +4028,7 @@ export class HiddenPowerTypeAttr extends VariableMoveTypeAttr {
 export class TeraBlastTypeAttr extends VariableMoveTypeAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
     if (user.isTerastallized()) {
-      move.type = user.getTeraType(); 
-      //changes move type to tera type
+      move.type = user.getTeraType(); //changes move type to tera type
       return true;
     }
 
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -4037,6 +4037,7 @@ export class HiddenPowerTypeAttr extends VariableMoveTypeAttr {
 }
 
 /**
+ * Changes the type of Tera Blast to match the user's tera type
  * @extends VariableMoveTypeAttr
  */
 export class TeraBlastTypeAttr extends VariableMoveTypeAttr {
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -3774,8 +3774,8 @@ export class TeraBlastPowerAttr extends VariablePowerAttr {
    */
     const power = args[0] as Utils.NumberHolder;
     if (user.isTerastallized() && move.type === Type.STELLAR) {
-      power.value = 200;
       //200 instead of 100 to reflect lack of stellar being 2x dmg on any type
+      power.value = 200;
       return true;
     }
 
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -3763,14 +3763,10 @@ export class TeraBlastPowerAttr extends VariablePowerAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
     const power = args[0] as Utils.NumberHolder;
     
-    if (user.isTerastallized()) {
-      move.type = user.getTeraType(); 
-      //changes move type to tera type
-      if (move.type === Type.STELLAR) {
-        power.value = 200;
-        //200 instead of 100 to reflect lack of stellar being 2x dmg on any type
-        return true
-      }
+    if (user.isTerastallized() && move.type === Type.STELLAR) {
+      power.value = 200;
+      //200 instead of 100 to reflect lack of stellar being 2x dmg on any type
+      return true
     }
 
     return false;
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -3749,22 +3749,28 @@ export class PhotonGeyserCategoryAttr extends VariableMoveCategoryAttr {
 export class TeraBlastCategoryAttr extends VariableMoveCategoryAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
     const category = (args[0] as Utils.IntegerHolder);  
-    move.power = 80;
-
-    if (user.isTerastallized()) {
-      move.type = user.getTeraType();
-      //changes type to tera type
-      if (move.type === Type.STELLAR) {
-        if (move.id === Moves.TERA_BLAST) {
-          move.power = 200;
-          //200 instead of 100 to reflect lack of stellar being 2x dmg on any type
-        }
-      }
-    }
 
     if (user.isTerastallized() && user.getBattleStat(Stat.ATK, target, move) > user.getBattleStat(Stat.SPATK, target, move)) {
       category.value = MoveCategory.PHYSICAL;
       return true;
+    }
+
+    return false;
+  }
+}
+
+export class TeraBlastPowerAttr extends VariablePowerAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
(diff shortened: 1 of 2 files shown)


src/data/move.ts
```diff
@@ -3748,7 +3748,19 @@ export class PhotonGeyserCategoryAttr extends VariableMoveCategoryAttr {
 
 export class TeraBlastCategoryAttr extends VariableMoveCategoryAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    const category = (args[0] as Utils.IntegerHolder);
+    const category = (args[0] as Utils.IntegerHolder);  
+    move.power = 80;
```
torranx, line 3752: I think doing this will globally update the move's power. If you check `calculateBattlePower()` and other power manipulations, they are using Utils.NumberHolder

cgyu7: I considered using 
>const power = args[0] as Utils.NumberHolder
for calculating whether the move power will be 80 or 100, but line 3751 is already using Utils.NumberHolder and I thought and some others added that it might cause an issue to use it again in the same class. I could also just make a separate class for the damage calc if it'll cause issues with the phy/special dmg calc.

Another idea is treating the calculation similarly to how magnitude is calculated by having [80, 200] and have it choose one or the other based on the mon's tera type.

cgyu7:
I've split the damage and move type calculations into two new classes, so that they're better organized and each catered to doing one specific calc instead of overloading one class with multiple

src/data/move.ts
```diff
@@ -3748,7 +3748,7 @@ export class PhotonGeyserCategoryAttr extends VariableMoveCategoryAttr {
 
 export class TeraBlastCategoryAttr extends VariableMoveCategoryAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    const category = (args[0] as Utils.IntegerHolder);
+    const category = (args[0] as Utils.IntegerHolder);
```
DayKev, line 3751:
```suggestion
    const category = (args[0] as Utils.IntegerHolder);
```

src/data/move.ts
```diff
@@ -8721,7 +8747,10 @@ export function initMoves() {
     End Unused */
     new AttackMove(Moves.TERA_BLAST, Type.NORMAL, MoveCategory.SPECIAL, 80, 100, 10, -1, 0, 9)
       .attr(TeraBlastCategoryAttr)
-      .unimplemented(),
+      .attr(TeraBlastTypeAttr)
+      .attr(TeraBlastPowerAttr)
+      .attr(StatChangeAttr, [ BattleStat.ATK, BattleStat.SPATK, ], -1, true, (user, target, move) => user.isTerastallized() && user.isOfType(Type.STELLAR))
```
DayKev, line 8752:
```suggestion
      .attr(StatChangeAttr, [ BattleStat.ATK, BattleStat.SPATK ], -1, true, (user, target, move) => user.isTerastallized() && user.isOfType(Type.STELLAR))
```

DayKev (changes requested):
Seems good, just needs tests and tsdocs.

cgyu7:
> Seems good, just needs tests and tsdocs.

What tests do I need to do? And what do I do for the tsdocs?

torranx:
> What tests do I need to do? And what do I do for the tsdocs?

automated tests for Tera Blast's supposed behavior, refer to existing tests in the codebase if you don't know where to start. tsdocs/jsdocs -> document your classes

DayKev:
> > Seems good, just needs tests and tsdocs.
> 
> What tests do I need to do? And what do I do for the tsdocs?

As an example for docs, I'll use Destiny Bond's move attr:
```ts
/**
 * Applies {@linkcode BattlerTagType.DESTINY_BOND} to the user.
 *
 * @extends MoveEffectAttr
 */
export class DestinyBondAttr extends MoveEffectAttr {
  constructor() {
    super(true, MoveEffectTrigger.PRE_APPLY);
  }

  /**
   * Applies {@linkcode BattlerTagType.DESTINY_BOND} to the user.
   * @param user {@linkcode Pokemon} that is having the tag applied to.
   * @param target {@linkcode Pokemon} N/A
   * @param move {@linkcode Move} {@linkcode Move.DESTINY_BOND}
   * @param {any[]} args N/A
   * @returns true
   */
    user.scene.queueMessage(`${i18next.t("moveTriggers:tryingToTakeFoeDown", {pokemonName: getPokemonNameWithAffix(user)})}`);
    user.addTag(BattlerTagType.DESTINY_BOND, undefined, move.id, user.id);
    return true;
  }
}
```
The text between the `/** ... */` is tsdocs. If you use VSCode (or the opensource version) it should have functionality to format these automatically for you (just start typing `/**` above the object you want to add comments for and hit enter). You put a description of what the thing you're commenting on does, and then below that you put more specific descriptions for things like function parameters. Don't worry if you don't get the formatting perfect, we can suggest fixes for that afterward.

src/data/move.ts
```diff
@@ -3759,6 +3759,29 @@ export class TeraBlastCategoryAttr extends VariableMoveCategoryAttr {
   }
 }
 
+/**
+ * @extends VariablePowerAttr
```
DayKev, line 3763:
```suggestion
/**
 * Increases the power of Tera Blast if the user is Terastallized into Stellar type
 * @extends VariablePowerAttr
```

src/data/move.ts
```diff
@@ -4012,6 +4035,27 @@ export class HiddenPowerTypeAttr extends VariableMoveTypeAttr {
   }
 }
 
+/**
+ * @extends VariableMoveTypeAttr
```
DayKev, line 4039:
```suggestion
/**
 * Changes the type of Tera Blast to match the user's tera type
 * @extends VariableMoveTypeAttr
```

src/data/move.ts
```diff
@@ -3759,6 +3759,29 @@ export class TeraBlastCategoryAttr extends VariableMoveCategoryAttr {
...
+   * @param user {@linkcode Pokemon} Pokemon using the move
+   * @param target {@linkcode Pokemon} N/A
+   * @param move {@linkcode Move} {@linkcode Move.TERA_BLAST}
+   * @param {any[]} args N/A
+   * @returns true or false
+   */
+    const power = args[0] as Utils.NumberHolder;
+    if (user.isTerastallized() && move.type === Type.STELLAR) {
+      power.value = 200;
+      //200 instead of 100 to reflect lack of stellar being 2x dmg on any type
```
DayKev, line 3777:
```suggestion
      //200 instead of 100 to reflect lack of stellar being 2x dmg on any type
      power.value = 200;
```

DayKev (changes requested):
Now you need tests to confirm the move works as expected. For Tera Blast, there should probably be:
* "changes type to match user's tera type"
* "increases power if user is Stellar tera type"
* "uses the higher stat of the user's Atk and SpAtk for damage calculation"
* "causes stat drops if user is Stellar tera type"

Here's a template you can copy to start making the test with. Look at the other tests for examples when you don't know what to do (moves/abilities that do similar things to what you're testing for are good places to start, ie: the Battery ability modifies the power of a move).
```ts
import { allMoves } from "#app/data/move";
import GameManager from "#test/utils/gameManager";
import { Species } from "#enums/species";
import Phaser from "phaser";
import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
import { Abilities } from "#app/enums/abilities.js";
import { SPLASH_ONLY } from "../utils/testUtils";
import { Type } from "#app/data/type.js";

describe("Moves - Tera Blast", () => {
  let phaserGame: Phaser.Game;
  let game: GameManager;
  const moveToCheck = allMoves[Moves.TERA_BLAST];

  beforeAll(() => {
    phaserGame = new Phaser.Game({
      type: Phaser.HEADLESS,
    });
  });

  afterEach(() => {
    game.phaseInterceptor.restoreOg();
  });

  beforeEach(() => {

    game.override
      .battleType("single")
      .disableCrits()
      .starterSpecies(Species.FEEBAS)
      .moveset([Moves.TERA_BLAST])
      .ability(Abilities.BALL_FETCH)
      .startingHeldItems([{name: "TERA_SHARD", type: Type.FIRE}])
      .enemySpecies(Species.MAGIKARP)
      .enemyMoveset(SPLASH_ONLY)
      .enemyAbility(Abilities.BALL_FETCH)
      .enemyLevel(20);

    vi.spyOn(moveToCheck, "calculateBattlePower");
  });

  it("changes type to match user's tera type", async() => {
    await game.startBattle();

    // test goes here
    expect("test").toBe("test");
  }, 20000);

  it("increases power if user is Stellar tera type", async() => {
    game.override.startingHeldItems([{name: "TERA_SHARD", type: Type.STELLAR}]);
    await game.startBattle();

    // test goes here
  }, 20000);

  it("uses the higher stat of the user's Atk and SpAtk for damage calculation", async() => {
    await game.startBattle();

    // test goes here
  }, 20000);

  it("causes stat drops if user is Stellar tera type", async() => {
    game.override.startingHeldItems([{name: "TERA_SHARD", type: Type.STELLAR}]);
    await game.startBattle();

    // test goes here
  }, 20000);
});

```

src/data/move.ts
cgyu7: tsdoc update

DayKev (changes requested):
Well you got them started, but the tests weren't finished. Amusingly in the course of finishing the tests I discovered a bug.

src/test/moves/tera_blast.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    game.override.startingHeldItems([{name: "TERA_SHARD", type: Type.STELLAR}]);
+    const stellarTypeMultiplier = 2;
+    const stellarTypeDmgBonus = 20;
+    const basePower = moveToCheck.power;
+    
+    await game.startBattle([Species.CHIKORITA]);
+
+    game.doAttack(getMovePosition(game.scene, 0, Moves.TERA_BLAST));
+
+    expect(moveToCheck.calculateBattlePower).toBe((basePower + stellarTypeDmgBonus) * stellarTypeMultiplier);
```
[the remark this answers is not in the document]
cgyu7: I see. I was under the assumption that `doAttack()` would result in the attack and its effects following through after starting the battle. So the function itself only applies the attack when the game phase is in a place where the attack can happen? (in this case I assume `setTurnOrder()` allows for the attack to happen and `phaseInterceptor` to specifically trigger the move effect)

src/test/moves/tera_blast.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    vi.spyOn(moveToCheck, "calculateBattlePower");
+  });
+
+  it("changes type to match user's tera type", async() => {
+
+    await game.startBattle([Species.CHIKORITA]);
+
+    game.doAttack(getMovePosition(game.scene, 0, Moves.TERA_BLAST));
+    
+    expect(moveToCheck.type).toBe(Type.FIRE);
```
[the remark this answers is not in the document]
cgyu7: Would this be the best way to test moves that can change types? Or is there another way to directly check the move's type after the move effect is activated that isn't just checking type effectiveness against an enemy?

src/test/moves/tera_blast.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    await game.startBattle([Species.CHIKORITA]);
+    
+    const playerPokemon = game.scene.getPlayerPokemon()!;
+    playerPokemon.stats[Stat.ATK] = 100;
+    playerPokemon.stats[Stat.SPATK] = 0;
+
+    game.doAttack(getMovePosition(game.scene, 0, Moves.TERA_BLAST));
+
+    expect(moveToCheck.category).toBe(MoveCategory.PHYSICAL);
+
```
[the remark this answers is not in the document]
cgyu7: In theory for this case, the test for the move being physical is the fact that it should pass the check of toxic debris being procced after the hit? And currently the issue is that it doesn't proc due to the bug? In this case, is it like the test for the move type where there's not a direct test for the move's category after move effect proc?

cgyu7:
Thank you for all the help with the test. This is my first time contributing to PokeRogue so I'm still relatively unfamiliar with the code and test. Is there anything else I need to do still other than commit the changes to the test?

src/test/moves/tera_blast.test.ts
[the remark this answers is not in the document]
DayKev: I'm probably gonna see about renaming `doAttack()` to something like `prepareAttack()` to prevent this confusion in the future.

`setTurnOrder()` is used when we want to force specific Pokémon to move first/second, it's only needed sometimes (it's probably not necessary here but it allows me to shorten the time it takes for the test to complete because it doesn't matter if the opponent Pokémon attacks or not, so I can just make the player Pokémon move first, check the relevant conditions, and then end the test).

`phaseInterceptor.to()` tells the game to run like normal until it reaches the specific phase and then the rest of the test code continues to run at that point. Generally you'd use either "MoveEffectPhase" (this is where damage and effects happen for the current move being executed) or "TurnEndPhase" (which is, unsurprisingly, the end of the turn) depending on what needs to be checked.

src/test/moves/tera_blast.test.ts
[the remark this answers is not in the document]
DayKev: Unfortunately in this case the internals aren't really possible to `spyOn()`/etc due to the way the code is written, so it's necessary to instead check something else as a proxy (like if the move is super effective, which should theoretically only happen in this test setup if the type of Tera Blast properly changes to Fighting type, since that's the only type Normal is weak to and is the tera type of the player Pokémon in this test).

src/test/moves/tera_blast.test.ts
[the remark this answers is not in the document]
DayKev: Yeah, there's no way to directly look at the category of the move being used, so we have to again test something that *should* happen if the move is working properly. In this case something other than the move being tested is broken which prevents us from testing this attribute properly, so until that's fixed the test is skipped.

DayKev:
> Thank you for all the help with the test. This is my first time contributing to PokeRogue so I'm still relatively unfamiliar with the code and test. Is there anything else I need to do still other than commit the changes to the test?

With the test changes, everything should be done. After the next beta -> main merge happening in a couple days (if nothing breaks) then feature PRs like this one will start being evaluated/merged again, and it should be good to go (unless someone else notices a problem I missed).

DayKev (approved):

torranx (approved):

innerthunder (approved):

frutescens (approved):

innerthunder (approved):

podarsmarty:
image

hmm was playing and noticed an error regarding tera-blast.json not being available. had the offline client redownload files but it still isn't there. is this related to the recent changes?

PR #3472, opened 2024-08-10
[Misc][Bug] Add `isBatonPassable` property to `BattlerTag`s

DayKev:
## What are the changes?
Baton Pass is now more accurate in which properties it passes to other Pokémon.

Baton Pass currently passes `BattlerTag`s that it shouldn't to other Pokémon.

## What did change?
An `isTransferrable: boolean` property is added to `BattlerTag`s to be used by Baton Pass and the Baton item, and existing `BattlerTag`s that should be passed are marked accordingly.

## How to test the changes?
Have the opposing Pokémon apply the "infatuated" effect (or Salt Cure, or various other effects that shouldn't be passable) on your Pokémon, then use Baton Pass and verify that the effect was not passed.
`npm run test baton_pass`

## Checklist


innerthunder (approved):
lgtm, checked the values for each tag against https://bulbapedia.bulbagarden.net/wiki/Baton_Pass_(move)#Effect

Tempo-anon (approved):

src/data/battler-tags.ts
```diff
@@ -97,7 +99,7 @@ export interface TerrainBattlerTag {
  */
 export class RechargingTag extends BattlerTag {
   constructor(sourceMove: Moves) {
-    super(BattlerTagType.RECHARGING, [ BattlerTagLapseType.PRE_MOVE, BattlerTagLapseType.TURN_END ], 2, sourceMove);
+    super(BattlerTagType.RECHARGING, [ BattlerTagLapseType.PRE_MOVE, BattlerTagLapseType.TURN_END ], 2, sourceMove, undefined, false);
```
torranx, line 102: Do you need to set it to false if the default value is already `false`?

DayKev: I set them all before I'd decided what the default would be lol. I can go back and remove the `false` values if needed.

torranx: > I can go back and remove the `false` values if needed.

yes please

torranx (approved):

Tempo-anon:
Looks good. Can you update the flash fire test mentioned in https://github.com/pagefaultgames/pokerogue/pull/3458?

innerthunder (dismissed):

src/data/battler-tags.ts
```diff
@@ -787,7 +789,7 @@ export class AquaRingTag extends BattlerTag {
 /** Tag used to allow moves that interact with {@link Moves.MINIMIZE} to function */
 export class MinimizeTag extends BattlerTag {
   constructor() {
-    super(BattlerTagType.MINIMIZED, BattlerTagLapseType.TURN_END, 1, Moves.MINIMIZE, undefined);
+    super(BattlerTagType.MINIMIZED, BattlerTagLapseType.TURN_END, 1, Moves.MINIMIZE);
```
Xavion3, line 792: Bulbapedia contradicts itself on whether minimize can be passed with baton pass so this will need another source.

Xavion3 (dismissed):
In addition to the above, I'm not seeing nightmare or lock on being transferrable here.

src/data/battler-tags.ts
DayKev: No it doesn't? Both the https://bulbapedia.bulbagarden.net/wiki/Minimize_(move) page and https://bulbapedia.bulbagarden.net/wiki/Baton_Pass_(move)#Effect page say Minimize's extra effects aren't passed.

torranx (dismissed):

DayKev:
> In addition to the above, I'm not seeing nightmare or lock on being transferrable here.

Bulbapedia indicates Lock On is not able to be Baton Passed in Gen5+

As for Nightmare, you can't even Baton Pass if you're affected by Nightmare, at least not in the scenario where you try to Sleep Talk into Baton Pass while affected by Nightmare (which is the only scenario you can possibly use it in).

src/data/battler-tags.ts
Xavion3: It does on the status effects page https://bulbapedia.bulbagarden.net/wiki/Status_condition#Minimize where it explicitly calls out baton passing minimize as a way to take the boosted damage.

Xavion3:
Why can't you baton pass while asleep using sleep talk?

src/data/battler-tags.ts
DayKev: No double damage after baton pass:
image

Double damage without baton pass:
image

DayKev:
> Why can't you baton pass while asleep using sleep talk?

https://github.com/user-attachments/assets/9ee9a995-19e1-4457-a615-96a1c0bdfaeb

Xavion3:
If bulbapedia is this unreliable about what is transferrable should we be using it as a source at all as this did? Or should we be checking everything against smogon to decide what to do?

DayKev:
https://www.smogon.com/dex/sv/moves/baton-pass/

Are we supposed to check every single move manually on Smogon to see if it works with Baton Pass? That's not practical. No source is going to be perfect, and to expect perfection is unreasonable.

DayKev:
Bulbapedia also doesn't list it as transferrable actually, so I'm not sure why I marked it as such.

innerthunder (approved):

src/data/battler-tags.ts
```diff
@@ -39,13 +39,15 @@ export class BattlerTag {
   public turnCount: number;
   public sourceMove: Moves;
   public sourceId?: number;
+  public isTransferrable: boolean;
```
Tempo-anon, line 42: now that I think about it, can this variable be named like `isBatonPassable` to make it clear that this is just for baton pass and the baton item? isTransferrable makes me think that it's a boolean on whether or not the battler tag can be transferred at all

DayKev: Hm, I suppose so.

Tempo-anon (approved):

xsn34kzx (approved):

frutescens (approved):

PR #3478, opened 2024-08-10
[Enhancement] Add Passive / Cost Reduction - Can Unlock option and Egg purchasable filter

KimJeongSun:
## What are the changes?

- Added a "Can Unlock" status to the Passive, Cost Reduction filter.
  - This option allows users to easily filter Pokémon that are in an unlockable state.
- An Egg filter, which allows users to filter by purchasable eggs, has also been added to the Misc category.
- Additionally, the restriction that required passive to be unlocked in order to purchase eggs has been removed.

The current candy notification system has clear limitations, making it very difficult to locate specific Pokémon.
Plus, the part about removing restrictions for egg purchases has been confirmed by @damocleas , and since it is currently unnecessary, it has been removed.

## What did change?

- Declared the aforementioned filters and added conditions for them.
- During this process, introduced a new state, "UNLOCKABLE," which is now available in the radial filter.
- Additionally, the check condition in the egg purchase menu that required passive unlocking has been removed, allowing egg purchases without needing to unlock the passive.

### Screenshots/Videos

https://github.com/user-attachments/assets/c34a69c5-630a-4618-84c8-48237855df59

## Checklist
- [x] Are the changes visual?


MokaStitcher (changes requested):
All seems good for the unlockable filters.

For the egg filters, the logic is wrong as currently you can only buy eggs if you have unlocked the passive first, as shown here:

https://github.com/user-attachments/assets/d63560e4-c10d-49f6-8591-eda07bdea407

So if the passive hasn't been unlocked the pokemon should not show in this filter

returntoice (approved):

KimJeongSun:
@MokaStitcher 
As a result of our discussion, we've decided to remove the passive unlock restriction for egg purchases. 🎉 
With this update, the issue has been naturally resolved. 👍 
Thank you for reporting it! 🙇

damocleas (approved):
Approving so I don't get pinged for same-species egg restriction being removed :^)

MokaStitcher (approved):
all good now

PR #3483, opened 2024-08-11
[QoL] [Balance] Adjust ether weights for lower-PP moves

Fontbane:
## What are the changes?

- Adjusts the formula for ether/elixir weights in the item pool to only take a move into account if it has half its PP missing (rounded down). For example, having 4/5 PP on a single move will no longer put Ethers into the reward pool, but 3/5 PP will.

- Currently using a 5 PP move at all will automatically add ethers and elixirs to the reward pool which can take the place of other rewards
- Other healing items are usually to remedy something the player did wrong. Using a move with low PP isn't a bad decision.
- Currently players are discouraged from using low PP moves due to the resulting dilution of the reward pool
- No one is using an Elixir to restore 1 PP on Morning Sun be so fr

## What did change?

- Checks in the weight function for ethers/elixirs if the ppUsed of a move is greater than or equal to the move's max PP (including PP Ups if they've been applied) rounded down

## How to test the changes?

- Use a 5-8 PP move
- Note that Ethers will not show up until you've reached 3/5, 3/6, 4/7, or 4/8 PP on the move
- 10 PP moves will still produce an Ether at 5 PP as expected

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -1451,11 +1451,11 @@ const modifierPool: ModifierPool = {
       return thresholdPartyMemberCount;
     }, 3),
     new WeightedModifierType(modifierTypes.ETHER, (party: Pokemon[]) => {
-      const thresholdPartyMemberCount = Math.min(party.filter(p => p.hp && p.getMoveset().filter(m => m?.ppUsed && (m.getMovePp() - m.ppUsed) <= 5).length).length, 3);
+      const thresholdPartyMemberCount = Math.min(party.filter(p => p.hp && p.getMoveset().filter(m => m?.ppUsed && (m.getMovePp() - m.ppUsed) <= 5 && m.ppUsed > Math.floor(m.getMovePp()/2)).length).length, 3);
       return thresholdPartyMemberCount * 3;
     }, 9),
     new WeightedModifierType(modifierTypes.MAX_ETHER, (party: Pokemon[]) => {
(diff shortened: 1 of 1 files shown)


damocleas (approved):

src/modifier/modifier-type.ts
```diff
@@ -1451,11 +1451,11 @@ const modifierPool: ModifierPool = {
       return thresholdPartyMemberCount;
     }, 3),
     new WeightedModifierType(modifierTypes.ETHER, (party: Pokemon[]) => {
-      const thresholdPartyMemberCount = Math.min(party.filter(p => p.hp && p.getMoveset().filter(m => m?.ppUsed && (m.getMovePp() - m.ppUsed) <= 5).length).length, 3);
+      const thresholdPartyMemberCount = Math.min(party.filter(p => p.hp && p.getMoveset().filter(m => m?.ppUsed && (m.getMovePp() - m.ppUsed) <= 5 && m.ppUsed >= Math.floor(m.getMovePp()/2)).length).length, 3);
```
torranx, line 1454: nit: spacing
```suggestion
      const thresholdPartyMemberCount = Math.min(party.filter(p => p.hp && p.getMoveset().filter(m => m?.ppUsed && (m.getMovePp() - m.ppUsed) <= 5 && m.ppUsed >= Math.floor(m.getMovePp() / 2)).length).length, 3);
```

Fontbane: I'll be doing another PR in this file soon, I can do it then

frutescens (approved):
Same as what Torranx said re: spacing + please follow up in this PR when you make the PR that does fix the spacing.

Issue #3385, opened 2024-08-06
[Bug] Desolate Land persists when the Pokemon is switched out during the transition between a double battle & single battle


YamiTheYin:
### Describe the bug

Desolate Land persisted even when the Pokemon with the ability was switched out during the transition between a (wild) double battle to a (wild) single battle.

### User data export file

[data_YamitheYin.txt](https://github.com/user-attachments/files/16513378/data_YamitheYin.txt)

### Session export file

[sessionData1_YamitheYin.txt](https://github.com/user-attachments/files/16513386/sessionData1_YamitheYin.txt)

### Expected behavior

Harsh sunlight is supposed to dissipate when the only ability holder of Desolate Land leaves the field.

### Screenshots / Videos

https://github.com/user-attachments/assets/18aaacb4-e147-4a9b-8b7d-1238629261f3

### What OS did you observe the bug on?

PC/Windows

### Which browser do you use?

Chrome

Snailman11:
Link to discord's (Primal Groudon's Desolate Land/Scorching sun remains up after partner battles.)
https://discord.com/channels/1125469663833370665/1262588017353298060

PR #3493, opened 2024-08-11
[Bug] Fixed primal weather remaining active during transition from double to single battle

tlelievre26:
## What are the changes?
Fixes a bug where primal weather would not go away when the mon with the ability is forced back into it's ball during the transition from a double to a single battle.

Relevant issue: https://github.com/pagefaultgames/pokerogue/issues/3385

## What did change?

The `ReturnPhase` class that controls the forced switch going from a double to a single battle now calls the `applyPreSwitchOutAbAttrs` function for controlling how abilities behave during a switch.

## How to test the changes?

Used the session file from the relevant issue until I got to a double battle and observed that Desolate Land correctly got deactivated.

Unsure if this change would justify any automated tests so if I should add some please let me know!

## Checklist
- [ ] Are the changes visual?


DayKev:
Duplicate of #2243 ?

tlelievre26:
> Duplicate of #2243 ?

Seems like it, nobody had mentioned working on it in the Discord issue so I wasn't aware there was another fix for it.

DayKev:
I notice you also added `this.lastPokemon = this.getPokemon();` whereas the other one didn't, is this also a necessary change to fix something?

tlelievre26:
> I notice you also added `this.lastPokemon = this.getPokemon();` whereas the other one didn't, is this also a necessary change to fix something?

The other PR called `applyPreSwitchoutAbAttributes` in the `onEnd` function, which already has a variable defined for this:
https://github.com/pagefaultgames/pokerogue/blob/e46491098d9edb17db2fdb741190f873178971f7/src/phases.ts#L1742

I added the call to the `switchAndSummon` function instead because that's where the attributes get applied in the `SwitchSummonPhase`, which meant I needed to add a variable to hold `this.getPokemon()`

src/phases/return-phase.ts
```diff
@@ -1,4 +1,5 @@
 import BattleScene from "#app/battle-scene.js";
+import { applyPreSwitchOutAbAttrs, PreSwitchOutAbAttr } from "#app/data/ability.js";
```
DayKev, line 2:
```suggestion
import { applyPreSwitchOutAbAttrs, PreSwitchOutAbAttr } from "#app/data/ability";
```

DayKev (approved):

DayKev:
> Unsure if this change would justify any automated tests so if I should add some please let me know!

Probably, so I added the test lol.

DayKev (approved):

torranx:
I'm thinking should the test be in `weather_strong_winds.ts` instead?

DayKev:
> I'm thinking should the test be in `weather_strong_winds.ts` instead?

Possibly. The primals' abilities should have some tests eventually, and this weather is directly linked to Mega-Ray and its ability though.

frutescens:
Hi! Could you move the tests to arena/weather_strong_winds.test.ts instead?

Chapybara-jp (approved):

Issue #1475, opened 2024-05-28
Trainers can use unevolved mons late-game if those mons have alternate evolution requirements


DayKev:
I just fought the rival battle on floor 95 and he had a Feebas (so effectively -1 Pokémon). There should probably be some arbitrary level set for Pokémon with alternate evolution requirements so that if a trainer is given one of them at or above that level they'll be given the evolved form instead.

![image](https://github.com/pagefaultgames/pokerogue/assets/34855794/cb0fd749-b76b-4e4f-a517-efb982a00bd9)

Tempo-anon:
This should have been fixed. If it has not, please let us know.

PR #3499, opened 2024-08-12
[Balance] Make sure trainers are using fully evolved Pokémon by gym 3

DayKev:
## What are the changes?
Starting at the third gym, trainers will no longer have Pokémon that aren't fully evolved (except for intentionally weak trainers like the Breeder class).

Request from balance team member Blitzy (and also because it's silly that the Rival fight on 95 could have an unevolved Growlithe or Feebas).

## What did change?
Generated Pokémon for trainer teams are forced to be fully evolved starting at wave 80 (the third gym). A new `currentWave` parameter was added to `getSpeciesForLevel()` to facilitate this check.

## How to test the changes?
Start at wave 95 a few times and check the Rival's party via the console.

## Checklist


damocleas (dismissed):
Approving on my end, this needs to be fixed

src/data/pokemon-species.ts
```diff
@@ -730,6 +730,11 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
           evolutionChance = Math.min(0.65 * easeInFunc(Math.min(Math.max(level - evolutionLevel, 0), preferredMinLevel) / preferredMinLevel) + 0.35 * easeOutFunc(Math.min(Math.max(level - evolutionLevel, 0), preferredMinLevel * 2.5) / (preferredMinLevel * 2.5)), 1);
         }
       }
+      // (Most) Trainers shouldn't be using unevolved Pokemon by the third gym leader
```
frutescens, line 733: Could you include the note about Breeders' teams in this comment ? I think it would also be good to include `... by the third gym leader / wave 80` in the comment as well, so people have a more numerical idea.

frutescens, line 733:
```suggestion
      /* (Most) Trainers shouldn't be using unevolved Pokemon by the third gym leader / wave 80. Exceptions to this include Breeders, whose large teams are counterbalanced by the use of weaker pokemon */
```

frutescens (dismissed):

Snailman11:
Linking Kev's bug report (the lvl 95 Feebas), makes this easier to find
https://github.com/pagefaultgames/pokerogue/issues/1475

src/data/daily-run.ts
```diff
@@ -49,7 +49,7 @@ export function getDailyRunStarters(scene: BattleScene, seed: string): Starter[]
       const costSpecies = Object.keys(speciesStarters)
         .map(s => parseInt(s) as Species)
         .filter(s => speciesStarters[s] === cost);
-      const starterSpecies = getPokemonSpecies(getPokemonSpecies(Utils.randSeedItem(costSpecies)).getTrainerSpeciesForLevel(startingLevel, true, PartyMemberStrength.STRONGER));
+      const starterSpecies = getPokemonSpecies(getPokemonSpecies(Utils.randSeedItem(costSpecies)).getTrainerSpeciesForLevel(startingLevel, true, PartyMemberStrength.STRONGER, 0));
```
flx-sta, line 52: Can we separate that into 2 variables please? This ready absolutely horrible
e.g.
```suggestion
      const randPkmSpecies = getPokemonSpecies(Utils.randSeedInt(costSpecies));
      const starterSpecies = getPokemonSpecies(randPkmSpecies.getTrainerSpeciesForLevel(startingLevel, true, PartyMemberStrength.STRONGER, 0));
```

src/data/pokemon-species.ts
```diff
@@ -648,8 +648,8 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
     return this.getSpeciesForLevel(level, allowEvolving, false, (isBoss ? PartyMemberStrength.WEAKER : PartyMemberStrength.AVERAGE) + (gameMode?.isEndless ? 1 : 0));
   }
 
-  getTrainerSpeciesForLevel(level: integer, allowEvolving: boolean = false, strength: PartyMemberStrength): Species {
-    return this.getSpeciesForLevel(level, allowEvolving, true, strength);
+  getTrainerSpeciesForLevel(level: integer, allowEvolving: boolean = false, strength: PartyMemberStrength, currentWave: number): Species {
```
flx-sta, line 651: no default `0` here for `currentWave: number` ?

src/data/pokemon-species.ts
```diff
@@ -754,7 +759,7 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
 
     for (const weight of evolutionPool.keys()) {
       if (randValue < weight) {
-        return getPokemonSpecies(evolutionPool.get(weight)!).getSpeciesForLevel(level, true, forTrainer, strength); // TODO: is the bang correct?
+        return getPokemonSpecies(evolutionPool.get(weight)!).getSpeciesForLevel(level, true, forTrainer, strength, currentWave); // TODO: is the bang correct?
```
flx-sta, line 762: Is it possible to get rid of the bang `!` used behind `evolutionPool.get(weight)!` too? That would be good

src/data/pokemon-species.ts
DayKev: Oh yeah, I had wanted to make sure I didn't miss any invocations of it and forgot to add the default after.

src/data/pokemon-species.ts
DayKev: It would require `getPokemonSpecies()` to handle `undefined`; what should it do then? There's not really a "default" Pokémon. It could throw up an error I suppose?

flx-sta: I suppose there is no "prettier" way, yes

DayKev:
```ts
/**
 * Gets the {@linkcode PokemonSpecies} object associated with the {@linkcode Species} enum given
 * @param species The species to fetch
 * @returns The associated {@linkcode PokemonSpecies} object
 */
export function getPokemonSpecies(species: Species | Species[] | undefined): PokemonSpecies {
  if (species === undefined) {
    throw new Error("`species` must not be undefined in `getPokemonSpecies()`");
  }
  // If a special pool (named trainers) is used here it CAN happen that they have a array as species (which means choose one of those two). So we catch that with this code block
  if (Array.isArray(species)) {
    // Pick a random species from the list
    species = species[Math.floor(Math.random() * species.length)];
  }
  if (species >= 2000) {
    return allSpecies.find(s => s.speciesId === species)!; // TODO: is this bang correct?
  }
  return allSpecies[species - 1];
}
```
```
diff --git a/src/data/pokemon-species.ts b/src/data/pokemon-species.ts
index d27dfa46b3e..b2cebe44c0e 100644
--- a/src/data/pokemon-species.ts
+++ b/src/data/pokemon-species.ts
@@ -25,7 +25,15 @@ export enum Region {
   PALDEA
 }
 
-export function getPokemonSpecies(species: Species | Species[]): PokemonSpecies {
+/**
+ * Gets the {@linkcode PokemonSpecies} object associated with the {@linkcode Species} enum given
+ * @param species The species to fetch
+ * @returns The associated {@linkcode PokemonSpecies} object
+ */
+export function getPokemonSpecies(species: Species | Species[] | undefined): PokemonSpecies {
+  if (species === undefined) {
+    throw new Error("`species` must not be undefined in `getPokemonSpecies()`");
+  }
   // If a special pool (named trainers) is used here it CAN happen that they have a array as species (which means choose one of those two). So we catch that with this code block
   if (Array.isArray(species)) {
     // Pick a random species from the list
@@ -756,7 +764,7 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
 
     for (const weight of evolutionPool.keys()) {
       if (randValue < weight) {
-        return getPokemonSpecies(evolutionPool.get(weight)!).getSpeciesForLevel(level, true, forTrainer, strength, currentWave); // TODO: is the bang correct?
+        return getPokemonSpecies(evolutionPool.get(weight)).getSpeciesForLevel(level, true, forTrainer, strength, currentWave);
       }
     }
```
The potential change to `getPokemonSpecies()`.

flx-sta: I like it just 1 tiny complaint:

I’m not a fan of specific `=== undefined` checks. They broke my neck during the `strict-null` migration 🥲

Instead I’d prefer a simple `!species` check.

another thing I noticed is the bang behind the find inside this function but that’s something for another time

flx-sta (dismissed):
Fought rival Finn 3 times.
No unevolved pokemon. 
lgtm

frutescens (approved):

CodeTappert (approved):

PR #3508, opened 2024-08-12
[Move] Finish Alluring Voice, Burning Jealousy and Lash Out

DayKev:
Fully implements Alluring Voice, Burning Jealousy, and Lash Out.

Reduce unimplemented moves.

* Added `battleStatsIncrease: boolean` and `battleStatsDecrease: boolean` to `PokemonTurnData`
* Added `ConfuseIfBoostedAttr` class for Alluring Voice, which will confuse the target if their stats have been boosted this turn.
* Added `BurnIfBoostedAttr` class for Burning Jealousy, which will burn the target(s) if their stats have been boosted this turn.

### Screenshots/Videos

https://github.com/user-attachments/assets/b240fcbb-b95c-4811-bec2-8a1059f46406

## How to test the changes?
`npm run test alluring_voice`, `npm run test burning_jealousy`, `npm run test lash_out`
For Alluring Voice and Burning Jealousy, the following overrides should suffice for testing:
```ts
const overrides = {
  OPP_LEVEL_OVERRIDE: 10,
  OPP_SPECIES_OVERRIDE: Species.REGIELEKI,
  OPP_MOVESET_OVERRIDE: Array(4).fill(Moves.HOWL),
  OPP_ABILITY_OVERRIDE: Abilities.ICE_SCALES,
  STARTER_SPECIES_OVERRIDE: Species.SHUCKLE,
  MOVESET_OVERRIDE: [Moves.ALLURING_VOICE, Moves.BURNING_JEALOUSY],
```
Make sure to use Burning Jealousy before Alluring Voice so the enemy can't hit itself in confusion.
For Lash Out:
```ts
const overrides = {
  OPP_LEVEL_OVERRIDE: 10,
  OPP_SPECIES_OVERRIDE: Species.REGIELEKI,
  OPP_MOVESET_OVERRIDE: Array(4).fill(Moves.HOWL),
  OPP_ABILITY_OVERRIDE: Abilities.FUR_COAT,
  STARTER_SPECIES_OVERRIDE: Species.SHUCKLE,
  MOVESET_OVERRIDE: [Moves.LASH_OUT],
```
You'll need to watch the console output to verify the power is increased.

## Checklist
  - This is an updated version of #1619


frutescens:
I think that instead of a tag, it might be more straightforward to add an additional variable to PokemonTurnData that stores a list of the differences in battle stats within the turn. 
something like... 
PokemonTurnData.battleStatsChange = [+2, 0, 0, 0, 0, 0]

src/test/moves/alluring_voice.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    game.override
+      .battleType("single")
+      .disableCrits()
+      .enemySpecies(Species.MAGIKARP)
+      .enemyAbility(Abilities.ICE_SCALES)
+      .enemyMoveset(Array(4).fill(Moves.HOWL))
+      .startingLevel(10)
+      .enemyLevel(10)
+      .starterSpecies(Species.FEEBAS)
+      .ability(Abilities.BALL_FETCH)
```
schmidtc1, line 39: I believe it's preferred that we use `Abilities.NONE` instead of `Abilities.BALL_FETCH`

DayKev: If you use `Abilities.NONE` that disables the override.

DayKev:
> I think that instead of a tag, it might be more straightforward to add an additional variable to PokemonTurnData that stores a list of the differences in battle stats within the turn. something like... PokemonTurnData.battleStatsChange = [+2, 0, 0, 0, 0, 0]

Done. 👍

src/field/pokemon.ts
```diff
@@ -4235,6 +4235,7 @@ export class PokemonBattleData {
   public berriesEaten: BerryType[] = [];
   public abilitiesApplied: Abilities[] = [];
   public abilityRevealed: boolean = false;
+  public statsBoostedFirstTurn: boolean = false;
```
schmidtc1, line 4238: Is this used anywhere?

DayKev: Doesn't look like it.

src/field/pokemon.ts
```diff
@@ -4245,15 +4245,16 @@ export class PokemonBattleSummonData {
 }
 
 export class PokemonTurnData {
-  public flinched: boolean;
-  public acted: boolean;
+  public flinched: boolean = false;
+  public acted: boolean = false;
   public hitCount: integer;
```
schmidtc1, line 4250: Could you change all of these `integer` variables to `number`, while you're working with this class.

schmidtc1 (approved):
Lgtm

frutescens (dismissed):

ElliottSimmonds:
Thanks for continuing this! Good work 🙂

src/phases/stat-change-phase.ts
```diff
@@ -85,6 +85,14 @@ export class StatChangePhase extends PokemonPhase {
...
 
       for (const stat of filteredStats) {
+        if (levels.value > 0 && pokemon.summonData.battleStats[stat] + levels.value <= 6) {
+          if (!pokemon.turnData) {
+            // Temporary fix for missing turn data struct on turn 1
+            pokemon.resetTurnData();
+          }
+          pokemon.turnData.battleStatsChange[stat] += levels.value;
+        }
+
```
xsn34kzx, line 95: Looking at the Bulbapedia pages for both moves, it sounds like a given stat stage only needs to be "raised" once during the turn rather than necessarily being collectively positive, if that makes sense. 

The scenario is basically, if somehow the target had its a random stat stage raised by 1 and then lowered by 1 in the same turn, should either move get their secondary effect? Under this implementation, they wouldn't since that stat stage wouldn't be above 0 (assuming the rest are 0). 

If the raise occurring at all during the turn is fine, we can probably make `battleStatsChange` into a single boolean that gets set to true any time a positive stat stage change occurs (which would then need to be checked against Contrary to see if inverted stat stages count or not). Probably gonna need a little more exploration since Bulbapedia doesn't really outline these edge cases outside of the immediate description.

xsn34kzx (commented):
Had a concern about an edge case concerning both moves and this implementation. Other than that, I think everything else looks good!

src/phases/stat-change-phase.ts
DayKev: Ah, you're right.

> Burning Jealousy/Lash Out still trigger their effects if the relevant stat boosts/drops are no longer boosted/dropped at the time the user attacks ([Anubis](https://www.smogon.com/forums/posts/8515366/))
- https://www.smogon.com/forums/threads/sword-shield-battle-mechanics-research.3655528/

DayKev: Also I've learned that Lash Out exists and is basically the inverse of Burning Jealousy so I can add that move to this PR too.

src/test/moves/lash_out.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Moves } from "#enums/moves";
+import { Species } from "#enums/species";
+import GameManager from "#test/utils/gameManager";
+import { getMovePosition } from "#test/utils/gameManagerUtils";
+import Phaser from "phaser";
+import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+
+const TIMEOUT = 20 * 1000;
+
+describe("Moves - Burning Jealousy", () => {
```
schmidtc1, line 13:
```suggestion
describe("Moves - Lash Out", () => {
```
Copy error here

src/data/move.ts
```diff
@@ -8517,10 +8568,10 @@ export function initMoves() {
     new AttackMove(Moves.SKITTER_SMACK, Type.BUG, MoveCategory.PHYSICAL, 70, 90, 10, 100, 0, 8)
       .attr(StatChangeAttr, BattleStat.SPATK, -1),
     new AttackMove(Moves.BURNING_JEALOUSY, Type.FIRE, MoveCategory.SPECIAL, 70, 100, 5, 100, 0, 8)
-      .target(MoveTarget.ALL_NEAR_ENEMIES)
-      .partial(),
+      .attr(StatusIfBoostedAttr, StatusEffect.BURN)
+      .target(MoveTarget.ALL_NEAR_ENEMIES),
     new AttackMove(Moves.LASH_OUT, Type.DARK, MoveCategory.PHYSICAL, 75, 100, 5, -1, 0, 8)
-      .partial(),
+      .attr(MovePowerMultiplierAttr, (user, target, move) => user.turnData.battleStatsDecreased ? 2 : 1),
```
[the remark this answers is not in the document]
DayKev: Just checked, Haze doesn't cause the power to double (and that makes sense, because it doesn't cause a `StatChangePhase` to occur).

schmidtc1 (approved):
Lgtm!

xsn34kzx (dismissed):

frutescens (dismissed):

frutescens (approved):

frutescens (approved):
Already tested + approved in the past

innerthunder (approved):

Issue #3218, opened 2024-07-29
[Bug] Some prompt options no longer loop-around the cursor


Anusien:
### Describe the bug

On the team selection screen, when selecting a Pokemon you get a list of items like Add to Party, Toggle IVs, etc. You used to be able to wrap around this list: pushing up at the top went to the bottom and pushing down at the bottom went to the top. This no longer works.

### Expected behavior

Can circularly wrap lists in both directions.

### What OS did you observe the bug on?

Mac/OSX

### Which browser do you use?

Chrome

flx-sta:
@Anusien would you be able to provide a list (and most preferably a video) of all the ones that can't do that (anymore?)

Anusien:
I don't know that a video would help. It would just be me hitting the up or down key at a menu and having nothing happen. But here's the list I've found so far:

Broken:
* Opening menu
* Will you switch Y/N (I don't know that this one worked before tbh)
* Use IV Scanner Y/N (I don't know that this one worked before tbh)
* Select a game mode (I don't know that this one worked before tbh)
* The menu when you select a Pokemon when starting a game: Add to Party, Toggle IVs...
* Select nature. menu when starting a game
* Use candy men when starting a game
* Your party is full. Release a Pokemon to make room for Wild X?

Working:
* Pokeball selection
* Choose a Pokemon: the send out screen
* Do what with this Pokemon?
* Scrolling through a Pokemon's move list
* Hitting ESC and getting the menu with Game Settings, Achievements, Stats...
* Do what with this Pokemon: the menu when you're in the shop, select an item like Soul Dew, and then select a Pokemon

flx-sta:
Could you adjust the title to say `[Bug] Some prompt options no longer loop-around the cursor`

Snailman11:
Can't replicate, probably fixed

PR #3518, opened 2024-08-12
[QoL][UI] Add cursor wrap around to option select menus

MokaStitcher:
## What are the changes?
Enable cursor wrap around for some menus that didn't have it

Some menus especially in the starter select UI are starting to have a lot of options and not being able to wrap around quickly to the top or bottom is annoying.
Closes issue #3218 

## What did change?
Update the `AbstractOptionSelectUiHandler` to enable cursor wrap around. All of the following menus that didn't have them now have it (there may be more):
- Opening menu
- Main menu: Manage Data / Import Session / Export Session
- Settings: Language choice / Controller choice
- All Yes/No menus
- Game mode choice
- Starter UI: Pokemon menu
- Starter UI: Manage Nature menu
- Starter UI: Manage Move menus (menu with current moves and menu with moves to pick from)
- Starter UI: Use Candies menu

Note:
Since all menus in the game don't necessarily use that handler class it's possible that some still don't wrap around. It might be good to make an inventory of these at some point

### Screenshots/Videos

https://github.com/user-attachments/assets/de6f6e60-c07f-44ed-85a3-fda7b30523c8

## How to test the changes?
Run the game and try to wrap around in as many menus as possible, including those with potentially a lot of options like the Manage Nature menu in stater select ui

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/ui/abstact-option-select-ui-handler.ts
+++ b/src/ui/abstact-option-select-ui-handler.ts
@@ -197,11 +197,15 @@ export default abstract class AbstractOptionSelectUiHandler extends UiHandler {
       case Button.UP:
         if (this.cursor) {
           success = this.setCursor(this.cursor - 1);
+        } else if (this.cursor === 0) {
+          success = this.setCursor(options.length -1);
         }
         break;
       case Button.DOWN:
         if (this.cursor < options.length - 1) {
           success = this.setCursor(this.cursor + 1);
+        } else {
+          success = this.setCursor(0);
         }
         break;
       }
@@ -268,11 +272,13 @@ export default abstract class AbstractOptionSelectUiHandler extends UiHandler {
     let isScroll = false;
     const options = this.getOptionsWithScroll();
     if (changed && this.config?.maxOptions && this.config.options.length > this.config.maxOptions) {
-      const optionsScrollTotal = options.length;
       if (Math.abs(cursor - this.cursor) === options.length - 1) {
+        // Wrap around the list
+        const optionsScrollTotal = this.config.options.length;
         this.scrollCursor = cursor ? optionsScrollTotal - (this.config.maxOptions - 1) : 0;
         this.setupOptions();
       } else {
+        // Move the cursor up or down by 1
         const isDown = cursor && cursor > this.cursor;
         if (isDown) {
           if (options[cursor].label === scrollDownLabel) {


frutescens (commented):
Looks good. Have you checked if the language selection menu in settings still works? Have you also checked if the Summary/Yes/No option when catching Pokemon with a full party still works as well?

MokaStitcher:
> Looks good. Have you checked if the language selection menu in settings still works? Have you also checked if the Summary/Yes/No option when catching Pokemon with a full party still works as well?

they do @frutescens (and wrap around, not sure if that's new for these)

frutescens:
They should since their menus are based off of the AbstractOptionUiHandler class as well. Thank you for doing this! I'll send my approval in now :)

frutescens (approved):

torranx (approved):
looks good but can you check if it still works on audio settings? also just noticed the typo on the file name kekw

src/ui/abstact-option-select-ui-handler.ts
```diff
@@ -268,11 +272,13 @@ export default abstract class AbstractOptionSelectUiHandler extends UiHandler {
     let isScroll = false;
     const options = this.getOptionsWithScroll();
     if (changed && this.config?.maxOptions && this.config.options.length > this.config.maxOptions) {
-      const optionsScrollTotal = options.length;
```
KimJeongSun, line 271: With the addition of rename and favorite options to the Pokémon menu, I believed that the scroll wrap feature was essential this time (though, admittedly, I didn't implement it in advance. Sorry about that! 🫠 ).
I'm thankful @MokaStitcher for addressing this! 🙏
The code looks clear and great overall. However, there's a part I didn't quite understand and would like to ask about. 🤔
Why does this need to be moved and defined differently?

MokaStitcher:
> looks good but can you check if it still works on audio settings? also just noticed the typo on the file name kekw

@torranx yep the settings are fine, I believe they use a different handler altogether

src/ui/abstact-option-select-ui-handler.ts
MokaStitcher: I moved the declaration in the if statement since it was only used there

Then, we get in the if statement if we want to wrap around. When we want to wrap to the end of the list we need to be careful if there are more elements in the menu than can be shown on screen at once. In this case we need to know how many elements are not shown because we scrolled down. This is what `this.scrollCursor` is for 

The math to know how much we need to scroll down to get to the end of the list should be `{total number of elements in the list} - {how many can actually be shown at once}`
Before the computation was wrong because it used the number of elements currently displayed on screen (`this.getOptionsWithScroll().length`) instead of the total number of options in the menu (`config.options.length`). So if scrolling was needed it would always set scrollcursor  to 1 instead of the correct number

(Looking at it now it's a bit weird that this code can set scrollcursor to a negative value lol. The rest of the code seems to handle it fine but maybe I should set it to 0 if that's the case, to be safe)

Hope that makes sense

KimJeongSun: great! good to hear that! 👍

PR #3529, opened 2024-08-13
[Enhancement] Add support for simulated calls to Abilities

innerthunder:
## What are the changes?

No user-facing changes -- this just adds a `simulated` argument to all abilities' `apply` functions in support of future implementation of simulated calls.

As a part of an upcoming AI rework, the enemy will need to be able to simulate outcomes for each of its possible actions with no adverse effects on the battle itself. For many abilities, this would require a distinction between an in-battle `apply` and a simulated `apply`. This PR adds that distinction.

## What did change?

- `data/ability`: All ability attributes now have a `simulated` argument for their `apply` function (and all variants of `apply`, e.g. `applyPreDefend`). The exported `applyXAttrs` functions also now have the option to run quietly (i.e. without the ability UI flyouts) with a similar `simulated` argument. **[WIP]**
- Fixed calls to `applyXAttrs` in `data/move`, `field/pokemon`, `phases`, etc. to reflect the new `simulated` argument (This arg is `false` for all existing calls to `applyXAttrs` atm).

### Screenshots/Videos

N/A

## How to test the changes?

simulated call tests are WIP

## Checklist


Xavion3 (commented):
This looks fine from a quick read, I tried to keep an eye on the major parts, but as hundreds of related one line changes I can't be 100% sure that something wasn't missed.

Xavion3 (approved):

src/data/ability.ts
```diff
@@ -2126,9 +2205,9 @@ export class PostSummonCopyAbilityAbAttr extends PostSummonAbAttr {
   private target: Pokemon;
   private targetAbilityName: string;
 
-  applyPostSummon(pokemon: Pokemon, passive: boolean, args: any[]): boolean {
+  applyPostSummon(pokemon: Pokemon, passive: boolean, simulated: boolean, args: any[]): boolean {
     const targets = pokemon.getOpponents();
-    if (!targets.length) {
+    if (simulated || !targets.length) {
       return false;
```
DayKev, line 2211: Shouldn't this also be `return simulated;` (such as slightly below in `PostSummonTransformAbAttr`)?

src/data/ability.ts
```diff
@@ -3008,12 +3122,16 @@ export class PostTerrainChangeAddBattlerTagAttr extends PostTerrainChangeAbAttr
...
+    if (simulated || !this.terrainTypes.find(t => t === terrain)) {
       return false;
     }
 
-    return pokemon.addTag(this.tagType, this.turnCount);
+    if (simulated) {
+      return pokemon.canAddTag(this.tagType);
+    } else {
+      return pokemon.addTag(this.tagType, this.turnCount);
+    }
```
DayKev, line 3134: If `simulated` is true, won't this second `if (simulated) { ... }` block never be reached?

src/data/ability.ts
```diff
@@ -4260,99 +4430,98 @@ export function applyBattleStatMultiplierAbAttrs(attrType: Constructor<BattleSta
...
+  pokemon: Pokemon, terrain: TerrainType, simulated: boolean = false, ...args: any[]): Promise<void> {
+  return applyAbAttrsInternal<PostTerrainChangeAbAttr>(attrType, pokemon, (attr, passive) => attr.applyPostTerrainChange(pokemon, passive, simulated, terrain, args), args, false, simulated);
 }
 
 export function applyCheckTrappedAbAttrs(attrType: Constructor<CheckTrappedAbAttr>,
-  pokemon: Pokemon, trapped: Utils.BooleanHolder, otherPokemon: Pokemon, isQuiet: boolean, messages: string[], ...args: any[]): Promise<void> {
-  return applyAbAttrsInternal<CheckTrappedAbAttr>(attrType, pokemon, (attr, passive) => attr.applyCheckTrapped(pokemon, passive, trapped, otherPokemon, args), args, false, isQuiet, messages);
+  pokemon: Pokemon, trapped: Utils.BooleanHolder, otherPokemon: Pokemon, isQuiet: boolean, messages: string[], simulated: boolean = false, ...args: any[]): Promise<void> {
+  return applyAbAttrsInternal<CheckTrappedAbAttr>(attrType, pokemon, (attr, passive) => attr.applyCheckTrapped(pokemon, passive, simulated, trapped, otherPokemon, args), args, false, simulated, messages);
```
DayKev, line 4514: There's still an `isQuiet` param but it's unused now from what I can see, shouldn't it be removed then?

innerthunder: removed

src/data/ability.ts
innerthunder: Removed the first check for `simulated`

src/data/ability.ts
innerthunder: I changed this but it's a bit different from `PostSummonTransformAbAttr` since there's an extra condition that can make this apply return `false` further down

frutescens (approved):

DayKev (approved):

PR #3532, opened 2024-08-13
[Bug] Palafin should not swap to Hero form when switching due to faint

DayKev:
## What are the changes?
Palafin will no longer swap to its Hero form when switching out due to fainting.

The current behavior is incorrect. cf https://bulbapedia.bulbagarden.net/wiki/Zero_to_Hero_(Ability)

## What did change?
The form change checks whether Palafin is fainted or not. Additional tests were also added.

## How to test the changes?
Have Palafin use Memento while in base form (with another Pokémon to switch to in the back) and then check whether Palafin swaps forms after switching to the unfainted Pokémon.

## Checklist

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -5225,7 +5225,7 @@ export function initAbilities() {
       .attr(NoTransformAbilityAbAttr)
       .attr(NoFusionAbilityAbAttr)
       .attr(PostBattleInitFormChangeAbAttr, () => 0)
-      .attr(PreSwitchOutFormChangeAbAttr, () => 1)
+      .attr(PreSwitchOutFormChangeAbAttr, (pokemon) => pokemon.hp > 0 ? 1 : pokemon.formIndex)
       .bypassFaint(),
     new Ability(Abilities.COMMANDER, 9)
       .attr(UncopiableAbilityAbAttr)
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -5225,7 +5225,7 @@ export function initAbilities() {
       .attr(NoTransformAbilityAbAttr)
       .attr(NoFusionAbilityAbAttr)
       .attr(PostBattleInitFormChangeAbAttr, () => 0)
-      .attr(PreSwitchOutFormChangeAbAttr, (pokemon) => pokemon.hp > 0 ? 1 : pokemon.formIndex)
+      .attr(PreSwitchOutFormChangeAbAttr, (pokemon) => !pokemon.isFainted() ? 1 : pokemon.formIndex)
       .bypassFaint(),
     new Ability(Abilities.COMMANDER, 9)
       .attr(UncopiableAbilityAbAttr)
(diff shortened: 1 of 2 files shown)


src/data/ability.ts
```diff
@@ -5225,7 +5225,7 @@ export function initAbilities() {
       .attr(NoTransformAbilityAbAttr)
       .attr(NoFusionAbilityAbAttr)
       .attr(PostBattleInitFormChangeAbAttr, () => 0)
-      .attr(PreSwitchOutFormChangeAbAttr, () => 1)
+      .attr(PreSwitchOutFormChangeAbAttr, (pokemon) => pokemon.hp > 0 ? 1 : pokemon.formIndex)
       .bypassFaint(),
```
flx-sta, line 5229: Should `faint` also be checked?

What is the purpose of `bypassFaint()` now?

src/test/abilities/zero_to_hero.test.ts
```diff
@@ -26,41 +29,83 @@ describe("Abilities - ZERO TO HERO", () => {
...
-    const moveToUse = Moves.SPLASH;
-    game.override.battleType("single");
-    game.override.ability(Abilities.ZERO_TO_HERO);
-    game.override.moveset([moveToUse]);
-    game.override.enemyMoveset([Moves.TACKLE, Moves.TACKLE, Moves.TACKLE, Moves.TACKLE]);
+    game.override
+      .battleType("single")
+      .moveset(SPLASH_ONLY)
+      .enemyMoveset(SPLASH_ONLY)
+      .enemyAbility(Abilities.BALL_FETCH);
```
flx-sta, line 36: `game.override.ability(Abilities.ZERO_TO_HERO);`

got lost during the migration

src/data/ability.ts
DayKev: Oh I didn't notice it had `.bypassFaint()`, that's probably what was causing the problem then.

DayKev: Nevermind, after removing that it won't revert to base form properly if it's fainted, that must be why they added that in the first place.

Should I change it to `pokemon.hp > 0 && pokemon.status?.effect !== StatusEffect.FAINT`?

src/test/abilities/zero_to_hero.test.ts
DayKev: There's no need to override the ability because Palafin can only has Zero to Hero as an ability. This also avoids putting ZtH on the irrelevant Pokémon.

src/data/ability.ts
flx-sta: there is `pokemon.isFainted()`

```ts
  isFainted(checkStatus?: boolean): boolean {
    return !this.hp && (!checkStatus || this.status?.effect === StatusEffect.FAINT);
  }
  ```

src/test/abilities/zero_to_hero.test.ts
flx-sta: okay. 
maybe add a test then that ensures that it always has that ability then

src/data/ability.ts
DayKev: Oh of course, maybe I shouldn't do this at 3am. 🤔

src/test/abilities/zero_to_hero.test.ts
DayKev: There's already tests for the other `AbAttr`s it has (the ones that prevent swapping/etc) in different test files.

flx-sta (approved):
lgtm 🚀

PR #3534, opened 2024-08-13
[Ability] Implement Forecast

torranx:
## What are the changes?
implements forecast that only 10% of users will use 😞 

castform is cutiepatootie

## What did change?
TBA

### Screenshots/Videos
TBA

## How to test the changes?
TBA

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -14,7 +14,7 @@ import { ArenaTagSide, ArenaTrapTag } from "./arena-tag";
 import { Stat, getStatName } from "./pokemon-stat";
 import { BerryModifier, PokemonHeldItemModifier } from "../modifier/modifier";
 import { TerrainType } from "./terrain";
-import { SpeciesFormChangeManualTrigger } from "./pokemon-forms";
+import { SpeciesFormChangeManualTrigger, SpeciesFormChangeWeatherSuppressedFormTrigger, SpeciesFormChangeWeatherTrigger } from "./pokemon-forms";
 import i18next from "i18next";
 import { Localizable } from "#app/interfaces/locales.js";
 import { Command } from "../ui/command-ui-handler";
@@ -26,6 +26,7 @@ import { ArenaTagType } from "#enums/arena-tag-type";
 import { BattlerTagType } from "#enums/battler-tag-type";
 import { Moves } from "#enums/moves";
 import { Species } from "#enums/species";
+import BattleScene from "#app/battle-scene.js";
 
 export class Ability implements Localizable {
   public id: Abilities;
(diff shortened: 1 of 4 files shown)


flx-sta:
10%? Optimism.

jokes aside: according to Reddit some see it as hidden carry

DayKev:
Is it necessary to add logic to Air Lock and Cloud Nine? It seems to me that they should just be suppressing the weather, and not have any references/etc to whatever side-effects that may cause; and Forecast/other things that care about weather should be checking whether it's suppressed or not when deciding what their effect should be.

torranx:
> Is it necessary to add logic to Air Lock and Cloud Nine? It seems to me that they should just be suppressing the weather, and not have any references/etc to whatever side-effects that may cause;

There isn't a way to dynamically change form based on the weather changes so it necessitates manually triggering form changes on every weather changes, including suppressions. But if you happen to find one, let me know.

> Forecast/other things that care about weather should be checking whether it's suppressed or not when deciding what their effect should be.

It is already checking conditions, but as mentioned above, we need to manually trigger form changes.

DayKev:
Maybe whenever the weather state is modified in some way (weather change, suppression change, etc) each Pokémon on the field should be triggered to check for any weather-based effects they need to modify.
```ts
// pseudocode
arena.activePokemon().each(pokemon => pokemon.checkWeatherEffects());

// in the pokemon class
checkWeatherEffects() {
  applyAbAttrs(weatherBasedAbAttr1, arena.weather, args);
  // other stuff that needs to happen immediately on weather change goes here
}
```
Probably something to be done in a future [weather] refactoring PR.

src/test/abilities/forecast.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    await game.startBattle([Species.CASTFORM]);
+
+    game.doAttack(getMovePosition(game.scene, 0, Moves.RAIN_DANCE));
+    await game.phaseInterceptor.to(TurnEndPhase);
+
+    expect(game.scene.getPlayerPokemon()?.formIndex).toBe(RAINY_FORM);
+    expect(game.scene.getEnemyPokemon()?.formIndex).not.toBe(RAINY_FORM);
+  });
+
+  it("cannot be copied", async () => {
```
DayKev, line 177: You don't need a test for `UncopiableAbilityAbAttr`, test(s) already exist for that `AbAttr`.

DayKev:
> Maybe whenever the weather state is modified in some way (weather change, suppression change, etc) each Pokémon on the field should be triggered to check for any weather-based effects they need to modify.

After looking into it, I notice that it's basically already done this way (which after thinking about it yes of course it is lol) *except* for weather suppression, and also it seems like only Castform and Cherrim do something immediately when the weather is [un]suppressed at the moment. Thanks Gamefreak.

DayKev:
I think Castform is supposed to revert to normal form when it switches out, but that doesn't happen currently.

src/data/pokemon-forms.ts
```diff
@@ -839,7 +912,24 @@ export const pokemonFormChanges: PokemonFormChanges = {
...
+    new SpeciesFormChange(Species.CASTFORM, "rainy", "snowy", new SpeciesFormChangeWeatherTrigger(Abilities.FORECAST, [WeatherType.HAIL, WeatherType.SNOW]), true),
+    new SpeciesFormChange(Species.CASTFORM, "sunny", "", new SpeciesFormChangeRevertWeatherFormTrigger(Abilities.FORECAST, [WeatherType.NONE, WeatherType.SANDSTORM, WeatherType.STRONG_WINDS, WeatherType.FOG]), true),
+    new SpeciesFormChange(Species.CASTFORM, "rainy", "", new SpeciesFormChangeRevertWeatherFormTrigger(Abilities.FORECAST, [WeatherType.NONE, WeatherType.SANDSTORM, WeatherType.STRONG_WINDS, WeatherType.FOG]), true),
+    new SpeciesFormChange(Species.CASTFORM, "snowy", "", new SpeciesFormChangeRevertWeatherFormTrigger(Abilities.FORECAST, [WeatherType.NONE, WeatherType.SANDSTORM, WeatherType.STRONG_WINDS, WeatherType.FOG]), true),
+    new SpeciesFormChange(Species.CASTFORM, "sunny", "", new SpeciesFormChangeActiveTrigger(), true),
+    new SpeciesFormChange(Species.CASTFORM, "rainy", "", new SpeciesFormChangeActiveTrigger(), true),
+    new SpeciesFormChange(Species.CASTFORM, "snowy", "", new SpeciesFormChangeActiveTrigger(), true),
```
innerthunder, line 931:
```suggestion
    new SpeciesFormChange(Species.CASTFORM, "sunny", "", new SpeciesFormChangeActiveTrigger(false), true),
    new SpeciesFormChange(Species.CASTFORM, "rainy", "", new SpeciesFormChangeActiveTrigger(false), true),
    new SpeciesFormChange(Species.CASTFORM, "snowy", "", new SpeciesFormChangeActiveTrigger(false), true),
```
Currently, Castform would revert to Normal Form upon reentering the field. This moves that form change to when Castform exits the field.

innerthunder (commented):
This probably needs review from @Xavion3 to check how this would impact Castform in challenges

Xavion3:
It was decided for various reasons that form changes which only occur during combat such as Castform and Meloetta would not count against a monotype challenge, so Castform is currently treated as always legal for mono normal regardless of form. If you implement a way to obtain the other Castform forms while they're not active in battle this may not function correctly anymore, but as long as they're restricted to only applying while active in battle the existing safeguards should hold as Castform was accounted for pre-emptively.

src/data/pokemon-forms.ts
torranx: I've tested both and they're the same- they change form outside of combat but a message appears that it reverted. also it already defaults to `false`:

```
export class SpeciesFormChangeActiveTrigger extends SpeciesFormChangeTrigger {
  public active: boolean;

  constructor(active: boolean = false) {
    super();
    this.active = active;
  }

    return pokemon.isActive(true) === this.active;
  }
}
```

flx-sta (approved):

innerthunder (approved):

innerthunder:
code looks good to me, but fair warning that this will have conflicts with #3529 if that gets in first

DayKev (approved):
Played through the first 10 or so waves with rain + Castform and it seemed to work properly.

Well, you'll have to fix the tests now that the `doAttack()` change finally got merged, but other than that seems good.

src/data/ability.ts
```diff
@@ -3014,6 +3078,45 @@ export class PostWeatherChangeAbAttr extends AbAttr {
...
+   * Calls {@linkcode Arena.triggerWeatherBasedFormChangesToNormal | triggerWeatherBasedFormChangesToNormal} when the
+   * weather changed to form-reverting weather, otherwise calls {@linkcode Arena.triggerWeatherBasedFormChanges | triggerWeatherBasedFormChanges}
+   * @param {Pokemon} pokemon the Pokemon that changed the weather
+   * @param passive n/a
+   * @param weather n/a
+   * @param args n/a
+   * @returns whether the form change was triggered
+   */
+  applyPostWeatherChange(pokemon: Pokemon, passive: boolean, simulated: boolean, weather: WeatherType, args: any[]): boolean {
+    if (!simulated && pokemon.species.speciesId === Species.CASTFORM && this.ability === Abilities.FORECAST) {
```
innerthunder, line 3105: Can you change this so that when `simulated` and conditions to change form normally would be met, this would return `true` (instead of simulated calls always returning `false`)?

innerthunder (approved):

DayKev (approved):

frutescens (approved):

PR #3556, opened 2024-08-15
[Balance] Ability-ignoring effects no longer ignore the source's own Abilities

innerthunder:
## What are the changes?

This change makes it so that if a Pokemon has an ability or uses a move that ignores ability effects, those ability-ignoring effects do not apply to the user's own abilities. For example, if a Pokemon has Mold Breaker as an ability and Contrary as a passive ability, it using Overheat would now raise the user's Sp. Atk instead of ignoring Contrary and lowering it.

This change was requested by the Balance team, fitting in line with previous changes to Neutralizing Gas (#1662 ). Also, this interaction has been [reported as a bug](https://discord.com/channels/1125469663833370665/1267021266271866890) before in the Discord, which indicates that players may find this interaction unintuitive as it stands.

## What did change?

- `field/arena`: New field `Arena.ignoreAbilitySource` tracks the source of active ability-ignoring effects by battler index, and is set within `setIgnoreAbilities`.
- `field/pokemon`: `Pokemon.canApplyAbility` now checks that `arena.ignoreAbilitySource` does not match the Pokemon's battler index before ignoring abilities' effects.
- `phases`: Small adjustments reflecting the changes to `Arena.setIgnoreAbilities()`

### Screenshots/Videos

Clanging Scales from a Pokemon with Mold Breaker + Contrary after the changes:

https://github.com/user-attachments/assets/49e5dbd3-54a8-4666-8fd8-b236902ebc6b

## How to test the changes?

tested via overrides

## Checklist
- [x] Are the changes visual?


src/field/arena.ts
```diff
@@ -536,8 +537,13 @@ export class Arena {
...
   }
 
-  setIgnoreAbilities(ignoreAbilities: boolean = true): void {
+  setIgnoreAbilities(ignoreAbilities: boolean, ignoreAbilitySource: BattlerIndex | null = null): void {
     this.ignoreAbilities = ignoreAbilities;
+    if (ignoreAbilities) {
+      this.ignoreAbilitySource = ignoreAbilitySource;
+    } else {
+      this.ignoreAbilitySource = null;
+    }
```
DayKev, line 546:
```suggestion
    this.ignoreAbilitySource = ignoreAbilities ? ignoreAbilitySource : null;
```

Xavion3:
It's unintuitive partially because Neutralising gas is bugged to not negate abilities like it should which creates an inconsistency, we can easily fix that by just reverting the change which broke it.

Also this is not a quality of life change, this is a significant change in how an ability works with significant balance and gameplay ramifications. This is even a nerf under some circumstances.

The bug report was also thrown off by the text of mold breaker not lining up with the effect, but this change does not solve that problem so that's not really a good excuse either. Unless we want to make it not ignore stuff like lightning rod.

Xavion3 (dismissed):
As above, this currently fails to actually solve the problem laid out in the reasons for this PR.

src/field/arena.ts
```diff
@@ -536,8 +537,9 @@ export class Arena {
     }
   }
 
-  setIgnoreAbilities(ignoreAbilities: boolean = true): void {
+  setIgnoreAbilities(ignoreAbilities: boolean, ignoreAbilitySource: BattlerIndex | null = null): void {
```
schmidtc1, line 540: To me, `ignoreAbilitySource` reads like a `boolean`. Maybe change the name to something like `abilityBattlerIndex`, `abilitySourceBattlerIndex`, etc.

Fontbane: Or `ignore**d**AbilitySource`

DayKev:
> It's unintuitive partially because Neutralising gas is bugged to not negate abilities like it should which creates an inconsistency, we can easily fix that by just reverting the change which broke it.
> 
> Also this is not a quality of life change, this is a significant change in how an ability works with significant balance and gameplay ramifications. This is even a nerf under some circumstances.
> 
> The bug report was also thrown off by the text of mold breaker not lining up with the effect, but this change does not solve that problem so that's not really a good excuse either. Unless we want to make it not ignore stuff like lightning rod.

So we need balance team comments then.

innerthunder:
Closing this as there is at least some evidence (via Sunsteel Strike/Moongeist Beam) that ability-ignoring effects apply to the user's ability in mainline; it's just that these effects are never available alongside ignorable abilities (hence Shadow Shield and Full Metal Body not being ignorable).

I may reopen this if the Game Design/Balance teams decide custom implementation is needed to remove potential anti-synergy between a Pokemon's passive and ability, but as it stands I'd say the current implementation is more accurate to mainline.

src/field/arena.ts
innerthunder: maybe something along the lines of `attackerIndex`? It's not necessarily an ability that creates this effect (see Moongeist Beam/Sunsteel Strike), and `ignoredAbilitySource` kind of sounds like it's referring to the Pokemon whose ability is ignored, not the one who's doing the ignoring.

damocleas (approved):

DayKev (approved):

flx-sta (approved):

PR #3557, opened 2024-08-15
[Bug] Fix some damage formulas processed with ceil instead of floor

KimJeongSun:
## What are the changes?

Now, Charizard with odd hp can withstand Stealth Rock twice.
When the `Wonder Guard` ability is active, a failed Jump Kick will cause fainting. ( because it has min damage 1 now )

Originally, I had no plans to do this, but I am updating to prevent future confusion among other developers regarding damage calculations.

## What did change?

Unified the use of ceil, floor, and minimum 1 values in the existing damage calculation formulas. Implemented this function as `toIntValue` in `utils.ts` and revised the overall damage calculation formulas to refer to it.
Added `damage_calculation.test.ts` to immediately check the related changes. (Sorry for the longer wait to run the tests!)

## Checklist
- [ ] Are the changes visual?


flx-sta:
Since it's a very reptitive pattern of `Math.max(Math.floor(...), 1)` we should think about a helper/utils method to be used that's descriptive.
Something like 
```ts
function toDmgValue(damage: number, minDamage: number = 1) {
  return Math.max(Math.floor(damage), minDamage);
}
```
<sub>This is just an idea for the function. Name and implementation are subject to opinionated changes</sub>

KimJeongSun:
@flx-sta 
Actually, during the first modification, I was aware of that part, so I considered modifying it to apply min floor inside the damage function. 

However, I hesitated because it was (cumbersome) and the code didn't seem very elegant.

Adding a new function to utils seems like a better idea. I've updated it.

```

/**
 * Convert a float value to an integer value.
 *
 * @param value - The float value to convert.
 * @param minValue - The minimum integer value to return. Defaults to 1.
 * @returns The converted value as an integer.
 */

export function toIntValue(value: number, minValue: number = 1) {
  return Math.max(Math.floor(value), minValue);
}
```

DayKev (approved):

flx-sta:
@KimJeongSun please don't crucify me but the name `toIntValue()` is a bit too generic and might be misleading in it's purpose. I know that there is a thorough jsdocu but the name is really dangerous I feel... Maybe someone elses opinion can resolve this.

KimJeongSun:
@flx-sta 
Changing the name isn't really a problem at all. Don't worry.
If you are concerned about that, changing the function name to something more specific could be a possible solution.

here are a few options.
1. **floatToDamageInt**
2. **ensureIntDamage**
3. **calculateDamageInt**
4. **getIntDamageValue**

flx-sta:
5. `toDmgValue` (lol)

-----

Link to the discord discussion:
[#dev-corner -> 1273892728635261022](https://discord.com/channels/1125469663833370665/1176874654015684739/1273892728635261022)

KimJeongSun:
updated to `toDmgValue`

```
/**
 * Convert a float value to an integer damage value.
 *
 * This function is used in the context of a Pokémon battle game to calculate the actual integer damage value from a float result.
 * Many damage calculation formulas involve various parameters and result in float values.
 * The actual damage applied to a Pokémon's HP must be an integer.
 * This function helps in ensuring that by flooring the float value and enforcing a minimum damage value.
 *
 * @param value - The float value to convert.
 * @param minValue - The minimum integer value to return. Defaults to 1.
 * @returns The converted value as an integer.
 */

export function toDmgValue(value: number, minValue: number = 1) {
  return Math.max(Math.floor(value), minValue);
}
```

innerthunder (approved):
lgtm

DayKev (approved):

src/utils.ts
```diff
@@ -548,3 +548,19 @@ export function capitalizeString(str: string, sep: string, lowerFirstChar: boole
...
+ * The actual damage applied to a Pokémon's HP must be an integer.
+ * This function helps in ensuring that by flooring the float value and enforcing a minimum damage value.
+ *
+ * @param value - The float value to convert.
+ * @param minValue - The minimum integer value to return. Defaults to 1.
+ * @returns The converted value as an integer.
+ */
+
+export function toDmgValue(value: number, minValue: number = 1) {
+  return Math.max(Math.floor(value), minValue);
```
flx-sta, line 565:
```suggestion
 * @param minValue - The minimum integer value to return. Defaults to 1.
 * @returns The converted value as an integer.
 */
export function toDmgValue(value: number, minValue: number = 1) {
  return Math.max(Math.floor(value), minValue);
```

src/test/moves/tackle.test.ts
```diff
@@ -76,6 +76,6 @@ describe("Moves - Tackle", () => {
     await game.phaseInterceptor.runFrom(EnemyCommandPhase).to(TurnEndPhase);
     const hpLost = hpOpponent - game.scene.currentBattle.enemyParty[0].hp;
     expect(hpLost).toBeGreaterThan(0);
-    expect(hpLost).toBe(4);
+    expect(hpLost).toBeLessThan(4);
```
torranx, line 79: why whas this change?

src/utils.ts
```diff
@@ -548,3 +548,18 @@ export function capitalizeString(str: string, sep: string, lowerFirstChar: boole
   }
   return null;
 }
+/**
+ * Convert a float value to an integer damage value.
+ *
+ * This function is used in the context of a Pokémon battle game to calculate the actual integer damage value from a float result.
```
torranx, line 554: This is redundant on the first line.

src/test/moves/tackle.test.ts
KimJeongSun: While correcting the damage calculation formula, I confirmed that the tackle damage was coming out as 3. I didn't want to check in detail how often the tackle damage comes out as 3 because it was bothersome. Additionally, the main purpose of the test code was not to verify the damage, but to confirm that tackle is ineffective against ghost types. So, I set the assert function a bit more loosely

src/utils.ts
KimJeongSun: I wrote the first line as a brief explanation and the detailed explanation starting from the third line. If you don't like this, you can remove the first line. Is that okay?

src/utils.ts
```diff
@@ -548,3 +548,18 @@ export function capitalizeString(str: string, sep: string, lowerFirstChar: boole
...
+ * Many damage calculation formulas involve various parameters and result in float values.
+ * The actual damage applied to a Pokémon's HP must be an integer.
+ * This function helps in ensuring that by flooring the float value and enforcing a minimum damage value.
```
torranx, line 557: I feel like these are unnecessary comments. It should already be evident on the first line. (That's just me tho). Take it with a grain of salt.

torranx (approved):
looks good to me, just have a couple nits

src/utils.ts
flx-sta: It seems redundant. Lemme just give the confirmation here :D

KimJeongSun: i removed 👍  ( i will go and whip hard to my translator )

src/utils.ts
KimJeongSun: i removed redundant comment line (first line)

flx-sta (approved):
lgtm 🚀 
(playest via charizard + stealth rocks)


We should for after the release though

innerthunder (approved):

DayKev (approved):

DayKev (approved):

PR #3567, opened 2024-08-16
[Test] Replace `doAttack()` with `move.select()` in tests

DayKev:
## What are the changes?
There are no user-facing changes.

The function being named `doAttack()` can cause confusion, leading people to believe it will execute the attack when actually it's just selecting it in the menus and not running the phases forward.

## What did change?
`doAttack()` was replaced with `move.select()` and the ability to choose a target was added to the function as well, negating the need to manually call `selectTarget()` after choosing a move.

## How to test the changes?
`npm run test`

## Checklist


innerthunder (approved):

flx-sta:
Technically it shouldn't have the word `attack` in it since it can also be a healing move or something else.

More correct would be `selectMove()` or something.
We could also simplify the argument cause in the end (almost) every call is:
```ts
game.selectMove(getMovePosition(game.scene, n, Moves.X));
```
which is kind of redundant.

```ts
game.selectMove(n, Moves.X)
``` 
as a supported way to call it would be convenient

torranx:
I agree with @flx-sta on this one. But the one I particularly don't like is the word `prepare`. 
If we decide the change it to `selectMove()`, I vote to move it to `moveHelper.ts` instead.

flx-sta:
> I agree with @flx-sta on this one. But the one I particularly don't like is the word `prepare`. If we decide the change it to `selectMove()`, I vote to move it to `moveHelper.ts` instead.

Love the idea.

```ts

/**
 * Helper to handle a Pokemon's move
 */
export class MoveHelper extends GameManagerHelper {
  
  //...
  
  /**
   * Select the move to be used by the given Pokemon(-index)
   * @param move the move to use
   * @param pkmIndex the pokemon index. Relevant for double-battles only (defaults to 0)
   */
  select(move: Moves, pkmIndex: 0 | 1 = 0) {
    this.game.doAttack(getMovePosition(this.game.scene, pkmIndex, move));
  }
  ```
  
>  ⚠️ I did decide to swap move & index here since index is most times `0` and rarely `1` thus making it optional to make usage even more simplified
  
----

An example use case would then be 
```ts
game.move.select(Moves.SPLASH);
```

or in a double battle if the 2nd pokemon is supposed to use it
```ts
game.move.select(Moves.SPLASH, 1);
```

flx-sta (changes requested):
You imported every phases file with `.js` extension

DayKev:
> You imported every phases file with `.js` extension

Whoops, forgot to babysit the auto-import.

torranx (commented):
I notice you're using 2 methods, `selectMove` and `select` which if I'm not mistaken, are both the same. My initial thought was we remove the `doAttack` and replace it with `game.move.select`.

flx-sta:
I have to double down on what @torranx mentioned:
It might be better to move the `game.selectMove()` body into `game.move.select()` body. Otherwise people have an option of using an alternative way resulting in "chaos" again. 

Additionally if possible it might also be good to move `game.doSelectTarget()` too (Only if applicable)

DayKev:
Replaced `selectMove()` with `move.select()`, also combined the target selection PR into this one since I'm refactoring this anyway.

flx-sta:
We should probably get this PR rolling asap as the work to be done will just get more and more. Just gotta make sure that the most up to date beta branch was merged before merging it

DayKev:
The only changes since last time I merged beta were some translations it seems.

flx-sta (approved):

innerthunder (approved):

src/test/utils/gameManager.ts
```diff
@@ -192,38 +192,22 @@ export default class GameManager {
   }
 
   /**
-   * Emulate a player attack
-   * @param movePosition the index of the move in the pokemon's moveset array
+   * Emulate a player's target selection after a move is chosen, usually called automatically by {@linkcode MoveHelper.select}.
```
frutescens, line 195: Could you extend this comment and state that this occurs during a CommandPhase of a battle ?

DayKev: Well `selectTarget()` triggers during `SelectTargetPhase` but I added a note to this and `move.select()` mentioning which Phase they trigger in.

torranx (approved):
well-discussed and looks good to me 👍

PR #3588, opened 2024-08-16
[Bug][Beta] Fix Pokemon info flyout disappearing on switch-out

### LINKED ISSUE #3571: [Bug] Missing/Gone health-bar UI
## Describe the bug

Sometimes the health-bar UI (including the type and all) is misplaced in `beta` right now

This should be fixed before the release!

## Reproduction Steps

> Reported by @DayKev 
- just switching your pokemon back and forth will eventually trigger it

## Expected behavior

The UI element should move back into screen properly

## Screenshots / Videos

image

## Additional context

Discord-Messages:
- https://discord.com/channels/1125469663833370665/1176874654015684739/1273930570824552530

innerthunder:
## What are the changes?

This fixes a bug in beta that causes a Pokemon's info flyout to not appear correctly after the Pokemon switches or is returned before a Trainer battle.

closes #3571 

## What did change?

- `phases.ts`: Removed a possible race condition with calls to `Pokemon.hideInfo` in `SwitchSummonPhase` by adding a 50 ms delay

## How to test the changes?

In game:
- Switch Pokemon while quickly mashing inputs
- Enter a Trainer battle (e.g. Wave 5) while quickly mashing inputs

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -1643,9 +1643,9 @@ export class SwitchSummonPhase extends SummonPhase {
       ease: "Sine.easeIn",
       scale: 0.5,
       onComplete: () => {
-        // 250ms delay on leaveField is necessary to avoid calling hideInfo() twice
+        // 300ms delay on leaveField is necessary to avoid calling hideInfo() twice
         // and double-animating the stats panel slideout
-        this.scene.time.delayedCall(250, () => pokemon.leaveField(!this.batonPass));
+        this.scene.time.delayedCall(300, () => pokemon.leaveField(!this.batonPass));
         this.scene.time.delayedCall(750, () => this.switchAndSummon());
       }
     });


DayKev (approved):
Switched between Pokémon for ~1m and it didn't break, seems good.

src/phases.ts
```diff
@@ -1643,9 +1643,9 @@ export class SwitchSummonPhase extends SummonPhase {
       ease: "Sine.easeIn",
```
frutescens, line 1643: Shouldn't the duration be 300 too then?

frutescens: same with the Pokemon tint animation

snoozbuster (approved):
Seems like an ok fix. When I wrote this originally it didn't occur to me that lining this animation up exactly with the animation in hideInfo might produce inconsistent results. I would like to know how exactly these two pieces of code racing causes the issue, though...

src/phases.ts
frutescens: nevermind

frutescens (approved):

innerthunder:
> Seems like an ok fix. When I wrote this originally it didn't occur to me that lining this animation up exactly with the animation in hideInfo might produce inconsistent results. I would like to know how exactly these two pieces of code racing causes the issue, though...

This is `Pokemon.hideInfo()`, which was called directly in `SwitchSummonPhase` and then called again in `Pokemon.leaveField()`, originally 500 ms later:

```ts
hideInfo(): Promise<void> {
  return new Promise(resolve => {
    if (this.battleInfo.visible) {
      this.scene.tweens.add({
        targets: [ this.battleInfo, this.battleInfo.expMaskRect ],
        x: this.isPlayer() ? "+=150" : `-=${!this.isBoss() ? 150 : 246}`,
        duration: 500,
        ease: "Cubic.easeIn",
        onComplete: () => {
          if (this.isPlayer()) {
            this.battleInfo.expMaskRect.x -= 150;
          }
          this.battleInfo.setVisible(false);
          this.battleInfo.setX(this.battleInfo.x - (this.isPlayer() ? 150 : !this.isBoss() ? -150 : -198));
          resolve();
        }
      });
    } else {
      resolve();
    }
  });
}
```
This only plays the animation if the info container is visible, but the callback after the animation is what's responsible for making the container invisible in the first place. So you had 2 async pieces of code in the race of
1. play info-hiding animation for 500 ms --> make the info container invisible
2. play another animation for 250 ms --> wait 250 ms --> if info container is visible, play info-hiding animation

So if (1) didn't finish before (2) checked the info container's visibility, the animation to hide the container gets played twice. And since the animation changes the container's position based on relative x-position, the final position of the container in this case is unpredictable (but could very well be off the screen).

Fontbane:
LGTM

Issue #3471, opened 2024-08-10
[Bug] Parting shot doesn't apply debuff even if no one to switch out to


DustinLin:
### Describe the bug

From smoogon and showdown, parting shot's debuffs still apply even if there is no one to switch out to (both if the party is just 1 pokemon or if all others are fainted). 

Pr #3412 wrote some tests, but most of them don't pass

### Expected behavior

The move shouldn't fail

### Screenshots / Videos

The move shouldn't fail

### What OS did you observe the bug on?

PC/Windows, Mac/OSX

### Which browser do you use?

Firefox

Snailman11:
https://github.com/pagefaultgames/pokerogue/issues/5379
Replaced with

PR #3589, opened 2024-08-17
[BUG] - fixing parting shot

DustinLin:
## What are the changes?

Making parting shot do the expected behavior (passing all the tests in #3412 )

Closes issue #3471 

## What did change?

Created a new move attribute tag `PartingShotAttr` that implements the behavior of parting shot. The move is unique due to the fact that certain attributes should/shouldn't `apply()` based on the `getCondition()`. Namely, we still want to try all stat changes (and if it fails then the resulting messages/abilities will show correctly), but *don't* want to prompt switch-out at the same time.

## How to test the changes?

`overrides.ts` and `npm test src/tests/move/parting_shot.test.ts`

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -4877,6 +4877,7 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
 
   	// Check if the move category is not STATUS or if the switch out condition is not met
       if (!this.getSwitchOutCondition()(user, target, move)) {
+        console.log("switching out condition is false in the apply");
         return resolve(false);
       }
 
@@ -4885,6 +4886,7 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
 	  const switchOutTarget = this.user ? user : target;
 	  if (switchOutTarget instanceof PlayerPokemon) {
         switchOutTarget.leaveField(!this.batonPass);
+        console.log("switching out , what isgoing on here");
 
         if (switchOutTarget.hp > 0) {
           user.scene.prependToPhase(new SwitchPhase(user.scene, switchOutTarget.getFieldIndex(), true, true), MoveEndPhase);
@@ -4974,6 +4976,47 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
       ret = ret / 2 + (Phaser.Tweens.Builders.GetEaseFunction("Sine.easeOut")(Math.min(Math.abs(battleStatTotal), 10) / 10) * (battleStatTotal >= 0 ? 10 : -10));
     }
     return ret;
+  }
+}
+
+/**
+ * Attr used by parting shot, it is a combo of ForceSwitchOut and StatChange with a special getCondition()
+ */
+export class PartingShotAttr extends ForceSwitchOutAttr {
+  private statChange: StatChangeAttr;
+  private canLowerStats: boolean;
+
+  constructor(user?: boolean, batonPass?: boolean) {
+    super(user, batonPass);
+    this.statChange = new StatChangeAttr([ BattleStat.ATK, BattleStat.SPATK ], -1, false, null, true, true, MoveEffectTrigger.PRE_APPLY);
+    this.canLowerStats = true;
+  }
+
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): Promise<boolean> {
+    // apply stat change, conditionally apply switch-out
+    this.statChange.apply(user, target, move, args);
+    if (this.canLowerStats) {
+      return super.apply(user, target, move, args);
+    } else {
+      return new Promise(resolve => {
+        resolve(false);
+      });
+    }
+  }
+
+  getCondition(): MoveConditionFunc {
+    return (user, target, move) => {
+      // conditions on if move should fail or not don't depend on if user is able to switch
+      // getCondition() is called before move is applied: move will only switch out if canLowerStats
+      if (target.hasAbility(Abilities.CLEAR_BODY, true, false) ||
+          (target.summonData.battleStats[0] === -6 && target.summonData.battleStats[2] === -6) || target.scene.arena.findTagsOnSide(t => t.tagType === ArenaTagType.MIST, ArenaTagSide.ENEMY).length > 0
+      ) {
+        this.canLowerStats = false;
+      } else {
+        this.canLowerStats = true;
+      }
+      return true;
+    };
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/data/move.ts
```diff
@@ -5123,6 +5122,47 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
   }
 }
 
+/**
+ * Attr used by parting shot, it is a combo of ForceSwitchOut and StatChange with a special getCondition()
+ */
+export class PartingShotAttr extends ForceSwitchOutAttr {
+  private statChange: StatStageChangeAttr;
+  private canLowerStats: boolean;
```
DayKev, line 5130:
```suggestion
  private statChange: StatStageChangeAttr = new StatStageChangeAttr([ Stat.ATK, Stat.SPATK ], -1, false, null, true, true, MoveEffectTrigger.PRE_APPLY);
  private canLowerStats: boolean = true;
```

src/data/move.ts
```diff
@@ -5123,6 +5122,47 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
...
+
+  constructor(user?: boolean, batonPass?: boolean) {
+    super(user, batonPass);
+    this.statChange = new StatStageChangeAttr([ Stat.ATK, Stat.SPATK ], -1, false, null, true, true, MoveEffectTrigger.PRE_APPLY);
+    this.canLowerStats = true;
```
DayKev, line 5135:
```suggestion
```

src/data/move.ts
```diff
@@ -5123,6 +5122,43 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
...
+        resolve(false);
+      });
+    }
+  }
+
+  getCondition(): MoveConditionFunc {
+    return (user, target, move) => {
+      // conditions on if move should fail or not don't depend on if user is able to switch
+      // getCondition() is called before move is applied: move will only switch out if canLowerStats === true
+      if (target.hasAbility(Abilities.CLEAR_BODY, true, false) ||
```
innerthunder, line 5150:
```suggestion
      if (target.hasAbilityWithAttr(ProtectStatAbAttr) ||
```
The existing condition doesn't include White Smoke, while this covers both abilities

src/data/move.ts
```diff
@@ -5123,6 +5122,43 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
...
+          (target.getStatStage(Stat.ATK) === -6 && target.getStatStage(Stat.SPATK) === -6) || target.scene.arena.findTagsOnSide(t => t.tagType === ArenaTagType.MIST, ArenaTagSide.ENEMY).length > 0
```
innerthunder, line 5151: The stat stage checks here need to account for Contrary:
> The user will no longer switch out if the target already has both -6 Attack and Special Attack stat stages (unless the target's Ability is [Contrary](https://bulbapedia.bulbagarden.net/wiki/Contrary_(Ability)), in which case the user will not switch out if the target has both +6 Attack and Special Attack stat stages).

ben-lear (dismissed):
Did not test, lgtm for beta though

innerthunder (changes requested):
I'm not convinced that a new attribute is necessary for Parting Shot, as I think the intended behavior can be achieved by adding a condition to the existing attributes for the move:

```ts
new StatusMove(Moves.PARTING_SHOT, Type.DARK, 100, 20, -1, 0, 6)
    .attr(StatStageChangeAttr, [ Stat.ATK, Stat.SPATK ], -1, false, null, true, true, MoveEffectTrigger.PRE_APPLY)
    .attr(ForceSwitchOutAttr, true, false)
    .condition(/* PartingShotCondition with your getCondition logic */)
    .soundBased(),
```

If the added condition returns `false`, the entire move fails, so in theory checking if the move can change stats through that should produce the desired behavior.

src/data/move.ts
DustinLin: thanks, added a test for that too

src/data/move.ts
DustinLin: thanks, added a test for that too

DustinLin:
> I'm not convinced that a new attribute is necessary for Parting Shot, as I think the intended behavior can be achieved by adding a condition to the existing attributes for the move:
> 
> ```ts
> new StatusMove(Moves.PARTING_SHOT, Type.DARK, 100, 20, -1, 0, 6)
>     .attr(StatStageChangeAttr, [ Stat.ATK, Stat.SPATK ], -1, false, null, true, true, MoveEffectTrigger.PRE_APPLY)
>     .attr(ForceSwitchOutAttr, true, false)
>     .condition(/* PartingShotCondition with your getCondition logic */)
>     .soundBased(),
> ```
> 
> If the added condition returns `false`, the entire move fails, so in theory checking if the move can change stats through that should produce the desired behavior.

The problem is that adding `.conditions()` would make the move fail (instead of going though and doing all the ability checks, etc.). For example: if we were to have the move auto-fail when the enemy had clear body, then there would be no ShowAbilityPhase to give the user feedback that the enemy blocked the stat lowering effect with clear body.

innerthunder:
> The problem is that adding `.conditions()` would make the move fail (instead of going though and doing all the ability checks, etc.). For example: if we were to have the move auto-fail when the enemy had clear body, then there would be no ShowAbilityPhase to give the user feedback that the enemy blocked the stat lowering effect with clear body. The same would happen for things like Mist as well.

You probably would have to apply `ProtectStatAbAttr` within the condition instead of using `hasAbilityWithAttr`, then. It'd probably lead to some redundant checks, but at least it'd be easier to get the right fail/cancel messages...

My main concern is I just want to avoid nested attributes (like the `StatStageChangeAttr` in your new attribute) when possible. When it comes to the inner attribute, you're either foregoing its conditions and other functions (how does the enemy AI make sense of the inner attribute?), or you're writing a lot of duplicate code.

I know other move attributes are guilty of using nesting (I'm working on rewriting charging moves partially for this reason lol), but it's definitely a pattern that shouldn't be continued.

DustinLin:
I think a root issue here is that we can't extend multiple classes. Here parting shot can be thought of as having ForceSwitch and StatChange-ing attributes, but since it can only inherit 1 of them the other is just a instance variable. If `PartingShotAttr` can `extend` both `ForceSwitchOutAttr` and `StatStageChangeAttr` then we can reuse the different `.apply()`'s implemented. This is what the current solution does anyways (without the proper inheritance).

According to typescript, [mixins](https://www.typescriptlang.org/docs/handbook/mixins.html) are the way to resolve wanting to inherit multiple classes, but I haven't seen any thing like this in the codebase so far.

DayKev:
What we need is to copy the `conditionalAttr()` functionality from the `Ability` class to the `Move` class and then use that for the `StatStageChangeAttr`.

SirzBenjie:
Unfortunately, the codebase has moved on too far from the codebase and porting the PR would be nontrivial. If there's interest in revisiting it, then reopen it. For now, I'm closing.

PR #3590, opened 2024-08-17
[DOCS] adding JSDocs to `arena.ts`

DustinLin:
## What are the changes?

nothing

was looking through these functions and noticed there were no docs

## What did change?

some docs/comments were added

## Checklist


src/field/pokemon.ts
```diff
@@ -4222,7 +4222,7 @@ export interface AttackMoveResult {
 }
 
 export class PokemonSummonData {
-  public battleStats: integer[] = [ 0, 0, 0, 0, 0, 0, 0 ];
+  public battleStats: integer[] = [ 0, 0, 0, 0, 0, 0, 0 ]; // [Atk, Def, SpAtk, SpDef, Spd, Acc]
```
DayKev, line 4225:
```suggestion
  public battleStats: integer[] = [ 0, 0, 0, 0, 0, 0, 0 ]; // [Atk, Def, SpAtk, SpDef, Spd, Acc, Eva]
```
Not 100% sure on it being accuracy and then evasion but regardless, it's missing.

DustinLin: looking at `data/battle-stat.ts` seems to support your guess

frutescens (dismissed):

src/field/arena.ts
```diff
@@ -301,8 +301,8 @@ export class Arena {
 
   /**
    * Attempts to set a new weather to the battle
-   * @param weather new weather to set of type WeatherType
-   * @param hasPokemonSource is the new weather from a pokemon
+   * @param weather {@linkcode WeatherType} new weather to set of type WeatherType
+   * @param hasPokemonSource boolean if the new weather is from a pokemon
```
flx-sta, line 305:
```suggestion
   * @param hasPokemonSource true if the new weather is from a pokemon
```

you don’t need to put types into JS doc as we are using ts. This is just redundant

src/field/arena.ts
```diff
@@ -301,8 +301,8 @@ export class Arena {
 
   /**
    * Attempts to set a new weather to the battle
-   * @param weather new weather to set of type WeatherType
-   * @param hasPokemonSource is the new weather from a pokemon
+   * @param weather {@linkcode WeatherType} new weather to set of type WeatherType
```
flx-sta, line 304:
```suggestion
   * @param weather {@linkcode WeatherType} new {@linkcode WeatherType} to set
```
you don’t need to put types into JS doc as we are using ts. This is just redundant

src/field/arena.ts
```diff
@@ -289,8 +289,8 @@ export class Arena {
 
   /**
    * Sets weather to the override specified in overrides.ts
-   * @param weather new weather to set of type WeatherType
-   * @returns true to force trySetWeather to return true
+   * @param weather {@linkcode WeatherType} new weather to set of type WeatherType
+   * @returns boolean - true to force trySetWeather to return true
```
flx-sta, line 293:
```suggestion
   * @returns true to force trySetWeather to return true
```
you don’t need to put types into JS doc as we are using ts. This is just redundant

src/field/arena.ts
```diff
@@ -289,8 +289,8 @@ export class Arena {
 
   /**
    * Sets weather to the override specified in overrides.ts
-   * @param weather new weather to set of type WeatherType
-   * @returns true to force trySetWeather to return true
+   * @param weather {@linkcode WeatherType} new weather to set of type WeatherType
```
flx-sta, line 292:
```suggestion
   * @param weather new {@linkcode WeatherType} to set
```
you don’t need to put types into JS doc as we are using ts. This is just redundant

flx-sta (changes requested):
Some minor NITs

src/field/arena.ts
DustinLin: I find them quite useful though for referring to types in vscode, for example in this case vscode gives you a direct link to the `WeatherType` enum, without the linkcode you have to take an extra step to search/find it

screenshot

flx-sta: No offense but there is no extra steps needed. Look above 
`(method) Arena.trySetWeather(weather:  WeatherType, hasPokemonSource: boolean)`

I understand that you mean the internal link but I didn't remove that for the WeatherType. I'm aware of that feature. I just rephrased it to be less redundant

DustinLin: thanks, for clarifying i get you (remove extra text, no the linkcode itself)

flx-sta (dismissed):

DayKev (changes requested):
https://github.com/DustinLin/pokerogue/pull/1

DustinLin:
thanks @DayKev , merged in

flx-sta (approved):

frutescens (approved):

DayKev (approved):

PR #3605, opened 2024-08-17
[Hotfix] End Biome Catch Problems 

frutescens:
Users are unable to catch Pokemon in the End Biome because of a bad conditional + an error in how gameMode.hasChallenge() determines active challenges.

Against user experience. 

gameMode.isFreshStartChallenge() would always return true because gameMode.hasChallenge() would only check for the id of a challenge object and not check if it was activated or not (based on its value). The challenge object regardless of activation would always be present, meaning that isFreshStartChallenge() / hasChallenge() would always return true. This also cleans up the conditional used to block catches for easier reading. 

### Screenshots/Videos
Fresh Start Uncaught Pokemon
Screenshot 2024-08-17 at 2 50 09 PM
Fresh Start Caught Pokemon
Screenshot 2024-08-17 at 2 51 36 PM

MonoGen Challenge Caught Pokemon
https://github.com/user-attachments/assets/d1038b79-1647-41e3-83f1-43cac1a35a15

MonoGen Challenge Uncaught Pokemon
image

## How to test the changes?
- Fresh Start --> End Biome - Player should not be able to catch Pokemon
- Challenges/Classic --> End Biome - Player should be able to catch previously hatched Pokemon + Player should not be able to catch unknown Pokemon
- Endless/Daily/Endless Spliced --> End Biome - Player should not be able to catch Pokemon

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/game-mode.ts
+++ b/src/game-mode.ts
@@ -62,7 +62,7 @@ export class GameMode implements GameModeConfig {
    * @returns true if the game mode has that challenge
    */
   hasChallenge(challenge: Challenges): boolean {
-    return this.challenges.some(c => c.id === challenge);
+    return this.challenges.some(c => c.id === challenge && c.value !== 0);
   }
 
   /**
--- a/src/phases.ts
+++ b/src/phases.ts
@@ -2036,7 +2036,11 @@ export class CommandPhase extends FieldPhase {
       }
       break;
     case Command.BALL:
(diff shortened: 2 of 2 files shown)


frutescens:
Works with Classic

innerthunder (approved):

DayKev (approved):

f-fsantos (approved):

PR #3606, opened 2024-08-17
[Bug] Don't reset turn count or used moves array at the start of a new wave

DayKev:
Moves like Fake Out and Gigaton Hammer can no longer be spammed at the start of every wave.

`battleSummonData` was being reset at the start of every wave, meaning any moves that check whether it's the first turn a Pokémon was sent out (ie: Fake Out) or whether a move can't be used twice in a row (ie: Gigaton Hammer) would think their conditions were erroneously fulfilled whenever a new wave started.

The call to `pokemon.resetBattleSummonData()` was removed from `BattleEndPhase`.

## How to test the changes?
Use overrides to give your pokemon Fake Out and Gigaton Hammer, start a new run, and spam them.

## Checklist


frutescens (approved):
Looks good. 

https://github.com/user-attachments/assets/88117416-1448-4c7b-b9fc-8adb26fa2c4b


https://github.com/user-attachments/assets/155602bc-8e5b-4c2b-863e-e1a1db019ef9

damocleas (approved):
approving this fix on my end if needed for spamming the big moves

PR #3616, opened 2024-08-18
[Bug] Fixed issue with falsy issue within condition to get a stat for IV scanner

Opaque02:
There was a change with the... stat enum that made HP now be 0. This meant that in some cases where the HP enum was being called in a condition for checking top IVs, it would trigger as falsy and cause issues. This PR fixes that

There's a discord thread about this bug [here](https://discord.com/channels/1125469663833370665/1274420818465525760)

Due to a stat change enum, this line of code (which would previously get the highest stat to show for the IV scanner) could now have a shownStat of 0, which is HP:

```if (shownStat) {```

Due to this, if HP is the highest stat, instead of continuing the if statement to get the highest stat, it would skip the if check and go to the next stat. However, since the next stat would be HP, it would fail again, and end up showing ??? for all stats instead of the correct values. It would sometimes show an odd number of ??? if HP wasn't the highest; basically once HP was hit, no new stats would be shown. This means if you have all 3 scanners and HP was 3rd highest, it would show you the first 2 and ??? for the other 4

### Screenshots/Videos
Before: 

image

After:

image

Before: 

image

After:

image

## How to test the changes?
Check the discord thread for some sample session data files you can download and test out

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/ui/battle-message-ui-handler.ts
+++ b/src/ui/battle-message-ui-handler.ts
@@ -226,7 +226,7 @@ export default class BattleMessageUiHandler extends MessageUiHandler {
             highestIv = ivs[s];
           }
         });
-        if (shownStat) {
+        if (shownStat !== null && shownStat !== undefined) {
           shownStats.push(shownStat);
           statsPool.splice(statsPool.indexOf(shownStat), 1);
         }


MokaStitcher (approved):
lgtm

f-fsantos (approved):

PR #3627, opened 2024-08-18
[Enhancement] Make the scrollbar cursor size more natural in the starter select UI

RedstonewolfX:
Improves on the scroll bar graphic by making it resize like a proper scroll bar.

The scroll bar is sized differently, to correctly show the size of the starters list.

I simply wanted the scroll bar to reflect the actual size of the window.
Genesect doesn't need its own page, trust me.

Instead of `page` in `scroll-bar.ts`  representing the number of 9x9 windows shown, it represents the number of rows onscreen. When Gen 5 is shown, 10 rows (9 full rows + 1) are visible, so the scroll bar should take up 9/10 of its rectangle.
Other values using `page` are updated to reflect this new value range.

### Screenshots/Videos
Gen 5 before.
image
Gen 5 after.
image

All starters before.
image
All starters after.
image

## How to test the changes?
Simply open the starter menu (press New Game). The difference will be immediately apparent.
It's most obvious when filtered to show Generation 5.
This does not rely on any `overrides`, and doesn't add any tests, nor should it affect any as it is a purely visual change.

## Checklist
    - How do I check this?
*I didn't run any tests because I cannot see any world in which these 3 lines of code break one, but I can run the tests if required*
- [x] Are the changes visual?

### DIFF
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -2415,7 +2415,7 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
       }
     });
 
-    this.starterSelectScrollBar.setPages(Math.max(Math.ceil(this.filteredStarterContainers.length / 9)), 1);
+    this.starterSelectScrollBar.setPages(Math.max(Math.ceil(this.filteredStarterContainers.length / 9), 1));
     this.starterSelectScrollBar.setPage(0);
 
     // sort
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -2415,7 +2415,7 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
       }
     });
 
-    this.starterSelectScrollBar.setPages(Math.ceil(this.filteredStarterContainers.length / 9));
+    this.starterSelectScrollBar.setPages(Math.max(Math.ceil(this.filteredStarterContainers.length / 9)), 1);
     this.starterSelectScrollBar.setPage(0);
 
     // sort
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -2415,7 +2415,7 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
       }
     });
 
-    this.starterSelectScrollBar.setPages(Math.ceil((this.filteredStarterContainers.length - 81) / 9) + 1);
+    this.starterSelectScrollBar.setPages(Math.ceil(this.filteredStarterContainers.length / 9));
     this.starterSelectScrollBar.setPage(0);
 
     // sort
--- a/src/ui/scroll-bar.ts
+++ b/src/ui/scroll-bar.ts
@@ -29,8 +29,8 @@ export class ScrollBar extends Phaser.GameObjects.Container {
 
   setPages(pages: number): void {
     this.pages = pages;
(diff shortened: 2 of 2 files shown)


KimJeongSun:
I like this update! 👍 
But actually, even though I committed the code, it's not originally my code. (I took it over from @FredeX who created it initially.) 
If it's okay, could you explain the changes in the logic before and after the update?

RedstonewolfX:
Hello! Sorry I didn't reply to this sooner, I wasn't checking because with the new backlog of PRs I didn't expect anyone to get back to this yet

Originally, the `pages` variable in the Scroll Bar was the number of pages worth of Pokémon onscreen (1 = between 1 and 81 Pokémon, 2 = between 82 and 162 Pokémon, etc.)
This changes it to be the number of visible _rows_ of Pokémon, creating a more precise value
Since the current page number is also multiplied by 9, this causes the scroll bar to display as normal without needing any further edits.
image

src/ui/starter-select-ui-handler.ts
```diff
@@ -2415,7 +2415,7 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
       }
     });
 
-    this.starterSelectScrollBar.setPages(Math.ceil((this.filteredStarterContainers.length - 81) / 9) + 1);
+    this.starterSelectScrollBar.setPages(Math.ceil(this.filteredStarterContainers.length / 9));
```
KimJeongSun, line 2418:
```suggestion
    this.starterSelectScrollBar.setPages(Math.max(Math.ceil(this.filteredStarterContainers.length / 9)), 1);
```

Shouldn't we ensure that the minimum value is set to 1 to prevent division by 0 issues in setPages?

KimJeongSun:
Except for the suggestion I mentioned above, the function logic seems to be implemented as you described, so it looks good.  👍 
However, for a double-check, I'll test the contents of the PR this evening and provide another review.

RedstonewolfX:
Merged the commit suggestion, so that error should be gone.
Let me know how the test goes

src/ui/starter-select-ui-handler.ts
```diff
@@ -2415,7 +2415,7 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
       }
     });
 
-    this.starterSelectScrollBar.setPages(Math.ceil((this.filteredStarterContainers.length - 81) / 9) + 1);
+    this.starterSelectScrollBar.setPages(Math.max(Math.ceil(this.filteredStarterContainers.length / 9)), 1);
```
KimJeongSun, line 2418:
```suggestion
    this.starterSelectScrollBar.setPages(Math.max(Math.ceil(this.filteredStarterContainers.length / 9), 1));
```
Sorry, but the parentheses in the code I just suggested were wrong. This is the correct way to fix it.

KimJeongSun: Please be a bit more skeptical of me next time! 😄

KimJeongSun (approved):

DayKev (approved):
Works as advertised.

flx-sta:
@RedstonewolfX 
> There is no overlap with another 
> How do I check this?

By searching the issues and asking around. I can assure you there is no overlap.

> I didn't run any tests because I cannot see any world in which these 3 lines of code break one, but I can run the tests if required

The CI is running them anyway. Technically this checkbox is redundant

flx-sta (approved):
lgtm 🚀 

https://github.com/user-attachments/assets/52e9de70-8cdf-4af0-898f-73788b4d7522

PR #3638, opened 2024-08-19
[Misc] Improve settings behavior: notify users of data loss risk and no reloads unnecessary

Vassiat:
No reload if there are no real changes in settings and warn for data loss in battle (If refused, revert the changes that need to be reloaded).

Small improvement in user experience, as when saving and exiting it also warns about data loss, I think it's fair that data reloading also warns, besides, if it warns, it means it can be reversed.

The `abstract-settings-ui-handler` class is now extended from `message-ui-handler`, the `reloadSettings` variable has changed from an array filtering options with `reloadRequired` to an array that stores the indices of options with `reloadRequired` or null for those that don't have it, integration for messages in settings, and a function that restricts reloading if data loss is possible in battle.

### _Podemos ver al usuario saliendo con los mismos datos que salió y igualmente hacer la "recarga necesaria"_
https://github.com/user-attachments/assets/866a0673-b17c-4900-bc85-e5dd24f34477

## After
https://github.com/user-attachments/assets/8439a90c-dd28-40d8-8377-3abc13e8fc58

https://github.com/user-attachments/assets/3f486a2c-d6b3-4be9-83a4-4c6907fe64d1

## Checklist
- [ ] ~Have I considered writing automated tests for the issue?~
- [x] Are the changes visual?

### DIFF
--- a/src/ui/settings/abstract-settings-ui-handler.ts
+++ b/src/ui/settings/abstract-settings-ui-handler.ts
@@ -2,7 +2,7 @@ import BattleScene from "../../battle-scene";
 import { hasTouchscreen, isMobile } from "../../touch-controls";
 import { TextStyle, addTextObject } from "../text";
 import { Mode } from "../ui";
-import UiHandler from "../ui-handler";
+import MessageUiHandler from "../message-ui-handler";
 import { addWindow } from "../ui-theme";
 import {Button} from "#enums/buttons";
 import {InputsIcons} from "#app/ui/settings/abstract-control-settings-ui-handler.js";
@@ -14,10 +14,11 @@ import i18next from "i18next";
 /**
  * Abstract class for handling UI elements related to settings.
  */
-export default class AbstractSettingsUiHandler extends UiHandler {
+export default class AbstractSettingsUiHandler extends MessageUiHandler {
   private settingsContainer: Phaser.GameObjects.Container;
   private optionsContainer: Phaser.GameObjects.Container;
   private navigationContainer: NavigationMenu;
+  private settingsMessageBoxContainer: Phaser.GameObjects.Container;
 
   private scrollCursor: integer;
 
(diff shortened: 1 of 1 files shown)


src/ui/settings/abstract-settings-ui-handler.ts
```diff
@@ -32,7 +33,7 @@ export default class AbstractSettingsUiHandler extends UiHandler {
 
   private cursorObj: Phaser.GameObjects.NineSlice | null;
 
-  private reloadSettings: Array<Setting>;
+  private reloadSettings: Array<integer | null>;
```
torranx, line 36: moving forward, we want to use `number` instead of `integer`

src/ui/settings/abstract-settings-ui-handler.ts
```diff
@@ -139,6 +139,22 @@ export default class AbstractSettingsUiHandler extends UiHandler {
...
     this.settingsContainer.add(cancelText);
 
+    this.settingsMessageBoxContainer = this.scene.add.container(0, 130);
+    this.settingsMessageBoxContainer.setName("settings-message-box");
+    this.settingsMessageBoxContainer.setVisible(false);
+
+    const settingsMessageBox = addWindow(this.scene, 0, 0, (this.scene.game.canvas.width/6) - 2, 48);
+    this.settingsMessageBoxContainer.add(settingsMessageBox);
+
+    const messageText = addTextObject(this.scene,8,8,"",TextStyle.WINDOW,{ maxLines: 2 });
```
torranx, line 149: please properly add spacing after commas

src/ui/settings/abstract-settings-ui-handler.ts
```diff
@@ -209,11 +231,37 @@ export default class AbstractSettingsUiHandler extends UiHandler {
 
     let success = false;
 
+    const progressLosing = () => {
```
torranx, line 234: can we add documentation here?

src/ui/settings/abstract-settings-ui-handler.ts
```diff
@@ -411,4 +462,10 @@ export default class AbstractSettingsUiHandler extends UiHandler {
     }
     this.cursorObj = null;
   }
+
+  showText(text: string, delay?: number|null, callback?: Function, callbackDelay?: number, prompt?: boolean, promptDelay?: number): void {
+    this.settingsMessageBoxContainer.setVisible(!!text);
```
torranx, line 467: nit: code style preference
```suggestion
    this.settingsMessageBoxContainer.setVisible(text.length > 0);
```

src/ui/settings/abstract-settings-ui-handler.ts
```diff
@@ -14,10 +14,11 @@ import i18next from "i18next";
 /**
  * Abstract class for handling UI elements related to settings.
  */
-export default class AbstractSettingsUiHandler extends UiHandler {
+export default class AbstractSettingsUiHandler extends MessageUiHandler {
```
torranx, line 17: are there no unintended effects of changing the base class?

Vassiat: As such, it's still coming from UiHandler because MessageUiHandler (which only serves to display text letter by letter) extends AwaitableUiHandler (which only adds button processing for tutorials), which in turn extends UiHandler. It's also using its own showText function, not the one from the UI, so there are no issues.

torranx (dismissed):

SirzBenjie:
Really old and obsolete. If we want this, we can revisit it in another PR that maybe looks at this one.

Issue #3570, opened 2024-08-16
[Bug] Aerilate + population bomb results in only one hit vs ghost types


Fubs:
### Describe the bug

I have a pokemon with the ability Aerilate and the move Population Bomb.

https://bulbapedia.bulbagarden.net/wiki/Aerilate_(Ability)
https://bulbapedia.bulbagarden.net/wiki/Population_Bomb_(move)

Population Bomb is a normal type move that hits 10 times. Aerilate should make Population Bomb become flying type and receive a 20% power boost. However, when attacking a ghost type pokemon, it seems that only the first hit of Population Bomb works correctly. The target will take one hit, and then the move will simply end.

### User data export file

[data.txt](https://github.com/user-attachments/files/16635952/data.txt)

### Session export file

[session.txt](https://github.com/user-attachments/files/16635961/session.txt)

### Expected behavior

A pokemon with the abiility Aerilate using Population Bomb against a ghost type target should result in up to 10 hits. All 10 hits should be considered a flying-type move affected by both STAB and the 20% boost from Aerilate.

### Screenshots / Videos

https://github.com/user-attachments/assets/b9ee14c6-2b78-4371-8e75-534935710064

### What OS did you observe the bug on?

Linux

### Which browser do you use?

Firefox

Snailman11:
Link to Discord's bug report (Aerilate, Population Bomb Bug) https://discord.com/channels/1125469663833370665/1254821254490361989

frutescens:
Should be resolved with the next release. Thank you for reporting this.

PR #3643, opened 2024-08-19
[Refactor][Move][Ability] Fixing up MoveTypeChangeAttr implementation

frutescens:
The user will be able to paralyze ground-type pokemon and unable to paralyze ghost-types if the Pokemon involved has the ability Normalize. If the user with a non-Normalize Move Type Change ability uses a normal type multi-hit move against a ghost user, it will be able to hit ghost types multiple times.

Just a small improvement to code logic. Fixing up issue #3570 

Abilities that change the types of moves are applied when executing status moves.
Replaced StatusMoveTypeImmunity attribute with RespectTypeImmunity attribute. This checks if an opposing Pokemon has a type that is immune to the current type of the move. 

### Screenshots/Videos
https://github.com/user-attachments/assets/f38b6a90-36fc-43cb-ad96-3cc8d8afda14

https://github.com/user-attachments/assets/4553e8fe-4408-49bd-b7d2-e71cb893a627

https://github.com/user-attachments/assets/190b73aa-ab79-486e-9769-8ee74d436b38

### Screenshots/Videos of Aerilate + Population Bomb

https://github.com/user-attachments/assets/8dc5a389-6811-4646-be01-a42a4dfc00f0

## How to test the changes?
Use Thunder Wave with a Pokemon with the ability Normalize --> Ground Types no longer immune, Ghost Types immune
Use Thunder Wave with a Pokemon without the above ability --> Ground Types immune, Ghost Types not immune

## Checklist
- [x] Are the changes visual?


src/field/pokemon.ts
```diff
@@ -1961,8 +1961,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     const cancelled = new Utils.BooleanHolder(false);
     const power = move.calculateBattlePower(source, this);
     const typeless = move.hasAttr(TypelessAttr);
-
-    const typeMultiplier = new Utils.NumberHolder(!typeless && (moveCategory !== MoveCategory.STATUS || move.getAttrs(StatusMoveTypeImmunityAttr).find(attr => types.includes(attr.immuneType)))
+    const typeEffectiveness = new Utils.IntegerHolder(1);
+    applyMoveAttrs(RespectTypeImmunityAttr, source, this, move, typeEffectiveness);
+    const typeMultiplier = new Utils.NumberHolder(!typeless && (moveCategory !== MoveCategory.STATUS || typeEffectiveness.value === 0)
       ? this.getAttackTypeEffectiveness(move, source, false, false)
```
schmidtc1, line 1967: Is this ternary still necessary, if you have moved it to the move attribute instead?

frutescens: Yes, since it checks the value of typeEffectiveness.

src/data/move.ts
```diff
@@ -726,6 +726,8 @@ export default class Move implements Localizable {
    */
   calculateBattlePower(source: Pokemon, target: Pokemon): number {
     if (this.category === MoveCategory.STATUS) {
+      // The type of a status move is usually irrelevant but Thunder Wave is an exception and found in many movesets.
```
schmidtc1, line 729: nit: "and found in many movesets" is not necessary to mention.

src/test/moves/thunder_wave.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  );
+
+  test(
+    "Thunder Wave does not paralyze if the Pokemon is already status-ed",
+    async () => {
+      game.override.enemySpecies(Species.MAGIKARP);
+      await game.startBattle();
+
+      const enemyPokemon: EnemyPokemon = game.scene.getEnemyPokemon()!;
+      enemyPokemon.status = new Status(StatusEffect.BURN);
```
schmidtc1, line 79: is there a reason for using `enemyPokemon.status` instead of `game.override.enemyStatusEffect(StatusEffect.Burn)`

frutescens: Ah, I forgot about that lol.

src/test/moves/thunder_wave.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      game.override.ability(Abilities.NORMALIZE);
+      game.override.enemySpecies(Species.HAUNTER);
+      await game.startBattle();
+
+      const enemyPokemon: EnemyPokemon = game.scene.getEnemyPokemon()!;
+
+      game.doAttack(getMovePosition(game.scene, 0, Moves.THUNDER_WAVE));
+      await game.phaseInterceptor.to(TurnEndPhase);
+
+      expect(enemyPokemon.status).toBeUndefined();
```
schmidtc1, line 118:
```suggestion
      expect(enemyPokemon.status).toBeUndefined();
      game.doAttack(getMovePosition(game.scene, 0, Moves.THUNDER_WAVE));
      await game.phaseInterceptor.to(TurnEndPhase);

      expect(enemyPokemon.status).toBeUndefined();
```
Might be safe to expect status to be undefined before as well. This is applicable to all the other tests as well.

KimJeongSun (approved):
LGTM!

src/test/moves/thunder_wave.test.ts
frutescens: I think that even though this is a relatively small check + only adds a few milliseconds to the test, I think that because these Pokemon do not have items (thanks to earlier work in #3358) + the opponent Pokemon do not have abilities or passives that prevent Paralysis / lead to early recovery from Paralysis, the check is a little extra. I know its only a few ms but I am trying to be more efficient in my unit tests.

schmidtc1 (approved):

src/data/move.ts
```diff
@@ -726,6 +726,8 @@ export default class Move implements Localizable {
    */
   calculateBattlePower(source: Pokemon, target: Pokemon): number {
     if (this.category === MoveCategory.STATUS) {
+      // The type of a status move is usually irrelevant but Thunder Wave is a common exception.
+      applyPreAttackAbAttrs(MoveTypeChangeAttr, source, target, this, new Utils.NumberHolder(1));
```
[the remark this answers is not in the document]
frutescens: Yes, the last argument of applyPreAttackAbAttrs is where battle power is stored. It also changes the type of the move if ability conditions are met, which is why it is placed there. Here, the function is used only for its abilities to change the move's typing, which is why `new Utils.NumberHolder(1)` acts as a temporary address for battle power.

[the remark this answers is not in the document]
frutescens: `export class MoveTypeChangeAttr extends PreAttackAbAttr {
  constructor(
    private newType: Type,
    private powerMultiplier: number,
    private condition?: PokemonAttackCondition
  ) {
    super(true);
  }

  applyPreAttack(pokemon: Pokemon, passive: boolean, defender: Pokemon, move: Move, args: any[]): boolean {
    if (this.condition && this.condition(pokemon, defender, move)) {
      move.type = this.newType;
      if (args[0] && args[0] instanceof Utils.NumberHolder) {
        args[0].value *= this.powerMultiplier;
      }
      return true;
    }

    return false;
  }
}`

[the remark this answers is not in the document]
torranx: my point is status moves will not in any way have power so I don't see why this should be included here? It's counterintuitive and it's adding a completely different logic when it should care about the move's power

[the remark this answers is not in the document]
frutescens: It is used to change the type of the move itself. The ability attribute application also changes the type of the move as well. The use of it here doesn't take into account the BP multiplier produced (as you see, Utils.IntegerHolder(1) goes literally unused). You can contrast it with how actual attacking moves use the BP multiplier just a few lines down.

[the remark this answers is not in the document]
torranx: > It is used to change the type of the move itself. The ability attribute application also changes the type of the move as well.
That doesn't answer my argument, this function is solely used for **calculating move power**. Evident by the fact that we are returning `-1` early when the move is of status move. 

The use of it here doesn't take into account the BP multiplier produced (as you see, Utils.IntegerHolder(1) goes literally unused). You can contrast it with how actual attacking moves use the ability attribute / the resulting BP multiplier just a few lines down.

You said it yourself, it doesnt take into account the bp multiplier so why are you hijacking this function to add a **completely different logic**?

src/test/moves/thunder_wave.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  afterEach(() => {
+    game.phaseInterceptor.restoreOg();
+  });
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override.battleType("single");
+    game.override.starterSpecies(Species.PIKACHU);
+    game.override.moveset([Moves.THUNDER_WAVE]);
+    game.override.enemyMoveset([Moves.SPLASH, Moves.NONE, Moves.NONE, Moves.NONE]);
```
flx-sta, line 34: use chaining pls
```suggestion
    game.override.battleType("single")
      .starterSpecies(Species.PIKACHU)
      .moveset([Moves.THUNDER_WAVE])
      .enemyMoveset([Moves.SPLASH, Moves.NONE, Moves.NONE, Moves.NONE]);
```

src/test/moves/thunder_wave.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    vi.spyOn(allMoves[Moves.THUNDER_WAVE], "accuracy", "get").mockReturnValue(100);
```
flx-sta, line 35: use the `MoveHelper` instead:

```suggestion
    game.move.forceHit()
```
<sub>Something like that. Just check how it's used elsewhere</sub>

src/test/moves/thunder_wave.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      expect(enemyPokemon.status?.effect).not.toBe(StatusEffect.PARALYSIS);
+    },
+    TIMEOUT
+  );
+
+  test(
+    "Thunder Wave's effectiveness is based on its type",
+    async () => {
+      game.override.ability(Abilities.NORMALIZE);
+      game.override.enemySpecies(Species.DIGLETT);
```
flx-sta, line 93: Use Chaining
```suggestion
      game.override.ability(Abilities.NORMALIZE).enemySpecies(Species.DIGLETT);
```

src/test/moves/thunder_wave.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      expect(enemyPokemon.status).toBeUndefined();
+    },
+    TIMEOUT
+  );
+
+  test(
+    "Thunder Wave does not paralyze if the Pokemon is already status-ed",
+    async () => {
+      game.override.enemySpecies(Species.MAGIKARP);
+      game.override.enemyStatusEffect(StatusEffect.BURN);
```
flx-sta, line 76: Use chaining

```suggestion
      game.override.enemySpecies(Species.MAGIKARP).enemyStatusEffect(StatusEffect.BURN);
```

src/test/moves/thunder_wave.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      expect(enemyPokemon.status?.effect).toBe(StatusEffect.PARALYSIS);
+    },
+    TIMEOUT
+  );
+
+  test(
+    "Thunder Wave respects type immunities",
+    async () => {
+      game.override.ability(Abilities.NORMALIZE);
+      game.override.enemySpecies(Species.HAUNTER);
```
flx-sta, line 110: use chaining:
```suggestion
      game.override.ability(Abilities.NORMALIZE).enemySpecies(Species.HAUNTER);
```

src/test/moves/thunder_wave.test.ts
flx-sta, line 1: We are generally using `it()` instead of `test()`.

I'd prefer if you'd use the same

flx-sta (changes requested):

torranx:
I know I've said this to you before but I'll say this again: please don't immediately resolve comments that are actually not resolved/have no consensus made - also to allow others to chime in with what they think.

frutescens:
@torranx I did let you respond, but even after explaining my code and how it was different, you continued to repeat things that were easily disproven with a read of MoveTypeChangeAttr and move.ts' calculateBattlePower()'s code. I didn't want to continue repeating myself to someone who wasn't listening and that 'Resolve Conversation' was me peacefully exiting a pointless discussion. No offense, but it didn't seem like you wanted consensus in the first place.

torranx:
> @torranx I did let you respond, but even after explaining my code and how it was different, you continued to repeat things that were easily disproven with a read of MoveTypeChangeAttr and move.ts' calculateBattlePower()'s code. I didn't want to continue repeating myself to someone who wasn't listening and that 'Resolve Conversation' was me peacefully exiting a pointless discussion. No offense, but it didn't seem like you wanted consensus in the first place.

A mere response does not equate to a comment being resolved. I wasn't repeating things and was explaining to you as well what I meant. It seems to me that you're trying to say you were always right because you already explained your side.

You resolved the comment after your response on this:

> my point is status moves will not in any way have power so I don't see why this should be included here? It's counterintuitive and it's adding a completely different logic when it should only care about the move's power

How come I was repeating myself and didn't want to reach consensus? lol I was literally asking why it should be included in that function. 

If you can't take someone else's feedback and create a healthy conversation with it, that's a you problem. Anyways, I'm leaving this discussion here.

src/data/move.ts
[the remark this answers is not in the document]
flx-sta: @torranx Is this what you were aiming for ?
https://github.com/pagefaultgames/pokerogue/pull/3643/commits/56b1f0e676fa8677562632dca99753e2d57fab94

src/test/moves/thunder_wave.test.ts
```diff
@@ -32,7 +31,7 @@ describe("Moves - Thunder Wave", () => {
       .starterSpecies(Species.PIKACHU)
       .moveset([Moves.THUNDER_WAVE])
       .enemyMoveset([Moves.SPLASH, Moves.NONE, Moves.NONE, Moves.NONE]);
-    game.move.forceHit()
+    game.move.forceHit();
```
flx-sta, line 34: I'm not sure it will work as simple like that. It might have to be called at a different spot. You probably wanna look at other tests that use it

src/data/move.ts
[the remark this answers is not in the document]
torranx: sort of yes, I will let you (and other devs) to decide where it should be placed. I only wanted that to be moved from the function because it is out of scope.

flx-sta (approved):

frutescens:
@torranx I was looking at this bug https://github.com/pagefaultgames/pokerogue/issues/3570 and I was thinking that it might be better to split MoveTypeChangeAbAttr into one attribute that changes the type and another that provides the BP multiplier. Would that interfere with calculateBasePower()?

torranx:
> @torranx I was looking at this bug #3570 and I was thinking that it might be better to split MoveTypeChangeAbAttr into one attribute that changes the type and another that provides the BP multiplier. Would that interfere with calculateBasePower()?

I would think so, yes. It needs to change type and multiply power simultaneously IIRC.

src/phases/move-effect-phase.ts
```diff
@@ -63,6 +63,7 @@ export class MoveEffectPhase extends PokemonPhase {
       }
 
       user.lapseTags(BattlerTagLapseType.MOVE_EFFECT);
+      applyPreAttackAbAttrs(MoveTypeChangeAttr, user, this.getTarget() ?? null, move);
```
innerthunder, line 66: If I'm reading this right, the move will be its changed type when passed into `target.apply` with this here, which I think would cause Aerilate (and similar abilities) to never be able to apply the power boost

innerthunder (changes requested):

src/phases/move-effect-phase.ts
frutescens: Discussed on Discord and fixed with a new conditional.

src/data/ability.ts
```diff
@@ -4915,7 +4915,7 @@ export function initAbilities() {
     new Ability(Abilities.STRONG_JAW, 6)
       .attr(MovePowerBoostAbAttr, (user, target, move) => move.hasFlag(MoveFlags.BITING_MOVE), 1.5),
     new Ability(Abilities.REFRIGERATE, 6)
-      .attr(MoveTypeChangeAttr, Type.ICE, 1.2, (user, target, move) => move.type === Type.NORMAL),
+      .attr(MoveTypeChangeAttr, Type.ICE, 1.2, (user, target, move) => move.defaultType === Type.NORMAL && !move.hasAttr(VariableMoveTypeAttr)),
```
innerthunder, line 4918: Can you change `MoveTypeChangeAttr` to `MoveTypeChangeAbAttr` while you're at it?

frutescens: Will do!

innerthunder (approved):
lgtm, but I think others should re-review before merging

KimJeongSun (approved):

src/data/move.ts
```diff
@@ -1082,13 +1082,16 @@ export class PreMoveMessageAttr extends MoveAttr {
...
-  constructor(immuneType: Type) {
-    super(false);
-
-    this.immuneType = immuneType;
+/**
+ * This allows type-sensitive moves like Thunder Wave to be affected by abilities like Normalize, etc.
+ * In addition, this avoids the use of conditionals and can allow for the different type matchups found in Inverse Mode.
+ */
+export class RespectTypeImmunityAttr extends MoveAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
```
torranx, line 1090: docs on this please, esp the `args`

src/data/move.ts
```diff
@@ -1082,13 +1082,16 @@ export class PreMoveMessageAttr extends MoveAttr {
...
+    //Apply type change to move
+    const typeEffectiveness = target.getAttackTypeEffectiveness(move.type, user);
+    args[0].value = typeEffectiveness;
```
torranx, line 1093:
```suggestion
    args[0].value = target.getAttackTypeEffectiveness(move.type, user);
```

src/field/pokemon.ts
```diff
@@ -1971,8 +1971,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     const cancelled = new Utils.BooleanHolder(false);
     const power = move.calculateBattlePower(source, this);
     const typeless = move.hasAttr(TypelessAttr);
-
-    const typeMultiplier = new Utils.NumberHolder(!typeless && (moveCategory !== MoveCategory.STATUS || move.getAttrs(StatusMoveTypeImmunityAttr).find(attr => types.includes(attr.immuneType)))
+    const typeEffectiveness = new Utils.IntegerHolder(1);
```
torranx, line 1974:
```suggestion
    const typeEffectiveness = new Utils.NumberHolder(1);
```

torranx (commented):
- couple minor comments
- ~not sure if there is no unintended side effect of `MoveTypeChangeAbAttr` because its being called in 2 different places~
- I personally don't like how there are both `typeMultiplier` and `typeEffectiveness` - there is no clear distinction between the two

other than those mentioned above, looks good

src/data/move.ts
```diff
@@ -1082,13 +1082,19 @@ export class PreMoveMessageAttr extends MoveAttr {
...
-    this.immuneType = immuneType;
+/**
+ * This allows type-immunity sensitive moves like Thunder Wave to be affected by abilities like Normalize, etc.
+ * In addition, this avoids the use of conditionals and can allow for the different type matchups found in Inverse Mode.
+ * @param user {@linkcode Pokemon} the user of the move
+ * @param target {@linkcode Pokemon} the target of the move used
+ * @param move {@linkcode Move} the move used
+ * @param args[0] the destination of the calculated type multiplier of a match-up between the target's types and the move's type
+ * @returns true
+ */
```
torranx, line 1093: I think this should be right above `apply()`, at least the params and returns

frutescens:
Closing since #3704 will fix things.

PR #3647, opened 2024-08-19
[Enhancement] Delete the existing starter preferences when new save data is imported

MokaStitcher:
When importing a save file, the local starter preferences for shiny variant, form, nickname, etc. are deleted.

At the moment the settings for preferred gender, form, shiny, variant, nickname and favorite status for starters are saved locally and not on the server. This means that when importing a different save data those preferences were kept, resulting in Pokemon potentially showing a shiny variant or form that is not actually unlocked, or the favorite ribbon and nickname that was set previously.
This is only useful for beta/local play where importing save data is possible, but would really help in preventing weird behavior when testing things out for the starter select screen

Creates a delete() method for StarterPrefs and calls it after importing new game data (not when importing session or other type of Game Data)
Note: I added a safeguard to make sure that we are in local or beta, but technically it shouldn't be necessary. let me know if I should keep or remove it

### Screenshots/Videos
Before: Nidoran has a nickname, Caterpie and Vulpix are set to show shiny tier 1 and 2 respectively
__before_1

Before, but after importing a different save file: Caterpie and Vulpix show variant colors that have not been actually unlocked for this new save, Nidoran still has its nickname
__before_2

After: Nidoran has a nickname, Vulpix and Rattata are set to show shiny tier 1 and 2 respectively, Rattata is a "favorite" Pokemon
_after_1

After importing a different save file: Nidoran's nickname, Vulpix and Rattata shiny and favorite status are all back to default
_after_2

## How to test the changes?
- If you don't want to mess with your local data, do all this in a new incognito window
- It's best to test with save data with a good amount of things unlocked like [this one](https://discord.com/channels/1125469663833370665/1267339517371875361/1275142522603180104)
- Play around with the starter preferences for nickname, favorite, form, gender, shiny and variant (I'm not exactly sure when they are saved right now but if you start a run after changing stuff it will be saved for sure).
- Go to Main Menu > Manage Data > Import Data and import a different save file like [this one](https://discord.com/channels/1125469663833370665/1267339517371875361/1275141551957217384)
- After it's done reloading, check that the preferences you set previously are gone
- If you want you can also double check that this does not happen when importing Session Data

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/system/game-data.ts
+++ b/src/system/game-data.ts
@@ -233,6 +233,10 @@ export class StarterPrefs {
       localStorage.setItem(`starterPrefs_${loggedInUser?.username}`, pStr);
     }
   }
+
+  static delete(): void {
+    localStorage.removeItem(`starterPrefs_${loggedInUser?.username}`);
+  }
 }
 
 export interface StarterDataEntry {
@@ -1315,6 +1319,11 @@ export class GameData {
             this.scene.ui.showText(`Your ${dataName} data will be overridden and the page will reload. Proceed?`, null, () => {
               this.scene.ui.setOverlayMode(Mode.CONFIRM, () => {
                 localStorage.setItem(dataKey, encrypt(dataStr, bypassLogin));
+
+                // If we imported main game data, delete the existing starter preferences
+                if (dataType === GameDataType.SYSTEM && (Utils.isLocal || Utils.isBeta)) {
+                  StarterPrefs.delete();
+                }
 
                 if (!bypassLogin && dataType < GameDataType.SETTINGS) {
                   updateUserInfo().then(success => {


KimJeongSun (approved):
LGTM!

AsdarDevelops:
A question that probably stems from ignorance:

Wouldn't it be better if instead, the exporting functionality fetched your preferences and attached them to the save data, and then the importing functionality would override earlier preferences with the ones from the new imported data?

I playtest a bunch in our Mystery Encounters beta (using my live save data) and this change, if I understand correctly, would wipe the preferences clean, thus making me have to setup the team anew anytime I want to start a new run to test for new MEs balance.

MokaStitcher:
closing PR as no longer necessary since the fixes in PR #3870 address this as a side effect

PR #3654, opened 2024-08-19
[Refactor][Misc] Languages in one array for no repeat key/label

Vassiat:
The user will no longer see the active language in the list of available languages.

I was trying to find another issue and it became difficult because the language keys were located in both `setting-display-ui-handler` and `system/settings`. I saw that I could reduce some lines and improve the implementation of new languages, so I did it.

I also added the functionality to remove the current language from the available options for the user, as it doesn't make much sense to choose the selected one, right? I didn't want to make another PR just for this, so I included it here, but I can remove it since it's not the main goal of this PR.

The language logic will be removed from `setting-display-ui-handler`, making it a bit less complex to add a new language.

## Checklist
- [ ] ~Have I considered writing automated tests for the issue?~
- [ ] ~If I have text, did I add placeholders for them in locales?~
- [x] Are the changes visual?

### DIFF
--- a/src/system/settings/settings.ts
+++ b/src/system/settings/settings.ts
@@ -79,6 +79,67 @@ export interface Setting {
   isHidden?: () => boolean
 }
 
+type LanguageType = {
+  key:string|`${string}-${string}`;
+  label:string;
+  handler?:()=>{};
+  value?:string;
+};
+
+const languages: Array<LanguageType> = [
+  {
+    key:"en",
+    label: "English",
+  },
+  {
+    key:"es",
+    label: "Español",
+  },
+  {
+    key:"it",
+    label: "Italiano",
+  },
+  {
+    key:"fr",
+    label: "Français",
+  },
+  {
+    key:"de",
+    label: "Deutsch",
+  },
+  {
+    key:"pt-BR",
+    label: "Português (BR)",
+  },
+  {
+    key:"zh-CN",
+    label: "简体中文",
+  },
+  {
+    key:"zh-TW",
+    label: "繁體中文",
+  },
+  {
+    key:"ko",
+    label: "한국어",
+  },
+  {
+    key:"ko-KR",
+    label: "한국어",
+  },
+  {
+    key:"ja",
+    label: "日本語",
+  },
+  // {
+  //   key:"ca-ES",
+  //   label: "Català",
+  // },
+];
+
+// default is English
+const currentLanguage = languages.find((locale) => i18next.resolvedLanguage! === locale.key)?.label ?? "English";
+
 /**
  * Setting Keys for existing settings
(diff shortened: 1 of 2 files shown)


flx-sta:
lgtm but could you add some test-coverage for this?
Basically spying on the output of array that is being displayed

Vassiat:
About a possible integration of dynamic imports to avoid having to manually integrate different languages every time. However, I would need to change the TypeScript configuration for the "module" option, as TypeScript complains about having an `await` at the top-level with the current configuration (despite that it works with the current configuration). After testing, it seems that using "ES2022" works, but I'm not sure what kind of compatibility effects it will have.

the current implementation:

``` ts
import { caEsConfig } from "#app/locales/ca_ES/config.js";
import { deConfig } from "#app/locales/de/config.js";
import { enConfig } from "#app/locales/en/config.js";
import { esConfig } from "#app/locales/es/config.js";
import { frConfig } from "#app/locales/fr/config.js";
import { itConfig } from "#app/locales/it/config.js";
import { koConfig } from "#app/locales/ko/config.js";
import { jaConfig } from "#app/locales/ja/config.js";
import { ptBrConfig } from "#app/locales/pt_BR/config.js";
import { zhCnConfig } from "#app/locales/zh_CN/config.js";
import { zhTwConfig } from "#app/locales/zh_TW/config.js";

// ...
i18next.init({
  resources: {
    // ...
        en: {
          ...enConfig
        },
        es: {
          ...esConfig
        },
        fr: {
          ...frConfig
        },
        it: {
          ...itConfig
        },
        de: {
          ...deConfig
        },
        "pt-BR": {
          ...ptBrConfig
        },
        "zh-CN": {
          ...zhCnConfig
        },
        "zh-TW": {
          ...zhTwConfig
        },
        ko: {
          ...koConfig
        },
        ja: {
          ...jaConfig
        },
        "ca-ES": {
          ...caEsConfig
        }
      },
    // ...
});
```

And the proposed implementation:

``` ts
const resources = {};

const modules = languagesKeys.map((lang) => {
  return import(`../locales/${lang.replace("-", "_")}/config`).then((module) => {
    resources[lang] = module.default;
  });
});

await Promise.all(modules);

// ...
i18next.init({
  // ...
  resources: resources,
  // ...
});
```

Vassiat:
> lgtm but could you add some test-coverage for this? Basically spying on the output of array that is being displayed

I'm still learning about testing and I'm not sure if I've implemented it correctly. can you please review my test for this refactoring to ensure it's correct?

SirzBenjie:
This is really good and we probably want to port this to the newer state of the codebase. But the codebase has changed substantially at this point, and this should probably be reopened as a new PR. Apologies, @Vassiat

PR #3665, opened 2024-08-20
[Bug] Fix off-by-one errors in some random number calls

DayKev:
Confusion will last 2-5 turns now instead of 2-4, trapping moves like Fire Spin will last 4-5 turns now instead of only 4, etc.
Basically, every move/effect that had a min-max turn range will actually last between the minimum and maximum amount of turns now instead of between min and max - 1.
Also fixes a bug where a move with 99 accuracy couldn't miss, as if it had 100 accuracy.

There was some usage of the random number functions that was incorrect.

Some instances of `randSeedInt(range, min)` are replaced with `randSeedIntRange(min, max)` when they were intended to be random between min and max inclusive and min was a number greater than zero.
Some tsdocs are added to the various `randSeedInt()` functions as well.

## How to test the changes?
Liberal use of the browser devtools console and/or `console.log()` I guess?

## Checklist


src/battle-scene.ts
```diff
@@ -957,6 +957,7 @@ export default class BattleScene extends SceneBase {
     this.offsetGym = this.gameMode.isClassic && this.getGeneratedOffsetGym();
   }
 
+  // What is the purpose of this function? Why not just use Battle.randSeedInt() directly?
   randBattleSeedInt(range: integer, min: integer = 0): integer {
     return this.currentBattle?.randSeedInt(this, range, min);
   }
```
DayKev, line 963: See code comment.

src/battle-scene.ts
```diff
@@ -2550,7 +2552,7 @@ export default class BattleScene extends SceneBase {
         if (mods.length < 1) {
           return mods;
         }
-        const rand = Math.floor(Utils.randSeedInt(mods.length));
+        const rand = Math.floor(Utils.randSeedInt(mods.length)); // Why is there a floor() call around a number that can only be an integer?
```
DayKev, line 2555: See code comment.

src/data/daily-run.ts
```diff
@@ -41,7 +41,7 @@ export function getDailyRunStarters(scene: BattleScene, seed: string): Starter[]
 
     const starterCosts: integer[] = [];
     starterCosts.push(Math.round(3.5 + Math.abs(Utils.randSeedGauss(1))));
-    starterCosts.push(Utils.randSeedInt(9 - starterCosts[0], 1));
+    starterCosts.push(Utils.randSeedInt(9 - starterCosts[0], 1)); // RNG: 1 to (5? 6?)
```
DayKev, line 44: So... how exactly does this work?

src/modifier/modifier-type.ts
```diff
@@ -979,7 +979,7 @@ class SpeciesStatBoosterModifierTypeGenerator extends ModifierTypeGenerator {
       }
 
       if (totalWeight !== 0) {
-        const randInt = Utils.randSeedInt(totalWeight, 1);
+        const randInt = Utils.randSeedInt(totalWeight, 1); // Why is the minimum 1?
```
DayKev, line 982: See code comment.

src/utils.ts
```diff
@@ -82,6 +82,7 @@ export function randInt(range: integer, min: integer = 0): integer {
   return Math.floor(Math.random() * range) + min;
 }
 
+// Is this only seeded if called via `Battle:randSeedInt()`?
 export function randSeedInt(range: integer, min: integer = 0): integer {
```
DayKev, line 86: See code comment.

src/data/daily-run.ts
DayKev, line 44:
```suggestion
    starterCosts.push(Utils.randSeedInt(9 - starterCosts[0], 1));
```

src/utils.ts
```diff
@@ -82,6 +82,7 @@ export function randInt(range: integer, min: integer = 0): integer {
   return Math.floor(Math.random() * range) + min;
 }
 
+// Is this only seeded if called via `Battle:randSeedInt()`?
```
DayKev, line 85: See code comment, wanna make sure I'm not misunderstanding this.

src/utils.ts
```diff
@@ -82,12 +82,13 @@ export function randInt(range: integer, min: integer = 0): integer {
...
 export function randSeedInt(range: integer, min: integer = 0): integer {
   if (range <= 1) {
     return min;
   }
   return Phaser.Math.RND.integerInRange(min, (range - 1) + min);
-}
+} //
```
DayKev, line 91: See code comment, wanna make sure I'm not misunderstanding this.

src/battle-scene.ts
PigeonBar: Probably just for convenience, since the `currentBattle?.randSeedInt()` function requires passing a `scene` argument.

src/battle-scene.ts
PigeonBar: I think it should be safe to remove the `floor()`.

src/utils.ts
PigeonBar: My understanding is that there is global RNG that gets seeded via `Scene:setSeed()` and `Scene:resetSeed()`. It generally gets seeded at the beginning of every wave based on the wave number and the run's stored seed. This global RNG is used when `Utils.randSeedInt()` is _not_ called via `Battle:randSeedInt()`, and it is used for stuff like generating wild Pokemon, shop item choices, and some egg attributes. Meanwhile, each `Battle` keeps track of its own RNG seed (which is generated using the global RNG seed), and calling `Battle:randSeedInt()` temporarily uses the battle's RNG seed. This RNG is used for most in-battle moves, Poke Ball catch attempts, etc..

DayKev: So calling `Utils.randSeedInt()` directly uses the global seed, calling `Battle:randSeedInt()` (or the passthrough function `BattleScene:randBattleSeedInt()`) calls `Utils.randSeedInt()` but causes it to use the current battle's local seed instead of the global seed for that call.

src/battle-scene.ts
```diff
@@ -2611,7 +2613,7 @@ export default class BattleScene extends SceneBase {
...
```
DayKev, line 2616:
```suggestion
        const rand = Utils.randSeedInt(mods.length);
```

src/battle-scene.ts
DayKev: Another thing to be fixed by making a global `scene` variable...

src/battle-scene.ts
```diff
@@ -960,6 +960,7 @@ export default class BattleScene extends SceneBase {
     this.offsetGym = this.gameMode.isClassic && this.getGeneratedOffsetGym();
   }
 
+  // What is the purpose of this function? Why not just use Battle.randSeedInt() directly?
```
DayKev, line 963:
```suggestion
```

frutescens (approved):

src/battle-scene.ts
```diff
@@ -855,7 +855,7 @@ export default class BattleScene extends SceneBase {
     overrideModifiers(this, false);
     overrideHeldItems(this, pokemon, false);
     if (boss && !dataSource) {
-      const secondaryIvs = Utils.getIvsFromId(Utils.randSeedInt(4294967295));
+      const secondaryIvs = Utils.getIvsFromId(Utils.randSeedInt(4294967296));
```
Tempo-anon, line 858: 4294967296 is 2^32 in case we wanted to document that in a comment or something.

src/battle-scene.ts
```diff
@@ -1112,7 +1122,8 @@ export default class BattleScene extends SceneBase {
             doubleTrainer = false;
           }
         }
-        newTrainer = trainerData !== undefined ? trainerData.toTrainer(this) : new Trainer(this, trainerType, doubleTrainer ? TrainerVariant.DOUBLE : Utils.randSeedInt(2) ? TrainerVariant.FEMALE : TrainerVariant.DEFAULT);
+        const variant = doubleTrainer ? TrainerVariant.DOUBLE : (Utils.randSeedInt(2) ? TrainerVariant.FEMALE : TrainerVariant.DEFAULT);
```
Tempo-anon, line 1125: nit: grrrrrr chained ternaries...

src/battle-scene.ts
```diff
@@ -1112,7 +1122,8 @@ export default class BattleScene extends SceneBase {
...
+        newTrainer = trainerData !== undefined ? trainerData.toTrainer(this) : new Trainer(this, trainerType, variant);
```
Tempo-anon, line 1126: nit: once again going to suggest putting () to make it clearer
`newTrainer = (trainerData !== undefined ? trainerData.toTrainer(this) : new Trainer(this, trainerType, variant));`

Tempo-anon (approved):

src/battle-scene.ts
DayKev: Maybe I should just replace it directly with `2^32`?

src/battle-scene.ts
DayKev:
```ts
        let variant: TrainerVariant;
        if (doubleTrainer) {
          variant = TrainerVariant.DOUBLE;
        } else {
          if (Utils.randSeedInt(2)) {
            variant = TrainerVariant.FEMALE;
          } else {
            variant = TrainerVariant.DEFAULT;
          }
        }
```
🤔

src/battle-scene.ts
Tempo-anon: Then it'd have to make the calculation every time wouldn't it? Slight optimization to have the number already precomputed

src/battle-scene.ts
Tempo-anon: Looks neater. `if (Utils.randSeedInt(2))` is still something cursed though. Maybe `Utils.randSeedInt(2) === 1` and then a comment or something to explain it's 50%

torranx (dismissed):

innerthunder (approved):

torranx (approved):

frutescens (approved):

PR #3678, opened 2024-08-21
[Test] Add `forceEnemyMove` Game Manager util

innerthunder:

No user-facing changes; this just adds a tool to help with integration testing

Unit tests at the moment are very limited in terms of controlling the enemy's actions. The vast majority of tests rely on using overrides to fill enemy movesets with four copies of the same move to force that move, and if the enemy needs to use more than one unique move during a test, moveset overrides would have to be set with every turn. The limitations from this lack of control are especially apparent, though, in tests with double battles:
- If the enemy is using a single-target move, there is currently no way to force the enemy to target a specific Pokemon on the field.
- Forcing moves via overrides is impractical when enemies in a double battle need to use different moves in one turn.

The `forceEnemyMove` util gives the test writer much more control in these situations while also cutting down on the enemy moveset overrides needed to carry out complex test cases.

- `test/utils/gameManager`:
  - New `forceEnemyMove` util function forces the next enemy selecting a move to use a specific move. This also includes an optional `target` argument to force the enemy to use the given move against the given target (if the move is single-targeted).
  - Renamed `forceOpponentToSwitch` to `forceEnemyToSwitch` to match conventions with gameManager's enemy override functions as well as `forceEnemyMove`.
- `phases/enemy-command-phase`: Added getter for `fieldIndex` in support of `forceEnemyMove`
- TODO: update unit tests to incorporate `forceEnemyMove` and reduce calls to `override.enemyMoveset`

### Screenshots/Videos

N/A

## How to test the changes?

`npm run test`

## Checklist
- [x] Have I considered writing automated tests for the issue? (...)
- [n/a] If I have text, did I add placeholders for them in locales?


src/test/utils/gameManager.ts
```diff
@@ -42,6 +42,9 @@ import { TitlePhase } from "#app/phases/title-phase.js";
 import { TurnEndPhase } from "#app/phases/turn-end-phase.js";
 import { TurnInitPhase } from "#app/phases/turn-init-phase.js";
 import { TurnStartPhase } from "#app/phases/turn-start-phase.js";
+import { Moves } from "#app/enums/moves.js";
+import { EnemyCommandPhase } from "#app/phases/enemy-command-phase.js";
+import { getMoveTargets } from "#app/data/move.js";
```
flx-sta, line 47:
```suggestion
import { Moves } from "#app/enums/moves";
import { EnemyCommandPhase } from "#app/phases/enemy-command-phase";
import { getMoveTargets } from "#app/data/move";
```
💀

src/test/utils/gameManager.ts
```diff
@@ -247,7 +250,27 @@ export default class GameManager {
...
+  /**
+   * Forces the next enemy selecting a move to use the given move in its moveset against the
+   * given target (if applicable).
+   * @param moveId {@linkcode Moves} the move the enemy will use
+   * @param target {@linkcode BattlerIndex} the target on which the enemy will use the given move
+   */
+  async forceEnemyMove(moveId: Moves, target?: BattlerIndex) {
+    // Wait for the next EnemyCommandPhase to start
+    await this.phaseInterceptor.to(EnemyCommandPhase, false);
+    const enemy = this.scene.getEnemyField()[(this.scene.getCurrentPhase() as EnemyCommandPhase).getFieldIndex()];
```
flx-sta, line 262: I'm not a fan of this `this.scene.getCurrentPhase() as EnemyCommandPhase`. Instead the method should be re-written:
```ts
  /* Phase Functions */
  getCurrentPhase<P extends Phase = Phase>(): P | null {
    return this.currentPhase ? this.currentPhase as P : null;
  }
```

So it can be called like this:
```ts
this.scene.getCurrentPhase<EnemyCommandPhase>();
```

src/phases/enemy-command-phase.ts
```diff
@@ -83,4 +83,8 @@ export class EnemyCommandPhase extends FieldPhase {
 
     this.end();
   }
+
+  getFieldIndex(): integer {
```
flx-sta, line 87:
```suggestion
  getFieldIndex(): number {
```

flx-sta (changes requested):

src/test/utils/gameManager.ts
innerthunder: Is this PR the place to change this? This pattern's used a lot within the codebase...

flx-sta: > The PR is self-contained and cannot be split into smaller PRs?

I didn't know the extend of that change. All good

DayKev (approved):

Tempo-anon (approved):

frutescens (approved):

PR #3681, opened 2024-08-21
[Bug] Fix Eviolite Weight Condition

xsn34kzx:

`Eviolite` should no longer get pushed into reward rolls when an unevolved Pokemon already has one.

`Eviolite` was never intended to have a weight above 0 if any unevolved Pokemon in the party already had one.

Changed the predicate of the last `some` function done on a given Pokemon's held items to check each item's `ModifierType`'s `id` directly instead of using `instanceof` (which, for whatever reason, failed regardless of the modifier thrown at it).

Worth noting that this may also apply to `Toxic Orb` and `Flame Orb`'s weight functions since they both also rely on `instanceof` for recognizing if a Pokemon already has either.

### Screenshots/Videos

Note that for both videos, session data provided on Discord was used that guaranteed `Eviolite` as a reward with only one unevolved Pokemon (`Galarian Linoone`) holding `Eviolite` already. As such, it's the only possible Pokemon to trigger the original weight function in both cases.

#### Before

https://github.com/user-attachments/assets/1a53b29f-bd84-4a5f-bbf0-583e73ee42a3

#### After

https://github.com/user-attachments/assets/43d8d271-1bf4-4f98-8c80-74e53d8c4969

## How to test the changes?

The only natural/ideal way to test is essentially to play with an unevolved Pokemon, give it `Eviolite`, and see if `Eviolite` ever gets rolled.

## Checklist
- [x] Are the changes visual?


flx-sta:
@xsn34kzx Mind splitting this whole condition into more readable sections too? It looks quite cluttery

flx-sta (changes requested):
code looks good but is lacking test coverage 💀

Snailman11:
Link to Discord's Bug report (Eviolite in item pool even though only unevolved pokemon in party already has one)
https://discord.com/channels/1125469663833370665/1274532840666697860

frutescens (approved):
Could a test be possibly implemented? I feel like Eviolite not appearing in shop pools is similar to that phrase '"If a tree falls in a forest and no one is around to hear it, does it make a sound?"

xsn34kzx:
@frutescens @flx-sta Following up on the test thing, it's funny because I wrote a test to check if it worked, and it did; however, I reverted the actual fix here, and it stilled passed (which I guess makes sense since `instanceof` as is *should* work without this fix). So this bug is in a weird state of limbo for me as far as what the actual issue is. I have more info on Discord on the night I tried it if we wanna deliberate on it more there.

xsn34kzx:
I can't really manually test the fix on the original save (given the time passing and changes since that save was brought in), but I provide a bit more reason as to why testing was "impossible" at the time. The issue can more than likely be traced to the fact that `instanceof` checks the prototype chain of given class, which may not be preserved for whatever reason when an item is loaded back into the game. This would explain why unit testing whether the item would spawn or not was not reproducing the bug, since that item was usually "freshly" generated and added as an item in the same session. Overall though, I think using the `id` is much more "foolproof" just by nature of it being consistent and persistent (we need to generate the item in the first place) instead of relying on `instanceof` directly.

DayKev:
Btw there are some new item pool test functions you can use which should hypothetically be accurate, check out the Daily Mode test and `src/test/utils/helpers/modifiersHelper.ts`.

xsn34kzx:
Testing wasn't necessary the issue itself back when I wrote a test, it was the fact that both the original condition (seen in the first video) and the new condition (the one in the second video) both passed checking the weight when all Pokemon in the party have Eviolite, which is why I think it's a load issue with making sure those modifier classes have the correct prototype chain for `instanceof` to work. Basically, I think it relies pretty much on how modifiers are built on load.

DayKev (approved):

frutescens (approved):

PR #3708, opened 2024-08-23
[Bug] Fixes Unseen Fist ignoring abilities

schmidtc1:

Unseen Fist + direct contact moves allow the user to ignore abilities (e.g. Sturdy)

It's a bug
Discord report: https://discord.com/channels/1125469663833370665/1274675083872436235

Adds a break in the switch case, so that it doesn't fall through to `IGNORE_PROTECT` when checking if a move should ignore abilities.

### Screenshots/Videos

https://github.com/user-attachments/assets/3307af07-dd60-4cf6-accf-c57dd0a7c26b

## How to test the changes?

Give pokemon Unseen Fist + direct contact move. Give opponent Sturdy, see that it no longer KOs through Sturdy

## Checklist
- [n/a] If I have text, did I add placeholders for them in locales?
- [x] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -595,6 +595,7 @@ export default class Move implements Localizable {
           return true;
         }
       }
+      break;
     case MoveFlags.IGNORE_PROTECT:
       if (user.hasAbilityWithAttr(IgnoreProtectOnContactAbAttr) &&
           this.checkFlag(MoveFlags.MAKES_CONTACT, user, target)) {
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -601,6 +601,7 @@ export default class Move implements Localizable {
           this.checkFlag(MoveFlags.MAKES_CONTACT, user, target)) {
         return true;
       }
+      break;
     }
 
     return !!(this.flags & flag);
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -601,7 +601,6 @@ export default class Move implements Localizable {
           this.checkFlag(MoveFlags.MAKES_CONTACT, user, target)) {
         return true;
       }
-      break;
     }
 
     return !!(this.flags & flag);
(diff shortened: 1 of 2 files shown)


innerthunder (dismissed):

src/data/move.ts
```diff
@@ -595,6 +595,7 @@ export default class Move implements Localizable {
           return true;
         }
       }
+      break;
     case MoveFlags.IGNORE_PROTECT:
```
torranx, line 599: shouldn't this have `break` too?

innerthunder: It's the last case in the switch block, so it doesn't *really* matter, but it'd be nice to have if people (like me) implement new cases and forget to add the break later.

schmidtc1: Yea, I'll add it to be safe

schmidtc1:
Accidentally pushed my overrides in https://github.com/schmidtc1/pokerogue/commit/2589faf69edd67883272761ab3ddf59a54e8c013. 
Reverted and should be fixed

torranx (approved):

innerthunder (approved):

frutescens (commented):
Tested and works. Does this change also apply to Pokemon with Fluffy? I tried testing it but it was hard to tell. 

https://github.com/user-attachments/assets/79f7d038-fce8-49a6-afe8-8524ff5900ef

schmidtc1:
> Tested and works. Does this change also apply to Pokemon with Fluffy? I tried testing it but it was hard to tell.
> 
>  Screen.Recording.2024-08-27.at.11.56.05.AM.mov

Yes this also applies to Fluffy. Essentially Unseen Fist was acting as if it were Mold Breaker.

frutescens (approved):

PR #3716, opened 2024-08-23
[Bug] Skip Eternatus dialogue again

frutescens:
The user will be now able to skip Eternatus' dialogue again. 

My mess. 

The function for Eternatus' dialogue now checks if the dialogue set to be skipped and has been seen by the user before playing the dialogue. If it is set to be skipped, it skips the dialogue and begins the battle like before. 

### Screenshots/Videos

https://github.com/user-attachments/assets/dcbb92e4-5548-4611-b464-468aef613cd3

## How to test the changes?
Fight Eternatus with the settings to skip dialogue on. And then fight Eternatus with the settings to skip dialogue off.
Also, check if a new seenDialogue entry has been created.

## Checklist
- [x] Are the changes visual?


src/phases/encounter-phase.ts
```diff
@@ -358,22 +358,30 @@ export class EncounterPhase extends BattlePhase {
...
-        const cycleCount = count.toLocaleString() + ordinalIndex;
-        const encounterDialogue = i18next.t(`${(this.scene.gameData.gender === PlayerGender.FEMALE) ? "PGF" : "PGM"}battleSpecDialogue:encounter`, {cycleCount: cycleCount});
-        this.scene.ui.showDialogue(encounterDialogue, enemy?.species.name, null, () => {
+        const localizationKey = ((this.scene.gameData.gender === PlayerGender.FEMALE) ? "PGF" : "PGM")+"battleSpecDialogue:encounter";
+        if (this.scene.skipSeenDialogues && this.scene.gameData.getSeenDialogues()[localizationKey] === true) {
+          console.log(`Dialogue ${localizationKey} skipped`);
           this.doEncounterCommon(false);
-        });
+        } else {
+          const count = 5643853 + this.scene.gameData.gameStats.classicSessionsPlayed;
```
flx-sta, line 366: why `5643853` as a minum + the sessions?

src/phases/encounter-phase.ts
```diff
@@ -358,22 +358,30 @@ export class EncounterPhase extends BattlePhase {
...
+          //The two lines below check if English ordinals (1st, 2nd, 3rd, Xth) are used and determine which one to use.
+          //Otherwise, it defaults to an empty string.
+          //As of 08-07-24: Spanish and Italian default to the English translations
+          const ordinalUse = ["en", "es", "it"];
+          const currentLanguage = i18next.resolvedLanguage ?? "en";
+          const ordinalIndex = (ordinalUse.includes(currentLanguage)) ? ["st", "nd", "rd"][((count + 90) % 100 - 10) % 10 - 1] ?? "th" : "";
+          const cycleCount = count.toLocaleString() + ordinalIndex;
```
flx-sta, line 373: shouldn't the ordinal be saved a translation key, rather than some hard-coded strings?
there is a tool at hand for that
https://www.i18next.com/translation-function/plurals
```ts
{
  "key_zero": "zero",
  "key_one": "singular",
  "key_two": "two",
  "key_few": "few",
  "key_many": "many",
  "key_other": "other"
}
```

----

for yours it would be 
```

```json
{
  "ordinal_one": "st",
  "ordinal_two": "nd",
  "ordinal_three": "rd",
  "ordinal_other": "th"
}
```

For the count you could provide the formular instead of the actual count so it can handle stuff like `21`, `22` etc.

That way you also don't have to check the language that is being resolved

flx-sta (changes requested):

src/phases/encounter-phase.ts
frutescens: 5643853 was the original number used.

src/phases/encounter-phase.ts
frutescens: this can handle 21, 22, etc. + other languages don't seem to use these ordinals. 
the only reason why spanish and italian are there is because they aren't translated yet

flx-sta: I think it should be part of i18n never the less since it's language specific.
I'm personally not a fan of language specific code being hard-coded

frutescens: I don't mind adding it since I do need to learn how to make translation keys with the new format but this is an English-only problem.

src/phases/encounter-phase.ts
```diff
@@ -358,22 +358,30 @@ export class EncounterPhase extends BattlePhase {
...
-        //As of 08-07-24: Spanish and Italian default to the English translations
-        const ordinalUse = ["en", "es", "it"];
-        const currentLanguage = i18next.resolvedLanguage ?? "en";
-        const ordinalIndex = (ordinalUse.includes(currentLanguage)) ? ["st", "nd", "rd"][((count + 90) % 100 - 10) % 10 - 1] ?? "th" : "";
-        const cycleCount = count.toLocaleString() + ordinalIndex;
-        const encounterDialogue = i18next.t(`${(this.scene.gameData.gender === PlayerGender.FEMALE) ? "PGF" : "PGM"}battleSpecDialogue:encounter`, {cycleCount: cycleCount});
-        this.scene.ui.showDialogue(encounterDialogue, enemy?.species.name, null, () => {
+        const localizationKey = ((this.scene.gameData.gender === PlayerGender.FEMALE) ? "PGF" : "PGM")+"battleSpecDialogue:encounter";
+        if (this.scene.skipSeenDialogues && this.scene.gameData.getSeenDialogues()[localizationKey] === true) {
+          console.log(`Dialogue ${localizationKey} skipped`);
```
xsn34kzx, line 363: Should this still be here? Not sure what our overall policy/feelings are about having new `console.log`s since this can be somewhat practical for the general player base but, at the same time, it also shows the raw translation keys.

src/phases/encounter-phase.ts
xsn34kzx: +1 to doing as much language-based decisions as possible through i18next. I'd imagine it would be possible to not fallback / not append an ordinal to languages that don't use them (or in this case, don't have a translation key entry) somewhere in the i18next library. The main concern here is mostly having strings (and translations) in the general code base, making this snippet of code messier over time.

xsn34kzx (commented):
Threw a few logistical comments!

src/phases/encounter-phase.ts
frutescens: This is how ui.showDialogue() behaves re: skipping dialogue.

xsn34kzx: Ah, gotcha, should be fine then.

xsn34kzx (approved):
Basically what I had in mind, lgtm!

flx-sta (approved):

src/phases/encounter-phase.ts
DayKev: Also, `console.log()` statements aren't executed on main.

DayKev:
Okay so apparently if you override directly to wave 200 things break slightly. Also the ordinal thing is missing, should be a "th" after the number there.
image

Also this is almost definitely unrelated but apparently reloading during the wave 200 dialog (this time vs boss Nidoking lol) causes this to happen (see left side of screen):
image

The dialog does skip now though, so there's that.

src/phases/encounter-phase.ts
```diff
@@ -358,22 +358,28 @@ export class EncounterPhase extends BattlePhase {
...
-        //The two lines below check if English ordinals (1st, 2nd, 3rd, Xth) are used and determine which one to use.
-        //Otherwise, it defaults to an empty string.
-        //As of 08-07-24: Spanish and Italian default to the English translations
-        const ordinalUse = ["en", "es", "it"];
-        const currentLanguage = i18next.resolvedLanguage ?? "en";
-        const ordinalIndex = (ordinalUse.includes(currentLanguage)) ? ["st", "nd", "rd"][((count + 90) % 100 - 10) % 10 - 1] ?? "th" : "";
-        const cycleCount = count.toLocaleString() + ordinalIndex;
-        const encounterDialogue = i18next.t(`${(this.scene.gameData.gender === PlayerGender.FEMALE) ? "PGF" : "PGM"}battleSpecDialogue:encounter`, {cycleCount: cycleCount});
-        this.scene.ui.showDialogue(encounterDialogue, enemy?.species.name, null, () => {
+        const localizationKey = ((this.scene.gameData.gender === PlayerGender.FEMALE) ? "PGF" : "PGM")+"battleSpecDialogue:";
```
MokaStitcher, line 361: Can't do a full review right now, but there's a function "shouldSkipDialog" in ui.ts that handles adding the PGM/PGF stuff to the key and checks if the dialog should get skipped, could you call it instead of doing the check yourself?

frutescens:
My bad, I fixed the issues now and it should work :)

https://github.com/user-attachments/assets/713cf848-e4b5-424d-b975-189b2fc9895f

src/phases/encounter-phase.ts
frutescens: good idea!

src/phases/encounter-phase.ts
```diff
@@ -358,22 +358,28 @@ export class EncounterPhase extends BattlePhase {
...
+        const encounterLocalizationKey = "battleSpecDialogue:encounter";
+        if (this.scene.ui.shouldSkipDialogue(encounterLocalizationKey)) {
+          // Logging mirrors logging found in dialogue-ui-handler
+          console.log(`Dialogue ${encounterLocalizationKey} skipped`);
           this.doEncounterCommon(false);
-        });
+        } else {
+          const localizationKey = ((this.scene.gameData.gender === PlayerGender.FEMALE) ? "PGF" : "PGM")+encounterLocalizationKey;
+          const count = 5643853 + this.scene.gameData.gameStats.classicSessionsPlayed;
+          //The line below checks if an English ordinal is necessary or not based on whether an entry for encounterLocalizationKey exists in the language or not.
```
DayKev, line 369:
```suggestion
          const localizationKey = ((this.scene.gameData.gender === PlayerGender.FEMALE) ? "PGF" : "PGM") + encounterLocalizationKey;
          const count = 5643853 + this.scene.gameData.gameStats.classicSessionsPlayed;
          // The line below checks if an English ordinal is necessary or not based on whether an entry for encounterLocalizationKey exists in the language or not.
```

DayKev (dismissed):

KimJeongSun (dismissed):
it looks working in shared case with Korean option

MokaStitcher (approved):
lgtm! (tested in french + a bit of english for the ordinals, and with both player genders)

CodeTappert (dismissed):

src/phases/encounter-phase.ts
```diff
@@ -358,24 +358,29 @@ export class EncounterPhase extends BattlePhase {
...
-        const ordinalUse = ["en", "es", "it"];
-        const currentLanguage = i18next.resolvedLanguage ?? "en";
-        const ordinalIndex = (ordinalUse.includes(currentLanguage)) ? ["st", "nd", "rd"][((count + 90) % 100 - 10) % 10 - 1] ?? "th" : "";
-        const cycleCount = count.toLocaleString() + ordinalIndex;
-        const genderIndex = this.scene.gameData.gender ?? PlayerGender.UNSET;
-        const genderStr = PlayerGender[genderIndex].toLowerCase();
-        const encounterDialogue = i18next.t("battleSpecDialogue:encounter", { context: genderStr, cycleCount: cycleCount });
-        this.scene.ui.showDialogue(encounterDialogue, enemy?.species.name, null, () => {
+        const localizationKey = "battleSpecDialogue:encounter";
+        if (this.scene.ui.shouldSkipDialogue(localizationKey)) {
```
MokaStitcher, line 362: if what PR #3879 did stays as is you would need to call shouldSkipDialogue with the gender specific key now
however i'm not sure it was correct to remove the gender stuff from there, but my comment there came in after the PR was already merged :(

MokaStitcher (changes requested):
just something to update I think in regards to PR #3879

src/phases/encounter-phase.ts
flx-sta: @MokaStitcher the `localizationKey` is never changed so the skip logic works as such:

If you have the gender `Boy` and you've seen the dialogue the key is saved as `seen` without any gender suffix.

If you now change your gender to `Girl` you still wouldn't see the dialogue again (even if it's different) as the key is ultimate WITHOUT the gender. Which is an expected behavior. 
The difference in the key only exists in the `json` files, no in the i18next call logic.

MokaStitcher: if that's how we want to have it work going forward i guess it's fine, but before if you changed gender you would see the dialogue once before it got skipped again, since potentially it is quite different between genders, especially for the rival

MokaStitcher: @flx-sta but also the thing is [here](https://github.com/flx-sta/pokerogue/blob/c3e5c21463bd0f08924dad71f6db5d45122cbd65/src/ui/ui.ts#L320) unless I'm mistaken it stores the one with context as seen, so if you play exclusively as female it wouldn't then skip the male dialogs? I'll have to test it out though, maybe i'm wrong

MokaStitcher: in the context of this specific dialog it's fine with the way it's done though you're right.
the issues i have are with the rest of the dialog related to the context PR but this is not the place to discuss them, sorry

MokaStitcher (approved):

DayKev (approved):

innerthunder (approved):

PR #3725, opened 2024-08-24
[QoL] Test dialogue menu option

Opaque02:
This will allow testers and the localisation team to open a new menu that lets them see the output of any i18 key, along with the ability to add options for more outputs. This works on any language

See [this discord post for info](https://discord.com/channels/1125469663833370665/1176874654015684739/1276865740355534849)

Added a new test dialogue under the manage data section that lets you enter a i18 key (i.e. gameMode:endlessSpliced) and it shows a message with the i18 output. It also has an autocomplete functionality on your writing so you can see what possible options are around (this was entirely done by @Vassiat - thanks again, it's so epic), and it shows the output in a message of the same width and style as a dialogue (thanks @MokaStitcher)

A bit more details:
- Firstly, the autocomplete is awesome - thanks again Vassiat. It checks against what you're typing and returns a box that you can move through that shows all the possible options based on what you've typed, and updates live
- For the drop down autocomplete, you can move up/down using the arrow keys, and once you've found the option, you need to press enter to select it. This is because the Z/A button are also bound to z on your keyboard, meaning if you went to type charizard, once you typed "z" it would select whatever you've got selected
- This works with any i18 key in any language. Once you type your i18 key you _need_ to press enter (or use the "Check" button) to see it - it's for the same reason as above
- You can press space to add a new section. This new section lets you add values into variables. Each variable is added in the order it appears in the key. This can be done as many times as you want; adding a key, pressing space, adding another key and pressing space will then let you add another key. Once you add a space, you'll get a new autocomplete section so you can always some autocomplete options for what you're looking into. This is explained more below

A bit on how the keys actually work:

Let's look at a key from i18. As an example, within menu.json is this line of code: `"usernameAlreadyUsed": "The provided username is already in use"`. This means that the i18 key for this would be menu:usernameAlreadyUsed - if you run this through i18 you'll get "The provided username is already in use". This is the most basic way to use this dialogue.

As a bit of an addition, if we now look at move.json, we'll see the following:

`"pound": {
    "name": "Pound",
    "effect": "The target is physically pounded with a long tail, a foreleg, or the like."
  }`

The way this works is that move:pound has some properties - so for this, if we wanted the name, we would check for move:pound.name, whereas if we wanted the description we would use move:pound.description.

If you are ever unsure what the key part should be (i.e. the bit before the : - so move, menu etc), you can have a look at the language config.ts files for each locale, like so:

image

As you can see from above, if I wanted to get something that's in the command-ui-handler.json file, I would use commandUiHandler: to get it.

Next, let's look at a more complicated key with a variable. Again within the menu.json is this key: `"stoppedEvolving": "{{pokemonName}} stopped evolving."`. Now here we have a couple of options. The first is to run this through our dialogue test menu by using menu:stoppedEvolving, which will give us "{{pokemonName}} stopped evolving." However, since this one has something surrounded by {{ }}, that makes it a variable. That means if we tell i18 to tell us what this key is if the pokemonName is "Charmander", it'll instead return "Charmander stopped evolving."

This process works for multiple variables too; consider the i18 key `"wasReverted": "{{fusionName}} was reverted to {{pokemonName}}."`, which is found in the party-ui-handler.json file. Here there are two variables - {{fusionName}} and {{pokemonName}} - if you provide both, they'll both be added, but if you don't provide one/both, those keys will just stay as what they currently are. For example if you wanted to use it and provided the pokemonName of Abra, you would have "{{fusionName}} was reverted to Abra."

To add variables using the test menu, you want to press space. Doing so will let you add a variable (with a brand new autocomplete functionality to it). The variables should be entered in the order of how they're shown in the key - while you don't say which variable is which, the first variable added is the first variable in the translated text, the second variable for the second and so on. As above, if a variable isn't added it just uses the {{ }} value from the i18 string, and if you add too many variables, the extra ones don't get used. 

You can add as many variables as you want, with each one showing its own autocomplete box, but the autocomplete box only works on the _last_ variable you've got, so if you've made a mistake in a previous variable, you can edit it, but the autocomplete box won't show and update for it.

Lastly, if you try to use a key that doesn't exist, it'll do its best to translate or just use the whole key; this may be needed to test edge cases, though you should try to stick to using i18 keys where possible.

### Screenshots/Videos
Below is a video of the menu using a simple key - gameMode:dailyRun:

https://github.com/user-attachments/assets/c420a294-4531-4ed6-887c-a2889ab1a431

Below is a video of a prompt that has 2 variables. You can see me add no variables, then add one (where you can see the popup appear again), and then add a second one (where once again the popup is added). Also apologies about the speed - my computer is semi dying so the speed may be due to that, but I've tested on my phone and don't have any speed issues, so hopefully it's just a me thing:

https://github.com/user-attachments/assets/826f6422-cc26-4951-b51c-99ac56e9f72e

Finally, below is the same as above but in a different language:

https://github.com/user-attachments/assets/b60200d8-35f8-4299-b50d-88b962840470

## How to test the changes?
Download this PR, open the menu, open the test dialogue submenu and go have fun :)

If you do test this and the performance isn't great, let me know and I'll see if there's anything I can do to make it better

## Checklist
- [x] Are the changes visual?


src/ui/test-dialogue-ui-handler.ts
```diff
@@ -0,0 +1,149 @@
+import { FormModalUiHandler } from "./form-modal-ui-handler";
+import { ModalConfig } from "./modal-ui-handler";
+import i18next from "i18next";
+import { PlayerPokemon } from "#app/field/pokemon.js";
```
DayKev, line 4:
```suggestion
import { PlayerPokemon } from "#app/field/pokemon";
```

Opaque02: Will be in next commit

flx-sta:
This is sick. Wow

josericardo-fo:
This is amazing!

PR #3733, opened 2024-08-24
[Dev] Cleanup some opponent overrides + add new opp form override

torranx:
none

no existing override for opponent's forms + `integer` type phase out in `overrides.ts`

new override `OPP_FORM_OVERRIDES`, moved other overrides to `EnemyPokemon` class

## How to test the changes?
use similar to `STARTER_FORM_OVERRIDES`

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -842,22 +842,12 @@ export default class BattleScene extends SceneBase {
     if (Overrides.OPP_SPECIES_OVERRIDE) {
       species = getPokemonSpecies(Overrides.OPP_SPECIES_OVERRIDE);
     }
-    const pokemon = new EnemyPokemon(this, species, level, trainerSlot, boss, dataSource);
 
     if (Overrides.OPP_LEVEL_OVERRIDE !== 0) {
-      pokemon.level = Overrides.OPP_LEVEL_OVERRIDE;
+      level = Overrides.OPP_LEVEL_OVERRIDE;
     }
 
-    const speciesId = pokemon.species.speciesId;
-
-    if (speciesId in Overrides.OPP_FORM_OVERRIDES &&
-      Overrides.OPP_FORM_OVERRIDES[speciesId] &&
-      pokemon.species.forms[Overrides.OPP_FORM_OVERRIDES[speciesId]]) {
-      pokemon.formIndex = Overrides.OPP_FORM_OVERRIDES[speciesId] ?? 0;
-    }
-    if (Overrides.OPP_GENDER_OVERRIDE !== null) {
-      pokemon.gender = Overrides.OPP_GENDER_OVERRIDE;
-    }
+    const pokemon = new EnemyPokemon(this, species, level, trainerSlot, boss, dataSource);
 
     overrideModifiers(this, false);
     overrideHeldItems(this, pokemon, false);
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -843,13 +843,22 @@ export default class BattleScene extends SceneBase {
       species = getPokemonSpecies(Overrides.OPP_SPECIES_OVERRIDE);
     }
     const pokemon = new EnemyPokemon(this, species, level, trainerSlot, boss, dataSource);
+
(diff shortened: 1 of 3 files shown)


src/battle-scene.ts
```diff
@@ -843,13 +843,22 @@ export default class BattleScene extends SceneBase {
...
     }
     const pokemon = new EnemyPokemon(this, species, level, trainerSlot, boss, dataSource);
+
     if (Overrides.OPP_LEVEL_OVERRIDE !== 0) {
       pokemon.level = Overrides.OPP_LEVEL_OVERRIDE;
     }
 
+    const speciesId = pokemon.species.speciesId;
+
+    if (speciesId in Overrides.OPP_FORM_OVERRIDES &&
```
schmidtc1, line 853: Would it not be better to modify the EnemyPokemon class to take in the formIndex, like PlayerPokemon does?

torranx: the pre-existing overrides are placed there so I figured I'd do the same - but I can recheck

src/field/pokemon.ts
```diff
@@ -3746,6 +3746,18 @@ export class EnemyPokemon extends Pokemon {
...
 
+    if (Overrides.OPP_GENDER_OVERRIDE) {
+      this.gender = Overrides.OPP_GENDER_OVERRIDE;
+    }
+
+    const speciesId = this.species.speciesId;
+
+    if (speciesId in Overrides.OPP_FORM_OVERRIDES &&
+      Overrides.OPP_FORM_OVERRIDES[speciesId] &&
+      this.species.forms[Overrides.OPP_FORM_OVERRIDES[speciesId]]) {
```
DayKev, line 3757:
```suggestion
    if (speciesId in Overrides.OPP_FORM_OVERRIDES
      && Overrides.OPP_FORM_OVERRIDES[speciesId]
      && this.species.forms[Overrides.OPP_FORM_OVERRIDES[speciesId]]) {
```
I think this makes it a bit easier to read here.

DayKev (approved):

schmidtc1 (approved):

src/field/pokemon.ts
```diff
@@ -3746,6 +3746,18 @@ export class EnemyPokemon extends Pokemon {
...
     }
 
+    if (Overrides.OPP_GENDER_OVERRIDE) {
+      this.gender = Overrides.OPP_GENDER_OVERRIDE;
+    }
+
+    const speciesId = this.species.speciesId;
+
+    if (speciesId in Overrides.OPP_FORM_OVERRIDES
+      && Overrides.OPP_FORM_OVERRIDES[speciesId]
```
chaosgrimmon, line 3756: Does this mean that you can't override the value to be `0`, because zeroes are falsy?
Thinking about cases like Florges Red, Furfrou Natural, or Minior Red

torranx: > Does this mean that you can't override the value to be `0`, because zeroes are falsy? Thinking about cases like Florges Red, Furfrou Natural, or Minior Red

Yes, and we only use this override if we want the `formIndex` to not be `0` (as this is the default form). Also this is mirrored from the pre-existing form override for starters

flx-sta (approved):

src/field/pokemon.ts
chaosgrimmon: If that's how the other code is then fair enough, but when the forms are purely aesthetic, such as the above species, there isn't a "default" form in the same way that a Mega or a GMax _isn't_ a standard form.
As cosmetic forms are randomly distributed, not being able to ensure the `0`th form complicates testing them somewhat.

If it's probably not worth the devtime to make it work, that's totally understandable, just wanted to be sure in order to document the overrides clearly for Art Tech purposes.

chaosgrimmon (approved):
lgtm

src/field/pokemon.ts
torranx: I don't see what you mean, when you override a species, it is by default, set to formIndex `0` and in the cases you mentioned, those Pokemon forms are at index `0`.

PR #3740, opened 2024-08-24
[Bug] Fix transfer item text disappearing

EmberCM:
The Summary, Transfer, etc text will be visible again

Cause I broke it initially in #2221 

Only check the item quantity before `transferMode` because that is when the item quantities are shown.
The `option` that is being checked is an index so it is used for Summary, Transfer, Rename, etc. Since the item modifiers are using the same index system, it tried checking those but the item modifier was undefined because it didn't exist at that index so it skipped it and didn't display any text.

### Screenshots/Videos
before
after

## How to test the changes?
Have a Pokemon with a lot of items (6-7 different kinds, because PartyOption.TRANSFER is 5 and SUMMARY is 6), then try transferring an item from that Pokemon to another.

## Checklist
- [x] Are the changes visual?


frutescens (approved):

innerthunder (approved):

flx-sta (approved):
playtest & code-review

PR #3767, opened 2024-08-26
[Bug] Show variant icons when forms share masterlist entry

chaosgrimmon:
When a Pokemon's forms rely on the same variant `_masterlist` entry, instead of having one per form, variant icons will now correctly appear.

Reported [here](https://discord.com/channels/1125469663833370665/1238339524778790922/1275188605316501596) and [here](https://discord.com/channels/1125469663833370665/1229465263532019712/1274491742821027894).

Although #3322 addressed a different set of variant icons not appearing, it in turn caused the current batch of issues.

The way that variant icons work is that the game does not know to load them unless it finds an entry in the `_masterlist` that tells it to. When each form is loaded, the affected code determines the name of the form that the game looks for in the `_masterlist`.

#3322 replaced `formKey` with `formSpriteKey`. This distinction allows multiple forms to use the same assets. This is generally useful when they are indistinguishable at icon-scale, or select cases like Sinistea or Poltchageist, where the marks of authenticity (or lack thereof) are not visible in any sprites. It also can be used to reduce the number of characters used to name each asset for that form, by giving it a shorter alias, though at the expense of clarity.
Since generally a form's `formSpriteKey` was used on the `_masterlist`, instead of its `formKey`, the assumption was that the solution simply was that simple.

Unfortunately, many forms have a `formSpriteKey` of `null`. This appears to be the cause of the following function.
```
  getFormSpriteKey(_formIndex?: integer) {
    return this.formSpriteKey !== null ? this.formSpriteKey : this.formKey;
  }
```
By using this function instead, all of the variant icons previously malfunctioning now appear as intended.

#### Before
image
image
image

#### After
image
image
image

And some other variant icons tested:
image
image
image
image

## How to test the changes?
Use an unlocked savefile to test variant icons, or perhaps use overrides to do specific alternative icon tests.

Note:
Something about the new starter select sorting system produces an "Variant Icon Missing" warning every time the sort criteria changes, if that variant tier is the one currently preserved as a preference for that Pokemon. This is unrelated, and the game is only able to do so if it can register the icon's key on the `_masterlist` to realise that it is missing.
Just set that Pokemon's `R: Shiny` to not shiny and it should go away.

## Checklist
~~- [] Have I considered writing automated tests for the issue?~~
- [x] Are the changes visual?

### DIFF
--- a/src/data/pokemon-species.ts
+++ b/src/data/pokemon-species.ts
@@ -301,7 +301,7 @@ export abstract class PokemonSpeciesForm {
     let variantDataIndex: integer | string = this.speciesId;
     const species = getPokemonSpecies(this.speciesId);
     if (species.forms.length > 0 && formIndex !== undefined) {
-      formkey = species.forms[formIndex]?.formSpriteKey;
+      formkey = species.forms[formIndex]?.getFormSpriteKey(formIndex);
       if (formkey) {
         variantDataIndex = `${this.speciesId}-${formkey}`;
       }


DayKev (approved):

frutescens (approved):
Seems good.
Screenshot 2024-08-26 at 4 50 01 PM
Screenshot 2024-08-26 at 4 49 46 PM
Screenshot 2024-08-26 at 4 49 26 PM

torranx (approved):

Issue #3722, opened 2024-08-24
[Bug] No Indication for Sleep Move Failure on Electric Terrain


Snailman11:
### Describe the bug

"xxxxx was protected by electric terrain"

https://replay.pokemonshowdown.com/gen9customgame-2186597941

-this could also apply to Bulldoze/Earthquake and Grassy Terrain, and Priority with Psychic

### Expected behavior

There should be a message to indicate that Electric terrain has nullified the move

### Screenshots / Videos

https://github.com/user-attachments/assets/f1a7685e-071a-4e64-9dd7-e7f329bdac11

### What OS did you observe the bug on?

Other

### If other please specify

gameplay

### Which browser do you use?

Other

### If other please specify

gameplay

jessemchung:
Hello, I am looking for an easy first issue to work on.  Could I work on this?

Snailman11:
> Hello, I am looking for an easy first issue to work on. Could I work on this?

Sure

jessemchung:
@Snailman11 Wait, are you a maintainer?  I read online I should ask them before working

@innerthunder Are you a maintainer?  Can I just start working on it?

innerthunder:
Go for it! I know there are quite a few cases of move-cancelling effects missing "activation" messages, so it'd be great to have people work towards sorting those messages out.

If you have any other questions about the developer workflow or how to submit pull requests, feel free to ask in [the Discord's dev channel](https://discord.com/channels/1125469663833370665/1176874654015684739).

jessemchung:
@innerthunder 
Gotcha, yeah, so I am looking into the fix and I suspect adding a new trigger file might be helpful (similar to the ability-trigger.ts file) but can also add a new entry to the battle.ts file for specific terrains.  Not sure so will ask in the dev channel.

Will also need to make adjustments to this case
`      if (!typeMultiplier.value) {
        this.scene.queueMessage(i18next.t("battle:hitResultNoEffect", { pokemonName: getPokemonNameWithAffix(this) }));
      }
`

I also probably need to make changes to
`export class DrowsyTag extends BattlerTag {`

Mainly adding a new queue Message 

Also need to watch here
```
    case StatusEffect.SLEEP:
      if (this.isGrounded() && this.scene.arena.terrain?.terrainType === TerrainType.ELECTRIC) {
        // add statement here.  But not sure
        return false;
      }
```

Will need to try to figure out the override feature

jessemchung:
[...]

PR #3775, opened 2024-08-26
[Bug] Added fix for 3722 so that text correctly appears when sleep fails in electric terrain

jessemchung:
This fix is for this issue:
https://github.com/pagefaultgames/pokerogue/issues/3722

The user will now, when attempting to put a pokemon to sleep on electric terrain, will receive a message why it failed

I needed to add fixes to drowsy (from yawn) as well as to the tryStatusEffect 

Makes it more similar to the mainline games and pokemon showdown

A simple change was added in the drowsy state check so that the message will appear.  I also had to add logic to the final

### Screenshots/Videos
See below

## How to test the changes?

I used override, I honestly am not sure how best to test yet.  I set my pokemon to Tapu Koko with spore and yawn as moves and the opponent as well to test.   Below is the basic override to see things. 
```
  readonly STARTER_SPECIES_OVERRIDE: Species | number = Species.TAPU_KOKO;
  readonly ABILITY_OVERRIDE: Abilities = Abilities.NONE;
  readonly PASSIVE_ABILITY_OVERRIDE: Abilities = Abilities.NONE;
  readonly STATUS_OVERRIDE: StatusEffect = StatusEffect.NONE;
  readonly GENDER_OVERRIDE: Gender | null = null;
  readonly MOVESET_OVERRIDE: Array<Moves> = [Moves.SPORE, Moves.YAWN];
  readonly SHINY_OVERRIDE: boolean = false;
  readonly VARIANT_OVERRIDE: Variant = 0;
```

## Checklist
- I set the target branch to beta
- Not sure how best to look at this but I confirmed in the issue that I would fix it
- Yes
- Yes
- I considered it but decided no for now
- Text was already there.  Just no triggering logic
-  Yes
    - No?  But I honestly am not sure how my code might be causing issues.  
    

> FAIL  |main| src/test/moves/flower_shield.test.ts > Moves - Flower Shield > does not increase defense of a pokemon in semi-vulnerable state
> AssertionError: expected +0 to be 1 // Object.is equality
> 
> - Expected
> + Received
> 
> - 1
> + 0
> 

- [x] Are the changes visual?
  - Yes

Original Issue on display

https://github.com/user-attachments/assets/9658fa79-3638-4274-9a0a-b6f48862a5ec

Fixes both with spore and sing to showcase 

https://github.com/user-attachments/assets/92bb4ba5-5a34-4019-8069-9d0a88a61027

https://github.com/user-attachments/assets/044916ef-c235-4aaa-b343-54e8c079406c


torranx:
Shouldn't you be adding in `terrain.isMoveTerrainCancelled` instead?

src/field/pokemon.ts
```diff
@@ -2767,6 +2767,14 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
 
   trySetStatus(effect: StatusEffect | undefined, asPhase: boolean = false, sourcePokemon: Pokemon | null = null, cureTurn: integer | null = 0, sourceText: string | null = null): boolean {
     if (!this.canSetStatus(effect, asPhase, false, sourcePokemon)) {
+      switch (effect) {
+      case StatusEffect.SLEEP:
+        if (this.isGrounded() && this.scene.arena.terrain?.terrainType === TerrainType.ELECTRIC) {
+          this.scene.queueMessage(i18next.t("terrain:defaultBlockMessage", { pokemonNameWithAffix: getPokemonNameWithAffix(this), terrainName: getTerrainName(TerrainType.ELECTRIC) }));
```
schmidtc1, line 2773: You do not need to add this switch here. `queueMessage` can simply be added inside the condition in `canSetStatus` here:
https://github.com/pagefaultgames/pokerogue/blob/762aa9e49b8df80d7ab2bea1a7074f5b6da53be3/src/field/pokemon.ts#L2739-L2741

schmidtc1 (changes requested):

src/field/pokemon.ts
schmidtc1: Disregard this in favor of @torranx's suggestion, I should stop reviewing while I'm sleep deprived

jessemchung:
> Shouldn't you be adding in `terrain.isMoveTerrainCancelled` instead?

@torranx Wait, so, terrain.isMoveTerrainCancelled seems like it is intent on nullifying the move entirely like how psychic terrain nullfies all priority moves.

In the case of electric terrain, what about relic song from Meloetta?  That move would still deal damage just not put the target to sleep?  Also, what about effects that put the pokemon to sleep like effect spore?

src/field/pokemon.ts
jessemchung: I think it has something to do with getTargetBenefitScore also checking canSetStatus which causes a message in their to occur multiple times

schmidtc1: I'm pretty sure that occurs due to canSetStatus being used by the enemy AI calculations. Then you may need to use the `quiet` boolean to prevent text showing up during those calculations:
```typescript
 if (this.isGrounded() && this.scene.arena.terrain?.terrainType === TerrainType.ELECTRIC) { 
   if (!quiet) {
     this.scene.queueMessage(i18next.t("terrain:defaultBlockMessage", { pokemonNameWithAffix: getPokemonNameWithAffix(this), terrainName: getTerrainName(TerrainType.ELECTRIC) }));
   }
   return false; 
 } 
```

schmidtc1: I think that's the correct approach. I'd assume the apply function should be sending false for the `quiet` boolean in there, since we are actually applying the effect rather than simulating it like `getTargetBenefitScore` does.

jessemchung: Gotcha, yeah looking at the situation, I tested with will o wisp and that at some point also needs to have the text appear.  In any case, seems good to go.  Tested with this quiet and things are working correctly.

frutescens:
Bug Issue #3722

Snailman11:
Will this fix the protected by message not displaying for status on Misty Terrain?

https://github.com/user-attachments/assets/3593b1e3-90b5-41c9-ab47-2a56f9aadc39

jessemchung:
> Will this fix the protected by message not displaying for status on Misty Terrain? -Looking through the code and I think that's unlikely. I'll open another bug report. #3898
> Misty.mp4

It will not, that would also be a very similar fix though.

jessemchung:
@schmidtc1 so the issue so far is that for some reason, setting quiet to false causes toxic damage to not work correctly.  Need to understand why still.

schmidtc1:
> @schmidtc1 so the issue so far is that for some reason, setting quiet to false causes toxic damage to not work correctly. Need to understand why still. I suspect this might be better to adjust the if statement to check which phase it is instead of checking the quiet property as for some reason toxic damage is tied to that variable. At least, this way, fewer changes would be needed

My guess is that the asPhase variable is being set to false and used as the quiet variable passed to `canSetStatus`, since you are using `trySetStatus` in the apply function. I came across this issue a while ago and had a PR to make a separate quiet parameter in `trySetStatus`, but it never went through. 

Imo, a separate quiet parameter should still be implemented to pass to canSetStatus, instead of using asPhase in place of the quiet parameter. 

As an alternative solution, you could maybe use `canSetStatus` in the if statement, and `trySetStatus` inside the block. But you'd have to play around with that to see if it works.

jessemchung:
Right now that you mentioned it, I believe you are right.  This part is probably the part causing a few issues.  I have an idea for a fix.  The code will be simple but a more elegant solution is likely a good idea

```
    if (asPhase) {
      this.scene.unshiftPhase(new ObtainStatusEffectPhase(this.scene, this.getBattlerIndex(), effect, cureTurn, sourceText!, sourcePokemon!)); // TODO: are these bangs correct?
      return true;
    }

```

jessemchung:
@schmidtc1 Hey quick question, what happened to that pull request? Also, I did find a solution sort of related to this

    As an alternative solution, you could maybe use canSetStatus in the if statement, and trySetStatus inside the block. But you'd have to play around with that to see if it works.

Now, I tried this and the problem is this chunk of code that causes problems if run multiple times (see below code block). Now, one option is to have the quiet variable work as it seems like it should but I don't think that is necessary. The big issue is that canSetStatus probably should only check if it is possible to setStatus and not handle other applications. As a result, the solution I have now is to split that function into two pieces. One part only checks if it is possible for the pokemon to be statused and the other handles the other logic like below. I ran the tests and everything passes.  


```
    applyPreSetStatusAbAttrs(StatusEffectImmunityAbAttr, this, effect, cancelled, quiet);

    const userField = this.getAlliedField();
    userField.forEach(pokemon => applyPreSetStatusAbAttrs(UserFieldStatusEffectImmunityAbAttr, pokemon, effect, cancelled, quiet));
```

schmidtc1:
> @schmidtc1 Hey quick question, what happened to that pull request? Also, I did find a solution sort of related to this
> 
> ```
> As an alternative solution, you could maybe use canSetStatus in the if statement, and trySetStatus inside the block. But you'd have to play around with that to see if it works.
> ```
> 
> Now, I tried this and the problem is this chunk of code that causes problems if run multiple times (see below code block). Now, one option is to have the quiet variable work as it seems like it should but I don't think that is necessary. The big issue is that canSetStatus probably should only check if it is possible to setStatus and not handle other applications. As a result, the solution I have now is to split that function into two pieces. One part only checks if it is possible for the pokemon to be statused and the other handles the other logic like below. I ran the tests and everything passes.
> 
> ```
>     const cancelled = new Utils.BooleanHolder(false);
>     applyPreSetStatusAbAttrs(StatusEffectImmunityAbAttr, this, effect, cancelled, quiet);
> 
>     const userField = this.getAlliedField();
>     userField.forEach(pokemon => applyPreSetStatusAbAttrs(UserFieldStatusEffectImmunityAbAttr, pokemon, effect, cancelled, quiet));
> ```

It's a bit messed up and outdated due to some beta merge issues, but here it is: #2663

jessemchung:
Gotcha.  Sorry to hear that pull request didn't get pulled in. Looks like a ton of work went into it

src/data/battler-tags.ts
```diff
@@ -822,11 +822,11 @@ export class DrowsyTag extends BattlerTag {
   }
 
   canAdd(pokemon: Pokemon): boolean {
-    const checkTerrain = pokemon.scene.arena.terrain?.terrainType !== TerrainType.ELECTRIC || !pokemon.isGrounded();
-    if (checkTerrain === false) {
+    const isNotElectricTerrainProtected = pokemon.scene.arena.terrain?.terrainType !== TerrainType.ELECTRIC || !pokemon.isGrounded();
```
schmidtc1, line 825: I feel like `isElectricTerrainProtected` makes more sense, and then refactoring to fit that name

jessemchung: I see, I was thinking about that as a possibility in which case, the return would be return !isElectricTerrainProtected

And then isElectricTerrainProtocted could be
`(pokemon.scene.arena.terrain?.terrainType === TerrainType.ELECTRIC && !pokemon.isGrounded();)`

src/data/battler-tags.ts
```diff
@@ -822,7 +822,11 @@ export class DrowsyTag extends BattlerTag {
   }
 
   canAdd(pokemon: Pokemon): boolean {
-    return pokemon.scene.arena.terrain?.terrainType !== TerrainType.ELECTRIC || !pokemon.isGrounded();
+    const isElectricTerrainProtected = pokemon.scene.arena.terrain?.terrainType === TerrainType.ELECTRIC && !pokemon.isGrounded();
```
Tempo-anon, line 825: nit:  `const isElectricTerrainProtected = )pokemon.scene.arena.terrain?.terrainType === TerrainType.ELECTRIC && !pokemon.isGrounded());`

src/field/pokemon.ts
```diff
@@ -2804,6 +2808,14 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       break;
     }
 
+
+    return true;
+  }
+
+
+  canSetStatus(effect: StatusEffect | undefined, quiet: boolean = false, overrideStatus: boolean = false, sourcePokemon: Pokemon | null = null): boolean {
```
Tempo-anon, line 2816: Add documentation to this new function

src/field/pokemon.ts
```diff
@@ -2804,6 +2808,14 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    const checker = this.checkIfCanSetStatus(effect, quiet, overrideStatus, sourcePokemon);
```
Tempo-anon, line 2817: nit: rename variable name because `checker` makes it seem like it's an object that is checking something rather than a boolean

src/data/move.ts
```diff
@@ -1953,6 +1953,11 @@ export class StatusEffectAttr extends MoveEffectAttr {
           return false;
         }
       }
+
+      if (!pokemon.status || (pokemon.status.effect === this.effect && moveChance < 0)) {
+        pokemon.checkIfCanSetStatus(this.effect, false, false, user);
+      }
+
       if ((!pokemon.status || (pokemon.status.effect === this.effect && moveChance < 0))
```
Tempo-anon, line 1961: Is this the same `if`? If so, why can't this `checkIfCanSetStatus` call just go in the existing if?

Tempo-anon (changes requested):

src/field/pokemon.ts
jessemchung: Where do I add documentation for the new function?

src/data/move.ts
jessemchung: I just realized yes it can.  It just has to go before pokemon.trySetStatus because that function prevents messages from showing up.  I got thrown off initially as I tried to put it in the conditional block but that didn't work.

Fixed now

src/data/battler-tags.ts
jessemchung: Gotcha, added paranthesis

src/field/pokemon.ts
jessemchung: Nevermind, figured it out.  I wasn't sure what you meant since the original function canSetStatus did not have any documentation.  Added documentation both for that function and the new function.

Tempo-anon:
Looks like some other changes were caught up in this commit? https://github.com/pagefaultgames/pokerogue/pull/3775/commits/754d3775396cbfcd6b9b981403df2b6bed69c10c

jessemchung:
> Looks like some other changes were caught up in this commit? [754d377](https://github.com/pagefaultgames/pokerogue/commit/754d3775396cbfcd6b9b981403df2b6bed69c10c)

@Tempo-anon Yeah, I don't know but this situation is basically unsalvagable. Attempt to pull down beta made things worse.  Made a new pull request

https://github.com/pagefaultgames/pokerogue/pull/3972

innerthunder:
closing this in favor of the new PR without conflicts

PR #3781, opened 2024-08-26
[Dev] Clarify Pichu Evolution behaviour in code

CodeTappert:

None. This does not change anything to the user

Because this was not clear from reading the code. (For example scatterbug is clear because pre and post have the same form key)

I basically just wrote in the code that spiky pichu can only evolve into partner and non spiky into non partner (like it already is)

### Screenshots/Videos

None

## How to test the changes?

Evolve pichu in both forms and see the outcome

## Checklist
- [ ] Are the changes visual?


frutescens (approved):
Tested + Seems to evolve correctly. 

From normal Pichu
image

PR #3785, opened 2024-08-26
[Bug] Fix Inconsistency with stat boost when breaking through boss shields + tests

MokaStitcher:
Breaking a boss' shields one by one or all at a time will result in the same total number of stats boosts for the boss, while before it was inconsistent
On the backend, this PR also makes a couple small adjustments to the damage calculation for shields, overrides and test suites

Breaking a boss' shield increases a random stat by 1, or by 2 if it's the last shield.
However, with how it's implemented currently, when breaking the last shield the extra boost is applied to for all broken shields if several of them were broken in a single hit.

This is especially visible for the final boss fight since the stats boosts are retained between phases:
- If you down its shields one by one they will give boosts of 1, 1 and 2 for the last one, for a total of 4 when entering the second phase.
- If you down its 3 shields at once they will each give +2, for a total of 6 when entering the second phase
This inconsistency is not desirable especially since it actively punishes doing good on the fight.
Instead, no matter how the fight goes the stats up from the 3 broken shields should always be 4. [The change has been approved by damo](https://discord.com/channels/1125469663833370665/1276155940801871934/1276160400471359560)

### Pokemon.ts
* Changed the damage calculation for the broken shields to use `Utils.toDmgValue` instead of rounding, for consistency with how the damage in handled in the rest of the game (see PR #3557 )
* Fixed the stat boosts inconsistency when breaking multiple shields at once
### Overrides & Tests
* `battle-scene.ts`: The species of an enemy pokemon influence whether it's a boss or not. However this is computed before the Species override is applied. I changed it so that when the enemy Species is overriden the `isBoss` value is also updated to take the new Species into account
* The above change just required a small update in `src/test/moves/fusion_flare.test.ts` because the Reshiram that was used is now treated as a boss (like all Legendary Pokemon are by default in the game), and so it had its passive which prevented freezing and broke the test
* Creates an override `OPP_HEALTH_SEGMENTS_OVERRIDE` that allows forcing a Pokemon to be a boss (or not) with the given amount of health segments
* Added tests for the boss shields

### Screenshots/Videos
Before, breaking shields one by one on Eternatus results in +4 stats boosts total:

https://github.com/user-attachments/assets/f0ada1c4-4e84-4448-b6b7-ae99b991966a

Before, breaking all shields at once on Eternatus results in +6 stats boosts total (the fact that all the boosts went to sp atk is unrelated and just unlucky):

https://github.com/user-attachments/assets/f0080e16-6cc8-4e23-ad82-540be0ab508d

After, breaking all shields at once on Eternatus results in +4 stats boosts total:

https://github.com/user-attachments/assets/b891454e-4b17-44e0-96f6-8b338c6044d5

## How to test the changes?

Run tests:
```
npm run test src/test/boss-pokemon.test.ts
```

Here's the [save file](https://discord.com/channels/1125469663833370665/1276155940801871934/1277668090183815220) used for the "Before" videos. Use `Throat Chop` to take the shields down one by one or `Dragon Hammer` to take them all down at once

You can use the following override to test the Eternatus fight. Use `Ancient Power` to break one shield at a time and `Psyko` to break all of them at once
``` 
  STARTING_WAVE_OVERRIDE: 200,
  STARTING_BIOME_OVERRIDE: Biome.END,
  STARTING_LEVEL_OVERRIDE: 400,
  STARTER_SPECIES_OVERRIDE: Species.MEWTWO,
  MOVESET_OVERRIDE: [ Moves.PSYCHIC, Moves.ANCIENT_POWER, Moves.SUPER_FANG, Moves.FALSE_SWIPE],
  OPP_MOVESET_OVERRIDE: SPLASH_ONLY
```

Use the following override to test a random boss fight. You can use `Super Fang` or `False Swipe` to bring shields down without killing the boss
```
  OPP_SPECIES_OVERRIDE: Species.RATICATE,
  OPP_LEVEL_OVERRIDE: 200,
  OPP_HEALTH_SEGMENTS_OVERRIDE: 6,
  OPP_MOVESET_OVERRIDE: SPLASH_ONLY,
  STARTING_LEVEL_OVERRIDE: 400,
  STARTER_SPECIES_OVERRIDE: Species.MEWTWO,
  ABILITY_OVERRIDE: Abilities.NO_GUARD,
  MOVESET_OVERRIDE: [ Moves.PSYCHIC, Moves.ANCIENT_POWER, Moves.SUPER_FANG, Moves.FALSE_SWIPE]
```

## Checklist
- [X] The PR is self-contained and cannot be split into smaller PRs? (more or less i guess xd)
- [X] Are the changes visual?


KimJeongSun (approved):

frutescens (approved):

src/battle-scene.ts
```diff
@@ -841,12 +841,13 @@ export default class BattleScene extends SceneBase {
...
+      level = Overrides.OPP_LEVEL_OVERRIDE;
+    }
     if (Overrides.OPP_SPECIES_OVERRIDE) {
       species = getPokemonSpecies(Overrides.OPP_SPECIES_OVERRIDE);
-    }
-
-    if (Overrides.OPP_LEVEL_OVERRIDE !== 0) {
-      level = Overrides.OPP_LEVEL_OVERRIDE;
+      // The fact that a Pokemon is a boss or not can change based on its Species and level
+      boss = this.getEncounterBossSegments(this.currentBattle.waveIndex, level, species) > 1;
```
Tempo-anon, line 850: This will always result in true once level is over 100 https://github.com/pagefaultgames/pokerogue/blob/main/src/battle-scene.ts#L1331 since `ret` will be at least 3

Tempo-anon (changes requested):

src/battle-scene.ts
MokaStitcher: only if the wave is a boss wave (so a multiple of 10) or if the pokemon is a legendary/sublegendary/mythical, otherwise the `getEncounterBossSegments` will return 0 before reaching that line https://github.com/pagefaultgames/pokerogue/blob/762aa9e49b8df80d7ab2bea1a7074f5b6da53be3/src/battle-scene.ts#L1326C7-L1326C13

Tempo-anon: Ah ok you're right. I remember going through and documenting battle-scene.ts in the past but looks like that PR got lost somewhere. Looks good to me then

src/field/pokemon.ts
```diff
@@ -4203,17 +4203,13 @@ export class EnemyPokemon extends Pokemon {
...
-      case 2:
-        if (this.bossSegments >= 5) {
-          statLevels++;
-        }
-        break;
+      // increase the boost if the boss has at least 3 segments and we passed last shield
+      if (this.bossSegments >= 3 && this.bossSegmentIndex === 1) {
+        statLevels++;
+      }
+      // increase the boost if the boss has at least 5 segments and we passed the second to last shield
```
Tempo-anon, line 4210: Should this only apply to the classic final boss? Won't this also affect endless?

src/phases/encounter-phase.ts
```diff
@@ -26,6 +26,7 @@ import { ScanIvsPhase } from "./scan-ivs-phase";
 import { ShinySparklePhase } from "./shiny-sparkle-phase";
 import { SummonPhase } from "./summon-phase";
 import { ToggleDoublePositionPhase } from "./toggle-double-position-phase";
+import Overrides from "#app/overrides";
```
Tempo-anon, line 29: should this be `{ Overrides }` for consistency?

src/test/moves/fusion_flare.test.ts
```diff
@@ -27,7 +27,7 @@ describe("Moves - Fusion Flare", () => {
     game.override.moveset([fusionFlare]);
     game.override.startingLevel(1);
 
-    game.override.enemySpecies(Species.RESHIRAM);
+    game.override.enemySpecies(Species.RATTATA);
```
Tempo-anon, line 30: Is this because the Reshiram is making things weird? Can it be overriden to not be a boss?

src/field/pokemon.ts
MokaStitcher: Yeah it should apply in endless too
I didn't change the logic of which shields give the extra boost and when even though I agree it makes little sense for endless in this state. But it would be the job of the balance team to decide if it needs to change and damo said I shouldn't change it for now, just fix the inconsistency

src/test/moves/fusion_flare.test.ts
MokaStitcher: yeah because of the change in the override from your first comment Reshiram is now treated as a boss so it had its passive which prevented freezing. It didn't seem like it needed to be a Reshiram for the test so I put Rattata in but I could definitely put it back and force Reshiram to non-boss instead with the new override

src/phases/encounter-phase.ts
MokaStitcher: i believe it doesn't work with `{ Overrides }` because of the way they are defined
```
export default {
  ...defaultOverrides,
  ...overrides
} satisfies InstanceType<typeof DefaultOverrides>;
```

not too familiar with how it works exactly but vscode is not happy with me if I put the brackets in

Tempo-anon (approved):

PR #3790, opened 2024-08-26
[Localisation] [PT-BR] Re-Added Portuguese PKMN names

AsdarDevelops:
I think during the migration it was assumed to be untranslated (like Catalan) but in reality it simply really looks like English (same thing for Spanish PKMN names). Even though I'm adding all the keys, the issue was happening only with Type Null and the Paradox mons. Adding all the keys so it's nice and "completed" from the GL user interface.


josericardo-fo (approved):

DayKev (approved):

flx-sta:
> I think during the migration it was assumed to be untranslated (like Catalan) but in reality it simply really looks like English (same thing for Spanish PKMN names). Even though I'm adding all the keys, the issue was happening only with Type Null and the Paradox mons. Adding all the keys so it's nice and "completed" from the GL user interface.

that was exactly what i thought. My bad

flx-sta (approved):

PR #3841, opened 2024-08-27
[Bug] Prevent daily runs from breaking due to very (un)lucky RNG

DayKev:
In a way... none.

It was possible (though extremely unlikely) for the RNG to generate a 9+ cost Pokémon as the first starter in Daily Run mode, thus causing the third Pokémon to be a 0 cost (which doesn't exist) and crashing the game.

The cost of the first starter in Daily Run mode is now capped at 8.

## How to test the changes?
Start a Daily Run and see that the game doesn't crash. et voilà

## Checklist

### DIFF
--- a/src/data/daily-run.ts
+++ b/src/data/daily-run.ts
@@ -40,7 +40,7 @@ export function getDailyRunStarters(scene: BattleScene, seed: string): Starter[]
     }
 
     const starterCosts: integer[] = [];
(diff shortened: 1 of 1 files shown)


torranx (approved):

frutescens (approved):

schmidtc1 (approved):

Tempo-anon (approved):
This is so funny. Has something like this ever happened before? I'd also suggest adding a comment to explain why there's a cap of 8 here for anyone else who comes here later

PR #3847, opened 2024-08-27
[Test] Add test for final boss fight phase switch

MokaStitcher:
No changes hopefully, this just adds a test and makes a tiny refactor along the way

There was a test marked as 'todo' to check that the final boss switched phase properly when its final shield gets broken.
Since there have been issues in the past with it not working with status damage, tests were added for both status and direct damage

- Implemented test marked as 'todo' for final boss switching to its second phase in `src/test/final_boss.test.ts`
- This required being able to skip through the dialog after the first phase. To do this I updated the helper function `gameManager.runToFinalBossEncounter` to mock `ui.shouldSkipDialogue` instead of what it was doing before, so that all dialog gets skipped and not just the one before the fight
- This meant in `ui.showDialog` the function had to actually be called (the code was already doing the same checks as shouldSkipDialogue rather than calling it), so I did that

### Screenshots/Videos
N/A

## How to test the changes?

`npm run test src/test/final_boss.test.ts`

You could play a bit and make sure the dialogue gets skipped if the setting is enabled like before

## Checklist
- [ ] Are the changes visual?


DayKev:
Couple comments/changes here: https://github.com/MokaStitcher/pokerogue/pull/1

PR #3861, opened 2024-08-28
Fun And Games! Encounter

ben-lear:
https://github.com/user-attachments/assets/00a1d17e-d3b2-4e01-b902-dee6380bb7c9

https://github.com/user-attachments/assets/023c2dc5-5987-4eb4-9fde-979dae3d4e56

https://github.com/user-attachments/assets/175357d7-0e5b-46cf-9ea9-02cd5b3db76d


src/battle-scene.ts
```diff
@@ -254,7 +254,7 @@ export default class BattleScene extends SceneBase {
   public pokemonInfoContainer: PokemonInfoContainer;
   private party: PlayerPokemon[];
   public mysteryEncounterData: MysteryEncounterData = new MysteryEncounterData(null);
-  public lastMysteryEncounter: MysteryEncounter;
+  public lastMysteryEncounter?: MysteryEncounter;
```
DayKev, line 257: You mark it as optional but then use `!` everywhere to say it's always defined, what's the reason for that?

ben-lear: `currentBattle.mysteryEncounter` is the one that is defined as optional and is mostly declared as defined everywhere. Reason being, mysteryEncounter will *not* be defined in standard battles, but during mystery encounters it's *always* defined. So once the conditional check has been made in a couple of the critical path locations (`battle-scene` mostly) to ensure it's defined, during actual ME specific paths, there is no need to continue doing conditional checks for the ME object.

`lastMysteryEncounter` is one that *should* be conditionally checked everywhere it is used. Because its sole purpose is just to finish unloading/destroying the remainder of any visual objects, etc.

PR #3870, opened 2024-08-28
[Bug] Starter select default attributes fixes

MokaStitcher:
- [bug] It is no longer possible to start a run with a Pokemon form that wasn't unlocked
- [bug] It is no longer possible to start a run with a Pokemon shiny variant that wasn't unlocked
- [qol] The last used variant preference is no longer deleted when turning shiny off. When turning shiny back on again it will go back to that variant instead of defaulting the the base variant

During the last dev cycle there were some important changes to the starter select UI (#3373 ) which resulted in default selections for starter Pokemon being partially broken
- Female only Pokemon would default to male
- Pokemon with multiple forms would default to the first one even if it wasn't unlocked
- Pokemon with only shiny Tier 2 or 3 unlocked would default to shiny Tier 1, etc.

There were hotfixes for these (PR #3601 / PR #3617 ) which worked fine for the display in the UI, but when starting the run the wrong attribute was still set. But after that if you quit and started another run with the same Pokemon, the correct attributes would be here this time.

This PR addresses this issue as well as a couple other edge cases due to unexpected dex attributes

- Reverted most of #3617 because the fix for default forms happened too late down the chain to be fully effective
- Updated `getCurrentDexProps` to properly set the default form, gender, nature, ability and shiny variant if there is no starter preference for them (initial version of this fix by @Opaque02 )
- All the issues that occurred mean that it is possible that some players have invalid preferred attributes set for some of their Pokemon (stored locally). There was already code responsible for checking for invalid preferences in `setSpecies`, which is called way too often for what we want to do. It also didn't delete the invalid preferences outside of its scope, which means that it didn't actually impact the actual attributes used when starting the run.
I moved all those checks so that they are executed once for all starters when opening the UI instead and actually delete the invalid preferences. Now any preference that are used everywhere else down the line should be valid
- Checks for unlocked variants where consolidated across the UI to account for the very weird edge case where a Pokemon has a T2 or T3 variant flag without the shiny flag
- Moved some const declarations around for better readability

### Screenshots/Videos

All before videos/screenshot are taken from current main

**Default Form**
Before: In this save file Unown form A is not unlocked. But when opening the UI for the first time it shows form A. When going to Unown it switches to form B, but when adding it to the team it still shows form A. When starting the run form A is used. After that, quitting the run and going back to make a new game would show form B from the start and all would work fine

https://github.com/user-attachments/assets/60d1e15a-8a4f-4eea-ae90-44f75c2a91b7

After, when opening the same save file for the first time:

https://github.com/user-attachments/assets/0d7cbabd-e051-4839-a1e2-9f9bb44f2644

**Default Variant**
Before: In this save file Solrock and Relicanth were only caught once, with shiny tier 3 and 2. In this case it shouldn't be possible to turn off the shiny state, but currently that's how it is by default. If you turn it on, part of the UI is not set to the correct variant but instead defaults to the base shiny, which isn't unlocked. When starting the run like this, Relicanth would have it base shiny colors. Like for Unown, quitting and starting a new game after that would fix it

_variant_before

After: It is no longer possible to switch shiny state on/off for these mons, and the proper variant is shown and used when starting a run

_variant_after

**Other shiny edge case**
Before: one person shared a save file where it was possible to switch to the shiny form for Zamazenta even if they had never caught its shiny. Somehow their dex data has the flag for variant 2, which means Zamazenta would show up in the shiny filter and they could switch to it and start a run with it (without the luck)
_zama_before

After: no longer possible to switch to shiny, Pokemon no longer shows up in shiny filter
_zama_after_1
_after_zama2

## How to test the changes?

[Save file with Unown and Zamazenta from the examples](https://discord.com/channels/1125469663833370665/1267339517371875361/1278397459462226082)
[Save file with Solrock and Relicanth from the examples](https://discord.com/channels/1125469663833370665/1267339517371875361/1278397521647108222)

**General Testing**
* open the game, preferably with a save that doesn't have everything unlocked
* play around trying to make sure that the default selections for gender, form, etc. are in line with what's unlocked in that file
* make sure that when starting the run the correct attributes are used

**To test the clearing of invalid starter preferences:**
* import the [save file with everything](https://github.com/pagefaultgames/pokerogue/blob/main/src/test/utils/saves/everything.prsv)
* change a bunch of settings on the starters (nature, gender, form, shiny state, variant, ability, ...)
* Exit the starter UI to save the preferences
* Import a different save file
* Make sure that the variants, forms, genders, etc that are not unlocked in that save file are not shown in the UI or used when starting a run

For example you could
* Load up everything.prsv
* Switch Relicanth shiny state twice, so that it's set to "off" in the starter preferences
* Exit starter UI to save the starter preferences
* Load up the Relicanth save from above
* Make sure that Relicanth is showing as shiny T2

## Checklist
- [ ] There is no overlap with another PR? (this is based off of #3621 which should get closed)
- [X] Are the changes visual?


frutescens (approved):

KimJeongSun (approved):
LGTM!

innerthunder (approved):

PR #3871, opened 2024-08-28
[Localization] Use gender context for achievments/achv

flx-sta:

Effectively none

Right now we have the inconvenient situation of having a `achv-male.json` and `achv-female.json`.
I wanted to reduce the redundancy by merging those by using the [i18n `context`](https://www.i18next.com/translation-function/context) key.

Since `en` is the single source of truth for the language keys all `achv.name` that have need a different key for female translations had to be added as an empty `"name_female": null` entry.

<sub>This change might also create new translation "voids" in [GitLocalize](https://gitlocalize.com/repo/9839). That's unfortunately a necessary evil</sub>

To be able to merge `male` and `female` achv files I used the file-compare tool of my IDE. All Asian languages had no differences between m & f.

### Screenshots/Videos
<sup>Don't worry about the `invalid date`. This is because I "cheater" the achievments</sup>
imageimage

## How to test the changes?
- Start the game in a language that has gender specific achievment names
    - e.g. `de`, `es`, `pt_BR`, `fr`
- Check the achievements (I hope you have some unlocked. I have no data with many unlocked :(... If you have some feel free to share)
- Change your gender to a different one
- Check them again => Should render the gender context string

## Checklist
- [x] Are the changes visual?


CodeTappert:
To be fair the .ts code had a good solution for it...

flx-sta:
> To be fair the .ts code had a good solution for it...

If you have a similar solution based on `json` feel free to share.
We require `json` files to be able to utilize lazy-load.

The last step we have to do is eliminate the `config.ts` file so a solution that happens inside there wouldn't work either

CodeTappert:
> > To be fair the .ts code had a good solution for it...
> 
> If you have a similar solution based on `json` feel free to share.
> We require `json` files to be able to utilize lazy-load.
> 
> The last step we have to do is eliminate the `config.ts` file so a solution that happens inside there wouldn't work either

Where is it currently decided which file should be used? I remeber doing Something with PGM and PGF.
We can just write in the code there for dialogue to use PGM always for language that have the same dialoue for male and female

flx-sta:
> https://github.com/pagefaultgames/pokerogue/pull/3871#issuecomment-2316275831

This would mean more hard-coded language-specific stuff.
This is not good practice and should be avoided at all cost. All language specific behavior should be living behind the i18n module and it's features

frutescens (approved):

torranx (approved):

PR #3697, opened 2024-08-22
[Localization] Change a few Evil Admin lines to sound more coherent. 

Blitz425:
## What are the changes the user will see?
New lines for select evil admins.

## Why am I making these changes?
Some of the lines used had grammar issues, or straight up did not make sense in the context of Rogue or at all. Aqua Admin Matt was using a pre battle line when he is defeated for some reason as an example. Majority of these lines are taken directly from Bulbapedia. 

## What are the changes from a developer perspective?
Archer, Courtney, Matt, Zinzolin, and Rood all got changes to their dialogue. 

## How to test the changes?
Set an override to an Admin spawn and see.
## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [ ] Have I considered writing automated tests for the issue?
- [ ] If I have text, did I add placeholders for them in locales?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [ ] Have I provided screenshots/videos of the changes?

PR #3895, opened 2024-08-29
Revert "[Localization] Change a few Evil Admin lines to sound more coherent. "

CodeTappert:
Reverts pagefaultgames/pokerogue#3697


CodeTappert:
We do a new PR instead

PR #3897, opened 2024-08-29
[Move] False Swipe and Hold Back modification

EmberCM:
False Swipe will continue to hit the enemy even at 1 hp and will no longer fail

Someone mentioned it in [Discord](https://discord.com/channels/1125469663833370665/1232191501723762709)

Remove the move condition

## How to test the changes?
```ts
const overrides = {
  MOVESET_OVERRIDE: [Moves.FALSE_SWIPE]
```

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -1261,10 +1261,6 @@ export class SurviveDamageAttr extends ModifiedDamageAttr {
     return Math.min(damage, target.hp - 1);
   }
 
-  getCondition(): MoveConditionFunc {
-    return (user, target, move) => target.hp > 1;
-  }
```
torranx, line 1266: Why is it removed completely? Isn't setting the condition to be `target.hp >= 1` better?

torranx:
Please also add tests

src/data/move.ts
schmidtc1: `getCondition` gives the possibility of failing, but we don't want it to fail. It should always hit, but just reduce damage so that HP doesn't go past 1.

torranx: Ok I've just read the actual move description

PR #3900, opened 2024-08-29
[Bug] skip-seen-dialogue storing strings instead of i18n keys

flx-sta:

Their previously stored seen dialogues will still apply correctly.

Skip Seen dialogues used to store the i18n keys but I've accidentally changed that to the actual string. Meaning if you change the language all dialogues are considered "not seen" 

I've made sure that skipSeenDialogues stores & checks the i18n keys instead of the actual resolved string

### Screenshots/Videos

https://github.com/user-attachments/assets/920d8ca4-105a-4884-8312-71c43aba78c1

## How to test the changes?
- Make sure to skip seen dialogues (settings -> general)
- Set starting wave to wave 5.
- Wait for the first dialogues
- reload
- start same game again
- YOu shouldn't see the dialogue this time
- change language
- reload
- same again same result -> shouldn't see it.
- Check the localstorage -> seenDialogues key

## Checklist
    - I tried but it was an absolute pain the a** as dialogues are skipped by default in tests
- [x] Are the changes visual?


CodeTappert (dismissed):

DayKev:
Loaded into wave 8 in Spanish -> all the dialog was blank.
When doing the same on beta, dialog shows as English (untranslated) rather than blank.

flx-sta:
> Loaded into wave 8 in Spanish -> all the dialog was blank. When doing the same on beta, dialog shows as English (untranslated) rather than blank.

Good catch.
this should fix it:
https://github.com/pagefaultgames/pokerogue/pull/3900/commits/ae34c686b115047662d86a8e33eae309a77096f0

frutescens:
Sent my fix in for the rival dialogue from game over

CodeTappert (dismissed):

Tempo-anon (approved):
Localization causing more issues :goomed:

src/phases/game-over-phase.ts
```diff
@@ -136,12 +136,16 @@ export class GameOverPhase extends BattlePhase {
...
 
-            if (!this.scene.ui.shouldSkipDialogue(message)) {
+            if (!this.scene.ui.shouldSkipDialogue(dialogueKey)) {
               this.scene.ui.fadeIn(500).then(() => {
-                this.scene.charSprite.showCharacter(`rival_${this.scene.gameData.gender === PlayerGender.FEMALE ? "m" : "f"}`, getCharVariantFromDialogue(miscDialogue.ending[this.scene.gameData.gender === PlayerGender.FEMALE ? 0 : 1])).then(() => {
-                  this.scene.ui.showDialogue(message, this.scene.gameData.gender === PlayerGender.FEMALE ? trainerConfigs[TrainerType.RIVAL].name : trainerConfigs[TrainerType.RIVAL].nameFemale, null, () => {
+                const genderIndex = this.scene.gameData.gender ?? PlayerGender.UNSET;
+                const genderStr = PlayerGender[genderIndex].toLowerCase();
+                // Dialogue has to be retrieved so that the rival's expressions can be loaded and shown via getCharVariantFromDialogue
+                const dialogue = i18next.t(dialogueKey, {context: genderStr});
```
torranx, line 146: nit
```suggestion
                const dialogue = i18next.t(dialogueKey, { context: genderStr });
```

torranx (approved):
lgtm, has minor nit

MokaStitcher:
all seems good now, the only weird thing is all these missing key messages in console when displaying dialog that uses multiple boxes like the for the rivals (I think it's because showDialogue is called for each bit of the dialogue with the actual text to display and not the key, and since you're using 
`const text = i18next.t(i18nKey, { context: genderStr });` outside of the if that checks if the key exists it displays this error

Screenshot 2024-08-30 at 18 12 50

torranx (approved):

flx-sta:
@MokaStitcher thanks for making me aware of that.
I've fixed the behavior for that

https://github.com/user-attachments/assets/679e661b-76c1-45a9-9b82-5c3dd147b3b7

torranx (approved):

MokaStitcher (approved):
all good now! 👍

frutescens (approved):

Tempo-anon (approved):

PR #3618, opened 2024-08-18
[Refactor] Break phases.ts up

f-fsantos:
## What are the changes the user will see?

None.

## Why am I making these changes?

Dev Experience.

## What are the changes from a developer perspective?

Changes to phases are now isolated.

## How to test the changes?

Play the game and explore the different phases.

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I considered writing automated tests for the issue?
- [] If I have text, did I add placeholders for them in locales?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [ ] Are the changes visual?
  - [ ] Have I provided screenshots/videos of the changes?

PR #3906, opened 2024-08-29
[Bug][Beta] Double vouchers fix

frutescens:
Hopefully, users will not get double golden vouchers.  

This bug is blocking beta. 

PR #1940 :  This fixed the issue of 1st Classic Clear voucher ids from being undefined. However, recent changes from #3618 led to this problem resurfacing, meaning that user accounts be considered to have two 1st Classic Clear vouchers - one with an actual id, the other with an undefined id. When title phase validated the vouchers, it would validate the newly created 'undefined' 1st Classic Clear id, which held the same information as the one as the actual id, and give the account another voucher.

This was fixed by moving initAchievements() above initVouchers() since initVouchers() depends on initAchievements(). Then, a conditional was added to the 1st classic clear achievement to prevent future validations of the achievement / associated voucher. This doesn't interfere with 1st classic clears. However, although this prevents vouchers from generating an 'undefined' 1st classic clear voucher again, those who were present during the accidental beta merge from last week may already have a valid 'undefined' one and still receive a second golden voucher.  

## How to test the changes?
Check the contents of vouchers -> console.log()
Check if a new user is able to clear classic and receive the achievement/voucher. 

## Checklist


flx-sta:
@frutescens I'm a bit confused what that `ordinals` code is doing inside this. Or is that intended?

DayKev:
He probably forgot to switch to `beta` before making the changes.

flx-sta:
> He probably forgot to switch to `beta` before making the changes.

image

frutescens:
Whoops! Sorry, I didn't realize I was another branch when I was making this PR and I had to run out immediately afterwards.

torranx (approved):

Tempo-anon (approved):

flx-sta (approved):

PR #1845, opened 2024-06-05
[Balance] Gives Gengar Levitate back, Cursed Body moved to Ability 2

damocleas:
## What are the changes?
Gengar given Levitate in Ability slot 1, Cursed Body moved to Ability slot 2.
Haunter and Gastly given Cursed Body on Ability slot 2 to match and be an easier to notice difference.

## Why am I doing these changes?
Because it was a requested feature, and is a revert to a nerf added in gen 7.

## What did change?
Gengar given Levitate in Ability slot 1, Cursed Body moved to Ability slot 2.
Haunter and Gastly given Cursed Body on Ability slot 2 to match.

### Screenshots/Videos
Cursed Body *Ability slot 2*
image
Cursed Body *Ability slot 2*
image
Levitate *Ability Slot 1*
image

## How to test the changes?
Start with a Levitate Gastly, evolve through to Gengar.
Set Ability 2 to test Cursed Body.

## Checklist
- [X] There is no overlap with another PR?
- [X] The PR is self-contained and cannot be split into smaller PRs?
- [X] Have I provided a clear explanation of the changes?
- [X] Have I tested the changes (manually)?
    - [X] Are all unit tests still passing? (`npm run test`)
- [X] Are the changes visual?
  - [X] Have I provided screenshots/videos of the changes?

PR #3910, opened 2024-08-30
[UI] [QoL] [Enhancement] Exclude redundant species from certain filters in starter select

Fontbane:

- In the starter select UI, when filtering for mons you don't have the hidden ability of, mons who don't have an HA at all will be excluded
- When filtering for mons you don't have passive or any cost reduction unlocked for yet, starters who don't have candy (Pikachu et al) will be excluded

- The "Hidden Ability- No" option is useful for seeing which mons you haven't found the hidden ability of
- It's redundant and possibly misleading to show the player mons with nonexistent hidden abilities alongside mons with hidden abilities the player just hasn't found
- Same applies even moreso to passives and cost reductions
- This makes the filters more useful for giving an overview of goals for the player

- Adds a speciesHasHiddenAbility check for the HA filter
- The condition for the exclude state now requires speciesHasHiddenAbility to be true as well as it not being unlocked
- Adds a isStarterProgressable check for a few other filters, whether that starter gets candy/egg moves/etc
- Adds an enum SortCriteria in dropdown.ts for the criteria by which the player can sort the starters

### Screenshots/Videos

Before:
Screenshot_868
Note the presence of Mew, which does not have a Hidden Ability at all.

After:
image
Note the absence of Mew between Mewtwo and Cyndaquil, as well as the rows starting with Drowzee appearing to have shifted one slot- Gastly has no hidden ability (unless #1845 is reconsidered), so Onix takes its place on that row and everything else shifts to the left.

Before:
Screenshot_871
Note that many starters without candy, such as the Hitmons, Chansey, Snorlax etc. are shown

After:
Screenshot_869
Bye bye

Before:
Screenshot_872
Same as passive

After:
Screenshot_870

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -39,13 +39,13 @@ import { Species } from "#enums/species";
 import {Button} from "#enums/buttons";
 import { EggSourceType } from "#app/enums/egg-source-types.js";
 import AwaitableUiHandler from "./awaitable-ui-handler";
-import { DropDown, DropDownLabel, DropDownOption, DropDownState, DropDownType } from "./dropdown";
+import { DropDown, DropDownLabel, DropDownOption, DropDownState, DropDownType, SortCriteria } from "./dropdown";
 import { StarterContainer } from "./starter-container";
 import { DropDownColumn, FilterBar } from "./filter-bar";
 import { ScrollBar } from "./scroll-bar";
 import { SelectChallengePhase } from "#app/phases/select-challenge-phase.js";
 import { TitlePhase } from "#app/phases/title-phase.js";
-import { Abilities } from "#app/enums/abilities.js";
+import { Abilities } from "#app/enums/abilities";
 
 export type StarterSelectCallback = (starters: Starter[]) => void;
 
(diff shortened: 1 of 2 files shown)


src/ui/starter-select-ui-handler.ts
```diff
@@ -45,6 +45,7 @@ import { DropDownColumn, FilterBar } from "./filter-bar";
 import { ScrollBar } from "./scroll-bar";
 import { SelectChallengePhase } from "#app/phases/select-challenge-phase.js";
 import { TitlePhase } from "#app/phases/title-phase.js";
+import { Abilities } from "#app/enums/abilities.js";
```
flx-sta, line 48:
```suggestion
import { Abilities } from "#app/enums/abilities";
```

flx-sta (changes requested):

DayKev (dismissed):

KimJeongSun (approved):
LGTM!

torranx (approved):

DayKev (approved):

flx-sta (approved):

Issue #3842, opened 2024-08-27
[Bug] Shedinja Doesn't Have Access to Nincada's Egg moves by Memory Mushroom/Relearn.


Snailman11:
### Describe the bug

Discord Bug Report: (Shedinja cannot learn egg moves from mushroom, even if nincada was brought from starter screen)
https://discord.com/channels/1125469663833370665/1277793919677235254

-Evolve a Nincada with egg moves unlocked into Ninjask and Shedinja at level 20, observe that Shedinja can't remember them with a Memory Mushroom

https://github.com/user-attachments/assets/bd743f31-328a-4a8a-9195-305a979f9e61

### Expected behavior

Shedinja is able to relearn Nincada's Egg moves

### What OS did you observe the bug on?

Other

### If other please specify

gameplay

### Which browser do you use?

Other

### If other please specify

gameplay

ckincaid17:
I'm going to work on this 🙂

PR #3915, opened 2024-08-30
[Bug] Let Shedinja relearn egg moves when evolved from a starter

ckincaid17:
After evolution, Shedinja's "met" biome and level will match that of the Nincada it evolved from. This will allow Shedinja to use Memory Mushrooms to relearn any egg moves the user has unlocked for Nincada, if Shedinja evolved from a starter Nincada.

https://github.com/pagefaultgames/pokerogue/issues/3842
This change makes Shedinja's relearn behavior the same as any other Pokemon evolved from a starter.

When Nincada evolves, the evolution is handled specially by creating a second Nincada in the user's party and setting many of its attributes to be the same as the original Nincada's. This PR makes it so the second Nincada also copies metadata about where and when the original one was met.

### Screenshots/Videos
Nincada before evolution:
Screen Shot 2024-08-31 at 12 23 14 PM

Shedinja being able to learn an egg move after evolution:

https://github.com/user-attachments/assets/811b518a-d71f-4e2f-820b-c52f6f358d0f

## How to test the changes?
 - Automated: Added a `metBiome` check in the Shedinja evolution test.
 - Manual: Evolve a Nincada under the standard conditions that lead to Shedinja (at least one empty spot in party and Pokeball in the bag) and verify that the Shedinja's met information on its summary screen matches what Nincada's was.

## Checklist
- [x] ~~If I have text, did I make it translatable and add a key in the English locale file(s)?~~
- [x] Are the changes visual?


src/field/pokemon.ts
```diff
@@ -3598,6 +3598,9 @@ export class PlayerPokemon extends Pokemon {
         newPokemon.moveset = this.moveset.slice();
         newPokemon.moveset = this.copyMoveset();
         newPokemon.luck = this.luck;
+        newPokemon.metLevel = this.metLevel;
```
ckincaid17, line 3601: Shedinja's gender is different from Nincada's in the recording I attached. Can copy that over too if we care.

DayKev: It seems Shedinja is supposed to be "gender-unknown"/genderless so not copying it over is correct.

DayKev (approved):

Tempo-anon (commented):
1. Technically Shedinja should be genderless. We should probably change that and fortunately I don't think that will cause many grayball bugs
2. This is technically a gameplay balance change so should get the balance team's opinion on it

ckincaid17:
1. Yeah for sure, I can fix Shedinja's gender in a follow-up PR. Would like to keep this one scoped to egg moves
2. This feels like a correctness + QoL fix imo (every other mon can learn its prevo's eggs moves), and using this quirk as an opportunity to balance Shedinja feels weird--plus you can get around this by just not deleting any egg moves until after the evolution--but I'll ask for a balance team review

Tempo-anon (approved):

PR #3918, opened 2024-08-30
[Localization] remove all `"<key>_female": null` entries in `locales/en`

flx-sta:

none

GitLocalize does not seem to like `null` values on keys!

I removed all `"<key>_female": null` entries in `locales/en`

### Screenshots/Videos

check the file changes 

## How to test the changes?

Just play basically. Especially dialogues.

## Checklist
- ~~[ ] Are the changes visual?~~
  - ~~[ ] Have I provided screenshots/videos of the changes?~~

### DIFF
--- a/src/locales/en/achv.json
+++ b/src/locales/en/achv.json
@@ -10,19 +10,19 @@
   },
   "10K_MONEY": {
     "name": "Money Haver",
-    "name_female": null
+    "name_female": "Money Haver"
   },
   "100K_MONEY": {
     "name": "Rich",
-    "name_female": null
+    "name_female": "Rich"
   },
   "1M_MONEY": {
     "name": "Millionaire",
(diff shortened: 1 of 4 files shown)


CodeTappert (approved):

AsdarDevelops (approved):
This adds redundancy to English keys by adding female as well when the string is gender neutral in English. Doing it this way rather than with "null" allows GitLocalize to bring this to translators as different things to translate.

It's also very nice to use in GitLocalize as the order in which these are shown makes the copy paste (for the cases where gender doesn't alter the string) much simpler than the old system with two differentiated files.

Tempo-anon (approved):

frutescens (approved):

flx-sta:
Waiting to be merged AFTER the next release

PR #3936, opened 2024-08-31
[Bug] Fix some trapping moves' interactions with Ghost-type Pokemon

### LINKED ISSUE #1915: [BUG] Infestation incorrectly interacts with Ghost-type Pokémon
**Describe the bug**
Infestation when used on a Ghost-type Pokémon deals damage, but its secondary effect is ignored.

**To Reproduce**
1. Send a Pokémon with Infestation vs a Ghost-type
2. Click on 'Infestation'
3. See error

**Expected behavior**
Infestation should deal damage and do damage over time to the Ghost-type. The reason this effect is ignored is because it also has a trapping effect, which is ignored by Ghost-types. The trapping is correctly ignored, but the per-turn damage should still take effect.

**Device**
PC (Firefox)

innerthunder:

This fixes two issues related to how certain trapping moves interact with Ghost-type Pokemon:
1. If a Ghost-type Pokemon is hit with a binding move (e.g. Infestation), it is now correctly afflicted with the damaging effect (while still able to switch or run), rather than cancelling the move's secondary effects altogether.
2. Ghost-type Pokemon are no longer able to use No Retreat multiple times in succession.

closes #1915 

- `data/battler-tags`: New `NoRetreatTag` that extends `TrappedTag` but removes the non-Ghost-type condition for adding. This condition is also removed from `DamagingTrapTag`, as this condition can instead be found in `Pokemon#isTrapped`
- `data/move`: Changed No Retreat to make use of its new tag type.
- `field/pokemon`: Added an `isTrapped` helper function to `Pokemon` which returns `true` when the `Pokemon` is under the effects of a trapping move or ability. This supports simulated calls to Shadow Tag, etc. and can forward ability trigger messages to UI elements via the `trappedAbMessages` argument.
- `phases/command-phase` and `phases/enemy-command-phase`: Rewrote checks for trapping using `Pokemon#isTrapped`

### Screenshots/Videos

**Showing Infestation's effects on a Ghost-type vs. on a non-Ghost-type:**

https://github.com/user-attachments/assets/fe9016af-7811-4755-9643-518f91e90cbb

**Showing No Retreat fails after the first use by a Ghost-type (but the Ghost-type can still switch):**

https://github.com/user-attachments/assets/e80da678-c5e2-41a9-9dde-30644e56970c

## How to test the changes?

To reproduce the video tests, add the following overrides to `src/overrides.ts`:
```ts
const overrides = {
  STARTER_SPECIES_OVERRIDE: Species.GENGAR,
  MOVESET_OVERRIDE: [Moves.SPLASH, Moves.NO_RETREAT],
  OPP_MOVESET_OVERRIDE: Array(4).fill(Moves.INFESTATION)
```

## Checklist
- [x] Are the changes visual?


Tempo-anon (approved):

src/phases/command-phase.ts
```diff
@@ -1,10 +1,8 @@
 import BattleScene from "#app/battle-scene.js";
```
DayKev, line 1: Since you're here, can you remove the `.js` from all the imports?

src/phases/enemy-command-phase.ts
```diff
@@ -1,9 +1,6 @@
 import BattleScene from "#app/battle-scene.js";
 import { BattlerIndex } from "#app/battle.js";
```
DayKev, line 2:
```suggestion
import BattleScene from "#app/battle-scene";
import { BattlerIndex } from "#app/battle";
```

src/phases/enemy-command-phase.ts
```diff
@@ -1,9 +1,6 @@
...
-import { applyCheckTrappedAbAttrs, CheckTrappedAbAttr } from "#app/data/ability.js";
-import { TrappedTag } from "#app/data/battler-tags.js";
 import { Command } from "#app/ui/command-ui-handler.js";
```
DayKev, line 3:
```suggestion
import { Command } from "#app/ui/command-ui-handler";
```

PR #3943, opened 2024-08-31
[Misc] New Utils function for Beta/Local logging

frutescens:
None. 

This will be helpful in case we want to test things on beta/local and check output/contents without interfering with the live experience. 

New Utils function that checks if the user is on beta or local before outputting a console.log. Function name up for debate.

## How to test the changes?
Use the function in Beta or Local to see console output. Change conditional to be false to not see console output.

## Checklist


DayKev:
`console.log()` statements are already ignored on main, this doesn't actually do anything lol.

PR #3952, opened 2024-09-01
[Bug] Fix reloads causing RNG inconsistencies

PigeonBar:

Reloads should no longer cause RNG inconsistencies.

RNG inconsistencies are not ideal, especially for Daily Runs, where all players should experience the same RNG results. This fix is meant as a quick, temporary fix until Xavion's planned RNG rework.

This "quick fix" simply calls `scene.resetSeed()` at a consistent point at the beginning of each encounter, after the opponent team has been randomly generated, so that the RNG is reset to a consistent state regardless of whether the session was reloaded.
This PR also adds some tools for performing reloads in unit tests.
Also, some further comments:

1. Due to the new RNG resets, some battle mechanics will now share the exact same RNG calls as opponent Pokémon generation, which could lead to issues of the form "Covet is more effective against Species A than Species B". If this is too problematic, we could also try generating the opponent team using `scene.executeWithSeedOffset()` (which isn't ideal, but Xavion's RNG rework will fix this).

2. If anyone knows a more "proper" way to perform reloads in unit tests, please feel free to add it to this PR.

3. (Edit: This PR now also changes Moody and Acupressure to use seeded RNG instead of unseeded RNG.)

### Before the changes:

https://github.com/user-attachments/assets/51c70ac5-6a57-4566-b173-de8ae8970fba

Lapras steals 0 items before the reload and then steals 3 items after the reload, which is an unwanted inconsistency.

### After the changes:

https://github.com/user-attachments/assets/bf3e7319-d8c9-44ec-bbde-83235ed5b593

Both before and after the reload, Lapras steals a Sharp Beak and a Liechi Berry on the same hits, which shows RNG consistency.

## How to test the changes?

1. Load a file, and defeat the opponents to progress to the next wave without reloading.
2. Spam Covet/Thief/Dire Claw/another move that calls RNG.
3. Record the RNG results from Step 2.
4. Save and quit, then reload the session.
5. Spam Covet/Thief/Dire Claw/another move that calls RNG.
6. Record the RNG results from Step 5.
7. The RNG results from Steps 3 and 6 should be the same as each other.

## Checklist
- [x] Are the changes visual?


src/test/reload.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    const preReloadRngState = Phaser.Math.RND.state();
+
+    await game.reload.reloadSession();
+
+    const postReloadRngState = Phaser.Math.RND.state();
+
+    expect(preReloadRngState).toBe(postReloadRngState);
+  }, 20000);
+
+  it("should not have RNG inconsistencies after a biome switch", async () => {
```
DayKev, line 40: Biome swaps after every 10th wave in Classic.

src/test/reload.test.ts
```diff
@@ -0,0 +1,137 @@
+import { Species } from "#app/enums/species.js";
+import { GameModes, getGameMode } from "#app/game-mode.js";
+import GameManager from "#test/utils/gameManager";
+import { afterEach, beforeAll, beforeEach, describe, expect, it } from "vitest";
+import { SPLASH_ONLY } from "./utils/testUtils";
+import { Moves } from "#app/enums/moves.js";
+import { EnemyCommandPhase } from "#app/phases/enemy-command-phase.js";
+import { DamagePhase } from "#app/phases/damage-phase.js";
```
DayKev, line 8:
```suggestion
import { Moves } from "#app/enums/moves";
import { Species } from "#app/enums/species";
import { GameModes, getGameMode } from "#app/game-mode";
import GameManager from "#test/utils/gameManager";
import { afterEach, beforeAll, beforeEach, describe, expect, it } from "vitest";
import { SPLASH_ONLY } from "./utils/testUtils";
```
Removed the `.js` and now-unused imports (also my IDE re-ordered them but w/e).

src/test/reload.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      .enemyLevel(1000)
+      .disableTrainerWaves()
+      .moveset([Moves.KOWTOW_CLEAVE])
+      .enemyMoveset(SPLASH_ONLY);
+    await game.classicMode.startBattle();
+    game.scene.gameMode = getGameMode(GameModes.ENDLESS);
+
+    // Transition from Wave 15 to Wave 16 in order to trigger biome switch
+    game.move.select(Moves.KOWTOW_CLEAVE);
+    await game.phaseInterceptor.runFrom(EnemyCommandPhase).to(DamagePhase);
```
DayKev, line 54:
```suggestion
    await game.phaseInterceptor.to("DamagePhase");
```
You don't need to use `runFrom()`.

src/test/reload.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    const preReloadRngState = Phaser.Math.RND.state();
+
+    await game.reload.reloadSession();
+
+    const postReloadRngState = Phaser.Math.RND.state();
+
+    expect(preReloadRngState).toBe(postReloadRngState);
+  }, 20000);
+
+  it("should not have RNG inconsistencies at a Daily run double battle", async () => {
```
DayKev, line 79: All of these tests past the first one seem redundant, what code are you testing that's not already tested by the one above this?

src/test/utils/phaseInterceptor.ts
```diff
@@ -16,6 +16,7 @@ import { MoveEffectPhase } from "#app/phases/move-effect-phase";
 import { MoveEndPhase } from "#app/phases/move-end-phase";
 import { MovePhase } from "#app/phases/move-phase";
 import { NewBattlePhase } from "#app/phases/new-battle-phase";
+import { NewBiomeEncounterPhase } from "#app/phases/new-biome-encounter-phase.js";
```
DayKev, line 19:
```suggestion
import { NewBiomeEncounterPhase } from "#app/phases/new-biome-encounter-phase";
```

src/test/reload.test.ts
PigeonBar: Some of my attempted fixes failed specifically after a biome switch (the 2nd test) and at trainer fights (the 5th test) due to them calling RNG slightly differently. Also, the baseline Daily Run test (the 3rd test) is useful because RNG inconsistencies are mainly a concern for Daily Runs. The other tests (including the Classic mode one) are "just in case" tests and can be removed if desired.

src/test/reload.test.ts
PigeonBar: For the biome switch test, I tried Classic 10 -> 11 but had issues with getting past the "level cap increased" prompt, so I am instead using Endless 15 -> 16.

PigeonBar: Update: I got a Wave 10 -> 11 test to work.

PigeonBar:
By the way, is there a reason why `getSessionSaveData()` from `game-data.ts` is private? It would probably be better for `reloadHelper.ts` to be able to call `getSessionSaveData()` directly from `game-data.ts` so that the reload helper can automatically account for any new information stored in the session data (e.g., the upcoming Mystery Encounter stuff).

DayKev:
Dunno, that one could probably be made public.

src/test/reload.test.ts
DayKev: For now it's fine. There's some issue with memory leaks in tests so I'm trying to identify redundant tests that can be skipped if it becomes a problem again.

PigeonBar:
`getSessionSaveData()` is now public.

Thank you for all your feedback so far!

DayKev (dismissed):

Tempo-anon:
Moody and acupressure should be using seeded rng

DayKev:
> Moody and acupressure should be using seeded rng

All tests use seeded RNG by default (with the same seed).

PigeonBar:
> Moody and acupressure should be using seeded rng

Thanks, I changed them to use seeded RNG.

DayKev (dismissed):

Tempo-anon:
Thanks for fixing moody and acupressure, but now you have some conflicts in data/ability and data/move that need to be resolved

Tempo-anon (approved):

PR #3957, opened 2024-09-01
[Enhancement/Item] Separate form change items into non-rare and rare

Madmadness65:

This PR splits the form change items into a non-rare group and rare group, similar to how evolution items are split. Both normal and rare form change items currently have the same chance of appearing within their respective tiers (Ultra and Rogue).

Also as a minor inclusion "Blank Memory" has been renamed to "Normal Memory" to be in line with the naming conventions of the other memories, but this item is still unused anyways so it doesn't make a big difference.

Certain form changing items, notably Arceus' Plates, Silvally's Memories, and Genesect's Drives, are not of the same strength as  most other form change items, such as Mega Stones or Max Mushrooms. As such, it was decided that they should be made more common than those other items.

Like evolution items, form change items are now filtered as rare and not rare based on the ID number of the item. Anything less than ID number 100 is considered "rare", while anything above is not.

**It is important to note that the form change items not only got shifted around due to this (causing existing saves to potentially have the wrong form change items now), but some IDs (specifically 65 through 77) no longer exist due to some items being moved to above ID number 100. Ideally this should be the only time a shift this major happens with the form change items.**

### Screenshots/Videos

image
An example of a Plate now being in Ultra tier.

image
An example of an unchanged form change item still being in Rogue tier.
## Checklist
- [x] Are the changes visual?


src/modifier/modifier-type.ts
```diff
@@ -1083,7 +1083,8 @@ class FormChangeItemModifierTypeGenerator extends ModifierTypeGenerator {
           }
         }
         return formChangeItemTriggers;
-      }).flat().flatMap(fc => fc.item))];
+      }).flat())
+      ].flat().flatMap(fc => fc.item).filter(i => (!!i && i < 100) === rare);
```
DayKev, line 1087: Is the `!!` necessary? Shouldn't it already be coerced to a boolean?

Madmadness65: Tested it without the `!!`, and as far as I can tell it still works correctly. It was just a case of following what was done in the evolution item pool code, and since it did work as intended, I didn't think to change it.

DayKev (approved):

PR #3964, opened 2024-09-02
[Balance] Double base shiny odds, adjusted Shiny Charm to match

DayKev:
More shinies!

image

Shiny threshold doubled from 32 to 64. Shiny Charm effectiveness reduced by half to compensate.

## How to test the changes?
Play the game.

## Checklist
- ~[ ] Are the changes visual?~ yes but no

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1511,7 +1511,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
    *
    * The exact mechanic is that it calculates E as the XOR of the player's trainer ID and secret ID
    * F is calculated as the XOR of the first 16 bits of the Pokemon's ID with the last 16 bits
-   * The XOR of E and F are then compared to the thresholdOverride (default case 64) to see whether or not to generate a shiny
+   * The XOR of E and F are then compared to the `shinyThreshold` (or `thresholdOverride` if set) to see whether or not to generate a shiny
    * @param thresholdOverride number that is divided by 2^16 (65536) to get the shiny chance
    * @returns true if the Pokemon has been set as a shiny, false otherwise
    */
@@ -1527,7 +1527,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     const E = this.scene.gameData.trainerId ^ this.scene.gameData.secretId;
     const F = rand1 ^ rand2;
 
-    const shinyThreshold = new Utils.IntegerHolder(64);
+    /** 64/65536 -> 1/1024 */
+    const baseShinyChance = 64;
+    const shinyThreshold = new Utils.IntegerHolder(baseShinyChance);
     if (thresholdOverride === undefined) {
       if (this.scene.eventManager.isEventActive()) {
         shinyThreshold.value *= this.scene.eventManager.getShinyMultiplier();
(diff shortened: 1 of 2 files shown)


damocleas (approved):
damo vs korea part 17 postponed

src/field/pokemon.ts
```diff
@@ -1511,7 +1511,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
    *
    * The exact mechanic is that it calculates E as the XOR of the player's trainer ID and secret ID
    * F is calculated as the XOR of the first 16 bits of the Pokemon's ID with the last 16 bits
-   * The XOR of E and F are then compared to the thresholdOverride (default case 32) to see whether or not to generate a shiny
+   * The XOR of E and F are then compared to the thresholdOverride (default case 64) to see whether or not to generate a shiny
```
Tempo-anon, line 1514: We should probably refactor this so there isn't just a magic number floating around

src/modifier/modifier.ts
```diff
@@ -2193,7 +2193,7 @@ export class ShinyRateBoosterModifier extends PersistentModifier {
   }
 
   apply(args: any[]): boolean {
-    (args[0] as Utils.IntegerHolder).value *= Math.pow(2, 2 + this.getStackCount());
+    (args[0] as Utils.IntegerHolder).value *= Math.pow(2, 1 + this.getStackCount());
```
Tempo-anon, line 2196: This change should also be reflected in the PR title and description

src/field/pokemon.ts
DayKev:
```ts
  /**
   * Function that tries to set a Pokemon shiny based on the trainer's trainer ID and secret ID
   * Endless Pokemon in the end biome are unable to be set to shiny
   *
   * The exact mechanic is that it calculates E as the XOR of the player's trainer ID and secret ID
   * F is calculated as the XOR of the first 16 bits of the Pokemon's ID with the last 16 bits
+  * The XOR of E and F are then compared to the `shinyThreshold` (or `thresholdOverride` if set) to see whether or not to generate a shiny
   * @param thresholdOverride number that is divided by 2^16 (65536) to get the shiny chance
   * @returns true if the Pokemon has been set as a shiny, false otherwise
   */
  trySetShiny(thresholdOverride?: integer): boolean {
    // Shiny Pokemon should not spawn in the end biome in endless
    if (this.scene.gameMode.isEndless && this.scene.arena.biomeType === Biome.END) {
      return false;
    }

    const rand1 = (this.id & 0xFFFF0000) >>> 16;
    const rand2 = (this.id & 0x0000FFFF);

    const E = this.scene.gameData.trainerId ^ this.scene.gameData.secretId;
    const F = rand1 ^ rand2;

+   /** 64/65536 -> 1/1024 */
+   const baseShinyChance = 64;
+   const shinyThreshold = new Utils.IntegerHolder(baseShinyChance);
    if (thresholdOverride === undefined) {
      if (this.scene.eventManager.isEventActive()) {
        shinyThreshold.value *= this.scene.eventManager.getShinyMultiplier();
      }
      if (!this.hasTrainer()) {
        this.scene.applyModifiers(ShinyRateBoosterModifier, true, shinyThreshold);
      }
    } else {
      shinyThreshold.value = thresholdOverride;
    }

    this.shiny = (E ^ F) < shinyThreshold.value;
+   if ((E ^ F) < baseShinyChance) {
      console.log("REAL SHINY!!");
    }

    if (this.shiny) {
      this.initShinySparkle();
    }

    return this.shiny;
  }
```
Like this?

DayKev: 1: 👍 
2: Doesn't look like it, I think `NumberHolder` is required for the `applyStuffAndThings(ModifierClass, ...)` functions to work.
3: Probably not, I'll remove it.

Tempo-anon (approved):

PR #3972, opened 2024-09-02
 [Bug] Fix sleep move failure text not appearing on Electric Terrain

jessemchung:
This fix is for this issue:
https://github.com/pagefaultgames/pokerogue/issues/3722

The user will now, when attempting to put a pokemon to sleep on electric terrain, will receive a message why it failed

I needed to add fixes to drowsy (from yawn) as well as to the tryStatusEffect 

Makes it more similar to the mainline games and pokemon showdown

A simple change was added in the drowsy state check so that the message will appear.  I also had to add logic to the final

### Screenshots/Videos
See below

## How to test the changes?

I used override, I honestly am not sure how best to test yet.  I set my pokemon to Tapu Koko with spore and yawn as moves and the opponent as well to test.   Below is the basic override to see things. 
```
  readonly STARTER_SPECIES_OVERRIDE: Species | number = Species.TAPU_KOKO;
  readonly ABILITY_OVERRIDE: Abilities = Abilities.NONE;
  readonly PASSIVE_ABILITY_OVERRIDE: Abilities = Abilities.NONE;
  readonly STATUS_OVERRIDE: StatusEffect = StatusEffect.NONE;
  readonly GENDER_OVERRIDE: Gender | null = null;
  readonly MOVESET_OVERRIDE: Array<Moves> = [Moves.SPORE, Moves.YAWN];
  readonly SHINY_OVERRIDE: boolean = false;
  readonly VARIANT_OVERRIDE: Variant = 0;
```

## Checklist
- I set the target branch to beta
- Not sure how best to look at this but I confirmed in the issue that I would fix it
- Yes
- Yes
- I considered it but decided no for now
- Text was already there.  Just no triggering logic
-  Yes
    - No?  But I honestly am not sure how my code might be causing issues.  
    

> FAIL  |main| src/test/moves/flower_shield.test.ts > Moves - Flower Shield > does not increase defense of a pokemon in semi-vulnerable state
> AssertionError: expected +0 to be 1 // Object.is equality
> 
> - Expected
> + Received
> 
> - 1
> + 0
> 

- [x] Are the changes visual?
  - Yes

Original Issue on display

https://github.com/user-attachments/assets/9658fa79-3638-4274-9a0a-b6f48862a5ec

Fixes both with spore and sing to showcase 

https://github.com/user-attachments/assets/92bb4ba5-5a34-4019-8069-9d0a88a61027

https://github.com/user-attachments/assets/044916ef-c235-4aaa-b343-54e8c079406c

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -2861,10 +2861,10 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
    * @param quiet a boolean value that is sometimes used as the asPhase boolean which in this case affects that.
    * @param overrideStatus a boolean value to determine whether a status should be overriden.  If true, the pokemon cannot be statused essentially
    * @param sourcePokemon {@linkcode Pokemon} The pokemon causing the status.  The target pokemon is 'this'.
-   * @returns true if the pokemon can be affected by the status, false if not.
+   * @returns true if the pokemon can be affected by the status, false if not.  Also, while not a return value, it will apply applyPreSetStatusAbAttrs which means this function should be used sparingly compared to checkIfCanSetStatus
    */
   canSetStatus(effect: StatusEffect | undefined, quiet: boolean = false, overrideStatus: boolean = false, sourcePokemon: Pokemon | null = null): boolean {
-    const canSetStatus = this.checkIfCanSetStatus(effect, quiet, overrideStatus, sourcePokemon);
+    const canSetStatusResult = this.checkIfCanSetStatus(effect, quiet, overrideStatus, sourcePokemon);
 
     const cancelled = new Utils.BooleanHolder(false);
     applyPreSetStatusAbAttrs(StatusEffectImmunityAbAttr, this, effect, cancelled, quiet);
@@ -2876,7 +2876,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       return false;
     }
 
-    return canSetStatus;
+    return canSetStatusResult;
   }
 
   trySetStatus(effect: StatusEffect | undefined, asPhase: boolean = false, sourcePokemon: Pokemon | null = null, cureTurn: integer | null = 0, sourceText: string | null = null): boolean {
--- a/src/data/dialogue.ts
+++ b/src/data/dialogue.ts
@@ -1,6 +1,6 @@
 import { BattleSpec } from "#enums/battle-spec";
 import { TrainerType } from "#enums/trainer-type";
-import {trainerConfigs} from "./trainer-config";
+import { trainerConfigs } from "./trainer-config";
 
 export interface TrainerTypeMessages {
     encounter?: string | string[],
@@ -704,6 +704,20 @@ export const trainerTypeDialogue: TrainerTypeDialogue = {
         "dialogue:zinzolin.victory.1",
         "dialogue:zinzolin.victory.2",
         "dialogue:zinzolin.victory.3",
+      ]
+    }
+  ],
+  [TrainerType.ROOD]: [
+    {
+      encounter: [
+        "dialogue:rood.encounter.1",
+        "dialogue:rood.encounter.2",
+        "dialogue:rood.encounter.3",
+      ],
+      victory: [
+        "dialogue:rood.victory.1",
+        "dialogue:rood.victory.2",
+        "dialogue:rood.victory.3",
       ]
     }
   ],


src/field/pokemon.ts
```diff
@@ -2841,6 +2852,19 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+  /**
+   * Checks if it is possible for a pokemon to be afflicted with the given status
+   * @param effect {@linkcode StatusEffect} the status effect being checked.
+   * @param quiet a boolean value that is sometimes used as the asPhase boolean which in this case affects that.
+   * @param overrideStatus a boolean value to determine whether a status should be overriden.  If true, the pokemon cannot be statused essentially
+   * @param sourcePokemon {@linkcode Pokemon} The pokemon causing the status.  The target pokemon is 'this'.
+   * @returns true if the pokemon can be affected by the status, false if not.
+   */
+  canSetStatus(effect: StatusEffect | undefined, quiet: boolean = false, overrideStatus: boolean = false, sourcePokemon: Pokemon | null = null): boolean {
+    const canSetStatus = this.checkIfCanSetStatus(effect, quiet, overrideStatus, sourcePokemon);
```
Tempo-anon, line 2867: nit: I like the name but now it is the same as the function's. Maybe `canSetStatusResult`?

jessemchung: Makes sense.  Changed to canSetStatusResult

src/field/pokemon.ts
```diff
@@ -2775,7 +2775,15 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
-  canSetStatus(effect: StatusEffect | undefined, quiet: boolean = false, overrideStatus: boolean = false, sourcePokemon: Pokemon | null = null): boolean {
+  /**
+   * Checks if it is possible for a pokemon to be afflicted with the given status
+   * @param effect {@linkcode StatusEffect} the status effect being checked.
+   * @param quiet a boolean value that is sometimes used as the asPhase boolean which causes unshifting the phase.
+   * @param overrideStatus a boolean value to determine whether a status should be overriden.  If true, the pokemon cannot be statused essentially
+   * @param sourcePokemon {@linkcode Pokemon} The pokemon causing the status.  The target pokemon is 'this'.
+   * @returns true if the pokemon can be affected by the status, false if not.
+   */
+  checkIfCanSetStatus(effect: StatusEffect | undefined, quiet: boolean = false, overrideStatus: boolean = false, sourcePokemon: Pokemon | null = null): boolean {
```
DayKev, line 2786: Why did you split this into 2 functions?

jessemchung: So long story short, the issue was trying to get the message to only show up when it is supposed to.  This led me to try investigating quiet as an option but quiet has to be set to false in the place I needed the message to show up (move.ts line 1965).  This is because quiet is being used as asPhase I believe.

Furthermore this part of that code proved problematic as well as this part prevented me from running the function more than once so I couldn't just run a check when needed.
```
    applyPreSetStatusAbAttrs(StatusEffectImmunityAbAttr, this, effect, cancelled, quiet);

    const userField = this.getAlliedField();
    userField.forEach(pokemon => applyPreSetStatusAbAttrs(UserFieldStatusEffectImmunityAbAttr, pokemon, effect, cancelled, quiet));

```

One option I considered was adding a brand new propety to canSetStatus (something like showMessages) but that seemed unnecessary and more importantly @schmidtc1 mentioned attempting to do something like this a while ago but it not being successful.  It seemed like for future proofing, splitting that function into two pieces would be better as that way, if checking if a pokemon was statusable was needed somewhere, it could be done easily

jessemchung: Also, I just retested to make sure.  For some reason, it just doesn't work when it is the message is in canSetStatus. if I try to put

`      this.scene.queueMessage(i18next.t("terrain:defaultBlockMessage", { pokemonNameWithAffix: getPokemonNameWithAffix(this), terrainName: getTerrainName(TerrainType.ELECTRIC) }));`

Inside canSetStatus it just doesn't show the message (I think it has something to do with the fact that that function is called multiple times).  That's when schmidtc1 pointed out earlier attempts at fixing this..

DayKev: 🍝 strikes again...

This function needs a better name than just being a longer version of the other function's name.

flx-sta (changes requested):
Something went wrong with creating the branch.
There is more commits showing up than expected.
My assumption would be, that you branched off of `main` instead of `beta`

jessemchung:
> Something went wrong with creating the branch. There is more commits showing up than expected. My assumption would be, that you branched off of `main` instead of `beta`

@fix-sta

I think that was what I originally did and I have had trouble getting it right since but the changes should be correct with regards to comparison to beta.  Is there something else I should do to fix this?

src/field/pokemon.ts
DayKev: Seems good enough for now.

flx-sta:
I'm putting this into draft so you can resolve the branch issues

SirzBenjie:
I am going to close this since it is very old and does not seem to have activity. Feel free to reopen if there is desire to revisit this.

PR #3973, opened 2024-09-02
[Balance] Changed escape calculation

Opaque02:
At the moment, the escape chance is kinda random - there's times where when you try to escape and fail, and while it makes sense for the more escape chances to increase your odds of escaping, with the current formula it could actually _drop_ your chances of escaping

The current escape mechanics are kinda random and not great - this PR aims to fix it as per [this thread](<https://discord.com/channels/1125469663833370665/1279772765192454145/1279772767499063316>), [this post](<https://discord.com/channels/1125469663833370665/1279772765192454145/1280034335537500193>) and [this post](<https://discord.com/channels/1125469663833370665/1279772765192454145/1280040821957328907>) by Mega

For reference, the current formula is based on gen 3 and 4:

image

With this formula, because of the mod 256 at the end, it means that the escape chance could vary wildly depending on how many escape chances you have, as well as the speed ratio - your speed divided by the enemy's speed. This can be seen below:

image

The above shows the escape chance (as a mod 256) for a fixed speed ratio but for a changing number of rerolls. You can see that at 9 rerolls you have an escape chance of almost 256, but at reroll 10, it's quite close to 0, which feels bad to the player. Conversely, the below is for a fixed number of rerolls but a changing speed ratio:

image

As you can see above, at specific speed ratios, this drops very suddenly to 0. It's worth noting that because the formula has the speed ratio * 128 and then mod 256, this means that every multiple of 256, the odds reset. This is why the graphs above only go from 0 to 2; once they go past 2, they repeat.

Also worth noting for the above, the shape and values are the same for all the values of speed ratios/rerolls etc, but the starting point may shift, but it doesn't change the outcomes, just _when_ those outcomes occur.

The new formula does _not_ follow mainline games, but has been approved by Megalith for PR.

The escape chance formula has been changed. Before telling how, there's some backstory to it.

Firstly, there is a different interaction for bosses vs non bosses. It is much harder to escape from a boss compared to a non boss. There's a maximum escape chance, but the value is much smaller for bosses. The higher your speed ratio, the higher your escape chance. The speed ratio is calculated as the total speed of all your pokemon on the field that are alive divided by the total speed of enemy pokemon on the field that are alive. For example, if you have two pokemon with 40 and 60 speed, and the enemy has 2 pokemon with 70 and 80 speed, your total speed is 100 and the total enemy's speed is 150. If your 40 speed pokemon dies, you now have 60 speed vs the enemy's 150; however if the enemy's 80 speed pokemon died instead you now have 100 speed vs the enemy's 70.

The higher your total speed ratio the higher the chance to escape, but it caps out at a specific speed ratio value; this is discussed below. Also, you get a bonus per previously failed escape attempt. The full rules are below:

- There's a minimum of a 5% escape chance. 
- If you're fighting a non-boss pokemon, the max escape chance is 95%; if you're fighting a boss pokemon the max escape chance is 25%
- The base escape chance (if you and the enemy have the same speed) is 28% for non boss pokemon and 8% for boss pokemon
- The speed ratio cap (the amount of your speed you can have before it stops benefiting you) is 4x for non boss pokemon and 6x for boss pokemon
- If at least one pokemon is a boss pokemon, you use the values for the boss calculation
- The bonus for previous escape attempts is 10% per attempt for non boss pokemon vs 2% per attempt for boss pokemon

If you'd like to see it as a table, this is it:

image

These can also be calculated using a formula. For non boss pokemon it can be calculated using:

image

For boss pokemon, it can be calculated by the following:

image

Where the speedRatio is the sum of speed for all of your alive pokemon on the field divided by the sum of the speed for all enemies on the field, and the previousAttempts is the total number of previously failed attempts. 

Once this escapeChance has been calculated, it is rounded to the nearest integer and then capped at a minimum of 5% and a maximum of 25% for bosses or 95% for non bosses

### Screenshots/Videos
(none)

## How to test the changes?
Download this PR and make sure running away still works

## Checklist
- [ ] Are the changes visual?


frutescens (dismissed):
What a great write-up! Love the charts

src/phases/attempt-run-phase.ts
```diff
@@ -53,4 +56,40 @@ export class AttemptRunPhase extends PokemonPhase {
...
+
+    let isBoss = false;
+    for (let e = 0; e < enemyField.length; e++) {
+      isBoss = isBoss || enemyField[e].isBoss(); // this line checks if any of the enemy pokemon on the field are bosses; if so, the calculation for escaping is different
+    }
+
+    const speedRatio = playerSpeed / enemySpeed; // this gets the speed ration between you and the average enemy speed
+    const speedCap = isBoss ? 6 : 4; // this gets the speed cap depending whether there's a boss or not in the enemy field
+    const minChance = 5; // this is the minimum chance to escape
+    const maxChance = isBoss ? 25 : 95; // this gets the max escape chance depending whether there's a boss or not in the enemy field
```
damocleas, line 88:
```suggestion
    const maxChance = isBoss ? 25 : 95; // this gets the max escape % chance depending whether there's a boss or not in the enemy field
```

src/phases/attempt-run-phase.ts
```diff
@@ -53,4 +56,40 @@ export class AttemptRunPhase extends PokemonPhase {
...
+     */
+
+    let isBoss = false;
+    for (let e = 0; e < enemyField.length; e++) {
+      isBoss = isBoss || enemyField[e].isBoss(); // this line checks if any of the enemy pokemon on the field are bosses; if so, the calculation for escaping is different
+    }
+
+    const speedRatio = playerSpeed / enemySpeed; // this gets the speed ration between you and the average enemy speed
+    const speedCap = isBoss ? 6 : 4; // this gets the speed cap depending whether there's a boss or not in the enemy field
+    const minChance = 5; // this is the minimum chance to escape
```
damocleas, line 87:
```suggestion
    const minChance = 5; // this is the minimum % chance to escape
```

damocleas (approved):

damocleas (approved):

DayKev:
Dangit damo I was converting the comments to tsdoc comments (and fixing stuff like that) and now 
I have to deal with local merge conflicts :despair:

DayKev (approved):

damocleas:
> Dangit damo I was converting the comments to tsdoc comments (and fixing stuff like that) and now I have to deal with local merge conflicts :despair:

goobed

DayKev:
@damocleas you are also a goob because you forgot to re-approve the PR :pikastare:

innerthunder (approved):

frutescens (approved):

PR #3987, opened 2024-09-02
[UI] Run history fixes

frutescens:
- MonoType Challenge Rules will be colored by the type's color
- Challenge Rules will automatically wrap around
- Held Item Modifiers in non-Endless modes will be moved slightly upwards to avoid passing borders
- Held Item Modifier Stacks should be the correct color? 

My feature. 
I think the UI should be adaptive so all users can enjoy this feature. 

Word wrap added to the Phaser Text Object. 
Color/Shadow added to MonoType variable. 
Boolean changed for retrieving modifier icon. 

### Screenshots/Videos
Challenge Rules 
image
image
image

Held Items:
image

## How to test the changes?
Have more than 3 challenges active.
Have more than 18 unique held items on a Pokemon in a non-endless/endless spliced mode

## Checklist
- [x] Are the changes visual?


src/ui/run-info-ui-handler.ts
```diff
@@ -15,6 +15,7 @@ import { Challenges } from "#enums/challenges";
 import { getLuckString, getLuckTextTint } from "../modifier/modifier-type";
 import RoundRectangle from "phaser3-rex-plugins/plugins/roundrectangle.js";
 import { Type, getTypeRgb } from "../data/type";
+import { TypeColor, TypeShadow } from "#app/enums/color.js";
```
DayKev, line 18: Could you remove the `.js` from the imports while you're here?

DayKev (approved):

PR #4001, opened 2024-09-03
Starter select defaults to shiny with highest variant again

MokaStitcher:
Shiny with highest variant being enabled by default on starter select is back!

Since #3373, the starter UI has defaulted to Pokemon being non shiny, and the lowest unlocked variant being selected by default when enabling shiny.
This has caused some frustration and confusion for players especially since the preferred setting is kept only locally, so switching device, deleting cookies or getting a new variant requires to change the setting manually each time. 

* updated `getCurrentDexProps` to return the highest unlocked variant
* updated the logic when enabling / disabling shiny state
* updated starter select test to account for the change

### Screenshots/Videos
Before:
__before

After: ✨ 
__after

## How to test the changes?

There are (somewhat incomplete) tests for the UI that check for switching shiny and variants:
`npm run test src/test/ui/starter-select.test.ts`

Load a save file with some shinies like [this](https://discord.com/channels/1125469663833370665/1267339517371875361/1278397459462226082) or  [this](https://discord.com/channels/1125469663833370665/1267339517371875361/1278397521647108222)

(keep in mind that if you had changed the settings before for some Pokemon the game will use that so the default values will not be applied for these Pokemon. To avoid this you can do your testing in an incognito window)

Play around in starter select with the shiny pokemon, add them to the team, change variant, form, disable shiny, make sure the icons get updated properly and the correct variant and shiny state is used when starting a run

## Checklist
- ~~[] If I have text, did I make it translatable and add a key in the English locale file(s)?~~
- [X] Are the changes visual?

### DIFF
--- a/src/test/ui/starter-select.test.ts
+++ b/src/test/ui/starter-select.test.ts
@@ -53,9 +53,6 @@ describe("UI - Starter select", () => {
       const handler = game.scene.ui.getHandler() as StarterSelectUiHandler;
       handler.processInput(Button.RIGHT);
       handler.processInput(Button.LEFT);
-      handler.processInput(Button.CYCLE_SHINY);
-      handler.processInput(Button.V);
-      handler.processInput(Button.V);
       handler.processInput(Button.ACTION);
       game.phaseInterceptor.unlock();
     });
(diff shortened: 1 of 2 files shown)


KimJeongSun (commented):
elite shiny union vs normal pokemon brotherhood!

frutescens (approved):

DayKev (approved):

Opaque02:
Code looks good and runs as expected!

Opaque02 (approved):

PR #4007, opened 2024-09-03
[Bug] Fix for unpausing evolutions for secondary Pokémon in a fusion

seanrmon:

Evolutions can now be unpaused for unevolved secondary Pokémon of a fusion.

This bug was simply an oversight where the UI would not provide the "Unpause evolutions" option if the primary Pokémon of a fusion is fully evolved.

Now we check if the pokemon:
Has paused evolution
AND
(
A. Is not fully evolved
OR
B. Is a fusion AND if the fused pokémon not fully evolved
)
To ensure consistency, I used the same logic that exists in the check to see whether a secondary Pokémon's evolution items could appear in the shop reward
modifier-type.ts line 781 :
`} else if (pokemon.isFusion() && pokemon.fusionSpecies && pokemonEvolutions.hasOwnProperty(pokemon.fusionSpecies.speciesId) &&  ...`

### Screenshots/Videos

**Before:**
image
image
**After:**
image
I was then able to evolve the pokemon normally.
As expected, the option should still not show up for evolved pokemon:
image

## Checklist
- [x] Are the changes visual?


flx-sta:
I'm not entirely sure this fixes the bug. I mean what if you have a fusion of 
1. A fully evolved Pokemon
2. An unevolved Pokemon

That would mean only the fusion Pokemon has evolutions paused. I don't this this logic can account for that as there is only a general `pause` for the whole pokemon.

flx-sta (changes requested):
Doesn't work in the scenario is described previously

https://github.com/user-attachments/assets/35daa2f4-da08-431c-b755-438e93089807

seanrmon:
Thanks for linking an issue to this and your insights, but unfortunately I'm not sure what you mean.

> Doesn't work in the scenario is described previously
> freecompress-Screen.Recording.2024-09-03.at.1.02.41.PM.mp4
> 
> Here is the session-file AFTER splicing them
> 
> [sessionData3_Guest.txt](https://github.com/user-attachments/files/16854972/sessionData3_Guest.txt) change file-extension to `.prsv` to import

Thanks for the reply, is that demonstration using the beta branch or my own commit?

flx-sta:
> Thanks for the reply, is that demonstration using the beta branch or my own commit?

This is running on your PR. I checked it out and did all the testing on the state of this PR

seanrmon:
OK I think I understand you now; the unpause option shouldn't show up unless you're in the item shop "check team" screen, that's always the case as far as I know

seanrmon:
I will check that now, but I followed your steps and can confirm I got it working on the "check team" screen using my fix
image

flx-sta:
@seanrmon I am 100% sure about it.

Here is my scenario again:

You have 2 Pokemon (unspliced for now)
1. Charizard (level 36)
2. Squirtle (level 36)

Charizard is fully evolved. 
Squirtle isn't and evolutions have been `paused`.

You now splice those two.

You are no longer able to unpause evolutions for Squirtle now.

seanrmon:
Ah I understand - I thought that was working as indended but if that's not the behaviour you want I can change it

flx-sta:
[595-missing-fusion-unpause-session_before-splice.txt](https://github.com/user-attachments/files/16857740/595-missing-fusion-unpause-session_before-splice.txt)
<sup>change file-extension to `.prsv` to import</sup>

https://github.com/user-attachments/assets/127bc73c-0889-4d76-a728-f5703f8a9565

flx-sta:
I'm going to put this back into Draft. 
Keep In mind I'm working on a solution for the same Issue right now. Idk if you want to "compete" as one might have to throw their solution away.

seanrmon:
OK well I have a quick solution done up. If you want to use it it's up to yourself. 


https://github.com/user-attachments/assets/42e7f0c1-e1d0-4e10-9d5e-bc1dfbe3a3df



This check will see if parent A or parent B has "pauseEvolutions" set to true, and if either or both of them do, then "pauseEvolutions" will be set to true for the resulting fusion pokemon. Is this the behaviour that you are looking for?

seanrmon:
@flx-sta are you happy enough with this fix?

DayKev:
Let's say you fuse Charmander and Squirtle, can you individually pause/unpause evolutions for either of them without affecting the other?

seanrmon:
> Let's say you fuse Charmander and Squirtle, can you individually pause/unpause evolutions for either of them without affecting the other?

You can not individually pause/unpause evolutions for each fused Pokémon once it is fused, the Pokémon only has one "pause evolutions" boolean. I think this makes sense and I think making it an individual check would be messier

flx-sta:
@seanrmon I did some development in this and decided to drop it.
Make sure you are testing all the edge-cases e.g. If you have a fully evolved primary Pokemon and merge it with an unevolved secondary, does the secondary still evolve?
Cause that scenario made my give up as the current code doesn't support that even remotely. But I might have missed something. lmk if you found a way

seanrmon:
Unrelated but there's another bug in the fuse() function:
```
if (!this.isFainted()) {
        // If this Pokemon hasn't fainted, make sure the HP wasn't set over the new maximum
        this.hp = Math.min(this.hp, maxHp);
        this.status = getRandomStatus(this.status, pokemon.status); // Get a random valid status between the two
      } else if (!pokemon.isFainted()) {
        // If this Pokemon fainted but the other hasn't, make sure the HP wasn't set to zero
        this.hp = Math.max(this.hp, 1);
        this.status = pokemon.status; // Inherit the other Pokemon's status
      }
```
should be 
```
if ((!this.isFainted())&&(!pokemon.isFainted())) {
        // If this Pokemon hasn't fainted, make sure the HP wasn't set over the new maximum
        this.hp = Math.min(this.hp, maxHp);
        this.status = getRandomStatus(this.status, pokemon.status); // Get a random valid status between the two
      } else if (!pokemon.isFainted()) {
        // If this Pokemon fainted but the other hasn't, make sure the HP wasn't set to zero
        this.hp = Math.max(this.hp, 1);
        this.status = pokemon.status; // Inherit the other Pokemon's status
      }
```
I've seen this happen in-game where a resulting fusion pokemon has the "Fainted" status, even though it's at max HP. You don't need to do anything in the "else" case because in that scenario the only secondary pokemon is fainted, so we don't need to do anything

flx-sta:
Good catch about the bug. We should add that information to a possible existing issue. I'm trying to find that right now

-----

Regarding your "covering it" explanation.
I'd prefer to see a video proofing your point

flx-sta:
[4007_charizard-squirtle.prsv](https://github.com/user-attachments/files/16912284/4007_charizard-squirtle.txt)
<sup>Change file-extension to `.prsv` to import it</sup>

flx-sta (approved):

PR #4009, opened 2024-09-03
[P1 Bug][UI/UX] Address shop cursor target feedbacks

torranx:
- entering daily run's rewards phase should no longer soft lock when the shop cursor target is set to `Shop`
- changed "Items" label to "Rewards"
- set the default option to `Rewards`
- set the default option to be left-most/first option

Addressing feedbacks:
- P1: Game can softlock if it's set to "Shop" before buyable shop items introduced - in Daily Run mode
- Some of the terms are ambiguous. "Items" can refer to the buyable items, the free ones, or transferring items. Someone suggested "Rewards"
- Move the default option back to the original please (Rewards)
- The default most common option (Rewards) should be the first option

- added an index mapping of options-enum to avoid the risk of mismatching the option keys and the shop cursor targets enum
- added condition to check the current game mode and the current set shop cursor target option

### Screenshots/Videos
Updated options
Screenshot 2024-09-04 at 4 02 25 AM

Daily run no longer softlocks when shop cursor target is set to `Shop`

https://github.com/user-attachments/assets/30f61a8c-c947-41a0-b753-5f35bbd4e97c

## How to test the changes?
set settings, enter rewards/shop phase

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/locales/de/settings.json
+++ b/src/locales/de/settings.json
@@ -100,7 +100,7 @@
   "moveTouchControls": "Bewegung Touch Steuerung",
   "shopOverlayOpacity": "Shop Overlay Deckkraft",
   "shopCursorTarget": "Shop-Cursor Ziel",
-  "items": "Items",
+  "rewards": "Items",
   "reroll": "Neu rollen",
   "shop": "Shop",
   "checkTeam": "Team überprüfen"
--- a/src/locales/ja/settings.json
+++ b/src/locales/ja/settings.json
@@ -100,7 +100,7 @@
   "moveTouchControls": "タッチ移動操作",
   "shopOverlayOpacity": "ショップオーバレイ不透明度",
   "shopCursorTarget": "ショップカーソル初位置",
-  "items": "アイテム",
+  "rewards": "アイテム",
   "reroll": "選択肢変更",
   "shop": "ショップ",
   "checkTeam": "手持ちを確認"
--- a/src/locales/ja/settings.json
+++ b/src/locales/ja/settings.json
@@ -100,7 +100,7 @@
   "moveTouchControls": "タッチ移動操作",
   "shopOverlayOpacity": "ショップオーバレイ不透明度",
   "shopCursorTarget": "ショップカーソル初位置",
-  "rewards": "アイテム",
+  "rewards": "ご褒美",
   "reroll": "選択肢変更",
   "shop": "ショップ",
   "checkTeam": "手持ちを確認"
--- a/src/system/settings/settings.ts
+++ b/src/system/settings/settings.ts
@@ -635,7 +635,7 @@ export const Setting: Array<Setting> = [
     key: SettingKeys.Shop_Cursor_Target,
     label: i18next.t("settings:shopCursorTarget"),
     options: SHOP_CURSOR_TARGET_OPTIONS,
-    default: ShopCursorTarget.REWARDS,
+    default: 0,
     type: SettingType.DISPLAY
   },
   {
--- a/src/system/settings/settings.ts
+++ b/src/system/settings/settings.ts
@@ -25,6 +25,7 @@ const VOLUME_OPTIONS: SettingOption[] = new Array(11).fill(null).map((_, i) => i
   value: "Mute",
   label: getTranslation("settings:mute")
 });
+
 const SHOP_OVERLAY_OPACITY_OPTIONS: SettingOption[] = new Array(9).fill(null).map((_, i) => {
   const value = ((i + 1) * 10).toString();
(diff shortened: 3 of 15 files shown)


f-fsantos (dismissed):

f-fsantos (approved):

src/test/items/dire_hit.test.ts
```diff
@@ -13,6 +13,7 @@ import { Button } from "#app/enums/buttons";
 import { CommandPhase } from "#app/phases/command-phase";
 import { NewBattlePhase } from "#app/phases/new-battle-phase";
 import { TurnInitPhase } from "#app/phases/turn-init-phase";
+import { ShopCursorTarget } from "#app/enums/shop-cursor-target.js";
```
DayKev, line 16:
```suggestion
import { ShopCursorTarget } from "#app/enums/shop-cursor-target";
```

src/test/items/temp_stat_stage_booster.test.ts
```diff
@@ -16,6 +16,7 @@ import ModifierSelectUiHandler from "#app/ui/modifier-select-ui-handler";
 import { TurnInitPhase } from "#app/phases/turn-init-phase";
 import { BattleEndPhase } from "#app/phases/battle-end-phase";
 import { EnemyCommandPhase } from "#app/phases/enemy-command-phase";
+import { ShopCursorTarget } from "#app/enums/shop-cursor-target.js";
```
DayKev, line 19:
```suggestion
import { ShopCursorTarget } from "#app/enums/shop-cursor-target";
```

frutescens:
What happens to everyone's settings before the PR?

torranx:
> What happens to everyone's settings before the PR?

You mean after? The game won't break but I think it will reset back to the default option because I changed the key REROLL_TARGET to SHOP_CURSOR_TARGET

src/locales/ja/settings.json
```diff
@@ -100,7 +100,7 @@
   "moveTouchControls": "タッチ移動操作",
   "shopOverlayOpacity": "ショップオーバレイ不透明度",
   "shopCursorTarget": "ショップカーソル初位置",
-  "items": "アイテム",
+  "rewards": "アイテム",
```
Chapybara-jp, line 103:
```suggestion
  "rewards": "ご褒美",
```

Chapybara-jp (dismissed):

src/locales/de/settings.json
```diff
@@ -100,7 +100,7 @@
   "moveTouchControls": "Bewegung Touch Steuerung",
   "shopOverlayOpacity": "Shop Overlay Deckkraft",
   "shopCursorTarget": "Shop-Cursor Ziel",
-  "items": "Items",
+  "rewards": "Items",
```
CodeTappert, line 103:
```suggestion
  "rewards": "Belohnungen",
```


Please check if thats to long

torranx:
> In my local instance, I set the cursor to Reroll from Check Team and then checked out this PR. It then changed my settings to Shop. 

That is an expected behavior with what we have now since the setting's key changed.

> I do think there should be a way to transition current settings to old b/c otherwise we'll run into similar issues from before.

What similar issues from before exactly, only "issue" here is that the users will only have to re-set the option they want? I agree there should be a way for us to transition a key to another.

src/locales/de/settings.json
torranx: it is
Screenshot 2024-09-04 at 5 54 23 AM

src/system/settings/settings.ts
```diff
@@ -53,6 +55,40 @@ const AUTO_DISABLED: SettingOption[] = [
...
+    label: i18next.t("settings:shop")
+  },
+  {
+    value: "Reroll",
+    label: i18next.t("settings:reroll")
+  },
+  {
+    value: "Check Team",
+    label: i18next.t("settings:checkTeam")
+  }
```
Tempo-anon, line 74: minor nit: this and the switch below have all the values in different orders from the way the enum is defined. Ideally they should all be in the same order as they are displayed in game

Tempo-anon (approved):

src/system/settings/settings.ts
torranx: that is what the index mapping I added is for, the enums matches the rows in the modifier select phase ie `0` being the row where Reroll is - so the only thing I can change here is to match the switch with the value of `options`

Tempo-anon (approved):

DayKev (approved):

f-fsantos (approved):

PR #4013, opened 2024-09-03
[Test] Add Dragon Cheer tests

torranx:
none

adding tests to https://github.com/pagefaultgames/pokerogue/pull/3959 now that it is possible

dragon cheer test

## How to test the changes?
`npm run test dragon_cheer`

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/test/moves/dragon_cheer.test.ts
+++ b/src/test/moves/dragon_cheer.test.ts
@@ -1,5 +1,4 @@
 import { BattlerIndex } from "#app/battle";
-import { DragonCheerTag } from "#app/data/battler-tags";
 import { Type } from "#app/data/type";
 import { Moves } from "#app/enums/moves";
 import { Species } from "#app/enums/species";
@@ -37,7 +36,6 @@ describe("Moves - Dragon Cheer", () => {
   it("increases the user's allies' critical hit ratio by one stage", async () => {
     await game.classicMode.startBattle([Species.DRAGONAIR, Species.MAGIKARP]);
 
-    const magikarp = game.scene.getPlayerField()[1];
     const enemy = game.scene.getEnemyField()[0];
 
     vi.spyOn(enemy, "getCritStage");
@@ -47,19 +45,14 @@ describe("Moves - Dragon Cheer", () => {
 
     await game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.PLAYER_2, BattlerIndex.ENEMY, BattlerIndex.ENEMY_2]);
 
-    // After Dragon Cheer
-    await game.phaseInterceptor.to("MoveEndPhase");
-    expect(magikarp.getTag(DragonCheerTag)).toBeDefined();
-
     // After Tackle
-    await game.phaseInterceptor.to("MoveEndPhase");
+    await game.phaseInterceptor.to("TurnEndPhase");
     expect(enemy.getCritStage).toHaveReturnedWith(1); // getCritStage is called on defender
   }, TIMEOUT);
 
   it("increases the user's Dragon-type allies' critical hit ratio by two stages", async () => {
     await game.classicMode.startBattle([Species.MAGIKARP, Species.DRAGONAIR]);
 
-    const dragonair = game.scene.getPlayerField()[1];
     const enemy = game.scene.getEnemyField()[0];
 
     vi.spyOn(enemy, "getCritStage");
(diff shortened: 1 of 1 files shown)


src/test/moves/dragon_cheer.test.ts
```diff
@@ -0,0 +1,117 @@
...
+  });
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override
+      .battleType("double")
+      .enemyAbility(Abilities.BALL_FETCH)
+      .enemyMoveset(SPLASH_ONLY)
+      .enemyLevel(20)
+      .startingWave(4)
```
DayKev, line 34:
```suggestion
```
Why start on wave 4?

flx-sta:
I know tests are sometimes redundant but some of those seem like they could be merged. especially the first few lines are almost identical and could be put into `beforeEach()` while exposting the properly variables in the correct scope (of describe)

src/test/moves/dragon_cheer.test.ts
torranx: 4 is always a double battle if im not mistaken

DayKev: I don't think so? Regardless, that's what the battle type override is for.

torranx: > I don't think so? Regardless, that's what the battle type override is for.

battle type override only overrides the battle to be double, it does not guarantee that 2 opposing pokemon spawn - the tests does not require 2 opposing pokemon to be clear

torranx:
> I know tests are sometimes redundant but some of those seem like they could be merged. especially the first few lines are almost identical and could be put into `beforeEach()` while exposting the properly variables in the correct scope (of describe)

when you say "first few lines", I'm assuming these:

```
await game.classicMode.startBattle([Species.DRAGONAIR, Species.MAGIKARP]);

const magikarp = game.scene.getPlayerField()[1];
const enemy = game.scene.getEnemyField()[0];

vi.spyOn(enemy, "getCritStage");

game.move.select(Moves.DRAGON_CHEER, 0);
game.move.select(Moves.TACKLE, 1, BattlerIndex.ENEMY);
```

you cant get the pokemon without starting the battle first

src/test/moves/dragon_cheer.test.ts
DayKev: Wait what? Since when? I've never had issues with there not being 2 opposing Pokémon when using the override. The game should always spawn 2 opposing Pokémon if it's a double battle.

torranx: It's either that or it has been updated/fixed already lol

torranx: Checked with only the battle type override and it now spawns 2 pokemon regardless of wave - I'm could sweat it wasnt before

DayKev: I just checked and wave 4 is not special, it's not always a double battle.
Also not running into a time where the game doesn't spawn 2 Pokémon when the override is set to "double" either.

flx-sta:
@torranx I mean you can start the battle in `beforeEach()` too. That's what I meant

torranx:
> @torranx I mean you can start the battle in `beforeEach()` too. That's what I meant

that would require setting the beforeEach()'s callback to be `async`, will that not cause side effects?

src/test/moves/dragon_cheer.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    vi.spyOn(enemy, "getCritStage");
+
+    game.move.select(Moves.DRAGON_CHEER, 0);
+    game.move.select(Moves.TACKLE, 1, BattlerIndex.ENEMY);
+
+    await game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.PLAYER_2, BattlerIndex.ENEMY, BattlerIndex.ENEMY_2]);
+
+    // After Dragon Cheer
+    await game.phaseInterceptor.to("MoveEndPhase");
+    expect(magikarp.getTag(DragonCheerTag)).toBeDefined();
```
DayKev, line 52: Imo you don't need to check for the `BattlerTag` being set, just test that the outcome of the battle is as expected (ie: the crit stage test below). Though it's not a big issue in the end.

DayKev (approved):

flx-sta (approved):

frutescens (approved):

PR #4023, opened 2024-09-04
[Enhancement] [UI/UX] Add ability and passive tooltips to starter select screen

brandonbay:
On the starter select screen, hovering the cursor over a Pokémon's Ability or Passive will show a tooltip with that Ability's description. Pressing the Toggle Ability key while the tooltip is visible will update its text.

Selecting a different Pokémon or deselecting one altogether will hide any visible tooltip.

There is no way to view the details of the selected Ability and Passive on the starter select screen. Move details can be viewed in the "Manage Moves" menu, and Natures are simple enough that even without the menu, their impact is clear. However, Abilities and Passives are opaque and require a visit to an external website if a player is not already familiar with their effect.

Since Abilities are a short list of three and Passives are single values, creating a menu to view these details seemed like overkill. A tooltip is a simple way to get the information on the screen

The change mostly just leverages the existing tooltip system, calling `setInteractive` on `this.pokemonAbilityText` and `this.pokemonPassiveText`. A mock for `setInteractive` was also added to mockText, to keep the tests passing.

New tooltip management functions `getTooltip` and `editTooltip` have been added, to support the case of the Toggle Ability key. As the tooltip is shown in an event handler, having the ability to edit it in place when it is currently visible is essential for this task. Knowing its current visibility and text state is also useful, as without this information, the egg candy count tooltip would also updated to show the Ability text when toggling through the Ability list.

Tooltips will also now reverse on the y-axis if they would fall off the bottom of the screen, and their maximum width has been increased from 696 to 850.

The egg candy count tooltip previously remained on the screen when changing the selected Pokémon, but it did not update its value. Now, it will disappear, and as before, the user will have to re-hover the sprite to see the new value.

Mobile tooltips are as functional as they ever are.

#### Desktop interaction:
ndiRXr3DVZ

#### Mobile still:
Screenshot_20240904_122334_Chrome

## How to test the changes?
The tests for the starter select screen continue to pass with the updated mock. No new tests were added to ensure the tooltips' presence.

To manually test, simply visit the starter select screen and hover your mouse over a Pokémon's Ability or Passive.

## Checklist
- [x] Are the changes visual?


src/ui/starter-select-ui-handler.ts
```diff
@@ -1921,6 +1925,18 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
               }
             } while (newAbilityIndex !== this.abilityCursor);
             starterAttributes.ability = newAbilityIndex; // store the selected ability
+
+            const { visible: tooltipVisible, content: oldTooltip } = (this.scene as BattleScene).ui.getTooltip();
```
MokaStitcher, line 1929: looks good at a glance, will take a closer look when i have more time

I don't think the casts to BattleScene are necessary in `starter-select-ui-handler` [(scene is defined as a BattleScene for UIHandlers)](https://github.com/pagefaultgames/pokerogue/blob/370468002946c433f25a75dd758dd5bd6ec5e002/src/ui/ui-handler.ts#L10) so it would be nice to remove them to make the code easier to read

MokaStitcher:
Played with it a little, this is really a nice addition!

Is it intended that when the passive is not unlocked we can't see the description?
I'd argue that it's the case in which having the tooltip is the most important, so that we can decide if we want to purchase it :)

Second thing, how hard would it be to snap the tooltip to the bottom of the game if it's too big?
For example here the tooltip for Wo-chien's passive gets cut off, and it's far from having the longest ability description in the game. Illuminate and Anger Shell have pretty long descriptions for example, if you want to check them

Screenshot 2024-09-04 at 19 40 24

Another option would be to make the tooltip a little wider, maybe a mix of both would be good.

brandonbay:
I did choose to hide the description when the passive is locked. I considered that it might be useful when unlocking, but also wondered if maybe keeping it more of a "surprise" was better Roguelike behaviour and less visual clutter as you move your mouse around.

It would be very easy to always show it, regardless of unlocked/enabled state of the passive. I don't feel too strongly either way, so if anyone else agrees with you, I'm happy to change that and always show it.

As for adjusting the tooltip size, I will look into it. So far, I've only used what was already there, so I'm not sure how hard it would be or how much impact it would have on the other tooltips in the game, but it would quite obviously be a useful change.

MokaStitcher:
I asked the balance team and they said that we should show the tooltip even if the passive is locked, so you're good to make the change! :)

brandonbay:
Passive tooltips will always show!

Tooltips are now a little wider, and will show above the cursor if they would otherwise fall off the bottom of the screen.

image

Opaque02:
So one issue on mobile is that while the hover over does work, it's hard to tell if you're hovering over the ability or the passive, especially if your thumb is in the way and you can't tell which is which. Maybe we need some kind of better visual distinction between the two?

Secondly, the interactive portion turns off when changing pokemon - it can be a little jarring that changing abilities works and updates accordingly, but because of it turning off when changing pokemon, if you've got your mouse (or finger) in the interactive section and change pokemon, you need to move your mouse/finger out of the interactive section and put it back in. Maybe there's a way we can make it persistent while changing pokemon too? Something to consider

brandonbay:
It's easy enough to just toss the ability name into the tooltip title for clarity. Good enough?

Persisting the tooltip while changing Pokémon wasn't as hard as I thought, so that's working for abilities, passives, and candy counts now. All it took was an extra `activeTooltip` property on the `StarterSelectUiHandler` class.
brave_xT07UTbCsi

DayKev (approved):

MokaStitcher:
I'm going to nitpick again I'm sorry ;;
The changes are great, but sometimes when moving from ability to passive there will be a brief time where the passive shows below, then jumps up

https://github.com/user-attachments/assets/2c774f37-869c-4ee5-9e12-e01bd2278be3

Maybe the tooltip could just always be shown above the cursor to avoid this issue?
Or the tooltip location could get computed based on the lowest point of the ability instead of the cursor location so that it doesn't jump in the middle of it, if that makes sense 🤔

Opaque02:
> It's easy enough to just toss the ability name into the tooltip title for clarity. Good enough?
> 
> Persisting the tooltip while changing Pokémon wasn't as hard as I thought, so that's working for abilities, passives, and candy counts now. All it took was an extra `activeTooltip` property on the `StarterSelectUiHandler` class. brave_xT07UTbCsi [ brave_xT07UTbCsi ](https://private-user-images.githubusercontent.com/16120332/364646215-3f1b4e7e-3107-4c92-b89e-0be5873f02e9.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjU1ODE5NzgsIm5iZiI6MTcyNTU4MTY3OCwicGF0aCI6Ii8xNjEyMDMzMi8zNjQ2NDYyMTUtM2YxYjRlN2UtMzEwNy00YzkyLWI4OWUtMGJlNTg3M2YwMmU5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA5MDYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwOTA2VDAwMTQzOFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTg3YmI1YmEzYTRiM2QwZGQ2MDNhZTUyNTU0NDY4ZWU0ZmY4YWQ4ZjU3NzZiYTQ4MDViNWU2MTBjMjY5YmJiYmEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.b3g1HtwljXsUBLIDkCISHQGdTjNM2CgqQ8wbT33NXtU) [ ](https://private-user-images.githubusercontent.com/16120332/364646215-3f1b4e7e-3107-4c92-b89e-0be5873f02e9.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjU1ODE5NzgsIm5iZiI6MTcyNTU4MTY3OCwicGF0aCI6Ii8xNjEyMDMzMi8zNjQ2NDYyMTUtM2YxYjRlN2UtMzEwNy00YzkyLWI4OWUtMGJlNTg3M2YwMmU5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA5MDYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwOTA2VDAwMTQzOFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTg3YmI1YmEzYTRiM2QwZGQ2MDNhZTUyNTU0NDY4ZWU0ZmY4YWQ4ZjU3NzZiYTQ4MDViNWU2MTBjMjY5YmJiYmEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.b3g1HtwljXsUBLIDkCISHQGdTjNM2CgqQ8wbT33NXtU)

This looks good, thanks!

src/test/utils/mocks/mocksContainer/mockText.ts
```diff
@@ -197,6 +197,10 @@ export default class MockText implements MockGameObject {
     this.color = color;
   });
 
+  setInteractive(hitArea, hitAreaCallback, dropZone) {
+    return null;
+  }
```
flx-sta, line 202: Please use 
```suggestion

  setInteractive = vi.fn();
```

instead. That way the method is already a mock/spy and can be mocked/spied (obviously)

brandonbay: Fixed, including in the other places mocking this function.

brandonbay:
> I'm going to nitpick again I'm sorry ;; The changes are great, but sometimes when moving from ability to passive there will be a brief time where the passive shows below, then jumps up
> 
>  Screen.Recording.2024-09-06.at.00.00.51.mov 
> Maybe the tooltip could just always be shown above the cursor to avoid this issue? Or the tooltip location could get computed based on the lowest point of the ability instead of the cursor location so that it doesn't jump in the middle of it, if that makes sense 🤔
> 
> edit: oh wait, your changes for the location of the tooltip impact all tooltips in the game right? So the first solution is probably off the table, unsure about the second one

@MokaStitcher This is consistent with how tooltips already work side to side, and would impact things everywhere. I would argue a larger redesign of tooltip placement is outside the scope of this change, and should be logged as an issue for later.

brave_D30B8HO5FM

DayKev (approved):

MokaStitcher (approved):

Tempo-anon (approved):
This looks very nice. Let's get this into beta to test and then see if any changes are needed.

PR #2699, opened 2024-06-29
[Refactor/Bug/Move] Overhaul Stats and Battle Items, Implement Several Stat Moves

xsn34kzx:
## What are the user-related changes?

- "Natural" stat-related fixes to moves, abilities, and other miscellaneous things:
  - [`Beast Boost`](https://bulbapedia.bulbagarden.net/wiki/Beast_Boost_(Ability)) now considers all effective stats when choosing a stat stage to boost, considers temporarily overridden stats due to moves like `Speed Swap`, and honors order preference when choosing a stat stage to boost (based on the highest stat).
  - `Imposter` and `Transform` now have their copied stats reflected in effective stat calculations (related to #2564 but does not fully address it). This has effects in enemy AI, how damage is calculated, and much more.
  - `White Smoke`/`Clear Body` no longer push multiple text prompts announcing their effects for instances where multiple stat stage decreases occur (i.e., getting hit by `Tickle` or `Parting Shot`).
  - When a transformed Pokemon leveled, the level-up phase would display and use the copied stats as the basis for calculating new stats; now, the Pokemon's true stats are always reflected.
  - When a boss segment is cleared, all effective stats are now weighted properly, which means that the chance of the boss getting the appropriate stat stage raises is more accurate.
- Implements several moves to essentially demo the overhaul:
  - [`Power Split`](https://bulbapedia.bulbagarden.net/wiki/Power_Split_(move))
  - [`Guard Split`](https://bulbapedia.bulbagarden.net/wiki/Guard_Split_(move))
  - [`Speed Swap`](https://bulbapedia.bulbagarden.net/wiki/Speed_Swap_(move))
  - [`Guard Swap`](https://bulbapedia.bulbagarden.net/wiki/Guard_Swap_(move))
  - [`Power Swap`](https://bulbapedia.bulbagarden.net/wiki/Power_Swap_(move))
- Overhauls the existing battle items (at request of the balance team):
  - Only one of a given type of item can be held at a time.
    - e.g., if you already have an `X Attack`, you can't have another `X Attack` but grabbing an `X Accuracy` happens without a hitch.
  - To accommodate the above change, grabbing another item of the same type will now "refresh" the battle count of the one currently obtained.
    - e.g., taking a new `X Defense` when you have an `X Defense` at 3 battles left will bring that existing `X Defense` back to 5 battles left but won't add the new one to the modifier list.
[...]

PR #4032, opened 2024-09-05
[Item/Balance] Overhaul Lapsing Modifiers

xsn34kzx:

- General Changes
  - Both lures (all tiers) and battle items have their battle counters adjusted such that they begin with solid green outline and, as they lapse over battles, turn to dark red.
  - Both items, when "refreshed", play a sound cue (`se/restore`) to indicate that they've changed. 
  - Both items now have updated English entries to describe their new behavior.
- Changes to Lures
  - Lures now mirror the behavior of battle items in that multiple lures of the same type can no longer be held at once and getting another will only "refresh" the battle count of the existing same-type lure; you can still have every tier of lure at the same time with stacking applications to double battle chance. 
  - To maintain the consequential behavior of having multiple lures at a given time, lures now provide a 4x chance to provide a double battle. As a result, there are new double battle probabilities for owning each lure:
    - One unique lure -> 50% double battle chance, 12.5% double boss battle chance
    - Two unique lures -> 100% double battle chance, 50% double boss battle chance
    - Three unique lures -> 100% double battle chance, 100% double boss battle chance

This PR rounds out the battle item changes made in #2699 by providing up-to-date descriptions and provides better visual feedback for what is essentially new behavior to the game entirely ("refreshing" lapsing modifiers instead of stacking multiple on top of one another). 

Originally, lures were left alone due to being out of scope of the main intent of #2699 and nerfing battle items in general; as such, the battle count reset logic in `add()` had to be duplicated for both `Dire Hit` and the rest of the battle items. By including lures, the code can now make use of polymorphism in the extended class, `LapsingPersistentModifier`, which makes all classes involved more clean overall. In addition, a `maxBattles` field was added in order to keep track of what value to refresh back to aid in accurate linear interpolation

A helper function `hslToHex()` was added to `Utils` to essentially allow for on-the-fly hex code generation with varying hue, saturation, and lightness. In this specific case, linear interpolation is used on hue (formula below) while keeping saturation and lightness constant to achieve the gradual hue shifting on the items' battle count text:

$$ \text{hue} = m \cdot \frac{\text{battleCount}}{\text{maxBattles}} + n $$

where $m + n$ is the ending value and $n$ is the starting value. Note that these can be tweaked and adjusted as needed; none of the colors are hard-coded.

Unit tests were added for lures and updated for battle items to reflect the changes (note that visual changes are not tested). In addition to this, an `item` option was added to the test boilerplate script.

#### Example - `Dire Hit` Lapsing
| 5 | 4 | 3 | 2 | 1 |
| --- | --- | --- | --- | --- |
|  |  |  |  |  |

#### Example - Save Slot

saveSlot

#### Examples - In-Battle

| Fresh Lapsing Modifiers | Lapsing Modifiers after 4 Battles |
| --- | --- |
| earlyBattle | lateBattle |

https://github.com/user-attachments/assets/7af1e819-1fdd-441d-8366-81525d9d425f

## How to test the changes?

The most fun way to test the new behavior and observe the new visual effect is by using the following overrides and starting a new Classic run:
```ts
const overrides = {
  ITEM_REWARD_OVERRIDE: [
    { name: "TEMP_STAT_STAGE_BOOSTER", type: Stat.ATK },
    { name: "DIRE_HIT" },
    { name: "SUPER_LURE" }
  ],
  STARTING_MODIFIER_OVERRIDE: [
    { name: "TEMP_STAT_STAGE_BOOSTER", type: Stat.ATK },
    { name: "TEMP_STAT_STAGE_BOOSTER", type: Stat.DEF },
    { name: "TEMP_STAT_STAGE_BOOSTER", type: Stat.SPATK },
    { name: "TEMP_STAT_STAGE_BOOSTER", type: Stat.SPDEF },
    { name: "TEMP_STAT_STAGE_BOOSTER", type: Stat.SPD },
    { name: "TEMP_STAT_STAGE_BOOSTER", type: Stat.ACC },
    { name: "DIRE_HIT" },
    { name: "LURE" },
    { name: "SUPER_LURE" },
    { name: "MAX_LURE" }
  ],
  STARTING_LEVEL_OVERRIDE: 100
```

The item rewards are used to refresh any items of the same type if applicable.

## Checklist
- [x] Are the changes visual?


damocleas (approved):

create-test-boilerplate.js
```diff
@@ -40,6 +40,9 @@ if (type === 'move') {
 } else if (type === 'ability') {
     dir = path.join(__dirname, 'src', 'test', 'abilities');
     description = `Abilities - ${formattedName}`;
+} else if (type === "item") {
+    dir = path.join(__dirname, 'src', 'test', 'items');
+    description = `Items - ${formattedName}`;
```
torranx, line 45: Please update the file docs the `console.error` too

Adri1 (approved):

src/modifier/modifier-type.ts
```diff
@@ -1348,19 +1355,22 @@ export const modifierTypes = {
...
   DIRE_HIT: () => new class extends ModifierType {
     getDescription(_scene: BattleScene): string {
-      return i18next.t("modifierType:ModifierType.TempStatStageBoosterModifierType.description", { stat: i18next.t("modifierType:ModifierType.DIRE_HIT.extra.raises") });
+      return i18next.t("modifierType:ModifierType.TempStatStageBoosterModifierType.description", {
+        stat: i18next.t("modifierType:ModifierType.DIRE_HIT.extra.raises"),
+        amount: i18next.t("modifierType:ModifierType.TempStatStageBoosterModifierType.extra.stage")
+      });
     }
-  }("modifierType:ModifierType.DIRE_HIT", "dire_hit", (type, _args) => new Modifiers.TempCritBoosterModifier(type)),
+  }("modifierType:ModifierType.DIRE_HIT", "dire_hit", (type, _args) => new Modifiers.TempCritBoosterModifier(type, 5)),
```
flx-sta, line 1373: whats the deal with the `5` here?

innerthunder (approved):

flx-sta (approved):
lgtm 
image

damocleas (approved):

src/modifier/modifier-type.ts
xsn34kzx: Pretty much both battle items and Dire Hit have `5` for `maxBattles` going into the actual modifier itself rather than saved anywhere in the `ModifierType` classes since they don't vary in item descriptions like lures do.

PR #4040, opened 2024-09-05
[QoL] Username finder

Opaque02:
There will be a new icon (coming soon) that users can click on the login screen that will help them see their username if they've forgotten it

Walker wants it

image

Added a new icon to the login screen. This icon will check the localStorage and filter out any keys that start with "data_{username}". This is a key for a save file, so it gets all of the save file names. It then filters out the "data_" to just have a list of all usernames that have logged in. If there's less than 2, it will show you a list of usernames on the device. If there's more than two, it will show you an error that you have too many save files to use this, and not show any. This is to avoid people gaining access to someone else's account on a shared device.

Also, there's a bug in live at the moment on the login screen. There's some setup that's done for the login screen that makes it so the discord and google link have a link to go to. However, this is done within the setup of the login ui handler. Normally this isn't an issue, but it's got an interactive section which is what allows it to be clicked/tapped on to activate. However, if you go to register screen, it disables the interactive part so you don't click it while it's not there. The issue is if you go back to the login screen, the setup bit isn't run so the interactive part isn't added back, and the buttons then do nothing. I have fixed this by moving the interactive logic to be called whenever the login screen is shown instead of just once within the setup. You can see this on beta/live at the moment by logging out, clicking "Register", then once that loads go back to the login and try to click on the discord/google images; they will not work any more.

### Screenshots/Videos
Below is the title screen with the icon (to be changed soon):

image

Below is a video on my dev env where I show I've only got a single save file and the icon works. Then I add a new fake save file to show two users on a device and show both showing up. Finally I then add a third save file to show more than 2 usernames on file and show it failing:

https://github.com/user-attachments/assets/28c09498-f17c-4b27-b1e1-be705498d93d

## How to test the changes?
Download this PR. If you have access to multiple account you can test it that way, but otherwise within your dev environment, open F12 and go to the console. You do not need to have the game paused for this.

You should be able to type out `Object.keys(localStorage)` and it'll return an array; expand it out and you should see something like data_Guest. This is your guest data on the machine. You can now click the button in the top left and an option popup will appear with "Guest". This shows that Guest has data on your machine. You can either click A/space/Z etc or X/cancel etc to go past the popup.

Next open [this file](https://github.com/user-attachments/files/16884801/Fake.data.txt). It has some fake data that you can use for extra accounts. Copy _all_ the text in this file (starting from "eyJ0cmF" and ending with "zU4Mzd9Cg==" and go back to the console. Now type `localStorage.data_{username} = "{pasted fake data}"`. Replace the {username} with a username of your choice, and the {pasted fake data} with the data from the file above. If you'd like to test it worked, type `Object.keys(localStorage)` once more, and expand the shown array. You should now see a data_Guest and a data_{username}. Once done, you can click the icon in the top left again to now see two usernames - Guest and {username} from before. 

Repeat the above step, but use a different username. Pressing the icon now will show you an error saying "You have too many save files to use this", and no shown option to see the usernames. 

If you would like to delete some user data in order to test this out some more, you can type into the console `delete localStorage.data_{username}`, with the username being the save data you want to delete from your dev environment.

## Checklist
- [x] Are the changes visual?


f-fsantos (approved):

torranx (approved):
looks good

PR #4048, opened 2024-09-05
[QoL] Adding challenge arrows

Opaque02:
When going to start a challenge, left/right arrows will now show up for each challenge to show you that you can use the arrow keys to change them. When you get to the end, the appropriate arrow will disappear to let you know you've hit the end.

Smewkie suggested it on discord, here we are
image

Added in some arrows based on the cursor and reverse cursor icons. They get placed onto the challenge ui handler, and the value text of the challenge (i.e. the generation number, type, on/off etc) is now centered between the two arrows. When you get to the max/min range of the challenge, it makes hides the arrows

When you go to start a challenge, once the start cursor is visible, the challenge icons will tint to show that you can't change the challenges any more. This may change to hide the cursors as per frutescens' comment, so keep an eye out for this

This also goes through the values for each challenge and bases the arrow placement on the widest text box available. This works per language, so if some languages have a longer value than others, it will still auto place the arrows and center the value text. Lastly, there's also a check for the max width of the arrow going past the edge of the challenge container. If this happens, the right arrow will be as far right as it can go, even if it means the text will overlap. 

### Screenshots/Videos
Below is a video of me going through the challenges screen and showing the arrows on display and showing it in the different languages:

https://github.com/user-attachments/assets/ada5ec37-b63e-408d-bcac-a68698364c9c

As you can see above, the position of the arrows change depending on the text in the challenges!

Below is a video of the arrows tinting when you go to hit start, and untinting when you go back to the challenge:

https://github.com/user-attachments/assets/3f1b4d07-980c-441b-b0cb-0f7e59000d09

## How to test the changes?
Download this PR and go to start a new challenge run. Try to change the values on the challenges and see how they go. This does _not_ handle other languages well yet, so that's the next thing to work on for it

## Checklist
- [x] Are the changes visual?


src/ui/challenges-select-ui-handler.ts
```diff
@@ -135,14 +142,31 @@ export default class GameChallengesUiHandler extends UiHandler {
...
+      const rightArrow = this.scene.add.image(0, 0, "cursor");
+      rightArrow.setName(`challenge-right-arrow-${i}`);
+      rightArrow.setOrigin(0, 0);
+      rightArrow.setScale(0.75);
+      rightArrow.setVisible(false);
+      this.valuesContainer.add(rightArrow);
+
       const value = addTextObject(this.scene, 0, 28 + i * 16, "", TextStyle.SETTINGS_LABEL);
       value.setName(`challenge-value-text-${i}`);
+      //value.setX((rightArrow.x - (leftArrow.x + leftArrow.width)) / 2);
```
DayKev, line 161: You missed some commented out code.

frutescens (changes requested):
When the player has pressed their first button.ACTION, could you hide the cursor arrows to signify that the challenges have been locked in? If they cancel out of confirming, then they should reappear.

src/data/challenge.ts
torranx, line 1: while you're at it, can you change the `integer` types to `number` from the functions you changed?

src/ui/challenges-select-ui-handler.ts
Opaque02: Fixed in next commit

src/data/challenge.ts
Opaque02: Done in next commit

src/ui/challenges-select-ui-handler.ts
```diff
@@ -203,16 +242,34 @@ export default class GameChallengesUiHandler extends UiHandler {
...
+      if (challengeLabel.rightArrow.x + challengeLabel.rightArrow.width + this.optionsBg.rightWidth + this.arrowSpacing > this.optionsWidth) {
+        // if we go out of bounds of the box, set the x position as far right as we can without going past the box, with this.arrowSpacing to allow a small gap between the arrow and border
+        challengeLabel.rightArrow.setX(this.optionsWidth - this.arrowSpacing - this.optionsBg.rightWidth - challengeLabel.rightArrow.width);
+      }
+
+      // this line of code gets the center point between the left and right arrows from their left side (Arrow.x gives left point), taking into account the width of the arrows
+      const xLocation = challengeLabel.rightArrow.x - (challengeLabel.rightArrow.x - (challengeLabel.leftArrow.x + challengeLabel.leftArrow.width)) / 2 - 0.5;
       if (challenge.id === Challenges.SINGLE_TYPE) {
-        this.monoTypeValue.setPositionRelative(this.challengeLabels[i].label, 113, 8);
+        this.monoTypeValue.setPositionRelative(challengeLabel.label, 112, 8);
```
MokaStitcher, line 261:
```suggestion
```
setting the relative position is not needed anymore since the X and Y are set directly below

src/ui/challenges-select-ui-handler.ts
```diff
@@ -203,16 +242,34 @@ export default class GameChallengesUiHandler extends UiHandler {
...
+      challengeLabel.rightArrow.setVisible(challenge.value !== challenge.maxValue);
+
+      // this check looks to make sure that the arrows and value textbox don't take up too much space that they'll clip the right edge of the options background
+      if (challengeLabel.rightArrow.x + challengeLabel.rightArrow.width + this.optionsBg.rightWidth + this.arrowSpacing > this.optionsWidth) {
+        // if we go out of bounds of the box, set the x position as far right as we can without going past the box, with this.arrowSpacing to allow a small gap between the arrow and border
+        challengeLabel.rightArrow.setX(this.optionsWidth - this.arrowSpacing - this.optionsBg.rightWidth - challengeLabel.rightArrow.width);
+      }
+
+      // this line of code gets the center point between the left and right arrows from their left side (Arrow.x gives left point), taking into account the width of the arrows
+      const xLocation = challengeLabel.rightArrow.x - (challengeLabel.rightArrow.x - (challengeLabel.leftArrow.x + challengeLabel.leftArrow.width)) / 2 - 0.5;
```
MokaStitcher, line 259:
```suggestion
      // this line of code gets the center point between the left and right arrows from their left side (Arrow.x gives middle point), taking into account the width of the arrows
      const xLocation = Math.round((challengeLabel.leftArrow.x + challengeLabel.rightArrow.x + challengeLabel.leftArrow.displayWidth) / 2);;
```

leftArrow has its xOrigin to 0.5, so x is the middle point, which means the xLocation calculation can be simplified since challengeLabel also has the same xOrigin. also rounding it to avoid subpixel stuff

MokaStitcher (changes requested):
suggested some modifications to the placement calculations, otherwise looks good!

src/ui/challenges-select-ui-handler.ts
Opaque02: Fixed in next commit

MokaStitcher (approved):
👍

DayKev (approved):

src/data/challenge.ts
```diff
@@ -511,29 +507,25 @@ export class SingleGenerationChallenge extends Challenge {
...
-      overrideValue = this.value;
-    }
-    if (this.value === 0) {
+  getDescription(overrideValue?: number): string {
+    const value = overrideValue ?? this.value;
+    if (value === 0) {
       return i18next.t("challenges:singleGeneration.desc_default");
     }
-    return i18next.t("challenges:singleGeneration.desc", { gen: i18next.t(`challenges:singleGeneration.gen_${this.value}`) });
+    return i18next.t("challenges:singleGeneration.desc", { gen: i18next.t(`challenges:singleGeneration.gen_${value}`) });
```
flx-sta, line 528: Why isn't `this.geti18nKey()` used here?

src/data/challenge.ts
```diff
@@ -511,29 +507,25 @@ export class SingleGenerationChallenge extends Challenge {
...
    */
-  getDescription(overrideValue?: integer): string {
-    if (overrideValue === undefined) {
-      overrideValue = this.value;
-    }
-    if (this.value === 0) {
+  getDescription(overrideValue?: number): string {
+    const value = overrideValue ?? this.value;
+    if (value === 0) {
       return i18next.t("challenges:singleGeneration.desc_default");
```
flx-sta, line 526: Why isn't `this.geti18nKey()` used here?

flx-sta (approved):
Just some NITs.
Nothing wild

src/data/challenge.ts
DayKev: Probably just an oversight of the original author (or it wasn't a thing at the time), I added it to my challenge PR.

PR #4054, opened 2024-09-05
[Refactor] data/pokemon-species.ts 

frutescens:
None.

Pokemon Species, unlike individual Pokemon, have fixed qualities so they should not be editable + I think every code needs some form of clean up one way or another. 

All Pokemon Species class variables except for name, generation, and form Index have been set to readonly. They are still public, but can't be changed. Generations and Form Index have been set to protected and helper functions have been created to help with their retrieval/editing. 

## How to test the changes?
Run the game as normal.

## Checklist
    - [x] Are all unit tests still passing? (`npm run test`)\


src/data/pokemon-species.ts
```diff
@@ -178,15 +178,47 @@ export abstract class PokemonSpeciesForm {
...
+   */
+  getGeneration(): number {
+    return this.generation;
+  }
+
+  /**
+   * Method to set the generation of a specific form
+   * @param the generation of the form involved
+   */
+  setGeneration(generation: number) {
```
DayKev, line 193: What's the purpose of making it protected and then adding a public function that performs no extra logic to set the value?

Also you can use `get` / `set` (https://www.typescriptlang.org/docs/handbook/2/classes.html#getters--setters) instead of making functions like this, though again it seems pointless to do either here.

DayKev: Also if you had used `get`/`set` you would have only needed to change one file instead of 10 btw.

frutescens: It is because some Pokemon have cross-generational forms. Thank you for the link though. I'll go fix it up LOL.

DayKev: But your get/set functions don't actually perform any logic, it's identical to doing `generation = 3;` and `formIndex = 1;` without the functions.
Also you didn't change the `formIndex` functions to `get`/`set`.

DayKev (approved):

PR #4062, opened 2024-09-06
[Misc] Make `OPP_MOVESET_OVERRIDE` fully override the enemy's moveset

DayKev:
N/A

To get rid of the annoying need to use `Array(4).fill(Moves.MOVENAME)` when overriding the enemy's moveset.

The enemy's moveset array will be exactly set to the moveset passed to the override, rather than only modifying the existing moveset array (which could potentially leave some of the Pokémon's regular moves in the array).

### Screenshots/Videos
No override (opposing Pokémon has 2 default moves in its moveset array):
image

With override set to `[Moves.SPLASH]` (opposing Pokémon only has 1 move in its moveset array):
image

## How to test the changes?
Modify `overrides.ts` as follows and enter any battle.
```ts
const overrides = {
  OPP_MOVESET_OVERRIDE: [Moves.SPLASH]
```

## Checklist


torranx (commented):
looks good but I would prefer it to accept `Moves` or `Moves[]`

DayKev:
Both the player and enemy moveset overrides are now optionally arrays.

innerthunder:
The `SPLASH_ONLY` util isn't needed anymore after this, right? I suppose you can get rid of it with this PR...

DayKev:
Good point.

innerthunder (approved):

torranx (commented):
minor nit: there are movesets set to single item arrays, otherwise looks good

PR #4065, opened 2024-09-06
[Balance] Rogue ball increase catch chance tinted pokemon

Opaque02:
Rogue balls now have a 5x chance to catch pokemon instead of 3x if the pokemon has a tinted pokeball

Damo wanted it
image

The main thing is that the code to find out if a pokemon has a tinted pokeball has been moved to the pokemon class under ".isPokeballTinted()". With this change, it's now possible to call from within the pokeball class by adding a pokemon field to the getPokeballCatchMultiplier. I've also added a boolean to getPokeballCatchMultiplier which is used to get the description. This is an override that lets you get the boosted rate without having to enter a pokemon as an input, which is used in the shops for the description.

Lastly, I've moved some of the text around in the shop so that for pokeballs, the number of pokeballs you have shows under the pokeball so the description section can be open for the catch rate and description if needed. The locales PR is located [here](https://github.com/pagefaultgames/pokerogue-locales/pull/35)

### Screenshots/Videos
This wurmple below as a tinted pokeball because I didn't have the non-shiny form of it:
image

As you can see, it has a catch rate of 5 with rogue balls:
image

The wurmple below has a tinted pokeball because I don't have the ability:
image

Once again, it has a boosted catch rate of 5:
image

Below is a wurmple that does not have a tinted pokeball:
image

As you can see, it has a normal catch rate of 3:
image

In the shop, this is now what you see for rogue balls:

image

## How to test the changes?
Download the PR and change some overrides.

Firstly, the below override will let you add 5 rogue balls to the shop to test the wording on the screen. Add this to the top section of overrides: 

```
const overrides = {
	ITEM_REWARD_OVERRIDE: [{ name: "ROGUE_BALL", count: 5 }],

```

Secondly, you can change the pokeball overrides section to something like this:

```
readonly POKEBALL_OVERRIDE: { active: boolean; pokeballs: PokeballCounts } = {
    active: true,
    pokeballs: {
      [PokeballType.POKEBALL]: 5,
      [PokeballType.GREAT_BALL]: 0,
      [PokeballType.ULTRA_BALL]: 0,
      [PokeballType.ROGUE_BALL]: 99,
      [PokeballType.MASTER_BALL]: 99,
    },
  };
```

This will give you master balls and rogue balls you can use to test the catch rate.

## Checklist
- [x] Are the changes visual?


src/locales/en/modifier-type.json
```diff
@@ -2,7 +2,8 @@
   "ModifierType": {
     "AddPokeballModifierType": {
       "name": "{{modifierCount}}x {{pokeballName}}",
-      "description": "Receive {{pokeballName}} x{{modifierCount}} (Inventory: {{pokeballAmount}}) \nCatch Rate: {{catchRate}}"
+      "description": "Receive {{pokeballName}} x{{modifierCount}} (Inventory: {{pokeballAmount}}) \nCatch Rate: {{catchRate}}",
+      "catchRateGenerator": "{{normalCatchRate}}x ({{boostedCatchRate}}x for uncaught Pokemon)"
```
Tempo-anon, line 6: Uncaught is not quite the same as the tinted pokeball since the tinted pokeball relates to both the specific pokemon form's dex attributes (like having owned a male charizard but not a female one) as well as the root form's ability index

Adri1, line 6:
```suggestion
      "catchRateGenerator": "{{normalCatchRate}}x ({{boostedCatchRate}}x for uncaught Pokémon)"
```

Opaque02: Thanks! Fixed in next commit!

Opaque02: Ah, that's a very good point - what wording would we want to use instead? "{{boostedCatchRate}}x for tinted pokeballs? We're a little limited on space is the only issue?

src/data/pokeball.ts
```diff
@@ -48,7 +49,9 @@ export function getPokeballName(type: PokeballType): string {
   return ret;
 }
 
-export function getPokeballCatchMultiplier(type: PokeballType): number {
+export function getPokeballCatchMultiplier(type: PokeballType, enemyPokemon?: Pokemon, getBoostedDescription?: boolean): number {
+  const pokemon = enemyPokemon ?? null;
+  const description = getBoostedDescription ?? false;
```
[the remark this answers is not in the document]
Opaque02: Would doing it the way I've already got it not be safer for things as a just in case? Or would it not make a difference in this case with how I've got it?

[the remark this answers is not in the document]
DayKev: `null` and `undefined` are both treated the same way in this scenario (for `enemyPokemon`), and both also evaluate to `false` when used in an if statement (for `getBoostedDescription`).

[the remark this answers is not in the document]
Opaque02: Perfect! I think I was in the mindset of fixing the challenges which used a similar change to what I did here - let me push a  change through now!

src/locales/en/modifier-type.json
Opaque02: @Tempo-anon it should now be clearer - let me know what you think

src/data/pokeball.ts
```diff
@@ -57,6 +58,12 @@ export function getPokeballCatchMultiplier(type: PokeballType): number {
   case PokeballType.ULTRA_BALL:
     return 2;
   case PokeballType.ROGUE_BALL:
+    /* making rogue balls have a higher chance to catch pokemon if they have a tinted pokeball (i.e. at least one thing is new for them)
+       * you can also get the boosted type for a description (i.e. for modifier-type)
+       */
```
DayKev, line 63:
```suggestion
    // Rogue balls have a higher chance to catch pokemon if they have a tinted pokeball icon (i.e. at least one thing is new for them).
    // You can also get the boosted type for a description (i.e. for modifier-type).
```
> You can also get the boosted type for a description (i.e. for modifier-type).

What exactly does this mean?

src/field/pokemon.ts
```diff
@@ -3783,6 +3783,37 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
     return rootForm.getAbility(abilityIndex) === rootForm.getAbility(currentAbilityIndex);
   }
+
+  /* Checks whether or not the pokemon's pokeball will be tinted.
+   * This checks the dex attributes (gender, shiny/non shiny, variant and forms)
+   * These values are checked against the exact pokemon you're catching (i.e. catching charizard won't compare against charmander)
+   * It also checks your abilities against the starter version
+   * If any are "new", the pokemon's pokeball is tinted
+   * @returns true if the pokemon's pokeball should be tinted, meaning at least one thing is new
+   */
```
DayKev, line 3793:
```suggestion
  /*
   * Checks whether or not the pokemon's pokeball icon will be tinted.
   * This checks the dex attributes (gender, shiny/non shiny, variant and forms).
   * These values are checked against the exact pokemon you're catching (i.e. catching charizard won't compare against charmander).
   * It also checks your abilities against the starter version.
   * If you are missing any, the pokemon's pokeball icon is tinted.
   * @returns `true` if the pokemon's pokeball icon should be tinted, meaning at least one thing is new
   */
```

src/modifier/modifier-type.ts
```diff
@@ -180,15 +181,32 @@ class AddPokeballModifierType extends ModifierType {
     return i18next.t("modifierType:ModifierType.AddPokeballModifierType.name", {
       "modifierCount": this.count,
       "pokeballName": getPokeballName(this.pokeballType),
+      "pokeballAmount": `${this.currentAmount}`,
     });
   }
 
+  addAmount(currentAmount: number) {
+    this.currentAmount = currentAmount;
+  }
```
DayKev, line 190: This function doesn't seem to be named properly (or it's not working as it should), it's setting an amount and not adding it.

Tempo-anon:
How does this work with fresh start since theoretically this would mean the more full your dex is the technically harder fresh start would be

src/field/pokemon.ts
```diff
@@ -3783,6 +3783,37 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    let playerOwnsThisAbility = false;
+    // Check if the player owns ability for the root form
+    if ((ownedAbilityAttrs & AbilityAttr.ABILITY_1) > 0 && this.hasSameAbilityInRootForm(0)) {
+      playerOwnsThisAbility = true;
+    }
+    if ((ownedAbilityAttrs & AbilityAttr.ABILITY_2) > 0 && this.hasSameAbilityInRootForm(1)) {
+      playerOwnsThisAbility = true;
+    }
+    if ((ownedAbilityAttrs & AbilityAttr.ABILITY_HIDDEN) > 0 && this.hasSameAbilityInRootForm(2)) {
+      playerOwnsThisAbility = true;
```
Tempo-anon, line 3812: I ended up just taking this whole chunk into its own helper function in https://github.com/pagefaultgames/pokerogue/pull/4307 since it seems to be needed in the pokemon-info-container as well

src/data/pokeball.ts
Opaque02: The boosted type for the description in shops. Since normally it's just listed as "Catch rate: 3x", there needs to be a way to get the boosted value without a pokemon. The getBoostedDescription boolean does that, so if that's passed, then the it's for the shop description.

src/field/pokemon.ts
Opaque02: Ah good point - want me to change it based on your or make a new change once yours goes in?

src/modifier/modifier-type.ts
Opaque02: True, fixed in next commit

Opaque02:
> How does this work with fresh start since theoretically this would mean the more full your dex is the technically harder fresh start would be

As per Damo on discord, only for fresh start rogue balls should be at a constant 5x rate; I'll get this updated

EnochG1 (changes requested):
If you modify, not add new text,
all translations become mistranslated if you do not modify the text in all locale files to changed English text.
(I thing wrong translation is worse than untranslated right English)
I would request to change all locales/{lang}/modifire-type.json.

Opaque02:
> If you modify, not add new text, all translations become mistranslated if you do not modify the text in all locale files to changed English text. (I thing wrong translation is worse than untranslated right English) I would request to change all locales/{lang}/modifire-type.json.

@EnochG1 - can you elaborate a bit more on what I should change? We don't need to put the new keys in manually these days, so  just want to double check about what I should change for this one

Tempo-anon:
4307 has been merged in. You can keep the `isTintedPokeball` function but you can now refactor it based on the other helper function I added. (And can also update pokemon-info-container and battle-info to use the new `isTintedPokeball`)

As for a constant 5x for fresh start... I'm not sure. Sure it works but there's a bunch of edge cases for that. Would it really be that broken for them to be a constant 5x boost? Pokeballs in general feel pretty underpowered for the tiers they're in. (5 ultra balls is the equivalent of 10 pokeballs has to compete with things like reviver seeds, rare evo items, mints, competitive items and 5 rogue balls which is the equivalent of 15-25 pokeballs has to compete with form change items and really competitive items like leftovers, shell bell, king's rock, etc)

src/field/pokemon.ts
Opaque02: Fixed and updated - should be in latest commit

MokaStitcher (changes requested):
Haven't reviewed the rest yet, but it seems Skorupi's sprite got accidentally modified during of the merges probably, please fix :pray:

chaosgrimmon:
> Haven't reviewed the rest yet, but it seems Skorupi's sprite got accidentally modified during of the merges probably, please fix 🙏

Yeah, those had to be hotfixed for Main because they'd cause a crash when used as a fusion material. The preferable solution would be to keep the .pngs, and use the .json [here instead](https://discord.com/channels/1125469663833370665/1290008413975351317/1290938914122633247)

This one might be the download link directly:
https://cdn.discordapp.com/attachments/1290008413975351317/1290938913774501909/451.json?ex=671d434d&is=671bf1cd&hm=f289a034970205777c842f39687efa0e47c4f05c8caabf643a291eafd5099cac&

MokaStitcher:
> The preferable solution would be to keep the .pngs, and use the .json here instead

Yeah but I don't think it should be part of this PR, I'd rather have a separate PR to fix Skorupi and just revert the changes from this PR

DayKev:
> > The preferable solution would be to keep the .pngs, and use the .json here instead
> 
> Yeah but I don't think it should be part of this PR, I'd rather have a separate PR to fix Skorupi and just revert the changes from this PR

It's just a git history issue, the change isn't part of this PR (it's in the f7715d7 merge commit).

src/modifier/modifier-type.ts
```diff
@@ -179,15 +180,32 @@ class AddPokeballModifierType extends ModifierType {
...
+      });
+    } else if (this.pokeballType !== PokeballType.MASTER_BALL) {
+      catchRate = `${getPokeballCatchMultiplier(this.pokeballType)}x`;
+    }
     return i18next.t("modifierType:ModifierType.AddPokeballModifierType.description", {
       "modifierCount": this.count,
       "pokeballName": getPokeballName(this.pokeballType),
-      "catchRate": getPokeballCatchMultiplier(this.pokeballType) > -1 ? `${getPokeballCatchMultiplier(this.pokeballType)}x` : "100%",
-      "pokeballAmount": `${scene.pokeballCounts[this.pokeballType]}`,
+      "catchRate": catchRate,
```
PigeonBar, line 208: The master ball's catch rate is no longer being displayed:

20241030 Master Catch Rate

PigeonBar:
ME's have their own copy of the capture attempt function, which probably also needs updating (Line 376):
https://github.com/pagefaultgames/pokerogue/blob/b3a0ca94753d13dd110c17f1856872a7b2dc9890/src/data/mystery-encounters/utils/encounter-pokemon-utils.ts#L369-L380

Tempo-anon:
I still think this change while is cool in theory, seems to be quite complicated as it involves changing a LOT of files, not to mention a lot of locale changes and thoughts on how to future proof it for various challenges.

Hell, just make rogue balls 5x or 10x across the board no matter what because they suck compared to the other things in the rogue tier.

DayKev:
> ME's have their own copy of the capture attempt function, which probably also needs updating (Line 376):
> 
> https://github.com/pagefaultgames/pokerogue/blob/b3a0ca94753d13dd110c17f1856872a7b2dc9890/src/data/mystery-encounters/utils/encounter-pokemon-utils.ts#L369-L380

Wait what? Why?

PigeonBar:
> > ME's have their own copy of the capture attempt function, which probably also needs updating (Line 376):
> > https://github.com/pagefaultgames/pokerogue/blob/b3a0ca94753d13dd110c17f1856872a7b2dc9890/src/data/mystery-encounters/utils/encounter-pokemon-utils.ts#L369-L380
> 
> Wait what? Why?

Right now, I think it's just used for Safari Zone

MokaStitcher:
> It's just a git history issue, the change isn't part of this PR (it's in the [f7715d7](https://github.com/pagefaultgames/pokerogue/commit/f7715d7d10469cce8ac829e6c367bec794901535) merge commit).

No, it's an issue with this branch. Here's how the Skorupi sheet looks in beta (taller than it is wide)
Screenshot_20241030-225049~2

Here's how it is on Opaquer's branch (wider than it is tall. This is the one that got added then reverted some time ago. It shouldn't get back in as it is broken, and has nothing to do with this PR anyway. Something went wrong during one of the merges to this branch, it's fine as long as a new commit it made that reverts the changes)
Screenshot_20241030-225535~2

SirzBenjie:
Closing this as it is an old draft and it has many conflicts. If this behavior is desired, it can be reopened.

Issue #3991, opened 2024-09-03
[Bug] Scrappy prevents Super Effective attacks against Ghost types in Inverse challenge


YamiTheYin:
### Describe the bug

Fighting is x2 effective against Ghost with inverse rules. However a Pokemon with the Scrappy ability is unable to land a super effective Fighting type attack on a Ghost type Pokemon (with no other types that should otherwise change effectiveness).

### Session export file

[sessionData1_YamitheYin.txt](https://github.com/user-attachments/files/16842650/sessionData1_YamitheYin.txt)

### User data export file

[data_YamitheYin.txt](https://github.com/user-attachments/files/16842653/data_YamitheYin.txt)

### Expected behavior

<img width="445" alt="image" src="https://github.com/user-attachments/assets/18f9b68b-1864-460c-8e5e-e6bfbbe8ceeb">

### Screenshots / Videos

https://github.com/user-attachments/assets/1506fd0d-b134-464e-ad67-927cdab9f5f7

### What OS did you observe the bug on?

PC/Windows

### Which browser do you use?

Chrome

KimJeongSun:
it is fixed on #4067

flx-sta:
@KimJeongSun the PR is still open. Only close if merged into `beta` (or `main`) please

PR #4067, opened 2024-09-06
[Bug] Fix scrappy (+ some immunity move and ability) in inverse battle

KimJeongSun:

Scrappy, Mind's Eye abilities and the moves Foresight, Odor Sleuth, Miracle Eye no longer make moves that should deal 2x damage instead deal 1x damage in inverse battle.
These abilities and moves no longer affect type matchups in inverse battles!

In Inverse Battles, these moves and abilities should not nullify type immunities (e.g., Normal -> Ghost, Psychic -> Dark), so they should have no effect. ( which are already 2x damage in inverse battle )
However, the existing code caused these moves and abilities to forcibly make the attacks deal 1x damage against these types, which is a bug.

The code has been modified so that these moves and abilities do not alter type effectiveness in Inverse Mode. 
Additionally, new test cases for Inverse Battles have been added ( + including tests to confirm that Thunder Wave can affect Ground-type Pokémon ) (+ inverse battle test code refactoring! requested by @torranx )

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/test/battle/inverse_battle.test.ts
+++ b/src/test/battle/inverse_battle.test.ts
@@ -1,8 +1,5 @@
 import { BattlerIndex } from "#app/battle";
-import { allMoves } from "#app/data/move";
 import { Type } from "#app/data/type";
-import { MoveEndPhase } from "#app/phases/move-end-phase";
-import { TurnEndPhase } from "#app/phases/turn-end-phase";
 import { Abilities } from "#enums/abilities";
 import { ArenaTagType } from "#enums/arena-tag-type";
 import { Challenges } from "#enums/challenges";
@@ -11,7 +8,8 @@ import { Species } from "#enums/species";
 import { StatusEffect } from "#enums/status-effect";
 import GameManager from "#test/utils/gameManager";
 import Phaser from "phaser";
-import { afterEach, beforeAll, beforeEach, describe, expect, it } from "vitest";
+import { SPLASH_ONLY } from "#test/utils/testUtils";
+import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
 
 const TIMEOUT = 20 * 1000;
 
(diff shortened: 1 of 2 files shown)


Snailman11:
Closes [Bug] Scrappy prevents Super Effective attacks against Ghost types in Inverse challenge
[#3991](https://github.com/pagefaultgames/pokerogue/issues/3991)

src/test/battle/inverse_battle.test.ts
```diff
@@ -200,4 +218,35 @@ describe("Inverse Battle", () => {
...
+    game.override
+      .ability(Abilities.SCRAPPY)
+      .enemySpecies(Species.GASTLY);
+
+    await game.challengeMode.startBattle();
+
+    const player = game.scene.getPlayerPokemon()!;
+    const enemy = game.scene.getEnemyPokemon()!;
+
+    expect(enemy.getMoveEffectiveness(player, allMoves[Moves.TACKLE])).toBe(2);
```
torranx, line 232: you should not manually call the function but instead use spyOn on the function, for reference you can check `tera_shell.ts`

src/test/battle/inverse_battle.test.ts
```diff
@@ -200,4 +218,35 @@ describe("Inverse Battle", () => {
...
+    await game.challengeMode.startBattle();
+
+    const player = game.scene.getPlayerPokemon()!;
+    const enemy = game.scene.getEnemyPokemon()!;
+
+    game.move.select(Moves.FORESIGHT);
+    await game.setTurnOrder([BattlerIndex.ENEMY, BattlerIndex.PLAYER]);
+    await game.phaseInterceptor.to(TurnEndPhase);
+
+    expect(enemy.getMoveEffectiveness(player, allMoves[Moves.TACKLE])).toBe(2);
```
torranx, line 250: same here

torranx (commented):
nit: I personally don't like how the tests are numbered, but should not block this PR

src/test/battle/inverse_battle.test.ts
KimJeongSun: Are there any advantages to using the provided code, or are there disadvantages to the existing test code method? 

When I tried it just now, the suggested method seems to be influenced by the priority of the Pokémon. Additionally, it appears impossible to check values for 0x (no effect) attacks.

KimJeongSun: updated! can you review it again?

innerthunder (approved):

DayKev (approved):
The tests have some redundant code that could be removed but that can be done in a later PR.

PR #4070, opened 2024-09-06
[Bug] Fix incorrect defensive properties on Terastallized Pokemon

innerthunder:
This fixes issues related to the defensive typing resolution of Terastallized Pokemon:
- Tera-Grass Pokemon are now correctly immune to powder moves
- Tera-Dark Pokemon are now correctly immune to Prankster-boosted status moves

This was reported [in the Discord](https://discord.com/channels/1125469663833370665/1281673920029921321)

- `field/pokemon`:
  - `Pokemon#getTypes` now only returns the Pokemon's Tera type when evaluating a Tera'd Pokemon's types defensively
  - `Pokemon#getMoveEffectiveness` uses `getTypes(true, true)` instead of `getTypes()` when evaluating `move.isTypeImmune(...)`. This means that for added immunities for powder moves and Prankster-boosted status moves:
    1. The defending Pokemon's Tera type is no longer ignored.
    2. If the defending Pokemon is Tera'd, these immunities are based *only* on the Tera type.

## How to test the changes?
`npm run test effectiveness` (5 new tests)

## Checklist
- [x] Are the changes visual?


DayKev:
Does the extra grass type added by Forest's Curse currently work correctly with powder moves? If not, would this fix that? What about what + tera types?

innerthunder:
> Does the extra grass type added by Forest's Curse currently work correctly with powder moves? If not, would this fix that? What about what + tera types?

It should work correctly, as changes to `summonData.types` are already accounted for. Also, Tera types are the only types evaluated when the Pokemon is terastallized and defending against a move, so afaik Forest's Curse behaves as Bulba says:

> Forest's Curse will always fail if the target is [Terastallized](https://bulbapedia.bulbagarden.net/wiki/Terastal_phenomenon), regardless of its type, and its effect is removed upon Terastallizing.

DayKev (approved):

frutescens (approved):

torranx (approved):

PR #4074, opened 2024-09-06
[Feature] Toggle Pokemon cries (during battle)

flx-sta:

A new setting will be available to turn PKM cries on/off during battle. 
> [!WARNING]
> This will not affect cries in the starter-select screen or in the evolution phase!

It's a highly requested feature and can also speed up tests

The `BattleScene` has a new field called `enablePokemonCries` (Same old...).
This fields value is checked during `faintCry()`, `encounter-phase` and `summon-phase`.
Some callback delay values are automatically adjusted too as we don't want the player to have to wait for a cry that isn't even being played.

### Screenshots/Videos

image

## How to test the changes?

- Adjust the setting to on or off
- Play 1 wave and make sure that both your pokemon and the enemy pokemon is fainted to check if the faint cries are also (not) playing

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/locales/en/settings.json
+++ b/src/locales/en/settings.json
@@ -104,5 +104,5 @@
   "reroll": "Reroll",
   "shop": "Shop",
   "checkTeam": "Check Team",
-  "pokemonCries": "Pokemon Cries"
+  "pokemonCries": "Pokémon Cries"
 }
--- a/src/locales/en/settings.json
+++ b/src/locales/en/settings.json
@@ -103,5 +103,6 @@
   "rewards": "Rewards",
   "reroll": "Reroll",
   "shop": "Shop",
-  "checkTeam": "Check Team"
+  "checkTeam": "Check Team",
+  "pokemonCries": "Pokemon Cries"
 }
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -140,6 +140,8 @@ export default class BattleScene extends SceneBase {
   public enableMoveInfo: boolean = true;
   public enableRetries: boolean = false;
   public hideIvs: boolean = false;
+  /** Toggle Pokemon cries */
+  public enablePokemonCries: boolean = true;
   /**
    * Determines the condition for a notification should be shown for Candy Upgrades
    * - 0 = 'Off'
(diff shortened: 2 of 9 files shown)


Adri1 (dismissed):

src/locales/en/settings.json
```diff
@@ -103,5 +103,6 @@
   "rewards": "Rewards",
   "reroll": "Reroll",
   "shop": "Shop",
-  "checkTeam": "Check Team"
+  "checkTeam": "Check Team",
+  "pokemonCries": "Pokemon Cries"
```
Adri1, line 107:
```suggestion
  "pokemonCries": "Pokémon Cries"
```

Adri1 (approved):

frutescens (changes requested):
Wouldn't it make more sense to go to the function cry() in pokemon-species.ts and disable the cry there? This way, cries will be disabled at the source. Also, wouldn't it be better to check / set scene.fieldVolume to 0 and use that in pokemon-species.ts' cry() instead of creating a new setting?

frutescens:
Mm, I did take a look at faintCry() and fusionFaintCry() and they do seem fairly clunky. It is surprising that they're very different from cry() but I do get why it's not so easily cut and dry.

DayKev:
If it only affects cries during battle and not elsewhere (ie: evolution) then maybe the label in the settings should be "Pokémon Cries In Battle [y/n]"

DayKev:
Btw I checked the test speed-up vs beta and:

Before:
image
After:
image

Not incredible or anything, but it's decent. However, when one or more Pokémon faint:

Before:
image
After:
image

Insane. `FaintPhase` no longer takes multiple whole seconds (!).

flx-sta:
> Wouldn't it make more sense to go to the function cry() in pokemon-species.ts and disable the cry there? This way, cries will be disabled at the source. Also, wouldn't it be better to check / set scene.fieldVolume to 0 and use that in pokemon-species.ts' cry() instead of creating a new setting?

This might be something to consider… yeah.
Is field only cries and faint cries anyway ?

regarding not changing `cry()`: cause I didn’t want it to affect evolutions & starter select cry’s. But if that is affected by “field” volume too I would probably cut it out too

frutescens:
Yes, field volume covers all cries + battle sounds. I think this feature is useful for devs, but not so useful for users (although I could see performance arguments for putting similar checks for other sound cues like BGM)

flx-sta:
@frutescens may take over this feature

PR #4080, opened 2024-09-07
[Dev] Save Data Version Migration

xsn34kzx:

Ideally, none; existing run history and sessions should retain all their existing modifiers and held items when transitioning from the existing `main` branch to the incoming `beta` branch.

Because of #2699 and #4032, vitamins, battle items, lures, and `White Herb` would break any outdated sessions on parse and load; this PR essentially aims to coalesce save migration into a set of functions called once for migrating over old data into new formats, naming schemes, etc.

Essentially created three separate functions that follow `SessionSaveData`, `SystemSaveData`, and the `settings` object respectively on load. Each function uses a `switch-case` block intentionally with no `break`s to let patches apply from the top-down in order of older to newer versions. By the end of that block, all necessary save data migration should have been applied, and the game version is manually incremented.

### Screenshots/Videos

TBD

## How to test the changes?

Load up an outdated session, system data, or settings and ensure that no errors or irregular behavior occurs (e.g., missing battle items, lures, or vitamins).

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/system/version-converter.ts
+++ b/src/system/version-converter.ts
@@ -1,0 +1,73 @@
+import { allSpecies } from "#app/data/pokemon-species.js";
+import { AbilityAttr, defaultStarterSpecies, DexAttr, SessionSaveData, SystemSaveData } from "./game-data";
+import { SettingKeys } from "./settings/settings";
+
+const LATEST_VERSION = process.env.npm_package_version!;
+
+export function sessionVersionConverter(data: SessionSaveData) {
+  const curVersion = data.gameVersion;
+  if (curVersion !== LATEST_VERSION) {
+    switch (data.gameVersion) {
+    case "1.0.4":
+      // Session data fixing would go here
+    }
+  }
+}
+
+export function systemVersionConverter(data: SystemSaveData) {
+  const curVersion = data.gameVersion;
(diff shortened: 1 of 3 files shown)


src/system/version-converter.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export function applySessionDataPatches(data: SessionSaveData) {
+  const curVersion = data.gameVersion;
+  if (curVersion !== LATEST_VERSION) {
+    switch (curVersion) {
+    case "1.0.0":
+    case "1.0.1":
+    case "1.0.2":
+    case "1.0.3":
+    case "1.0.4":
```
frutescens, line 15: May be good to use comments to provide approximate dates for future developers

frutescens (approved):
image

src/system/version-converter.ts
xsn34kzx: I agree for when we're about to release since the biggest problem with the earlier versions is that there is no approximate date for any of them or when they changed (not that those version numbers matter too much since they weren't incremented for other releases). We can either do it in this PR or in the `beta` merge 🫡

f-fsantos (approved):

PR #4098, opened 2024-09-08
[Test] Fix Safeguard test that relied on a now-fixed bug

DayKev:
N/A

The test is broken.

The enemy Pokémon is now damaged before using Rest so that it will properly activate.

## How to test the changes?
`npm run test safeguard`

## Checklist

### DIFF
--- a/src/test/moves/safeguard.test.ts
+++ b/src/test/moves/safeguard.test.ts
@@ -38,7 +38,7 @@ describe("Moves - Safeguard", () => {
   });
 
   it("protects from damaging moves with additional effects", async () => {
-    await game.startBattle();
+    await game.classicMode.startBattle();
     const enemy = game.scene.getEnemyPokemon()!;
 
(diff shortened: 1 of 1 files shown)


Tempo-anon (approved):

torranx (commented):
can you please provide more info on "relied on a now-fixed bug"? I checked the files changed and all I see is the addition of `.classicMode` and `damageAndUpdate(1)` so it's not apparent to me

DayKev:
Statuses like Burn and Poison used to tick immediately when weren't supposed to (ie: because you got it from a Flame Orb or you just switched in to Toxic Spikes), so the enemy Pokémon in the test would be damaged and Rest would succeed. When that got fixed, the enemy Pokémon in the test no longer was damaged when it tried to use Rest and so the move failed (which meant it wasn't asleep, which the test was checking for).

PR #4104, opened 2024-09-08
[Hotfix] Properly whitelisting fused Pokemon that happen to have Pokemon with alt. forms

frutescens:
No crashes hopefully.

My feature, my mistake. 

I originally got the method to retrieve fusion cries from fusionFaintCry(). However, this method was incorrect and would return the wrong cry key if the Pokemon was not in its default form. This did not lead to any awkward situations originally.

However, when I made my garbage collection, the wrong cry key retrieved with this method meant that the wrong key could be whitelisted and cries like 'cry/115-mega' would not be whitelisted and end up deleted during garbage collection even though the latter would be actively in use.  

## How to test the changes?
Hatch eggs with a fused Pokemon with both parts having different forms as a lead. Should not crash.  

## Checklist


Tempo-anon (approved):

DayKev:
You forgot to base the branch of main instead of beta.

Snailman11:
Discord Bug Report: (Egg opening summary screen crashes game to restart)
https://discord.com/channels/1125469663833370665/1282185799542116442

PR #4105, opened 2024-09-08
[Hotfix] Allow Fake Out to be used at the start of every wave again

DayKev:
Fake Out and First Impression can be used at the start of every wave.

Request from @damocleas due to player feedback.

The active Pokémon's turn count is set to 1 at the start of every wave, so first-turn-only moves won't fail.

## How to test the changes?
`npm run test fake_out`

## Checklist
- ~[ ] **I'm using `beta` as my base branch**~

### DIFF
--- a/src/phases/battle-end-phase.ts
+++ b/src/phases/battle-end-phase.ts
@@ -23,6 +23,16 @@ export class BattleEndPhase extends BattlePhase {
       this.scene.unshiftPhase(new GameOverPhase(this.scene, true));
     }
 
+    /**
+     * Allow Fake Out and First Impression to be used at the start of every wild battle
+     * Note: This is specifically a buff to those moves and not normally expected behavior
+     */
+    for (const pokemon of this.scene.getField()) {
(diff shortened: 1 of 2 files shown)


innerthunder (approved):
lgtm 🚀

damocleas (approved):

damocleas:
Please test with mat block too

DayKev:
Mat Block also works.

src/phases/battle-end-phase.ts
```diff
@@ -23,6 +23,16 @@ export class BattleEndPhase extends BattlePhase {
       this.scene.unshiftPhase(new GameOverPhase(this.scene, true));
     }
 
+    /**
+     * Allow Fake Out and First Impression to be used at the start of every wild battle
```
Tempo-anon, line 27: If you just want to make a change to two specific moves wouldn't this make more sense as a move attribute rather than doing something that might cause unintended behavior by manually changing a Pokemon's turns out?

Tempo-anon (changes requested):

DayKev:
image

PR #4106, opened 2024-09-08
[Bug] Obstruct et al no longer protect against status moves

DayKev:
[Burning Bulwark](https://dex.pokemonshowdown.com/moves/burningbulwark), [King's Shield](https://dex.pokemonshowdown.com/moves/kingsshield), [Obstruct](https://dex.pokemonshowdown.com/moves/obstruct) and [Silk Trap](https://dex.pokemonshowdown.com/moves/silktrap) no longer protect against status moves.

These moves are not supposed to protect against status moves but currently they do.

A new `BattlerTag` class `DamageProtectedTag` was created. `ContactStatStageChangeProtectedTag` and `ContactBurnProtectedTag` now inherit from that instead of `ProtectedTag`, and the protection check in `MoveEffectPhase` was updated to account for the change.

## How to test the changes?
`npm run test obstruct`

## Checklist


src/test/moves/obstruct.test.ts
```diff
@@ -0,0 +1,71 @@
+import { Moves } from "#app/enums/moves";
+import { Stat } from "#app/enums/stat.js";
```
ben-lear, line 2: nit: import .js

ben-lear, line 2:
```suggestion
import { Stat } from "#app/enums/stat";
```

ben-lear (changes requested):

src/test/moves/obstruct.test.ts
DayKev: :despair:

innerthunder (approved):
lgtm, though maybe a better name for the new tag would be `AttackMoveProtectedTag` (or just `AttackProtectedTag`)?

schmidtc1 (approved):

src/data/battler-tags.ts
```diff
@@ -1180,6 +1180,8 @@ export class ProtectedTag extends BattlerTag {
   }
 }
 
+export class DamageProtectedTag extends ProtectedTag {}
```
Tempo-anon, line 1183: Can we get some tsdoc on these new classes

src/data/battler-tags.ts
```diff
@@ -1215,7 +1217,7 @@ export class ContactDamageProtectedTag extends ProtectedTag {
   }
 }
 
-export class ContactStatStageChangeProtectedTag extends ProtectedTag {
+export class ContactStatStageChangeProtectedTag extends DamageProtectedTag {
```
Tempo-anon, line 1220: Each of these classes should have a tsdoc that explains what they're doing and which moves they're associated with

Tempo-anon (changes requested):

Tempo-anon (approved):
Baneful bunker is somehow different. Gamefreak why.

PR #4113, opened 2024-09-08
[Bug] Fixing seed sower uturn switchout bug for trainer battles

DustinLin:

For trainer battles, the game currently freezes/crashes when an enemy pokemon uses a switch-out move on a pokemon with the Seed Sower ability. It doesn't now:

- current `beta` behavior

https://github.com/user-attachments/assets/33950ed6-b720-4de5-8ef6-4e6be307256a

The fix will probably solve any other animations that play during a switch out (eg: other similar abilities)

Bug reported on discord

The problem is that the seed-sower animation is attempting to play/target a pokemon that currently isn't on the field (they are in the middle of switching out). Currently the animation functions in `battle-anims.ts` use `getPlayerPokemon()` and `getEnemyPokemon()` to grab sprites to play the animations. When the sprite is not visible, it crashes the game.

I've refactored the solution from #3346 (which solved the animation problem for wild battles) to work in this case as well.

Finally i've updated the `dragon_tail` tests with the refactoring

### Screenshots/Videos

`beta` behavior from above

- `pr` for single trainer battles

https://github.com/user-attachments/assets/d35f2eda-767a-4f3e-8a3d-ac740da3cc15

- `pr` for double trainer battles

https://github.com/user-attachments/assets/5b9eaf1b-8ebd-46a5-9ec3-8bed2974eaf6

For sanity I included the wild versions as well
- `pr` for single wild battles

https://github.com/user-attachments/assets/cfe3bd34-4d5e-4f5f-9de9-f7065ed7668d

- `pr` for double wild battles

https://github.com/user-attachments/assets/aeb3ba1e-9d71-48d0-a1a9-b66ebe4e2bd7

## How to test the changes?

Can try single/double trainer battles (wave 8, 25)

## Checklist
- [x] Are the changes visual?


src/battle-scene.ts
```diff
@@ -749,6 +749,14 @@ export default class BattleScene extends SceneBase {
     return this.getPlayerField().find(p => p.isActive());
   }
 
+  /**
+   * Finds the first {@linkcode Pokemon.isActive()} pokemon from the player that isn't also currently switching out
+   * @returns Either the first {@linkcode PlayerPokemon} satisfying, or undefined if no player pokemon on the field satisfy
+   */
```
ben-lear, line 755: Nit: the name for these functions does not seem to describe their true purpose accurately. Perhaps “getNonSwitchPlayerPokemon”?

src/battle-scene.ts
```diff
@@ -766,6 +774,14 @@ export default class BattleScene extends SceneBase {
     return this.getEnemyField().find(p => p.isActive());
   }
 
+  /**
+   * Finds the first {@linkcode Pokemon.isActive()} pokemon from the enemy that isn't also currently switching out
+   * @returns Either the first {@linkcode EnemyPokemon} satisfying, or undefined if no player pokemon on the field satisfy
+   */
+  getEnemyPokemonOnScreen(): EnemyPokemon | undefined {
```
ben-lear, line 781: Nit: the name for these functions does not seem to describe their true purpose accurately. Perhaps “getNonSwitchEnemyPokemon”?

ben-lear (dismissed):
A couple nits but otherwise lgtm

src/data/battle-anims.ts
```diff
@@ -887,7 +887,7 @@ export abstract class BattleAnim {
                 const setSpritePriority = (priority: integer) => {
                   switch (priority) {
                   case 0:
-                    scene.field.moveBelow(moveSprite as Phaser.GameObjects.GameObject, scene.getEnemyPokemon() || scene.getPlayerPokemon()!); // TODO: is this bang correct?
+                    scene.field.moveBelow(moveSprite as Phaser.GameObjects.GameObject, scene.getEnemyPokemonOnScreen() || scene.getPlayerPokemonOnScreen()!); // this bang is correct to ensure a GameObject is passed in as a second arg (as opposed to undefined)
```
DayKev, line 890: That's not what `!` does, it doesn't force anything to be defined it just tells the ts compiler that it's incorrect in thinking something can possibly be `undefined` (which isn't true in this case, those functions can explicitly be `undefined`).

DustinLin: i see, then the question to ask is: will there be a case where all pokemon sprites aren't visible on the screen?

The original code thinks no, and I can't think of a case either. It could be the case that both pokemon at off screen (like fly or something), but things like seed sower animation shouldn't be playing at those times either.

src/battle-scene.ts
DustinLin: good point, since the function looks at the switchout status (and not directly the visibility of the pokemon on screen), I should change it

src/battle-scene.ts
```diff
@@ -749,6 +749,14 @@ export default class BattleScene extends SceneBase {
     return this.getPlayerField().find(p => p.isActive());
   }
 
+  /**
+   * Finds the first {@linkcode Pokemon.isActive()} pokemon from the player that isn't also currently switching out
```
torranx, line 753: nit:
```suggestion
   * Finds the first {@linkcode Pokemon.isActive() | active PlayerPokemon} that isn't also currently switching out
```

src/battle-scene.ts
```diff
@@ -766,6 +774,14 @@ export default class BattleScene extends SceneBase {
     return this.getEnemyField().find(p => p.isActive());
   }
 
+  /**
+   * Finds the first {@linkcode Pokemon.isActive()} pokemon from the enemy that isn't also currently switching out
```
torranx, line 778: nit:
```suggestion
   * Finds the first {@linkcode Pokemon.isActive() | active EnemyPokemon} that isn't also currently switching out
```

src/field/pokemon.ts
```diff
@@ -1974,11 +1974,11 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
   }
 
   /**
-   * sets if the pokemon has fled (implies it's a wild pokemon)
+   * sets if the pokemon is switching out (if it's a enemy wild implies it's going to flee)
    * @param status - boolean
    */
-  setWildFlee(status: boolean): void {
-    this.wildFlee = status;
+  setSwitchOut(status: boolean): void {
```
torranx, line 1980:
```suggestion
  setSwitchOutStatus(status: boolean): void {
```

DustinLin: thanks, that's a better name, will change

DayKev (approved):

DayKev (approved):

PR #3326, opened 2024-08-04
[Move] Heal Block

frutescens:
## What are the changes?
This implements Heal Block and provides a test for it. 
Link to Old PR: https://github.com/pagefaultgames/pokerogue/pull/3047 

## Why am I doing these changes?
I think it would be nice for it to be implemented. 

## What did change?
- HealBlockTag (Duration: 5 Turns)
- PokemonHealPhase now accounts for HealBlockTag 
- Not implemented yet: moves restricted by heal block 
- Partially implemented: Tests 
- Lines for Localization 

Waiting for this implementation to finish up. 

### Screenshots/Videos
https://github.com/user-attachments/assets/08b71f5d-3325-437d-b4d4-975b7f803973

## How to test the changes?
Use Heal Block during a Pokemon battle and see if the affected pokemon can heal itself. 

## Checklist
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #4120, opened 2024-09-09
[Move] Implement Heal Block

frutescens:
Old PRs: #3326 #3047 

Heal Block will be implemented. 
Psychic Noise will have its secondary effect implemented - Heal Block will be applied to the opponent for two turns. 

Attempt to get rid of partial tags. 

HealBlockTag prevents healing in HealPhase. The Regenerator ability is unaffected by Heal Block because it does not use Heal Phase for its healing. but also extends upon the MoveRestrictionBattlerTag to restrict the use of healing moves while the tag is still in effect. The move restriction is based on if the move is a status move and has a TRIAGE_MOVE flag. This avoids adding another redundant flag since attacking moves that heal are not banned with Heal Block. In addition, Heal Block-exempt moves like Pain Split do not have the Triage Flag. There is one exception to this condition however - Pollen Puff.

Pollen Puff has two different effects depending on the move's target. If it targets an opponent, it is a 90 BP attack. If it targets an ally, it is a healing move. Heal Block prevents the affected from using Heal Block towards allies, but allows for it to be used against opponents. This means that it cannot be disabled in Command Phase, like most moves, and must be disabled in Select Target Phase. MoveRestrictionBattlerTag was created based on preventing the use of moves in Command Phase, not Select Target Phase, so I added new functions and parameters to check that Pollen Puff had an allowed use case. Pollen Puff has an attribute StatusCategoryOnAllyAttr but because it was never applied in-game until now, it didn't work. This attribute was fixed to check if the user and the target are allies. Arguably, these new additions should help with Taunt's implementation. 

Psychic Noise and Heal Block share similar implementations. However, Psychic Noise is allowed to hit the user again even when Heal Block is active while Heal Block (the move) should fail when the user already has an Heal Block tag. 

### Screenshots/Videos
Pollen Puff

https://github.com/user-attachments/assets/30824664-95f4-498d-933e-ba250bc70dbd

## How to test the changes?
Use heal block on a Pokemon and damage it afterwards to see if it can heal or not.

## Checklist
- [x] Are the changes visual?


DayKev:
Fixed the git history and added the code from the previous PR (also updated the tests to use the latest functions/etc).

DayKev:
Still needs implementation of disabling healing moves.

> During the effect, healing and draining moves are unusable
> ...
> Pain Split and the Regenerator Ability are unaffected.
- https://www.smogon.com/dex/sv/moves/heal-block/

src/data/battler-tags.ts
```diff
@@ -2178,6 +2190,81 @@ export class ExposedTag extends BattlerTag {
...
+  }
+
+  onActivation(pokemon: Pokemon): string {
+    return i18next.t("battle:battlerTagsHealBlock", { pokemonNameWithAffix: getPokemonNameWithAffix(pokemon) });
+  }
+
+  /**
+   * Checks if a move is disabled under Heal Block
+   * @param {Moves} move {@linkcode Moves} the move ID
+   * @returns {boolean} T/F if the move has a TRIAGE_MOVE flag and is a status move
```
DayKev, line 2218:
```suggestion
   * @returns `true` if the move has a TRIAGE_MOVE flag and is a status move
```

src/data/move.ts
```diff
@@ -3969,8 +3969,10 @@ export class StatusCategoryOnAllyAttr extends VariableMoveCategoryAttr {
    */
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
     const category = (args[0] as Utils.IntegerHolder);
+    // Need to check for strict equality or something better than this.
+    const isAlly = (user.isPlayer() && target.isPlayer()) || (!user.isPlayer() && !target.isPlayer());
```
ben-lear, line 3973: Can’t you check field index? If one is index PLAYER_1 and other is PLAYER_2, or ENEMY_1 and ENEMY_2. You can probably also do a battle.isDouble check to short circuit.

src/data/battler-tags.ts
```diff
@@ -2178,6 +2190,81 @@ export class ExposedTag extends BattlerTag {
...
+    super(BattlerTagType.HEAL_BLOCK, [ BattlerTagLapseType.PRE_MOVE, BattlerTagLapseType.TURN_END ], turnCount, sourceMove);
+  }
+
+  /**
+   * Uses the default onAdd method
+   */
+  override onAdd(pokemon: Pokemon): void {
+    super.onAdd(pokemon);
+  }
+
```
DayKev, line 2210:
```suggestion
```
You don't need to include a method that you're not changing.

src/data/battler-tags.ts
```diff
@@ -2178,6 +2190,81 @@ export class ExposedTag extends BattlerTag {
...
+    return false;
+  }
+
+  /**
+   * Checks if a move is disabled under Heal Block because of its choice of target
+   * Implemented b/c of Pollen Puff
+   * @param {Moves} move {@linkcode Moves} the move ID
+   * @param {Pokemon} user {@linkcode Pokemon} the move user
+   * @param {Pokemon} target {@linkcode Pokemon} the target of the move
+   * @returns {boolean} the move cannot be used b/c the target is an ally
```
DayKev, line 2233:
```suggestion
   * @returns `true` if the move cannot be used because the target is an ally
```

src/data/move.ts
innerthunder, line 3973:
```suggestion
    const isAlly = (user.isPlayer() === target.isPlayer());
```

src/phases/select-target-phase.ts
```diff
@@ -17,6 +19,14 @@ export class SelectTargetPhase extends PokemonPhase {
     const move = turnCommand?.move?.move;
     this.scene.ui.setMode(Mode.TARGET_SELECT, this.fieldIndex, move, (targets: BattlerIndex[]) => {
       this.scene.ui.setMode(Mode.MESSAGE);
+      const fieldSide = this.scene.getField();
+      const user = fieldSide[this.fieldIndex];
+      const moveObject = allMoves[move!];
+      if (moveObject && user.isMoveTargetRestricted(moveObject.id, user, fieldSide[targets[0]])) {
+        const errorMessage = user.getRestrictingTag(move!, user, fieldSide[targets[0]])!.selectionDeniedText(user, moveObject.id);
+        user.scene.queueMessage(i18next.t(errorMessage, { moveName: moveObject.name }), 0, true);
+        targets.length = 0;
```
innerthunder, line 28: Not sure why you're overriding `targets.length` here. Seems like it would have some weird side effects.

innerthunder (dismissed):
You might want to look into modifying `getMoveTargets` in `data/move.ts` to add the Pollen Puff interaction. Enforcing it at the UI level (or within the select target phase in general) still allows enemies to target their allies under Heal Block.

src/data/move.ts
frutescens: Figured something else out.

src/phases/select-target-phase.ts
frutescens: It's to move the user back to Command Phase.

frutescens:
@innerthunder I'm confused about some of your comments because I think it's not accurate to how the move actually works. I attached a video of Pollen Puff now, so hopefully it'll answer some of your comments.

src/data/move.ts
frutescens: This isn't necessary anymore.

src/phases/select-target-phase.ts
innerthunder: I think it'd be better to set `targets` to `[]`, then.

innerthunder:
> @innerthunder I'm confused about some of your comments because I think it's not accurate to how the move actually works. I attached a video of Pollen Puff now, so hopefully it'll answer some of your comments.

Doesn't showdown just prevent you from selecting the ally when they're Heal Blocked? I haven't looked into how it works in mainline (like when the ally target is hit with Psychic Noise).

In any case, the way it works right now probably requires some changes to the AI to have the enemy properly respond to Heal Block, but you don't have to worry about it too much.

frutescens:
Showdown's implementation is technically correct. However, in PokeRogue, doing so means placing TargetSelectPhase into a while-loop or something similar until the user chooses a valid target or exits out (targets = []), which I feel will lead to more unintended effects than booting the user back into command phase.

innerthunder (approved):

DayKev (approved):

SangaraSorama (approved):

PR #4126, opened 2024-09-09
[Hotfix] fix UI transition causing unusual freezes after egg summary shown

### LINKED ISSUE #4111: [Bug] Game crash/softlock after the first egg summary in a new run
### Describe the bug

When starting a run from "New Game", the game will crash during the reward phase after seeing the egg summary screen.
This does not happen if less than 5 eggs hatched since there is no egg summary screen
This does not happen if the game is loaded from an existing save, so exiting the run or reloading the page at any point will result in the bug not happening

From my observation there are two cases:
- If the wave has a shop/reward phase after it, the game crashes and there is an error in the console (see screenshots)
- If not, the game will be softlocked. Pressing buttons result in both the "error" and "action" sound effect to be played, but nothing happening, and no error in the console

I'm not 100% sure this is the same bug, but I'd say the probably is pretty high

### Session export file

_No response_

### User data export file

_No response_

### Expected behavior

The game should behave the same when starting a new run and loading a run (and not crash :p)

### Screenshots / Videos

Wave with shop - crash
Screenshot 2024-09-08 at 18 03 51

Wave without shop - softlock, but no error
Screenshot 2024-09-08 at 17 52 46

Mac/OSX

### If other please specify

_No response_

### Which browser do you use?

Firefox

### If other please specify

_No response_

### Additional context

This was tested on current main branch, but I also experienced this issue on beta before the merge but I wasn't able to reproduce it back then

j-diefenbach:
Fixes #4111 

Hopefully no more crashes :)

removed redundant / buggy clears and UI transitions from the egg summary UI handler (these were not quite the right use case for them and interfered with other UI logic - breaking the next UI handler thus causing either the shop freeze or a no-error freeze).
Instead egg summary handler performs the only transition back to message mode, then calling phase.end() after completion.

### Screenshots/Videos

Hard to exactly demonstrate freeze but console logs show ordering of events
Freeze (bad, can't continue):
image
With fix (good, game continues to play):
image
Freeze from wave 10 (no visible errors here but still freezes)
image

NOTE: will get better videos demonstrating but for now just test for urself

## How to test the changes?

Same as #4111 but checkout this branch
## Checklist
- [X] **I'm using `MAIN` as my base branch**
- [X] tests are still passing

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -2765,8 +2765,8 @@ export default class BattleScene extends SceneBase {
     playerParty.forEach(p => {
       keys.push(p.getSpriteKey(true));
       keys.push(p.getBattleSpriteKey(true, true));
-      keys.push("cry/" + p.species.getCryKey(p.species.formIndex));
-      if (p.fusionSpecies && p.getSpeciesForm() !== p.getFusionSpeciesForm()) {
+      keys.push("cry/" + p.species.getCryKey(p.formIndex));
+      if (p.fusionSpecies) {
         keys.push("cry/"+p.fusionSpecies.getCryKey(p.fusionFormIndex));
       }
     });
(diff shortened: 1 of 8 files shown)


MokaStitcher:
tested and the crashes seem to be fixed!
Only got a missing cry issue for a specific pokemon, but causing no crash and not directly related to this fix, I forwarded it to frutescens

MokaStitcher (approved):
can we remove all the console logging stuff that was added?

MokaStitcher (changes requested):
meant to request changes sorry '-'
can we remove all the console logging stuff that was added?

MokaStitcher (approved):

DayKev (approved):
Started a new run, had 10 eggs hatch after beating wave 1 and went through the egg summary screen, then beat wave 2. No crashes/freezes. I assume that means the fix is working?

Issue #4107, opened 2024-09-08
[Bug] Crash on using TM


ByThePowerOfScience:
### Describe the bug

I used the TM for Payback on my Chespin in slot 2 (1-indexed), and the game went to a black screen.  Checking the console revealed the following stacktrace:
```
TypeError: t.usedTMs is undefined
    start https://pokerogue.net/assets/index-xHZ0xkVd.js:15689
    onActionInput https://pokerogue.net/assets/index-xHZ0xkVd.js:15406
    processInput https://pokerogue.net/assets/index-xHZ0xkVd.js:15412
    processInput https://pokerogue.net/assets/index-xHZ0xkVd.js:15722
    buttonAb https://pokerogue.net/assets/index-xHZ0xkVd.js:16250
    5 https://pokerogue.net/assets/index-xHZ0xkVd.js:16250
    listenInputs https://pokerogue.net/assets/index-xHZ0xkVd.js:16250
    emit https://pokerogue.net/assets/index-xHZ0xkVd.js:1
    gamepadButtonDown https://pokerogue.net/assets/index-xHZ0xkVd.js:16250
    emit https://pokerogue.net/assets/index-xHZ0xkVd.js:1
    update https://pokerogue.net/assets/index-xHZ0xkVd.js:3239
    update https://pokerogue.net/assets/index-xHZ0xkVd.js:3243
    refreshPads https://pokerogue.net/assets/index-xHZ0xkVd.js:3247
    update https://pokerogue.net/assets/index-xHZ0xkVd.js:3247
    emit https://pokerogue.net/assets/index-xHZ0xkVd.js:1
    updatePoll https://pokerogue.net/assets/index-xHZ0xkVd.js:3031
    preRender https://pokerogue.net/assets/index-xHZ0xkVd.js:3027
    emit https://pokerogue.net/assets/index-xHZ0xkVd.js:1
    step https://pokerogue.net/assets/index-xHZ0xkVd.js:466
    step https://pokerogue.net/assets/index-xHZ0xkVd.js:470
    d https://pokerogue.net/assets/index-xHZ0xkVd.js:1036
[index-xHZ0xkVd.js:16376:31258](https://pokerogue.net/assets/index-xHZ0xkVd.js)
    onerror https://pokerogue.net/assets/index-xHZ0xkVd.js:16376
```

### Session export file

[...]

PR #4132, opened 2024-09-09
[Hotfix] Fix crash when teaching TM to starter in new run

Fontbane:

Fixes issue #4107 

- For some reason usedTMs is not initialized by starter select
- There was a stray usedTms in PlayerPokemon structure
- Checks if usedTMs is a thing before pushing to it, if not it initializes it

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -3524,7 +3524,6 @@ export default interface Pokemon {
 
 export class PlayerPokemon extends Pokemon {
   public compatibleTms: Moves[];
-  public usedTms: Moves[];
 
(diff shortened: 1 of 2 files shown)


DayKev:
`usedTMs` should probably be removed from `Pokemon` rather than `PlayerPokemon` since `EnemyPokemon` don't interact with TMs.

Fontbane:
> `usedTMs` should probably be removed from `Pokemon` rather than `PlayerPokemon` since `EnemyPokemon` don't interact with TMs.

The reason it was in PlayerPokemon was because I tried that initially when implementing it, and there were complications with it.

DayKev (approved):

Issue #4100, opened 2024-09-08
[Bug] Lures Are Infinite Duration


OreoCupcakes:
### Describe the bug

Lures refresh back to their initial starting value (10, 15, 30) upon a reload of the game.

### Session export file

[sessionData1.zip](https://github.com/user-attachments/files/16920981/sessionData1.zip)

### Expected behavior

Sessions should save the lure value and keep track of it during a reload.

### What OS did you observe the bug on?

PC/Windows

### Which browser do you use?

Firefox

Snailman11:
Link to Discord's Bug Report (Lures refresh on reload)
https://discord.com/channels/1125469663833370665/1282186962869944391

PR #4114, opened 2024-09-08
[Hotfix] Fix reloads causing lures to refresh, and fix game version

PigeonBar:
Note: This is copied from #4109 in order to re-point it to the main branch cleanly.

## What are the changes the user will see?
Fixes #4100, so reloading the game will no longer cause lures to refresh their durations.

May also introduce other fixes and/or side effects that I am not aware of.

## Why am I making these changes?

A fix for #4100.

## What are the changes from a developer perspective?

The version in `package.json` and `package-lock.json` is now updated to the game's current version of v1.0.5. This means that all game data saved during v1.0.5 will be correctly labelled as v1.0.5, so `version-converter.ts` will no longer treat them as v1.0.4 data and will no longer interfere with them.

`version-converter.ts` is also revised to automatically use the latest version in `package.json`.

### Before the changes

https://github.com/user-attachments/assets/d758cecf-c883-436c-a9b3-cdecc208f06b

After the reload, the lure's duration gets reset to 10.

### After the changes

https://github.com/user-attachments/assets/864ab28f-05f7-4682-96a2-edf8c59e7f0a

After the reload, the lure's duration stays at 7.

## How to test the changes?

1. While playing a run, obtain a lure.
2. Advance to the next wave to count down the lure's duration.
3. Reload the session.
4. The lure's duration should not refresh.

## Checklist
- ~~[ ] **I'm using `beta` as my base branch**~~
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [ ] Have I considered writing automated tests for the issue?
- ~~[ ] If I have text, did I make it translatable and add a key in the English locale file(s)?~~
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #4146, opened 2024-09-10
[Hotfix] Fix Lures Refreshing on Reload

xsn34kzx:

Lures will no longer "refresh" when reloading save data or refreshing the page after data is saved after a wave finishes.

This was an oversight in the save migration algorithm; it was never intended to run more than once, which is essentially forced by the fact that the version number cannot be incremented past `1.0.4` at the moment.

The reason why lures were being set back to their `maxBattles` on reload is because the initial conversion from old lures to new lures takes the first argument of the `ModifierData` `args` array and setting it to the `battleCount`. This effectively takes the old `battleCount` to the new `battleCount` for accurate conversion. However, a side effect of the version number never incrementing is that the conversion would run again on reload/data parsing. As such, the `maxBattles` field would be put into `battleCount` on subsequent reloads because the conversion still thinks the game data is transitioning from `1.0.4` to a new version, which is why simply incrementing the version to `1.0.5` is also a valid fix for this scenario. To get around that though, the length of the `args` array can be checked since it was initially length 1, becoming a length 2 array on conversion. This prevents the conversion from re-running more than once after conversion is done without *needing* the version increment. In addition to this, I also made sure to convert any lingering `White Herb`s on enemies purely for conversion accuracy.

Worth noting that we still need to increment version at some point because this can definitely cause issues in the future with conversion patches (as it did here).

### Screenshots/Videos

N/A

## How to test the changes?

Grab a lure in a run, save the game manually or let the game get to the next wave, and reload. Loading up the save after the fact should retain the `battleCount` it left off with instead of going back to its fresh maximum.

## Checklist
  - https://github.com/pagefaultgames/pokerogue/pull/4114

### DIFF
--- a/src/system/version-converter.ts
+++ b/src/system/version-converter.ts
@@ -31,7 +31,7 @@ export function applySessionDataPatches(data: SessionSaveData) {
 
           // From [ stat, battlesLeft ] to [ stat, maxBattles, battleCount ]
           m.args = [ newStat, 5, m.args[1] ];
-        } else if (m.className === "DoubleBattleChanceBoosterModifier") {
+        } else if (m.className === "DoubleBattleChanceBoosterModifier" && m.args.length === 1) {
           let maxBattles: number;
           switch (m.typeId) {
           case "MAX_LURE":
@@ -53,6 +53,8 @@ export function applySessionDataPatches(data: SessionSaveData) {
       data.enemyModifiers.forEach((m) => {
         if (m.className === "PokemonBaseStatModifier") {
           m.className = "BaseStatModifier";
+        } else if (m.className === "PokemonResetNegativeStatStageModifier") {
+          m.className = "ResetNegativeStatStageModifier";
         }
       });
     }


src/system/version-converter.ts
```diff
@@ -53,6 +53,8 @@ export function applySessionDataPatches(data: SessionSaveData) {
       data.enemyModifiers.forEach((m) => {
         if (m.className === "PokemonBaseStatModifier") {
           m.className = "BaseStatModifier";
+        } else if (m.className === "PokemonResetNegativeStatStageModifier") {
```
ben-lear, line 56: can we not do `instanceof` checks?

ben-lear (approved):

ben-lear (approved):

DayKev (approved):

Snailman11:
Fixes #4100 
(Lures are infinite Duration)

torranx (approved):

DayKev (approved):

PR #4149, opened 2024-09-10
[Bug] Run History displays Pokemon that have their natures changed during the run

frutescens:
Users will see accurate nature information. 
Button will not work when not visible.

image
image

New ternary. 
New conditional check for held items input

### Screenshots/Videos
Will test soon. 

## How to test the changes?
Change Pokemon nature with mints.
Lose run.
Pokemon's nature should be the changed one.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/ui/run-info-ui-handler.ts
+++ b/src/ui/run-info-ui-handler.ts
@@ -526,7 +526,9 @@ export default class RunInfoUiHandler extends UiHandler {
       // Contains Name, Level + Nature, Ability, Passive
       const pokeInfoTextContainer = this.scene.add.container(-85, 3.5);
       const textContainerFontSize = "34px";
-      const pNature = getNatureName(pokemon.nature);
+      // This checks if the Pokemon's nature has been overwritten during the run and displays the change accurately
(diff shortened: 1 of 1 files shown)


DayKev:
In this case it actually makes sense to have the actual nature separate from the override nature as future-proofing in case we implement stuff like the berries that check base nature instead of mint nature for additional effects. The one time the 🍝 isn't actually crazy (well... the modifiers system is still an insane amount of 🍝 but that's another story lol).

DayKev (dismissed):

DayKev:
Btw I wonder if making the nature appear in a different color if you used a mint on the Pokémon would be worthwhile. 🤔

MokaStitcher (commented):
tested, seems to work fine. Just a small suggestion, not vital

If anyone wants a test file: metapod's nature was changed from brave to lonely here: 
[updated_nature_session.txt](https://github.com/user-attachments/files/16947977/updated_nature_session.txt)

MokaStitcher:
> Btw I wonder if making the nature appear in a different color if you used a mint on the Pokémon would be worthwhile. 🤔

I don't even understand why the old nature is remembered in the first place, is it needed anywhere or was that a "just in case" kind of thing?

Issue #4134, opened 2024-09-09
[Bug][UI] Egg summary not showing HA icon for some Pokemon


MokaStitcher:
### Describe the bug

When hatching eggs the hidden ability icon does not show at all for Pokemon with a single ability + HA

Ferroseed is also one of these Pokemon but it's a special case since its HA is the same as its normal ability. And in its case the HA icon is always shown even if the HA wasn't the one unlocked with the thing

### User data export file

Bunch of eggs for testing:  [data_ha_singleability_eggs.txt](https://github.com/user-attachments/files/16931667/_ha_singleability_eggs.txt)
- Bulbasaur (has single ability + HA already unlocked) - HA icon does not work
- Charmanders (has single ability + HA not unlocked) - HA icon does not work
- Machop (has 2 abilities + HA not unlocked) - HA icon works
- Bellsprout (has 2 abilities + HA already unlocked) - HA icon works

Ferroseed eggs: [data_ha_ferroseed_eggs.txt](https://github.com/user-attachments/files/16931804/_ha_ferroseed_eggs.txt)

### Expected behavior

The HA icon should show up for all hatched Pokemon who have their HA
The HA icon should not show up for hatched Ferroseeds if they don't have their HA

### Screenshots / Videos

Bulbasaur and Charmander don't show HA icon:

<img width="420" alt="_egg_ha_bulbasaur" src="https://github.com/user-attachments/assets/8c490538-5a2b-480d-93a8-d60912defd69">
<img width="275" alt="egg_ha_charmander" src="https://github.com/user-attachments/assets/836b611e-6172-4eb5-9f53-76803fb8dde4">

Ferroseed special case: only 2 of these have the HA, but all show the icon

<img width="821" alt="_egg_ferroseed" src="https://github.com/user-attachments/assets/66c3ccef-2847-4a70-a560-e01035d357a0">

### What OS did you observe the bug on?

Mac/OSX

### Which browser do you use?

Firefox

### Additional context

Maybe this comes from the changes here that were not taken into account properly? https://github.com/pagefaultgames/pokerogue/pull/3138
In any case the conditions to display the icon or not should probably just be the same as the one that changes the ability text color, which seems to be working fine

PR #4150, opened 2024-09-10
[Bug][UI] Fix HA icon not always showing in egg summary screen

MokaStitcher:
In egg summary, HA icon is now shown for all Pokemon who hatched with their HA, while before in some cases it wasn't
The icons for unlocks now show on top of the Pokemon sprite as well, while before it could get hidden by the Pokemon sprite or cursor

Fix issue #4134 
TDLR:
* problem 1: for some Pokemon species the HA icon is not showing even if the Pokemon has its HA
* problem 2: for Ferroseed the HA icon is always showing, even if it doesn't have its HA

- Check for the presence of the HA by using the `Pokemon.abilityIndex` instead of calling `Pokemon.hasAbility` which would always return true in the case of Ferroseed since its HA is the same as its normal ability
- (note: this also fixes issue 1, which was caused by the fact that by default `Pokemon.hasAbility` checks that the ability applies to the current battle conditions, so for some HAs like weather ones it would return false even if the Pokemon has the HA. If it wasn't for Ferroseed's special case doing `displayPokemon.hasAbility(displayPokemon.species.abilityHidden, false, true)` instead of `displayPokemon.hasAbility(displayPokemon.species.abilityHidden)` would have been just fine to fix that)
- removed ternary in the HA check for `src/ui/pokemon-info-container.ts` since the HA is always at index 2 since PR 3138
- renamed icon variables for better code readability
- display the icons on top of the Pokemon and cursor and move them around a little to better match the starter UI

### Screenshots/Videos
Before: HA icon doesn't show for Bulbasaur & Charmander
_egg_ha_bulbasaur
After: 
_after_bulb

Before: HA icon always shows for Ferossed
_egg_ferroseed
After:
_after_fer

Before / After of icon placement:
1_current _after_icons

## How to test the changes?

Use override:
`EGG_IMMEDIATE_HATCH_OVERRIDE: true`

Load one of these, clear a wave and check egg summary:
- [Save with Bulbasaur & Charmander eggs](https://github.com/user-attachments/files/16944432/_ha_missing_icons.txt)
- [Save with Ferosseed eggs](https://github.com/user-attachments/files/16944433/_ha_ferroseed_eggs.txt)
- [Save with eggs for Pokemon with varying sprite sizes](https://github.com/user-attachments/files/16944467/_ha_various.txt)

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/ui/egg-summary-ui-handler.ts
+++ b/src/ui/egg-summary-ui-handler.ts
@@ -29,8 +29,10 @@ export default class EggSummaryUiHandler extends MessageUiHandler {
   private summaryContainer: Phaser.GameObjects.Container;
   /** container for the mini pokemon sprites */
   private pokemonIconSpritesContainer: Phaser.GameObjects.Container;
-  /** container for the icons displayed alongside the mini icons (e.g. shiny, HA capsule) */
+  /** container for the icons displayed on top of the mini pokemon sprites (e.g. shiny, HA capsule) */
   private pokemonIconsContainer: Phaser.GameObjects.Container;
+  /** container for the elements displayed behind the mini pokemon sprites (e.g. egg rarity bg) */
+  private pokemonBackgroundContainer: Phaser.GameObjects.Container;
   /** hatch info container that displays the current pokemon / hatch (main element on left hand side) */
   private infoContainer: PokemonHatchInfoContainer;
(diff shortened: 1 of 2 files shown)


KimJeongSun (approved):
LGTM!

PR #4159, opened 2024-09-10
[Hotfix] Fix System Data Conversion Failure 2

xsn34kzx:

For any users who could not load into the game due to an error (reading `caughtAttr` on `undefined`) within `applySystemDataPatches`, they should be able to now.

Users were essentially blocked from being able to enter the game due to an null access, forcing them into the "Failed to load system data" screen.

Added more `undefined`/`null` guards to prevent trying to access fields in objects that weren't initialized yet during the save migration process.

### Screenshots/Videos

N/A

## How to test the changes?

Similar to #4123, the best way would be to use the latest system data provided in [this bug report](https://discord.com/channels/1125469663833370665/1282387337162264626), place it into your own local storage `user_Guest` data, and make sure the data is decrypted during parsing. The game should load as normal without prompting any console errors.

Extraneous errors in the process that are entirely unrelated may include JSON parsing errors (not decrypting the incoming data), malformed UTF-8 (decrypting the data and saving that decrypted data to local storage but forgetting to remove the decryption portion), etc.

## Checklist

### DIFF
--- a/src/system/version-converter.ts
+++ b/src/system/version-converter.ts
@@ -76,7 +76,7 @@ export function applySystemDataPatches(data: SystemSaveData) {
       if (data.starterData) {
         // Migrate ability starter data if empty for caught species
         Object.keys(data.starterData).forEach(sd => {
-          if (data.dexData[sd].caughtAttr && !data.starterData[sd].abilityAttr) {
+          if (data.dexData[sd]?.caughtAttr && (data.starterData[sd] && !data.starterData[sd].abilityAttr)) {
             data.starterData[sd].abilityAttr = 1;
           }
         });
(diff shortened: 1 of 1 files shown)


src/system/version-converter.ts
```diff
@@ -104,9 +104,11 @@ export function applySystemDataPatches(data: SystemSaveData) {
       // --- PATCHES ---
 
       // Fix Starter Data
-      if (data.starterData) {
-        for (const starterId of defaultStarterSpecies) {
+      for (const starterId of defaultStarterSpecies) {
+        if (data.starterData[starterId]?.abilityAttr) {
```
flx-sta, line 108: this will include a false on `0` or `""`. 
Is that intended?

src/system/version-converter.ts
```diff
@@ -104,9 +104,11 @@ export function applySystemDataPatches(data: SystemSaveData) {
...
           data.starterData[starterId].abilityAttr |= AbilityAttr.ABILITY_1;
+        }
+        if (data.dexData[starterId]?.caughtAttr) {
```
flx-sta, line 111: this will include a false on `0` or `""`. 
Is that intended?

src/system/version-converter.ts
flx-sta: >  These are ported patches from before save migration (just littered around the initSystem function)
> \- SN34KZ, Discord

Legacy as is!

src/system/version-converter.ts
flx-sta: >  These are ported patches from before save migration (just littered around the initSystem function)
> \- SN34KZ, Discord

legacy as-is

flx-sta (approved):

ben-lear (approved):

CodeTappert (approved):

PR #4160, opened 2024-09-10
[Misc] eslint rule to prevent `.js` imports 

flx-sta:

To make checks fail when an `.js` import *snuck* into the PR 

- `eslint-import-x` module added (the original one (non-x) is still struggling to adopt es-lint 9 <sub>ridicules</sub>).
- add eslint rule to throw error on file-extensions in imports, except for `json`.
- rename ~~`account.spec.ts`~~ to `account.test.ts` to match common pattern
- removed any leftover `.js` imports that have sneaked in again

## Checklist


ben-lear (approved):

innerthunder (approved):
Seems good. No ".js" imports to be found after the rule was applied

CodeTappert (approved):

PR #4168, opened 2024-09-11
[Misc][Move] Add `waveTurnCount` to `PokemonBattleData`

DayKev:
Fake Out and First Impression (and Mat Block) can be used at the start of every wave again.

Slight buff for a couple of moves (requested by @damocleas), and new field in `PokemonBattleData` that can be used for other things later.

`waveTurnCount: number = 1` added as a field to `PokemonBattleData`.
`FirstMoveCondition` updated to use the new field, allowing moves like Fake Out to be used at the start of every wave even if the Pokémon has been out for multiple turns.

## How to test the changes?
`npm run test fake_out`

## Checklist
  - Modified version of #4105 


damocleas (approved):

flx-sta (approved):

Tempo-anon:
To be honest, not a huge fan with a specific hardcoded case just for certain moves and not others. Things should be consistent and not just have kneejerk changes based on vibes, and it seems a lot of complaining about fake out has died off now too.

damocleas:
> To be honest, not a huge fan with a specific hardcoded case just for certain moves and not others. Things should be consistent and not just have kneejerk changes based on vibes, and it seems a lot of complaining about fake out has died off now too.

you do not see all the feedback I do, there are still consistent complaints and it will feel even worse when we have a real doubles mode

innerthunder (approved):

PR #2732, opened 2024-06-30
[Balance] Rework some evolutions

Fontbane:
## What are the changes?

- **Evolution changes listed on [this spreadsheet](https://docs.google.com/spreadsheets/d/1A_inHbG9e6TffeMcFDTXa7QAtAMuzsdczQJyAJg-0Ok/edit?usp=sharing)**
- Many item+trade evos use their appropriate item instead of Linking Cord
- Bisharp requires Leader's Crest to evolve (**sprite requires atlas to be repacked!**)
- Cosmog evolves into Cosmoem at level 23, which evolves into Solgaleo by using the Sun Flute, or Lunala by using the Moon Flute (**sprites requires atlas to be repacked!**)
- Gligar, Sneasel require Razor Fang, Razor Claw
- Gimmighoul evolves when you collect 10 Nuggets, Big Nuggets, or Relic Gold
- Biome specific evolutions changed to time-based or use an item
- Alolan Raichu and Hisuian Lilligant are evolved by using the Shiny Stone. Alolan Exeggutor is evolved using the Sun Stone.
- Friendship evolution adjustments
- You can pause evolutions for any mon at any time. Paused evolutions won't put their evo item in reward pool except Slowpoke who is too slow to turn off his Linking Cord magnet, and Eevee who likes rocks.
- Moved Syrupy Apple, Masterpiece Teacup, Chipped Pot, Galarica Cuff, Galarica Wreath, Auspicious Armor, Malicious Armor from rare evolution item pool to regular evo item pool

## Why am I doing these changes?

- For balance
- Biome evos are restrictive, inconvenient, unintuitive
- Gimmighoul is too easy to evolve
- Bisharp evolves too quickly
- Friendship evolutions are kind of all over the place
- Getting evolution items when you plan not to evolve something is annoying (cough Duraludon)

## What did change?

- Many evolution methods
- Added localization entries for Pausing evolutions as well as all the new items
- Items added (most already had sprites but were unused):
  - Razor Fang
  - Razor Claw
  - Electirizer
  - Magmarizer
  - Protector
  - Dragon Scale
  - Reaper Cloth
  - Prism Scale
  - Sachet
  - Whipped Dream
  - Leader's Crest (NEW! Requires item atlas to be recompiled!)

## Checklist
- [X] There is no overlap with another PR?
- [X] The PR is self-contained and cannot be split into smaller PRs?
- [X] Have I provided a clear explanation of the changes?
- [ ] Have I tested the changes (manually)?
[...]

PR #4190, opened 2024-09-12
[Balance] Tweak trainer evolution delay

Fontbane:

- Adjusted minimum evolution levels so trainers are more likely to use evolved mons at appropriate waves.
- Remove temporary hardcoded 100% chance for trainer mons to be evolved at floor 80

- The admins at 66 were still using unevolved mons occasionally
- Trainer mons not set as "weaker" are pretty much hardcoded to evolve at/after wave 80. To my understanding this is more of a temporary measure until we figure out how to actually fix evolution delays. I removed it so this can be tested as an alternative.

- In the function getStrengthLevelDiff, the values for average and strong mons were lowered from 10 and 5 to 8 and 4, respectively. These are nice, neat powers of 2. This shrinks the margins in which a mon can remain unevolved on an admin or gym leader team.

## How to test the changes?

Face trainers particularly in the range between the 2 evil team admin fights and see if there's any unevolved mons who should be evolved. Mons with Very Long evolution delays might take longer to evolve (obviously), though a lot of those will be changed with #2732 

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/pokemon-species.ts
+++ b/src/data/pokemon-species.ts
@@ -666,9 +666,9 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
     case PartyMemberStrength.WEAK:
       return 20;
     case PartyMemberStrength.AVERAGE:
-      return 10;
+      return 8;
     case PartyMemberStrength.STRONG:
-      return 5;
+      return 4;
     default:
       return 0;
     }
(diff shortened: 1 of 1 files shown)


src/data/pokemon-species.ts
```diff
@@ -666,9 +666,9 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
     case PartyMemberStrength.WEAK:
       return 20;
     case PartyMemberStrength.AVERAGE:
-      return 10;
+      return 8;
```
Tempo-anon, line 669: can you document `getStrengthLevelDiff` and what the number means

Tempo-anon (changes requested):
Please add documentation and explanation, as well as before/after's

Tempo-anon (approved):

PR #4213, opened 2024-09-13
[Bug] Fix `SwapStatStagesAttr` Oversight

xsn34kzx:

`Guard Swap` and `Power Swap` will now swap their intended stat stages instead of swapping all stat stages like `Heart Swap`.

`Guard Swap` and `Power Swap` were not behaving as intended.

There was a mistake in the loop that swaps each stat stage: instead of looping over the given list of stats in the `SwapStatStagesAttr` constructor, it would always loop over `BATTLE_STATS`, effectively swapping all stat stages regardless of what was passed in. As such, the `for` loop was changed to use the class's `stats` field instead of `BATTLE_STATS`.

In addition to the changes, the unit tests for `Guard Swap` and `Power Swap` were made to be more rigorous in that it checks the expected value of all stat stages (since this oversight slipped by because it only checked DEF and SPDEF for both the player and the enemy). Along with that change, a unit test was made for `Heart Swap` as well.

### Screenshots/Videos

Each move was used in the following order to test if the moves worked correctly now:

- Initial `Power Swap` from player (does nothing), `Shell Smash` twice from enemy, `Power Swap` from player

<div align=center>
powerSwapFirst
</div>

- `Shell Smash` from enemy, `Guard Swap` from player
<div align=center>
guardSwapSecond
</div>

- `Shell Smash` from enemy, `Heart Swap` from player
<div align=center>
heartSwapLast
</div>

## How to test the changes?

The following overrides when starting a new Classic run can be used to replicate the move demonstration in the screenshots:

```ts
const overrides = {
  MOVESET_OVERRIDE: [ Moves.GUARD_SWAP, Moves.POWER_SWAP, Moves.HEART_SWAP ],
  OPP_MOVESET_OVERRIDE: Moves.SHELL_SMASH,
```

## Checklist
~- [ ] If I have text, did I make it translatable and add a key in the English locale file(s)?~
- [x] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -2836,7 +2836,7 @@ export class SwapStatStagesAttr extends MoveEffectAttr {
    */
   apply(user: Pokemon, target: Pokemon, move: Move, args: any []): boolean {
     if (super.apply(user, target, move, args)) {
-      for (const s of BATTLE_STATS) {
+      for (const s of this.stats) {
         const temp = user.getStatStage(s);
         user.setStatStage(s, target.getStatStage(s));
         target.setStatStage(s, temp);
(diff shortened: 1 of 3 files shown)


Snailman11:
Link to Discord's Bug Report: (guard swap bugged)
https://discord.com/channels/1125469663833370665/1284001737631600790

Tempo-anon (approved):

PR #3975, opened 2024-09-02
[Enhancement][EnemyAI] Add support for simulated damage calculations and "Search for KO" move filtering

innerthunder:
## What are the changes the user will see?
The AI won't use random status moves instead of going for a KO anymore.

## Why am I making these changes?
In the mainline games and many popular ROM Hacks, the enemy AI uses damage calculations to limit their move selection to moves that can KO a player Pokemon when possible. Until the recently added support for simulated ability calls, it wasn't possible for the enemy AI in PokeRogue to do so without introducing major side effects to the game state. However, with that support now added, it's now possible to decouple damage modifiers from the central `apply` function and silently calculate damage to implement this behavior.

## What are the changes from a developer perspective?
`field/pokemon`:
- All parts of the damage calculation in `Pokemon#apply` (except for critical hit resolution) are now moved to the function `Pokemon#getAttackDamage`, which takes several parameters:
  - `ignoreAbility`: If `true`, ignores the defending Pokemon's abilities during the calculation. Currently used for damage calculations where the player Pokemon's ability is unknown to the enemy.
  - `ignoreSourceAbility`: If `true`, ignores the attacking Pokemon's abilities during the calculation. Currently unused, but can be useful for when the enemy is assessing its defensive matchup against a player Pokemon whose ability is "unknown."
  - `simulated`: If `true`, suppresses all effects that would trigger messages or ability popups, or otherwise change the game state. This is only set to `false` when this function is called in `apply`.
- A few bugs were fixed in the move to `getAttackDamage`, the most notable being that the STAB damage multiplier now correctly stacks the effects of the attacker's type(s) and Tera type. All damage multipliers internal to the calculation are also documented with comments.
- If one or more of the enemy's moves KO at least one player Pokemon, `EnemyPokemon#getNextMove` will only evaluate the moves that can KO for selection.

`test/enemy_command.test`: New probabilistic test for `EnemyPokemon#getNextMove`. Currently only checks that status moves are never used when a KO is possible.

`test/battle/damage_calculation.test`: Added tests for `Pokemon.getAttackDamage`
- Attacks should deal 1 damage at minimum
- Fixed-damage moves aren't affected by damage multipliers (specifically testing against type multipliers and Multiscale)
- One-hit KO moves aren't affected by damage multipliers (same as above)

#### Move Choice Analysis: Before and After "Search for KO"

The test in `enemy_command.test` simulates the following battle scenario:
- The player Pokemon is a Level 1 Magikarp
[...]

PR #4246, opened 2024-09-14
[Bug] Golden punch bug fix

Opaque02:
Sometimes when using the golden punch it picks it up as NaN money to add to your score, which breaks your money and doesn't let you buy stuff. Looks like this was added accidentally in #3975 

In [main](https://github.com/pagefaultgames/pokerogue/blob/main/src/field/pokemon.ts), the damage field here is a number holder, whereas in [beta](https://github.com/pagefaultgames/pokerogue/blob/beta/src/field/pokemon.ts#L2665) it's a number. This means sometimes when the golden punch is being applied, it would use 95 and try convert it to a number holder, which fails and returns NaN during the [apply calculation](https://github.com/pagefaultgames/pokerogue/blob/beta/src/modifier/modifier.ts#L2418) when trying to convert `args[1] as Utils.IntegerHolder`

I changed pokemon.ts for the golden punch modifier to pass through damage as a new util integer holder instead of a number

### Screenshots/Videos
In beta after using a move - groudon has a golden punch equipped:

image

After this patch after using a move:

image

## How to test the changes?
Download this PR and use a file with a golden punch. You can either use overrides or you can use a save file with a golden punch, like [this one](https://discord.com/channels/1125469663833370665/1176874654015684739/1284416753560191026)

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -2662,7 +2662,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
           const attackResult = { move: move.id, result: result as DamageResult, damage: damage, critical: isCritical, sourceId: source.id, sourceBattlerIndex: source.getBattlerIndex() };
           this.turnData.attacksReceived.unshift(attackResult);
           if (source.isPlayer() && !this.isPlayer()) {
-            this.scene.applyModifiers(DamageMoneyRewardModifier, true, source, new Utils.IntegerHolder(damage));
+            this.scene.applyModifiers(DamageMoneyRewardModifier, true, source, new Utils.NumberHolder(damage));
           }
         }
       }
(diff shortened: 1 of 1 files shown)


src/field/pokemon.ts
```diff
@@ -2662,7 +2662,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
           const attackResult = { move: move.id, result: result as DamageResult, damage: damage, critical: isCritical, sourceId: source.id, sourceBattlerIndex: source.getBattlerIndex() };
           this.turnData.attacksReceived.unshift(attackResult);
           if (source.isPlayer() && !this.isPlayer()) {
-            this.scene.applyModifiers(DamageMoneyRewardModifier, true, source, damage);
+            this.scene.applyModifiers(DamageMoneyRewardModifier, true, source, new Utils.IntegerHolder(damage));
```
innerthunder, line 2665: Can you change this and the modifier to use `NumberHolder` instead?

innerthunder (changes requested):

src/field/pokemon.ts
Opaque02: Updated in latest commit

DayKev (approved):

innerthunder (approved):

MokaStitcher (approved):

PR #4248, opened 2024-09-14
[Bug] Run History Not Saving Correctly

frutescens:
Their run history may work again.

Guh.

The method to retrieve the last wave's information was not updated while game-data.ts' getSessionSaveData() was. This would mean that when the parseSessionSaveData hit the changed line, it would read an undefined and explode, meaning that it would fail to read the rest of the data.

## How to test the changes?
Finish a run and check if run history is logged correctly. 

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/phases/game-over-phase.ts
+++ b/src/phases/game-over-phase.ts
@@ -238,7 +238,7 @@ export class GameOverPhase extends BattlePhase {
       gameVersion: this.scene.game.config.gameVersion,
       timestamp: new Date().getTime(),
       challenges: this.scene.gameMode.challenges.map(c => new ChallengeData(c)),
-      mysteryEncounterType: this.scene.currentBattle.mysteryEncounter?.encounterType,
+      mysteryEncounterType: this.scene.currentBattle.mysteryEncounter?.encounterType ?? -1,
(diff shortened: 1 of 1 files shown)


DayKev (approved):

MokaStitcher:
I can't reproduce the issue on beta.
Lost a run to final fight, beat final fight, lost to a wild encounter, to a mystery encounter, and none caused any issue with the run history.
Need more info to know if this actually fixes stuff

src/ui/run-info-ui-handler.ts
```diff
@@ -93,6 +93,7 @@ export default class RunInfoUiHandler extends UiHandler {
     const run = args[0];
     // Assigning information necessary for the UI's creation
     this.runInfo = this.scene.gameData.parseSessionData(JSON.stringify(run.entry));
+    console.log(this.runInfo);
```
ben-lear, line 96: Do we still want this log?

ben-lear (approved):

src/ui/run-info-ui-handler.ts
frutescens: whoops, my bad

PR #4266, opened 2024-09-15
 [Bug] Using default animation for errors that occur.

podarsmarty:
- Fetch errors will no longer result in a crashed game

- These errors are prevalent in offline builds that crash the game. I expect this to be possible on the live version but more rare and transient

- Adds error handling to promise so that default animations do not occur ONLY for bad responses

### Screenshots/Videos
N/A

## How to test the changes?
- Build and package offline with move animation json file for a pokemon being used deleted. Will crash before and work properly after

## Checklist
     I took a look but for tests the entirely of fetch is mocked out. would need to add mock configurations to add unit tests
- [ ] Are the changes visual?


src/data/battle-anims.ts
```diff
@@ -515,13 +514,23 @@ export function initMoveAnim(scene: BattleScene, move: Moves): Promise<void> {
...
+            useDefaultAnim(move, defaultMoveAnim, error);
+            return resolve();
           });
       };
       fetchAnimAndResolve(move);
     }
   });
 }
 
+function useDefaultAnim(move: Moves, defaultMoveAnim: Moves, ...optionalParams: any[]) {
```
flx-sta, line 528: NIT:
missing ts-docs

src/enums/moves.ts
```diff
@@ -1873,3 +1873,9 @@ export enum Moves {
   /**{@link https://bulbapedia.bulbagarden.net/wiki/Malignant_Chain_(move) | Source} */
   MALIGNANT_CHAIN,
 }
+
+export class MovesUtil {
```
flx-sta, line 1877: NIT:
Missing ts-docs

src/enums/moves.ts
```diff
@@ -1873,3 +1873,9 @@ export enum Moves {
...
+  public static filename(move: Moves) {
```
flx-sta, line 1878: NIT:
Missing ts-docs

src/enums/moves.ts
flx-sta: Also I'd prefer to move new classes into their own files.

To keep it simple though you might want to consider simply creating an exported function/method in `utils.ts`

flx-sta (changes requested):

src/data/battle-anims.ts
```diff
@@ -515,13 +514,30 @@ export function initMoveAnim(scene: BattleScene, move: Moves): Promise<void> {
             } else {
               resolve();
             }
+          })
+          .catch(error => {
+            useDefaultAnim(move, defaultMoveAnim, error);
+            return resolve();
```
flx-sta, line 520: We should probably print at least a log, better yet a warning here
```suggestion
          .catch(error => {
            console.warn(`No move animation found for ${Utils.animationFileName(move)}! Falling back to default.`);
            useDefaultAnim(move, defaultMoveAnim, error);
            return resolve();
```

flx-sta (changes requested):
sorry found 1 more.
(make sure to not always force push 👀 )

podarsmarty:
> sorry found 1 more. (make sure to not always force push 👀 )

ahh no worries. will fix. my bad, saw you were merging, figured id rebase on beta and squash my commits so forced

src/data/battle-anims.ts
podarsmarty: @flx-sta hmm the useDefaultAnim function already logs an error to the console

podarsmarty: renamed method to make that behavior more clear

flx-sta:
I find it rather odd that you decided to move the logging into the defaultAnim function (although you did rename it).
But technically the functions (original) context has nothing to with the logging. It would be better to have the logging inside the actual `catch()`.
Also `console.error` seems to elevated as it's not breaking anything. It's just an inconvenience, a warning to say so. Which implies wanting to use `console.warn()`

podarsmarty:
> I find it rather odd that you decided to move the logging into the defaultAnim function (although you did rename it). But technically the functions (original) context has nothing to with the logging. It would be better to have the logging inside the actual `catch()`. Look at it this way: Technically you could be calling `defaultAnim()` on purpose without it being an error/warning, right?
> 
> Also `console.error` seems to elevated as it's not breaking anything. It's just an inconvenience, a warning to say so. Which implies wanting to use `console.warn()`

i had moved it since the function is called in two places that would both warrant a log. figured it be better to share rather than copy pasting it. yea its not breaking, i had left it as an error the way it originally was, i can switch it over to a warn and duplicate the logging though.

however, i'd argue you should only be calling defaultAnim() in an error state since everything should have an animation defined the way it is setup

flx-sta (dismissed):
Thanks for elaborating your approach.

src/data/battle-anims.ts
```diff
@@ -515,13 +515,41 @@ export function initMoveAnim(scene: BattleScene, move: Moves): Promise<void> {
...
       fetchAnimAndResolve(move);
     }
   });
 }
 
+/**
+ * Populates the default animation for the given move.
+ *
+ * @param move the move to populate an animation for
+ * @param defaultMoveAnim the default move to use as the default animation
```
DayKev, line 534:
```suggestion
 * @param defaultMoveAnim the move to use as the default animation
```

src/data/battle-anims.ts
```diff
@@ -515,13 +515,41 @@ export function initMoveAnim(scene: BattleScene, move: Moves): Promise<void> {
...
+ */
+function useDefaultAnim(move: Moves, defaultMoveAnim: Moves) {
+  populateMoveAnim(move, moveAnims.get(defaultMoveAnim));
+}
+
+/**
+ * Helper method to call when an move animation is missing.
```
DayKev, line 541:
```suggestion
 * Helper method for printing a warning to the console when a move animation is missing.
```

src/data/battle-anims.ts
```diff
@@ -515,13 +515,41 @@ export function initMoveAnim(scene: BattleScene, move: Moves): Promise<void> {
...
+ *
+ * @param move the move to populate an animation for
+ * @param optionalParams parameters to add to the error logging
+ *
+ * @remarks use {@link useDefaultAnim} to use a default animation
```
DayKev, line 546:
```suggestion
 * @remarks use {@linkcode useDefaultAnim} to use a default animation
```

flx-sta (approved):

DayKev (approved):

PR #4286, opened 2024-09-16
[Localization] [ALL] Localized status condition

CodeTappert:

For now only german users will see a difference. But this offers the option to have localized status icons (see screenshots). Its of course also changed in the party and summary. If the icons are used anywhere else please let me know. My search with the IDE didnt show any more use

Lugiad askes me and i also wanted todo this a long time ago

It behaves the same as the type icons. It loads the image based on the i18n language. Fallback of course as always is english

### Screenshots/Videos

Only german is different for now:
Before:
image

After:
image

## How to test the changes?

PLay in german. Get your or the enemy pokemon to a different status effect.
Or play in any other language and verify that nothing has changed visually

## Checklist
- [ ] Are the changes visual?


flx-sta:
Something that was probably discussed before:

why aren't we using a text-container/-box here?  (obviously adjust the color accordingly with the status).

This approach seems quite inconvenient and static

CodeTappert:
> _Something that was probably discussed before:_
> 
> Why aren't we using a text-container/-box here? (obviously adjust the color accordingly with the status).
> 
> This approach seems quite inconvenient and static

Because there is no font in the world that looks good that small. it needs to be pixelart

flx-sta (approved):
lgtm.
I'm not 100% a fan of putting placeholders for all the languages that are missing it but I'm ultimately fine with it

image

Adri1 (dismissed):

josericardo-fo (approved):

Adri1 (approved):

CuriousRookie:
it seems that `statuses_ca_ES`, `statuses_pt_BR`, `statuses_zh_CN`, `statuses_zh_TW` does not match the value of `i18next.resolvedLanguage`

[https://github.com/pagefaultgames/pokerogue/blob/6626df27ba0b4a64916bb008fabcd640feeae4e3/src/loading-scene.ts#L241-L254](https://github.com/pagefaultgames/pokerogue/blob/6626df27ba0b4a64916bb008fabcd640feeae4e3/src/loading-scene.ts#L241-L254)

PR #4297, opened 2024-09-17
[Move][Mirror] Update HitTagAttr attributes v2

flx-sta:
### Mirror of #3082

> [!WARNING]
> Copied PR template/comment from #3082

---

## What are the changes?
Particular moves had outdated effects, or had effects removed due to not being in the most recent games.
It adjusts Whirlwind affecting Fly, Steamroller affecting Minimize, and Malicious Moonsault affecting Minimize.

These cause the game to behave in a way that those familiar with canonical mechanics might not expect.

Whirlwind was only supposed to be able to hit Pokemon in the semi-invulnerable turn of Fly in Gen II, and so that behaviour is getting removed. In fact, it is currently also capable of hitting Pokemon in the semi-invulnerable turn of Bounce and Sky Drop, since those all share the FLYING BattlerTag.
image
Steamroller and Malicious Moonsault were effective against Minimized Pokemon as of the last time they were canonically implemented, and so that functionality is being restored.
image

## What did change?
Whirlwind lost the attribute allowing it to hit Pokemon in the FLYING turn of the two-turn moves Fly, Bounce, and Sky Drop.
Steamroller and Malicious Moonsault gained the attribute that bypasses accuracy checks on Minmized targets, as well as the attribute that does double damage against them.

## Checklist
- [x] Have I considered writing automated tests 


chaosgrimmon:
Can confirm that flx kept me looped in and we concepted the tests collaboratively.
Full disclosure, I'm pretty sure I didn't actually get to testing the changes for my old draft PR. It _should_ work exactly as well as other moves that have the same properties, but like, it very much was not a polished, usable PR until flx dug it back out of the attic.

src/data/move.ts
```diff
@@ -8259,6 +8258,8 @@ export function initMoves() {
     new AttackMove(Moves.LEAF_TORNADO, Type.GRASS, MoveCategory.SPECIAL, 65, 90, 10, 50, 0, 5)
       .attr(StatStageChangeAttr, [ Stat.ACC ], -1),
     new AttackMove(Moves.STEAMROLLER, Type.BUG, MoveCategory.PHYSICAL, 65, 100, 20, 30, 0, 5)
+      .attr(MinimizeAccuracyAttr)
```
Tempo-anon, line 8261: Does the MinimizeAccuracyAttr not already also basically do the same thing as HitsTagAtr minimized? Looking at the code it seems like the HitsTagAttr only doubles the damage while MinimizeAccuracyAttr means it never misses. lol

Tempo-anon (changes requested):
Can I request a name change for two Attrs? Could also potentially copy some logic to make the damage doubling part of the anti-minimize tag since all moves that always hit minimize targets seem to double damage but I guess for simplicity's sake this can be done first

`HitsTagAttr` -> `DealsDoubleDamageToTagAttr`
`MinimizeAccuracyAttr` -> `AlwaysHitsMinimizeAttr`

src/data/move.ts
flx-sta: I think it's better to keep it separate for potential race-conditions

Tempo-anon (approved):

src/data/move.ts
Tempo-anon: agree

DayKev:
Moves that use it that deal 2x damage:
- Gust
- Stomp
- Body Slam
- Surf
- Earthquake
- Magnitude
- Twister
- Whirlpool
- Dragon Rush
- Heavy Slam
- Heat Crash
- Steamroller
- Flying Press

Moves that use it that don't deal 2x damage:
- Thunder
- Fissure
- Sky Uppercut
- Smack Down
- Hurricane
- Thousand Arrows

PR #4323, opened 2024-09-19
[Localization] French adjustments (+German)

Adri1:
French and Behemoth adjustments in text / French UI

To have French and German adjusted

None

## Checklist


src/locales/de/move.json
```diff
@@ -3121,11 +3121,11 @@
   },
   "behemothBlade": {
     "name": "Gigantenhieb",
-    "effect": "Der Anwender wird zu einem riesigen Schwert und greift das Ziel an. Dynamaximierte Ziele erleiden doppelten Schaden."
+    "effect": "Der Anwender wird zu einem riesigen Schwert und greift das Ziel an."
```
CodeTappert, line 3124: ? Why

Adri1: > ? Why

This effect don't exists in Rogue, but yet French and German are the only languages that have it left in their descriptions

PR #4331, opened 2024-09-19
[QOL] Localizable Trainer Name + Cleffa Species name can be used in all 3 pokemon nicknames

CodeTappert:

Ähm? Localized name i guess and maybe funnier nicknames

Because the first one is a oversight, and the second one a qol

The name of the breeder is localizable and the species name can now be used in all 3 nicknames

### Screenshots/Videos

None since this doesnt impact the game unless somebody actual does localization

## How to test the changes?

Code review

## Checklist
- [ ] Are the changes visual?


CodeTappert:
My local test is just seeing if everything works as before

Adri1 (approved):

src/data/trainer-config.ts
```diff
@@ -911,7 +911,7 @@ export class TrainerConfig {
     if (!getIsInitialized()) {
       initI18n();
     }
-    this.name = i18next.t(`trainerNames:${name.toLowerCase()}`);
+    this.name = i18next.t(`trainerNames:${name.toLowerCase().replace(/\s/g, "_")}`);
```
ben-lear, line 914: Would it be simpler to change this function to just accept the raw i18n key for the trainerName? So instead of passing “Expert Pokemon Breeder”, just pass “expert_pokemon_breeder”

ben-lear: Also wasn’t this function already being used by Bug-Type Superfan or something? Just checking to make sure other cases aren’t broken

CodeTappert: This functions was not used by the breeder. But the winstrates. 

The way it currently is implemented (after this pr) is like it is for all trainers that have a initFor Function

ben-lear (approved):

flx-sta (approved):
lgtm

Issue #3673, opened 2024-08-20
[Bug] Moves That would be nullified by Move/type Absorbing Abilities Can Miss Incorrectly


Snailman11:
### Describe the bug

https://replay.pokemonshowdown.com/gen9customgame-2184170758

-Tested on Showdown

https://discord.com/channels/1125469663833370665/1125894949020381285/1274511272855207977

-Found out about it from Discord

Abilities such as Storm drain, Volt Absorb, and Flash Fire do not absorb moves when the incoming attack misses.

### Expected behavior

Move accuracy should be checked after immunity abilities are considered.

### Screenshots / Videos

https://github.com/user-attachments/assets/b6b74812-6638-4f14-85e5-6ed6e1142117

### What OS did you observe the bug on?

Other

### If other please specify

gameplay

### Which browser do you use?

Other

### If other please specify

gameplay

PR #4337, opened 2024-09-20
[Bug] Flash Fire/etc now trigger even if the attack would miss

PrabbyDD:
Fix for #3673
Should now properly use absorbing abilities even when attack misses under the hood

bugfix

modified several tests (such as flash fire, sap sipper, volt absorb, etc...) to account for when a pokemon misses an attack that should be absorbed

In ability.ts added a getter to get type immunity in TypeImmunityAbAttr
In move-effect-phase.ts added a check during hit checks for if a pokemon is immune to attacking move due to ability 

### Screenshots/Videos

https://github.com/user-attachments/assets/273a14ac-8469-40bd-a7ba-0b0e824b324c

Same settings but changed to water gun as attack to see that volt absorb doesnt get confused for water absorb
https://github.com/user-attachments/assets/bb8714c3-d86c-412c-b3c1-b85627832c14

## How to test the changes?
Use override files and change hitcheck function to return false
Also use the unit tests I modified

## Checklist
- [X ] **I'm using `beta` as my base branch**
- [ X] There is no overlap with another PR?
- [ X] The PR is self-contained and cannot be split into smaller PRs?
- [X ] Have I provided a clear explanation of the changes?
- [ X] Have I considered writing automated tests for the issue?
- [X ] Have I tested the changes (manually)?
    - [X ] Are all unit tests still passing? (`npm run test`)
- [X ] Are the changes visual?
  - [X ] Have I provided screenshots/videos of the changes?


src/test/abilities/flash_fire.test.ts
```diff
@@ -126,4 +126,23 @@ describe("Abilities - Flash Fire", () => {
...
+
+    const enemyPokemon = game.scene.getEnemyPokemon()!;
+
+    game.move.select(Moves.FLAMETHROWER);
+    await game.setTurnOrder([BattlerIndex.ENEMY, BattlerIndex.PLAYER]);
+    await game.phaseInterceptor.to("MoveEffectPhase");
+
+    await game.move.forceMiss();
+    await game.phaseInterceptor.to("BerryPhase", false);
+    expect(enemyPokemon.hp).toBe(enemyPokemon.getMaxHp());
```
DayKev, line 146: This will always be true no matter what because the enemy Pokémon is never damaged.
Same for the other tests.

PrabbyDD: ahh your right, ill have to think of a more creative test

DayKev (approved):

DayKev (approved):

PR #4352, opened 2024-09-20
[Bug] Fix crash when using enemy move override

DayKev:
Potentially some missing animation issues will no longer cause a crash... or it's possible the issue is only caused specifically by the override.

It's really annoying when the game crashes while testing. 

The game will crash less when using the enemy moveset override.

### Screenshots/Videos
Before fix:
image

## How to test the changes?
```ts
const overrides = {
  OPP_MOVESET_OVERRIDE: Moves.SPLASH
```
Start a new Classic run with any random Pokémon, make sure not to faint the enemy on the first turn.

## Checklist


src/data/battle-anims.ts
```diff
@@ -1279,7 +1279,7 @@ export class MoveAnim extends BattleAnim {
   getAnim(): AnimConfig {
     return moveAnims.get(this.move) instanceof AnimConfig
       ? moveAnims.get(this.move) as AnimConfig
-      : moveAnims.get(this.move)![this.user?.isPlayer() ? 0 : 1] as AnimConfig; // TODO: is this bang correct?
+      : moveAnims.get(this.move)?.[this.user?.isPlayer() ? 0 : 1] as AnimConfig;
```
innerthunder, line 1282: Why the `.` here?

ben-lear (approved):

src/data/battle-anims.ts
ben-lear: Optional chaining requires it, even if you’re about to index an areay

DayKev: That's the syntax for using a `?` before `[]`, you have to add the `.` for whatever reason. 🤷
I asked the same question recently lmao.

flx-sta: this is necessary when using the ternary

flx-sta (approved):
it fixes it, but now the enemy pokemon doesn't perform any animations anymore (which is obviously the cause, but why?)

Only when using overrides!

torranx (approved):
lgtm

PR #4368, opened 2024-09-21
[Bug] Preventing Duplicate Species in Trainer Battles

frutescens:
Users will no longer have to fight trainers with repeated Pokemon. The game will try to supply unique party members instead. 

Quid-pro-quo

## Screenshots
Can't prove definitely that no repeats will occur, but ran through Wave 112 multiple times. 
Console Output: 
image
Enemy Parties (Wave 112):
image
image
image
image
image

A new function was created 'checkDuplicateSpecies' that checks if the Pokemon Species passed is already found in the party. If it is, it prompts the genNewPartyMemberSpecies to reroll again. An infinite loop is already prevented in the code as the game allows up to 10 retries before settling with the Pokemon chosen. 

## How to test the changes?
Choose any trainer that lacks fixed party members. Would recommend Wave 112.
Evil team leaders and the rivals are not a good choice b/c the all/majority of their party is set in stone.  

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/field/trainer.ts
+++ b/src/field/trainer.ts
@@ -425,12 +425,30 @@ export default class Trainer extends Phaser.GameObjects.Container {
       }
     }
 
+    // Prompts reroll of party member species if species already present in the enemy party
+    if (this.checkDuplicateSpecies(ret)) {
+      console.log("Duplicate species detected, prompting reroll...");
+      retry = true;
+    }
+
     if (retry && (attempt || 0) < 10) {
       console.log("Rerolling party member...");
       ret = this.genNewPartyMemberSpecies(level, strength, (attempt || 0) + 1);
     }
 
     return ret;
+  }
+
+  /**
+   * Checks if the enemy trainer already has the Pokemon species in their party
+   * @param {PokemonSpecies} species {@linkcode PokemonSpecies}
+   * @returns `true` if the species is already present in the party
+   */
+  checkDuplicateSpecies(species: PokemonSpecies): boolean {
+    const currentPartySpecies = this.scene.getEnemyParty().map(p => {
+      return p.species.speciesId;
+    });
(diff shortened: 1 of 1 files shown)


src/field/trainer.ts
```diff
@@ -425,6 +425,12 @@ export default class Trainer extends Phaser.GameObjects.Container {
...
 
+    // Prompts reroll of party member species if species already present in the enemy party
+    if (this.checkDuplicateSpecies(ret)) {
+      console.log("Duplicate species detected, prompting reroll...");
+      retry = true;
+    }
+
     if (retry && (attempt || 0) < 10) {
       console.log("Rerolling party member...");
       ret = this.genNewPartyMemberSpecies(level, strength, (attempt || 0) + 1);
```
DayKev, line 436:
```suggestion
    if (retry && (attempt ?? 0) < 10) {
      console.log("Rerolling party member...");
      ret = this.genNewPartyMemberSpecies(level, strength, (attempt ?? 0) + 1);
```
Might as well fix nearby.

frutescens:
Seems to work with Wave 114 (Evil Team admins too)
image

DayKev (approved):

innerthunder (approved):
I think I'd rather have the pool filtered to unselected species instead of forcing rerolls in the future, but this seems like an ok temporary fix for now.

xsn34kzx:
+1 to @innerthunder's comment, I feel like the root cause is the pool not "shrinking" as species get picked during team generation (this probably falls under a trainer generation rewrite anyway). In any case, lgtm!

CodeTappert:
image

Seems fine. I will test some more

xsn34kzx (approved):

CodeTappert (approved):

Issue #4300, opened 2024-09-17
 Mystery Encounters beta bugs/breakfix 2


flx-sta:
Please add any bugs, session saves, screenshots, logs, etc. here in comments so they can be tracked under this issue

This is a follow up issue for #4240

VaylenObscuras:
ME with the choice of 3 trainer battles. Went for "mightiest" and was surprised to see that my delphox retained its Quiver Dance boosts. I was under impression that boosts should be reset when an encounter happens.

PigeonBar:
(Originally from https://github.com/pagefaultgames/pokerogue/issues/4240#issuecomment-2354448815)

Sometimes, if you save and quit a session which had encountered ME's, and then start a new run, the game can incorrectly apply and save `mysteryEncounterSaveData.encounteredEvents` from the old session to the new session. This can cause the game to refuse to spawn ME's due to the "do not spawn an ME within 3 waves of the last one" check.

I was able to continue reproducing this bug after applying the changes from https://github.com/pagefaultgames/rogueserver/pull/52. In fact, this bug is probably more common now, and it appears to no longer fix itself after a reload.

![20240917 Encountered Events](https://github.com/user-attachments/assets/b678ebb0-2541-4c19-9440-d699ca992b11)

PigeonBar:
In Department Store Sale, there is a weird edge case that happens when you have a party of Pokemon that cannot learn any TMs, then you choose to go to the TM counter:

https://github.com/user-attachments/assets/edf51bce-6a6d-4068-8dbd-cfce99e7a9bb

PigeonBar:
The game allows you to open the menu while scrolling the ending dialogue text after an ME (this is easier to trigger at lower game speeds), which then allows you to use the egg gacha exploit to save any meta-progression that the ME awards. Here is an example of saving an epic egg from A Trainer's Test:

https://github.com/user-attachments/assets/3022aa9d-44a3-4929-a175-eb6c33579aac

PigeonBar:
If the Mysterious Chest's unlucky option KO's your only healthy Pokemon, and you click "Yes" to retry the battle, it now crashes the game:

https://github.com/user-attachments/assets/8f583055-c329-4089-91a1-d3717e0bc130

Snailman11:
Health Ui Disappears After "Weird Dream" ME

> (beta) the “mysterious women” mystery event removes Pokémon health bars
https://discord.com/channels/1125469663833370665/1285854633247637514

[...]

PR #4369, opened 2024-09-21
[Bugs/Balance] Fix various ME bugs and small balance adjustments

ben-lear:
Tracking Issue: https://github.com/pagefaultgames/pokerogue/issues/4300

## Balance Changes

- Teleporting Hijinks: If the player encounters below wave 50, the wild boss that spawns will *NOT* have Atk/SpAtk boosts and instead only have Defenses/Spd boosted.
- The Expert Pokemon Breeder:
  - Breeder's team evolutions have been bumped from wave 30/45/60 > 45/60/75
  - Breeder's team has no hard-coded items (except Steel Terrastalization on Ace)
  - If the player loses the battle, instead of Game Over, the player's selected pokemon instead loses all friendship, remains KOed, and the player gets no rewards.
- Weird Dream:
  - If player selects the "Leave" option, party levels are decreased by 12.5% now (previously 20%)
  - Newly transformed Pokemon movesets were changed from [3 previous pokemon moves + 1 new pokemon egg move] >[2 previous pokemon moves + 1 new pokemon egg move + 1 new pokemon favored STAB move]

## Bugs Squashed

- Renamed all ME assets to use a standardized naming convention
- Fixed crash that happens when selecting "Yes" for Retry Battle during ME
- Fixed MEs that require a Pokemon's held item to also check that the held item is transferrable
  - This solves issues such as giving Evolution/Form change items in Delibir-dy, etc.
- Fixed the way Shuckle Juice stacks (it still has max 2 stacks) so that positive and negative Juices do not get combined
- Fixed the way ME custom typing/ability overrides interact with fusion Pokemon and battle effects such as Burn Up
- Refactored Pokemon selection filtering for MEs that require the player to select a Pokemon.
  - If a Pokemon is unselectable because it is invalid for a challenge, will display a more appropriate message
- Fixed issues where Save + Quit or beating a run and immediately starting a new run would corrupt the new run session data with previous run's ME session data
- Can no longer *ever* open the menu if the MessageUiHandler is in the middle of animating text on the screen. This applies globally across the game.
- Fixed issues where a Pokemon at exactly half HP would die to Shady Vitamin Dealer (now Pokemon must be at 51% health or higher)
- Fixed Phione and a few other edge case legends/mythicals spawning in Safari Zone and Pokemon Salesman
- Fiery Fallout now properly applies burn where applicable
- Fixed MEs that required certain numbers of legal/unfainted mons to check properly for challenge legality (An Offer You Can't Refuse)
- Fixed Weird Dream not passing pokerus to new transformed Pokemon

## Still Outstanding/Unconfirmed Bugs
See tracking Issue: https://github.com/pagefaultgames/pokerogue/issues/4300

## Videos/Screenshots

New Breeder battle loss logic:

https://github.com/user-attachments/assets/2ca77cf1-fd68-4ded-a751-191431ebb0ca

Menu isn't exploitable anymore (spammed menu button whole recording)

chrome_Z2aF09E9qX


damocleas (approved):

CodeTappert (approved):

src/data/mystery-encounters/utils/encounter-pokemon-utils.ts
```diff
@@ -773,3 +782,23 @@ export async function addPokemonDataToDexAndValidateAchievements(scene: BattleSc
   scene.gameData.updateSpeciesDexIvs(pokemon.species.getRootSpeciesId(true), pokemon.ivs);
   return scene.gameData.setPokemonCaught(pokemon, true, false, false);
 }
+
```
frutescens, line 785: Can't comment directly, but can 'getHighestStatPlayerPokemon()' and similar functions use isAllowedinBattle() instead? Just in case a player has an ineligible Pokemon in their party that has high/low stats/qualities.

frutescens (changes requested):

src/data/mystery-encounters/utils/encounter-pokemon-utils.ts
ben-lear: refactored all those methods to be more consistent now

damocleas (approved):

frutescens (approved):

CodeTappert (approved):

PR #4384, opened 2024-09-23
[Tests] fix shell side arm test flakyness

flx-sta:
 
The `src/test/moves/shell_side_arm.test.ts` test should no be no more flaky

## How to test the changes?

Run the tests as many times as possible an see if they ever fail.

## Checklist


innerthunder (changes requested):
Overlap with #4383.

Also, Shell Side Arm's implementation relies on the accuracy of `getBaseDamage`, so mocking that function isn't the way to go about fixing this imo

flx-sta:
> Overlap with #4383.
> 
> Also, Shell Side Arm's implementation relies on the accuracy of `getBaseDamage`, so mocking that function isn't the way to go about fixing this imo

I neither agree nor disagree. It depends on the scope of your test

PR #4344, opened 2024-09-20
[Feature] Show app version on title screen

fransdhinta:
## What are the changes the user will see?

- Now the title screen will display the app version.

## Why am I making these changes?

- Because users need to know which version is running this game
- As a user, I would be happy to know that the game version is updated, meaning the game is still maintained.
- An issue mentioned this feature https://github.com/pagefaultgames/pokerogue/issues/3909#issue-2495887371
## What are the changes from a developer perspective?

- Show app version below the title

### Screenshots/Videos

before:
before
after:
after

## How to test the changes?

Just load the game, go to title screen

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [ ] Have I considered writing automated tests for the issue?
- [ ] If I have text, did I make it translatable and add a key in the English locale file(s)?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #4388, opened 2024-09-23
[Dev] Bump Game Version, Overhaul Version Migration

xsn34kzx:

Sessions around version `1.0.0` had battle items with higher `battleCounts` than `maxBattles`, causing a minor issue with `battleCount` coloring and having more battles left than intended, which has been in fixed. Besides this, no other changes should be seen besides the version changing from `1.0.4` to `1.1.0`.

Version migration was put together in a way that was quick but not easily maintainable for future changes or patch additions, since line bloat would easily take over the file as time goes on. Additionally, there was a halt on versioning at all given that the server would force the client to reload on receiving outdated sessions, which is no longer an issue with real-time version migration. With the companion server changes, versioning can also now be resumed and migration can be done without repeats (causing scenarios like #4100).

- Utilizes [`npm version minor --force --no-git-tag-version`](https://docs.npmjs.com/cli/v9/commands/npm-version) to increment the version up to `1.1.0`.
- Moved all patches into their own separate files all under the `versions` folder
  - e.g, all `1.0.4` and below patches exist in `src/system/version_migration/versions/v1_0_4.ts` with specific arrays per each data type (`sessionMigrators`, `systemMigrators`, etc.)
- `src/system/version_migration/version_converter.ts` now holds several `VersionConverter` classes that better reuse and encapsulate version parsing, comparison, and migration.
  - Each `VersionConverter` supports application of functions at a granular level with differentiation based on major, minor, and patch values in addition to having different patching timelines. This means that `SessionSaveData` could have a lot of different version specific patches migrating from `1.0.4` to `1.1.0` to `1.X.X` with multiple `case` entries within the existing `switch` blocks, but that doesn't need to be the case for `SystemSaveData` since it has its own `applyMigration` function. 

### Screenshots/Videos

For reference, this is the phase that was removed.

outdatedPhase

## How to test the changes?

If on hand, any save data or session data can be imported to check if the game will successfully migrate from its previous version to the latest version without hiccup. 

For the server changes, a local mirror of the server using Docker connected to the frontend would need to be installed in order to see that newer versions beyond `1.0.4` will load without causing error or flagging the removed `OutdatedPhase`.

## Checklist
  -  https://github.com/pagefaultgames/pokerogue/pull/4344
- [x] Are the changes visual?


flx-sta (commented):
So apart from the typical NITs I'd put here (missing override, missing docs. etc) I have a general concern with the approach.

I think it should be possible to design the script in a way that you don't have to import every single migrator file into the main file

```ts
import * as v1_0_4 from "./versions/v1_0_4"; // ⚠️ BAD
```

Instead it should be done via `import()` (and the `await` that comes with it).

There is two benefits to that:
1.  it's makes the design more intuitive as the next developer doesn't have to worry about importing the file (but naming it properly obv.)
2. The chunks that vite generates for the build are structured better because the code doesn't have to have all version migrators constantly loaded but instead only loads the ones that are needed.

src/system/version_migration/versions/v1_0_4.ts
```diff
@@ -0,0 +1,137 @@
...
+
+
+export const systemMigrators = [
+  /**
+   * Migrate ability starter data if empty for caught species
+   * @param data {@linkcode SystemSaveData}
+   */
+  function migrateAbilityData(data: SystemSaveData) {
+    if (data.starterData && data.dexData) {
+      // Migrate ability starter data if empty for caught species
```
DayKev, line 13:
```suggestion
```
You probably don't need to copy the comment that's 4 lines above lol.

flx-sta:
Another edge case to consider is if in a single file the functions have to be executed in a specific order. What guarantees that ?

xsn34kzx:
> Another edge case to consider is if in a single file the functions have to be executed in a specific order. What guarantees that ?

@flx-sta Given that functions are organized in lists for the moment (and that the previous system data migration functions were in a "order", though it's unclear how much that order mattered previously), they should all execute as they appear in the files they're written in. Nothing necessarily enforces order though as far as the IDE complaining, we would have to review and make sure that any new migration functions get tacked onto the end of the list.

src/system/version_migration/versions/v1_0_4.ts
xsn34kzx: Copy-paste leftover, will remove 🫡

Xavion3:
Can we avoid using the nested switch approach? Just like one file per version with loading the functions from them into some kind of list or map feels like it'd work so much better.

xsn34kzx:
Reduced some of the hacky-ness and went back to function calls for initializing the `VersionConverter` objects and running their migration protocols. I also added a "pre-processing" method for any data sanitation that always runs regardless of version for instances like money sanitizing. I also took out the `switch-case` statements in favor of `if` statements they're much smaller and more flexible for what my concern is with using maps (multi-layered maps to discern major, minor, and patch numbers or having to have multiple entries that point to the same starter function since, for all intents and purposes, `1.0.0` up to `1.0.4` have been coalesced into a  "blob" version that would all start at the `1.0.4` migration function.

Still trying to figure out where to fit in @flx-sta's comment since I think it's a valid optimization as version files become more plentiful and store more functions. I tried making some things `async` with `await`s, but things proved to not work since the process relies heavily on making sure things happen one after another strictly.

So far, I think this is a solid draft and works consistently as far as my testing with older files (session and system data).

DayKev (approved):

flx-sta:
@xsn34kzx here is something interesting for your further development:

https://www.npmjs.com/package/semver

flx-sta:
And here is an example of such an "async" migration spit out by ChatGPT:

<details>
    <summary> Migration approach by ChatGPT </summary>

By organizing migration scripts in files named after each version, you'll have a clear and maintainable version history for your savegame migration process.

### Implementation Plan:

1. **Directory Structure**:
   Organize your migration files in a dedicated folder, something like:

   ```
   migrations/
     v1_0_1.ts
     v1_0_2.ts
     v1_1_0.ts
     v1_2_0.ts
   ```

2. **Migration File Template**:
   Each file will export a migration function that takes the current savegame and returns the migrated version.

   Example of `v1_0_1.ts`:
   ```typescript
   export const migrateToV1_0_1 = (savegame: any) => {
     // Apply changes for version 1.0.1
     savegame.playerData.nickname = "NewPlayer";
     savegame.version = "1.0.1";
     return savegame;
   };
   ```

3. **Migration Loader**:
   Create a loader that dynamically imports and runs the necessary migration files based on the savegame's current version. You could automate the migration sequence using file imports.

   ```typescript
   async function migrateSavegame(savegame: any, targetVersion: string): Promise<any> {
     const migrations = [
       'v1_0_1', 'v1_0_2', 'v1_1_0', 'v1_2_0' // Add all migrations in order
     ];

     for (const migration of migrations) {
       if (savegame.version === migration.replace('_', '.')) {
         const { migrateToV1_0_1 } = await import(`./migrations/${migration}`);
         savegame = migrateToV1_0_1(savegame);
       }
     }

     if (savegame.version !== targetVersion) {
       throw new Error('Migration failed to reach the latest version.');
     }

     return savegame;
   }
   ```

4. **Version Comparison**:
   Ensure that your system properly handles version comparison. You can use string comparison, or more robust methods like `semver`.

   ```typescript
   import semver from 'semver';

   // Ensure versions are compared correctly
   if (semver.lt(savegame.version, '1.2.0')) {
     // apply migrations
   }
   ```

5. **Migration Order**:
   Be careful with the order in which migrations are applied. Since you already plan to use filenames representing versions, they should be applied sequentially in the correct order.

This structure will make it easy to:
- Add new migrations without changing existing code.
- Track changes across versions clearly.
- Apply migrations automatically based on the version.

</details>

DayKev (dismissed):

DayKev (dismissed):

DayKev (approved):

torranx (approved):
cool change, would love to see the condition for error handling (`error.startsWith("session out of date")` not rely on the `string` error message soonTM. anyways, looks good to me

f-fsantos (approved):

PR #4397, opened 2024-09-23
[Enhancement] Remove unnecessary loading of trainer sprites in loading scene

ben-lear:
Removes ~5 MB of image data (not counting actual heap usage) from being loaded on every session load. These images are already loaded on demand when trainers are spawned, there is no need to keep all of these assets loaded for every session.

https://github.com/user-attachments/assets/bcc05970-703f-4062-a205-55375ae9504e

https://github.com/user-attachments/assets/0b9c7509-2cc4-44fc-a130-a0b1fee7ebc4

https://github.com/user-attachments/assets/4bc9c14d-b30e-4bc8-ac14-c8811b491897

### DIFF
--- a/src/loading-scene.ts
+++ b/src/loading-scene.ts
@@ -1,5 +1,4 @@
 import { GachaType } from "./enums/gacha-types";
-import { trainerConfigs } from "./data/trainer-config";
 import { getBiomeHasProps } from "./field/arena";
 import CacheBustedLoaderPlugin from "./plugins/cache-busted-loader-plugin";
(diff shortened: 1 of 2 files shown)


ben-lear:
> (no access to a computer to do it right now, but it would be nice to check that this doesn't mess up the run history since it uses trainer sprites if the run was lost to a trainer)

Good catch, fixed that and working now.

https://github.com/user-attachments/assets/5beef7fc-5540-46ac-944f-4a5e83575c7b

Issue #4400, opened 2024-09-24
[P2] Toxic misses semi-invulnerable targets even if the user is Poison-type


Lneacx:
### Describe the bug

Toxic, when used by a Poison-type Pokémon, should be guaranteed to hit targets who are semi-invulnerable due to Fly, Dig, etc. This behavior is described in the Bulbapedia pages for [Toxic](https://bulbapedia.bulbagarden.net/wiki/Toxic_(move)#Generation_VI_onwards) and [Semi-invulnerability](https://bulbapedia.bulbagarden.net/wiki/Semi-invulnerable_turn#Moves_that_can_hit_semi-invulnerable_Pok%C3%A9mon).

In PokéRogue, Poison-types using Toxic miss semi-invulnerable targets.

### Session export file

This session features a Poison-type starter with Toxic facing an opponent who uses only moves that grant semi-invulnerability.
[sessionData_Guest.txt](https://github.com/user-attachments/files/17113355/sessionData_Guest.txt)

### Expected behavior

Toxic should hit even if the opponent is semi-invulnerable, because the user is Poison-type.

Repro steps with the session file (I'm assuming the RNG seed is the same as in my testing):
1. On the first turn, have Toxapex Run. This will fail. Bidoof will respond with a semi-invulnerable move, probably Dig.
2. On the second turn, while Bidoof is still underground, use Toxic. I expect this to hit, but instead Bidoof avoids it.

### What OS did you observe the bug on?

PC/Windows, Linux

### Which browser do you use?

Chrome

PR #4401, opened 2024-09-24
[Bug] Fix Toxic to hit through semi-invlnerability when used by a Poison-type.

Lneacx:

Currently, using Toxic on a flying / digging / diving / etc. enemy will always miss, even if the user is Poison-type. After this change, a Poison-type using Toxic will hit those enemies.

As described in #4400, in the main series games Toxic is guaranteed to hit when used by a Poison-type, even against semi-invulnerable enemies. By making this interaction work how it does in the main series games, I believe Toxic's behavior in PokéRogue will line up better with player expectations.

I edited the logic for `hitCheck` in `src/phases/move-effect-phase.ts` to ignore semi-invulnerability if the user is Poison-type and the move has Toxic's Accuracy attribute (in addition to the existing exception for `HitsTagAttr`).
I considered updating the data for Toxic or `ToxicAccuracyAttr` in `data/move.ts` instead because I'd rather not change game logic if I don't have to. But as far as I can tell, the only ways for a move to bypass semi-invulnerability is by having an `HitsTagAttr` or updating `hitCheck`. I don't think adding a `HitsTagAttr` to Toxic would help, because Toxic's ability to hit semi-invuln foes varies based on the user's type, which is not something that `HitsTagAttr` supports.

### Screenshots/Videos

Before: [Screen recording 2024-09-24 6.10.18 AM.webm](https://github.com/user-attachments/assets/a7d91330-0d1f-4f85-aaa5-bf428e6004ca)
After: [Screen recording 2024-09-24 6.11.55 AM.webm](https://github.com/user-attachments/assets/6b1dd7bd-8eff-422e-b5d2-000afe9ba765)

## How to test the changes?

Manually: Import the session file https://github.com/user-attachments/files/17113355/sessionData_Guest.txt, and observe that Toxic now hits through Dig whereas it did not before.
Automated Tests: I wrote some unit tests in toxic.test.js to verify that Toxic's sure-hit behavior works only for Poison-types.

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1201,7 +1201,7 @@ export default class BattleScene extends SceneBase {
 
       // Check for mystery encounter
       // Can only occur in place of a standard (non-boss) wild battle, waves 10-180
-      if (this.isWaveMysteryEncounter(newBattleType, newWaveIndex, mysteryEncounterType)) {
+      if (this.isWaveMysteryEncounter(newBattleType, newWaveIndex, mysteryEncounterType) || newBattleType === BattleType.MYSTERY_ENCOUNTER || !isNullOrUndefined(mysteryEncounterType)) {
         newBattleType = BattleType.MYSTERY_ENCOUNTER;
         // Reset base spawn weight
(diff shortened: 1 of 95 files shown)


Lneacx:
Ah, I accidentally did some weirdness with my commits so closing the PR for now while I correct it

PR #4430, opened 2024-09-25
[P2][Beta] Fix Sappy Seed applying its secondary effect against targets with Substitute

innerthunder:
This fixes a bug where Sappy Seed incorrectly applies its secondary effect when the target has put in a Substitute.

Reported [in the Discord](https://discord.com/channels/1125469663833370665/1284249564429160549)

- `data/move`: The "seeded" effect now has its own move attribute extending `AddBattlerTagAttr`. It now checks that the move does not hit its target's substitute before adding the tag.
- `test/moves/substitute.test`: New integration test for Sappy Seed hitting into a player Pokemon's substitute.

## How to test the changes?
`npm run test substitute`

## Checklist

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -4581,11 +4581,24 @@ export class AddBattlerTagAttr extends MoveEffectAttr {
   }
 }
 
+/**
+ * Adds a {@link https://bulbapedia.bulbagarden.net/wiki/Seeding | Seeding} effect to the target
+ * as seen with Leech Seed and Sappy Seed.
+ * @extends AddBattlerTagAttr
+ */
 export class LeechSeedAttr extends AddBattlerTagAttr {
   constructor() {
     super(BattlerTagType.SEEDED);
   }
 
+  /**
+   * Adds a Seeding effect to the target if the target does not have an active Substitute.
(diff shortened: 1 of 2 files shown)


src/data/move.ts
```diff
@@ -4581,6 +4581,30 @@ export class AddBattlerTagAttr extends MoveEffectAttr {
...
+export class LeechSeedAttr extends AddBattlerTagAttr {
+  constructor() {
+    super(BattlerTagType.SEEDED);
+  }
+
+  /**
+   * Adds a Seeding effect to the target if the target does not have an active Substitute.
+   * @param user the {@linkcode Pokemon} using the move
+   * @param target the {@linkcode Pokemon} targeted by the move
+   * @param move the {@linkcode Move} invoking this effect
```
torranx, line 4598: can you add types here please

innerthunder: the types are linked?

DayKev (approved):

frutescens (approved):

Tempo-anon (approved):
Discussed elsewhere, but maybe we should add a new flag to battler tags to see if they bypass substitutes

PR #4445, opened 2024-09-26
[P2] Fix Toxic to bypass semi-invulnerability when used by a Poison-type.

Lneacx:
This is a duplicate of #4401, which I closed because I wrecked my branch accidentally.

Currently, using Toxic on a flying / digging / diving / etc. enemy will always miss, even if the user is Poison-type. After this change, a Poison-type using Toxic will hit those enemies.

As described in #4400, in the main series games Toxic is guaranteed to hit when used by a Poison-type, even against semi-invulnerable enemies. By making this interaction work how it does in the main series games, I believe Toxic's behavior in PokéRogue will line up better with player expectations.

I edited the logic for `hitCheck()` in `src/phases/move-effect-phase.ts` to ignore semi-invulnerability if the user is Poison-type and the move has Toxic's Accuracy attribute (in addition to the existing exception for HitsTag attr).

I considered an alternative, which I'm open to implementing instead:
* Create a HitsTag attribute with a boolean condition function:
   * Add the condition either by adding a new field directly to `HitsTagAttr` or by extending it. Have `hitCheck()` evaluate the condition to decide whether to actually apply the `HitsTagAttr`. For each semi-invuln tag, add an attribute to Toxic to hit that tag, with a condition that checks if the user is Poison-type.
    * I like this change over my current approach because it doesn't hardcode a Toxic attr into `hitCheck()` by name. But this approach feels weird because it adds a new field to HitsTag attr that is used just by Toxic; I don't think any other HitsTag attr users would need a condition now or in the future.

### Screenshots/Videos

Before: [Screen recording 2024-09-24 6.10.18 AM.webm](https://github.com/user-attachments/assets/a7d91330-0d1f-4f85-aaa5-bf428e6004ca)
After: [Screen recording 2024-09-24 6.11.55 AM.webm](https://github.com/user-attachments/assets/6b1dd7bd-8eff-422e-b5d2-000afe9ba765)

## How to test the changes?

You can manually test this by importing this session file (https://github.com/user-attachments/files/17113355/sessionData_Guest.txt) and observing that Toxic now hits through Dig whereas it did not before.
I also wrote some unit tests in `toxic.test.ts` to verify that Toxic's sure-hit behavior works only for Poison-types.

## Checklist
- [x] Are the changes visual?


src/test/moves/toxic.test.ts
```diff
@@ -0,0 +1,77 @@
+import { BerryPhase } from "#app/phases/berry-phase";
```
DayKev, line 1:
```suggestion
```

src/test/moves/toxic.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      .enemySpecies(Species.MAGIKARP)
+      .enemyMoveset(Moves.SPLASH);
+  });
+
+  it("should be guaranteed to hit if user is Poison-type", async () => {
+    vi.spyOn(allMoves[Moves.TOXIC], "accuracy", "get").mockReturnValue(0);
+    await game.classicMode.startBattle([Species.TOXAPEX]);
+
+    game.move.select(Moves.TOXIC);
+    await game.phaseInterceptor.to(BerryPhase, false);
```
DayKev, line 39:
```suggestion
    await game.phaseInterceptor.to("BerryPhase", false);
```

src/test/moves/toxic.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    expect(game.scene.getEnemyPokemon()!.status?.effect).toBe(StatusEffect.TOXIC);
+  });
+
+  it("may miss if user is not Poison-type", async () => {
+    vi.spyOn(allMoves[Moves.TOXIC], "accuracy", "get").mockReturnValue(0);
+    await game.classicMode.startBattle([Species.UMBREON]);
+
+    game.move.select(Moves.TOXIC);
+    await game.phaseInterceptor.to(BerryPhase, false);
```
DayKev, line 49:
```suggestion
    await game.phaseInterceptor.to("BerryPhase", false);
```

src/test/moves/toxic.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  it("should miss semi-invulnerable targets if user is not Poison-type", async () => {
+    vi.spyOn(allMoves[Moves.TOXIC], "accuracy", "get").mockReturnValue(-1);
+    game.override.enemyMoveset(Moves.FLY);
+    await game.classicMode.startBattle([Species.UMBREON]);
+
+    game.move.select(Moves.TOXIC);
+    await game.setTurnOrder([BattlerIndex.ENEMY, BattlerIndex.PLAYER]);
+    await game.phaseInterceptor.to(BerryPhase, false);
```
DayKev, line 73:
```suggestion
    await game.phaseInterceptor.to("BerryPhase", false);
```

DayKev (approved):

DayKev:
This is the simpler implementation, and it seems like either or requires some "hardcoding". Lemme see if there are any other "if type then always hit" interactions.

DayKev:
Awkwardly it seems Toxic is the only move that conditionally hits semi-invulnerable targets. I suppose a condition function parameter could be added to `HitsTagAttr`...
Actually this is yet another case for conditional move attrs, but that's potentially a big refactor. For now I think this simple approach is fine.

DayKev (approved):

Issue #4351, opened 2024-09-20
[Feature] Add username visual


flx-sta:
## Describe the Feature

- The username should be displayed on the title screen
   - There should be an option to hide the username for screenshots, videos & streamers

flx-sta:
Username visible in summary screen

PR #4456, opened 2024-09-26
[Enhancement] Add username to title screen + add possibility to hide it

prisca-c:

User will see his username appear on title screen above `connected user` label .
User will also be able to hide username from settings menu.

This is a PR for issue #4351 !
Some ppl can switch from an account to another for challenge purpose or wont share their username when sharing their game. I guess the issue #4351 's has been created for these kind of issues.

Added a text object inside `src/ui/title-ui-handler.ts` to hold/display the usename.
To handle username changes on title screen , it listen to an event `sessionUpdate` using phaser eventEmiter, this event is called when title screen is loaded at the benning and also when user switch  `Hidden Username` 's states on settings menu.
I've added new entry inside `src/system/settings.ts` and a public  property to display Hidden Username setting inside the General setting menu.

### Screenshots/Videos

`Before`

title-before
settings-before

`After`

title
settings

## How to test the changes?

Just login in or not (bypass or not), user not logged in will show "Guest" on title, if you want to hide it just go to general setting on bottom, it is called `Hidden Username` in EN or `Pseudo caché` in FR (Sorry did not translate other languaga since i dont want to make translation mistake).

## Checklist
- [x] If I have text, did I make it translatable and add a key in the English locale file(s)? (Only EN and FR )
- [x] Are the changes visual?

### DIFF
--- a/src/ui/title-ui-handler.ts
+++ b/src/ui/title-ui-handler.ts
@@ -6,6 +6,7 @@ import { TextStyle, addTextObject, getTextStyleOptions } from "./text";
 import { getSplashMessages } from "../data/splash-messages";
 import i18next from "i18next";
 import { TimedEventDisplay } from "#app/timed-event-manager";
+import {loggedInUser} from "#app/account";
 
 export default class TitleUiHandler extends OptionSelectUiHandler {
   /** If the stats can not be retrieved, use this fallback value */
@@ -15,6 +16,7 @@ export default class TitleUiHandler extends OptionSelectUiHandler {
   private playerCountLabel: Phaser.GameObjects.Text;
   private splashMessage: string;
   private splashMessageText: Phaser.GameObjects.Text;
+  private usernameText: Phaser.GameObjects.Text;
   private eventDisplay: TimedEventDisplay;
 
   private titleStatsTimer: NodeJS.Timeout | null;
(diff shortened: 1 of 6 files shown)


src/ui/title-ui-handler.ts
```diff
@@ -43,18 +45,35 @@ export default class TitleUiHandler extends OptionSelectUiHandler {
       this.titleContainer.add(this.eventDisplay);
     }
 
+    this.usernameText = addTextObject(
+      this.scene,
+      (this.scene.game.canvas.width / 6) - 2,
+      (this.scene.game.canvas.height / 7) - 576 * getTextStyleOptions(TextStyle.WINDOW, this.scene.uiTheme).scale,
+      loggedInUser?.username || "Guest",
```
torranx, line 52: can we store this in a variable? since there are 2 occurrences of this

torranx (commented):
looks good but I don't agree with having a dedicated option for this, reasons being:
- Addtl settings should have a reasonable "merit" - I don't think a setting specific for screenshots, videos and streamers is enough justification, considering the username is to be placed in the Title Phase
- With the above said, we are convoluting the settings with UI options

prisca-c:
> I don't agree with having a dedicated option for this, reasons being:
> 
> * Addtl settings should have a reasonable "merit" - I don't think a setting specific for screenshots, videos and streamers is enough justification, considering the username is to be placed in the Title Phase
> * With the above said, we are convoluting the settings with UI options

Yes i understand your point !
Will make some changes according to your return

Tempo-anon:
I'm not sure if this is a needed feature. Having one's username on the title screen sort of sticks out and also requires a new settings option to have it be on/off. Getting one's username in-game is not that difficult as all it entails is opening up a Pokemon's summary screen. The given usecase of people having specific accounts for specific purposes seems like it's too much of an edge case to warrant this being a feature to me.

PR #4463, opened 2024-09-27
[Bug] [Balance] Gimmighoul evolution adjustments/fixes

Fontbane:
## What are the changes the user will see?

- Fixes count of Treasure tracker resetting to 1 after resetting the game, and bug which allows the player to receive free balls by keeping Gimmighoul unevolved and picking up Nuggets
- Allows Amulet Coin, Golden Punches, and Golden balls to count towards Gimmighoul evolution

## Why am I making these changes?

- Desired by balance team
- Avoids confusion for player
- Fixes bug introduced by #2732 

## What are the changes from a developer perspective?

- Checks the number of gold modifiers when checking the evo counter
- Changes the max stat limit for the tracker so you don't get extra balls after hitting it

### Screenshots/Videos

After 8 nuggets, 1 golden ball and 1 amulet coin:
image

## Checklist
- [X] **I'm using `beta` as my base branch**
- [X] There is no overlap with another PR?
- [X] The PR is self-contained and cannot be split into smaller PRs?
- [X] Have I provided a clear explanation of the changes?
- [X] Have I considered writing automated tests for the issue?
- [ ] If I have text, did I make it translatable and add a key in the English locale file(s)?
- [X] Have I tested the changes (manually)?
    - [X] Are all unit tests still passing? (`npm run test`)
- [X] Are the changes visual?
  - [X] Have I provided screenshots/videos of the changes?

PR #4473, opened 2024-09-27
[P2 Bug] [Balance] Adjust/fix Gimmighoul evolution

Fontbane:

- Re-implements #4463, reverting the last commit that broke it
- Fixes issue where Treasure stack could be full and players got free red balls
- Fixes issue where Treasure counter would disappear after saving (if your save was affected by this, pick up a Nugget, Big Nugget, or Relic Gold and the tracker will reappear with the correct number)

- When I posted #4463 it worked, but came under scrutiny specifically for the use of a magic number in the max stack count function. I pushed a commit to make it not be a magic number, but neglected to test it due to the pressure at the time. This broke it, leading to its reversion in #4472. The issue was that the max stack limit was too low.
- I've restored it to a functioning version so it can be re-implemented, as it fixes a bug on a popular mon.
- I also fixed the saving bug while I was at it

- Overrides getIconStackText for the evo tracker to display as red once you met the requirement
- Max stack count for the treasure tracker item is 999

### Screenshots/Videos
Screenshot_923

## How to test the changes?

- Use Gimmighoul, pick up treasure items with its evolution paused/cancelled until it goes beyond 10 on the tracker.

## Checklist
- [ ] Are the changes visual?


src/modifier/modifier.ts
```diff
@@ -852,26 +852,45 @@ export class EvoTrackerModifier extends PokemonHeldItemModifier {
...
+    if (this.getStackCount() >= this.required) {
+      text.setTint(0xf89890);
+    }
+    text.setOrigin(0, 0);
+
+    return text;
+  }
+
+  getMaxHeldItemCount(pokemon: Pokemon): integer {
+    this.stackCount = pokemon.evoCounter + pokemon.getHeldItems().filter(m => m instanceof DamageMoneyRewardModifier).length + pokemon.scene.findModifiers(m => m instanceof MoneyMultiplierModifier || m instanceof ExtraModifierModifier).length;
```
DayKev, line 892:
```suggestion
    this.stackCount = pokemon.evoCounter + pokemon.getHeldItems().filter(m => m instanceof DamageMoneyRewardModifier).length
      + pokemon.scene.findModifiers(m => m instanceof MoneyMultiplierModifier || m instanceof ExtraModifierModifier).length;
```

src/modifier/modifier.ts
```diff
@@ -852,26 +852,45 @@ export class EvoTrackerModifier extends PokemonHeldItemModifier {
...
-    return this.required;
+  getIconStackText(scene: BattleScene, virtual?: boolean): Phaser.GameObjects.BitmapText | null {
+    if (this.getMaxStackCount(scene) === 1 || (virtual && !this.virtualStackCount)) {
+      return null;
+    }
+
+    const pokemon = scene.getPokemonById(this.pokemonId);
+
+    this.stackCount = pokemon
+      ? pokemon.evoCounter + pokemon.getHeldItems().filter(m => m instanceof DamageMoneyRewardModifier).length + pokemon.scene.findModifiers(m => m instanceof MoneyMultiplierModifier || m instanceof ExtraModifierModifier).length
```
DayKev, line 878:
```suggestion
      ? pokemon.evoCounter + pokemon.getHeldItems().filter(m => m instanceof DamageMoneyRewardModifier).length
        + pokemon.scene.findModifiers(m => m instanceof MoneyMultiplierModifier || m instanceof ExtraModifierModifier).length
```

DayKev (approved):

Tempo-anon (approved):

PR #4485, opened 2024-09-28
sanitize all money when saving session data to server

ben-lear:
Handles any accidental corrupted saves with non-integer money values and sanitizes the value before saving to server.

### DIFF
--- a/src/system/game-data.ts
+++ b/src/system/game-data.ts
@@ -957,7 +957,7 @@ export class GameData {
       enemyModifiers: scene.findModifiers(() => true, false).map(m => new PersistentModifierData(m, false)),
       arena: new ArenaData(scene.arena),
       pokeballCounts: scene.pokeballCounts,
-      money: scene.money,
+      money: Math.floor(scene.money),
       score: scene.score,
       waveIndex: scene.currentBattle.waveIndex,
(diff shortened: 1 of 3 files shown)


src/system/game-data.ts
```diff
@@ -1047,7 +1047,7 @@ export class GameData {
             scene.pokeballCounts = Overrides.POKEBALL_OVERRIDE.pokeballs;
           }
 
-          scene.money = sessionData.money || 0;
+          scene.money = Math.floor(sessionData.money || 0);
```
DayKev, line 1050: Using `||` is actually probably more dangerous than `??` but w/e.

DayKev (approved):

src/system/version-converter.ts
```diff
@@ -6,6 +6,10 @@ const LATEST_VERSION = "1.0.5";
 
 export function applySessionDataPatches(data: SessionSaveData) {
   const curVersion = data.gameVersion;
+
+  // Always sanitize money as a safeguard
+  data.money = Math.floor(data.money);
+
```
xsn34kzx, line 12: Just to note it down for later, this *should* be enough (can look into it later) but better to be safe than sorry for now with flooring as much as possible.

xsn34kzx (approved):

PR #4494, opened 2024-09-28
[Bug] Prevent MEs incorrectly spawning on existing session saves

ben-lear:
Server is returning value of `0` for `mysteryEncounterType` on sessions that existed before the update (`0` is a valid ME, but an ME is not valid for the session state) which causes Mysterious Challengers to spawn on old saves.

This removes the early exit on the check for a ME so that it always must succeed on the roll to spawn a ME (which is seed offset, so refreshes/resets should never affect it).

This will *not* completely solve the fact that *if* a ME randomly determines to spawn on that first wave, it will be forced into Mysterious Challengers. There's not really any workaround to that.


MokaStitcher (approved):
lgtm

PR #4496, opened 2024-09-28
fix Black Sludge on session reload

ben-lear:
Fixes session data for Black Sludge item missing its shop modifier.


MokaStitcher (approved):
tested, works for new saves but not old ones

Tempo-anon (approved):

PR #4502, opened 2024-09-29
[Enhancement] Set appropriate evil team on monogen challenges, attempt 2

Madmadness65:

This is a redo of #4355, looking to address the two major issues with it that were noticed, that being wave 115 and 165 set reward tiers were not being given (which has been fixed), and generation 3 and 7 evil teams randomly switching between the two teams between waves (which should be fixed now).

Other than that, the premise of this PR is still the same: the evil team you face in monogen challenges will now be determined by what generation you choose.

Being able to set evil teams like the Elite Four/Champions is a desired change. The necessary fixes could not be put in before the previous beta to main merge, so this PR will look to rectify that.

The set waves on 35, 62, 64, 66, 112, 114, 115, and 165 are now split nine ways to accommodate the nine generations of teams. Team Rocket is intentionally duplicated to handle both generation 1 and 2 monogen. Generation 3 and 7 have special handling to split the evil teams up based on a `randSeedInt` value, which is set once when the challenge is loaded.

### Screenshots/Videos

https://github.com/user-attachments/assets/e5d9ee22-0e7b-4b02-83ba-48e6df492fbe

## How to test the changes?

Do some monogen challenge runs. You can also override the waves to the evil team fights to immediately test them.

## Checklist
- [x] Are the changes visual?


src/data/challenge.ts
```diff
@@ -109,6 +110,7 @@ export abstract class Challenge {
   public maxValue: integer; // The maximum strength of the challenge.
   public severity: integer; // The current severity of the challenge. Some challenges have multiple severities in addition to strength.
   public maxSeverity: integer; // The maximum severity of the challenge.
+  public rand: integer; // A random number applied to the challenge.
```
Tempo-anon, line 113: Should probably change all the `integer` to `number`s

src/data/challenge.ts
```diff
@@ -447,7 +451,187 @@ export class SingleGenerationChallenge extends Challenge {
...
+          if (this.rand !== 2) {
+            trainerTypes = [ TrainerType.UNKNOWN, TrainerType.UNKNOWN, TrainerType.UNKNOWN, TrainerType.UNKNOWN, TrainerType.UNKNOWN, TrainerType.UNKNOWN, TrainerType.AETHER_GRUNT ];
+            break;
+          } else {
+            trainerTypes = [ TrainerType.UNKNOWN, TrainerType.UNKNOWN, TrainerType.UNKNOWN, TrainerType.UNKNOWN, TrainerType.UNKNOWN, TrainerType.UNKNOWN, TrainerType.SKULL_GRUNT ];
+            break;
+          }
+        }
+        break;
+      case 62:
```
Tempo-anon, line 481: 62, 64, 112, and 114 are all the same so this can be ported into a helper function

src/data/challenge.ts
```diff
@@ -122,6 +124,7 @@ export abstract class Challenge {
     this.maxValue = maxValue;
     this.severity = 0;
     this.maxSeverity = 0;
+    this.rand = 0;
```
Tempo-anon, line 127: Why not just have the variable here save the evil team for all runs instead of just for gens 3 and 7?

Madmadness65: I set `rand` that way in case it were to be used by any other existing or future challenges. Plus, I wouldn't set that value in the abstract class anyways, it's fine being set within the challenge classes themselves.

Tempo-anon: In terms of future proofing, I think it would be better if we have individual variables for tracking things like `evilTeamThisRun` rather than a generic `rand` that's used for all various randomization needed for future challenges.

PR #4511, opened 2024-09-30
[Move] Fully Implement the Pledge Moves

innerthunder:
This fully implements the moves Water Pledge, Fire Pledge, and Grass Pledge.

Less (P) moves

- [x] `field/pokemon`: `Pokemon.turnData.combiningPledge` added to track which Pledge move (if any) is being used in combination with the Pokemon's selected move this turn.
- [x] `data/move`:
  - [x] Created `AwaitCombinedPledgeAttr`, which cancel's the user's Pledge move if an ally will use another unique Pledge move later in the turn.
  - [x] Created attributes for Pledge moves' type change, power boost, and guaranteed STAB boost when used in combination.
  - [x] Created `AddPledgeEffectAttr`, which adds an arena tag to the field when a Pledge move is used in combination with another unique Pledge move and successfully hits a target.
- [x] `data/arena-tag`: Added tags for the secondary effects of combined Pledge moves, including
  - [x] Fire/Grass Pledge: A "sea of fire" added to the target's side of the field, dealing damage to each affected non-Fire-type Pokemon equal to 1/8 of their max HP (rounded down, minimum 1 damage) at the end of each turn for 4 turns.
  - [x] Water/Fire Pledge: A "rainbow" added to the user's side of the field, doubling the secondary effect chance of the user and its ally's moves for 4 turns. This effect stacks with Serene Grace except when applying to a move's chance of flinching its target.
  - [x] Grass/Water Pledge: A "swamp" added to the target's side of the field, quartering the Speed of each Pokemon on that side for 4 turns.
- [x] `test/moves/pledge_moves.test`: New integration test coverage for the Pledge moves' behavior when used individually and in combination. 

### Screenshots/Videos

Fire/Grass Pledge Combo. Note that this video includes a text error that has since been fixed.

https://github.com/user-attachments/assets/93b14f70-08c5-454a-8a84-cb9130082552

## How to test the changes?
`npm run test pledge_moves` (Only basic interactions so far)

## Checklist
- [x] Are the changes visual?


Xavion3:
Is the `combiningPledge` field in the turn data actually assigned in this PR? A quick search doesn't reveal it.

Also a quick glance at the code makes me wonder what happens if in a single battle a pokemon uses a pledge move in a turn while already preparing to use a different pledge move. I'm not sure that can happen right now, but it's theoretically possible once we have more moves implemented.

We'll probably want test cases for the turn order altering effects too, since those are a notable aspect of this move and I can't see that in a test case or in the video.

SangaraSorama (dismissed):
For the locales folder.

src/data/move.ts
```diff
@@ -2685,6 +2692,55 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
...
+      return false;
+    }
+
+    const overridden = args[0] as Utils.BooleanHolder;
+
+    const allyMovePhase = user.scene.findPhase<MovePhase>(phase => phase instanceof MovePhase && phase.pokemon.isPlayer() === user.isPlayer());
+    if (allyMovePhase) {
+      const allyMove = allyMovePhase.move.getMove();
+      if (allyMove !== move && allyMove.hasAttr(AwaitCombinedPledgeAttr)) {
+        [user, user.getAlly()].forEach(p => p.turnData.combiningPledge = move.id);
```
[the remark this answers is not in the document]
innerthunder: This line doesn't run in that case because it requires an ally's `MovePhase` to exist in the phase queue (i.e. it would fail the `if (allyMovePhase)` check. There's also a test in the `pledge_moves` test suite that takes place in a single battle ("Fire Pledge - should not combine with an enemy's Pledge move").

src/data/move.ts
```diff
@@ -2685,6 +2692,62 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
...
+  override apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    if (user.turnData.combiningPledge) {
+      // "The two moves have become one!\nIt's a combined move!"
+      user.scene.queueMessage(i18next.t("moveTriggers:combiningPledge"));
+      return false;
+    }
+
+    const overridden = args[0] as Utils.BooleanHolder;
+
+    const allyMovePhase = user.scene.findPhase<MovePhase>(phase => phase instanceof MovePhase && phase.pokemon.isPlayer() === user.isPlayer());
```
DayKev, line 2724:
```suggestion
    const allyMovePhase = user.scene.findPhase<MovePhase>((phase) => phase instanceof MovePhase && phase.pokemon.isPlayer() === user.isPlayer());
```

src/data/move.ts
```diff
@@ -2685,6 +2692,62 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
...
+    if (allyMovePhase) {
+      const allyMove = allyMovePhase.move.getMove();
+      if (allyMove !== move && allyMove.hasAttr(AwaitCombinedPledgeAttr)) {
+        [user, user.getAlly()].forEach(p => p.turnData.combiningPledge = move.id);
```
DayKev:
```suggestion
        [user, user.getAlly()].forEach((p) => p.turnData.combiningPledge = move.id);
```

src/data/move.ts
```diff
@@ -2685,6 +2692,62 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
...
+
+        // "{userPokemonName} is waiting for {allyPokemonName}'s move..."
+        user.scene.queueMessage(i18next.t("moveTriggers:awaitingPledge", {
+          userPokemonName: getPokemonNameWithAffix(user),
+          allyPokemonName: getPokemonNameWithAffix(user.getAlly())
+        }));
+
+        // Move the ally's MovePhase (if needed) so that the ally moves next
+        const allyMovePhaseIndex = user.scene.phaseQueue.indexOf(allyMovePhase);
+        const firstMovePhaseIndex = user.scene.phaseQueue.findIndex(phase => phase instanceof MovePhase);
```
DayKev, line 2738:
```suggestion
        const firstMovePhaseIndex = user.scene.phaseQueue.findIndex((phase) => phase instanceof MovePhase);
```

src/data/move.ts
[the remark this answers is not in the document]
Xavion3: What if it is somehow called in a single battle such as due to the instruct case?

innerthunder:
> A quick glance at the code makes me wonder what happens if in a single battle a pokemon uses a pledge move in a turn while already preparing to use a different pledge move. I'm not sure that can happen right now, but it's theoretically possible once we have more moves implemented.

Just echoing things discussed in the Discord here. Instruct seems like the "problem move" for this interaction, and I suspect that in this situation, the current code would cause the following behavior to occur:
- Suppose the player Pokemon used Fire Pledge last turn and is planning on using Grass Pledge this turn.
- The enemy Pokemon is set to use Instruct on the player Pokemon this turn.
- If the enemy goes first in the turn, the player is instructed to use Fire Pledge, which then combines with the player's Grass Pledge.
- If the enemy goes second, the player uses Grass Pledge, then is instructed to use Grass Pledge again with no combined effect

@Snailman11 tested the case where the enemy outspeeds the player on Showdown [here](https://replay.pokemonshowdown.com/gen9customgame-2213090750-mu32biwczh04bviv7v0a04pmi661i8tpw), showing behavior consistent with the current implementation. We should take Showdown's implementation with a grain of salt, though, as Instruct has been known to be bugged in Showdown for certain interactions in the past.

> We'll probably want test cases for the turn order altering effects too, since those are a notable aspect of this move and I can't see that in a test case or in the video.

Just implemented the turn order altering effect, along with a new test for it.

src/data/move.ts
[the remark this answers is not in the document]
innerthunder: Ah, I see what you mean now. It's not totally correct to assume the combined moves are coming from different Pokemon... I'll look into fixing it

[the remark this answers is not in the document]
innerthunder: Just replaced references to `user.getAlly()` with `allyMovePhase.pokemon`, which should be able to account for the user combining its own Pledges. Not sure if that's the desired behavior (it should be consistent with Showdown, at least), but the code should be more stable now.

src/data/move.ts
```diff
@@ -4303,6 +4399,47 @@ export class MatchUserTypeAttr extends VariableMoveTypeAttr {
   }
 }
 
+/**
+ * Changes the type of a Pledge move based on the Pledge move combined with it.
+ * @extends VariableMoveTypeAttr
+ */
+export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
+  override apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    const moveType = args[0];
```
DayKev, line 4408: Why doesn't this one need `as NumberHolder` but `stabMultiplier` does?

src/test/moves/pledge_moves.test.ts
```diff
@@ -0,0 +1,272 @@
+import { BattlerIndex } from "#app/battle";
+import { ArenaTagSide } from "#app/data/arena-tag";
+import { allMoves, FlinchAttr } from "#app/data/move";
+import { Type } from "#app/data/type";
+import { ArenaTagType } from "#app/enums/arena-tag-type";
+import { Stat } from "#app/enums/stat";
```
DayKev, line 6:
```suggestion
import { ArenaTagType } from "#enums/arena-tag-type";
import { Stat } from "#enums/stat";
```

src/data/move.ts
innerthunder: Is this the standard convention/style choice for the codebase? I see lines like this written both with and without the `()` throughout the code...

DayKev: I believe they're all supposed to use `()`. Older code ofc just did whatever.

src/data/move.ts
innerthunder: I just switched the other attributes to also use `instanceof` checks.

DayKev (dismissed):

torranx (dismissed):
didn't see anything wrong - lgtm

src/test/moves/pledge_moves.test.ts
```diff
@@ -0,0 +1,272 @@
...
+    game.override
+      .battleType("double")
+      .startingLevel(100)
+      .moveset([Moves.FIRE_PLEDGE, Moves.GRASS_PLEDGE, Moves.WATER_PLEDGE, Moves.SPLASH])
+      .enemySpecies(Species.SNORLAX)
+      .enemyLevel(100)
+      .enemyAbility(Abilities.BALL_FETCH)
+      .enemyMoveset(Moves.SPLASH);
+  });
+
```
Tempo-anon, line 40: There should be a test case for if one of the Pokemon is incapacitated (paralysis, sky drop) resulting in no move used at all

src/data/move.ts
```diff
@@ -3721,6 +3784,45 @@ export class LastMoveDoublePowerAttr extends VariablePowerAttr {
...
+      return true;
+    }
+    return false;
+  }
+}
+
+/**
+ * Applies STAB to the given Pledge move if the move is part of a combined attack.
+ */
+export class CombinedPledgeStabBoostAttr extends MoveAttr {
```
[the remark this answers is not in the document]
Xavion3: It actually affects all of them, because the way pledge moves work mean that the pokemon actually making the attack doesn't necessarily use the same type of move as the one they clicked.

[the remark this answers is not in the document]
Tempo-anon: Well in the example of like a Charizard using fire pledge and a Squirtle (slower) using water pledge, the slower pokemon (the one doing the attack) will almost always still get the STAB boost since they natively get STAB for the pledge moves they can learn. The only times this wouldn't apply are for metronome, Mew, and maybe if wacky interactions change the typings of a Pokemon.

package-lock.json
```diff
@@ -7,6 +7,7 @@
 		"": {
 			"name": "pokemon-rogue-battle",
 			"version": "1.0.4",
+			"hasInstallScript": true,
```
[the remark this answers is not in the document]
innerthunder: Discussed on the Discord already, but this has been appearing for me whenever I run `npm install` locally. It seems like other devs also had this generated but would roll back the changes. I asked @f-fsantos about it and he said it was fine to push this.

src/data/arena-tag.ts
```diff
@@ -1025,6 +1026,81 @@ class ImprisonTag extends ArenaTrapTag {
...
+ * Damages all non-Fire-type Pokemon on the given side of the field at the end
+ * of each turn for 4 turns.
+ */
+class FireGrassPledgeTag extends ArenaTag {
+  constructor(sourceId: number, side: ArenaTagSide) {
+    super(ArenaTagType.FIRE_GRASS_PLEDGE, 4, Moves.FIRE_PLEDGE, sourceId, side);
+  }
+
+  override onAdd(arena: Arena): void {
+    // "A sea of fire enveloped your/the opposing team!"
```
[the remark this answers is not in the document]
innerthunder: I wrote these comments mostly to keep track of messages as I was researching them before adding the keys, but I figured they were helpful enough to keep 😅

src/data/move.ts
```diff
@@ -3721,6 +3784,45 @@ export class LastMoveDoublePowerAttr extends VariablePowerAttr {
...
+export class CombinedPledgePowerAttr extends VariablePowerAttr {
+  override apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    const power = args[0];
+    if (!(power instanceof Utils.NumberHolder)) {
+      return false;
+    }
+    const combinedPledgeMove = user.turnData.combiningPledge;
+
+    if (combinedPledgeMove && combinedPledgeMove !== move.id) {
+      power.value *= 150/80;
```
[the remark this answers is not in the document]
innerthunder: Exactly. Some power boosts (e.g. Normalize) apply before `VariablePowerAttr`s, so this being multiplicative makes sure those boosts aren't lost.

src/test/moves/pledge_moves.test.ts
```diff
@@ -0,0 +1,272 @@
...
+      await game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.PLAYER_2, BattlerIndex.ENEMY, BattlerIndex.ENEMY_2]);
+      await game.phaseInterceptor.to("MoveEndPhase", false);
+
+      expect(firePledge.calculateBattlePower).toHaveLastReturnedWith(80);
+      expect(playerPokemon[0].getMoveType).toHaveLastReturnedWith(Type.FIRE);
+    }
+  );
+
+  it(
+    "Fire Pledge - should not combine with an ally using Fire Pledge",
```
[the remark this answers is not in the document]
innerthunder: From Fire Pledge's page:

> When two allies attempt to use Fire Pledge and either [Water Pledge](https://bulbapedia.bulbagarden.net/wiki/Water_Pledge_(move)) or [Grass Pledge](https://bulbapedia.bulbagarden.net/wiki/Grass_Pledge_(move)) on the same turn, the ally moving first will not itself use a move, but instead the ally moving second will use a combined attack with a power of 150 and an additional effect immediately after it...

This implies that the allies must use two different Pledges for them to combine moves. [Smogon's movedex entry](https://www.smogon.com/dex/sv/moves/fire-pledge/) also backs this up.

src/data/move.ts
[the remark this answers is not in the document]
innerthunder: It matters in more cases than that. For example:
1. Player has Charizard using Fire Pledge and Venusaur using Grass Pledge
2. Charizard goes first and waits for Venusaur's move
3. Venusaur uses Grass Pledge, which becomes Fire type when combined with Fire Pledge. Despite changing type, the combined attack still gains STAB.

Mew is a pretty extreme case where this applies, seeing as it doesn't share types with *any* of the Pledges, but it applies more subtly in a lot of other cases

src/test/moves/pledge_moves.test.ts
```diff
@@ -0,0 +1,296 @@
...
+      await game.forceEnemyMove(Moves.SPORE, BattlerIndex.PLAYER_2);
+      await game.forceEnemyMove(Moves.SPLASH);
+
+      await game.setTurnOrder([BattlerIndex.ENEMY, BattlerIndex.PLAYER, BattlerIndex.PLAYER_2, BattlerIndex.ENEMY_2]);
+
+      await game.phaseInterceptor.to("BerryPhase", false);
+
+      enemyPokemon.forEach((p) => expect(p.hp).toBe(p.getMaxHp()));
+    }
+  );
```
Xavion3, line 295: This is theoretically inconsistent if two enemy pokemon immune to fire were rolled right?

DayKev: The enemy species is already set to be Snorlax at the top of the test suite (line 35).

innerthunder: Yeah, the `beforeEach` block makes it so all enemies in every test are Snorlax with an irrelevant ability

DayKev (approved):

src/data/move.ts
```diff
@@ -2687,6 +2694,62 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
...
+      return false;
+    }
+
+    const overridden = args[0] as Utils.BooleanHolder;
+
+    const allyMovePhase = user.scene.findPhase<MovePhase>((phase) => phase instanceof MovePhase && phase.pokemon.isPlayer() === user.isPlayer());
+    if (allyMovePhase) {
+      const allyMove = allyMovePhase.move.getMove();
+      if (allyMove !== move && allyMove.hasAttr(AwaitCombinedPledgeAttr)) {
+        [user, allyMovePhase.pokemon].forEach((p) => p.turnData.combiningPledge = move.id);
```
DayKev, line 2730:
```suggestion
        [ user, allyMovePhase.pokemon ].forEach((p) => p.turnData.combiningPledge = move.id);
```

src/data/move.ts
```diff
@@ -3762,6 +3825,45 @@ export class LastMoveDoublePowerAttr extends VariablePowerAttr {
...
+export class CombinedPledgePowerAttr extends VariablePowerAttr {
+  override apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    const power = args[0];
+    if (!(power instanceof Utils.NumberHolder)) {
+      return false;
+    }
+    const combinedPledgeMove = user.turnData.combiningPledge;
+
+    if (combinedPledgeMove && combinedPledgeMove !== move.id) {
+      power.value *= 150/80;
```
DayKev, line 3841:
```suggestion
      power.value *= 150 / 80;
```

src/test/moves/pledge_moves.test.ts
```diff
@@ -0,0 +1,296 @@
...
+  afterEach(() => {
+    game.phaseInterceptor.restoreOg();
+  });
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override
+      .battleType("double")
+      .startingLevel(100)
+      .moveset([Moves.FIRE_PLEDGE, Moves.GRASS_PLEDGE, Moves.WATER_PLEDGE, Moves.SPLASH])
```
DayKev, line 34:
```suggestion
      .moveset([ Moves.FIRE_PLEDGE, Moves.GRASS_PLEDGE, Moves.WATER_PLEDGE, Moves.SPLASH ])
```

src/test/moves/pledge_moves.test.ts
```diff
@@ -0,0 +1,296 @@
...
+      .enemySpecies(Species.SNORLAX)
+      .enemyLevel(100)
+      .enemyAbility(Abilities.BALL_FETCH)
+      .enemyMoveset(Moves.SPLASH);
+  });
+
+  it(
+    "Fire Pledge - should be an 80-power Fire-type attack outside of combination",
+    async () => {
+      await game.classicMode.startBattle([Species.BLASTOISE, Species.CHARIZARD]);
```
DayKev, line 44:
```suggestion
      await game.classicMode.startBattle([ Species.BLASTOISE, Species.CHARIZARD ]);
```

src/test/moves/pledge_moves.test.ts
```diff
@@ -0,0 +1,296 @@
...
+      const firePledge = allMoves[Moves.FIRE_PLEDGE];
+      vi.spyOn(firePledge, "calculateBattlePower");
+
+      const playerPokemon = game.scene.getPlayerField();
+      vi.spyOn(playerPokemon[0], "getMoveType");
+
+      game.move.select(Moves.FIRE_PLEDGE, 0, BattlerIndex.ENEMY);
+      game.move.select(Moves.SPLASH, 1);
+
+      await game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.PLAYER_2, BattlerIndex.ENEMY, BattlerIndex.ENEMY_2]);
```
DayKev, line 55:
```suggestion
      await game.setTurnOrder([ BattlerIndex.PLAYER, BattlerIndex.PLAYER_2, BattlerIndex.ENEMY, BattlerIndex.ENEMY_2 ]);
```

src/test/moves/pledge_moves.test.ts
```diff
@@ -0,0 +1,296 @@
...
+
+      expect(firePledge.calculateBattlePower).toHaveLastReturnedWith(80);
+      expect(playerPokemon[0].getMoveType).toHaveLastReturnedWith(Type.FIRE);
+    }
+  );
+
+  it(
+    "Fire Pledge - should not combine with an ally using Fire Pledge",
+    async () => {
+      await game.classicMode.startBattle([Species.BLASTOISE, Species.CHARIZARD]);
```
DayKev, line 66:
```suggestion
      await game.classicMode.startBattle([ Species.BLASTOISE, Species.CHARIZARD ]);
```

DayKev (dismissed):

torranx (dismissed):

DayKev (approved):

src/test/moves/pledge_moves.test.ts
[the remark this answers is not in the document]
Tempo-anon: Ah ok seems I misread. I was interpreting it as two of the same pledge would still boost the base power to 150

Tempo-anon (approved):

PR #4533, opened 2024-10-01
[P1] Fusion with Skorupi in second slot crashes game

pom-eranian:
The latest fix on skorupi's sprites causes a crash if the fusion has skorupi in second slot.
Hotfix to main reverted this update, while the cause of crash is investigated and files fixed accordingly.

image

## Checklist
- [] There is no overlap with another PR?
- [] The PR is self-contained and cannot be split into smaller PRs?
~~- [ ] Have I considered writing automated tests for the issue?~~
- [] Have I tested the changes (manually)?
- [x] Are the changes visual?
  - [] Have I provided screenshots/videos of the changes?


Snailman11:
Link to The Discord Post (Spliced endless run fails to continue to new wave)
https://discord.com/channels/1125469663833370665/1290008413975351317

MokaStitcher:
If and when the hotfix gets merged instead of this you can probably change this PR into the actual fix for the sprite and move it to draft while it's being investigated
I'd like to look more into why that code even gets called when fusion palettes are disabled too. Fixing that would at least allow players to continue playing if another issue like this one happens again

PR #4534, opened 2024-10-01
[Ability] Implement Unburden

muscode13:
Unburden implemented

Ability is not implemented

`PostItemLostAbAttr ` - Ability attribute that triggers when the Pokemon loses an item
`PostItemLostApplyBattlerTagAbAttr ` - Extends `PostItemLostAbAttr` and applies a BattlerTagType to the Pokemon when activated.
`BattlerTagType.UNBURDEN` - Doubles the Pokemon's effective speed.

## How to test the changes?
- Use Pokemon with Unburden in the game and give them berries. Also have a pokemon with Magician steal items from them in double battles
- run `npm run test unburden`

## Checklist
- [x] There is no overlap with another PR? - I believe the other Unburden PR was closed
- [ ] Are the changes visual?


Xavion3:
This should give a battler tag that doubles speed, not directly interact with combat stages.

muscode13:
Will do!

muscode13:
Alright I switched it to tags! I wasn't sure if it should keep doubling speed for each item lost, but I can implement that if you guys want.

Tempo-anon (changes requested):
I think for simplicity's sake of implementation a battler tag is a fine way of going about implementation. Any item lost/used up will grant the full doubled speed boost.

Right now I believe it's missing some documentation and edge case interactions though.

src/data/battler-tags.ts
```diff
@@ -1537,6 +1537,19 @@ export class AbilityBattlerTag extends BattlerTag {
...
+/**
+ * Tag used by Unburden to double speed
+ * @extends AbilityBattlerTag
+ */
+export class UnburdenTag extends AbilityBattlerTag {
+  constructor() {
+    super(BattlerTagType.UNBURDEN, Abilities.UNBURDEN, BattlerTagLapseType.CUSTOM, 1);
+  }
+  onAdd(pokemon: Pokemon): void {
+    pokemon.setStat(Stat.SPD, pokemon.getStat(Stat.SPD, false) * 2, false);
```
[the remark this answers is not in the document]
muscode13: I did onAdd here since I saw that Gorilla Tactics, which provides a similar boost to a stat, had it that way. I'm not sure if there is a functional difference though. I can change it to be persistent instead of an onAdd if that if preferred though

src/data/battler-tags.ts
```diff
@@ -1537,6 +1537,19 @@ export class AbilityBattlerTag extends BattlerTag {
   }
 }
 
+/**
+ * Tag used by Unburden to double speed
+ * @extends AbilityBattlerTag
+ */
+export class UnburdenTag extends AbilityBattlerTag {
```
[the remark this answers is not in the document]
muscode13: I'll double check and add a test case for that. I assumed that abilities like that did the heavy lifting of the logic.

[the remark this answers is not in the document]
Xavion3: They'll make it so the game won't acknowledge the pokemon as having unburden while neutralised, but they can't remove existing things like this so easily.

muscode13:
Alright I added documentation to the attributes, renamed them, added a test for neutralizing abilities, and accounted for getEffectiveStat

Xavion3:
Don't double the speed twice, current it doubles in both getEffectiveStat and onAdd

muscode13:
> Don't double the speed twice, current it doubles in both getEffectiveStat and onAdd

Oh right forgot about that. Got a fix, just trying to get the Grip Claw to actually steal in my tests now

muscode13:
OK should only be doubling in one place now

muscode13:
Any updates? Been waiting for a review for almost a week now

src/data/ability.ts
```diff
@@ -3790,6 +3792,117 @@ export class PostDancingMoveAbAttr extends PostMoveUsedAbAttr {
   }
 }
 
+/**
+ * Ability attribute for Unburden, triggers upon berry consumption
```
Tempo-anon, line 3796: Is this just for consumption? Or is it also for all removal like stealing, fling, knocked off, and incinerate? The name of the class seems like it doesn't match up with the functionality (checking berries eaten)

src/data/ability.ts
```diff
@@ -3790,6 +3792,117 @@ export class PostDancingMoveAbAttr extends PostMoveUsedAbAttr {
...
+    if (pokemon.getTag(BattlerTagType.UNBURDEN)) {
+      return false;
+    }
+
+    pokemon.addTag(BattlerTagType.UNBURDEN);
+    return true;
+  }
+
+  getCondition(): AbAttrCondition {
+    return (pokemon: Pokemon) => pokemon.battleData.berriesEaten.length !== 0;
```
Tempo-anon, line 3825: Is berriesEaten the right tag when there's other cases for berry removal?

src/data/ability.ts
```diff
@@ -3790,6 +3792,117 @@ export class PostDancingMoveAbAttr extends PostMoveUsedAbAttr {
...
+  }
+
+  getCondition(): AbAttrCondition {
+    return (pokemon: Pokemon) => pokemon.turnData.itemsLost > 0;
+  }
+
+}
+
+/**
+ * Ability attribute for Unburden, triggers upon an item being lost while attacking (Pickpocket)
```
Tempo-anon, line 3869: Also for things that remove an item when attacking? Things like fling, natural gift, etc

src/field/pokemon.ts
```diff
@@ -938,6 +938,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       if (this.status && this.status.effect === StatusEffect.PARALYSIS) {
         ret >>= 1;
       }
+      if (this.getTag(BattlerTagType.UNBURDEN)) {
```
Tempo-anon, line 941: This should have a check here to make sure there is no neutralizing gas on the field

Tempo-anon (changes requested):
I think functionally for the most part it is fine. The only real blocker I see is to have the neutralizing gas interaction. 

Should we also slap a (P) on unburden for the weird edge cases of gaining/losing this ability though?

src/data/ability.ts
muscode13: Just for a pokemon eating their berry. I'll reword the documentation. Other stuff like knock off and incinerate are handled by the UnburdenDefendingItemRemovedAbAttr, which tracks berry's lost through turnData.itemsLost. I did forget about fling, I'll double check that that causes a boost.

src/field/pokemon.ts
muscode13: Not sure if this is required. No other ability has a check for neutralizing gas, including slow start which is a couple lines above the Unburden speed modifier. I believe the canApplyAbility already addresses this in the Pokemon class

src/data/ability.ts
muscode13: I checked and all of them are unimplemented. The only one I found was stuff cheeks. To make them work with unburden, you'd just have to iterate the turnData.itemsLost when the move is used when they get implemented

muscode13:
Refactored my implementation in a way that I believe is more in line with the rest of the codebase. Let me know what you think!

src/data/ability.ts
```diff
@@ -4872,6 +4908,11 @@ export function applyPostFaintAbAttrs(attrType: Constructor<PostFaintAbAttr>,
   return applyAbAttrsInternal<PostFaintAbAttr>(attrType, pokemon, (attr, passive) => attr.applyPostFaint(pokemon, passive, simulated, attacker, move, hitResult, args), args, false, simulated);
 }
 
+export function applyPostItemLostAbAttrs(attrType: Constructor<PostItemLostAbAttr>,
+  pokemon: Pokemon, passive: boolean, simulated: boolean = false, args: any[]): Promise<void> {
```
innerthunder, line 4912:
```suggestion
  pokemon: Pokemon, passive: boolean, simulated: boolean = false, ...args: any[]): Promise<void> {
```
Matches other `applyAbAttrs` functions. This also means you don't have to pass empty arrays as arguments (e.g. in the new code in `battle-scene.ts`)

src/field/pokemon.ts
```diff
@@ -945,6 +945,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
         if (this.status && this.status.effect === StatusEffect.PARALYSIS) {
           ret >>= 1;
         }
+        if (this.getTag(BattlerTagType.UNBURDEN)) {
```
innerthunder, line 948:
```suggestion
        if (this.getTag(BattlerTagType.UNBURDEN) && !this.scene.getField(true).some(pokemon => pokemon !== this && pokemon.hasAbilityWithAttr(SuppressFieldAbilitiesAbAttr))) {
```
For the Neutralizing Gas interaction (please also add a unit test for neut gas Pokemon entering the field when Unburden has already activated)

innerthunder (changes requested):

src/field/pokemon.ts
muscode13: I'll commit this in, but I am confused why Unburden needs an explicit check for this when abilities like Gorilla Tactics and Slow Start do not.

src/battle-scene.ts
```diff
@@ -2593,9 +2593,19 @@ export default class BattleScene extends SceneBase {
           const addModifier = () => {
             if (!matchingModifier || this.removeModifier(matchingModifier, !target.isPlayer())) {
               if (target.isPlayer()) {
-                this.addModifier(newItemModifier, ignoreUpdate, playSound, false, instant).then(() => resolve(true));
+                this.addModifier(newItemModifier, ignoreUpdate, playSound, false, instant).then(() => {
+                  if (source) {
```
torranx, line 2597: sorry if this was answered before but are these necessary checks? iirc the attributes already does that

src/test/abilities/unburden.test.ts
torranx, line 1: nit: can we add proper line breaks here please, its hard to read

src/data/ability.ts
```diff
@@ -5215,6 +5251,11 @@ export function applyPostFaintAbAttrs(attrType: Constructor<PostFaintAbAttr>,
   return applyAbAttrsInternal<PostFaintAbAttr>(attrType, pokemon, (attr, passive) => attr.applyPostFaint(pokemon, passive, simulated, attacker, move, hitResult, args), args, false, simulated);
 }
 
+export function applyPostItemLostAbAttrs(attrType: Constructor<PostItemLostAbAttr>,
+  pokemon: Pokemon, passive: boolean, simulated: boolean = false, ...args: any[]): Promise<void> {
```
torranx, line 5255: I don't think parameter `passive` is necessary here since the attribute itself does not check whether or not the pokemon has passive?

src/battle-scene.ts
muscode13: If I don't check for source here I get a typescript error: Argument of type 'Pokemon | null | undefined' is not assignable to parameter of type 'Pokemon'. I could change the typing within my AbAttr to reflect this if you'd prefer that.

muscode13: Also, source is the pokemon with Unburden here

torranx: ah I see, this is ok then 👍

src/test/abilities/unburden.test.ts
torranx, line 1: nit: consistent line break after each test cases

torranx (commented):
sorry for being nitpicky, but otherwise looks good to me

torranx (approved):
lgtm!

innerthunder (approved):

PR #4543, opened 2024-10-02
[Move] Triple Arrows effect chance for stat change is now 50%

DayKev:
Triple Arrows will now cause the defense drop 50% of the time instead of 30%, and is no longer marked as "(P)".

The implementation was incomplete.

A new field `effectChanceOverride` was added to `MoveEffectAttr` and to the constructor of `StatStageChangeAttr` which can override the move's secondary effect chance for that attr.

## How to test the changes?
`npm run test triple_arrows`

## Checklist


frutescens (approved):

src/data/move.ts
```diff
@@ -9290,9 +9293,8 @@ export function initMoves() {
     new AttackMove(Moves.TRIPLE_ARROWS, Type.FIGHTING, MoveCategory.PHYSICAL, 90, 100, 10, 30, 0, 8)
       .makesContact(false)
       .attr(HighCritAttr)
-      .attr(StatStageChangeAttr, [ Stat.DEF ], -1)
-      .attr(FlinchAttr)
-      .partial(),
+      .attr(StatStageChangeAttr, [ Stat.DEF ], -1, undefined, undefined, undefined, undefined, undefined, undefined, 50)
```
innerthunder, line 9296: imo it'd be easier to read/interpret this if these args were filled in with the default values, especially since TSDocs aren't normally upfront about default values

src/test/moves/triple_arrows.test.ts
innerthunder, line 1: Can you write another test or two for proof that both effect chances are modified by Serene Grace/Sheer Force? It seems correctly implemented just from looking at the code, but more coverage always helps.

src/data/move.ts
DayKev: I left them as `undefined` to make explicit that they weren't being modified by the move. Also some don't have default parameters and are just optional so I'd have to leave those as `undefined` anyway.
I could add tsdocs to `StatStageChangeAttr` to display the parameters.

image

innerthunder (dismissed):

frutescens (approved):

innerthunder (approved):

torranx (approved):

frutescens (approved):

PR #4545, opened 2024-10-02
[Move] Improve Future Sight & Doom Desire (still partial)

DayKev:
Future Sight and Doom Desire improvements:
- Enable Future Sight / Doom Desire to hit if mon switches out / dies
- Fail if mon is already being targeted by one of these moves
- Enables Future Sight / Doom Desire to hit through protection moves.

Not fixed:
- Cannot be used on multiple Pokémon in doubles, only the first target will be hit.
- ???

The moves were barely implemented.

Added checks/code in `MovePhase` and `MoveEffectPhase` for delayed attacks.

## How to test the changes?
`npm run test future_sight`

## Checklist
  - Update of #1590 


Snailman11:
I think this 
Fixes https://github.com/pagefaultgames/pokerogue/issues/1568
[BUG] Future Sight does not behave like the mainline games

DayKev:
It fixes some of the issues but not all of them. I suppose the original issue could be closed and a new, updated issue could be opened afterwards with the remaining bugs.

frutescens:
re: Cannot be used on multiple Pokémon in doubles, only the first target will be hit.

Could creating a BattlerTag class function onSwitch() stop these current restrictions? Well, it's out of scope.

DayKev:
> re: Cannot be used on multiple Pokémon in doubles, only the first target will be hit.
> 
> Could creating a BattlerTag class function onSwitch() stop these current restrictions? Well, it's out of scope.

I tried swapping to a `BattlerTag` implementation (handling normal switching is fairly easy) but the problem is when a Pokémon faints (and I didn't test with non-normal switching, such as with U-Turn or Roar etc). If there was a way to apply arena tags to a field index instead of the whole side... that's probably what the final solution will need, but that's potentially a lot of refactoring.

frutescens:
Yeah, I ran into a similar issue with Imprison.

src/data/move.ts
```diff
@@ -2670,13 +2678,18 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
   }
 
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): Promise<boolean> {
+    // Edge case for the move applied on a pokemon that has fainted
+    if (!target) {
+      return new Promise(resolve => resolve(true));
+    }
+    const side = target.isPlayer() ? ArenaTagSide.PLAYER: ArenaTagSide.ENEMY;
```
DayKev, line 2685:
```suggestion
    const side = target.isPlayer() ? ArenaTagSide.PLAYER : ArenaTagSide.ENEMY;
```
Huh? How did eslint not catch this before?

src/test/moves/future_sight.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  afterEach(() => {
+    game.phaseInterceptor.restoreOg();
+  });
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override
+      .startingLevel(50)
+      .moveset([Moves.FUTURE_SIGHT, Moves.SPLASH])
```
DayKev, line 26:
```suggestion
      .moveset([ Moves.FUTURE_SIGHT, Moves.SPLASH ])
```

src/test/moves/future_sight.test.ts
```diff
@@ -0,0 +1,45 @@
...
+      .battleType("single")
+      .enemySpecies(Species.MAGIKARP)
+      .enemyAbility(Abilities.STURDY)
+      .enemyMoveset(Moves.SPLASH);
+  });
+
+  it("hits 2 turns after use, ignores user switch out", async () => {
+    await game.classicMode.startBattle([Species.FEEBAS, Species.MILOTIC]);
```
DayKev, line 34:
```suggestion
    await game.classicMode.startBattle([ Species.FEEBAS, Species.MILOTIC ]);
```

src/phases/move-effect-phase.ts
```diff
@@ -66,54 +71,54 @@ export class MoveEffectPhase extends PokemonPhase {
...
-         * @see {@linkcode hitCheck}
-         */
+       * Stores results of hit checks of the invoked move against all targets, organized by battler index.
+       * @see {@linkcode hitCheck}
+       */
       const targetHitChecks = Object.fromEntries(targets.map(p => [ p.getBattlerIndex(), this.hitCheck(p) ]));
       const hasActiveTargets = targets.some(t => t.isActive(true));
 
       /** Check if the target is immune via ability to the attacking move */
       const isImmune = targets[0].hasAbilityWithAttr(TypeImmunityAbAttr) && (targets[0].getAbility()?.getAttrs(TypeImmunityAbAttr)?.[0]?.getImmuneType() === user.getMoveType(move));
```
PigeonBar, line 108:
```suggestion
      const isImmune = targets[0]?.hasAbilityWithAttr(TypeImmunityAbAttr) && (targets[0]?.getAbility()?.getAttrs(TypeImmunityAbAttr)?.[0]?.getImmuneType() === user.getMoveType(move));
```
Taking the opportunity to fix the crash that occurs if the target for Future Sight's hit is missing (e.g., fainted)

PigeonBar:
I think the "Future Sight hits immediately when called by Metronome or Copycat" bug is occurring due to Future Sight using `PokemonMove.virtual` to keep track of whether it is currently the predict turn or the hit turn (see `new PokemonMove(this.sourceMove!, 0, 0, true)`): https://github.com/pagefaultgames/pokerogue/blob/f7797603a1511177b36c57a8c9c7d413c2fb22aa/src/data/arena-tag.ts#L763-L783
Not sure what's the best way to fix this, maybe pass another argument into `MoveEffectPhase` specifically for whether or not it is specifically the hit turn of a delayed attack move?

src/data/move.ts
```diff
@@ -2746,13 +2754,18 @@ export class DelayedAttackAttr extends OverrideMoveEffectAttr {
   }
 
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): Promise<boolean> {
+    // Edge case for the move applied on a pokemon that has fainted
+    if (!target) {
+      return new Promise(resolve => resolve(true));
```
innerthunder, line 2759:
```suggestion
      return Promise.resolve(true)
```

src/data/move.ts
```diff
@@ -8043,7 +8057,8 @@ export function initMoves() {
       .attr(StatStageChangeAttr, [ Stat.SPDEF ], -1)
       .ballBombMove(),
     new AttackMove(Moves.FUTURE_SIGHT, Type.PSYCHIC, MoveCategory.SPECIAL, 120, 100, 10, -1, 0, 2)
-      .partial() // Complete buggy mess
+      .partial()
+      .ignoresProtect()
       .attr(DelayedAttackAttr, ArenaTagType.FUTURE_SIGHT, ChargeAnim.FUTURE_SIGHT_CHARGING, i18next.t("moveTriggers:foresawAnAttack", { pokemonName: "{USER}" })),
```
innerthunder, line 8062:
```suggestion
      .attr(DelayedAttackAttr, ArenaTagType.FUTURE_SIGHT, ChargeAnim.FUTURE_SIGHT_CHARGING, i18next.t("moveTriggers:foresawAnAttack", { pokemonName: "{USER}" }))
      .ignoresProtect()
      .partial(), // cannot be used on multiple Pokemon on the same side in a double battle
```

src/data/move.ts
```diff
@@ -8350,7 +8365,8 @@ export function initMoves() {
       .attr(ConfuseAttr)
       .pulseMove(),
     new AttackMove(Moves.DOOM_DESIRE, Type.STEEL, MoveCategory.SPECIAL, 140, 100, 5, -1, 0, 3)
-      .partial() // Complete buggy mess
+      .partial()
+      .ignoresProtect()
       .attr(DelayedAttackAttr, ArenaTagType.DOOM_DESIRE, ChargeAnim.DOOM_DESIRE_CHARGING, i18next.t("moveTriggers:choseDoomDesireAsDestiny", { pokemonName: "{USER}" })),
```
innerthunder, line 8370:
```suggestion
      .attr(DelayedAttackAttr, ArenaTagType.DOOM_DESIRE, ChargeAnim.DOOM_DESIRE_CHARGING, i18next.t("moveTriggers:choseDoomDesireAsDestiny", { pokemonName: "{USER}" }))
      .ignoresProtect()
      .partial(), // cannot be used on multiple Pokemon on the same side in a double battle
```

innerthunder (approved):

PR #4549, opened 2024-10-02
[Move] Update Tera Starstorm (still Partial), Readd Partial tag to Tera Blast

torranx:

[Tera Starstorm](https://bulbapedia.bulbagarden.net/wiki/Tera_Starstorm_(move)) updated implemention
Why (P)? - both do not ignore abilities that affect stats

partially implemented move, missing the ff when used by Terapagos in its Stellar Form:
- Tera Starstorm becomes [Stellar-type](https://bulbapedia.bulbagarden.net/wiki/Stellar_(type))
- targets both opponents in a [Double Battle](https://bulbapedia.bulbagarden.net/wiki/Double_Battle).

- Changed `TeraBlastCategoryAttr` class name to `TeraMoveCategoryAttr` to align its usecase
- new class `TeraStarstormTypeAttr` to handle type changing logic for the move
- added special targeting for the move
- misc: updated magic numbers to instead use enums for expanding force

## Checklist
- [ ] Are the changes visual?


Xavion3:
Fusions exist, so this will need to account for those.

That and also this is still partial for the same reasons tera blast is, terablastcategoryattr was only a partial implementation when I first implemented, and it doesn't look like anyone has fixed it since.

src/test/moves/tera_starstorm.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    game.move.select(Moves.TERA_STARSTORM, 0, BattlerIndex.ENEMY);
+    game.move.select(Moves.TERA_STARSTORM, 1);
+
+    await game.setTurnOrder([BattlerIndex.PLAYER, BattlerIndex.PLAYER_2, BattlerIndex.ENEMY, BattlerIndex.ENEMY_2]);
+
+    const enemyField = game.scene.getEnemyField();
+
+    // Pokemon other than Terapagos should not be affected - only hits one target
+    await game.phaseInterceptor.to("MoveEndPhase");
+    expect(enemyField.some(pokemon => pokemon.isFullHp())).toBe(true);
```
DayKev, line 64: Why are you not directly checking that the enemy you didn't target is at full hp?

torranx: that's another way to do it, but ultimately it results to the same expectation

src/data/move.ts
```diff
@@ -9617,8 +9641,10 @@ export function initMoves() {
       .attr(ElectroShotChargeAttr)
       .ignoresVirtual(),
     new AttackMove(Moves.TERA_STARSTORM, Type.NORMAL, MoveCategory.SPECIAL, 120, 100, 5, -1, 0, 9)
-      .attr(TeraBlastCategoryAttr)
-      .partial(),
+      .attr(TeraMoveCategoryAttr)
+      .attr(TeraStarstormTypeAttr)
+      .attr(VariableTargetAttr, (user, target, move) => (user.fusionSpecies?.speciesId === Species.TERAPAGOS || user.species.speciesId === Species.TERAPAGOS) && user.isTerastallized() ? MoveTarget.ALL_NEAR_ENEMIES : MoveTarget.NEAR_OTHER)
+      .partial(), // Does not ignore abilities that affect stats
```
DayKev, line 9647:
```suggestion
      .partial(), // TeraMoveCategoryAttr does not ignore abilities that affect stats
```
Also should put it on Tera Blast too.

DayKev (approved):

src/data/move.ts
torranx: this is already planned to be in a separate pr 👍

DayKev: Tbh it should just be in this PR, they're related moves and you're already changing something in Expanding Force too.

torranx: updated

DayKev (dismissed):

src/data/move.ts
```diff
@@ -3901,7 +3901,7 @@ export class PhotonGeyserCategoryAttr extends VariableMoveCategoryAttr {
   }
 }
 
-export class TeraBlastCategoryAttr extends VariableMoveCategoryAttr {
+export class TeraMoveCategoryAttr extends VariableMoveCategoryAttr {
```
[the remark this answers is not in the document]
torranx: decided to stick with the `TeraMoveCategoryAttr` since this is inlined with the other attributes + `TeraMoveUseHigherAttackStat` is confusing to me, nevertheless, I've added docs

Tempo-anon (approved):

PR #4555, opened 2024-10-03
[Balance] Add Memory Mushroom to Shop

innerthunder:
- Memory Mushrooms are now purchasable in the shop after Wave 80. They currently cost 4x the base item cost (equivalent to 2 Revives). When the player selects Memory Mushroom from the shop, the purchase isn't made until the player finalizes the choice to learn a move. After using a Memory Mushroom from the shop, the player is returned to the rewards screen with the same choices as before the purchase was made.
- When the player chooses to use a TM on a Pokemon but cancels the learning of the move, the player is returned to the rewards screen with the same reward options

Requested by Balance Team

- `modifier/modifier-type`: Added Memory Mushroom to the shop item pool in the third tier (Wave 51+)
- `modifier/modifier`: `RememberMoveModifier` (i.e. Memory Mushroom) now takes a `cost` argument. If `cost` is defined, the player spends that much money if the move is learned.
- `phases/select-modifier-phase`: Now queues a copy of itself in the `phaseQueue` when a TM or Memory Mushroom is selected. This allows the player to return to the rewards screen if they decide against learning a move during the `LearnMovePhase`. The cost of these items (when applicable) is also forwarded to `LearnMovePhase`, so the player doesn't spend money unless they actually learn the move.
- `phases/learn-move-phase`: The `fromTM` parameter is now expanded to a `LearnMoveType` enum with 3 possible values:
  - `LEARN_MOVE`: For default behavior, moves learned from level-up, evolution, and other non-item sources.
  - `TM`: For moves obtained from TM rewards during the `SelectModifierPhase`.
  - `MEMORY`: For moves obtained from Memory Mushroom, either as a reward or shop purchase. This type also uses a `cost` parameter to deduct money from the player when applicable.
  The `learnMove` function has been updated in support of the new `LearnMoveType`s.

### Screenshots/Videos

**Buying a Memory Mushroom + Cancelling a TM**

https://github.com/user-attachments/assets/c4c89e5e-3820-4322-97b0-b61f81ae353c

## How to test the changes?
Start a run in Classic (or Endless, etc.). Starting at Wave 81, Memory Mushroom should appear in the shop. Try to interact with TMs and Memory Mushroom as they appear as rewards. When you use a TM/Memory Mushroom on a Pokemon but then decline replacing a move for it, you should return to the shop/rewards screen with the same options. Doing the same with a Memory Mushroom from the shop menu should not result in lost money.

## Checklist
- [x] Are the changes visual?


Madmadness65:
Ideally the Max Ether should be moved to the upper row of items, that way there's an even 7 items on each row.

image

innerthunder:
> Ideally the Max Ether should be moved to the upper row of items, that way there's an even 7 items on each row.

I tried setting the row limit to 7 and now it's off-center...

image

PigeonBar:
Re-running `SelectModifierPhase` after buying/cancelling a Memory Mushroom causes the game to call RNG, which can influence the items you get from rerolls. In this clip, rerolling normally gives me Poke Balls + TM Dream Eater + X Attack, but rerolling after cancelling a Memory Mushroom gives me X Attack + Lansat Berry + Poke Balls.

https://github.com/user-attachments/assets/0d4cd4fd-aabf-42b2-ae4a-2cd2e722f669

DayKev:
> Re-running `SelectModifierPhase` after buying/cancelling a Memory Mushroom causes the game to call RNG, which can influence the items you get from rerolls. In this clip, rerolling normally gives me Poke Balls + TM Dream Eater + X Attack, but rerolling after cancelling a Memory Mushroom gives me X Attack + Lansat Berry + Poke Balls.

That's probably because of this code:
https://github.com/pagefaultgames/pokerogue/blob/a259ccfc34fbca7b9fdc7dc9f5eee5172791b091/src/phases/select-modifier-phase.ts#L31-L32

There probably needs to be an extra condition after `!this.rerollCount` that checks if it's a re-run SMP and if so don't run `this.updateSeed()`.

PigeonBar:
After some more investigation, it turns out that `regenerateModifierPoolThresholds` also calls RNG because it randomly generates items for some reason.

DayKev:
That can be wrapped around a `if (!this.isCopy) {}` as well then.

PigeonBar:
For ME options where item rewards and rerolls are meant to be disabled, buying/cancelling a Memory Mushroom will re-enable rerolling for item rewards:

https://github.com/user-attachments/assets/0e11ab09-5905-489c-9deb-f09025fa7a38

(This can probably be fixed by passing `customModifierSettings.rerollMultiplier` to the `copy()` of the SMPhase.)

innerthunder:
Both of the above issues should be fixed now!

DayKev (approved):

Madmadness65:
> I tried setting the row limit to 7 and now it's off-center...
> 
> EDIT: it also seems like the cursor doesn't align with the shop icons furretthonk

This can be fixed by also editing this line in `modifier-select-ui-handler` as well as the shop options row limit const:
https://github.com/pagefaultgames/pokerogue/blob/0a4c12387b5992e1f14f53a0f8e6b493a0390ad0/src/ui/modifier-select-ui-handler.ts#L214
Changing `(rowOptions.length + 2)` to `(rowOptions.length + 0.75)` visually fixes the shop items being offset.

innerthunder:
> This can be fixed by also editing this line in `modifier-select-ui-handler` as well as the shop options row limit const:
> 
> https://github.com/pagefaultgames/pokerogue/blob/0a4c12387b5992e1f14f53a0f8e6b493a0390ad0/src/ui/modifier-select-ui-handler.ts#L214
> 
> Changing `(rowOptions.length + 2)` to `(rowOptions.length + 0.75)` visually fixes the shop items being offset, at least when the shop options are fully populated. You'll have to tinker with the values on the lines under it to get it to a reasonable position visually for when the shop has fewer items.

This whole formula for `sliceWidth` is so confusing to me... What's it even trying to accomplish? I'm guessing it's trying to get the width of a slice when dividing the screen's width into `rowOptions.length + 2` sections? That's what the x-assignment for the shop options seems to suggest, at least...

damocleas (approved):

DayKev (approved):

Blitz425 (approved):

DayKev:
Oh for some reason I saw an approval by Blitzy and thought "yeah he's a jdev seems good to me" :pikastare:

PR #4558, opened 2024-10-03
[P3] Fix start button cursor not being cleared properly in starter select

MokaStitcher:
Fix a bug in Starter select where sometimes there was a cursor on a Pokemon and on the Start button at the same time

[Report on Discord](https://discord.com/channels/1125469663833370665/1267339517371875361/1291263337404960844) courtesy of Snailman

The issue was that when starting a game from the Start button the cursor would not get cleared, so when coming back to starter select later it would still be here, leading the game to interpret inputs as if the start button was selected
- in `popStarter`, make sure that the Start button cursor gets cleared once the team is empty, so it's no longer possible to be on the button if the team is empty. Since `popStarter` is called in `handler.clear()`, the start cursor now always gets cleared when leaving starter UI
- Took the opportunity to rework navigation between the filters, the party and the start button:
  - Before: moving up/down from the last filter (Sort) would bring you to the list of filtered Pokemon. Now, as long as there is at least one Pokemon in the team, it takes you there
  - Moving down from the Start button now brings you to the filters instead of the first Pokemon in the team

### Screenshots/Videos
Before:

https://github.com/user-attachments/assets/9fa92b64-bbd8-4c33-a7fb-afd89ffbecbf

After:

https://github.com/user-attachments/assets/9361bc33-e63b-47ca-816c-a6feb1b4d7ea

## How to test the changes?
Cursor reset, variant 1:
- New Game
- Add a Pokemon to the team
- Navigate to the start button
- Start the run and cancel once in the save slot selection
- New Game. The cursor should no longer be on the start button

Cursor reset, variant 2:
- New Game
- Add Pokemon(s) to the team
- Navigate to the start button
- Press X multiple times to remove Pokemon from the team. The cursor should not stay on the start button when the team is empty

Navigation
- New Game
- Move your cursor around in various states of the party (empty or not), and filters (empty or not)

## Checklist
- [X] Are the changes visual?


Opaque02 (dismissed):
LGTM! Nice cleanup of code!

DayKev (dismissed):

PR #4562, opened 2024-10-03
[P3] Fix persisting sleep animation when sprite is already loaded

### LINKED ISSUE #4465: [Bug] Save + Quitting does not reset sleeping mons' animation speed
### Describe the bug

--_F5 does reset the speed properly, but if a pokémon gets put to sleep, which lowers the pokémon's animation speed, then you save+quit and return to the battle, the pokémon will start the battle at the same low animation speed whether or not it is currently asleep._
_This compounds every time, which may result in ridiculously low animation speeds if repeated enough time._

--_In my experience this effect persists between waves too. If you KO a mon while asleep, the next mon that appears in the slot will have slower animations_

-- _Yup, this is not reset between waves, forgot to mention that in the post. Catching the mon also doesn't reset it, so chaining sleep+capture repeatedly does the same thing_
_Note that this is not overall slowdown, just the specific mon's animation getting slower and slower, nothing to do with actual lag, especially since other animations play perfectly fine._

https://discord.com/channels/1125469663833370665/1285351284924485713

### Session export file

_No response_

### User data export file

_No response_

### Expected behavior

Pokemon animation speed should reset when quitting the game, or if the pokemon exist the field through KO or being caught.

### Screenshots / Videos

_No response_

acelynnzhang:
Pokémon animation speeds will function normally after reloading a session that was saved when an enemy Pokémon was asleep.

Fixes #4465

Calling PokemonSpeciesForm.loadAssets will reset the framerate when previously nothing would be done if the sprite was already loaded.

### Screenshots/Videos

Before:

https://github.com/user-attachments/assets/a1380c69-f779-4d9f-bd46-35985d3ddea7

After:

https://github.com/user-attachments/assets/823e126c-8a86-4991-a884-b6e360c5b882

## How to test the changes?

- Import session using the session data [linked here](https://discord.com/channels/1125469663833370665/1285351284924485713/1289283674986254457)
- Use Spore
- Save + quit
- Load session again

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/pokemon-species.ts
+++ b/src/data/pokemon-species.ts
@@ -484,6 +484,8 @@ export abstract class PokemonSpeciesForm {
             frameRate: 12,
             repeat: -1
           });
+        } else {
+          scene.anims.get(spriteKey).frameRate = 12;
         }
         let spritePath = this.getSpriteAtlasPath(female, formIndex, shiny, variant).replace("variant/", "").replace(/_[1-3]$/, "");
         const useExpSprite = scene.experimentalSprites && scene.hasExpSprite(spriteKey);


chaosgrimmon:
Seems like a straightforward response to `Quit to menu` recycling the scene?
I very much do dislike the fact that framerates are fixed for the entire species though.

pom-eranian (approved):
Confirmed, fps is reset.  [video](https://github.com/user-attachments/assets/2879f007-367f-4dde-8094-050b551a77ca)

acelynnzhang:
> I very much do dislike the fact that framerates are fixed for the entire species though.

Are there any cases where the framerate is not 12? I referenced [this code](https://github.com/pagefaultgames/pokerogue/blob/160b730921d8d58b925ae0c2c5d353d5ee3b4505/src/field/pokemon.ts#L3441 ) when getting rid of the effects of sleep.

chaosgrimmon:
@acelynnzhang 

Also, if you happen to know which part of the code causes the framerate reduction, it'd save me a fair bit of time looking for it.

acelynnzhang:
> Also, if you happen to know which part of the code causes the framerate reduction, it'd save me a fair bit of time looking for it.
[Yep!](https://github.com/pagefaultgames/pokerogue/blob/160b730921d8d58b925ae0c2c5d353d5ee3b4505/src/field/pokemon.ts#L3398)

chaosgrimmon:
> Are there any cases where the framerate is not 12?

Great question, and I suppose the easy answer is previously, _when it was reduced by a member of the Species being Asleep_.

(Since the code is in _`pokemon-species`_, it affects every single future instantiation of that Species, provided that the scene was still in use. Because this doesn't affect existing instances, it did not appear to cause problems if, say, you Slept only one enemy Pokemon in a double battle and they were the same Species, but upon Quit+Reload, both would have their assets loaded, causing _both_ to be slower than intended)

https://github.com/user-attachments/assets/32285868-cc03-4ce0-98ef-d79a1a17b3f9

https://github.com/user-attachments/assets/6613fef5-b669-4a8e-b59f-d4a46a600d38

chaosgrimmon (approved):
tested, lgtm

https://github.com/user-attachments/assets/47a441c3-f699-4c4f-a194-4fe490fbefc1

chaosgrimmon:
> [Yep!](https://github.com/pagefaultgames/pokerogue/blob/160b730921d8d58b925ae0c2c5d353d5ee3b4505/src/field/pokemon.ts#L3398)

That's really interesting given the statement in the Issue, if the actual code should only be setting framerate to a flat `4`.
> This compounds every time, which may result in ridiculously low animation speeds if repeated enough time.

acelynnzhang:
> (the reason I had to ask is in fact because it seemed like it wasn't getting slower upon repeated attempts)

I was also unable to see the compounding in my testing. I used DevTools to track the framerate changes.

PR #4564, opened 2024-10-03
[Beta][Test] Fix Scale Shot flaky test

Fontbane:

- Scale Shot unit test was incorrectly setting the player species instead of opponent species leading to undesired randomness
- Test would fail when a Fairy-type opponent generated

- Overrides opponent species correctly

## How to test the changes?

- `npm run test scale_shot`

## Checklist
- [ ] Are the changes visual?


src/test/moves/scale_shot.test.ts
```diff
@@ -36,11 +36,12 @@ describe("Moves - Scale Shot", () => {
...
       .enemyMoveset(Moves.SKILL_SWAP)
-      .enemyLevel(5);
+      .enemyLevel(3);
   });
 
   it("applies stat changes after last hit", async () => {
-    await game.classicMode.startBattle([Species.FORRETRESS]);
+    game.override.enemySpecies(Species.FORRETRESS);
+    await game.classicMode.startBattle();
     const minccino = game.scene.getPlayerPokemon()!;
```
torranx, line 45: this is not guaranteed to spawn a minccino but a random mon instead since it is not specified in `startBattle()`

src/test/moves/scale_shot.test.ts
```diff
@@ -36,11 +36,12 @@ describe("Moves - Scale Shot", () => {
...
     game.move.select(Moves.SCALE_SHOT);
     await game.phaseInterceptor.to(MoveEffectPhase);
```
torranx, line 47: this is another cause of a flaky test since it is not guaranteed which pokemon moves first.
see `setTurnOrder`

src/test/moves/scale_shot.test.ts
```diff
@@ -54,7 +55,8 @@ describe("Moves - Scale Shot", () => {
   });
 
   it("unaffected by sheer force", async () => {
-    await game.classicMode.startBattle([Species.WOBBUFFET]);
+    game.override.enemySpecies(Species.WOBBUFFET);
+    await game.classicMode.startBattle();
     const minccino = game.scene.getPlayerPokemon()!;
```
torranx, line 60: same concern above

src/test/moves/scale_shot.test.ts
```diff
@@ -54,7 +55,8 @@ describe("Moves - Scale Shot", () => {
...
     const wobbuffet = game.scene.getEnemyPokemon()!;
     wobbuffet.setStat(Stat.HP, 100, true);
```
torranx, line 62: this should not write to the actual stats

src/test/moves/scale_shot.test.ts
```diff
@@ -54,7 +55,8 @@ describe("Moves - Scale Shot", () => {
   });
 
   it("unaffected by sheer force", async () => {
```
torranx, line 57: wouldnt this be simplified by spying on `calculateBattlePower`?
also shouldn't the scale shot user be the one having Sheer Force? its set to the enemy currently

src/test/moves/scale_shot.test.ts
```diff
@@ -36,11 +36,12 @@ describe("Moves - Scale Shot", () => {
       .enemyAbility(Abilities.SHEER_FORCE)
```
torranx, line 36: this should be inside the specific test case

src/test/moves/scale_shot.test.ts
```diff
@@ -36,11 +36,12 @@ describe("Moves - Scale Shot", () => {
...
       .enemyPassiveAbility(Abilities.STALL)
```
torranx, line 37: unnecessary passive abilities are set

src/test/moves/scale_shot.test.ts
```diff
@@ -36,11 +36,12 @@ describe("Moves - Scale Shot", () => {
...
       .enemyMoveset(Moves.SKILL_SWAP)
```
torranx, line 38: its generally best to use SPLASH to avoid any interaction that may result in a different result

torranx (changes requested):
could do some line breaks

src/test/moves/scale_shot.test.ts
Fontbane: Forretress uses Mirror Coat which has negative priority

src/test/moves/scale_shot.test.ts
Fontbane: Wobbuffet has Stall

src/test/moves/scale_shot.test.ts
Fontbane: Turn one you use it normally, then Wobbuffet swaps Sheer Force onto you, then you use it with Sheer Force

src/test/moves/scale_shot.test.ts
Fontbane: This guarantees Minccino moves first

src/test/moves/scale_shot.test.ts
Fontbane: This is done to give Minccino Sheer Force for the second turn

src/test/moves/scale_shot.test.ts
torranx: I guess that could work but then again a helper function exists for this which does rely on any abilities

src/test/moves/scale_shot.test.ts
torranx: what's the need to use it normally on turn 1 when the test should only be testing that it doesn't affect sheer force? that is just unnecessary complication imo

src/test/moves/scale_shot.test.ts
torranx: see [comment here](https://github.com/pagefaultgames/pokerogue/pull/4564/files#r1786778827).
also if an override is for a specific test case, that should be placed there

src/test/moves/scale_shot.test.ts
torranx: sorry concern being this is not guaranteed to spawn a mincinno

src/test/moves/scale_shot.test.ts
Fontbane: Integrated your suggestions, thanks

src/test/moves/scale_shot.test.ts
```diff
@@ -30,45 +32,54 @@ describe("Moves - Scale Shot", () => {
...
+    game.override.enemySpecies(Species.FORRETRESS);
+
+    await game.classicMode.startBattle([ Species.MINCCINO ]);
     const minccino = game.scene.getPlayerPokemon()!;
     game.move.select(Moves.SCALE_SHOT);
+
+    await game.setTurnOrder([ BattlerIndex.PLAYER, BattlerIndex.ENEMY ]);
+
     await game.phaseInterceptor.to(MoveEffectPhase);
     await game.phaseInterceptor.to(DamagePhase);
```
torranx, line 51: I don't think these are needed?

torranx (commented):
looks a lot better, except for 1 nit

src/test/moves/scale_shot.test.ts
Fontbane: I believe the effect, if activated on the first hit, would apply on that first MoveEffectPhase so I just wait for the second hit. It's gonna go through all 5 anyway.

torranx (approved):

ben-lear (approved):

src/test/moves/scale_shot.test.ts
innerthunder: There's no need for the `DamagePhase` line if that's the case.

DayKev (approved):

PR #3974, opened 2024-09-02
[Refactor] Refactor move phase and add documentation

DayKev:
Text mostly copied from #3064

## What are the changes?
~~Nothing functional, at least not intentionally.~~  Added documentation and cleaned up `MovePhase`.
Update: A bug introduced by the `strict-null` refactor (see [here](https://github.com/pagefaultgames/pokerogue/pull/3259/files#diff-7bfd86f885f7bd68a0b3b3a290cbfe482a6338ac569b6a7418716a863006537aR2657-R2658)) is fixed (multi-hit moves called by Metronome/etc only hit once).

## Why am I doing these changes?
It's extremely difficult to follow in its current state.  Adding documentation and cleaning up the code here ~~(and soon `MoveEffectPhase`)~~ will pay off in the short term, and will ultimately make it easier to game-plan a larger refactor.

## What did change?
Large sections of code were split out into methods and variables were renamed for readability; quite a few comments were added in places where I felt they'd be helpful.  A few things were shuffled around, but only in non-functional ways (e.g. moving a variable declaration closer to where it ends up being used).

## Checklist
- [x] **I'm using `beta` as my base branch**
- [ ] There is no overlap with another PR?
  - Update of #3064
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- ~[ ] Have I considered writing automated tests for the issue?~
- ~[ ] If I have text, did I make it translatable and add a key in the English locale file(s)?~
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- ~[ ] Are the changes visual?~
  - ~[ ] Have I provided screenshots/videos of the changes?~

PR #4576, opened 2024-10-04
[Bug] Metal Burst causes game to crash if intended target faints

seanrmon:

Metal burst will no longer cause game to crash if there are no valid targets.

I believe this was caused by [Refactor] Refactor move phase and add documentation (#3974), which is currently in use on the beta branch.

In `resolveCounterAttackTarget()`, if `attacker` is null then `this.targets[0]` does not get set and `this.scene.getField()[this.targets[0]].hp` will lead to a memory access violation.

### Screenshots/Videos

image

https://github.com/user-attachments/assets/2f3fd6ca-d992-48bc-b7b3-80aed08c1918

## How to test the changes?

[sessionData_metalburstcrash.txt](https://github.com/user-attachments/files/17261538/sessionData_metalburstcrash.txt)

Use Water Spout and Metal Burst. See video.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/phases/move-phase.ts
+++ b/src/phases/move-phase.ts
@@ -376,7 +376,7 @@ export class MovePhase extends BattlePhase {
         // account for metal burst and comeuppance hitting remaining targets in double battles
         // counterattack will redirect to remaining ally if original attacker faints
         if (this.scene.currentBattle.double && this.move.getMove().hasFlag(MoveFlags.REDIRECT_COUNTER)) {
-          if (this.scene.getField()[this.targets[0]].hp === 0) {
+          if (attacker?.isActive(true) && this.scene.getField()[this.targets[0]].hp === 0) {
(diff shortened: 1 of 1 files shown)


DayKev:
The solution that fixed this last time was
https://github.com/pagefaultgames/pokerogue/blob/f562a76332fcf76601ed215b48b11e2b484efaef/src/phases/move-phase.ts#L378-L382
being changed to `this.targets[0] = this.pokemon.turnData.attacksReceived[0].sourceBattlerIndex;` it seems (or maybe that change was made independently, I'm not 100% sure; the git history for this section got clobbered :regiDespair:).
Putting that code back fixes the problem (Metal Burst redirects instead of crashing), whereas yours stops the crash but causes Metal Burst to fail instead of redirecting.

DayKev:
I'll re-add the code that got reverted and add some tests to prevent this happening again.

DayKev:
Btw sorry to "steal" the PR like this, but since it's a crash issue I wanted to get it fixed w/tests ASAP.

PR #4583, opened 2024-10-04
[Refactor] Pokerogue API client

flx-sta:

n/a

Well you know how it is... the way it was designed in the code was bugging @Opaque02  & me.  I thought I fast and easy fix it. One thing leads to another and here we are..

> [!WARNING]
> **I couldn't check the discord/google unlink requests due to local limitations.**
> 
> *Yes, I did set up my own Discord Dev Application to link it to but it simply refused to save those discord IDs into the database. No idea why*
> image

All `Utils.apiFetch` and `Utils.apiPost` were replaced with a new `PokerogueApi` class which is structured into sub-apis

Every request & response from the API is properly typed now

### Screenshots/Videos

n/a

## How to test the changes?

1. Make sure you have the [rogueserver repo](https://github.com/pagefaultgames/rogueserver/) cloned
2. Start the local server
3. Make sure to set `VITE_BYPASS_LOGIN` to `0` in `env.development`
4. Start the client
5. Tests all the things that cause requests 

## Checklist
- [x] Have I considered writing automated tests for the issue? => there is some already


src/plugins/api/api-base.ts
```diff
@@ -0,0 +1,89 @@
...
+    let body: string | undefined = undefined;
+    const headers: HeadersInit = {};
+
+    if (bodyData) {
+      if (dataType === "json") {
+        body = typeof bodyData === "string" ? bodyData : JSON.stringify(bodyData);
+        headers["Content-Type"] = "application/json";
+      } else if (dataType === "form-urlencoded") {
+        if (bodyData instanceof Object) {
+          body = new URLSearchParams(Object.entries<any>(bodyData).map(([ k, v ]) => [ k, v.toString() ])).toString();
```
Opaque02, line 43: Seems to be something not quite working with this line for the discord linking. For context, my server was set up for the admin panel stuff, which I've repurposed for this PR. With my admin stuff PR I've got, I can use the admin link functionality and it updates the DB with the new discord id (it can be a fake discord id - you won't be able to log in but you can at least see the db be updated), but it seems like in this one it doesn't?

It seems like when you go to use admin-link here, the Object.entries(bodyData) is returning null, which then makes the body be null as well. Here's some screenshots of it in action:

Here I'm trying to make my username "Rand" have a discord id of 4
image

And here is the Object.entries for bodyData:
image

Another thing for it is within the debugger itself - here is the line where body is created and we see the values for body data:
image

Yet the next line has the body with a value of ""
image

For reference, this is my db value
image

And, using postman, I can update the value to something else:
image
image

flx-sta: Thanks for the detailed feedback.
<sub>I'm actually using something like Postman too called [Thunder Client](https://www.thunderclient.com/)</sub>

flx-sta: https://github.com/pagefaultgames/pokerogue/pull/4583/commits/30b6d4e6ffd92365abaf34cae9d0b1de7d7962f2
Fixed ☑️

Opaque02 (approved):
Looks good as far as I can tell - can't test some of the discord stuff but so far all looks good otherwise

src/system/game-data.ts
```diff
@@ -1383,7 +1378,15 @@ export class GameData {
...
+        let promise: Promise<any> = Promise.resolve(null);
+
+        if (dataType === GameDataType.SYSTEM) {
+          promise = pokerogueApi.savedata.system.get({ clientSessionId });
+        } else if (dataType === GameDataType.SESSION) {
+          promise = pokerogueApi.savedata.session.get({ slot: slotId, clientSessionId });
+        }
+
+        promise
           .then(response => response.text())
```
MokaStitcher, line 1390:
```suggestion
```

needs to be removed since `response.text()` is already called in `pokerogueApi.savedata.system.get`

no energy to review more of it right now sorry 😴

src/ui/admin-ui-handler.ts
```diff
@@ -148,6 +150,16 @@ export default class AdminUiHandler extends FormModalUiHandler {
...
+        const onFail = (error: string) => {
+          this.scene.ui.setMode(Mode.ADMIN, Object.assign(this.config, { errorMessage: error?.trim() }));
+          this.scene.ui.playError();
+        };
+        if (!this.inputs[0].text) {
+          return onFail("Username is required");
+        }
+        if (!this.inputs[1].text) {
+          return onFail("Discord Id is required");
+        }
```
Opaque02, line 162: So, the code for the admin panel does work correctly, which is good, however there's currently a console error that wasn't coming up before as far as I can tell: 
image

I've done some tracking and I think it's related to these lines here. Firstly, we replaced `const onFail` with `this.showMessage()`, which lets us show errors and success messages instead of just fail errors. The console error comes from it getting to line 157 (`if (!this.inputs[0].text)`); the `.text` is actually reading the property of `this.inputs[0].node.text`, but by this point the inputs have been destroyed, so `this.inputs[0].node` doesn't exist, giving us the error we see above.

I'm not 100% sure but I suspect this might be from a bad merge - deleting these lines seemed to fix in my local testing so far

src/ui/admin-ui-handler.ts
```diff
@@ -276,25 +288,60 @@ export default class AdminUiHandler extends FormModalUiHandler {
 
   private async adminSearch(adminSearchResult: AdminSearchInfo) {
     try {
-      const adminInfo = await Utils.apiFetch(`admin/account/adminSearch?username=${encodeURIComponent(adminSearchResult.username)}`, true);
-      if (!adminInfo.ok) { // error - if adminInfo.status === this.httpUserNotFoundErrorCode that means the username can't be found in the db
-        return { adminSearchResult: adminSearchResult, error: true, errorType: adminInfo.status === this.httpUserNotFoundErrorCode ? this.ERR_USERNAME_NOT_FOUND : this.ERR_GENERIC_ERROR };
+      // const adminInfo = await Utils.apiFetch(`admin/account/adminSearch?username=${encodeURIComponent(adminSearchResult.username)}`, true);
```
Opaque02, line 291: Debug line

src/ui/admin-ui-handler.ts
flx-sta: Yeah this is probably a leftover of a merge mistake.
If this is obsolete now I might aswell remove it completely.
Thanks for letting me know

src/plugins/api/pokerogue-session-savedata-api.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+   * Get a session savedata.
+   * @param params The {@linkcode GetSessionSavedataRequest} to send
+   * @returns The session as `string`
+   */
+  public async get(params: GetSessionSavedataRequest) {
+    try {
+      const urlSearchParams = this.toUrlSearchParams(params);
+      const response = await this.doGet(`/savedata/session/get?${urlSearchParams}`);
```
Opaque02, line 45: When I go to open the save session screen, it seems like there's a console error with this when a slot is empty:

image

Here's a screenshot of my response: 

image

flx-sta: This was always the case, wasn't it?

flx-sta: Yeah, same on live
image

PigeonBar:
These are now unused after the refactor, should we delete them or mark them as `@deprecated`?
https://github.com/pagefaultgames/pokerogue/blob/684fb93e398a7eaca367177893877b66ec900990/src/utils.ts#L273-L276

flx-sta:
> These are now unused after the refactor, should we delete them or mark them as `@deprecated`?
> 
> https://github.com/pagefaultgames/pokerogue/blob/684fb93e398a7eaca367177893877b66ec900990/src/utils.ts#L273-L276

@deprecated sounds good as a first step

src/system/game-data.ts
```diff
@@ -984,10 +983,9 @@ export class GameData {
       };
 
       if (!bypassLogin && !localStorage.getItem(`sessionData${slotId ? slotId : ""}_${loggedInUser?.username}`)) {
-        Utils.apiFetch(`savedata/session/get?slot=${slotId}&clientSessionId=${clientSessionId}`, true)
-          .then(response => response.text())
+        pokerogueApi.savedata.session.get({ slot: slotId, clientSessionId })
           .then(async response => {
-            if (!response.length || response[0] !== "{") {
+            if (!response && response?.length === 0 || response?.[0] !== "{") {
```
PigeonBar, line 988:
```suggestion
            if (!response || response?.length === 0 || response?.[0] !== "{") {
```
Is `||` intended here instead of `&&`?

PigeonBar:
The Discord and Google unlink requests should be partially testable by linking using the admin panel, then unlinking using the admin panel or by using Menu > Manage Data > Unlink.

PigeonBar:
I've done some testing and haven't found any issues so far

src/test/mystery-encounter/encounters/clowning-around-encounter.test.ts
```diff
@@ -247,7 +249,7 @@ describe("Clowning Around - Mystery Encounter", () => {
       });
     });
 
-    it("should randomize held items of the Pokemon with the most items, and not the held items of other pokemon", async () => {
+    it.skipIf(new TimedEventManager().isEventActive())("should randomize held items of the Pokemon with the most items, and not the held items of other pokemon", async () => {
```
PigeonBar, line 252: We no longer need to skip this test since #4775 is merged.

flx-sta:
@PigeonBar Applied all your suggested changes.
Thanks

PR #4585, opened 2024-10-05
[Bug] Trainer Party Generation checks Trainer's Static Pokemon List for Duplicates too

frutescens:
Trainers will generated hopefully with less duplicates in their parties.

Felt like it. 

When a trainer's party is generated, it seems that the random Pokemon of their party is generated first and then the static fixed Pokemon are added to their party. The current duplicate check did not check the static Pokemon involved so it would be possible for a trainer like Olivia to appear with two Carbinks since one Carbink was a static member of the party. 

## How to test the changes?
Generate Trainers. 

## Checklist


CodeTappert:
So as far as i understand this code change. It adds the static pokemon list first to the team and then the randoms now

frutescens:
> So as far as i understand this code change. It adds the static pokemon list first to the team and then the randoms now

No, it checks the static pokemon list when adding random Pokemon. The order of addition is still the same.

DayKev (approved):

innerthunder (approved):

damocleas (approved):

PR #4591, opened 2024-10-05
[Enhancement] Refactor Starter Species to use separate EggTier map

ben-lear:

No functional changes.

Creates a new Species `EggTier` map that can be used by Balance team to set `EggTier` separately from a species' starter cost. Allows for situations like Regice, where it is still obtainable from Epic eggs as a sub-legendary but has starter cost of 5, etc.

Not much changes, except an egg calculates its `EggTier` from the `speciesEggTier` map now instead of from `speciesStarterCosts`.

Also removed a redundant math clamping function in `utils.ts`

## How to test the changes?
Eggs are hatching with normal expected species. There are no balance changes in this PR.

## Checklist
- [X] Are the changes visual?


damocleas (approved):

src/utils.ts
```diff
@@ -38,7 +38,13 @@ export function shiftCharCodes(str: string, shiftCount: integer) {
...
 
-export function clampInt(value: integer, min: integer, max: integer): integer {
+/**
+ * Clamps a number
+ * @param value
+ * @param min
+ * @param max
+ */
+export function clampNumber(value: number, min: number, max: number): number {
   return Math.min(Math.max(value, min), max);
```
DayKev, line 48: Tbh this function is unnecessary, there's already `Phaser.Math.Clamp(value, min, max)`, just use that.
```suggestion
  return Phaser.Math.Clamp(value, min, max);
```

src/utils.ts
```diff
@@ -38,7 +38,13 @@ export function shiftCharCodes(str: string, shiftCount: integer) {
   return newStr;
 }
 
-export function clampInt(value: integer, min: integer, max: integer): integer {
+/**
+ * Clamps a number
+ * @param value
+ * @param min
+ * @param max
+ */
```
DayKev, line 46:
```suggestion
```

src/data/balance/species-egg-tiers.ts
```diff
@@ -0,0 +1,603 @@
+import { Species } from "#enums/species";
+import { EggTier } from "#enums/egg-type";
+
+/**
+ * Map of all starters and their respective {@link EggTier}, which determines which type of egg the starter hatches from.
```
DayKev, line 5:
```suggestion
 * Map of all starters and their respective {@linkcode EggTier}, which determines which type of egg the starter hatches from.
```

src/utils.ts
ben-lear: we should just remove entirely then

DayKev (approved):

torranx (approved):
lgtm

damocleas (approved):

PR #4596, opened 2024-10-06
[Misc] Miscellaneous code cleanup

DayKev:
N/A

Just purifying some heretical code. 🙏

- Replaced `.` with `#app` in `overrides.ts` imports.
- Added a missing semicolon and missing explanation comment in `eslint.config.js`.
- `phases/move-effect-phase.ts`:
  - Fix indentation
  - Renamed `getTarget` to `getFirstTarget` for clarity.
  - Converted namespace import (`Utils`) to named imports.
  - Added `public`/etc to class methods.
- `field/pokemon.ts`:
  - Renamed `generateVariant()` to `generateShinyVariant()` for clarity.
  - Marked various methods as `public`/etc.
  - Added tsdocs to `isOfType()`.
  - Updated various tsdocs and comments (and removed an obsolete comment).
  - Cleaned up the monstrosity that was a 450+ line of code (Xavion please...) into something readable.
- `utils.ts`:
  - Removed an unused function `randSeedEasedWeightedItem()`.
  - Marked `IntegerHolder`, `FixedInt`, and `fixedInt()` as deprecated.
  - Updated various tsdocs and comments.

## How to test the changes?
Play the game normally.

## Checklist


src/field/pokemon.ts
```diff
@@ -2163,8 +2175,20 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
         // Sqrt the weight of any damaging moves with overlapping types. This is about a 0.05 - 0.1 multiplier.
         // Other damaging moves 2x weight if 0-1 damaging moves, 0.5x if 2, 0.125x if 3. These weights double if STAB.
         // Status moves remain unchanged on weight, this encourages 1-2
-        movePool = baseWeights.filter(m => !this.moveset.some(mo => m[0] === mo?.moveId)).map(m => [ m[0], this.moveset.some(mo => mo?.getMove().category !== MoveCategory.STATUS && mo?.getMove().type === allMoves[m[0]].type) ? Math.ceil(Math.sqrt(m[1])) : allMoves[m[0]].category !== MoveCategory.STATUS ? Math.ceil(m[1] / Math.max(Math.pow(4, this.moveset.filter(mo => (mo?.getMove().power!) > 1).length) / 8, 0.5) * (this.isOfType(allMoves[m[0]].type) ? 2 : 1)) : m[1] ]); // TODO: is this bang correct?
-      } else { // Non-trainer pokemon just use normal weights
+        const movesetFilter = (m: [Moves, number]) => {
```
ben-lear, line 2178: nit: can just convert what was previously the inlined anonymous function to a bracketed anonymous function. No need to convert to a variable as it isn't re-used.

DayKev: Good point, I was focused on making it readable and missed I could do it simpler that way.

src/utils.ts
```diff
@@ -474,14 +465,16 @@ export function hslToHex(h: number, s: number, l: number): string {
...
-If the lang is not in the function, it usually means that lang is going to use the default english version
-This function is used in:
-- summary-ui-handler.ts: If the lang is not available, it'll use types.json (english)
-English itself counts as not available
+/**
+ * This function returns `true` if the current lang is available for some functions.
+ * If the lang is not in the function, it usually means that lang is going to use the default english version
+ *
+ * This function is used in:
+ * - `summary-ui-handler.ts`: If the lang is not available, it'll use `types.json` (english).
```
MokaStitcher, line 473: putting aside the fact that this does not seems like an ideal way to handle localized images at all, it seems like this is out of date and it's only used when loading the atlas images for types and statuses and in 

https://github.com/pagefaultgames/pokerogue/blob/5d0b36132061bae767dafdd3f27c6c1be12264df/src/utils.ts#L616-L618

that is called whenever said type or status icons are needed

To be more true to what it actually does maybe it should be something more like this:
```
This function returns `true` if all localized images used by the game have been added for the given language
```
and rename it to `hasAllLocalizedSprites` or something to that effect

MokaStitcher (commented):
~~that move weight calculation killed me~~

Just have issues with `Utils.verifyLang` otherwise lgtm

MokaStitcher (dismissed):
lgtm

PR #4599, opened 2024-10-06
[P3] Fix input blocking on repeated MEs without refresh

ben-lear:
If a user encounters multiple MEs without refreshing their session, the second+ ME will have initial 1 second input blocking as intended before the player can select an option.

Previously input blocking was not working as intended.

Reset the input blocking variable when the ME UI is cleared.

### Screenshots/Videos

Before:

chrome_7JCs92tJgz

After:

chrome_KB2gAvwhrI

Opening View Party does NOT re-enable input blocking

chrome_LXNMfUhEaX

# IMPORTANT

Safari Zone will also input block on options now (this may not be desired, up for discussion):

chrome_CYgASELCK4

This does NOT affect Wobbuffet game ME.

## How to test the changes?
`MYSTERY_ENCOUNTER_RATE_OVERRIDE: 256,
  MYSTERY_ENCOUNTER_OVERRIDE: MysteryEncounterType.MYSTERIOUS_CHEST`

## Checklist
- [X] Are the changes visual?


MokaStitcher:
This seems to work, but a bit inconsistently

For example if you access the party menu by selection an option of the ME and back out, the input will be blocked again. https://github.com/user-attachments/assets/7c47d94a-6fd6-4031-8e4f-174e4c226c48

This is not too bad, it's best to block too much than not enough IMO, but this + the Safari Zone issue makes me wonder if we shouldn't look for a better solution. Ideally blocking should only happen once when the ME appears. MEs can potentially do a lot of different things with the UI that would break the intended behavior, like GTS here. So the handler is maybe not the right place to hold onto the `blockInput` attribute 

The other issue I found is a bit more worrying: the fix doesn't seem to work all the time. For example in the GTS ME if you make a trade then the next ME will not have its input blocked. If you leave you're good though.

https://github.com/user-attachments/assets/d94823a9-9994-461a-837b-f490cf00ab5a

Does this mean that somehow the ME handler does not get cleared in this case, and possibly with it the ME Mode is still in the stack? I did not look into it further, but it might be worth checking. Maybe other MEs have similar issues too?

SirzBenjie:
PR is too old; as Moka points out, the solution is brittle and is best done elsewhere.

PR #4606, opened 2024-10-07
[Bug] Syrup Bomb effect is removed when user leaves the field

DayKev:
When the Pokémon that used Syrup Bomb leaves the field, the end-of-turn speed lowering effect will no longer occur.

The current behavior is incorrect.

Added a check to `SyrupBombTag` to see if the source Pokémon is still on the field and remove the tag if it isn't.

## How to test the changes?
`npm run test syrup_bomb`

## Checklist

### DIFF
--- a/src/data/battler-tags.ts
+++ b/src/data/battler-tags.ts
@@ -2636,16 +2636,16 @@ export class ImprisonTag extends MoveRestrictionBattlerTag {
 /**
  * Battler Tag that applies the effects of Syrup Bomb to the target Pokemon.
  * For three turns, starting from the turn of hit, at the end of each turn, the target Pokemon's speed will decrease by 1.
- * The tag can also expire by taking the target Pokemon off the field.
+ * The tag can also expire by taking the target Pokemon off the field, or the Pokemon that originally used the move.
  */
 export class SyrupBombTag extends BattlerTag {
-  constructor() {
-    super(BattlerTagType.SYRUP_BOMB, BattlerTagLapseType.TURN_END, 3, Moves.SYRUP_BOMB);
+  constructor(sourceId: number) {
+    super(BattlerTagType.SYRUP_BOMB, BattlerTagLapseType.TURN_END, 3, Moves.SYRUP_BOMB, sourceId);
   }
 
   /**
    * Adds the Syrup Bomb battler tag to the target Pokemon.
-   * @param {Pokemon} pokemon the target Pokemon
+   * @param pokemon - The target {@linkcode Pokemon}
    */
   override onAdd(pokemon: Pokemon) {
     super.onAdd(pokemon);
@@ -2654,12 +2654,12 @@ export class SyrupBombTag extends BattlerTag {
 
   /**
    * Applies the single-stage speed down to the target Pokemon and decrements the tag's turn count
-   * @param {Pokemon} pokemon the target Pokemon
-   * @param {BattlerTagLapseType} _lapseType
+   * @param pokemon - The target {@linkcode Pokemon}
+   * @param _lapseType - N/A
    * @returns `true` if the turnCount is still greater than 0 | `false` if the turnCount is 0 or the target Pokemon has been removed from the field
    */
   override lapse(pokemon: Pokemon, _lapseType: BattlerTagLapseType): boolean {
-    if (!pokemon.isActive(true)) {
+    if (!pokemon.isActive(true) || (this.sourceId && !pokemon.scene.getPokemonById(this.sourceId)?.isActive(true))) {
       return false;
     }
     pokemon.scene.queueMessage(i18next.t("battlerTags:syrupBombLapse", { pokemonNameWithAffix: getPokemonNameWithAffix(pokemon) })); // Custom message in lieu of an animation in mainline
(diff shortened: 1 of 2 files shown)


src/data/battler-tags.ts
```diff
@@ -2654,12 +2654,12 @@ export class SyrupBombTag extends BattlerTag {
...
    * Applies the single-stage speed down to the target Pokemon and decrements the tag's turn count
-   * @param {Pokemon} pokemon the target Pokemon
-   * @param {BattlerTagLapseType} _lapseType
+   * @param pokemon - The target {@linkcode Pokemon}
+   * @param _lapseType - N/A
    * @returns `true` if the turnCount is still greater than 0 | `false` if the turnCount is 0 or the target Pokemon has been removed from the field
    */
   override lapse(pokemon: Pokemon, _lapseType: BattlerTagLapseType): boolean {
-    if (!pokemon.isActive(true)) {
+    if (!pokemon.isActive(true) || (this.sourceId && !pokemon.scene.getPokemonById(this.sourceId)?.isActive(true))) {
```
innerthunder, line 2662: I don't think the extra condition here is needed. Tags are already removed by source ID when the source switches out:

https://github.com/pagefaultgames/pokerogue/blob/160b730921d8d58b925ae0c2c5d353d5ee3b4505/src/phases/switch-summon-phase.ts#L68-L69

Just adding a `sourceId` to this tag should be enough on its own to implement the desired behavior. You can look at `DamagingTrapTag`'s implementation for reference.

src/data/battler-tags.ts
```diff
@@ -2673,12 +2673,8 @@ export class SyrupBombTag extends BattlerTag {
...
 /**
  * Retrieves a {@linkcode BattlerTag} based on the provided tag type, turn count, source move, and source ID.
- *
- * @param {BattlerTagType} tagType the type of the {@linkcode BattlerTagType}.
- * @param turnCount the turn count.
- * @param {Moves} sourceMove the source {@linkcode Moves}.
- * @param sourceId the source ID.
- * @returns {BattlerTag} the corresponding {@linkcode BattlerTag} object.
+ * @param sourceId - The ID of the pokemon adding the tag
+ * @returns The corresponding {@linkcode BattlerTag} object.
```
innerthunder, line 2677: Seems like some lines were deleted here on accident?

innerthunder (dismissed):

frutescens (approved):
Approved -> hopefully w/ the comments restored

src/data/battler-tags.ts
DayKev: Alright, wasn't sure if the original condition needed to be there so I left it in originally.

src/data/battler-tags.ts
DayKev: I left out the ones that were self-evident. Comments like "`@param sourceMove the source move`" aren't useful (and are potentially even anti-useful, making it take longer to find relevant information). I did forget to move the `@linkcode Moves` part into the description above though, lemme fix that.

DayKev: Oh wait that's the enum, nevermind; that's probably why I didn't move it in the first place.

frutescens (approved):

src/data/battler-tags.ts
```diff
@@ -2654,15 +2654,16 @@ export class SyrupBombTag extends BattlerTag {
...
-   * @param {Pokemon} pokemon the target Pokemon
-   * @param {BattlerTagLapseType} _lapseType
-   * @returns `true` if the turnCount is still greater than 0 | `false` if the turnCount is 0 or the target Pokemon has been removed from the field
+   * @param pokemon - The target {@linkcode Pokemon}
+   * @param _lapseType - N/A
+   * @returns `true` if the `turnCount` is still greater than `0`; `false` if the `turnCount` is `0` or the target or source Pokemon has been removed from the field
    */
   override lapse(pokemon: Pokemon, _lapseType: BattlerTagLapseType): boolean {
-    if (!pokemon.isActive(true)) {
+    if (this.sourceId && !pokemon.scene.getPokemonById(this.sourceId)?.isActive(true)) {
```
MokaStitcher, line 2662: I'm not too familiar with BattlerTags, what was the purpose of the `!pokemon.isActive(true)` condition before?
Based on the comment I would've assumed it checks that the target is still active, in which case this probably shouldn't be removed and the check should be something like `if(!pokemon.isActive(true) || ((this.sourceId && !pokemon.scene.getPokemonById(this.sourceId)?.isActive(true)))` to check for both target and source, not just source?

frutescens: As the original implementer, I can explain.... so I was paranoid that the game would attempt to lower the affected Pokemon's (the original target) stats when the Pokemon was off the field and worried that it could lead to a crash. Kev is more familiar with the nuances of the battle system + summonData/battleSummonData so I trust that the right changes were made :)

innerthunder: When a Pokemon switches out (or otherwise leaves the field), all of its tags are removed via `Pokemon.resetSummonData`, so the original `!pokemon.isActive(true)` condition shouldn't ever be met anyway.

This condition is still kind of redundant, too. When a Pokemon leaves the field, any tag with that Pokemon as the source across the field is removed.

src/data/battler-tags.ts
innerthunder: Imo it's better to leave them in, even if they seem self-evident. On VSCode, at least, it makes it a bit easier to understand what types are required by the parameters, and the extra info isn't really obscuring much

src/data/battler-tags.ts
MokaStitcher: Ah I see, thank you

> The new condition here is still kind of redundant, too. When a Pokemon leaves the field, any tag with that Pokemon as the source across the field is removed.

If I remove the condition the test doesn't pass anymore, so there seems to be more to it than that? (also tested manually, and indeed if the condition is removed the tag doesn't get cleared if the source is switched out)

MokaStitcher (approved):
lgtm

src/data/battler-tags.ts
DayKev: The types are already shown in VSCode though?
image
The comments provide extra context beyond the types of the parameters (since TS automatically adds those to the tsdoc display, as opposed to JS). All the parameters I didn't list have self-documenting names.

PR #4609, opened 2024-10-07
Undo mistaken hotfix to Weird Dream ME

ben-lear:
The bug that was reported was in beta, not production as originally thought. So now prod is actually broken.

image

Rollback changes from the hotfix (not undoing the commit so we don't have to force push)


Tempo-anon (approved):
💢

Issue #4594, opened 2024-10-06
[Bug] Pollen Puff Damages Ally Before Healing


Snailman11:
### Describe the bug

When used in a double battle against an ally, Pollen Puff damages the ally before healing it.

Link to Discord's Bug Report (pollen puff does not heal)
https://discord.com/channels/1125469663833370665/1289355443394842686

### Reproduction

Use Pollen Puff on an Ally

### Expected behavior

It should only heal when used against an ally in doubles.

### Screenshots / Videos

https://github.com/user-attachments/assets/74a1df55-dee1-413a-942b-d3778b50e7ca

frutescens:
Might be a while before I can get to it, but the Heal Block-related changes are unrelated to this. Somehow the move's category change / handling changed during the time before the most-recent beta. Those are the notes from my quick research into it when the bug was first discovered.

PR #4615, opened 2024-10-08
[Bug] Pollen Puff behavior fix

frutescens:
Cat in front of keyboard will update later. Sorry.

Pollen Puff won't damage an ally before healing it. 

#4594 

image
New Parameter in pokemon.apply() that takes in the move category. 

### Screenshots/Videos

https://github.com/user-attachments/assets/c9547755-754f-48e8-abb2-6a084501422a

## How to test the changes?
Use Pollen Puff.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -2680,12 +2680,13 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
    * Applies the results of a move to this pokemon
    * @param source The {@linkcode Pokemon} using the move
    * @param move The {@linkcode Move} being used
-   * @param moveCategory the {@linkcode MoveCategory} of the move used
    * @returns The {@linkcode HitResult} of the attack
    */
-  apply(source: Pokemon, move: Move, moveCategory: MoveCategory): HitResult {
+  apply(source: Pokemon, move: Move): HitResult {
     const defendingSide = this.isPlayer() ? ArenaTagSide.PLAYER : ArenaTagSide.ENEMY;
-    if (moveCategory === MoveCategory.STATUS) {
+    const moveCategory = new Utils.NumberHolder(move.category);
+    applyMoveAttrs(VariableMoveCategoryAttr, source, this, move, moveCategory);
+    if (moveCategory.value === MoveCategory.STATUS) {
       const cancelled = new Utils.BooleanHolder(false);
       const typeMultiplier = this.getMoveEffectiveness(source, move, false, false, cancelled);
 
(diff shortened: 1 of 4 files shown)


src/field/pokemon.ts
```diff
@@ -2680,11 +2680,12 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
    * Applies the results of a move to this pokemon
    * @param source The {@linkcode Pokemon} using the move
    * @param move The {@linkcode Move} being used
+   * @param moveCategory the {@linkcode MoveCategory} of the move used
    * @returns The {@linkcode HitResult} of the attack
    */
-  apply(source: Pokemon, move: Move): HitResult {
+  apply(source: Pokemon, move: Move, moveCategory: MoveCategory): HitResult {
```
Tempo-anon, line 2686: Adding a third param here causes issues on line 2686 of `pokemon.ts` `apply(source: Pokemon, move: Move, moveCategory: MoveCategory): HitResult {` and line 168 of `sheer_force.test.ts` `applyPostDefendAbAttrs(PostDefendTypeChangeAbAttr, target, user, move, target.apply(user, move));`

Luckily those are the only two places that need updating I believe

Tempo-anon (changes requested):

src/field/pokemon.ts
frutescens: fix

src/phases/move-effect-phase.ts
```diff
@@ -192,7 +192,9 @@ export class MoveEffectPhase extends PokemonPhase {
              * made visible to the user until the resulting {@linkcode DamagePhase}
              * is invoked.
              */
-          const hitResult = !isProtected ? target.apply(user, move) : HitResult.NO_EFFECT;
+          const moveCategory = new Utils.IntegerHolder(move.category);
```
MokaStitcher, line 195:
```suggestion
          const moveCategory = new Utils.NumberHolder(move.category);
```
I think we're trying to move away from using integer since there's no actual difference with number

These lines may need to be updated too, but i think that's it?
https://github.com/pagefaultgames/pokerogue/blob/d8c914c7686c0fa4c74609ebed97c03f65a46f3e/src/data/move.ts#L4110-L4114

frutescens: Ok, fixed!

src/field/pokemon.ts
innerthunder, line 2686: I feel like things would be organized better if the move's final category was calculated within this function rather than in `MoveEffectPhase` (or, better yet, make a `getMoveCategory` method similar to `Pokemon.getMoveType`). Scope-wise, move category isn't used anywhere else in `MoveEffectPhase`, and I don't think that's likely to change with what's left to implement.

frutescens: Fixed.

innerthunder (approved):

DayKev (approved):

Tempo-anon (approved):

Issue #4604, opened 2024-10-07
[Bug] Trick Room Does Not End when Used Again While Active


Snailman11:
### Describe the bug

Trick Room continues to be present if the move is used while the Trick Room Field Effect is Active.

Magic Room and Wonder Room should have this interaction as well, however, they are not yet implemented.

Link to Discord (Trick Room not reverting when used twice)
https://discord.com/channels/1125469663833370665/1292692532542701659

### Reproduction

Use Trick Room with Trick Room active

### Expected behavior

https://replay.pokemonshowdown.com/gen9customgame-2217571587-wnyqjg92072t6yfc94dbav1z7ayhu2rpw
Trick Room should get rid of Trick Room

> Similar to Magic Room and Wonder Room, using Trick Room while it is already in effect will end it immediately.
-Bulba

### Screenshots / Videos

Trick Room does not remove Trick Room

https://github.com/user-attachments/assets/e303b81e-0536-40a6-b150-bbd9e8cb1fbd

Beta

https://github.com/user-attachments/assets/3db2100a-0051-4113-ad79-ec032da0c8e1

### Session export file

@
[TrickRoomTrickRoom.txt](https://github.com/user-attachments/files/17273546/TrickRoomTrickRoom.txt)

### Additional context

https://github.com/pagefaultgames/pokerogue/pull/3394
Wonder Room PR does not seem to mention that this should occur

https://github.com/pagefaultgames/pokerogue/pull/4343
Trick Room Challenges

PR #3528, opened 2024-08-13
[Bug] Fix first-turn status damage and arena hazards

ElizaAlex:
## What are the changes?

- Adjusted MoveEffectPhase.start() so that ENEMY_SIDE targeted moves no longer occur twice per use in double battles.
- Improved documentation / code readability
- Fixed Toxic / Poison dealing damage immediately when applied.
- Fixed Hazards not persisting through save
- Updated Toxic Orb test

I wasn't able to recreate toxic spikes not being absorbed, or the double message, though I didn't attempt to until after fixing the double battle problem. 

## Why am I doing these changes?
Addresses bugs compiled in #3240 

## What did change?
- Adjusted MoveEffectPhase to only apply the effects of ENEMY_SIDE targeted moves a single time, even in double battles. Previously this caused a bug where you would apply two layers of toxic spikes (and presumably spikes)
- Fixed bug where toxin/poison/burn damage would apply prematurely by adding a new phase scheduled after weather effects to check for and queue status damage.
- As a result of the Poison/Toxic/Burn changes, Toxic and Flame Orb now behave as they should, with the damage not triggering until the second turn. 
- Fixed ArenaTags not being loaded properly after a reset by basing them off of the handling of BattlerTags. Previously the ArenaData class constructor never successfully recovered saved tags from the JSON string. 
- Updated Toxic Orb test to reflect changes.
- ~~Refactored MoveEffectPhase in Phases.ts to improve readability, heavily reducing nested code by breaking the section into various documented helper functions~~ Separated out into #3541 

### Screenshots/Videos
Only one layer of Spikes/Toxic Spikes are applied even in double battles:

https://github.com/user-attachments/assets/ad1dc723-475f-453b-9fa8-8de296159438

Only one status affliction message is displayed per target:

https://github.com/user-attachments/assets/5aabe286-e59a-4393-b570-80ebf1fc4ed5

Hazards now persist through a reload:

https://github.com/user-attachments/assets/97e50058-3215-4ad2-b2cb-690f39fb9173

Toxic Spikes are correctly absorbed by grounded poison types:

https://github.com/user-attachments/assets/2c7845ee-9fd4-4928-8c28-20a11e1d439d

Additionally, Pokemon no longer take damage immediately upon being poisoned by Toxic Spikes. (The damage they take upon entering the field is due to the normal Spikes)

## How to test the changes?

1. Use overrides to force double battles, and to give your starter(s) of choice toxic spikes
[...]

Issue #3240, opened 2024-07-30
[Bug] Toxic Spikes implementation issues


innerthunder:
### Describe the bug

The move Toxic Spikes has several bugs in its implementation. This report compiles the bugs reported in other issues and in the Discord.

### Unreported issues with Toxic Spikes (data/evidence below)
1. When used in a double battle with both enemies active, Toxic Spikes sets two layers of spikes
2. After using Toxic Spikes in a previous double battle, the status affliction message appears twice for each affected enemy

### Other reported bugs related to Toxic Spikes for reference:
1. Toxic Spikes is removed from the field after reloading the save (https://github.com/pagefaultgames/pokerogue/issues/684)
3. Grounded Poison-type Pokemon do not absorb Toxic Spikes upon entering the field (https://github.com/pagefaultgames/pokerogue/issues/935)
4. Pokemon take (toxic) poison damage immediately after being poisoned by Toxic Spikes (https://github.com/pagefaultgames/pokerogue/issues/1320)

### Session export file

In this save, Toxic Spikes was used once by the player's Toxapex in a previous double battle. You can observe the duplicate messages at the start of battle as well as the fact that enemies are afflicted with Toxic Poison rather than normal poison.

[sessionData1_innerthunder (1).zip](https://github.com/user-attachments/files/16422919/sessionData1_innerthunder.1.zip)

### Expected behavior

- Using Toxic Spikes successfully in a double battle should add 1 layer of spikes to the enemy's side of the field
- The "__ was poisoned by Toxic Spikes!" message should only show once per affected enemy as they enter the field

### Screenshots / Videos

The start of Turn 1 in a battle after Toxic Spikes was used once in a double battle:
![image](https://github.com/user-attachments/assets/1ed87d3e-3047-45b1-a3b9-2ef9c8a9ff58)

### What OS did you observe the bug on?

PC/Windows

### Which browser do you use?

Chrome

Snailman11:
Here is a link to the Toxic Spikes bugs posted on discord;
1. Does not save after reload (Hazards)
[⁠Reloading save removes spikes](https://discord.com/channels/1125469663833370665/1267188711511031838)
Does not get absorbed by Poison types
[⁠Toxic Spikes not being absorbed](https://discord.com/channels/1125469663833370665/1267643732878954538)
Early poison tick damage
[⁠Toxic spikes and early damage](https://discord.com/channels/1125469663833370665/1241770168665903174)

[...]

PR #4628, opened 2024-10-10
[Bug] Room-Moves cancel out existing Room-Moves Arena Tags

frutescens:
Trick Room will end immediately if the move is used while the Trick Room Field Effect is Active. 
Magic Room and Wonder Room should have this interaction as well, however, they are not yet implemented.

Squishing bugs... #4604 

New parameter

### Screenshots/Videos
TBD

## How to test the changes?
Use Trick Room twice

## Checklist
- [x] Are the changes visual?


DayKev:
There's already an `onOverlap()` function, you should use that instead of adding a new field.

frutescens:
> There's already an `onOverlap()` function, you should use that instead of adding a new field.

Oh I added the new parameter because this bug would affect Wonder Room and Magic Room as well.

DayKev:
> > There's already an `onOverlap()` function, you should use that instead of adding a new field.
> 
> Oh I added the new parameter because this bug would affect Wonder Room and Magic Room as well.

You can just do `override onOverlap(...) { whatever.removeTag(thisTag); }`, you don't need to add an extra `removeOnOverlap` field to check in the `onAdd()` function.

PigeonBar:
This PR seems to be running into an issue where the game tries to apply the tag once for each Pokemon on the field, causing Trick Room to cancel itself out immediately:

https://github.com/user-attachments/assets/a106ea79-ed5e-4aac-8bea-506d8cca7a04

(See also: #3240, a similar bug for Toxic Spikes)

frutescens:
> This PR seems to be running into an issue where the game tries to apply the tag once for each Pokemon on the field, causing Trick Room to cancel itself out immediately:
> 
>  20241011.Trick.Room.Cancel.mp4 
> (See also: #3240, a similar bug for Toxic Spikes)

I added a new PR that might fix the issue.

PigeonBar:
Just tested it, Trick Room still cancels itself out immediately. (`substitute.test.ts` happens to use Trick Room, so you can use that for now to see if Trick Room is cancelling itself.)

It looks like fixing it will be a bit messy, though:
1. #3528 tried to fix moves with `MoveTarget.ENEMY_SIDE` (e.g., Spikes) applying multiple times in a turn, but the fix was accidentally deleted from that PR, so this will need its own fix, probably outside the scope of this PR. (See #4645.)
2. Meanwhile, Trick Room uses `MoveTarget.BOTH_SIDES` instead, so preventing it from applying multiple times will require the `MoveTarget.ENEMY_SIDE` fix to also include `MoveTarget.BOTH_SIDES`.

frutescens:
That's tricky...

DayKev:
> but the fix was accidentally deleted from that PR

Ack, I'll add it back in to the existing MEPhase PR.

PR #4629, opened 2024-10-10
[Bug] Fix enemy used PP flyout

### LINKED ISSUE #4622: [Beta][P3] Enemy used PP flyout does not count down
### Describe the bug

When an enemy uses a move for the first time the battle flyout updates and shows the move and correct PP, but every time the move is used after that the PP does not change

### Reproduction

Start any battle and stall while the opposing Pokemon uses its moves

### Expected behavior

The PP count show update for every move used

### Screenshots / Videos

Battle flyout after 4 turns on main:
_pp_main

Battle flyout the same 4 turns on beta:
_pp_beta

### Session export file

For the same battle as the screenshot, keep using dragon dance

[sessionData3_pp.txt](https://github.com/user-attachments/files/17309982/sessionData3_pp.txt)

### User data export file

_No response_

### Additional context

I confirmed that this has been introduced with PR #3974, probably an issue with the `MoveUsedEvent` not being called in the same way as before

DayKev:
The battle info flyout will properly show how much PP has been used by the enemy again.

Fixes #4622 

Fixed line 229 to use `this.move.ppUsed` instead of `ppUsed`.

### Screenshots/Videos
<details>
  <summary>After Fix</summary>

image

</details>

## How to test the changes?
Enter a battle and have the enemy use the same move at least twice.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/phases/move-phase.ts
+++ b/src/phases/move-phase.ts
@@ -8,10 +8,6 @@ import { SpeciesFormChangePreMoveTrigger } from "#app/data/pokemon-forms";
 import { getStatusEffectActivationText, getStatusEffectHealText } from "#app/data/status-effect";
 import { Type } from "#app/data/type";
 import { getTerrainBlockMessage } from "#app/data/weather";
-import { Abilities } from "#app/enums/abilities";
-import { BattlerTagType } from "#app/enums/battler-tag-type";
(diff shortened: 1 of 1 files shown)


Tempo-anon (approved):
Style refactoring in the same PR as bugfix :pikamad:

PR #4632, opened 2024-10-10
[Bug] Fix Early Bird

DayKev:
Early Bird will now appropriately reduce the number of sleep turns in all cases.

Discord bug report:
image

Sleep's `cureTurn` refactored to `turnsRemaining` and Early Bird now reduces it by an extra 1 each turn (implementation inspired by/copied from Showdown's).
Relevant changes in `ability.ts` and `move-phase.ts`.

## How to test the changes?
`npm run test early_bird`
`npm run test status-effect`
`npm run test will_o_wisp`

## Checklist


src/data/ability.ts
```diff
@@ -4192,18 +4192,31 @@ export class RedirectTypeMoveAbAttr extends RedirectMoveAbAttr {
...
+/**
+ * Used by Early Bird
+ * @param statusEffect - The {@linkcode StatusEffect} to check for
+ * @see {@linkcode apply}
+ */
 export class ReduceStatusEffectDurationAbAttr extends AbAttr {
-  private statusEffect: StatusEffect;
-
-  constructor(statusEffect: StatusEffect) {
+  constructor(
```
Tempo-anon, line 4201: Huh, I don't think this quite matches the style of other classes right? Usually the variable is declared outside the constructor

src/data/ability.ts
```diff
@@ -4192,18 +4192,31 @@ export class RedirectTypeMoveAbAttr extends RedirectMoveAbAttr {
...
 
-  apply(pokemon: Pokemon, passive: boolean, simulated: boolean, cancelled: Utils.BooleanHolder, args: any[]): boolean {
+  /**
+   * Applies the duration reduction of the `AbAttr`
+   * @param args - The args passed to the `AbAttr`:
+   * - `[0]` - The {@linkcode StatusEffect} of the Pokemon
+   * - `[1]` - The number of turns remaining until the status is healed
+   * @returns `true` if the ability was applied
+   */
+  apply(_pokemon: Pokemon, _passive: boolean, _simulated: boolean, _cancelled: Utils.BooleanHolder, args: any[]): boolean {
```
Tempo-anon, line 4214: The documentation for this method is not too clear. Is this just supposed to tick down the number of turns remaining an extra time?

src/data/status-effect.ts
```diff
@@ -1,22 +1,23 @@
...
-    this.effect = effect;
-    this.turnCount = turnCount === undefined ? 0 : turnCount;
-    this.cureTurn = cureTurn!; // TODO: is this bang correct?
-  }
+  constructor(
+    public effect: StatusEffect,
+    /** Only used by the Toxic status effect */
+    public turnCount: number = 0,
+    /** Only used by the Sleep status effect */
+    public turnsRemaining?: number
```
Tempo-anon, line 13: nit: if these two variables are only used for two statuses, I think it'll be more readable for now if their names reflected it, like `toxicTurnCount` and `turnsRemainingToSleep`

src/phases/obtain-status-effect-phase.ts
```diff
@@ -9,26 +9,24 @@ import { PokemonPhase } from "./pokemon-phase";
 import { PostTurnStatusEffectPhase } from "./post-turn-status-effect-phase";
 
 export class ObtainStatusEffectPhase extends PokemonPhase {
```
Tempo-anon, line 11: I am now aware that all these new phase files have no documentation. 🐸 REEEEEEEEEEEEEEEEE

src/test/abilities/early_bird.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    expect(player.getLastXMoves(1)[0].result).toBe(MoveResult.FAIL);
+
+    game.move.select(Moves.SPLASH);
+    await game.toNextTurn();
+
+    expect(player.status?.effect).toBeUndefined();
+    expect(player.getLastXMoves(1)[0].result).toBe(MoveResult.SUCCESS);
+  });
+
+  it("reduces 3-turn sleep to 1 turn", async () => {
```
Tempo-anon, line 62: nit: should also add a check like 4 turn sleep to 2 turn

Tempo-anon (commented):
Bulbapedia and Smogon are once again disagreeing on the exact implementation of early bird. Bulbapedia just says that when sleep is applied the turns to sleep is cut in half upon application while Smogon says that the turns to sleep counter ticks down an extra time. :pikastare: This will affect interactions like neutralizing gas.

src/data/ability.ts
DayKev: There were a couple done this way (which is how I figured out it was possible), and when the variables are just assigned directly with no logic applied this is the better way imo. No need to do:
```ts
class myClass {
  public myVar: firstType;
  public anotherVar: secondType[];
  // ...

  constructor(myVar: firstType, anotherVar: secondType[]/* ... */) {
    super();
    this.myVar = myVar;
    this.anotherVar = anotherVar;
    // ...
  }
}
```
when you can just do:
```ts
class myClass {
  constructor(
    public myVar: firstType,
    public anotherVar: secondType[],
    // ...
  ) {
    super();
  }
}

src/data/ability.ts
DayKev: Yeah. Lemme update the tsdocs.

src/test/abilities/early_bird.test.ts
DayKev: Is there anything that sets 4 turn sleep? Default random sleep (ie: from Spore, etc) is 1-3 turns, and Rest is always 2 turns.

src/data/ability.ts
torranx: There really is not much of a difference, the first one just offers explicitness esp for more complex/larger classes. Readability-wise, it depends on the dev but personally I prefer the first one. But considering that the _code-style we follow_ is the former, I'd say keep it that way for consistency. Unless you're willing to rewrite everything 😄.

DayKev: > Unless you're willing to rewrite everything

If there's a way to do it [mostly-]automatically, I'd love to do that lmao. I wonder if there's a prettifier that can do it... 🤔 

> There really is not much of a difference, the first one just offers explicitness esp for more complex/larger classes.

I prefer minimizing the amount of duplicate/boilerplate code that's necessary. Why write something three times when we only need to do it once?

src/phases/obtain-status-effect-phase.ts
DayKev: :despair:

src/data/ability.ts
torranx: You're technically not writing it multiple times. Key difference is it separates property declaration/initialization _and_ assignment. TBC,  I'm not entirely against this change, but this is _currently_ against our code-style and would like to have others weigh in on this.

innerthunder: Personally I find it easier to read the current code-style, and I'd rather have things kept consistent when possible

Tempo-anon:
So do we have confirmation of how early bird works? Does it just halve the initial turn count set for sleep (which I think would be much simpler to implement) or does it tick down sleep counter twice?

DayKev:
> So do we have confirmation of how early bird works? Does it just halve the initial turn count set for sleep (which I think would be much simpler to implement) or does it tick down sleep counter twice?

I'm not sure how to check how it works in SV's code, though I suppose we could try to look at the gen 3/4 decomps (and maybe BDSP?) and just assume they never changed how it worked in later gens. Absent that I'd prefer matching Showdown (which is what I copied).

DayKev:
@Tempo-anon Well I looked up the code of Emerald and it does indeed seem to reduce sleep turns by 2 rather than divide the turns remaining by half:

https://github.com/pret/pokeemerald/blob/a55c75d350969624745a357711ba01edf8ce9590/src/battle_util.c#L2013-L2016

Tempo-anon:
dang, I was hoping it'd be just halving sleep turn count on activation which would have made this a lot simpler to implement

innerthunder (approved):

PR #4410, opened 2024-09-25
[Feature][UI] Save Preview

frutescens:
## What are the changes the user will see?
Users can use Button.RIGHT to see more information about their runs. 

## Why am I making these changes?
I thought it would be neat. 

## What are the changes from a developer perspective?
Run Info uses an enum to determine its purpose and chooses to display/not display certain information. 
Trainer Parties are not displayed in this mode. Only the trainer sprites are.
New function added to ME Utils to facilitate getting ME locale strings without having to generate an entire ME.

### Screenshots/Videos
Full UX

https://github.com/user-attachments/assets/4e7b8fde-5dce-44db-9de5-4ef35496bda1

Save Previews:
<details>
  <summary>Mystery Encounters</summary>

image

</details>
<details>
  <summary>Trainer Double</summary>

image

</details>
<details>
  <summary>Trainer Single</summary>

image

</details>
<details>
  <summary>Wild Double</summary>

image

</details>

---

<details>
  <summary>Got the scrolling to work now</summary>

https://github.com/user-attachments/assets/3d9591ee-e371-49f9-92e0-9374b7b56538

</details>

## How to test the changes?
Press right on any session save slot. 
Check in a variety of situations / languages -> Important to see if it works for all trainers :) 

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #4633, opened 2024-10-10
[P3 Beta][UI] Fix loading behavior introduced with save preview

MokaStitcher:
(Beta only) No more weirdness when clicking "Load Game" with slow network
(Beta only) Fix slot not always being reverted to the right place when moving the cursor

In PR #4410 the way session slots were displayed while loading was altered in a few problematic ways:
- the save slots would only get shown once they have finished loading instead of showing an empty slot with "loading" message
- during loading it was still possible to move the cursor (like it was before), but due to the above point it would cause visual issues and bugs
- the save slots would get loaded sequentially instead of in parallel
This is especially visible when playing on a device with slowish internet connection, and wasn't noticed before because our testing was made locally
This PR reverts some of these changes to get back to the old behavior of the UI + fixes a small issue with the slots not always moving back to the correct spot when moving the cursor

- reverted most changes to `populateSessionSlots` from PR #4410:
  - add the slot to the UI before loading it instead of after, so that it shows as "loading..." like before
  - remove `await` so that all slots can load in parallel
- when a slot has finished loading, call `setCursor` if needed so that the slot gets shifted and the right arrow for the preview gets shown
- Since it's possible to move the cursor, exit the view or load a session while others are still loading, added a safeguard after a session finished loading to make sure the view is still active before making any changes to it. (this is actually a bug that was already present on main with the old behavior and would cause errors in the console, but no crash)

### Screenshots/Videos
<details><summary>Current behavior on main (what we want):</summary>

https://github.com/user-attachments/assets/444bca13-f481-499c-95d7-2cb964b4c61b

</details>

<details><summary>Before, beta:</summary>

https://github.com/user-attachments/assets/90e9c672-df4c-4e3a-b500-f7802a619896

</details>

<details><summary>After:</summary>

https://github.com/user-attachments/assets/ed96bb68-319c-4f67-a26e-bf596ac5458c

</details>

<details><summary>Slot position issue, after: (see comments for before video)</summary> 

https://github.com/user-attachments/assets/5eb0abaf-dceb-4661-aef1-7fb193d9e533

</details>

## How to test the changes?
The easiest way is to add a random delay after a session finished loading, 
line 405 of `save-slot-select-ui-handler.ts`:
`await new Promise(r => setTimeout(r, 500 * Math.random() * 5));`

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/ui/save-slot-select-ui-handler.ts
+++ b/src/ui/save-slot-select-ui-handler.ts
@@ -12,7 +12,7 @@ import { Mode } from "./ui";
 import { addWindow } from "./ui-theme";
 import { RunDisplayMode } from "#app/ui/run-info-ui-handler";
 
-const sessionSlotCount = 5;
+const SESSION_SLOTS_COUNT = 5;
 
 export enum SaveSlotUiMode {
   LOAD,
@@ -84,12 +84,10 @@ export default class SaveSlotSelectUiHandler extends MessageUiHandler {
     this.saveSlotSelectCallback = args[1] as SaveSlotSelectCallback;
 
     this.saveSlotSelectContainer.setVisible(true);
-    this.populateSessionSlots()
-      .then(() => {
-        this.setScrollCursor(0);
-        this.setCursor(0);
-      });
+    this.populateSessionSlots();
 
+    this.setScrollCursor(0);
+    this.setCursor(0);
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


DayKev (approved):

flx-sta:
https://github.com/user-attachments/assets/315ce821-001e-4aed-8782-b2b3bfc95e8a

Could you also fix this bug when you are already working on it?

@MokaStitcher

MokaStitcher:
@flx-sta Fixed, thank you! (description edited, or about to be)

DayKev (approved):

Issue #4537, opened 2024-10-01
[P3] "Rapid increase" reported when max number was reached


flx-sta:
### Describe the bug

When the max number of a stat increase was reached, it doesn't show an info that it can't increase further but reports that it was `rapidly increased`

### Reproduction

1. Pick an attack that increases stats
2. Keep using it till the max number (+6) is reached
3. Keep using it and read what the message says (see video)

### Expected behavior

It should tell the player that it can't be increased any further instead

### Screenshots / Videos

https://github.com/user-attachments/assets/6ca597f9-84fa-4334-8229-a2da2a5cbbe4

### Additional context

- Apparently it was introduced with the stats refactor
- Discord Threads:
    - https://discord.com/channels/1125469663833370665/1282189311101501463

PigeonBar:
Just wanted to add that this also occurs for stat drops:

https://github.com/user-attachments/assets/492c7de7-30fc-4713-af94-8404f238aced

PR #4635, opened 2024-10-10
[P3] Fix "Won't Go Any Lower/Higher" Not Appearing

xsn34kzx:

Lowering or raising a stat stage below -6 or beyond +6 will no longer display the wrong message ("X severely fell" or "X drastically rose").

#2699 had replaced the original `getBattleStatLevelChangeDescription`, which incorrectly translated `switch-case` to an `if-else` for the case where stat stages could no longer be lowered/raised. This caused the wrong stat stage change description to get pushed instead of the correct one.

Added `stages > 2` to the boolean that triggers the "X severely fell"/"X drastically rose" messages to prevent it from triggering in the case that `stages` is equal to 0, which indicates that a stat cannot be lowered or raised.

#### Before

Refer to #4537 for a visual of the bug.

#### After

https://github.com/user-attachments/assets/2534e640-fe92-4fa5-9d1d-e8fa4eb0bb78

## How to test the changes?

Use any stat stage lowering/increasing move (e.g., `Shell Smash`) and use keep using it until you reach the minimum/maximum most stat stage. Using the move one more time should now display the right message ("X won't go any higher/lower!") as opposed to the current one.

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/enums/stat.ts
+++ b/src/enums/stat.ts
@@ -50,7 +50,7 @@ export function getStatStageChangeDescriptionKey(stages: number, isIncrease: boo
     return isIncrease ? "battle:statRose" : "battle:statFell";
   } else if (stages === 2) {
     return isIncrease ? "battle:statSharplyRose" : "battle:statHarshlyFell";
-  } else if (stages <= 6) {
+  } else if (stages > 2 && stages <= 6) {
     return isIncrease ? "battle:statRoseDrastically" : "battle:statSeverelyFell";
   }
   return isIncrease ? "battle:statWontGoAnyHigher" : "battle:statWontGoAnyLower";


DayKev (approved):

flx-sta:
https://github.com/user-attachments/assets/315ce821-001e-4aed-8782-b2b3bfc95e8a

Could you also fix this bug when you are already working on it?

@MokaStitcher

flx-sta (approved):
lgtM

https://github.com/user-attachments/assets/8c301410-1507-45d3-8580-b9f65e6f4bc9

innerthunder (approved):

PR #4638, opened 2024-10-11
[P3] Fix "falsex" type hint for Sheer Cold

innerthunder:
Sheer Cold will no longer show a type hint of "falsex" against Ice-type Pokemon

Sheer Cold's code is old as hell but somehow worked with the new type effectiveness code, so I missed it earlier.

`data/move`: Sheer Cold's `IceNoEffectTypeAttr` no longer interprets the type effectiveness multiplier as a boolean, and now sets the type multiplier to 0 instead of `false`.

### Screenshots/Videos

image

## How to test the changes?
1. Use `src/overrides.ts` to give your Pokemon Sheer Cold and set the enemy species to an Ice-type.
2. Start a new game
3. Enable type hints
4. Sheer Cold's type hint should show as "0x" instead of "falsex"

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -4556,18 +4556,19 @@ export class WaterSuperEffectTypeMultiplierAttr extends VariableMoveTypeMultipli
 export class IceNoEffectTypeAttr extends VariableMoveTypeMultiplierAttr {
   /**
    * Checks to see if the Target is Ice-Type or not. If so, the move will have no effect.
-   * @param {Pokemon} user N/A
-   * @param {Pokemon} target Pokemon that is being checked whether Ice-Type or not.
-   * @param {Move} move N/A
-   * @param {any[]} args Sets to false if the target is Ice-Type, so it should do no damage/no effect.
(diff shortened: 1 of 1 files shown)


torranx (approved):
lol how was this not broken

Tempo-anon (commented):
What happens when you use sheer cold on an ice type? Does it do 0 damage or does it fail? 

Screencapped off showdown for what happens when you sheer cold an ice type
image

DayKev (approved):

DayKev:
> What happens when you use sheer cold on an ice type? Does it do 0 damage or does it fail?

It fails, which is handled in Move Phase iirc.

innerthunder:
> What happens when you use sheer cold on an ice type? Does it do 0 damage or does it fail?

Works as expected (ignore the missing Splash anim, that's an unrelated bug with overrides).

https://github.com/user-attachments/assets/372cc3e3-7454-46ca-87da-5621b5ccbe8e

PR #4640, opened 2024-10-11
[Refactor] Clean up `StatStageChangePhase` and `PokemonPhase`

DayKev:
N/A

Because it was there.

The `const end = () => { ... };` inside of the `start()` method was moved into its own class method.
An unused method `aggregateStatStageChanges()` was removed.
`player` was renamed to `isPlayer` in `PokemonPhase` for clarity.

## How to test the changes?
Use moves that alter stat stages.

## Checklist


DayKev:
Apparently if you rename a branch it closes the related PR instead of just updating it...

PR #4646, opened 2024-10-11
[Ability] Remove partials from Serene Grace and Shield Dust

frutescens:
Users will be less confused when using Pokemon with Serene Grace and Shield Dust. 

Making our shareholders happy. 

In PR #246 Secret Power's (as-of-now unimplemented) 30% secondary effect chance was cited as the reason for why Serene Grace and Shield Dust was labeled partial; however, the secondary effect modifications that Serene Grace/Shield Dust provide should affect that 30% secondary effect chance even though Secret Power is only partially implemented. 

## Checklist

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -4905,8 +4905,7 @@ export function initAbilities() {
       .attr(TypeImmunityAddBattlerTagAbAttr, Type.FIRE, BattlerTagType.FIRE_BOOST, 1)
       .ignorable(),
     new Ability(Abilities.SHIELD_DUST, 3)
-      .attr(IgnoreMoveEffectsAbAttr)
-      .partial(),
+      .attr(IgnoreMoveEffectsAbAttr),
     new Ability(Abilities.OWN_TEMPO, 3)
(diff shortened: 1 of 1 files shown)


DayKev (approved):

DayKev:
Maybe we should say that Sheer Force blocking Meloetta's form change is a `wontfix` and remove the partial from it too? 🤔

frutescens:
> Maybe we should say that Sheer Force blocking Meloetta's form change is a `wontfix` and remove the partial from it too? 🤔

It's not just Meloetta's form change; there's some other secondary effects that Sheer Force blocks. 
image

DayKev:
Ah, so shell bell and a handful of abilities (half of which are unimplemented :plains:).

innerthunder (approved):

frutescens:
We can always add the Partial label back after those items are implemented :yawn:

Tempo-anon (changes requested):
We can remove the partials but let's at least leave a comment here explaining any future TODOs so someone can just read a comment instead of chasing down this PR

DayKev:
Seems like #4649 supercedes this PR.

PR #4657, opened 2024-10-13
[Refactor] Add friendship related constants

Fontbane:

- None

- To make these constants easier to change if need be

- Adds 4 new `const`s to `src/data/balance/starters.ts`:
  - `CLASSIC_CANDY_FRIENDSHIP_MULTIPLIER`
  - `FRIENDSHIP_GAIN_VICTORY`
  - `FRIENDSHIP_GAIN_RARE_CANDY`
  - `FRIENDSHIP_LOSS_FAINT`
 - Replaces magic numbers with the appropriate constants where necessary

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/balance/starters.ts
+++ b/src/data/balance/starters.ts
@@ -2,6 +2,12 @@ import { Species } from "#enums/species";
 
 export const POKERUS_STARTER_COUNT = 5;
 
+// #region Friendship constants
+export const CLASSIC_CANDY_FRIENDSHIP_MULTIPLIER = 2;
+export const FRIENDSHIP_GAIN_VICTORY = 2;
+export const FRIENDSHIP_GAIN_RARE_CANDY = 5;
+export const FRIENDSHIP_LOSS_FAINT = 10;
+
 /**
  * Function to get the cumulative friendship threshold at which a candy is earned
  * @param starterCost The cost of the starter, found in {@linkcode speciesStarterCosts}
--- a/src/phases/faint-phase.ts
+++ b/src/phases/faint-phase.ts
@@ -19,6 +19,7 @@ import { SwitchPhase } from "./switch-phase";
 import { VictoryPhase } from "./victory-phase";
 import { SpeciesFormChangeActiveTrigger } from "#app/data/pokemon-forms";
 import { SwitchType } from "#enums/switch-type";
+import { FRIENDSHIP_LOSS_FAINT } from "#app/data/balance/starters";
 
 export class FaintPhase extends PokemonPhase {
   private preventEndure: boolean;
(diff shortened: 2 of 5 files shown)


src/data/balance/starters.ts
```diff
@@ -2,6 +2,12 @@ import { Species } from "#enums/species";
 
 export const POKERUS_STARTER_COUNT = 5;
 
+// #region Friendship constants
+export const CLASSIC_CANDY_FRIENDSHIP_MULTIPLIER = 2;
+export const FRIENDSHIP_GAIN_VICTORY = 2;
```
Tempo-anon, line 7: nit: I think this can be named something clearer since "victory" is a bit ambiguous. It's supposed to represent any enemy Pokemon fainting so long as the friendship gainer participated

src/data/balance/starters.ts
```diff
@@ -2,6 +2,12 @@ import { Species } from "#enums/species";
...
+export const FRIENDSHIP_GAIN_RARE_CANDY = 5;
+export const FRIENDSHIP_LOSS_FAINT = 10;
```
Tempo-anon, line 9: nit: would also name these `FRIENDSHIP_GAIN_ON_RARE_CANDY_USE` and `FRIENDSHIP_LOSS_ON_FAINT` just so it makes grammatical sense

src/data/balance/starters.ts
```diff
@@ -2,6 +2,12 @@ import { Species } from "#enums/species";
 
 export const POKERUS_STARTER_COUNT = 5;
 
+// #region Friendship constants
```
xsn34kzx, line 5:
```suggestion
// Friendship constants
```
NIT: Correct me if I'm wrong, but I don't think we use `region` folding anywhere in the code base.

src/phases/faint-phase.ts
```diff
@@ -127,7 +128,7 @@ export class FaintPhase extends PokemonPhase {
 
     pokemon.faintCry(() => {
       if (pokemon instanceof PlayerPokemon) {
-        pokemon.addFriendship(-10);
+        pokemon.addFriendship(FRIENDSHIP_LOSS_FAINT);
```
xsn34kzx, line 131:
```suggestion
        pokemon.addFriendship(-FRIENDSHIP_LOSS_FAINT);
```
Either the negative should be put on the constant or this should get a `-` attached to it here.

xsn34kzx (dismissed):
Minor NIT and a change to make sure that friendship doesn't get increased on faint by accident.

src/data/balance/starters.ts
DayKev: There's a couple, but it's not widely used.

src/phases/faint-phase.ts
Tempo-anon: lmao so true. Either `pokemon.addFriendship(-1 * FRIENDSHIP_LOSS_ON_FAINT)` or create a new `loseFriendship` method

damocleas (approved):

Tempo-anon (approved):

PR #4636, opened 2024-10-11
[Ability] Partially Implement Infiltrator

innerthunder:
## What are the changes the user will see?
This partially implements [Infiltrator](https://bulbapedia.bulbagarden.net/wiki/Infiltrator_(Ability)). Infiltrator now doesn't bypass the effects of Mist, but otherwise functions as expected.

## Why am I making these changes?
Less (N) moves (even though technically this should've been marked (P) after the Substitute implementation)

## What are the changes from a developer perspective?
- `data/ability`: New `InfiltratorAbAttr` flags Light Screen, Reflect, Aurora Veil, Safeguard, and Substitute as "bypassed" when they would otherwise be applied.
- `field/pokemon`: New `isSafeguarded` function to check if a Pokemon is under the effects of Safeguard when defending against another Pokemon's move.
- `data/arena-tag` and `data/move`: Relevant tags and functions have been updated to apply the attacker's `InfiltratorAbAttr` and cancel or bypass effects.
- `test/abilities/infiltrator.test`: New integration tests for Infiltrator against all relevant field effects and Substitute (Mist test currently skipped)

## How to test the changes?
`npm run test infiltrator`

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I considered writing automated tests for the issue?
- [x] If I have text, did I make it translatable and add a key in the English locale file(s)?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)
- [?] Are the changes visual?
  - [?] Have I provided screenshots/videos of the changes?

PR #4294, opened 2024-09-17
[Qol][Mirror] Modifiers type inference v2

flx-sta:
> [!NOTE]
> ### Mirror of #1747 
> <sub>Copied PR template/comment from #1747</sub>

---

## What are the changes?
Add type inference to all modifier classes

## Why am I doing these changes?
`as Type` assertions are dangerous and this make the user experience better by adding intellisense

## What did change?
1. Make applyModifier method accept generic modifier classes and get typeof args from inference
2. Refactor all places modifiers are used to follow the new type inference param scheme
3. Create a new utility type that maps to `new(...args: any[]) => SomeClass` 
4. Refactor all places that used `new(...args: any[]) => SomeClass` to now use `Utils.ClassType<T>`

## How to test the changes?
Try it out with intellisense.
Run the game in localhost to see the changes aren't breaking

## Checklist
- [x] There is no overlap with another PR?
- [ ] The PR is self-contained and cannot be split into smaller PRs?
  - The `Utils.ClassType<T>` alias can be its own separate PR 
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
    - [x] Are all unit tests still passing? (`npm run test`)

~~- [ ] Are the changes visual?~~
~~- [ ] Have I provided screenshots/videos of the changes?~~

PR #4659, opened 2024-10-14
[Refactor] Add type inference and support for simulated calls to `ArenaTag.apply`

innerthunder:
N/A

Simulated calls to arena tags are necessary for the Enemy AI to evaluate turn actions silently, especially when [Infiltrator](#4636) is implemented. With that change, I figured it was also a good time to add type inference to `ArenaTag.apply` along the lines of #4294 

`data/arena-tag`: The signature for `ArenaTag.apply` has changed from
```ts
apply(arena: Arena, args: any[]);
```
to
```ts
apply(arena: Arena, simulated: boolean, ...args: unknown[]);
```

The new `simulated` argument, much like the argument of the same name for abilities' `apply` functions, allows arena tags to be applied without any observable changes to game state. `args` has also been changed to use spread syntax, which allows `overrides` of the function in subclasses to expand `args` into multiple arguments.

`applyTagsOnSide` and `applyTags` have also been changed accordingly, along with all references to the two functions in the codebase.

### Screenshots/Videos
N/A

## How to test the changes?
For a specific example, you can check the screen tests:
`npm run test reflect`
`npm run test light_screen`
`npm run test aurora_veil`

## Checklist

### DIFF
--- a/src/data/arena-tag.ts
+++ b/src/data/arena-tag.ts
@@ -1,7 +1,7 @@
 import { Arena } from "#app/field/arena";
 import BattleScene from "#app/battle-scene";
 import { Type } from "#app/data/type";
-import * as Utils from "#app/utils";
+import { BooleanHolder, NumberHolder, toDmgValue } from "#app/utils";
 import { MoveCategory, allMoves, MoveTarget, IncrementMovePriorityAttr, applyMoveAttrs } from "#app/data/move";
 import { getPokemonNameWithAffix } from "#app/messages";
 import Pokemon, { HitResult, PokemonMove } from "#app/field/pokemon";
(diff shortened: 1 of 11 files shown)


src/data/arena-tag.ts
```diff
@@ -249,38 +257,35 @@ export class ConditionalProtectTag extends ArenaTag {
...
-   * @param arena The arena containing this tag
-   * @param args\[0\] (Utils.BooleanHolder) Signals if the move is cancelled
-   * @param args\[1\] (Pokemon) The Pokemon using the move
-   * @param args\[2\] (Pokemon) The intended target of the move
-   * @param args\[3\] (Moves) The parameters to the condition function
-   * @param args\[4\] (Utils.BooleanHolder) Signals if the applied protection supercedes protection-ignoring effects
-   * @returns
+   * @param arena the {@linkcode Arena} containing this tag
+   * @param simulated `true` if the tag is applied quietly; `false` otherwise.
+   * @param args the arguments for this tag's application:
```
DayKev, line 264: You forgot to update this with the named params.

src/data/arena-tag.ts
```diff
@@ -563,13 +575,12 @@ export class IonDelugeTag extends ArenaTag {
   /**
    * Converts Normal-type moves to Electric type
    * @param arena n/a
-   * @param args
-   * - `[0]` {@linkcode Utils.NumberHolder} A container with a move's {@linkcode Type}
+   * @param simulated n/a
+   * @param moveType a {@linkcode Utils.NumberHolder | NumberHolder} containing a move's {@linkcode Type}
```
DayKev, line 579: > {@linkcode Utils.NumberHolder | NumberHolder}

ew
Can you just convert the `* as Utils` from a namespace import to named imports (vscode's refactor tool should be able to handle it for you)?

torranx (approved):

PR #4661, opened 2024-10-14
[Bug] Potential fix for Expert Breeder's Pokemon being invisible

PigeonBar:

(Original bug report: https://discord.com/channels/1125469663833370665/1289869659297812520)

The Expert Pokemon Breeder's ace should never be invisible.

(Note: The linked bug report also mentioned some crashes when starting the Expert Pokemon Breeder. However, I was never able to reproduce a crash. So, I guess that this PR could also fix the crashes, but we can't know for sure.)

Fixing a bug, potentially fixing a crash.

The ME options for Expert Pokemon Breeder now properly return a Promise so that the game can wait for the Promise to be resolved before starting the battle. This Promise includes loading all assets needed for the battle, so this change should cause the game to wait for all assets to be loaded before starting the battle.

Also updated some tests.

## How to test the changes?

Modify `encounter-phase.utils.ts:340` to add a 5000ms delay to asset loading, like this:

    loadEnemyAssets.push(new Promise(resolve => {
      setTimeout(() => {
        enemyPokemon.loadAssets().then(() => {
          resolve();
        });
      }, 5000);
    }));

Also: `npm run test expert-breeder`

### Screenshots/Videos

Before the changes:

https://github.com/user-attachments/assets/b9bce9ca-dff9-4194-9921-8cb75e639c61

After the changes:

https://github.com/user-attachments/assets/c0f1f6aa-be7c-4a9e-be5f-67237e00876d

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
+++ b/src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
@@ -140,7 +140,7 @@ describe("The Expert Pokémon Breeder - Mystery Encounter", () => {
           }));
         }
 
-        return scene.currentBattle?.enemyParty || [];
+        return scene.currentBattle?.enemyParty ?? [];
       });
 
       await runMysteryEncounterToEnd(game, 1, undefined, true);
@@ -219,7 +219,7 @@ describe("The Expert Pokémon Breeder - Mystery Encounter", () => {
           }));
         }
 
-        return scene.currentBattle?.enemyParty || [];
+        return scene.currentBattle?.enemyParty ?? [];
       });
 
       await runMysteryEncounterToEnd(game, 2, undefined, true);
@@ -298,7 +298,7 @@ describe("The Expert Pokémon Breeder - Mystery Encounter", () => {
           }));
         }
 
-        return scene.currentBattle?.enemyParty || [];
+        return scene.currentBattle?.enemyParty ?? [];
       });
 
       await runMysteryEncounterToEnd(game, 3, undefined, true);
--- a/src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
+++ b/src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
@@ -124,10 +124,31 @@ describe("The Expert Pokémon Breeder - Mystery Encounter", () => {
       });
     });
 
-    it("should start battle against the trainer", async () => {
+    it("should start battle against the trainer with correctly loaded assets", async () => {
       await game.runToMysteryEncounter(MysteryEncounterType.THE_EXPERT_POKEMON_BREEDER, defaultParty);
+
+      let successfullyLoaded = false;
+      vi.spyOn(scene, "getEnemyParty").mockImplementation(() => {
+        const ace = scene.currentBattle?.enemyParty[0];
+        if (ace) {
+          // Pretend that loading assets takes an extra 500ms
+          vi.spyOn(ace, "loadAssets").mockImplementation(() => new Promise(resolve => {
+            setTimeout(() => {
+              successfullyLoaded = true;
+              resolve();
+            }, 500);
+          }));
+        }
+
+        return scene.currentBattle?.enemyParty || [];
+      });
+
       await runMysteryEncounterToEnd(game, 1, undefined, true);
 
+      // Check that assets are successfully loaded
+      expect(successfullyLoaded).toBe(true);
+
+      // Check usual battle stuff
       expect(scene.getCurrentPhase()?.constructor.name).toBe(CommandPhase.name);
       expect(scene.currentBattle.trainer).toBeDefined();
       expect(scene.currentBattle.mysteryEncounter?.encounterMode).toBe(MysteryEncounterMode.TRAINER_BATTLE);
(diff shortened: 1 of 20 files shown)


src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
```diff
@@ -124,10 +124,31 @@ describe("The Expert Pokémon Breeder - Mystery Encounter", () => {
...
+          // Pretend that loading assets takes an extra 500ms
+          vi.spyOn(ace, "loadAssets").mockImplementation(() => new Promise(resolve => {
+            setTimeout(() => {
+              successfullyLoaded = true;
+              resolve();
+            }, 500);
+          }));
+        }
+
+        return scene.currentBattle?.enemyParty || [];
```
DayKev, line 143:
```suggestion
        return scene.currentBattle?.enemyParty ?? [];
```

src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
```diff
@@ -182,10 +203,31 @@ describe("The Expert Pokémon Breeder - Mystery Encounter", () => {
...
```
DayKev, line 222:
```suggestion
        return scene.currentBattle?.enemyParty ?? [];
```

src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
```diff
@@ -240,10 +282,31 @@ describe("The Expert Pokémon Breeder - Mystery Encounter", () => {
...
```
DayKev, line 301:
```suggestion
        return scene.currentBattle?.enemyParty ?? [];
```

src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
```diff
@@ -124,10 +124,31 @@ describe("The Expert Pokémon Breeder - Mystery Encounter", () => {
...
 
-    it("should start battle against the trainer", async () => {
+    it("should start battle against the trainer with correctly loaded assets", async () => {
       await game.runToMysteryEncounter(MysteryEncounterType.THE_EXPERT_POKEMON_BREEDER, defaultParty);
+
+      let successfullyLoaded = false;
+      vi.spyOn(scene, "getEnemyParty").mockImplementation(() => {
+        const ace = scene.currentBattle?.enemyParty[0];
+        if (ace) {
+          // Pretend that loading assets takes an extra 500ms
```
DayKev, line 134: Tests currently override delays to be 1ms, might need to disable that for these tests. Lemme look into making some helper functions for that.

src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
PigeonBar: I used `||` instead of `??` because that's also the implementation of the original `scene.getEnemyParty()`, should it be changed there as well?

https://github.com/pagefaultgames/pokerogue/blob/e962ac1f182d33e6f06fef1858c49b7ffb90c4b9/src/battle-scene.ts#L791-L793

src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
PigeonBar: `inputs.test.ts` also ultimately relies on `setTimeout`, so using `setTimeout` here should be fine.

MokaStitcher:
lgtm
Wondering though, it seems like some MEs use `return initBattleWithEnemyConfig(scene, config);` and others `await initBattleWithEnemyConfig(scene, config);`
Do you why that is and if there's a ultimately a difference between these? If there isn't it probably should be all changed to one or the other

src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
DayKev: True; though that test might also just be broken lmao.

src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
DayKev: Yeah, though I have a PR changing that already, it just hasn't been merged yet.

src/test/mystery-encounter/encounters/the-expert-breeder-encounter.test.ts
DayKev: Well the tests increase in time taken appropriate to the timeout they're using so, I guess it's fine. 🤷 
image

PigeonBar:
> lgtm Wondering though, it seems like some MEs use `return initBattleWithEnemyConfig(scene, config);` and others `await initBattleWithEnemyConfig(scene, config);` Do you know why that is and if there's a ultimately a difference between these? If there isn't it probably should be all changed to one or the other

I've changed it to use `await` more consistently, so that it's easier to tell that we're using these as Promises.

DayKev (approved):

DayKev:
Seems to work, also tested by using browser devtools to simulate slow network in addition to the method mentioned in the OP.

PR #4663, opened 2024-10-14
[Bug Fix] Attacks that miss against pokemon in semi invul state that have abilities such as volt absorb will not trigger while semi invulnerable

PrabbyDD:
This interaction will now work properly. For example, if thunderbolt is used against a pokemon with volt absorb, and that pokemon with volt absorb is in a semi invulnerable state, the attack will just miss instead of triggering the ability. 

related to #4337 

Bug fix

added check in move-effect-phase.ts for if pokemon is in semi invulnerable state

### Screenshots/Videos

https://github.com/user-attachments/assets/2d704a4f-2116-41cb-b91b-2278f09e9eb4

## How to test the changes?
Override files, run the updated volt absorb test file, and manually change return value of hitcheck() to false. 

## Checklist
- [X ] **I'm using `beta` as my base branch**
- [X ] There is no overlap with another PR?
- [ X] The PR is self-contained and cannot be split into smaller PRs?
- [X ] Have I provided a clear explanation of the changes?
- [X ] Have I considered writing automated tests for the issue?
- [X ] Have I tested the changes (manually)?
    - [X ] Are all unit tests still passing? (`npm run test`)
- [X ] Are the changes visual?
  - [X ] Have I provided screenshots/videos of the changes?

### DIFF
--- a/src/phases/move-effect-phase.ts
+++ b/src/phases/move-effect-phase.ts
@@ -151,7 +151,7 @@ export class MoveEffectPhase extends PokemonPhase {
 
           /** Is the pokemon immune due to an ablility, and also not in a semi invulnerable state?  */
           const isImmune = target.hasAbilityWithAttr(TypeImmunityAbAttr) && (target.getAbility()?.getAttrs(TypeImmunityAbAttr)?.[0]?.getImmuneType() === user.getMoveType(move))
-              && !targets[0].getTag(SemiInvulnerableTag);
+              && !target.getTag(SemiInvulnerableTag);
 
           /**
              * If the move missed a target, stop all future hits against that target
--- a/src/phases/move-effect-phase.ts
+++ b/src/phases/move-effect-phase.ts
@@ -99,8 +99,9 @@ export class MoveEffectPhase extends PokemonPhase {
       const targetHitChecks = Object.fromEntries(targets.map(p => [ p.getBattlerIndex(), this.hitCheck(p) ]));
       const hasActiveTargets = targets.some(t => t.isActive(true));
 
-      /** Check if the target is immune via ability to the attacking move */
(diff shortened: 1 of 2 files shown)


Snailman11:
@PrabbyDD Make sure to link your fixes with the bugs




Link to Original Discord Bug Report (aqua step (by quaq) heals wild politoed in bounce. If it can't be hit, shouldn't it not heal?)
https://discord.com/channels/1125469663833370665/1290401734593875969

Tempo-anon:
what if the user or target have no guard?

src/phases/move-effect-phase.ts
```diff
@@ -148,8 +149,9 @@ export class MoveEffectPhase extends PokemonPhase {
               && (hasConditionalProtectApplied.value || (!target.findTags(t => t instanceof DamageProtectedTag).length && target.findTags(t => t instanceof ProtectedTag).find(t => target.lapseTag(t.tagType)))
               || (this.move.getMove().category !== MoveCategory.STATUS && target.findTags(t => t instanceof DamageProtectedTag).find(t => target.lapseTag(t.tagType))));
 
-          /** Is the pokemon immune due to an ablility?  */
-          const isImmune = target.hasAbilityWithAttr(TypeImmunityAbAttr) && (target.getAbility()?.getAttrs(TypeImmunityAbAttr)?.[0]?.getImmuneType() === user.getMoveType(move));
+          /** Is the pokemon immune due to an ablility, and also not in a semi invulnerable state?  */
+          const isImmune = target.hasAbilityWithAttr(TypeImmunityAbAttr) && (target.getAbility()?.getAttrs(TypeImmunityAbAttr)?.[0]?.getImmuneType() === user.getMoveType(move))
+              && !targets[0].getTag(SemiInvulnerableTag);
```
DayKev, line 154: This one uses `target` and not `targets[0]` for the ability attr checks, shouldn't the tag check match that?

PrabbyDD: You are right, will submit an updated version

PrabbyDD:
> what if the user or target have no guard?

No guard seemingly works fine. It bypasses the semi invul state, thus registering a hit, and triggering the ability.

PR #4666, opened 2024-10-16
[UI][QoL] Cursor defaults to Fight at the start of each new wave

frutescens:
Cursor will default to FIGHT at the start of a new wave.

People requested it. 

New conditional that decides on cursor placement. 

### Screenshots/Videos

https://github.com/user-attachments/assets/85bbdae2-bc82-44ff-999d-b9281bd409af

## How to test the changes?
Start new waves through BALL or RUN. Cursor should return to FIGHT. 

## Checklist
- [x] Are the changes visual?


DayKev (dismissed):

Opaque02 (changes requested):
So seems like there's a small issue with this - when going between waves the first turn does move to fight, as expected which is good. If you then do something other than fight, like use a pokeball, if it fails, next turn (turn 2) will then start at fight instead of ball. Using a ball again makes the cursor stay on ball for turn 3+ though, which is good. Here's a video of it in action

https://github.com/user-attachments/assets/71734ed7-12d0-4c9b-816c-de87e594dbbb

I've changed the pokeball catch rate to 0, which makes the ball instantly fail, hence why they just jump out with no animation. You can see turn 1 it starts on fight, then I choose ball and try catch it, which fails. Turn 2 it starts on fight again, and I choose ball. Turn 3 it's on ball and stays that way onwards. The next battle the same thing happens

I did a very minor amount of digging (literally like 30 seconds) - it seems when the ball is chosen, the show() function is run for the command-ui-handler BEFORE turn 2. This means it's still on turn 1, and currentBattle.starter is still false. Because of this, it sets the cursor to fight, so then when turn 2 comes around, it "remembers" the last position of the cursor (which is now fight because it just got set there), and continues from there. I could be wrong - as I said it was literally 30 seconds or less that I looked into it - but it might be a place to start?

frutescens:
Proof of fix :)

https://github.com/user-attachments/assets/bc83b041-a326-49ad-a9ac-48058a6666e6

DayKev (dismissed):

Opaque02 (dismissed):
So things from last video look to be fixed, which is good, but now there's a new issue. It shows up in two places I've found: 

Firstly it seems like if you go to change pokemon mid fight, and choose a pokemon, the pokemon comes out and the fight option is selected, which is correct. However, the first case of this happens if you go to choose a pokemon and then CANCEL instead of selecting a pokemon, the command ui handler is shown again but the cursor is invisible, even though it's still on the pokemon option. Moving it to a valid spot (up/right) shows the cursor again. See below for a video:

https://github.com/user-attachments/assets/2355f988-2f1e-4d94-9143-a3e7be0930f9

The second case of this happens in a similar thing with the fight option. If you choose a move, the next turn the cursor will be on fight which is good, but if you press cancel, the cursor is now invisible again. Once again, moving somewhere valid (right/down) fixes this. Here is a video of this:

https://github.com/user-attachments/assets/4b9976d2-987d-4e51-a6f6-91801de59d50

I haven't breakpoint'd things to try and figure out why but I'm guessing maybe removing the setCursor within the show() now may have an adverse affect if you don't select an action and clear() the handler and then show() it again? Not 100% sure though, this is a complete guess

frutescens:
https://github.com/user-attachments/assets/fd1b8766-fe86-4cc6-bd5d-6da0e2cbe341

DayKev (approved):

frutescens:
Proof of @Opaque02 's cursor fix


https://github.com/user-attachments/assets/546a4a70-05d1-4f8f-a9ce-76a81e6133be

innerthunder (approved):

f-fsantos (approved):

ahofmeister:
I would appreciate, if this is an option (like the shop cursor) to decide whether it is fight or run (or even something else). I tend to run away in higher floors with a runaway mon.

PR #4676, opened 2024-10-16
[Bug Fix] Fix for Speed Boost is Activated on the Turn a Pokemon is Switched In, and When a Pokemon Fails to Escape #4353

### LINKED ISSUE #4353: [Bug] Speed Boost is Activated on the Turn a Pokemon is Switched In, and When a Pokemon Fails to Escape
### Describe the bug

-If you switch pokemon, or are affected by a move that switches a Pokémon out ex. U-turn, Flip-Turn, Teleport, Chilly Reception, Roar etc , and a Pokemon with Speed Boost is selected, Speed Boost will activate on turn end.

-Failing to run away still allows Speed Boost to be activated at the end of the round.

Link to Discord's Bug Report (Speed Boost activating upon being U-Turned in)
https://discord.com/channels/1125469663833370665/1277568529763270679

### Session export file

_No response_

### User data export file

_No response_

### Expected behavior

> Speed Boost will not activate on the turn the Pokémon with the Ability is switched in (even if it gains the Ability after switching in). It will activate if the Pokémon is switched in due to a previous Pokémon fainting.

https://youtu.be/bTuVs_p2qE8?t=318

https://replay.pokemonshowdown.com/gen9customgame-2207035451-k4d5fnnye0m9tygyd3u63mygcyc19hlpw

> Starting from Generation V, Speed Boost will not activate at the end of the turn if the user failed to escape from a wild encounter during that turn.

### Screenshots / Videos

Boost after U-turn

https://github.com/user-attachments/assets/4168c9a0-b1cd-4405-9404-417dcf61356a

Boosts when failed running away

https://github.com/user-attachments/assets/041a81c2-9e58-4a84-956b-f05982dc0c06

Standard Switch-in

https://github.com/user-attachments/assets/22fbb52c-8d3e-47e4-8179-3969ac585987

Other

### If other please specify

gameplay

### Which browser do you use?

Other

### If other please specify

gameplay

### Additional context

_No response_

https://github.com/user-attachments/assets/449ecebf-7b98-40b7-acf2-b192705d6fee

PrabbyDD:
Speed boost will no longer trigger if the pokemon failed to escape or if it was just switched in on any turn via any method except the initial switch in question. 

Fixes #4353 

I have added a couple of variables to determine if a pokemon was just switched in or failed an escape, and modified several phase files in order to achieve the intended fix. 

### Screenshots/Videos

<details>
  <summary>No speed boost on failed run away</summary>

https://github.com/user-attachments/assets/93ca04cd-a3bc-4ee3-b26a-2175d42d8ad2

</details>
<details>
  <summary>No speed boost on switch</summary>

https://github.com/user-attachments/assets/e6fdff1e-1208-4ae2-bbd2-8d8837ffdb2e

</details>
<details>
  <summary>No speed boost on battle start initial switch in</summary>

https://github.com/user-attachments/assets/1518f82c-2f72-4d4e-be5d-242d8b762e9a

</details>
<details>
  <summary>No speed boost on switch from U-Turn</summary>

https://github.com/user-attachments/assets/3445cf02-935c-4391-a09b-0be4a93c8c71

</details>
<details>
  <summary>No speed boost on switch from Teleport</summary>

https://github.com/user-attachments/assets/f597bef4-995b-4ea2-a59d-595ade89eb08

</details>

## How to test the changes?
Override files, and running test files

## Checklist
- [ ] Are the changes visual?


PrabbyDD:
I have not written unit tests yet, I will probably today or tomorrow, I just wanted to put this pr out there because I feel like my code is kinda shady here and im not sure if there might be unintended consequences messing with the turn end and other phase files like this.

src/field/pokemon.ts
```diff
@@ -105,6 +105,8 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
   public pauseEvolutions: boolean;
   public pokerus: boolean;
   public switchOutStatus: boolean;
+  public switchedInThisTurn: boolean;
+  public failedRunAway: boolean;
```
innerthunder, line 109: Put these in `turnData`, please.

src/phases/turn-end-phase.ts
```diff
@@ -37,7 +38,16 @@ export class TurnEndPhase extends FieldPhase {
...
+      if (!pokemon.hasAbility(Abilities.SPEED_BOOST)) {
+        applyPostTurnAbAttrs(PostTurnAbAttr, pokemon);
+      } else if (pokemon.hasAbility(Abilities.SPEED_BOOST) && !pokemon.switchedInThisTurn && !pokemon.failedRunAway) {
+        pokemon.setSwitchedInStatus(false);
+        pokemon.failedRunAway = false;
+        applyPostTurnAbAttrs(PostTurnAbAttr, pokemon);
+      } else {
+        pokemon.failedRunAway = false;
+        pokemon.setSwitchedInStatus(false);
+      }
```
innerthunder, line 50: does this also apply to Moody?

innerthunder (changes requested):

src/phases/turn-end-phase.ts
PrabbyDD: Bulbapedia doesnt saying anything about moody acting the same way as speed boost

DayKev: Based on the Showdown code Moody doesn't act the same as Speed Boost.

src/field/pokemon.ts
PrabbyDD: I am trying to put these variables in TurnData, but am running into an issue where I am trying to set these variables in turndata while turndata is undefined. I am not sure where turnData is being initialized with an actual object, and its happening after I try to access these variables.

PrabbyDD: I guess it would kinda make sense that it wouldnt work, because switching is between turns, so maybe the turn data for the switched in pokemon doesnt exist yet?

src/phases/switch-summon-phase.ts
```diff
@@ -105,6 +105,10 @@ export class SwitchSummonPhase extends SummonPhase {
     const party = this.player ? this.getParty() : this.scene.getEnemyParty();
     const switchedInPokemon = party[this.slotIndex];
     this.lastPokemon = this.getPokemon();
+    if (this.switchType !== SwitchType.INITIAL_SWITCH) {
+      switchedInPokemon.setSwitchedInStatus(true);
+    }
+    this.lastPokemon.setSwitchedInStatus(false);
```
innerthunder: This needs to go in `onEnd`. The superclass `SummonPhase`'s `onEnd` function resets the turn data of the Pokemon switching in, so the switched in Pokemon is only guaranteed to have its turn data defined after `super.onEnd()` is called.

src/field/pokemon.ts
innerthunder: See my other comment below

src/phases/switch-summon-phase.ts
PrabbyDD: I updated the solution - I still had left it in the same spot but tried to save the turn data after the switch summon phase because the turn data wasnt defined in this file because it was before the new pokemon enters, which I guess is before its turn begins.

src/data/ability.ts
```diff
@@ -3594,7 +3594,15 @@ export class PostTurnStatStageChangeAbAttr extends PostTurnAbAttr {
...
-      pokemon.scene.unshiftPhase(new StatStageChangePhase(pokemon.scene, pokemon.getBattlerIndex(), true, this.stats, this.stages));
+      if (!pokemon.hasAbility(Abilities.SPEED_BOOST)) {
+        pokemon.scene.unshiftPhase(new StatStageChangePhase(pokemon.scene, pokemon.getBattlerIndex(), true, this.stats, this.stages));
+        this.showAbility = true;
+      } else if (pokemon.hasAbility(Abilities.SPEED_BOOST) && !pokemon.turnData.switchedInThisTurn && !pokemon.turnData.failedRunAway) {
+        pokemon.scene.unshiftPhase(new StatStageChangePhase(pokemon.scene, pokemon.getBattlerIndex(), true, this.stats, this.stages));
+        this.showAbility = true;
+      } else {
+        this.showAbility = false;
+      }
```
innerthunder:
```suggestion
      if (!pokemon.hasAbility(Abilities.SPEED_BOOST) || (!pokemon.turnData.switchedInThisTurn && !pokemon.turnData.failedRunAway)) {
        if (!simulated) {
           pokemon.scene.unshiftPhase(new StatStageChangePhase(pokemon.scene, pokemon.getBattlerIndex(), true, this.stats, this.stages));
        }
        return true;
      } 
      return false;
```
You should remove the `if (!simulated)` block encapsulating this, too... GitHub is being annoying about me removing it lol

innerthunder (approved):

src/data/ability.ts
```diff
@@ -3594,7 +3594,13 @@ export class PostTurnStatStageChangeAbAttr extends PostTurnAbAttr {
 
   applyPostTurn(pokemon: Pokemon, passive: boolean, simulated: boolean, args: any[]): boolean {
     if (!simulated) {
-      pokemon.scene.unshiftPhase(new StatStageChangePhase(pokemon.scene, pokemon.getBattlerIndex(), true, this.stats, this.stages));
+      if (!pokemon.hasAbility(Abilities.SPEED_BOOST)) {
```
DayKev, line 3597: Seems like this would cause issues if the Pokémon has both Speed Boost and Moody, since you're not checking whether Speed Boost is the ability currently activating but instead just whether the Pokémon has that ability at all.

src/phases/attempt-run-phase.ts
```diff
@@ -14,6 +14,9 @@ export class AttemptRunPhase extends PokemonPhase {
     super(scene, fieldIndex);
   }
 
+  /** For testing purposes: this is to force the pokemon to fail and escape */
+  public forceFailEscape = false;
+
```
DayKev, line 19: I think technically this should be put before the `constructor()` block, not a big deal though.

src/data/ability.ts
PrabbyDD: Im not sure what you mean in this instance - is it possible for a pokemon to have 2 active abilities at the same time?

innerthunder: Yes. `hasAbility` checks the Pokemon's Ability *and* Passive, so you have to account for having 2 different post-turn abilities at the same time

PrabbyDD: So does that mean a pokemon can have both abilities, but only 1 can be active? So that means instead of checking if it has that ability with hasAbility, I need to check if speed boost is active?

innerthunder: For the record, I don't think this breaks anything atm, since Speed Boost is the only ability that uses this attribute. But this could introduce problems with scalability.

Suppose there was another ability that used this attribute, and you have a Pokemon `source` that has that ability and Speed Boost as a passive. When this attribute is applied for the new ability, `source.hasAbility(Abilities.SPEED_BOOST)` would return `true`, so the new ability wouldn't trigger its stat change on the turn `source` is switched in.

The way I see it, you have 2 options to make the code more robust:
1. Make a field in this attribute class (say, `appliesOnSwitch`), set it to `true` for Speed Boost, and check that instead of doing the `hasAbility` checks.
2. Rename this attribute to `SpeedBoostAbAttr` and only have it apply to Speed Boost (meaning you no longer need the `hasAbility` checks). Future abilities with similar effects would have to use another attribute or rewrite this attribute at the time of their implementation.

I think I'd personally prefer option 2 since there really isn't another ability in the game that would use this attribute, but I don't feel strongly about going one way or the other.

PrabbyDD: Should be good now.

src/data/ability.ts
```diff
@@ -3579,22 +3579,26 @@ export class MoodyAbAttr extends PostTurnAbAttr {
   }
 }
 
-export class PostTurnStatStageChangeAbAttr extends PostTurnAbAttr {
+export class SpeedBoostAbAttr extends PostTurnAbAttr {
   private stats: BattleStat[];
   private stages: number;
```
innerthunder, line 3584: These should be removed since we're always using the same values for them now

innerthunder (changes requested):

src/phases/attempt-run-phase.ts
```diff
@@ -10,6 +10,10 @@ import { NewBattlePhase } from "./new-battle-phase";
 import { PokemonPhase } from "./pokemon-phase";
 
 export class AttemptRunPhase extends PokemonPhase {
+
+  /** For testing purposes: this is to force the pokemon to fail and escape */
+  public forceFailEscape = false;
```
innerthunder, line 15: Not a blocker, but this should probably be turned into an override eventually.

innerthunder (dismissed):
Looking good for the most part now. Thanks for bearing with our suggestions and responding quickly 😅

src/phases/attempt-run-phase.ts
PrabbyDD: eventually :P

PrabbyDD:
ahhhh i broke it after doing the merge conflict

innerthunder (approved):

DayKev (approved):

Issue #4667, opened 2024-10-16
[Bug] No Exp Gained after Switching in a Pokemon as the Opponent Faints


Snailman11:
### Describe the bug

Pokemon do not get awarded Exp when switched in as the opponent faints.

Link to Discord's Bug Report (Didn't get exp from boss in daily)
https://discord.com/channels/1125469663833370665/1285054783908216893

### Reproduction

Toxic an Opponent and switch in a new Pokemon as the opponent faints. No Exp will be gained.

### Expected behavior

Exp Should be awarded to the Pokemon that is switched in.

### Screenshots / Videos

![Screenshot (403)](https://github.com/user-attachments/assets/24cdf305-fb61-45c6-aab6-fe018534bafc)

https://github.com/user-attachments/assets/00a7fe56-91df-49b7-88cf-3c704505c5a9

**Toxic on Alolan Vulpix, dies as Bonsly is switched in. Alolan Vulpix uses Imprison on that turn.**

![EXP 1](https://github.com/user-attachments/assets/3782c7a8-feaf-4e22-8caf-415cf4a21469)

![EXP 2](https://github.com/user-attachments/assets/26230af3-c3f6-4eaa-b6ee-b4c791c53cf3)

![EXP 3](https://github.com/user-attachments/assets/fdf31d74-86f8-4d3c-bebe-9ae357c1d5dd)

PR #4687, opened 2024-10-18
[Bug fix] No Exp Gained after Switching in a Pokemon as the Opponent Faints

PrabbyDD:
Fix for #4667 

Will now properly give xp to all participants for this specific case

bug fix

Added in faint-phase a loop that adds the current pokemon to the participants list, which also happens in TurnInitPhase. 
If the battle ends as a pokemon is being switched in (such as with a poisoned enemy), then the next turnInitPhase doesnt happen and the current pokemon is not added. This fixes that. 

### Screenshots/Videos

https://github.com/user-attachments/assets/cd3d4688-43db-4311-9519-bdf3712e7a5d

## How to test the changes?
Override files, browser debugger

## Checklist
- [x] Are the changes visual?


src/phases/faint-phase.ts
```diff
@@ -65,6 +65,15 @@ export class FaintPhase extends PokemonPhase {
       }
     }
 
+    /** In case the current pokemon was just switched in, make sure it is counted as participating in the combat */
+    this.scene.getField().forEach((pokemon, i) => {
```
DayKev, line 69: You could probably use `getPlayerField()` and skip the `isPlayer()` check below.

PrabbyDD: I just did what was done in turn-init-phase. Should I still change it?

DayKev: `TurnInitPhase` also does things to the enemy's pokemon, but here you're only acting on the player's pokemon, so you can just use `getPlayerField()`.

PrabbyDD: I think it should be fixed now.

src/phases/faint-phase.ts
```diff
@@ -65,6 +65,15 @@ export class FaintPhase extends PokemonPhase {
       }
     }
 
+    /** In case the current pokemon was just switched in, make sure it is counted as participating in the combat */
+    this.scene.getPlayerField().forEach((pokemon, i) => {
+      if (pokemon?.isActive()) {
```
innerthunder, line 70:
```suggestion
      if (pokemon?.isActive(true)) {
```

innerthunder (approved):

PR #4689, opened 2024-10-19
[Test] Add a way for tests to force activation of Statuses

DayKev:
N/A

This allows the testing of more interactions with status effects (specifically Paralysis and Freeze).
#4680 needed a way to test the interaction of Frenzy moves and Paralysis.

Added `STATUS_ACTIVATION_OVERRIDE` to `overrides.ts`.
Added the helper function `forceStatusActivation()` in `MoveHelper` to manipulate `STATUS_ACTIVATION_OVERRIDE` per-move.
Added an example test of the Paralysis status effect to demonstrate usage of `forceStatusActivation()`.

## How to test the changes?
`npm run test status-effect`

## Checklist

### DIFF
--- a/src/phases/move-phase.ts
+++ b/src/phases/move-phase.ts
@@ -11,6 +11,7 @@ import { getTerrainBlockMessage } from "#app/data/weather";
 import { MoveUsedEvent } from "#app/events/battle-scene";
 import Pokemon, { MoveResult, PokemonMove, TurnMove } from "#app/field/pokemon";
 import { getPokemonNameWithAffix } from "#app/messages";
+import Overrides from "#app/overrides";
 import { BattlePhase } from "#app/phases/battle-phase";
 import { CommonAnimPhase } from "#app/phases/common-anim-phase";
 import { MoveEffectPhase } from "#app/phases/move-effect-phase";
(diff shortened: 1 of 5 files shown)


src/phases/move-phase.ts
```diff
@@ -168,21 +170,37 @@ export class MovePhase extends BattlePhase {
...
       switch (this.pokemon.status.effect) {
       case StatusEffect.PARALYSIS:
-        if (!this.pokemon.randSeedInt(4)) {
+        if (this.statusActivationOverride === false) {
+          break;
+        }
+        if (!this.pokemon.randSeedInt(4) || this.statusActivationOverride) {
           activated = true;
-          this.cancelled = true;
+          this.cancel();
```
MokaStitcher, line 178: not feeling too strongly about it, but all these `this.cancel()` could be moved line 208 to the `if(activated)` block instead of doing it for each status effect?

MokaStitcher (approved):
lgtm

Out of curiosity, was Sleep excluded from this because there are already ways to know if it will activate via `status.cureTurn`?

src/phases/move-phase.ts
DayKev: Ah good point, that'd probably be better since they're all contingent on the same condition.

DayKev:
> lgtm
> 
> Out of curiosity, was Sleep excluded from this because there are already ways to know if it will activate via `status.cureTurn`?

Sleep doesn't have any elements of RNG to it (once the Pokémon has fallen asleep that is), so there's no need to put in any methods of checking/manipulating RNG for it.

MokaStitcher (approved):

MokaStitcher (approved):

src/phases/move-phase.ts
```diff
@@ -24,6 +24,8 @@ import { StatusEffect } from "#enums/status-effect";
 import i18next from "i18next";
 
 export class MovePhase extends BattlePhase {
+  /** Used for tests, allows paralysis and freeze to be forced to activate */
+  public statusActivationOverride: boolean | null = null;
```
torranx, line 28: why is this a property instead of being defined in `overrides.ts` like all the other overrides? that way, it can also be used for manual testing without going through finding this property

DayKev: Oh good point, I was just thinking from the perspective of the automated tests.

torranx (dismissed):

PR #4693, opened 2024-10-20
[Bug] The pokemon that set an arena hazard shouldn't ignore it

### LINKED ISSUE #935: Arena hazards are incorrectly ignored by the Pokemon that set them
Not sure if it's intended but arena trap doesn't work against teleport, Skorupi also not absorbing poison spikes.

Edit: Block also does not stop teleport.

DayKev:
Arena hazards will properly affect the Pokémon that set them (which can happen due to the hazards being swapped with Court Change, for example). This also means Toxic Spikes will be removed if the Pokémon that set them was a poison type and switches in to them.

Fixes #935 

Removed the `this.sourceId === pokemon.id` check in `ArenaTrapTag.apply()`.

## How to test the changes?
`npm run test toxic_spikes`

## Checklist

### DIFF
--- a/src/data/arena-tag.ts
+++ b/src/data/arena-tag.ts
@@ -626,7 +626,7 @@ export class ArenaTrapTag extends ArenaTag {
    * @returns `true` if this hazard affects the given Pokemon; `false` otherwise.
    */
   override apply(arena: Arena, simulated: boolean, pokemon: Pokemon): boolean {
-    if (this.sourceId === pokemon.id || (this.side === ArenaTagSide.PLAYER) !== pokemon.isPlayer()) {
+    if ((this.side === ArenaTagSide.PLAYER) !== pokemon.isPlayer()) {
       return false;
     }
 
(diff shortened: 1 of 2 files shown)


src/data/arena-tag.ts
```diff
@@ -626,7 +626,7 @@ export class ArenaTrapTag extends ArenaTag {
    * @returns `true` if this hazard affects the given Pokemon; `false` otherwise.
    */
   override apply(arena: Arena, simulated: boolean, pokemon: Pokemon): boolean {
```
Tempo-anon, line 628: Why is this named ArenaTrapTag? Surely something like HazardTag would make more sense?

Tempo-anon (approved):
Why was there a `this.sourceId === pokemon.id` case to begin with? You'll always be targeting the opponent's field and all that did was cause problems for things that reflect hazards like court change and magic bounce

src/data/arena-tag.ts
DayKev: The answer to both questions is "because sam". 🤣

Issue #4706, opened 2024-10-22
[Bug] Defeats from some Mystery Encounters are not displayed in Run History correctly


frutescens:
### Describe the bug

Defeats from some Mystery Encounters are not parsed correctly by the Run History UI. This issue looks through all 22 original Mystery Encounters and checks if their lose conditions are correctly parsed by the UI currently or not. 

### Affected Mystery Encounters
- [x] Trainer MEs <-- Covered by #4716 
- [x] Mysterious Chest <-- Covered by #4716 as well. 
- [ ] Fiery Fallout and Lost at Sea <-- Still drafting an idea

### Reproduction

To reproduce this bug, please refer to the labeled session data attached.

### Expected behavior

The trainer + trainer's team / wild Pokemon that defeated the player.

frutescens:
### Mysterious Challenges
Status: Broken
[Run History File -> mysteriousChallenges.txt](https://github.com/user-attachments/files/17482935/mysteriousChallenges.txt)

Snailman11:
Link to Discord's Bug Report (Run history trainer not showing up)
https://discord.com/channels/1125469663833370665/1298383872358092872
**-Bug Type Superfan**

PR #4716, opened 2024-10-24
[Bug][UI] Fixes to Run History Logging

frutescens:
Run History will reflect the state of held items at the start of the wave, not the end of the wave <-- a popular player request.
Run History will also display losses caused from trainer battles in Mystery Encounters correctly. 
Certain edge case losses will also have proper handling as well. 
Additionally, some Biomes in certain languages are too long for the first line. This causes problems in Session Preview. The line will be moved downwards as a result. 

Addressing issues found here <-- https://github.com/pagefaultgames/pokerogue/issues/4706

New function that gets the desired information to create a sessionSaveData object that fits Run History's needs.

Balance Approved --> Mysterious Chest requires at least two valid Pokemon in the party before spawning. 
Balance Approved (01/11/2024) --> Fiery Fallout + Lost at Sea MEs will not faint Pokemon in the party and will set Pokemon to an HP of 1 instead. 

### Screenshots/Videos
image

## How to test the changes?
- Case 1 - Lose in a trainer battle within a Mystery Encounter
- Case 2 - Lose or use items during the battle and lose the battle
- Case 3 - Check the session preview of a run that is in a Biome with a very long name.

## Checklist
- [x] Are the changes visual?


MokaStitcher:
I'm a bit worried about the biome text being hidden/hard to read with bigger trainer sprites like those in the screenshots of PR #4410 . Can you try and see how it looks?
But if that can work I think it'd be nice if both Wave and Biome text were centered horizontally within the window (yes, it's mostly because i want to save myself work for RTL languages compatibility :p )

src/data/mystery-encounters/utils/encounter-pokemon-utils.ts
```diff
@@ -288,7 +288,10 @@ export function applyDamageToPokemon(scene: BattleScene, pokemon: PlayerPokemon,
   if (damage <= 0) {
     console.warn("Healing pokemon with `applyDamageToPokemon` is not recommended! Please use `applyHealToPokemon` instead.");
   }
-
+  // If a Pokemon would faint from the damage applied, its HP is instead set to 1.
+  if (pokemon.hp - damage <= 0) {
+    damage = pokemon.hp - 1;
+  }
```
PigeonBar, line 294: We might want to add a check for if the Pokemon is already fainted, so we don't accidentally revive it to 1 HP

src/phases/game-over-phase.ts
```diff
@@ -108,7 +114,9 @@ export class GameOverPhase extends BattlePhase {
             this.scene.gameData.gameStats.dailyRunSessionsWon++;
           }
         }
-        this.scene.gameData.saveRunHistory(this.scene, this.scene.gameData.getSessionSaveData(this.scene), this.victory);
+        this.getRunHistoryEntry().then(runHistoryEntry => {
+          this.scene.gameData.saveRunHistory(this.scene, runHistoryEntry, this.victory);
+        });
```
PigeonBar, line 119: Since `getRunHistoryEntry()` is async, we need to be careful so that the rest of the game over process (specifically, `PostGameOverPhase` doesn't delete the session data before `getRunHistoryEntry()` is finished retrieving and processing the session data.

One potential solution would be to keep track of the Promise:
```suggestion
        const runHistoryEntryPromise = this.getRunHistoryEntry().then(runHistoryEntry => {
          this.scene.gameData.saveRunHistory(this.scene, runHistoryEntry, this.victory);
        });
```

Then, replace the `this.end()` later with `runHistoryEntryPromise.then(() => this.end())`.

src/data/mystery-encounters/utils/encounter-pokemon-utils.ts
frutescens: Ah, good point!

frutescens:
image

src/phases/game-over-phase.ts
```diff
@@ -109,7 +115,8 @@ export class GameOverPhase extends BattlePhase {
             this.scene.gameData.gameStats.dailyRunSessionsWon++;
           }
         }
-        this.scene.gameData.saveRunHistory(this.scene, this.scene.gameData.getSessionSaveData(this.scene), this.isVictory);
+
```
MokaStitcher, line 118:
```suggestion
```

src/phases/game-over-phase.ts
```diff
@@ -212,5 +222,34 @@ export class GameOverPhase extends BattlePhase {
       this.firstRibbons.push(getPokemonSpecies(pokemon.species.getRootSpeciesId(forStarter)));
     }
   }
+
+  /**
+   * Retrieves the session's data to log its TBD
+   * @returns
+   */
+  private async getRunHistoryEntry(): Promise<SessionSaveData> {
```
MokaStitcher, line 230: I wonder if there is a better way to do this that would involve not rewriting almost the whole `getSessionSaveData` function, which is something we tried to move away from in the past.
Could we use `this.scene.gameData.getSessionSaveData(this.scene)` like before and just update the things we want to be based on the pre wave data?

MokaStitcher (commented):
nits + attempt at making the text centering a bit clearer (UI coding via Github PR suggestions, what could go wrong?)

Tested the ME stuff, found no issue.
Have not tested the modifier change yet, would like the code for it to not be rewriting so much if possible

flx-sta (dismissed):
image
image

DayKev:
Sitrus proc'd in battle, run history shows both still existing. 👍
image

Will approve after comments addressed/changes made.

src/overrides.ts
```diff
@@ -176,6 +176,7 @@ class DefaultOverrides {
   // -------------------------
 
   /** 1 to 256, set to null to ignore */
+  /** Requirements: STARTING_WAVE_OVERRIDE > 10 && STARTING_BIOME_OVERRIDE !== Biome.TOWN && MYSTERY_ENCOUNTER_RATE_OVERRIDE === 256 */
```
DayKev, line 179:
```suggestion
  /**
   * `1` (almost never) to `256` (always), set to `null` to disable the override
   *
   * Note: Make sure `STARTING_WAVE_OVERRIDE > 10`, otherwise MEs won't trigger
   */
```

src/phases/game-over-phase.ts
frutescens: I would say that having it separate from getSessionSaveData() allows us to know exactly what we are getting, but doing it the way you suggested also allows for the same thing. The reason why I felt it was better to have a separate function was because the issue with run history being unable to depict losses from mystery encounter trainers was from a change in getSessionSaveData()

src/ui/run-info-ui-handler.ts
```diff
@@ -302,8 +303,13 @@ export default class RunInfoUiHandler extends UiHandler {
...
     }
-
-    this.runResultContainer.add(runStatusText);
+    const currentBiomeText = addTextObject(this.scene, -0.5, 0, `${getBiomeName(this.runInfo.arena.biome)}`, TextStyle.WINDOW, { fontSize: "60px" });
+    const runResultWindow = this.runResultContainer.getByName("Run_Result_Window") as Phaser.GameObjects.Image;
+    const windowCenterX = (runResultWindow.getTopRight().x - runResultWindow.x) / 2;
+    currentBiomeText.setPosition(windowCenterX - currentBiomeText.displayWidth / 2, runResultWindow.getBottomCenter().y - 15);
+    runStatusText.setX(windowCenterX - 5 - currentBiomeText.displayWidth / 2);
+    this.runResultContainer.add([ runStatusText, currentBiomeText ]);
+    console.log(windowCenterX);
```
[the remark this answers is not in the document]
frutescens: It does, but when I did that, the words were slightly off center. I'll try again though.

CodeTappert (dismissed):

DayKev (approved):

src/phases/game-over-phase.ts
```diff
@@ -212,5 +221,34 @@ export class GameOverPhase extends BattlePhase {
       this.firstRibbons.push(getPokemonSpecies(pokemon.species.getRootSpeciesId(forStarter)));
     }
   }
+
+  /**
+   * Slightly modified version of {@linkcode GameData.getSessionSaveData}.
+   * @returns A promise containing the {@linkcode SessionSaveData}
+   */
+  private async getRunHistoryEntry(): Promise<SessionSaveData> {
```
DayKev, line 229: Does this need to be async?

src/phases/game-over-phase.ts
DayKev: Something to be looked at in the future. :pikasmile:

src/ui/run-info-ui-handler.ts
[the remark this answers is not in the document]
MokaStitcher: ended up making a PR to your fork for it :p https://github.com/frutescens/pokerogue/pull/12

src/phases/game-over-phase.ts
frutescens: Yes. It takes some time.

DayKev (approved):

MokaStitcher (approved):

PR #4727, opened 2024-10-26
[P3] Prevent Unintended ME Migration Application

xsn34kzx:

N/A

Moving the migration function within `v1_1_0.ts`'s `sessionMigrators` array to `v1_0_4.ts`'s `sessionMigrators`

As is, the Mystery Encounter migration will be repeated once the game version gets bumped to a version above `v1.1.0`; although the migration itself is idempotent, moving it into the `v1.0.4` migration prevents repeated applications as we move to higher versions for correctness/intended migration behavior.

### Screenshots/Videos

N/A

## How to test the changes?

Load up any old session data and ensure it does not crash/stall the game.

## Checklist
- ~[] Have I considered writing automated tests for the issue?~
- ~[] If I have text, did I make it translatable and add a key in the English locale file(s)?~
- ~[] Are the changes visual?~
  - ~[] Have I provided screenshots/videos of the changes?~

### DIFF
--- a/src/system/version_migration/versions/v1_0_4.ts
+++ b/src/system/version_migration/versions/v1_0_4.ts
@@ -1,6 +1,7 @@
 import { SettingKeys } from "../../settings/settings";
 import { AbilityAttr, defaultStarterSpecies, DexAttr, SystemSaveData, SessionSaveData  } from "../../game-data";
 import { allSpecies } from "../../../data/pokemon-species";
+import { CustomPokemonData } from "#app/data/custom-pokemon-data";
 
 export const systemMigrators = [
   /**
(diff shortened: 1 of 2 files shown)


DayKev (approved):

PigeonBar:
I think the main issue is that the main branch can still label save data as v1.0.5 in _local storage_. Here is a screenshot of loading today's Daily Run on pokerogue.net:

20241026 Main v1 0 5

As a result, we would have to make sure that the ME version migration also runs on v1.0.5 files (and probably all v1.0.x files in general, to be safe).

MokaStitcher:
> I think the main issue is that the main branch can still label save data as v1.0.5 in _local storage_. Here is a screenshot of reloading today's Daily Run on pokerogue.net:

Ah that makes sense thank you. If current main uses 1.0.5 then all saves with ME should be labeled 1.0.5, so it should be moved to a converter for that version instead of the 1.0.4 one

src/system/version_migration/versions/v1_0_4.ts
```diff
@@ -131,5 +132,28 @@ export const sessionMigrators = [
...
+      if (pokemon["mysteryEncounterPokemonData"]) {
+        pokemon.customPokemonData = new CustomPokemonData(pokemon["mysteryEncounterPokemonData"]);
+        pokemon["mysteryEncounterPokemonData"] = null;
+      }
+      if (pokemon["fusionMysteryEncounterPokemonData"]) {
+        pokemon.fusionCustomPokemonData = new CustomPokemonData(pokemon["fusionMysteryEncounterPokemonData"]);
+        pokemon["fusionMysteryEncounterPokemonData"] = null;
+      }
+      pokemon.customPokemonData = pokemon.customPokemonData ?? new CustomPokemonData();
+      if (pokemon["natureOverride"] && pokemon["natureOverride"] >= 0) {
```
DayKev, line 153: This will never run if the nature override is `0` because of the first half of the `if`. Should be `!isNullOrUndefined(pokemon["natureOverride"]) && ...` instead.

MokaStitcher: lol you're right, how did we miss that last time

MokaStitcher (approved):

PR #4738, opened 2024-10-27
[Bug] Silvally/Arceus should not change form without its ability

DayKev:
Arceus and Silvally will require their signature abilities to change form.

The behavior is incorrect.

Added a check for the existence of the ability to the form changes.

### Screenshots/Videos

<details>
  <summary>Sky Plate with Huge Power</summary>

image
image
image

</details>

## How to test the changes?
```ts
const overrides = {
  ITEM_REWARD_OVERRIDE: [{ name: "FORM_CHANGE_ITEM", type: FormChangeItem.SKY_PLATE }],
  ABILITY_OVERRIDE: Abilities.PURE_POWER
```
Try to use the plate on Arceus.

## Checklist


Tempo-anon:
What about other Pokemon that change types/forms with ability like Castform?

DayKev:
> What about other Pokemon that change types/forms with ability like Castform?

`new SpeciesFormChange(Species.CASTFORM, "", "sunny", new SpeciesFormChangeWeatherTrigger(Abilities.FORECAST, [ WeatherType.SUNNY, WeatherType.HARSH_SUN ]), true),`
`new SpeciesFormChange(Species.CHERRIM, "overcast", "sunshine", new SpeciesFormChangeWeatherTrigger(Abilities.FLOWER_GIFT, [ WeatherType.SUNNY, WeatherType.HARSH_SUN ]), true),`

frutescens (approved):

MokaStitcher:
This brings some fusion/form change weirdness to the table

1) If we fuse Silvally as 1st part of the fusion, so it loses the ability:
  * the ROMs don't spawn, so that's good.
  * If I had given it a ROM beforehand I'm allowed to disable it, it will devolve, and lose the typing. Is that fine?
  * After that I have the option to reactivate it, but it doesn't do anything
  * If I have another Silvally on the team that makes a ROM spawn as a reward, I can give it to the fusion. Nothing happens. But the item is here and marked as activated (it can have more ROM at a time than one in this "enabled" state)

I wonder if things should be adapted somewhat to prevent giving the item if the Pokemon doesn't have the ability. And to prevent deactivating the item if it doesn't have the ability

2) If we fuse Silvally as 2nd part of the fusion, so it keeps the ability:
  * ROM still doesn't spawn
  * If I had given it a ROM beforehand I'm allowed to disable/reenable it, but nothing happens in both cases
  * if I get a ROM spawn from another Silvally I can't give it to the fusion, so it will never change form (I think that's a known thing for form changes for 2nd part of the fusion though)

I guess most of these are fine if we accept that the 2nd part of the fusion can't change form. I'm not too familiar with the whole fusion forms/evolution issues

DayKev:
> This brings some fusion/form change weirdness to the table
> 
> 1. If we fuse Silvally as 1st part of the fusion, so it loses the ability:
> 
> * the ROMs don't spawn, so that's good.
> * If I had given it a ROM beforehand I'm allowed to disable it, it will devolve, and lose the typing. Is that fine?

I think so? It should probably revert to base type automatically before fusing, not sure how easy that'd be though...

> * After that I have the option to reactivate it, but it doesn't do anything. Again, good
> * If I have another Silvally on the team that makes a ROM spawn as a reward, I can give it to the fusion. Nothing happens. But the item is here and marked as activated (it can have more ROM at a time than one in this "enabled" state)
> 
> I wonder if things should be adapted somewhat to prevent giving the item if the Pokemon doesn't have the ability.

Yeah, maybe the item spawn weight should check for both the species and the ability?

>  And to prevent deactivating the item if it doesn't have the ability

True, might need to look at the party UI code to find out where these are displayed.
 
> 2. If we fuse Silvally as 2nd part of the fusion, so it keeps the ability:
> 
> * ROM still doesn't spawn
> * If I had given it a ROM beforehand I'm allowed to disable/reenable it, but nothing happens in both cases
> * if I get a ROM spawn from another Silvally I can't give it to the fusion, so it will never change form (I think that's a known thing for form changes for 2nd part of the fusion though)
> 
> I guess most of these are fine if we accept that the 2nd part of the fusion can't change form. I'm not too familiar with the whole fusion forms/evolution issues

Yeah I think all of that is the same behavior as before this PR, since currently the ability isn't checked for anywhere, just the species.

> 3. (if I fuse Silvally with itself it also doesn't evolve lol)

Fusions...

Madmadness65 (approved):

damocleas (approved):

PR #4756, opened 2024-10-29
[Refactor] Adding Type, Current HP, Max HP, & Status to updateGameInfo()

ThePsychedelicSeal:
I am looking to create an automatic overlay for streamers that will display basic party information. The script will update on the command phase and at the end of a wave. Altogether it will show:

- Species
- Type
- Level
- Current HP / Max HP
- Status Condition

I have attached a rough idea of what the end product (for my stream) will look like. I will make the code publicly available for users to edit according to their preferences.
stream mockup 1

The average user will not see any changes to gameplay. The info will be accessible through window.gameInfo.

I added updateGameInfo() to the start of the command-phase execution. With this, to grab current HP, I went into pokemon.ts to add a getter/setter. Finally, in battle-scene.ts, I added the titled attributes to gameInfo.

To test, someone will just need to run window.gameInfo in console and the new information will be populated to the party array that already included name and level.

No automated tests, no overrides used, and should be straightforward to test.

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -2965,7 +2965,7 @@ export default class BattleScene extends SceneBase {
       biome: this.currentBattle ? getBiomeName(this.arena.biomeType) : "",
       wave: this.currentBattle?.waveIndex || 0,
       party: this.party ? this.party.map(p => {
-        return { name: p.name, level: p.level };
+        return { name: p.name, type: p.type[0], level: p.level, currentHP: p.hp, maxHP: p.stats[0], status: p.status };
       }) : [],
       modeChain: this.ui?.getModeChain() ?? [],
     };
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -2965,7 +2965,7 @@ export default class BattleScene extends SceneBase {
       biome: this.currentBattle ? getBiomeName(this.arena.biomeType) : "",
       wave: this.currentBattle?.waveIndex || 0,
       party: this.party ? this.party.map(p => {
-        return { name: p.name, type: p.type[0], level: p.level, currentHP: p.hp, maxHP: p.stats[0], status: p.status };
+        return { name: p.name, type: p.getTypes(), level: p.level, currentHP: p.hp, maxHP: p.getMaxHp(), status: p.status };
       }) : [],
       modeChain: this.ui?.getModeChain() ?? [],
     };
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -89,7 +89,13 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
   public exp: integer;
   public levelExp: integer;
   public gender: Gender;
-  public hp: integer;
+  private _hp: number;
+  public get hp() {
+    return this._hp;
+  }
+  public set hp(hp: number) {
+    this._hp = hp;
+  }
   public stats: integer[];
   public ivs: integer[];
   public nature: Nature;
@@ -181,6 +187,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     }
     this.exp = dataSource?.exp || getLevelTotalExp(this.level, species.growthRate);
     this.levelExp = dataSource?.levelExp || 0;
+
     if (dataSource) {
       this.id = dataSource.id;
       this.hp = dataSource.hp;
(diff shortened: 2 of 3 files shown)


src/field/pokemon.ts
```diff
@@ -89,7 +89,13 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
   public exp: integer;
   public levelExp: integer;
   public gender: Gender;
-  public hp: integer;
+  private _hp: number;
```
f-fsantos, line 92: Is there a point in changing this?

f-fsantos (changes requested):

src/field/pokemon.ts
ThePsychedelicSeal: This was a suggestion from @flx on discord. The currentHP addition will not update without the get/set attached to _hp. If not a problem, great, but I can try testing with it as "public" or "integer" if that is what you are asking.

PigeonBar: After trying it myself, I don't think changing hp to use get/set functions is needed anymore for the currentHP to update; the added `this.scene.updateGameInfo();` to `command-phase.ts` should be enough.

ThePsychedelicSeal: You are correct. I just added a new commit that restores the original pokemon.ts.

src/battle-scene.ts
```diff
@@ -2965,7 +2965,7 @@ export default class BattleScene extends SceneBase {
       biome: this.currentBattle ? getBiomeName(this.arena.biomeType) : "",
       wave: this.currentBattle?.waveIndex || 0,
       party: this.party ? this.party.map(p => {
-        return { name: p.name, level: p.level };
+        return { name: p.name, type: p.getTypes(), level: p.level, currentHP: p.hp, maxHP: p.getMaxHp(), status: p.status };
```
MokaStitcher, line 2968:
```suggestion
        return { name: p.name, types: p.getTypes(), level: p.level, currentHP: p.hp, maxHP: p.getMaxHp(), status: p.status };
```

FYI, getTypes() will not return the Pokemon's tera type if applicable.
If it's something you would potentially be interested in getting you should check the optional parameters of the function, or alternatively getTeraType()

src/field/pokemon.ts
```diff
@@ -181,6 +181,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     }
     this.exp = dataSource?.exp || getLevelTotalExp(this.level, species.growthRate);
     this.levelExp = dataSource?.levelExp || 0;
+
```
MokaStitcher, line 184:
```suggestion
```
could just revert this since you're not touching that file anymore

MokaStitcher (commented):
nits

src/field/pokemon.ts
ThePsychedelicSeal: This was resolved in the most recent commit. Thanks!

src/battle-scene.ts
ThePsychedelicSeal: Thank you for the tip. I added getTeraType() and changed "type" to "types". 
This also reminded me I need Form info for Megas and G-Max, so added that as well in the most recent commit.

DayKev:
image

Currently the new fields you added are numbers/etc, did you want human-readable output (ie: "Water" instead of `10` and "Sleep" instead of `4`, etc)?

(Also minor nit, `tera` should be `teraType`).

DayKev:
Also in the future make sure you make a new branch instead of using the `beta` branch for your PR, it messes things up for people trying to test the PR when you do it this way.

ThePsychedelicSeal:
@DayKev if I could display the type/status in a string instead of an array value that would be awesome, but I am unsure how to do it. I will change "tera" to "teraType" however in the next commit.

DayKev:
> @DayKev if I could display the type/status in a string instead of an array value that would be awesome, but I am unsure how to do it. I will change "tera" to "teraType" however in the next commit.

Submitted a PR to change the numbers into strings (so "WATER" instead of `10` etc): https://github.com/ThePsychedelicSeal/pokerogue/pull/1
image

ThePsychedelicSeal:
@DayKev Thank you for the help! I merged the PR.

DayKev (approved):

flx-sta (approved):
lgtm
image

Tempo-anon:
Doesn't affect players so I'm fine with this.

PR #4766, opened 2024-10-31
[Refactor] Create global scene variable

DayKev:
N/A

Currently to access the active scene you need to awkwardly pass it around thorough various functions/etc, and if it's not already being passed to the context you're working in you have to refactor the function/etc to add a scene passaround parameter/whatever and then potentially update every instance of it throughout the code.

All the random `scene` function params, class fields, etc are ~~obliterated~~ removed and replaced with a single global scene variable.

## How to test the changes?
Play the game, run the tests.

## Checklist


src/phases/evolution-phase.ts
```diff
@@ -189,75 +189,75 @@ export class EvolutionPhase extends Phase {
                             }
                           });
 
-                          SoundFade.fadeOut(this.scene, evolutionBgm, 100);
+                          SoundFade.fadeOut(evolutionBgm, 100);
```
MokaStitcher, line 192:
```suggestion
                          SoundFade.fadeOut(gScene, evolutionBgm, 100);
```
The issue seems to be here. Some of the calls to SoundFade.fadeOut kept the scene, but for others it got removed like here and in Pokemon.ts, hatch phase. But it's a Phaser thing and still needs the scene
I can't test too much but changing this file fixed the evolution test. And leaving the hatch scene as it is crashes the game during hatching, so I think it's safe to say that is the problem

src/phases/egg-hatch-phase.ts
```diff
@@ -285,14 +285,14 @@ export class EggHatchPhase extends Phase {
     this.canSkip = false;
     this.hatched = true;
     if (this.evolutionBgm) {
-      SoundFade.fadeOut(this.scene, this.evolutionBgm, Utils.fixedInt(100));
+      SoundFade.fadeOut(this.evolutionBgm, Utils.fixedInt(100));
```
MokaStitcher, line 288:
```suggestion
      SoundFade.fadeOut(gScene, this.evolutionBgm, Utils.fixedInt(100));
```
(see comment in evolution phase)

src/field/pokemon.ts
```diff
@@ -3171,29 +3171,29 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
       duration = Math.min(duration, fusionCry.totalDuration * 1000);
       fusionCry.destroy();
       scene.time.delayedCall(Utils.fixedInt(Math.ceil(duration * 0.4)), () => {
         try {
-          SoundFade.fadeOut(scene, cry, Utils.fixedInt(Math.ceil(duration * 0.2)));
-          fusionCry = this.getFusionSpeciesForm().cry(scene, Object.assign({ seek: Math.max(fusionCry.totalDuration * 0.4, 0) }, soundConfig));
-          SoundFade.fadeIn(scene, fusionCry, Utils.fixedInt(Math.ceil(duration * 0.2)), scene.masterVolume * scene.seVolume, 0);
+          SoundFade.fadeOut(cry, Utils.fixedInt(Math.ceil(duration * 0.2)));
+          fusionCry = this.getFusionSpeciesForm().cry(Object.assign({ seek: Math.max(fusionCry.totalDuration * 0.4, 0) }, soundConfig));
+          SoundFade.fadeIn(fusionCry, Utils.fixedInt(Math.ceil(duration * 0.2)), scene.masterVolume * scene.seVolume, 0);
```
MokaStitcher, line 3196:
```suggestion
          SoundFade.fadeOut(gScene, cry, Utils.fixedInt(Math.ceil(duration * 0.2)));
          fusionCry = this.getFusionSpeciesForm().cry(Object.assign({ seek: Math.max(fusionCry.totalDuration * 0.4, 0) }, soundConfig));
          SoundFade.fadeIn(gScene, fusionCry, Utils.fixedInt(Math.ceil(duration * 0.2)), scene.masterVolume * scene.seVolume, 0);
```
(see comment in evolution phase)

src/field/pokemon.ts
```diff
@@ -3318,9 +3318,9 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
           frameProgress -= frameThreshold;
         }
         if (i === transitionIndex) {
-          SoundFade.fadeOut(this.scene, cry, Utils.fixedInt(Math.ceil((duration / rate) * 0.2)));
-          fusionCry = this.scene.playSound(fusionCryKey, Object.assign({ seek: Math.max(fusionCry.totalDuration * 0.4, 0), rate: rate }));
-          SoundFade.fadeIn(this.scene, fusionCry, Utils.fixedInt(Math.ceil((duration / rate) * 0.2)), this.scene.masterVolume * this.scene.seVolume, 0);
+          SoundFade.fadeOut(cry, Utils.fixedInt(Math.ceil((duration / rate) * 0.2)));
+          fusionCry = gScene.playSound(fusionCryKey, Object.assign({ seek: Math.max(fusionCry.totalDuration * 0.4, 0), rate: rate }));
+          SoundFade.fadeIn(fusionCry, Utils.fixedInt(Math.ceil((duration / rate) * 0.2)), gScene.masterVolume * gScene.seVolume, 0);
```
MokaStitcher, line 3323:
```suggestion
          SoundFade.fadeOut(gScene, cry, Utils.fixedInt(Math.ceil((duration / rate) * 0.2)));
          fusionCry = gScene.playSound(fusionCryKey, Object.assign({ seek: Math.max(fusionCry.totalDuration * 0.4, 0), rate: rate }));
          SoundFade.fadeIn(gScene, fusionCry, Utils.fixedInt(Math.ceil((duration / rate) * 0.2)), gScene.masterVolume * gScene.seVolume, 0);
```
(see comment in evolution phase)

MokaStitcher (commented):
Found the sound issue I think (see comment in evolution phase)
Haven't checked anything else yet. Did the ME tests pass because you skipped the ones that were failing for now?

torranx:
Can you please update the "Why" section on the ff:
- why can you say its easier development?
- what does this solve?
- why is this the approach we made (ie global variable vs using something from phaser's `SceneManager` (I'm not too knowledgeable on phaser but I've seen some scene getters on docs)? 

I agree that it's🍝 to slap the `BattleScene` on everything but I just wanna understand this change doesn't turn out to be the same.

torranx (commented):
Can you please update the "Why" section on the ff:
- why can you say its easier development?
- what does this solve?
- why is this the approach we made (ie global variable vs using something from phaser's `SceneManager` (I'm not too knowledgeable on phaser but I've seen some scene getters on docs)? 

I agree that it's🍝 to slap the `BattleScene` on everything but I just wanna understand this change doesn't turn out to be the same.

src/battle.ts
```diff
@@ -1,4 +1,4 @@
-import BattleScene from "./battle-scene";
+import { gScene } from "#app/battle-scene";
```
Tempo-anon, line 1: not a fan of `gScene` as the name. `globalScene` is more clear, or even just `scene` if it will be the only one.

src/phases/evolution-phase.ts
DayKev: Ah, right, for some reason I was thinking it only needed to be changed in one place since most things still worked after that change.

DayKev:
I sure love how impossible it is to batch apply suggestions to a large PR, thanks GitHub.

src/battle.ts
DayKev: `globalScene` is fine. I didn't want to go with just `scene` to prevent overlap with any remaining compatibility uses of `scene`.

DayKev:
> Can you please update the "Why" section on the ff:

Yeah, I just wanted to get the PR up before sleeping so I could get some feedback (esp on the sound issue).

> * why can you say its easier development?
> * what does this solve?

Currently to access the scene you need to awkwardly pass it around thorough various functions/etc, and if it's not already being passed to the context you're working in you have to refactor the function/etc to add a `scene` passaround parameter/whatever and then potentially update every instance of it throughout the code.

> * why is this the approach we made (ie global variable vs using something from phaser's `SceneManager` (I'm not too knowledgeable on phaser but I've seen some scene getters on docs)?
> 
> I agree that it's🍝 to slap the `BattleScene` on everything but I just wanna understand this change doesn't turn out to be the same.

I'm not too familiar with Phaser's tool(s) that we have access to for this, and their documentation permalinks are broken when using Google as I've just realized when trying to look it up. If you know of a better way that'd be great. I went with what seemed like the simplest method.

DayKev:
> Found the sound issue I think (see comment in evolution phase) Haven't checked anything else yet. Did the ME tests pass because you skipped the ones that were failing for now?

Yeah I disabled them (for now) and marked them with `// TODO: ...`. There was an issue that broke a bunch of tests (since a lot of ME tests are very similar), plus a strange intermittent(?) issue with the `SelectModifierPhase` tests and an issue I don't understand in the "bug superfan" ME test.

General ME tests issue: https://github.com/pagefaultgames/pokerogue/pull/4766/commits/55d1b1402af26ec841189ec077469e40c4dfbace#diff-b25c954eb59b69c9bdf7245abdfcb9428f37a4d2b7c68bcf2177aab1810889d8R129

SMP issue: https://github.com/pagefaultgames/pokerogue/pull/4766/commits/55d1b1402af26ec841189ec077469e40c4dfbace#diff-6dcfc99d6f0b54ca275b4778e9f553ec1ae6cbee7698442de4731ef524409df2R16

"bug superfan" ME issue: https://github.com/pagefaultgames/pokerogue/pull/4766/commits/55d1b1402af26ec841189ec077469e40c4dfbace#diff-dfe138746b004afe72d3f75d913fc5a5c84a9d52879f12a11afb1200c4efd3d9R386

I tested the ME issues in-game and everything worked fine, so it's a test-specific issue, I'm just not sure what's going on.

MokaStitcher (commented):
Stopping for now, but this should fix the bug superfan encounter issue at least (explanation in second comment bc I keep forgetting github doesn't organize them by line)

DayKev:
New issue discovered: async bullshit!
<details>
  <summary>broken stuff</summary>

image
image

https://github.com/user-attachments/assets/66de7ba0-7733-44e6-afb4-b27f7b917c6c

</details>
<details>
  <summary>not all the stuff is broken though!</summary>

image

</details>

Some of the i18n just randomly doesn't load at the right time.

PigeonBar:
Take this with several grains of salt, but:
1. The async malfunctioning seems to be caused by stuff being imported too early. Everything importing `globalScene` seems to cause `battle-scene.ts` to also need to import from `settings.ts` in a way that initializes its various objects before `i18next` is ready.
2. I was able to stop this issue by moving `globalScene` to a new file `global-scene.ts` using VSCode's refactor tools:

```
import BattleScene from "./battle-scene";

// BattleScene.name;

export let globalScene: BattleScene;

export function initGlobalScene(scene: BattleScene): void {
  globalScene = scene;
}
```

3. This solution seems to solve the locales issue by delaying the imports in `battle-scene.ts` until they are truly needed. As further evidence, uncommenting the commented line causes locales to break again.

DayKev:
Huh, I tried the opposite (making sure `battle-scene.ts` didn't import `settings.ts`) but I guess that could do it. I wonder if that'll fix anything else...

DayKev:
Okay I think I found the issue with the `SelectModifierPhase` tests, the way they're mocking the set up is incorrect in some way. Using the standard test setup (`game.classicMode.startBattle([ Species.ABRA, Species.VOLCARONA ]);` etc) instead of whatever they're doing to manually run a SMPhase fixed the tests. Though two of them I couldn't get to work yet because making the UI do things in tests is hell...

DayKev:
Well whatever, the two SMPhase tests I couldn't fix weren't actually testing the thing they were supposed to be testing anyway.

DayKev:
Everything is fixed!
🎉

frutescens (approved):
approving to bite the merge conflict bullet

DayKev:
If only GitHub wasn't dumb and could fix these fake merge conflicts automatically instead of needing me to do it locally.

frutescens (dismissed):
This is needed for cleaning up RNG too.

Madmadness65 (approved):

Issue #4768, opened 2024-10-31
[Bug] Autofilling username and password appears with black text, making it hard to read on the background


NayanaAg:
### Describe the bug

When you type text into the username / password fields to log in, the text is white. However, if you log in and use the autofill after saving your username / password, it appears in black / dark grey instead which is quite hard to read. The font also looks different (smaller? not sure if it's a different font).

### Reproduction

1. launch pokerogue
2. click username field
3. select "autofill" option from dropdown

### Expected behavior

I would expect consistent styling for username and password text regardless of whether you choose to type it manually or use autofill. 

### Screenshots / Videos

Typing normally: 
![Screenshot 2024-10-30 095315](https://github.com/user-attachments/assets/c782f39d-75d8-46b4-b657-27965f0d69fa)

Autofill: ![Screenshot 2024-10-30 094953](https://github.com/user-attachments/assets/2c6e6fe5-5a6b-409e-9d0f-f8768a912712)

Snailman11:
Which Browser and password saver are you using?

NayanaAg:
I've tried it on Chrome & Opera with their default password saver, not a special plugin or anything

PR #4773, opened 2024-11-01
[UI/UX] Autofill login text colour changed to white

NayanaAg:
When a user autofills their username and password on the login page, the text will be white (the same as if you typed the username and password by hand) instead of the browser's default black.

The black text on dark grey is bad for accessibility. It also stops you from being able to confirm if the autofill is correct, especially if you changed your information recently, because you can' actually read what got filled out.

https://github.com/pagefaultgames/pokerogue/issues/4768 

From an accessibility standpoint, white text is much easier to read on the login form background.

Added a new CSS property to index.css for the webkit text fill colour in autofill inputs
(thanks to Swain in the discord for helping me with this!)

(there was no code for this before, so it's an addition as opposed to a modification)

### Screenshots/Videos

Before: [black autofill text in login form](https://postimg.cc/qzR9fp6b)

[](https://postimg.cc/xXNhCgRL)

## How to test the changes?
You can test it by trying an autofill password, I know you can do it by setting the bypassLogin in battle-scene.ts to false but I'm not sure of the steps after that. 

I did not make use of the `src/overrides.ts` file.
I did not introduce any automated tests.
The reviewer will need to use the browser default autofill or some other plugin.

## Checklist
- [x] Are the changes visual?


index.css
```diff
@@ -67,6 +67,11 @@ input:-internal-autofill-selected {
   background-clip: text;
 }
 
+/*So autofill text appears in white instead of default black*/
```
torranx, line 70: nit: comment can be omitted

torranx (commented):
small nit. looks good otherwise

MokaStitcher:
Can you please show how this looks with the legacy theme?
Did you check it in multiple browsers?

torranx:
> Did you check it in multiple browsers?

- [:autofill](https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill#browser_compatibility) has full browser support
- [-webkit-text-fill-color](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-fill-color) has full browser support

NayanaAg:
Oh looks like it doesn't work well with legacy because of the white background, I hadn't tried it before

MokaStitcher:
> Oh looks like it doesn't work well with legacy because of the white background, I hadn't tried it before

yeah I figured. Not sure if there's an easy solution for it atm unfortunately. Except maybe using a middle-of-the-road color.
Although the ideal solution would be to set this color through Phaser, not CSS, like other colors in the game

torranx:
> yeah I figured. Not sure if there's an easy solution for it atm unfortunately. Except maybe using a middle-of-the-road color. Although the ideal solution would be to set this color through Phaser, not CSS, like other colors in the game

I don't think an HTML input text's property can be set from phaser.

One way to fix this is to add a data attribute to the input fields based on the current UI setting and use that to target the element in CSS

MokaStitcher:
> I don't think an HTML input text's property can be set from phaser?

I think it's possible in the general sense, the input element is created through Phaser and we're able to do some HTML stuff to it directly. Or if needed we have access to the document and can retrieve things from there directly. But since in this case we would need to set style based on the autofill pseudo class, it's not as simple as just changing the color

> One way to fix this is to add a data attribute to the input fields based on the current UI setting and use that to target the element in CSS

Yes, this is already what we do to handle which touch buttons get shown in the various UI modes through CSS
https://github.com/pagefaultgames/pokerogue/blob/a70f0860e1c0a88b29b10d64160e2e2c7511b9fa/src/ui/ui.ts#L493

We can probably do something similar in this case

(let's ignore the fact that I EDITED your comment on accident instead of quoting it, my brain is mush. I think I put it back properly 💀 )

NayanaAg:
I'd be happy to edit the CSS if you guys let me know how! I'm not super experienced but I was also wondering if there's a way to target CSS based on which theme the person loads in.

torranx:
> I'd be happy to edit the CSS if you guys let me know how! I'm not super experienced but I was also wondering if there's a way to target CSS based on which theme the person loads in.

Please see discussion above, that should give you an idea how to go about it.

damocleas:
Mirrored at the above.

PR #4728, opened 2024-10-26
Event

f-fsantos:

PR #4775, opened 2024-11-01
[Bug] Take weight into account when getting the tier of a modifier

MokaStitcher:
None / Correct tier displayed for Soothe Bell if it is added as a fixed reward in shop

Event PR #4728 put Soothe Bell in two tiers, Great and Rogue tier. During events Rogue tier weight's is 0 so the item gets treated as Great tier, the rest of the time it is in Rogue tier as before, with weight 0 in Great Tier.
MEs use code that looks for the tier of a given modifier type in the item pool. But this code does not take the weight into account, which makes it think that Soothe Bell is a Great Tier item at all times, even if no event is ongoing.
This has notably caused the Clowning Around ME test to fail
This issue would also cause the Soothe Bell to appear visually as a Great Tier item after the event is done if set as a fixed shop reward by a ME

This PR should address both the issue of events impacting tests and the modifier tier issue, allowing us to leave the pool changes in and to have more items in multiple tiers in the future if we want.

- Mock `TimedEventManager` in tests to disable events
- Update `withTierFromPool` function in `modifier-type` to take in account the modifier weight and skip over tiers where the weight is 0

### Screenshots/Videos
Before, with no event ongoing, if an ME set Soothe Bell as a fixed reward it would appear as Great Tier rarity
image

After, with no event ongoing, it appears as Rogue Tier:
image

## How to test the changes?
Change line 250 of  `clowning-around-encounter.test.ts` to run the test XX times in a row
```it.only.each(Array.from({ length: XX }))("should ...```

`npm run test src/test/mystery-encounter/encounters/clowning-around-encounter.test.ts`

## Checklist
- [ ] There is no overlap with another PR? at least one other PR that disables the Clowning Around ME
- [ ] The PR is self-contained and cannot be split into smaller PRs? eh...
- [ ] Are the changes visual?

### DIFF
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -19,7 +19,7 @@ import { Unlockables } from "#app/system/unlockables";
 import { getVoucherTypeIcon, getVoucherTypeName, VoucherType } from "#app/system/voucher";
 import PartyUiHandler, { PokemonMoveSelectFilter, PokemonSelectFilter } from "#app/ui/party-ui-handler";
 import { getModifierTierTextTint } from "#app/ui/text";
-import { formatMoney, getEnumKeys, getEnumValues, IntegerHolder, NumberHolder, padInt, randSeedInt, randSeedItem } from "#app/utils";
+import { formatMoney, getEnumKeys, getEnumValues, IntegerHolder, isNullOrUndefined, NumberHolder, padInt, randSeedInt, randSeedItem } from "#app/utils";
 import { Abilities } from "#enums/abilities";
 import { BattlerTagType } from "#enums/battler-tag-type";
 import { BerryType } from "#enums/berry-type";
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/modifier/modifier-type.ts [hunk 1]
```diff
@@ -121,18 +121,41 @@ export class ModifierType {
...
+  withTierFromPool(poolType: ModifierPoolType = ModifierPoolType.PLAYER, party?: PlayerPokemon[], rerollCount: number = 0): ModifierType {
+    let defaultTier: undefined | ModifierTier;
     for (const tier of Object.values(getModifierPoolForType(poolType))) {
       for (const modifier of tier) {
         if (this.id === modifier.modifierType.id) {
-          this.tier = modifier.modifierType.tier;
-          return this;
+          let weight: number;
+          if (modifier.weight instanceof Function) {
+            weight = party ? (modifier.weight as Function)(party, rerollCount) : 0;
```
innerthunder, line 139: I figure `as Function` isn't necessary here because of the `instanceof` check above?

innerthunder (approved):
Just 1 nit, otherwise looks good

src/modifier/modifier-type.ts
```diff
@@ -121,18 +121,41 @@ export class ModifierType {
...
+          } else {
+            weight = modifier.weight as number;
+          }
+          if (weight > 0) {
+            this.tier = modifier.modifierType.tier;
+            return this;
+          } else if (!isNullOrUndefined(defaultTier)) {
```
PigeonBar, line 146:
```suggestion
          } else if (isNullOrUndefined(defaultTier)) {
```

src/test/utils/mocks/mockTimedEventManager.ts
```diff
@@ -0,0 +1,21 @@
+import { TimedEventManager } from "#app/timed-event-manager";
+
+/** Mock TimedEventManager so that ongoing events don't impact tests */
+export class MockTimedEventManager extends TimedEventManager {
+  constructor() {
+    super();
+  }
+
```
DayKev, line 8:
```suggestion
```
Empty constructors can be left out.

src/modifier/modifier-type.ts
DayKev, line 139:
```suggestion
            weight = party ? modifier.weight(party, rerollCount) : 0;
```
Yeah.

src/modifier/modifier-type.ts
```diff
@@ -121,18 +121,41 @@ export class ModifierType {
...
     for (const tier of Object.values(getModifierPoolForType(poolType))) {
       for (const modifier of tier) {
         if (this.id === modifier.modifierType.id) {
-          this.tier = modifier.modifierType.tier;
-          return this;
+          let weight: number;
+          if (modifier.weight instanceof Function) {
+            weight = party ? (modifier.weight as Function)(party, rerollCount) : 0;
+          } else {
+            weight = modifier.weight as number;
```
DayKev, line 141:
```suggestion
            weight = modifier.weight;
```
Same here, TS already knows it's a number because of the `if` statement.

src/modifier/modifier-type.ts
MokaStitcher: LOL, thank you 💀

DayKev (approved):

xsn34kzx (approved):

PR #4643, opened 2024-10-11
[Bug] Prevent wild mons fleeing with U-turn, Flip Turn, Volt Switch

Fontbane:
## What are the changes the user will see?

- Fixes a bug or oversight where wild mons will flee after using U-turn, Flip Turn or Volt Switch.

## Why am I making these changes?

- In the main series games, these moves do not allow wild mons to flee. Teleport is the only self-switching move that allows fleeing.

## What are the changes from a developer perspective?

- Adds another check to the ForceSwitchOutAttr

### Screenshots/Videos

- Player using U-turn and switching, wild Weedle using U-turn and not fleeing, wild Rookidee using Teleport and fleeing
LoadSceneDemo7

## Checklist
- [X] **I'm using `beta` as my base branch**
- [X] There is no overlap with another PR?
- [X] The PR is self-contained and cannot be split into smaller PRs?
- [X] Have I provided a clear explanation of the changes?
- [X] Have I considered writing automated tests for the issue?
- [ ] If I have text, did I make it translatable and add a key in the English locale file(s)?
- [X] Have I tested the changes (manually)?
    - [X] Are all unit tests still passing? (`npm run test`)
- [X] Are the changes visual?
  - [X] Have I provided screenshots/videos of the changes?

PR #4777, opened 2024-11-02
[Move][P2] Add "no switch-in" fail condition for Shed Tail and Baton Pass

innerthunder:
Shed Tail and Baton Pass should now correctly fail when used with no available Pokemon in the user's party to switch in

Following up on #4643, which didn't completely fix status moves

`data/move`: Added a new condition for Shed Tail and Baton Pass requiring the user to have at least one available Pokemon in their party and not on the field. Note that `Pokemon.isActive()` also checks for challenge-eligibility, so the new condition will not be met by party Pokemon not allowed in battle.

## How to test the changes?
`npm run test shed_tail` (1 new test)

## Checklist
- [?] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -5839,7 +5839,6 @@ export class ForceSwitchOutAttr extends MoveEffectAttr {
   }
 }
 
-
 export class ChillyReceptionAttr extends ForceSwitchOutAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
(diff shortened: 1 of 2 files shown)


torranx (approved):

DayKev (approved):

PigeonBar:
Parting Shot, and possibly also Chilly Reception, should also not allow wild Pokemon to flee.

Source: [Reddit](https://www.reddit.com/r/PokeMoonSun/comments/72ryx4/how_to_get_parting_shot_on_smeargle/)

PR #4783, opened 2024-11-03
[Move] Fully Implement Round

innerthunder:
Round is now fully implemented with the power boost and turn order manipulation effects.

less (P) moves pikasmile

- `data/move`: 2 new attributes for Round: one for the power boost from previous Round uses, and the other for manipulating turn order
- `field/pokemon`: New `turnData` field to track if the Pokemon is using Round as a followup to another Pokemon's Round.
- `test/moves/round.test`: New unit test checking turn order manipulation and power boost.

### Screenshots/Videos

**Both allies using Round in a double battle:**

https://github.com/user-attachments/assets/6f755b67-530e-47eb-a30c-52d4162e3674

Note that Shuckle moves immediately after Voltorb [despite having the lowest base Speed out of any evolved Pokemon](https://bulbapedia.bulbagarden.net/wiki/List_of_fully_evolved_Pok%C3%A9mon_by_base_stats).

From the logs for both uses of Round in this turn, we can see that the power of the second Round was doubled compared to the first:

image

## How to test the changes?
`npm run test round`

## Checklist
- [x] Are the changes visual?

PS: I did some [testing on Showdown](https://replay.pokemonshowdown.com/gen9nationaldexdoubles-2236675389-h3lu9njwif81cup98cp0ferebchzvoepw) to determine that Round must actually be used (not just selected) in order to apply the combo effect. The turn order manipulation uses a self-targeted `POST_APPLY` trigger, so the current implementation should match Showdown's behavior.

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -4157,6 +4157,60 @@ export class CombinedPledgeStabBoostAttr extends MoveAttr {
   }
 }
 
+/**
+ * Variable Power attribute for {@link https://bulbapedia.bulbagarden.net/wiki/Round_(move) | Round}.
+ * Doubles power if another Pokemon has previously selected Round this turn.
+ * @extends VariablePowerAttr
+ */
+export class RoundPowerAttr extends VariablePowerAttr {
+  override apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    const power = args[0];
+    if (!(power instanceof Utils.NumberHolder)) {
+      return false;
+    }
+
+    if (user.turnData?.joinedRound) {
+      power.value *= 2;
+      return true;
+    }
+    return false;
+  }
+}
+
+/**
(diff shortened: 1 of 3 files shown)


torranx (commented):
Does this account for [Round can now hit Pokémon even if they are behind a substitute.](https://bulbapedia.bulbagarden.net/wiki/Round_(move)#:~:text=Round%20can%20now%20hit%20Pok%C3%A9mon%20even%20if%20they%20are%20behind%20a%20substitute.)?

innerthunder:
> Does this account for [Round can now hit Pokémon even if they are behind a substitute.](https://bulbapedia.bulbagarden.net/wiki/Round_(move)#:~:text=Round%20can%20now%20hit%20Pok%C3%A9mon%20even%20if%20they%20are%20behind%20a%20substitute.)?

Round is sound-based, so it should hit through substitutes (this should be true even in beta now).

torranx (approved):
looks good to me

PR #4789, opened 2024-11-04
[Move][Refactor] Create `Move.getPriority` + Fully Implement Upper Hand

innerthunder:
Upper Hand should now work against attacks boosted by Gale Wings, Triage, etc.

More implementation + consolidating some recurring code for move priority calculations

- `data/move`:
  - New function `Move.getPriority` returns the move's final priority after applying relevant move and ability attributes (e.g. from Grassy Glide or Prankster)
  - Rewrote Upper Hand's condition into a `UpperHandCondition` class such that it now uses the new `getPriority` function and facilitates condition-based effect scoring for future AI plans.
- Rewrote several instances where move priority is independently calculated to use the new `getPriority` function instead.

## How to test the changes?
`npm run test upper_hand` (new)

## Checklist
- [?] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -8,7 +8,7 @@ import { Constructor, NumberHolder } from "#app/utils";
 import * as Utils from "../utils";
 import { WeatherType } from "./weather";
 import { ArenaTagSide, ArenaTrapTag, WeakenMoveTypeTag } from "./arena-tag";
-import { allAbilities, AllyMoveCategoryPowerBoostAbAttr, applyAbAttrs, applyPostAttackAbAttrs, applyPostItemLostAbAttrs, applyPreAttackAbAttrs, applyPreDefendAbAttrs, BlockItemTheftAbAttr, BlockNonDirectDamageAbAttr, BlockOneHitKOAbAttr, BlockRecoilDamageAttr, ConfusionOnStatusEffectAbAttr, FieldMoveTypePowerBoostAbAttr, FieldPreventExplosiveMovesAbAttr, ForceSwitchOutImmunityAbAttr, HealFromBerryUseAbAttr, IgnoreContactAbAttr, IgnoreMoveEffectsAbAttr, IgnoreProtectOnContactAbAttr, InfiltratorAbAttr, MaxMultiHitAbAttr, MoveAbilityBypassAbAttr, MoveEffectChanceMultiplierAbAttr, MoveTypeChangeAbAttr, PostDamageForceSwitchAbAttr, PostItemLostAbAttr, ReverseDrainAbAttr, UncopiableAbilityAbAttr, UnsuppressableAbilityAbAttr, UnswappableAbilityAbAttr, UserFieldMoveTypePowerBoostAbAttr, VariableMovePowerAbAttr, WonderSkinAbAttr } from "./ability";
+import { allAbilities, AllyMoveCategoryPowerBoostAbAttr, applyAbAttrs, applyPostAttackAbAttrs, applyPostItemLostAbAttrs, applyPreAttackAbAttrs, applyPreDefendAbAttrs, BlockItemTheftAbAttr, BlockNonDirectDamageAbAttr, BlockOneHitKOAbAttr, BlockRecoilDamageAttr, ChangeMovePriorityAbAttr, ConfusionOnStatusEffectAbAttr, FieldMoveTypePowerBoostAbAttr, FieldPreventExplosiveMovesAbAttr, ForceSwitchOutImmunityAbAttr, HealFromBerryUseAbAttr, IgnoreContactAbAttr, IgnoreMoveEffectsAbAttr, IgnoreProtectOnContactAbAttr, InfiltratorAbAttr, MaxMultiHitAbAttr, MoveAbilityBypassAbAttr, MoveEffectChanceMultiplierAbAttr, MoveTypeChangeAbAttr, PostDamageForceSwitchAbAttr, PostItemLostAbAttr, ReverseDrainAbAttr, UncopiableAbilityAbAttr, UnsuppressableAbilityAbAttr, UnswappableAbilityAbAttr, UserFieldMoveTypePowerBoostAbAttr, VariableMovePowerAbAttr, WonderSkinAbAttr } from "./ability";
 import { AttackTypeBoosterModifier, BerryModifier, PokemonHeldItemModifier, PokemonMoveAccuracyBoosterModifier, PokemonMultiHitModifier, PreserveBerryModifier } from "../modifier/modifier";
 import { BattlerIndex, BattleType } from "../battle";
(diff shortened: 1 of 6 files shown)


src/data/move.ts
```diff
@@ -7384,6 +7393,21 @@ export class FirstMoveCondition extends MoveCondition {
   }
 }
 
+export class UpperHandCondition extends MoveCondition {
```
Tempo-anon, line 7396: Add some documentation here to explain the condition quickly and easily

src/data/move.ts
```diff
@@ -7384,6 +7393,21 @@ export class FirstMoveCondition extends MoveCondition {
...
+  constructor() {
+    super((user, target, move) => {
+      const targetCommand = user.scene.currentBattle.turnCommands[target.getBattlerIndex()];
+
+      return !!targetCommand
+        && targetCommand.command === Command.FIGHT
+        && !target.turnData.acted
+        && !!targetCommand.move?.move
+        && allMoves[targetCommand.move.move].category !== MoveCategory.STATUS
+        && allMoves[targetCommand.move.move].getPriority(target) > 0;
```
Tempo-anon, line 7406: bulbapedia says the priority condition is being between +1 and +3, though I don't know if that's necessary since upper hand is +3 and another condition is requiring the target to have not moved yet

src/phases/turn-start-phase.ts
```diff
@@ -98,26 +98,22 @@ export class TurnStartPhase extends FieldPhase {
         const aMove = allMoves[aCommand.move!.move];
         const bMove = allMoves[bCommand!.move!.move];
 
-        // The game now considers priority and applies the relevant move and ability attributes
-        const aPriority = new Utils.IntegerHolder(aMove.priority);
-        const bPriority = new Utils.IntegerHolder(bMove.priority);
+        const aUser = this.scene.getField(true).find(p => p.getBattlerIndex() === a)!;
+        const bUser = this.scene.getField(true).find(p => p.getBattlerIndex() === b)!;
```
Tempo-anon, line 102: I know you're not responsible for this, but a[thing] and b[thing] variable naming is ugh, especially when a and b are sorted

src/data/move.ts
innerthunder: Yeah, I think that's just a byproduct of the other condition requiring the target to not have moved. There's no mention of this required priority range on [Smogon](https://www.smogon.com/dex/sv/moves/upper-hand/).

src/phases/turn-start-phase.ts
innerthunder: Idk, it seems like the convention for `sort` functions is to use `a` and `b` for the points of comparison (especially when `a` and `b` are numbers/indexes), and the variable naming just mirrors that. I'll leave it up to another PR to change it :snoremax:

src/phases/turn-start-phase.ts
```diff
@@ -98,26 +98,22 @@ export class TurnStartPhase extends FieldPhase {
...
         // Otherwise, the game returns the user of the move with the highest priority.
-        const isSameBracket = Math.ceil(aPriority.value) - Math.ceil(bPriority.value) === 0;
-        if (aPriority.value !== bPriority.value) {
+        const isSameBracket = Math.ceil(aPriority) - Math.ceil(bPriority) === 0;
+        if (aPriority !== bPriority) {
           if (isSameBracket && battlerBypassSpeed[a].value !== battlerBypassSpeed[b].value) {
             return battlerBypassSpeed[a].value ? -1 : 1;
           }
-          return aPriority.value < bPriority.value ? 1 : -1;
+          return aPriority < bPriority ? 1 : -1;
```
Tempo-anon, line 116: nit: `return (aPriority < bPriority) ? 1 : -1`

src/test/moves/upper_hand.test.ts
```diff
@@ -0,0 +1,84 @@
...
+      .ability(Abilities.BALL_FETCH)
+      .battleType("single")
+      .disableCrits()
+      .enemySpecies(Species.MAGIKARP)
+      .enemyAbility(Abilities.BALL_FETCH)
+      .enemyMoveset(Moves.QUICK_ATTACK)
+      .startingLevel(100)
+      .enemyLevel(100);
+  });
+
```
Tempo-anon, line 36: Should also have test cases for the move failing when
* The opponent does not use a priority attack
* The opponent uses a priority attack and acts before the upper hand

Tempo-anon (commented):
LGTM. Just a minor nit and looking for a few more tests

src/test/moves/upper_hand.test.ts
innerthunder: > * The opponent does not use a priority attack

There are already 2 test for fail conditions: 1 for when the opponent uses a non-priority attack, and the other when the opponent uses a status move (currently Splash, but I'll change it to a priority status move like Baby-Doll Eyes).

> * The opponent uses a priority attack and acts before the upper hand

Adding this one in the next commit 👍

Tempo-anon (approved):

src/test/moves/upper_hand.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    await game.phaseInterceptor.to("BerryPhase");
+
+    expect(feebas.getLastXMoves()[0].result).toBe(MoveResult.SUCCESS);
+    expect(magikarp.isFullHp()).toBeFalsy();
+    expect(feebas.isFullHp()).toBeTruthy();
+  });
+
+  it("should fail if the target has already moved", async () => {
+    game.override.enemyMoveset(Moves.TACKLE);
+    vi.spyOn(allMoves[Moves.TACKLE], "priority", "get").mockReturnValue(4);
```
DayKev, line 88:
```suggestion
    game.override.enemyMoveset(Moves.ENDURE);
```
The protect moves are priority class 4, you can just use one of those.

src/test/moves/upper_hand.test.ts
```diff
@@ -0,0 +1,100 @@
+import { allMoves } from "#app/data/move";
+import { MoveResult } from "#app/field/pokemon";
+import { Abilities } from "#enums/abilities";
+import { Moves } from "#enums/moves";
+import { Species } from "#enums/species";
+import GameManager from "#test/utils/gameManager";
+import Phaser from "phaser";
+import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
```
DayKev, line 8:
```suggestion
import { afterEach, beforeAll, beforeEach, describe, expect, it } from "vitest";
```

src/test/moves/upper_hand.test.ts
innerthunder: I wanted the move to be an attack so that the increased priority would be the only factor in making Upper Hand fail.

DayKev (approved):

PR #4790, opened 2024-11-04
[Feature] Add GROWTH sorting option to dropdown in Starter Select UI

afa7789:

New option at dropdown in starter selection to explore growth rate pokemons

I wanted to sort by growth rate to see which mon`s are faster/slower

Added just a column, and a case in the sorting part

## How to test the changes?

not sure yet.

## Checklist
- [] If I have text, did I make it translatable and add a key in the English locale file(s)?
no I did not. 
- [ ] Are the changes visual?

### DIFF
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -490,6 +490,7 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
       new DropDownOption(this.scene, SortCriteria.CANDY, new DropDownLabel(i18next.t("filterBar:sortByCandies"))),
       new DropDownOption(this.scene, SortCriteria.IV, new DropDownLabel(i18next.t("filterBar:sortByIVs"))),
       new DropDownOption(this.scene, SortCriteria.NAME, new DropDownLabel(i18next.t("filterBar:sortByName")))
+      new DropDownOption(this.scene, SortCriteria.GROWTH, new DropDownLabel(i18next.t("filterBar:sortByGrowth")))
     ];
     this.filterBar.addFilter(DropDownColumn.SORT, i18next.t("filterBar:sortFilter"), new DropDown(this.scene, 0, 0, sortOptions, this.updateStarters, DropDownType.SINGLE));
     this.filterBarContainer.add(this.filterBar);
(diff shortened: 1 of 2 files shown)


MokaStitcher:
Thank you for your contribution

You need to also make a PR to the [translation repo](https://github.com/pagefaultgames/pokerogue-locales) to add the "sortByGrowth" string you want to use. You only need to add it for English. Link this PR in the locale one, and the other way around too.

We also need before and after screenshots or videos for these kind of changes
If you don't know how to run the game locally to test things we can help here or in the dev corner in the official discord

On the PR itself, I'm not entirely sure that sorting by growth rate makes complete sense since it's not as simple as "this Pokemon grows faster than this other one", but I could be convinced

src/ui/starter-select-ui-handler.ts
```diff
@@ -490,6 +490,7 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
       new DropDownOption(this.scene, SortCriteria.CANDY, new DropDownLabel(i18next.t("filterBar:sortByCandies"))),
       new DropDownOption(this.scene, SortCriteria.IV, new DropDownLabel(i18next.t("filterBar:sortByIVs"))),
       new DropDownOption(this.scene, SortCriteria.NAME, new DropDownLabel(i18next.t("filterBar:sortByName")))
```
MokaStitcher, line 492:
```suggestion
      new DropDownOption(this.scene, SortCriteria.NAME, new DropDownLabel(i18next.t("filterBar:sortByName"))),
```
Code fails to compile

afa7789:
Hey Moka.

Im gonna do it later, just need some time due to work.

I dont actually think it matters, but it would be good to have, I wanted to
see which of my mins had a faster growth rate to be a supporter character.

And had to browse one by one so why not right ? I also strongly believe
sorting is great thing to have.

I think we could also have a sorting by egg moves ( number of unlocked eggs
moves ). ( Could be useful ).

Had some trouble locally setting it up, still going over the code base.

innerthunder (changes requested):
I think this would make more sense as a filtering option than as a sorting option since there isn't a clear linear relationship between all growth rates. Sorting could be misleading for the user when it comes to interpreting the meaning of "Erratic" and "Fluctuating" growth rates, for example, compared to other growth rates.

CodeTappert:
I do not really think this is necessary.

afa7789:
Makes sense.
Okay we cant scrap the PR then.

PR #4793, opened 2024-11-05
[Move] Corrosive Gas

muscode13:
Corrosive Gas will be usable in game

Corrosive Gas is currently unimplemented

Added some extra checks to the RemoveHeldItemAttr to make it work correctly when Corrosive Gas is used

### Screenshots/Videos

https://github.com/user-attachments/assets/7f708731-e2ac-4abe-bc75-5a16baff52d7

## How to test the changes?

- Use `npm run test corrosive_gas`
- Use Corrosive Gas in-game

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -2374,8 +2374,10 @@ export class RemoveHeldItemAttr extends MoveEffectAttr {
    * @returns {boolean} True if an item was removed
    */
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    if (!this.berriesOnly && target.isPlayer()) { // "Wild Pokemon cannot knock off Player Pokemon's held items" (See Bulbapedia)
-      return false;
+    if (move.id !== Moves.CORROSIVE_GAS) {
+      if (!this.berriesOnly && target.isPlayer()) { // "Wild Pokemon cannot knock off Player Pokemon's held items" (See Bulbapedia)
+        return false;
+      }
     }
 
     if (move.hitsSubstitute(user, target)) {
@@ -2407,6 +2409,8 @@ export class RemoveHeldItemAttr extends MoveEffectAttr {
 
       if (this.berriesOnly) {
         user.scene.queueMessage(i18next.t("moveTriggers:incineratedItem", { pokemonName: getPokemonNameWithAffix(user), targetName: getPokemonNameWithAffix(target), itemName: removedItem.type.name }));
+      } else if (move.id === Moves.CORROSIVE_GAS) {
+        user.scene.queueMessage(i18next.t("moveTriggers:corrosiveGasItem", { pokemonName: getPokemonNameWithAffix(user), targetName: getPokemonNameWithAffix(target), itemName: removedItem.type.name }));
       } else {
         user.scene.queueMessage(i18next.t("moveTriggers:knockedOffItem", { pokemonName: getPokemonNameWithAffix(user), targetName: getPokemonNameWithAffix(target), itemName: removedItem.type.name }));
       }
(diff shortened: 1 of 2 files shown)


src/test/moves/corrosive_gas.test.ts
```diff
@@ -0,0 +1,110 @@
...
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override
+      .moveset([ Moves.SPLASH, Moves.CORROSIVE_GAS ])
+      .battleType("double")
+      .enemySpecies(Species.MAGIKARP)
+      .enemyMoveset(Moves.SPLASH);
+  });
+
+  it("should remove enemy and player items", async () => {
```
Tempo-anon, line 30: Is this test misnamed? It should not remove the user's items

src/test/moves/corrosive_gas.test.ts
```diff
@@ -0,0 +1,110 @@
+import { Moves } from "#enums/moves";
+import { Species } from "#enums/species";
+import GameManager from "#test/utils/gameManager";
+import Phaser from "phaser";
+import { afterEach, beforeAll, beforeEach, describe, expect, it } from "vitest";
+
+describe("Moves - Corrosive Gas", () => {
```
Tempo-anon, line 7: Add tests for
* Not removing items from sticky hold users
* Not removing items if it hits a substitute
* Should probably not affect form change items
* Items removed cannot be reobtained with recycle/harvest
* Items should come back after the battle

src/data/move.ts
```diff
@@ -10034,7 +10038,7 @@ export function initMoves() {
       .makesContact(false),
     new StatusMove(Moves.CORROSIVE_GAS, Type.POISON, 100, 40, -1, 0, 8)
       .target(MoveTarget.ALL_NEAR_OTHERS)
-      .unimplemented(),
+      .attr(RemoveHeldItemAttr, false),
```
Tempo-anon, line 10041: I think this is slightly different from the other uses of this Attr since Bulbapedia says the items are returned after the battle (even if the affected Pokemon swap out)

src/data/move.ts
```diff
@@ -2374,8 +2374,10 @@ export class RemoveHeldItemAttr extends MoveEffectAttr {
    * @returns {boolean} True if an item was removed
    */
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    if (!this.berriesOnly && target.isPlayer()) { // "Wild Pokemon cannot knock off Player Pokemon's held items" (See Bulbapedia)
-      return false;
+    if (move.id !== Moves.CORROSIVE_GAS) {
+      if (!this.berriesOnly && target.isPlayer()) { // "Wild Pokemon cannot knock off Player Pokemon's held items" (See Bulbapedia)
```
Tempo-anon, line 2378: nit: not sure if I like this extra nested if for a special case. Corrosive gas should be different enough from the other item removal moves (because it returns the items) that it should probably have its own Attr

Tempo-anon (changes requested):

src/test/moves/corrosive_gas.test.ts
muscode13: I'll rename the test but I meant the ally pokemon

innerthunder:
Putting my comments from the Discord here for visibility:
image

Also, seeing that we're going with a custom implementation, you might need to make another locale PR updating Corrosive Gas's move description.

muscode13:
@innerthunder Added the changes you suggested. Let me know what you think! (Also added new videos to show the tint)

innerthunder:
> @innerthunder Added the changes you suggested. Let me know what you think! (Also added new videos to show the tint)

Hmm... would it be possible to darken both the Pokemon icon and the item icon? I think just darkening the item is a bit difficult to notice.

SirzBenjie:
I'm working on updating this PR to work.

SirzBenjie:
I am still working on shepherding this PR, but it is not so easy, and there have been non-trivial changes.
I am going to hold off on until _after_ our modifier rework, as that _should_ make this substantially easier to port.

emdeann:
Marking this as a draft until items rework

PR #4794, opened 2024-11-05
[Move] Implement Grudge

frutescens:
Grudge will now be usable.

Shareholder value up.

New Battler Tag added for Grudge + additional code in FaintPhase to handle Grudge's special effect. 

Localization PR here: [Link](https://github.com/pagefaultgames/pokerogue-locales/pull/45)
Test will be made. 

### Screenshots/Videos

https://github.com/user-attachments/assets/9d3bff04-f2dc-4a5d-81df-b01aecbd10f7

Please note that video was taken pre-localization. 

## How to test the changes?
1. With a faster Pokemon, use Grudge
2. Faint the Pokemon that just used Grudge
3. Profit? 

## Checklist
- [x] Are the changes visual?


src/data/battler-tags.ts
```diff
@@ -2796,6 +2796,36 @@ export class PowerTrickTag extends BattlerTag {
   }
 }
 
+export class GrudgeTag extends BattlerTag {
```
Tempo-anon, line 2799: Needs doc

Tempo-anon (commented):
Probably want tests for:
* Not activating on damage that occurs at end of turn like leech seed, sand/hail, and future sight
* Does not activate against struggle (would that even do anything)?
* Pokemon hit by multiple Pokemon, grudge should set the correct move to 0
* Reviver seed shenanigans

Research required:
* If a Pokémon with Truant uses Grudge then loafs around the next turn, then is defeated before using a move the second turn, will Grudge take effect?

src/data/battler-tags.ts
frutescens: It's in draft ),:

frutescens:
> Research required:
> 
> * If a Pokémon with Truant uses Grudge then loafs around the next turn, then is defeated before using a move the second turn, will Grudge take effect?

Grudge will lapse before then since it only lasts within the turn it is used in.

frutescens:
> Probably want tests for:
> 
> * Not activating on damage that occurs at end of turn like leech seed, sand/hail, and future sight
> * Does not activate against struggle (would that even do anything)?
> * Pokemon hit by multiple Pokemon, grudge should set the correct move to 0
> * Reviver seed shenanigans
> 
> Research required:
> 
> * If a Pokémon with Truant uses Grudge then loafs around the next turn, then is defeated before using a move the second turn, will Grudge take effect?

Right now, Grudge does not activate against Struggle or indirect damage
In addition, Grudge should handle Pokemon getting hit by multiple Pokemon correctly since there is only 1 'sourcePokemon' found in faint-phase.

Grudge mirrors the behavior of Destiny Bond in this case since Reviver Seed will probably cause similar problems.

src/phases/faint-phase.ts
```diff
@@ -33,16 +33,22 @@ export class FaintPhase extends PokemonPhase {
    */
   private destinyTag?: DestinyBondTag;
 
+  /**
+   * Grudge tag belonging to the currently fainting Pokemon, if applicable
+   */
+  private grudgeTag?: GrudgeTag;
+
   /**
    * The source Pokemon that dealt fatal damage and should get KO'd by Destiny Bond, if applicable
```
innerthunder, line 42:
```suggestion
   * The source Pokemon that dealt fatal damage
```

src/data/battler-tags.ts
```diff
@@ -2827,6 +2827,45 @@ export class PowerTrickTag extends BattlerTag {
...
+   * @returns `false` if Grudge activates its effect or lapses
+   */
+  override lapse(pokemon: Pokemon, lapseType: BattlerTagLapseType, sourcePokemon?: Pokemon): boolean {
+    if (lapseType === BattlerTagLapseType.CUSTOM && sourcePokemon) {
+      if (sourcePokemon.isActive() && pokemon.isOpponent(sourcePokemon)) {
+        const lastMove = pokemon.turnData.attacksReceived[0];
+        const lastMoveData = sourcePokemon.getMoveset().find(m => m?.moveId === lastMove.move);
+        if (lastMoveData && lastMove.move !== Moves.STRUGGLE) {
+          lastMoveData.ppUsed = lastMoveData.getMovePp();
+          pokemon.scene.queueMessage(i18next.t("battlerTags:grudgeOnLapse", { pokemonNameWithAffix: getPokemonNameWithAffix(pokemon), moveName: lastMoveData.getName() }));
```
innerthunder, line 2859: nit: I think `battlerTags:XLapse` (instead of `XOnLapse`) is the convention for locale keys in these lapse functions.

innerthunder (commented):
Some minor updates to comments that used to specifically reference Destiny Bond. LGTM otherwise

DayKev (commented):
Seems good, just some nits then it's good to go.
https://github.com/frutescens/pokerogue/pull/10

src/data/battler-tags.ts
frutescens: Fixed and updated both PRs

innerthunder (approved):

PR #4797, opened 2024-11-05
[Misc] Clean up various phases (part 1)

DayKev:
N/A

General code clean up.

- Removed some redundant code.
- Used default parameters and improved typing instead of `!!` and `??`.
- Cleaned up some leftover `strict-null` `TODO`s.
- Replaced `integer` with `number`.
- Replaced namespace `* as Utils` imports with named imports.
- Added some documentation.
- Improved readability.
- Removed unnecessary constructors.

## Checklist


innerthunder (commented):
Can you fix the `Utils` imports in these files?

Madmadness65 (approved):

src/phases/message-phase.ts
```diff
@@ -3,9 +3,9 @@ import { Phase } from "#app/phase";
 
 export class MessagePhase extends Phase {
   private text: string;
-  private callbackDelay: number | null;
-  private prompt: boolean | null;
-  private promptDelay: number | null;
+  private callbackDelay?: number | null;
+  private prompt?: boolean | null;
+  private promptDelay?: number | null;
```
SirzBenjie, line 8: If we're making them optional, why keep the explicit null? Is there a spot where they are intentionally provided with null?

SirzBenjie (approved):

src/phases/message-phase.ts
DayKev: This was just the simplest way to resolve the `!`s. This also matches the params of `BattleScene#queueMessage`:
```ts
  queueMessage(
    message: string,
    callbackDelay?: number | null,
    prompt?: boolean | null,
    promptDelay?: number | null,
    defer?: boolean | null,
  ) {
    const phase = new MessagePhase(message, callbackDelay, prompt, promptDelay);
```

Issue #4744, opened 2024-10-28
[Bug] Freeze Dry Does not Affect Soaked Shedinja


Snailman11:
### Describe the bug

Freeze Dry is not hitting through Wonder Guard, when it should as it hits supereffectively against Wonder Guard under Soak

Link to Discord's Bug Report
https://discord.com/channels/1125469663833370665/1275243963883716608

### Reproduction

Soak a Shedinja and use Freeze Dry

### Expected behavior

https://replay.pokemonshowdown.com/gen9customgame-2232285675-9hrhfwzgfdb5oerp7h9fp72yhsooziopw
It should hit through Wonder Guard

### Screenshots / Videos

https://github.com/user-attachments/assets/533bd438-2563-465d-b635-e1d9df1e17b5

PR #4798, opened 2024-11-05
[Bug] Fixing Incorrect Freeze Dry Interaction With Soaked Wonder Guard Target

geeilhan:

The Player was not able to damage a target with "Wonder Guard" even though the used move forced super effectiveness against the soaked target (forced water type). Now freeze dry damages the target correctly.

I saw the [Issue](https://github.com/pagefaultgames/pokerogue/issues/4744) (#4744 ) and wanted to fix it

Should the Player Pokemon use Freeze Dry (which has the [`WaterSuperEffectTypeMultiplierAttr`](https://github.com/pagefaultgames/pokerogue/blob/13841765f177a2e6626683967457adcfc965b631/src/data/move.ts#L4889-L4903) on a target with the [`NonSuperEffectiveImmunityAbAttr`](https://github.com/pagefaultgames/pokerogue/blob/13841765f177a2e6626683967457adcfc965b631/src/data/ability.ts#L508-L529) it now checks if the target is of Type [Water](https://github.com/pagefaultgames/pokerogue/blob/13841765f177a2e6626683967457adcfc965b631/src/data/type.ts#L13). If that is the case then the move's super effective damage calculation is forced even if ice moves are usually not very effective against water type Pokemons.

### Screenshots/Videos

Before Fix:

https://github.com/user-attachments/assets/dd5fe1eb-35bd-4abd-8c48-88d3f3803d30

After Fix:

https://github.com/user-attachments/assets/13abffa2-ae5e-48d6-8c15-c0c5883a3616

## How to test the changes?

Manual tests:
```
const overrides = {
  OPP_SPECIES_OVERRIDE: Species.SHEDINJA,
  MOVESET_OVERRIDE: [ Moves.SOAK, Moves.FREEZE_DRY, Moves.EXTREME_SPEED ],
  OPP_MOVESET_OVERRIDE: Moves.SPLASH
```

Automated tests: [here](https://github.com/pagefaultgames/pokerogue/commit/42a1d5db06e77246d24b1838a5be8c9c64461c05)

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -7,7 +7,7 @@ import { Weather, WeatherType } from "./weather";
 import { BattlerTag, GroundedTag } from "./battler-tags";
 import { StatusEffect, getNonVolatileStatusEffects, getStatusEffectDescriptor, getStatusEffectHealText } from "./status-effect";
 import { Gender } from "./gender";
-import Move, { AttackMove, MoveCategory, MoveFlags, MoveTarget, FlinchAttr, OneHitKOAttr, HitHealAttr, allMoves, StatusMove, SelfStatusMove, VariablePowerAttr, applyMoveAttrs, IncrementMovePriorityAttr, VariableMoveTypeAttr, RandomMovesetMoveAttr, RandomMoveAttr, NaturePowerAttr, CopyMoveAttr, MoveAttr, MultiHitAttr, SacrificialAttr, SacrificialAttrOnHit, NeutralDamageAgainstFlyingTypeMultiplierAttr, FixedDamageAttr } from "./move";
+import Move, { AttackMove, MoveCategory, MoveFlags, MoveTarget, FlinchAttr, OneHitKOAttr, HitHealAttr, allMoves, StatusMove, SelfStatusMove, VariablePowerAttr, applyMoveAttrs, IncrementMovePriorityAttr, VariableMoveTypeAttr, RandomMovesetMoveAttr, RandomMoveAttr, NaturePowerAttr, CopyMoveAttr, MoveAttr, MultiHitAttr, SacrificialAttr, SacrificialAttrOnHit, NeutralDamageAgainstFlyingTypeMultiplierAttr, FixedDamageAttr, WaterSuperEffectTypeMultiplierAttr } from "./move";
 import { ArenaTagSide, ArenaTrapTag } from "./arena-tag";
 import { BerryModifier, HitHealModifier, PokemonHeldItemModifier } from "../modifier/modifier";
 import { TerrainType } from "./terrain";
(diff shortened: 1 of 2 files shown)


src/data/ability.ts
```diff
@@ -512,6 +512,16 @@ export class NonSuperEffectiveImmunityAbAttr extends TypeImmunityAbAttr {
 
   applyPreDefend(pokemon: Pokemon, passive: boolean, simulated: boolean, attacker: Pokemon, move: Move, cancelled: Utils.BooleanHolder, args: any[]): boolean {
     if (move instanceof AttackMove && pokemon.getAttackTypeEffectiveness(attacker.getMoveType(move), attacker) < 2) {
```
DayKev, line 514: The type multiplier is already being passed via `applyPreDefendAbAttrs(TypeImmunityAbAttr, this, source, move, cancelledHolder, simulated, typeMultiplier);` in `pokemon.ts`, the Wonder Guard `AbAttr` should be using that instead of calculating it independently. That would be the proper fix.
https://github.com/pagefaultgames/pokerogue/blob/13841765f177a2e6626683967457adcfc965b631/src/field/pokemon.ts#L1639

geeilhan: Thanks I will fix it that way. I also felt that my approach was a bit too forced. I kind of get a little bit lost in all those similar sounding functions.

torranx (commented):
would this change warrant the removal of the `edgeCase` in freeze dry?

DayKev:
> would this change warrant the removal of the `edgeCase` in freeze dry?

No, Freeze-Dry itself would need to be refactored.

src/data/ability.ts
geeilhan: The typeMultiplier is given to applyPreDefentAbAttrs correctly (2x) but the parameter is forced to 0 if 

https://github.com/pagefaultgames/pokerogue/blob/13841765f177a2e6626683967457adcfc965b631/src/field/pokemon.ts#L1679

does not return 2. This function only checks the types of attack and defender without considering Attr and special move effects so `getAttackTypeEffectiveness` returns 0.5 instead of 2.

I changed the if condition

https://github.com/pagefaultgames/pokerogue/blob/13841765f177a2e6626683967457adcfc965b631/src/data/ability.ts#L514

so the given parameter in args (typeMultiplier) is used.

geeilhan:
I have added 

https://github.com/geeilhan/pokerogueFork/blob/027579b183f8f853ff96adbe6d2bede66d13b9bd/src/data/ability.ts#L514-L516

instead of using only `const modifierValue = (args[0] as Utils.NumberHolder).value` in case args is empty. This should usually not be the case for `TypeImmunityAbAttr` but I added it just in case but I don't think it is necessary. 
Should I change `modifierValue` back to `const modifierValue = (args[0] as Utils.NumberHolder).value`?

DayKev (approved):

PR #4804, opened 2024-11-06
[Dev] Remove logging for api requests

MokaStitcher:
none

This was added with the API refactor, but may result in people accidentally sharing stuff they shouldn't when trying to troubleshoot

removed log

## Checklist
- [ ] Are the changes visual?

### DIFF
--- a/src/plugins/api/api-base.ts
+++ b/src/plugins/api/api-base.ts
@@ -69,8 +69,6 @@ export abstract class ApiBase {
       "Content-Type": config.headers?.["Content-Type"] ?? "application/json",
     };
 
-    console.log(`Sending ${config.method ?? "GET"} request to: `, this.base + path, config);
-
     return await fetch(this.base + path, config);
   }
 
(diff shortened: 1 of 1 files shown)


DayKev (approved):

flx-sta:
Can we wrap this with ` if (import.meta.env.DEV)` instead?

```ts
if (import.meta.env.DEV) {
  console.log(`Sending ${config.method ?? "GET"} request to: `, this.base + path, config);
}
```

DayKev:
> Can we wrap this with ` if (import.meta.env.DEV)` instead ?
> 
> ```ts
> if (import.meta.env.DEV) {
>   console.log(`Sending ${config.method ?? "GET"} request to: `, this.base + path, config);
> }
> ```

That wouldn't do anything, the log already doesn't run on main and does run on beta and this doesn't change that. It'd need to check if it's running locally instead to stop the log running on beta server.

flx-sta:
> > ....
> 
> That wouldn't do anything, the log already doesn't run on main and does run on beta and this doesn't change that. It'd need to check if it's running locally instead to stop the log running on beta server.

It would. Basically the code will only run on development mode

frutescens (approved):

flx-sta (approved):

DayKev (approved):

Issue #4610, opened 2024-10-07
[Bug] Game crash involving Ditto, Reviver Seed, and Sketch


Fubs:
### Describe the bug

I have a ditto with the impostor ability. The ditto is holding a reviver seed, knows sketch as one of its base moves (unused up until this point). Opposing team includes a G-Max charizard, an aegislash, a zacian, and a cinderace.

Ditto switches transforms into G-Max charizard, and KO's the opposing G-Max charizard. Then opponent sends in zacian. Zacian uses magical torque, which inflicts confusion on the ditto but does not KO it. Ditto KO's the zacian, and opponent switches in aegislash. Ditto KO's aegislash, and opponent switches in cinderace.

At this point, opponent's pokemon is cinderace, and my ditto (transformed into G-Max charizard) is still confused.

The first turn after cinderace comes out, ditto's confusion procs and it hurts itself. The opposing cinderace uses bolt strike and KO's the ditto. Ditto's reviver seed activates and ditto is revived. Ditto is now alive and back in its base form.

The second turn, ditto attempts to use sketch to copy bolt strike. The game writes this glitchy message:
```
Ditto sketched
!
```
And then after selecting "fight" on the next turn, the menu area becomes blank, but my cursor can still move aroud, as if the moves become invisible. Upon selecting a move though, the game crashes to a black screen.

### Reproduction

Set up this situation:
- A ditto that:
  - has the impostor ability
  - is transformed into G-Max charizard
  - is holding a reviver seed
  - is at low hp (enough to get one hit KO'd)
  - is confused

- A cinderace that:
  - knows bolt strike
  - is at low hp
  - is paralyzed
 
On the first turn: have ditto hurt itself from confusion and get KO'd by bolt strike. Then ditto will be revived back to its base form from the reviver seed. On the second turn: ditto uses sketch to copy bolt strike, and the game should glitch out as described in the previous section.

### Expected behavior

The revived ditto should be able to copy bolt strike with sketch without crashing the game.

### Screenshots / Videos

https://github.com/user-attachments/assets/812a8196-0a14-4059-99a4-8fe59f9b49c4

### Session export file

[session.txt](https://github.com/user-attachments/files/17284748/session.txt)

### User data export file

[data.txt](https://github.com/user-attachments/files/17284751/data.txt)

### Additional context

Browser/OS: Firefox on linux

Snailman11:
Prob just need rev seed and sketch, if sketch is trying to sketch the move which KO'd it.

PR #4806, opened 2024-11-06
[Bug] Prevents crash from using Sketch against a lost turn 

frutescens:
Sketch should no longer crash the game when it tries to copy a move from a lost turn. 

A bug a day...
See #4610 

Added some conditional checks to SketchAttr, but more research is needed to be done on what Sketch can copy and what it cannot. Throwing into draft.

Will write a test as well. 

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -6748,7 +6748,7 @@ export class SketchAttr extends MoveEffectAttr {
       return false;
     }
 
-    const targetMove = target.getLastXMoves(target.battleSummonData.turnCount)
+    const targetMove = target.getLastXMoves(target.battleSummonData.turnCount).reverse()
       .find(m => m.move !== Moves.NONE && m.move !== Moves.STRUGGLE && !m.virtual);
     if (!targetMove) {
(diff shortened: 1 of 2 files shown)


PigeonBar:
I've looked into this a bit before (but haven't found the time to write code yet). I think the correct behaviour is that if the target lost a turn (i.e., its move was cancelled), then the Sketch user should sketch the target's most recent move that was not cancelled. For example, in the following screenshot, Magikarp uses Round and then falls asleep for a few turns (while clicking Splash for those sleep turns), and then Squirtle uses Sketch, which sketches Round (instead of sketching Splash or failing):

20241106 Sketch Vs  Cancelled

frutescens:
Yeah, I realized that this is more of a complicated fix than I thought it was, but I think I have an idea on how to tackle it starting tomorrow.

DayKev (approved):

DayKev:
Added moveset checks into the Sketch tests (it was easier to push directly than make a PR lol).

Issue #4289, opened 2024-09-17
[Bug] Gigaton Hammer and Blood Moon can be Selected and fails, when Chosen Successively.


Snailman11:
### Describe the bug

-Gigaton Hammer and Blood Moon can be chosen on the cooldown turn. 
Instead of not being able to be chosen, (selecting them should send them back to the player menu) the moves fail.

-When encored into Gigaton Hammer or Blood Moon, the move fails on the second turn. After the second turn, the user is able to execute Gigaton/Blood instead of being forced to struggle.

-Calling Gigaton or Blood Moon with Sleep talk does not let the user use the move normally the turn after it was summoned.

Link to Discord's Bug Report (blood moon and gigaton hammer can be selected when they shouldnt)
https://discord.com/channels/1125469663833370665/1285167628373069854

### Expected behavior

> Gigaton Hammer inflicts damage. The user cannot select Gigaton Hammer in succession unless the move failed. If the player attempts to use Gigaton Hammer twice in a row, they will be prevented from doing so, and will be allowed to select another move. Once another move is used, Gigaton Hammer can be used again.
> 
> If the user cannot select another move (due to Choice items or having no other usable moves), it will use Struggle every other turn.
> 
> Instruct and Sleep Talk allow Gigaton Hammer to be used twice in a row. Encore allows Gigaton Hammer to be used twice in a row, but on subsequent turns of the Encore the user will be forced to Struggle.
> 

https://youtu.be/SgLDCml8UZY?feature=shared

When Called, Gigaton Hammer and Blood Moon's counter should be separate from normal usage of the move.

https://replay.pokemonshowdown.com/gen9customgame-2203228437-sw7lpgwynfqjgsyjncpemd36pot0r3lpw

### Screenshots / Videos

No message/return to menu if Gigaton is selected successively

https://github.com/user-attachments/assets/25d69758-25f3-485e-8fcb-5bf449a1b37a

Encore does not allow a second usage of the move

https://github.com/user-attachments/assets/340eab93-d512-4262-8fe4-e8139d3024f9

Sleep talk does let it be used repeatedly (correct), but prevents the user from using it normally after waking up.

https://github.com/user-attachments/assets/180dbbaa-5d5e-4d7e-a4c7-dc58f5058b19

### What OS did you observe the bug on?

Other

### If other please specify

gameplay

### Which browser do you use?

Other

### If other please specify

gameplay

### Additional context

[...]

Issue #3503, opened 2024-08-12
List of Partial / Unimplemented Moves and Abilities


frutescens:
Last Updated: 2024-10-04

## Partial Moves

- [ ] Rage #3667 
- [ ] Pursuit #3411 
- [ ] Uproar #2763
- [ ] Secret Power
- [ ] Torment -> needs Uproar to be fully implemented
- [ ] Round
- [ ] Terablast
- [ ] Trick Or Treat
- [ ] Forest's Curse
- [ ] Spectral Thief
- [ ] Sunsteel Strike
- [ ] Moongeist Beam
- [ ] Photon Geyser
- [ ] Stuff Cheeks
- [ ] Dragon Darts
- [ ] Shell Side Arm
- [x] Triple Arrows #4543 
- [ ] Order Up
- [x] Tera Starstorm #4549
- [ ] Tera Blast
- [ ] Upper Hand
- [ ] Sky Drop #1803
- [ ] Freeze-Dry
- [ ] Future Sight #1568
- [ ] Rollout #3717
- [ ] Ice Ball #3717
 
## Unimplemented Moves 
- [ ] Bide
- [ ] Trick https://github.com/pagefaultgames/pokerogue/pull/3367 (might be stale)
- [ ] Recycle
- [ ] Magic Coat
- [ ] Grudge
- [ ] Snatch
- [ ] Natural Gift #3247 
- [ ] Embargo
- [ ] Fling #2200 
- [ ] Power Trick https://github.com/pagefaultgames/pokerogue/pull/2658 (might be stale)
- [ ] Me First
- [ ] Switcheroo
- [ ] Lunar Dance
- [ ] Wonder Room #3394 
- [ ] Telekinesis #4506 
- [ ] Magic Room
- [ ] Ally Switch
- [ ] Quash
- [ ] Bestow
- [ ] Ion Deluge
- [ ] Fairy Lock
- [ ] Powder #3662 
- [ ] Instruct
- [ ] Corrosive Gas

## Partial Abilities
- [ ] Shield Dust
- [ ] Serene Grace
- [ ] Sheer Force
- [ ] Harvest
- [ ] Victory Star
- [ ] Cheek Pouch
- [ ] Sweet Veil
- [ ] Shields Down
- [ ] Power Construct
- [ ] Corrosion
- [ ] Wandering Spirit
- [ ] Neutralizing Gas
- [ ] Protosynthesis
- [ ] Good as Gold
- [ ] Quark Drive
[...]

PR #4813, opened 2024-11-07
[Move] Partially Implement Instruct

Bertie690:
[Locales PR](https://github.com/pagefaultgames/pokerogue-locales/pull/47)
Instruct will actually work and calls the target's last used move correctly.
It currently does not play nice with Gigaton Hammer, Blood Moon and Torment, but those are related to the janky implementation of those moves currently. #4289 
Also pledge jank happens currently.
Also the enemy cannot really use it atm but whatevs

Because I want funny klefki prankster instructing power trick shuckle rock slide. 

Added `RepeatMoveAttr` for moves that force targets to repeat moves themselves. (This is distinct from `CopyMoveAttr` which has the user use a target's previous move)

### Screenshots/Videos
Instruct repeating an ally's rock slide & subsequently failing on enemy Unown due to not having made moves yet:

https://github.com/user-attachments/assets/d4ee43d5-d324-4f5e-824d-76aac5e26ea1

pp after instruction
image
It sorta works on pledges... a bit
https://github.com/user-attachments/assets/60490c80-57bd-43cd-8d0b-66ea511803bb
Instruct fails if used on Instruct/Outrage (Full list of moves on Bulbapedia):
https://github.com/user-attachments/assets/b9a22fa2-ab67-4d1d-bd6b-81b855fa545a
https://github.com/user-attachments/assets/59e6ef89-5a7c-477e-a2d0-469ce402b5c7

## How to test the changes?
Run `npm run test instruct`

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -6572,6 +6572,49 @@ export class CopyMoveAttr extends OverrideMoveEffectAttr {
...
+    if (!movesetMove || movesetMove.virtual) {
+      user.scene.queueMessage(i18next.t("battle:attackFailed"));
+      return false;
+    }
+
+    const moveTargets = getMoveTargets(target, lastMove?.move!);
+    if (!moveTargets.targets.length) {
+      user.scene.queueMessage(i18next.t("battle:attackFailed"));
+      return false;
+    }
```
innerthunder, line 6601: These need to be covered in this attribute's `getCondition` function. There might be some overlap with the `lastMoveCopiableCondition` you already have, but doing this bypasses the usual failure handling in `phases/move-phase.ts`

innerthunder (changes requested):
Get rid of all the changes to overrides, `package-lock.json`, and `package.json`.

Bertie690:
> Get rid of all the changes to overrides, `package-lock.json`, and `package.json`.

Yessiree will do that at home

Bertie690:
Requesting #3503 to be updated with this PR

Bertie690:
I am officially doen now.
Have the other 2 videos on hard drive but can't upload because clipchamp is being a royal !@#?$ ATM.

src/data/move.ts
```diff
@@ -6586,6 +6586,124 @@ export class CopyMoveAttr extends OverrideMoveEffectAttr {
...
+   *
+   * @param user {@linkcode Pokemon} that used the attack
+   * @param target {@linkcode Pokemon} targeted by the attack
+   * @param move {@linkcode Move} being used
+   * @param args N/A
+   * @returns {boolean} true if the move succeeds
+   */
+  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+    // get the last move used (excluding status based failures) as well as the corresponding moveset slot
+    const lastMove = target.getLastXMoves(-1).find(m => m.move !== Moves.NONE);
```
innerthunder, line 6606:
```suggestion
    const lastMove = target.getLastXMoves().find(m => m.move !== Moves.NONE);
```

src/data/move.ts
```diff
@@ -6586,6 +6586,124 @@ export class CopyMoveAttr extends OverrideMoveEffectAttr {
...
+    const movesetMove = target.getMoveset().find(m => m?.moveId === lastMove?.move);
+    const moveTargets = lastMove?.targets;
+
+    user.scene.queueMessage(i18next.t("moveTriggers:instructingMove", {
+      userPokemonName: getPokemonNameWithAffix(user),
+      targetPokemonName: getPokemonNameWithAffix(target)
+    }));
+    target.getMoveQueue().push({ move: lastMove?.move!, targets: moveTargets!, ignorePP: false });
+    target.scene.unshiftPhase(new MovePhase(target.scene, target, moveTargets!, movesetMove!, false, false));
```
innerthunder, line 6615: This seems like it could easily be broken if the queue is somehow already populated. Maybe unshift the move queue entry instead of pushing it?

src/data/move.ts
Bertie690: Correct me if I'm wrong, but doesn't calling `getlastXMoves()` with an empty argument default to only grabbing the singular last move used?
We want to get the most recent move used that isn't a full paralysis, sleep turn, etc. (hence why I made it check em all for safety).

src/data/move.ts
innerthunder: Mirror Move does this for the user itself, which basically means it's safe to assume the move queue is empty when it's used. This is slightly different in that the targeted Pokemon may have a populated move queue when it's Instructed.

I know that a lot of the moves that add to the move queue are already set for Instruct to fail against, but I think changing this to `unshift` for both Instruct and Mirror Move would be the most robust way to implement them.

innerthunder:
Btw, please don't use your fork's `beta` branch for PRs. It makes it hard for us to checkout your branch for playtesting without affecting our own `beta` branch.

The best practice for dev work is to make a new branch for every PR/feature, named differently from any of the existing branches on upstream.

Bertie690:
> Btw, please don't use your fork's `beta` branch for PRs. It makes it hard for us to checkout your branch for playtesting without affecting our own `beta` branch.
> 
> The best practice for dev work is to make a new branch for every PR/feature, named differently from any of the existing branches on upstream.

I completely forgot about this initially. Sorry!

Bertie690:
Closing PR to change commit branch.

PR #4816, opened 2024-11-08
[Move] Implement Ally Switch 

frutescens:
The user can now use Ally Switch in battle. 

Shareholder value up. 

New AllySwitchAttr (child of MoveEffectAttr)
Will add a test. 

Localization PR here: https://github.com/pagefaultgames/pokerogue-locales/pull/46

### Screenshots/Videos

https://github.com/user-attachments/assets/2d4bea48-e642-49b7-8575-a239fac2161c

## How to test the changes?
1. Go into a doubles battle and use Ally Swap. 

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -6748,7 +6748,7 @@ export class SketchAttr extends MoveEffectAttr {
       return false;
     }
 
-    const targetMove = target.getLastXMoves(target.battleSummonData.turnCount)
+    const targetMove = target.getLastXMoves(target.battleSummonData.turnCount).reverse()
       .find(m => m.move !== Moves.NONE && m.move !== Moves.STRUGGLE && !m.virtual);
     if (!targetMove) {
       return false;
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -6748,7 +6748,7 @@ export class SketchAttr extends MoveEffectAttr {
       return false;
     }
 
-    const targetMove = target.getLastXMoves()
+    const targetMove = target.getLastXMoves(target.battleSummonData.turnCount)
       .find(m => m.move !== Moves.NONE && m.move !== Moves.STRUGGLE && !m.virtual);
     if (!targetMove) {
       return false;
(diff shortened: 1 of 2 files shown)


src/data/move.ts
```diff
@@ -6976,6 +6977,34 @@ export class SuppressAbilitiesIfActedAttr extends MoveEffectAttr {
   }
 }
 
+export class AllySwitchAttr extends MoveEffectAttr {
```
Tempo-anon, line 6980: where doc pikaemote

src/data/move.ts
```diff
@@ -9100,8 +9129,9 @@ export function initMoves() {
       .attr(AddArenaTagAttr, ArenaTagType.QUICK_GUARD, 1, true, true)
       .condition(failIfLastCondition),
     new SelfStatusMove(Moves.ALLY_SWITCH, Type.PSYCHIC, -1, 15, -1, 2, 5)
-      .ignoresProtect()
-      .unimplemented(),
+      .attr(AllySwitchAttr)
+      .condition(failIfSingleBattle)
```
Tempo-anon, line 9133: Hey if we have a `failIfSingleBattle` how come it isn't on helping hand as well?

src/test/moves/sketch.test.ts
```diff
@@ -50,4 +52,28 @@ describe("Moves - Sketch", () => {
     expect(playerPokemon?.getLastXMoves()[0].result).toBe(MoveResult.SUCCESS);
     // Can't verify if the player Pokemon's moveset was successfully changed because of overrides.
   });
+
+  it("Sketch should retrieve the most recent valid move from its target history", async () => {
```
Tempo-anon, line 56: this looks like it got wrongly caught up in this PR since ally swap was not used

Tempo-anon (commented):
Looks like in the video that some animations are not correct. When your Pokemon swap places the animation for the dragon dance still plays in the original spot.

Additionally for testing purposes
* There should be a test for using and not using ally switch and seeing if the opponent's moves still target the same ones* 
* Additional uses of ally switch should have a decreased chance of success like Protect
* If a Pokemon targets its ally but then is ally switched the move will fail (including acupressure)

*Bulbapedia says certain interactions (Snipe Shot, Stalwart, Propeller Tail) ignore the ally switch and will still hit the original target.

src/data/move.ts
frutescens: this is in draft.

src/test/moves/sketch.test.ts
frutescens: yeah, but the PR wit those commits should be in soon since its ready

Issue #2675, opened 2024-06-28
[BUG] Unable to catch imposter ditto in spliced mode


3ae3ae:
**Describe the bug**

I have a bug where I can't catch the imposter ditto in spliced mode.
I tried to catch the ditto with the master ball and the game froze the moment I caught it.
It was the same when I tried with a ball other than the master ball.
Tried refreshing the page and the same thing happened.

**To Reproduce**

1. meet the imposter ditto in spliced mode.
2. throw the monster ball at ditto.
3. If he catches it, the game freezes.

**Expected behavior**

The game should not freeze when you catch a ditto.

**Screenshots / Videos**

![Screenshot_20240628_085318_Firefox](https://github.com/pagefaultgames/pokerogue/assets/83870598/adb0cb6b-bc22-481f-9d07-3f8b90374170)

**Device**

Android Firefox

**Additional context**

#2481
#2635
Ditto once transformed into my Reshiram, which has Fire, Flying, Ghost, and Normal-type moves, it killed my Normal/Ghost-type WigglytuffShedinja.
This bug has not been reproduced, but may be related to this.

Issue #2635, opened 2024-06-27
[BUG] Imposters breaks the game


kleenxfeu:
**Describe the bug**
Upon switching a guzzlord spliced with an imposter ditto, the game locks up. It doesn't freeze per say because the sprites are still moving up and down, but I can't do anything.
However, if the switch happens when the promp pops up at the start of a battle, it doesn't freeze. In this case, opening up the summary screen of the transformed species (in this case guzz/ditto transformed into claydol), I have a black screen, the game freezes.

I don't remember it being an issue with a simple ditto.

Snailman11:
Grand Transform Aggregated Issues (Fusion Transform Softlock, Delayed Visual, Form Not Saved)
https://discord.com/channels/1125469663833370665/1267976989957029900/1267976989957029900


This one is the same  @as Issue 2481(Spliced Transform) https://github.com/pagefaultgames/pokerogue/issues/2481

Issue #2481, opened 2024-06-21
[BUG] Spliced Transform


MetallurgyBlockworks:
**Describe the bug**
A lock-up occurs if a spliced Ditto transforms into another Pokemon during a double battle. The battle continues normally, but once the Dittosplice gains experience it locks up.

This bug has been replicated on two different floors transforming into four different Pokemon during double battles.

**To Reproduce**
1. Splice a Shiny 1 Ditto with a Spidops
2. Enter a double battle
3. Transform into either opposing Pokemon
4. Win the battle

**Expected behavior**
The Ditto gains the experience and life moves on.

**Screenshots / Videos**
![image](https://github.com/pagefaultgames/pokerogue/assets/162029849/9946683e-cb7d-4cd3-89b6-3447dbd85d31)

![image](https://github.com/pagefaultgames/pokerogue/assets/162029849/55f3a642-0f0f-4ed3-ba0d-25a34cb57cf4)

![image](https://github.com/pagefaultgames/pokerogue/assets/162029849/815091cc-ae2a-4d6d-9af1-f1331681db45)

![image](https://github.com/pagefaultgames/pokerogue/assets/162029849/27e60215-66f9-4093-951e-61cc238a1a1e)

![image](https://github.com/pagefaultgames/pokerogue/assets/162029849/e33fc8ee-48b8-482a-8f2a-5ef3b7ce942a) (Game is frozen at this screen.)

SivWatt:
Bumped into this bug splicing shiny Ditto with another Pokemon (Solrock #338). 
This will always happen when winning any fight with transformed spliced Ditto in court. 
For me it happened in classic wave 200.

Snailman11:
Grand Transform Aggregated Issues (Fusion Transform Softlock, Delayed Visual, Form Not Saved, Stats not copied(imposter), pp not reduced)
https://discord.com/channels/1125469663833370665/1267976989957029900/1267976989957029900

LoaneShark:
Can confirm that it is still happening, in my case I spliced an imposter ditto into the second slot (first slot is a shiny furret with pickup passive). It was also during a classic mode run.

To be more specific, the game softlocks if any of these 3 things happen happen:
1) You check the summary of the ditto fusion while it is transformed (and on the field)
2) You win a battle / try to change areas while the ditto fusion is transformed (and therefore on the field)
3) You try to teach it a move or encounter an event that checks your moves while it is transformed

[...]

PR #4824, opened 2024-11-08
[Bug] Transform and Imposter will now fail when either Pokemon is fused

DayKev:
Transform and Imposter will fail if the user or the target is a fused Pokémon.

When either the user or target of Transform/Imposter is a fused Pokémon, the game tends to crash. With Imposter it can even break a session save because of an Imposter Pokémon immediately trying to transform when reloading a save.

Transform and Imposter now have checks that cause them to fail if either the user or the target is a fused Pokémon.

## How to test the changes?
Start a session of "Endless (Spliced)" with Imposter Ditto and try to use Transform. To test for the player being fused, use these overrides and fuse Ditto as the second Pokémon onto another Pokémon:
```ts
const overrides = {
  ITEM_REWARD_OVERRIDE: [{ name: "DNA_SPLICERS" }],
  MOVESET_OVERRIDE: [ Moves.TRANSFORM, Moves.SURF ]
```

## Checklist

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -2463,12 +2463,15 @@ export class PostSummonCopyAllyStatsAbAttr extends PostSummonAbAttr {
   }
 }
 
+/**
+ * Used by Imposter
+ */
 export class PostSummonTransformAbAttr extends PostSummonAbAttr {
   constructor() {
     super(true);
   }
 
-  async applyPostSummon(pokemon: Pokemon, passive: boolean, simulated: boolean, args: any[]): Promise<boolean> {
+  async applyPostSummon(pokemon: Pokemon, _passive: boolean, simulated: boolean, _args: any[]): Promise<boolean> {
     const targets = pokemon.getOpponents();
     if (simulated || !targets.length) {
       return simulated;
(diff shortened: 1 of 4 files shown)


src/data/move.ts
```diff
@@ -6986,18 +6986,25 @@ export class SuppressAbilitiesIfActedAttr extends MoveEffectAttr {
...
+/**
+ * Used by Transform
+ */
 export class TransformAttr extends MoveEffectAttr {
   async apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): Promise<boolean> {
     if (!super.apply(user, target, move, args)) {
       return false;
     }
 
+    // transforming from or into fusion pokemon causes various problems (including crashes and save corruption)
```
Tempo-anon, line 6998: I was thinking the condition could just be added where the move is defined.
```ts
new StatusMove(Moves.TRANSFORM, Type.NORMAL, -1, 10, -1, 0, 1)
      .attr(TransformAttr)
      .condition((user, target, move) => !target.getTag(BattlerTagType.SUBSTITUTE && !user.fusionSpecies && target.fusionSpecies)) // Disables transforming from or into fused Pokemon because it causes various sprite issues leading to freezing
      .ignoresProtect(),
```

src/data/ability.ts
```diff
@@ -2475,17 +2478,29 @@ export class PostSummonTransformAbAttr extends PostSummonAbAttr {
...
+      pokemon.scene.executeWithSeedOffset(() => {
+        if (targets[0].fusionSpecies) {
+          target = targets[1];
+          return;
+        } else if (targets[1].fusionSpecies) {
+          target = targets[0];
+          return;
+        }
+        target = Utils.randSeedItem(targets);
+      }, pokemon.scene.currentBattle.waveIndex);
```
Tempo-anon, line 2490: A comment would be good here since at a glance this is not too clear. Essentially what this does is that if there are any fused Pokemon in a double battle it automatically grabs the other one. If both are fused then it grabs targets[1] but then later on line 2497 it will fail

Tempo-anon (commented):
Code looks fine to me though I had a few nits. Has this been tested? There should be some uploaded session data files that can be used for testing

src/data/move.ts
DayKev: Ah, that might be better.

frutescens:
Do we know why Transform/Imposter on fused Pokemon cause such issues?

DayKev:
> Do we know why Transform/Imposter on fused Pokemon cause such issues?

🤷
Probably because fusions are a mess of 🍝 
There's some refactoring I wanted to do with fusions to simplify things but I haven't gotten around to it yet.

src/data/ability.ts
DayKev: Comment added. 👍

DayKev:
> Has this been tested

Yep, I tested both cases (user and target fused) with both Transform and Imposter.

Snailman11:
Band-Aid Fix for
https://github.com/pagefaultgames/pokerogue/issues/2635
 which is the same issue as
https://github.com/pagefaultgames/pokerogue/issues/2481
and 
https://github.com/pagefaultgames/pokerogue/issues/2675

DayKev:
Added a check to prevent Ditto from being generated as part of a fusion (for "Endless (Spliced)" mode mainly).

frutescens (approved):

PR #4840, opened 2024-11-10
[Move] Fully implement Freeze Dry

geeilhan:
Freeze-Dry now does the correct amount of damage.

While I was working on the Bug Fix of Freeze-Dry interacting with a soaked wonder guard target I noticed that "Freeze-Dry" was not fully implemented and its interaction with "Electrify" as well as "Normalize" was faulty.
(#3503)

If the Player Pkm gets hit by `Electrify` before its attack the type of `Freeze-Dry`  changes to electric. As a result the damage multiplier should change according to the [Wiki](https://bulbapedia.bulbagarden.net/wiki/Freeze-Dry_(move)).
If the move is of type electric it naturally has a super effectiveness against water/flying type pokemon (e.g.Gyarados). After already having a 4x multiplier against Gyarados the [`WaterSuperEffectTypeMultiplierAttr`](https://github.com/pagefaultgames/pokerogue/blob/2968059814dd714a1a855006adc6016b1f081a1b/src/data/move.ts#L4941-L4955) was forcefully applied even though the multiplier was correct which resulted in a 16x damage multiplier at the end of the calculation. 
A similar thing happened with `Freeze-Dry` under the effect of `Normalize`.

I have changed `WaterSuperEffectTypeMultiplierAttr` so it considers if `Normalize` is in effect or if the move has been affected by `Electrify`.

### Screenshots/Videos
Note: If you look at the right hand side of the recording you can see the mulitplier value through `console.log`

<details>
  <summary>Before Changes Electrify + Freeze Dry (Multiplier should be 4x but is 16x after calculation)</summary>

https://github.com/user-attachments/assets/c4c19396-cd43-46a0-8eaf-dbda3ad378a0

</details>
<details>
  <summary>Before Changes Normalize + Freeze Dry (Multiplier should be 2x but is 4x)</summary>

https://github.com/user-attachments/assets/6d41e0bb-6172-44e4-9a6c-8dfc63452a21

</details>
<details>
  <summary>After Changes Electrify + Freeze Dry</summary>

https://github.com/user-attachments/assets/a5fb177c-e516-44fc-aab7-2d8b573e7d82

</details>
<details>
  <summary>After Changes Normalize + Freeze Dry</summary>

https://github.com/user-attachments/assets/de826c00-45dc-42bc-9a8a-65fec492d9c4

</details>

## How to test the changes?
I used this to test Normalize:
```
const overrides = {
  MOVESET_OVERRIDE: [ Moves.FREEZE_DRY, Moves.ELECTRIFY, Moves.SEED_BOMB, Moves.SOAK ],
  ABILITY_OVERRIDE: Abilities.NORMALIZE,
  OPP_MOVESET_OVERRIDE: Moves.SPLASH ,
  OPP_SPECIES_OVERRIDE: Species.GYARADOS,
  OPP_ABILITY_OVERRIDE: Abilities.BALL_FETCH,
  OPP_LEVEL_OVERRIDE: 10,
```
as well as this for Electriy:
```
const overrides = {
  MOVESET_OVERRIDE: [ Moves.FREEZE_DRY, Moves.ELECTRIFY, Moves.SEED_BOMB, Moves.SOAK ],
  OPP_MOVESET_OVERRIDE: Moves.ELECTRIFY ,
  OPP_SPECIES_OVERRIDE: Species.GYARADOS,
  OPP_ABILITY_OVERRIDE: Abilities.BALL_FETCH,
  OPP_LEVEL_OVERRIDE: 10,
```

Also some automated tests have been added to `src/test/moves/freeze_dry.test.ts` in order to test 0.25x dmg multiplier as well as 4x dmg

## Checklist
- [x] Are the changes visual?


src/data/move.ts
```diff
@@ -4945,7 +4945,22 @@ export class WaterSuperEffectTypeMultiplierAttr extends VariableMoveTypeMultipli
       const effectivenessAgainstWater = new Utils.NumberHolder(getTypeDamageMultiplier(move.type, Type.WATER));
       applyChallenges(user.scene.gameMode, ChallengeType.TYPE_EFFECTIVENESS, effectivenessAgainstWater);
       if (effectivenessAgainstWater.value !== 0) {
-        multiplier.value *= 2 / effectivenessAgainstWater.value;
+        /**
+         * During Normalize the given multiplier value against water Pkm is 1x but should be 2x.
+         *
+         * For more information about special interactions visit [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Freeze-Dry_(move))
+         */
+        if (user.getAbility().id === Abilities.NORMALIZE) {
```
Tempo-anon, line 4953: Shouldn't this be a hasAbility check in case the user has a passive?

src/test/moves/freeze_dry.test.ts
Tempo-anon, line 1: Is there a test for inverse battles in here? Bulbapedia says the move should do 2x against water even during inverse battles

src/data/move.ts
```diff
@@ -4945,7 +4945,22 @@ export class WaterSuperEffectTypeMultiplierAttr extends VariableMoveTypeMultipli
       const effectivenessAgainstWater = new Utils.NumberHolder(getTypeDamageMultiplier(move.type, Type.WATER));
       applyChallenges(user.scene.gameMode, ChallengeType.TYPE_EFFECTIVENESS, effectivenessAgainstWater);
       if (effectivenessAgainstWater.value !== 0) {
-        multiplier.value *= 2 / effectivenessAgainstWater.value;
```
Tempo-anon, line 4948: nit: Since this attr is only used for a single move, I'd be okay with renaming it and also adding documentation to it

```ts
/**
 * Attribute used for Freeze Dry that deals super effective damage against water even during inverse or if the move's type is changed
 */
export class FreezeDryAttr extends VariableMoveTypeMultiplierAttr
```

src/data/move.ts
innerthunder: Yeah, that'd also check if the ability is suppressed (or ignored, but that's not really relevant here)

geeilhan: Thanks will change this to hasAbility

src/test/moves/freeze_dry.test.ts
geeilhan: I did not know inverse battles are implemented in here. I will have to revise the code in case of inverse battles

geeilhan: The super effectiveness against water type works for the inverse battle but not for the other types. Will need a little bit more time to figure that out.

src/data/move.ts
```diff
@@ -4934,14 +4934,29 @@ export class NeutralDamageAgainstFlyingTypeMultiplierAttr extends VariableMoveTy
   }
 }
 
-export class WaterSuperEffectTypeMultiplierAttr extends VariableMoveTypeMultiplierAttr {
+export class FreezeDryAttr extends VariableMoveTypeMultiplierAttr {
```
Tempo-anon, line 4937: nit: Would also like an overview TSdoc on the class, even if it's a quick 1 sentence summary

src/data/move.ts
```diff
@@ -4934,14 +4934,29 @@ export class NeutralDamageAgainstFlyingTypeMultiplierAttr extends VariableMoveTy
...
+         */
+        if (user.hasAbility(Abilities.NORMALIZE)) {
+          multiplier.value *= 2;
+          return true;
+        }
+
+        /**
+         * If move is of type electric it already has super effectiveness against water types and multiplier does not need to recalculate its value.
+         */
+        if (user.getMoveType(move) !== Type.ELECTRIC) {
```
Tempo-anon, line 4957: nit: I know this isn't a thing, but would it make sense for this to check for electric and grass for future proofing purposes?

src/test/moves/freeze_dry.test.ts
```diff
@@ -126,4 +141,55 @@ describe("Moves - Freeze-Dry", () => {
...
+    vi.spyOn(enemy, "getMoveEffectiveness");
+
+    game.move.select(Moves.FREEZE_DRY);
+    await game.setTurnOrder([ BattlerIndex.ENEMY, BattlerIndex.PLAYER ]);
+    await game.phaseInterceptor.to("BerryPhase");
+
+    expect(enemy.getMoveEffectiveness).toHaveReturnedWith(0.25);
+  });
+
+  it("should deal 2x damage to Water type during inverse battle", async () => {
```
Tempo-anon, line 177: Can there also be some tests for normalize/electrify interactions under inverse?

innerthunder (commented):
I feel like there's a cleaner way to do this. The Normalize and Electrify interactions kind of suggest that Freeze Dry directly sets the type multiplier to 2 against the Water type (while the normal type matchup still applies for the target's other types). The way it behaves in inverse battles seems to suggest the same. 

I'm not totally sure if this is actually the case and, if so, how well the attribute's parent class supports this, but I'd rather avoid the individual ability checks if possible.

src/data/move.ts
geeilhan: I used to have grass type in here too but when I was thinking back the wiki only shows special interaction between freeze-dry and normalize/electrify so I removed the Type.GRASS check

geeilhan: I can still change it so it also checks for GRASS type. No incorrect behaviour happens if I do so.

geeilhan:
> I feel like there's a cleaner way to do this. The Normalize and Electrify interactions kind of suggest that Freeze Dry directly sets the type multiplier to 2 against the Water type (while the normal type matchup still applies for the target's other types). The way it behaves in inverse battles seems to suggest the same.
> 
> I'm not totally sure if this is actually the case and, if so, how well the attribute's parent class supports this, but I'd rather avoid the individual ability checks if possible.

The other classes that have `VariableMoveTypeMultiplierAttr` as a parent also force values like this. This seems to be a characteristic of this class. Either way it looks like l can't go around checking the conditions like this. 
Moving this check somewhere else would make `VariableMoveTypeMultiplierAttr` children classes obsolete.

src/data/move.ts
geeilhan: Added documentation to `FreezeDryAttr` and `apply()`

src/test/moves/freeze_dry.test.ts
geeilhan: Added basic test cases for these edge cases.

src/data/move.ts
```diff
@@ -4970,16 +4970,42 @@ export class NeutralDamageAgainstFlyingTypeMultiplierAttr extends VariableMoveTy
...
-    if (target.isOfType(Type.WATER)) {
-      const effectivenessAgainstWater = new Utils.NumberHolder(getTypeDamageMultiplier(move.type, Type.WATER));
-      applyChallenges(user.scene.gameMode, ChallengeType.TYPE_EFFECTIVENESS, effectivenessAgainstWater);
-      if (effectivenessAgainstWater.value !== 0) {
-        multiplier.value *= 2 / effectivenessAgainstWater.value;
+    if (target.isOfType(Type.WATER) && multiplier.value !== 0) {
+      const multipleTypes = new Utils.BooleanHolder(target.getTypes().length > 1);
+
+      if (multipleTypes) {
+        const effectivenessAgainstTarget = new Utils.NumberHolder(getTypeDamageMultiplier(move.type, target.getTypes().filter(types => types !== Type.WATER)[0]));
```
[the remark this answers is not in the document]
geeilhan: Thanks for review

src/data/move.ts
```diff
@@ -4971,16 +4971,42 @@ export class NeutralDamageAgainstFlyingTypeMultiplierAttr extends VariableMoveTy
...
-      const effectivenessAgainstWater = new Utils.NumberHolder(getTypeDamageMultiplier(move.type, Type.WATER));
-      applyChallenges(user.scene.gameMode, ChallengeType.TYPE_EFFECTIVENESS, effectivenessAgainstWater);
-      if (effectivenessAgainstWater.value !== 0) {
-        multiplier.value *= 2 / effectivenessAgainstWater.value;
+    if (target.isOfType(Type.WATER) && multiplier.value !== 0) {
+      const multipleTypes = (target.getTypes().length > 1);
+
+      if (multipleTypes) {
+        const nonWaterType = target.getTypes().filter(type => type !== Type.WATER)[0];
+        const effectivenessAgainstTarget = new Utils.NumberHolder(getTypeDamageMultiplier(move.type, nonWaterType));
```
innerthunder, line 5000:
```suggestion
        const effectivenessAgainstTarget = new Utils.NumberHolder(getTypeDamageMultiplier(user.getMoveType(move), nonWaterType));
```
obtains the move's type after applying Normalize, etc.

src/test/moves/freeze_dry.test.ts
```diff
@@ -126,4 +141,112 @@ describe("Moves - Freeze-Dry", () => {
 
     expect(enemy.getMoveEffectiveness).toHaveReturnedWith(2);
   });
+
+  it("should deal 4x damage to water/flying type under Electrify", async () => {
```
innerthunder, line 145: This might cause false positives since Electric and Ice have the same effectiveness against Flying. It wouldn't catch what I suggested a fix for above.

Maybe another test for "should deal 0x damage to Water/Ghost (i.e. Jellicent) under Normalize" would help? Alternatively, you could test for Water/Ground types and Electrify...

innerthunder (changes requested):

src/test/moves/freeze_dry.test.ts
geeilhan: added both test cases (Water/Ghost under Normalize and Water/Ground under Electrify)

geeilhan: Thanks for the review!

innerthunder (approved):

DayKev (approved):

PR #4842, opened 2024-11-10
[QoL][Sprite] Set default fps to 10 instead of 12 on pokemon anim config 

pom-eranian:
- All the official sprites were, at some point in the past, slowed down considerably by duplicating all frames.  At the same time some unofficial sprites move too fast. Lowering the framerate slightly +  deleting those extra frames will get a more consistent style as a result.
- It matches the aseprite default frame duration, so the preview will by default match the final result, as it should.

The impressive improvement of rendering 2 less frames per second per pokemon. Tweens may vary.

### Screenshots/Videos
https://github.com/user-attachments/assets/bed25921-8b5e-41e8-902a-3cf1b99c2420

## Checklist
~~- [ ] Have I considered writing automated tests for the issue?~~
- [x] Are the changes visual?


Tempo-anon (commented):
Are there any other places where frameRate will need to be changed? https://github.com/search?q=repo%3Apagefaultgames%2Fpokerogue%20framerate&type=code

Will any sprite sheets need updating to account for this change?

MokaStitcher:
This code seems to only apply to Player Pokemon, or am I misunderstanding?

https://github.com/pagefaultgames/pokerogue/blob/073f85c8cb332cb65c0a7223f6e338e3151e1ed1/src/field/pokemon.ts#L420

DayKev (approved):

innerthunder (approved):

PR #4847, opened 2024-11-11
[UI] Make ME text legible when using the legacy UI theme

MokaStitcher:
Readable text in MEs when using legacy theme

Being able to read the game's text is good

Default theme is unchanged
Legacy theme had the following changes, only affecting MEs:
- white text to black
- money text from pale yellow to gold
- option with fulfilled requirement from green to gold

Implementation:
- `encounter-dialogue-utils.getEncounterText` now properly uses the current ui theme when calling `getTextColor`
- `text.ts`:
  - added new TextStyles `ME_OPTION_DEFAULT` and `ME_OPTION_SPECIAL` for ME option colors
  - Added new TextStyle `MONEY_WINDOW`. Currently the default style for Money is a pale yellow in both themes. That doesn't work when the money has to be displayed in a game window since the themes have different colored windows (dark in default theme, white in legacy theme)
  - Currently the MONEY TextStyle is applied automatically by i18n based on the localized string. Changing the style of all money values in MEs would mean having to update a lot of locale strings, for something that will most likely need to change again with the future UI rework. The TextStyles are a mess, the scope of this PR is not to fix that but only to make MEs work in legacy theme. So, for now, `getTextColor` was changed to take a parameter of whether the text is to get displayed within a game window, in which case it replaces the MONEY textstyle with MONEY_WINDOW
- `run-info-ui-handler.ts` also displays money in a window, making it hard to read in legacy theme. It now uses the MONEY_WINDOW TextStyle as well

### Screenshots/Videos

Before:
before_legacy before_default

After:
after_legacy after_default

## How to test the changes?
Overrides:
Default Overrides to get MEs every wave
```
  STARTING_WAVE_OVERRIDE: 11,
  MYSTERY_ENCOUNTER_RATE_OVERRIDE: 255,
```

Example specific override for the one in the screenshots:
```
  ABILITY_OVERRIDE: Abilities.INTIMIDATE,
  MYSTERY_ENCOUNTER_OVERRIDE: MysteryEncounterType.AN_OFFER_YOU_CANT_REFUSE
```

## Checklist
- [X] Are the changes visual?

### DIFF
--- a/src/data/mystery-encounters/utils/encounter-dialogue-utils.ts
+++ b/src/data/mystery-encounters/utils/encounter-dialogue-utils.ts
@@ -7,22 +7,22 @@ import i18next from "i18next";
 /**
  * Will inject all relevant dialogue tokens that exist in the {@linkcode BattleScene.currentBattle.mysteryEncounter.dialogueTokens}, into i18n text.
  * Also adds BBCodeText fragments for colored text, if applicable
- * @param scene
  * @param keyOrString
  * @param primaryStyle Can define a text style to be applied to the entire string. Must be defined for BBCodeText styles to be applied correctly
- * @param uiTheme
  */
-export function getEncounterText(scene: BattleScene, keyOrString?: string, primaryStyle?: TextStyle, uiTheme: UiTheme = UiTheme.DEFAULT): string | null {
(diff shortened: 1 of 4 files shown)


flx-sta (approved):
image

PR #4839, opened 2024-11-10
[Bug] Added `failIfSingleBattle` condtion to Doubles-only moves

frutescens:
## What are the changes the user will see?
Helping Hand and other NearAlly moves should show a failure message.

## Why am I making these changes?
I don't know either.
Note to self: Link bug report too

## What are the changes from a developer perspective?
The declarations of Helping Hand and other doubles-only moves (I don't know, it's late) now has the failIfSingleCondtional

### Screenshots/Videos

https://github.com/user-attachments/assets/e25450de-04ea-41b8-af95-a5a4fa6d4e53

## How to test the changes?
Force a Pokemon to use Helping Hand in Single battle. Failure message should appear. 

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Are the changes visual?
  - [x] Have I provided screenshots/videos of the changes?

PR #4848, opened 2024-11-11
[Beta][Bug] Move Phase not Ending Correctly 

frutescens:
Move Phase should end correctly if the move fails. 

My bug...

Removed the conditional requiring the PP Used to be greater than the move's max PP. 

### Screenshots/Videos
Situation 1: 

https://github.com/user-attachments/assets/1949709a-4c22-43c9-ad33-5bff3522c799

Situation 2:

https://github.com/user-attachments/assets/ae164a1c-a0ec-4b04-b951-947fe9395833

## How to test the changes?
Situation 1:
1. Faint a Pokemon
2. Progress through the next waves as normal.

Situation 2:
1. Attempt to run away from a double battle with a Pokemon that has the ability run away, but have the first Pokemon use a move. 

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/phases/move-phase.ts
+++ b/src/phases/move-phase.ts
@@ -120,10 +120,12 @@ export class MovePhase extends BattlePhase {
     console.log(Moves[this.move.moveId]);
 
     // Check if move is unusable (e.g. because it's out of PP due to a mid-turn Spite).
-    if (!this.canMove(true) && (this.pokemon.isActive(true) || this.move.ppUsed >= this.move.getMovePp())) {
-      this.fail();
-      this.showMoveText();
-      this.showFailedText();
+    if (!this.canMove(true)) {
+      if (this.pokemon.isActive(true)) {
+        this.fail();
+        this.showMoveText();
+        this.showFailedText();
+      }
       return this.end();
     }
 


MokaStitcher:
What's the reason behind the PP check removal?
In the original PR it was changed from && to ||
And now fully removed, is it because it's not necessary anymore to check that here? The comment specifically refers to pp stuff too so if it's not needed we might as well remove it

frutescens:
> What's the reason behind the PP check removal? In the original PR it was changed from && to || And now fully removed, is it because it's not necessary anymore to check that here? The comment specifically refers to pp stuff too so if it's not needed we might as well remove it

It was because the failure check from before specifically checked if the Pokemon was attempting to use a charge move, but had the charge move's PP decrease to 0 during its "charging state" from Spite so it would print out the failure message in that only specific situation. However, the charge move would be unable to be used (move.canUse() -> false) and move phase would end regardless, so I decided that it would be better to remove the check as a whole.

MokaStitcher:
For reference. Was introduced in PR #4839 
Related bug reports on discord [here](https://discord.com/channels/1125469663833370665/1305619397091655720) and [here](https://discord.com/channels/1125469663833370665/1305607125879488677)

PR #4852, opened 2024-11-12
[Misc] Per-Session Biome Tracking

frutescens:
None, ideally.

Necessary groundwork for Run History 2.0 and Biome Rework. 

biomeTracker, an object, has been added to session data. 
It is appended onto from newArena() in battle-scene

Problem: Fails to grab this.scene.currentBattle when starting a run because it isn't initialized yet + leads to the game freezing on boot up. Temporary workaround for Daily Mode though. 

## How to test the changes?
Recommend logging the results on console and setting waves per biome change in Endless to 2 or 3. 

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -10,7 +10,7 @@ import { initCommonAnims, initMoveAnim, loadCommonAnimAssets, loadMoveAnimAssets
 import { Phase } from "#app/phase";
 import { initGameSpeed } from "#app/system/game-speed";
 import { Arena, ArenaBase } from "#app/field/arena";
-import { GameData } from "#app/system/game-data";
+import { GameData, BiomeSessionData } from "#app/system/game-data";
 import { addTextObject, getTextColor, TextStyle } from "#app/ui/text";
 import { allMoves } from "#app/data/move";
 import { MusicPreference } from "#app/system/settings/settings";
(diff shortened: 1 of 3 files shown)


src/battle-scene.ts
```diff
@@ -1273,7 +1274,7 @@ export default class BattleScene extends SceneBase {
     if (!waveIndex && lastBattle) {
       const isWaveIndexMultipleOfTen = !(lastBattle.waveIndex % 10);
       const isEndlessOrDaily = this.gameMode.hasShortBiomes || this.gameMode.isDaily;
-      const isEndlessFifthWave = this.gameMode.hasShortBiomes && (lastBattle.waveIndex % 5) === 0;
+      const isEndlessFifthWave = this.gameMode.hasShortBiomes && (lastBattle.waveIndex % 2) === 0;
```
Tempo-anon, line 1277: The variable name no longer matches the code

src/battle-scene.ts
```diff
@@ -1323,7 +1324,22 @@ export default class BattleScene extends SceneBase {
     return this.currentBattle;
   }
 
-  newArena(biome: Biome): Arena {
+  newArena(biome: Biome, isReset: boolean = true): Arena {
```
Tempo-anon, line 1327: Need documentation for `newArena` , also not sure about the name `isReset`. Maybe something like `userHasReset`?

Also as a quick sanity check, the only times a user should go into a new biome is:
* Every 10 waves in classic/daily mode
* The teleporter ME in classic mode
* Special boss wave 50 in daily mode
* Special boss wave every 50 floors in endless mode
* Every 5 floors otherwise in endless mode

frutescens: newArena() is called to generate the title phase background. 
Yes.. it's spaghetti.

src/battle-scene.ts
frutescens: forgot to revert after testing.

PR #4853, opened 2024-11-12
[P3][Beta] Fix failed charge moves not displaying failed text

PigeonBar:

If a charge move fails, the game will correctly display "But it failed!"

#4839 introduced more bugs than we expected.

Charge moves no longer check for `move.getFailedText()` to return a defined value in order to display the failure text. (My understanding is that `move.getFailedText()` is intended to check for custom failure text, and it returns `null` if the move should use the default failure text of "But it failed!".)

For reference, here is the code for `showFailedText()`:
https://github.com/pagefaultgames/pokerogue/blob/4802f512ff7925037ecb90bf4d19505d1831d2a4/src/phases/move-phase.ts#L550-L552

### Screenshots/Videos

<details>
<summary> Before the changes: Charge move silently fails </summary>

https://github.com/user-attachments/assets/139ee999-4a3d-4038-a4b0-11d0c970ed09

</details>
<details>
<summary> After the changes: Charge move displays "But it failed!" </summary>

https://github.com/user-attachments/assets/6885a035-2a93-401d-a75c-813c5e5c0c34

</details>

## How to test the changes?

Fail to use a charge move (e.g., use Sky Drop against a Pokemon with a Substitute, or use Fly while Gravity is active.)

## Checklist
- [x] Are the changes visual?

### DIFF
--- a/src/phases/move-phase.ts
+++ b/src/phases/move-phase.ts
@@ -378,10 +378,8 @@ export class MovePhase extends BattlePhase {
       this.pokemon.pushMoveHistory({ move: this.move.moveId, targets: this.targets, result: MoveResult.FAIL, virtual: this.move.virtual });
 
       const failureMessage = move.getFailedText(this.pokemon, targets[0], move, new BooleanHolder(false));
-      if (failureMessage) {
-        this.showMoveText();
-        this.showFailedText(failureMessage);
-      }
+      this.showMoveText();
+      this.showFailedText(failureMessage ?? undefined);
 
       // Remove the user from its semi-invulnerable state (if applicable)
       this.pokemon.lapseTags(BattlerTagLapseType.MOVE_EFFECT);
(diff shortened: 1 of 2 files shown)


src/phases/move-phase.ts
```diff
@@ -378,10 +378,8 @@ export class MovePhase extends BattlePhase {
       this.pokemon.pushMoveHistory({ move: this.move.moveId, targets: this.targets, result: MoveResult.FAIL, virtual: this.move.virtual });
 
       const failureMessage = move.getFailedText(this.pokemon, targets[0], move, new BooleanHolder(false));
-      if (failureMessage) {
-        this.showMoveText();
-        this.showFailedText(failureMessage);
-      }
+      this.showMoveText();
+      this.showFailedText(failureMessage ?? undefined);
```
PigeonBar, line 382: (Changes `null` to `undefined` if necessary, in order to satisfy the compiler)

DayKev (approved):

innerthunder (approved):

PR #4858, opened 2024-11-12
[Bug] Moved code preventing MBH's transfer to post modifier generation 

frutescens:
MBH will no longer be able to be stolen from the opponent. 

Bug report https://discord.com/channels/1125469663833370665/1300228174151680040

Code moved down to correct location. 

## How to test the changes?
Try to steal MBH when generated. 

## Checklist

### DIFF
--- a/src/phases/encounter-phase.ts
+++ b/src/phases/encounter-phase.ts
@@ -141,10 +141,6 @@ export class EncounterPhase extends BattlePhase {
         } else if (!(battle.waveIndex % 1000)) {
           enemyPokemon.formIndex = 1;
           enemyPokemon.updateScale();
-          const bossMBH = this.scene.findModifier(m => m instanceof TurnHeldItemTransferModifier && m.pokemonId === enemyPokemon.id, false) as TurnHeldItemTransferModifier;
-          this.scene.removeModifier(bossMBH!);
-          bossMBH?.setTransferrableFalse();
-          this.scene.addEnemyModifier(bossMBH!);
         }
       }
 
@@ -220,6 +216,18 @@ export class EncounterPhase extends BattlePhase {
       if (!this.loaded && battle.battleType !== BattleType.MYSTERY_ENCOUNTER) {
         regenerateModifierPoolThresholds(this.scene.getEnemyField(), battle.battleType === BattleType.TRAINER ? ModifierPoolType.TRAINER : ModifierPoolType.WILD);
         this.scene.generateEnemyModifiers();
+        // This checks if the current battle is an Endless E-Max battle/Classic final boss and sets the MBH held by the boss to untransferrable
+        if (this.scene.currentBattle.waveIndex % 1000 || battle.battleSpec === BattleSpec.FINAL_BOSS) {
+          const enemyPokemon = this.scene.getEnemyPokemon();
+          if (enemyPokemon) {
+            const bossMBH = this.scene.findModifier(m => m instanceof TurnHeldItemTransferModifier && m.pokemonId === enemyPokemon.id, false) as TurnHeldItemTransferModifier;
+            if (bossMBH) {
+              this.scene.removeModifier(bossMBH!);
+              bossMBH?.setTransferrableFalse();
+              this.scene.addEnemyModifier(bossMBH!);
+            }
+          }
+        }
       }
 
       this.scene.ui.setMode(Mode.MESSAGE).then(() => {


src/phases/encounter-phase.ts
```diff
@@ -220,6 +216,18 @@ export class EncounterPhase extends BattlePhase {
       if (!this.loaded && battle.battleType !== BattleType.MYSTERY_ENCOUNTER) {
         regenerateModifierPoolThresholds(this.scene.getEnemyField(), battle.battleType === BattleType.TRAINER ? ModifierPoolType.TRAINER : ModifierPoolType.WILD);
         this.scene.generateEnemyModifiers();
+        // This checks if the current battle is an Endless E-Max battle/Classic final boss and sets the MBH held by the boss to untransferrable
+        if (this.scene.currentBattle.waveIndex % 1000 || battle.battleSpec === BattleSpec.FINAL_BOSS) {
```
flx-sta, line 220: NIT: Can we give those magic checks some names?
Something like `this.scene.currentBattle.isEndlessFinalBoss()` and insteadof of `battle.battleSpec === BattleSpec.FINAL_BOSS` something like `battle.isClassicFInalBoss()`. 

Next person to stumble upon it would have an easier time to understand it.

src/phases/encounter-phase.ts
```diff
@@ -220,6 +216,18 @@ export class EncounterPhase extends BattlePhase {
...
+          const enemyPokemon = this.scene.getEnemyPokemon();
+          if (enemyPokemon) {
+            const bossMBH = this.scene.findModifier(m => m instanceof TurnHeldItemTransferModifier && m.pokemonId === enemyPokemon.id, false) as TurnHeldItemTransferModifier;
+            if (bossMBH) {
+              this.scene.removeModifier(bossMBH!);
+              bossMBH?.setTransferrableFalse();
+              this.scene.addEnemyModifier(bossMBH!);
```
flx-sta, line 227: These `!` and the `?` shouldn't be necessary anymore.

```suggestion
              this.scene.removeModifier(bossMBH);
              bossMBH.setTransferrableFalse();
              this.scene.addEnemyModifier(bossMBH);
```

 
_WE don't want to add new bangs. They were merely a desperate measure to speed up the adoption of strict-null_

flx-sta (changes requested):

src/phases/encounter-phase.ts
MokaStitcher: I thought this existed already because I had seen something like that in the endless boss test, but the numbers are actually redefined in the test.
It test uses `Minor` for 250/500/750 and `Major` for 1000, which could be a name idea. Probably need a name for the 50 bosses too though
https://github.com/pagefaultgames/pokerogue/blob/b6b756a1620c488557e9c6bf5e2b0990242e138e/src/test/endless_boss.test.ts#L7-L10

flx-sta (dismissed):
Except that 1 comment lgtm

src/phases/encounter-phase.ts
flx-sta: In gaming you usually would call those `mini bosses` I'd say

DayKev: > I thought this existed already

https://github.com/pagefaultgames/pokerogue/blob/b6b756a1620c488557e9c6bf5e2b0990242e138e/src/game-mode.ts#L238-L261

src/battle.ts
```diff
@@ -438,6 +445,32 @@ export default class Battle {
   isBattleMysteryEncounter(): boolean {
     return this.battleType === BattleType.MYSTERY_ENCOUNTER;
   }
+
+  /**
+   * @returns `true` if the current battle is against classic mode's final boss
+   */
+  isBattleClassicFinalBoss(): boolean {
```
DayKev, line 452: This already exists btw:
https://github.com/pagefaultgames/pokerogue/blob/b6b756a1620c488557e9c6bf5e2b0990242e138e/src/game-mode.ts#L206-L223

frutescens: screaming and crying

DayKev (approved):

PigeonBar:
It turns out that even with this fix, MBH becomes stealable again **after reloading**, because when the session data gets saved and reloaded, `PersistentModifierData` doesn't keep track of whether or not each individual modifier was set to be transferrable.

https://github.com/user-attachments/assets/51ef6305-25d7-41aa-b4d1-a790a6f1d046

frutescens:
@PigeonBar I think I decided just to stop a potential transfer of MBH at the heart of the code - tryTransferModifier(). Not sure if it leads to other problems though.

PigeonBar:
The issue is that `tryTransferHeldItemModifier()` is used in many places in the code. For example, your new changes would prevent transferring an MBH between party members during the rewards screen.

Some possible other approaches to fixing the bug:
1. Give held items an `isStealable` property separate from `isTransferrable` -- This is needed anyway in the long run, but would also require more care and effort.
2. Set the MBH to be non-transferrable during `EncounterPhase.end()` -- this runs after the MBH is generated, and it also runs no matter whether or not there was a reload.

DayKev (dismissed):

src/battle-scene.ts
```diff
@@ -2584,61 +2584,68 @@ export default class BattleScene extends SceneBase {
...
+      let removeOld = true;
+      if (matchingModifier) {
+        const maxStackCount = matchingModifier.getMaxStackCount(target.scene);
+        if (matchingModifier.stackCount >= maxStackCount) {
+          return resolve(false);
         }
+        const countTaken = Math.min(transferQuantity, itemModifier.stackCount, maxStackCount - matchingModifier.stackCount);
+        itemModifier.stackCount -= countTaken;
+        newItemModifier.stackCount = matchingModifier.stackCount + countTaken;
         removeOld = !itemModifier.stackCount;
```
MokaStitcher, line 2612:
```suggestion
```
not necessary since it's done a few lines below outside the if/else statement

MokaStitcher (changes requested):
I haven't tested more as I already found issues with the things I tested :(
- Magician allows to steal MBH
- Knock Off allows to knock it off (that can't be intended right?)
- Thief seems to crash the game when the steal chance procs
image

At this point I'm really starting to think adding a `stealable` attribute for modifiers that is distinct from the `transferrable` one may be the way to go, rather than adding MBH-specific checks scattered in various places in the code :(

frutescens:
@MokaStitcher Since there's not much time left, I decided to the code that worked before the October update instead. It's clunky and not elegant but I believe it will do the job.

DayKev (approved):

DayKev:
> At this point I'm really starting to think adding a `stealable` attribute for modifiers that is distinct from the `transferrable` one may be the way to go

I've been saying that for a long time. :)

PR #4862, opened 2024-11-13
[Balance] No more double 50x bosses in Endless + cleaning up RNG

frutescens:
Wave 50x bosses will no longer be encountered as double battles. 
Little to no impact on Endure Tokens + Thief/Covet however.

Balance request.

New conditional added to battle-scene.ts that cancels out a double battle if it is a wave 50x boss battle.
Endure Token RNG brought into standard format with the rest of the game. It no longer relies on percentages (0.02 -> 2) and RandSeedInt(100) is used to determine the target's endure chance instead. This ensures that the Endure token is tied to the current battle's seed, not the overall run's seed.  Also brought in Thief/Covet chance rolls to be in line with the game too. 

## How to test the changes?
1. Endless bosses should no longer generate as doubles on floors 50, 100, 150...
2. Test Endure Token behavior and report changes. 
3. Test Covet/Thief and report changes. 

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1241,6 +1241,11 @@ export default class BattleScene extends SceneBase {
       newDouble = false;
     }
 
+    // Disable double battles on Endless/Endless Spliced Wave 50x boss battles (Introduced 1.2.0)
+    if (this.gameMode.isEndlessBoss(newWaveIndex)) {
+      newDouble = false;
+    }
+
     const lastBattle = this.currentBattle;
 
     const maxExpLevel = this.getMaxExpLevel();
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1233,16 +1233,16 @@ export default class BattleScene extends SceneBase {
       newDouble = !!double;
     }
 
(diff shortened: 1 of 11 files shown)


PigeonBar:
It turns out there's an old bug with the `isEndlessBoss()` function, this should be `return waveIndex % 50 === 0 && ...` instead. Thankfully `isEndlessBoss()` is not used anywhere else in the code.

https://github.com/pagefaultgames/pokerogue/blob/0c521bbe0828746ace0dd7310340fb6340bee6e5/src/game-mode.ts#L238-L241

src/battle-scene.ts
```diff
@@ -1241,6 +1241,11 @@ export default class BattleScene extends SceneBase {
       newDouble = false;
     }
 
+    // Disable double battles on Endless/Endless Spliced Wave 50x boss battles (Introduced 1.2.0)
+    if (this.gameMode.isEndlessBoss(newWaveIndex)) {
+      newDouble = false;
+    }
+
```
DayKev, line 1248: This should probably be moved to before the override checks.

MokaStitcher (approved):

DayKev (approved):

innerthunder (approved):

damocleas (approved):

PR #4866, opened 2024-11-14
[P1] Fix System Data Conversion Failure 3

xsn34kzx:

For any users who could not load into the game due to an error (reading `starterData` on `null`) within `fixStarterData`, they should be able to now.

Users were essentially blocked from being able to enter the game due to a null access, forcing them into the "Failed to load system data" screen.

Added a `null` guard to prevent the `fixStarterData` function from running if `starterData` is `null`.

### Screenshots/Videos

N/A

## How to test the changes?

Steps from #4159:
The best way would be to use the latest system data provided in [this bug report](https://discord.com/channels/1125469663833370665/1306275020779032598), place it into your own local storage `user_Guest` data, and make sure the data is decrypted during parsing. The game should load as normal without prompting any console errors.

Extraneous errors in the process that are entirely unrelated may include JSON parsing errors (not decrypting the incoming data), malformed UTF-8 (decrypting the data and saving that decrypted data to local storage but forgetting to remove the decryption portion), etc.

## Checklist

### DIFF
--- a/src/system/version_migration/versions/v1_0_4.ts
+++ b/src/system/version_migration/versions/v1_0_4.ts
@@ -46,12 +46,14 @@ export const systemMigrators = [
    * @param data {@linkcode SystemSaveData}
    */
   function fixStarterData(data: SystemSaveData) {
-    for (const starterId of defaultStarterSpecies) {
-      if (data.starterData[starterId]?.abilityAttr) {
-        data.starterData[starterId].abilityAttr |= AbilityAttr.ABILITY_1;
-      }
-      if (data.dexData[starterId]?.caughtAttr) {
-        data.dexData[starterId].caughtAttr |= DexAttr.FEMALE;
+    if (data.starterData !== null) {
+      for (const starterId of defaultStarterSpecies) {
+        if (data.starterData[starterId]?.abilityAttr) {
+          data.starterData[starterId].abilityAttr |= AbilityAttr.ABILITY_1;
+        }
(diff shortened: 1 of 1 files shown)


DayKev (approved):

src/system/version_migration/versions/v1_0_4.ts
```diff
@@ -46,12 +46,14 @@ export const systemMigrators = [
    * @param data {@linkcode SystemSaveData}
    */
   function fixStarterData(data: SystemSaveData) {
-    for (const starterId of defaultStarterSpecies) {
-      if (data.starterData[starterId]?.abilityAttr) {
-        data.starterData[starterId].abilityAttr |= AbilityAttr.ABILITY_1;
-      }
-      if (data.dexData[starterId]?.caughtAttr) {
-        data.dexData[starterId].caughtAttr |= DexAttr.FEMALE;
+    if (data.starterData !== null) {
```
DayKev, line 49:
```suggestion
    if (!isNullOrUndefined(data.starterData)) {
```
Actually thinking about it, maybe it should be this?

Will require the import as well (also replace all those `../..` in the other imports lol).

MokaStitcher (dismissed):
Approving since technically it solves the issue.
Will likely need a separate PR to move the starterData migration to the same spot as the rest of the version migration as discussed [on discord](https://discord.com/channels/1125469663833370665/1282387337162264626/1306640311027499040)

DayKev (approved):

MokaStitcher (dismissed):

PR #4872, opened 2024-11-15
[Move] Clean up Glaive Rush implementation

innerthunder:
No user-facing changes aside from maybe removing some oddly-timed fail messages when using Glaive Rush.

Glaive Rush is currently implemented with two tags, both exclusive to the move, and a condition that has since been implemented in a cleaner way with `AddBattlerTagAttr`'s `cancelOnFail` param. This cleans up both of these aspects of the move in preparation for cleaner move scoring with the AI rework.

- Glaive Rush's `ALWAYS_GET_HIT` and `RECEIVE_DOUBLE_DAMAGE` tags have been condensed into a single `GLAIVE_RUSH` type tag. This tag also now has its own tag class `GlaiveRushTag` instead of directly using the `BattlerTag` class.
- Glaive Rush's tag attribute now sets `cancelOnFail` to `true`, replacing Glaive Rush's `condition` function.

## How to test the changes?
`npm run test glaive_rush`

## Checklist
- [x] Are the changes visual?


src/data/battler-tags.ts
```diff
@@ -2489,6 +2489,12 @@ export class AutotomizedTag extends BattlerTag {
   }
 }
 
+export class GlaiveRushTag extends BattlerTag {
```
DayKev, line 2492: I didn't create a `GlaiveRushTag` class intentionally because it's not necessary, you're not doing anything with it; the base `BattlerTag` class already contains all the functionality being used (and there's plenty of tags already doing the same thing, see the ones directly above and below them in the `Switch` block as examples).

Actually I don't like combining the tags themselves either, now it can never be reused in the future.

innerthunder: I think in the long run, it'd be better if every effect had its own tag class. With the AI rework, for example, we're planning on adding some scoring functions within certain tags to improve the AI's decision making. I can't say for sure that Glaive Rush will be included in this, but structuring its tag this way allows for more flexibility to that end.

As for the merge, I could see the tags being recycled in the future, so I could go either way. The thing is that we've already had opportunities to reuse it (like `ALWAYS_GET_HIT` for Telekinesis), and we just haven't done so.

DayKev: > I think in the long run, it'd be better if every effect had its own tag class. With the AI rework, for example, we're planning on adding some scoring functions within certain tags to improve the AI's decision making. I can't say for sure that Glaive Rush will be included in this, but structuring its tag this way allows for more flexibility to that end.

But there's no need to do this pre-emptively, we don't need to add a bunch of empty classes that "might" get used. The ones that end up needing them can have classes added later.

> As for the tag merge, I could see the tags being recycled in the future, so I could go either way on it. The thing is that we've already had opportunities to reuse it (like `ALWAYS_GET_HIT` for Telekinesis), and we just haven't done so.

Telekinesis' accuracy bypass doesn't apply to OHKO moves or during Fly/etc but Glaive Rush does. It also has an `onAdd` message, though that could probably be part of the move instead of the tag.

PR #4880, opened 2024-11-16
[Misc] Rename `DamagePhase` to `DamageAnimPhase`

DayKev:
N/A

The phase name is misleading as it doesn't actually calculate damage or anything, just play the animations for damage numbers appearing.

Renamed `DamagePhase` to `DamageAnimPhase` and the file from `damage-phase.ts` to `damage-anim-phase.ts`.

## Checklist


src/phases/damage-anim-phase.ts
```diff
@@ -1,11 +1,11 @@
-import BattleScene from "#app/battle-scene";
-import { BattlerIndex } from "#app/battle";
-import { BattleSpec } from "#app/enums/battle-spec";
-import { DamageResult, HitResult } from "#app/field/pokemon";
-import * as Utils from "#app/utils";
-import { PokemonPhase } from "./pokemon-phase";
+import type BattleScene from "#app/battle-scene";
+import { type BattlerIndex } from "#app/battle";
+import { BattleSpec } from "#enums/battle-spec";
+import { type DamageResult, HitResult } from "#app/field/pokemon";
```
innerthunder, line 4: Why the `type`s here?

DayKev: Type-only imports are not added to the transpiled js. All imports should be type-only when they are only used as type hints (btw VSCode users, enable "prefer type only imports" from your settings which for some insane reason isn't enabled by default).

innerthunder (approved):

PR #4887, opened 2024-11-16
[Refactor] Rewrite Move Effect Phase

### LINKED ISSUE #3958: [Bug] Prankster Roar/Whirlwind Bugs Out On Dark Types/Dazzling/etc.
### Describe the bug

Prankster gives priority to Roar/Whirlwind, turning it from -6 to -5. Dark types, Dazzling, Queenly Majesty, Armor Tail, and Psychic Terrian all give immunity to priority moves.
When using the move against a mon with immunity, it will "roar" the target and make it untargetable, but the target will still be in the battle and will attack the following turn.

### Session export file

_No response_

### User data export file

_No response_

### Expected behavior

Prankster Roar/Whirlwind against immune targets should just fail the move immediately like all other priority move fails.

### Screenshots / Videos

https://github.com/user-attachments/assets/5e113c95-03c3-4b36-b6ae-baf0e6c89213

PC/Windows

### If other please specify

_No response_

### Which browser do you use?

Firefox

### If other please specify

_No response_

### Additional context

_No response_

### LINKED ISSUE #4499: [Bug] Fixed damage moves dealing critical hits
### Describe the bug

(First reported at https://discord.com/channels/1125469663833370665/1125894949020381285/1289739808125751307)

Fixed-damage moves, such as Super Fang and Night Shade, can score critical hits. It appears that this bug is purely visual and does not cause the moves to deal extra damage.

### Reproduction

To reproduce this bug easily, use overrides to give Farfetch'd a leek, a scope lens, and some fixed-damage moves.

### Expected behavior

According to https://bulbapedia.bulbagarden.net/wiki/Category:Moves_that_deal_direct_damage, fixed-damage moves should not be able to score critical hits.

### Screenshots / Videos

https://github.com/user-attachments/assets/edfa3eb1-2e2f-41f1-b6b4-59295ec0b3ed

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

### LINKED ISSUE #4645: [Bug] Entry hazards still get set twice during double battles
### Describe the bug

Part of PR #3528 was designed to fix the bugs involving Spikes and Toxic Spikes getting set twice. However, partway through its development (https://github.com/pagefaultgames/pokerogue/pull/3528/commits/89b5893f82dfa97ef061e21fed46b8d40a700eb5), this fix was accidentally deleted, so Spikes and Toxic Spikes still get set twice:

https://github.com/user-attachments/assets/d8d7b3ed-19c0-49a5-a101-d210050ed790

### Reproduction

Use Spikes or Toxic Spikes during a double battle.

### Expected behavior

Spikes or Toxic Spikes should only be set once per turn.

### Screenshots / Videos

_No response_

### Session export file

_No response_

### User data export file

_No response_

### Additional context

Original bug: #3240

### LINKED ISSUE #4764: [Bug] Certain Move and Ability Effects (BattlerTags) Work through Substitute (PENDING)
### Describe the bug

Toxic Chain, Magician, Stench

PostAttackAbAttr

Circumvents Substitute

Salt Cure, Syrup Bomb, Octolock, Tar Shot, anchor shot, throat chop nightmare, 

AddBattlerTagAttr seems to circumvent Substitute in cases when it shouldn't, when not a sound move and used by an opposing mon

### Reproduction

Use one of the moves listed above, or have one of those abilities activate against a pokemon behind a Substitute

### Expected behavior

https://replay.pokemonshowdown.com/gen9customgame-2233812802-wmedsp6w3du79eyvz6h9ta9oolffjbppw

### Screenshots / Videos

_No response_

### Session export file

_No response_

### User data export file

_No response_

### Additional context

I HAVE NOT FINISHED THIS YET

innerthunder:
> [!IMPORTANT]
> I have left the project and am no longer maintaining this PR. Leaving this up for reference in case someone else wants to mirror it and fix it up.
> \- innerthunder

- Move animations will no longer play if the move does not successfully hit a target (fixes #3958 and other visual inconsistencies)
- Semi-invulnerability, protection, and move immunity effects are now checked in the correct (mainline) order, fixing odd interactions when multiple of these effects are present.
- Meloetta's form change upon using Relic Song now visually triggers after all other effects of the move.
- Field effects such as Trick Room should not be invoked multiple times in a single use of move. This is an extension of the fix to Spikes and Toxic Spikes applying multiple layers in #3528 
  - This might also fix #4645 
- Fixed-damage moves should no longer appear to critically hit (fixes #4499)
- Substitute should now properly block most secondary effects against the source (see below)

I figure it's better to list the changes before justifying them...
- `phases/move-effect-phase`:
  - `hitCheck` now returns a `HitCheckEntry`, which consists of a `HitCheckResult` and the type effectiveness multiplier of the phase's move against the given target. The `HitCheckResults` this function can now return are as follows:
    - `HIT`: The move successfully hit the target
    - `NO_EFFECT`: The target is immune to the move.
    - `NO_EFFECT_NO_MESSAGE`: The target is immune to the move, and the immunity effect plays its own message instead of the default "It doesn't affect {target}!"
    - `PROTECTED`: The target protected itself against the move.
    - `MISS`: The move failed the accuracy check and missed the target.
    - `ERROR`: The move failed unexpectedly (e.g. the move's user is `undefined`)

    Hit check entries against each of the move's targets are cached in the phase's new `hitChecks` field to be used throughout the phase's other functions. Hit check entries are initialized in a `PENDING` state, but each hit check should be overwritten before effects apply.
  - If no hit check results in a `HIT`, the phase will end without playing the move's animation (unless the move has a `POST_TARGET` effect like Explosion).
  - `Pokemon.apply` has been moved from the `Pokemon` class to `MoveEffectPhase.applyMove`. Critical hit resolution and damage calculations are still done within `Pokemon` methods. `PRE_APPLY` move effects are now triggered before this method is called (but only when the move successfully hits).
  - `POST_APPLY` and `HIT` move effect triggers have been condensed into a single `POST_APPLY` trigger (WIP). Target-facing secondary effects generally no longer trigger when the move hits a substitute (fixes #4764 )
  - Helper functions for different move effect trigger types have been generalized into a single `triggerMoveEffect` function.
  - Removed `MoveEffectPhase.stopMultiHit` and `Pokemon.stopMultiHit`. Multi-hit early stopping is now done with the help of added code in `MEPhase.end` (and modifications to `hitCount` and `hitsLeft` in some edge cases)
  - Other minor adjustments to field-targeting moves (not a complete fix), immunity checks, and general cleanup.

Aside from fixing the bugs listed above (among others), this should allow for cleaner implementations of moves and abilities that interact with the phase for several reasons:
- Using a `Pokemon` function to apply move damage brought about unnecessary scope issues, especially having to do with type immunity checks and multi-hit logic. Moving this function to `MoveEffectPhase` allows developers to modify most, if not all, aspects of a move's effects in one place.
- Move effect trigger types are needlessly convoluted, and the distinction between `POST_APPLY` and `HIT` move effect triggers has become increasingly unclear. It'd be nice to condense trigger types that are ultimately redundant.
- `PRE_APPLY` move effects currently have no way of actually applying before the move's effectiveness and/or damage were calculated. This fixes that.
- Condensing all hit checks (immunity, protection, and accuracy) into one place will help with some otherwise complicated implementations, including Dragon Darts and Sky Drop.
- This should bring field effects closer to not requiring a target to apply

## How to test the changes?
`npm run test` (still fixing some tests)

## Checklist
- [?] Are the changes visual?


src/data/move.ts
```diff
@@ -5487,16 +5453,13 @@ export class ConfuseAttr extends AddBattlerTagAttr {
       return false;
     }
 
-    if (!move.hitsSubstitute(user, target)) {
-      return super.apply(user, target, move, args);
-    }
     return false;
```
PigeonBar, line 5456:
```suggestion
    return super.apply(user, target, move, args);
```

innerthunder: oops... I caught something like this for flinching, but I guess there isn't much test coverage for confusion.

PR #4900, opened 2024-11-17
[Misc] Follow up to Wimp Out bug fix

frutescens:
Follow up to #4899 with test from @PigeonBar + some comments 

Test added for checking that Wimp Out behaves correctly in this situation. 

## How to test the changes?
npm run test wimp_out

## Checklist

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -4951,9 +4951,10 @@ class ForceSwitchOutHelper {
       }
     /**
      * For wild Pokémon battles, the Pokémon will flee if the conditions are met (waveIndex and double battles).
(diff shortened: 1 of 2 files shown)


src/data/ability.ts
```diff
@@ -4951,6 +4951,7 @@ class ForceSwitchOutHelper {
       }
     /**
      * For wild Pokémon battles, the Pokémon will flee if the conditions are met (waveIndex and double battles).
+     * It will not flee if it is a wave 10x boss battle or a battle in a Mystery Encounter
```
PigeonBar, line 4954:
```suggestion
     * It will not flee if it is a Mystery Encounter with fleeing disabled (checked in `getSwitchOutCondition()`) or if it is a wave 10x wild boss
```

PR #4896, opened 2024-11-17
[Balance] Multi Lens damage reduction on fixed-damage moves

PigeonBar:
## What are the changes the user will see?

Multi Lens will apply proper damage reduction to fixed-damage moves.
Affected moves: Sonic Boom, Dragon Rage, Super Fang, Nature's Madness, Ruination, Counter, Mirror Coat, Metal Burst, Comeuppance, Seismic Toss, Night Shade, Psywave.

**Note**: This implementation technically causes `TargetHalfHpDamageAttr` (e.g. Super Fang) to deal slightly less damage if used with Multi Lens.
With 1 Multi Lens, the damage dealt is around 45% (1 - 0.625 * 0.875) of the target's HP.
With 2 Multi Lens, the damage dealt is around 43% (1 - 0.75 * 0.875 * 0.875) of the target's HP.

## Why am I making these changes?

Requested by balance, due to recent Multi Lens changes now allowing Multi Lens to activate on fixed-damage moves.

## What are the changes from a developer perspective?

1. `PokemonMultiHitModifier` now gets applied to fixed-damage moves.
2. Rewrote `PokemonMultiHitModifier` slightly to account for interaction with Parental Bond on fixed-damage moves.
3. Added a test.

### Screenshots/Videos

(Note: Cubone has 2 Multi Lens and Parental Bond in these videos. Hits 1-3 are Multi Lens, Hit 4 is Parental Bond. Ignore the critical hit on the first hit, it's a visual bug unrelated to Multi Lens: #4499)
<details>
<summary> Before the changes: Multi Lens does not apply damage reduction to Seismic Toss </summary>

https://github.com/user-attachments/assets/d5135e82-075a-4ced-ab1a-62a7a348cad1

</details>
<details>
<summary> After the changes: Multi Lens applies damage reduction to Seismic Toss </summary>

https://github.com/user-attachments/assets/54edaf02-18c2-40a2-a51f-040ec14d7aad

</details>

## How to test the changes?

`npm run test parental_bond`
`npm run test multi_lens`

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I considered writing automated tests for the issue?
- ~~[ ] If I have text, did I make it translatable and add a key in the English locale file(s)?~~
- [x] Have I tested the changes (manually)?
[...]

PR #4914, opened 2024-11-19
[BUGFIX] Fixed Super Fang interaction with Multi Lens

Bertie690:
[RIP funny math formula, you will be missed.](https://www.desmos.com/calculator/wdngrksdfz)

Super Fang and company now deal correct damage with multi lens (ie 50%), split between the appropriate amount of hits linearly.
For an example of how it works: The 1st hit of super fang with 2 lenses does `(target.hp/2)*0.5=0.25*hp` damage, leaving the enemy down to 75% hp after the first hit. The remaining 2 multi lens hits each get half the difference between the leftover (75%) and the target (50%) amounts, for 12.5% damage each. 

Parental Bond works similarly, but it changes the target damage amount to 0.25 instead (the hp% the target would have left after 2 halvings).

BECAUSE I COULD.

Adds a check in `TargetHalfHpDamageAttr.apply()` for subsequent repeated hits that splits the damage of hits after the first one linearly to make them add up to the desired amount. The first hit is unaffected as the following strikes are intended to adjust for it appropriately.
Also adds a check in pokemon.ts to skip multiplying the move damage for the subsequent hits of such moves because they're already pre computed and I _really_ don't want to deal with reciprocals...

### Screenshots/Videos
_oh god please no I'm not spending another hour filming these just for someone to tell me to refactor them 12 hours later_

## How to test the changes?
`npm run test multi_lens`
Check the last few tests at the end.
Alternately use the following overrides:
```ts
const overrides = {
  MOVESET_OVERRIDE: Moves.SUPER_FANG,
  STARTING_HELD_ITEMS_OVERRIDE: [{ name: "MULTI_LENS", count: 2 }],
  OPP_LEVEL_OVERRIDE: 100,
  OPP_SPECIES_OVERRIDE: Species.BLISSEY,
  OPP_MOVESET_OVERRIDE: Moves.SPLASH,
```
and start a new game with a Koffing and _literally any other starter without PB_.

## Checklist
- ~~If I have text, did I make it translatable and add a key in the English locale file(s)?~~
- [x] Are the changes visual?


src/field/pokemon.ts
```diff
@@ -2618,10 +2618,34 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
-    const fixedDamage = new Utils.IntegerHolder(0);
+    // If the attack deals fixed damage, return a result with that much damage
+    const fixedDamage = new Utils.NumberHolder(0);
     applyMoveAttrs(FixedDamageAttr, source, this, move, fixedDamage);
     if (fixedDamage.value) {
+      const lensCount = source.getHeldItems().find(i => i instanceof PokemonMultiHitModifier)?.getStackCount() ?? 0;
+      // Apply damage fixing for hp cutting moves on multi lens hits (NOT PARENTAL BOND)
+      if (move.hasAttr(TargetHalfHpDamageAttr) &&
+        (source.turnData.hitCount === source.turnData.hitsLeft ||
+          source.turnData.hitCount - source.turnData.hitsLeft !== lensCount + 1)) {
```
DayKev, line 2629:
```suggestion
      if (move.hasAttr(TargetHalfHpDamageAttr)
        && (source.turnData.hitCount === source.turnData.hitsLeft
          || source.turnData.hitCount - source.turnData.hitsLeft !== lensCount + 1)) {
```

src/test/items/multi_lens.test.ts
```diff
@@ -135,4 +135,78 @@ describe("Items - Multi Lens", () => {
...
+      .enemyMoveset(Moves.SPLASH)
+      .enemyLevel(100000)
+      .enemySpecies(Species.BLISSEY); // allows for unrealistically high levels of accuracy
+
+    await game.classicMode.startBattle([ Species.MAGIKARP ]);
+
+    const enemyPokemon = game.scene.getEnemyPokemon()!;
+
+    game.move.select(Moves.SUPER_FANG);
+    await game.forceEnemyMove(Moves.SPLASH);
```
DayKev, line 169:
```suggestion
```
This can be omitted if the enemy has only one move (or in cases where we don't care what move the enemy uses).

src/test/items/multi_lens.test.ts
```diff
@@ -135,4 +135,78 @@ describe("Items - Multi Lens", () => {
...
```
DayKev, line 188:
```suggestion
```

src/test/items/multi_lens.test.ts
```diff
@@ -135,4 +135,78 @@ describe("Items - Multi Lens", () => {
...
+    await game.setTurnOrder([ BattlerIndex.PLAYER, BattlerIndex.ENEMY ]);
+    await game.phaseInterceptor.to("MoveEndPhase", true);
```
DayKev, line 171:
```suggestion
    await game.phaseInterceptor.to("MoveEndPhase");
```

src/test/items/multi_lens.test.ts
```diff
@@ -135,4 +135,78 @@ describe("Items - Multi Lens", () => {
...
+      .enemyLevel(100000)
+      .enemySpecies(Species.BLISSEY); // allows for unrealistically high levels of accuracy
+
+    await game.classicMode.startBattle([ Species.MAGIKARP ]);
+
+    const enemyPokemon = game.scene.getEnemyPokemon()!;
+
+    game.move.select(Moves.SUPER_FANG);
+    await game.setTurnOrder([ BattlerIndex.PLAYER, BattlerIndex.ENEMY ]);
+    await game.phaseInterceptor.to("MoveEndPhase", true);
```
DayKev, line 152:
```suggestion
    await game.phaseInterceptor.to("MoveEndPhase");
```
The default is `true`.

src/test/items/multi_lens.test.ts
```diff
@@ -135,4 +135,78 @@ describe("Items - Multi Lens", () => {
...
+      .enemySpecies(Species.BLISSEY); // allows for unrealistically high levels of accuracy
+
+    await game.classicMode.startBattle([ Species.MAGIKARP ]);
+
+    const enemyPokemon = game.scene.getEnemyPokemon()!;
+
+    game.move.select(Moves.SUPER_FANG);
+    await game.forceEnemyMove(Moves.SPLASH);
+    await game.setTurnOrder([ BattlerIndex.PLAYER, BattlerIndex.ENEMY ]);
+    await game.phaseInterceptor.to("MoveEndPhase", true);
```
DayKev, line 190:
```suggestion
    await game.phaseInterceptor.to("MoveEndPhase");
```

src/test/items/multi_lens.test.ts
```diff
@@ -135,4 +135,78 @@ describe("Items - Multi Lens", () => {
...
+      .enemyMoveset(Moves.SPLASH)
+      .enemyLevel(100000)
+      .enemySpecies(Species.BLISSEY); // allows for unrealistically high levels of accuracy
+
+    await game.classicMode.startBattle([ Species.MAGIKARP ]);
+
+    const enemyPokemon = game.scene.getEnemyPokemon()!;
+
+    game.move.select(Moves.SUPER_FANG);
+    await game.forceEnemyMove(Moves.SPLASH);
```
DayKev, line 207:
```suggestion
```

src/test/items/multi_lens.test.ts
```diff
@@ -135,4 +135,78 @@ describe("Items - Multi Lens", () => {
...
+    await game.setTurnOrder([ BattlerIndex.PLAYER, BattlerIndex.ENEMY ]);
+    await game.phaseInterceptor.to("MoveEndPhase", true);
```
DayKev, line 209:
```suggestion
    await game.phaseInterceptor.to("MoveEndPhase");
```

DayKev:
Seems fine to me, just some style comments.

src/field/pokemon.ts
```diff
@@ -2618,10 +2618,34 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
     if (fixedDamage.value) {
+      const lensCount = source.getHeldItems().find(i => i instanceof PokemonMultiHitModifier)?.getStackCount() ?? 0;
+      // Apply damage fixing for hp cutting moves on multi lens hits (NOT PARENTAL BOND)
+      if (move.hasAttr(TargetHalfHpDamageAttr)
+        && (source.turnData.hitCount === source.turnData.hitsLeft
+        || source.turnData.hitCount - source.turnData.hitsLeft !== lensCount + 1)) {
+        // Do some unholy math to make the moves' damage values add up to 50%
+        // Values obtained courtesy of WolframAlpha and Desmos Graphing Calculator
+        // (https://www.desmos.com/calculator/wdngrksdfz)
+        let damageMulti = 0;
```
PigeonBar, line 2633: I would like to reiterate my strong concerns about potential bugs arising from hard-coding a solution instead of finding a cleaner solution. For example, Super Fang now deals basically no damage if the user is not holding Multi Lens.

https://github.com/user-attachments/assets/95ac7e49-3b49-4252-bb53-e111894f7d25

innerthunder:
I'm not a fan of hard-coding the fixed damage values like this. We just updated Multi-Lens to reduce its tech debt, and this change feels like we're going backwards on that. Also, the balance team has made comments implying that Multi-Lens' maximum stack count is subject to change, and this change could potentially get in the way of that.

Some options that wouldn't increase tech debt:
1. We leave the interaction as is. This ultimately makes Multi-Lens a nerf to Super Fang's damage output, but obviously it would be the easiest option for devs.
2. We revert the changes from #4896. This would probably make fixed-damage moves too overpowered, but would definitely make the Super Fang interaction more stable.
3. We restrict Multi-Lens' effect on all fixed-damage moves. I don't think players would enjoy this change, at least in the short term, but this restriction has been in place before, and it wouldn't be difficult to add.
4. We restrict Multi-Lens' effect on Super Fang, Ruination, and Nature's Madness. This would "fix" the interaction while preserving the interactions of other fixed-damage moves, but it wouldn't be very intuitive to the player (I could see some "bug" reports come from this one)

DayKev:
> We restrict Multi-Lens' effect on Super Fang, Ruination, and Nature's Madness. This would "fix" the interaction while preserving the interactions of other fixed-damage moves, but it wouldn't be very intuitive to the player (I could see some "bug" reports come from this one)

This is what I prefer tbh.

> We leave the interaction as is. This ultimately makes Multi-Lens a nerf to Super Fang's damage output, but obviously it would be the easiest option for devs.

This would be fine too.

Tempo-anon:
Appreciate the math and work done on this but will also echo the sentiment that hardcoding it is not the way to go

Bertie690:
I get the message (hardcoding = not future proofed = bad), but at the same time, I do think these hard coded values serve a purpose - as a "stopgap" fix that fixes stuff now and can be cleaned up later. 
I will almost certainly work on a cleaner implementation of this stuff later (likely using some sort of "target" damage tracker like pigeon-bar suggested), but frankly that can be a task for next week's me, not today's me. Yeah it may break in the future, but IMHO what matters is that it works _now_.

Bertie690:
Also, yes I know the end result is the same as "just turn off multi lens for hp cutting moves", but the actual gameplay effects of it being single hit are not nonexistent (iron barbs, boss hp bars, etc.)

innerthunder:
> Thanks for the feedback everyone. I do get the message (hardcoding = not future proofed = bad), but at the same time, I do think these hard coded values serve a purpose - as a "stopgap" fix that fixes stuff now and can be cleaned up later. I can work on a cleaner implementation of this stuff later (likely using some sort of "target" damage tracker like pigeon-bar suggested), but frankly that can be a task for next week's me, not today's me. (I'm just sharing my opinion FYI - feel free to pick it apart.)

I think when it comes to an open-source project where development is largely based on "best effort," it's generally best to avoid stopgap fixes when possible. The cleanup for "temporary" fixes tends to fall into low priority, so it's likely that a stopgap sticks around until it ends up causing problems itself.

Bertie690:
I'll get started on reworking it then.

src/field/pokemon.ts
```diff
@@ -2622,34 +2622,14 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
-          default:
-            damageMulti = 0.5;
-            break;
-        }
-
-        fixedDamage.value = this.hp * damageMulti;
+      // Don't apply the multi lens damage penalty for subsequent hits of a hp cutting move
+      // those get handled separately in move.ts
+      if (!(move.hasAttr(TargetHalfHpDamageAttr)
+        && source.turnData.hitCount !== source.turnData.hitsLeft)) {
```
PigeonBar, line 2628: Multi Lens multiplier applying to only the first hit of Super Fang is unintuitive and could lead to accidental mistakes by future devs who need to work with Multi Lens. Please change it so that the Multi Lens multiplier also applies to the bonus hits granted by Multi Lens (which is how it works for all other moves).

Since the intent of Multi Lens is based on splitting damage across multiple hits, an intuitive way to implement this would be for `TargetHalfHpDamageAttr` to keep track of an `initialHp` that gets reset before non-Multi Lens hits, then set damage to `initialHp / 2` (which will then be split up by Multi Lens).

DayKev (approved):

frutescens (approved):
> Screenshots/Videos

> oh god please no I'm not spending another hour filming these just for someone to tell me to refactor them 12 hours later

Approving - however, I do want to emphasize that even though providing screenshots/video proof is annoying and time-consuming, it is important and good practice to do so. The burden of proof is on the PR's creator... so please remember this in the future.

PR #4922, opened 2024-11-19
[Tests] Enemy Held Item overrides are now applied after generateEnemyModifiers()

frutescens:
None.

This change is needed to test item-related moves and abilities without having unwanted items interfere with test conditions. 

overrideModifiers and overrideHeldItems are now applied after generateEnemyModifiers() in encounter-phase.ts. Previously, they were applied in addEnemyPokemon(), but addEnemyPokemon() was run before generateEnemyModifiers() so the items in the override would be given but additional items could also be generated afterwards. 

## Checklist


frutescens:
These flakey tests...

DayKev (approved):

PR #4941, opened 2024-11-26
[Move] Full Stuff Cheeks Implementation

geeilhan:

Stuff Cheeks is now fully implemented. (#3503)

Added a new class `MoveSelectCondition` inside `move.ts` which checks if a move is selectable in the move menu. It contains the condition (`UserMoveConditionFunc`) that allows selecting a move. 

`StuffCheeksCondition` extends `MoveSelectCondition` and contains the condition that fails if the user does not hold at least 1 berry. If the condition fails a `StuffCheeksTag`, which extends `MoveRestrictionTag`, is added. This tag gets removed at the end of the turn. 

Also added
`private selectableConditions: MoveSelectCondition[] = [];`
which saves the condition for the move.

## Screenshots/Videos

Before implementation:

[stuff-cheeks-implementation-before.webm](https://github.com/user-attachments/assets/1423affc-25c2-42e2-810d-c9ad31e6d556)

After Implementation:

[stuff-cheeks-implementation-after.webm](https://github.com/user-attachments/assets/0ea4676e-c707-4b00-89d5-fc7e62eaa533)

## How to test the changes?

Added automated tests

## Checklist


geeilhan:
Currently working on refactoring this so I use BattlerTags (MoveRestricitonBattlerTag) instead of making the move directly unselectable

geeilhan:
I think something is wrong with the tests. `npm run test` doesn't work after syncing with beta branch

geeilhan:
If this gets merged I would like to use `MoveSelectCondition` for Fling and Switcheroo

src/data/move.ts
```diff
@@ -7722,6 +7752,64 @@ export function applyMoveChargeAttrs(attrType: Constructor<MoveAttr>, user: Poke
...
+    super((user, move) => this.selectableCondition(user));
+  }
+
+  /**
+   * Checks if the user is holding a berry
+   * @param user {@linkcode Pokemon} whose berries to check
+   * @returns true if the user is holding a berry, otherwise false
+   */
+  private selectableCondition(user: Pokemon): boolean {
+    return user.getHeldItems().filter(m => m instanceof BerryModifier, user.isPlayer()).length > 0;
```
DayKev, line 7792:
```suggestion
    return user.getHeldItems().filter(m => m instanceof BerryModifier).length > 0;
```
The original function this is copied from was `scene.findModifiers()` which took a `isPlayer: boolean` param, but `.filter()` is a different function and passing `user.isPlayer()` to it doesn't do anything here.

src/data/move.ts
```diff
@@ -7722,6 +7752,64 @@ export function applyMoveChargeAttrs(attrType: Constructor<MoveAttr>, user: Poke
...
+   */
+  apply(user: Pokemon, move: Move): boolean {
+    if (!this.selectableCondition(user)) {
+      user.addTag(BattlerTagType.STUFF_CHEEKS, 0, move.id);
+    }
+    return this.func(user, move);
+  }
+}
+
+const hasBerryCondition: MoveConditionFunc = (user: Pokemon, target: Pokemon, move: Move) => user.getHeldItems().filter(m => m instanceof BerryModifier, user.isPlayer()).length > 0;
```
DayKev, line 7811:
```suggestion
const hasBerryCondition: MoveConditionFunc = (user: Pokemon, target: Pokemon, move: Move) => user.getHeldItems().filter(m => m instanceof BerryModifier).length > 0;
```

src/data/move.ts
geeilhan: Thanks for the hint!

SirzBenjie:
@geeilhan Apologies that this never made it in. Your PRs are always a treat to view. I think the refactor of moves are implemented, combined with the upcoming modifier rework, means this PR is unfortunately too outdated.
An implementation will be easier once the modifier rework comes. I'm going to close this.

PR #4982, opened 2024-12-10
[Move] Add makeContact To Physical Part Of Shell Side Arm

geeilhan:
(#3503)

The user now makes physical contact if Shell Side Arm does physical damage.

Now calls `makesContact()` if move turns physical.

Removed `.partial()` 

Added test that checks for contact

## Screenshots/Videos

Before changes vs Rough Skin:

[shell-side-arm-implementation-before.webm](https://github.com/user-attachments/assets/b858a4ad-c51f-48c2-ae12-5e86841b4739)

After changes vs Rough Skin:

[shell-side-arm-implementation-after.webm](https://github.com/user-attachments/assets/13967dd3-e747-4091-a9ac-34aac31340a7)

## How to test the changes?

Give enemy Pokemon Rough Skin (Ability) and use Side Shell Arm. Make sure the enemy Pokemon has more SPDEF than DEF.

```
const overrides = {
  OPP_SPECIES_OVERRIDE: Species.SNORLAX,
  OPP_ABILITY_OVERRIDE: Abilities.ROUGH_SKIN,
  OPP_LEVEL_OVERRIDE: 20,
  MOVESET_OVERRIDE: [ Moves.SHELL_SIDE_ARM ],
```

## Checklist


Tempo-anon:
Hey geeilhan, are you on the discord server? I have some questions

Tempo-anon (changes requested):
I don't think this works as it's making contact even when special. Can you add the following test?
```ts
  it("should not make contact if the move becomes special", async () => {
    game.override.enemySpecies(Species.SLOWBRO).enemyAbility(Abilities.ROUGH_SKIN);

    await game.classicMode.startBattle([Species.XURKITREE]);

    const player = game.scene.getPlayerPokemon()!;

    game.move.select(Moves.SHELL_SIDE_ARM);
    await game.toNextTurn();

    expect(player.getMaxHp()).toBe(player.hp);
  });
```

geeilhan:
> I don't think this works as it's making contact even when special. Can you add the following test?
> 
> ```ts
>   it("should not make contact if the move becomes special", async () => {
>     game.override.enemySpecies(Species.SLOWBRO).enemyAbility(Abilities.ROUGH_SKIN);
> 
>     await game.classicMode.startBattle([Species.XURKITREE]);
> 
>     const player = game.scene.getPlayerPokemon()!;
> 
>     game.move.select(Moves.SHELL_SIDE_ARM);
>     await game.toNextTurn();
> 
>     expect(player.getMaxHp()).toBe(player.hp);
>   });
> ```

Thanks for 

> I don't think this works as it's making contact even when special. Can you add the following test?
> 
> ```ts
>   it("should not make contact if the move becomes special", async () => {
>     game.override.enemySpecies(Species.SLOWBRO).enemyAbility(Abilities.ROUGH_SKIN);
> 
>     await game.classicMode.startBattle([Species.XURKITREE]);
> 
>     const player = game.scene.getPlayerPokemon()!;
> 
>     game.move.select(Moves.SHELL_SIDE_ARM);
>     await game.toNextTurn();
> 
>     expect(player.getMaxHp()).toBe(player.hp);
>   });
> ```

Thanks for the feedback. Will implement this too

Tempo-anon:
That is weird. For reference I was testing out something like 
```ts
override apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
    const category = args[0] as NumberHolder;

    const predictedPhysDmg = target.getBaseDamage(user, move, MoveCategory.PHYSICAL, true, true);
    const predictedSpecDmg = target.getBaseDamage(user, move, MoveCategory.SPECIAL, true, true);

    // Random chance of being physical or special if predicted damage is tied
    if (predictedPhysDmg > predictedSpecDmg || (predictedPhysDmg === predictedSpecDmg && user.randSeedInt(2) === 0)) {
      category.value = MoveCategory.PHYSICAL;
      move.makesContact();
      console.log("CONTACT");
      return true;
    }
    move.makesContact(false);
    console.log("NO CONTACT");
    return false;
  }
```
and when I `npm run test shell_side_arm` I get the following output
```
stdout | src/test/moves/shell_side_arm.test.ts > Moves - Shell Side Arm > should not make contact if the move becomes special
[ { targets: [ 2 ], multiple: false }, 'Xurkitree' ]
[ 'setMode', 'MESSAGE (=0)', [] ]
[ '%cStart Phase EnemyCommandPhase', 'color:green;' ]

stdout | src/test/moves/shell_side_arm.test.ts > Moves - Shell Side Arm > should not make contact if the move becomes special
[ '%cStart Phase TurnStartPhase', 'color:green;' ]

stdout | src/test/moves/shell_side_arm.test.ts > Moves - Shell Side Arm > should not make contact if the move becomes special
[ '%cStart Phase MovePhase', 'color:green;' ]

stdout | src/test/moves/shell_side_arm.test.ts > Moves - Shell Side Arm > should not make contact if the move becomes special
[ 'SHELL_SIDE_ARM' ]
[ '%cStart Phase MessagePhase', 'color:green;' ]

stdout | src/test/moves/shell_side_arm.test.ts > Moves - Shell Side Arm > should not make contact if the move becomes special
[ 'Xurkitree used Shell Side Arm!' ]
[ '%cStart Phase MoveEffectPhase', 'color:green;' ]

stdout | src/test/moves/shell_side_arm.test.ts > Moves - Shell Side Arm > should not make contact if the move becomes special
[ 'NO CONTACT' ]
[ 'crit stage: +0' ]
[ 'NO CONTACT' ]
[ 'base damage', 161.8754098360656, 'Shell Side Arm', 90, 387, 183 ]
[ 'damage', 161, 'Shell Side Arm' ]
[ '%cStart Phase DamageAnimPhase', 'color:green;' ]

stdout | src/test/moves/shell_side_arm.test.ts > Moves - Shell Side Arm > should not make contact if the move becomes special
[ '%cStart Phase DamageAnimPhase', 'color:green;' ]

stdout | src/test/moves/shell_side_arm.test.ts > Moves - Shell Side Arm > should not make contact if the move becomes special
[ '%cStart Phase ShowAbilityPhase', 'color:green;' ]

stdout | src/test/moves/shell_side_arm.test.ts > Moves - Shell Side Arm > should not make contact if the move becomes special
[ '%cStart Phase MessagePhase', 'color:green;' ]

stdout | src/test/moves/shell_side_arm.test.ts > Moves - Shell Side Arm > should not make contact if the move becomes special
[ "Wild Slowbro's Rough Skin\nhurt its attacker!" ]
[ '%cStart Phase MoveEndPhase', 'color:green;' ]

AssertionError: expected 276 to be 242 // Object.is equality
```

So it's definitely reaching the line where it's logging NO CONTACT but it's still having contact.

geeilhan:
> That is weird. 
> So it's definitely reaching the line where it's logging NO CONTACT but it's still having contact.

Yes I found the same issue so I thought that maybe MAKES_CONTACT is set when the constructor is called or somewhere else at the beginning but it not set anywhere

geeilhan:
> > That is weird.
> > So it's definitely reaching the line where it's logging NO CONTACT but it's still having contact.
> 
> Yes I found the same issue so I thought that maybe MAKES_CONTACT is set when the constructor is called or somewhere else at the beginning but it not set anywhere

So the solution for the tests to work was to move the no-contact-test to the top of the automated tests...

geeilhan (commented):
Tests should work now

Tempo-anon:
Yep looks like tests pass with that. Strangely, replacing the code block
```ts
    if (move.hasFlag(MoveFlags.MAKES_CONTACT)) {
      move.makesContact(false);
    }
```
with just `move.makesContact(False);` results in the test failing 🤔

Tempo-anon (dismissed):

src/data/move.ts
```diff
@@ -4612,13 +4612,19 @@ export class ShellSideArmCategoryAttr extends VariableMoveCategoryAttr {
...
+
+    if (shouldUsePhysical) {
       category.value = MoveCategory.PHYSICAL;
+      move.makesContact();
       return true;
     }
+    /** MoveFlags are not reset every turn so if this flag is set it needs to be reset if the move is a special attack */
+    if (move.hasFlag(MoveFlags.MAKES_CONTACT)) {
+      move.makesContact(false);
+    }
```
flx-sta, line 4627: should probably wrap it in a descriptive method (private?).

e.g.
```ts
private resetFlags() {
// MoveFlags are not reset every turn so if this flag is set it needs to be reset if the move is a special attack
 if (move.hasFlag(MoveFlags.MAKES_CONTACT)) {
      move.makesContact(false);
    }
}

geeilhan:
> Yep looks like tests pass with that. Strangely, replacing the code block
> 
> ```ts
>     if (move.hasFlag(MoveFlags.MAKES_CONTACT)) {
>       move.makesContact(false);
>     }
> ```
> 
> with just `move.makesContact(False);` results in the test failing 🤔

I think that is because `move.makesContact(False);` XORs the flag MAKES_CONTACT and not removes it. 
This would basically set the MAKES_CONTACT flag in some situations instead of removing the flag if `move.makesContact(False);` is called unconditionally.

geeilhan:
> Yep looks like tests pass with that. Strangely, replacing the code block
> 
> ```ts
>     if (move.hasFlag(MoveFlags.MAKES_CONTACT)) {
>       move.makesContact(false);
>     }
> ```
> 
> with just `move.makesContact(False);` results in the test failing 🤔

This does seem weird

src/data/move.ts
geeilhan: Added this method as descriptive method.

SirzBenjie:
So the fix for this works but it is a bit hacky in that it adds state to the move attribute itself.
I think we can port this over when we eventually revisit fixing shell side arm. Hopefully sooner rather than later.

PR #5028, opened 2024-12-21
Toggle cursor run

Wlowscha:
A new on/off toggle is added to the Game Settings/Display menu, which is on by default corresponding to the current game behavior. 
Game behavior is only affected after a user switches the toggle off. The result is that the cursor does not reset after fleeing or catching a wild pokémon, unless the new battle is a trainer, mystery encounter, or is in the End.

Several users have expressed that the ability to let the cursor not reset (not go back to "Fight" at the beginning of each battle) would positively affect their experience. Having this as a toggle allows to leave experience unchanged for all other users.

A new variable commandCursorReset has been added to the BattleScene class in battle-scene.ts.
In command-phase.ts, at the beginning of each battle, this variable is checked to decide whether the cursor should reset.
A new toggle has been added in settings.ts.

## Screenshots/Videos
Screenshot of the new toggle in menu (without the changes to locales implemented yet).
new_toggle
Screenshot of tests passing.
tests_passing

## How to test the changes?
I made use of overrides to increase the frequency of mystery encounters and make sure they did not lead to anomalous behaviour.
The most direct way to test the changes is start a new run, switch the toggle off, and start a new battle after running or catching a wild pokémon.

## Checklist
- [x ] **I'm using `beta` as my base branch**
- [x ] There is no overlap with another PR?
- [x ] The PR is self-contained and cannot be split into smaller PRs?
- [x ] Have I provided a clear explanation of the changes?
- [x ] Have I tested the changes manually?
- [x ] Are all unit tests still passing? (`npm run test`)
- [x ] Have I provided screenshots/videos of the changes (if applicable)?

- [x ] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?
  - [x ] If so, please leave a link to it here: https://github.com/pagefaultgames/pokerogue-locales/pull/87
- [x ] Has the translation team been contacted for proofreading/translation?

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -147,6 +147,7 @@ export default class BattleScene extends SceneBase {
   public damageNumbersMode: integer = 0;
   public reroll: boolean = false;
   public shopCursorTarget: number = ShopCursorTarget.REWARDS;
+  public commandCursorReset: boolean = true;
   public showMovesetFlyout: boolean = true;
   public showArenaFlyout: boolean = true;
   public showTimeOfDayWidget: boolean = true;
--- a/src/phases/command-phase.ts
+++ b/src/phases/command-phase.ts
@@ -35,8 +35,14 @@ export class CommandPhase extends FieldPhase {
     this.scene.updateGameInfo();
 
     const commandUiHandler = this.scene.ui.handlers[Mode.COMMAND];
+
+    // If one of these conditions is true, we always reset the cursor to Command.FIGHT
+    const cursorResetBattle = this.scene.currentBattle.battleType === BattleType.MYSTERY_ENCOUNTER ||
+                              this.scene.currentBattle.battleType === BattleType.TRAINER ||
+                              this.scene.arena.biomeType === Biome.END;
+
     if (commandUiHandler) {
-      if (this.scene.currentBattle.turn === 1 || commandUiHandler.getCursor() === Command.POKEMON) {
+      if ((this.scene.currentBattle.turn === 1 && (this.scene.commandCursorReset || cursorResetBattle)) || commandUiHandler.getCursor() === Command.POKEMON) {
         commandUiHandler.setCursor(Command.FIGHT);
       } else {
         commandUiHandler.setCursor(commandUiHandler.getCursor());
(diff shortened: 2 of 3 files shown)


Snailman11:
Could you record a video of the changes the player would experience during a run?

DayKev:
I think it would make more sense to call the option "Cursor Memory" and then default to off (current behavior).

DayKev:
Also this makes more sense to put in the "General" category than "Display".

Wlowscha:
> Could you please record a video of the changes the player would experience during a run?

Here are two videos: 

The first shows how the cursor goes back to FIGHT when the toggle is on and instead stays on RUN when the toggle is off.
https://github.com/user-attachments/assets/545d43ae-59fb-4b3b-b0cb-3a7a6d4ae658

The second video shows how the toggle goes back to FIGHT for a trainer battle, even though the toggle is off.
https://github.com/user-attachments/assets/18627b7a-e6fa-48b3-80b4-6915b2fd16fb

Wlowscha:
> I think it would make more sense to call the option "Cursor Memory" and then default to off (current behavior).

This sounds like a good idea, the current behavior corresponding to the "off" state is probably more intuitive.
Could it be confusing for users to only call it "Cursor Memory" as opposed to, for example, "Battle Cursor Memory"?


> Also this makes more sense to put in the "General" category than "Display".

I put it in Display because that's where the toggle for the position of the Shop Cursor also is, but I don't mind moving it around if it makes more sense to have it under General.

DayKev:
> > I think it would make more sense to call the option "Cursor Memory" and then default to off (current behavior).
> 
> This sounds like a good idea, the current behavior corresponding to the "off" state is probably more intuitive. Could it be confusing for users to only call it "Cursor Memory" as opposed to, for example, "Battle Cursor Memory"?

Hm, possibly. I think either way is fine.

> 
> > Also this makes more sense to put in the "General" category than "Display".
> 
> I put it in Display because that's where the toggle for the position of the Shop Cursor also is, but I don't mind moving it around if it makes more sense to have it under General.

Ah, that one should also probably be moved at some point lol.

Wlowscha:
@DayKev I have made changes as you requested. In particular, I changed the name of the variables to Cursor Memory while inverting the logic, and I moved the option to the general settings (see picture below).
general_settings

DayKev (approved):
Seems to work. Just needs the locales PR to be merged https://github.com/pagefaultgames/pokerogue-locales/pull/87

Then it's up to @Madmadness65 and @SangaraSorama

DayKev:
Locales PR merged, just waiting on approval from @Madmadness65

Madmadness65 (approved):

Issue #4963, opened 2024-12-03
[Bug] Destiny Bond, Innards Out Crashes when Activated while Future Sight User not on Field


Snailman11:
### Describe the bug

If Future Sight KOs/faints a Pokémon with Innards Out or with Destiny Bond Active, and the user of Future Sight is not on the Field, the game freezes.

Link to Discord's Bug Report (Destiny bond + future sight crash)
https://discord.com/channels/1125469663833370665/1312241949499064331

### Reproduction

Use Croagunk for best results and use ROAR on the first turn.
-KO or Phase or Let the AI arbitrarily switch after it uses Future Sight

```
const overrides = {
STARTING_WAVE_OVERRIDE: 25,
STARTING_LEVEL_OVERRIDE: 5,
MOVESET_OVERRIDE: [Moves.ROAR, Moves.SPLASH, Moves.DESTINY_BOND],
ABILITY_OVERRIDE: Abilities.PRANKSTER,
OPP_SPECIES_OVERRIDE: Species.WINGULL,
OPP_LEVEL_OVERRIDE: 5,
OPP_MOVESET_OVERRIDE: [Moves.FUTURE_SIGHT]
}

```

### Expected behavior

> Destiny Bond does not activate from indirect damage such as weather damage, damage from Leech Seed, and Future Sight.
https://bulbapedia.bulbagarden.net/wiki/Destiny_Bond_(move)
-Bulba

### Screenshots / Videos

![Screenshot (460)](https://github.com/user-attachments/assets/9d054e78-a5f3-4182-aba2-8233618d1229)

**Future Sight and Destiny Bond**

https://github.com/user-attachments/assets/97d66ac8-00e5-4c27-a41c-faa55d570c74

**Future Sight and Innards Out Crashes**

https://github.com/user-attachments/assets/47a1e910-719e-442a-bb58-85ad312911ac

**Future Sight and Aftermath does not Crash**

https://github.com/user-attachments/assets/1b4024ba-d95d-45a0-bc23-d1815f55ec58

### Additional context

I have not checked this from the Point of View of the Player's Future Sight.

Snailman11:
Reddit reported this occurring with Gulp Missile
https://www.reddit.com/r/pokerogue/comments/1hbkdi3/the_game_breaks_when_cramorants_gulp_missle/

PR #5044, opened 2024-12-27
[Bug] Introducing a check for pokémon being on field to activate abilities

Wlowscha:
Abilities such as Neutralizing Gas stop working after the user is out of the field, whether this is due to switching, a move such as roar, or an ability such as wimp out.
Additionally, Future Sight will now only hit once when used by a Pokémon with Parental Bond, if the user switches out before Future Sight hits (intended behavior).

It is well known that the effect of Neutralizing Gas persists after a mon is hit by a forced switch move, see e.g. discord issue
https://discord.com/channels/1125469663833370665/1320888834334130186
The change brings the gameplay in line with player expectations.

In the function canApplyMove(...), a new check is introduced to confirm that the ability user is on the field.

A few tests failed, because of not correctly adding mons to game.scene.field (Effectiveness) or because of checking if a mon not on the field has an ability (Pastel Veil). As far as my understanding goes, the issue in this case is within the tests themselves, and they have been changed accordingly. In particular, hasAbility is only used on mons on the field elsewhere, which suggests that is the only intended use case.

The test for Parental Bond interacting with Future Sight when the user leaves the field has been updated to reflect the expected result (according to a TODO in the test itself).

Tests for abilities that increase the rates of double battles have been refactored to be more robust and deterministic. The previous versions were liable to return a positive due to naturally generating a double battle rather than due to the ability itself.

The only situation where tests actually failed is for abilities which activate before switching, when the switch is forced by a move or ability. In this case, the switched mon is removed from game.scene.field by the leaveField method _before_ the SwitchSummonPhase, which is when these abilities are triggered. To accomodate this case, all abilities with a preSwitch trigger have been given a new bypassOnField attribute, which lets them work as intended (see for example the tests for Wimp Out and U-turn, in particular their interaction with Regenerator).
Note that a better fix might come from refactoring the order of calls in ForceSwitchOutHelper and dependent functions, to ensure that the leaveField method is always called _after_ all preSwitch abilities have triggered. This would remove the need for a bypassOnField attribute altogether, but would require significant intervention on the codebase.

## Screenshots/Videos

Before this PR (wonder guard doesn't come back on after Koffing is roared out):
https://github.com/user-attachments/assets/513388b7-d144-4550-bdd0-15f40d11fb85

After this PR (Neutralizing gas stops working, and Wonder Guard comes back up as expected):
https://github.com/user-attachments/assets/683d34fd-9ec0-4f0a-82b4-52b2275433e7

## How to test the changes?
Use the following override:
const overrides = {
  ABILITY_OVERRIDE: Abilities.WONDER_GUARD,
  MOVESET_OVERRIDE: [ Moves.ROAR, Moves.TACKLE, Moves.SPLASH ],
  OPP_SPECIES_OVERRIDE: Species.KOFFING,
  STARTING_BIOME_OVERRIDE: Biome.SLUM,

Go through battles until you find an encounters where one Koffing has Neutralizing Gas, while the other has a different ability. Use Roar on the Neutralizing Gas Koffing. After doing this, your active mons should be immune to Tackle as Wonder Guard is active again.

## Checklist
- [x ] **I'm using `beta` as my base branch**
- [x ] There is no overlap with another PR?
- [x ] The PR is self-contained and cannot be split into smaller PRs?
- [x ] Have I provided a clear explanation of the changes?
- [x ] Have I tested the changes manually?
- [x ] Are all unit tests still passing? (`npm run test`)
  - [x ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [x ] Have I provided screenshots/videos of the changes (if applicable)?

Are there any localization additions or changes? No


Snailman11:
If your PR is ready, you can call it out in
https://discord.com/channels/1125469663833370665/1320074624096145458/1321925787049398312

for increased visibility

Wlowscha:
This PR also fixes the issue of Storm Drain persisting after the user is roared out, also pointed out here:
https://discord.com/channels/1125469663833370665/1320888834334130186

Video for proof:
https://github.com/user-attachments/assets/d2615ca6-8077-4bab-ba23-d3bdb616f32b

To replicate, use overrides

const overrides = {
  MOVESET_OVERRIDE: [ Moves.ROAR, Moves.WATER_GUN, Moves.SPLASH ],
  OPP_SPECIES_OVERRIDE: Species.GASTRODON,
  STARTING_BIOME_OVERRIDE: Biome.SWAMP,

and find an encounter with one Storm Drain and one Sticky Hold Gastrodon.

Snailman11:
https://github.com/pagefaultgames/pokerogue/blob/10e0f9f0deaf7645ce059b1cd5b5120e6c1fd148/src/data/move.ts#L8686
      .partial() // cannot be used on multiple Pokemon on the same side in a double battle, hits immediately when called by Metronome/etc, should not apply ##abilities or held items if user is off the field

Perhaps Abilities specifically could be removed from Future Sight's section in move.ts

Wlowscha:
> https://github.com/pagefaultgames/pokerogue/blob/10e0f9f0deaf7645ce059b1cd5b5120e6c1fd148/src/data/move.ts#L8686
> 
> 
> .partial() // cannot be used on multiple Pokemon on the same side in a double battle, hits immediately when called by Metronome/etc, should not apply **abilities** or held items if user is off the field
> Perhaps Abilities specifically could be removed from Future Sight's section in move.ts

This PR would fix the issue with abilities belonging to the user. I believe there are some interactions with abilities which belong to the _target_ which still break the game if the future sight user is off the field, see e.g. #4963  
So the text mentioning abilities is still relevant, but I can update it for more clarity :)

Issue #5086, opened 2025-01-05
[Bug] Defog does not Clear Opposing Safeguard and Mist


Snailman11:
### Describe the bug

Defog does not clear Enemy Safeguard or Mist

https://github.com/pagefaultgames/pokerogue/blob/1be95e8a1a36ec8ac9ac8b112d45b0abc61af496/src/data/move.ts#L9218
-No Attributes that clear Safeguard or Defog are present in Defog's section.

### Reproduction

Use Defog while Safeguard or Mist is active on the opposing side.

### Expected behavior

> Defog now clears away Spikes, Toxic Spikes, Stealth Rock, and Sticky Web on both the user's and target's side of the field. In addition to Light Screen, Reflect, Safeguard, and Mist, it also clears away Aurora Veil from the target's side of the field only.

### Screenshots / Videos

Defog should Clear away Enemy Mist and Safeguard

https://github.com/user-attachments/assets/cf0ee479-142e-40c6-8d4c-7961a6a00cea

PR #5107, opened 2025-01-10
[Bug] Fixed defog not removing the target's Safeguard and Mist

Zain-A-Abbas:
Defog will now successfully remove the Safeguard and Mist from the target side.

Fixes https://github.com/pagefaultgames/pokerogue/issues/5086.

~~Added `RemoveSafeguard` and `RemoveMist` classes to `moves.ts`, and has them be called by Defog with both sides set to false.~~ Used `RemoveArenaTagsAttr` for Defog in `moves.ts`. Added `defog.test.ts` as well.

## How to test the changes?
Using defog on an enemy who has Safeguard or Mist active would verify the changes work.

## Checklist

### DIFF
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -5814,60 +5814,6 @@ export class RemoveArenaTrapAttr extends MoveEffectAttr {
 }
 
 
-export class RemoveSafeguard extends MoveEffectAttr {
-
-  private targetBothSides: boolean;
-
-  constructor(targetBothSides: boolean = false) {
-    super(true, { trigger: MoveEffectTrigger.PRE_APPLY });
-    this.targetBothSides = targetBothSides;
-  }
-
-  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-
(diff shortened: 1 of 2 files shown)


DayKev:
So a couple comments:
- Rather than make a new class for each of these tags, the existing class (`RemoveArenaTagsAttr`) should probably be updated to properly account for the issue
- Move and Ability additions/changes/bugfixes need tests added (you can use `npm run create-test` to create the test file if one doesn't already exist for the relevant move or ability), look at the other tests as an example (if you can find one that handles a similar scenario that's a good place to start).

Zain-A-Abbas:
Made the requested changes. Let me know if anything else is needed or if I made a mistake.

src/data/move.ts
```diff
@@ -5813,6 +5813,7 @@ export class RemoveArenaTrapAttr extends MoveEffectAttr {
   }
 }
 
+
```
DayKev, line 5816:
```suggestion
```

DayKev (approved):

PR #5121, opened 2025-01-13
[Balance] Fix off by one error in Generate random biome and Daily Mode now has its own function

Jimmybald1:
Endless now allows random traverse (random biome from End at every 50th wave) to Island and Laboratory. Also fixed the off by one error that led to the weights not being accurate with the results.
Daily mode now has its own function to get a random biome for its starting biome. It also now uses a custom weight table approved by the Balance Team:
image
This table is weighted higher for biomes that have a lot of exit paths or lead to a large amount of final paths in general.

It is simply not fun to have biomes be unavailable when there is no balance reason to have them be unavailable. And Daily Mode had no business using a function that was balanced around having 250+ waves.

Another list with biome weights. Which means it would have to be included in the biome rework as well. It is still useful to have this happen before the biome rework, because the off by one and the blocked biomes based on an enum value would have had to be fixed anyway.

## Screenshots/Videos
Previously unavailable starting biomes are now available:
image
image

New rough odds by simply executing the function 100,000x:
image

## How to test the changes?
Start a few dailies with a different seed (there are no overrides for this sadly) and previously unavailable biomes will show up now.
To confirm the weights are accurate I wrote a simple loop that called the function and counted the results. I don't think testing how often biomes show up is a good idea in a test?

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1865,7 +1865,7 @@ export default class BattleScene extends SceneBase {
 
   generateRandomBiome(waveIndex: integer): Biome {
     const relWave = waveIndex % 250;
-    const biomes = Utils.getEnumValues(Biome).slice(1, Utils.getEnumValues(Biome).filter(b => b >= 50).length * -1);
+    const biomes = Utils.getEnumValues(Biome).filter(b => b !== Biome.TOWN && b !== Biome.END);
     const maxDepth = biomeDepths[Biome.END][0] - 2;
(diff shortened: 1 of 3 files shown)


Snailman11:
Link to Discord's Bug Report (RNG Issues in Endless Biome Pathing (or otherwise without map))
https://discord.com/channels/1125469663833370665/1304143826977554452

damocleas (approved):

DayKev (approved):

PR #4838, opened 2024-11-10
[Misc] Move Revival Blessing func from `PlayerPokemon` to the move's attr

DayKev:
## What are the changes the user will see?
N/A

## Why am I making these changes?
A move shouldn't be adding arbitrary methods to the `Pokemon` classes.

## What are the changes from a developer perspective?
Removed `revivalBlessing()` method from the `PlayerPokemon` class and moved its functionality to Revival Blessing's move attr.
Removed an unused parameter from `RevivalBlessingAttr`'s constructor.
Updated some tsdocs.
Added tests for Revival Blessing.

## How to test the changes?
`npm run test revival_blessing`

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I considered writing automated tests for the issue?
- ~[ ] If I have text, did I make it translatable and add a key in the English locale file(s)?~
- ~[ ] Have I tested the changes (manually)?~
    - [x] Are all unit tests still passing? (`npm run test`)
- ~[ ] Are the changes visual?~
  - ~[ ] Have I provided screenshots/videos of the changes?~

PR #5141, opened 2025-01-18
[Bug] Fix #2769 Revival Blessing Softlock in doubles

### LINKED ISSUE #2769: [BUG] Revival Blessing Breaks Double Battles
**Describe the bug**

When battling against two trainers in a double battle, if a Pokemon is revived with revival blessing while a slot is open, it instantly returns to the field. At the end of the turn, another pokemon is sent out, making the revived Pokemon untargetable, eventually leading to only attacking ones own Pokemon.

**To Reproduce**

1. Go to a trainer double battle
2. Defeat One Pokemon before the Revival Blessing Mon acts and leave it up, if it clicks Revival Blessing, the bug occurs
3. Make sure Trainer has an extra Pokemon to throw out.
4. See extra mon Error
5. Defeat all Pokemon
6. See Unwinnable Error

**Expected behavior**

When a Pokemon is revived, either it doesn't instantly return to the field, or doesn't allow for a second Pokemon to be added on top of it at end of turn.

Screenshots / Videos**

https://github.com/pagefaultgames/pokerogue/assets/174589289/e0c6c1a5-62f7-482e-add2-b034c35b9e19

Screenshot_20240703_164806_Samsung Internet

https://github.com/pagefaultgames/pokerogue/assets/174589289/709ab987-ee63-43cc-8f2d-6f7158646fdb

**Device**

Playing on Samsung Internet on a Galaxy [S24]

**Additional context**

emdeann:
- Revival Blessing doesn't cause a softlock in doubles
- Pokemon who were fainted and revived in the same turn can't still use their moves
- Enemy pokemon will get placed in the correct location after the last remaining pokemon uses Revival Blessing

Fixes #2769 

The game softlocked because the Revival Blessing implementation queued a `SwitchSummonPhase` if the ally Pokemon was fainted. This worked fine for the player, but failed for the enemy because the enemy immediately queues a switch from `FaintPhase` and the revival blessing implementation failed to clear it.

i.e.: The enemy has 3 pokemon: One gets fainted and `FaintPhase` queues a switch into that slot. Then the other uses Revival Blessing and also queues a switch into the same slot, resulting in two mons getting sent into the same slot and a softlock.

I added a check to remove a queued switch phase from the enemy if revival blessing is used, which fixes the issue. I also fixed two other bugs with the implementation:
- The `MovePhase` was not cancelled by revival blessing, so if the same Pokemon fainted and was revived before moving it would still get its move off, which didn't match mainline behavior
- In the case where only one enemy was left, revival blessing wouldn't reposition it from the center, so you'd end up with one mon in the center and one on the right instead of on the left and right

In the process of doing this I also moved the player `revivalBlessing()` function from pokemon.ts to move.ts because what was it doing there in the first place

## Screenshots/Videos
No more weird switch in

https://github.com/user-attachments/assets/dada5975-bcb1-4ffb-97fb-10d25aba287b

## Checklist
  - Daykev's [#4838](https://github.com/pagefaultgames/pokerogue/pull/4838) does the same refactor I did (but pre global scene) so it would become redundant


Madmadness65 (approved):

damocleas (approved):

SirzBenjie:
This will conflict with #5283

emdeann:
Waiting to fix conflicts here until the above is merged, agreed that it should definitely go first

DayKev:
Looks good. Can you write a test for this? Should be good to go after that.

Wlowscha (approved):

DayKev (approved):

PR #5146, opened 2025-01-18
[Bug] Fix moves/abilities which modify abilities

### LINKED ISSUE #3905: [Bug] Skill Swap, Role Play, Entrainment, Trace, and Neutralizing Gas(P) Do not Reactivate abilities when Gaining them or After Suppression Wears off
### Describe the bug

The above abilities and moves don't "activate" abilities when they should, ex. Intimidate, Download

### Session export file

_No response_

### User data export file

_No response_

### Expected behavior

(1a, 2, 4)
https://replay.pokemonshowdown.com/gen9customgame-2190559634-saoooyiccf7ranxqatw26ueli22pn57pw
(1b)
https://replay.pokemonshowdown.com/gen9customgame-2190582913-oc7pvy340gfps3t9w92gzsa3kafwgozpw
(3)  https://replay.pokemonshowdown.com/gen9customgame-2190602831-7cpl0g8xo9bw72mk1t1aekva8o8iht5pw

1a. Trace activates an ability when copying it.
1b. Trace waits for a copiable ability to enter the field on the opposing side and copies it.
2. Skill Swap "activates" both abilities.
3. Role Play and Entrainment activate gained abilities, (fails if the ability are the same - Implemented)
4. Neutralizing Gas leaving the field "activates" the ability.

> 1a. While a Pokémon with Trace is in battle, if at any time an opponent has an Ability that can be copied by Trace, Trace will activate and replace itself with that Ability. (Doesn't actually explicitly mention that the new ability gets activated, but it does)

> 1b. While a Pokémon with Trace is in battle, if at any time an adjacent opponent has an Ability that can be copied by Trace, Trace will activate and replace itself with that Ability.

> 2. If both Pokémon have the same Ability, Skill Swap will succeed. If it is an Ability such as [Intimidate](https://bulbapedia.bulbagarden.net/wiki/Intimidate_(Ability)) that activates when it is gained, it will activate for both Pokémon after being gained this way.

> 4. Once Neutralizing Gas is no longer applying, Abilities that activate when gained (i.e. Abilities that activate when a Pokémon is switched in, such as Intimidate) immediately pop up an activation message regardless of whether they have already done so, with the exceptions of Unnerve (which still applies but without displaying a message) and Imposter (which only activates when switched in, not when gained). If Neutralizing Gas stops applying because the Pokémon with Neutralizing Gas is about to switch out, those Abilities activate before the switch occurs (e.g. if Intimidate is activated this way, it will decrease the Attack of the Pokémon with Neutralizing Gas before it switches out, rather than decreasing the Attack of the Pokémon who switches in to replace it).

### Screenshots / Videos

**Trace (Intimidate)**

https://github.com/user-attachments/assets/e7ccffc8-b31b-49b7-b96c-4b4e383557f4

**Skill Swap (Intimidate and Download)**

https://github.com/user-attachments/assets/59b845a3-ba15-4e06-92db-a563af157592

**Role Play (Intimidate)**

https://github.com/user-attachments/assets/b30244ba-fc0f-4aaa-8914-0ce7c5f0d9bb

**Entrainment (Download)**

https://github.com/user-attachments/assets/e3ffe1c9-44e6-4ef0-8aa2-46764b622d55

**Neutralizing Gas (Intimidate)**

https://github.com/user-attachments/assets/f387dded-3c91-4e49-b91a-36c961ead9f8

**(Delta Stream does not reactivate)**

https://github.com/user-attachments/assets/f3ec8241-d96c-4885-b0bc-aa82b4cbf695

**Neut. Gas into Neut. Gas (Ability should apply as Neut gas switches out)**

https://github.com/user-attachments/assets/5a3dfcc0-3cac-4d40-954a-1dfcc9662725

Other

### If other please specify

gameplay

### Which browser do you use?

Other

### If other please specify

gameplay

### Additional context

_No response_

emdeann:
Moves like skill swap/entrainment and abilities like trace properly activate gained abilities.

i.e. A Pokemon gains intimidate and intimidate immediately activates - any ability which activates on summon should also activate on obtaining it for the first time.

To match mainline behavior - will *mostly* fix #3905 but I haven't touched neutralizing gas as that's a different thing.

- Added a `setTempAbility()` function to pokemon.ts which automatically applies the new ability if needed and clears primal weather (if the original ability was a primal weather ability). This should be used in place of directly modifying `pokemon.summonData` in most cases where an ability is changed during the battle.
- Added a couple helper functions to ability.ts as well as slightly modifying `applyAbAttrsInternal`. This was necessary because I needed a way to ignore passive abilities (they shouldn't be reapplied when the main ability changes) without rewriting entire ability application methods. I just added a boolean `considerPassive` which defaults to true but can be turned off so that `applyAbAttrsInternal` doesn't call the apply function with the passive ability.
- Modified all existing moves/abilities which change abilities to use `setTempAbility()` (at least, I think I got them all)
- Added automated tests for the above moves/abilities

I am aware that the ability bar popouts/ability text are very janky for a lot of this, but that problem goes way beyond these changes. Ideally an ability would have a lot more agency over when exactly its message/bar get displayed, but that's a major change and not in my scope here

## Screenshots/Videos
Skill Swap + Intimidate:

https://github.com/user-attachments/assets/42c891ba-ee94-4e31-b206-08dc746b843f

Trace + Intimidate:

https://github.com/user-attachments/assets/3b0e2c13-add8-4b34-9345-1d0f9146f84e

Role Play + Intimidate + Desolate Land ending

https://github.com/user-attachments/assets/c4dee323-759c-473e-b02e-18834c1b9400

...etc
Let me know if there's another video you want me to grab, the amount of possible test cases here kinda hurts my head

## How to test the changes?
Added automated tests

## Checklist


Madmadness65 (dismissed):
Seems to be working as intended from the testing I did locally, but should probably get a second opinion too.

Xavion3:
It looks like in the trace mp4 that intimidate shows up and activates before trace does, this feels likely to induce confusion and probably not what we want.

emdeann:
@Xavion3 Absolutely is an issue but such issues exist throughout the entire game and I don't feel it makes sense to address here.

> I am aware that the ability bar popouts/ability text are very janky for a lot of this, but that problem goes way beyond these changes. Ideally an ability would have a lot more agency over when exactly its message/bar get displayed, but that's a major change and not in my scope here

i.e. for the trace issue you mentioned, fixing it would require the trace message to be displayed before it is applied, but that's not possible because `target` isn't defined until the ability is applied. imo fixing that in a way that isn't hacky and specific to every individual ability would require a major refactor of how abilities are shown. Once that's done this would be easy to fix

Wlowscha:
According to https://bulbapedia.bulbagarden.net/wiki/Imposter_(Ability), _"Imposter only activates when the Pokémon with the Ability is sent out; if it fails to transform into an opponent, it will not activate unless the Pokémon with Imposter is withdrawn and sent out again. Imposter will also not activate when obtained during battle."_ Is this currently supported?

emdeann:
Illusion isn't implemented yet, but I added a `NoMidTurnActivationAttr` and tacked it onto illusion's definition. Any future ability which is both post-summon and only activates one (why, game freak) can just have this attr added to it

Madmadness65 (approved):

Xavion3:
Can we possibly go for just a separate attribute for "trigger when gained" effects? It'd achieve the same effect, but be a positive restriction instead of a negative one. This would be better future proofed for any other weird abilities or potential new abilities that do one thing on gain that isn't exactly the same as what happens on summon, and avoid introducing edge cases into summon attributes by making those not actually based purely on summoning anymore.

Also it's currently an undocumented edge case, so at minimum that'd have to be fixed even if we go with the current less rigorous approach.

emdeann:
I thought about doing it that way and I really didn't prefer it because afaik every single post-summon ability activates on gain and vice versa (illusion is better described as a "pre summon" ability as in that PR). As a model it might make more sense to make a new attribute, but I wouldn't consider it less rigorous to make the hypothetical edge case handle its own business as opposed to changing every existing post summon ability

Wlowscha:
I agree with Xavion on this point. One solution would be to have a `MidTurnAbAttr`, called by `applyMidTurnAbAttr`, and then a derived class `MidTurnSummonAbAttr` (or something to this effect) which, as part of its `.apply`, checks that the ability also has a `PostSummonAbAttr` and calls that. Then we can give `MidTurnSummonAbAttr` to all the abilities that need it, and exclude the few cases that don't, without having to define a separate attribute for each `PostSummonAbAttr`.

On a separate note, I feel that "MidTurn" is not very informative for what is doing---when I read it, it makes me wonder _what_ is happening in the middle of the turn. Since this is always tied to receiving or reactivating an ability, maybe something like "OnGain" would be more descriptive. This is more of a personal gripe, however.

src/data/move.ts
```diff
@@ -7526,11 +7526,12 @@ export class SwitchAbilitiesAttr extends MoveEffectAttr {
...
-    target.summonData.ability = tempAbilityId;
+    const tempAbility = user.getAbility();
 
     globalScene.queueMessage(i18next.t("moveTriggers:swappedAbilitiesWithTarget", { pokemonName: getPokemonNameWithAffix(user) }));
+
+    user.setTempAbility(target.getAbility());
+    target.setTempAbility(tempAbility);
     // Swaps Forecast/Flower Gift from Castform/Cherrim
     globalScene.arena.triggerWeatherBasedFormChangesToNormal();
     // Swaps Forecast/Flower Gift to Castform/Cherrim (edge case)
```
Wlowscha, line 7537: `globalScene.arena.triggerWeatherBasedFormChanges();` here should not be needed anymore, since both Forecast and Flower Gift have a PostSummon ability attribute which will be called by setTempAbility.

Madmadness65 (dismissed):

emdeann:
@Wlowscha I agree with your point on naming, went ahead and changed that. 

The way I'm looking at "on gain" vs. "post summon" is that they literally _are_ the same thing - when a pokemon is sent out it "gains" the ability for the first time and the effects are no different than if it gains the ability during the battle. It would almost make more sense to have the base attribute named "OnGain" instead of "PostSummon" in that sense, it's more general. If there's some ability I'm not thinking of that goes against that idea then I'd happily change it, but I just don't think it makes sense right now.

also apologies to mad who has approved this pr at least 50 times

Madmadness65 (dismissed):

Xavion3:
So, a question I've got upon reviewing this properly now and how it approaches it.

Why have the temporary ability code at all? Wouldn't it make more sense to just proc the newly gained onGain abilities instead?

SirzBenjie (approved):
I really like this PR.
You added a tiny bit of code to have a ton of impact. The method name is clear and it is intuitive to use and understand.
I wouldn't change a thing.

Xavion3:
So, realised some merits, but facing a different issue.

Is there any way to make this code support passives changing? Since we can do that now as of 1.6? (might've been 1.5, the versions blur a bit.)

SirzBenjie:
> So, realised some merits, but facing a different issue
> Is there any way to make this code support passives changing? Since we can do that now as of 1.6? (might've been 1.5, the versions blur a bit.)

It was 1.6

src/data/ability.ts
```diff
@@ -4852,50 +4868,69 @@ async function applyAbAttrsInternal<TAttr extends AbAttr>(
...
+) {
+  const ability = passive ? pokemon.getPassiveAbility() : pokemon.getAbility();
+  if (gainedMidTurn && ability.getAttrs(attrType).some(attr => attr instanceof PostSummonAbAttr && !attr.shouldActivateOnGain())) {
+    return;
+  }
 
-      globalScene.setPhaseQueueSplice();
+  for (const attr of ability.getAttrs(attrType)) {
+    const condition = attr.getCondition();
+    if ((condition && !condition(pokemon))) {
```
DayKev, line 4901:
```suggestion
    if (condition && !condition(pokemon)) {
```

DayKev (approved):
Well it adds even more `apply*AbAttrs` functions but whatever, just means more work for whoever wants to take care of that mess later lol

DayKev (approved):

Wlowscha (approved):

PR #5129, opened 2025-01-15
[Move] Improve implementation of Rage Fist damage increase

geeilhan:
## What are the changes the user will see?

None except higher damage numbers

## What are the changes from a developer perspective?

Added `hitRecCount` to `CustomPokemonData` which helps in calculating the hitCount of the current arena.
`prevHitCount ` was also added to `PokemonBattleData` to calculate the value that needs to be added to `hitRecCount`.

`hitRecCount` is reset in `doEncounter()` in the `new-encounter-biome-phase.ts` (entering new arena), as well as in the `encounter-phase.ts` should the encounter be a trainer battle. This prevents the counter from being reset if the pokemon should faint and then be revived or if the pokemon got switched out.

What is still missing is the increase in `hitCount` in `PokemonBattleData` should the pokemon hurt itself during confusion.

## Screenshots/Videos

There are no visual changes. The move worked as intended and only the proper damage calculation was missing

## How to test the changes?

Automated tests implemented. 
Some automated tests are currently marked as todo but these test have been tested manually and passed.

For manual tests use:

```
const overrides = {
  OPP_SPECIES_OVERRIDE: Species.MAGIKARP,
  OPP_LEVEL_OVERRIDE: 1,
  OPP_MOVESET_OVERRIDE: [ Moves.DOUBLE_SLAP ],
  MOVESET_OVERRIDE: [ Moves.RAGE_FIST, Moves.SPLASH ],
  STARTING_LEVEL_OVERRIDE: 100,
  STARTING_WAVE_OVERRIDE: 4
} satisfies Partial<InstanceType<typeof DefaultOverrides>>;
```

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test`)
  - [x] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?

PR #5165, opened 2025-01-22
[Bug] Mage Rage Fist ignore confusion self-hits

geeilhan:
Continuation of this [Rage Fist PR](https://github.com/pagefaultgames/pokerogue/pull/5129)
(#3503)

Confusion self inflicted hits does not count to the increased power of rage fist

Added a `confHitCount` to `pokemon.battleData` which only counts self inflicted hits due to confusion. this number gets subtracted from `hitCount` in order to get the correct `hitCount` value for rage fist

## Screenshots/Videos

For reference look at the videos from [this PR](https://github.com/pagefaultgames/pokerogue/pull/5129)

## How to test the changes?

Spam Splash until you get confused then use Rage Fist and pray for a self hit 
```
const overrides = {
  OPP_SPECIES_OVERRIDE: Species.MAGIKARP,
  OPP_LEVEL_OVERRIDE: 1,
  OPP_MOVESET_OVERRIDE: [ Moves.CONFUSE_RAY, Moves.TACKLE ],
  MOVESET_OVERRIDE: [ Moves.RAGE_FIST, Moves.SPLASH ],
  STARTING_LEVEL_OVERRIDE: 100,
  STARTING_WAVE_OVERRIDE: 1,
```

## Checklist


SirzBenjie (commented):
code looks good, but for cases like these I like to see automated tests. I can write, if necessary :)

As in, can we get an automated test specifically for ensuring confusion hits don't increase damage? I think you can mock the confusion check to force the mon to hit itself.

geeilhan:
> code looks good, but for cases like these I like to see automated tests. I can write, if necessary :)
> 
> As in, can we get an automated test specifically for ensuring confusion hits don't increase damage? I think you can mock the confusion check to force the mon to hit itself.

That would be nice. I couldn't write tests for the confusion case because I didn't know how to mock it.

SirzBenjie:
> > code looks good, but for cases like these I like to see automated tests. I can write, if necessary :)
> > As in, can we get an automated test specifically for ensuring confusion hits don't increase damage? I think you can mock the confusion check to force the mon to hit itself.
> 
> That would be nice. I couldn't write tests for the confusion case because I didn't know how to mock it.

Okay, I'll add it 🫡

damocleas:
has conflicts, and above comment still it seems

DayKev:
Fixed in #5655

PR #5169, opened 2025-01-23
[Bug] Fix miniblackhole

Xavion3:
## What are the changes the user will see?
Miniblackhole will actually work as intended.

## Why am I making these changes?
Miniblackhole accidentally used the wrong variable name at one point so was working incorrectly.

## What are the changes from a developer perspective?
Slightly changed mbh code.

## Screenshots/Videos
This is difficult to show because it's a slight change to the randomisation method.

## How to test the changes?
Overrides to give yourself leftovers and a bunch of berries and the opponent miniblackhole, repeatedly start and play runs and it should consistently prefer leftovers to berries.

## Checklist
- [X] **I'm using `beta` as my base branch**
- [X] There is no overlap with another PR?
- [X] The PR is self-contained and cannot be split into smaller PRs?
- [X] Have I provided a clear explanation of the changes?
- [X] Have I tested the changes manually?
- [X] Are all unit tests still passing? (`npm run test`)
  - [ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?

PR #5183, opened 2025-01-25
Revert "[Bug] Fix miniblackhole"

damocleas:
Reverts pagefaultgames/pokerogue#5169

### DIFF
--- a/src/modifier/modifier.ts
+++ b/src/modifier/modifier.ts
@@ -3183,12 +3183,12 @@ export abstract class HeldItemTransferModifier extends PokemonHeldItemModifier {
           break;
         }
       }
-      const randItemIndex = pokemon.randSeedInt(tierItemModifiers.length);
-      const randItem = tierItemModifiers[randItemIndex];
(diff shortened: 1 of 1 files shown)


DayKev:
Was there a bug?

Snailman11:
Is the block of code working as intended?

DayKev (approved):

Blitz425 (approved):

Blitz425:
Hard agree on this revert, a fundamental change affecting _the_ reoccurring big boss shouldn't be a last minute addition when update is soon.

EscaShark:
> Is the block of code working as intended?

In Xav's PR, yes, it is working "as intended". But the problem is that this is what was "intended" a loooong time ago and, in the opinion of the majority of balance team members that have spoken up about it, not a good thing for the game right now. Ergo the existence of the bug is actually better for the game. This is also reflected in the bugfix receiving nothing but negative feedback from players.

Down the line the bugged block of code should probably be removed altogether, unless consensus on this issue were to flip. But as for now I'd rather have ugly code than code that does something neither the playerbase nor the balance team intends to happen.

PR #5200, opened 2025-01-28
[Ability][Move] Last Respects Refactor and Full Implementation

geeilhan:
(#3503 )

Supreme Overlord now works correctly

Supreme Overlord now always calculates how many fainted Pokemon are in the party. Therefore `globalScene.currentBattle.playerFaints` does not need to be used so it does not matter if this counter resets every wave.

Changed documentation for `globalScene.currentBattle.playerFaints`.

## Screenshots/Videos

No visuals but added automated test for Supreme Overlord.

## How to test the changes?

Includes automated tests.
For manual tests use: 
```
const overrides = {
  MOVESET_OVERRIDE: [ Moves.TACKLE, Moves.EXPLOSION ],
  ABILITY_OVERRIDE: Abilities.SUPREME_OVERLORD,
```
and start with 2-3 starter Pokemon. Use Explosion to make Pokemon faint for damage increase.

## Checklist


src/data/ability.ts
```diff
@@ -6319,8 +6319,7 @@ export function initAbilities() {
     new Ability(Abilities.SHARPNESS, 9)
       .attr(MovePowerBoostAbAttr, (user, target, move) => move.hasFlag(MoveFlags.SLICING_MOVE), 1.5),
     new Ability(Abilities.SUPREME_OVERLORD, 9)
-      .attr(VariableMovePowerBoostAbAttr, (user, target, move) => 1 + 0.1 * Math.min(user.isPlayer() ? globalScene.currentBattle.playerFaints : globalScene.currentBattle.enemyFaints, 5))
-      .partial(), // Counter resets every wave instead of on arena reset
+      .attr(VariableMovePowerBoostAbAttr, (user, target, move) => 1 + 0.1 * Math.min(user.isPlayer() ? globalScene.getPlayerParty().filter(pokemon => pokemon.isFainted()).length : globalScene.getEnemyParty().filter(pokemon => pokemon.isFainted()).length, 5)),
```
DayKev, line 6322: The reason it wasn't already done this way (as I found out when I tried to make the same change) is because rather than check how many party members are currently fainted, it should remember the number of fainted pokemon even if you revive all of them via items or moves like revival blessing etc. This change would make it so if you revive all your party members then SO won't have a damage boost anymore.

> This Pokemon's moves have their power multiplied by 1+(X*0.1), where X is the total number of times any Pokemon has fainted on the user's side when this Ability became active, and X cannot be greater than 5.

- https://dex.pokemonshowdown.com/abilities/supremeoverlord

geeilhan: PR (#5201) would solve this issue then. since #5201 changes playerFaints/enemyFaints so it does not reset the count every wave but every new arena stage. I will change this PR to Draft and open it again once #5201 is merged

Xavion3:
This isn't actually a bug, that's how supreme overlord/last respects are actually meant to work. It was just an outstanding decision on how to translate per battle effects into pokerogue that got it labeled (P). It's not about working correctly or incorrectly. Also this shouldn't be separate from #5201 , they're the same code.

This can be worked on, but it's a far bigger topic than just whether it works, because right now they do (sort of) work correctly, we just haven't fully defined what correctly is. It might have to involve devs, balance, and design to hammer out what exactly we want to do, because *every* effect that is based on the idea of lasting a single battle can be potentially impacted here.

geeilhan:
> This isn't actually a bug, that's how supreme overlord/last respects are actually meant to work. It was just an outstanding decision on how to translate per battle effects into pokerogue that got it labeled (P). It's not about working correctly or incorrectly. Also this shouldn't be separate from #5201 , they're the same code.
> 
> This can be worked on, but it's a far bigger topic than just whether it works, because right now they do (sort of) work correctly, we just haven't fully defined what correctly is. It might have to involve devs, balance, and design to hammer out what exactly we want to do, because _every_ effect that is based on the idea of lasting a single battle can be potentially impacted here.

Should I put both PRs in Draft then? I think this will take time for the devs/balance team/design team to iron out then

DayKev:
I would consolidate both PRs into one (which you choose doesn't matter). SO and LR are supposed to work the same, the only difference is SO's cap is 5 and LR's cap is something crazy like 100.

geeilhan:
Combined both PRs.

Xavion3:
If it's going to be tracked longer, probably just stick it in the arena instead of mucking up the battle.

Also remember if you want it to last across multiple waves the save/loading code has to support it so it can actually save the progress and that doesn't work unless you put it in system data or arena data or the like.

geeilhan:
> If it's going to be tracked longer, probably just stick it in the arena instead of mucking up the battle.
> 
> Also remember if you want it to last across multiple waves the save/loading code has to support it so it can actually save the progress and that doesn't work unless you put it in system data or arena data or the like.

Oh yeah didn't think about loading/saving. I will test that out thanks

Xavion3:
If you're going to make it linger between battles, we should probably think about just making it part of the arena data instead of the battle data, since it's not actually tied to battles anymore so it doesn't make much sense to store there. This would also stop the need for the oddities with having to pass them back into the constructor every wave.

geeilhan:
> If you're going to make it linger between battles, we should probably think about just making it part of the arena data instead of the battle data, since it's not actually tied to battles anymore so it doesn't make much sense to store there. This would also stop the need for the oddities with having to pass them back into the constructor every wave.

Only enemyFaints is in battleData and playerFaints is in arena. Since reloading a session would reset e.g. the trainer battle (-> reset enemyFaints count).

SirzBenjie (dismissed):

src/field/arena.ts
```diff
@@ -52,12 +57,13 @@ export class Arena {
 
   public readonly eventTarget: EventTarget = new EventTarget();
 
-  constructor(biome: Biome, bgm: string) {
+  constructor(biome: Biome, bgm: string, playerFaints?: number) {
```
DayKev, line 60:
```suggestion
  constructor(biome: Biome, bgm: string, playerFaints: number = 0) {
```

src/field/arena.ts
```diff
@@ -52,12 +57,13 @@ export class Arena {
...
     this.biomeType = biome;
     this.tags = [];
     this.bgm = bgm;
     this.trainerPool = biomeTrainerPools[biome];
     this.updatePoolsForTimeOfDay();
+    this.playerFaints = playerFaints ?? 0;
```
DayKev, line 66:
```suggestion
    this.playerFaints = playerFaints;
```

src/battle.ts
```diff
@@ -115,7 +117,7 @@ export default class Battle {
 
   private rngCounter: number = 0;
 
-  constructor(gameMode: GameMode, waveIndex: number, battleType: BattleType, trainer?: Trainer, double?: boolean) {
+  constructor(gameMode: GameMode, waveIndex: number, battleType: BattleType, trainer?: Trainer, double?: boolean, enemyFaints?: number) {
```
DayKev, line 120:
```suggestion
  constructor(gameMode: GameMode, waveIndex: number, battleType: BattleType, trainer: Trainer = null, double: boolean = false, enemyFaints: number = 0) {
```

src/battle.ts
```diff
@@ -125,6 +127,7 @@ export default class Battle {
       ? new Array(double ? 2 : 1).fill(null).map(() => this.getLevelForWave())
       : trainer?.getPartyLevels(this.waveIndex);
     this.double = double ?? false;
+    this.enemyFaints = enemyFaints ?? 0;
```
DayKev, line 130:
```suggestion
    this.double = double;
    this.enemyFaints = enemyFaints;
```

SirzBenjie:
> > If you're going to make it linger between battles, we should probably think about just making it part of the arena data instead of the battle data, since it's not actually tied to battles anymore so it doesn't make much sense to store there. This would also stop the need for the oddities with having to pass them back into the constructor every wave.
> 
> Only enemyFaints is in battleData and playerFaints is in arena. Since reloading a session would reset e.g. the trainer battle (-> reset enemyFaints count). But I could also move enemyFaints to arena if that would be better.

For now, we should not let enemy wild pokemon benefit from this:
image

SirzBenjie (approved):

SirzBenjie:
> > If you're going to make it linger between battles, we should probably think about just making it part of the arena data instead of the battle data, since it's not actually tied to battles anymore so it doesn't make much sense to store there. This would also stop the need for the oddities with having to pass them back into the constructor every wave.
> 
> Only enemyFaints is in battleData and playerFaints is in arena. Since reloading a session would reset e.g. the trainer battle (-> reset enemyFaints count). But I could also move enemyFaints to arena if that would be better.

The chosen implementation is intuitive in my opinion. Keeps SoC in check. It may be a bit awkward wrt the constructor, but I think th

> I would consolidate both PRs into one (which you choose doesn't matter). SO and LR are supposed to work the same, the only difference is SO's cap is 5 and LR's cap is something crazy like 100.

Actually, SO and LR are slightly different in that SO is a one-off boost that occurs on summon.

SirzBenjie (approved):

src/test/moves/last_respects.test.ts
```diff
@@ -0,0 +1,219 @@
...
+    await game.toNextTurn();
+
+    game.move.select(Moves.LAST_RESPECTS);
+    await game.setTurnOrder([ BattlerIndex.PLAYER, BattlerIndex.ENEMY ]);
+    await game.phaseInterceptor.to(MoveEffectPhase);
+
+    expect(move.calculateBattlePower).toHaveReturnedWith(basePower + (2 * 50));
+  });
+
+  it("should have 200 power if an ally fainted twice and another one once", async () => {
```
DayKev, line 70: Should probably have the same test for Supreme Overlord.

DayKev (approved):

src/test/moves/last_respects.test.ts
geeilhan: Added tests

Xavion3:
> Given that, it makes the most sense to have it tracked per-battle and not need it to be passed to the next one.
Just implement that exception? Like, we just track them both in the same place in the arena, but we can hardcode in a nerf to wild pokemon for now if we want by just not giving wild pokemon the boosts.

No need to implement it not track information it should be tracking just to nerf wild pokemon.

src/battle-scene.ts
```diff
@@ -1361,6 +1362,9 @@ export default class BattleScene extends SceneBase {
         this.arena.updatePoolsForTimeOfDay();
       }
       if (resetArenaState) {
+        this.resetEnemyFaintCount();
+        this.arena.resetPlayerFaintCount();
```
DayKev, line 1366: Oh yeah, this should be called from within `Arena.resetArenaEffects()` instead of out here.

src/battle-scene.ts
```diff
@@ -1344,6 +1344,7 @@ export default class BattleScene extends SceneBase {
     if (newBattleType === BattleType.MYSTERY_ENCOUNTER) {
       // Will generate the actual Mystery Encounter during NextEncounterPhase, to ensure it uses proper biome
       this.currentBattle.mysteryEncounterType = mysteryEncounterType;
+      this.arena.resetPlayerFaintCount();
```
DayKev, line 1347: This one isn't necessary because if it's a Mystery Encounter then `resetArenaState` a few lines below this is set to `true` and that `if (resetArenaState) {` block is executed.

DayKev: Actually wait, maybe that's not true, hm.

DayKev: Okay after reading the code again my first comment should be correct, since L1340 is `this.currentBattle = new Battle(this.gameMode, newWaveIndex, newBattleType, newTrainer, newDouble);` which sets `this.currentBattle.battleType` to `BattleType.MYSTERY_ENCOUNTER` by the time `resetArenaState` is set on L1358. So this extra `resetPlayaerFaintCount()` shouldn't be necessary.

src/battle-scene.ts
geeilhan: I think I can just remove resetEnemyFaintCount() since every time a new Battle is created the enemyFaints counter gets set to 0 anyways. 

Tested it and enemyFaints counts correctly and resets every new Battle.

Xavion3:
Yeah, the more I think on this the more I think we really should just be consistent on how we track this data. If we want to specifically nerf wild battles, we can just code in an exception for those, although making the behaviour inconsistent should probably be avoided anyway.

SirzBenjie:
> Yeah, the more I think on this the more I think we really should just be consistent on how we track this data. If we want to specifically nerf wild battles, we can just code in an exception for those, although making the behaviour inconsistent should probably be avoided anyway.

I'm not sure how the behavior is inconsistent. Could you elaborate on that? 
It's just tracked at the granularity needed for the side that's using it.
For enemies, the fainted counter is tracked per wave. For the player, it's per arena, resetting on trainer battles (anytime the Pokemon go back in their pokeballs).

Wild pokemon still get the boost when an ally of theirs faints in a double battle. They just don't get the boost from earlier waves.

src/data/move.ts
```diff
@@ -10916,8 +10916,7 @@ export function initMoves() {
       .attr(ConfuseAttr)
       .recklessMove(),
     new AttackMove(Moves.LAST_RESPECTS, Type.GHOST, MoveCategory.PHYSICAL, 50, 100, 10, -1, 0, 9)
-      .partial() // Counter resets every wave instead of on arena reset
-      .attr(MovePowerMultiplierAttr, (user, target, move) => 1 + Math.min(user.isPlayer() ? globalScene.currentBattle.playerFaints : globalScene.currentBattle.enemyFaints, 100))
+      .attr(MovePowerMultiplierAttr, (user, target, move) => 1 + Math.min(user.isPlayer() ? globalScene.arena.playerFaints : globalScene.currentBattle.enemyFaints, 100))
```
SirzBenjie, line 10919: Not for this PR, but the code here will to use/create an onSummon attribute that gives its boost once based on a snapshot of the number of ally mons that have fainted to get SO to be fully implemented.

SirzBenjie (approved):

DayKev (approved):

PR #5219, opened 2025-01-30
[Bug][Move] Shed Tail now rounds incurred damage up, not down

EscaShark:

Using Shed Tail now rounds the incurred damage up instead of down, as per https://bulbapedia.bulbagarden.net/wiki/Shed_Tail_(move)

Minor bugfix (accidentally discovered this while testing for another Shed Tail-related bug and it felt easy enough to fix).

Added an extra argument to AddSubstituteAttr with associated private variable. Setting roundUp to true will cause the incurred damage to be rounded up. Currently, Shed Tail has roundUp set to true and Substitute has roundUp set to false.

Both apply and getCondition were changed to use math.ceil instead of math.floor whenever roundUp is set to true.

## Screenshots/Videos
Before:

https://github.com/user-attachments/assets/1e85c88b-cdfb-49e1-b0a6-90300f66d97b

After:

https://github.com/user-attachments/assets/6b57bc66-c037-404c-9cd2-cd615c6d66bf

Note that in both videos, Substitute is used first to confirm that Substitute damage is still rounded down.

## How to test the changes?

Bring any mon with an odd HP stat and use Shed Tail.

## Checklist


EscaShark:
Looks like I might have to rewrite some tests for this first (since they were seemingly built with the incorrect assumption of rounding down in mind)

src/data/move.ts
```diff
@@ -1691,11 +1691,14 @@ export class HalfSacrificialAttr extends MoveEffectAttr {
 export class AddSubstituteAttr extends MoveEffectAttr {
   /** The ratio of the user's max HP that is required to apply this effect */
   private hpCost: number;
+  /** Whether the damage taken should be rounded up (Shed Tail rounds up) */
+  private roundUp: boolean;
 
-  constructor(hpCost: number = 0.25) {
+  constructor(hpCost: number = 0.25, roundUp: boolean) {
```
DayKev, line 1697:
```suggestion
  constructor(hpCost: number, roundUp: boolean) {
```

bavalpey:
It's curious how this happened. All percentage based effects round up -- spikes, super fang, etc.

Worth looking at whether those have the proper behavior.

SirzBenjie:
It's curious why this was written this way in the first place. All % based damage effects round down — spikes, super fang, belly drum, substitute, etc..

Worth investigating whether this bug is unique to shed tail or whether those are implemented properly.

DayKev:
> It's curious why this was written this way in the first place. All % based damage effects round down — spikes, super fang, belly drum, substitute, etc..
> 
> Worth investigating whether this bug is unique to shed tail or whether those are implemented properly.

There's a `doDmgValue()` utility function that handles rounding all damage values properly. It was added after inconsistencies in damage rounding were discovered. At this point everything should be using it.

EscaShark:
> It's curious why this was written this way in the first place. All % based damage effects round down — spikes, super fang, belly drum, substitute, etc..
> 
> Worth investigating whether this bug is unique to shed tail or whether those are implemented properly.

Shed Tail uniquely rounds up, this is a common trivia tidbit among mons mechanic nerds. Reason is unknown but ig Game Freak drew the line at a max HP Orthworm being able to Shed Tail more than once without healing?

SirzBenjie:
> Shed Tail uniquely rounds up, this is a common trivia tidbit among mons mechanic nerds. Reason is unknown but ig Game Freak drew the line at a max HP Orthworm being able to Shed Tail more than once without healing?

Oh I completely misunderstood the issue. I had thought that she'd tail was incorrectly rounding up. Didn't realize she'd tail is unique. How odd.

Wlowscha (dismissed):

SirzBenjie:
This fail is not related to this PR

Madmadness65 (approved):

DayKev (approved):
Minor bugfixes 🎉

PR #5083, opened 2025-01-04
[UI/UX] Internal pokedex

Wlowscha:
## What are the changes the user will see?
A new "Pokedex" entry is available in the game menu, accessible from the title page or during a run. Selecting the option opens the Pokedex interface, which has two parts:

1) A display of all mons in a box-like container, identical to the starter-select menu, but displaying evolutions as well as starters. In addition to the filter bars from the starter-select menu, new text forms allow to search mons with a given name, move, or ability. Bringing the cursor on a mon displays the name, number and sprite of the base form. By default, icons do not include shiny overlays, candy icons and abilities, but these can be toggled on with C.

2) Clicking on a mon will open a second page, where a second menu allows details such as level moves, egg moves, biomes etc to be displayed.
For egg moves and abilities, the ones that have not yet been unlocked are shown in darker text, but still visible to the user.
The Evolutions entry shows a list of prevolution, evolutions, and forms available during a run such as megas and g-max forms. Selecting one brings the user to another page for that specific evolution stage or form.
What options exactly should be available in what shape is still to be finalized.

It should also be possible for users to buy candy upgrades directly from the dex page, without needing to interrupt a run and go through the starter select screen. This is not implemented yet but the logic is already there, it is a matter of picking a button.

## Why am I making these changes?
An internal dex is a key functionality. Having to leave the game in order to look up information can be bothersome; in the case of mobile users, it can lead to the game shutting down and reloading altogether. Moreover, the dex can implement some functionalities, such as buying candies, which are currently unavailable during a run and require to go back to the title screen.
Most importantly, a functional internal dex reduces the need to rely on external sources, and ensures that data is always fully up to date.

## What are the changes from a developer perspective?
Balance:
Helper functions have been added in biomes.ts, pokemon-evolutions.ts and tms.ts, to collect information that the pokedex needs to access in a compact form.

Ui:
The ui/abstact-option-select-ui-handler file has been modified to change the color of individual options, as well as to allow the insertion of lines which are skipped by the cursor (such as headers or descriptions). The new version replicates the old behavior in all cases, but allows for more customization of the menus if required.

The ui/dropdown.ts file has been extended to include a scrollbar when there are too many options (needed for the biome filter).

A new ui/pokedex-ui-scan.ts has been added to manage the auto-complete interface that selects names, moves or abilities.

The ui/pokedex-ui-handler.ts and ui/pokedex-page-ui-handler.ts files have been added, replicating and changes functionality from starter-select-ui-handler.ts.

[...]

PR #5223, opened 2025-01-30
[Misc] New data structures for pokedex

Wlowscha:
No visible changes.

In the internal pokedex #5083 it is necessary to access the list of biomes for a given species, and the list of TMs for a given species. This information is not directly available since the respective data structures work in reverse (biome pools which assign species to biomes, and arrays which assign species to tm moves). This PR introduces two new structures which are created as the game starts, so that it is not necessary to repeat the calculation every time a new species is selected in the dex.

No code changes but addition of two new data structures.
catchableSpecies in biomes.ts is an interface which assigns to each species id an array of {biome, time of day, tier).
speciesTmMoves is an integer which assigns to each species id the array of its tm moves.

## Checklist


src/data/balance/tms.ts
```diff
@@ -68433,6 +68433,31 @@ export const tmSpecies: TmSpecies = {
...
+      }
+      flipped[speciesKey].push(moveKey);
+    }
+  }
+  return flipped;
+}
+
+export const speciesTmMoves: SpeciesTmMoves = flipTmSpecies(tmSpecies);
+
+
```
DayKev, line 68460:
```suggestion
```

DayKev (approved):

src/data/balance/tms.ts
```diff
@@ -68433,6 +68433,31 @@ export const tmSpecies: TmSpecies = {
   ],
 };
 
+interface SpeciesTmMoves {
+  [key: integer]: Moves[]
+}
+
+function flipTmSpecies(tmSpecies: TmSpecies): SpeciesTmMoves {
```
Fontbane, line 68440: A more precise term would be "transpose". Also, is the parameter necessary? The function would be able to call tmSpecies without it.

src/data/balance/tms.ts
```diff
@@ -68433,6 +68433,31 @@ export const tmSpecies: TmSpecies = {
...
+  const flipped: SpeciesTmMoves = {};
+
+  for (const move in tmSpecies) {
+    const moveKey = Number(move);
+    const speciesList = tmSpecies[move];
+
+    for (const species of speciesList) {
+      const speciesKey = Number(species);
```
Fontbane, line 68448: Does this... work? Wouldn't some of these be arrays for forms? What happens for, let's say, Upper Hand on Lycanroc?

Wlowscha: Simple... it breaks! Thanks for pointing it out, I'll fix this :)

Wlowscha: It's fixed now.

src/data/balance/tms.ts
Wlowscha: It's changed :)

src/data/balance/passives.ts
```diff
@@ -1,7 +1,15 @@
 import { Abilities } from "#app/enums/abilities";
 import { Species } from "#app/enums/species";
 
-export const starterPassiveAbilities = {
+export interface PassiveAbilities {
+  [key: integer]: Abilities
```
DayKev, line 5:
```suggestion
  [key: number]: Abilities
```
Btw you should use `number` instead of `integer` (`integer` is just an alias for `number` and doesn't actually provide any additional type safety annoyingly).

Madmadness65 (approved):

DayKev (approved):
Btw I meant you should change all instances of `integer` with `number` lol.

PR #5225, opened 2025-01-31
[Ability] [Move] Implement Magic Bounce and Magic Coat

SirzBenjie:

Magic Bounce and magic coat are both fully implemented, no longer marked as (N), and all interactions (other than extremely specific niche scenarios) should work as expected. 

Magic bounce is on several pokemon, yet does not do anything at all. Mega Absol and Mega Sableye end up not having any ability.

Okay, there's a decent amount here, as there's quite a lot of components that need to be considered to get magic bounce working properly.

### `src/phases/move-effect-phase.ts`

Lots of changes in this file that are needed for magic bounce. I won't go through every line here, but rather the overall.

Added a new field `reflected` that is true if the initiating move was reflected by the effects of magic bounce, to prevent infinite magic bounce loops, and modified the constructor to take this new parameter and set the flag accordingly.

Extracted out some logic that is also needed for magic bounce from the `hitCheck` method to their own functions, to adhere to DRY principles.

Modified `start` to properly handle magic bounce in all of the cases. See the newly added unit tests for an exhaustive list of the behaviors.
Notably, removed the `Prevent ENEMY_SIDE targeted moves from occurring twice in double battles` logic from the per target loop to occur prior to looping over the targets, as we need to know which target, if any, will bounce the enemy side move.
Moved some other variable definitions around (such as `isCommanding) without modifying them as they are needed earlier than their original place.

Added a `queuedPhases` array that will store new phases that should be added (in order) after the next `Move End` phase, in order to insert new move phases initiated by magic bounce. This `queudPhases` structure should probably be used for other moves that need to add phases after the move is ended, though that is outside the scope of this PR.

### `src/phases/move-phase.ts`
Added a new field `reflected` that is true if the initiating move was reflected by the effects of magic bounce. This is checked before initiating dancer, as Feather Dance does not trigger dancer if magic bounce was bounced.

Modified the `showMoveText` method to check the `reflected` tag when determining the message in order to display the "bounced the move back" text instead of the "used the move" text.
Modified the constructor call to `MoveEffectPhase` to pass along the `reflected` field.

### `src/battle-scene.ts`

Refactored `unshiftPhase` to mirror javascript's `array.unshift` method, allowing it to take a variable list of phases to unshift.
Refactored `prependToPhase` and `appendToPhase` to also allow taking a `Phase []` instead of just `Phase`.

Necessary in order to insert the move-phases initiated from magic bounce after the `move-end` phase 

### `src/data/ability.ts`

Applied the new `ReflectStatusMoveAbAttr` and removed the `unimplemented` marker on `Magic Bounce`. Added an `edgeCase` marker on it as well, as there are some niche interactions that are currently not handled (and due to issues with moves outside the scope of this PR).

### `src/data/battler-tags.ts`

- Added the `MagicCoatTag` class for the Magic Coat move that lasts until the end of the turn and initiates the `battlerTags:magicCoatOnAdd` message when added.  
- Added a new case for `getBattlerTag` to construct and return this when getting the `Magic_Coat` tag.

### `src/data/move.ts`

Removed `unimplemented` from `Magic Coat`, and have it add the new `MAGIC_COAT` battler tag. Gave it the `failIfLastCondition` condition. Marked it as `edgeCase` for the same reasons as the Magic Bounce ability.

Added a new move flag, `REFLECTABLE` and the `reflectable()` method which adds it when called.  
Modified `getTagTargetBenefitScore` to mark the `Magic_Coat` tag have a benefit score of 3 in line with similar move effects.  

Added the new `REFLECTABLE` tag to each move listed on [bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Magic_Coat_(move)).
Note that Sappy Seed is marked as having the reflectable flag in the game code of scarlet/violet, but not in pokerogue, as per Kev's response:

image

### `src/enums/battler-tag-type.ts`
Added a new enum member for Magic Coat.

### Tests

Added `src/test/abilities/magic_bounce.test.ts` with many tests for magic bounce. 4 of them are marked as TODO, related to edge cases. Despite these TODO tests, these are incredibly niche circumstances that do not warrant marking the move as `partial`.

Likewise added `src/test/moves/magic_coat.test.ts` that repeats most of the checks for magic bounce, along with a few things specific to magic coat as a move.

## Screenshots/Videos

There are way, way too many specific scenarios to give evidence for every one. Every case has an automated test added.
Here are videos showing magic bounce working properly.

### Demonstration of magic bounce properly bouncing growl back to both opponents in a double battle, per target that has magic bounce. Also only bouncing back spikes once:

https://github.com/user-attachments/assets/e8b598fd-366c-49a0-baf3-cb2b1e220e38

### Demonstration of magic bounce bouncing normally, respecting the immunities, and doing evasion / accuracy check against new target

https://github.com/user-attachments/assets/3c56e9a3-1220-467c-b45e-6d0802adc2b0

### Demonstration of magic coat

https://github.com/user-attachments/assets/45bcc9f6-13fc-463e-952b-937b1029d83f

### Demonstration of magic coat bypassing move immunities
https://github.com/user-attachments/assets/f3ddcf03-8e75-4468-b608-46b541e7ca65

## How to test the changes?

`npx vitest run src/test/moves/magic_coat.test.ts src/test/abilities/magic_bounce.test.ts`

Overrides I used for various test videos:
soundproof / magic coat
```ts
const overrides = {
  OPP_ABILITY_OVERRIDE: Abilities.SOUNDPROOF,
  OPP_MOVESET_OVERRIDE: [ Moves.MAGIC_COAT ],
  MOVESET_OVERRIDE: [ Moves.SPORE, Moves.GROWL ]
```

Fly / spikes
```ts
const overrides = {
  OPP_MOVESET_OVERRIDE: [ Moves.FLY ],
  OPP_ABILITY_OVERRIDE: Abilities.MAGIC_BOUNCE,
  OPP_SPECIES_OVERRIDE: Species.REGIELEKI,
  STARTING_LEVEL_OVERRIDE: 100,
  MOVESET_OVERRIDE: [ Moves.SPIKES, Moves.CHARM, Moves.SPLASH ],
  ABILITY_OVERRIDE: Abilities.STALL,
```

## Checklist
- [x] The PR is self-contained and cannot be split into smaller PRs?  (see below)

  - [x] If so, please leave a link to it here: pagefaultgames/pokerogue-locales/pull/108 and pagefaultgames/pokerogue-locales/pull/110 and pagefaultgames/pokerogue-locales/pull/118

## Additional Notes

I didn't want to split this into multiple PRs even though there is a sizeable code change. The logic for magic coat and magic bounce is exactly the same, (other than magic coat adding its own tag). Decoupling these into separate PRs would require one change being merged before the other.

I saw a few bugs not related to magic bounce, but that will be triggered with this magic bounce, that I was tempted to fix, but opted not to in order to keep the PR self contained. One of these is the erroneous behavior for lock on and kin.


src/test/abilities/magic_bounce.test.ts
```diff
@@ -149,6 +149,17 @@ describe("Abilities - Magic Bounce", () => {
     expect(game.scene.arena.getTagOnSide(ArenaTagType.SPIKES, ArenaTagSide.PLAYER)!["layers"]).toBe(1);
   });
 
+  it("should not bounce spikes when the target is in the semi-invulnerable state", async () => {
```
SirzBenjie, line 152: https://github.com/user-attachments/assets/6a0784d2-7046-4bf0-99e3-1d167f83ac62

Wlowscha (approved):

SirzBenjie:
Needs to be re-reviewed with new fix for:
https://bsky.app/profile/f-buffalo.bsky.social/post/3lhd6oo55zc2a

src/phases/move-phase.ts
```diff
@@ -543,7 +547,7 @@ export class MovePhase extends BattlePhase {
       return;
     }
 
-    globalScene.queueMessage(i18next.t("battle:useMove", {
+    globalScene.queueMessage(i18next.t(this.reflected ? "battle:magicCoatExe" : "battle:useMove", {
```
DayKev, line 550: "Exe"?
I think `magicCoatTriggered` (or "Activated") might be better, though regardless you made a mistake when adding the key to the locales submodule (you added it as `MagicCoatExe` instead of `magicCoatExe`) that would need to be fixed.

SirzBenjie: Oof of course I did that.
MagicCoatExe is the name of the message in dumps in game files. I'm not the best at naming things so I went with that.

DayKev: Alright, in that case it should be changed to `magicCoatTriggered` (or `magicCoatActivated` if you think it fits better).

Madmadness65 (approved):

PR #5235, opened 2025-02-01
[Balance] Removing Pain Split from boss move pool

Wlowscha:
Bosses don't spawn with Pain Split in their movepool anymore.

Pain Split is a move which is directly detrimental to bosses, since the calculation uses the entire hp and the self damage ignores shields. Bosses are already made to be immune to Pain Split when used by the player's mons, but can still hit themselves. Removing Pain Split from the pool follows the same logic by which Explosion and other self-destructing moves are removed.

Moves with the HpSplitAttr are filtered out when looking at potential moves to form a movepool for bosses.

## Checklist


src/field/pokemon.ts
```diff
@@ -2240,6 +2240,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     // Bosses never get self ko moves
```
damocleas, line 2240: Update comment

damocleas (approved):

DayKev (approved):

PR #5250, opened 2025-02-03
[Refactor] Replace all instances of  `integer` with `number`

Wlowscha:
None, this is an internal refactor.

The `integer `type defined in `phaser.d.ts` is identical to `number `and therefore redundant. Having both types causes confusion, especially as "integer" implies to be stricter than "number" when in fact it is not. Removing `integer `makes the code more uniform and removes a source of confusion for new contributors.

Everywhere, `integer` is replaced by `number`. The exceptions are some comments (where it makes sense to still refer to "integer" in some cases) and the `node_modules` folder which is in .gitignore.

## Checklist
- [x] Are all unit tests still passing? (`npm run test`)``


DayKev (approved):
Should be good, GitHub won't display all the files but it shouldn't be possible for this to break anything.

damocleas (approved):

damocleas:
might need to update src/data/balance/tms since your other PR was merged before this

damocleas (approved):

Madmadness65 (approved):

PR #5264, opened 2025-02-07
[Refactor] [Docs] Minor refactor of `move.checkFlags` into `move.doesFlagEffectApply`

SirzBenjie:
Technically, this fixes an unreported bug where moves called by metronome would improperly not ignore abilities if the user had mold breaker.

@Wlowscha and I had a long discussion in discord about the fact that the `checkFlags` method inside moves was confusing.
This class is the ideal place, adhering to the Separation of Concerns design principal, for moves to check whether a flag of theirs can be applied in a given situation (given that the existing structure of the code does not have move instances for moves that are in flight).

However, the name of the method is potentially confusing, and has likely contributed to the method not being used in cases where it probably should be (note, refactoring those uses is not in the scope of this PR). 

The signature of the method was also changed in a proactive effort to keep the function's signature from looking like
```ts
getBaseDamage(source: Pokemon,
  move: Move,
  moveCategory: MoveCategory,
  ignoreAbility: boolean = false,
  ignoreSourceAbility: boolean = false,
  ignoreAllyAbility: boolean = false,
  ignoreSourceAllyAbility: boolean = false,
  isCritical: boolean = false,
  simulated: boolean = true): number
```

which is very unwieldly to invoke in cases where you just want to override the default for a later parameter (e.g., in the above case, `simulated`).

There are likely circumstances where this method would want to be called with additional options to check. For example, in my implementation of Magic Bounce which would like to have its logic for checking the `reflectable` flag reside here, rather than in the body of the `move-effect` phase. Using the parameter destructuring pattern keeps code churn small and improves readability. It also has the added benefit of making the arguments at callsites more explicit w.r.t. the parameter they correspond to.

`moves.checkFlags` has been renamed to `moves.doesFlagEffectApply`. Its signature has similarly been changed to use parameter destructring.
The check for `this.followUp` when determining whether or not to set the `ignoreAbilities` global in `move-phase` has been removed, as the check is now done inside this method. The check logic for `MoveFlags.IGNORE_ABILITIES` has been modified to return `false` if the source of the flag was from a move that was being used as a follow up.

## Screenshots/Videos
N/A
## How to test the changes?
N/A

## Checklist

  - ~~[ ] If so, please leave a link to it here: ~~

### DIFF
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -660,7 +660,6 @@ export default class Move implements Localizable {
           // by another move, such as via metronome.
         }
         return this.hasFlag(MoveFlags.IGNORE_ABILITIES) && !isFollowUp;
-        break;
       case MoveFlags.IGNORE_PROTECT:
         if (user.hasAbilityWithAttr(IgnoreProtectOnContactAbAttr)
           && this.doesFlagEffectApply({ flag: MoveFlags.MAKES_CONTACT, user, target: null })) {
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -1,4 +1,4 @@
-import { ChargeAnim, MoveChargeAnim } from "../battle-anims";
+import { ChargeAnim, initMoveAnim, loadMoveAnimAssets, MoveChargeAnim } from "../battle-anims";
 import {
   CommandedTag,
   EncoreTag,
@@ -618,12 +618,25 @@ export default class Move implements Localizable {
 
   /**
    * Checks if the move flag applies to the pokemon(s) using/receiving the move
+   *
+   * This method will take the `user`'s ability into account when reporting flags, e.g.
+   * calling this method for {@linkcode MoveFlags.MAKES_CONTACT | MAKES_CONTACT}
+   * will return `false` if the user has a {@linkcode Abilities.LONG_REACH} that is not being suppressed.
+   *
+   * To simply check whether the move has a flag, use {@linkcode hasFlag}.
    * @param flag {@linkcode MoveFlags} MoveFlag to check on user and/or target
    * @param user {@linkcode Pokemon} the Pokemon using the move
    * @param target {@linkcode Pokemon} the Pokemon receiving the move
+   * @param isFollowUp (defaults to `false) `true` if the move was used as a follow up
    * @returns boolean
+   * @see {@linkcode hasFlag}
    */
-  checkFlag(flag: MoveFlags, user: Pokemon, target: Pokemon | null): boolean {
+  doesFlagEffectApply({ flag, user, target = null, isFollowUp = false }: {
+    flag: MoveFlags;
+    user: Pokemon;
+    target?: Pokemon | null;
+    isFollowUp?: boolean;
+  }): boolean {
     // special cases below, eg: if the move flag is MAKES_CONTACT, and the user pokemon has an ability that ignores contact (like "Long Reach"), then overrides and move does not make contact
     switch (flag) {
       case MoveFlags.MAKES_CONTACT:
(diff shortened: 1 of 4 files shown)


Wlowscha (dismissed):

Madmadness65 (approved):

Wlowscha (dismissed):

src/data/move.ts
```diff
@@ -698,12 +698,30 @@ export default class Move implements Localizable {
...
+   * **Note:** This method only checks if the move should have effectively have the flag applied to its use.
+   * It does *not* check whether the flag will trigger related effects.
+   * For example using this method to check {@linkcode MoveFlags.WIND_MOVE}
+   * will not consider {@linkcode Abilities.WIND_RIDER | Wind Rider }.
+   *
+   * To simply check whether the move has a flag, use {@linkcode hasFlag}.
    * @param flag {@linkcode MoveFlags} MoveFlag to check on user and/or target
    * @param user {@linkcode Pokemon} the Pokemon using the move
    * @param target {@linkcode Pokemon} the Pokemon receiving the move
+   * @param isFollowUp (defaults to `false) `true` if the move was used as a follow up
```
DayKev, line 715:
```suggestion
   * @param isFollowUp (defaults to `false`) `true` if the move was used as a follow up
```

src/data/move.ts
```diff
@@ -698,12 +698,30 @@ export default class Move implements Localizable {
...
    * @returns boolean
-   */
-  checkFlag(flag: MoveFlags, user: Pokemon, target: Pokemon | null): boolean {
+   * @see {@linkcode hasFlag}
+   */
+  doesFlagEffectApply({ flag, user, target = null, isFollowUp = false }: {
+    flag: MoveFlags;
+    user: Pokemon;
+    target?: Pokemon | null;
```
DayKev, line 722:
```suggestion
    target?: Pokemon;
```

src/data/move.ts
```diff
@@ -718,11 +736,13 @@ export default class Move implements Localizable {
...
           }
+          // Sunsteel strike, Moongeist beam, and photon geyser will not ignore abilities if invoked
+          // by another move, such as via metronome.
         }
-        break;
+        return this.hasFlag(MoveFlags.IGNORE_ABILITIES) && !isFollowUp;
       case MoveFlags.IGNORE_PROTECT:
         if (user.hasAbilityWithAttr(IgnoreProtectOnContactAbAttr)
-          && this.checkFlag(MoveFlags.MAKES_CONTACT, user, null)) {
+          && this.doesFlagEffectApply({ flag: MoveFlags.MAKES_CONTACT, user, target: null })) {
```
DayKev, line 745:
```suggestion
          && this.doesFlagEffectApply({ flag: MoveFlags.MAKES_CONTACT, user })) {
```

src/data/moves/move.ts
```diff
@@ -639,11 +657,13 @@ export default class Move implements Localizable {
...
           }
+          // Sunsteel strike, Moongeist beam, and photon geyser will not ignore abilities if invoked
+          // by another move, such as via metronome.
         }
-        break;
+        return this.hasFlag(MoveFlags.IGNORE_ABILITIES) && !isFollowUp;
       case MoveFlags.IGNORE_PROTECT:
         if (user.hasAbilityWithAttr(IgnoreProtectOnContactAbAttr)
-          && this.checkFlag(MoveFlags.MAKES_CONTACT, user, null)) {
+          && this.doesFlagEffectApply({ flag: MoveFlags.MAKES_CONTACT, user, target: null })) {
```
DayKev, line 666:
```suggestion
          && this.doesFlagEffectApply({ flag: MoveFlags.MAKES_CONTACT, user })) {
```

src/data/moves/move.ts
```diff
@@ -619,12 +619,30 @@ export default class Move implements Localizable {
...
    * @param flag {@linkcode MoveFlags} MoveFlag to check on user and/or target
    * @param user {@linkcode Pokemon} the Pokemon using the move
    * @param target {@linkcode Pokemon} the Pokemon receiving the move
+   * @param isFollowUp (defaults to `false) `true` if the move was used as a follow up
    * @returns boolean
-   */
-  checkFlag(flag: MoveFlags, user: Pokemon, target: Pokemon | null): boolean {
+   * @see {@linkcode hasFlag}
+   */
+  doesFlagEffectApply({ flag, user, target = null, isFollowUp = false }: {
```
DayKev, line 640:
```suggestion
  doesFlagEffectApply({ flag, user, target, isFollowUp = false }: {
```

src/data/moves/move.ts
```diff
@@ -619,12 +619,30 @@ export default class Move implements Localizable {
...
+    flag: MoveFlags;
+    user: Pokemon;
+    target?: Pokemon | null;
```
DayKev, line 643:
```suggestion
    target?: Pokemon;
```
Also change `Move#hitsSubstitute()`'s `target` param to be optional instead of `| null` (that's the reason this param is `| null`, but it's unnecessary).

DayKev (approved):

emdeann (dismissed):

Wlowscha (approved):

DayKev (approved):

PR #5265, opened 2025-02-07
[Bug][Ability] Fix mold breaker effect lingering if the user's move runs out of PP

SirzBenjie:
Mold breaker will now no longer suppress abilities when its user uses a move with its last pp.

I was asked to look into this odd interaction with mold breaker
image

In `move-phase`'s `end` method, `MoveEndPhase` will *always* be unshifted, instead of only if the move was not called via a follow up and an erroneous check to the `canMove` call (which would previously return `false` after the move's pp had been deducted, as the `canMove` check also looks at the remaining pp).

Instead, `this.followUp` is now passed as an argument to the `MoveEndPhase` constructor which is called  unconditionally by `end`.
 
`MoveEndPhase` takes a new optional parameter - `wasFollowUp` that is checked before lapsing `BattlerTagLapseType.AFTER_MOVE`.
I don't know the interactions that would cause this lapse to need to not happen when a move is called via a follow up (something with Encore it looks like). This behavior, however, is consistent with what would happen prior to this PR.

## Screenshots/Videos
Note me highlighting the lack (or presence of, in the fixed version) of the MoveEndPhase in the console.
<details><summary>Before the fix</summary>

https://github.com/user-attachments/assets/61e02a15-cb15-4e35-9f5c-0bcaa5afcf6f
</details>

<details><summary>After the fix</summary>

https://github.com/user-attachments/assets/b70211d7-ea2b-46db-8035-77b74ad76213
</details>

## How to test the changes?
Boy oh boy is testing this complicated. We went through a lot of discussion in the `#endless-grinding` channel on discord to figure out exactly how to trigger this bug. (first message link here: https://discord.com/channels/1125469663833370665/1225619031156064317/1337131126992539725)

In short, I was able to reproduce this after a lot of effort with the following set up:

Alternatively, you can have an ally with the double-lure abilities/passive where one of them is ignorable (e.g. watchog with illuminate / no guard).

For the lure test (which the video depicts), you'll need to get yourself into a situation where you have run out of PP at the same time as you KO the last enemy pokemon in a double battle with your mold breaker user, AND make sure your watchog ally has used its move that turn already. (If a move is used after the PP drops, the ignoreAbilities arena flag will be cleared by that move's MoveEndPhase).
The wave that you do this on has to be before there will be a trainer or boss, as these will nullify the lures. I did my test on wave 109.
When you KO the last pokemon, without this bugfix, the `ignoreAbilities` modifier will be true. Since No Guard is an ignorable ability, when checking for abilities that have the lure effect, it will be ignored. When moving on to the next battle, if it isn't a double battle, that means that no guard's lure ability didn't trigger.

Alternatively, you may be able to whip up a scenario with a slightly easier way to test it, though I didn't end up testing it this way:
Get a pokemon to have mold breaker and an easily-confirmed ignorable passive, like earth eater / levitate. Make sure your ally pokemon does not have mold breaker or an intervening ability. Have your ally have a single-target ground type move, and have it go after your mold breaker ally. Have your mold breaker ally use a damaging move on its last PP, and then have its ally use a ground move on your mold breaker pokemon. This must happen before any enemy moves intervene, so it's best if you use a spread move and KO both opponents.

Before the change, mold breaker would have erroneously still been active, and so the ground move would go through levitate / earth eater.
With the change, this should not happen.

## Checklist

  - ~~[ ] If so, please leave a link to it here:~~

### DIFF
--- a/src/phases/move-end-phase.ts
+++ b/src/phases/move-end-phase.ts
@@ -1,13 +1,20 @@
 import { globalScene } from "#app/global-scene";
 import { BattlerTagLapseType } from "#app/data/battler-tags";
 import { PokemonPhase } from "./pokemon-phase";
+import type { BattlerIndex } from "#app/battle";
 
 export class MoveEndPhase extends PokemonPhase {
+  private wasFollowUp: boolean;
+  constructor(battlerIndex: BattlerIndex, wasFollowUp: boolean = false) {
+    super(battlerIndex);
+    this.wasFollowUp = wasFollowUp;
+  }
+
   start() {
     super.start();
 
     const pokemon = this.getPokemon();
-    if (pokemon.isActive(true)) {
+    if (!this.wasFollowUp && pokemon.isActive(true)) {
       pokemon.lapseTags(BattlerTagLapseType.AFTER_MOVE);
     }
 
--- a/src/phases/move-end-phase.ts
+++ b/src/phases/move-end-phase.ts
@@ -14,7 +14,7 @@ export class MoveEndPhase extends PokemonPhase {
     super.start();
 
     const pokemon = this.getPokemon();
-    if (!this.wasFollowUp && pokemon.isActive(true)) {
+    if (!this.wasFollowUp && pokemon?.isActive(true)) {
       pokemon.lapseTags(BattlerTagLapseType.AFTER_MOVE);
     }
 
--- a/src/phases/move-phase.ts
+++ b/src/phases/move-phase.ts
@@ -465,8 +465,7 @@ export class MovePhase extends BattlePhase {
   }
 
   /**
-   * Queues a {@linkcode MoveEndPhase} if the move wasn't a {@linkcode followUp} and {@linkcode canMove()} returns `true`,
-   * then ends the phase.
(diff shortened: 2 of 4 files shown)


Snailman11:
Link to Discord's Bug Report (no guard and illuminate on the same mon dont stack their lure effects)
https://discord.com/channels/1125469663833370665/1296386119604375583

and
(Wonder Guard does not work)
https://discord.com/channels/1125469663833370665/1251966075906556025

SirzBenjie:
Note: The [current failing test](https://github.com/pagefaultgames/pokerogue/blob/6c4dedb73e4cbc516cbbd16bab9149c3760cbc50/src/test/abilities/dancer.test.ts#L31) had some assumptions that are invalidated by this PR. The functionality of the test still works as expected, though the automated methodology to advance the phases along causes issues. Specifically, the test expects dancer-invoked moves to not have `MoveEndPhase` (which they did not before, which is actually another way to trigger this bug).

DayKev (approved):

DayKev (approved):

Madmadness65 (approved):

Wlowscha (approved):

PR #5266, opened 2025-02-07
[Move] [Bug] Fix super-niche edgecase with mega gengar and telekinesis

SirzBenjie:
Most likely none at all. But mega gengar can no longer be baton passed the effect of telekinesis.

Because I was reading code and saw a comment saying this interaction wasn't handled, and knew it would be a one-line fix.
Also because no matter how insignificant the change, I am obsessed with matching cartridge functionality.

Added an automated test to telekinesis' unit tests for this interaction.
Inside of `pokemon`'s `transferSummon` method where it checks if a tag is baton passable, add a check if the tag is telekinesis and the pokemon is mega gengar.

## Screenshots/Videos
This is too niche. Just run the tests :)

## How to test the changes?
```
npx vitest run src/test/moves/telekinesis.test.ts -t "should not be baton passed onto a mega gengar"
```

## Checklist

  - ~~[ ] If so, please leave a link to it here: ~~

### DIFF
--- a/src/data/battler-tags.ts
+++ b/src/data/battler-tags.ts
@@ -2871,7 +2871,7 @@ export class SyrupBombTag extends BattlerTag {
 /**
  * Telekinesis raises the target into the air for three turns and causes all moves used against the target (aside from OHKO moves) to hit the target unless the target is in a semi-invulnerable state from Fly/Dig.
  * The first effect is provided by {@linkcode FloatingTag}, the accuracy-bypass effect is provided by TelekinesisTag
- * The effects of Telekinesis can be baton passed to a teammate. Unlike the mainline games, Telekinesis can be baton-passed to Mega Gengar.
+ * The effects of Telekinesis can be baton passed to a teammate.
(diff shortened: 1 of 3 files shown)


DayKev (approved):

Madmadness65 (approved):

Xavion3:
Doesn't this not handle the case where mega-gengar is the second pokemon in a fusion?

DayKev:
If M-Gengar is the second mon in the fusion then it doesn't matter. It's only disallowed in the first place because GameFreak wanted to be silly, and as the second in the fusion it's no longer the body.

EscaShark:
Does this also apply to Diglett and Dugtrio?

SirzBenjie:
> Does this also apply to Diglett and Dugtrio?

For some strange reason, those mons *are* able to receive the effect via baton pass. Only mega Gengar doesn't. I suspect this has to do with internal nuance in the logic in Pokemon. But with this PR, it matches base.

damocleas (approved):

PR #5267, opened 2025-02-07
[Refactor] Align ability display with mainline

### LINKED ISSUE #5063: [Bug] Imposter Flyout Shows Gained Ability Instead of Imposter
### Describe the bug

The Flyout when Imposter activates does not show Imposter, instead, it shows the Ability of the Pokemon that Imposter transformed into.

### Reproduction

Have Imposter activate against a Wild Pokemon (Start a run with Imposter Ditto)

### Expected behavior

It should show Imposter and not the Ability of the Pokemon it transforms into.

https://github.com/user-attachments/assets/d9dde229-17cf-4741-a5a7-f5aa04d4d50d

### Screenshots / Videos

https://github.com/user-attachments/assets/edc8f675-4f0d-4969-8b04-93cdd9d01a0c

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

emdeann:
- Abilities correctly display before their effects happen, rather than after
- The show/hide animations of the ability bar now block phase changes, so it no longer gets desynced with the battle
- When multiple abilities go off in succession, they now display properly instead of overlapping or other weirdness
- Generally, it is more clear when an ability is activating and what its immediate effect is

There may still be desync between effects and ability flyouts for abilities which make visual changes outside of phases. For example, the arena will turn yellow before the flyout for Electric Terrain.  In an ideal world the ability effect execution would be put in a phase of its own to avoid this, but due to the huge amount of abilities being applied without a phase change (i.e. passing a BooleanHolder and returning its value immediately after) this isn't feasible. A reasonable workaround is to require abilities which display visuals to do so in a phase rather than directly, so they can be synced properly.

To make ability display as a whole stop feeling janky

`apply` (and friends) of `AbAttr` is now split into two functions: `canApply`, which handles the logic of whether an ability can apply or not in the current situation, and `apply`, which actually applies the ability's effects. This is necessary to allow "looking into the future" to see whether an ability will succeed.

`applyAbAttrsInternal` now takes an additional function parameter `successFunc`, which is treated essentially identically to `applyFunc`. The various forms of `apply...AbAttrs` are edited to match this.

`Arena` now has `canSet(Weather/Terrain)` in addition to `trySet(Weather/Terrain)` to allow checking those conditions without actually changing the arena.

`Phase.start()` and `MessagePhase` no longer do anything ability bar related. It's all handled by the ability-related phases now

_`ShowAbilityPhase` now stores the pokemon and ability names as strings rather than storing the object, because abilities which directly modify properties may change the pokemon object before it's shown (i.e. imposter would show the copied ability from the opponent, not "Imposter")._

^ This should be fixed anyway after promise removal. However, the bar doesn’t _need_ the entire pokémon object, so it may be better to just keep it storing info as strings to avoid issues down the road.

`HideAbilityPhase` now handles hiding the ability. This is done as a separate phase as opposed to just a timer on the tween in `AbilityBar` because in mainline, the ability shows -> the ability's effects happen -> the ability hides. That is how it's implemented here; if `apply` adds a phase (stat change, etc.) the bar won't be hidden until that phase finishes. Worth noting that `HideAbilityPhase` is extremely similar to `ShowAbilityPhase` to the point where it might be worth making them one class to avoid duplication of code.

`AbilityBar` is modified to flyout at either the left or the right, depending on whether the player or enemy's ability is activating.
Open to suggestions on the positioning of the bar on the right if anyone cares, I stay away from anything ui related for good reason

Fixes #5063 

### Old ability bar:

<details><summary>Video 1</summary>
<p>

https://github.com/user-attachments/assets/44f98d2a-ee2c-4f68-bfed-6d70df1c4163

</p>
</details> 
<details><summary>Video 2</summary>
<p>

https://github.com/user-attachments/assets/92822eda-2eec-4e91-9ef1-6bf24e622bd4

</p>
</details> 

### New:

<details><summary>PostSummon</summary>
<p>

https://github.com/user-attachments/assets/2c900086-14a7-4086-8af3-471ae1f7e103

</p>
</details> 
<details><summary>PreDefend (Bulletproof)</summary>
<p>

https://github.com/user-attachments/assets/80995143-e86a-43ed-9a21-4f24dd9a8c6b

</p>
</details> 
<details><summary>Trace</summary>
<p>

https://github.com/user-attachments/assets/ca596a27-366b-4330-944e-cd90a5077080

</p>
</details> 
<details><summary>Transform</summary>
<p>

https://github.com/user-attachments/assets/da37e08f-326c-4937-bd75-f49fe7c980ed

</p>
</details>

## How to test the changes?
Play the game with any ability which causes an ability bar flyout

## Checklist


SirzBenjie:
There are so many bug reports related to these flyouts.
Let me know if there's anything I can do to help with this PR.

emdeann:
@SirzBenjie most helpful thing is get #5283 in asap 🙏 so i can work on conflicts, that's the main thing I'm waiting on. I'll probably also ask for your expert opinion on automated tests, since I've only written them for move and ability effects. This for sure needs them but I'm not sure where to go on that

src/phases/show-ability-phase.ts
```diff
@@ -1,14 +1,20 @@
...
+  private pokemonName: string;
+  private abilityName: string;
 
   constructor(battlerIndex: BattlerIndex, passive: boolean = false) {
     super(battlerIndex);
 
     this.passive = passive;
+    // Set these now as the pokemon object may change before the queued phase is run
+    this.pokemonName = getPokemonNameWithAffix(this.getPokemon());
+    this.abilityName = passive ? this.getPokemon().getPassiveAbility().name : this.getPokemon().getAbility().name;
```
DayKev, line 17:
```suggestion
    this.abilityName = (passive ? this.getPokemon().getPassiveAbility() : this.getPokemon().getAbility()).name;
```
Could do this, not sure if it's better. 🤔

src/battle-scene.ts
```diff
@@ -3407,6 +3420,52 @@ export default class BattleScene extends SceneBase {
...
+
+    if (cancelled.value) {
+      return false;
+    }
+
+    const matchingModifier = this.findModifier(
+      (m) =>
+        m instanceof PokemonHeldItemModifier &&
+        m.matchType(mod) &&
+        m.pokemonId === target.id,
```
DayKev, line 3444:
```suggestion
        m instanceof PokemonHeldItemModifier
        && m.matchType(mod)
        && m.pokemonId === target.id,
```

src/battle-scene.ts
```diff
@@ -3595,6 +3654,11 @@ export default class BattleScene extends SceneBase {
     });
   }
 
+  canRemoveModifier(modifier: PersistentModifier, enemy: boolean = false): boolean {
```
DayKev, line 3657: This doesn't really check if a modifier can be removed, it just checks if it exists. It should be renamed to something like `hasModifier`.

src/phase.ts
```diff
@@ -2,9 +2,7 @@ import { globalScene } from "#app/global-scene";
 
 export class Phase {
   start() {
-    if (globalScene.abilityBar.shown) {
-      globalScene.abilityBar.resetAutoHideTimer();
-    }
+
```
DayKev, line 5:
```suggestion
```

src/ui/ability-bar.ts
```diff
@@ -33,69 +26,50 @@ export default class AbilityBar extends Phaser.GameObjects.Container {
...
     this.setVisible(false);
-    this.shown = false;
   }
 
-  showAbility(pokemon: Pokemon, passive: boolean = false): void {
-    this.abilityBarText.setText(`${i18next.t("fightUiHandler:abilityFlyInText", { pokemonName: getPokemonNameWithAffix(pokemon), passive: passive ? i18next.t("fightUiHandler:passive") : "", abilityName: !passive ?  pokemon.getAbility().name : pokemon.getPassiveAbility().name })}`);
-
-    if (this.shown) {
-      return;
+  startTween(config: any, text?: string): Promise<void> {
```
DayKev, line 31:
```suggestion
  public async startTween(config: any, text?: string): Promise<void> {
```

src/ui/ability-bar.ts
```diff
@@ -33,69 +26,50 @@ export default class AbilityBar extends Phaser.GameObjects.Container {
...
+            config.onComplete();
+          }
+          resolve();
+        }
+      });
+    });
+  }
 
-    globalScene.fieldUI.bringToTop(this);
+  showAbility(pokemonName: string, abilityName: string, passive: boolean = false): Promise<void> {
```
DayKev, line 49:
```suggestion
  public async showAbility(pokemonName: string, abilityName: string, passive: boolean = false): Promise<void> {
```

src/ui/ability-bar.ts
```diff
@@ -33,69 +26,50 @@ export default class AbilityBar extends Phaser.GameObjects.Container {
...
-    if (this.autoHideTimer) {
-      clearInterval(this.autoHideTimer);
-    }
-
-    if (this.tween) {
-      this.tween.stop();
-    }
-
-    this.tween = globalScene.tweens.add({
+  hide(): Promise<void> {
```
DayKev, line 64:
```suggestion
  public async hide(): Promise<void> {
```

DayKev (commented):
Didn't look at `ability.ts` yet

src/phases/show-ability-phase.ts
emdeann: Little smoother, I like it

SirzBenjie:
> @SirzBenjie most helpful thing is get #5283 in asap 🙏 so i can work on conflicts, that's the main thing I'm waiting on. I'll probably also ask for your expert opinion on automated tests, since I've only written them for move and ability effects. This for sure needs them but I'm not sure where to go on that

SirzBenjie:
Whoops, I fat fingered the close button.

Anyway, this PR is looking very good. I think that tests for UI flyouts can't be in scope here. The UI testing infra is pretty hacky and it needs a rewrite imo.

emdeann:
Yeah I agree, like I mentioned on dc earlier it's mostly about the ability effect tests in the suite you're making. The actual flyout is very simple and is not where I anticipate many issues

SirzBenjie (dismissed):

damocleas (approved):

Madmadness65 (approved):

damocleas (approved):

damocleas (approved):

Madmadness65 (approved):

DayKev (approved):

Snailman11:
Fixes (neutralizing gas no longer gives a popup)
https://discord.com/channels/1125469663833370665/1348832257099563069
Where Postsummon abilities that set arena conditions (Drought, n gas, E surge don't display their flyouts.)

PR #5282, opened 2025-02-08
[UI/UX] Pokedex updates batch

Wlowscha:
This PR includes various additions and bug fixes for the internal dex. In particular:
- In the main page, the F button is now used to display form icons whenever possible.
- The dexForDevs option now displays full unlocks in the main page as expected.
- Passive abilities and abilities of alternate forms were not filtered properly, this is now fixed.
- Pikachu was crashing the game due to not having a passive, this is now fixed.
- Candy upgrades can be bought even when looking at evolutions and forms.
- Species that were previously treated as evolved starters now properly display egg moves
- Selecting the pokedex option from the starter select screen now does not play the error sound
- If a mon has no TM moves, selecting the option in the dex page shows an error message.
- The menu in the pokedex page is positioned properly for localized text
- Removed a spurious line that cleared phases when exiting the dex, causing a variety of bugs in battle
- Dark icons when filtering egg moves and passives which are not unlocked

## Screenshots/Videos
<details><summary>Showing forms in the main page</summary>
<p>

https://github.com/user-attachments/assets/01dcb54c-53f6-4cc6-97c5-851d1f26f471

</p>
</details> 

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [ ] The PR is self-contained and cannot be split into smaller PRs?
- [ ] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test`)
- [x] Have I provided screenshots/videos of the changes (if applicable)?
  - [x] Have I made sure that any UI change works for both UI themes (default and legacy)?

Are there any localization additions or changes? If so:
- [x] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?
  - [x] If so, please leave a link to it here: https://github.com/pagefaultgames/pokerogue-locales/pull/117
- [x] Has the translation team been contacted for proofreading/translation?

PR #121, opened 2024-04-14
Implement sucker punch

Xavion3:
Now fails if the target has already acted or has selected a status move

I've tested this a bit and it seems to work, I also confirmed you can sucker punch a sleeping target that has selected an attacking move so the move doesn't have to be possible as intended.

PR #5294, opened 2025-02-10
[Refactor] Add support for showing separate species form names

Fontbane:

- Adds support for regional forms to be displayed as being so
- For example, certain contexts will say "Paldean Tauros" as opposed to just "Tauros"
- Also does this for Eternal Flower Floette and Bloodmoon Ursaluna

- Might be useful for #5282 
- Fixes issues with form names in the dex and condenses some reused code into a useful function
- Gives the player more information when a mon hatches

- Adds `getExpandedSpeciesName` and `getFormNameToDisplay` functions to the `PokemonSpecies` class
- `getExpandedSpeciesName` returns, for example, "Alolan Ninetales" or "Bloodmoon Ursaluna"
- `getFormNameToDisplay` gets form names even for mons that aren't considered forms in Rogue and are instead separate species. This includes regional variants, Eternal Flower Floette, and Bloodmoon Ursaluna. It takes 2 optional parameters, the formIndex and a boolean for whether to append the species name at the end.
- Removes `getFormString` function in `pokedex-page-ui-handler`

## Screenshots/Videos

<details>
<summary>Hatching Bloodmoon Ursaluna</summary>

Screenshot_1058

image

</details>

<details>
<summary> Hatching Eternal Flower Floette </summary>

image
image

</details>

<details>
<summary>Hatching Alolan Geodude</summary>

image
image

</details>

<details>
<summary>Various Dex views</summary>

image
image
image

</details>

<details>
<summary>Various starter select views</summary>

image
image
image

</details>

Paldean Tauros shows its breed as its form instead:
image

Zen Mode G-Darm is simply shown as "Zen Mode", while normal mode is called "Paldean Form". These are without locale changes.
image
image

Evolution screen (no locale changes, would say "Alolan Ninetales")
image

## Checklist

- [X] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?
  - [X] If so, please leave a link to it here: [#121](https://github.com/pagefaultgames/pokerogue-locales/pull/121) MERGED
- [X] Has the translation team been contacted for proofreading/translation?


src/data/pokemon-species.ts
```diff
@@ -690,6 +690,62 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
...
+    }
+  }
+
+  /**
+ * Find the form name for species with just one form (regional variants, Floette, Ursaluna)
+ * @param species the species to check
+ * @returns the pokemon-form locale key for the single form name ("Alolan Form", "Eternal Flower" etc)
+ */
+  getFormNameToDisplay(formIndex: number = 0, key?: string, append: boolean = false): string {
+    const formKey = key ?? (this.forms?.[formIndex!]?.formKey);
```
Wlowscha, line 719: Are there any cases where you would want to pass a different formKey from the one associated to the formIndex?

Fontbane: This is really just so either a form index or a key can be passed in. With only form index, there would be cases where you get a form key, then get the form index for that key to pass into the function, which gets the key for that index, which is kind of silly to me. If you have the key already before calling the function, it might as well be used instead of having to look it up again.

Wlowscha: Yeah but that forces you to look up the formKey separately if you already have the index and also want to set append to true. And even if you have the formKey, you are forced to look up the formIndex anyway (like in the examples in the dex below). I think just passing the string would reduce the risk of inconsistencies

src/data/pokemon-species.ts
```diff
@@ -690,6 +690,62 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
...
+    const formText = Utils.capitalizeString(formKey, "-", false, false) || "";
+    const speciesName = Utils.capitalizeString(Species[this.speciesId], "_", true, false);
+    let ret: string = "";
+
+    const region = this.getRegion();
+    if (this.speciesId === Species.ARCEUS) {
+      ret = i18next.t(`pokemonInfo:Type.${formText?.toUpperCase()}`);
+    } else if ([ SpeciesFormKey.MEGA_X, SpeciesFormKey.MEGA_Y, SpeciesFormKey.GIGANTAMAX_RAPID, SpeciesFormKey.GIGANTAMAX_SINGLE ].includes(formKey)) {
```
Wlowscha, line 727:
```suggestion
    } else if ([ SpeciesFormKey.MEGA_X, SpeciesFormKey.MEGA_Y, SpeciesFormKey.GIGANTAMAX_RAPID, SpeciesFormKey.GIGANTAMAX_SINGLE ].includes(formKey as SpeciesFormKey)) {
```

src/data/pokemon-species.ts
Fontbane: Yeah, I suppose checking for key being an empty string would make the lookup skip redundant. Will fix.

src/data/pokemon-species.ts
```diff
@@ -690,6 +690,62 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
...
+      }
+    } else if (this.speciesId === Species.ETERNAL_FLOETTE) {
+      ret = i18next.t("pokemonForm:floetteEternalFlower");
+    } else if (this.speciesId === Species.BLOODMOON_URSALUNA) {
+      ret = i18next.t("pokemonForm:ursalunaBloodmoon");
+    } else {
+      const regionalName = i18next.t(`pokemonForm:regionalForm${Region[region]}`);
+      ret = regionalName;
+    }
+    return ret + (append ? this.name : "");
```
Wlowscha, line 746: With the way it is right now, the form name and the name are connected with no space in between--this is bad for english but might be good for some other languages, so probably this needs to be localized as well.

Something like 
`return append ? i18next.t(`pokemonForm:appendName`, { form: ret, name: this.name }) : ret`

alcremie

src/ui/pokedex-page-ui-handler.ts
```diff
@@ -1366,13 +1328,14 @@ export default class PokedexPageUiHandler extends MessageUiHandler {
...
+                      const preFormIndex: number = preSpecies?.forms.find(f => f.formKey === pre.preFormKey)?.formIndex ?? 0;
 
                       const conditionText: string = pre.description;
 
                       options.push({
                         label: pre.preFormKey ?
-                          this.getFormString(pre.preFormKey, preSpecies ?? this.species, true) :
-                          this.getRegionName(preSpecies ?? this.species),
+                          (preSpecies ?? this.species).getFormNameToDisplay(preFormIndex, true) :
+                          (preSpecies ?? this.species).getExpandedSpeciesName(),
```
Wlowscha, line 1338: This is displaying regional forms incorrectly: since they do note have a formKey, the name is not appended
rapidashgalar

src/ui/pokemon-info-container.ts
```diff
@@ -218,23 +217,7 @@ export default class PokemonInfoContainer extends Phaser.GameObjects.Container {
...
-        const i18key = `pokemonForm:${speciesName}${formText}`;
-        if (i18next.exists(i18key)) {
-          formName = i18next.t(i18key);
-        } else {
-          const rootSpeciesName = Utils.capitalizeString(Species[pokemon.species.getRootSpeciesId()], "_", true, false);
-          const i18RootKey = `pokemonForm:${rootSpeciesName}${formText}`;
-          formName = i18next.exists(i18RootKey) ? i18next.t(i18RootKey) : formText;
-        }
-      }
+      const formName = pokemon.species.getFormNameToDisplay(pokemon.formIndex);
```
SirzBenjie, line 220: +10 points for extracting this logic out into a common function.

src/data/pokemon-species.ts
```diff
@@ -690,6 +690,62 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
...
+    const formKey = this.forms?.[formIndex!]?.formKey;
+    const formText = Utils.capitalizeString(formKey, "-", false, false) || "";
+    const speciesName = Utils.capitalizeString(Species[this.speciesId], "_", true, false);
+    let ret: string = "";
+
+    const region = this.getRegion();
+    if (this.speciesId === Species.ARCEUS) {
+      ret = i18next.t(`pokemonInfo:Type.${formText?.toUpperCase()}`);
+    } else if ([ SpeciesFormKey.MEGA_X, SpeciesFormKey.MEGA_Y, SpeciesFormKey.GIGANTAMAX_RAPID, SpeciesFormKey.GIGANTAMAX_SINGLE ].includes(formKey as SpeciesFormKey)) {
+      return i18next.t(`battlePokemonForm:${formKey}`, { pokemonName: "" });
```
SirzBenjie, line 728: Am I allowed to be nit picky about style? I know we don't mandate formatting, but imo for cases like these especially, the `else if` should really go on the next line after the closing curly brace. The indentation matches the body and it's throwing me off.

(this is an observation only and should not hold up this PR)

src/data/pokemon-species.ts
```diff
@@ -690,6 +690,62 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
...
+      const i18key = `pokemonForm:${speciesName}${formText}`;
+      if (i18next.exists(i18key)) {
+        ret = i18next.t(i18key);
+      } else {
+        const rootSpeciesName = Utils.capitalizeString(Species[this.getRootSpeciesId()], "_", true, false);
+        const i18RootKey = `pokemonForm:${rootSpeciesName}${formText}`;
+        ret = i18next.exists(i18RootKey) ? i18next.t(i18RootKey) : formText;
+      }
+    } else if (this.speciesId === Species.ETERNAL_FLOETTE) {
+      ret = i18next.t("pokemonForm:floetteEternalFlower");
```
SirzBenjie, line 739: Why is the existing structure hardcoding the check for this form name? Surely this `i18RootKey` should be a property that exists on the pokemon itself rather than being hardcoded here. So hardcoding `pokemonForm:floetteEternalFlower` is a code smell.
I think hardcoding keys like this is unsustainable and it shouldn't be added in new code. Computing this root key should be attached to a pokemon instance.

This is not just a nit and *should* hold up this PR, in my opinion. I'd like some additional feedback from other devs on this.

SirzBenjie (changes requested):
The code logic looks good to me, but I have some nits with the specific implementation choices. The long if / else if chain in `getFormNameToDisplay` indicates to me that the logic is too conflated and is violating SoC.

src/data/pokemon-species.ts
DayKev: The current eslint settings mandate the `else if` being on the same line as the brackets. I've wanted to change it too but never got around to asking about it or anything.

src/data/pokemon-species.ts
DayKev: Iirc the handling of this mon is a bit jank because of gamefreak and us matching their data in terms of forms/etc. That's not to say it can't be improved though.

SirzBenjie: Yeah but this doesn't really address my issue which is about where this data should reside. I  really don't think the i18RootKey should be constructed here.

src/data/pokemon-species.ts
Wlowscha: suggestion
```
return ret + (append ? ` ${this.name}` : ""); 
```

src/data/pokemon-species.ts
Wlowscha: I think it's fine in this case. Even if we had a separate method or attribute, it would anyway only be used once here. The options are: 
- Have a separate method to generate the key using this same logic (which seems to defeat the purpose)
- Have a method which is overridden for specific species and forms (but that would require introducing subclasses for species and forms, which is a massive refactoring) 
- Have an attribute which is set when creating each instance (meaning that we would take the logic of what this function does automatically in one place, and do it manually across different instances).
None of these solutions seem ideal. I believe it's more expressive to simply deal with the exceptions where it's required, especially since we might have to come back to this later on, and change the logic, if there are localisation issues.

src/data/pokemon-species.ts
Fontbane: image
I'M ALWAYS RIGHT

src/data/pokemon-species.ts
Fontbane: The issue is that for our purposes it's a separate species and not a form at all, but we want to show a form name for it. fwiw I do have something slightly more robust for this in mind, but it'll have to wait for next patch because it requires more locale work.

src/data/pokemon-species.ts
SirzBenjie: Next patch changing this incoming.

src/data/pokemon-species.ts
SirzBenjie: >   None of these solutions seem ideal. I believe it's more expressive to simply deal with the exceptions where it's required, especially since we might have to come back to this later on, and change the logic, if there are localisation issues.

This makes a lot of sense, thanks for engaging! I think you're right.
Especially since this has been given critical thought. PR is good to go for me

SirzBenjie (dismissed):

src/data/pokemon-species.ts
```diff
@@ -690,6 +690,62 @@ export default class PokemonSpecies extends PokemonSpeciesForm implements Locali
...
+  getFormNameToDisplay(formIndex: number = 0, append: boolean = false): string {
+    const formKey = this.forms?.[formIndex!]?.formKey;
+    const formText = Utils.capitalizeString(formKey, "-", false, false) || "";
+    const speciesName = Utils.capitalizeString(Species[this.speciesId], "_", true, false);
+    let ret: string = "";
+
+    const region = this.getRegion();
+    if (this.speciesId === Species.ARCEUS) {
+      ret = i18next.t(`pokemonInfo:Type.${formText?.toUpperCase()}`);
+    } else if ([ SpeciesFormKey.MEGA_X, SpeciesFormKey.MEGA_Y, SpeciesFormKey.GIGANTAMAX_RAPID, SpeciesFormKey.GIGANTAMAX_SINGLE ].includes(formKey as SpeciesFormKey)) {
```
Wlowscha, line 727: suggestion
``
} else if ([ SpeciesFormKey.MEGA, SpeciesFormKey.MEGA_X, SpeciesFormKey.MEGA_Y, SpeciesFormKey.PRIMAL, SpeciesFormKey.GIGANTAMAX, SpeciesFormKey.GIGANTAMAX_RAPID, SpeciesFormKey.GIGANTAMAX_SINGLE, SpeciesFormKey.ETERNAMAX ].includes(formKey as SpeciesFormKey)) {
``

Wlowscha, line 727:
```
suggestion
    } else if ([ SpeciesFormKey.MEGA, SpeciesFormKey.MEGA_X, SpeciesFormKey.MEGA_Y, SpeciesFormKey.PRIMAL, SpeciesFormKey.GIGANTAMAX, SpeciesFormKey.GIGANTAMAX_RAPID, SpeciesFormKey.GIGANTAMAX_SINGLE, SpeciesFormKey.ETERNAMAX ].includes(formKey as SpeciesFormKey)) {
```

Fontbane: Isn't that just all of them lol

SirzBenjie: Can we slap this array in a constant somewhere in the file instead of hardcoding it here?

Wlowscha: We could just go with an `Object.values(SpeciesFormKey) as string[]`, but that requires removing these keys:
```
  ORIGIN = "origin",
  INCARNATE = "incarnate",
  THERIAN = "therian",
```
and replacing them with the string where they show up in the code (which is very few places...)
It can be done, I just wouldn't like to break something accidentally right now.

I'm thinking later on we should anyway take a look at all the forms and the way they are stored, which is a bit inconsistent, and also make extra sure that the localization works for all forms and languages, maybe this cleanup can be part of that.

Fontbane: I think we're too close to merge to get too deep with these. Next update I plan to do a larger sweep of form names.

Wlowscha (approved):

src/data/pokemon-species.ts
bavalpey: Sounds good.

SirzBenjie (approved):

src/data/pokemon-species.ts
SirzBenjie: That makes sense. This is more of a refactor and should probably not hold up this PR.

Issue #5311, opened 2025-02-13
[Bug] Several mons which require knowing a move to evolve can't evolve before certain level


Fontbane:
### Describe the bug

As in the main series, several species evolve by leveling up while knowing a certain move. However, some of these also have an evolution level set here. **That means they don't evolve until the level they actually learn the move at.** This is probably meant to help with wild/trainer evolution delays, but there's a couple issues with it:
- Some of these moves can be learned outside of level up. For example, Clobbopus can learn Taunt via TM (previously unavailable due to being unimplemented). There could also be Mystery Events that teach these moves, and the player could use fusion to teach them.
- Level up learnsets are subject to change, and in fact are planned to change soon.
- Wild evolution delay is... very problematic for so many reasons that it probably requires its own issue

Affected species:
- Tangela (34, Ancient Power)
- Lickitung (32, Rollout)
- Aipom (32, Double Hit)
- Yanma (33, Ancient Power)
- Girafarig (32, Twin Beam)
- Dunsparce (32, Hyper Drill)
- Stantler (25, Psyshield Bash)
- Tsareena (28, Stomp)
- Clobbopus (35, Taunt) (this one's especially bad because Taunt is a TM, I'm posting a fix soon)
- Hisuian Qwilfish (28, Barb Barrage)
- Primeape (35, Rage Fist)

### Reproduction

Level up one of the affected mons with the necessary move in its learnset before the listed level. It won't evolve.

### Expected behavior

The affected mons can evolve before the listed level if they know the move.

Snailman11:
Was this discussed at all with other balance members? I feel like having some consistency for evolution rather than it being partially luck-based during early waves is completely reasonable.

Fontbane:
> Was this discussed at all with other balance members? I feel like having some consistency for evolution rather than it being partially luck-based during early waves is completely reasonable.
> 
> Is there some planned, in progress alternative to how evolution delays currently function to enable these types of mons in a way that Bal wants them to if the level restrictions are removed?

I don't think this behavior is intended. Evolution delays are a continuing pain in the neck for balance team and in need of a refactor. The level requirements for these evolutions definitely reads as "jank creating arbitrary player friction" rather than a consciously decided-upon restriction.

Snailman11:
<https://github.com/pagefaultgames/pokerogue/pull/5312>
Desirable behaviour according to balance by 3 to 1 majority,

PR #5312, opened 2025-02-13
[Bug] Fix Clobbopus being unable to evolve before lv 35

Fontbane:

- Fixes an issue where Clobbopus was unable to evolve before level 35, even if it knew Taunt

- Fixes one of the species affected by #5311 
- I felt this one was important as its evolution move is a TM, so this case is much more feasible for a player to encounter
- The other mons will require a slight refactor of evolution delay

- Changes Clobbopus's evolution level to 1 and gives it `SpeciesWildEvolutionDelay.LONG`

## How to test the changes?

```ITEM_REWARD_OVERRIDE: [{ "name": "TM_GREAT", type: Moves.TAUNT }]```
This will give you the TM for Taunt in your rewards. Start with Clobbopus and it should evolve by wave 5.

## Checklist

### DIFF
--- a/src/data/balance/pokemon-evolutions.ts
+++ b/src/data/balance/pokemon-evolutions.ts
@@ -1668,7 +1668,7 @@ export const pokemonEvolutions: PokemonEvolutions = {
     new SpeciesEvolution(Species.APPLETUN, 1, EvolutionItem.SWEET_APPLE, null, SpeciesWildEvolutionDelay.LONG)
   ],
   [Species.CLOBBOPUS]: [
-    new SpeciesEvolution(Species.GRAPPLOCT, 35, null, new MoveEvolutionCondition(Moves.TAUNT)/*Once Taunt is implemented, change evo level to 1 and delay to LONG*/)
+    new SpeciesEvolution(Species.GRAPPLOCT, 1, null, new MoveEvolutionCondition(Moves.TAUNT), SpeciesWildEvolutionDelay.LONG)
   ],
   [Species.SINISTEA]: [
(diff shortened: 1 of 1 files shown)


damocleas:
Being able to luck into Grapploct super early is not desirable behavior

Blitz425:
If this is how it works in mainline, I see absolutely no issue with this implementation

Fontbane:
> Being able to luck into Grapploct super early is not desirable behavior

How is that any different from lucking into an item evolution?

SirzBenjie (approved):
Codewise, looks good.

I'll abstain from balance considerations in my approval.

EscaShark:
Can you alter the PR to include every other mon where this applies as well? This is most visible on Clobbopus because Taunt is a TM, but technically all of these could get their moves earlier through fuse tutoring. I don't think splitting them all into different PRs is warranted when it's the same effective discussion every time.

It should also be noted that weirdly, Bonsly and Mime Jr. do evolve at any level despite not getting their evo moves at level 1, so I don't think this was ever a hard rule to begin with.

damocleas:
> > Being able to luck into Grapploct super early is not desirable behavior
> 
> How is that any different from lucking into an item evolution? This restriction is _completely arbitrary_ and only present because Taunt used to be (N).

I'm more of a fan of restricting that slightly so the first 10 waves aren't instantly evolved but otherwise idc, also the above, should just try to catch them all if there's more

damocleas:
Evo delay + this is there to not let you get a mon earlier than it learns the move naturally, can be revisited later (closing off of discord convo, didn't need to stay open this long)

Wlowscha:
Closing this for now makes sense, it will need to be revisited in the wider context of a rework of how evo delays work in general, plus the other evos mentioned in the issue above

PR #5319, opened 2025-02-14
[Bug][Hotfix] Display caught forms in dex as intended

Wlowscha:
All caught forms are displayed and their data can be accessed.

There was a bug where all forms except the default one would be marked as uncaught. This is now fixed.

The `getFullUnlocksData` function (introduced to filter out impossible unlocks from old save data) was not taking into account the contribution to the `DexAttr.caughtData` bigint from caught forms. This is fixed by looping over forms and adding incremental contributions.

## Screenshots/Videos
Showing that forms are displayed properly with the hotfix:
vivis
unowns
pikacool
pikas

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "pokemon-rogue-battle",
-	"version": "1.6.0",
+	"version": "1.6.1",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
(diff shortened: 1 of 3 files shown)


DayKev (approved):

damocleas:
before
image
after
image
seems to fix it

damocleas (approved):

PR #5330, opened 2025-02-15
[Misc] Added a Daily Run Seed Override to the overrides. Only works locally.

Jimmybald1:

-

Makes testing a specific new daily run much easier. Instead of searching where in the code the seed is set, now you can just use the override.

Added:
DAILY_RUN_SEED_OVERRIDE: string

The override only works locally.

## Screenshots/Videos

-

## How to test the changes?

Set the override to a string and the daily run should change to that seed. (locally)
DAILY_RUN_SEED_OVERRIDE: ""

## Checklist


src/overrides.ts
```diff
@@ -47,6 +47,7 @@ class DefaultOverrides {
   // -----------------
   /** a specific seed (default: a random string of 24 characters) */
   readonly SEED_OVERRIDE: string = "";
+  readonly DAILY_RUN_SEED_OVERRIDE: string = "";
```
DayKev, line 50:
```suggestion
  readonly DAILY_RUN_SEED_OVERRIDE: string | null = null;
```
You used `!isNullOrUndefined()` as the check for activating the override but `""` is neither of those.

DayKev (approved):

damocleas (approved):

Issue #5113, opened 2025-01-12
[Bug] Splash does not Display a Message when Used


Snailman11:
### Describe the bug

Splash does not display a message when used.

### Reproduction

Use Splash, view the no message in the textbox.

### Expected behavior

https://github.com/user-attachments/assets/afcd42d5-f288-4dca-8be1-b7d88fd2813c

It should display "But nothing happened!" immediately after Splash is used.

https://abcboy101.github.io/poke-corpus/#q=dp.324.795&s=QPNnFvRRZc3T9w

### Screenshots / Videos

https://github.com/user-attachments/assets/f6ebe2c3-8c2b-4a83-8bb9-a6d66a04b530

Jacky3003:
Hello, wondering if I could pick this up as a first issue to work on.

Snailman11:
> Hello, wondering if I could pick this up as a first issue to work on.

You probably could

SirzBenjie:
Fixed by #5392

PR #5340, opened 2025-02-15
[Bug] Fix splash not displaying a message when used

Jacky3003:

When splash is used, the user should be able to see the message "But nothing happened!".

The following problem this PR fixes is an issue that is linked [here](https://github.com/pagefaultgames/pokerogue/issues/5113). This change makes the behaviour of splash to be more consistent with its usage in the mainline games.

A new class has been added in `src/data/battler-tags.ts` to allow for the display of the message. Localisation also needs to be done for languages that are not included in the attached localisation PR found [here](https://github.com/pagefaultgames/pokerogue-locales/pull/125). 
## Screenshots/Videos

https://github.com/user-attachments/assets/88b5ee31-e527-420d-9640-871490ca9939
## How to test the changes?

Override the move set for a pokemon to include splash in `src/overrides.ts`, then test in a run accordingly.

## Checklist
  - [X] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?

- [X] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?
  - [X] If so, please leave a link to it here: https://github.com/pagefaultgames/pokerogue-locales/pull/125

### DIFF
--- a/src/data/battler-tags.ts
+++ b/src/data/battler-tags.ts
@@ -1646,6 +1646,16 @@ export class CenterOfAttentionTag extends BattlerTag {
   }
 }
 
+export class SplashTag extends BattlerTag {
+  constructor() {
+    super(BattlerTagType.SPLASH, BattlerTagLapseType.TURN_END, 1, Moves.SPLASH, undefined, false);
+  }
+  onAdd(pokemon: Pokemon): void {
+    super.onAdd(pokemon);
+    globalScene.queueMessage(i18next.t("battlerTags:splashLapse"));
+  }
+}
+
 export class AbilityBattlerTag extends BattlerTag {
   public ability: Abilities;
 
@@ -3000,6 +3010,8 @@ export class MagicCoatTag extends BattlerTag {
  */
 export function getBattlerTag(tagType: BattlerTagType, turnCount: number, sourceMove: Moves, sourceId: number): BattlerTag {
   switch (tagType) {
+    case BattlerTagType.SPLASH:
+      return new SplashTag();
     case BattlerTagType.RECHARGING:
       return new RechargingTag(sourceMove);
     case BattlerTagType.BEAK_BLAST_CHARGING:
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -8791,6 +8791,7 @@ export function initMoves() {
     new AttackMove(Moves.PSYWAVE, Type.PSYCHIC, MoveCategory.SPECIAL, -1, 100, 15, -1, 0, 1)
       .attr(RandomLevelDamageAttr),
     new SelfStatusMove(Moves.SPLASH, Type.NORMAL, -1, 40, -1, 0, 1)
+      .attr(AddBattlerTagAttr, BattlerTagType.SPLASH)
       .condition(failOnGravityCondition),
     new SelfStatusMove(Moves.ACID_ARMOR, Type.POISON, -1, 20, -1, 0, 1)
       .attr(StatStageChangeAttr, [ Stat.DEF ], 2, true),
--- a/src/enums/battler-tag-type.ts
+++ b/src/enums/battler-tag-type.ts
@@ -95,4 +95,5 @@ export enum BattlerTagType {
   ENDURE_TOKEN = "ENDURE_TOKEN",
   POWDER = "POWDER",
   MAGIC_COAT = "MAGIC_COAT",
+  SPLASH = "SPLASH"
 }
--- a/src/overrides.ts
+++ b/src/overrides.ts
@@ -32,7 +32,11 @@ import { WeatherType } from "#enums/weather-type";
  * }
  * ```
  */
-const overrides = {} satisfies Partial<InstanceType<typeof DefaultOverrides>>;
+const overrides = {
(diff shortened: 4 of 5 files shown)


Jacky3003:
Wondering if there is a better solution for the secret power test case (just a side note for those reviewing).

SirzBenjie (changes requested):
Okay, so this can be implemented without having a battler tag.
The way to do it is to actually add a move attribute for having a custom fail message. Coincidentally, I'm working on the flavor text for focus punch right now, and needed to add such an attribute.

Battler tags are meant to be used for things like effects that have something happen when they wear off -- Beak Blast, recharging from recharge moves, etc. It's sort of overkill to use a battler tag for splash.

It might actually be better for me to merge this into the PR I'm working on for focus punch. You already did the locales PR for the flavor text.

SirzBenjie:
> Wondering if there is a better solution for the secret power test case (just a side note for those reviewing).

Just curious, what is the relationship between Secret Power and this PR? I didn't see it mentioned.

Madmadness65 (changes requested):
You don't need all this to add the flavor text message for Splash. Here's a video of my quick implementation:

https://github.com/user-attachments/assets/d679268e-ac6d-4bb0-b90d-d26300d768a3

And here is the code block I quickly wrote for it, in `move.ts` :
```ts
export class SplashAttr extends MoveEffectAttr {
  constructor() {
    super();
  }

    globalScene.queueMessage("But nothing happened!"); // Replace this with the proper i18n localization string
    return true;
  }
}
```
Therefore, all that you need to do is add `.attr(SplashAttr)` to Splash and the message works.

SirzBenjie:
> You don't need all this to add the flavor text message for Splash. Here's a video of my quick implementation:
> 
> And here is the code block I quickly wrote for it, in `move.ts` :
> 
> ```ts
> export class SplashAttr extends MoveEffectAttr {
>   constructor() {
>     super();
>   }
> 
>   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
>     globalScene.queueMessage("But nothing happened!"); // Replace this with the proper i18n localization string
>     return true;
>   }
> }
> ```
> 
> Therefore, all that you need to do is add `.attr(SplashAttr)` to Splash and the message works.


Actually, I think this is what is technically correct. Splash being used can't count as it failing, since it plays the animation. So Madmadness is actually more correct and this should have no overlap with `getFailedMessage`

Jacky3003:
> Just curious, what is the relationship between Secret Power and this PR? I didn't see it mentioned.

@SirzBenjie For the secret power test case, splash was used as a placeholder move to test a niche interaction with serene grace and the pledge moves:
```typescript
      game.override
        .moveset([ Moves.FIRE_PLEDGE, Moves.WATER_PLEDGE, Moves.SECRET_POWER, Moves.SPLASH ])
        .ability(Abilities.SERENE_GRACE)
        .enemyMoveset([ Moves.SPLASH ])
        .battleType("double");
      await game.classicMode.startBattle([ Species.BLASTOISE, Species.CHARIZARD ]);

      const sereneGraceAttr = allAbilities[Abilities.SERENE_GRACE].getAttrs(MoveEffectChanceMultiplierAbAttr)[0];
      vi.spyOn(sereneGraceAttr, "apply");

      game.move.select(Moves.WATER_PLEDGE, 0, BattlerIndex.ENEMY);
      game.move.select(Moves.FIRE_PLEDGE, 1, BattlerIndex.ENEMY_2);

      await game.phaseInterceptor.to("TurnEndPhase");

      let rainbowEffect = game.scene.arena.getTagOnSide(ArenaTagType.WATER_FIRE_PLEDGE, ArenaTagSide.PLAYER);
      expect(rainbowEffect).toBeDefined();

      rainbowEffect = rainbowEffect!;
      vi.spyOn(rainbowEffect, "apply");

      game.move.select(Moves.SECRET_POWER, 0, BattlerIndex.ENEMY);
      game.move.select(Moves.SPLASH, 1);

      expect(sereneGraceAttr.apply).toHaveBeenCalledOnce();
      expect(sereneGraceAttr.apply).toHaveLastReturnedWith(true);

      expect(rainbowEffect.apply).toHaveBeenCalledOnce();
```
 found out it broke when using other status moves (like idef, for example). It's probably not relevant to the commit and seems like a separate issue. Also is the plan still to merge this into the flavor text for focus punch? If not I'll do the implementation that madmadness suggested on this PR.

SirzBenjie:
> Also is the plan still to merge this into the flavor text for focus punch? If not I'll do the implementation that madmadness suggested on this PR.

Sorry for the late reply on this.  Once I figured out that the splash message is actually part of its use and not a special fail message (I had always thought that splash was literally "failing"), my Focus Punch PR is no longer related.
It does, however, seem like Madmadnes has incorporated the suggested changes into a new PR #5392

bavalpey:
> Seems like this PR can be closed then. Thank you for the help @SirzBenjie @Madmadness65

Thank you for the contribution!!

SirzBenjie:
> Seems like this PR can be closed then. Thank you for the help @SirzBenjie @Madmadness65

Thank you for your contribution!!

PR #5341, opened 2025-02-16
[Bug] [Move] Add focus punch lost focus message

### LINKED ISSUE #4696: [Bug] Focus Punch Missing Post-Hit Message
### Describe the bug

It is missing the message it gets after being hit

### Reproduction

[FocusPunch.txt](https://github.com/user-attachments/files/17452368/FocusPunch.txt)
Use Earth Power as the Minior on the first turn

### Expected behavior

If the user is hit before the attack, it will say, "<Pokémon> used Focus Punch!" before it says the "<Pokémon> lost its focus and couldn't move!". Wild Pokémon executing this move while the player unsuccessfully attempts to flee will always display the charge-up animation before executing the move.

Generation V onwards
It no longer states "<Pokémon> used Focus Punch!" preceding "<Pokémon> lost its focus and couldn't move!"

### Screenshots / Videos

https://github.com/user-attachments/assets/bc99c44b-5282-4a22-bd4b-8a6058a4f895

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

SirzBenjie:
If a mon gets hit after tightening its focus, when it attempts to use the move, it will now display the proper message.
Fixes #4696.

Is very important bugfix.

Also, it sort of refactors the `getFailedMessage` method which had an additional argument that wasn't actually used and made no sense anyway.

Adds a new MoveAttribute, `PreUseInterruptAttr`.
Its constructor takes a `MoveConditionFunc` and an optional string message or a function that will compute a string from the user, target, and move.

Its apply method just returns the result of evaluating the condition supplied in the constructor.
It overrides the `getFailedMessage` function.
This will once again check the condition (which is unfortunate but necessary because we have no better way of assigning attributes to moves that are in flight and avoiding this would needlessly complicate the logic of the `move-phase`) returning the (evaluated) message if the condition is met.
Otherwise, returns undefined.

**Note** applying this attribute precludes failure due to weather moves, type immunities, or other move conditions. That is, the failure will be attributed to the interrupt attribute and none of the others will need to be checked.

I also modified the `getFailedMessage` to return `undefined` instead of `null` as this is consistent with how JavaScript methods use undefined return values (e.g. see `array.find`).
It also makes it nicer to pass in default values elsewhere.

Focus Punch has had its condition replaced with this new attribute, and the old condition is now passed into the condition of the constructor, along with the message using the i18next key called `lostFocus`
For reference, this was:
```ts
condition((user, target, move) => !user.turnData.attacksReceived.find(r => r.damage))
```

## Screenshots/Videos
<details><summary>Losing focus and not displaying the message</summary>

https://github.com/user-attachments/assets/b4571c53-8f1e-4fc9-87cd-88ab245bacae
</details>

## How to test the changes?
```
const overrides = {
  MOVESET_OVERRIDE: [ Moves.QUICK_ATTACK ],
  OPP_MOVESET_OVERRIDE: [ Moves.FOCUS_PUNCH ],
  OPP_SPECIES_OVERRIDE:  Species.MAGIKARP,
  OPP_ABILITY_OVERRIDE: Abilities.STURDY,
```

For automated test:
```
npx vitest run src/test/moves/focus_punch.test.ts -t "should replace"
```

## Checklist

  - [x] If so, please leave a link to it here: https://github.com/pagefaultgames/pokerogue-locales/pull/126


Madmadness65 (dismissed):

src/test/moves/focus_punch.test.ts
```diff
@@ -130,4 +131,15 @@ describe("Moves - Focus Punch", () => {
       expect(game.scene.phaseQueue.find(phase => phase instanceof MoveHeaderPhase)).toBeDefined();
     }
   );
+  it("should replace the 'but it failed' text when it the user gets hit", async () => {
```
Wlowscha, line 134:
```suggestion
it("should replace the 'but it failed' text when the user gets hit", async () => {
```

damocleas (dismissed):

src/data/move.ts
```diff
@@ -1320,6 +1317,54 @@ export class PreMoveMessageAttr extends MoveAttr {
...
 }
 
+/**
+ * Attribute for moves that can be conditionally interrupted to be considered to
+ * have failed before their "useMove" message is displayed. Currently used by
+ * Focus Punch.
+ * @extends MoveAttr
+ */
+export class PreUseInterruptAttr extends MoveAttr {
+  protected message: string | ((user: Pokemon, target: Pokemon, move: Move) => string) | undefined;
```
DayKev, line 1327:
```suggestion
  protected message?: string | ((user: Pokemon, target: Pokemon, move: Move) => string);
```

DayKev (commented):
Just one nit.

damocleas (approved):

PR #5364, opened 2025-02-18
[Bug] Fix tera type access

Xavion3:
Some parts of the code weren't correctly acknowleding tera type overrides, like terapagos was displaying as tera normal sometimes.

The code accessed tera type directly instead of using getTeraType in a couple of places, this is now fixed.

The code accessed tera type directly instead of using getTeraType in a couple of places, this is now fixed.

## Screenshots/Videos
image

## How to test the changes?
Terastallize terapagos. It should say tera stellar.

## Checklist

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -2831,7 +2831,7 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       stabMultiplier.value += 0.5;
     }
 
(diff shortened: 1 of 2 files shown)


SirzBenjie (approved):

DayKev:
If `.getTeraType()` is supposed to be used instead of `.teraType` then you should make the `teraType` field `protected`.

Xavion3:
You still need access to the actual tera type for save/load code thanks to fusions, you should just use getTeraType any time the type a pokemon is terastallizing into is important, which is almost all of them.

DayKev:
You could add a param to `getTeraType()` that makes it return the "raw" value if the param is `true`.

Madmadness65 (approved):

PR #3364, opened 2024-08-05
[Balance] Rebalance many Gym Leader teams, add Koga

Fontbane:
## What are the changes?

- Adds Koga as a Gym Leader
  - Koga is found in the Swamp biome. Roxie has been moved to Slum biome.
- Changes the teams of the following Gym Leaders:
  - Giovanni (now Ground-type as in main series)
  - Bugsy, Whitney, Chuck, Jasmine
  - Brawly, Norman, Juan
  - Gardenia, Maylene, Crasher Wake, Fantina, Volkner
  - Byron had an internal change that probably shouldn't be noticed in-game
  - Cilan, Cress, Cheren, Lenora, Roxie, Burgh, Clay
  - Clemont
  - Kabu, Bea, Allister, Raihan
  - Katy, Brassius, Iono, Kofu, Tulip, Ryme, Grusha
  - Larry (Gym Leader) had his Dunsparce and Starly swapped
- Adjusts the teams of the following Elite Four members who had their Gym teams adjusted:
  - Koga
  - Bea
  - Allister
  - Raihan
- Certain Gym Leaders and E4s can be set to terastalize the first of their signature species. For example, if Iono has this configuration turned on, her Misdreavus (set to first slot) will be Tera Electric. This allows Paldean Trainers to use their Tera mons from SV.  The following Trainers have this turned on:
  - Katy
  - Brassius
  - Iono
  - Kofu
  - Larry
  - Ryme
  - Tulip
  - Grusha
  - Lacey
  - As far as changelogs are concerned, you can word this like "Gave Paldean Gym Leaders their Terastalized off-type aces from the main series"

## Why am I doing these changes?

- To make Gym Leaders a little more balanced and faithful to their main series appearances
- Move some mons around to avoid repeats between Gym Leaders and their region's E4 or champion
- Give Paldea Gym Leaders their tera mons
- Every other Gym Leader is in this, even ones that retired or joined the E4, so Koga should be in too

## What did change?

- Split Koga into "koga_gym" and "koga"
- Documented getRandomTeraModifiers function
- Added a new boolean, "forceTeraFirstSlot" to the TrainerConfig class and a new method, "setForceTeraFirstSlot()"
-  Added new parameter to getRandomTeraModifiers for whether to force terastalization for the mon in the last slot (first signature species)
- Added functionality for giving Tera modifier to last enemy party member before deciding random teras (I'm aware Tera mechanics will change at some point, but this technically doesn't change how the Tera modifier is applied, just who it's applied to)
[...]

PR #5367, opened 2025-02-18
[Balance] Paldean Gym Leader, E4, and other Tera-related trainer adjustments

Fontbane:

- Paldean Gym Leaders and E4 members now Terastallize their ace (including before floor 100)
- Because of this, these trainers had their teams adjusted to be closer to canon
- Adjustments to Lance:
  - Lance's slot of Tyranitar/Garchomp/Kommo-o will now be Tera Dragon rather than just Tera to their primary type
  - His Tyranitar has Unnerve, Garchomp has Rough Skin, and Kommo-o has Soundproof
 - Flint will Tera his Steelix/Lopunny slot to Fire-type when fought
 - Prevents Shedinja from Terastallizing to non-Bug type on trainers

- Tera rework in, this was desired to go alongside it
- Split off my massive PR #3364 

- Changed specialty type from array to just one
- initForGymLeader and initForEliteFour now have a "teraSlot" parameter, of which slot they should Tera. This accepts negative numbers, so -1 would be the last party member regardless of party size. Adjusted setRandomTeraModifiers code to accommodate this.
  - Also filtered Shedinja of non-Bug specialty types from the list of valid party members to tera.
  - If the trainer has more than one Tera allowed (unused), it will pick the specified slot first
- Gave admins the "specialtyType" parameter as well, in case we need it at some point (currently does nothing afaik)
- Sorted the species pools for E4 members (the variable slots) to be in natdex order just for organization
- Gym Leader team changes:
  - [On this spreadsheet (under construction)](https://docs.google.com/spreadsheets/d/1h2eYmmJG3dyWBpNmnekBTMXVyt6IM3Zf0-XA4vAY15I/edit?usp=sharing)

## Screenshots/Videos

Brassius
Screenshot_1080

## Checklist
  - Trainer forms matching specialty type was split to #5409
  - _This PR initially contained changes to moveset generation to account for tera type, which have been removed for now and recorded as the issue_ #5410
  - _It also initially removed the need to reroll evolutions to fit specialty type. This was removed for now and I left a comment with instructions for removing it._


src/battle-scene.ts
```diff
@@ -1523,6 +1523,25 @@ export default class BattleScene extends SceneBase {
         }
     }
 
+    if (species.speciesId === Species.ROTOM // Set Rotom form for Trainers to match specialty type
+      && this.currentBattle.battleType === BattleType.TRAINER && !isNullOrUndefined(this.currentBattle.trainer)
+      && [ Type.FLYING, Type.GHOST, Type.FIRE, Type.GRASS, Type.WATER, Type.ICE ].includes(this.currentBattle.trainer.config.specialtyType)) {
```
DayKev, line 1528:
```suggestion
    if (
      species.speciesId === Species.ROTOM // Set Rotom form for Trainers to match specialty type
      && this.currentBattle.battleType === BattleType.TRAINER
      && !isNullOrUndefined(this.currentBattle.trainer)
      && [ Type.FLYING, Type.GHOST, Type.FIRE, Type.GRASS, Type.WATER, Type.ICE ].includes(this.currentBattle.trainer.config.specialtyType)
    ) {
```

src/field/pokemon.ts
```diff
@@ -2348,10 +2348,11 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       weightMultiplier += 0.4;
     }
     const baseWeights: [Moves, number][] = movePool.map(m => [ m[0], Math.ceil(Math.pow(m[1], weightMultiplier) * 100) ]);
+    const checkTera = this.teraType !== Type.UNKNOWN && !this.isOfType(this.teraType); // Whether the mon has an off-type Tera. If so, moves matching Tera type (and Tera Blast) are considered STAB
```
Xavion3, line 2351: Use getTeraType, don't directly access teraType unless you've got a good reason to.

Xavion3: This applies to all the changes, I just only commented it on this line.

Madmadness65 (dismissed):

Madmadness65 (approved):

damocleas (approved):
Approving balance part

Madmadness65 (approved):

Xavion3:
I'm fixing the underlying issues here as this isn't actually attempting to fix the problem and introducing a save/reload bug is bad. So for now temporarily stuck until the actual bug is fixed.

Xavion3:
Save/loads fixed in #5414 rendering this partially obsolete. setRandomTeras no longer has a reason to be removed (not that it had a reason to be removed anyway).

Madmadness65 (dismissed):

src/data/trainer-config.ts
```diff
@@ -251,7 +254,7 @@ export class TrainerConfig {
   public partyMemberFuncs: PartyMemberFuncs = {};
   public speciesPools: TrainerTierPools;
   public speciesFilter: PokemonSpeciesFilter;
-  public specialtyTypes: Type[] = [];
+  public specialtyType: Type = Type.UNKNOWN;
```
Xavion3, line 257: If it's going to be an optional field, just make it an optional field.

src/data/trainer-config.ts
```diff
@@ -594,18 +597,21 @@ export class TrainerConfig {
...
   /**
-   * Sets random pokemon from the trainers team to instant tera. Uses their specialty types is they have one.
-   * @param count The amount of pokemon to have instant tera
+   * Sets random pokemon from the trainer's team to instant tera. Also sets Tera type to specialty type and checks for Shedinja as appropriate.
+   * @param count A callback (yucky) to see how many teras should be used
+   * @param slot Optional, a specified slot that should be terastallized. Wraps to match party size (-1 will get the last slot and so on).
    * @returns this
    */
-  setRandomTeraModifiers(count: () => integer): TrainerConfig {
+  setRandomTeraModifiers(count: () => number, slot?: number): TrainerConfig {
```
Xavion3, line 605: Just make count a number if you want, it made sense as a callback when we were messing with things like the elite four getting multiple teras.

Xavion3 (changes requested):
As above, some of this seems like it really shouldn't be approached this way.

src/data/trainer-config.ts
Fontbane: > Also if we're going to rework this function to this degree, it should just be two functions.

That's crazy because I basically did that but you complained about too much trainer code being refactored

DayKev (approved):

Madmadness65 (approved):

PR #5374, opened 2025-02-19
[Test] [Refactor] Automated Test Infra Refactor

SirzBenjie:
None

vitest 3 will help improve performance, which is necessary as I am working towards adding many unit tests.

Projects with the "src" structure generally have the tests folder as a sibling of the src folder, not a child. There are many benefits to this. I can go into them here, if you want.

I am working on implementing some of the optimizations that were done in https://github.com/Despair-Games/poketernity/pull/158 and https://github.com/Despair-Games/poketernity/pull/247.
Specifically, we should no longer be creating a new Global scene for every test (why the hell were we in the first place.. who knows).

I am also suggesting that we add [vitest caching](https://github.com/raegen/vite-plugin-vitest-cache) so that tests are only re-run based on source files or dependencies changing. The way this works is that the build files for each test are hashed, and then those hashes are stored in a cache.
This should make use of Github's cache action, which allows for caching items from workflows.

Moved the tests out of `src/test` and into `./test`
Updated all of the paths in the root folders to respect this: `tsconfig.json`, `vitest.workspaces.ts`, and `vitest.config.ts`.
Updated the imports inside the tests to refer to "#app" and "#test" when necessary, instead of using relative imports.

[TBD] Update workflow file to use caching and --no-isolate

## Screenshots/Videos
N/A

## How to test the changes?
Well, just run any test you want and see that it works.

## Checklist
  - ~~[] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?~~
- ~~[] Have I provided screenshots/videos of the changes (if applicable)?~~

  - ~~[ ] If so, please leave a link to it here:~~


DayKev (commented):
Btw you can remove the pre-test step now since the PR that added `--no-isolate` also added an overrides sanitization step which you already copied over.

PR #5406, opened 2025-02-24
[Bug] Fix #4972 Status-Prevention Abilities do not Cure Status

### LINKED ISSUE #4972: [Bug] Status-Prevention Abilities do not Cure Status
### Describe the bug

Immunity and Pastel Veil does not cure the Poison status after a Pokémon with Immunity is Poisoned.

Limber does not cure Paralysis after a Pokémon with Limber is Paralyzed

Insomnia does not cure Sleep after a Pokémon is put to Sleep

Water Veil does not cure Burns after a Pokémon is Burned.

Magma Armor probably does not cure Freeze, but I don't want to deal with the 10% proc.

Oblivious does not cure Taunt

Own Tempo does not cure Confusion

Link to Discord's Bug Report (Immunity ability does not stop Toxic Orb poisoning)
https://discord.com/channels/1125469663833370665/1254195713626083408

### Reproduction

Use Will-O-Wisp, Thunder Wave, Spore, and Toxic, or Oblivious and Own Tempo with a Pokémon that has Mold Breaker on the respective Ability.

This should also apply with Neutralizing Gas, which has issues with abilities reactivating (#3905) 

### Expected behavior

Toxic Poison should immediately be cured when the ability is reactivated
 https://replay.pokemonshowdown.com/gen9customgame-2257080241-fj97ehzhf1116bpgf33dkdn2tsz0f20pw

https://bulbapedia.bulbagarden.net/wiki/Category:Abilities_that_can_heal_non-volatile_status_conditions

https://bulbapedia.bulbagarden.net/wiki/Category:Abilities_that_can_heal_volatile_status_conditions

### Screenshots / Videos

Immunity + Mold Breaker

https://github.com/user-attachments/assets/39fb8558-88ff-446a-a0ce-79263b9cf25a

Insomnia + Mold Breaker

https://github.com/user-attachments/assets/86e91042-1149-48ac-ad50-427ae40d8f4f

Limber + Mold Breaker

https://github.com/user-attachments/assets/7f5a165a-5b3c-47a4-9829-f9f808380bdb

Pastel Veil + Mold Breaker

https://github.com/user-attachments/assets/0e0a30f9-b276-4ddb-9d06-db4fdcd7b07f

Water Veil + Mold Breaker

https://github.com/user-attachments/assets/1074720f-2ecd-4c96-903f-25da8c807fd4

Oblivious + Mold Breaker

https://github.com/user-attachments/assets/006cfb9d-3a42-4fbc-ba9c-8a88a036ee31

Own Tempo + Mold Breaker

https://github.com/user-attachments/assets/67f2b53e-84e9-44da-a0dc-ca9fe9cff33e

### Session export file

_No response_

### User data export file

_No response_

### Additional context

**PURIFYING SALT SHOULD NOT CURE STATUS UPON BEING GAINED/REACTIVATED, unlike other abilities with StatusEffectImmunityAbAttr**
> If the Pokémon has a status condition and gains this Ability, the status condition will not be cured and remains in effect.
-Bulba

-When being Switched out, if the status preventing ability is suppressed, then the status condition will not be healed.

https://replay.pokemonshowdown.com/gen9customgame-2257437771

-If Trace would give a Pokemon an Ability that prevents the status condition, then the status will be cured at the end of the round.

Normal switch against T spikes, Trace gaining Immunity
https://replay.pokemonshowdown.com/gen9customgame-2257432270

Slow U-Turn against T-Spikes, takes damage, then is cured
https://replay.pokemonshowdown.com/gen9customgame-2257434225

Fast U-turn against T-Spikes, does not take damage and is cured
https://replay.pokemonshowdown.com/gen9customgame-2257436315

-It seems that it waits for another Pokémon to make a move and cures itself at the end of that move. otherwise, it cures at the {very start of the next/very end of the] round.

-Check functionality for
Immunity is unsuppressed or gained by trace from faint switch in

<https://replay.pokemonshowdown.com/gen9doublescustomgame-2290281986>
Sweet Veil does not Heal Sleep like Insomnia, it only prevents it.

emdeann:
The abilities: Immunity, Insomnia, Limber, Magma Armor, Oblivious, Own Tempo, Thermal Exchange, Vital Spirit, Water Bubble, and Water Veil now correctly remove corresponding status effects when they are gained. I determined that other status prevention abilities do not reapply, so they aren't on here (but let me know if I missed something).

These abilities also immediately proc and remove their corresponding statuses if the status was applied while the ability was otherwise active (i.e. mold breaker)

Fixes #4972 

This reactivation part was trivial now that onGain activation as a whole is supported. I added `PostSummonHealStatusAbAttr` and `PostSummonRemoveBattlerTagAbAttr` and attached them to the relevant abilities. These will automatically be called in any "gaining" circumstance (swapping, copying, un-suppression once the neut gas fix is in, etc.)

This is not some hacky fix - these abilities _are_ post-summon and will activate immediately on summon if the pokemon has the corresponding status.

The Mold Breaker/etc. fix is much more arduous as it's not inherently supported. What I went with instead of trying to rework mold breaker was to attach a `PostApply(Status/BattlerTag)` attribute to each ability that immediately removes the respective status if the ability is now active again. This feels a little messy because now each of these abilities have a lot of attributes - if mold breaker were able to call onGain similar to neutralizing gas this would be alleviated, but that is nontrivial due to mold breaker only selectively impacting abilities (I can't think of a case besides this where an ability would need to proc after a mold breaker move). 

It also means I have to call `arena.setIgnoreAbilities(false)` before calling the PostApply attrs - this shouldn't be an issue as any mold breaker effects are over _after_ a status/tag has already been applied, but again feels less than ideal. Unsure if I can do better here or if it's just down to the implementation of mold breaker abilities.

In doing this I also made status clearing into a phase to avoid the status icon desyncing from the rest of the battle. Could that be a separate PR? Probably, but it's relevant to making this work nicely and will be necessary for the future ability display stuff anyway.

Open to lots of feedback on how I did any of the mold breaker stuff as it feels a little janky to me and I hate janky
## Screenshots/Videos

https://github.com/user-attachments/assets/2c0c7165-21f4-4b85-baab-be67c84aab20

https://github.com/user-attachments/assets/95c019af-ba2b-42bb-9c7c-ae94400f2d61

https://github.com/user-attachments/assets/dfad79dc-5391-4ce2-b352-7f6a2f080e4a

## How to test the changes?
Automated tests

## Checklist


src/data/ability.ts
```diff
@@ -2208,6 +2208,28 @@ export class PostSummonAddBattlerTagAbAttr extends PostSummonAbAttr {
   }
 }
 
+export class PostSummonRemoveBattlerTagAbAttr extends PostSummonAbAttr {
+  private immuneTags: BattlerTagType[];
+
+  /**
+   * @param immuneEffects - The status effects to which the Pokémon is immune.
```
DayKev, line 2215:
```suggestion
   * @param immuneTags - The {@linkcode BattlerTagType | battler tags} the Pokémon is immune to.
```

src/data/ability.ts
```diff
@@ -2403,6 +2425,37 @@ export class PostSummonTerrainChangeAbAttr extends PostSummonAbAttr {
   }
 }
 
+export class PostSummonHealStatusAbAttr extends PostSummonAbAttr {
+  private immuneEffects: StatusEffect[];
+  private statusHealed: StatusEffect;
+
+  /**
+   * @param immuneEffects - The status effects to which the Pokémon is immune.
```
DayKev, line 2433:
```suggestion
   * @param immuneEffects - The {@linkcode StatusEffect}s the Pokémon is immune to.
```

test/abilities/oblivious.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    await game.classicMode.startBattle([ Species.FEEBAS ]);
+    const enemy = game.scene.getEnemyPokemon();
+    enemy?.addTag(BattlerTagType.TAUNT); // sourceID needs to be defined
+    expect(enemy?.getTag(BattlerTagType.TAUNT)).toBeTruthy();
+
+    game.move.select(Moves.SKILL_SWAP);
+    await game.phaseInterceptor.to("BerryPhase");
+
+    expect(enemy?.status).toBeFalsy();
```
DayKev, line 49: You also made the battler tag tests check for status by mistake.

test/abilities/oblivious.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  it("should remove taunt when gained", async () => {
+    game.override.ability(Abilities.OBLIVIOUS)
+      .enemyAbility(Abilities.BALL_FETCH)
+      .moveset(Moves.SKILL_SWAP)
+      .enemyMoveset(Moves.SPLASH),
+
+    await game.classicMode.startBattle([ Species.FEEBAS ]);
+    const enemy = game.scene.getEnemyPokemon();
+    enemy?.addTag(BattlerTagType.TAUNT); // sourceID needs to be defined
```
DayKev, line 43: Is this comment inaccurate?

emdeann: Evidently don't copy paste without enough sleep

SirzBenjie (changes requested):
This PR is structured well and the code is good.
However, I want to firmly push back against adding new ability attributes unless it shows to be absolutely necessary. It seems much better to add a field to the `StatusEffectImmunityAbAttr` and then check if the flag is present instead of having yet another ability attribute.
I think this is a problem because we have so many ability attributes, even for those that are related. It's quite easy to forget to add an ability attribute (which has already been done in several instances and has been the sole source of a couple of bugs).

Then, you can add the applyPostSetStatus method to that class, for instance.

I like the new phase, though. That's good.
I think the mold breaker stuff needs to be handled/initiated in the MoveEndPhase.

emdeann:
Agreed on all the Mold Breaker stuff.

I still prefer to keep the two distinct attributes though - "heal when gained" and "prevent" are two distinctly different things. In this case it's a good thing that you'd have to intentionally add the extra attribute because it should be considered when adding a status prevention ability whether it also activates on summon or not (several do not).  

The other thing is that `StatusEffectImmunityAbAttr` is not PostSummon nor can be made PostSummon, which means it couldn't take advantage of the existing activate on gain infrastructure which it absolutely should be doing.

DayKev:
> In this case it's a good thing that you'd have to intentionally add the extra attribute because it should be considered when adding a status prevention ability whether it also activates on summon or not (several do not).

If it's a required parameter then you still have to do that. It doesn't matter in this case, but it would achieve the same effect (and arguably be better at it, because you would be explicitly asked to define it instead of needing to know about both `AbAttr`s).

emdeann:
Fair enough, if there's a reasonable way to get the abilities to activate on gain without adding an extra attribute then I'm happy doing it that way. Unsure what that would look like though

src/data/ability.ts
```diff
@@ -6893,13 +6969,15 @@ export function initAbilities() {
...
       .attr(StatusEffectImmunityAbAttr, StatusEffect.BURN)
+      .attr(PostSummonHealStatusAbAttr, StatusEffect.BURN)
       .ignorable(),
     new Ability(Abilities.ANGER_SHELL, 9)
       .attr(PostDefendHpGatedStatStageChangeAbAttr, (target, user, move) => move.category !== MoveCategory.STATUS, 0.5, [ Stat.ATK, Stat.SPATK, Stat.SPD ], 1)
       .attr(PostDefendHpGatedStatStageChangeAbAttr, (target, user, move) => move.category !== MoveCategory.STATUS, 0.5, [ Stat.DEF, Stat.SPDEF ], -1)
       .condition(getSheerForceHitDisableAbCondition()),
     new Ability(Abilities.PURIFYING_SALT, 9)
       .attr(StatusEffectImmunityAbAttr)
+      .attr(PostSummonHealStatusAbAttr)
```
Wlowscha, line 6980: According to Bulbapedia "If the Pokémon has a status condition and gains this Ability, the status condition will not be cured and remains in effect." https://bulbapedia.bulbagarden.net/wiki/Purifying_Salt_(Ability)

emdeann: oh yeah you can see in the PR I didn’t even list purifying salt, I just thought I already removed the attribute. Will fix when I’m home or you can

SirzBenjie (approved):

SirzBenjie:
I've approved it because it's in good shape and it's been sitting for a while.
I still think that adding required parameters to the statusImmunity attribute would be a bit better. However, this would also be straightforward to update should the need arise.

Unicornpowerstar (approved):

src/phases/reset-status-phase.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Phase which handles resetting a Pokemon's status to none
+ *
+ * This is necessary to perform in a phase primarly to ensure that the status icon disappears at the correct time in the battle
+ */
+export class ResetStatusPhase extends BattlePhase {
+  private pokemon: Pokemon;
+  private affectConfusion: boolean;
+  private reloadAssets: boolean;
```
DayKev, line 14:
```suggestion
  private readonly pokemon: Pokemon;
  private readonly affectConfusion: boolean;
  private readonly reloadAssets: boolean;
```

src/phases/move-end-phase.ts
```diff
@@ -2,18 +2,34 @@ import { globalScene } from "#app/global-scene";
...
+    this.targets = targets;
+  }
+
   start() {
     super.start();
 
     const pokemon = this.getPokemon();
     if (pokemon.isActive(true)) {
-      pokemon.lapseTags(BattlerTagLapseType.AFTER_MOVE);
+      pokemon.  lapseTags(BattlerTagLapseType.AFTER_MOVE);
```
DayKev, line 22:
```suggestion
      pokemon.lapseTags(BattlerTagLapseType.AFTER_MOVE);
```

src/phases/obtain-status-effect-phase.ts
```diff
@@ -44,6 +46,12 @@ export class ObtainStatusEffectPhase extends PokemonPhase {
               this.sourceText ?? undefined,
             ),
           );
+          if (this.statusEffect && this.statusEffect !== StatusEffect.FAINT) {
```
DayKev, line 49:
```suggestion
          if (!isNullOrUndefined(this.statusEffect) && this.statusEffect !== StatusEffect.FAINT) {
```
Also needs imports updated.

src/phases/move-end-phase.ts
emdeann: lol i was waiting for the time i accidentally committed one of these

SirzBenjie (dismissed):

DayKev (dismissed):

damocleas (approved):

DayKev (approved):

PR #5420, opened 2025-02-26
[Refactor] Refactor moves.ts part 1

SirzBenjie:
N/A

This is part 1 of my moves.ts refactor. I am breaking up this huge file into several smaller files
However, because so many files touch that, I am going to do this incrementally over multiple PRs in order for them to be self-contained.
One big refactor PR is going to be too much.

`src/data/moves.ts` is more than 10k lines. The amount of distinct classes in it is too much. There is no clear seperation of concerns. It initializes moves, defines all of the attributes, has enums, has some interactions.. It's just not reasonable.
These different parts need to be extracted into their own files. Perhaps this will help with the cyclical import issue.

A new `moves` directory has been added to `src/data`, and `src/data/moves.ts` has been moved there.
All of the enums that were defined in `moves.ts` have been moved to their own files in `src/enums`
All of the `invalidXmoves` constants have been extracted to a new `invalid-moves.ts` file.

They are also now `ReadonlySets` because their only use was to check for containment, for which Set is the proper data structure, _not_ an array.

The `Type` enum has been renamed to `PokemonType` because you should never name anything `Type` as this has meaning in a programming context, and is confusing.

All relevant imports and references have been updated to reflect the these changes.

## Screenshots/Videos
N/A

## How to test the changes?
N/A

## Checklist


Fontbane:
The enum file names should follow this-naming-convention rather than CamelCase. Also, some of them could be combined into one file tbh. If we're renaming the `Type` enum the file should be renamed accordingly.

bavalpey:
> The enum file names should follow this-naming-convention rather than CamelCase. Also, some of them could be combined into one file tbh. If we're renaming the `Type` enum the file should be renamed accordingly.

No, we really shouldn't be combining the enums into one file. Files are a free abstraction at build time. They reduce code churn, avoid importing excess code, etc. It's better this way anyway. Only disadvantage is having more different files in the system.

SirzBenjie:
> The enum file names should follow this-naming-convention rather than CamelCase. Also, some of them could be combined into one file tbh. If we're renaming the `Type` enum the file should be renamed accordingly.

No, we really shouldn't be combining the enums into one file. Files are a free abstraction at build time. They reduce code churn, avoid importing excess code, etc. It's better this way anyway. Only disadvantage is having more different files in the system.

DayKev (approved):

Wlowscha (approved):

PR #5426, opened 2025-02-27
[Refactor] Move playerFaints to arena data

Fontbane:

- None

- In #5200, `playerFaints` was tracked oddly. It's stored by the `Arena` class but not by `ArenaData`. This meant it wouldn't save. So instead, `playerFaints` was added straight to `sessionSaveData`, and the `Arena` constructor as well as `BattleScene.newArena()` required an extra parameter just to pass `sessionSaveData.playerFaints`.

- Puts player faint count in `ArenaData` class so it's saved and loaded from there instead of at the root of session data.
- Removes unnecessary parameter from `Arena` constructor and `BattleScene.newArena()`

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1794,8 +1794,8 @@ export default class BattleScene extends SceneBase {
     return this.currentBattle;
   }
 
-  newArena(biome: Biome, playerFaints?: number): Arena {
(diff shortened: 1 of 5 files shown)


Fontbane:
Had a better idea actually

PR #5448, opened 2025-02-28
[Bug] Display Pokémon name in Focus Punch lost focus message

### LINKED ISSUE #5442: [Bug] Focus Punch Failed Message Not Showing Pokemon Name
### Describe the bug

After a pokemon is hit by move while focus punch is prepared, it shows the focus punch failed message as updated in #4696. However, instead of actually saying the pokemon name, it says {{pokemonName}}.

### Reproduction

In a battle against an opposing pokemon, hit them with a move while it has focus punch prepared.

### Expected behavior

The message properly includes the pokemon name. So in the case of my screenshot, "Hariyama lost its focus and couldn't move!"

### Screenshots / Videos

Image

### Session export file

[sessionData2_wrhyme29.prsv.txt](https://github.com/user-attachments/files/19020016/sessionData2_wrhyme29.prsv.txt)

### User data export file

_No response_

### Additional context

_No response_

Wlowscha:
Fixes #5442 , also reported [on Discord](https://discord.com/channels/1125469663833370665/1345085765440503920)

After #5341 , the locales string was generated in the code without passing the pokemonName argument, this is now fixed.

## Screenshots/Videos

https://github.com/user-attachments/assets/8add358c-f674-4a7c-a6ea-61c97553e07d

magikarplostfocus

best_focus_puncher

## How to test the changes?
E.g. with this override:
```
const overrides = {
  MOVESET_OVERRIDE: [ Moves.FOCUS_PUNCH ],
  OPP_MOVESET_OVERRIDE: [ Moves.QUICK_ATTACK ],
  OPP_SPECIES_OVERRIDE:  Species.MAGIKARP,
  OPP_ABILITY_OVERRIDE: Abilities.STURDY,
```

## Checklist

### DIFF
--- a/test/moves/focus_punch.test.ts
+++ b/test/moves/focus_punch.test.ts
@@ -140,6 +140,6 @@ describe("Moves - Focus Punch", () => {
     await game.phaseInterceptor.to("MessagePhase", false);
     const consoleSpy = vi.spyOn(console, "log");
     await game.phaseInterceptor.to("MoveEndPhase", true);
-    expect(consoleSpy).nthCalledWith(1, i18next.t("moveTriggers:lostFocus"));
+    expect(consoleSpy).nthCalledWith(1, i18next.t("moveTriggers:lostFocus", { pokemonName: "Charizard" }));
   });
 });
--- a/src/data/move.ts
+++ b/src/data/move.ts
@@ -9384,7 +9384,7 @@ export function initMoves() {
       .attr(BypassBurnDamageReductionAttr),
     new AttackMove(Moves.FOCUS_PUNCH, Type.FIGHTING, MoveCategory.PHYSICAL, 150, 100, 20, -1, -3, 3)
       .attr(MessageHeaderAttr, (user, move) => i18next.t("moveTriggers:isTighteningFocus", { pokemonName: getPokemonNameWithAffix(user) }))
-      .attr(PreUseInterruptAttr, i18next.t("moveTriggers:lostFocus"), user => !!user.turnData.attacksReceived.find(r => r.damage))
+      .attr(PreUseInterruptAttr, (user, target, move) => i18next.t("moveTriggers:lostFocus", { pokemonName: getPokemonNameWithAffix(user) }), user => !!user.turnData.attacksReceived.find(r => r.damage))
       .punchingMove(),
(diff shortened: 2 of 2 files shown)


SirzBenjie (approved):
Wow, thanks! This was a gross oversight by me.

test/moves/focus_punch.test.ts
```diff
@@ -140,6 +140,6 @@ describe("Moves - Focus Punch", () => {
     await game.phaseInterceptor.to("MessagePhase", false);
     const consoleSpy = vi.spyOn(console, "log");
     await game.phaseInterceptor.to("MoveEndPhase", true);
-    expect(consoleSpy).nthCalledWith(1, i18next.t("moveTriggers:lostFocus"));
+    expect(consoleSpy).nthCalledWith(1, i18next.t("moveTriggers:lostFocus", { pokemonName: "Charizard" }));
```
SirzBenjie, line 143: I don't know if tests are ever run with different locales, but this shouldn't be hardcoded as Charizard? Probably best to stick to the getPokemonWithAffix, or to force locales language perhaps.

Wlowscha: I would say that the test should give the correct result ("Charizard") regardless of what the inner logic of the function that determines the name is, even if it's changed later on. Not sure about locales but it looks like english is always used for tests...?

bavalpey: > I would say that the test should give the correct result ("Charizard") regardless of what the inner logic of the function that determines the name is

This isn't a hill I care about enough, but I think it's worth pointing out that this actually makes the test not a unit test anymore -- instead of testing the focus punch message, we're now additionally testing the functionality of `getPokemonNameWithAffix`.

It's really not a big deal, though. In my test overhaul, I am going to address these sorts of antipatterns anyway, so it's effectively a nonissue.

SirzBenjie: > I would say that the test should give the correct result ("Charizard") regardless of what the inner logic of the function that determines the name is

This isn't a hill I care about enough, but I think it's worth pointing out that this actually makes the test not a unit test anymore -- instead of testing the focus punch message, we're now additionally testing the functionality of getPokemonNameWithAffix.

It's really not a big deal, though. In my test overhaul, I am going to address these sorts of antipatterns anyway, so it's effectively a nonissue.

damocleas (approved):

PR #5476, opened 2025-03-03
[Balance] Make Pickup and Honey Gather Unsuppressable

damocleas:
Pickup and Honey Gather should no longer be suppressed from Ability Suppression Moves/Abilities like Gastro Acid, and Neutralizing Gas from an ally in a double battle.

This was reported directly in the balance channels, and was not the first time I had heard of the issue, but was previously dismissed.

After testing it seems these do not get suppressed by the opposing pokemon at the end of battle correctly, but your ally can still suppress both in a double battle, making you keep (usually support pokemon) benched in the back of the party to get the effect out of it in double battles.

Added the Unsuppressable tags to Pickup and Honey Gather.

These are notably now the only abilities with this tag that do NOT have the Uncopyable and Unswappable tags

## How to test the changes?
- Set a Pokemon on your side of the field in a double battle to neutralizing gas and the other to Pickup or Honey Gather
- Make sure opposing pokemon has an item that can be picked up, no max stacks of anything on your pickup pokemon
- Get to the end of battle with the Pickup / Honey Gatherer still alive, should now activate

## Checklist

### DIFF
--- a/src/data/ability.ts
+++ b/src/data/ability.ts
@@ -6186,7 +6186,8 @@ export function initAbilities() {
       .attr(ProtectStatAbAttr, Stat.ATK)
       .ignorable(),
     new Ability(Abilities.PICKUP, 3)
-      .attr(PostBattleLootAbAttr),
+      .attr(PostBattleLootAbAttr)
+      .attr(UnsuppressableAbilityAbAttr),
     new Ability(Abilities.TRUANT, 3)
       .attr(PostSummonAddBattlerTagAbAttr, BattlerTagType.TRUANT, 1, false),
     new Ability(Abilities.HUSTLE, 3)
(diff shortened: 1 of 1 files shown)


Snailman11:
Does this actually practically change anything? Could you mention exactly what that would be?

DayKev (approved):
Wait were these actually getting suppressed? That's kinda funny lol

damocleas:
> Wait were these actually getting suppressed? That's kinda funny lol

Need to triple check and get a video but I was told that Pickup wasn't working on Mold Breaker Pokemon, added Honey Gather too to prevent cases where it wouldn't make sense to since these both have custom effects that apply at the end of battle.

Either way this still matters for a balance side since this affects double battles where it is your partner Pokemon with either ability, and your other mon has neutralizing gas, and it won't activate.

damocleas:
Updated the description to be more accurate after thorough testing, it seems the reporter was having confirmation bias.

Madmadness65 (approved):

PR #5502, opened 2025-03-11
[BUG] Fix Locked Reroll not using Luck Score

### LINKED ISSUE #4308: [Bug] Formula used to calculate the chance of items upgrading in the shop is not the same with and without Lock Rarities selected
### Describe the bug

When item tier upgrade count is calculated for items initially presented in the shop and whenever the shop is rerolled without Lock Rarities enabled the following code is run (Starting at line 2151 in src/modifier/modifier-type.ts)
```ts
    if (player && tierValue) {
      const partyLuckValue = getPartyLuckValue(party);
      const upgradeOdds = Math.floor(128 / ((partyLuckValue + 4) / 4));
      let upgraded = false;
      do {
        upgraded = Utils.randSeedInt(upgradeOdds) < 4;
        if (upgraded) {
          upgradeCount++;
        }
      } while (upgraded);
    }
```
Whereas when item tier upgrade count is calculated whenever the shop is rerolled with Lock Rarities enabled it instead runs the next else if block which uses the following code to calculate upgrade count (Starting at line 2184 in src/modifier/modifier-type.ts)
```ts
    if (tier < ModifierTier.MASTER) {
      const partyShinyCount = party.filter(p => p.isShiny() && !p.isFainted()).length;
      const upgradeOdds = Math.floor(32 / ((partyShinyCount + 2) / 2));
      while (modifierPool.hasOwnProperty(tier + upgradeCount + 1) && modifierPool[tier + upgradeCount + 1].length) {
        if (!Utils.randSeedInt(upgradeOdds)) {
          upgradeCount++;
        } else {
          break;
        }
      }
      tier += upgradeCount;
    }
```

### Session export file

_No response_

### User data export file

_No response_

### Expected behavior

It seems odd that this is written using essentially two completely different methods. The first method uses the total party luck value that is based on the total value of all your shinies added together whereas the second only takes into account the number of your party pokemon that are shiny. In my opinion this should probably be changed so that both places use the first method so that reroll upgrade chances are the same.

### Screenshots / Videos

_No response_

PC/Windows

### If other please specify

_No response_

### Which browser do you use?

Chrome

### If other please specify

_No response_

### Additional context

After finding this bug while looking at the code I cloned main to my pc and ran it locally to test and make sure I was reading the code correctly below you can see the break points that I put in to see which section of code was running in each case

When rerolling without Lock Rarities only the first breakpoint is hit
image

When rerolling with Lock Rarities only the second breakpoint is hit
image

Scoooom:
When locking your reroll using Lock Capsule, it now correctly uses your Luck Value

It is currently simply counting the number of shiny pokemon you have. I suspect this was an older method of handling luck. This will also fix #4308

Stop using partyShinyCount and instead use getPartyLuckValue.

I will note, I was unsure if using the old upgradeOdds method, or the old one which uses different Math.floor formula.

```
Math.floor(32 / ((partyShinyCount + 2) / 2));
```

VS 

```
Math.floor(128 / ((partyLuckValue + 4) / 4));
```

## Screenshots/Videos
N/A

## How to test the changes?
Check that locked rerolls are giving the proper upgrades. I will note this change may massively affect endless runs where locked rerolls occur most often.

## Checklist


SirzBenjie (approved):

damocleas (approved):

damocleas:
math and all that seems correct, testing is impossible to check without just "feeling lucky"

Madmadness65 (approved):

DayKev (approved):

PR #5344, opened 2025-02-16
[Enhancement] Add form changes for Solgaleo, Lunala, & Marshadow cosmetic forms

Madmadness65:
## What are the changes the user will see?

Solgaleo, Lunala, and Marshadow will be able to form change into their unique cosmetic forms via their signature moves. These are custom form change methods since the Z-Moves these forms are usually attached to are not planned to be implemented.

## Why am I making these changes?

The graphics for these forms were unused, and I wanted to find a reasonable way to implement them, since these forms will not impact gameplay.

## What are the changes from a developer perspective?

Added PokemonForm entries for Solgaleo and Lunala so they can use their unique sprites. Entries were added to the form changes to change form when their signature move is used, Party icons needed to be added for Radiant Sun Solgaleo and Full Moon Lunala. Existing variant palettes exist for both forms, but they do lack party icons for the variants.

## Screenshots/Videos

<details><summary>Marshadow form changing into Zenith form</summary>

https://github.com/user-attachments/assets/6368d918-84c3-4c8a-9e3e-2ee558eb9193

</details>
<details><summary>Solgaleo and Lunala form changing into Radiant Sun and Full Moon forms respectively</summary>

https://github.com/user-attachments/assets/75cf5724-b191-4a11-9b49-45c8830c9ebc

</details>

## How to test the changes?

Use a Solgaleo, Lunala, or Marshadow with Sunsteel Strike, Moongeist Beam, or Spectral Thief respectively. Switch out the Pokemon to see that it reverts back to normal when recalled.

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test`)
  ~- [ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?~
- [x] Have I provided screenshots/videos of the changes (if applicable)?
  ~- [ ] Have I made sure that any UI change works for both UI themes (default and legacy)?~

Are there any localization additions or changes? If so:
- [x] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?
  - [x] If so, please leave a link to it here: https://github.com/pagefaultgames/pokerogue-locales/pull/127
- [x] Has the translation team been contacted for proofreading/translation?

PR #5512, opened 2025-03-12
Revert "[Enhancement] Add form changes for Solgaleo, Lunala, & Marshadow cosmetic forms"

DayKev:
Reverts pagefaultgames/pokerogue#5344


EscaShark:
I left some of the reasoning for this on the previous comment, but will reiterate, and include some things said in contributor-chat about it as well.

In my opinion, it is at best unnecessary and at worst detrimental to add these custom formes, for these reasons:

- They are not actually formes. These Pokémon do not transform in mainline series nor does any dex entry recognize these cosmetic changes as formes (unlike, say, Inactive Xerneas). They are really just part of move animations, no different from Mega Sharpedo with its teeth extended, Iron Valiant with its blade drawn, or Overdrive Zekrom. This might be intentional, but Madness said in the original PR I quote "These are custom form change methods since **the Z-Moves these forms are usually attached to** are not planned to be implemented." This seems to imply that he believes these are actual formes that are simply tied to Z-Moves, erroneously--on both the signature Z-Moves of these mons and the non-Z signature moves, these are only part of a move animation, not an actual forme.
- I recognize that assets for these "formes" exist. However, that does not give us an obligation to use them. If I (attempt to) add a Mega Sharpedo with extended teeth to the sprite database tomorrow, the contributor team is not forced to find a use case for it. There's not even a fallacy of sunk costs since to my knowledge the Radiant Sun, Full Moon and Zenith sprites aren't even made by the Rogue team ourselves. It makes more sense to remove an unnecessary asset than find a confusing, custom use case for it.
- All three "formes" do not have animations, and Zenith does not have shiny variants. It is aesthetically unpleasing to go from an animated sprite to a non-animated one in the middle of a battle. In Zenith's case it's more egregious: transforming would remove your visual shininess. These are in my opinion disqualifying for custom formes meant to be cosmetics.
- It is also aesthetically weird for Lunala to remain in an awkward "charging up Moongeist Beam" forme all the time after using it once. One would expect a purely cosmetic change to at least look good. A possibility to immediately revert once the move animation is finished was brought up, but this would slow gameplay for these three mons down significantly (needing two transformations every time they want to use their signature move) as well as not fixing any of the other concerns.
- unicorn_power's planned animations for Solgaleo and Lunala already refer to these "formes", so even if this is meant as something like a "Hey, we know Radiant Sun Solgaleo exists!", we already have other ways to do so.

Due to these formes not actually existing, the reasoning for adding them seemingly being erroneous to begin with, and multiple complications being brought with implementing them, I believe it is not a good idea to do this and strongly support reverting #5344.

PR #5513, opened 2025-03-12
[Bug] Activate PostSummon Abilities in Speed and Priority Order

emdeann:
At the start of battles, abilities will be activated in speed order instead of simply iterating through the pokemon on the field. The ability of a faster enemy will activate before a slower player pokemon. In addition, abilities with priority activation (Neutralizing Gas, Unnerve, As One, etc.) will activate before or after all others depending on the level of priority. Because there are passives, this means that a Pokemon's two abilities could activate at different times if they have differing levels of priority. 

To match mainline behavior

The implementation of this behavior (after several iterations, lol) is entirely based around dynamic updating and portability to other phase types. My hope is that with the structure set up here it will be easy to do things like dynamically order move phases based on speed changes.

The most important addition is a new abstract class type, `PhasePriorityQueue`. This is the structure for a queue which internally reorders its elements before returning the top when it is popped. The inheriting class provides its implementation for `reorder` with specifics on how to order the elements.

For `PostSummonPhase`s, the subclass is `PostSummonPhasePriorityQueue` which implements reordering based on (1) priority and (2) pokemon effective speed.

`BattlePhase` now stores an array of all priority queues (right now it is of length 1, of course) and a parallel array representing the type of phase to put in each queue. It has also been updated with functions to push/pop from priority queues. Another goal of this implementation is to be noninvasive, i.e. I can naively push a `PostSummonPhase` and it will be processed automatically in order. To this end, `pushPhase` has been updated to redirect to `pushDynamicPhase` if the phase in question is of a dynamic type (again, currently just `PostSummonPhase`).

A byproduct of this implementation is that `conditionalQueue` is now checked entirely during `shiftPhase` instead of just its first element.

The actual activation of phases from the priority queue uses a placeholder in the form of the new `ActivatePriorityQueuePhase`. This is added to `phaseQueue` whenever a dynamic phase is added to a priority queue. When it is activated, it simply tells `BattleScene` to unshift the top phase from the matching priority queue. This way no ordering or activation timing is touched relative to any other phases - whenever a `PostSummonPhase` would have been activated before, the top phase from `PostSummonPhasePriorityQueue` is activated instead. This likewise ensures that the same _number_ of `PostSummonPhase`s are activated as are pushed, so there is no room for error there.

The specific implementation for PostSummon becomes a bit more complex due to the fact that pokemon can have two abilities and each ability may have a different priority. To allow for this, the `PostSummonActivateAbilityPhase` is added as a child of `PostSummonPhase` which serves only to activate the Pokemon's abilities which match its priority level. These are automatically queued by `PostSummonPhasePriorityQueue` when a phase is pushed and are ordered in the same manner (where a regular `PostSummonPhase` has priority 0). They are queued by default after the pokemon's regular `PostSummonPhase`, so due to stable sorting will stay behind it if they also have priority zero. 

Due to the above no ability application happens in the default `PostSummonPhase` anymore aside from Commander.

Ability definitions are modified to account for priority existing. (It also looks like I made a couple random formatting changes to ability stuff, this is either (A) added doc to `applyAbAttrsInternal` because I initially added a parameter there but decided against it or (B) the new linter/formatter)

`getAbilityPriorities` is added to `Pokemon` to return all priorities from the Pokemon's abilities to make life a little easier.

aaaand tests

## Screenshots/Videos

<details><summary>Faster enemy activates first</summary>
<p>

https://github.com/user-attachments/assets/3155bb78-d415-482a-9a37-ce6bb01c6031

</p>
</details> 
<details><summary>Faster player activates first</summary>
<p>

https://github.com/user-attachments/assets/4b2811a7-1f35-4e76-a956-74c0edd333e1

</p>
</details> 
<details><summary>Double battle correct order</summary>
<p>

https://github.com/user-attachments/assets/e150c5a6-ddd0-41f6-bc53-655fee3382ae

</p>
</details> 

_Speeds: Charmander - 12, Zigzagoon - 12, Hoothoot - 11, Bulbasaur - 10_

<details><summary>Slower Pokemon with priority ability activates first</summary>
<p>

https://github.com/user-attachments/assets/50920dd8-b1fe-48c9-809a-da41dcee6192

</p>
</details> 
<details><summary>Normal and passive of the same Pokemon are ordered differently if different priority</summary>
<p>

https://github.com/user-attachments/assets/3efd6769-d4a9-44f6-b5d1-fab95359a024

</p>
</details> 
<details><summary>Priority abilities activate before hazards</summary>
<p>

https://github.com/user-attachments/assets/ff47cd94-ef9f-4bf1-88b7-4361a06102c5

</p>
</details> 

<details><summary>Ability order dynamically updates based on speed changes</summary>
<p>

https://github.com/user-attachments/assets/2127a585-b4d2-4d42-8a61-a787013f1989

_These are all zero priority abilities and the player is faster until Slow Start activates, leading the opponent's ability to activate before the player's passive. (The main ability will activate before the passive if they have the same priority, so this particular interaction couldn't happen if Slow Start was the passive)_

</p>
</details> 

## How to test the changes?
Added automated tests. This is obviously also testable in any situation where multiple abilities are activated at the start of a battle.

## Checklist


SirzBenjie (approved):
So I would have wished for a better data structure than a queue as it sorts itself. However, I realize that the number of elements here are at most 4 pokemon, so the data structure itself shouldn't actually matter that much. Also, the handling with trick room requiring the way the data is sorted to be re-ordered complicates this matter.

This is a much needed rewrite, and is done in a pretty robust way. I believe this priority queue can be reused for move order.

emdeann:
Basically my thoughts exactly Benjie, I’m fully aware this is not an efficient solution but I went for readability and ease of use over that because these structures are going to be holding so few elements

src/field/pokemon.ts
```diff
@@ -2271,6 +2271,10 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     return false;
   }
 
+  public getAbilityPriorities(): number[] {
```
SirzBenjie, line 2274:
```suggestion
  public getAbilityPriorities(): [number, number] {
```

SirzBenjie (dismissed):

src/phases/post-summon-phase.ts
```diff
@@ -25,12 +26,39 @@ export class PostSummonPhase extends PokemonPhase {
...
+
+  public getPriority() {
+    return 0;
+  }
+}
+
+/**
+ * Helper to {@linkcode PostSummonPhase} which applies abilities
+ */
+export class PostSummonActivateAbilityPhase extends PostSummonPhase {
```
DayKev, line 45: Each phase should be in its own file.

src/data/phase-priority-queue.ts
```diff
@@ -0,0 +1,102 @@
...
+function isTrickRoom(): boolean {
+  const speedReversed = new BooleanHolder(false);
+  globalScene.arena.applyTags(TrickRoomTag, false, speedReversed);
+  return speedReversed.value;
+}
+
+/**
+ * Enum representation of the phase types held by implementations of {@linkcode PhasePriorityQueue}
+ */
+export enum DynamicPhaseType {
```
DayKev, line 100: Enums should be in their own file.

SirzBenjie (approved):

src/enums/dynamic-phase-type.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Enum representation of the phase types held by implementations of {@linkcode PhasePriorityQueue}
+ */
+
```
DayKev:
```suggestion
```
Thanks VSCode.

src/phases/post-summon-activate-ability-phase.ts
```diff
@@ -0,0 +1,28 @@
+import type { BattlerIndex } from "#app/battle";
+import { applyPostSummonAbAttrs, PostSummonAbAttr } from "#app/data/ability";
+import { PostSummonPhase } from "#app/phases/post-summon-phase";
+
+/**
+ * Helper to {@linkcode PostSummonPhase} which applies abilities
+ */
+
```
DayKev:
```suggestion
```

damocleas:
bumping this, needs conflicts resolved

SirzBenjie (approved):

DayKev (dismissed):

SirzBenjie (dismissed):

src/data/abilities/ability.ts
```diff
@@ -86,6 +86,7 @@ export class Ability implements Localizable {
   public name: string;
   public description: string;
   public generation: number;
+  public postSummonPriority: number;
```
SirzBenjie: Shouldn't this be `readonly`

DayKev (approved):

PR #5529, opened 2025-03-16
[UI/UX] Grey options in Pokédex for uncaught mons

Wlowscha:

It is now sufficient to have seen a Pokémon to be able to see its stats, moves, abilities, and evolutions. Moreover, a Pokémon is considered "seen" by the Pokédex as soon as its starter has been unlocked, even if the specific evolution or form has not been yet seen in game. In this case, it is also possible to access IVs and natures.

The options that cannot be selected now appear as dark text. Incidentally, this includes the TM moves for mons that cannot learn TMs, such as Ditto.

EDIT: the changes now also affect the visibility of icons in the main Pokédex screen, for consistency.

Note that the changes, in the current form, also affect all starter-selectable forms (e.g., catching a single Unown form is enough to make all 26 visible in the Dex). But again, this should be a desirable change with the passive split.

Similar features have been requested on discord. With the passive split #5531 it becomes essential to be able to check a Pokémon's passive even when it has not yet been caught, so long as the starter has been unlocked.

Changed the menu to a BBCodeText. This is perfectly safe and has never broken anything ever.

Introduced a new `isSeen()` function which returns `true `if the mon has been seen _or_ if the starter is unlocked. Changed the logic for menu options to check the `isSeen `condition. For IVs and natures it is still necessary to check if the starter is actually unlocked, which leads to slightly clunky logic.

## Screenshots/Videos

Pokémon seen, starter not unlocked:
lugia_1
lugia_2

Pokémon not seen (all options are dark and cannot be selected):
arceus_1

Form not seen, but starter is unlocked:
venusaur_1
venusaur_2

Evolution not seen, but starter is unlocked:
flareon_1
flareon_2

Ditto:
ditto_1

## How to test the changes?
Open the Pokédex and check if the menu breaks somehow in any case.

## Checklist


Wlowscha:
It also would be good to grey out the option in those few cases that need it---e.g. mons which don't learn any TMs.

src/ui/pokedex-page-ui-handler.ts
```diff
@@ -696,11 +696,39 @@ export default class PokedexPageUiHandler extends MessageUiHandler {
...
 
     return true;
   }
 
+  getMenuText(): string {
+    const isFormCaught = this.isFormCaught();
+    const isSeen = this.speciesStarterDexEntry?.seenAttr;
+    if (this.species.speciesId === Species.ODDISH) {
+      console.log(isFormCaught, isSeen);
+    }
```
DayKev, line 711:
```suggestion
```
I think you left in some testing code.

Madmadness65 (approved):

damocleas (approved):
works fine in test mashing

PR #5533, opened 2025-03-17
[Bug] [UI/UX] Status moves now play a No Effect Message Against Immune Type Pokemon

### LINKED ISSUE #5084: [Bug] Moves that Inflict Non-Volatile Status Missing Fail Messages
### Describe the bug

Will-O-Wisp, Toxic, Thunder Wave, Spore, Sleep Power, Stun Spore, and Poison Power don't play a failure message for when a Pokémon already possesses a Non-volatile status condition, and when a Pokémon already possesses a specific, matching non-volatile status condition.

### Reproduction

Use Spore against a Poisoned Pokémon.

Use Spore against a Sleeping Pokémon.

### Expected behavior

https://github.com/user-attachments/assets/85dae0a5-2914-4088-ba6a-8e648d5d47b7

https://abcboy101.github.io/poke-corpus/#q=already+paralyzed&s=xcosif3z0hQzcQ

Each Status-inflicting move has its own failure message when used against a Pokemon with a matching Status condition.

"	[VAR 0102(0000)] is already paralyzed!" 

Furthermore, it seems that the message varies with the type of target.

### Screenshots / Videos

No "The wild xxxxx is already paralyzed" when Thunder wave is used against a Paralyzed Pokemon.

https://github.com/user-attachments/assets/ef28f3f0-da2d-49f7-838e-237408b4f7ee

No "But it Failed" when a Status move is used against a Pokémon with a different kind of status condition.

https://github.com/user-attachments/assets/b0996f87-15a3-49d2-bb8a-735020b90fca

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

### LINKED ISSUE #5085: [Bug] Moves that apply Non-Volatile Status do not play a No Effect Message when Used Against an Immune Type.
### Describe the bug

When using Will-O-Wisp, Thunder Wave, Toxic, or Poison Gas against a Pokémon whose type is immune to that Status condition, no "It doesn't affect" message plays.

### Reproduction

Use a non-volatile status move against a Pokémon whose Type is immune to it.

### Expected behavior

"It doesn't affect the xxxxxxxx" should play when a Pokémon is targeted by a status condition move they're invulnerable to.

https://github.com/user-attachments/assets/1d452aee-9a29-46f7-bc0a-19577ad5e157

### Screenshots / Videos

https://github.com/user-attachments/assets/8ea1cd13-bfee-485e-a4b9-15320b6d51fa

### Session export file

_No response_

### User data export file

_No response_

### Additional context

Spore/Powder moves against Grass Types plays a message

https://github.com/user-attachments/assets/af365898-33c5-450f-9b83-089302804ca4

shadownocas:
The user will now be able to observe a "It doesn't affect XXXX" message when status moves like: Will-O-Wisp, Thunder Wave, Toxic, or Poison Gas are used against a Pokemon whose type is immune to that Status condition.

I have decided to implement this change to improve the user experience since they will now be able to get real-time feedback of how certain moves are affecting other Pokemon.

Fixes #5085 
Fixes #5084 

In the file `move.ts`:
Updated `isTypeImmune()` function to receive another parameter `move: Move` and to include the cases in which the status moves listed in the bug are being used against the corresponding immune type pokemon.
In the file `pokemon.ts`:
Updated the function call of `isTypeImmune()` to correspond with the newly added parameter.

## Screenshots/Videos
Before: No "It doesn't affect XXXX" message is displayed.

https://github.com/user-attachments/assets/31b6f8b5-f674-4532-982b-b1442ae692c0

After: "It doesn't affect XXXX" message is correctly displayed.

https://github.com/user-attachments/assets/a8fb9175-f703-4651-ab24-abdb58a36889

## How to test the changes?
To test the changes you can use the `src/overrides.ts` file and use:
For Will-O-Wisp:
  `MOVESET_OVERRIDE: Moves.WILL_O_WISP,`
  `OPP_SPECIES_OVERRIDE: Species.CHARMANDER,`

For Thunder Wave:
  `MOVESET_OVERRIDE: Moves.THUNDER_WAVE,`
  `OPP_SPECIES_OVERRIDE: Species.PIKACHU,`

For Toxic:
  `MOVESET_OVERRIDE: Moves.TOXIC,`
  `OPP_SPECIES_OVERRIDE: Species.GASTLY,`

For Poison Gas:
  `MOVESET_OVERRIDE: Moves.POISON_GAS,`
  `OPP_SPECIES_OVERRIDE: Species.DIALGA,`

## Checklist


src/data/moves/move.ts
```diff
@@ -339,6 +339,26 @@ export default class Move implements Localizable {
...
+      case PokemonType.POISON:
+        if (move.id === Moves.TOXIC) {
+          return true;
+        }
+        break;
+      case PokemonType.STEEL:
+        if (move.id === Moves.POISON_GAS) {
+          return true;
+        }
+        break;
```
DayKev, line 361:
```suggestion
      case PokemonType.POISON:
      case PokemonType.STEEL:
        if ([Moves.TOXIC, Moves.POISON_GAS, Moves.POISON_POWDER].includes(move.id)) {
          return true;
        }
        break;
```
I'm not sure this is the best method though, ideally it would tie in to `Pokemon.canSetStatus()` or something.

shadownocas: Thank you for reviewing my code. I'll change my code to implement what you suggested.

emdeann (commented):
Have to second Kev here - it shouldn't be the move's responsibility to determine something like _why_ a status failed, otherwise we'd have to keep coming back to this code to update cases when new abilities, statuses, etc. get added (even now you're only considering certain abilities - what about Purifying Salt, Water Bubble, and so on?). 

`canSetStatus` is the centralized place for these checks - doing this anywhere but there isn't sustainable. There's already a `quiet` parameter passed there to determine whether messages get sent, so you should be able to use that to queue the message from that function instead.

DayKev:
Yeah something like "if a Status move tried to set a status and failed, display the no effect message". Not sure the best way to do that though due to the way the code is structured, would need to see if there's a good spot to do it cleanly.

shadownocas:
Hi! thank you for replying. I took into consideration what both of you said and i agree that the best approach would prbably be to change the CanSetStatus function to determine when to display the message. That being said i have a proposed fix that involves removing the return false of that function and setting a boolean "isImmune" and in the end of the function i would check the boolean, display the message "It doesn't affect XXXXX" if its true and the quiet parameter is set to false to then finally return the false.

emdeann:
The general idea is correct - you will need to queue the message instead of immediately returning true or false. Just make sure you maintain a couple considerations:

- You don't want to remove all the efficiency from the method (i.e. if you purely use a boolean and return it at the very end, you'll be doing things like checking ability attributes for every run through of the function, even when it isn't necessary. These are much more expensive than the type checks in the switch statement.
- If you're going to display the messages in `canSetStatus` (which you should), you should do it right and cover every case. That means that the proper message should be displayed for an immunity due to misty terrain, electric terrain, safeguard, etc. That would include removing message activations from other places if they exist as well (for example, `StatusEffectAttr` shouldn't be sending the safeguard message for the same reason discussed here earlier)

I'm aware my second point expands the scope of the pr a little more, but I don't think it'll be too bad and I really don't want to have a half-finished `canSetStatus` function - that would just be even more confusing than it is now.

shadownocas:
Thank you for your input! Concerning your first point, i agree and i have in mind another way that wont hinder the efficiency of the base code and will make the code more intuitive. Regarding your second consideration, i will take a look at the code and change it so that the messages regarding status moves are more centralized in the canSetStatus function and all the cases are taken into consideration.

src/field/pokemon.ts
```diff
@@ -5300,6 +5300,15 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
 
+  messageIsImmune(quiet: boolean): boolean {
+    if(!quiet){
+      globalScene.queueMessage(
+        i18next.t("abilityTriggers:moveImmunity", {pokemonNameWithAffix: getPokemonNameWithAffix(this),})
+      );
+    }
+    return false;
+  }
+
```
emdeann, line 5311:
```suggestion
  queueImmuneMessage(quiet: boolean): void {
    if(!quiet) {
      globalScene.queueMessage(
        i18next.t("abilityTriggers:moveImmunity", {pokemonNameWithAffix: getPokemonNameWithAffix(this),})
      );
    }
  }
```
This method shouldn't be used as a shortcut for when you're returning false because that makes its usages less clear. Just return false after calling the method

I would also suggest that you add a parameter to this method for the status attempting to be applied, because the message should be different depending on whether the status being applied is the same status that the pokemon already has.

For example, if your pokemon is already paralyzed and you try to paralyze it, you'll get "{pokemon} is already paralyzed!". If you try to give it a different status then you get the standard immune message.

src/field/pokemon.ts
```diff
@@ -5427,6 +5429,16 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     );
 
     if (cancelled.value) {
+      return this.messageIsImmune(quiet);
```
emdeann, line 5432:
```suggestion
      return false;
```
We should trust the abilities to queue their own messages, that's a standard part of their activation. (this may have not worked before, but it will now because `quiet` is properly being passed as false)

src/field/pokemon.ts
```diff
@@ -5361,30 +5363,30 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
               defType,
             );
             if (cancelImmunity.value) {
-              return false;
+              return this.messageIsImmune(quiet);
```
emdeann, line 5366:
```suggestion
              return false;
```
You're in a `.map` call here, not the greater function. `cancelImmunity` is used to check if the user has corrosion, so the result here would be playing "It doesn't effect {pokemon}" when corrosion successfully lands a poison on a poison/steel type. This we clearly don't want

src/field/pokemon.ts
shadownocas: Alright, my thought process for that was to make the samllest addiciton to the base code. Now i can see how it could make it harder to follow the logic of the function by making those changes.

src/field/pokemon.ts
shadownocas: When i was making the change i actually didn't notice it was inside the mapping function and the tests i ran also didn't show any error. So it was actually my mistake, thank you for pointing it out.

src/field/pokemon.ts
```diff
@@ -5648,6 +5660,20 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     }
 
     if (cancelled.value) {
+      this.messageIsImmune(quiet, effect);
+      return false;
+    }
```
emdeann, line 5665: Should be no message here as described before

src/data/moves/move.ts
```diff
@@ -2446,10 +2446,7 @@ export class StatusEffectAttr extends MoveEffectAttr {
         return false;
       }
```
emdeann, line 2447: This block should be removed as its checked in `canSetStatus` and only in the latter will the message play

src/field/pokemon.ts
```diff
@@ -5516,6 +5517,18 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     );
   }
 
+  messageIsImmune(quiet: boolean, effect?: StatusEffect): void {
```
emdeann, line 5520: I would prefer this method to be named `queueImmuneMessage` so it lines up with how other things are named (and is a bit clearer in what it's doing)

emdeann (commented):
I fixed merge conflicts for you and and added a couple of last minute comments. Looks good after those changes

We will have to iterate on this structure to add messages like the terrain immunities, but that's not something that has to be done immediately. I'd rather get this in asap

src/data/moves/move.ts
shadownocas: Yeah, i was working on this right now. This change was submitted on a previous commit and showed up when i did git rebase.

src/field/pokemon.ts
shadownocas: Alright, i will change it. Thank you for the suggestion.

src/field/pokemon.ts
shadownocas: Alright, removed it

shadownocas:
Thank you for your feedback! I've incorporated the requested changes, please let me know if further revisions are needed.

emdeann (approved):
Looks great!

PR #5446, opened 2025-02-28
[Bug][Hotfix] Restrict Use Candies option in the Pokédex

Wlowscha:
## Why am I making these changes?
It was pointed out that using candies in the Pokédex during the shop phase can be exploited to dupe rewards for catching mons. #5444 

## What are the changes from a developer perspective?
Introducing a `canUseCandies `condition in the dex which is only `true` in whitelisted phases.

## Screenshots/Videos
The option does not show up anymore, even though Pidgey has more than enough candies:
no-candies-for-pidgey

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test`)
  - [ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [x] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (default and legacy)?

PR #5538, opened 2025-03-18
[UI/UX] Option to see Pokedex entry after catching wild Pokémon

Wlowscha:
The Summary/Yes/No menu that shows up after catching a wild Pokémon now also shows a "Pokédex" option, which allows the player to see the Pokédex entry of the newly caught Pokémon.

Note that, thanks to #5446 candies cannot be bought when accessing the Pokédex at this point.

This is a nice qol feature and has been requested by multiple users on discord.

The Pokédex page ui now takes as argument a function which is called after leaving the page. This is necessary to call promptRelease() again, otherwise the confirm menu does not show up. The behavior is replicated from the summary ui handler.

The confirm ui handler now takes an extra argument when called with the "fullParty" option, which is called when selecting the Pokedex option. The indices in args have to be shifted accordingly (this is quite horrible, and overall it would be a good thing to refactor the ui handlers to avoid using these args, but this is beyond the scope of this PR).

## Screenshots/Videos

https://github.com/user-attachments/assets/da467002-81b9-477d-8e4c-5b0b1735ac6c

Works also after Mystery Encounters:

https://github.com/user-attachments/assets/7c92d59b-fb38-4152-9ebd-7c71e73d92b0

## How to test the changes?
Start a run (preferably on a new save) and catch some Pokémon.

## Checklist


damocleas (approved):

src/data/mystery-encounters/utils/encounter-pokemon-utils.ts
```diff
@@ -741,6 +741,26 @@ export async function catchPokemon(
...
+                      pokemon.formIndex,
+                      attributes,
+                      null,
+                      () => {
+                        globalScene.ui.setMode(Mode.MESSAGE).then(() => {
+                          promptRelease();
+                        });
+                      },
+                    );
+                  },
```
emdeann, line 763: Everything else looks good, but I don't like how we're adding identical functions to two different places here. Are we content just making it a little worse now and tackling ME duplication of code as a whole sometime later?

Madmadness65 (approved):

emdeann (approved):

PR #5547, opened 2025-03-21
[Bug][Refactor] Fix Unsuppressable Abilities being Unreplaceable

emdeann:
Changing the ability of a pokemon with Honey Gather/Pickup will no longer fail.

Additionally Flower Gift and Forecast are no longer replaceable in any form. This differs from mainline behavior but... mainline behavior is that skill swap works but wandering spirit doesn't. Makes the most sense to have these be the same as any other form change ability.

After Honey Gather/Pickup were made unsuppressable they were also accidentally unswappable due to `UnsuppressableAbilityAbAttr` being used as "if the ability has this attribute, it can't be replaced". 

I initially just changed ability setting functions to check `UnswappableAbilityAbAttr` instead, but this has edge cases like Neutralizing Gas which is not swappable but _is_ replaceable.

It ends up making more sense to use `copiable`, `suppressable`, and `replaceable` instead of `swappable` (an ability is swappable if it is both copiable and replaceable). I edited ability swapping logic to consider these changes.

I also changed these from being attributes to being fields (essentially flags) on the ability object, added by `.unsuppressable()`, etc. similar to the already existing `.ignorable()`/`.bypassFaint()`/etc
## How to test the changes?
Automated test

## Checklist


SirzBenjie (dismissed):
I really like removing attributes. This is great.

src/data/ability.ts
```diff
@@ -57,6 +57,9 @@ export class Ability implements Localizable {
   public generation: number;
   public isBypassFaint: boolean;
   public isIgnorable: boolean;
+  public isSuppressable: boolean;
+  public isCopiable: boolean;
+  public isReplaceable: boolean;
```
DayKev, line 62:
```suggestion
  public isSuppressable: boolean = true;
  public isCopiable: boolean = true;
  public isReplaceable: boolean = true;
```

src/data/ability.ts
```diff
@@ -68,6 +71,10 @@ export class Ability implements Localizable {
     this.attrs = [];
     this.conditions = [];
 
+    this.isSuppressable = true;
+    this.isCopiable = true;
+    this.isReplaceable = true;
```
DayKev, line 76:
```suggestion
```

src/data/ability.ts
```diff
@@ -68,6 +71,10 @@ export class Ability implements Localizable {
...
+
     this.localize();
   }
```
DayKev, line 80:
```suggestion

  public get isSwappable(): boolean {
    return this.isCopiable && this.isReplaceable;
  }
```

src/data/ability.ts
```diff
@@ -1138,7 +1160,7 @@ export class PostDefendAbilitySwapAbAttr extends PostDefendAbAttr {
 
   override canApplyPostDefend(pokemon: Pokemon, passive: boolean, simulated: boolean, attacker: Pokemon, move: Move, hitResult: HitResult | null, args: any[]): boolean {
     return move.checkFlag(MoveFlags.MAKES_CONTACT, attacker, pokemon)
-      && !attacker.getAbility().hasAttr(UnswappableAbilityAbAttr) && !move.hitsSubstitute(attacker, pokemon);
+      && attacker.getAbility().isReplaceable && attacker.getAbility().isCopiable && !move.hitsSubstitute(attacker, pokemon);
```
DayKev, line 1163:
```suggestion
      && attacker.getAbility().isSwappable && !move.hitsSubstitute(attacker, pokemon);
```

src/data/moves/move.ts
```diff
@@ -7469,7 +7466,7 @@ export class SwitchAbilitiesAttr extends MoveEffectAttr {
   }
 
   getCondition(): MoveConditionFunc {
-    return (user, target, move) => !user.getAbility().hasAttr(UnswappableAbilityAbAttr) && !target.getAbility().hasAttr(UnswappableAbilityAbAttr);
+    return (user, target, move) => [user, target].every(pkmn => pkmn.getAbility().isReplaceable && pkmn.getAbility().isCopiable);
```
DayKev, line 7469:
```suggestion
    return (user, target, move) => [user, target].every(pkmn => pkmn.getAbility().isSwappable);
```

DayKev (commented):
Seems good.

SirzBenjie (approved):

SirzBenjie (approved):

damocleas (approved):

damocleas (approved):

Madmadness65 (approved):

SirzBenjie (approved):

PR #5563, opened 2025-03-27
[Bug] Fix Extra Ability Activation Flyouts during AI Decisions

emdeann:
Abilities won't randomly activate while the AI is making decisions regarding certain moves it has that have secondary effects. This currently only affects Shield Dust which actually won't be displaying anyway, but it's still a good change to make for the future.

Abilities shouldn't display during simulated calls

`getMoveChance` is called by several implementations of `getTargetBenefitScore` which were correctly passing `showAbility` as false, but that parameter wasn't used properly in the function. It was passed as the fifth arg to an ability attribute that only used two args and not used at all in a second ability attribute call. 

I changed the calls to use `!showAbility` as the value to pass for `simulated` so the abilities will know whether to display or not.
## Screenshots/Videos
<p>

https://github.com/user-attachments/assets/05edaffe-bc8a-467f-a852-37006c5bb023

There is an extra activation before the turn, then the expected one when the flinch is prevented.

</p>
</details> 

<p>

https://github.com/user-attachments/assets/f2a94e27-713c-4365-9e2b-94c6c2853db0

No extra activation
</p>
</details> 

## How to test the changes?
Use Shield Dust while the enemy has fake out (one of the moves with the above impl of `getTargetBenefitScore`) and observe it not randomly activating at the start of the turn.

## Checklist

### DIFF
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -1233,7 +1233,7 @@ export class MoveEffectAttr extends MoveAttr {
   getMoveChance(user: Pokemon, target: Pokemon, move: Move, selfEffect?: Boolean, showAbility?: Boolean): number {
     const moveChance = new Utils.NumberHolder(this.effectChanceOverride ?? move.chance);
 
-    applyAbAttrs(MoveEffectChanceMultiplierAbAttr, user, null, false, moveChance, move, target, selfEffect, showAbility);
+    applyAbAttrs(MoveEffectChanceMultiplierAbAttr, user, null, !showAbility, moveChance, move);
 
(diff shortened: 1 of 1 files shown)


Madmadness65 (approved):

DayKev (approved):

Snailman11:
Does this affect Heavy Metal?
https://discord.com/channels/1125469663833370665/1270224310501507094

Snailman11:
Link to Discord's Bug Report (Shield Dust text bug? flyout ai move selection)
https://discord.com/channels/1125469663833370665/1350993312408145961

PR #5589, opened 2025-03-30
[Refactor] Refactor ability file part 1

SirzBenjie:
N/A

`src/data/abilities.ts` is far too large and has strayed beyond being a single-purpose file. It is the source of many cyclic imports. It madly violates the single responsibility principal.

Created a new folder for ability related data, `src/data/abilities`, and updated imports.
Moved `src/data/ability.ts` to that subfolder.
Created a new file meant to hold the universal data lists (right now, just allAbilities) so that files that need access to allAbilities can import it alone instead of having to import abilities.ts (which commonly leads to cyclical imports).

Extracted out the Ability class into its own file, called `ability-class.ts`. (This is part 1 of the ability file refactor. The next parts will extract out things like ability attributes into their own file, and eventually we will have abilities.ts be the same as ability-class.ts).

Extracted out some types that were defined in `ability.ts` to their own file in the `@types` folder.

Part 2 will remove the ability attributes from this file and place them in their own file. (Note, I don't plan on splitting up each attribute into its own file, but rather group related ones together). As a result, we will also be able to remove the initAbilities method into its own file, similar do what Poketernity has done.

## Screenshots/Videos
N/A

## How to test the changes?
N/A

## Checklist


Wlowscha (dismissed):

Wlowscha (approved):

biome.jsonc
```diff
@@ -32,6 +32,7 @@
       // TODO: these files are too big and complex, ignore them until their respective refactors
       "src/data/moves/move.ts",
       "src/data/ability.ts",
```
DayKev, line 34:
```suggestion
```
Well, the file was moved so...

DayKev (approved):

DayKev (approved):

Wlowscha (approved):

PR #4799, opened 2024-11-05
[Item] Add Weather and Terrain Extender Item

xsn34kzx:
> [!NOTE]
> The sprite for the item was included in https://github.com/pagefaultgames/pokerogue/pull/5542.

## What are the changes the user will see?

Adds the functionality for `Mystical Rock`, an item that has the following properties:
- Tier: Ultra
- Max Stack: 2
- Effect: Adds two turns towards weather and terrain moves per stack.
- Spawn Conditions: The item will be pushed to the item reward pool with a weight of 10 if any Pokémon in the party have the following moves and abilities:
  - Moves
    -  Sunny Day
    - Rain Dance
    - Sandstorm
    - Snowscape
    - Hail
    - Chilly Reception
    - Electric Terrain
    - Psychic Terrain
    - Grassy Terrain
    - Misty Terrain
  - Abilities
    - Drought
    - Orichalcum Pulse
    - Drizzle
    - Sand Stream
    - Sand Spit
    - Snow Warning
    - Electric Surge
    - Hadron Engine
    - Psychic Surge
    - Grassy Surge
    - Seed Sower
    - Misty Surge

## Why am I making these changes?

Requested by the @pagefaultgames/balance-team, this original item was made to primarily add the field effect extension effects of several mainline items ([Heat Rock](https://bulbapedia.bulbagarden.net/wiki/Heat_Rock), [Damp Rock](https://bulbapedia.bulbagarden.net/wiki/Damp_Rock), [Smooth Rock](https://bulbapedia.bulbagarden.net/wiki/Smooth_Rock), [Icy Rock](https://bulbapedia.bulbagarden.net/wiki/Icy_Rock), and [Terrain Extender](https://bulbapedia.bulbagarden.net/wiki/Terrain_Extender)) into one item and to curb reward pool bloat while providing extra utility for weather-based and terrain-based strategies in a given run.

## What are the changes from a developer perspective?

The `FieldEffectModifier` class extending `PokemonHeldItemModifier` in `src/modifier/modifier.ts` was written to achieve the effect of the item.

In `src/modifier/modifier-type.ts`, a `MYSTICAL_ROCK` entry was made in the `modifierTypes` array along with an accompanying `WeightedModifierType` entry in `ModifierTier.Ultra` array within the `modifierPool` to essentially provide the spawning conditions explained above.

[...]

PR #5606, opened 2025-04-01
[Bug] Prevent `Mystical Rock` after Max Stack

xsn34kzx:

Players should no longer get `Mystical Rock` from the rewards screen when any applicable Pokémon in the party already has the maximum stack count for the item.

The item should not be pushed when the applicable Pokémon already have the maximum stack count, which makes it "take up" an item reward slot.

A `for` loop is added to the weight function for `Mystical Rock` that goes through each held item of a given Pokémon, checks if it has `Mystical Rock`, and checks if the current stack count is equal to the max stack count. Only if this check is `false` will the weight function continue checking for relevant abilities and moves dictated previously.

## Screenshots/Videos

N/A

## How to test the changes?

Use the following override in `src/overrides.ts`
```ts
const overrides = {
  STARTING_HELD_ITEMS_OVERRIDE: [
    { name: "MYSTICAL_ROCK", count: 2 }
  ]
```
and start a new Classic Mode run with a Pokémon who has a relevant ability or move listed in #4799 (i.e., Groudon). Using `console.log` statements with the weight function within and after the `isHoldingMax` `if` block can be used to see what the end result weight of the item in the rewards pool is (logging after the block means the weight is 0, logging within the block means the weight is 10).

## Checklist
  - ~[ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?~
- ~[ ] Have I provided screenshots/videos of the changes (if applicable)?~
  - ~[ ] Have I made sure that any UI change works for both UI themes (default and legacy)?~

- ~[ ] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?~
  - ~[ ] If so, please leave a link to it here:~ 
- ~[ ] Has the translation team been contacted for proofreading/translation?~

### DIFF
--- a/src/modifier/modifier-type.ts
+++ b/src/modifier/modifier-type.ts
@@ -2823,37 +2823,48 @@ const modifierPool: ModifierPool = {
       modifierTypes.MYSTICAL_ROCK,
       (party: Pokemon[]) => {
         return party.some(p => {
-          const moveset = p.getMoveset(true).map(m => m.moveId);
+          let isHoldingMax = false;
+          for (const i of p.getHeldItems()) {
+            if (i.type.id === "MYSTICAL_ROCK") {
+              isHoldingMax = i.getStackCount() === i.getMaxStackCount();
+              break;
+            }
+          }
 
-          const hasAbility = [
-            Abilities.DRIZZLE,
-            Abilities.ORICHALCUM_PULSE,
-            Abilities.DRIZZLE,
-            Abilities.SAND_STREAM,
-            Abilities.SAND_SPIT,
-            Abilities.SNOW_WARNING,
-            Abilities.ELECTRIC_SURGE,
-            Abilities.HADRON_ENGINE,
-            Abilities.PSYCHIC_SURGE,
-            Abilities.GRASSY_SURGE,
-            Abilities.SEED_SOWER,
-            Abilities.MISTY_SURGE,
-          ].some(a => p.hasAbility(a, false, true));
+          if (!isHoldingMax) {
+            const moveset = p.getMoveset(true).map(m => m.moveId);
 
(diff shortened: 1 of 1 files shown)


xsn34kzx:
Discord Bug Report (Mystical Rock keeps spawning after reaching max stacks in relevant mon)
https://discord.com/channels/1125469663833370665/1356632465220960326

src/modifier/modifier-type.ts
```diff
@@ -2823,37 +2823,48 @@ const modifierPool: ModifierPool = {
...
-            Moves.CHILLY_RECEPTION,
-            Moves.ELECTRIC_TERRAIN,
-            Moves.PSYCHIC_TERRAIN,
-            Moves.GRASSY_TERRAIN,
-            Moves.MISTY_TERRAIN,
-          ].some(m => moveset.includes(m));
-
-          return hasAbility || hasMoves;
+          let isHoldingMax = false;
+          for (const i of p.getHeldItems()) {
```
DayKev, line 2827: Instead of a for loop can't you search directly if a pokemon is holding a specific item? I don't know if it'll matter but late endless this could loop over a ton of items. Then again maybe the performance would be similar for both approaches, not sure.

xsn34kzx: Yeah, I thought about doing that, but I assumed that it probably uses a `for` loop to go over every item anyway and/or uses `instanceof` which has been unreliable in the past with the Orbs. I imagine this is more performant just because it's inline and does the boolean check at the same time, but it's definitely not a huge difference.

DayKev (approved):
Code looks good. 👍

damocleas (approved):

PR #5283, opened 2025-02-09
[Refactor] Remove Promises from moves and abilities

SirzBenjie:
## What are the changes the user will see?
N/A, this is an internal refactor.
Potentially, though, game will be more stable due to fewer dangling promises.

## Why am I making these changes?
Move phase in desperate need of refactor. Lots of dangling promises, and violated assumptions about how effects work.

I also wanted to champion the draft PR from #4887 and talked to InnerThunder about it. They strongly suggested first removing the promises from moves and abilities. So I undertook this task.

This is part of an effort to slowly but surely remove all dangling promises from the codebase. There are countless instances of the attributes from moves and abilities being called as though they are synchronous. In fact, they were not, and this has led to dozens, if not hundreds, of dangling promises.

So, I cherry-picked the sets of commits that did just this from Despair-Games/poketernity/pull/18

## What are the changes from a developer perspective?
There's... a lot. I'm quoting what InnerThunder wrote over on the fork's PR.

> The output of `MoveAttr.apply` as well as `AbAttr.apply` and its offshoots is now strictly boolean, no longer supporting Promises.
> * The functions `applyMoveAttrs` and `applyXAbAttrs` are now synchronous with a `void` return type.
> * Similarly, all modifiers' `apply` functions are now synchronous
> * Several modifier functions in `BattleScene` have also been made synchronous, including
> * `addModifier` and `addEnemyModifier`
> * `tryTransferHeldItemModifier`
> * `updateModifiers`
> * Some phases were created in support of previously async move and ability effects:
>   * `PokemonTransformPhase` (for Transform and Imposter)
>   * `RevivalBlessingPhase` (for Revival Blessing)
>   * `MoveAnimPhase` (currently for Future Sight and Doom Desire, but can be used to synchronously play any move animation in the future)
> * `LoadMoveAnimPhase` to synchronously load animation(s) for a given move (now used for Metronome and Nature Power).

**Note, I had to modify the `CallMoveAttr` when merging the changes, as that was not implemented over in the fork.**

> PS: The commit history is messed up, no way to avoid it... but the important files to review are:
> 
> * `battle-scene.ts` (modifier functions)
> * `data/ability.ts`
> * `data/move.ts`
> * `modifier/modifier.ts`
> * `field/pokemon.ts` (specifically `Pokemon.fuse`)
> * In `phases`:
>   
>   * `move-effect-phase.ts`
>   * `load-move-anim-phase.ts` (new)
>   * `move-anim-phase.ts` (new)
[...]

PR #5618, opened 2025-04-02
[Bug] Fix uses of getAlly()

emdeann:
Less crashes when the ally isn't on the field (one that's been noticed right now is with Plus/Minus)

^

`Pokemon.getAlly()` was not annotated as returning `undefined` even though it does this whenever the ally is undefined. This results in potential crashes in various places where its values was assumed to be defined (these were somewhat hidden behind promises until #5283).

I annotated `getAlly()` properly, then went through and fixed all of its uses to respect the possibility of it being undefined. I am being very careful with this - in `apply` for abilities I'm checking for undefined again even though it would be reasonable to use a bang instead (because `canApply` has already passed), and I'm always using `isNullOrUndefined` instead of evaluating the objects directly. Neither of these things are necessary but I like it best this way

## Screenshots/Videos
Nothing really relevant, I could show plus/minus working but I just pinky promise that they do

## How to test the changes?
Ensure all existing tests run, check abilities and moves that have been modified to ensure that no visible changes exist. The weirdest stuff is probably in the target selection switch statement - it wasn't failing before so this shouldn't be an issue, but take a look to make sure

## Checklist


src/data/ability.ts
```diff
@@ -2866,8 +2867,9 @@ export class CommanderAbAttr extends AbAttr {
     // another Pokemon, this effect cannot apply.
 
     // TODO: Should this work with X + Dondozo fusions?
-    return globalScene.currentBattle?.double && pokemon.getAlly()?.species.speciesId === Species.DONDOZO
-           && !(pokemon.getAlly().isFainted() || pokemon.getAlly().getTag(BattlerTagType.COMMANDED));
+    const ally = pokemon.getAlly();
+    return globalScene.currentBattle?.double && !Utils.isNullOrUndefined(ally) && ally.species.speciesId === Species.DONDOZO
+           && !(ally.isFainted() || ally.getTag(BattlerTagType.COMMANDED));
```
DayKev, line 2872:
```suggestion
    return globalScene.currentBattle?.double
      && ally?.species.speciesId === Species.DONDOZO
      && !(ally.isFainted() || ally.getTag(BattlerTagType.COMMANDED));
```

SirzBenjie (approved):
It's code like this that makes me wish javascript had decorators like python.

LGTM.

src/phases/revival-blessing-phase.ts
```diff
@@ -42,8 +42,12 @@ export class RevivalBlessingPhase extends BattlePhase {
             true,
           );
 
-          if (globalScene.currentBattle.double && globalScene.getPlayerParty().length > 1) {
-            const allyPokemon = this.user.getAlly();
+          const allyPokemon = this.user.getAlly();
+          if (
+            globalScene.currentBattle.double &&
+            globalScene.getPlayerParty().length > 1 &&
+            !Utils.isNullOrUndefined(allyPokemon)
```
SirzBenjie, line 49: This is one of those times where if this ever returns false something seriously went wrong. Lol

SirzBenjie (approved):

DayKev (approved):

src/data/ability.ts
SirzBenjie: Explicitly checking for a double battle here is unnecessary. If the pokemon has an ally, then the battle must be a double battle. This is an invariant for `getAlly` to not return undefined.

SirzBenjie: Also I think the check for `isNullOrUndefined` is better because it's intentinoal. And if dondozo were to somehow be removed from the species array, then we are now checking two undefineds against each other

DayKev: If `Species.DONDOZO` evaluates to `undefined` then the entire game is broken lol.

PR #5623, opened 2025-04-03
[Balance] Fix Depot Agent trainer type lacking Pokémon

Madmadness65:

Fixes the Depot Agent trainer type to have party templates and a species filter (of Ground type) so that the trainer type appears properly when encountered. 

Also removes a stray duplicate Barboach from the Fisherman's rare species pool, as it already exists in his uncommon species pool.

The Depot Agent trainer type is marked to appear in the Metropolis biome as an uncommon trainer, but it was not given any potential Pokémon to appear with, either as a species pool or filter, so now they have one.

Gave the trainer type four potential party templates to pick from, and gave it a species filter of Ground-type Pokémon to use, as that's generally what they use in canon.

## Screenshots/Videos

image

image

## How to test the changes?

Force a battle with the Depot Agent trainer type.

## Checklist
  ~- [ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?~
  ~- [ ] Have I made sure that any UI change works for both UI themes (default and legacy)?~


DayKev (approved):

Fontbane:
Should we just get rid of this trainer type? There's like 2 of them in the whole series, and Metropolis has a lot of trainer types already.

damocleas:
might be in agreement this trainer type isn't really needed, probably

DayKev:
Does this trainer's pokemon pool + type overlap with a different one? Would it make more sense to have those trainer(s) appear here instead?

Madmadness65:
> Does this trainer's pokemon pool + type overlap with a different one? Would it make more sense to have those trainer(s) appear here instead?

The Ground-type species filter does not overlap with any other existing generic trainers currently, so it does fill a specific trainer type niche in this way, even if there isn't much of an original source to work with.

DayKev:
In that case I think it's fine to have this since it increases trainer variety.

Fontbane:
Maybe it can go somewhere else then, but even so I'd rather use other trainer classes. We're missing some staples anyway like Ruin Maniac.

damocleas (approved):

PR #5628, opened 2025-04-04
[Refactor] Add enum for evolution condition

Fontbane:

- None (yet)
- Slightly reduced memory usage

- Exposes more information about the evolution to other files. Makes it easier for other files to check what kind of condition an evolution has. This will be useful, for example, if we want to add a ChallengeType which changes all evolutions with a certain condition, without needing to use instanceof.
- Some of the EvolutionCondition classes were making unnecessary i18next calls and did some odd stuff (like using string literals "day" and "night" instead of the `TimeOfDay` enum)

- Adds the `EvolutionConditionType` enum
- Changes EvolutionCondition constructors to pass this
- Adds `getConditionType` function to the `SpeciesFormEvolution` class
- Removed unnecessary `enforceFunc` from evolutions (was just used for the gender condition)
- Adds a `get` for `description` for evolutions and their conditions
  - This means the descriptions are made as-needed and not constantly loaded.
    - This cuts down on a lot of strings, but they're strings so it's not that much memory. It's still a little.
    - This was needed because the evolutions are initialized before moves (which descriptions rely on for names)

## Checklist

  - [ ] If so, please leave a link to it here: https://github.com/pagefaultgames/pokerogue-locales/pull/160


src/field/pokemon.ts
```diff
@@ -6985,9 +6985,6 @@ export class EnemyPokemon extends Pokemon {
             pe.speciesId === speciesId &&
             (!pe.evoFormKey || pe.evoFormKey === this.getFormKey()),
         );
-        if (evolution?.condition?.enforceFunc) {
-          evolution.condition.enforceFunc(this);
-        }
```
SirzBenjie, line 6990: Can you explain what removing these is doing? Doesn't this change behavior?

SirzBenjie:
This technically adds a circular dependency with the reference to `#allMoves`, but this structure should be moved to its own file along the other `allX` structs.

src/data/balance/pokemon-evolutions.ts
```diff
@@ -140,215 +90,258 @@ export class FusionSpeciesFormEvolution extends SpeciesFormEvolution {
...
-      super(() => globalScene.arena.getTimeOfDay() === TimeOfDay.DUSK || globalScene.arena.getTimeOfDay() === TimeOfDay.NIGHT);
-      this.timesOfDay = [ TimeOfDay.DUSK, TimeOfDay.NIGHT ];
     } else {
-      super(() => false);
-      this.timesOfDay = [];
+      super((() => globalScene.arena.getTimeOfDay() === TimeOfDay.DUSK || globalScene.arena.getTimeOfDay() === TimeOfDay.NIGHT), EvolutionConditionType.NIGHT);
+      this.timesOfDay = [ TimeOfDay.DUSK, TimeOfDay.NIGHT ];
     }
-    this.description = i18next.t("pokemonEvolutions:timeOfDay", { tod: i18next.t(`pokemonEvolutions:${tod}`) });
+  }
```
Xavion3, line 127: Can we just make this function accept a list of time of days? There's already precedent for pokemon that evolve with time of day conditions beyond just day and night.

src/data/balance/pokemon-evolutions.ts
```diff
@@ -140,215 +90,258 @@ export class FusionSpeciesFormEvolution extends SpeciesFormEvolution {
...
     this.amount = amount;
-    this.description = i18next.t("pokemonEvolutions:friendship");
+  }
+  
+  get description(): string {
+    return i18next.t("pokemonEvolutions:friendship");
   }
 }
 
 class FriendshipTimeOfDayEvolutionCondition extends SpeciesEvolutionCondition {
```
Xavion3, line 158: Why does this class exist at all? Can we just remove it?

src/field/pokemon.ts
Fontbane: This is unused, it's meant specifically to enforce that gender condition evos don't change gender when they evolve, which is redundant.

PR #5635, opened 2025-04-06
Fix #2735: Hazard moves incorrectly require targets

### LINKED ISSUE #2735: [BUG] Hazard moves (Spikes etc.) incorrectly require targets
**Describe the bug**

The move Spikes is currently treated as targeting all opposing Pokémon. This matters in Double Battles, or if the target has already knocked itself out before the move is executed.

This was noticed with Spikes, but may apply to other entry hazard moves such as Toxic Spikes, Stealth Rock and Sticky Web.

This has the following implications:

- Using Spikes against two Pokémon may result in two layers of Spikes being set. The message will be displayed multiple times accordingly.
- Spikes fails if there is no target.

**To Reproduce**

1. Ensure a Pokémon in your party knows one of the moves listed above.
2. Enter a Double Battle.
3. Use Spikes (or another similar move) with multiple opposing Pokémon, _or_ have one Pokémon knock out all opposing Pokémon before your other Pokémon executes Spikes in the same turn.
4. Hold V to verify the number of layers of Spikes have been set.

**Expected behavior**

As per the original games, these moves should target the opposing side, not all opposing Pokémon. 

- Spikes should not fail if the opposing Pokémon has already been knocked out.
- Spikes should only set one layer each time the move is used, not two.

**Device**

Firefox Developer Edition v128.0b9 on Windows 11

didas72:
Hazard moves should no longer require targets to successfully execute.

To allow hazard type moves to properly work in double battles after the enemy pokémon fainted, instead of displaying a failure message and not executing.
This Fixes #2735

To implement the fix, 4 source files were changed, accompanied by the addition of a test on another file. The test is detailed in the corresponding section.

The bulk of the logic change is in `move-effect-phase.ts`, while `battle-anim.ts`, `move.ts` and `move-phase.ts` only got supporting changes.

Under the function `start` in `move-effect-phase.ts`, additional logic was added to check for the combination of being a Hazard move AND not having targets. If this condition verifies, the code will use a 'dummy pokémon' as a target, to allow most other functions to remain unchanged while allowing the move to execute. The 'dummy' passed is the move's user. This may sound weird, I will detail why later.
Given that the 'target' pokémon is now in the 'wrong' side of the arena, a somewhat hard-coded check is performed in `move.ts` to revert this side effect.
The remainder of the changes include small verifications to bypass certain checks performed by the existing code, to allow the move to proceed without the expected targets.

Why the 'user' as a dummy pokémon? After several attempts to use more reasonable options, I found that this was the only viable option. For example, creating a new instance of pokémon meant that Sprites were not instantiated. The 'user' pokémon is always correctly setup when the move executes. Which saves a lot of work and leaves the code decluttered, as I would have to set up an entire new pokémon in an unexpected place otherwise.

It is also worth noting that some changes to `move-effect-phase.ts` were formatting changes done by the linter.

## Screenshots/Videos
https://github.com/user-attachments/assets/962d5093-8c7d-4b40-b637-c88bfb8aa3ae

## How to test the changes?
To test manually and observe the behaviour shown in the video, one can set the overrides to include:
```
MOVESET_OVERRIDE: [Moves.SPIKES, Moves.EARTHQUAKE],
OPP_SPECIES_OVERRIDE: Species.DIGLETT,
STARTING_LEVEL_OVERRIDE: 50
```
After that, launch a new save with any two starting pokémons. Instruct the first one to use EARTHQUAKE to faint both enemies and the second one to use SPIKES.

There is also a new automated test, kept under `spikes.test.ts`, named `should work when all targets fainted`.

## Checklist
  - [X] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?

### DIFF
--- a/src/data/battle-anims.ts
+++ b/src/data/battle-anims.ts
@@ -1428,7 +1428,9 @@ export class MoveAnim extends BattleAnim {
   public move: Moves;
 
   constructor(move: Moves, user: Pokemon, target: BattlerIndex, playOnEmptyField = false) {
-    super(user, globalScene.getField()[target], playOnEmptyField);
+    /** Set target to the user pokemon as a dummy if no target is found to avoid crashes */
+    const dummyPokemon = user;
+    super(user, globalScene.getField()[target] ?? dummyPokemon, playOnEmptyField);
 
     this.move = move;
   }
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -5928,7 +5928,7 @@ export class AddArenaTagAttr extends MoveEffectAttr {
     }
 
     if ((move.chance < 0 || move.chance === 100 || user.randSeedInt(100) < move.chance) && user.getLastXMoves(1)[0]?.result === MoveResult.SUCCESS) {
-      const side = (this.selfSideTarget ? user : target).isPlayer() ? ArenaTagSide.PLAYER : ArenaTagSide.ENEMY;
+      const side = ((this.selfSideTarget ? user : target).isPlayer() != (move.hasAttr(AddArenaTrapTagAttr) && target == user)) ? ArenaTagSide.PLAYER : ArenaTagSide.ENEMY;
       globalScene.arena.addTag(this.tagType, this.turnCount, move.id, user.id, side);
       return true;
     }
@@ -5977,7 +5977,7 @@ export class RemoveArenaTagsAttr extends MoveEffectAttr {
 export class AddArenaTrapTagAttr extends AddArenaTagAttr {
   getCondition(): MoveConditionFunc {
     return (user, target, move) => {
-      const side = (this.selfSideTarget ? user : target).isPlayer() ? ArenaTagSide.PLAYER : ArenaTagSide.ENEMY;
+      const side = (this.selfSideTarget != user.isPlayer()) ? ArenaTagSide.ENEMY : ArenaTagSide.PLAYER;
       const tag = globalScene.arena.getTagOnSide(this.tagType, side) as ArenaTrapTag;
       if (!tag) {
         return true;
--- a/src/phases/move-effect-phase.ts
+++ b/src/phases/move-effect-phase.ts
@@ -26,6 +26,7 @@ import {
 } from "#app/data/battler-tags";
 import type { MoveAttr } from "#app/data/moves/move";
 import {
+  AddArenaTrapTagAttr,
   applyFilteredMoveAttrs,
   applyMoveAttrs,
   AttackMove,
@@ -209,12 +210,12 @@ export class MoveEffectPhase extends PokemonPhase {
       targets.some(t => t.hasAbilityWithAttr(ReflectStatusMoveAbAttr) || !!t.getTag(BattlerTagType.MAGIC_COAT));
 
     /**
-     * If no targets are left for the move to hit (FAIL), or the invoked move is non-reflectable, single-target
+     * If no targets are left for the move to hit and it is not a hazard move (FAIL), or the invoked move is non-reflectable, single-target
      * (and not random target) and failed the hit check against its target (MISS), log the move
      * as FAILed or MISSed (depending on the conditions above) and end this phase.
      */
     if (
-      !hasActiveTargets ||
+      (!hasActiveTargets && !move.hasAttr(AddArenaTrapTagAttr)) ||
       (!mayBounce &&
         !move.hasAttr(VariableTargetAttr) &&
         !move.isMultiTarget() &&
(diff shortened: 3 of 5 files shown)


src/data/moves/move.ts
```diff
@@ -5977,7 +5977,7 @@ export class RemoveArenaTagsAttr extends MoveEffectAttr {
 export class AddArenaTrapTagAttr extends AddArenaTagAttr {
   getCondition(): MoveConditionFunc {
     return (user, target, move) => {
-      const side = (this.selfSideTarget ? user : target).isPlayer() ? ArenaTagSide.PLAYER : ArenaTagSide.ENEMY;
+      const side = (this.selfSideTarget != user.isPlayer()) ? ArenaTagSide.ENEMY : ArenaTagSide.PLAYER;
```
DayKev, line 5980:
```suggestion
      const side = (this.selfSideTarget !== user.isPlayer()) ? ArenaTagSide.ENEMY : ArenaTagSide.PLAYER;
```

src/data/moves/move.ts
```diff
@@ -5928,7 +5928,7 @@ export class AddArenaTagAttr extends MoveEffectAttr {
     }
 
     if ((move.chance < 0 || move.chance === 100 || user.randSeedInt(100) < move.chance) && user.getLastXMoves(1)[0]?.result === MoveResult.SUCCESS) {
-      const side = (this.selfSideTarget ? user : target).isPlayer() ? ArenaTagSide.PLAYER : ArenaTagSide.ENEMY;
+      const side = ((this.selfSideTarget ? user : target).isPlayer() != (move.hasAttr(AddArenaTrapTagAttr) && target == user)) ? ArenaTagSide.PLAYER : ArenaTagSide.ENEMY;
```
DayKev, line 5931:
```suggestion
      const side = ((this.selfSideTarget ? user : target).isPlayer() !== (move.hasAttr(AddArenaTrapTagAttr) && target === user)) ? ArenaTagSide.PLAYER : ArenaTagSide.ENEMY;
```

src/data/battle-anims.ts
```diff
@@ -1428,7 +1428,9 @@ export class MoveAnim extends BattleAnim {
   public move: Moves;
 
   constructor(move: Moves, user: Pokemon, target: BattlerIndex, playOnEmptyField = false) {
-    super(user, globalScene.getField()[target], playOnEmptyField);
+    /** Set target to the user pokemon as a dummy if no target is found to avoid crashes */
+    const dummyPokemon = user;
+    super(user, globalScene.getField()[target] ?? dummyPokemon, playOnEmptyField);
```
DayKev, line 1433:
```suggestion
    // Set target to the user pokemon if no target is found to avoid crashes
    super(user, globalScene.getField()[target] ?? user, playOnEmptyField);
```
Since all you're doing is assigning `user` to the `dummyPokemon` variable you can just use `user` directly instead for the same result.
Same with the other time in `move-effect-phase.ts`.

src/phases/move-effect-phase.ts
```diff
@@ -239,18 +240,32 @@ export class MoveEffectPhase extends PokemonPhase {
...
+       * The user pokemon is used as a dummy target, to ensure all code still runs, but it's values are not used.
+       */
+      let dummyPokemon = user;
+
+      // Prevent ENEMY_SIDE targeted moves from occurring twice in double battles
+      // and check which target will magic bounce.
+      // The dummy pokemon is used in case the move is a hazard and there are no targets.
+      const trueTargets: Pokemon[] =
+        !hasActiveTargets && move.hasAttr(AddArenaTrapTagAttr)
+          ? [dummyPokemon]
```
DayKev, line 267:
```suggestion
          ? [user]
```

src/phases/move-effect-phase.ts
```diff
@@ -239,18 +240,32 @@ export class MoveEffectPhase extends PokemonPhase {
...
+
+      /**
+       * In the event that the move is a hazard move, there may be no target and the move should still succeed.
+       * The user pokemon is used as a dummy target, to ensure all code still runs, but it's values are not used.
+       */
+      let dummyPokemon = user;
+
+      // Prevent ENEMY_SIDE targeted moves from occurring twice in double battles
+      // and check which target will magic bounce.
+      // The dummy pokemon is used in case the move is a hazard and there are no targets.
```
DayKev, line 264:
```suggestion
      // In the event that the move is a hazard move, there may be no target and the move should still succeed.
      // In this case, the user is used as the "target" to prevent a crash.
      // This should not affect normal execution of the move otherwise.
```

src/phases/move-effect-phase.ts
```diff
@@ -239,18 +240,32 @@ export class MoveEffectPhase extends PokemonPhase {
...
+      playOnEmptyField,
+    ).play(move.hitsSubstitute(user, this.getFirstTarget()!), () => {
+      /** Has the move successfully hit a target (for damage) yet? */
+      let hasHit = false;
+
+      /**
+       * In the event that the move is a hazard move, there may be no target and the move should still succeed.
+       * The user pokemon is used as a dummy target, to ensure all code still runs, but it's values are not used.
+       */
+      let dummyPokemon = user;
```
DayKev, line 260:
```suggestion
```

src/phases/move-effect-phase.ts
```diff
@@ -239,18 +240,32 @@ export class MoveEffectPhase extends PokemonPhase {
...
-        let hasHit = false;
-
-        // Prevent ENEMY_SIDE targeted moves from occurring twice in double battles
-        // and check which target will magic bounce.
-        const trueTargets: Pokemon[] =
-          move.moveTarget !== MoveTarget.ENEMY_SIDE
+    const playOnEmptyField =
+      (globalScene.currentBattle?.mysteryEncounter?.hasBattleAnimationsWithoutTargets ?? false) ||
+      (!hasActiveTargets && move.hasAttr(AddArenaTrapTagAttr));
+    // Move animation only needs one target, unless it is a hazard move
```
DayKev, line 246:
```suggestion
    // Move animation only needs one target. The attacker is used as a fallback.
```

Madmadness65 (approved):

DayKev (approved):
Seems to work.

SirzBenjie (approved):
Tested this, and it works.

Fantastic.

PR #5656, opened 2025-04-13
Held item refactor

Wlowscha:
## Checklist

### DIFF
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -121,6 +121,7 @@ import { MoveFlags } from "#enums/MoveFlags";
 import { MoveEffectTrigger } from "#enums/MoveEffectTrigger";
 import { MultiHitType } from "#enums/MultiHitType";
 import { invalidAssistMoves, invalidCopycatMoves, invalidMetronomeMoves, invalidMirrorMoveMoves, invalidSleepTalkMoves } from "./invalid-moves";
+import { applyHeldItemAttrs, AttackTypeBoosterHeldItemAttr } from "#app/modifier/held-items";
 
 type MoveConditionFunc = (user: Pokemon, target: Pokemon, move: Move) => boolean;
 type UserMoveConditionFunc = (user: Pokemon, move: Move) => boolean;
@@ -837,7 +838,7 @@ export default class Move implements Localizable {
 
     if (!this.hasAttr(TypelessAttr)) {
       globalScene.arena.applyTags(WeakenMoveTypeTag, simulated, this.type, power);
-      globalScene.applyModifiers(AttackTypeBoosterModifier, source.isPlayer(), source, this.type, power);
+      applyHeldItemAttrs(AttackTypeBoosterHeldItemAttr, source, this.type, power)
     }
 
     if (source.getTag(HelpingHandTag)) {
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -259,6 +259,7 @@ import { MoveFlags } from "#enums/MoveFlags";
 import { timedEventManager } from "#app/global-event-manager";
 import { loadMoveAnimations } from "#app/sprites/pokemon-asset-loader";
 import { ResetStatusPhase } from "#app/phases/reset-status-phase";
+import { allHeldItems, HeldItemType } from "#app/modifier/held-items";
 
 export enum LearnMoveSituation {
   MISC,
@@ -340,6 +341,8 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
   public usedTMs: Moves[];
 
   private shinySparkle: Phaser.GameObjects.Sprite;
+
+  private heldItems: [HeldItemType, number][] = [];
 
   constructor(
     x: number,
@@ -1184,6 +1187,22 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       m => m instanceof PokemonHeldItemModifier && m.pokemonId === this.id,
       this.isPlayer(),
     ) as PokemonHeldItemModifier[];
+  }
+
+  getHeldItems2(): [HeldItemType, number][] {
+    return this.heldItems;
+  }
+
+  addHeldItem(itemType: HeldItemType, stack: number) {
+    const maxStack = allHeldItems[itemType].getMaxStackCount();
+
+    const existing = this.heldItems.find(([type]) => type === itemType);
+
+    if (existing) {
+      existing[1] = Math.min(existing[1] + stack, maxStack);
+    } else {
+      this.heldItems.push([itemType, Math.min(stack, maxStack)]);
+    }
   }
 
   updateScale(): void {
--- a/src/modifier/held-items.ts
+++ b/src/modifier/held-items.ts
@@ -1,0 +1,230 @@
+import type Pokemon from "#app/field/pokemon";
+import { globalScene } from "#app/global-scene";
+import type { Localizable } from "#app/interfaces/locales";
+import type { Constructor, NumberHolder } from "#app/utils";
+import { PokemonType } from "#enums/pokemon-type";
+import i18next from "i18next";
+
+export enum HeldItemType {
+  NONE,
+
+  SITRUS_BERRY = 1,
+  LEPPA_BERRY,
+
+  SILK_SCARF = 101,
+  BLACK_BELT,
+  SHARP_BEAK,
+  POISON_BARB,
+  SOFT_SAND,
+  HARD_STONE,
+  SILVER_POWDER,
+  SPELL_TAG,
+  METAL_COAT,
+  CHARCOAL,
+  MYSTIC_WATER,
+  MIRACLE_SEED,
+  MAGNET,
+  TWISTED_SPOON,
+  NEVER_MELT_ICE,
+  DRAGON_FANG,
+  BLACK_GLASSES,
+  FAIRY_FEATHER,
+}
+
+export class HeldItem implements Localizable {
+  //  public pokemonId: number;
+  public type: HeldItemType;
+  public maxStackCount: number;
+  public isTransferable = true;
+  public isStealable = true;
+  public isSuppressable = true;
+  public attrs: HeldItemAttr[];
+
+  public name: string;
+  public description: string;
+  public icon: string;
+
+  constructor(type: HeldItemType, maxStackCount = 1, name, description, icon) {
+    this.type = type;
+    this.maxStackCount = maxStackCount;
+
+    this.isTransferable = true;
+    this.isStealable = true;
+    this.isSuppressable = true;
+
+    this.name = name;
+    this.description = description;
+    this.icon = icon;
+  }
+
+  //TODO: we might want to change things to make this work... otherwise it's pointless
+  // to derive from Localizable.
+  localize(): void {}
+
+  //  get name(): string {
+  //    return i18next.t(`modifierType:AttackTypeBoosterItem.${AttackTypeBoosterItem[this.moveType]?.toLowerCase()}`);
+  //  }
+
+  //  getDescription(): string {
+  //    return
+  //  }
+
+  untransferable(): HeldItem {
+    this.isTransferable = false;
+    return this;
+  }
+
+  unstealable(): HeldItem {
+    this.isStealable = false;
+    return this;
+  }
+
+  unsuppressable(): HeldItem {
+    this.isSuppressable = false;
+    return this;
+  }
+
+  attr<T extends Constructor<HeldItemAttr>>(AttrType: T, ...args: ConstructorParameters<T>): HeldItem {
+    const attr = new AttrType(...args);
+    this.attrs.push(attr);
+
+    return this;
+  }
+
+  /**
+   * Get all ability attributes that match `attrType`
+   * @param attrType any attribute that extends {@linkcode AbAttr}
+   * @returns Array of attributes that match `attrType`, Empty Array if none match.
+   */
+  getAttrs<T extends HeldItemAttr>(attrType: Constructor<T>): T[] {
+    return this.attrs.filter((a): a is T => a instanceof attrType);
+  }
+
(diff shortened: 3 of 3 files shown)


Wlowscha:
https://github.com/user-attachments/assets/c08b367c-b79a-4db3-bdaf-fefd6dc8e4c4

Documenting progress: HeldItems now fully functional, although still a work in progress.

Instead of rewriting everything, the new HeldItem logic is being coupled with the old ModifierType logic for rewards (which will slowly be phased out).

The video shows the two separate instances of Leftover rewards. The new instance does not yet show up in the summary, but works in game. The old instance does not work anymore. Tests have not yet been updated, so the Leftovers test appropriately fails.

src/enums/held-items.ts
```diff
@@ -0,0 +1,88 @@
...
+type HeldItemValue = typeof HeldItems[HeldItemName];
+
+// Use a type-safe reducer to force number keys and values
+export const HeldItemNames: Record<HeldItemValue, HeldItemName> = Object.entries(HeldItems).reduce(
+  (acc, [key, value]) => {
+    acc[value as HeldItemValue] = key as HeldItemName;
+    return acc;
+  },
+  {} as Record<HeldItemValue, HeldItemName>
+);
```
SirzBenjie, line 88: I really like this.

src/modifier/modifier-type.ts
```diff
@@ -1328,6 +1399,63 @@ export class FusePokemonModifierType extends PokemonModifierType {
...
+      if (pregenArgs && pregenArgs.length === 1 && pregenArgs[0] in PokemonType) {
+        return new AttackTypeBoosterReward(pregenArgs[0] as PokemonType, TYPE_BOOST_ITEM_BOOST_PERCENT);
+      }
+
+      const attackMoveTypes = party.flatMap(p =>
+        p
+          .getMoveset()
+          .map(m => m.getMove())
+          .filter(m => m instanceof AttackMove)
+          .map(m => m.type),
```
SirzBenjie, line 1414: Can you put a TODO comment here that makes it consider the type of variable-type moves / abilities
This would let something like a mon with refrigerate and normal moves allow nevermeltice to generate.

src/modifier/all-held-items.ts
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  unsuppressable(): HeldItem {
+    this.isSuppressable = false;
+    return this;
+  }
+
+  getMaxStackCount(): number {
+    return this.maxStackCount;
+  }
```
SirzBenjie, line 67: Are we able to make this into a getter, or, better yet, have maxStackCount just be a readonly property that doesn't need this method?

I think this is mirroring the existing logic; can we fix it?

src/modifier/all-held-items.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  createIcon(stackCount: number, _forSummary?: boolean): Phaser.GameObjects.Container {
+    const container = globalScene.add.container(0, 0);
+
+    const item = globalScene.add.sprite(0, 12, "items");
+    item.setFrame(this.getIcon());
+    item.setOrigin(0, 0.5);
```
SirzBenjie, line 74: If you merge beta into this branch, which has the fixes to mocks from #5696 you can (and should) use phaser chaining

```ts
const item = globalScene.add.sprite(0, 12, "items").setFrame(this.getIcon()).setOrigin(0, 0.5);
```

src/modifier/all-held-items.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  getIconStackText(stackCount: number): Phaser.GameObjects.BitmapText | null {
+    if (this.getMaxStackCount() === 1) {
+      return null;
+    }
+
+    const text = globalScene.add.bitmapText(10, 15, "item-count", stackCount.toString(), 11);
+    text.letterSpacing = -0.5;
+    if (stackCount >= this.getMaxStackCount()) {
+      text.setTint(0xf89890);
```
SirzBenjie, line 93: I don't like magic numbers. Can we make this a constant somewhere? (Is it not already)?

src/modifier/all-held-items.ts
```diff
@@ -0,0 +1,224 @@
...
+    }
+  }
+}
+
+export const allHeldItems = {};
+
+export function initHeldItems() {
+  // SILK_SCARF, BLACK_BELT, etc...
+  for (const [typeKey, heldItemType] of Object.entries(attackTypeToHeldItem)) {
+    const pokemonType = Number(typeKey) as PokemonType;
```
SirzBenjie, line 219: The call to `Number` here seems fishy. I'm sure there's a way to make typescript not find this necessary.

src/modifier/modifier-type.ts
```diff
@@ -1328,6 +1399,63 @@ export class FusePokemonModifierType extends PokemonModifierType {
...
+          if (attackMoveTypeWeights.get(t)! < 3) {
+            // attackMoveTypeWeights.has(t) was checked before
+            attackMoveTypeWeights.set(t, attackMoveTypeWeights.get(t)! + 1);
+          } else {
+            continue;
+          }
+        } else {
+          attackMoveTypeWeights.set(t, 1);
+        }
+        totalWeight++;
```
SirzBenjie, line 1433:
```suggestion
      for (const t of attackMoveTypes) {
        const weight = attackMoveTypeWeights.get(t) ?? 0;
        if (weight < 3) {
          attackMoveTypeWeights.set(t, weight + 1)
          totalWeight++
        }
```

Shorter way of writing the same thing.

src/modifier/all-held-items.ts
DayKev, line 1: This file seems to have a bunch of different things in it. Imo it should be broken apart somewhat.

Wlowscha:
newHeldItemSummary

Held items now showing up properly in summary :)

src/modifier/modifier-type.ts
Wlowscha: I guess what we would need is a method on Pokemon which takes the move as input and returns its type. Unless more information is needed which is not on the Pokemon... How does Plasma Fists work, for example? Although it would not be relevant here specifically.

src/modifier/all-held-items.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  localize(): void {
+    this.name = this.getName();
+    this.description = this.getDescription();
+    this.icon = this.getIcon();
+  }
+
+  getName(): string {
+    return "";
+  }
```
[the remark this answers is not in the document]
DayKev: There shouldn't be both a `name` field and a `getName` method if `getName` does not return `name`.

Also in that case you can just use a getter (`get name(): string { ... }`) so that `item.name` can be used instead of `item.getName()`.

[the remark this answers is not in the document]
SirzBenjie: Yeah we should use getters

src/modifier/modifier-type.ts
SirzBenjie: I can worry about the implementation, I just want a comment here, as it's a place that wants to know about the move type.

src/modifier/modifier-bar.ts
DayKev, line 1: This is a UI file? It should be in `src/ui/` then.

Wlowscha: That makes sense... I'll move it when I actually get to the point of refactoring it :)

src/items/held-items/attack-type-booster.ts
```diff
@@ -3,8 +3,16 @@ import { PokemonType } from "#enums/pokemon-type";
...
-import { allHeldItems } from "../all-held-items";
+import { HeldItem, ITEM_EFFECT } from "#app/items/held-item";
+
+export interface ATTACK_TYPE_BOOST_PARAMS {
+  /** The pokemon with the item */
+  pokemon: Pokemon;
+  /** The resolved type of the move */
+  moveType: PokemonType;
+  /** Holder for the damage value */
+  movePower: NumberHolder;
```
SirzBenjie, line 15: Doesn't need to be a number holder, a caller can just keep around the reference to the object it uses as params and then read it later on.

Wlowscha:
Changed several things in battle-scene.ts:
- Made a separate addHeldItem function which adds an item and updates modifiers in globalScene (still needs sounds)
- Also made a separate addFormChangeItem, which also incorporates FormChangeHeldItemModifier's obsolete .apply() method
- Removed held-item related logic from addModifier and similar functions
- Removed removePartyMemberModifiers (no need to erase modifiers since they are attached to the pokemon)
- Same applies to clearEnemyHeldItemModifiers
- Refactored tryTransferHeldItem and canTransferHeldItem (thanks to #5932 for much better understanding of the code)

src/items/held-items/item-steal.ts
```diff
@@ -0,0 +1,155 @@
...
+import { HeldItem, ITEM_EFFECT } from "../held-item";
+import { getPokemonNameWithAffix } from "#app/messages";
+import { allHeldItems } from "../all-held-items";
+import { globalScene } from "#app/global-scene";
+
+export interface ITEM_STEAL_PARAMS {
+  /** The pokemon with the item */
+  pokemon: Pokemon;
+  /** The pokemon to steal from (optional) */
+  target?: Pokemon;
```
Bertie690, line 15: Why is this optional?

src/items/held-items/incrementing-stat.ts
```diff
@@ -0,0 +1,61 @@
+import type Pokemon from "#app/field/pokemon";
+import { HeldItem, ITEM_EFFECT } from "../held-item";
+import { Stat } from "#enums/stat";
+import type { NumberHolder } from "#app/utils/common";
+
+export interface INCREMENTING_STAT_PARAMS {
+  /** The pokemon with the item */
+  pokemon: Pokemon;
+  stat: Stat;
+  statHolder: NumberHolder;
```
Bertie690: No need for a number holder if we're embedding it inside an object

src/items/held-items/hit-heal.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Applies {@linkcode HitHealModifier}
+   * @param pokemon The {@linkcode Pokemon} that holds the item
+   * @returns `true` if the {@linkcode Pokemon} was healed
+   */
+  apply(params: HIT_HEAL_PARAMS): boolean {
+    const pokemon = params.pokemon;
+    const stackCount = pokemon.heldItemManager.getStack(this.type);
+    if (pokemon.turnData.totalDamageDealt && !pokemon.isFullHp()) {
```
Bertie690:
```suggestion
    if (pokemon.turnData.totalDamageDealt > 0 && !pokemon.isFullHp()) {
```

src/battle-scene.ts
```diff
@@ -2694,35 +2682,46 @@ export default class BattleScene extends SceneBase {
...
-      }
-      if (!ignoreUpdate) {
-        this.updateModifiers(false, instant);
-        resolve();
-      } else {
-        resolve();
-      }
-    });
+  addEnemyModifier(modifier: PersistentModifier, ignoreUpdate?: boolean) {
+    (modifier as PersistentModifier).add(this.enemyModifiers, false);
```
Bertie690: Unneeded assertion on already persistent modifier

src/battle-scene.ts
```diff
@@ -2619,15 +2618,8 @@ export default class BattleScene extends SceneBase {
     let success = false;
     const soundName = modifier.type.soundName;
     this.validateAchvs(ModifierAchv, modifier);
-    const modifiersToRemove: PersistentModifier[] = [];
     if (modifier instanceof PersistentModifier) {
       if ((modifier as PersistentModifier).add(this.modifiers, !!virtual)) {
```
Bertie690: Unneeded assertion

Bertie690 (changes requested):
I know this is still a WIP, but just wanted to point out a few things that I saw.

Wlowscha:
Comments on the Delibirdy encounter:
- Code for backup option was duplicated four times, it is now a single function which assigns a shell_bell
- List of items excluded from option 3 included vitamins and shuckle juice... which is pointless since they are not transferable anyway. Having removed them, the list is the same as the one from option 2, but I am leaving them separate for now.

src/items/held-item-data-types.ts
SirzBenjie, line 1: `src/items/held-item-data-types.ts` should probably go in `src/@types`

src/modifier/modifier-type.ts
emdeann: Late to the party here lol but `Pokemon#getMoveType` does already exist and considers everything including plasma fists

src/enums/held-item-id.ts
DayKev, line 1: I think the lower 2 bytes were supposed to be the category so that there would be room to increase the max number of items per category (in the unlikely event we need to do so).
```
SITRUS_BERRY: 0x0101,
LUM_BERRY: 0x0201,
ENIGMA_BERRY: 0x0301,
...

REVIVER_SEED: 0x0102,
WHITE_HERB: 0x0202,

SILK_SCARF: 0x0103,
BLACK_BELT: 0x0203,
...
```

src/items/all-held-items.ts
```diff
@@ -0,0 +1,216 @@
...
+import {
+  EvolutionStatBoostHeldItem,
+  SpeciesStatBoostHeldItem,
+  type STAT_BOOST_PARAMS,
+} from "./held-items/stat-booster";
+import { type SURVIVE_CHANCE_PARAMS, SurviveChanceHeldItem } from "./held-items/survive-chance";
+import { type TURN_END_HEAL_PARAMS, TurnEndHealHeldItem } from "./held-items/turn-end-heal";
+import { type TURN_END_STATUS_PARAMS, TurnEndStatusHeldItem } from "./held-items/turn-end-status";
+
+export const allHeldItems: Record<HeldItemId, HeldItem> = {};
```
DayKev: This definition (and others like it) should go in `src/data/data-lists.ts`

PR #5666, opened 2025-04-16
[GitHub] Use `.nvmrc` for pages workflow

DayKev:
N/A

The pages workflow was the only one not using the `.nvmrc` file.

```diff
-         node-version: 22
+         node-version-file: "pokerogue_docs/.nvmrc"
```

## Checklist


DayKev:
Oh even though the tests don't run they're still marked as required... I wonder if there's a way to fix that.

SirzBenjie:
> Oh even though the tests don't run they're still marked as required... I wonder if there's a way to fix that.

Well, honestly they should run if we change the .nvmrc

PR #5677, opened 2025-04-17
[Bug][UI/UX] Fix type hint after enemy disappears

### LINKED ISSUE #4540: [P3] Type hint does not update when `Teleport` is used
### Describe the bug

The enemy Abra used `Teleport` thus not being on the battlefield anymore.
Yet my attacks were still showing type-effectiveness against the pokemon left (ghost is not very effective against normal afaik)

### Reproduction

1. Enforce double fight
2. HAve a pokemon with `Teleport` on the enemy team
3. Have a pokemon with a very-effective move against ONLY the pokemon with teleport
4. Let the pokemon teleport away
5. Type hint will still show very-effective even though the 'very-effective' enemy is not present anymore

### Expected behavior

It should update the type-hints for only the pokemon left on the field

### Screenshots / Videos

image

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

SirzBenjie:
Users with the type hint setting enabled will no longer see improper type hints.

As a side-effect of the fix for this:
Frisk, Download, Anticipation, Forewarn, and imposter will no longer consider an opponent that has fled as a valid target for the ability.
Mini-Black-Hole will no longer be able to steal items from an opponent that has fled.

Fixes #4540 

`pokemon#getOpponents` now takes an optional boolean paramter, `onField`, just like the `isActive` method, that defaults to true.
There were many instances of `getOpponents` being called. I investigated each one and determined if it was appropriate to keep it as the default.

## Screenshots/Videos

<details><summary> After fix</summary>

https://github.com/user-attachments/assets/0f2bf2a6-5680-4e28-9f9a-67d97096a884
</details>

## How to test the changes?
Run the new unit test:
`npm run test:silent test/ui/type-hints.test.ts -- -t "should show the proper hint for a move in doubles after one of the enemy pokemon flees"`

To test manually, use these overrides:

```ts
const overrides = {
  OPP_SPECIES_OVERRIDE: Species.ABRA,
  MOVESET_OVERRIDE: [Moves.SOAK, Moves.FAKE_OUT, Moves.SHADOW_BALL, Moves.THUNDERBOLT],
  OPP_MOVESET_OVERRIDE: [Moves.TELEPORT],
  OPP_ABILITY_OVERRIDE: Abilities.BALL_FETCH,
```

Use soak and fake out on the same abra. The other abra will teleport. Then, enter the "fight" action and notice how shadow ball is no longer given the supereffective type hint.

## Checklist


Madmadness65 (dismissed):

src/data/moves/move.ts
```diff
@@ -8190,7 +8190,7 @@ export type MoveTargetSet = {
 
 export function getMoveTargets(user: Pokemon, move: Moves, replaceTarget?: MoveTarget): MoveTargetSet {
```
DayKev, line 8191: Shouldn't move targeting only consider Pokémon that are on the field?

DayKev:
Uh, Fake Out didn't flinch the Abra...
<p>

https://github.com/user-attachments/assets/96d0197e-a7ed-45dc-9bf3-8d5b30d01ec6

</p>
</details>

DayKev:
Just checked and the same happens on current beta too.

DayKev:
Fake Out works correctly in single battles though...

src/data/moves/move.ts
SirzBenjie: So this was defensive programming. When moves retarget due to another pokemon coming in, it looks at the targets that are in the array - that is, the field positions. When the move actually resolves, it does check if the pokemon is still on the field.

Madmadness65 (approved):

DayKev (approved):

PR #5678, opened 2025-04-17
[Refactor][Move] Refactor move effect phase

SirzBenjie:
Too many to list. Please see #4887 and https://github.com/Despair-Games/poketernity/pull/37

Primarily, though, animations no longer play when a move fails to do any of its effects, mirroring that of mainline.

The move effect phase is complicated, and things that should happen in it are not happening there.

This refactor is very necessary in order to more easily implement some of the remaining unimplemented moves, like dragon darts, without incurring major code debt.

- Renamed `faintPhase`'s `preventEndure` field to `preventInstantRevive`, as it had nothing to do with endure and was only used for reviver seed.
- Moved `pokemon#apply` to the move-effect phase
- Moved the check for applying the `Reflectable` tag into `move#doesFlagEffectApply` and out of the move-effect phase
- Created a new enum, `HitCheckResult` that stores the result of the hit check for the move effect phase
- Refactored the `getAttackDamage` and friends to use the object syntax instead of the huge parameter list.
- `postDefendAbAttr` abilities are no longer responsible for checking whether a substitute was hit; the move effect phase has its own check for substitute and will only attempt to trigger those abilities if the target did not have a substitute
- An animation is only queued if there is at least one successful hit target, or if there is a post apply effect of the move (e.g. for self-destruct)
- The move effect phase constructor is no longer passed an instance of `pokemonMove`. Instead, it is passed the `move` itself
  - The `move` field of the move effect phase has accordingly had its type changed and call sites updated
- Added `pokemon#getMoveCategory`, which takes the logic previously done by `pokemon#apply` for determining whether a move is physical or special
- Added `pokemon#getCriticalHitResult` which extracted the same logic out of the `apply` 
- Moved lots of the logic in `moveEffectPhase#start` into distinct, purposeful functions

## Screenshots/Videos

<details><summary> Flamethrower animation not playing against flash fire, but other animations still playing</summary>

https://github.com/user-attachments/assets/85c82939-b243-4580-8545-488c9d158024
</details>

## How to test the changes?
To test move animations properly not displaying on ability immunity:
```ts
const overrides = {
  OPP_ABILITY_OVERRIDE: Abilities.FLASH_FIRE,
  MOVESET_OVERRIDE: [Moves.FLAMETHROWER],
```

Otherwise, there's nothing specific to test other than making sure moves / abilities work as you expect them.

## Checklist


Madmadness65 (approved):

Madmadness65 (dismissed):

src/constants.ts
```diff
@@ -9,3 +9,8 @@ export const SESSION_ID_COOKIE_NAME: string = "pokerogue_sessionId";
 
 /** Max value for an integer attribute in {@linkcode SystemSaveData} */
 export const MAX_INT_ATTR_VALUE = 0x80000000;
+
+/** The min and max waves for mystery encounters to spawn in classic mode */
+export const CLASSIC_MODE_MYSTERY_ENCOUNTER_WAVES: [number, number] = [10, 180] as const;
+/** The min and max waves for mystery encounters to spawn in challenge mode */
+export const CHALLENGE_MODE_MYSTERY_ENCOUNTER_WAVES: [number, number] = [10, 180] as const;
```
damocleas, line 16: Would these not be 11 and 179?

damocleas (approved):

emdeann (approved):
I don't have the time to go through this with a super fine tooth comb but generally looks good, changes are great and I haven't run into any new bugs while testing. Would like to get in asap for testing

test/abilities/galvanize.test.ts
```diff
@@ -54,21 +53,23 @@ describe("Abilities - Galvanize", () => {
     await game.phaseInterceptor.to("BerryPhase", false);
 
     expect(playerPokemon.getMoveType).toHaveLastReturnedWith(PokemonType.ELECTRIC);
-    expect(enemyPokemon.apply).toHaveReturnedWith(HitResult.EFFECTIVE);
+    expect(spy).toHaveReturnedWith(1);
```
DayKev, line 56: Can't this be `HitCheckResult.HIT` instead of `1`?

src/constants.ts
DayKev: True, but that would require investigating the ME code to make sure it's using the proper comparison operators before changing this.

DayKev (approved):

Madmadness65 (approved):

src/constants.ts
SirzBenjie: this just reloacted the existing content. I didn't do this. If needed to change, would be a different PR.

src/phases/move-effect-phase.ts
```diff
@@ -115,23 +261,24 @@ export class MoveEffectPhase extends PokemonPhase {
...
+      }
+      if (!user.scene) {
+        /*
+         * This happens if the Pokemon that used the delayed attack gets caught and released
+         * on the turn the attack would have triggered. Having access to the global scene
+         * in the future may solve this entirely, so for now we just cancel the hit
+         */
+        super.end();
+        return;
+      }
```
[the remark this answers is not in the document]
SirzBenjie: I don't think that anything that I modified in this PR would made whatever required doing that no longer necessary.
I also don't want to make another change for this to have it get re-approved.
We can put that sort of trim in another pr.

test/abilities/galvanize.test.ts
SirzBenjie: Yes, will patch in a follow-up PR,

src/phases/move-effect-phase.ts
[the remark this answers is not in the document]
DayKev: Yeah, just letting you know (that's why I approved the PR still).

PR #5682, opened 2025-04-19
[Docs] Added changes to markdown files + comment fixes

Bertie690:
N/A
Split up stuff from prior PR, plus i like good docs
Test boilerplate changes have been moved to the kebab case pr because that one makes more sense for it than I

ReloadHelper now helpfully indicates a reloaded session in the console log

## Screenshots/Videos
N/A
## How to test the changes?
N/A

## Checklist


Madmadness65 (approved):

SirzBenjie:
I added a commit that fixed up the comments.md file which was actually incorrect. Incorrect synatx was being used.
We are using `tsdoc`, which is distinct from `jsdoc`. It's important to note that `tsdoc` is distinct from jsdoc with unique syntax, and some things do not render properly.
VSCode even struggles to get TSDoc to render properly.

Anyway, you need to be adding a `-` after the name of a parameter.

SirzBenjie (dismissed):

damocleas:
needs conflict resolved

test/testUtils/gameManager.ts
```diff
@@ -512,8 +517,9 @@ export default class GameManager {
   }
 
   /**
-   * Command an in-battle switch to another Pokemon via the main battle menu.
-   * @param pokemonIndex the index of the pokemon in your party to switch to
+   * Command an in-battle switch to another {@linkcode Pokemon} via the main battle menu.
+   * @param pokemonIndex the 0-indexed position of the party pokemon to switch to.
+   * Should generally never be called with 0 as that will just select your current active pokemon.
```
DayKev, line 522:
```suggestion
   * @param pokemonIndex - The 0-indexed position of the party pokemon to switch to.
   * Should never be called with 0 as that will select the currently active pokemon.
```

test/testUtils/gameManager.ts
```diff
@@ -540,7 +546,7 @@ export default class GameManager {
    * of the party UI, where you just need to navigate to a party slot and press
    * Action twice - navigating any menus that come up after you select a party member
    * is not supported.
-   * @param slot the index of the pokemon in your party to switch to
+   * @param slot the 0-indexed position of the pokemon in your party to switch to
```
DayKev, line 549:
```suggestion
   * @param slot 0 The 0-indexed position of the pokemon in your party to switch to
```

test/testUtils/gameManager.ts
```diff
@@ -575,7 +581,8 @@ export default class GameManager {
   /**
    * Intercepts `TurnStartPhase` and mocks {@linkcode TurnStartPhase.getSpeedOrder}'s return value.
    * Used to manually modify Pokemon turn order.
-   * Note: This *DOES NOT* account for priority, only speed.
+
+   * Note: This *DOES NOT* account for priority.
    * @param {BattlerIndex[]} order The turn order to set
```
DayKev, line 586:
```suggestion
   * @param order - The turn order to set
```

Madmadness65 (dismissed):

test/testUtils/gameManager.ts
```diff
@@ -439,8 +443,8 @@ export default class GameManager {
   }
 
   /**
-   * Checks if the player has won the battle.
-   * @returns True if the player has won, otherwise false.
+   * Check if the player has won the battle.
+   * @returns whether the player has won the battle (all opposinc Pokemon have been fainted)
```
DayKev, line 447:
```suggestion
   * @returns whether the player has won the battle (all opposing Pokemon have been fainted)
```

test/testUtils/gameManager.ts
```diff
@@ -458,8 +462,8 @@ export default class GameManager {
 
   /**
    * Checks if the current mode matches the target mode.
-   * @param mode - The target mode.
-   * @returns True if the current mode matches the target mode, otherwise false.
+   * @param mode - The target {@linkcode Mode} to check.
```
DayKev, line 465:
```suggestion
   * @param mode - The target {@linkcode UiMode} to check.
```

SirzBenjie (approved):

create-test-boilerplate.js
```diff
@@ -72,14 +72,15 @@ async function runInteractive() {
 
   const type = typeAnswer.selectedOption.toLowerCase();
   // Convert fileName from kebab-case or camelCase to snake_case
```
emdeann, line 74:
```suggestion
  // Convert fileName to kebab-case
```

create-test-boilerplate.js
```diff
@@ -72,14 +72,15 @@ async function runInteractive() {
...
+  // Ex: "Cud Chew/Cud-Chew" --> "cud-chew"
   const fileName = fileNameAnswer.userInput
-    .replace(/-+/g, "_") // Convert kebab-case (dashes) to underscores
     .replace(/([a-z])([A-Z])/g, "$1_$2") // Convert camelCase to snake_case
```
emdeann, line 77:
```suggestion
    .replace(/([a-z])([A-Z])/g, "$1-$2") // Convert camelCase to kebab-case
```

create-test-boilerplate.js
```diff
@@ -72,14 +72,15 @@ async function runInteractive() {
...
-    .replace(/\s+/g, "_") // Replace spaces with underscores
+    .replace(/\s+/g, "-") // Replace spaces with underscores
```
emdeann, line 78:
```suggestion
    .replace(/\s+/g, "-") // Replace spaces with dashes
```

create-test-boilerplate.js
```diff
@@ -130,9 +131,9 @@ describe("${description}", () => {
   beforeEach(() => {
     game = new GameManager(phaserGame);
     game.override
-      .moveset([ Moves.SPLASH ])
-      .ability(Abilities.BALL_FETCH)
-      .battleType("single")
+      .moveset(Moves.${type === "move" ? attrName : "SPLASH"})
+      .ability(Abilities.${type === "ability" ? attrName : "BALL_FETCH"})
```
emdeann, line 135: I don't actually love this. You're not always going to name the test after a specific move or ability, and even if you are you're not always going to want the player pokemon to be overridden with it by default. more/less personal preference though

create-test-boilerplate.js
SirzBenjie: Welp we'll have to rename our test files then lol. We've been using underscores for tests ever since this was written.

create-test-boilerplate.js
SirzBenjie: I agree, yeah.

Bertie690: A few cents:
1. The way I see it, the code is a starting point that is expected to be changed. While you may not always _want_ the player to have the ability/move, the same argument could be applied to disabling crits or defaulting to single battles.
2. I've seen plenty of tests that keep the starting placeholder overrides with ball fetch and splash... and then override the exact same stuff in each individual test case (see the water bubble test for example). I know it's probably a matter of personal preference, but it just doesn't look nice to me personally to see code be literally useless like that. (Then again, IDK what could be done to encourage people short of adding a big "Change Me!" comment or similar)
At the end of the day I don't really care too much about it so won't be too hung up if it gets reverted

create-test-boilerplate.js
Bertie690: That will be for the test fixes PR my friend

create-test-boilerplate.js
DayKev: I think it's fine to do it this way, this would likely reduce the need to edit these values (though there's always tests that need to do something different regardless of which way it's done).

emdeann: my opinion remains that I'd rather anything that impacts the test (i.e. not a placeholder like ball fetch) have to be manually set to avoid having some rogue override someone doesn't notice make a test nondeterministic. The other defaults make sense in that there is basically no scenario where my test is going to fail because I didn't realize crits were off or whatever, but it's entirely possible that I just don't realize an ability is set by default and that has a chance to mess with something

emdeann: but this is really splitting hairs here

SirzBenjie (approved):

docs/linting.md
```diff
@@ -1,34 +1,65 @@
-# Biome
+# Linting & Formatting
 
-## Key Features
+> "Any fool can write code that a computer can understand. Good programmers write code that humans can understand."
```
Bertie690, line 3: this should prob go in the comments file actually lol

Madmadness65 (approved):

SirzBenjie (approved):

Bertie690:
hey benjie can you wait to merge it?
I want to find a better quote for the formatting doc and move the existing one to the comments doc (where it was supposed to be)

i may or may not have just realized I had hit ctrl+v in the wrong file lol

PR #5693, opened 2025-04-23
[Refactor][UI/UX] Cleanup battle-info ui code

SirzBenjie:
N/A

There are mumblings of wanting to change parts of the battle ui. I took a look at the ui file to try to get an understanding of what is going on there, and it's in quite poor shape.

Created a new directroy, `src/ui/battle-info`.
Moved the old `battle-info.ts` file into it.
Converted `BattleInfo` into an abstract class, as it should never be instantiated itself.

I did not change any code logic, not even stuff that looked fishy like this. This was entirely just moving things around and cleaning up poor syntax to replace with equivalent code:

I didn't even clean up suspicious logic, such as this:
```ts
const baseVariant = !doubleShiny ? pokemon.getVariant(true) : pokemon.variant;
```
As that is not the point of this PR

- Made it so all methods now pass the `cognitiveComplexity` biome lint, by splitting up methods into smaller methods.

- Made it so that all of the phaser methods that return `this` for chaining actually use chaining. This shortens the code and makes it clearer what is going on.

- Moved methods into the `PlayerBattleInfo` and `EnemyBattleInfo` class into their own files.
  - Any method of `BattleInfo` class that had functionality surrounded by ``(if !playerInfo)`` (or not negated) had that logic moved to an overridden method in the respective class that calls the base class's method before doing that logic.

- Removed `updateEffectiveness` from `Pokemon` and only into `EnemyPokemon` class, as we never look at effectiveness for the player pokemon.

- Created a new function in `utils/common.ts` (I didn't know a better spot to put this) for getting the shiny descriptor for a variant.

Previously, there was this nasty method contributing ~12 points of cognitive complexity:
```ts
doubleShiny || baseVariant ? `${baseVariant === 2 ? i18next.t("common:epicShiny") : baseVariant === 1 ? i18next.t("common:rareShiny") : i18next.t("common:commonShiny")}${doubleShiny ? `/${pokemon.fusionVariant === 2 ? i18next.t("common:epicShiny") : pokemon.fusionVariant === 1 ? i18next.t("common:rareShiny") : i18next.t("common:commonShiny")}` : ""}`
```

I updated `pokemon-info-container`, which also had the exact same pattern, to use this method. That's why that file was touched here as well.

Instead of that long ternary chain, I just made a short and sweet method that resolves to the i18n key for the variant.
## Screenshots/Videos
N/A

## How to test the changes?
N/A

## Checklist


SirzBenjie:
Everything broke when I tried to do this. It's dubious errors. I'm giving up on refactoring UI code.
Maybe this can be re-opened...

PR #5696, opened 2025-04-23
[Refactor][UI/UX] Cleanup battle-info ui code

SirzBenjie:
N/A

There are mumblings of wanting to change parts of the battle ui. I took a look at the ui file to try to get an understanding of what is going on there, and it's in quite poor shape.

Created a new directroy, `src/ui/battle-info`.
Moved the old `battle-info.ts` file into it.
Converted `BattleInfo` into an abstract class, as it should never be instantiated itself.

I did not change any code logic. This was entirely just moving things around and cleaning up poor syntax to replace with equivalent code:

I didn't even clean up suspicious logic, such as this:
```ts
const baseVariant = !doubleShiny ? pokemon.getVariant(true) : pokemon.variant;
```
As that is not the point of this PR

- Made it so all methods now pass the `cognitiveComplexity` biome lint, by splitting up methods into smaller methods.

- Made it so that all of the phaser methods that return `this` for chaining actually use chaining. This shortens the code and makes it clearer what is going on.

- Moved methods into the `PlayerBattleInfo` and `EnemyBattleInfo` class into their own files.
  - Any method of `BattleInfo` class that had functionality surrounded by ``(if !playerInfo)`` (or not negated) had that logic moved to an overridden method in the respective class that calls the base class's method before doing that logic.

- Removed `updateEffectiveness` from `Pokemon` and only into `EnemyPokemon` class, as we never look at effectiveness for the player pokemon.

- Created a new function in `utils/common.ts` (I didn't know a better spot to put this) for getting the shiny descriptor for a variant.
- Modified all of the `GameObject` mocks in `test/testUtils/mocksContainer` to properly return the `this` parameter as required for certain methods. Also added a few mocked methods that were missing and necessary.

Previously, there was this nasty method contributing ~12 points of cognitive complexity:
```ts
doubleShiny || baseVariant ? `${baseVariant === 2 ? i18next.t("common:epicShiny") : baseVariant === 1 ? i18next.t("common:rareShiny") : i18next.t("common:commonShiny")}${doubleShiny ? `/${pokemon.fusionVariant === 2 ? i18next.t("common:epicShiny") : pokemon.fusionVariant === 1 ? i18next.t("common:rareShiny") : i18next.t("common:commonShiny")}` : ""}`
```

I updated `pokemon-info-container`, which also had the exact same pattern, to use this method. That's why that file was touched here as well.

Instead of that long ternary chain, I just made a short and sweet method that resolves to the i18n key for the variant.
## Screenshots/Videos
N/A

## How to test the changes?
N/A

## Checklist

### DIFF
--- a/src/ui/battle-info/battle-info.ts
+++ b/src/ui/battle-info/battle-info.ts
@@ -43,7 +43,6 @@ export default abstract class BattleInfo extends Phaser.GameObjects.Container {
   protected levelContainer: Phaser.GameObjects.Container;
   protected hpBar: Phaser.GameObjects.Image;
   protected levelNumbersContainer: Phaser.GameObjects.Container;
-  protected hpNumbersContainer: Phaser.GameObjects.Container;
   protected type1Icon: Phaser.GameObjects.Sprite;
   protected type2Icon: Phaser.GameObjects.Sprite;
   protected type3Icon: Phaser.GameObjects.Sprite;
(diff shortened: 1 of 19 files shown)


DayKev:
<details><summary>Beta 84a2ce9</summary>
<p>

image

</p>
</details> 
<details><summary>This PR ce0a166</summary>
<p>

image

</p>
</details>

src/utils/common.ts
```diff
@@ -566,3 +567,18 @@ export function animationFileName(move: Moves): string {
...
+ * @returns The localized shiny descriptor
+ */
+export function getShinyDescriptor(variant: Variant): string {
+  switch (variant) {
+    case 2:
+      return i18next.t("common:epicShiny");
+    case 1:
+      return i18next.t("common:shiny");
+    case 0:
+      return i18next.t("common:normal");
```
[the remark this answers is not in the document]
Wlowscha: Also, note that the same logic for the hover description on the shiny icon is present in SummaryUiHandler as well. So the new function should also be used here. Since the logic is apparently the same in all three cases, maybe it would be best to have a common function that directly returns the hover text.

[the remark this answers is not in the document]
Wlowscha: shiny-rare-wrong

Just to show what goes wrong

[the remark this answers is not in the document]
SirzBenjie: > Since the logic is apparently the same in all three cases, maybe it would be best to have a common function that directly returns the hover text.

Agreed yeah. Another PR, though.

src/utils/common.ts
```diff
@@ -576,3 +577,18 @@ export function animationFileName(move: Moves): string {
...
+ * @returns The localized shiny descriptor
+ */
+export function getShinyDescriptor(variant: Variant): string {
+  switch (variant) {
+    case 2:
+      return i18next.t("common:epicShiny");
+    case 1:
+      return i18next.t("common:rareShiny");
+    case 0:
+      return i18next.t("common:commonShiny");
```
xsn34kzx, line 592: Is there a precedent/preference for descending or ascending order on `switch` blocks?

src/ui/fight-ui-handler.ts
```diff
@@ -391,7 +391,7 @@ export default class FightUiHandler extends UiHandler implements InfoToggle {
 
     const opponents = (globalScene.getCurrentPhase() as CommandPhase).getPokemon().getOpponents();
     opponents.forEach(opponent => {
-      opponent.updateEffectiveness(undefined);
+      (opponent as EnemyPokemon).updateEffectiveness(undefined);
```
xsn34kzx, line 394:
```suggestion
      (opponent as EnemyPokemon).updateEffectiveness();
```

src/ui/battle-info/enemy-battle-info.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  // #region Type effectiveness hint objects
+  protected effectivenessContainer: Phaser.GameObjects.Container;
+  protected effectivenessWindow: Phaser.GameObjects.NineSlice;
+  protected effectivenessText: Phaser.GameObjects.Text;
+  protected currentEffectiveness?: string;
+  // #endregion
+
+  override get statOrder(): Stat[] {
+    return [Stat.HP, Stat.ATK, Stat.DEF, Stat.SPATK, Stat.SPDEF, Stat.ACC, Stat.EVA, Stat.SPD];
```
xsn34kzx, line 27: Would it break anything to make this array an immutable, static class field in this and the `PlayerBattleInfo` class?

xsn34kzx (commented):
Super minor suggestions and questions since I've been out of the game for a minute.

src/ui/fight-ui-handler.ts
```diff
@@ -279,7 +279,7 @@ export default class FightUiHandler extends UiHandler implements InfoToggle {
       this.moveInfoOverlay.show(pokemonMove.getMove());
 
       pokemon.getOpponents().forEach(opponent => {
-        opponent.updateEffectiveness(this.getEffectivenessText(pokemon, opponent, pokemonMove));
+        (opponent as EnemyPokemon).updateEffectiveness(this.getEffectivenessText(pokemon, opponent, pokemonMove));
```
emdeann, line 282: Given that we're assuming here that `pokemon` is a `PlayerPokemon` and its opponents are `EnemyPokemon`, I would prefer this to look like `PlayerPokemon.getOpponents()` returning an array of `EnemyPokemon` instead of having to do this type assertion. but that is debatably at best in scope here

src/ui/battle-info/enemy-battle-info.ts
SirzBenjie: I wanted to make it an immutable static class field.
However, we can't actually make it static _and_ have this exist in the base battle info class. And all battle info classes need to have a stat order.

I played around with it and it was really cumbersome.
Plus, this achieves the same effect.

src/utils/common.ts
SirzBenjie: I haven't seen a preference one way or the other in the codebase.

src/ui/battle-info/enemy-battle-info.ts
xsn34kzx: Ah, makes sense then.

Wlowscha (approved):

src/ui/fight-ui-handler.ts
SirzBenjie: This makes sense, but I feel as though it should be in another PR. This one is bloated as is.

Madmadness65 (approved):

DayKev:
Is there any way to manually generate a diff between the deleted `battle-info.ts` file and the new files? I want to just evaluate the differences and not comment on all the other nonsense that wasn't touched.

src/ui/battle-info/battle-info.ts
```diff
@@ -0,0 +1,655 @@
...
+  protected statsBox: Phaser.GameObjects.Sprite;
+  protected statValuesContainer: Phaser.GameObjects.Container;
+  protected statNumbers: Phaser.GameObjects.Sprite[];
+
+  get statOrder(): Stat[] {
+    return [];
+  }
+
+  /** Helper method used by the constructor to create the tera and shiny icons next to the name */
+  private constructIcons() {
```
DayKev, line 63: Why separate out these specifically?

src/ui/battle-info/battle-info.ts
```diff
@@ -0,0 +1,655 @@
...
+    this.lastLevelExp = -1;
+    this.lastLevel = -1;
+
+    // Initially invisible and shown via Pokemon.showInfo
+    this.setVisible(false);
+
+    this.box = globalScene.add.sprite(0, 0, this.getTextureName()).setName("box").setOrigin(1, 0.5);
+    this.add(this.box);
+
+    this.nameText = addTextObject(player ? -115 : -124, player ? -15.2 : -11.2, "", TextStyle.BATTLE_INFO)
```
DayKev, line 128: There's still a bunch of `player ? a : b`, shouldn't that also be split and moved into the subclasses?

src/ui/battle-info/player-battle-info.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  override initInfo(pokemon: PlayerPokemon): void {
+    super.initInfo(pokemon);
+    this.setHpNumbers(pokemon.hp, pokemon.getMaxHp());
+    this.expMaskRect.x = (pokemon.levelExp / getLevelTotalExp(pokemon.level, pokemon.species.growthRate)) * 510;
+    this.lastExp = pokemon.exp;
+    this.lastLevelExp = pokemon.levelExp;
+
+    this.statValuesContainer.setPosition(8, 7);
+  }
```
DayKev, line 57:
```suggestion
  }

```

src/ui/battle-info/player-battle-info.ts
```diff
@@ -0,0 +1,211 @@
...
+import { Stat } from "#enums/stat";
+import BattleInfo from "./battle-info";
+
+export class PlayerBattleInfo extends BattleInfo {
+  protected player: true = true;
+  protected hpNumbersContainer: Phaser.GameObjects.Container;
+
+  override get statOrder(): Stat[] {
+    return [Stat.ATK, Stat.DEF, Stat.SPATK, Stat.SPDEF, Stat.ACC, Stat.EVA, Stat.SPD];
+  }
```
DayKev, line 14:
```suggestion
  }

```

src/ui/battle-info/battle-info.ts
```diff
@@ -0,0 +1,655 @@
...
+
+      if (this.lastStatus !== StatusEffect.NONE) {
+        this.statusIndicator.setFrame(StatusEffect[this.lastStatus].toLowerCase());
+      }
+
+      this.statusIndicator.setVisible(!!this.lastStatus).setPositionRelative(this.nameText, xOffset, 11.5);
+    }
+  }
+
+  /** Update the pokemon name inside the container ,*/
```
DayKev, line 427:
```suggestion
  /** Update the pokemon name inside the container */
```

src/ui/battle-info/battle-info.ts
```diff
@@ -0,0 +1,655 @@
...
+      .on("pointerover", () =>
+        globalScene.ui.showTooltip(
+          "",
+          `${pokemon.species.getName(pokemon.formIndex)}/${pokemon.fusionSpecies?.getName(pokemon.fusionFormIndex)}`,
+        ),
+      )
+      .on("pointerout", () => globalScene.ui.hideTooltip());
+  }
+
+  /** Called by {@linkcode initInfo} to initialize the shiny icon
```
DayKev, line 267:
```suggestion
  /**
   * Called by {@linkcode initInfo} to initialize the shiny icon
```

src/ui/battle-info/battle-info.ts
```diff
@@ -0,0 +1,655 @@
...
+    this.offset = offset;
+
+    this.x += 10 * (this.offset === this.player ? 1 : -1);
+    this.y += 27 * (this.offset ? 1 : -1);
+    this.baseY = this.y;
+  }
+
+  //#region Update methods and helpers
+
+  /** Update the status icon to match the pokemon's current status
```
DayKev, line 411:
```suggestion
  /**
   * Update the status icon to match the pokemon's current status
```

src/ui/battle-info/player-battle-info.ts
```diff
@@ -0,0 +1,211 @@
...
+            });
+            return;
+          }
+          resolve();
+        },
+      });
+    });
+  }
+
+  /** Updates the info on the info bar.
```
DayKev, line 169:
```suggestion
  /**
   * Updates the info on the info bar.
```

xsn34kzx (approved):

src/ui/battle-info/battle-info.ts
SirzBenjie: Good point! I should separate out more of them! That will make it a bit easier to split out the differences.

SirzBenjie: Okay I split them out and this also helped remove almost all instances of the checks against `isPlayer`.
I couldn't really decide on a nice way to handle it for the `setTypes` method.

SirzBenjie:
> The HP number for the player UI doesn't fade in/out any more, instead appearing and disappearing instantly when C is pressed/released.

Thanks for pointing this out. The caught indicator and champion ribbon on enemy pokemon also had the same issue.
I fixed both by ensuring they are positioned below the stat info container.

I also addressed each of the comments from your review.

DayKev (approved):

Wlowscha (approved):

damocleas (approved):

PR #5720, opened 2025-04-27
[Dev] Enable Biome checking of `pokemon.ts`

DayKev:
N/A

All TypeScript files should be checked with Biome. Performance is not significantly impacted by parsing the file.

- `src/field/pokemon.ts` removed from Biome's ignore list and Biome was applied to it.
- A rule to disallow the use of the `integer` type alias was added.

## Checklist

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -57,7 +57,7 @@
       },
       "style": {
         "noVar": "error",
-        "useEnumInitializers": "off", // large enums like Moves/Species would make this cumbersome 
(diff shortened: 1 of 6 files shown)


Madmadness65 (dismissed):

SirzBenjie (dismissed):

damocleas:
has conflicts

Wlowscha (approved):

SirzBenjie (approved):

Wlowscha (approved):

PR #5723, opened 2025-04-27
[Refactor] Create utility function `makeArray`

DayKev:
N/A

Code simplification.

A new utility function `makeArray(input)` is added.
```ts
// old
if (!Array.isArray(input)) {
  input = [input];
}

// new
input = makeArray(input);
```

## Checklist


src/utils/common.ts
```diff
@@ -601,3 +601,10 @@ export function deepMergeObjects(a: Object, b: Object) {
     }
   }
 }
+
+export function makeArray<T>(input: T | T[]): T[] {
+  if (!Array.isArray(input)) {
+    return [input];
+  }
+  return input;
```
SirzBenjie, line 609: This way gives you a bit more power with typescript inference, using type predicates. It's a lot nicer, as it forces the return type of the method to be exactly the type that was put in, rather than a generic array. This preserves tuples, if the input was known to be a tuple. It's great.
```suggestion
export function makeArray<T extends any[]>(input: T): T;
export function makeArray<T>(input: T): T | [T] {
  return Array.isArray(input) ? input : [input];
```

DayKev: image

DayKev: Tried a couple things but it didn't work, not sure how to fix it.

SirzBenjie: I forgot to have the second overload when pasting here. Edited it now

SirzBenjie (dismissed):

SirzBenjie (dismissed):

SirzBenjie (approved):

Wlowscha (approved):

PR #5731, opened 2025-04-29
[Misc] Add decrypt-save.js utility script

SirzBenjie:
None

Javi needed help decrypting a save for account support related reasons. We have no mechanism to do this that doesn't involve loading up the game.

Created a new script, `scripts/decrypt-save.js` that decrypts a save.
If we ever change how saves are encrypted, this will need to be updated.

## How to test the changes?
Take one of the saves from `tests` and then run it.

## Checklist

### DIFF
--- a/scripts/decrypt-save.js
+++ b/scripts/decrypt-save.js
@@ -1,0 +1,153 @@
+import pkg from "crypto-js";
+const { AES, enc } = pkg;
+// biome-ignore lint: This is how you import fs from node
+import * as fs from "node:fs";
+
+const SAVE_KEY = "x0i2O7WRiANTqPmZ";
+
+/**
+ * A map of condensed keynames to their associated full names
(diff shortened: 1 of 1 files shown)


scripts/decrypt-save.js
```diff
@@ -0,0 +1,153 @@
...
+    process.exit(1);
+  }
+}
+
+function main() {
+  let args = process.argv.slice(2);
+  if (args.includes("-h") || args.includes("--help") || args.length === 0) {
+    printUsage();
+    process.exit(0);
+  }
```
xsn34kzx, line 123:
```suggestion
```
Is there a reason for doing this twice (here and on L129) or is it fine toss out this first `if` block?

SirzBenjie: That's definitely an oversight. Thanks!

damocleas (approved):

xsn34kzx (approved):

PR #5497, opened 2025-03-08
[GitHub] Add path filters to avoid unnecessarily re-running tests

SirzBenjie:
## What are the changes the user will see?
N/A

## Why am I making these changes?
Part of the test refactor (sort of). We do not need to re-run tests in commits that only change files that cannot impact tests (e.g. the README file). These tests are long running.

## What are the changes from a developer perspective?

In summary, the test github action will only re-run when source code files, test files, any file that can impact the build, or the test workflow file itself has changes.

Accomplished by added the following [paths filter](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) to `.github/workflows/tests.yml`
```yaml
paths:
      # src and test files
      - "src/**"
      - "test/**"
      - "public/**" # PLEASE REVIEW -- Does it make sense to include this? Can modifying these break tests? (REMOVE COMMENT AFTER REVIEW)
      # Workflows that can impact tests
      - ".github/workflows/test*.yml"
      # top-level files
      - "package*.json"
      - ".nvrmc" # Updates to node version can break tests
      - "vite.*.ts" # vite.config.ts, vite.vitest.config.ts, vitest.workspace.ts
      - "tsconfig*.json" # tsconfig.json tweaking can impact compilation
      - "global.d.ts" # I think this can impact tests? (REMOVE COMMENT AFTER REVIEW)
      - ".env.*" # PLEASE REVIEW -- can these impact test behavior?
      # Blanket negations for files that cannot impact tests
      - "!**/*.py" # No .py files
      - "!**/*.sh" # No .sh files
      - "!**/*.md" # No .md files
      - "!**/.git*" # .gitkeep and family
```

## Screenshots/Videos
N/A

## How to test the changes?
You'd have to make a commit and open a pull request that only changes a file, like README.md, to verify that the tests aren't automatically re-run. The, make a commit and open a pull request that changes any source code file and verify tests are still re-run.

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- ~~[ ] Have I tested the changes manually?~~
- ~~[ ] Are all unit tests still passing? (`npm run test`)~~
  - ~~[ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?~~
[...]

PR #5744, opened 2025-05-02
[GitHub] Make tests skip if no filters are matched

SirzBenjie:
N/A

#5497 added path filters to only run tests if any of the changed files could possibly have modified them.
However, due to idiosyncracies with GitHub, this cuases required checks to be "pending" -- they would never have been triggered.

This would prevent PRs that don't actually need tests to run from being able to be merged unless someone with permissions did a bypass checks and merge.

We now use https://github.com/dorny/paths-filter to cause the test job to be skipped if none of the changed files match the filters.

I moved the filters into `.github/test-filters.yml` and outside of the workflow file. Also deduplicated. This was allowed due to the paths filter action.

## Screenshots/Videos

N/A

## How to test the changes?
N/A

## Checklist
- [x] Have I tested the changes manually? Yes I followed a similar setup on my own repo in another spot
- ~~[ ] Are all unit tests still passing? (`npm run test:silent`)~~


.github/test-filters.yml
```diff
@@ -0,0 +1,19 @@
...
+  # Workflows that can impact tests
+  - ".github/workflows/test*.yml"
+  - ".github/test-filters.yml"
+  # top-level files
+  - "package*.json"
+  - ".nvrmc" # Updates to node version can break tests
+  - "vite*" # vite.config.ts, vite.vitest.config.ts, vitest.workspace.ts
+  - "tsconfig*.json" # tsconfig.json tweaking can impact compilation
+  - "global.d.ts"
+  - ".env.*"
```
DayKev, line 14: I just realized that this won't catch `.env` itself, only `.env.mode` files.

SirzBenjie: Easy to fix. Though it shouldn't matter; tests don't use that env file.

DayKev (dismissed):

Madmadness65 (approved):

DayKev (approved):

Wlowscha (approved):

damocleas (approved):

PR #5786, opened 2025-05-06
[Bug] Lock Capsule not locking rarities consistently

### LINKED ISSUE #5626: [Bug] Lock Capsule not locking rarities consistently
### Describe the bug

In endless mode, when i lock rarities in the reward screen and reroll a few times, one of the locked rares will not be locked correctly and turn into a blue item.
Big nugget -> TM138 - Aerial Ace
This behaviour can be seen in the last two uploaded screenshots

### Reproduction

I noticed this the first time today and am not sure, whether it is a rare incident or happening after an update.
I will mention it if it happens again, my session data and run will be included.

### Expected behavior

Locking rarities in the rewards screen correctly locks all rarities.

### Screenshots / Videos

Image

Image

Image

Image

Image

Image

Image

Image

Image

### Session export file

[sessionData_KetschupGo.txt](https://github.com/user-attachments/files/19600602/sessionData_KetschupGo.txt)
[sessionData1_KetschupGo.txt](https://github.com/user-attachments/files/19600603/sessionData1_KetschupGo.txt)

### User data export file

[runHistoryData_KetschupGo.txt](https://github.com/user-attachments/files/19600609/runHistoryData_KetschupGo.txt)
[data_KetschupGo.txt](https://github.com/user-attachments/files/19600608/data_KetschupGo.txt)

### Additional context

I uploaded 2 session files as i am not sure which one is correct, i hope you can figure it out. Sorry for that.
Also, i checked the dev console and no messages were shown that indicate an error.

lxy-lxy-lxy:
Fix #5626, I add lockModifierTiers to check, this is current result:
Snipaste_2025-05-06_17-38-59
—————————————————————————————————————————————————————————
This is issue #5626 
image


src/modifier/modifier-type.ts
```diff
@@ -3287,6 +3287,8 @@ export function getModifierTypeFuncById(id: string): ModifierTypeFunc {
   return modifierTypes[id];
 }
 
+let lockModifierTiers: null | boolean = null;
+
```
DayKev, line 3291:
```suggestion
```

src/modifier/modifier-type.ts
```diff
@@ -3312,6 +3314,7 @@ export function getPlayerModifierTypeOptions(
   const options: ModifierTypeOption[] = [];
   const retryCount = Math.min(count * 5, 50);
   if (!customModifierSettings) {
+    lockModifierTiers = !!modifierTiers && modifierTiers.length >= count;
```
DayKev, line 3317:
```suggestion
```

src/modifier/modifier-type.ts
```diff
@@ -3322,6 +3325,8 @@ export function getPlayerModifierTypeOptions(
         ),
       );
     });
+    // init lockModifierTiers
+    lockModifierTiers = null;
```
DayKev, line 3329:
```suggestion
```

src/modifier/modifier-type.ts
```diff
@@ -3640,7 +3645,7 @@ function getNewModifierTypeOption(
       }
       tier += upgradeCount;
     }
-  } else if (retryCount === 10 && tier) {
+  } else if (retryCount === 10 && tier && lockModifierTiers !== true) {
```
DayKev, line 3648:
```suggestion
  } else if (retryCount >= 100 && tier) {
```

DayKev (commented):
Technically the bug is intentional (to prevent rare cases where it might keep rerolling for an extremely long time, if not forever), but the max reroll count of 10 is probably too low. Increasing the max reroll count to 100 would prevent it from happening except in the case where something is broken and there's no other choice but to drop down a tier.

lxy-lxy-lxy:
> Technically the bug is intentional (to prevent rare cases where it might keep rerolling for an extremely long time, if not forever), but the max reroll count of 10 is probably too low. Increasing the max reroll count to 100 would prevent it from happening except in the case where something is broken and there's no other choice but to drop down a tier.

Oh okay, now I get it!

lxy-lxy-lxy:
After testing this can be resolved

DayKev (approved):

PR #5806, opened 2025-05-09
[Bug][Hotfix] Fix crashes when loading save with a transformed pokemon

emdeann:
Saves will run, hopefully. This introduces a new bug where the sprite of the transformed pokemon will not display properly. This is obviously preferrable to the current situation

P0

The main issue causing the crash was that `summonData.speciesForm` wasn't properly being constructed in newly-created `Pokemon` objects, so it was a raw object that crashed on method calls. This was caused by the cud chew PR removing `summonDataPrimer`, but not properly redirecting the construction of `speciesForm` from the primer to the actual pokemon object

I also changed the icon-related code to default to ignoring transformations and correctly sync the atlas keys and individual icon keys. I did this because I thought the icons being wacky was related to the data saving issues, but this is evidently not the case. (It actually did work properly for the save selection screen before, but that was essentially pure luck and balanced on a house of cards)
## How to test the changes?
Grab a save from https://discord.com/channels/1125469663833370665/1369529068529127546 or related reports and see if it loads

## Checklist


DayKev:
It fixed some scenarios, but it seems there's still issues.
image
[sessionData2 - transform crash pr5806.prsv.txt](https://github.com/user-attachments/files/20117485/sessionData2.-.transform.crash.pr5806.prsv.txt)

DayKev (approved):
Fixes the bug it sets out to fix

PR #5808, opened 2025-05-09
[Bug][Hotfix] Fix Transformed Sprites not loading properly

emdeann:
A transformed pokemon at the lead of the party will display the correct sprite when loading a save. Hopefully also prevents a related phaser crash

Should fix https://discord.com/channels/1125469663833370665/1368847213224132618

Similarly to icons, the code for grabbing the atlas and the code for grabbing the sprite key weren't synced in terms of their use of `ignoreOverride`, so one would grab the transformed species and the other would grab the base species (i.e. ditto), leading to mixups. No idea how it ever worked correctly, in all honesty.

As an example, `loadAssets()` is called when generating pokemon from save data with no arguments, and `ignoreOverride` defaults to true in that method. `ignoreOverride` is then passed (as `true`) when getting the atlas, but *not* passed (defaulting to false) when getting the actual sprite key. Genuinely, no idea

I changed the call to `getBattleSpriteKey()` to actually use the `ignoreOverride` parameter, as well as changed the relevant `loadAssets()` call to pass `false` to `ignoreOverride`

## Screenshots/Videos
<p>

https://github.com/user-attachments/assets/7ed1ee19-a0b0-4499-b4da-214e1d936483

</p>
</details> 

<p>

https://github.com/user-attachments/assets/c8832579-ed1a-46be-999d-3ba8e1ae899e

</p>
</details> 

## How to test the changes?
Grab a broken ditto save and see if the sprite loads in correctly

## Checklist

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -909,7 +909,6 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       const originalWarn = console.warn;
       // Ignore warnings for missing frames, because there will be a lot
       console.warn = () => {};
-      // NEED REVIEW: is `this.isPlayer()` always the correct check for whether to get the front of back sprite?
       const battleSpriteKey = this.getBattleSpriteKey(this.isPlayer(), ignoreOverride);
       const battleFrameNames = globalScene.anims.generateFrameNames(battleSpriteKey, {
         zeroPad: 4,
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -909,19 +909,23 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       const originalWarn = console.warn;
       // Ignore warnings for missing frames, because there will be a lot
       console.warn = () => {};
-      const battleFrameNames = globalScene.anims.generateFrameNames(this.getBattleSpriteKey(), {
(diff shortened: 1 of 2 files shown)


src/field/pokemon.ts
```diff
@@ -909,19 +909,23 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       const originalWarn = console.warn;
       // Ignore warnings for missing frames, because there will be a lot
       console.warn = () => {};
-      const battleFrameNames = globalScene.anims.generateFrameNames(this.getBattleSpriteKey(), {
+      // NEED REVIEW: is `this.isPlayer()` always the correct check for whether to get the front of back sprite?
```
DayKev, line 912: Only the player should ever have a Pokémon back sprite.

DayKev (approved):

src/field/pokemon.ts
emdeann: i realized if you go deep enough into that function it defaults to `isPlayer()` anyway

DayKev: Lmao of course it does.

SirzBenjie (approved):
Our sprite code is built on a house of cards.
I have been trying to improve it.
This is a good catch.
From my understanding, this should hold up.

Madmadness65 (approved):

PR #5692, opened 2025-04-23
[Move Bug] Fixed dancer move failure interactions

Bertie690:
# THIS PR CANNOT BE MERGED UNTL #5810 IS MERGED
Once it is I will squash all my changes into a single commit based off of it.
The reason this PR looks so huge is it being based off of the latter PR.

## What are the changes the user will see?
Dancer (in keeping with mainline behavior), no longer works if the move fails to affect/is protected against/misses all its targetstargets.
Also flinches will not bork dancer.

Fixes #5052
Fixes #5833
Fixes #5867

## Why am I making these changes?
because i'm the only person on this planet insane and devoted enough to this completely meaningless stuff to do something about it

## What are the changes from a developer perspective?
- Moved Dancer procs inside MoveEffectPhase to allow access to hit check results, and made PostMoveUsedAbAttr take a HitChecks parameter to check the success/failure of the invoked move.
- Added tests apropos of Dancer for both this PR and #5810.

Reworked early bird code to be less jank

Assorted code cleanups & doc fixes

## Screenshots/Videos
TODO

## How to test the changes?
`npm run test -- dancer.test.ts`
overrides if you really want to I guess

## Checklist
- [x] I'm using beta as my base branch
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [] Have I tested the changes manually?
- [ ] Are all unit tests still passing? (`npm run test`)
- [x] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?

PR #5810, opened 2025-05-10
[Refactor/Bug] Added `MoveUseType` + refactored MEP

### LINKED ISSUE #5748: [Bug] Certain moves checking last move used have incorrect status interactions
### Describe the bug

There are a decent amount of moves doing something vaguely similar to this - calling `getLastXMoves(1)[0]` to check the very last entry in the target's move history.

Image

HOWEVER, this also causes these moves to fail if the last status-induced failures (paralysis, confusion, etc).

Affected move attributes include but are not limited to:
- `CopyMoveAttr` (Mirror Move/Copycat)
- `ReducePpMoveAttr` (Spite/Eerie Spell)
- `ResistLastMoveTypeAttr` (Conversion 2) (May or may not actually fail on cart; too lazy to check showdown atm)

### Reproduction

```ts
const overrides = {
  MOVESET_OVERRIDE: [Moves.SPITE, Moves.RECOVER],
  OPP_MOVESET_OVERRIDE: [Moves.TACKLE],
  OPP_STATUS_OVERRIDE: StatusEffect.PARALYSIS
```
spam spite until the enemy mon gets full paralyzed.
(Won't actually run out of pp due to overrides)

### Expected behavior

Image
The moves should skip status whiffs as they technically don't count as using a move at all.

### Screenshots / Videos

https://github.com/user-attachments/assets/8db789d8-1805-47e2-93de-ea5fd7c72eb2

### Session export file

_No response_

### User data export file

_No response_

### Additional context
Fixed in #5692

### LINKED ISSUE #5809: [Move Bug] After You, Quash and charging moves always consume PP
### Describe the bug

Currently, Quash, After You and charging moves incorrectly reset the `virtual` and `ignorePP` flags on the resulting `MovePhase` to `false`, effectively meaning they always consume PP and never count as virtually invoked (for Last Resort), even when used via Metronome or Copycat.
Former are bugs in the actual move attribute moving the phase around, latter is a bug with `MoveChargePhase` itself

### Reproduction

Since local dev moveset overrides forcibly fix PP counts to a set value, you'd have to use a starter with a 2 turn charging move (like electro shot/solarbeam) and either a starter learning copycat/mirror move (happiny, etc) or an enemy overridden to spam quash/after you.

### Expected behavior

Not to consume PP. 
I have confirmation for copycat [on showdown](https://replay.pokemonshowdown.com/gen9doublesou-2356043948-4923hbjrwy3q6rey25m2spaw5cei1awpw), and I'd heavily presume quash/after you to function similarly.

### Screenshots / Videos

Image

### Session export file

_No response_

### User data export file

_No response_

### Additional context

Fixed in dancer refactor PR (pull request will be made shortly)

Bertie690:
Not much, other than minor bugfixes

Fixes #5809
Fixes #5748

Refactors: Needed for dancer PR #5692.

Bugfixes: trivial to fix with new use type

### MoveUseType
- A new enum (`MoveUseType`) has been added, replacing the optional `virtual`/`reflected`/etc booleans inside `TurnMove`, `MovePhase` and `MoveEffectPhase`. Each one (more or less) inherits the exclusions of the types before it, inspired by the fact that all of the prior booleans effectively "implied" one another (`ignorePP` was always true when `virtual` was, etc.) 
  - I highly recommend looking at `move-use-type.ts` **BEFORE** the rest of the changes (or at least reading the next section)
  - `MoveUseType.INDIRECT` is unused and will be exponded upon in the dancer PR. 

#### Equivalence Table for MEP
The use of `>` vs `>=`  is intentional for both extensibility and IDE editor highlighting. The ONLY checks that use `===` are specific checks for edge cases.

| Boolean combination |     `MoveUseType` check    |
| --------------------|----------------------------|
| `if (virtual)`      | `>= MoveUseType.INDIRECT`  |
| `if (!virtual)`     | `< MoveUseType.INDIRECT`   |
| `if (ignorePP)`     | `>= MoveUseType.IGNORE_PP` |
| `if (!ignorePP)`    | `< MoveUseType.IGNORE_PP`  |
| `if (reflected)`    | `>= MoveUseType.REFLECTED` |

### Other changes
- Fixed several bugs with move selection via the new utility function `getLastNonVirtualMove`. 
- Added tests for said changes

## Screenshots/Videos
TBD (a lot of this shit is hard to get videos due to order of operations)

## How to test the changes?
check the many new automated tests

## Checklist
  - pls i already split this up once and reverted half a dozen comment fixes

### DIFF
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -1,8 +1,8 @@
-import { HitResult, MoveResult, PlayerPokemon } from "#app/field/pokemon";
+import { HitResult, MoveResult, PlayerPokemon, type TurnMove } from "#app/field/pokemon";
 import { BooleanHolder, NumberHolder, toDmgValue, isNullOrUndefined, randSeedItem, randSeedInt, type Constructor } from "#app/utils/common";
 import { getPokemonNameWithAffix } from "#app/messages";
 import { BattlerTagLapseType, GroundedTag } from "#app/data/battler-tags";
-import { getNonVolatileStatusEffects, getStatusEffectDescriptor, getStatusEffectHealText } from "#app/data/status-effect";
+import { getNonVolatileStatusEffects, getStatusEffectDescriptor, getStatusEffectHealText, type Status } from "#app/data/status-effect";
 import { Gender } from "#app/data/gender";
 import {
   AttackMove,
@@ -67,8 +67,8 @@ import { BerryUsedEvent } from "#app/events/battle-scene";
 
 
 // Type imports
-import type { EnemyPokemon, PokemonMove } from "#app/field/pokemon";
-import type Pokemon from "#app/field/pokemon";
+import { EnemyPokemon, PokemonMove } from "#app/field/pokemon";
+import Pokemon from "#app/field/pokemon";
 import type { Weather } from "#app/data/weather";
 import type { BattlerTag } from "#app/data/battler-tags";
 import type { AbAttrCondition, PokemonDefendCondition, PokemonStatStageChangeCondition, PokemonAttackCondition, AbAttrApplyFunc, AbAttrSuccessFunc } from "#app/@types/ability-types";
@@ -76,6 +76,7 @@ import type { BattlerIndex } from "#app/battle";
 import type Move from "#app/data/moves/move";
 import type { ArenaTrapTag, SuppressAbilitiesTag } from "#app/data/arena-tag";
 import { SelectBiomePhase } from "#app/phases/select-biome-phase";
+import { MoveUseType } from "#enums/move-use-type";
 import { noAbilityTypeOverrideMoves } from "../moves/invalid-moves";
 
 export class BlockRecoilDamageAttr extends AbAttr {
@@ -1067,7 +1068,7 @@ export class PostDefendMoveDisableAbAttr extends PostDefendAbAttr {
   }
 
   override canApplyPostDefend(pokemon: Pokemon, passive: boolean, simulated: boolean, attacker: Pokemon, move: Move, hitResult: HitResult | null, args: any[]): boolean {
-    return attacker.getTag(BattlerTagType.DISABLED) === null
+    return !isNullOrUndefined(attacker.getTag(BattlerTagType.DISABLED))
       && move.doesFlagEffectApply({flag: MoveFlags.MAKES_CONTACT, user: attacker, target: pokemon}) && (this.chance === -1 || pokemon.randSeedInt(100) < this.chance);
   }
 
@@ -1246,7 +1247,7 @@ export class MoveTypeChangeAbAttr extends PreAttackAbAttr {
 
   /**
    * Determine if the move type change attribute can be applied
-   * 
+   *
    * Can be applied if:
    * - The ability's condition is met, e.g. pixilate only boosts normal moves,
    * - The move is not forbidden from having its type changed by an ability, e.g. {@linkcode Moves.MULTI_ATTACK}
@@ -1262,7 +1263,7 @@ export class MoveTypeChangeAbAttr extends PreAttackAbAttr {
    */
   override canApplyPreAttack(pokemon: Pokemon, _passive: boolean, _simulated: boolean, _defender: Pokemon | null, move: Move, _args: [NumberHolder?, NumberHolder?, ...any]): boolean {
     return (!this.condition || this.condition(pokemon, _defender, move)) &&
-            !noAbilityTypeOverrideMoves.has(move.id) && 
+            !noAbilityTypeOverrideMoves.has(move.id) &&
             (!pokemon.isTerastallized ||
               (move.id !== Moves.TERA_BLAST &&
               (move.id !== Moves.TERA_STARSTORM || pokemon.getTeraType() !== PokemonType.STELLAR || !pokemon.hasSpecies(Species.TERAPAGOS))));
@@ -2334,18 +2335,18 @@ export class PostSummonStatStageChangeAbAttr extends PostSummonAbAttr {
       // phase list (which could be after CommandPhase for example)
       globalScene.unshiftPhase(new StatStageChangePhase(pokemon.getBattlerIndex(), true, this.stats, this.stages));
     } else {
-      for (const opponent of pokemon.getOpponents()) {
-        const cancelled = new BooleanHolder(false);
-        if (this.intimidate) {
-          applyAbAttrs(IntimidateImmunityAbAttr, opponent, cancelled, simulated);
-          applyAbAttrs(PostIntimidateStatStageChangeAbAttr, opponent, cancelled, simulated);
+    for (const opponent of pokemon.getOpponents()) {
+      const cancelled = new BooleanHolder(false);
+      if (this.intimidate) {
+        applyAbAttrs(IntimidateImmunityAbAttr, opponent, cancelled, simulated);
+        applyAbAttrs(PostIntimidateStatStageChangeAbAttr, opponent, cancelled, simulated);
 
-          if (opponent.getTag(BattlerTagType.SUBSTITUTE)) {
-            cancelled.value = true;
-          }
+        if (opponent.getTag(BattlerTagType.SUBSTITUTE)) {
+          cancelled.value = true;
         }
-        if (!cancelled.value) {
-          globalScene.unshiftPhase(new StatStageChangePhase(opponent.getBattlerIndex(), false, this.stats, this.stages));
+      }
+      if (!cancelled.value) {
+        globalScene.unshiftPhase(new StatStageChangePhase(opponent.getBattlerIndex(), false, this.stats, this.stages));
         }
       }
     }
@@ -4398,12 +4399,12 @@ export class PostBiomeChangeTerrainChangeAbAttr extends PostBiomeChangeAbAttr {
  * @extends AbAttr
  */
 export class PostMoveUsedAbAttr extends AbAttr {
-  canApplyPostMoveUsed(
+    canApplyPostMoveUsed(
     pokemon: Pokemon,
     move: PokemonMove,
     source: Pokemon,
     targets: BattlerIndex[],
-    simulated: boolean,
+        simulated: boolean,
     args: any[]): boolean {
     return true;
   }
@@ -4413,7 +4414,7 @@ export class PostMoveUsedAbAttr extends AbAttr {
     move: PokemonMove,
     source: Pokemon,
     targets: BattlerIndex[],
-    simulated: boolean,
+        simulated: boolean,
     args: any[],
   ): void {}
 }
@@ -4446,16 +4447,17 @@ export class PostDancingMoveAbAttr extends PostMoveUsedAbAttr {
     move: PokemonMove,
     source: Pokemon,
     targets: BattlerIndex[],
-    simulated: boolean,
+        simulated: boolean,
     args: any[]): void {
     if (!simulated) {
       // If the move is an AttackMove or a StatusMove the Dancer must replicate the move on the source of the Dance
+       // TODO: fix in main dancer PR (currently keeping this purely semantic rather than actually fixing bug)
       if (move.getMove() instanceof AttackMove || move.getMove() instanceof StatusMove) {
         const target = this.getTarget(dancer, source, targets);
-        globalScene.unshiftPhase(new MovePhase(dancer, target, move, true, true));
+        globalScene.unshiftPhase(new MovePhase(dancer, target, move, MoveUseType.FOLLOW_UP));
       } else if (move.getMove() instanceof SelfStatusMove) {
         // If the move is a SelfStatusMove (ie. Swords Dance) the Dancer should replicate it on itself
-        globalScene.unshiftPhase(new MovePhase(dancer, [ dancer.getBattlerIndex() ], move, true, true));
+        globalScene.unshiftPhase(new MovePhase(dancer, [ dancer.getBattlerIndex() ], move, MoveUseType.FOLLOW_UP))
       }
     }
   }
@@ -4897,7 +4899,7 @@ export class BlockRedirectAbAttr extends AbAttr { }
  * @see {@linkcode apply}
  */
 export class ReduceStatusEffectDurationAbAttr extends AbAttr {
-  private statusEffect: StatusEffect;
+private statusEffect: StatusEffect;
 
   constructor(statusEffect: StatusEffect) {
     super(false);
@@ -4906,7 +4908,7 @@ export class ReduceStatusEffectDurationAbAttr extends AbAttr {
   }
 
   override canApply(pokemon: Pokemon, passive: boolean, simulated: boolean, args: any[]): boolean {
-    return args[1] instanceof NumberHolder && args[0] === this.statusEffect;
+        return args[1] instanceof NumberHolder && args[0] === this.statusEffect;
   }
 
   /**
@@ -5545,6 +5547,7 @@ class ForceSwitchOutHelper {
    *
    * @param pokemon The {@linkcode Pokemon} attempting to switch out.
    * @returns `true` if the switch is successful
+   * TODO: Make this actually cancel pending move phases on the switched out target
    */
   public switchOutLogic(pokemon: Pokemon): boolean {
     const switchOutTarget = pokemon;
@@ -5554,7 +5557,7 @@ class ForceSwitchOutHelper {
      * - If the Pokémon is still alive (hp > 0), and if so, it leaves the field and a new SwitchPhase is initiated.
      */
     if (switchOutTarget instanceof PlayerPokemon) {
-      if (globalScene.getPlayerParty().filter((p) => p.isAllowedInBattle() && !p.isOnField()).length < 1) {
+      if (globalScene.getPlayerParty().every(p => !p.isActive(true))) {
         return false;
       }
 
@@ -5887,7 +5890,7 @@ export function applyPostMoveUsedAbAttrs(
   move: PokemonMove,
   source: Pokemon,
   targets: BattlerIndex[],
-  simulated = false,
+    simulated = false,
   ...args: any[]
 ): void {
   applyAbAttrsInternal<PostMoveUsedAbAttr>(
@@ -6897,8 +6900,9 @@ export function initAbilities() {
       .ignorable(),
     new Ability(Abilities.ANALYTIC, 5)
       .attr(MovePowerBoostAbAttr, (user, target, move) => {
-        const movePhase = globalScene.findPhase((phase) => phase instanceof MovePhase && phase.pokemon.id !== user?.id);
-        return isNullOrUndefined(movePhase);
+        // Boost power if all other Pokemon have already moved (no other moves are slated to execute)
+        const laterMovePhase = globalScene.findPhase((phase) => phase instanceof MovePhase && phase.pokemon.id !== user?.id);
+        return isNullOrUndefined(laterMovePhase);
       }, 1.3),
     new Ability(Abilities.ILLUSION, 5)
       // The Pokemon generate an illusion if it's available
--- a/src/data/arena-tag.ts
+++ b/src/data/arena-tag.ts
@@ -30,6 +30,7 @@ import { MoveEffectPhase } from "#app/phases/move-effect-phase";
 import { PokemonHealPhase } from "#app/phases/pokemon-heal-phase";
 import { StatStageChangePhase } from "#app/phases/stat-stage-change-phase";
 import { CommonAnimPhase } from "#app/phases/common-anim-phase";
+import { MoveUseType } from "#enums/move-use-type";
 
 export enum ArenaTagSide {
   BOTH,
@@ -892,7 +893,7 @@ export class DelayedAttackTag extends ArenaTag {
 
     if (!ret) {
       globalScene.unshiftPhase(
-        new MoveEffectPhase(this.sourceId!, [this.targetIndex], allMoves[this.sourceMove!], false, true),
+        new MoveEffectPhase(this.sourceId!, [this.targetIndex], allMoves[this.sourceMove!], MoveUseType.REFLECTED), // Reflected ensures this doesn't check status, use PP or be copied
       ); // TODO: are those bangs correct?
     }
 
--- a/src/data/battler-tags.ts
+++ b/src/data/battler-tags.ts
@@ -44,12 +44,17 @@ import { EFFECTIVE_STATS, getStatKey, Stat, type BattleStat, type EffectiveStat 
 import { StatusEffect } from "#enums/status-effect";
 import { WeatherType } from "#enums/weather-type";
 import { isNullOrUndefined } from "#app/utils/common";
+import { MoveUseType } from "#enums/move-use-type";
 
 export enum BattlerTagLapseType {
   FAINT,
   MOVE,
   PRE_MOVE,
   AFTER_MOVE,
+  /**
+   * TODO: Stop treating this like a catch-all "semi invulnerability" tag;
+   * we may want to use this for other stuff later
+   */
   MOVE_EFFECT,
   TURN_END,
   HIT,
@@ -60,7 +65,7 @@ export enum BattlerTagLapseType {
 
 export class BattlerTag {
   public tagType: BattlerTagType;
-  public lapseTypes: BattlerTagLapseType[];
+  public lapseTypes: BattlerTagLapseType[]; // TODO: Make this a set
   public turnCount: number;
   public sourceMove: Moves;
   public sourceId?: number;
@@ -93,7 +98,7 @@ export class BattlerTag {
   onOverlap(_pokemon: Pokemon): void {}
 
   /**
-   * Tick down this {@linkcode BattlerTag}'s duration.
+   * Trigger and tick down this {@linkcode BattlerTag}'s duration.
    * @returns `true` if the tag should be kept (`turnCount` > 0`)
    */
   lapse(_pokemon: Pokemon, _lapseType: BattlerTagLapseType): boolean {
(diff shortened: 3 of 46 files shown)


src/data/mystery-encounters/encounters/absolute-avarice-encounter.ts
```diff
@@ -303,7 +304,7 @@ export const AbsoluteAvariceEncounter: MysteryEncounter = MysteryEncounterBuilde
           sourceBattlerIndex: BattlerIndex.ENEMY,
           targets: [BattlerIndex.ENEMY],
           move: new PokemonMove(Moves.STUFF_CHEEKS),
-          ignorePp: true,
+          useType: MoveUseType.IGNORE_PP,
```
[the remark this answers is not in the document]
Bertie690: Idk man I just copy paste code
My instinct says we should remove it and always ignore PP consumption for battle start moves, but that's a topic for another day and another pr
~~I'm already worried about code bloat as is lol~~

test/mystery-encounter/encounters/bug-type-superfan-encounter.test.ts
```diff
@@ -116,37 +116,25 @@ const POOL_3_POKEMON: { species: Species; formIndex?: number }[] = [
...
 const POOL_4_POKEMON = [Species.GENESECT, Species.SLITHER_WING, Species.BUZZWOLE, Species.PHEROMOSA];
 
-const PHYSICAL_TUTOR_MOVES = [
-  Moves.MEGAHORN,
-  Moves.ATTACK_ORDER,
-  Moves.BUG_BITE,
-  Moves.FIRST_IMPRESSION,
-  Moves.LUNGE
-];
+const PHYSICAL_TUTOR_MOVES = [Moves.MEGAHORN, Moves.ATTACK_ORDER, Moves.BUG_BITE, Moves.FIRST_IMPRESSION, Moves.LUNGE];
```
damocleas, line 119: why were the changes here made?

Bertie690: probably biome jank lol

src/data/abilities/ability.ts
```diff
@@ -1067,7 +1068,7 @@ export class PostDefendMoveDisableAbAttr extends PostDefendAbAttr {
   }
 
   override canApplyPostDefend(pokemon: Pokemon, passive: boolean, simulated: boolean, attacker: Pokemon, move: Move, hitResult: HitResult | null, args: any[]): boolean {
-    return attacker.getTag(BattlerTagType.DISABLED) === null
+    return !isNullOrUndefined(attacker.getTag(BattlerTagType.DISABLED))
```
DayKev, line 1071: Isn't this reversing the result?
Before: `true` if `getTag` is `null`
After: `true` if `getTag` is not `null`

src/data/arena-tag.ts
```diff
@@ -892,7 +893,7 @@ export class DelayedAttackTag extends ArenaTag {
 
     if (!ret) {
       globalScene.unshiftPhase(
-        new MoveEffectPhase(this.sourceId!, [this.targetIndex], allMoves[this.sourceMove!], false, true),
+        new MoveEffectPhase(this.sourceId!, [this.targetIndex], allMoves[this.sourceMove!], MoveUseType.REFLECTED), // Reflected ensures this doesn't check status, use PP or be copied
```
DayKev, line 896: So the name of this value isn't accurate then, since this is not a "reflected" move.

src/data/pokemon-species.ts
```diff
@@ -76,13 +76,14 @@ export const normalForm: Species[] = [
...
+ * @param species - The {@linkcode Species} to fetch.
+ * If an array of `Species` is passed (such as for named trainer spawn pools),
+ * one will be selected at random.
+ * @returns The {@linkcode PokemonSpecies} object associated with the given species.
  */
 export function getPokemonSpecies(species: Species | Species[]): PokemonSpecies {
-  // If a special pool (named trainers) is used here it CAN happen that they have a array as species (which means choose one of those two). So we catch that with this code block
   if (Array.isArray(species)) {
-    // Pick a random species from the list
+    // TODO: can't we just use normal int number gen rather than this junk
```
DayKev, line 86:
```suggestion
    // TODO: this should not be handled by this function
```

src/phases/pokemon-phase.ts
```diff
@@ -4,21 +4,23 @@ import type Pokemon from "#app/field/pokemon";
 import { FieldPhase } from "./field-phase";
 
 export abstract class PokemonPhase extends FieldPhase {
-  protected battlerIndex: BattlerIndex | number;
+  protected battlerIndex: BattlerIndex;
```
DayKev, line 7:
```suggestion
  protected battlerIndex: BattlerIndex | number;
```
This is `| number` because it (insanely) assumes that this is actually a pokemon ID and not a `BattlerIndex` if the number is greater than the possible `BattlerIndex` values. You could probably rename it to `battlerIndexOrId` for now so at least it's clear that's what is going on.

src/phases/pokemon-phase.ts
```diff
@@ -4,21 +4,23 @@ import type Pokemon from "#app/field/pokemon";
...
   public player: boolean;
   public fieldIndex: number;
 
-  constructor(battlerIndex?: BattlerIndex | number) {
+  constructor(battlerIndex?: BattlerIndex) {
```
DayKev, line 11:
```suggestion
  constructor(battlerIndex?: BattlerIndex | number) {
```

src/data/arena-tag.ts
Bertie690: Is it even an actual move at all? ATM we push the move history entry once the future sight hit resolves (which may or may not even be correct due to truant shenanigans)

src/phases/pokemon-phase.ts
Bertie690: this... is even worse than the bleeping "oops 0 PID means bork" issues with transferring items

src/data/battler-tags.ts
```diff
@@ -1147,29 +1142,26 @@ export class EncoreTag extends MoveRestrictionBattlerTag {
...
   canAdd(pokemon: Pokemon): boolean {
-    const lastMoves = pokemon.getLastXMoves(1);
-    if (!lastMoves.length) {
+    const lastMove = pokemon.getLastNonVirtualMove(false);
+    if (!lastMove) {
       return false;
     }
 
-    const repeatableMove = lastMoves[0];
+    const unEncoreableMoves = new Set<Moves>([
```
Wlowscha, line 1150: Wouldn't this find a better home in invalid-moves.ts?

Bertie690: yes, but i didn't really feel like bloating the PR too much. 
Moving invalid moves can (mostly) be its own thing.

src/phases/move-phase.ts
```diff
@@ -221,146 +225,145 @@ export class MovePhase extends BattlePhase {
...
-    const isDelayedAttack = this.move.getMove().hasAttr(DelayedAttackAttr);
+    const isDelayedAttack = move.hasAttr(DelayedAttackAttr);
     if (isDelayedAttack) {
-      // Check the player side arena if future sight is active
-      const futureSightTags = globalScene.arena.findTags(t => t.tagType === ArenaTagType.FUTURE_SIGHT);
-      const doomDesireTags = globalScene.arena.findTags(t => t.tagType === ArenaTagType.DOOM_DESIRE);
-      let fail = false;
+      // Check the player side arena if another delayed attack is active and hitting the same slot.
+      const delayedAttackTags = globalScene.arena.findTags(t =>
+        [ArenaTagType.FUTURE_SIGHT, ArenaTagType.DOOM_DESIRE].includes(t.tagType),
```
SirzBenjie, line 336: Can you please combine the check for these tags with the check for the target index please, so that we don't call `some` twice?
Also, let's avoid making arrays when we don't need to. 
Something along the lines of 
```ts
const delayedAttackTags = globalScene.arena.findTags(t => (t === ArenaTagType.FUTURE_SIGHT || t === ArenaTagType.DOOM_DESIRE) && (tag as DelayedAttackTag).targetIndex === currentTargetIndex)
```

src/phases/move-phase.ts
```diff
@@ -370,123 +373,125 @@ export class MovePhase extends BattlePhase {
...
      */
-    let failedDueToTerrain = false;
-    let failedDueToWeather = false;
-    if (success) {
-      const passesConditions = move.applyConditions(this.pokemon, targets[0], move);
-      failedDueToWeather = globalScene.arena.isMoveWeatherCancelled(this.pokemon, move);
-      failedDueToTerrain = globalScene.arena.isMoveTerrainCancelled(this.pokemon, this.targets, move);
-      success = passesConditions && !failedDueToWeather && !failedDueToTerrain;
-    }
+    const passesConditions = move.applyConditions(this.pokemon, targets[0], move);
```
SirzBenjie, line 382: It was intentional to only call this code in the case that the move was already successful. Please revert.
applyCondition need not be called if the move was already unsuccessful.

src/phases/turn-start-phase.ts
```diff
@@ -171,34 +173,19 @@ export class TurnStartPhase extends FieldPhase {
...
-              const playerPhase = new MovePhase(
-                pokemon,
-                turnCommand.targets || turnCommand.move!.targets,
-                move,
-                false,
-                queuedMove.ignorePP,
-              ); //TODO: is the bang correct here?
-              globalScene.pushPhase(playerPhase);
-            }
-          } else {
```
SirzBenjie, line 192: I may have missed something in the PR, but can you explain why removing this logic is OK?

src/system/version_migration/versions/v1_10_0.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+const fixMoveHistory: SessionSaveMigrator = {
+  version: "1.10.0",
+  migrate: (data: SessionSaveData): void => {
+    const mapTurnMove = (tm: OldTurnMove): TurnMove => ({
+      move: tm.move,
+      targets: tm.targets,
+      result: tm.result,
+      turn: tm.turn,
+      // NOTE: This currently mis-classifies Dancer and Magic Bounce-induced moves, but not much we can do about it tbh
```
SirzBenjie, line 32: Can you explain how it misclassifies them, please?
Honestly, it might be easier just to drop the moves from the history. I believe the impact is low.
@Xavion3

src/ui/fight-ui-handler.ts
SirzBenjie, line 1: Can you please summarize the changes in this file and why they are necessary?
Is it just refactoring? Are these changes mandatory for the PR to work?

SirzBenjie (changes requested):
One comment I have:

I very much dislike the numeric comparison check for MoveUseType. It means that you have to remember or constantly switch to the chart.

Instead, we should be writing methods, like `isFollowUpMove` that take a parameter and return a boolean. The body of these methods can use the comparison check, that's fine. I just do not like seeing the comparisons used, as it is not intuitive that the >= comparisons make sense.

src/data/abilities/ability.ts
```diff
@@ -4446,16 +4447,17 @@ export class PostDancingMoveAbAttr extends PostMoveUsedAbAttr {
     move: PokemonMove,
     source: Pokemon,
     targets: BattlerIndex[],
-    simulated: boolean,
+        simulated: boolean,
```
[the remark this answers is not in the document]
Bertie690: i genuinely have no idea

Bertie690:
> Instead, we should be writing methods, like `isFollowUpMove` that take a parameter and return a boolean. The body of these methods can use the comparison check, that's fine. I just do not like seeing the comparisons used, as it is not intuitive that the >= comparisons make sense.

Sure.
What about explicit equality checks though?

Bertie690 (commented):
> I may have missed something in the PR, but can you explain why removing this logic is OK? 

That function call was duplicated in the code block literally 5 lines below it, so I squished the above conditional together to save a nesting level.

src/ui/fight-ui-handler.ts
Bertie690: It was changing a nested `if` into a `switch case` and passing a parameter to `CommandPhase.handleCommand`

src/phases/turn-start-phase.ts
Bertie690: It was duplicated code with the branch 5 locs below it

I just merged the conditions and moved it down

src/data/moves/move.ts
```diff
@@ -6715,20 +6734,25 @@ export class FirstMoveTypeAttr extends MoveEffectAttr {
...
     const targets = moveTargets.multiple || moveTargets.targets.length === 1
       ? moveTargets.targets
-      : [ this.hasTarget ? target.getBattlerIndex() : moveTargets.targets[user.randBattleSeedInt(moveTargets.targets.length)] ]; // account for Mirror Move having a target already
-    user.getMoveQueue().push({ move: move.id, targets: targets, virtual: true, ignorePP: true });
+      : [this.hasTarget
+        ? target.getBattlerIndex()
+        : moveTargets.targets[user.randBattleSeedInt(moveTargets.targets.length)]];
     globalScene.unshiftPhase(new LoadMoveAnimPhase(move.id));
-    globalScene.unshiftPhase(new MovePhase(user, targets, new PokemonMove(move.id, 0, 0, true), true, true));
+    globalScene.unshiftPhase(new MovePhase(user, targets, new PokemonMove(move.id), MoveUseType.FOLLOW_UP));
```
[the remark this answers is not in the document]
SirzBenjie: Ah yes. That checks out.

src/phases/turn-start-phase.ts
SirzBenjie: Thanks, I see it now.

src/ui/fight-ui-handler.ts
SirzBenjie: That makes sense to me.
A second look at this and I believe the logic is the same.

Bertie690:
> > > Instead, we should be writing methods, like `isFollowUpMove` that take a parameter and return a boolean. The body of these methods can use the comparison check, that's fine. I just do not like seeing the comparisons used, as it is not intuitive that the >= comparisons make sense.

Helper functions have been delivered.

SirzBenjie (approved):

src/phases/move-phase.ts
```diff
@@ -221,93 +230,111 @@ export class MovePhase extends BattlePhase {
...
-                attr => attr instanceof HealStatusEffectAttr && attr.selfTarget && attr.isOfEffect(StatusEffect.FREEZE),
-              ) ||
-            (!this.pokemon.randBattleSeedInt(5) && Overrides.STATUS_ACTIVATION_OVERRIDE !== true) ||
-            Overrides.STATUS_ACTIVATION_OVERRIDE === false;
-
-          activated = !healed;
-          break;
-      }
+    // Skip for follow ups/reflected moves, no status condition or post turn statuses (e.g. Poison/Toxic)
+    if (!this.pokemon.status || this.pokemon.status?.isPostTurn() || this.useType >= MoveUseType.FOLLOW_UP) {
```
Wlowscha, line 237: Another inequality left here

Wlowscha (changes requested):

Wlowscha:
Not directly relevant to this PR, but having the enum called "moveUseType" is very confusing, especially when one of the entries is called "NORMAL". Would definitely support a renaming PR down the line.

SirzBenjie (dismissed):

damocleas (dismissed):

src/enums/move-use-type.ts
```diff
@@ -0,0 +1,152 @@
...
+ * Enum representing all the possible ways a given move can be executed.
+ * Each one inherits the properties (or exclusions) of all types preceding it.
+ * Properties newly found on a given use type will be **bolded**,
+ * while oddities breaking a previous trend will be listed in _italics_.
+
+ * Callers should refrain from performing non-equality checks on `MoveUseTypes` directly,
+ * instead using the available helper functions
+ * ({@linkcode isVirtual}, {@linkcode isIgnoreStatus}, {@linkcode isIgnorePP} and {@linkcode isReflected}).
+ */
+export enum MoveUseType {
```
DayKev:
```suggestion
export const MoveUseMode = {
```
etc

src/data/abilities/ability.ts
```diff
@@ -7173,7 +7174,13 @@ export function initAbilities() {
...
       .bypassFaint(),
     new Ability(AbilityId.DANCER, 7)
-      .attr(PostDancingMoveAbAttr),
+      .attr(PostDancingMoveAbAttr)
+      .edgeCase(),
+      /* Incorrect interations with:
+      * Petal Dance (should not lock in or count down timer; currently does both)
+      * Flinches (due to tag being removed earlier)
+      * Failed/protected moves (should not trigger if original move is protected against)
+      */
```
```suggestion
      // Incorrect interations with:
      // - Petal Dance (should not lock in or count down timer; currently does both)
      // - Flinches (due to tag being removed earlier)
      // - Failed/protected moves (should not trigger if original move is protected against)
      .edgeCase(),
```

src/data/abilities/ability.ts
```diff
@@ -7316,7 +7323,9 @@ export function initAbilities() {
       .bypassFaint()
       .edgeCase(), //  interacts incorrectly with rock head. It's meant to switch abilities before recoil would apply so that a pokemon with rock head would lose rock head first and still take the recoil
     new Ability(AbilityId.GORILLA_TACTICS, 8)
-      .attr(GorillaTacticsAbAttr),
+      .attr(GorillaTacticsAbAttr)
+      .edgeCase(),
+      // TODO: Verify whether Gorilla Tactics increases struggle's power or not
```
```suggestion
      // TODO: Verify whether Gorilla Tactics increases struggle's power or not
      .edgeCase(),
```

src/field/pokemon.ts
```diff
@@ -6757,10 +6800,9 @@ interface IllusionData {
 export interface TurnMove {
   move: MoveId;
   targets: BattlerIndex[];
+  useType: MoveUseType;
```
```suggestion
  mode: MoveUseMode;
```

src/field/pokemon.ts
```diff
@@ -6779,6 +6821,12 @@ export interface AttackMoveResult {
 export class PokemonSummonData {
   /** [Atk, Def, SpAtk, SpDef, Spd, Acc, Eva] */
   public statStages: number[] = [0, 0, 0, 0, 0, 0, 0];
+  /**
+   * A queue of moves yet to be executed, used by charging, recharging and frenzy moves.
+   * So long as this array is nonempty, this Pokemon's corresponding `CommandPhase` will be skipped over entirely
+   * in favor of using the queued move.
+   * TODO: Clean up a lot of the code surrounding the move queue. It's intertwined with the
```
```suggestion
   * TODO: Clean up a lot of the code surrounding the move queue.
```

src/field/pokemon.ts
```diff
@@ -6204,21 +6235,21 @@ export class EnemyPokemon extends Pokemon {
...
+    for (const [i, queuedMove] of moveQueue.entries()) {
+      const moveIndex = this.getMoveset().findIndex(m => m.moveId === queuedMove.move);
+      // If the queued move was called indirectly, ignore all PP and usability checks.
+      // Otherwise, ensure that the move being used is actually usable
+      // TODO: Virtual moves shouldn't use the move queue
+      if (
+        isVirtual(queuedMove.useType) ||
+        (moveIndex > -1 && this.getMoveset()[moveIndex].isUsable(this, isIgnorePP(queuedMove.useType)))
+      ) {
+        moveQueue.splice(0, i); // TODO: This may be redundant and definitely should not be done here
```
DayKev: Shouldn't this be `.splice(i - 1, i)`?

src/phases/move-charge-phase.ts
```diff
@@ -10,21 +10,32 @@ import { MovePhase } from "#app/phases/move-phase";
...
  */
 export class MoveChargePhase extends PokemonPhase {
   /** The move instance that this phase applies */
   public move: PokemonMove;
   /** The field index targeted by the move (Charging moves assume single target) */
   public targetIndex: BattlerIndex;
 
-  constructor(battlerIndex: BattlerIndex, targetIndex: BattlerIndex, move: PokemonMove) {
+  /** The {@linkcode MoveUseType} of the move that triggered the charge; passed on from move phase */
+  private useType: MoveUseType;
```
DayKev:
```suggestion
  private useMode: MoveUseMode;
```
Plus the rest of the file

src/phases/move-charge-phase.ts
```diff
@@ -53,29 +65,28 @@ export class MoveChargePhase extends PokemonPhase {
   /** Checks the move's instant charge conditions, then ends this phase. */
   public override end() {
     const user = this.getUserPokemon();
-    const move = this.move.getMove();
+    const move = this.move.getMove() as ChargingMove;
```
DayKev, line 69:
```suggestion
    // Checked for `ChargingMove` in `this.start()`
    const move = this.move.getMove() as ChargingMove;
```

src/ui/fight-ui-handler.ts
```diff
@@ -138,51 +139,58 @@ export default class FightUiHandler extends UiHandler implements InfoToggle {
...
 
-    if (button === Button.CANCEL || button === Button.ACTION) {
-      if (button === Button.ACTION) {
-        if ((globalScene.getCurrentPhase() as CommandPhase).handleCommand(this.fromCommand, cursor, false)) {
+    switch (button) {
+      case Button.CANCEL:
+        {
+          // Attempts to back out of the move selection pane are blocked in certain MEs
+          const { battleType, mysteryEncounter } = globalScene.currentBattle;
+          if (battleType === BattleType.MYSTERY_ENCOUNTER || !mysteryEncounter?.skipToFightInput) {
```
DayKev: You swapped the equality check when moving this.
```suggestion
          if (battleType !== BattleType.MYSTERY_ENCOUNTER || !mysteryEncounter?.skipToFightInput) {
```

test/moves/dig.test.ts
```diff
@@ -53,9 +53,24 @@ describe("Moves - Dig", () => {
     await game.phaseInterceptor.to("TurnEndPhase");
     expect(playerPokemon.getTag(BattlerTagType.UNDERGROUND)).toBeUndefined();
     expect(enemyPokemon.hp).toBeLessThan(enemyPokemon.getMaxHp());
+    expect(playerPokemon.getMoveQueue()).toHaveLength(0);
     expect(playerPokemon.getMoveHistory()).toHaveLength(2);
+  });
 
-    const playerDig = playerPokemon.getMoveset().find(mv => mv && mv.moveId === MoveId.DIG);
+  it("should deduct PP only on the 2nd turn of the move", async () => {
```
DayKev: Bulbapedia indicates it should work this way but that's not how it works on Showdown. Do you have a cart test somewhere?

test/moves/disable.test.ts
```diff
@@ -74,81 +81,99 @@ describe("Moves - Disable", () => {
...
+  });
 
-  it("disables NATURE POWER, not the move invoked by it", async () => {
-    game.override.enemyMoveset([MoveId.NATURE_POWER]);
-    await game.classicMode.startBattle();
+  it.each([
+    { name: "Nature Power", moveId: MoveId.NATURE_POWER },
+    { name: "Mirror Move", moveId: MoveId.MIRROR_MOVE },
+    { name: "Copycat", moveId: MoveId.COPYCAT },
+    { name: "Metronome", moveId: MoveId.METRONOME },
```
DayKev, line 130: You should probably override the called move from Metronome (and Nature Power?) so that it doesn't pick something that causes the test to fail.

DayKev (commented):
image

src/data/abilities/ability.ts
Bertie690: Does it matter whether it's single line or block?

src/field/pokemon.ts
Bertie690: No? We remove the first item each time until the first item can be used or none remain
Pls check i don't have it open rn and may be dumbo

src/data/abilities/ability.ts
DayKev: Not really, but it should go above the `.edgeCase()`; I just prefer the TSDoc style to only be used on actual TSDocs so I changed it at the same time.

src/phases/move-charge-phase.ts
Bertie690: ok I'll do it myself in a min and then resolve em

src/ui/fight-ui-handler.ts
Bertie690: mb

test/moves/dig.test.ts
Bertie690: not yet...
Skip the test for now?

test/moves/disable.test.ts
Bertie690: Such an override would be wholly redundant and irrelevant.
The only thing we're checking is "is the called move (whatever the hell it is) being disabled". 
What move it is makes no bearing on the result as it shouldn't even be CONSIDERED for disable.

src/data/abilities/ability.ts
Bertie690: This isn't tsdoc...?
It's a block comment with bullet points lol

src/data/abilities/ability.ts
Bertie690: you want this comment above the edge case but not the other?
IDRC myself but we should be consistent in the ENTIRE codebase fwiw

src/phases/move-charge-phase.ts
Bertie690: @DayKev do i rename the file as well?

src/field/pokemon.ts
DayKev: After taking another look I think it might need to be `.splice(0, 1)` (otherwise you're removing more and more entries each time as you traverse the array).

src/phases/move-charge-phase.ts
DayKev: Yeah, `move-use-mode.ts`.

src/field/pokemon.ts
Bertie690: Oh right brain fart duhrrrrr

src/data/abilities/ability.ts
DayKev: `/** */` is a [TS|JS]Doc comment block. `/* */` is a block comment.

src/data/abilities/ability.ts
DayKev: It should always be above (beside is okay, but can become a problem if the comment is too long).

test/moves/disable.test.ts
DayKev: What if it Metronomes into a move that prevents the player Pokémon from [successfully] using Disable? That's the issue.

src/data/abilities/ability.ts
Bertie690: Ok that'll be my next PR lol

src/data/abilities/ability.ts
Bertie690: it is a block comment lmfao
```ts
/* Incorrect interations with:
* Petal Dance (should not lock in or count down timer; currently does both)
* Flinches (due to tag being removed earlier)
* Failed/protected moves (should not trigger if original move is protected against)
*/
```

DayKev: Ah, I'm so used to everyone always doing `/**` for block comments I just automatically saw it as that lmao.

src/phases/move-phase.ts
```diff
@@ -210,95 +212,110 @@ export class MovePhase extends BattlePhase {
...
-          CommonAnim.POISON + (this.pokemon.status.effect - 1),
-        );
-      } else if (healed) {
-        globalScene.phaseManager.queueMessage(
-          getStatusEffectHealText(this.pokemon.status.effect, getPokemonNameWithAffix(this.pokemon)),
+    if (
+      this.useMode === MoveUseMode.INDIRECT &&
+      [StatusEffect.SLEEP, StatusEffect.FREEZE].includes(this.pokemon.status.effect)
+    ) {
+      // Dancer thaws out or wakes up a frozen/sleeping user prior to use
```
DayKev, line 210: Dancer is the only way this code path can possibly be reached?

src/phases/move-phase.ts
```diff
@@ -76,95 +78,95 @@ export class MovePhase extends BattlePhase {
...
   public start(): void {
     super.start();
 
-    console.log(MoveId[this.move.moveId]);
+    // Cover our bases - if we have no use type
+    if (!this.useMode) {
+      console.warn(`Unexpected MoveUseMode of ${this.useMode} during move phase!`);
+      // @ts-expect-error - useMode is readonly and shouldn't normally be assigned to
+      this.useMode = MoveUseMode.NORMAL;
+    }
```
DayKev: Is this really necessary? It's already a required param, you're basically protecting from JS spontaneously combusting or something.
```suggestion
```
(Deleted and reposted because GitHub was being weird).
Edit: GitHub is still being weird and not showing it in the file view... whatever.

DayKev (dismissed):

SirzBenjie (dismissed):
Still looking ok.

Should make sure the game runs as normal with this up.

DayKev (approved):

DayKev (dismissed):

damocleas (dismissed):

PR #5818, opened 2025-05-12
[Bug][UI/UX] Fix faint phase order

SirzBenjie:
A move that causes a target to faint will no longer have the target faint before other messages related to the move are displayed.

~~Fixes a bug caused by #5678~~
Was not actually caused by that PR.

Added a new private field to the move effect phase, `faintPhases`, that stores the faint phases that will be added by the move.
Instead of immediately unshifting the faint phase when the target faints, the faint phase is pushed to this array. Then, each of the faint phases are unshifted after all of the move effects could occur, inside of the `applyToTargets` method.

This ensures that the faint phases are unshifted _after_ all of the phases that are enqueued from a move's use (like the self-stat increase). Note that this must happen before POST_TARGET effects to ensure that self-destruct's self-faint occurs after the targets hit by the move faint. This is why the unshifting is done in the `applyToTargets` method.

## Screenshots/Videos

<details><summary>Current hotfix branch</summary>

https://github.com/user-attachments/assets/0d36b1bd-66c9-43d3-8003-5834c5ce4ea7
</details>

<details><summary>After fix</summary>

https://github.com/user-attachments/assets/502e1d6e-9d8a-46e6-bb5d-7f1e847e149b
</details>

## How to test the changes?
I use these overrides:
```ts
const overrides = {
  MOVESET_OVERRIDE: Moves.MYSTICAL_POWER,
  STARTING_LEVEL_OVERRIDE: 100
```

Just go into battle and use mystical power, and note that the stat boost message now properly appears before the target's faint animation has played.

## Checklist
- ~~[ ] **I'm using `beta` as my base branch**~~ Hotfix

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "pokemon-rogue-battle",
-  "version": "1.9.2",
+  "version": "1.9.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "pokemon-rogue-battle",
-      "version": "1.9.2",
+      "version": "1.9.3",
       "hasInstallScript": true,
       "dependencies": {
         "@material/material-color-utilities": "^0.2.7",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "pokemon-rogue-battle",
   "private": true,
-  "version": "1.9.2",
+  "version": "1.9.3",
   "type": "module",
   "scripts": {
     "start": "vite",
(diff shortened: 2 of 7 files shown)


SirzBenjie:
Actually, I just checked what happens on cartridge, and the stat drops do indeed happen *after* the mon faints. This means that this is not, in fact a bug, and actually the behavior before the MEP was improper. Who would've thought!

SirzBenjie:
The fix for this requires larger changes involving organizing the sequence of events for recoil and other things. This is a larger change.

PR #5821, opened 2025-05-14
[UI/UX] Party UI handler refactor (part I)

Wlowscha:
Hopefully none---this is supposed to be strictly a refactor.

The Party UI handler, like many UI scripts, is a behemoth and very difficult to parse. It is called from a variety of places in the code with different modes, callbacks and filters. These various options are processed in functions with multiple nested conditionals which make it very difficult to understand at a glance what is going on for a given option.

Take for example this line:
```
            } else if (this.cursor) {
              (globalScene.getCurrentPhase() as CommandPhase).handleCommand(
                Command.POKEMON,
                this.cursor,
                option === PartyOption.PASS_BATON,
              );
            }
```
This code snippet handles switching a Pokémon out when the `PartyUiMode` is `SWITCH `and the `PartyOption `is either `PASS_BATON `or `SEND_OUT`, a situation corresponding to switching out using the Pokémon command in battle (possibly holding the Baton item). This code does NOT apply to using the Baton Pass move (which is handled elsewhere, under a `PartyUiMode` named `FAINT_SWITCH`) nor does it apply to the optional switching at the beginning of a battle (handled by the `PartyUiMode `named `POST_BATTLE_SWITCH`). None of this information is apparent from the code. In fact, the reason this code block is executed is that, in this case, the `selectCallback `function is null, while it isn't when using `BATON_PASS`. Moreover, the correct functioning of the code also requires to call `this.clearOptions()`, which is done in a different place.

The aim of the refactor is to replace this with something like
```
      // This is used when switching out using the Pokemon command (possibly holding a Baton held item). In this case there is no callback.
      if ((option === PartyOption.PASS_BATON || option === PartyOption.SEND_OUT) && this.partyUiMode === PartyUiMode.SWITCH) {
        this.clearOptions();
        (globalScene.getCurrentPhase() as CommandPhase).handleCommand(
          Command.POKEMON,
          this.cursor,
          option === PartyOption.PASS_BATON,
        );
      }
```

## Why is (X) not included in this refactor?

This refactor is intended to be limited solely to the PartyUiHandler file. It is going to be followed by another PR focusing on renaming some variables, and possibly changing the way inputs are handled, to make e.g. passing callbacks to the handler more expressive.

Moving the PartySlot class to its own file is desirable, however doing so also requires moving the PartyUiMode enum to its own file (to avoid circular imports). Given that a bunch of things depend on it, this is also deferred to the following refactor PR.

There are two major changes in this PR:

1) `updateOptions`, which is responsible for assigning menu options to each `PartyUiMode`, has been heavily reorganized so that the list of options is written down immediately after the mode itself. For commonly shared options (like summary and pokedex), the `addCommonOptions()` function has been introduced. The only exception is the cancel option, plus any scrolling, which are the same for all modes and now handled by `addCancelAndScrollOptions`().

The logic for actually displaying the options (converting them to text, etc) has been moved to updateOptionsWindow() but not touched otherwise. 

2) `processInput`, which had a complexity of 204, has been broken up into various function. It checks whether the option menu is open and, if so, it outsources handling the input to `processOptionMenuInput`. The logic in the latter function has been heavily reorganized in order to better display the code executed depending on the selected option.

The logic for processing input when the menu is closed has not been changed yet.

## How to test the changes?
For switching, this override is useful:
```
const overrides = {
  STARTING_LEVEL_OVERRIDE: 200,
  ITEM_REWARD_OVERRIDE: [{name: "BATON"}],
  MOVESET_OVERRIDE: [Moves.BATON_PASS],
```
Other miscellaneous:
```
const overrides = {
  ITEM_REWARD_OVERRIDE: [{name: "DNA_SPLICERS"}],
  MOVESET_OVERRIDE: [Moves.REVIVAL_BLESSING],
```

Apart from these, any manual testing of the party ui during a run is useful. I have tested manually the following:
- Manual switch (done)
- Manual switch using baton (done)
- Switch after faint (done)
- Use Baton Pass (done)
- Use Revive (done)
- Use Revival Blessing (done)
- Use Ether (done)
- Apply healing item (done)
- Teach move with TM (done)
- Transfer held items (done)
- Summary, Pokedex, Unpause Evolution work (done)
- Splice / unsplice (done)
- Release/rename work (done)
But it is of course possible that I missed some edge cases.

In one instance, the summary menu seemed to be behaving weirdly (not showing the right sprite when scrolling up and down). However, I was not able to replicate this later.

## Checklist


src/ui/party-ui-handler.ts
```diff
@@ -355,6 +356,519 @@ export default class PartyUiHandler extends MessageUiHandler {
...
+  processPokedexOption(pokemon: Pokemon): boolean {
+    const ui = this.getUi();
+    ui.playSelect();
+    const attributes = {
+      shiny: pokemon.shiny,
+      variant: pokemon.variant,
+      form: pokemon.formIndex,
+      female: pokemon.gender === Gender.FEMALE,
+    };
+    ui.setOverlayMode(UiMode.POKEDEX_PAGE, pokemon.species, attributes).then(() => this.clearOptions());
```
SirzBenjie, line 375: This may not be the fault of the PR, but why are we keeping around dangling promises?
`.then`. means that `setOverlayMode` is an async method.
`processPokedexOptions` should therefore also be async.

SirzBenjie (commented):
I reviewed this, and the reorganization looks good.

I did not go over it with a fine-tooth comb, so I haven't verified that the logic remains exactly the same as it was before.
However, I did go over the code that was added, and it is substantially easier to reason about than what it was before. I have a good understanding of what is going on.

I have one comment about the async method that I think should be addressed.

src/ui/party-ui-handler.ts
Wlowscha: Does this comment also apply to `processSummaryOption` and `processUnspliceOption` then?

SirzBenjie (approved):

xsn34kzx (dismissed):

src/ui/party-ui-handler.ts
```diff
@@ -356,6 +357,519 @@ export default class PartyUiHandler extends MessageUiHandler {
...
+    if (option === PartyOption.UNSPLICE) {
+      return this.processUnspliceOption(pokemon);
+    }
+    if (option === PartyOption.RENAME) {
+      return this.processRenameOption(pokemon);
+    }
+    // This is only relevant for PartyUiMode.CHECK
+    // TODO: This risks hitting the other options (.MOVE_i and ALL) so does it? Do we need an extra check?
+    if (option >= PartyOption.FORM_CHANGE_ITEM && globalScene.getCurrentPhase() instanceof SelectModifierPhase) {
+      if (this.partyUiMode === PartyUiMode.CHECK) {
```
DayKev, line 752: This can just be part of the previous `if`, there's no need to nest them like this.

DayKev (commented):
Can you add `public/protected/private` to all the methods you added/changed?
---
Are there any logic changes or is this supposed to be purely moving code around?

Wlowscha:
> Can you add `public/protected/private` to all the methods you added/changed?
> 
> Are there any logic changes or is this supposed to be purely moving code around?

Added `private` to all the new methods.

It is supposed to be only moving code around, without changes in the logic. If there are any, they are not intentional.

DayKev (dismissed):

PR #5822, opened 2025-05-15
[Bug] IVs of trainers were incorrectly using the battle seed

### LINKED ISSUE #5820: [Bug] battleSeedState and rngCounter incorrectly incremented by IVs of trainers
### Describe the bug

Battle rng is completely different after the first reload. Which means that the enemy can do wildly different actions after the first reload.

In `Battle.ts` `randSeedInt()` the `this.battleSeedState` and `this.rngCounter` values are not properly reset or are not properly reloaded from session. This results in a different rng state after first reload.

### Reproduction

```ts
const overrides = {
  STARTING_WAVE_OVERRIDE: 5,
  DAILY_RUN_SEED_OVERRIDE: "Cv2MM6b3f/uNO36HFuyqUQ==",
```
- New Game > Daily Run
- Wave 5 trainer Houndoom selects Incinerate.
- Reload page and continue
- Wave 5 trainer Houndoom now selects Bite.

Reason:
`this.battleSeedState` is now undefined after reload and `this.rngCounter` went from 12 pre-Reload to 0 post-Reload.

### Expected behavior

When i checkout v1.8.5, the values are properly rese and `this.rngCounter` is only set to 0 once, on new class. So my assumption currently is that the battle class is not properly reset when the battle starts. Any rng rolls earlier in the wave, like generating encounters is affecting it now.

### Screenshots / Videos

_No response_

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

Jimmybald1:
Fixes #5820 

Reloading should not change the outcome of a battle anymore.

It took the better part of a year to eliminate all the rng changes that reloading had. They are now 99.99% gone, but one incorrect use of the battle seed and it was back worse than before.
This PR fixes the IVs calling the wrong seed function. But it also includes a rename of the battle seed function. This should hopefully prevent any incorrect calls in the future.

The battle seed functions should only be used for battle rng, never for pokemon generation. The rename of the functions should at least show developers that they are calling a different function than the global seed functions in common.ts.

Also, apparently biome didnt like trainer-config... I only changed the randSeedInt to randSeedIntRange there.

## Screenshots/Videos
Before this PR:

https://github.com/user-attachments/assets/28869540-fb3d-40fd-b2da-1d31f58d1869

After this PR:

https://github.com/user-attachments/assets/8f1274ac-aab1-4062-a23f-ca892ca83b89

See overrides below.
This puts you in a trainer fight where the enemy Houndoom either chooses Incinerate or Bite, depending on reloading. Without reload it chooses Incinerate, because there were 2x 6 IV calls to the battle seed while generating the enemies. After your first reload, it pulls the enemies from your session and does not have these IV calls. Now it chooses to use Bite.

After the fix, the Houndoom always uses bite, because the IV calls are not happening anymore.

## How to test the changes?
Example used in the video:
```ts
const overrides = {
  STARTING_WAVE_OVERRIDE: 5,
  DAILY_RUN_SEED_OVERRIDE: "Cv2MM6b3f/uNO36HFuyqUQ==",
```

## Checklist


Wlowscha (approved):
I've gone through and all looks good to me.

DayKev (approved):

PR #5829, opened 2025-05-15
[Bug] Fix Substitute sprite crash & revived Pokemon softlock

### LINKED ISSUE #5807: [Bug] After reload, Substitute is still active but has no sprite, causing a crash.
### Describe the bug

If you have a substitute still active from the previous wave and you reload the page. The sub wont be created properly and cause crashes/freezes with errors when it tries to interact with the sub sprite.

### Reproduction

Use Substitute (keep it alive).
Complete the wave.
On next wave reload the page.
When youre back you dont see a Substitute sprite anymore, but when you get hit it does still hit and fade your substitute.
If the Substitute never fades and you complete the wave again, it will cause a crash trying to interact with the sub sprite.

(so far i only encountered the bug when i also switched biome after the invisible sub happened)

```ts
const overrides = {
  STARTING_WAVE_OVERRIDE: 9,
  MOVESET_OVERRIDE: [Moves.SUBSTITUTE, Moves.TACKLE],
  OPP_MOVESET_OVERRIDE: [Moves.SPLASH],
```

### Expected behavior

Subsitute should load its sprite properly on reloads.

### Screenshots / Videos

Image

Image

Image

### Session export file

Currently already in the state with a spriteless Substitute:
[Substitute crash.txt](https://github.com/user-attachments/files/20123815/Substitute.crash.txt)

### User data export file

_No response_

### Additional context

Discord bug report: https://discord.com/channels/1125469663833370665/1370377949228040223

When i get hit and let the spriteless sub fade, i can proceed without errors like normal.

emdeann:
Reloading with a substitute will load the sprite correctly instead of crashing

Pokemon which are revived and immediately sent back out will not be treated as if they are switching out (resulting in a softlock)

Fixes #5807 

`resetSummonData` was doing some setup completely unrelated to summonData - when calls to it were removed from the `SummonPhase`, things like the sub sprite were never set properly. I split it into two methods - one which actually resets the summon data and one that does the misc field setup. 

I more or less took educated guesses on which method should be called where (because everything was calling `resetSummonData` previously, even if only one of the two behaviors of that method were needed). I haven't noticed anything terribly wrong in testing but when reviewing you should make sure the calls are logical.

## Screenshots/Videos

<details><summary>Substitute</summary>
<p>

https://github.com/user-attachments/assets/643ce639-e16e-47d4-b138-90e1e5bd2a2d

</p>
</details> 

<details><summary>Revive</summary>
<p>

https://github.com/user-attachments/assets/efaf4292-af72-4c96-912a-fd96c068ccff

</p>
</details> 

## How to test the changes?
Substitute - clear a wave with substitute up, then reload and ensure that the sprite is correctly displayed and does not crash

Revive - override the first pokemon in your party to use explosion, use it to ko the opponent and then use a revive on that pokemon immediately after. Check that the pokemon is able to use its moves as expected in the next wave

## Checklist


Wlowscha (approved):

SirzBenjie (approved):
This makes sense to me.

Played about 50 waves on the branch and didn't run into any issues.

Madmadness65 (approved):

PR #5837, opened 2025-05-18
[Bug] Prevent Sleep Talk from succeeding on the wake up turn

schmidtc1:

Sleep talk will no longer succeed on the wake up turn.

Bug

Removes conditional in apply, moves it to userSleptOrComatoseCondition, adds check for sleep turns remaining

## Screenshots/Videos

https://github.com/user-attachments/assets/d77bca31-c07c-4415-b27f-92c08abd2d25

## How to test the changes?

Use src/overrides, give mon Sleep Talk, Rest, and a self-damaging move. Observe Sleep Talk on wake up turn no longer succeeds.

## Checklist
  - [n/a] Have I made sure that any UI change works for both UI themes (default and legacy)?

- [n/a] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?
  - [n/a] If so, please leave a link to it here: 
- [n/a] Has the translation team been contacted for proofreading/translation?


Snailman11:
Link to Discord's Bug report
https://discord.com/channels/1125469663833370665/1369119275847127051

SirzBenjie (commented):
Unfortunately, I don't think that this way of fixing it will actually drill down at the real issue.

The issue seems to be _when_ this condition is checked, which I believe occurs at the start of the move phase, but before the check to wake the user up has occurred.

Adding a check against the user being asleep and having sleep turns remaining is unintuitive.
I don't think this is the fault of the PR.

Also, this does not seem like it would fix the bug where rest cannot be used the turn that the user wakes up.

schmidtc1:
> Unfortunately, I don't think that this way of fixing it will actually drill down at the real issue.
> 
> The issue seems to be _when_ this condition is checked, which I believe occurs at the start of the move phase, but before the check to wake the user up has occurred.
> 
> Adding a check against the user being asleep and having sleep turns remaining is unintuitive. I don't think this is the fault of the PR.
> 
> Also, this does not seem like it would fix the bug where rest cannot be used the turn that the user wakes up.

Revised my approach to target fixing the order in which wake up/condition is checked as you recommended:
`ResetStatus` was defaulting to queuing a phase, which is why wake up occurred after the condition was checked. By not queuing a phase, status is cleared once `healed` is set to true. Move condition is then checked afterwards. Tests seem to still be passing.

Tested Rest on wake up turn and seems to now be working. This approach also more accurately reflects the mainline game's sleep talk move, as prior to this change, we were not printing "User is fast sleep. User used sleep talk." We were just printing "User used sleep talk." and skipping the fast asleep message.

schmidtc1:
Closing out in favor of https://github.com/pagefaultgames/pokerogue/pull/5872, which is more extensive

PR #5868, opened 2025-05-23
[Misc] Add `phase#is` method to help reduce circular imports

SirzBenjie:
None

Many circular import cycles occur due to phases importing each other in some way. In many cases, this is entirely due to `instanceof` checks.

First, I moved the phase utility types out of `PhaseInterceptor` and into `phase-types.ts` so that they can be reused.

Adds a new method to `Phase`, called `is` that allows for files to check if a phase is a specific phase without having to import the phase itself to use the `instanceof` check.

I also converted a few phase classes into abstract classes that should have already been marked as such.

I added a new property, `phaseName` to each phase that is just the string name of the phase.

I then replaced all of the instances of `instanceof <specificPhase>` with a call to this method.

Note that this is _not_ the exact same as an `instanceof` check in the case that a phase is subclassed, as it checks the *exact* match of the phase name. This is OK for all of the existing uses of `instanceof` as they were never checking for a phase that could be subclassed.

This reduces circular imports from 585 to 574.

Note that the method also works as a type predicate to narrow the type of the var passed in in the same way that an instanceof check did.

## Screenshots/Videos
N/A

## How to test the changes?
N/A

## Checklist
- [x] Have I tested the changes manually? I did about 50 waves on super speed. Nothing out of the ordinary happened.

### DIFF
--- a/src/phases/encounter-phase.ts
+++ b/src/phases/encounter-phase.ts
@@ -47,6 +47,7 @@ import { WEIGHT_INCREMENT_ON_SPAWN_MISS } from "#app/data/mystery-encounters/mys
 import { getNatureName } from "#app/data/nature";
 
 export class EncounterPhase extends BattlePhase {
+  protected readonly phaseName = "EncounterPhase";
   private loaded: boolean;
 
   constructor(loaded = false) {
--- a/src/phases/encounter-phase.ts
+++ b/src/phases/encounter-phase.ts
@@ -47,7 +47,7 @@ import { WEIGHT_INCREMENT_ON_SPAWN_MISS } from "#app/data/mystery-encounters/mys
 import { getNatureName } from "#app/data/nature";
 
 export class EncounterPhase extends BattlePhase {
-  // Union type is necessary as these
+  // Union type is necessary as this is subclassed, and typescript will otherwise complain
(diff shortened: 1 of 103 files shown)


src/phases/encounter-phase.ts
```diff
@@ -47,6 +47,8 @@ import { WEIGHT_INCREMENT_ON_SPAWN_MISS } from "#app/data/mystery-encounters/mys
 import { getNatureName } from "#app/data/nature";
 
 export class EncounterPhase extends BattlePhase {
+  // Union type is necessary as these
```
DayKev, line 50: Btw you forgot to finish this comment.

src/phases/move-anim-phase.ts
```diff
@@ -12,6 +12,8 @@ export class MoveAnimPhase<Anim extends MoveAnim> extends Phase {
     super();
   }
 
+  protected readonly phaseName = "MoveAnimPhase";
```
DayKev, line 15: This should probably go before the constructor.

src/phases/encounter-phase.ts
SirzBenjie: Fixed

PR #5869, opened 2025-05-24
[Bug] Protect rng now resets on new waves and fixed to look at all turns in the same wave.

Jimmybald1:
Protect is no longer stuck at maximum 1 in 3 odds. So chaining multiple protects in a row will be far less likely.
Protect is now also guaranteed to succeed on the first turn the pokemon is out. It properly resets the rng on each new wave.

Protect was stuck on maximum 1 in 3 odds instead of becoming 3 times less likely on each subsequent protect. And in discussion with balance team, protect rng should reset on each new wave.

-

## How to test the changes?
```ts
const overrides = {
  SEED_OVERRIDE: "eUv9ArUWreKyj4MbtnNR8gmI",
  BATTLE_STYLE_OVERRIDE: "double",
  OPP_MOVESET_OVERRIDE: [Moves.TACKLE],
  MOVESET_OVERRIDE: [Moves.PROTECT, Moves.DETECT, Moves.CORE_ENFORCER],
  STARTING_WAVE_OVERRIDE: 2,
```

## Checklist

### DIFF
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -5780,20 +5780,35 @@ export class ProtectAttr extends AddBattlerTagAttr {
 
   getCondition(): MoveConditionFunc {
     return ((user, target, move): boolean => {
+      // Protect rng resets on new waves, it always succeeds.
+      if (user.tempSummonData.waveTurnCount === 1) {
+        return true;
+      }
+
       let timesUsed = 0;
-      const moveHistory = user.getLastXMoves();
+      const moveHistory = user.getLastXMoves(-1);
       let turnMove: TurnMove | undefined;
 
       while (moveHistory.length) {
         turnMove = moveHistory.shift();
+
         if (!allMoves[turnMove?.move ?? Moves.NONE].hasAttr(ProtectAttr) || turnMove?.result !== MoveResult.SUCCESS) {
           break;
         }
+
         timesUsed++;
+
+        // Break after first move used this wave.
+        // If no move was used on turn 1, then it would have broken in the attr check already.
+        if (turnMove?.turn === 1) {
+          break;
+        }
       }
+
       if (timesUsed) {
         return !user.randBattleSeedInt(Math.pow(3, timesUsed));
       }
+
       return true;
     });
   }


DayKev (approved):

damocleas (approved):

Jimmybald1:
Still an issue where pokeballs increment the turn count without any log in movehistory, causing the protect check to sometimes see previous waves moves as the last used move. This has to be fixed first.

Jimmybald1:
Fixed the additional issues now and thanks to Matthew/Bertie for helping out and providing a lot of tests.

src/data/moves/move.ts
```diff
@@ -5781,20 +5781,22 @@ export class ProtectAttr extends AddBattlerTagAttr {
...
 
-      while (moveHistory.length) {
-        turnMove = moveHistory.shift();
-        if (!allMoves[turnMove?.move ?? Moves.NONE].hasAttr(ProtectAttr) || turnMove?.result !== MoveResult.SUCCESS) {
+      for (const turnMove of user.tempSummonData.waveMoveHistory.reverse()) {
+        if (
+          // Quick & Wide guard increment the Protect counter without using it for fail chance
+          !(allMoves[turnMove.move].hasAttr(ProtectAttr) || 
+          [Moves.QUICK_GUARD, Moves.WIDE_GUARD].includes(turnMove.move)) || 
+          turnMove?.result !== MoveResult.SUCCESS
```
DayKev, line 5790:
```suggestion
          turnMove.result !== MoveResult.SUCCESS
```

src/data/moves/move.ts
```diff
@@ -5781,20 +5781,22 @@ export class ProtectAttr extends AddBattlerTagAttr {
   getCondition(): MoveConditionFunc {
     return ((user, target, move): boolean => {
       let timesUsed = 0;
-      const moveHistory = user.getLastXMoves();
-      let turnMove: TurnMove | undefined;
 
-      while (moveHistory.length) {
-        turnMove = moveHistory.shift();
-        if (!allMoves[turnMove?.move ?? Moves.NONE].hasAttr(ProtectAttr) || turnMove?.result !== MoveResult.SUCCESS) {
+      for (const turnMove of user.tempSummonData.waveMoveHistory.reverse()) {
```
DayKev, line 5785:
```suggestion
      for (const turnMove of user.tempSummonData.waveMoveHistory.slice().reverse()) {
```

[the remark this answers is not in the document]
DayKev: > Why are you trying to reverse it in the first place?

It walks the move history backwards until it finds a non-Protect move and then stops at that point, ignoring the rest of the move history, since Protect doesn't care about moves that happened before that point.

It should be doing `.slice().reverse()` like `Pokemon#getLastXMoves` though to make a copy of the array instead of reversing the original.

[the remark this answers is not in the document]
emdeann: I know that the intent is to look through the history backwards, my point is that you can just walk it backwards to begin with instead of reversing it and then walking it forwards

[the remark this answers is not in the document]
DayKev: How so? I didn't see any "reverse walking" functions here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

damocleas (approved):

DayKev (approved):

src/data/moves/move.ts
```diff
@@ -5781,20 +5781,22 @@ export class ProtectAttr extends AddBattlerTagAttr {
...
-        timesUsed++;
-      }
-      if (timesUsed) {
-        return !user.randBattleSeedInt(Math.pow(3, timesUsed));
+
+        timesUsed++
       }
-      return true;
+
+      // console.log(`Wave Move History: ${user.tempSummonData.waveMoveHistory.reverse().map(t => t.move)}\nTimes Used In Row: ${timesUsed}\nSuccess chance: 1 in ${Math.pow(3, timesUsed)}`)
```
xsn34kzx:
```suggestion
```
This comment should get axed, no?

src/data/moves/move.ts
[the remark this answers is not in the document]
emdeann: Probably C-style like ```for (let i = arr.length - 1; i >= 0; i--)```, it's less idiomatic I guess though? It just doesn't make sense to me to reverse the entire array to be able to iterate it backwards. again though not like it would end up mattering in this case

src/data/moves/move.ts
Jimmybald1: I thought it was useful for any future tester and reviewer to have easy access to it, I believe theres more instances of these kind of commented console logs?
It can be removed if other devs also dont like it. I dont mind much either way.

damocleas (approved):

DayKev (approved):

src/phases/battle-end-phase.ts
```diff
@@ -58,12 +58,6 @@ export class BattleEndPhase extends BattlePhase {
       globalScene.unshiftPhase(new GameOverPhase(true));
     }
 
-    for (const pokemon of globalScene.getField()) {
-      if (pokemon) {
-        pokemon.tempSummonData.waveTurnCount = 1;
-      }
-    }
```
SirzBenjie, line 65: Can you explain why removing this is necessary and what it impacts?
Or is it guaranteed that the resetWaveData is already being called?

src/data/moves/move.ts
SirzBenjie: It is true that the codebase has these latent comments. I don't particularly think they belong in the code; it doesn't seem to be that bad to recreate the item being logged. I would prefer it be removed.

src/data/moves/move.ts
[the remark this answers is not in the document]
SirzBenjie: It's worth noting that this is how we usually go through reversed arrays in the codebase. @emdeann is right that doing it that way is less efficient, as it has to allocate a new copy of the array.
If the contents of the array can't change, then there's no reason not to do the c-style loop.

Fontbane:
Disagree with the balance decision here. Not a fan of temp summon data in general.

src/phases/battle-end-phase.ts
DayKev: He moved it to `Pokemon#resetWaveData`, which should always be called when this would be iirc.

Bertie690: Correct. Felt a bit random that this of all things was being reset separately.

test/moves/protect.test.ts
```diff
@@ -26,92 +25,205 @@ describe("Moves - Protect", () => {
...
+    expect(charizard);
   });
 
-  test("should protect the user from attacks", async () => {
+  it.each<{ numTurns: number; chance: number }>([
+    { numTurns: 1, chance: 3 },
+    { numTurns: 2, chance: 9 },
+    { numTurns: 3, chance: 27 },
+    { numTurns: 4, chance: 81 },
+  ])("should have a 1/$chance success rate after $times successful uses", async ({ numTurns, chance }) => {
```
Bertie690: I goofed on the message lol...
```suggestion
  ])("should have a 1/$chance success rate after $numTurns successful uses", async ({ numTurns, chance }) => {
```

Bertie690 (dismissed):

DayKev (dismissed):

DayKev:
Removed the "reset counter on every wave" behavior due to further balance team discussion.

DayKev (dismissed):

damocleas (approved):

Bertie690 (dismissed):

DayKev (approved):

Wlowscha (approved):

PR #5883, opened 2025-05-29
[Refactor] Cleanup fight UI

SirzBenjie:
None, well, maybe a miniscule performance increase from one less unnecessary tween being created each time the info overlay is toggled.

Part of a series of PRs I am working on to cleanup the mess in our UI files.

Make `battleScene#addInfotoggles` take in multiple toggles.

Made `processInput` use a switch for all cases (it randomly wasn't before).
Made Phaser objects use method chaining.
Fixed an unnecessary tween that was being created in the `toggleInfo` method when the visibility of the battle info overlay was toggled.

## Screenshots/Videos
N/A

## How to test the changes?
Start a game, select fight, and then make sure it works the same as it always has (it will).

## Checklist
- [x] Are all unit tests still passing? (`npm run test:silent`) Needs #5884 


Wlowscha (approved):

xsn34kzx (dismissed):

src/ui/fight-ui-handler.ts
```diff
@@ -231,6 +234,64 @@ export default class FightUiHandler extends UiHandler implements InfoToggle {
     return !this.fieldIndex ? this.cursor : this.cursor2;
   }
 
+  /**@returns TextStyle according to percentage of PP remaining */
```
DayKev:
```suggestion
  /** @returns TextStyle according to percentage of PP remaining */
```

DayKev (dismissed):

PR #5886, opened 2025-05-29
[Refactor] Refactor select-modifier-phase.ts

Wlowscha:
None at all.

This is another refactor of a UI file to make the code easier to parse and change. Not quite as big as the party ui handler refactor, and should be easier to review since the logic was not quite as tangled, so it's mostly just moving logic to appropriate smaller functions.

The complexity in this file revolves around `modifierSelectCallback`, a function defined inside the `.start()` method of the `SelectModifierPhase` class. The reason why this is a function, and not itself a method of `SelectModifierPhase`, is that it needs to be passed to other ui handlers so they know what to do when a selection is canceled (for example, when deciding against learning a TM move).

The `modifierSelectCallback `essentially takes a value for both `rowCursor `and `cursor`, finds the corresponding option, and executed. This is done through two switch statements. Each possible selection in the switch statements has been converted to its own function (except for the check party option, which is a single line). The new functions are:
```
rerollModifiers();
openModifierTransferScreen(modifierSelectCallback);
toggleRerollLock();
selectRewardModifierOption(cursor, modifierSelectCallback);
selectShopModifierOption(rowCursor, cursor, modifierSelectCallback);
```
The latter two options convert the modifier selected by the user into an object that can be applied, then call the new `applyChosenModifier()`, which takes care of distinguishing which effect needs to take place. If the party screen must be opened, it calls either `openFusionMenu()` or `openModifierMenu()`, otherwise it directly calls `applyModifier()`. There is some naming overlap here which may need to be untangled.

This logic:
```
    globalScene.ui.setMode(
      UiMode.MODIFIER_SELECT,
      this.isPlayer(),
      this.typeOptions,
      modifierSelectCallback,
      this.getRerollCost(globalScene.lockModifierTiers),
    );
```
appeared in several places throughout the code, it is now its own method called `resetModifierSelect()`.

## How to test the changes?
Start a new run, win some battles, and play around with the modifier options.

## Checklist


SirzBenjie (dismissed):

src/phases/select-modifier-phase.ts
```diff
@@ -98,259 +80,300 @@ export class SelectModifierPhase extends BattlePhase {
...
+      return false;
+    }
+
+    return this.applyChosenModifier(modifierType, cost, modifierSelectCallback);
+  }
+
+  // Apply a chosen modifier: do an effect or open the party menu
+  private applyChosenModifier(modifierType: ModifierType, cost: number, modifierSelectCallback): boolean {
+    if (modifierType! instanceof PokemonModifierType) {
+      //TODO: is the bang correct?
```
DayKev:
```suggestion
    if (modifierType instanceof PokemonModifierType) {
```

src/phases/select-modifier-phase.ts
```diff
@@ -98,259 +80,300 @@ export class SelectModifierPhase extends BattlePhase {
...
+      if (modifierType instanceof FusePokemonModifierType) {
+        this.openFusionMenu(modifierType, cost, modifierSelectCallback);
+      } else {
+        this.openModifierMenu(modifierType, cost, modifierSelectCallback);
       }
+    } else {
+      this.applyModifier(modifierType!.newModifier()!); // TODO: is the bang correct?
```
DayKev:
```suggestion
      this.applyModifier(modifierType.newModifier());
```

src/phases/select-modifier-phase.ts
```diff
@@ -98,259 +80,300 @@ export class SelectModifierPhase extends BattlePhase {
...
+    );
+  }
+
+  // Function that determines how many reward slots are available
+  private getModifierCount(): number {
+    const modifierCountHolder = new NumberHolder(3);
+    if (this.isPlayer()) {
+      globalScene.applyModifiers(ExtraModifierModifier, true, modifierCountHolder);
+      globalScene.applyModifiers(TempExtraModifierModifier, true, modifierCountHolder);
+    }
```
DayKev: If there's a `this.isPlayer()` check it should be done at the top of `start()` and end the phase early if somehow things are *that* screwed up that a `SelectModifierPhase` was pushed for an enemy.
```suggestion
    globalScene.applyModifiers(ExtraModifierModifier, true, modifierCountHolder);
    globalScene.applyModifiers(TempExtraModifierModifier, true, modifierCountHolder);
```

src/phases/select-modifier-phase.ts
```diff
@@ -98,259 +80,300 @@ export class SelectModifierPhase extends BattlePhase {
...
+
+    // If custom modifiers are specified, overrides default item count
+    if (this.customModifierSettings) {
+      const newItemCount =
+        (this.customModifierSettings.guaranteedModifierTiers?.length || 0) +
+        (this.customModifierSettings.guaranteedModifierTypeOptions?.length || 0) +
+        (this.customModifierSettings.guaranteedModifierTypeFuncs?.length || 0);
```
DayKev:
```suggestion
        (this.customModifierSettings.guaranteedModifierTiers?.length ?? 0) +
        (this.customModifierSettings.guaranteedModifierTypeOptions?.length ?? 0) +
        (this.customModifierSettings.guaranteedModifierTypeFuncs?.length ?? 0);
```

src/phases/select-modifier-phase.ts
```diff
@@ -98,259 +80,300 @@ export class SelectModifierPhase extends BattlePhase {
...
+    }
 
-      return !cost!; // TODO: is the bang correct?
-    };
+    return modifierCountHolder.value;
+  }
+
+  // Function that resets the reward selection screen,
+  // e.g. after pressing cancel in the party ui or while learning a move
+  private resetModifierSelect(modifierSelectCallback) {
```
DayKev: Missed a param typing.

SirzBenjie (approved):

DayKev (commented):
Other than my other comment, 👍

src/phases/select-modifier-phase.ts
Wlowscha: Fixed!

xsn34kzx (approved):

DayKev (approved):

PR #5902, opened 2025-05-30
[Refactor] Added functions `isPlayer` and `isEnemy` to reduce circular imports

Bertie690:
N/A

Helps lay groundwork for circular import removal

Made `isPlayer` be a TS narrowing function
Added `isEnemy` as an equivalent enemy pokemon version
replaced prior `instanceof` checks with new function calls

## Screenshots/Videos
N/A

## How to test the changes?
tests

## Checklist


src/data/moves/move.ts
```diff
@@ -9868,7 +9866,7 @@ export function initMoves() {
         const lastEnemyFaint = globalScene.currentBattle.enemyFaintsHistory[globalScene.currentBattle.enemyFaintsHistory.length - 1];
         return (
           (lastPlayerFaint !== undefined && turn - lastPlayerFaint.turn === 1 && user.isPlayer()) ||
-          (lastEnemyFaint !== undefined && turn - lastEnemyFaint.turn === 1 && !user.isPlayer())
+          (lastEnemyFaint !== undefined && turn - lastEnemyFaint.turn === 1 &&user.isEnemy())
```
DayKev, line 9869:
```suggestion
          (lastEnemyFaint !== undefined && turn - lastEnemyFaint.turn === 1 && user.isEnemy())
```

src/system/pokemon-data.ts
```diff
@@ -140,7 +140,7 @@ export default class PokemonData {
     this.fusionLuck = source.fusionLuck ?? (source.fusionShiny ? source.fusionVariant + 1 : 0);
     this.fusionTeraType = (source.fusionTeraType ?? 0) as PokemonType;
 
-    this.boss = (source instanceof EnemyPokemon && !!source.bossSegments) || (!this.player && !!source.boss);
+    this.boss = (source.isEnemy() && !!source.bossSegments) || (!this.player && !!source.boss);
```
DayKev, line 143:
```suggestion
    this.boss = (source instanceof EnemyPokemon && !!source.bossSegments) || (!this.player && !!source.boss);
```
You aren't able to remove this yet (at least not this way).

src/system/pokemon-data.ts
```diff
@@ -5,7 +5,7 @@ import { Nature } from "#enums/nature";
 import { PokeballType } from "#enums/pokeball";
 import { getPokemonSpecies, getPokemonSpeciesForm } from "../data/pokemon-species";
 import { Status } from "../data/status-effect";
-import Pokemon, { EnemyPokemon, PokemonBattleData, PokemonMove, PokemonSummonData } from "../field/pokemon";
+import Pokemon, { PokemonBattleData, PokemonMove, PokemonSummonData } from "../field/pokemon";
```
DayKev, line 8:
```suggestion
import Pokemon, { EnemyPokemon, PokemonBattleData, PokemonMove, PokemonSummonData } from "../field/pokemon";
```

src/data/moves/move.ts
Bertie690: I swear I thought biome would fix that lol.
Oh well...

src/field/pokemon.ts
```diff
@@ -3303,7 +3305,8 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       this.battleInfo.setX(this.battleInfo.x + (this.isPlayer() ? 150 : !this.isBoss() ? -150 : -198));
       this.battleInfo.setVisible(true);
       if (this.isPlayer()) {
-        this.battleInfo.expMaskRect.x += 150;
+        // TODO: How do you get this to not require a private property access?
+        this["battleInfo"].expMaskRect.x += 150;
```
emdeann, line 3309: The issue here is that the base Pokemon class is doing things that are the job of the PlayerPokemon class. Really, `Pokemon` should never itself be calling `isPlayer`. 

I'm generally very against bypassing access modifiers like this but it's okay here as the way to fix it is certainly out of scope

emdeann (dismissed):

DayKev (dismissed):

SirzBenjie (commented):
I know this isn't done consistently in the codebase, but _please_ change the `type X` that appears in import lines and move them to their own line.
e.g.
``ts
import { type Foo, Bar, Baz } from 'path'```
to
```ts
import { Bar, Baz } from 'path'
import type { Foo } form 'path'
```
If I could make a lint rule for this, I would.

Bertie690:
Sure ig - not gonna be free for a fair bit

src/phases/faint-phase.ts
```diff
@@ -18,7 +18,7 @@ import { BattleSpec } from "#app/enums/battle-spec";
 import { StatusEffect } from "#app/enums/status-effect";
 import type { EnemyPokemon } from "#app/field/pokemon";
 import type Pokemon from "#app/field/pokemon";
-import { HitResult, PlayerPokemon, PokemonMove } from "#app/field/pokemon";
+import { HitResult, type PlayerPokemon, PokemonMove } from "#app/field/pokemon";
```
DayKev:
```suggestion
import { HitResult, PokemonMove } from "#app/field/pokemon";
import type { PlayerPokemon } from "#app/field/pokemon";
```

SirzBenjie (approved):

PR #5911, opened 2025-05-31
[Test/Ability] Cleaned up Magic Guard/battle tests; fixed innards out interaction

Bertie690:
Innards Out no longer incorrectly triggers and damages magic guard opponents, nor on one's own ally.

Was condensing some magic guard/battle tests into `it.each`es when I noticed no innards out test case.
Added one, it failed, did some digging around, etc etc.

Cleaned up Innards Out code to no longer trigger on ally friendly fire KOs & added basic tests
Added innards out tests

Cleaned up aforementioned tests to use `it.each`

## Screenshots/Videos
TODO
## How to test the changes?
`npm run test -- innards-out.test.ts magic_guard.test.ts`

## Checklist
  - I could've made them separate but would have to make tests anyways; the actual fix IMHO is too minor to warrant yet another blocker.

### DIFF
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -1246,7 +1246,7 @@ export class MoveTypeChangeAbAttr extends PreAttackAbAttr {
 
   /**
    * Determine if the move type change attribute can be applied
-   * 
+   *
    * Can be applied if:
    * - The ability's condition is met, e.g. pixilate only boosts normal moves,
    * - The move is not forbidden from having its type changed by an ability, e.g. {@linkcode Moves.MULTI_ATTACK}
@@ -1262,7 +1262,7 @@ export class MoveTypeChangeAbAttr extends PreAttackAbAttr {
    */
   override canApplyPreAttack(pokemon: Pokemon, _passive: boolean, _simulated: boolean, _defender: Pokemon | null, move: Move, _args: [NumberHolder?, NumberHolder?, ...any]): boolean {
     return (!this.condition || this.condition(pokemon, _defender, move)) &&
-            !noAbilityTypeOverrideMoves.has(move.id) && 
+            !noAbilityTypeOverrideMoves.has(move.id) &&
             (!pokemon.isTerastallized ||
               (move.id !== Moves.TERA_BLAST &&
               (move.id !== Moves.TERA_STARSTORM || pokemon.getTeraType() !== PokemonType.STELLAR || !pokemon.hasSpecies(Species.TERAPAGOS))));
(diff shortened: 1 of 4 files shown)


SirzBenjie (dismissed):

Madmadness65 (dismissed):

SirzBenjie (approved):

src/data/abilities/ability.ts
```diff
@@ -5351,20 +5359,32 @@ export class PostFaintContactDamageAbAttr extends PostFaintAbAttr {
...
 /**
- * Attribute used for abilities (Innards Out) that damage the opponent based on how much HP the last attack used to knock out the owner of the ability.
+ * Attribute used for abilities that damage opponents causing the user to faint
+ * equal to the amount of damage the last attack inflicted.
+ *
+ * Used for {@linkcode Abilities.INNARDS_OUT}.
  * @sealed
  */
 export class PostFaintHPDamageAbAttr extends PostFaintAbAttr {
-  override apply({ simulated, pokemon, move, attacker }: PostFaintAbAttrParams): void {
```
DayKev, line 5378: Why did you remove `simulated`?

test/abilities/magic_guard.test.ts
```diff
@@ -29,404 +28,142 @@ describe("Abilities - Magic Guard", () => {
...
+  it.each<{ name: string; move?: MoveId; enemyMove?: MoveId }>([
+    { name: "Non-Volatile Status Conditions", enemyMove: MoveId.TOXIC },
+    { name: "Volatile Status Conditions", enemyMove: MoveId.LEECH_SEED },
+    { name: "Crash Damage", move: MoveId.HIGH_JUMP_KICK },
+    { name: "Variable Recoil Moves", move: MoveId.DOUBLE_EDGE },
+    { name: "HP% Recoil Moves", move: MoveId.CHLOROBLAST },
+  ])("should prevent damage from $name", async ({ move = MoveId.SPLASH, enemyMove = MoveId.SPLASH }) => {
+    await game.classicMode.startBattle([SpeciesId.MAGIKARP]);
+    // force a miss on HJK
+    vi.spyOn(allMoves[MoveId.HIGH_JUMP_KICK], "accuracy", "get").mockReturnValue(0);
```
DayKev: HJK still hits because the opponent has No Guard. This isn't how you're supposed to force a miss anyway, you're supposed to use `await game.move.forceMiss()`. With the way you've done it here, you'd have to do `if (move === HJK) forceMiss()` or something.

Because we keep having issues like this when testing for a negative, we should probably make sure we add some guard `expect`s in these kinds of tests, like `expect(hjk).toHaveMissed()`

Bertie690: I agree - maybe a thought for later though

Wlowscha (approved):

SirzBenjie (approved):

PR #5919, opened 2025-06-01
[Refactor] Cleanup achvs ui handler

SirzBenjie:
None

Another UI file to be cleaned up.

Made `AchvUnlocks` and `VoucherUnlocks` interfaces public.
Cleaned up the logic in each of the methods.
Split up `processInput` into sub methods for each input that is pressed.
Used phaser object chaining.
Removed the `.setOrigin(0)` calls from items returned by `addWindow` as this method already sets origin to 0.

De-duplicated the logic common to `updateAchvIcons` and `updateVoucherIcons` into one method.

## Screenshots/Videos
N/A

## How to test the changes?
Open up a game and verify that the achievements menu works as it always has.

## Checklist


src/ui/achvs-ui-handler.ts
```diff
@@ -70,44 +70,35 @@ export default class AchvsUiHandler extends MessageUiHandler {
   setup() {
     const ui = this.getUi();
 
-    this.mainContainer = globalScene.add.container(1, -(globalScene.game.canvas.height / 6) + 1);
+    /** Width of the global canvas / 6 */
+    const WIDTH = globalScene.game.canvas.width / 6;
```
DayKev, line 74: Btw there's a `scaledCanvas` field you can use instead of needing to specify `/ 6` each time.
```ts
export class SceneBase extends Phaser.Scene {
  /**
   * Since everything is scaled up by 6 by default using the game.canvas is annoying
   * Until such point that we use the canvas normally, this will be easier than
   * having to divide every width and heigh by 6 to position and scale the ui
   * @readonly
   * @defaultValue
   * width: `320`
   * height: `180`
   */
  public readonly scaledCanvas = {
    width: 1920 / 6,
    height: 1080 / 6,
  };
```

xsn34kzx (approved):

src/ui/achvs-ui-handler.ts
SirzBenjie: Where is that?

DayKev: `SceneBase`: https://github.com/pagefaultgames/pokerogue/blob/beta/src/scene-base.ts
Which is extended by `BattleScene`

SirzBenjie: tbh, I'd rather not rely on that because that's a magic number and I'd prefer to refer to the canvas's width which, hopefully, may be able to change dynamically.

DayKev: The `/ 6` is also a magic number though.

Actually this is specifically not a magic number, that's the whole point of defining it in one place as a constant.

SirzBenjie: That makes sense. In that case, I think a better solution would be to open a separate PR which removes all uses of `globalScene.game.canvas.width / 6` and uses the scaledCanvas instead.
Imo, for now, consistency with the other parts of the codebase is better. I don't see scaledCanvas being used in like... any of the UI files.

PR #5931, opened 2025-06-04
[Bug] Status moves blocked by terrain now use correct message

### LINKED ISSUE #3722: [Bug] No Indication for Sleep Move Failure on Electric Terrain

### LINKED ISSUE #3898: [Bug] No Indication for Status Move Failure on Misty Terrain
### Describe the bug

https://replay.pokemonshowdown.com/gen9customgame-2190484548-11eut1hfmal1b0y1s7o2cjauij8qrzepw

When using a status move with Misty terrain active no message is displayed when it fails to inflict status.

### Session export file

_No response_

### User data export file

_No response_

### Expected behavior

When using a status move with Misty Terrain on the field, there should be a message that shows a pokemon was protected;
"xxxxx surrounds itself with a protective mist"

-Message should not appear in cases where a Pokémon would be given status by the secondary effect of a move.

### Screenshots / Videos

https://github.com/user-attachments/assets/e03dad4a-4ea6-4587-bba8-3f568771b590

Other

### If other please specify

gameplay

### Which browser do you use?

Other

### If other please specify

gameplay

### Additional context

Related to #3722 ([Bug] No Indication for Sleep Move Failure on Electric Terrain)

Bertie690:
> [!NOTE]
> **This does not have any locales changes.** The existing terrain block messages were already in the code, the issue merely was the status code not using them when they should.

Electric/Misty terrain will show correct messages for blockage

- Fixes #3898 
- Fixes #3722 

Moved terrain-based message funcs out of `weather.ts` and into `terrain.ts` where they should've been in the first place

cleaned up all 3 functions and made them return empty strings instead of `null` on error

renamed `Pokemon.queueImmuneMessage` and cleaned it up to display the terrain messages

Used said func inside confusion apply code

## Screenshots/Videos

**IGNORE THE CONFUSION NOT BEING BLOCKED. I FIXED IT AFTER THE FACT**
<p>

https://github.com/user-attachments/assets/4af561df-011e-44c7-8be4-e9871f9c929e

</p>
</details> 

## How to test the changes?
TBD - will make tests later

## Checklist


SirzBenjie (approved):
Looks good.

Madmadness65 (dismissed):

Snailman11:
Does this fix
#3898  and #3722

Madmadness65 (dismissed):

src/data/terrain.ts
```diff
@@ -96,3 +97,70 @@ export function getTerrainColor(terrainType: TerrainType): [number, number, numb
...
+      return i18next.t("terrain:mistyStartMessage");
+    case TerrainType.ELECTRIC:
+      return i18next.t("terrain:electricStartMessage");
+    case TerrainType.GRASSY:
+      return i18next.t("terrain:grassyStartMessage");
+    case TerrainType.PSYCHIC:
+      return i18next.t("terrain:psychicStartMessage");
+    default:
+      console.warn(`${terrainType} unexpectedly provided as terrain type to getTerrainStartMessage!`);
+      return "";
```
DayKev: Rather than having this in the switch statement as the `default` case, I think it'd be better to have it outside so that if a new terrain type is added in the future it'll cause TS/Biome/whatever to emit an error if no one remembers to add it to this switch.
Same with the other ones.

src/data/battler-tags.ts
```diff
@@ -678,7 +678,11 @@ export class ConfusedTag extends BattlerTag {
   }
 
   canAdd(pokemon: Pokemon): boolean {
-    return globalScene.arena.terrain?.terrainType !== TerrainType.MISTY || !pokemon.isGrounded();
+    const blockedByTerrain = !(pokemon.isGrounded() && globalScene.arena.terrain?.terrainType === TerrainType.MISTY);
+    if (blockedByTerrain) {
+      pokemon.queueStatusImmuneMessage(false, TerrainType.MISTY);
+    }
+    return blockedByTerrain;
```
DayKev: The condition in the `if` needs to be inverted from the return value, otherwise the message doesn't appear.
```suggestion
    const blockedByTerrain = pokemon.isGrounded() && globalScene.arena.terrain?.terrainType === TerrainType.MISTY;
    if (blockedByTerrain) {
      pokemon.queueStatusImmuneMessage(false, TerrainType.MISTY);
      return false;
    }
    return true;
```

DayKev:
> Does this fix #3898 and #3722

Yes, updated the PR description.

src/data/terrain.ts
Bertie690: Alr - how would we do that?

src/data/terrain.ts
```diff
@@ -96,3 +97,70 @@ export function getTerrainColor(terrainType: TerrainType): [number, number, numb
...
+  }
```
DayKev: Like this.
```suggestion
  }
  console.warn(`${terrainType} unexpectedly provided as terrain type to getTerrainStartMessage!`);
  return "";
```

Bertie690: @DayKev do we need a return statement? I think TS might compiler error if the branch is reached without a return.

DayKev: TS will recognize that we've exhaustively covered the possible enum values, so it won't complain if there's no `default: return "";` in the `switch` statement. Unless a new `TerrainType` value is added in the future, in which case we do want the error to occur so the new value has to be added to the `switch`.

Bertie690: Kev, it doesn't complain even if there IS one - we don't have exhaustiveness checks on atm

DayKev (approved):

PR #5934, opened 2025-06-04
[UI/UX] Adding more space in Egg Gacha for another languge

ShinigamiHolo:
Text that is not overlaping with frame or goes out of it (Russian)

Because, in this case, Russian language take more space than menu gives which results with text looking odd and going out of frame

Just some lines of code in "src/ui/egg-gacha-ui-handler.ts". New variable EGOS_width and switch statement added.
Easy to add new languages to expand menu if needed

## Screenshots/Videos
Changed
Changed
Original
Original

## How to test the changes?
Changing language settings 

## Checklist


DayKev (dismissed):

Madmadness65 (approved):

src/ui/egg-gacha-ui-handler.ts
```diff
@@ -206,7 +206,17 @@ export default class EggGachaUiHandler extends MessageUiHandler {
     this.eggGachaOptionsContainer = globalScene.add.container(globalScene.game.canvas.width / 6, 148);
     this.eggGachaContainer.add(this.eggGachaOptionsContainer);
 
-    this.eggGachaOptionSelectBg = addWindow(0, 0, 96, 16 + 576 * this.scale);
+    // Increase egg box width on certain languages
+    let EGOS_width = 0;
```
SirzBenjie: Can we change the variable name to something that isn't EGOS_width? It isn't immediately obvious that EGOS is an acronym for Egg Gacha Options Select. The name also uses an underscore, which is a bit jarring when most of the codebase is (hopefully) using camelCase.
The other thing is that the variable name is using an underscore, which we don't generally use in the codebase.

Given that the code is minified, I don't see a reason to avoid a longer name. Just `eggGachaOptionSelectWidth` is fine.

SirzBenjie (commented):
Changes are good and I apologize for the nitpicky comment, but...

src/ui/egg-gacha-ui-handler.ts
ShinigamiHolo: "EGOS_width" was used as placeholder just to check if this menu can be expanded, I didn't check how variables named usually so I left it as is, so change name however is fitting
If I need to edit it... Idk how to do it from here...

ShinigamiHolo:
How do I delete file from pull request..?

DayKev (approved):

Madmadness65 (approved):

PR #5936, opened 2025-06-05
[Bug] Fix Rattled Speed Stat Increase Delayed until End of Turn

### LINKED ISSUE #5516: [Bug] Rattled Speed Stat Increase Delayed until End of Turn
### Describe the bug

When activated by Intimidate, Rattled boosts Speed after all other Pokemon have been considered, instead of immediately.
- When switching, this results in the new Pokemon gaining the Speed boost instead of the Rattled Pokemon
- On Battle start, Rattled's speed boost is only handled after other Pokemon's stat modifications have been resolved.

Link to Discord's Bug Report (Rattled Intimidate Interaction)
https://discord.com/channels/1125469663833370665/1349072523756568607

### Reproduction

Start run with Gimmighoul - Chest, and 2 other Pokemon. 
Use False Swipe against an Opponent, and then use Trick Room.
- When the opponent switches, switch out Gimmighoul as well
```
    STARTING_WAVE_OVERRIDE: 25,

    STARTING_LEVEL_OVERRIDE: 100,

    MOVESET_OVERRIDE: [Moves.FALSE_SWIPE, Moves.TRICK_ROOM],

    OPP_ABILITY_OVERRIDE: Abilities.INTIMIDATE,	
	
```

```
    STARTING_WAVE_OVERRIDE: 8,

    STARTING_LEVEL_OVERRIDE: 100,

    MOVESET_OVERRIDE: [Moves.FALSE_SWIPE, Moves.TRICK_ROOM],

    OPP_ABILITY_OVERRIDE: Abilities.INTIMIDATE,
```

### Expected behavior

Rattled should boost speed immediately after it's triggered by Intimidate.

### Screenshots / Videos

Rattled 1v1

https://github.com/user-attachments/assets/4a680526-7c67-4f64-b40d-ee6c3b1fff04

Rattled 2v2

https://github.com/user-attachments/assets/fdcd08a5-b385-4ef2-9f77-eb7cf6e907dc

<details><summary>Defiant is correct</summary>

Defiant 2v2

https://github.com/user-attachments/assets/13f7954a-c56e-44b1-9dcb-20d028e53773

Defiant Entry is correct

https://github.com/user-attachments/assets/1477a432-1a0f-4442-a381-a91b79abebf6

</details> 

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

Gamez0:

Currently, when the opponent changes their Pokémon and Rattle is activated, the attack power is lowered and then the speed does not increase right away. This PR aims to fix this issue.

Fixes #5516 

- ability.ts

In the place where `pushPhase` was done, use `globalScene.findPhase`.
If there is a `MovePhase` or `SwitchSummonPhase` in the `phaseQueue`, `prependToPhase`is done based on the found phase,
and if there is none, `pushPhase` is done as before.

## How to test the changes?

1. Set overrides in src/overrides.ts
```ts
const overrides = {
  STARTING_WAVE_OVERRIDE: 25,

  STARTING_LEVEL_OVERRIDE: 100,

  MOVESET_OVERRIDE: [MoveId.FALSE_SWIPE, MoveId.TRICK_ROOM],

  OPP_ABILITY_OVERRIDE: AbilityId.INTIMIDATE,

  OPP_MOVESET_OVERRIDE: [MoveId.PIN_MISSILE] // Bug type move (optional)

```
2. Start battle

3. Check speed raise after attack falls

4. Use False Swipe against an Opponent, and then use Trick Room.

5. When the opponent switches, switch out Gimmighoul as well

## Checklist
~~[ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?~~

~~Are there any localization additions or changes? If so:
[ ] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?
  [ ] If so, please leave a link to it here: 
 [ ] Has the translation team been contacted for proofreading/translation?~~

### DIFF
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -2130,7 +2130,12 @@ export class PostIntimidateStatStageChangeAbAttr extends AbAttr {
 
   override apply(pokemon: Pokemon, passive: boolean, simulated:boolean, cancelled: BooleanHolder, args: any[]): void {
     if (!simulated) {
-      globalScene.pushPhase(new StatStageChangePhase(pokemon.getBattlerIndex(), false, this.stats, this.stages));
+      const newStatStageChangePhase = new StatStageChangePhase(pokemon.getBattlerIndex(), false, this.stats, this.stages)
+      if(globalScene.findPhase(m => m instanceof MovePhase)){
+        globalScene.prependToPhase(newStatStageChangePhase, MovePhase)
+      }else {
+        globalScene.pushPhase(newStatStageChangePhase);
+      }
     }
     cancelled.value = this.overwrites;
   }
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -2133,6 +2133,8 @@ export class PostIntimidateStatStageChangeAbAttr extends AbAttr {
       const newStatStageChangePhase = new StatStageChangePhase(pokemon.getBattlerIndex(), false, this.stats, this.stages)
       if(globalScene.findPhase(m => m instanceof MovePhase)){
         globalScene.prependToPhase(newStatStageChangePhase, MovePhase)
+      } else if(globalScene.findPhase(m => m instanceof SwitchSummonPhase)){
+        globalScene.prependToPhase(newStatStageChangePhase, SwitchSummonPhase)
       }else {
         globalScene.pushPhase(newStatStageChangePhase);
       }


Bertie690:
Can we get a few automated tests for this?

test/abilities/rattled.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      .moveset([MoveId.FALSE_SWIPE, MoveId.TRICK_ROOM])
+      .ability(AbilityId.RATTLED)
+      .battleType(BattleType.TRAINER)
+      .disableCrits()
+      .enemySpecies(SpeciesId.MAGIKARP)
+      .enemyAbility(AbilityId.INTIMIDATE)
+      .enemyMoveset(MoveId.PIN_MISSILE);
+  });
+
+  it("should reduce attack and then increase speed", async () => {
```
Bertie690:
```suggestion
  it("should trigger and boost speed immediately after Intimidate attack drop on initial send out", async () => {
```

test/abilities/rattled.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    await game.classicMode.runToSummon([SpeciesId.GIMMIGHOUL]);
```
Bertie690:
```suggestion
    // `runToSummon` used instead of `startBattle` to avoid skipping past initial "post send out" effects
    await game.classicMode.runToSummon([SpeciesId.GIMMIGHOUL]);
```

Use `startBattle` - `runToSummon` is a bit weird sometimes.
(IDK why it's even exported but ehhhh...)

test/abilities/rattled.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    const playerPokemon = game.scene.getPlayerPokemon();
+    await game.phaseInterceptor.to("StatStageChangePhase");
+
+    expect(playerPokemon!.getStatStage(Stat.ATK)).toBe(-1);
+    expect(playerPokemon!.getStatStage(Stat.SPD)).toBe(0);
+
+    await game.phaseInterceptor.to("StatStageChangePhase");
+
+    expect(playerPokemon!.getStatStage(Stat.ATK)).toBe(-1);
+    expect(playerPokemon!.getStatStage(Stat.SPD)).toBe(1);
```
Bertie690: There are new test utils for this fwiw
```suggestion
    const playerPokemon = game.field.getPlayerPokemon();
    await game.phaseInterceptor.to("StatStageChangePhase");

    expect(playerPokemon.getStatStage(Stat.ATK)).toBe(-1);
    expect(playerPokemon.getStatStage(Stat.SPD)).toBe(0);

    await game.phaseInterceptor.to("StatStageChangePhase");

    expect(playerPokemon.getStatStage(Stat.ATK)).toBe(-1);
    expect(playerPokemon.getStatStage(Stat.SPD)).toBe(1);
```

Gamez0:
@Bertie690 
I added a test to check the order of stat changes. (https://github.com/pagefaultgames/pokerogue/pull/5936/commits/400733442e284081d6650dcc08440f3bdaddaa4f, attack falls first then speed increases) 

It would be nice if there was a test for when the pokemon are swapped, but I don't know how to write a test code that forces the enemy trainer to change pokemon. There doesn't seem to be many examples, so could you please give me some?

test/abilities/rattled.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    game.phaseInterceptor.restoreOg();
+  });
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override
+      .moveset([MoveId.FALSE_SWIPE, MoveId.TRICK_ROOM])
+      .ability(AbilityId.RATTLED)
+      .battleType(BattleType.TRAINER)
+      .disableCrits()
```
Bertie690: Might be paranoid but better overly paranoid than not...?
```suggestion
      .disableCrits()
      .battleStyle("single")
```

src/data/abilities/ability.ts
```diff
@@ -2130,7 +2130,14 @@ export class PostIntimidateStatStageChangeAbAttr extends AbAttr {
 
   override apply(pokemon: Pokemon, passive: boolean, simulated:boolean, cancelled: BooleanHolder, args: any[]): void {
     if (!simulated) {
```
Bertie690, line 2625: Can we get an early return here?

src/data/abilities/ability.ts
```diff
@@ -2130,7 +2130,14 @@ export class PostIntimidateStatStageChangeAbAttr extends AbAttr {
...
-      globalScene.pushPhase(new StatStageChangePhase(pokemon.getBattlerIndex(), false, this.stats, this.stages));
+      const newStatStageChangePhase = new StatStageChangePhase(pokemon.getBattlerIndex(), false, this.stats, this.stages)
+      if(globalScene.findPhase(m => m instanceof MovePhase)){
+        globalScene.prependToPhase(newStatStageChangePhase, MovePhase)
+      } else if(globalScene.findPhase(m => m instanceof SwitchSummonPhase)){
+        globalScene.prependToPhase(newStatStageChangePhase, SwitchSummonPhase)
+      }else {
```
```suggestion
      if (globalScene.findPhase(m => m instanceof MovePhase)) {
        globalScene.prependToPhase(newStatStageChangePhase, MovePhase)
      } else if (globalScene.findPhase(m => m instanceof SwitchSummonPhase)) {
        globalScene.prependToPhase(newStatStageChangePhase, SwitchSummonPhase)
      } else {
```

Bertie690: Wouldn't it make more sense to use `unshiftPhase` to have it trigger immediately rather than "before the next action that may or may not occur"?

test/abilities/rattled.test.ts
Bertie690: EDIT - I realized this _has_ to use `runToSummon` as it checks for reactions to a start-of-battle send-out (whereas `startBattle` skips over everything before the first `CommandPhase`).

Changed review to add comment justifying usage

Bertie690:
> It would be nice if there was a test for when the pokemon are swapped, but I don't know how to write a test code that forces the enemy trainer to change pokemon. There doesn't seem to be many examples, so could you please give me some?

`game.forceEnemyToSwitch` and `game.doSwitchPokemon` in `gameManager.ts` are helper functions to force each side (enemy/player) to switch to a party member. Either works fine (though the latter requires specifying 2+ species in the `startBattle` array so you have smthng to swap to).

The switch test can _probably_ just check order of effects rather than actual stat boost application (since the first test does already).

src/data/abilities/ability.ts
Gamez0: You have a point. I modified this in this way because @DayKev first suggested that this part should be changed. I think it would be good to ask @DayKev for his opinion as well.

src/data/abilities/ability.ts
Gamez0: https://github.com/pagefaultgames/pokerogue/pull/5936/commits/92e45fba81c9bdb16464190f909dc3dba12c77ce Thank you @Bertie690. Please make sure I changed it the way you intended.

src/data/abilities/ability.ts
DayKev: I suggested `unshiftPhase` in the linked issue, afaik that's all that needs to be done.

Gamez0: @Bertie690 @DayKev 
When using `unshiftPhase`, I figure out rattled was triggered first and then intimated was triggered. From what I understand Rattled is the ability that reacts to certain types of moves. So it seems natural that the intimated effect would be shown first and then rattled would be triggered. I mentioned it in the description above. 
(This was one of the reason I wrote this test https://github.com/pagefaultgames/pokerogue/pull/5936/commits/400733442e284081d6650dcc08440f3bdaddaa4f)

> Why I didn't do it the suggested way:
https://github.com/pagefaultgames/pokerogue/issues/5516#issuecomment-2834364564, the speed increase phase due to rattle will occur before the attack decrease phase due to intimidation. So we need to check if it's before switching to a specific phase and do pushPhase or prependToPhase depending on the situation.

Gamez0:
Thank you @Bertie690. With your help I wrote another test. https://github.com/pagefaultgames/pokerogue/pull/5936/commits/6bf78cd732a5f020be005c3c6d8958cb6db22879

src/data/abilities/ability.ts
Bertie690: I'll try and tinker and see what I come up with.
Making it `unshiftPhase` would mean the flyout stays for the duration of the stat boost message phase.

SirzBenjie (approved):

src/data/abilities/ability.ts
DayKev: Ah, gotcha.

src/data/abilities/ability.ts
```diff
@@ -2129,9 +2129,20 @@ export class PostIntimidateStatStageChangeAbAttr extends AbAttr {
   }
 
   override apply(pokemon: Pokemon, passive: boolean, simulated:boolean, cancelled: BooleanHolder, args: any[]): void {
-    if (!simulated) {
-      globalScene.pushPhase(new StatStageChangePhase(pokemon.getBattlerIndex(), false, this.stats, this.stages));
+    if (simulated) {
+      cancelled.value = this.overwrites;
+      return
```
```suggestion
      return;
```

src/data/abilities/ability.ts
```diff
@@ -2129,9 +2129,20 @@ export class PostIntimidateStatStageChangeAbAttr extends AbAttr {
...
 
   override apply(pokemon: Pokemon, passive: boolean, simulated:boolean, cancelled: BooleanHolder, args: any[]): void {
-    if (!simulated) {
-      globalScene.pushPhase(new StatStageChangePhase(pokemon.getBattlerIndex(), false, this.stats, this.stages));
+    if (simulated) {
+      cancelled.value = this.overwrites;
+      return  
+    }
+
+    const newStatStageChangePhase = new StatStageChangePhase(pokemon.getBattlerIndex(), false, this.stats, this.stages)
```
```suggestion
    const newStatStageChangePhase = new StatStageChangePhase(pokemon.getBattlerIndex(), false, this.stats, this.stages);
```

src/data/abilities/ability.ts
```diff
@@ -2129,9 +2129,20 @@ export class PostIntimidateStatStageChangeAbAttr extends AbAttr {
...
+    if (globalScene.findPhase(m => m instanceof MovePhase)) {
+      globalScene.prependToPhase(newStatStageChangePhase, MovePhase)
```
```suggestion
      globalScene.prependToPhase(newStatStageChangePhase, MovePhase);
```

src/data/abilities/ability.ts
```diff
@@ -2129,9 +2129,20 @@ export class PostIntimidateStatStageChangeAbAttr extends AbAttr {
...
+    } else if (globalScene.findPhase(m => m instanceof SwitchSummonPhase)) {
+      globalScene.prependToPhase(newStatStageChangePhase, SwitchSummonPhase)
```
```suggestion
      globalScene.prependToPhase(newStatStageChangePhase, SwitchSummonPhase);
```

src/data/abilities/ability.ts
```diff
@@ -2129,9 +2129,20 @@ export class PostIntimidateStatStageChangeAbAttr extends AbAttr {
...
+    } else {
+      globalScene.pushPhase(newStatStageChangePhase);
     }
+
```
```suggestion

```

DayKev (approved):

DayKev (dismissed):

Bertie690:
PR REVIEW:
https://github.com/Gamez0/pokerogue/pull/1

src/data/abilities/ability.ts
Bertie690: OK so apparently this happens because the clown who did guard dog merged its intimidate immunity into `PostIntimidateStatStageChangeAbAttr` (instead of using `IntimidateImmunityAbAttr`), and thus triggered Rattled _BEFORE_ Intimidate to allow it to be canceled. 
Separating out the attributes for Guard Dog lets us trigger rattled after intim and use `unshiftPhase`.

Bertie690 (approved):
LGTM!
Short and sweet PR.

DayKev (approved):

Gamez0:
Hello @DayKev 
Just letting you know that I think this PR is ready to be merged. Please take a look when you have a chance. Thanks!

DayKev (dismissed):

SirzBenjie:
I can't believe this didn't get in. Sincere apologies, @Gamez0, we'll get this in for 1.10.8

SirzBenjie:
@emdeann I marked one of the tests on here as TODO because I didn't have time to figure out how to force rattled's speed stat stage increase to trigger after the intimidate one.

src/data/abilities/ability.ts
```diff
@@ -2633,16 +2633,14 @@ export class IntimidateImmunityAbAttr extends CancelInteractionAbAttr {
...
-  constructor(stats: readonly BattleStat[], stages: number, overwrites?: boolean) {
+  constructor(stats: readonly BattleStat[], stages: number) {
     super(true);
     this.stats = stats;
     this.stages = stages;
-    this.overwrites = !!overwrites;
   }
 
-  override apply({ pokemon, simulated, cancelled }: AbAttrParamsWithCancel): void {
+  override apply({ pokemon, simulated }: AbAttrParamsWithCancel): void {
```
```suggestion
  override apply({ pokemon, simulated }: AbAttrBaseParams): void {
```

test/abilities/rattled.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    expect(feebas).toHaveStatStage(Stat.ATK, -1);
+    expect(feebas).toHaveStatStage(Stat.SPD, 1);
+    // Nothing but show/hide ability phases should be visible
+    for (const log of game.phaseInterceptor.log) {
+      expect(log).toBeOneOf(["ShowAbilityPhase", "HideAbilityPhase", "StatStageChangePhase", "MessagePhase"]);
+    }
+  });
+
+  it("should activate after Intimidate from enemy switch", async () => {
```
Bertie690:
```suggestion
  // TODO: This is also flaky 
  it.todo("should activate after Intimidate from enemy switch", async () => {
```

src/data/abilities/ability.ts
```diff
@@ -2858,6 +2853,9 @@ export class PostSummonStatStageChangeAbAttr extends PostSummonAbAttr {
           this.stages,
         );
       }
+      // TODO: Ensure that the stat stage change phase applied here applies after intimidate's
+      // TODO: This will cause rattled to increase with scrappy if mon has both.
+      applyAbAttrs("PostIntimidateStatStageChangeAbAttr", params);
```
emdeann, line 2858:
```suggestion
if (!cancelled.value) {
  globalScene.phaseManager.unshiftNew(
    "StatStageChangePhase",
    opponent.getBattlerIndex(),
    false,
    this.stats,
    this.stages,
  );
        
  applyAbAttrs("PostIntimidateStatStageChangeAbAttr", params);
}
```
Isn't it pretty cut and dry that "PostIntimidateStatStageChangeAbAttr" should only activate if the intimidate actually happened? This would mean Guard Dog would have to be done differently of course

Bertie690: > Isn't it pretty cut and dry that "PostIntimidateStatStageChangeAbAttr" should only activate if the intimidate actually happened? 

No - the only 2 things that do this are Rattled (which we have no info on) and Guard Dog (which 100% does).

I'd rather not give Guard Dog a custom attribute whose only job is to add a side effect to `CancelInteractionAbAttr` 

Anywho might as well ask balteam about it

Bertie690: Also I'll request you hold off on this suggestion for now since it will almost certainly break guard dog as it was refactored in this PR.

We can discuss how it _should_ work later, but I'd rather not lump unrelated changes in a PR that should've been merged 2 update cycles ago.

emdeann: With the way this is written right now, Rattled will activate even when intimidate does not due to a substitute being up. That's not a balance team issue, it's strictly incorrect compared to mainline behavior.

What's the problem with giving guard dog a different attribute? Leaving it as "PostIntimidateStatStageChangeAbAttr" is dumb because it objectively does not happen "post intimidate". You could even just add a `replacement` stat stage to `IntimidateImmunityAttr`

emdeann: This isn't the type of thing I want to split into a separate PR because we'd be leaving it in a buggy state until some unknown later pr is merged

I really want to stress how nonsensical it is for an ability to have both `PostIntimidate` and `IntimidateImmunity`

emdeann:
Apologies that it's taken so long to get this PR in, things get missed unfortunately. I'd like to status check real quick:

The bug this is looking to fix doesn't exist anymore because I unknowingly fixed it as part of implementing dynamic turn order. However, I don't want to throw this PR away as there are still issues with this interaction. 

1. You moved the `applyAbAttrs("PostIntimidateStatStageChangeAbAttr", params);` call to below the phase application of Intimidate. This is good, but it needs to _only_ happen if the intimidate actually activated (i.e. inside the `if (!cancelled.value)` block. The reason is that Rattled (PostIntimidateStatStageChangeAbAttr) should not activate if intimidate was blocked by a substitute or another ability like Scrappy.

2. Guard Dog has a couple interaction issues that need to be fixed. The above change obviously would break it, because it itself cancels intimidate. 

The main idea here is that Guard Dog is emphatically _not_ an "intimidate immunity". It reverses the stat change of intimidate, but the intimidate is still applied. This is verifiable on cartridge by observing that a pokemon with Guard Dog will activate its held Adrenaline Orb when intimidated (unlike with scrappy or a substitute). 

What you need to do instead is create a new ability attribute for Guard Dog that reverses the stat change but does not cancel intimidate. It would be simple, something like 
```ts
export class FlipIntimidateAbAttr extends AbAttr {
  override canApply({cancelled}: AbAttrParamsWithCancel): boolean {
    return !cancelled.value;
  }

  override apply({cancelled}: AbAttrParamsWithCancel) {
    cancelled.value = true;
  }
}
```
and then
```ts
      if (!cancelled.value) {
        applyAbAttrs("FlipIntimidateAbAttr", {pokemon: opponent, cancelled: flipped})

        globalScene.phaseManager.unshiftNew(
          "StatStageChangePhase",
          opponent.getBattlerIndex(),
          false,
          this.stats,
          flipped.value ? this.stages * -1 : this.stages,
        );
 ```
This would allow correct interactions like Rattled activating with Guard Dog but not if the intimidate was cancelled. Note that balance team has requested that if a pokemon has an intimidate blocker like Scrappy and Guard Dog, Guard Dog should be the one to activate. The little example I gave hasn't considered that.

@Gamez0 Is this something you're willing to work on? I don't want to usurp your work but I do understand that this is quite different from the original PR, so I can take it over if you're not feeling it

Gamez0:
@emdeann Please feel free to take over. Sorry for late response.

PR #5170, opened 2025-01-23
[Bug] Fix Login Screen Buttons can be Pressed While Animating

ChrisLolz:
## What are the changes the user will see?

- When pressing the enter key or clicking the login screen buttons rapidly, the outside buttons will no longer go upwards.
- The login buttons will only work when the screen has finished animating
- Prevents closing the rename Pokemon screen while the UI is still moving. 
## Why am I making these changes?

Mainly to fix #5163
## What are the changes from a developer perspective?

- Added to check to see if the container still has tweens remaining in the action listeners of `FormModalUiHandler`, `LoginFormUiHandler`, `ModalUiHandler`, and `RegistrationFormUiHandler`. 
- Moved the creation of `infoContainer` into its own function to make setup more clean since `ExternalPartyContainer` is created on its own function already. 
## Screenshots/Videos

You can't really tell but I am spamming enter and clicking in these videos
Login screen:

https://github.com/user-attachments/assets/a070e9d7-b8a0-496c-b716-8cd5326010d9

Register screen:

https://github.com/user-attachments/assets/9b40ed7e-a3d9-46b7-a817-cef5132de116

Gear icon:

https://github.com/user-attachments/assets/ce23afdf-3acb-4c13-b54d-fb4f6c875d42

Renaming Pokemon

https://github.com/user-attachments/assets/4980c18c-c53d-4662-aea0-d520174e24b7

## How to test the changes?

I'm unable to host my own local server so I just changed line 30 and 31 to true of `login-phase.ts` to force the login screen. Otherwise just test manually by clicking the buttons.
## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test`)
  - [ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [x] Have I provided screenshots/videos of the changes (if applicable)?
  - [x] Have I made sure that any UI change works for both UI themes (default and legacy)?

Are there any localization additions or changes? If so:
- [ ] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?
  - [ ] If so, please leave a link to it here: 
- [ ] Has the translation team been contacted for proofreading/translation?

PR #5938, opened 2025-06-05
[UI/UX] Add change password ui

SirzBenjie:
There will now be a "Change Password" option that can be found under the Manage Data menu.
This opens up a form similar to the login screen that asks the user to enter a password, and then will submit the API request to the server upon a password change.

NOTE: The server backend has been changed so that all sessions are deactivated on changing the password. See https://github.com/pagefaultgames/rogueserver/compare/pagefaultgames:4d6de1e...pagefaultgames:3f648b5

The endpoint already exists in rogueserver, but there's just no way to access it.
I was also asked specifically if we could implement this by Javi.

- Added `CHANGE_PASSWORD_FORM` to the `UiMode` enum
- Created `change-password-form-ui-handler.ts` that follows the structure of the login/register forms (`login-form-ui-handler.ts` and `registration-form-ui-handler.ts`)
- Registered the new handler in the ui setup
- Tweaked the form-modal-ui-handler class to allow for behavior of the `cancel` button to be modified. This allows the password entry fields of the change password modal to be cleared in the event of a successful change or if the user presses "cancel".
- Hitting the change password endpoint will deactivate all active sessions. The endpoint will create a new session in the user's active window such that the user will not need to log out after using the change password form.

**Note**: There are a few structural with the UI code here. Fixing those is **not** the purpose of this PR. The changes in this PR also prefer to mirror the existing style of the UI implementation, mirroring the style of the UI code, even if that style is arguably inadequate. Having a consistent style is better.

## Screenshots/Videos

<details><summary>Changing password and then logging in</summary>

https://github.com/user-attachments/assets/6a290111-84fc-4d9e-8b9d-59be0de24669
</details>

<details><summary>Demonstration of mismatched passwords, too-short passwords, and clearing modal if dialogue is closed / reopened</summary>

https://github.com/user-attachments/assets/c8917e96-3c2b-4e8a-ae99-30dfc6e088d1
</details>

## How to test the changes?
You'll need a copy of rogueserver built. If you want to test this out, reach out to me on Discord and I can help set it up.

## Checklist
  - [ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes? Relies on server running, so... we can't?

  - [x] If so, please leave a link to it here: https://github.com/pagefaultgames/pokerogue-locales/pull/176


Snailman11:
1. I'd like to request feedback after the Password change attempt has succeeded, such as a sound or a message.

2. Need to see the response to failing to match the Change Password "Confirm Password" field with the "Password" field

src/ui/change-password-form-ui-handler.ts
```diff
@@ -0,0 +1,125 @@
+import type { InputFieldConfig } from "./form-modal-ui-handler";
+import { FormModalUiHandler } from "./form-modal-ui-handler";
+import type { ModalConfig } from "./modal-ui-handler";
+import { UiMode } from "#enums/ui-mode";
+import i18next from "i18next";
+// import type { OptionSelectItem } from "#app/ui/abstact-option-select-ui-handler";
```
DayKev:
```suggestion
```

src/ui/change-password-form-ui-handler.ts
```diff
@@ -0,0 +1,125 @@
...
+      };
+      // Upon pressing cancel, the inputs should be cleared
+      const originalCancelAction = this.cancelAction;
+      this.cancelAction = () => {
+        globalScene.ui.playSelect();
+        for (const input of this.inputs) {
+          input.setText("");
+        }
+        originalCancelAction?.();
+      };
```
DayKev, line 112: Don't you need to also re-assign `originalCancelAction` back to `this.cancelAction` inside here?

SirzBenjie: Nope. We aren't creating a temporary. We are augmenting the function to prepend additional logic to before the method.

DayKev: Doesn't that mean it'll keep adding the same "new" logic each time, over and over?

SirzBenjie: No, because the `cancelAction` is itself defined inside this `show` method. Specifically, it is defined in the call to `super.show()`



```ts
// filename: form-modal-ui-handler.ts
override show(args: any[]): boolean {
  if (super.show(args)) {
    this.inputContainers.map(ic => ic.setVisible(true));

    const config = args[0] as FormModalConfig;

    this.submitAction = config.buttonActions.length ? config.buttonActions[0] : null;
    this.cancelAction = config.buttonActions[1] ?? null;

    // snip
```

DayKev: It's the same for `submitAction` but you re-assign it to itself above, unlike for `cancelAction`. What's the difference?

SirzBenjie: The logic for `submitAction` was copied from the login UI.
Honestly, I don't fully understand why there's the check against the tweens in the modal container being 0, and then the re-setting of `submitAction`. From what I can tell, it appears to be redundant.
It warrants further investigation, but while the `submitAction` flow is something I ported, the `cancelAction` is something new that I wrote.

That further investigation is outside the scope of the PR, IMHO.

DayKev: The check against tweens being `0` is because otherwise you can click multiple times on the login button while it's animating and break it.
- #5170

SirzBenjie: Unless `show` is getting called twice without either `submit` or `cancel` being pressed, then it shouldn't be necessary.

Bertie690 (approved):

PR #5950, opened 2025-06-07
refactor: fix various interactions with Smack Down/Telekinesis/Gravity

Bertie690:
A whole host of undocumented issues will now be fixed:

- Smack Down/Thousand Arrows hitting a grounded, semi-invulnerable Pokemon (via No Guard or similar effects) will no longer knock the target down
- Thousand Arrows deals 2x damage against flying-types in Inverse Battles (instead of dealing 1x like prior; confirmed on showdown & bulbapedia)
- Mega Gengar receiving Telekinesis via Baton Pass will no longer retain telekinesis' ungrounding effect. (We previously only removed the accuracy boost and left the pokemon still "technically" airborne)
- (Requested by Balteam) Telekinesis' custom behavior for Mega Gengar is now also shared by G-Max Gengar, being un-passable and acively removed if gained by either.
- Gravity now correctly makes all Pokemon be considered as "grounded" while active. (We previously checked it _exclusively_ for Flying-type immunities and Levitate, leading to inconsistencies with hazards and similar effects)
- ~~Telekinesis now shows a message on removal~~ Moved to #7045 
- Telekinesis can coexist with Magnet Rise
- Inverse battles will no longer cause grounded Flying-types to take 1x damage from ground moves
- Ingrain no longer [prematurely removes Magnet Rise](https://replay.pokemonshowdown.com/gen9nationaldex-2502821440) - it simply nullifies the latter's effect until the user switches out.

Now, my initial goal for this was to make Gravity actively ground affected Pokemon (instead of being checked inside the type matchup code and effectively nowhere else).

However, the further I dug, the more botched the spaghetti became:
1. Ingrain (which should coexist with the battler tags whose effects it overrides) was forcibly and noisily removing said effects on move use.
2. Many moves had unnecessarily duplicated conditions (like Telekinesis explicitly checking for Ingrain's "healing" tag of all things)
3. The code in `Pokemon#getAttackTypeEffectiveness` resulted in grounded Flying-types taking 1x dmg from Ground-type moves inside Inverse Battles (instead of 2x).
4. Telekinesis shared the exact same `FloatingTag` as Magnet Rise, preventing them from coexisting and causing the aforementioned bugs with Mega Gengar + Baton Pass. 
  1. Moreover, the ability of `HitsTagAttr` to bypass `TypeImmuneTag`'s effects was _solely_ used to make Thousand Arrows bypass these tags' immunities, in a fairly hackjob manner to boot.
5. Levitate's move immunity attribute needed to explicitly checks for _every single effect_ that could forcibly ground the user.
6. THEY ADDED ANOTHER LEVITATE ABILITY IN LEGENDS Z-A MY LORD
The list goes on...

In light of this tangled mess (and the potential for Iron Ball to be added after the Modifier Rework), I decided to consolidate these checks to make them easier to modify later on and fix all these problems.

1. Moved the various additional attributes on Smack Down and Thousand Arrows (namely the `AddBattlerTagAttr` and `RemoveBattlerTagAttr` parts) into the effect of `SmackDownAttr`. Said attribute now ensures that its target is currently airborne (barring any potential semi-invulnerability) before forcibly grounding them and removing the aforementioned tags, as occurs in the mainline series. (You can't smack down a diving golduck.)
2. Added a Gravity check to `Pokemon#isGrounded` alongside a new parameter (`ignoreSemiInvulnerable`) to skip the semi-invuln check, alongside a small helper function (`Pokemon.isForciblyGrounded`) that checks for forced-grounding or forced-airborne effects.
3. Reworked the existing code inside `Pokemon#getAttackTypeEffectiveness` to rely on `Pokemon.isForciblyGrounded`; it now returns `0` for Ground moves if a Pokemon has an effect forcibly lifting it into the air (and the current move isn't Thousand Arrows). This consolidates the various checks for the aforementioned grounded checks into a single place, while making the addition of future effects (ex: Iron Ball) as simple as modifying a single function.
5. Moved Telekinesis to its own tag separate from `FloatingTag` to make it stack with Magnet Rise.
7. Reworked Levitate into a simple boolean holder setting ability.
8. Converted `BattlerTag#isBatonPassable` into a public function that takes the recipient Pokemon as a parameter, replacing the hardcoded (and broken) check for Mega Gengar + Telekinesis inside `Pokemon#transferSummon`.

Miscellaneous and varied docs updates to related code and tests
## Screenshots/Videos

https://github.com/user-attachments/assets/1501de7d-69f9-4087-8834-80961ca94d00

https://github.com/user-attachments/assets/4e8350c8-195f-434b-8cbc-fc4ad35e50ea

## How to test the changes?
`pnpm test:silent magnet-rise.test.ts telekinesis.test.ts fly-bounce.test.ts smack-down-thousand-arrows.test.ts ingrain.test.ts`

## Checklist
  - Many of these changes rely on each other or need special attention when taken all together, hence why I decided to tackle them all at once.
  - The only exception is #7045, which should be merged before this one if possible.
  - [x] Have I created new automated tests (`pnpm create-test`) or updated existing tests related to the PR's changes?


src/field/pokemon.ts
```diff
@@ -2263,13 +2263,28 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * @returns Whether this pokemon is currently grounded, as described above.
+   */
+  public isGrounded(ignoreSemiInvulnerable = false): boolean {
     return (
       !!this.getTag(GroundedTag) ||
       (!this.isOfType(PokemonType.FLYING, true, true) &&
         !this.hasAbility(AbilityId.LEVITATE) &&
         !this.getTag(BattlerTagType.FLOATING) &&
-        !this.getTag(SemiInvulnerableTag))
+        ignoreSemiInvulnerable || !this.getTag(SemiInvulnerableTag)
```
TijmenvanderKemp: `&&` [binds stronger](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence) than `||` so is this code correct? Shouldn't there be parentheses around your `||` operation?

Bertie690: yes oops

SirzBenjie:
Going to mark this as a draft until the tests pass.

Bertie690:
I still need to condense the groundedness tests into 1 large file

src/data/battler-tags.ts
```diff
@@ -2254,33 +2274,21 @@ export class SemiInvulnerableTag extends SerializableBattlerTag {
...
-    this.#immuneType = immuneType;
-  }
-}
-
 /**
- * Battler Tag that lifts the affected Pokemon into the air and provides immunity to Ground type moves.
- * @see {@link https://bulbapedia.bulbagarden.net/wiki/Magnet_Rise_(move) | MoveId.MAGNET_RISE}
- * @see {@link https://bulbapedia.bulbagarden.net/wiki/Telekinesis_(move) | MoveId.TELEKINESIS}
+ * Battler Tag that lifts the affected Pokemon into the air, providing immunity to Ground-type moves.
+ * Used by Magnet Rise.
```
DayKev: Why did you remove the `@see ...` sections?

SirzBenjie (dismissed):
This is an extremely high quality PR.

Madmadness65 (dismissed):

SirzBenjie (dismissed):

src/data/moves/move.ts
```diff
@@ -6361,7 +6359,9 @@ export class NeutralDamageAgainstFlyingTypeAttr extends MoveTypeChartOverrideAtt
     args: [multiplier: NumberHolder, types: readonly PokemonType[], moveType: PokemonType],
   ): boolean {
     const [multiplier, types] = args;
-    if (target.isGrounded() || !types.includes(PokemonType.FLYING)) {
+    // Thousand Arrows deals super-effective damage to flying types in Inverse Battles
+    // https://replay.pokemonshowdown.com/gen9nationaldex-2536869784-17h3xwyxhlb6uaj6n30acyyligy93s6pw
```
emdeann, line 6379: It's fine for quick checks but I'm not sure I want showdown links used as cites like this. It's not any more official than we are

src/data/arena-tag.ts
```diff
@@ -1226,25 +1226,30 @@ export class GravityTag extends SerializableArenaTag {
...
-              pokemonNameWithAffix: getPokemonNameWithAffix(pokemon),
-            }),
-          );
-        }
+      pokemon.removeTag(BattlerTagType.FLOATING);
+      pokemon.removeTag(BattlerTagType.TELEKINESIS);
+      if (pokemon.getTag(BattlerTagType.FLYING)) {
+        pokemon.removeTag(BattlerTagType.FLYING);
+        pokemon.addTag(BattlerTagType.INTERRUPTED);
+      }
```
emdeann, line 1248: I would make this into a helper if you're doing it more than once. Particularly because of the interaction with `FLYING`

emdeann (dismissed):
Same thing as the Dancer one, I think we should do 1.12.1 for this

src/data/moves/move.ts
Bertie690: Technically Thousand Arrows wasn't even legitimately obtainable in gen 6 (the only generation that even HAS inverse battles), so it's the best we got unless someone wants to hack/mod a 3DS

emdeann: I mean, if it's in the game it can be trivially tested on emulator regardless. But even if it's not, there's no reason to point to showdown as the golden source. At that point it would just be our own decision

src/data/arena-tag.ts
Bertie690: This would probably make most sense after the tag methods are made to take arrays instead of constructors

src/data/moves/move.ts
Bertie690: Fair enough.

Madmadness65 (dismissed):

src/field/pokemon.ts
```diff
@@ -2566,7 +2615,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
         : 1,
     );
 
-    if (this.getTypes({ returnOriginalTypesIfStellar: true }).find(t => move.isTypeImmune(source, this, t))) {
+    if (this.getTypes({ returnOriginalTypesIfStellar: true }).some(t => move.isTypeImmune(source, this, t))) {
```
DayKev: This should probably be `if (move.isTypeImmune(source, this)) {` and then the iterating over the target's types should happen in that function.

src/field/pokemon.ts
```diff
@@ -2441,17 +2447,59 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * - `true`: This Pokemon is forcibly grounded by an effect like Smack Down, Ingrain or Gravity.
+   * - `false`: This Pokemon is forcibly lifted off the ground by an effect like Magnet Rise or an active Levitate/Eelevate ability.
+   * - `undefined`: Neither of the above effects applies.
+   *
+   * @see {@linkcode isGrounded} - Main function that uses this
+   * @privateRemarks
+   * This method does not account for typing or semi-invulnerability as it is used to determine
+   * whether Ground-type moves should be nullified against airborne defenders (both of which are ignored by the latter).
+   */
+  private isForciblyGrounded(): boolean | undefined {
```
DayKev: I don't like the name of this method. An `isX` function should only ever return `true` or `false`, answering the question of whether "X" is true or not.
It should probably be split into `isForciblyGrounded` and `isForciblyUngrounded` (or whatever naming is best).

DayKev (dismissed):

Wlowscha:
As far as I can see, `isGrounded()` is never called in the code with `useIllusion` set to `true`. Is it really necessary to keep it as an argument?

DayKev:
> As far as I can see, `isGrounded()` is never called in the code with `useIllusion` set to `true`. Is it really necessary to keep it as an argument?

That parameter was added in this PR and he didn't go through the callsites to update them yet (hence the "TODO" comment). I don't know if there's anywhere it would be relevant, I didn't check.

src/field/pokemon.ts
```diff
@@ -2703,19 +2750,27 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
    * @param defenderType - The {@linkcode PokemonType} of the defender
+   * @param forciblyGrounded - Whether the user is forcibly grounded
    * @returns Whether the type immunity was bypassed
    */
   private checkIgnoreTypeImmunity({
     source,
     simulated,
     moveType,
     defenderType,
+    forciblyGrounded,
```
Wlowscha: Is this necessary to pass this as a parameter here, when we could just call `this.isForciblyGrounded()` inside the function?

src/data/abilities/ab-attrs.ts
```diff
@@ -206,6 +206,13 @@ abstract class CancelInteractionAbAttr extends AbAttr {
   }
 }
 
+/** @sealed */
+export class UngroundedAbAttr extends CancelInteractionAbAttr {
```
Wlowscha: I find this implementation to be fairly unintiuitive, since in order for the Pokémon to be ungrounded, we require that `cancelled` is equal to `true`. What are we cancelling? I would prefer a more explicit implementation as in #7564 .

src/field/pokemon.ts
DayKev: Good point, after the change I made there's no need to do this anymore. Updated it.

src/data/abilities/ab-attrs.ts
DayKev: I kinda prefer reuse when possible but it's not a big deal here so I made the change.

DayKev (approved):

Wlowscha (approved):

PR #5959, opened 2025-06-09
[Refactor] Remove circular deps 3

SirzBenjie:
# What are the changes the user will see?
None

To reduce circular dependencies.

- Moved many enums, and a few classes / methods to their own files.
- Added a `MoveAttr.is` method that takes in a string name of the move attributes
- Made all of the move methods that check for attributes take the string name of a move attribute, and then apply it.
- Added an `is` method to Moves so that files can check if a move is an instance of a move subclass without having to import the instance in order to use instanceof

**FOR REVIEWERS:**
Relevant files and methods to review
- `src/@types/move-types.ts`
- The 
The

## Screenshots/Videos
N/A

## How to test the changes?
Run tests, and also play a few runs to ensure the game runs as it always has.

## Checklist
- [x] The PR is self-contained and cannot be split into smaller PRs? Maybe it could but I'm not waiting for many smaller ones to get accepted for this.


src/data/moves/pokemon-move.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Compared to {@linkcode Move}, this class also tracks things like
+ * PP Ups recieved, PP used, etc.
+ * @see {@linkcode isUsable} - checks if move is restricted, out of PP, or not implemented.
+ * @see {@linkcode getMove} - returns {@linkcode Move} object by looking it up via ID.
+ * @see {@linkcode usePp} - removes a point of PP from the move.
+ * @see {@linkcode getMovePp} - returns amount of PP a move currently has.
+ * @see {@linkcode getPpRatio} - returns the current PP amount / max PP amount.
+ * @see {@linkcode getName} - returns name of {@linkcode Move}.
+ **/
+
```
DayKev: I wish VSCode didn't insert these random newlines when you use the "refactor" tool to move code to a different file.
```suggestion
 */
```

src/data/moves/move.ts
```diff
@@ -956,6 +978,9 @@ export default class Move implements Localizable {
 }
 
 export class AttackMove extends Move {
+  override is<K extends keyof MoveClassMap>(moveSort: K): this is MoveClassMap[K] {
```
DayKev: I don't think `moveSort` is a good variable name here. It should be either `moveType` or `moveKind` if you absolutely don't want `type` in the name.
(Same with the other instances of this).

src/data/moves/move.ts
```diff
@@ -147,6 +150,16 @@ export default class Move implements Localizable {
   private flags: number = 0;
   private nameAppend: string = "";
 
+  /**
+   * Check if the move is of the given subclass without requiring `instanceof`.
+   * 
+   * @param moveSort - The string name of the move to check against
+   * @returns Whether this move is of the provided type.
+   */
+  public is<K extends keyof MoveClassMap>(moveSort: K): this is MoveClassMap[K] {
```
DayKev: This should be `abstract` (actually this entire class should be `abstract` (and not a `default` export, I hate those in intra-project files) but that's probably for a different PR).

Also with the current class setup there's no reason for the subclasses to have `|| super.is(...)` in the `return`s.

src/@types/move-types.ts
```diff
@@ -0,0 +1,28 @@
+import type { AttackMove, StatusMove, SelfStatusMove, MoveAttrConstructorMap, MoveAttr } from "#app/data/moves/move";
+
+export type MoveAttrFilter = (attr: MoveAttr) => boolean;
+
+export type * from "#app/data/moves/move";
```
DayKev, line 13: Is there a way to enforce specific imports only happen from a certain file?

Also given type-only imports aren't emitted in the transpiled JS this is kind of irrelevant :shrug:
Well, it kind of obscures the actual place something is defined, which isn't very important... but will VSCode properly jump to the actual definition?

src/data/moves/apply-attrs.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * Module holding functions to apply move attributes.
+ * Must not import anything that is not a type.
+ */
+import type { MoveAttrString } from "#app/@types/move-types";
```
DayKev:
```suggestion
import type { MoveAttrFilter, MoveAttrString } from "#app/@types/move-types";
```

src/data/moves/apply-attrs.ts
```diff
@@ -0,0 +1,58 @@
...
+import type Pokemon from "#app/field/pokemon";
+import type { default as Move, ChargingMove, MoveAttr } from "./move";
+import type { MoveAttrFilter } from "#app/@types/move-types";
```
DayKev:
```suggestion
```

src/@types/move-types.ts
SirzBenjie: There's not a great way of ensuring specific imports only happen from a certain file. However, I found that vscode will default to the ones from `@types`, perhaps because it comes first alphabetically.

image

src/data/moves/move.ts
SirzBenjie: That's true. This is a remnant of what I had originally, which was also having just the plain `Move` in the `MoveClassMap`, and then the `is` method was checking there.
I realized there's no reason to check that it's just a `Move` so I removed it.

Will update.

Also made it so that `Move` class is abstract. Nothing was instantiating move, so it was free.

src/data/moves/move.ts
SirzBenjie: Agreed.
`moveType` is misleading, so I went with `moveKind`.

SirzBenjie:
Since this hasn't been merged and I have more time to work on it...
I'm going to give the `is` treatment to ability attributes.

src/@types/move-types.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Map of move subclass names to their respective classes.
+ * Does not include the ChargeMove subclasses. For that, use `ChargingMoveClassMap`.
+ *
+ * @privateremarks
+ * The `Never` field is necessary to ensure typescript does not improperly narrow a failed `is` guard to `never`.
```
DayKev: You mean this?
```suggestion
 * The `never` field (`declare private _: never`) in some classes is necessary
 * to ensure typescript does not improperly narrow a failed `is` guard to `never`.
```

src/@types/move-types.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+export type MoveClassMap = {
+  AttackMove: AttackMove;
+  StatusMove: StatusMove;
+  SelfStatusMove: SelfStatusMove;
+};
+
+/*
+ * Without the `Never: never` field, typescript will
+ */
```
DayKev: Looks like the comment got partially duplicated (probably when you were moving things around).
```suggestion
```

DayKev (approved):

Issue #5066, opened 2025-01-01
[Bug] DefDefattr (Psyshock, Psystrike, Secret Sword) do not Consider Fur Coat in Damage Calculation


Snailman11:
### Describe the bug

Psyshock, Psystrike, and Secret Sword do not consider the damage-reduction of Fur Coat against Physical attack during damage calculation.

### Reproduction

Use Psyshock, Psystrike, or Secret Sword against a Pokémon before and after they gain Fur Coat.

```
	MOVESET_OVERRIDE: [Moves.PSYSHOCK, Moves.ENTRAINMENT],
	ABILITY_OVERRIDE: Abilities.FUR_COAT,
	
	OPP_LEVEL_OVERRIDE: 10,
	OPP_MOVESET_OVERRIDE: [Moves.LIGHT_SCREEN]
```

### Expected behavior

https://replay.pokemonshowdown.com/gen9customgame-2273358942

Fur Coat should be considered in damage calculation, and reduce the damage done by those moves.

> Secret Sword inflicts damage. Secret Sword uses the target's Defense instead of its Special Defense to calculate damage. 

> Fur Coat doubles the Defense stat of the Pokémon with this Ability.

### Screenshots / Videos

Psyshock Fur Coat

https://github.com/user-attachments/assets/225ee184-050a-403e-a68e-deee22c51ce5

Pystrike Fur Coat

https://github.com/user-attachments/assets/b0bb236d-9486-48e6-9352-5343b9e7959c

### Additional context

DefDefattr

https://github.com/pagefaultgames/pokerogue/blob/206419c9f176cf43cb3e8639dfee175664afa5f5/src/data/move.ts#L4399

Fur Coat checks for a Physical-type Attack

https://github.com/pagefaultgames/pokerogue/blob/206419c9f176cf43cb3e8639dfee175664afa5f5/src/data/ability.ts#L5858

thisPieonFire:
I'm taking a stab at this one!

SirzBenjie:
This is not a bug, unfortunately. Showdown has this incorrect.

Fur coat cares about the damage category being used. It will matter for variable _category_ moves, like tera blast, photon geyser, and shell side arm. But for sacred sword / psyshock, it does not!

PR #5963, opened 2025-06-10
[Bug] [Ability] Fur Coat fix

sisyphos-spirit:
## Fix
Moves Psyshock, Psystryke and secret sword were not being afected by fur coat


Snailman11:
Please use the template provided

sisyphos-spirit:
> Please use the template provided

Sorry, I'm new and I was trying to figure out how to solve this https://github.com/pagefaultgames/pokerogue/issues/5066 but I think I ended up creating a new Issue.

src/data/abilities/ability.ts
```diff
@@ -8737,7 +8737,7 @@ export function initAbilities() {
       .attr(PokemonTypeChangeAbAttr),
     //.condition((p) => !p.summonData.abilitiesApplied.includes(AbilityId.PROTEAN)), //Gen 9 Implementation
     new Ability(AbilityId.FUR_COAT, 6)
-      .attr(ReceivedMoveDamageMultiplierAbAttr, (_target, _user, move) => move.category === MoveCategory.PHYSICAL, 0.5)
+      .attr(ReceivedMoveDamageMultiplierAbAttr, (_target, _user, move) => move.category === MoveCategory.PHYSICAL || move.id === MoveId.PSYSHOCK || move.id === MoveId.PSYSTRIKE || move.id === MoveId.SECRET_SWORD, 0.5)
```
TijmenvanderKemp: This checks for those specific moves, but if any move is introduced in the future that also always uses the Defense stat, this code will be incomplete again. Those moves have in common that they have the attribute DefDefAttr attached. It would be safer to check whether the move has that attribute using `move.hasAttr(DefDefAttr)`

TijmenvanderKemp: There is no test yet for the Fur Coat ability in the `test/abilities` folder. Could you add a test that tests:
1. That a physical move of your choice is affected
2. That a special move of your choice is not affected
3. That those three specific moves are also affected, even though they are Special?

TijmenvanderKemp (changes requested):

TijmenvanderKemp:
Congrats on your first PR btw :)

sisyphos-spirit:
The test is not passing yet, but I don't really know why. Did I created the test file correctly?

DayKev:
The tests are mostly correct, you just never reset the enemy's ability in the last test after each check, so it had Fur Coat for all of the attacks after the first one. I pushed a fix for the tests (you can take a look at the changes here [9049e1b](https://github.com/pagefaultgames/pokerogue/pull/5963/commits/9049e1b32bd19556760a181c3285b45a07304d12)).

sisyphos-spirit:
> The tests are mostly correct, you just never reset the enemy's ability in the last test after each check, so it had Fur Coat for all of the attacks after the first one. I pushed a fix for the tests (you can take a look at the changes here [9049e1b](https://github.com/pagefaultgames/pokerogue/pull/5963/commits/9049e1b32bd19556760a181c3285b45a07304d12)).

Ok, that makes total sense. Thanks!

src/data/abilities/ability.ts
```diff
@@ -8488,7 +8488,7 @@ export function initAbilities() {
       .attr(PokemonTypeChangeAbAttr),
     //.condition((p) => !p.summonData.abilitiesApplied.includes(AbilityId.PROTEAN)), //Gen 9 Implementation
     new Ability(AbilityId.FUR_COAT, 6)
-      .attr(ReceivedMoveDamageMultiplierAbAttr, (_target, _user, move) => move.category === MoveCategory.PHYSICAL, 0.5)
+      .attr(ReceivedMoveDamageMultiplierAbAttr, (_target, _user, move) => move.category === MoveCategory.PHYSICAL || move.hasAttr("DefDefAttr"), 0.5)
```
SirzBenjie, line 8491: This is _A_ solution, but not _the_ solution. And I'm not even sure what the best solution for this is, anyway.
Right now, for instance, I'm not sure if this would apply to Tera Blast being used as a physical attack. Or moongeist beam, or shell side-arm.

test/abilities/fur_coat.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override
+      .battleStyle("single")
+      .ability(AbilityId.BALL_FETCH)
+      .enemySpecies(SpeciesId.MILOTIC)
+      .enemyAbility(AbilityId.BALL_FETCH)
+      .enemyMoveset([MoveId.SPLASH])
+      .enemyLevel(20)
+      .disableCrits();
```
DayKev:
```suggestion
      .criticalHits(false);
```

src/data/abilities/ability.ts
emdeann: I think that doing this properly would involve changing these types of moves to actually change their category somewhere that can be checked (something like `Pokemon#getMoveCategory(move, target)` , similar to `getMoveType`?). Right now all something like psyshock does is manually change the defense value via a `NumberHolder` which is not conducive to things like this

test/abilities/fur_coat.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    await game.toEndOfTurn();
+    const damageBefore = enemy.getMaxHp() - enemy.hp;
+    // Give Fur Coat
+    game.field.mockAbility(enemy, AbilityId.FUR_COAT);
+    enemy.hp = enemy.getMaxHp();
+    // Use Tackle after Fur Coat
+    game.move.use(MoveId.TACKLE);
+    await game.toEndOfTurn();
+    const damageAfter = enemy.getMaxHp() - enemy.hp;
+    expect(damageAfter).toBeLessThan(damageBefore);
```
DayKev, line 48: Instead of just checking that the damage is lower, it should probably check that the damage is halved (`+/- 1` to account for rounding).

test/abilities/fur_coat.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    const attackDamageBefore = enemy.getMaxHp() - enemy.hp;
+
+    game.field.mockAbility(enemy, AbilityId.FUR_COAT);
+    enemy.hp = enemy.getMaxHp();
+
+    game.move.use(moveId);
+    await game.toEndOfTurn();
+
+    const attackDamageAfter = enemy.getMaxHp() - enemy.hp;
+    expect(attackDamageAfter).toBeLessThan(attackDamageBefore);
```
DayKev, line 88: Same here.

src/data/abilities/ability.ts
emdeann: wait... this method already exists exactly as i've described it. lol

All we need to do is use `Pokemon#getMoveCategory` instead of `move.category` in these checks

Snailman11:
If a mass-change to move.category is implemented then this could close
#6045

SirzBenjie:
@sisyphos-spirit
I'm closing this since it's been a few weeks without activity / response. If you wish to address the comments raised by @emdeann and @DayKev, then feel free to reopen and we can revisit this PR.

PR #5663, opened 2025-04-14
[Refactor] Remove circular imports part 1

SirzBenjie:
## What are the changes the user will see?
N/A

## Why am I making these changes?
We need to remove all sources of circular imports as they are causing issues with things being brittle and breaking.

## What are the changes from a developer perspective?
Cyclic dependency count reduced from 909 to 623.

Moved the `Mode` enum from `src/ui/ui.ts` and into its own file in `#enums/ui-mode.ts`, and renamed it to `UiMode`
This got rid of 136 cyclical imports

Created a new directory, `src/utils` and moved `src/utils.ts` to this folder, and renamed it to `common.ts`.
  - Moved the cookie functions out of this file and into their own file, `src/utils/cookies.ts`, 
  - Moved the `isBeta` utility variable to its own file.
This got rid of 8 cyclical dependencies, but was also required in order for removing things from battle-scene.ts to remove dependencies (please don't ask me why)

Created a new directory called `global-vars` for holding variables referenced by several places. I'm not good enough at naming or figuring out a better structure than to literally slap these things in their own file. 
 - Moved `bypassLogin` from `battle-scene.ts` to this folder
 - Moved `starterColors` from `battle-scene.ts` to this folder
This got rid of 142 cyclical dependencies

## Screenshots/Videos
N/A

### **Please actually test these changes by doing a run. Partially removing circular imports can cause unforeseen issues**

Run the automated tests. Also, actually manually launch the game to make sure you can get into the loading scene, as the last PR that tried to remove circular imports caused issues. (I did this myself and it works)

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs? Technically, it can be done 1-by-1, however, most of the PRs would have terrible merge conflicts. It's better to do this in one go
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test`)
  - [ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [ ] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (default and legacy)?

Are there any localization additions or changes? If so:
- [ ] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?
  - [ ] If so, please leave a link to it here: 
[...]

PR #5964, opened 2025-06-10
[Refactor] Remove circular dependencies (part 4)

SirzBenjie:
None

Removes the LAST bit of circular dependencies.
See #5959 and #5663 for a discussion.

- Moved some constants out of Mystery encounters and into `constants.ts`
- Created `initModifierPools()` and `initModifierTypes()` methods so that `modifierType` and modifiers, similar to allAbilities, is moved to `data-lists`.
- Applied the same treatment that was done to Move Attributes (in #5959) to Ability Attributes and Modifiers (and, to a lesser extend, ModifierTypes). The modifier types were not as extensive, as the modifier rework is impending.
- Moved a few more enums.
- Moved the `Ability` class back into the abilities.ts file. This was necessary as it needs to be able to have access to the list of all ability attributes.

**Files to review**
JUST look at the changes to `AbAttr` in `Abilities.ts`.
Then, look at modifiers.ts and the new files created in that folder.

That's it.

## Screenshots/Videos
N/A

## How to test the changes?
N/A

## Checklist
- [x] Have I tested the changes manually? Yeah. I tested many waves

### DIFF
--- a/src/enums/confirm-ui-mode.ts
+++ b/src/enums/confirm-ui-mode.ts
@@ -1,0 +1,13 @@
+// biome-ignore lint/correctness/noUnusedImports: Used in tsdoc
+import type ConfirmUiHandler from "#app/ui/confirm-ui-handler";
+
+/**
+ * Used by {@linkcode ConfirmUiHandler} to determine whether the cursor should start on Yes or No
+ */
+export const ConfirmUiMode = Object.freeze({
+    /** Start cursor on Yes */
+    DEFAULT_YES: 1,
+    /** Start cursor on No */
+    DEFAULT_NO: 2
+});
+export type ConfirmUiMode = typeof ConfirmUiMode[keyof typeof ConfirmUiMode];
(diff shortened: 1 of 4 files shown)


damocleas (approved):
mandatory review covered, just approving this because required

if any changes are made I do not have to re-review it clears the requirement

src/modifier/modifier.ts
```diff
@@ -164,6 +158,15 @@ export abstract class Modifier {
     this.type = type;
   }
 
+  /** */
```
DayKev: Forgot to put the docs.

src/modifier/modifier.ts
```diff
@@ -188,6 +191,9 @@ export abstract class PersistentModifier extends Modifier {
   public stackCount: number;
   public virtualStackCount: number;
 
+  /** */
```
DayKev: Same here.

DayKev (commented):
Seems good other than the 2 missing docs.

src/data/abilities/ability.ts
```diff
@@ -71,13 +69,227 @@ import type {
...
+    this.id = id;
+
+    this.nameAppend = "";
+    this.generation = generation;
+    this.attrs = [];
+    this.conditions = [];
+
+    this.isSuppressable = true;
+    this.isCopiable = true;
+    this.isReplaceable = true;
```
```suggestion
```
Actually might as well remove these now, since they're redundant (these already have default values).

PR #5972, opened 2025-06-11
[Dev] Enable `exactOptionalPropertyTypes` in `tsconfig.json`

Bertie690:
N/A

[tsconfig description](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes)

TL;DR setting property X to `undefined` is subtly different from it not being included.

This also significantly improves the type-safety of `Partial` and types with optional properties as you can no longer write stuff like this:

```ts
type StatMultsByStat = {
  [k in Stat]?: number
}

const a: StatMultsByStat = {
  Stat.HP: undefined,
  Stat.ATK: undefined,
// ...
}
```

Added thing to tsconfig.json

Fixed prior code to avoid assigning `undefined` to things where needed

Added `undefined` in all other cases

## Checklist


damocleas (approved):
approving because required, should clear if other changes are needed but I won't need to approve again

DayKev (approved):

DayKev (approved):

Bertie690:
OK apparently #6002 REALLY screwed this PR up due to the sheer amount of destructuring on interfaces with optional parameters.

PR #4311, opened 2024-09-18
[Challenge] Add Nuzlocke-related challenges

DayKev:
Todo:
- Change method of "one catch per biome" restriction
- Prevent fusions with fainted Pokémon in hardcore challenge
  - Would also affect splicing in other challenges, needs balance team discussion
- Restrict MEs that call `PartyHealPhase` from appearing during no autoheal challenge?

## What are the changes the user will see?
Adds 3 new challenges:
- "No Free Heal": Disables the automatic healing that occurs after every 10th wave, replacing it with a normal shop phase
- "Hardcore": Fainted Pokémon can't be revived
- "Limited Catch": Only the first wild Pokémon encounter
of every biome can be added to your current party (ie: from waves 1, 11, 21, ...)
  - Takes into account if the X1 wave was a ME, allowing you to catch the Pokémon on wave X2 instead

## Why am I making these changes?
Nuzlocke mode is a popular request/on the roadmap.

## What are the changes from a developer perspective?
6 new `ChallengeType`s were added:
- `NO_HEAL_PHASE`
Checks if the heal phase after every 10th wave should be activated. See `src/phases/party-heal-phase.ts`

- `SHOP_ITEM_BLACKLIST`
Filters the shop for specific items. See `src/modifier/modifier-type.ts` and `src/ui/modifier-select-ui-handler.ts`

- `RANDOM_ITEM_BLACKLIST`
Filters the random item selection for specific items. See `src/modifier/modifier-type.ts`

- `ADD_POKEMON_TO_PARTY`
Checks if the current caught Pokémon can be added to the party. See `src/phases/attempt-capture-phase.ts`

- `MOVE_BLACKLIST`
Checks if move the user wants to use is illegal for this challenge. See `src/phases/command-phase.ts`

- `PREVENT_REVIVE`
Checks if Pokémon can be revived. See `src/phases/party-heal-phase.ts`

### Screenshots/Videos
<details>
  <summary>Starter filter with "No Legends" challenge active</summary>

image

</details>
<details>
  <summary>Shop with "Hardcore" challenge active</summary>

image

</details>
<details>
  <summary>Lack of auto-heal with "No Free Heal" challenge active</summary>

https://github.com/user-attachments/assets/116755c0-a586-46cf-88a6-196cbbc9096e

</details>
<details>
  <summary>Caught Pokémon not going to party with "Limited Catch" challenge active</summary>

[...]

PR #5979, opened 2025-06-12
[Challenge] Nuzlocke related challenges and AI improvements

mati-soda:
Added 3 new Nuzlocke challenges:
- "No Free Heal": Disables the automatic healing that occurs after every 10th wave, replacing it with a normal shop phase
- "Hardcore": normal mode where fainted Pokémon can't be revived, but can stay on the player's party and an hard mode where the fainted pokemon is automatically deleted from the party.
- "Limited Catch": Only the first wild Pokémon encounter of every biome can be added to your current party (ie: from waves 1, 11, 21, ...). Takes into account if the X1 wave was a ME, allowing you to catch the Pokémon on wave X2 instead.

Enemy AI: smarter choices on whether it should sacrifice a pokemon or not based on the move effectiveness, health and speed.

Nuzlocke mode is a popular request/on the roadmap. Related to issue #4311.
Enemy AI is too predictable with it's switching, making late game too easy.

"No Free Heal": a new challenge `NO_AUTO_HEAL` of type `NO_HEAL_PHASE` was added, along with a function `applyNoHealPhase` to determine if healing should be skipped when the challenge is active (`challenge.ts`). The healing phase (`party-heal-phase.ts`) was modified to check this condition and skip healing if necessary. Additionally, the shop logic was updated so the shop appears every 10th wave when this challenge is active (`modifier-type.ts` and `victory-phase.ts`). 

"Hardcore": added a `HARDCORE` challenge with several associated types and apply functions in `challenge.ts`. These include: `RANDOM_ITEM_BLACKLIST` (filters reward items, `modifier-type.ts`), `SHOP_ITEM_BLACKLIST` (filters shop items, `modifier-select-ui-handler.ts`), `MOVE_BLACKLIST` (prevents the use of blacklisted moves, `pokemon.ts`), `DELETE_POKEMON` (deletes fainted Pokémon in hard mode, `battle-end-phase.ts`), `SHOULD_FUSE` (blocks fusion with dead Pokémon, `modifier.ts`), and `PREVENT_REVIVE` (prevents healing of fainted Pokémon, `party-heal-phase.ts`). 

"Limited Catch": introduced a new challenge `LIMITED_CATCH` with the type `ADD_POKEMON_TO_PARTY` and a function `applyAddPokemonToParty` to control when a Pokémon can be added to the party (`challenge.ts`). The Pokémon addition logic was also adjusted so that caught Pokémon update the Pokédex but are not added to the party if not eligible (`encounter-pokemon-utils.ts` and `attempt-capture-phase.ts`). 

Enemy AI was improved by altering `getMatchupScore` to better consider a Pokémon’s actual moves and their effectiveness rather than relying solely on Pokémon type, and new logic was added to decide when a Pokémon should be sacrificed or switched based on its HP and speed. (`pokemon.ts`)

## Screenshots/Videos

https://github.com/user-attachments/assets/04dd2ba3-bbfc-4ef0-be7b-669fcb483fbd

## How to test the changes?
Start a run with the selected challenges.

## Checklist
- [X] Are all unit tests still passing? (`npm run test:silent`)


EscaShark:
The enemy AI adjustment should ideally be a separate PR altogether.

mati-soda:
> The enemy AI adjustment should ideally be a separate PR altogether.

Just created another PR for the AI adjustments. Thank you, I have also removed the AI changes from this PR.

damocleas:
I don't think we need to be disabling Mystery Encounters for the end of wave heal challenge or anything, I think it's a good thing to be able to chance into.

And don't worry about the free Sacred Ash from "A Trainer's Test", I'll be making a change to it in #5960, so no need to worry about a forced Sacred Ash causing issues either

shadownocas:
Hi, thank you so much for the answer! I just changed what you requested. Let me know what I could improve.

mati-soda:
Thank you so much for the answer! I just removed the auto heal logiv from the Mystery Encounters. If something else needs to be changed let me know.

mati-soda:
@SirzBenjie Hello, sorry for bothering, could you review this pull request to see if it meets all the requirements and is ready to be merged?

damocleas:
I am not sure what to name it, but maybe the "No Free Heal" should be a 3 option thing with "No Shop / No Free Heal / No Shop + Free Heal", __No Shop__ is currently used in Daily Mode always, and is a good challenge to have with this as that is the main source of healing.

I have no clue how to name this, good to bring back to the discord.

damocleas (changes requested):
I am not sure what to name it, but after testing this, I feel like the "No Free Heal" should be a 3 option thing with "No Shop / No Free Heal / No Shop + Free Heal"

The last of the three is likely almost impossible in a game like this.. But the option being there seems interesting enough. Using only the items given to you by the item table.

__No Shop__ is currently used in Daily Mode always, and is a good challenge to have with this as that is the main source of healing in runs.

I have no clue how to name this, good to bring back to the discord.

We can probably add an achievement in a separate PR from this before next update that is Hardcore + Limited Catch for completing a "Nuzlocke"

src/data/challenge.ts
```diff
@@ -888,6 +968,123 @@ export class LowerStarterPointsChallenge extends Challenge {
...
+      canStay.value = true;
+    }
+    return true;
+  }
+
+  override applyShouldFuse(pokemon: Pokemon, pokemonToFuse: Pokemon, canFuse: BooleanHolder): boolean {
+    if (pokemon!.isFainted() || pokemonToFuse.isFainted()) {
+      canFuse.value = false;
+    }
+    return true;
```
Fontbane:
```suggestion
      canFuse.value = false;
      return true;
    }
    return false;
```
Should only return true if the challenge does something

src/data/challenge.ts
```diff
@@ -344,6 +346,84 @@ export abstract class Challenge {
...
+    return false;
+  }
+
+  /**
+   * An apply function for PREVENT_REVIVE. Derived classes should alter this.
+   * @param _canBeRevived {@link BooleanHolder} Whether it should revive the fainted Pokemon.
+   * @returns {@link boolean} if this function did anything.
+   */
+  applyRevivePrevention(_canBeRevived: BooleanHolder): boolean {
+    return true;
```
Fontbane:
```suggestion
    return false;
```

src/data/challenge.ts
```diff
@@ -888,6 +968,123 @@ export class LowerStarterPointsChallenge extends Challenge {
...
+  }
+
+  applyDeletePokemon(canStay: BooleanHolder): boolean {
+    if (this.value === 2) {
+      canStay.value = false;
+    } else {
+      canStay.value = true;
+    }
+    return true;
```
Fontbane:
```suggestion
      return true;
    }
    return false;
```

src/data/challenge.ts
```diff
@@ -888,6 +968,123 @@ export class LowerStarterPointsChallenge extends Challenge {
...
+
+  applyShopItemBlacklist(shopItem: ModifierTypeOption, isValid: BooleanHolder): boolean {
+    isValid.value = !this.itemBlackList.includes(shopItem.type.localeKey);
+    return true;
+  }
+
+  applyMoveBlacklist(move: PokemonMove, moveCanBeUsed: BooleanHolder): boolean {
+    const moveBlacklist = [MoveId.REVIVAL_BLESSING];
+    moveCanBeUsed.value = !moveBlacklist.includes(move.moveId);
+    return true;
```
Fontbane:
```suggestion
    const moveBlacklist = [MoveId.REVIVAL_BLESSING];
    if (moveBlacklist.includes(move.moveId)) {
      moveCanBeUsed.value = false;
      return true;
    }
    return false;
```

`moveBlacklist` can probably be a private class variable like `itemBlacklist` is

src/data/challenge.ts
```diff
@@ -888,6 +968,123 @@ export class LowerStarterPointsChallenge extends Challenge {
...
+
+  override applyAddPokemonToParty(waveIndex: number, canAddToParty: BooleanHolder): boolean {
+    const lastMystery = globalScene.lastMysteryEncounter?.encounterType;
+    if (lastMystery === undefined && !(waveIndex % 10 === 1)) {
+      canAddToParty.value = false;
+    }
+    if (!(waveIndex % 10 === 1) && !(!this.mysteryEncounterBlacklist.includes(lastMystery!) && waveIndex % 10 === 2)) {
+      canAddToParty.value = false;
+    }
+    return true;
```
Fontbane:
```suggestion
    if (waveIndex % 10 !== 1) {
      const lastMystery = globalScene.lastMysteryEncounter?.encounterType;
      if (isNullOrUndefined(lastMystery) || !(waveIndex % 10 === 2 && !this.mysteryEncounterBlacklist.includes(lastMystery!)) {
        canAddToParty.value = false;
        return true;
      }
    return false;
```
You'll have to import `isNullOrUndefined` for this, surprised it's not used by other challenges

Fontbane (commented):
Some nits/suggestions. The parts relating to modifiers might have to wait until after the modifier rework.
You'll have to document your functions.
The shop appearing on the 10th wave is cool.

mati-soda:
> I am not sure what to name it, but after testing this, I feel like the "No Free Heal" should be a 3 option thing with "No Shop / No Free Heal / No Shop + Free Heal"
> 
> The last of the three is likely almost impossible (gameplay wise) in a game like this for most players, but the option being there seems interesting enough. Using only the items given to you by the item table.
> 
> **No Shop** is currently used in Daily Mode (with the free heal on biome transitions active) always, and is a good challenge to have with this as that is the main source of healing in runs.
> 
> I have no clue how to name this, good to bring back to the discord.
> 
> We can probably add an achievement in a separate PR from this before next update that is Hardcore + Limited Catch for completing a "Nuzlocke"

Thank you for the feedback, just implemented the other two options for the No Free Heal challenge, I would be happy to discuss with you the name and the possible changes to the locales repository this will require.

mati-soda:
> Some nits/suggestions. The parts relating to modifiers might have to wait until after the modifier rework. You'll have to document your functions. The shop appearing on the 10th wave is pretty cool.
Thank you for reviewing the code, I just made the changes you suggested.

src/data/challenge.ts
```diff
@@ -889,6 +978,143 @@ export class LowerStarterPointsChallenge extends Challenge {
...
+
+/**
+ * Challenge that removes the ability to revive fallen pokemon
+ */
+export class HardcoreChallenge extends Challenge {
+  private itemBlackList = [
+    "modifierType:ModifierType.REVIVE",
+    "modifierType:ModifierType.MAX_REVIVE",
+    "modifierType:ModifierType.SACRED_ASH",
+    "modifierType:ModifierType.REVIVER_SEED",
```
Fontbane, line 1022: Actually I'm not sure it's necessary to blacklist Reviver Seed, since ~~it shouldn't be able to function anyway (due to `applyRevivePrevention`) and~~ it could still be held by a mon to be exchanged in a Mystery Encounter like Delibirdy.

damocleas: I don't see the problem with reviver seed working in this personally, it's an item made to prevent something being knocked out

if it gets disabled by the other thing then this needs to change

Fontbane: It appears Reviver Seed still functions normally. We've talked it out and decided it should be removed from the blacklist.

Wlowscha:
I have reviewed this and I like the looks of it overall. I have a couple of comments:

1) Answering Font's comment:
> The parts relating to modifiers might have to wait until after the modifier rework.

From what I have seen, the Nuzlocke challenges only interact with the shop/reward phase, which means there are no significant conflicts with #5656. Green lights from me on that front.

2) My understanding is that (so far at least) there are no locales changes associated with this PR. I believe several of the new challenge modes would benefit from customized messages, for example:
- When capturing a Pokémon which is then not added to the party;
- When a fainted Pokémon is deleted from the party;
- When a move is unusable due to being blacklisted;
and possibly more. It looks like the changes to `src/phases/command-phase.ts` are going in this direction, but a custom message would be preferable.

Can't wait to play nuzlocke mode on main :)

src/data/challenge.ts
DayKev: > I don't see the problem with reviver seed working in this personally, it's an item made to prevent something being knocked out

Well, that's not what it does, it revives the Pokémon immediately after it faints, similar to its origin game.

damocleas: we voted on it and it was decided it is allowed since the pokemon isn't actually knocked out, and is a very rare item anyways

DayKev: Leaving it in is fine, but it is literally knocked out in the code.

Xavion3:
Can we make the internal name of hardcore descriptive of its function? Probably something like NO_REVIVE? It's fine to call it whatever we want in the text, but hardcore isn't very illustrative of its function in the code.

I'm still reviewing the whole thing, but so far this has stood out to me.

Limited Catch also feels like it would be a great use of the 2d scaling, but for now a second level which is one catch per biome instead of just first catch per biome feels like it makes sense. We'd probably need to finally implement some of that arbitrary session data tracking though so it's trickier.

Fontbane:
> Limited Catch also feels like it would be a great use of the 2d scaling, but for now a second level which is one catch per biome instead of just first catch per biome feels like it makes sense. We'd probably need to finally implement some of that arbitrary session data tracking though so it's trickier.

In theory you'd only need to record the last wave you caught something. But I'd like catch tracking (like, how many mons you catch for every biome in a run) to go alongside biome tracking anyway which will come soon with the biome rework (hopefully). Can probably be added later.

damocleas:
closed in favor of #6186 which was built off of this

PR #5981, opened 2025-06-12
 [Balance] Implemented AI improvements

mati-soda:

Enemy AI makes smarter choices on whether it should sacrifice a pokemon or not based on the move effectiveness, health and speed.

Enemy AI is too predictable with it's switching, making late game too easy.

Enemy AI was improved by altering getMatchupScore to better consider a Pokémon’s actual moves and their effectiveness rather than relying solely on Pokémon type, and new logic was added to decide when a Pokémon should be sacrificed or switched based on its HP and speed. (pokemon.ts)

## Checklist
- [X] Are all unit tests still passing? (`npm run test:silent`)

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -2504,14 +2504,39 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       defScore *=
         1 / Math.max(this.getAttackTypeEffectiveness(enemyTypes[1], opponent, false, false, undefined, true), 0.25);
     }
+    atkScore *= 1.25; //give more value for the pokemon's typing
+    const moveset = this.moveset;
+    let moveAtkScoreLenght = 0;
+    for (const move of moveset) {
+      if (move.getMove().category === MoveCategory.SPECIAL || move.getMove().category === MoveCategory.PHYSICAL) {
+        atkScore += opponent.getAttackTypeEffectiveness(move.getMove().type, this, false, true, undefined, true);
+        moveAtkScoreLenght++;
+      }
+    }
+    atkScore = atkScore / (moveAtkScoreLenght + 1); //calculate the median for the attack score
     /**
      * Based on this Pokemon's HP ratio compared to that of the opponent.
      * This ratio is multiplied by 1.5 if this Pokemon outspeeds the opponent;
      * however, the final ratio cannot be higher than 1.
      */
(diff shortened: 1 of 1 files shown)


damocleas (approved):
might be fine, hard to test on my own though to feel anything

SirzBenjie (approved):
Code looks good from an implementation standpoint.
It warrants some testing to see how much of an improvement. But i don't think it can be worse than the existing...

src/field/pokemon.ts
```diff
@@ -2503,14 +2503,39 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
-      hpDiffRatio = Math.min(hpDiffRatio * 1.5, 1);
+    const hpRatio = this.getHpRatio();
+    const oppHpRatio = opponent.getHpRatio();
+    const isDying = hpRatio <= 0.2;
+    let hpDiffRatio = hpRatio + (1 - oppHpRatio);
+    if (isDying && this.isActive(true)) {
+      //It might be a sacrifice candidate if hp under 20%
+      const badMatchup = atkScore < 1.5 && defScore < 1.5;
+      if (!outspeed && badMatchup) {
+        //It might not be a worthy sacrifice if it doesn't outspeed or doesn't do enough damage
```
SirzBenjie, line 2524: What if it has priority moves?

SirzBenjie: Actually, if the mon has priority, the AI that makes move selection should know to weight priority moves heavily in the case that it has low HP against a faster opponent. Given that our AI move selection _isn't_ doing that, it doesn't make sense to consider priority here, either.

src/field/pokemon.ts
```diff
@@ -2503,14 +2503,39 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
       defScore *=
         1 / Math.max(this.getAttackTypeEffectiveness(enemyTypes[1], opponent, false, false, undefined, true), 0.25);
     }
+    atkScore *= 1.25; //give more value for the pokemon's typing
+    const moveset = this.moveset;
+    let moveAtkScoreLenght = 0;
```
```suggestion
    let moveAtkScoreLength = 0;
```

Umbraful: Thank you for pointing that out. Just commited the fix.

DayKev (approved):

Issue #5982, opened 2025-06-13
[Bug] Speed boost still acitvates on slower mon when failing to flee


Bertie690:
### Describe the bug

When failing to flee in doubles, Speed Boost still activates on the slower pokemon (the one NOT selected for the switch out checks) at the end of the turn.

![Image](https://github.com/user-attachments/assets/d02c9bb5-f721-4b18-bdbc-61a1beb84d44)

Looking at the code, it looks like this is only updated for 1 pokemon (rather than both).

(Come to think of it, this phase shouldn't take a pokemon AT ALL given it coaclesces the speed stats of all pokemon on field and does all its checks manually. If we're checking all the on-field pokemon manually, why do we not just check run away and such there?)

### Reproduction

```ts
const overrides = {
  ABILITY_OVERRIDE: AbilityId.SPEED_BOOST,
  OPP_HEALTH_SEGMENTS_OVERRIDE: 2,
  OPP_LEVEL_OVERRIDE: 100,
  BATTLE_STYLE_OVERRIDE: "double",
  OPP_MOVESET_OVERRIDE: MoveId.SPLASH,
} satisfies Partial<InstanceType<OverridesType>>;
```

spam running away

### Expected behavior

Should not boost spd of either pokemon
![Image](https://github.com/user-attachments/assets/493e948c-904d-4d0a-a017-26d559ce6eec)

### Additional context

Good first issue candidate, thhough arguably will require a minor refactor for attempt run phase & turn start phase

PR #5999, opened 2025-06-17
[BUG] Fix Speed Boost activating on slower Pokemon on failed run attempt in doubles battles.

jnotsknab:
User's slower pokemon will no longer gain speed boost when failing to run in a double battle.

- Fixes bug referenced in issue #5982,  caused by only checking **failedRunAway** bool for player's first pokemon in battle.
- In double battles speed boost is still applied to the second/slower Pokemon due to this check.
- Thus, the speed stat on the slower Pokemon could be exploited.

- We now reference the player's Pokemon directly from the returned array from the `getPlayerField()` method.
-  In the `AttemptRunPhase` class the else block entered on fail now checks if  `globalScene.currentBattle.double` is true.
- If so we set the second Pokemon's  **failedRunAway** bool to true.
- This change ensures the second Pokemon now fails the `canApplyPostTurn()` check in the `SpeedBoostAbAttr`class, preventing the speed boost from applying.

## Screenshots/Videos
Second mon gains speed boost, should not occur.

https://github.com/user-attachments/assets/222e9af4-ac29-4f2c-8ad8-9e57665c7e30

Second mon no longer gains the boost.

[After.webm](https://github.com/user-attachments/assets/8683e85e-56ae-4f7b-b438-b28646d547df)

## How to test the changes?

Use the following overrides and attempt to run away from battle, both of your Pokemon should no longer recieve a speed boost.

<pre><code>const overrides = { ABILITY_OVERRIDE: AbilityId.SPEED_BOOST, OPP_HEALTH_SEGMENTS_OVERRIDE: 2, OPP_LEVEL_OVERRIDE: 100, BATTLE_STYLE_OVERRIDE: "double", OPP_MOVESET_OVERRIDE: MoveId.SPLASH, } satisfies Partial&lt;InstanceType&lt;OverridesType&gt;&gt;;</code></pre>

## Checklist


jnotsknab:
I agree, while this PR fixes the immediate bug, the current run phase logic particularly the coupling to individual pokemon could use refactoring. 

I can follow up with a separate PR that implements a minor refactor to the run phase and decouples it from each pokemon. Once that’s in place, I can update this PR if needed to align with the altered implementation.

jnotsknab:
Merged into #6017, no longer relevant.

PR #6001, opened 2025-06-17
[i18n] Map "biome" namespace to the filename change to "biomes"

DayKev:
An update to the locales submodule is included as it is necessary.

Our code linter + formatter (Biome) expects that any file named "biome.json" is a config file, but our locales had files named "biome.json" that were not config files for Biome. They have been renamed "biomes.json" to prevent them from being erroneously considered as config files.

There should be no change to the development experience, "biome" should still be a valid locales namespace as it has been mapped to the file rename to prevent the need for any change from developers.

## Screenshots/Videos
image

## How to test the changes?
Load up any run to ensure the biome name is still properly localized.

## Checklist

  - [x] If so, please leave a link to it here: https://github.com/pagefaultgames/pokerogue-locales/pull/186

### DIFF
--- a/src/plugins/i18n.ts
+++ b/src/plugins/i18n.ts
@@ -101,6 +101,7 @@ const namespaceMap = {
   doubleBattleDialogue: "dialogue-double-battle",
   splashMessages: "splash-texts",
   mysteryEncounterMessages: "mystery-encounter-texts",
+  biome: "biomes",
 };
 
 //#region Functions


Madmadness65 (approved):

SirzBenjie:
Hopefully it's not too confusing that we have a namespace that differs from the actual filename.

DayKev:
> Hopefully it's not too confusing that we have a namespace that differs from the actual filename.

We already have a few actually, and I don't think it's been a problem so far?

PR #6002, opened 2025-06-17
[Refactor] [Ability] Ab attr apply type safety

SirzBenjie:
Hopefully, none.
It's possible that there are minor logic issues similar to the MEP refactor related to not translating all callsites with the exact same logic. I did check to ensure this was happening.

The existing system for applying abilities is cumbersome. Every new ability attribute that needs different parameters has to make a new apply method. There are also so many apply methods. What's worse, nothing is preventing an ability attribute from being called with the wrong apply method, meaning that typescript will not properly lint the parameters that are expected to be passed.
For example, a developer could unwittingly use the base `applyAbAttrs` to invoke a `preDefendAbAttr`. This ability attribute requires more parameters.

- Condensed almost all of the various `applyXAbAttrs` into just one method, `applyAbAttrs`, that is fully type safe.
- Left in `applyOnGainAbAttrs` and `applyOnLoseAbAttrs` as these can be used to apply different ability attributes that should be invoked without requiring the caller to be aware of all of the different ability attributes that must be invoked. This is different than the normal method of just using a base class, as these attributes may not share a base class that is allowed to be invoked by the generic `applyAbAttrs` method.
- Created `@types/ability-types.ts` to be a centralized place for the types related to abilities and their application
- Created `@types/type-helpers.ts` that exposes some helpful utility types that are used in ability attributes
- Removed the `args` attribute for all ability apply methods and replaced them with type-safe interfaces. This has numerous advantages including, but not limited to:
   - The tsdoc for the parameters only needs to appear in one place -- in the interface definition. This reduces duplication of comments, while preserving the ability to see information about the parameter in hover info
   - Ensures that all required parameters are present when invoking an ability attribute's method. For example, many ability attributes require a `BooleanHolder` for their logic. In some cases, ability attributes were checking that the proper entry in args existed and was a boolean holder. This can now be safely handled by typescript; these runtime checks are no longer necessary.
   - Opens up the possibility for things like booleans and numbers to no longer need to use a holder for mutation, as their values can be stored in an object (this can be a different PR)- Updated the callsites for all 
Other than the aforementioned changes to ability attributes, also did a few things
- Added TODOs inside many ability attributes when I encountered suspicious behavior that should be addressed (though not in this PR to keep the scope focused)
- Removed the ability for the `status` parameter in `Pokemon#canSetStatus` to be undefined. This was necessary for proper type safety when invoking the attributes of certain abilities.
   - This required updating the `Pokemon#trySetStatus` method to return false early if provided an undefined status.
   - Also required updating Psycho Shift's move attribute implementation.

## Screenshots/Videos
N/A

## How to test the changes?
Run our test suite.
Can also go through and ensure the game still plays smoothly.

This is a very disruptive PR, as it is impacting all instances where ability attributes are called, and the bodies of every single ability attribute.

## Checklist
  - ~~[ ] Have I created new automated tests (`npm run test:create`) or updated existing tests related to the PR's changes?~~

### DIFF
--- a/src/@types/ability-types.ts
+++ b/src/@types/ability-types.ts
@@ -27,7 +27,7 @@ export type AbAttrMap = {
 };
 
 /**
- * Subset of ability attribute classes that may be passed to {@linkcode applyAbAttrs } method
+ * Subset of ability attribute classes that may be passed to {@linkcode applyAbAttrs} method
  *
  * @remarks
  * Our AbAttr classes violate Liskov Substitution Principle.
--- a/src/@types/ability-types.ts
+++ b/src/@types/ability-types.ts
@@ -4,6 +4,9 @@ import type Pokemon from "#app/field/pokemon";
 import type { BattleStat } from "#enums/stat";
 import type { AbAttrConstructorMap } from "#app/data/abilities/ability";
 
+// biome-ignore lint/correctness/noUnusedImports: Used in a tsdoc comment
+import type { applyAbAttrs } from "#app/data/abilities/apply-ab-attrs";
+
 // Intentionally re-export all types from the ability attributes module
 export type * from "#app/data/abilities/ability";
 
@@ -25,3 +28,16 @@ export type AbAttrString = keyof AbAttrConstructorMap;
 export type AbAttrMap = {
   [K in keyof AbAttrConstructorMap]: InstanceType<AbAttrConstructorMap[K]>;
 };
+
+/**
+ * Subset of ability attribute classes that may be passed to {@linkcode applyAbAttrs } method
+ *
+ * @remarks
+ * Our AbAttr classes violate Liskov Substitution Principal.
+ *
+ * AbAttrs that are not in this have subclasses with apply methods requiring different parameters than
+ * the base apply method.
+ *
+ * Such attributes may not be passed to the {@linkcode applyAbAttrs} method
+ */
+export type CallableAbAttrString = Exclude<AbAttrString, "PreDefendAbAttr" | "PreAttackAbAttr">;
--- a/src/@types/ability-types.ts
+++ b/src/@types/ability-types.ts
@@ -1,17 +1,14 @@
-import type { AbAttr } from "#app/data/abilities/ability";
 import type Move from "#app/data/moves/move";
 import type Pokemon from "#app/field/pokemon";
 import type { BattleStat } from "#enums/stat";
 import type { AbAttrConstructorMap } from "#app/data/abilities/ability";
 
+// intentionally re-export all types from abilities to have this be the centralized place to import ability types
+export type * from "#app/data/abilities/ability";
+
 // biome-ignore lint/correctness/noUnusedImports: Used in a tsdoc comment
 import type { applyAbAttrs } from "#app/data/abilities/apply-ab-attrs";
 
-// Intentionally re-export all types from the ability attributes module
-export type * from "#app/data/abilities/ability";
-
-export type AbAttrApplyFunc<TAttr extends AbAttr> = (attr: TAttr, passive: boolean, ...args: any[]) => void;
-export type AbAttrSuccessFunc<TAttr extends AbAttr> = (attr: TAttr, passive: boolean, ...args: any[]) => boolean;
 export type AbAttrCondition = (pokemon: Pokemon) => boolean;
 export type PokemonAttackCondition = (user: Pokemon | null, target: Pokemon | null, move: Move) => boolean;
 export type PokemonDefendCondition = (target: Pokemon, user: Pokemon, move: Move) => boolean;
@@ -33,11 +30,17 @@ export type AbAttrMap = {
  * Subset of ability attribute classes that may be passed to {@linkcode applyAbAttrs } method
  *
  * @remarks
- * Our AbAttr classes violate Liskov Substitution Principal.
+ * Our AbAttr classes violate Liskov Substitution Principle.
  *
  * AbAttrs that are not in this have subclasses with apply methods requiring different parameters than
  * the base apply method.
  *
  * Such attributes may not be passed to the {@linkcode applyAbAttrs} method
  */
-export type CallableAbAttrString = Exclude<AbAttrString, "PreDefendAbAttr" | "PreAttackAbAttr">;
+export type CallableAbAttrString =
+  | Exclude<AbAttrString, "PreDefendAbAttr" | "PreAttackAbAttr">
+  | "PreApplyBattlerTagAbAttr";
+
+export type AbAttrParamMap = {
+  [K in keyof AbAttrMap]: Parameters<AbAttrMap[K]["apply"]>[0];
+};
--- a/src/@types/type-helpers.ts
+++ b/src/@types/type-helpers.ts
@@ -1,0 +1,34 @@
+/*
+ * A collection of custom utility types that aid in type checking and ensuring strict type conformity
+ */
+
+// biome-ignore lint/correctness/noUnusedImports: Used in a tsdoc comment
+import type { AbAttr } from "./ability-types";
+
+/** Exactly matches the type of the argument, preventing adding additional properties.
+ * ⚠️ Should never be used with `extends`, as this will nullify the exactness of the type.
+ * As an example, used to ensure that the parameters of {@linkcode AbAttr#canApply} and {@linkcode AbAttr#getTriggerMessage} are compatible with
+ * the type of the apply method
+ *
+ * @typeParam T - The type to match exactly
+ */
+export type Exact<T> = {
+  [K in keyof T]: T[K];
+};
+
+/**
+ * Type hint that indicates that the type is intended to be closed to a specific shape.
+ *
+ * @remarks
+ * Can be used to ensure that a particular parameter does not change when subclassed, **But will only do so for one level of inheritance**
+ * ```ts
+ * type foo = { } // whatever
+ * class Foo {
+ *  method<T>(param: Closed<T, foo>)
+ * }
+ * ```
+ *
+ * @typeParam T - The type to check
+ * @typeParam Shape - The shape to match against
+ */
+export type Closed<X, Shape> = X extends Shape ? (Shape extends X ? X : never) : never;
--- a/src/@types/type-helpers.ts
+++ b/src/@types/type-helpers.ts
@@ -22,3 +22,11 @@ export type Exact<T> = {
  *
  */
 export type Closed<X> = X;
+
+/**
+ * Remove `readonly` from all properties of the provided type
+ * @typeParam T - The type to make mutable
+ */
+export type Mutable<T> = {
+  -readonly [P in keyof T]: T[P];
+};
--- a/src/@types/type-helpers.ts
+++ b/src/@types/type-helpers.ts
@@ -5,9 +5,12 @@
 // biome-ignore lint/correctness/noUnusedImports: Used in a tsdoc comment
 import type { AbAttr } from "./ability-types";
 
-/** Exactly matches the type of the argument, preventing adding additional properties.
+/**
+ * Exactly matches the type of the argument, preventing adding additional properties.
+ *
  * ⚠️ Should never be used with `extends`, as this will nullify the exactness of the type.
- * As an example, used to ensure that the parameters of {@linkcode AbAttr#canApply} and {@linkcode AbAttr#getTriggerMessage} are compatible with
+ *
+ * As an example, used to ensure that the parameters of {@linkcode AbAttr.canApply} and {@linkcode AbAttr.getTriggerMessage} are compatible with
  * the type of the apply method
  *
  * @typeParam T - The type to match exactly
@@ -19,7 +22,6 @@ export type Exact<T> = {
 /**
  * Type hint that indicates that the type is intended to be closed to a specific shape.
  * Does not actually do anything special, is really just an alias for X.
- *
  */
 export type Closed<X> = X;
 
--- a/src/@types/type-helpers.ts
+++ b/src/@types/type-helpers.ts
@@ -18,17 +18,7 @@ export type Exact<T> = {
 
 /**
  * Type hint that indicates that the type is intended to be closed to a specific shape.
+ * Does not actually do anything special, is really just an alias for X.
  *
- * @remarks
- * Can be used to ensure that a particular parameter does not change when subclassed, **But will only do so for one level of inheritance**
- * ```ts
- * type foo = { } // whatever
- * class Foo {
- *  method<T>(param: Closed<T, foo>)
- * }
- * ```
- *
- * @typeParam T - The type to check
- * @typeParam Shape - The shape to match against
  */
-export type Closed<X, Shape> = X extends Shape ? (Shape extends X ? X : never) : never;
+export type Closed<X> = X;
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -81,6 +81,7 @@ import type { ArenaTrapTag, SuppressAbilitiesTag } from "#app/data/arena-tag";
 import type { Constructor } from "#app/utils/common";
 import type { Localizable } from "#app/@types/locales";
 import { applyAbAttrs } from "./apply-ab-attrs";
+import type { Closed, Exact } from "#app/@types/type-helpers";
 
 export class Ability implements Localizable {
   public id: AbilityId;
@@ -135,7 +136,8 @@ export class Ability implements Localizable {
     if (!targetAttr) {
       return [];
     }
-    return this.attrs.filter((a): a is AbAttrMap[T] => a instanceof targetAttr);
+    // TODO: figure out how to remove the `as AbAttrMap[T][]` cast
+    return this.attrs.filter((a): a is AbAttrMap[T] => a instanceof targetAttr) as AbAttrMap[T][];
   }
 
   /**
@@ -220,6 +222,30 @@ export class Ability implements Localizable {
   }
 }
 
+/**
+ * Base set of parameters passed to every ability attribute's apply method
+ */
+export interface AbAttrBaseParams {
+  /**
+   * The pokemon that has the ability being applied
+   */
+  readonly pokemon: Pokemon;
+
+  /**
+   * Whether the ability application results in the interaction being cancelled
+   */
+  readonly cancelled: BooleanHolder;
+
+  /** Whether the ability's effects are being simulated.
+   * Used to prevent, for instance, messages flyouts from being displayed.
+   * Defaults to false.
+   */
+  readonly simulated?: boolean;
+
+  /** Whether the ability is the passive ability. Default false */
+  readonly passive?: boolean;
+}
+
 export abstract class AbAttr {
   public showAbility: boolean;
   private extraCondition: AbAttrCondition;
@@ -250,23 +276,21 @@ export abstract class AbAttr {
   }
 
   /**
-   * Applies ability effects without checking conditions
-   * @param _pokemon - The pokemon to apply this ability to
-   * @param _passive - Whether or not the ability is a passive
-   * @param _simulated - Whether the call is simulated
-   * @param _args - Extra args passed to the function. Handled by child classes.
-   * @see {@linkcode canApply}
+   * Apply ability effects without checking conditions.
+   *
+   * For a description of parameters, see {@linkcode AbAttrBaseParams}
+   * @see {@linkcode AbAttrBaseParams}
    */
-  apply(
-    _pokemon: Pokemon,
-    _passive: boolean,
-    _simulated: boolean,
-    _cancelled: BooleanHolder | null,
-    _args: any[],
-  ): void {}
+  apply(_params: AbAttrBaseParams): void {}
 
-  getTriggerMessage(_pokemon: Pokemon, _abilityName: string, ..._args: any[]): string | null {
+  // The NoInfer in the next two signatures enforces that the type of the _params operand
+  // is always compatible with the type of apply. This allows fewer fields, but never a parameter with more.
+  getTriggerMessage(_params: Exact<Parameters<this["apply"]>[0]>, _abilityName: string): string | null {
     return null;
+  }
+
+  canApply(_params: Exact<Parameters<this["apply"]>[0]>): boolean {
+    return true;
   }
 
   getCondition(): AbAttrCondition | null {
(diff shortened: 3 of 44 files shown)


src/@types/ability-types.ts
```diff
@@ -25,3 +25,22 @@ export type AbAttrString = keyof AbAttrConstructorMap;
...
+ * Subset of ability attribute classes that may be passed to {@linkcode applyAbAttrs } method
+ *
+ * @remarks
+ * Our AbAttr classes violate Liskov Substitution Principle.
+ *
+ * AbAttrs that are not in this have subclasses with apply methods requiring different parameters than
+ * the base apply method.
+ *
+ * Such attributes may not be passed to the {@linkcode applyAbAttrs} method
+ */
```
emdeann, line 39: It would be useful to clarify here that these types are still callable as long as it's the particular subclasses that are used

emdeann (dismissed):
Pretty much all looks good to me. I'd prefer to avoid the need for `ts-expect-error` but the logic is sound, seems like something TS should be able to handle. Tested ingame and seems to run fine. With so many minor callsite changes we may run into a bug or two after merge but that's okay

src/@types/ability-types.ts
SirzBenjie: But that means that the types _aren't_ callable. You can't call `applyAbAttr("PreAttackAbAttr")`, you'd have to call one of its subclasses.
I don't see how that would translate to the type being callable in that case?

SirzBenjie:
> Pretty much all looks good to me. I'd prefer to avoid the need for `ts-expect-error` but the logic is sound, seems like something TS should be able to handle. Tested ingame and seems to run fine. With so many minor callsite changes we may run into a bug or two after merge but that's okay

There isn't a good way of getting typescript to track the types together properly because of the use of generics here.
Even if we explicitly _cast_ the type of params to be the type of the params operand to the `attr`'s method, typescript will still error.
Consider this:
```ts
type paramType = Parameters<typeof attr["canApply"]>[0];
if ((condition && !condition(pokemon)) || !attr.canApply(params as paramType)) {
``` 
Typescript will _still_ emit an error on the `params` there. And this is with us literally telling it that the type of params _is_ the type of the first argument to its `canApply` method.
Despite this, typescript wants to ensure that `params` matches the type of the argument for _all_ ability attributes, even though this is not necessary.

I can get rid of the need for a suppression comment here by casting using `as any`. That is a lot better, so I will do that.

emdeann (dismissed):

emdeann (dismissed):

Madmadness65 (dismissed):

src/data/abilities/ability.ts
```diff
@@ -250,25 +286,23 @@ export abstract class AbAttr {
...
-   * Applies ability effects without checking conditions
-   * @param _pokemon - The pokemon to apply this ability to
-   * @param _passive - Whether or not the ability is a passive
-   * @param _simulated - Whether the call is simulated
-   * @param _args - Extra args passed to the function. Handled by child classes.
-   * @see {@linkcode canApply}
+   * Apply ability effects without checking conditions.
+   *
+   * For a description of parameters, see {@linkcode AbAttrBaseParams}
+   * @see {@linkcode AbAttrBaseParams}
```
DayKev: This TSDoc may not be ideal (unless you override it for every subclass using a different interface):
image

src/data/abilities/ability.ts
```diff
@@ -1128,75 +864,55 @@ export class MoveImmunityStatStageChangeAbAttr extends MoveImmunityAbAttr {
...
-    move: Move,
-    cancelled: BooleanHolder,
-    args: any[],
-  ): void {
-    super.applyPreDefend(pokemon, passive, simulated, attacker, move, cancelled, args);
+  override canApply(params: MoveImmunityAbAttrParams): boolean {
+    // TODO: Evaluate whether it makes sense to check against simulated here.
+    // We likely want to check 'simulated' when the apply method enqueues the phase
+    return !params.simulated && super.canApply(params);
+  }
```
DayKev, line 868: Yeah, `canApply` shouldn't be looking at the `simulated` param ever, that's only for making sure the game state isn't changed during AI calculations and thus preventing move/ability effects (and flyouts/messages/etc) from occurring.
I wonder if it should even be a param of `[ab]Attr`s at all, instead just being handled by the `apply[Ab|Move]Attrs` function on whether to call the `apply` methods or not.

src/data/abilities/ability.ts
```diff
@@ -2116,49 +1550,32 @@ export class FieldMultiplyStatAbAttr extends AbAttr {
...
-  ): void {
-    statValue.value *= this.multiplier;
+  apply({ statVal, hasApplied }: FieldMultiplyStatAbAttrParams): void {
+    statVal.value *= this.multiplier;
     hasApplied.value = true;
   }
 }
 
+export interface MoveTypeChangeAbAttrParams extends AugmentMoveInteractionAbAttrParams {
+  // TODO: Replace the number holder with a holder for the type.
```
DayKev, line 1567: What do you mean?

src/data/abilities/ability.ts
```diff
@@ -2285,59 +1655,51 @@ export class PokemonTypeChangeAbAttr extends PreAttackAbAttr {
...
+  hitCount?: NumberHolder;
+  /** Holder for the damage multiplier _of the current hit_ */
+  multiplier?: NumberHolder;
+}
+
 /**
  * Class for abilities that convert single-strike moves to two-strike moves (i.e. Parental Bond).
- * @param damageMultiplier the damage multiplier for the second strike, relative to the first.
  */
 export class AddSecondStrikeAbAttr extends PreAttackAbAttr {
```
DayKev: Note that by doing this the information is removed from the hover at the point of ability definition:
image

src/data/abilities/ability.ts
```diff
@@ -2348,6 +1710,16 @@ export class AddSecondStrikeAbAttr extends PreAttackAbAttr {
...
 }
 
+/**
+ * Common interface for parameters used by abilities that modify damage/power of a move before an attack
+ */
+export interface PreAttackModifyDamageAbAttrParams extends AugmentMoveInteractionAbAttrParams {
+  /**
+   * The power of the move or amount of damage. May be modified by ability application.
+   */
+  damage: NumberHolder;
```
DayKev: Maybe this should be 2 different interfaces, one for power and one for damage. Which one is being modified does matter for damage calculations (though not as much as in mainline since we don't do the excessive rounding they do).

src/data/abilities/ability.ts
```diff
@@ -2767,22 +2051,18 @@ export class PostAttackStealHeldItemAbAttr extends PostAttackAbAttr {
...
-    pokemon: Pokemon,
-    passive: boolean,
-    simulated: boolean,
-    defender: Pokemon,
-    move: Move,
-    hitResult: HitResult,
-    args: any[],
-  ): boolean {
+  override canApply(params: PostMoveInteractionAbAttrParams): boolean {
+    const { simulated, pokemon, opponent, move, hitResult } = params;
```
DayKev, line 2054: Isn't it a bit inconsistent to use `const {} = params;` here (and a couple other places) instead of doing the destructuring from within the params like you did elsewhere?

src/data/abilities/ability.ts
```diff
@@ -3227,45 +2424,35 @@ export class IgnoreOpponentStatStagesAbAttr extends AbAttr {
...
-  ): void {
-    (args[1] as BooleanHolder).value = true;
+  override apply({ ignored }: IgnoreOpponentStatStagesAbAttrParams): void {
+    ignored.value = true;
   }
 }
 
+/**
+ * Abilities with this attribute prevent the user from being affected by Intimidate.
+ * @sealed
```
DayKev, line 2445: Why is this particular `AbAttr` `@sealed`?

Also same question for the other ones, if the reasons are different.

src/data/abilities/ability.ts
```diff
@@ -4638,129 +3745,96 @@ export class ConditionalUserFieldProtectStatAbAttr extends PreStatStageChangeAbA
...
 }
 
-/**
- * Provides immunity to BattlerTags {@linkcode BattlerTag} to specified targets.
- */
-export class PreApplyBattlerTagImmunityAbAttr extends PreApplyBattlerTagAbAttr {
+// Intentionally not exported because this shouldn't be able to be passed to `applyAbAttrs`. It only exists so that
+// PreApplyBattlerTagImmunityAbAttr and UserFieldPreApplyBattlerTagImmunityAbAttr can avoid code duplication
+// while preserving type safety. (Since the UserField version require an additional parameter, target, in its apply methods)
+abstract class BaseBattlerTagImmunityAbAttr<P extends PreApplyBattlerTagAbAttrParams> extends PreApplyBattlerTagAbAttr {
```
DayKev, line 3779: Shouldn't the others like that also not be exported then?

src/data/abilities/ability.ts
```diff
@@ -4638,129 +3745,96 @@ export class ConditionalUserFieldProtectStatAbAttr extends PreStatStageChangeAbA
...
-    cancelled: BooleanHolder,
-    args: [Pokemon, ...any],
-  ): boolean {
-    return (
-      super.canApplyPreApplyBattlerTag(pokemon, passive, simulated, tag, cancelled, args) && this.condition(args[0])
-    );
+  override canApply(params: UserFieldBattlerTagImmunityAbAttrParams): boolean {
+    // the `!!params` here is to ensure the target is not null or undefined. This is defensive programming
+    // to guard against the case where
+    return super.canApply(params) && this.condition(params.target);
```
DayKev: Obsolete comment?

src/data/abilities/ability.ts
SirzBenjie: It's sealed because it shouldn't be subclassed without consideration; its apply method's parameters doesn't have a marker for `Closed`. It was a bit cumbersome to add this everywhere. Annotating with `sealed` is a lot easier.

src/data/abilities/ability.ts
SirzBenjie: Perhaps, but this PR actually _created_ this class, while it did not create the others. I wanted as minimal code changes as possible, and thought that changing the visibility of an existing class was too much for this PR.

src/data/abilities/ability.ts
SirzBenjie: `MoveTypeChange` abuses a `NumberHolder` to mutate a variable that is really a `PokemonType`. We should make this type safe, such that it only accepts a valid `PokemonType` instead of a NumberHolder.

src/data/abilities/ability.ts
SirzBenjie: This wouldn't work in the general case. Some times, you need to actually evaluate the `apply` method when simulating.
Simulated also isn't used exclusively for AI calculations.
For example, when we are looking at the type of a move for the purposes of the move effectiveness hint, we need to simulate abilities like galvanize. We actually _want_ to do the logic in the apply method so that we know what the type of the move is.

src/data/abilities/ability.ts
SirzBenjie: Can you elaborate on what you mean?

src/data/abilities/ability.ts
SirzBenjie: I don't destructure inside of the params when the method needs access to the whole `params` object itself (because it invokes `super`)

src/data/abilities/ability.ts
SirzBenjie: I actually do have 2 interfaces; there's PreAttackModifyPower as well. This comment just needs to be updated.

src/data/abilities/ability.ts
DayKev: Gotcha. Maybe a "todo" should be added to the others then.

src/data/abilities/ability.ts
DayKev: Ah, okay.

src/data/abilities/ability.ts
DayKev: Oh it's so you can re-use `params` too, okay.

src/data/abilities/ability.ts
DayKev: I was thinking it might be useful to remove both `NumberHolder` and `BooleanHolder` and instead have a generic `ValueHolder<T>`, would that be a good idea? I think it would solve this problem.

src/data/abilities/ability.ts
DayKev: Oh, that's a bit problematic.

src/data/abilities/ability.ts
DayKev: As shown in the screenshot, the inherited docs are pointing to the wrong interface (`TypeImmunityAbAttr#apply` uses the `TypeMultiplierAbAttrParams` interface, not the `AbAttrBaseParams` interface linked in the inherited TSDoc from `AbAttr#apply`).
Any subclass of `AbAttr` will thus have inaccurate TSDocs on the `apply` method.

src/@types/type-helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+// biome-ignore lint/correctness/noUnusedImports: Used in a tsdoc comment
+import type { AbAttr } from "./ability-types";
+
+/**
+ * Exactly matches the type of the argument, preventing adding additional properties.
+ *
+ * ⚠️ Should never be used with `extends`, as this will nullify the exactness of the type.
+ *
+ * As an example, used to ensure that the parameters of {@linkcode AbAttr.canApply} and {@linkcode AbAttr.getTriggerMessage} are compatible with
+ * the type of the apply method
```
Bertie690, line 14:
```suggestion
 * For instance, is used to ensure that the parameters of {@linkcode AbAttr.canApply} and {@linkcode AbAttr.getTriggerMessage} are compatible with
 * the type of the {@linkcode AbAttr.apply | apply} method, no matter the subclass.
```

src/@types/type-helpers.ts
```diff
@@ -0,0 +1,34 @@
+/*
+ * A collection of custom utility types that aid in type checking and ensuring strict type conformity
+ */
+
+// biome-ignore lint/correctness/noUnusedImports: Used in a tsdoc comment
+import type { AbAttr } from "./ability-types";
+
+/**
+ * Exactly matches the type of the argument, preventing adding additional properties.
```
Bertie690, line 9:
```suggestion
 * Exactly matches the type of the argument, preventing adding or removing properties.
```

src/@types/type-helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Type hint that indicates that the type is intended to be closed to a specific shape.
+ * Does not actually do anything special, is really just an alias for X.
+ */
+export type Closed<X> = X;
+
+/**
+ * Remove `readonly` from all properties of the provided type
+ * @typeParam T - The type to make mutable
```
Bertie690, line 30:
```suggestion
 * Remove `readonly` from all properties of the provided type.
 * @typeParam T - The type to make mutable.
```

src/@types/type-helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Exactly matches the type of the argument, preventing adding additional properties.
+ *
+ * ⚠️ Should never be used with `extends`, as this will nullify the exactness of the type.
+ *
+ * As an example, used to ensure that the parameters of {@linkcode AbAttr.canApply} and {@linkcode AbAttr.getTriggerMessage} are compatible with
+ * the type of the apply method
+ *
+ * @typeParam T - The type to match exactly
```
Bertie690, line 16:
```suggestion
 * @typeParam T - The type to match exactly.
```

src/data/abilities/ability.ts
```diff
@@ -81,6 +81,12 @@ import type { ArenaTrapTag, SuppressAbilitiesTag } from "#app/data/arena-tag";
 import type { Constructor } from "#app/utils/common";
 import type { Localizable } from "#app/@types/locales";
 import { applyAbAttrs } from "./apply-ab-attrs";
+import type { Closed, Exact } from "#app/@types/type-helpers";
+
+// biome-ignore lint/correctness/noUnusedImports: Used in TSDoc
+import type BattleScene from "#app/battle-scene";
+// biome-ignore lint/correctness/noUnusedImports: Used in TSDoc
```
Bertie690: Replace with range ignore

src/data/abilities/ability.ts
```diff
@@ -220,6 +227,35 @@ export class Ability implements Localizable {
...
 
+/** Base set of parameters passed to every ability attribute's apply method */
+export interface AbAttrBaseParams {
+  /** The pokemon that has the ability being applied */
+  readonly pokemon: Pokemon;
+
+  /**
+   * Whether the ability's effects are being simulated.
+   * Used to prevent, for instance, messages flyouts from being displayed.
+   * Defaults to false.
```
```suggestion
   * @defaultValue `false`.
```

src/data/abilities/ability.ts
```diff
@@ -220,6 +227,35 @@ export class Ability implements Localizable {
...
 }
 
+/** Base set of parameters passed to every ability attribute's apply method */
+export interface AbAttrBaseParams {
+  /** The pokemon that has the ability being applied */
+  readonly pokemon: Pokemon;
+
+  /**
+   * Whether the ability's effects are being simulated.
+   * Used to prevent, for instance, messages flyouts from being displayed.
```
```suggestion
   * Used to prevent message flyouts and other effects from being triggered.
```

src/data/abilities/ability.ts
```diff
@@ -220,6 +227,35 @@ export class Ability implements Localizable {
   }
 }
 
+/** Base set of parameters passed to every ability attribute's apply method */
+export interface AbAttrBaseParams {
+  /** The pokemon that has the ability being applied */
+  readonly pokemon: Pokemon;
+
+  /**
+   * Whether the ability's effects are being simulated.
```
```suggestion
   * Whether the ability's effects are being simulated (for instance, during AI damage calculations).
```

src/data/abilities/ability.ts
```diff
@@ -250,25 +286,21 @@ export abstract class AbAttr {
...
-    _simulated: boolean,
-    _cancelled: BooleanHolder | null,
-    _args: any[],
-  ): void {}
-
-  getTriggerMessage(_pokemon: Pokemon, _abilityName: string, ..._args: any[]): string | null {
+  apply(_params: AbAttrBaseParams): void {}
+
+  // The `Exact` in the next two signatures enforces that the type of the _params operand
+  // is always compatible with the type of apply. This allows fewer fields, but never a type with more.
```
Bertie690, line 296: > This allows fewer fields, but never a type with more.

If it has fewer fields than required, isn't that literally called "incompatible types"?

src/data/abilities/ability.ts
```diff
@@ -277,71 +309,47 @@ export abstract class AbAttr {
...
-    _args: any[],
-  ): void {
+  override apply({ cancelled }: AbAttrParamsWithCancel): void {
     cancelled.value = true;
   }
 }
 
+export interface DoubleBattleChanceAbAttrParams extends AbAttrBaseParams {
+  /** Holder for the chance of a double battle that may be modified by the ability */
+  chance: NumberHolder;
```
Bertie690, line 328: I suggest changing this to `threshold` or similar as that more clearly describes its function.

src/data/abilities/ability.ts
```diff
@@ -462,100 +457,87 @@ export class ClearTerrainAbAttr extends AbAttr {
...
-  ): boolean {
-    return true;
-  }
+/**
+ * Shared interface for AbAttrs that interact with a move that is being used by or against the user.
+ *
+ * Often extended by other interfaces to add more parameters.
+ * Used, e.g. by {@linkcode PreDefendAbAttr} and {@linkcode PostAttackAbAttr}
+ */
+// TODO: Consider making this not require `cancelled`, as many abilities do not do anything with the parameter.
```
Bertie690: it doesn't take cancelled though...????
this extends base params lmao

src/data/abilities/ability.ts
```diff
@@ -2364,38 +1733,37 @@ export class DamageBoostAbAttr extends PreAttackAbAttr {
...
   }
 }
 
+export interface PreAttackModifyPowerAbAttrParams extends AugmentMoveInteractionAbAttrParams {
+  /** Holds the base power of the move, which may be modified after ability application */
+  power: NumberHolder;
+}
+
+/*
+This base class *is* allowed to be invoked directly by `abAttrApply`.
```
```suggestion
This base class *is* allowed to be invoked directly by `applyAbAttrs`.
```

src/data/abilities/ability.ts
```diff
@@ -3071,32 +2288,25 @@ export class SynchronizeStatusAbAttr extends PostSetStatusAbAttr {
...
     if (!simulated && sourcePokemon) {
       sourcePokemon.trySetStatus(effect, true, pokemon);
     }
   }
 }
 
+/**
+ * Base class for abilities that apply an effect after the user knocks out an opponent in battle.
+ *
+ * Not to be confused with {@link PostKnockOutAbAttr}, which applies after any pokemon is knocked out in battle.
```
```suggestion
 * Not to be confused with {@linkcode PostKnockOutAbAttr}, which applies after _any_ pokemon is knocked out in battle (friend or foe).
```

Bertie690 (changes requested):
Extremely partial initial review (I think I got like to around Download before the lag got to me)

src/data/abilities/ability.ts
SirzBenjie: Thanks for pointing that out. It was leftover.

src/data/abilities/ability.ts
SirzBenjie: No. And this behavior is actually what we want. It allows us, for instance, to override the `apply` method's parameter signature, but not do it to `canApply`. This is OK, as we can just leverage the superclass definition of `canApply`, which will work since at worst it has fewer fields than we provide. This makes subclassing and inherited methods work.

src/@types/type-helpers.ts
SirzBenjie: This isn't a sentence; it doesn't need a period.

src/data/abilities/ability.ts
SirzBenjie: Yeah good point, I can do that now.

src/@types/type-helpers.ts
SirzBenjie: No, it allows for properties to be removed.

src/data/abilities/ability.ts
SirzBenjie: Yep, I wanted to do exactly that.

src/data/abilities/ability.ts
SirzBenjie: Hrmm. This is very not ideal. Better to remove that part of the TSDoc I suppose?

DayKev: I think so.

src/data/abilities/ability.ts
SirzBenjie: Actually, none of these classes ever need to be exported.... Not unless they need to for some use in tests.

Snailman11:
Link to Discord Bug Report (Status condition called "none.description" when blocked by ability)
https://discord.com/channels/1125469663833370665/1369197158506561638

PR #6004, opened 2025-06-18
[Dev] Update Biome to 2.0.0

DayKev:
N/A

Biome 2.0 has a bunch of new features we can take advantage of.

### Rule Changes / New Rules
```jsonc
// newly added recommended rule, can't be used, manually overridden/disabled
"noDocumentCookie": "off" // Firefox has minimal support for the "Cookie Store API"

// previously "warn"
"noExcessiveCognitiveComplexity": "info",

// new manually added rules
"useAdjacentGetterSetter": "error",
"noConstantBinaryExpression": "error",
"noTsIgnore": "error",
"noAwaitInLoop": "warn",
"useJsonImportAttribute": "off", // "Import attributes are only supported when the '--module' option is set to 'esnext', 'node18', 'nodenext', or 'preserve'. ts(2823)"
"useIndexOf": "error",
"useObjectSpread": "error",
"useNumericSeparators": "off", // TODO: enable?
"useIterableCallbackReturn": "warn", // TODO: refactor and make "error"
"noShadow": "warn" // TODO: refactor and make "error"

// only for test files
"noFloatingPromises": "error"
```

## Checklist


biome.jsonc
```diff
@@ -77,53 +113,70 @@
...
-                "use": "number"
-              }
-            }
-          }
-        }
+        "useAdjacentGetterSetter": "error",
+        "noConstantBinaryExpression": "error",
+        "noTsIgnore": "error",
+        "noAwaitInLoop": "warn",
+        "useJsonImportAttribute": "off", // "Import attributes are only supported when the '--module' option is set to 'esnext', 'node18', 'nodenext', or 'preserve'. ts(2823)"
```
SirzBenjie, line 140: Do we need to explicitly specify "off" for this, then? Is Biome not smart enough to ignore this rule when the module option isn't set to esnext / node18?

Madmadness65 (approved):

biome.jsonc
DayKev: It's not necessary to specify it but I wanted to put it in so it's clear why we can't use it (since if it worked we would use it).

DayKev:
> Are we able to get rid of ESlint now?

Biome's rule isn't 100% accurate yet, nor do they have an equivalent to https://typescript-eslint.io/rules/no-misused-promises/ yet either. Though I think it's probably good enough we could stop using it?

SirzBenjie:
> > Are we able to get rid of ESlint now?
> 
> Biome's rule isn't 100% accurate yet, nor do they have an equivalent to https://typescript-eslint.io/rules/no-misused-promises/ yet either. Though I think it's probably good enough we could stop using it?

I would like to stop using it.

DayKev:
> > > Are we able to get rid of ESlint now?
> > 
> > 
> > Biome's rule isn't 100% accurate yet, nor do they have an equivalent to https://typescript-eslint.io/rules/no-misused-promises/ yet either. Though I think it's probably good enough we could stop using it?
> 
> I would like to stop using it.

Removed all the ESLint packages + config and workflow.

SirzBenjie (approved):

Issue #5863, opened 2025-05-22
[Bug] Transform/Imposter Succeeds Against Transformed Pokemon


Snailman11:
### Describe the bug

Using Transform or activating the Imposter Ability against a Transformed Pokemon should result in the attempt failing. 

### Reproduction

Start with Ditto
"""
	OPP_SPECIES_OVERRIDE: Species.DITTO,

	OPP_ABILITY_OVERRIDE: Abilities.IMPOSTER,
"""

### Expected behavior

>  It also fails against Shadow Pokémon, if the user or the target are transformed
-Bulba

Transform fails against Transformed Pokemon
https://replay.pokemonshowdown.com/gen9customgame-2369601136

Imposter does not activate against Transformed Pokemon
https://replay.pokemonshowdown.com/gen9customgame-2369623281

### Screenshots / Videos

Transformed Transform

https://github.com/user-attachments/assets/8eb228cb-92aa-497b-bdd4-692439889881

Imposter

https://github.com/user-attachments/assets/155a7c66-fb3f-4c6e-abfa-6bd8e8f11d8c

PR #6008, opened 2025-06-19
[BUG] Fix Transform and Imposter issues

jnotsknab:

- The Imposter ability will no longer activate against a transformed pokemon, and the Transform move will fail if the user or target is transformed.
- Players will no longer see duplicate transformation messages.

- Fixes the bugs referenced in issue #5863 
- Transform would succeed when used by or against a Pokemon that had already transformed
- Imposter could trigger against an already transformed Pokemon, causing unintended behavior.
- `phase-manager.queueMessage` queues duplicate messages upon transformation, which is unintended.
- Transform and Imposter will now follow expected behavior from mainline games.

- Implemented the `TransformTag` class which extends `BattleTag`, and indicates if a Pokemon is transformed or not based on if the tag exists for a given Pokemon
- This tag is added to the user when **Transform** succeeds and assists in preventing the invalid scenarios mentioned above.
- In the `canApplyPostSummon` override method of the `PostSummonTransformAbAttr` class we now check if either user or target has a `TransformTag`
- If so the user or target has already transformed, so we return `false`, preventing the ability from triggering.
- Removed the `phase-manager.queueMessage` invocation for the targets transformation event as this is handled within the **PokemonTransformPhase**
- In the **MovePhase** we now check if the current `MoveId` matches `MoveId.TRANSFORM` and if either target or user has a `TransformTag`
- If so Transform fails and we display the appropriate messages and end the turn.
## Screenshots/Videos
Transform Before:

https://github.com/user-attachments/assets/6faa5fb9-c332-414e-982f-0beccac3e023

Imposter Before:

https://github.com/user-attachments/assets/755d5a9e-0784-41f7-8e09-e06e3fd44388

Transform After: 

[TransformAfter.webm](https://github.com/user-attachments/assets/777dfc00-24cd-4b24-9210-d1bfad6a69ca)

Imposter After:

[ImposterAfter.webm](https://github.com/user-attachments/assets/3a4e240e-b395-46de-9d94-fbb98865e8cd)

## How to test the changes?

Use the following overrides to test the changes, comment ability overrides out if need be to test fixes seperately

<pre><code>const overrides = {
    OPP_SPECIES_OVERRIDE: SpeciesId.DITTO,
    OPP_MOVESET_OVERRIDE: MoveId.TRANSFORM,
    OPP_ABILITY_OVERRIDE: AbilityId.IMPOSTER,
    ABILITY_OVERRIDE: AbilityId.IMPOSTER,
    STARTER_SPECIES_OVERRIDE: SpeciesId.MEW,
    MOVESET_OVERRIDE: MoveId.TRANSFORM

} satisfies Partial<InstanceType<OverridesType>>;</code></pre>

## Checklist


src/data/abilities/ability.ts
```diff
@@ -3909,8 +3909,14 @@ export class PostSummonTransformAbAttr extends PostSummonAbAttr {
   override canApplyPostSummon(pokemon: Pokemon, _passive: boolean, simulated: boolean, _args: any[]): boolean {
     const targets = pokemon.getOpponents();
     const target = this.getTarget(targets);
+    const user = pokemon;
 
-    if (target.summonData.illusion) {
+    //Prevents Imposter from triggering on a transformed target or if the user is already transformed
+    if (user?.getTag(BattlerTagType.TRANSFORM) || target?.getTag(BattlerTagType.TRANSFORM)) {
```
SirzBenjie, line 3915: There's no reason to define `user` here. Just use `pokemon` instead of `user`.

SirzBenjie: There's no reason to define `user = pokemon` here. Just use `pokemon` instead of `user` in `user.getTag`

src/data/battler-tags.ts
```diff
@@ -3221,6 +3221,25 @@ export class ImprisonTag extends MoveRestrictionBattlerTag {
...
+  }
+  /**
+   * Adds the Transform battler tag to the Pokemon transforming.
+   * @param pokemon - The {@linkcode Pokemon} transforming
+   */
+  override onAdd(pokemon: Pokemon): void {
+    super.onAdd(pokemon);
+  }
+}
+
```
SirzBenjie: I'm not sure if a _tag_ is necessary here. And indeed, I have a disdain for our tag system, and would like to avoid using it when possible.
We can already check if a pokemon is transformed by checking whether `pokemon.summonData.speciesForm` is defined or not. Sure, this isn't the most _intuitive_ way to do it. But we could just as easily expose a method on the `pokemon` class called `isTransformed` that just checks this.
It's better than having a tag.

src/data/abilities/ability.ts
jnotsknab: The check now uses pokemon directly and the exposed `isTransformed` method

src/data/battler-tags.ts
jnotsknab: `TransformTag` has now been removed completely in favor of the suggested implementation of an exposed 

`isTransformed` method within the `pokemon` class.

src/phases/move-phase.ts
```diff
@@ -339,6 +339,17 @@ export class MovePhase extends BattlePhase {
...
+      //Prevent using Transform if either the user or target is already transformed
+      this.move.getMove().id === MoveId.TRANSFORM &&
+      (targets[0].isTransformed() || this.pokemon.isTransformed())
+    ) {
+      this.showMoveText();
+      this.showFailedText();
+      this.end();
+      return;
+    }
+
```
DayKev: This check should be as a condition for Transform itself (on `TransformAttr` or something) and not in `MovePhase`.

jnotsknab: Check has been altered as a condition of  the `TransformAttr` class as requested and no longer references `MoveId.TRANSFORM` which is implicitly implied.

src/data/moves/move.ts
```diff
@@ -7610,14 +7610,13 @@ export class SuppressAbilitiesIfActedAttr extends MoveEffectAttr {
  */
 export class TransformAttr extends MoveEffectAttr {
   override apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    if (!super.apply(user, target, move, args)) {
+    if (!super.apply(user, target, move, args) || (target.isTransformed() || user.isTransformed())) {
```
DayKev: I think what you want is to do is something like this:
```ts
override canApply(...) {
  return super.canApply(...) && !(target.isTransformed() || user.isTransformed());
}
```
Then you shouldn't need to manually queue a failure message.

jnotsknab: `canApply` from my understanding does not implicity queue the fail message, it simply prevents the move from applying, so this would work in failing the move itself, but would not queue the message and provide feedback to the player, should `canApply` be responsible for queuing this message, or is it expected elsewhere?

DayKev: Oh, hm. I'll take a closer look.

Bertie690: @jnotsknab is correct.

`canApply` is solely a helper function for checking if the individual _effects_ of the attribute can apply (i.e. the user & target aren't fainted).
Any conditions that make the move actually _fail_ **must** go inside `getCondition` as otherwise the move is _still considered_ as having "succeeded" for stuff like Stomping Tantrum and co.

(Frankly, I kinda hate the way this system is setup wrt moves with multiple effects, but I digress...)

src/field/pokemon.ts
```diff
@@ -1062,6 +1062,10 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
     return species;
   }
 
+  isTransformed(): boolean {
```
Bertie690, line 1064:
```suggestion
  /** 
   * Getter function that returns whether this {@linkcode Pokemon} is currently transformed into another one 
   * (such as by the effects of {@linkcode MoveId.TRANSFORM} or {@linkcode AbilityId.IMPOSTER}. 
   * @returns Whether this Pokemon is currently transformed.
   */
  isTransformed(): boolean {
```

src/field/pokemon.ts
```diff
@@ -1062,6 +1062,34 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+        // Neither pokemon can be behind an illusion
+        target.summonData.illusion ||
+        this.summonData.illusion ||
+        // The target cannot be behind a substitute
+        target.getTag(BattlerTagType.SUBSTITUTE) ||
+        // Transforming to/from fusion pokemon causes various problems (crashes, etc.)
+        // TODO: Consider lifting restriction once bug is fixed
+        this.isFusion() ||
+        target.isFusion()
+      )
```
DayKev, line 1088:
```suggestion
      this.isTransformed() ||
      target.isTransformed() ||
      // Neither pokemon can be behind an illusion
      target.summonData.illusion ||
      this.summonData.illusion ||
      // The target cannot be behind a substitute
      target.getTag(BattlerTagType.SUBSTITUTE) ||
      // Transforming to/from fusion pokemon causes various problems (crashes, etc.)
      // TODO: Consider lifting restriction once bug is fixed
      this.isFusion() ||
      target.isFusion()
```

Bertie690 (changes requested):
comment fix

DayKev (approved):

DayKev (dismissed):

xsn34kzx (dismissed):

src/field/pokemon.ts
```diff
@@ -1056,6 +1056,39 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * (such as by the effects of {@linkcode MoveId.TRANSFORM} or {@linkcode AbilityId.IMPOSTER}.
+   * @returns Whether this Pokemon is currently transformed.
+   */
+  isTransformed(): boolean {
+    return this.summonData.speciesForm !== null;
+  }
+
+  /**
+   * Return whether this Pokemon can transform into an opposing Pokemon.
+   * @param target - The {@linkcode Pokemon} being transformed into.
```
Bertie690:
```suggestion
   * @param target - The {@linkcode Pokemon} being transformed into
```

Benjie apparently says period no go in param lists soooooo...

src/field/pokemon.ts
```diff
@@ -1318,31 +1351,25 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
   }
 
   /**
-   * Writes the value to the in-battle stage of the corresponding {@linkcode BattleStat} of the {@linkcode Pokemon}.
-   *
-   * Note that, if the value is not within a range of [-6, 6], it will be forced to the closest range bound.
-   * @param stat the {@linkcode BattleStat} whose stage is to be overwritten
-   * @param value the desired numeric value
+   * Sets this {@linkcode Pokemon}'s in-battle stat stage to the corresponding value.
+   * @param stat - The {@linkcode BattleStat} whose stage is to be overwritten.
```
Bertie690:
```suggestion
   * @param stat - The {@linkcode BattleStat} whose stage is to be overwritten
```

src/field/pokemon.ts
```diff
@@ -1318,31 +1351,25 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
-    return this.summonData ? this.summonData.statStages : [0, 0, 0, 0, 0, 0, 0];
+    return this.summonData.statStages;
   }
 
   /**
-   * Retrieves the in-battle stage of the specified {@linkcode BattleStat}.
-   * @param stat the {@linkcode BattleStat} whose stage is desired
-   * @returns the stage of the desired {@linkcode BattleStat} if available, 0 otherwise
+   * Retrieve the value of the given stat stage for this {@linkcode Pokemon}.
+   * @param stat - The {@linkcode BattleStat} to retrieve the stat stage for.
```
Bertie690:
```suggestion
   * @param stat - The {@linkcode BattleStat} to retrieve the stat stage for
```

src/field/pokemon.ts
```diff
@@ -1270,39 +1303,39 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
 
   /**
-   * Writes the value to the corrseponding {@linkcode PermanentStat} of the {@linkcode Pokemon}.
-   *
-   * Note that this does nothing if {@linkcode value} is less than 0.
-   * @param stat the desired {@linkcode PermanentStat} to be overwritten
-   * @param value the desired numeric value
-   * @param bypassSummonData write to actual stats (`true` by default) or in-battle overridden stats (`false`)
+   * Change one of this {@linkcode Pokemon}'s {@linkcode PermanentStat}s to the specified value.
+   * @param stat - The {@linkcode PermanentStat} to be overwritten.
```
Bertie690:
```suggestion
   * @param stat - The {@linkcode PermanentStat} to be overwritten
```

src/field/pokemon.ts
```diff
@@ -1270,39 +1303,39 @@ export default abstract class Pokemon extends Phaser.GameObjects.Container {
...
     }
     return this.stats;
   }
 
   /**
    * Retrieves the corresponding {@linkcode PermanentStat} of the {@linkcode Pokemon}.
-   * @param stat the desired {@linkcode PermanentStat}
-   * @param bypassSummonData prefer actual stats (`true` by default) or in-battle overridden stats (`false`)
-   * @returns the numeric value of the desired {@linkcode Stat}
+   * @param stat - The desired {@linkcode PermanentStat}.
```
Bertie690:
```suggestion
   * @param stat - The {@linkcode PermanentStat} to retrieve
```

src/phases/pokemon-transform-phase.ts
```diff
@@ -21,12 +21,13 @@ export class PokemonTransformPhase extends PokemonPhase {
     super(userIndex);
 
     this.targetIndex = targetIndex;
+
```
Bertie690:
```suggestion
```

SirzBenjie (approved):

PR #6012, opened 2025-06-19
[Test] Modernize Phase Interceptor, move UI handling to `PromptHandler`

Bertie690:
> [!IMPORTANT]
> This PR was initially created basing off of the [equivalent file from `poketernity`](<https://github.com/Despair-Games/poketernity/blob/beta/test/test-utils/phase-interceptor.ts>) for type safety (with a few deprecated functions removed here and there).
> I then proceeded to rewrite the entire file from scratch and then some.

Dependent on #6109 and #6108

N/A

Put simply, the phase interceptor was a good-ish idea executed in about the worst way possible. 

It used intervals everywhere with attached callbacks, was virtually incomprehensible even barring the lack of type signatures, and was a pain to debug.

- Cleaned up the code for the phase interceptor. It still does the same stuff, but is much more readable and sensible.
- Moved prompt handling code to a new class (`PromptHandler`)

### The Old system

After spending _far_ too much time looking at barely documented, practically unreadable :spaghetti:, I understand the current system for phase intercepting to go vaguely like this: 
1. Go through the giant `PHASES` array and set all phases' `start` methods to add their _constructors_ to an `onHold` array. (Why a queue was chosen when only 1 phase can be started at a time is beyond me.)
2. Once `to` is called, use a `Interval` to periodically check the currently on-hold phase.
   - If it matches the target phase, then call `run` as applicable and resolve.
   - If not, call `run` on the current phase.
   - If the queue is empty, skip and check again the next time the callback runs. This is presumably required due to the phase manager not being started until **AFTER** move animations are loaded in `BattleScene.create` (via a floating promise I may add). (Also see 5 for more)

3. When `run` is called, start a 2nd interval that:
   - Unshifts `onHold` every time it runs. (Again, only 1 phase can be started at a time, but I digress...)
   - Set `inProgress` to an object with the current phase's name and a callback to resolve the promise.
   - Manually call the corresponding phase's `start` method bound to the value of `scene.phaseManager.getCurrentPhase`.
4. Once the phase finishes or once `unlock` is called, trigger the callback and return control to the `to` interval. This will run the next phase in line (if the target was not reached). 
5. On changing UI modes, if the phase is within `endBySetMode`, trigger the callback and return control to the caller the next time the main event loop finishes (to prevent UI stalling). This also causes the `onHold` array to be emptied (so subsequent checks from the `to` interval will simply sit and wait until the phase is ended manually).

From what I can tell, the entire `errorInterceptor` system was designed _exclusively_ for clearing said intervals if a running phase crashed, as was the `onError` callback on `inProgress` and the overriding of `Phase.prototype.end`.

### Issues

There are several issues this presents:
1. Only works on certain phases or else requires duplicating an entire master list
2. Completely ignores the current `Phase.is` system, instead functioning based off of constructor names
3. Treats certain phases differently with respect to UI mode changes
4. Overriding the start methods of phases is much less effective than just changing the phase manager

> [!IMPORTANT]
> Unfortunately, I cannot remove the intervals entirely. Our current UI system, due to effectively hanging until an input is obtained, effectively means all handing of phases has to occur within intervals that run during the poll phase. I'd love to fix this, but we'd basically have to re-write all our UI logic.
> The best I can do is change them into `vi.waitUntil` calls to look better, but those still rely on intervals...

### The New system

With that said and done, here is a synopsis of how the new system works:
1. The prior prompt handling has been sequestered into a new class, `PromptHandler`.
2. All the prior method overwriting is done using overrides on private methods. Only 2 of these are required: 1 inside `PromptHandler` to wrap `UI.setModeInternal` and the other to delay starting the next phase when `PhaseManager.shiftPhase` is called. 
   - To facilitate this, the action of calling `newPhase.start` within the `PhaseManager` has been separated into a new method (`startCurrentPhase`).
3. All the phase interceptor intervals that actually block operations have been replaced with `vi.waitUntil` due to it having more error handling and better safety. 
   The one instance of a non-blocking interval (in `PromptHandler`) is stored in a static instance and cleared on test end. (For extra security, the class' constructor throws an error if it somehow wasn't cleared prior.)
   - This (along with point 1) effectively alleviates the need to call `restoreOg` after each test - all mocks and intervals will be reset before the next test runs, and any errors will naturally propagate up the chain and reject the promise. (The method itself has been converted to a no-op and removed from existing test boilerplates, with plans to remove it entirely in a future PR.)
4. **All phases work with the interceptor**. Rather than needing a giant list of eligible phases, the new system is effectively agnostic to what phases are being run, merely serving as a wrapper around the existing `is` and `start` methods.
5. The interceptor uses a private `state` variable to track what it is doing as a part of making its inner workings less opaque.

> [!NOTE]
> I would like to stress that this is **99% backwards compatible with the old phase interceptor.** 
> It may look a ~~little~~ lot different, but the core logic is still the same as before (override phase starting and check each phase in turn), just executed a lot cleaner and with a lot fewer questionable callbacks.
>
> Most changes to test files are effectively what _would_ have happened if phases were added to the big `PhaseInterceptor.PHASES` array. 
Furthermore, I sincerely hope this variant will be easier to make work with a future breakup of `BattleScene`. I have added TODO comments to relevant aspects of `game-wrapper.ts` questioning the veracity and necessity of various stubs that would make multi-scene interactions hard.

## How to test the changes?
Run tests, pretty much.
I added a new folder `test/test-utils/tests` containing both unit and integration tests for the new classes. 

## Important files to review
- `phase-interceptor.ts`
- `prompt-handler.ts`
- `phase-manager.ts`
- new test files
- `game-manager.ts`
- `vitest.setup.ts`
- `battle-scene.ts`

## Checklist

### DIFF
--- a/src/phases/title-phase.ts
+++ b/src/phases/title-phase.ts
@@ -18,20 +18,21 @@ import { getBiomeKey } from "#field/arena";
 import type { Modifier } from "#modifiers/modifier";
 import { getDailyRunStarterModifiers, regenerateModifierPoolThresholds } from "#modifiers/modifier-type";
 import { vouchers } from "#system/voucher";
-import type { SessionSaveData } from "#types/save-data";
 import type { OptionSelectConfig, OptionSelectItem } from "#ui/abstract-option-select-ui-handler";
 import { SaveSlotUiMode } from "#ui/save-slot-select-ui-handler";
 import { isLocalServerConnected } from "#utils/common";
 import { getPokemonSpecies } from "#utils/pokemon-utils";
 import i18next from "i18next";
 
+const NO_SAVE_SLOT = -1;
+
 export class TitlePhase extends Phase {
   public readonly phaseName = "TitlePhase";
   private loaded = false;
-  private lastSessionData: SessionSaveData;
+  // TODO: Make `end` take a `GameModes` as a parameter rather than storing it on the class itself
   public gameMode: GameModes;
 
-  start(): void {
+  async start(): Promise<void> {
     super.start();
 
     globalScene.ui.clearText();
@@ -44,30 +45,46 @@ export class TitlePhase extends Phase {
       globalScene.playBgm("title", true);
     }
 
-    globalScene.gameData
-      .getSession(loggedInUser?.lastSessionSlot ?? -1)
-      .then(sessionData => {
-        if (sessionData) {
-          this.lastSessionData = sessionData;
-          const biomeKey = getBiomeKey(sessionData.arena.biome);
-          const bgTexture = `${biomeKey}_bg`;
-          globalScene.arenaBg.setTexture(bgTexture);
-        }
-        this.showOptions();
-      })
-      .catch(err => {
-        console.error(err);
-        this.showOptions();
-      });
+    const lastSlot = await this.checkLastSaveSlot();
+    await this.showOptions(lastSlot);
   }
 
-  showOptions(): void {
+  /**
+   * If a user is logged in, check the last save slot they loaded and adjust various variables
+   * to account for it.
+   * @returns A Promise that resolves with the last loaded session's slot ID.
+   * Returns `NO_SAVE_SLOT` if not logged in or no session was found.
+   */
+  private async checkLastSaveSlot(): Promise<number> {
+    if (loggedInUser == null) {
+      return NO_SAVE_SLOT;
+    }
+    try {
+      const sessionData = await globalScene.gameData.getSession(loggedInUser.lastSessionSlot);
+      if (!sessionData) {
+        return NO_SAVE_SLOT;
+      }
+
+      globalScene.sessionSlotId = loggedInUser.lastSessionSlot;
+      // Set the BG texture to the last save's current biome
+      const biomeKey = getBiomeKey(sessionData.arena.biome);
+      const bgTexture = `${biomeKey}_bg`;
+      globalScene.arenaBg.setTexture(bgTexture);
+      return loggedInUser.lastSessionSlot;
+    } catch (err) {
+      console.error(err);
+      return NO_SAVE_SLOT;
+    }
+  }
+
+  private async showOptions(lastSessionSlot: number): Promise<void> {
     const options: OptionSelectItem[] = [];
-    if (loggedInUser && loggedInUser.lastSessionSlot > -1) {
+    // Add a "continue" menu if the session slot ID is >-1
+    if (lastSessionSlot > NO_SAVE_SLOT) {
       options.push({
         label: i18next.t("continue", { ns: "menu" }),
         handler: () => {
-          this.loadSaveSlot(this.lastSessionData || !loggedInUser ? -1 : loggedInUser.lastSessionSlot);
+          this.loadSaveSlot(lastSessionSlot);
           return true;
         },
       });
@@ -144,8 +161,9 @@ export class TitlePhase extends Phase {
         label: i18next.t("menu:loadGame"),
         handler: () => {
           globalScene.ui.setOverlayMode(UiMode.SAVE_SLOT, SaveSlotUiMode.LOAD, (slotId: number) => {
-            if (slotId === -1) {
-              return this.showOptions();
+            if (slotId === NO_SAVE_SLOT) {
+              console.warn("Attempted to load save slot of -1 through load game menu!");
+              return this.showOptions(slotId);
             }
             this.loadSaveSlot(slotId);
           });
@@ -174,21 +192,18 @@ export class TitlePhase extends Phase {
       noCancel: true,
       yOffset: 47,
     };
-    globalScene.ui.setMode(UiMode.TITLE, config);
+    await globalScene.ui.setMode(UiMode.TITLE, config);
   }
 
   // TODO: Make callers actually wait for the save slot to load
   private async loadSaveSlot(slotId: number): Promise<void> {
-    globalScene.sessionSlotId = slotId > -1 || !loggedInUser ? slotId : loggedInUser.lastSessionSlot;
+    // TODO: Do we need to `await` this?
     globalScene.ui.setMode(UiMode.MESSAGE);
     globalScene.ui.resetModeChain();
     try {
-      const success = await globalScene.gameData.loadSession(slotId, slotId === -1 ? this.lastSessionData : undefined);
+      const success = await globalScene.gameData.loadSession(slotId);
       if (success) {
         this.loaded = true;
-        if (loggedInUser) {
-          loggedInUser.lastSessionSlot = slotId;
-        }
         globalScene.ui.showText(i18next.t("menu:sessionSuccess"), null, () => this.end());
       } else {
         this.end();
@@ -319,6 +334,7 @@ export class TitlePhase extends Phase {
     });
   }
 
+  // TODO: Refactor this
   end(): void {
     if (!this.loaded && !globalScene.gameMode.isDaily) {
       globalScene.arena.preloadBgm();
@@ -357,6 +373,7 @@ export class TitlePhase extends Phase {
       }
     }
 
+    // TODO: Move this to a migrate script instead of running it on save slot load
     for (const achv of Object.keys(globalScene.gameData.achvUnlocks)) {
       if (vouchers.hasOwnProperty(achv) && achv !== "CLASSIC_VICTORY") {
         globalScene.validateVoucher(vouchers[achv]);
--- a/test/mystery-encounter/encounter-test-utils.ts
+++ b/test/mystery-encounter/encounter-test-utils.ts
@@ -150,12 +150,9 @@ async function handleSecondaryOptionSelect(game: GameManager, pokemonNo: number,
 /**
  * For any {@linkcode MysteryEncounter} that has a battle, can call this to skip battle and proceed to {@linkcode MysteryEncounterRewardsPhase}
  * @param game
- * @param runRewardsPhase
+ * @param runRewardsPhase - (Default `true`) Whether to start the `MysteryEncounterRewardsPhase`
  */
-export async function skipBattleRunMysteryEncounterRewardsPhase(
-  game: GameManager,
-  runRewardsPhase?: false | undefined,
-) {
+export async function skipBattleRunMysteryEncounterRewardsPhase(game: GameManager, runRewardsPhase?: false) {
   game.scene.phaseManager.clearPhaseQueue();
   game.scene.getEnemyParty().forEach(p => {
     p.hp = 0;
--- a/test/test-utils/helpers/classic-mode-helper.ts
+++ b/test/test-utils/helpers/classic-mode-helper.ts
@@ -12,24 +12,25 @@ import { SelectStarterPhase } from "#phases/select-starter-phase";
 import { TurnInitPhase } from "#phases/turn-init-phase";
 import { generateStarters } from "#test/test-utils/game-manager-utils";
 import { GameManagerHelper } from "#test/test-utils/helpers/game-manager-helper";
+import type { IntClosedRange } from "type-fest";
 
 /**
  * Helper to handle classic-mode specific operations.
  */
 export class ClassicModeHelper extends GameManagerHelper {
   /**
-   * Runs the classic game to the summon phase.
-   * @param species - An array of {@linkcode SpeciesId} to summon.
-   * @returns A promise that resolves when the summon phase is reached.
+   * Run the classic game to the summon phase.
+   * @param species - An array of {@linkcode SpeciesId} to summon
+   * @returns A Promise that resolves when the summon phase is reached.
    * @remarks
    * Do not use this when {@linkcode startBattle} can be used!
    */
   async runToSummon(species: SpeciesId[]): Promise<void>;
   /**
-   * Runs the classic game to the summon phase.
+   * Run the classic game to the summon phase.
    * Selects 3 daily run starters with a fixed seed of "test"
-   * (see `DailyRunConfig.getDailyRunStarters` in `daily-run.ts` for more info).
-   * @returns A promise that resolves when the summon phase is reached.
+   * (see `DailyRunConfig.getDailyRunStarters` in `daily-run.ts` for more info)
+   * @returns A Promise that resolves when the summon phase is reached.
    * @deprecated - Specifying the starters helps prevent inconsistencies from internal RNG changes.
    */
   // biome-ignore lint/style/useUnifiedTypeSignatures: Marks the overload for deprecation
@@ -64,15 +65,15 @@ export class ClassicModeHelper extends GameManagerHelper {
 
   /**
    * Transitions to the start of a battle.
-   * @param species - An array of {@linkcode SpeciesId} to start the battle with.
-   * @returns A promise that resolves when the battle is started.
+   * @param species - An array of {@linkcode SpeciesId}s with which to start the battle
+   * @returns A Promise that resolves when the battle is started.
    */
   async startBattle(species: SpeciesId[]): Promise<void>;
   /**
    * Transitions to the start of a battle.
    * Will select 3 daily run starters with a fixed seed of "test"
-   * (see `DailyRunConfig.getDailyRunStarters` in `daily-run.ts` for more info).
-   * @returns A promise that resolves when the battle is started.
+   * (see `DailyRunConfig.getDailyRunStarters` in `daily-run.ts` for more info)
+   * @returns A Promise that resolves when the battle is started.
    * @deprecated - Specifying the starters helps prevent inconsistencies from internal RNG changes.
    */
   // biome-ignore lint/style/useUnifiedTypeSignatures: Marks the overload for deprecation
@@ -109,7 +110,8 @@ export class ClassicModeHelper extends GameManagerHelper {
   /**
    * Queue inputs to switch at the start of the next battle, and then start it.
    * @param pokemonIndex - The 0-indexed position of the party pokemon to switch to.
-   * Should never be called with 0 as that will select the currently active pokemon and freeze
+   * @throws {@linkcode Error}
+   * Fails test if `pokemonIndex` is out of valid bounds
    * @returns A Promise that resolves once the battle has been started and the switch prompt resolved.
    * @remarks
    * This will temporarily set the current {@linkcode BattleStyle} to `SWITCH` for the duration
@@ -121,7 +123,7 @@ export class ClassicModeHelper extends GameManagerHelper {
    * await game.startBattleWithSwitch(1);
    * ```
    */
-  public async startBattleWithSwitch(pokemonIndex: number): Promise<void> {
+  public async startBattleWithSwitch(pokemonIndex: IntClosedRange<1, 5>): Promise<void> {
     this.game.settings.battleStyle(BattleStyle.SWITCH);
     this.game.onNextPrompt(
       "CheckSwitchPhase",
--- a/test/test-utils/helpers/prompt-handler.ts
+++ b/test/test-utils/helpers/prompt-handler.ts
@@ -62,7 +62,7 @@ export class PromptHandler extends GameManagerHelper {
     super(game);
 
     this.originalSetModeInternal = this.game.scene.ui["setModeInternal"];
-    // `any` assertion needed as we are mocking private property
+    // type assertion needed as we are mocking private property
     vi.spyOn(
       this.game.scene.ui as unknown as Pick<{ setModeInternal: UI["setModeInternal"] }, "setModeInternal">,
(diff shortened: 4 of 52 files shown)


src/phase-manager.ts
```diff
@@ -394,12 +394,17 @@ export class PhaseManager {
 
   /**
    * Helper method to start and log the current phase.
+   *
+   * @remarks
```
SirzBenjie: This comment seems more appropriate for `privateRemarks`

src/phases/title-phase.ts
```diff
@@ -168,27 +168,26 @@ export class TitlePhase extends Phase {
     globalScene.ui.setMode(UiMode.TITLE, config);
   }
 
-  loadSaveSlot(slotId: number): void {
+  // TODO: Make callers actually wait for the damn save slot to load
```
SirzBenjie:
```suggestion
  // TODO: Make callers actually wait for the save slot to load
```

SirzBenjie (approved):
It took me a while to review this PR and make sure I understood the new process. This is with the docs.
That was because our _old_ system really was quite convoluted. I have played around with this branch and it *feels* more ergonomic now. I can only hope it doesn't conflict with @emdeann's phase PR.

test/abilities/screen-cleaner.test.ts
```diff
@@ -24,57 +23,26 @@ describe("Abilities - Screen Cleaner", () => {
 
   beforeEach(() => {
     game = new GameManager(phaserGame);
-    game.override.battleStyle("single").ability(AbilityId.SCREEN_CLEANER).enemySpecies(SpeciesId.SHUCKLE);
+    game.override
+      .battleStyle("single")
+      .ability(AbilityId.SCREEN_CLEANER)
+      .enemySpecies(SpeciesId.SHUCKLE)
+      .weather(WeatherType.HAIL);
```
DayKev:
```suggestion
      .weather(WeatherType.SNOW);
```

test/moves/baton-pass.test.ts
```diff
@@ -74,14 +74,6 @@ describe("Moves - Baton Pass", () => {
 
     // check buffs are still there
     expect(game.field.getEnemyPokemon().getStatStage(Stat.SPATK)).toEqual(2);
-    // confirm that a switch actually happened. can't use species because I
-    // can't find a way to override trainer parties with more than 1 pokemon species
```
DayKev, line 78: You didn't replace this check.

test/mystery-encounter/encounters/bug-type-superfan-encounter.test.ts
```diff
@@ -364,7 +364,9 @@ describe("Bug-Type Superfan - Mystery Encounter", () => {
       await skipBattleRunMysteryEncounterRewardsPhase(game, false);
 
       expect(game).toBeAtPhase("MysteryEncounterRewardsPhase");
-      game.phaseInterceptor["prompts"] = []; // Clear out prompt handlers
+      // Clear out prompt handlers
+      // TODO: Is this even needed?
```
DayKev: You can just remove it and run the test, and see if it still passes.

test/setup/vitest.setup.ts
```diff
@@ -56,21 +57,28 @@ vi.mock(import("i18next"), async importOriginal => {
   return await importOriginal();
 });
 
-global.testFailed = false;
+//#endregion Mocking
+
+//#region Hooks
```
DayKev, line 64:
```suggestion
// #endregion Mocking

// #region Hooks
```

test/setup/vitest.setup.ts
```diff
@@ -56,21 +57,28 @@ vi.mock(import("i18next"), async importOriginal => {
...
+  clearInterval(PromptHandler.runInterval);
+  PromptHandler.runInterval = undefined;
 });
 
-afterAll(() => {
-  global.server.close();
-  MockConsole.printPostTestWarnings();
-  console.log(chalk.hex("#dfb8d8")("Closing i18n MSW server!"));
-});
+//#endregion Hooks
```
DayKev, line 86:
```suggestion
// #endregion Hooks
```

test/test-utils/game-manager.ts
```diff
@@ -171,15 +176,18 @@ export class GameManager {
    * @param mode - The mode to wait for.
    * @param callback - The callback function to execute on next prompt.
    * @param expireFn - Optional function to determine if the prompt has expired.
+   * @param awaitingActionInput - If true, will prevent the prompt from activating until the current {@linkcode AwaitableUiHandler}
+   * is awaiting input; default `false`
+   * @todo Remove in favor of {@linkcode promptHandler.addToNextPrompt}
```
DayKev:
```suggestion
   * @deprecated Remove in favor of {@linkcode promptHandler.addToNextPrompt}
```

test/test-utils/game-manager.ts
```diff
@@ -415,18 +411,26 @@ export class GameManager {
...
    */
-  isCurrentPhase(phaseTarget: PhaseClass | PhaseString) {
-    const targetName = typeof phaseTarget === "string" ? phaseTarget : phaseTarget.name;
-    return this.scene.phaseManager.getCurrentPhase().phaseName === targetName;
+  isCurrentPhase(phaseTarget: PhaseString): boolean;
+  /**
+   * Checks if the current phase matches the target phase.
+   * @param phaseTarget - The target phase.
+   * @returns Whether the current phase matches the target phase
+   * @deprecated - Use `PhaseString` instead
```
DayKev:
```suggestion
   * @deprecated Use `PhaseString` instead
```
Only `@param` and `@typeParam` use a hyphen for whatever reason.

Bertie690:
> That was because our old system really was quite convoluted.

this is exactly why i made the giant write up of what the two systems do.

test/test-utils/helpers/settings-helper.ts
```diff
@@ -1,53 +1,75 @@
...
+import { ExpNotification } from "#enums/exp-notification";
 import { PlayerGender } from "#enums/player-gender";
+import type { GameManager } from "#test/test-utils/game-manager";
 import { GameManagerHelper } from "#test/test-utils/helpers/game-manager-helper";
+import { getEnumStr } from "#test/test-utils/string-utils";
 import chalk from "chalk";
 
 /**
  * Helper to handle settings for tests
+ * @todo Why does this exist
```
DayKev: Wdym? It exists so settings can be manipulated by tests, abstracted away from needing to access the settings directly (after all, settings need to be refactored and that would require modifying every single test that touched them if this didn't exist).

test/test-utils/helpers/settings-helper.ts
```diff
@@ -1,53 +1,75 @@
...
-    this.log(`Gender set to: ${PlayerGender[gender]} (=${gender})`);
+    this.log(`Gender set to PlayerGender.${getEnumStr(PlayerGender, gender)}!`);
   }
 
   /**
    * Change the exp gains speed
-   * @param speed the {@linkcode ExpGainsSpeed} to set
+   * @param speed - the {@linkcode ExpGainsSpeed} to set
    */
   expGainsSpeed(speed: ExpGainsSpeed) {
```
DayKev:
```suggestion
  expGainsSpeed(speed: ExpGainsSpeed): this {
```

test/test-utils/helpers/settings-helper.ts
```diff
@@ -1,53 +1,75 @@
...
+    this.log(`Type Hints ${enable ? "enabled" : "disabled"}!`);
+    return this;
   }
 
   /**
    * Change the player gender
-   * @param gender the {@linkcode PlayerGender} to set
+   * @param gender - The {@linkcode PlayerGender} to set
    */
   playerGender(gender: PlayerGender) {
```
DayKev:
```suggestion
  playerGender(gender: PlayerGender): this {
```

test/test-utils/helpers/settings-helper.ts
```diff
@@ -1,53 +1,75 @@
...
     this.game.scene.gameData.gender = gender;
-    this.log(`Gender set to: ${PlayerGender[gender]} (=${gender})`);
+    this.log(`Gender set to PlayerGender.${getEnumStr(PlayerGender, gender)}!`);
   }
```
DayKev, line 68:
```suggestion
    return this;
  }
```

test/moves/baton-pass.test.ts
Bertie690: 2 reasons:
1. It is jank
2. I am deleting the file and centralizing the tests in my wimp out pr (whenever that may be lol)

test/setup/vitest.setup.ts
Bertie690: @SirzBenjie actually cautioned me AGAINST doing _this exact thing_ with the reasoning that it breaks the `#region` parsing for certain IDEs.
(benjie if you could confirm i remembered correctly that'd be great)

test/ai/ai-moveset-gen.test.ts
```diff
@@ -161,7 +161,8 @@ describe("Unit Tests - ai-moveset-gen.ts", () => {
   describe("", () => {
     //#region boilerplate
     let phaserGame: Phaser.Game;
-    let game: GameManager;
+    // TODO: Can we remove this if nothing uses it?
+    let _game: GameManager;
```
DayKev: Well, remove it and see if the tests break then.

test/mystery-encounter/encounter-test-utils.ts
```diff
@@ -193,7 +152,10 @@ async function handleSecondaryOptionSelect(game: GameManager, pokemonNo: number,
  * @param game
  * @param runRewardsPhase
```
DayKev:
```suggestion
 * @param runRewardsPhase - (Default `true`) Whether to start the `MysteryEncounterRewardsPhase`
```

test/mystery-encounter/encounter-test-utils.ts
```diff
@@ -193,7 +152,10 @@ async function handleSecondaryOptionSelect(game: GameManager, pokemonNo: number,
...
  */
-export async function skipBattleRunMysteryEncounterRewardsPhase(game: GameManager, runRewardsPhase = true) {
+export async function skipBattleRunMysteryEncounterRewardsPhase(
+  game: GameManager,
+  runRewardsPhase?: false | undefined,
+) {
```
DayKev:
```suggestion
export async function skipBattleRunMysteryEncounterRewardsPhase(game: GameManager, runRewardsPhase?: false) {
```
Making it optional should automatically add `undefined` to the type signature, and there's no reason to ever manually pass `undefined` to it.

test/setup/vitest.setup.ts
DayKev: Just leaving a note that it wasn't breaking anything, he just preferred it that way. So 🤷

test/test-utils/helpers/classic-mode-helper.ts
```diff
@@ -119,7 +122,7 @@ export class ClassicModeHelper extends GameManagerHelper {
    * ```
    */
   public async startBattleWithSwitch(pokemonIndex: number): Promise<void> {
```
DayKev:
```suggestion
  public async startBattleWithSwitch(pokemonIndex: number): Promise<void> {
    if (!isBetween(pokemonIndex, 1, 5)) {
      expect.fail(`Invalid party index (${pokemonIndex}) passed to \`startBattleWithSwitch\`: must be between 1 and 5 (inclusive)!`);
    }
```
Probably needs the import for `isBetween` added.

I think you can use `type-fest` to make the typing of `pokemonIndex` only 1-5 instead, if you want to look into that.

test/test-utils/helpers/classic-mode-helper.ts
```diff
@@ -110,7 +110,10 @@ export class ClassicModeHelper extends GameManagerHelper {
    * Queue inputs to switch at the start of the next battle, and then start it.
    * @param pokemonIndex - The 0-indexed position of the party pokemon to switch to.
    * Should never be called with 0 as that will select the currently active pokemon and freeze
```
DayKev:
```suggestion
   * @throws {@linkcode Error}
   * Fails test if `pokemonIndex` is out of valid bounds
```

test/test-utils/helpers/prompt-handler.ts
```diff
@@ -0,0 +1,172 @@
...
+  private readonly originalSetModeInternal: (typeof this.game.scene.ui)["setModeInternal"];
+
+  /** A {@linkcode NodeJS.Timeout | Timeout} containing an interval used to check prompts. */
+  public static runInterval?: NodeJS.Timeout;
+
+  constructor(game: GameManager) {
+    super(game);
+
+    this.originalSetModeInternal = this.game.scene.ui["setModeInternal"];
+    // `any` assertion needed as we are mocking private property
```
DayKev:
```suggestion
    // `as unknown as ...` assertion needed as we are mocking private property
```

test/test-utils/helpers/prompt-handler.ts
```diff
@@ -0,0 +1,172 @@
...
+      awaitingActionInput,
+    });
+  }
+
+  /**
+   * Wrapper function to add coloration to phase logs.
+   * @param args - Arguments to original logging function
+   */
+  // TODO: Move this to colors.ts & change color after mock console PR
+  private doLog(...args: unknown[]): void {
```
DayKev, line 169: Minor: I think `printLog` would be a better method name.
(Same for the other classes with `doLog`).

test/test-utils/helpers/prompt-handler.ts
```diff
@@ -0,0 +1,172 @@
...
+  ): ReturnType<typeof this.originalSetModeInternal> {
+    const mode = args[0];
+
+    this.doLog(
+      `UI mode changed from ${getEnumStr(UiMode, this.game.scene.ui.getMode())} to ${getEnumStr(UiMode, mode)}!`,
+    );
+    // TODO: Add `await` to this
+    const ret = this.originalSetModeInternal.apply(this.game.scene.ui, args);
+
+    const currentPhase = this.game.scene.phaseManager.getCurrentPhase()?.phaseName!;
```
DayKev:
```suggestion
    const currentPhase = this.game.scene.phaseManager.getCurrentPhase().phaseName;
```
I guess this was done before you removed `null` from `getCurrentPhase`?
Also it should have just been `getCurrentPhase()!.phaseName;` instead of `getCurrentPhase()?.phaseName!;` because the latter is a bit silly.

test/test-utils/mocks/mock-phase.ts
```diff
@@ -0,0 +1,12 @@
+import { Phase } from "#app/phase";
+
+/**
+ * A rudimentary mock of a phase used for unit tests.
+ * Ends upon starting by default.
+ */
+export abstract class mockPhase extends Phase {
```
DayKev:
```suggestion
export abstract class MockPhase extends Phase {
```

test/test-utils/phase-interceptor.ts
```diff
@@ -1,450 +1,224 @@
...
-  callback(): void;
-  onError(error: any): void;
-}
-
-interface onHoldStub {
-  name: string;
-  call(): void;
-}
-
+// biome-ignore-start lint/correctness/noUnusedImports: TSDoc imports
```
DayKev:
```suggestion
```

test/test-utils/phase-interceptor.ts
```diff
@@ -1,450 +1,224 @@
...
+import type { GameManager } from "#test/test-utils/game-manager";
+import type { PromptHandler } from "#test/test-utils/helpers/prompt-handler";
+import type { Constructor } from "#types/common";
+// biome-ignore-end lint/correctness/noUnusedImports: TSDoc imports
```
DayKev:
```suggestion
```

test/test-utils/phase-interceptor.ts
```diff
@@ -1,450 +1,224 @@
...
-   *
-   * CAUTION: If a phase and its subclasses (if any) both appear in this list,
-   * make sure that this list contains said phase AFTER all of its subclasses.
-   * This way, the phase's `prototype.start` is properly preserved during
-   * `initPhases()` so that its subclasses can use `super.start()` properly.
+   * The interceptor's current state.
+   * Possible values are the following:
+   * - `running`: The interceptor is currently running a phase.
+   * - `interrupted`: The interceptor has been interrupted by a UI prompt and is waiting for the caller to end it.
+   * - `idling`: The interceptor is not currently running a phase and is ready to start a new one.
```
DayKev:
```suggestion
   * @see {@linkcode StateType}
```
Duplicated documentation tends to get out of sync very quickly.

test/test-utils/tests/helpers/prompt-handler.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  let handler: AwaitableUiHandler;
+
+  let callback1: Mock;
+  let callback2: Mock;
+  let setModeCallback: Mock;
+  let checkModeCallback: Mock;
+
+  beforeEach(() => {
+    setModeCallback = vi.fn(() => console.log("set mode called!")).mockName("set mode callback");
+    checkModeCallback = vi.fn(() => console.log("check mode called!")).mockName("check mode callback 2");
```
DayKev:
```suggestion
    checkModeCallback = vi.fn(() => console.log("check mode called!")).mockName("check mode callback");
```

test/test-utils/tests/phase-interceptor/unit.test.ts
```diff
@@ -0,0 +1,150 @@
...
+import type { Phase } from "#app/phase";
+import { GameManager } from "#test/test-utils/game-manager";
+import { mockPhase } from "#test/test-utils/mocks/mock-phase";
+import type { Constructor } from "#types/common";
+import type { PhaseString } from "#types/phase-types";
+import Phaser from "phaser";
+import { beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+
+// TODO: Move these to `mock-phase.ts` if/when unit tests for the phase manager are created
+class applePhase extends mockPhase {
```
DayKev:
```suggestion
class ApplePhase extends mockPhase {
```
Same with the rest, classes are PascalCase.

test/setup/vitest.setup.ts
Bertie690: Seriously?

I was told that was required for it not to break.
I mean it's pathetically easy to enforce one or the other - a simple regex would do the trick.

test/test-utils/helpers/classic-mode-helper.ts
Bertie690: Yup it has an int range type
(Sadly doesn't work on negatives, but i'll live with what i have.)

test/test-utils/helpers/prompt-handler.ts
Bertie690:
```suggestion
    // type assertion needed as we are mocking private property
```

test/test-utils/helpers/prompt-handler.ts
Bertie690: that can be for another day

DayKev:
Notes for the future:
https://github.com/pagefaultgames/pokerogue/pull/6012#discussion_r2648878714
https://github.com/pagefaultgames/pokerogue/pull/6012#discussion_r2648887489

src/constants/colors.ts
```diff
@@ -16,6 +16,8 @@ export const NEW_TURN_COLOR = "#ffad00ff" as const;
 export const UI_MSG_COLOR = "#009dffff" as const;
 export const OVERRIDES_COLOR = "#b0b01eff" as const;
 export const SETTINGS_COLOR = "#008844ff" as const;
+export const PHASE_INTERCEPTOR_COLOR = "#ff7f50" as const;
+export const PROMPT_HANDLER_COLOR = "#003cbdff" as const;
```
SirzBenjie, line 20: Is the extra 00 at the beginning the alpha? Interesting. Is it required to have the extra 2 digits?

test/ai/ai-moveset-gen.test.ts
SirzBenjie: @Bertie690 Can you remove it locally and see if tests break. If they don't, push another commit that removes the line and the comment.

src/constants/colors.ts
Bertie690: AFAIK VS Code's hex color picker changes hex codes automatically from `RRGGBB` to `RRGGBBAA`.
(Basically all hex codes assume alphas of `FF` if ommitted, but no harm no foul I guess.)

test/ai/ai-moveset-gen.test.ts
Bertie690: sure

PR #6016, opened 2025-06-20
[Balance] Berry Juice and Old Gateau do not depend on the instance

Wlowscha:
These are the only held items in the game which, in principle, have different parameters for each instance. In practice, the values are fixed by the respective MEs. It makes sense therefore to remove the instance-dependent parameters, especially in light of #5656 .

~Berry Juice still takes a parameter, now a boolean `good` (indicating whether the effect is positive or negative). The value of the stat increase or decrease is fixed in the item.~
I have reverted to using numeric values for the item, otherwise a migrator would be needed.
Separate icons and locales keys for the two versions of the item are necessary.

Old Gateau takes no parameters. The stat increase is fixed at 20. The affected stats are decided by a function that now acts on the pokemon itself, so they could change if the pokemon evolves or changes form.
Since the new version of the item has no parameters, a migrator is not necessary.

## Checklist

Locales PR: https://github.com/pagefaultgames/pokerogue-locales/pull/189


src/data/mystery-encounters/encounters/weird-dream-encounter.ts
```diff
@@ -104,7 +103,7 @@ const EXCLUDED_TRANSFORMATION_SPECIES = [
 const SUPER_LEGENDARY_BST_THRESHOLD = 600;
 const NON_LEGENDARY_BST_THRESHOLD = 570;
 
-const OLD_GATEAU_STATS_UP = 20;
+//const OLD_GATEAU_STATS_UP = 20;
```
DayKev:
```suggestion
```

SirzBenjie (dismissed):
LGTM, though not keen on the names "good_berry_juice" and "bad_berry_juice".
At least not on what's displayed to the user. In the codebase, it's fine to call it that so that devs don't have to know that something like "sour = bad, sweet = positive" (or whatever we come up with)

Wlowscha:
<details><summary>Images</summary>
<p>

wrong sour
shuckle
sweet_juice
old_gateau

</p>
</details> 

Showing the new item descriptions, + correct icon showing up in the Shuckle encounter.

For some reason, the `berry_juice_bad` icon is showing up as the other one. This happened even when I tried to remove it entirely.

damocleas:
just deciding on new assets/locales name, meant to approve for balance related since nothing changes

DayKev (approved):

PR #6017, opened 2025-06-20
[Refactor] Minor run phase rework

jnotsknab:

Hopefully none other than the original bug referenced being fixed.

While fixing the bug introduced by issue #5982, it was revealed that the `AttemptRunPhase` itself had unnecessary logic and was a poor implementation. `AttemptRunPhase` should be a team based action rather than referring to a specific Pokemon, in turn these changes will also resolve PR #5999  which originally refereed to  the bug.

- Modified the `getPlayerField` method of the `BattleScene` class to accept a boolean parameter to only return an array of the active Pokemon on the field.
- Altered `AttemptRunPhase`  to now return the random battle seed based on the global battle scene rather than from an individual Pokemon.
- Introduced `calculateEscapeChance` in lieu of the original `attemptRunAway` method for reducing calculation bloat and better naming conventions.
- Intrinsically solved the speed boost bug by looping over the active player field array and setting each pokemon's turn data accordingly.
- Removed redundant references to pokemon and checks on whether its a doubles battle in the `TurnStartPhase`
## Screenshots/Videos

N/A
## How to test the changes?

Use the following overrides to ensure the speed bug no longer exists:

`const overrides = {
  ABILITY_OVERRIDE: AbilityId.SPEED_BOOST,
  OPP_HEALTH_SEGMENTS_OVERRIDE: 2,
  BATTLE_STYLE_OVERRIDE: "double",
  OPP_MOVESET_OVERRIDE: MoveId.SPLASH,
  RUN_SUCCESS_OVERRIDE: false } satisfies Partial<InstanceType<OverridesType>>;`

Other than that running away from battle should function as expected as the calculations themselves weren't changed.
## Checklist


src/battle-scene.ts
```diff
@@ -800,12 +800,14 @@ export default class BattleScene extends SceneBase {
   // TODO: Add `undefined` to return type
   /**
    * Returns an array of PlayerPokemon of length 1 or 2 depending on if in a double battle or not.
-   * Does not actually check if the pokemon are on the field or not.
+   * @param active - If true, returns only the PlayerPokemon that are currently active
```
Bertie690:
```suggestion
   * @param active - Whether to consider only active on-field pokemon ({@see {@linkcode Pokemon.isActive} for more information}); default `false`
```

src/battle.ts
```diff
@@ -94,6 +94,9 @@ export default class Battle {
   /** If the current battle is a Mystery Encounter, this will always be defined */
   public mysteryEncounter?: MysteryEncounter;
 
+  /**Tracks whether the last run attempt in battle failed*/
```
Bertie690:
```suggestion
  /** 
   * Tracker for whether the last run attempt failed.
   * @defaultValue `false`
   */
```

src/phases/attempt-run-phase.ts
```diff
@@ -1,33 +1,38 @@
...
+import { NumberHolder, randSeedInt } from "#app/utils/common";
 import { globalScene } from "#app/global-scene";
-
-export class AttemptRunPhase extends PokemonPhase {
+import { calculateEscapeChance } from "#app/utils/run-utils";
+import { FieldPhase } from "./field-phase";
+export class AttemptRunPhase extends FieldPhase {
   public readonly phaseName = "AttemptRunPhase";
   /** For testing purposes: this is to force the pokemon to fail and escape */
   public forceFailEscape = false;
```
Bertie690: I humbly request this constant be axed.

To keep a long story short, the original creator littered random debug constants like these all throughout the codebase, which (while fine as temp placeholders) should have _never_ made their way into actual production code.

src/phases/attempt-run-phase.ts
```diff
@@ -1,33 +1,38 @@
...
 
+  private getTeamRNG(range: number, min = 0) {
+    return globalScene.currentBattle ? globalScene.randBattleSeedInt(range, min) : randSeedInt(range, min);
```
Bertie690: Don't take my word for it, but I highly doubt `globalScene.currentBattle` can be `null` when this phase is started - running always happens _during a battle_, so there should be no concievable circumstance where flee checks should need to fallback on global RNG.
(That said, I am literally an overenthusiastic teenager so maybe @DayKev or someone else can chime in to confirm).

src/phases/attempt-run-phase.ts
```diff
@@ -1,33 +1,38 @@
...
+  }
+
   start() {
     super.start();
 
-    const playerField = globalScene.getPlayerField();
-    const enemyField = globalScene.getEnemyField();
+    //Increment escape attempts count on entry
```
Bertie690:
```suggestion
    // Increment escape attempts count on entry
```

src/phases/attempt-run-phase.ts
```diff
@@ -1,33 +1,38 @@
...
+    const enemyField = globalScene.getEnemyField();
 
+    const escapeRoll = this.getTeamRNG(100);
     const escapeChance = new NumberHolder(0);
 
-    this.attemptRunAway(playerField, enemyField, escapeChance);
+    calculateEscapeChance(activePlayerField, enemyField, escapeChance, currentAttempts);
 
-    applyAbAttrs("RunSuccessAbAttr", playerPokemon, null, false, escapeChance);
+    activePlayerField.forEach(p => {
```
Bertie690: 2 notes that you probably should just read and ignore:
1. This should check the player's party in speed order. HOWEVER, this problem is so systematic that I'd recommend adding a TODO for dean to fix during his PR (IIRC one of the other devs was starting work on this kinda stuff).
2. We need to break out of the loop early if the attribute makes run chance hit 100%. (Incidentally, this requires using a `for... of` instead of `forEach` since you can't break out of the callback manually).
    - Come to think of it, Run Away should _probably_ just use a `BooleanHolder` for pass/fail rather than applying to escape chance, but that can probably be something for another PR.

src/phases/attempt-run-phase.ts
Bertie690: In fact, given the escape rolls have been extracted out, any callers attempting to force a failed run attempt can simply mock the RNG function itself - there is _literally zero reason_ for this to exist anymore.

src/phases/battle-phase.ts
```diff
@@ -3,6 +3,19 @@ import { TrainerSlot } from "#enums/trainer-slot";
...
+
+      activePlayerField.forEach(p => {
+        p.turnData.failedRunAway = true;
+      });
+
+      //Reset flag for future run attempts
+      globalScene.currentBattle.failedRunAway = false;
+    }
+  }
+
```
Bertie690:
```suggestion
  start() {
    // Reset flag for run attempts
    globalScene.currentBattle.failedRunAway = false;
  }
```

Most of the above chunk is redundant for 2 reasons:
1. Clearing turn data is done _automatically_ in `turn-init-phase` before the command handling is even called.
2. Setting an already-false `boolean` to `false` does about as much as you'd expect it to (ie nothing).

The boolean resetting may not even be needed by the looks of it (though I'd rather not delve into the 🍝 that is `globalScene.newBattle` right now) - the easiest way to test it is just removing it and seeing if anything breaks!

src/phases/turn-start-phase.ts
```diff
@@ -209,27 +208,8 @@ export class TurnStartPhase extends FieldPhase {
...
-                  playerActivePokemon[0].getStat(Stat.SPD) > playerActivePokemon[1].getStat(Stat.SPD)
-                    ? playerActivePokemon[0]
-                    : playerActivePokemon[1];
-                // check if either active pokemon has the ability "Run Away"
-                const hasRunAway = playerActivePokemon.find(p => p.hasAbility(AbilityId.RUN_AWAY));
-                runningPokemon = hasRunAway !== undefined ? hasRunAway : fasterPokemon;
-              }
-            }
-            phaseManager.unshiftNew("AttemptRunPhase", runningPokemon.getFieldIndex());
+            //Team Based action, no need for checking individual pokemon or whether its doubles
```
Bertie690:
```suggestion
            // Running (like ball throwing) is a team action taking up both Pokemon's turns.
```

src/utils/run-utils.ts
```diff
@@ -0,0 +1,71 @@
...
+import type Pokemon from "#app/field/pokemon";
+import { Stat } from "#enums/stat";
+
+/**
+ * Calculates the chance for the player's team to successfully run away from battle.
+ *
+ * @param playerField       The player's currently active Pokémon
+ * @param enemyField        The enemy team's Pokémon
+ * @param escapeChance      A mutable NumberHolder to be assigned the final escape chance value
+ * @param escapeAttempts    The number of previous escape attempts made in the battle
```
Bertie690:
```suggestion
  * @param escapeAttempts - The number of prior failed escape attempts in the current battle.
  * @returns The final escape chance, as a percentage out of 100. 
```

### Remarks:
Rather than passing the on-field pokemon directly, it makes more sense to access them from `globalScene` instead.
Additionally, making the escape chance a direct return value rather than passed via a `NumberHolder` is both more sensible and plays nicer with mocks.

src/utils/run-utils.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+export function calculateEscapeChance(
+  playerField: PlayerPokemon[],
+  enemyField: EnemyPokemon[],
+  escapeChance: NumberHolder,
+  escapeAttempts: number,
+) {
+  /** Sum of the speed of all enemy pokemon on the field */
```
Bertie690: Can we get an override for escape success?

Something like this could go inside `overrides.ts`
```ts
/** 
 * If non-null, will override random flee attempts to always or never succeed by forcing {@linkcode calculateEscapeChance} to return 100% or 0%.
 * Set to `null` to disable.
 * 
 * Is overridden if either player Pokemon has {@linkcode AbilityId.RUN_AWAY | Run Away}.
 */
  RUN_SUCCESS_OVERRIDE: boolean | null = null
```

and something like this in the success chance:
```ts
  // Check for override, guaranteeing or forbidding random flee attempts as applicable.
  if (Overrides.RUN_SUCCESS_OVERRIDE !== null) {
    return Overrides.RUN_SUCCESS_OVERRIDE ? 100 : 0
```

src/utils/run-utils.ts
```diff
@@ -0,0 +1,71 @@
...
+  const enemySpeed = enemyField.reduce(
+    (total: number, enemyPokemon: Pokemon) => total + enemyPokemon.getStat(Stat.SPD),
```
Bertie690:
```suggestion
    (total, enemy) => total + enemy.getStat(Stat.SPD),
```

src/utils/run-utils.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @param escapeAttempts    The number of previous escape attempts made in the battle
+ */
+export function calculateEscapeChance(
+  playerField: PlayerPokemon[],
+  enemyField: EnemyPokemon[],
+  escapeChance: NumberHolder,
+  escapeAttempts: number,
+) {
+  /** Sum of the speed of all enemy pokemon on the field */
+  const enemySpeed = enemyField.reduce(
```
Bertie690:
```suggestion
  // Cf https://bulbapedia.bulbagarden.net/wiki/Escape#Generation_V_onwards
  // From gen 5 onwards, running takes the _base_ speed totals of both party sides.
  const enemySpeed = enemyField.reduce(
```

src/utils/run-utils.ts
```diff
@@ -0,0 +1,71 @@
...
+    (total: number, enemyPokemon: Pokemon) => total + enemyPokemon.getStat(Stat.SPD),
+    0,
+  );
+  /** Sum of the speed of all player pokemon on the field */
+  const playerSpeed = playerField.reduce(
+    (total: number, playerPokemon: Pokemon) => total + playerPokemon.getStat(Stat.SPD),
```
Bertie690:
```suggestion
  const playerSpeed = playerField.reduce(
    (total, player) => total + player.getStat(Stat.SPD),
```

src/utils/run-utils.ts
```diff
@@ -0,0 +1,71 @@
...
+   *   - The speed ratio cap should be 6x for bosses and 4x for non-bosses
+   *   - The "default" escape chance when your speed equals the enemy speed should be 8.33% for bosses and 27.5% for non-bosses
+   *
+   *  From the above, we can calculate the below values
+   */
+
+  let isBoss = false;
+  for (let e = 0; e < enemyField.length; e++) {
+    isBoss = isBoss || enemyField[e].isBoss(); // this line checks if any of the enemy pokemon on the field are bosses; if so, the calculation for escaping is different
+  }
```
Bertie690:
```suggestion
  /** Whether at least 1 pokemon on the enemy field is a boss. */
  const isBoss = enemyField.some(e => e.isBoss());
```

test/abilities/desolate-land.test.ts
```diff
@@ -145,6 +146,7 @@ describe("Abilities - Desolate Land", () => {
     expect(game.scene.arena.weather?.weatherType).toBe(WeatherType.HARSH_SUN);
 
     vi.spyOn(game.scene.getPlayerPokemon()!, "randBattleSeedInt").mockReturnValue(0);
+    vi.spyOn(globalScene, "randBattleSeedInt").mockReturnValue(0);
```
Bertie690, line 149: Something worth noting: It might make sense to add a `forceRunFail` helper to overrides.

Such a function would:
1. Queue inputs to run away from a fight (see the switch helper code for inspiration, just hitting cursor value of 4 once)
2. Mock the override for attempt running to the specified value.

smthng like
```ts
/**
 * Perform menu inputs to select the `RUN` option in-battle,
 * optionally mocking {@linkcode Overrides.RUN_SUCCESS_OVERRIDE | RUN_SUCCESS_OVERRIDE} for the duration of the ensuing `AttemptRunPhase`.
 * @param success - If provided, will force the run attempt to either succeed or fail; default `true`
 * @remarks
 * This does not account for Run Away, which bypasses the random roll altogether.
 */
  public tryRunAway(success?: boolean) {
    // smthng probably involving `game.setCursor`, `game.onNextPrompt` and/or `vi.spyOn().mockReturnValueOnce` idk
```

Bertie690:
One other thing I forgot - can we get a regression test for speed boost?

Just a little smthng checking that both player pokemon in doubles do indeed not get SPD boosts if a run attempt fails.

src/phases/attempt-run-phase.ts
DayKev: If `currentBattle` is `null` here then the game then something went horribly wrong, yeah. You don't need to worry about checking for/handling it (and if somehow it turns out `currentBattle` *can* be `null` here, well it'll be good to know about it so we can fix it).

src/phases/attempt-run-phase.ts
```diff
@@ -60,64 +65,10 @@ export class AttemptRunPhase extends PokemonPhase {
...
     } else {
-      playerPokemon.turnData.failedRunAway = true;
+      globalScene.currentBattle.failedRunAway = true;
       globalScene.phaseManager.queueMessage(i18next.t("battle:runAwayCannotEscape"), null, true, 500);
     }
 
     this.end();
   }
-
-  attemptRunAway(playerField: PlayerPokemon[], enemyField: EnemyPokemon[], escapeChance: NumberHolder) {
```
DayKev, line 67: What was the reason for moving this out of the `AttemptRunPhase` class? Was it in order to use it in the tests?

src/phases/attempt-run-phase.ts
SirzBenjie: @Bertie690 is correct.

src/phases/attempt-run-phase.ts
SirzBenjie: Can definitely be something for another PR. I wouldn't worry about splitting it in this one.

src/phases/attempt-run-phase.ts
SirzBenjie: I can't speak for the author, but I actually really _like_ this change. It makes the phase more purposeful.

test/abilities/desolate-land.test.ts
SirzBenjie: Sure, but not in this PR.

SirzBenjie (commented):
This looks great!
I agree with many of @Bertie690's suggestions. 
Thank you very much for the PR!

src/utils/run-utils.ts
```diff
@@ -0,0 +1,71 @@
+import type { PlayerPokemon, EnemyPokemon } from "#app/field/pokemon";
+import type { NumberHolder } from "#app/utils/common";
+import type Pokemon from "#app/field/pokemon";
+import { Stat } from "#enums/stat";
+
+/**
+ * Calculates the chance for the player's team to successfully run away from battle.
```
SirzBenjie:
```suggestion
 * Calculate the chance for the player's team to successfully run away from battle.
```

src/phases/attempt-run-phase.ts
DayKev: Well the purpose of the phase is to handle [attempting] running away from a battle by the player; the escape calculation only has relevance here.

jnotsknab: Mainly for seperation of concerns and added clarity in tests as the function performs a pure calculation which doesnt depend on the internal state. So, IMO centralizing the calculation into a utility felt cleaner and more aligned with structure conventions, however if the consensus leans more torward reintroducing it into the phase then I am not opposed.

src/phases/battle-phase.ts
jnotsknab: In theory this would work, but in my testing it seems that turn data is cleared after `failedRunAway` is set, so this causes the speed bug to persist, for now I just reintroduced it into the phase itself upon failing to prevent the bug while still satisfying the changes you requested. (Boolean resetting was indeed not needed)

src/phases/attempt-run-phase.ts
DayKev: I think it makes sense to keep the function in the same file, even if it's extracted from the class. It's not really a general-purpose utility function since only this phase and its tests have any use for it.

Also the `playerField` and `enemyField` params for the function are redundant, since you can just do `const playerField = globalScene.getPlayerField();` and `const enemyField = globalScene.getEnemyField();` from within the function itself and get the same result.

DayKev (dismissed):

src/phases/attempt-run-phase.ts
```diff
@@ -110,9 +113,14 @@ export class AttemptRunPhase extends PokemonPhase {
     /** Slope of the escape chance curve */
     const escapeSlope = (maxChance - minChance) / speedCap;
 
+    //   Check for override, guaranteeing or forbidding random flee attempts as applicable.
+    if (Overrides.RUN_SUCCESS_OVERRIDE !== null) {
+      return Overrides.RUN_SUCCESS_OVERRIDE ? 100 : 0;
+    }
+
```
Bertie690: Extremely minor nit - this early return should probably be at the start of the function (before the variables)

Bertie690 (approved):

DayKev (approved):

Bertie690 (approved):

PR #6030, opened 2025-06-24
Update voucher.ts

GnawOmiz:
Optimized Suboptimal Code

### DIFF
--- a/src/system/voucher.ts
+++ b/src/system/voucher.ts
@@ -12,28 +12,39 @@ export enum VoucherType {
   GOLDEN,
 }
 
+const VoucherTypeNames: Record<VoucherType, string> = {
+  [VoucherType.REGULAR]: "voucher:eggVoucher",
+  [VoucherType.PLUS]: "voucher:eggVoucherPlus",
+  [VoucherType.PREMIUM]: "voucher:eggVoucherPremium",
+  [VoucherType.GOLDEN]: "voucher:eggVoucherGold",
+};
+
+const VoucherTypeIcons: Record<VoucherType, string> = {
+  [VoucherType.REGULAR]: "coupon",
+  [VoucherType.PLUS]: "pair_of_tickets",
+  [VoucherType.PREMIUM]: "mystic_ticket",
+  [VoucherType.GOLDEN]: "golden_mystic_ticket",
+};
+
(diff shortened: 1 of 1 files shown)


src/system/voucher.ts
```diff
@@ -89,35 +73,34 @@ export interface Vouchers {
...
-    vouchers[key] = new Voucher(voucherType, `${i18next.t("voucher:defeatTrainer", { trainerName })} ${title}`);
+  for (const [key, config] of Object.entries(trainerConfigs)) {
+    if (
+      config.isBoss &&
+      config.getDerivedType() !== TrainerType.RIVAL &&
+      config.hasVoucher
+    ) {
+      const voucherType = config.moneyMultiplier < 10 ? VoucherType.PLUS : VoucherType.PREMIUM;
+      const title = config.title ? ` (${config.title})` : "";
+      vouchers[TrainerType[key as keyof typeof TrainerType]] = new Voucher(
```
DayKev, line 96:
```suggestion
      vouchers[TrainerType[key]] = new Voucher(
```
This type casting doesn't appear to be necessary.

DayKev (commented):
You need to run Biome to lint/format the file (`pnpm exec biome check --write --unsafe src/system/voucher.ts`) but otherwise it seems good.

DayKev:
> Optimized Suboptimal Code

You should not remove the PR template, you're supposed to fill it out (if some heading isn't relevant to the PR like the "what will the user experience" you can put "N/A" for that section).

src/system/voucher.ts
```diff
@@ -43,43 +54,16 @@ export class Voucher {
...
-      case VoucherType.REGULAR:
-        return AchvTier.COMMON;
-      case VoucherType.PLUS:
-        return AchvTier.GREAT;
-      case VoucherType.PREMIUM:
-        return AchvTier.ULTRA;
-      case VoucherType.GOLDEN:
-        return AchvTier.ROGUE;
-    }
+    return VoucherTiers[this.voucherType];
```
SirzBenjie, line 57: The switch is actually better because it uses less data and one less lookup.

SirzBenjie:
I'm closing this since comments haven't been addressed after a month.
We can accept it if there's interest in properly applying biome and addressing the comments.

PR #6031, opened 2025-06-25
[Dev] Mark the fixer for Biome's `useBlockStatements` as "safe"

DayKev:
N/A

Improved developer experience.

Biome's `lint/style/useBlockStatements` rule will automatically be applied without the developer needing to manually apply the fix (either in the IDE or by adding the `--unsafe` flag when running Biome from the command line).

## How to test the changes?
Put `if (globalScene) console.log("test");` in some file and run `pnpm biome` (or `pnpm exec biome check --write path/to/file.ts`). If you have your IDE configured to automatically apply Biome on saving a file then it should automatically be modified without you needing to run Biome from the command line.
```ts
// this
if (globalScene) console.log("test");

// should become this
if (globalScene) {
  console.log("test");
}
```

## Checklist

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -70,7 +70,10 @@
       },
       "style": {
         "useEnumInitializers": "off", // large enums like Moves/Species would make this cumbersome
-        "useBlockStatements": "error",
+        "useBlockStatements": {
+          "level": "error",
+          "fix": "safe"
+        },
         "useConst": "error",
         "useImportType": "error",
         "noNonNullAssertion": "off", // TODO: Turn this on ASAP and fix all non-null assertions in non-test files


SirzBenjie (approved):

SirzBenjie:
Worth noting that the reason this is marked unsafe likely has to do with variable scoping issues. So not to worry about making it a safe fix.

@DayKev Can we maybe also mark the switch declaration thing as safe (for what it's called)

DayKev:
> Worth noting that the reason this is marked unsafe likely has to do with variable scoping issues. So not to worry about making it a safe fix.
> 
> @DayKev Can we maybe also mark the switch declaration thing as safe (for what it's called)

It looks like they upgraded it to a safe fix in Biome 2.0: https://biomejs.dev/linter/rules/no-switch-declarations/
They beat us to it lol.

Bertie690 (approved):
LGTM

PR #6039, opened 2025-06-27
[Dev] Generate moveset when using player Species override

DayKev:
N/A

Improved testing workflow.

Player Pokémon will now generate an NPC moveset when the player species override is used (except for Keldeo, to prevent crashes) instead of having no moves. The moveset override will still apply normally if used.

## How to test the changes?
```ts
const overrides = {
  STARTER_SPECIES_OVERRIDE: SpeciesId.VENUSAUR,
```

## Checklist


src/field/pokemon.ts
```diff
@@ -5620,7 +5620,10 @@ export class PlayerPokemon extends Pokemon {
     }
 
     if (!dataSource) {
-      if (globalScene.gameMode.isDaily) {
+      if (
+        globalScene.gameMode.isDaily ||
+        (Overrides.STARTER_SPECIES_OVERRIDE && Overrides.STARTER_SPECIES_OVERRIDE !== SpeciesId.KELDEO)
```
DayKev, line 5714:
```suggestion
        // Keldeo is excluded due to crashes involving its signature move and the associated form change
        (Overrides.STARTER_SPECIES_OVERRIDE && Overrides.STARTER_SPECIES_OVERRIDE !== SpeciesId.KELDEO)
```

Wlowscha (approved):

PR #5925, opened 2025-06-03
[Bug] Fix: Mystery Encounters tooltip and description not scrolling properly

SmhMyHead:
## What are the changes the user will see?
The rewards tooltip and description of the Mystery Encounters will scroll properly now, the text won't get interrupted anymore.

## Why am I making these changes?
As far as I know there wasn't an issue created for this, but the bug has been going since version 1.8.

## What are the changes from a developer perspective?
No more arbitrary numbers on the description and tooltip Tweens. Now it calculates how much it should scroll and if it should scroll by checking descriptionTextObject.displayHeight (how tall the text is) and descriptionTextMaskRect.fillRect's height (how tall the box is).

The scroll speed / duration have maintained the "line height" used before to keep the same speed (10 for description and 11.2 for tooltip).

## Screenshots/Videos
<details><summary>English 1</summary>

https://github.com/user-attachments/assets/f57f1bc4-e8e0-4c62-a2b4-2b1d5393ead3

</details>
<details><summary>English 2</summary>

https://github.com/user-attachments/assets/599787fc-93d0-4f4a-a15e-b5e1e750a61b

</details>
<details><summary>Simplified Chinese</summary>

https://github.com/user-attachments/assets/ecb09cdb-544c-46f2-81d6-ddd7837f0da2

</details>
<details><summary>German</summary>

https://github.com/user-attachments/assets/52891613-6714-419d-ab3f-7837917b8e2a

</details>

## How to test the changes?
```typescript
const overrides = {
  MYSTERY_ENCOUNTER_RATE_OVERRIDE: 256,
  STARTING_WAVE_OVERRIDE: 11,
  STARTING_MONEY_OVERRIDE: 1000000,
  STARTING_LEVEL_OVERRIDE: 50,

} satisfies Partial<InstanceType<OverridesType>>;
```

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test:silent`)
 ~- [ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?~
- [x] Have I provided screenshots/videos of the changes (if applicable)?
  - [x] Have I made sure that any UI change works for both UI themes (default and legacy)?

~Are there any localization additions or changes? If so:~
[...]

PR #6042, opened 2025-06-28
refactor(ui): create a custom object for scrolling text

Wlowscha:
Scrolling text is used in various places in the code, most notably in move overlays and in the summary. Setting up scrolling text requires a few different steps which can be time-consuming to go through.

The goal of this PR is to create a standardized "scrolling text" object which can be used more easily wherever needed. Objects that already used scrolling text have been refactored to employ the new class, as examples of how it should be used.

Introduced a new `Scrolling text`, extanding `Container`, which has three methods:
- The constructor sets up a background element, plus the text. There is an option to show the background, in which case an offset is automatically added between the background and the text; otherwise, there is no offset.
- The `createMask` method creates a mask object which is used to crop the text to the desired area. This methods must be passed as input the _global_ coordinates of the scrolling text object. It is done this way because of two reasons: 1) the mask must be positioned relative to the scene, not to a parent container; 2) there is no simple way to recover the global coordinates from inside the `ScrollingText` container. The latter would be much more desirable if possible. This is a separate method because, in principle, if the position of the  `ScrollingText` container is changed, the position of the mask would have to be recalculated. In practice, this does not happen in the code anywhere, but could be the case later on.
- The `activate` method is responsible for starting the animation that makes the text scroll, while the mask ensures that it is correctly cropped.

In order to set up the tween doing the scrolling, we need to calculate the total text to display vs the currently shown text (size of the mask). For some reason, `text.displayHeight` does not correspond exactly to (height of a line * number of lines). This causes issues, in particular, when a custom line spacing is passed as a style option. This causes the scrolling distance to be wrong, which causes visual bugs. To fix this, I had to calculate the total height of the text manually:
```
    const lineHeight = this.text.style.lineHeight / 6;
    const displayHeight = lineHeight * Math.round(this.text.displayHeight / lineHeight);
    const scrollAmount = displayHeight - this.maskHeight;
```
It seems that `round` is the right function here; depending on the line spacing, the difference could be positive or negative, which makes either `floor` or `ceil` unreliable.

EDIT: The container now has its own `setText` method to update the text. This method has extra parameters to determine whether the height of the background box should be changed to fit the text, as well as whether the background should be anchored at the bottom when changing the height.

## How to test the changes?
Open the move overlay in the dex, check the summary during a run, etc.

To try out mystery encounters, you can use this override:
```
const overrides = {
  STARTING_WAVE_OVERRIDE: 11,
  STARTING_LEVEL_OVERRIDE: 9999,
```

## Checklist


Wlowscha:
Looks like https://github.com/pagefaultgames/pokerogue/pull/5925 is vey relevant too

src/ui/scrolling-text.ts
```diff
@@ -0,0 +1,110 @@
...
+This takes various coordinates:
+- The x and y coordinates relative to the parent container, this is typical behavior for Phaser.GameObjects.Container.
+- The width and height of the box; these are needed to create the background.
+The mask is not created right away (although this is possible in principle). Instead, we have a separate function,
+which takes as input the _global_ coordinates of scrolling text object. This is necessary to correctly position the mask in the scene.
+ */
+
+const BORDER = 8;
+
+export default class ScrollingText extends Phaser.GameObjects.Container {
```
Bertie690: No default class plz

Bertie690 (dismissed):

src/ui/scrolling-text.ts
Wlowscha:
```suggestion
export class ScrollingText extends Phaser.GameObjects.Container {
```

src/ui/containers/scrolling-text.ts
```diff
@@ -0,0 +1,110 @@
...
+  ) {
+    super(scene, x, y);
+
+    this.offsetX = hasBackground ? BORDER : 0;
+    this.offsetY = hasBackground ? BORDER - 2 : 0;
+
+    // Adding the background
+    this.descBg = addWindow(0, 0, width, height);
+    this.descBg.setOrigin(0, 0);
+    this.descBg.setVisible(hasBackground);
```
SirzBenjie:
```suggestion
    this.descBg = addWindow(0, 0, width, height).setOrigin(0, 0).setVisible(hasBackground);
```

src/ui/containers/scrolling-text.ts
```diff
@@ -0,0 +1,110 @@
...
+    const globalMaskX = globalX + this.offsetX;
+    const globalMaskY = globalY + this.offsetY;
+
+    const visibleWidth = this.descBg.width - (this.offsetX - 2) * 2;
+    this.maskHeight = (this.text.style.lineHeight / 6) * this.maxLineCount;
+    const visibleHeight = this.maskHeight;
+
+    const maskGraphics = scene.make.graphics({ x: 0, y: 0 });
+    maskGraphics.fillRect(globalMaskX, globalMaskY, visibleWidth, visibleHeight);
+    maskGraphics.setScale(6);
```
SirzBenjie: Can you also apply method chaining here (I would do a suggestion, but it wouldn't format properly with biome on web editor).

src/ui/containers/scrolling-text.ts
```diff
@@ -0,0 +1,110 @@
...
+import i18next from "i18next";
+import type BBCodeText from "phaser3-rex-plugins/plugins/bbcodetext";
+
+/*
+This takes various coordinates:
+- The x and y coordinates relative to the parent container, this is typical behavior for Phaser.GameObjects.Container.
+- The width and height of the box; these are needed to create the background.
+The mask is not created right away (although this is possible in principle). Instead, we have a separate function,
+which takes as input the _global_ coordinates of scrolling text object. This is necessary to correctly position the mask in the scene.
+ */
```
SirzBenjie: Should this be either a module comment or a doc comment on something?

Wlowscha: It can by anything you want it to be

fabske0: It should probably be a doc comment for the constructor. this is what it looks like rn

src/ui/containers/pokedex-info-overlay.ts
```diff
@@ -36,52 +32,32 @@ export class PokedexInfoOverlay extends Phaser.GameObjects.Container implements
 
   private readonly maskPointOriginX: number;
   private readonly maskPointOriginY: number;
```
fabske0: these are unused aside from `options`

fabske0 (changes requested):
The ME texts (descriptionand options) jump weirdly. Not always for the description. some Mes are fine

https://github.com/user-attachments/assets/cec46b7f-0bfd-4a0a-bc98-998c24699451

</details>

src/ui/handlers/summary-ui-handler.ts
```diff
@@ -1337,7 +1304,7 @@ export class SummaryUiHandler extends UiHandler {
 
     this.moveSelect = false;
     this.extraMoveRowContainer.setVisible(false);
-    this.moveDescriptionText.setText("");
+    this.moveDescription.text.setText("");
```
fabske0, line 1305: **Maybe** the scrollingText class could have a `setText()` method so there is no need for `.text.`

Very minor though

fabske0 (dismissed):

Wlowscha:
image

Now allows for resizing when text is shorter than the maximum number of lines

src/ui/containers/scrolling-text.ts
```diff
@@ -0,0 +1,205 @@
...
+}
+
+const BORDER = 8;
+
+export class ScrollingText extends Phaser.GameObjects.Container {
+  private readonly descBg: Phaser.GameObjects.NineSlice;
+  public text: BBCodeText;
+  private descScroll: Phaser.Tweens.Tween | null = null;
+  private readonly maxLineCount: number;
+  private anchorBottom: boolean;
```
fabske0: `anchorBottom` is unused

src/ui/containers/scrolling-text.ts
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * This method must be passed as input the global coordinates of the scrolling text object.
+   * It is done this way because of two reasons:
+   * 1) the mask must be positioned relative to the scene, not to a parent container;
+   * 2) there is no simple way to recover the global coordinates from inside the ScrollingText container.
+   * The latter would be much more desirable if possible.
+   *
+   * @param globalX
+   * @param globalY
+   */
```
fabske0, line 95: Param description would be nice

src/ui/containers/scrolling-text.ts
```diff
@@ -0,0 +1,205 @@
...
+   * Method to update the text after the scrolling object and mask have been created.
+   * if `fitHeight` is `true`, the background box is automatically adjusted to fit the text
+   * (only if the text is shorter than the maximum number of lines).
+   * When resizing, by default the top of the box is fixed. To change this behavior, and keep
+   * the bottom fixed, set `anchorBottom` to `true` instead.
+   *
+   * @param content
+   * @param fitHeight
+   * @param anchorBottom
+   */
```
fabske0, line 131: Param description would be nice

fabske0 (dismissed):
minor nits

fabske0 (approved):

PR #6061, opened 2025-07-02
[Dev] Remove warnings from `pnpm` interactions

SirzBenjie:
None

I'm tired of the spurious warnings emitted by pnpm

1. Removed `dependency-graph.js` which is completely unnecessary. The dependency graph for the files in our repository isn't useful, as there are so many files, that looking at it visually doesn't do much.
   - This allowed me to remove the `@hpcc-js/wasm` dev dependency, which itself emits warnings. See https://github.com/hpcc-systems/hpcc-js-wasm/issues/308
2. Changed the `postinstall` in `package.json` to directly update the submodule instead of invoking lefthook's post-merge command. There's no reason to use post-merge here, anyway
   - Some idiosyncracies in how `lefthook run` works require an invocation to manually press `enter` in order for it to continue. See https://github.com/evilmartians/lefthook/discussions/1060
3. Added `pnpm-workspace, and whitelisted the packages that have `postinstall` hooks themselves: msw, lefthook, and esbuild
   - This prevents the warning that mentions that `pnpm approve-builds` needs to be run. We actually *want* these `postinstall` scripts to run, as they setup the packages for proper usage. (e.g. lefthook's will install the hooks).

## Screenshots/Videos
N/A

## How to test the changes?
N/A

## Checklist
- ~~[ ] Are all unit tests still passing? (`pnpm test:silent`)~~


package.json
```diff
@@ -22,14 +22,13 @@
     "docs": "typedoc",
     "depcruise": "depcruise src test",
     "depcruise:graph": "depcruise src --output-type dot | node dependency-graph.js > dependency-graph.svg",
-    "postinstall": "lefthook install && lefthook run post-merge",
+    "postinstall": "lefthook install; git submodule update --init --recursive",
```
DayKev, line 24: That doesn't work.
image

DayKev:
```suggestion
    "postinstall": "lefthook install && git submodule update --init --recursive",
```
I think it still needs to be `&&`.

SirzBenjie: Looks to be an OS difference..

SirzBenjie: We don't _actually_ want it to be `&&` though. `lefthook` can fail to install and we still want to do the other thing.

SirzBenjie: Well there's not a cross-platform way to run two commands.. w/e

DayKev: > We don't _actually_ want it to be `&&` though. `lefthook` can fail to install and we still want to do the other thing.

Should it be the other way around then? `git submodule ... && lefthook install`?

DayKev (dismissed):

package.json
```diff
@@ -22,14 +22,13 @@
     "docs": "typedoc",
     "depcruise": "depcruise src test",
     "depcruise:graph": "depcruise src --output-type dot | node dependency-graph.js > dependency-graph.svg",
```
DayKev:
```suggestion
```
Almost forgot, we should remove the command that generates the graph too.

DayKev:
Now that this caused me to think about the wasm error, I think I know what it is: https://github.com/hpcc-systems/hpcc-js-wasm/issues/308#issuecomment-3032716588

Though it doesn't matter now because no one ever used the depcruiser graph script anyway and so we should still remove it.

DayKev (approved):

package.json
DayKev: Found a solution.

damocleas (approved):

PR #6067, opened 2025-07-06
[Bug] Fix anger point procing on every hit if first hit in multi hit was a crit

SirzBenjie:
Fixes 2 bugs related to anger point:

1. Anger point will no longer constantly emit its application message when the first strike of a multi-hit move is a critical hit
2. Anger point will now correctly maximize the user's attack (previously only added +6)

Fixes a bug related to anger point: https://discord.com/channels/1125469663833370665/1370628941060243466/1370628941060243466.

Made `PostDefendCritStatStageAbAttr` no longer subclass `PostDefendAbAttr` so that it could be applied separately from the other `PostDefend` abilities in the move effect phase, particularly, conditioned on the attack being a critical hit.
This allowed me to remove the janky check that was going on in the ability.

I also adjusted the ability attribute to give a +12 boost (from +6).

Also added tests for anger point.

## Screenshots/Videos
Trigger circumstances are too annoying to force. Also why I haven't manually tested.

## How to test the changes?
``pnpm test:silent tests/abilities/anger-point.test.ts``

## Checklist

### DIFF
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -1198,12 +1198,12 @@ export class PostDefendContactApplyTagChanceAbAttr extends PostDefendAbAttr {
  * Set stat stages when the user gets hit by a critical hit
  *
  * @privateremarks
- * It is the responsibility of the caller to ensure that this ability is only applied
+ * It is the responsibility of the caller to ensure that this ability attribute is only applied
  * when the user has been hit by a critical hit; such an event is not checked here.
  *
  * @sealed
  */
-export class PostDefendCritStatStageChangeAbAttr extends AbAttr {
+export class PostReceiveCritStatStageChangeAbAttr extends AbAttr {
   private stat: BattleStat;
   private stages: number;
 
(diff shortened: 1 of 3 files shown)


src/data/abilities/ability.ts
```diff
@@ -1194,7 +1194,15 @@ export class PostDefendContactApplyTagChanceAbAttr extends PostDefendAbAttr {
...
 
-export class PostDefendCritStatStageChangeAbAttr extends PostDefendAbAttr {
+/**
+ * Set stat stages when the user gets hit by a critical hit
+ *
+ * @privateremarks
+ * It is the responsibility of the caller to ensure that this ability is only applied
+ * when the user has been hit by a critical hit; such an event is not checked here.
+ *
+ * @sealed */
```
```suggestion
 * @sealed
 */
```

DayKev (approved):

src/data/abilities/ability.ts
```diff
@@ -1194,7 +1194,15 @@ export class PostDefendContactApplyTagChanceAbAttr extends PostDefendAbAttr {
...
+export class PostDefendCritStatStageChangeAbAttr extends AbAttr {
```
emdeann:
```suggestion
export class PostCritStatStageChangeAbAttr extends AbAttr {
```

I don't like this being named "PostDefend" if it's not actually subclassing `PostDefendAbAttr`

SirzBenjie: Well, it should still be discernable that it's for crits done to the user, not by the user.

DayKev (approved):

PR #6080, opened 2025-07-11
[Bug][UI/UX][Refactor] Fix empty movesets related to starter forms and general SSUI cleanup

### LINKED ISSUE #3937: [Bug] Form Changing Pokémon that need two pokémon cannot swap current moves with other current moves
### Describe the bug

The bug consists of three pokémon which need a secondary pokémon to change forms... 
The first instance reported was in this bug report which reported Calyrex with a unusual behaviour I looked into further: https://discord.com/channels/1125469663833370665/1279222639092563989
The bug affects currently to my knowledge Calyrex, Necrozma and Kyurem.
It prevents the user from rearraging the moves that were already on like any other pokémon by putting the opposite move you want to arrange in a certain way, it doesn't seem to affect any moves not on the pokémon being egg or normal ones in the non set moves... 

### Session export file

Not needed, its in the main server and easy to replicate

### User data export file

_No response_

### Expected behavior

It should let the moves legit swap between each other.

### Screenshots / Videos

https://github.com/user-attachments/assets/266bd168-3f8a-4de3-af60-10779a12cb17

PC/Windows

### If other please specify

_No response_

### Which browser do you use?

Chrome

### If other please specify

_No response_

### Additional context

_No response_

### LINKED ISSUE #6037: [Bug] Pokemon with Forms that have Specific Starter Movesets can Begin a Run with Empty Movesets, ~~Exiting Corrupts Saves Online~~
### Describe the bug

Indeedee, Rotom, and Pikachu can begin a run with no moves in their moveset. Reloading or refreshing on the ~~Online or Beta Website corrupts the save~~ (This is fixed by #6070) , requiring a new session be imported to be able to access that save slot again.

The Form-Moveset Pokemon above can have their Movesets enter an unmodifiable state. This is very important and required for triggering this bug.

- This does not require the Pokemon to already be in the Party.
- This does not occur when simply changing forms, the moveset must fail to be modified.
- Seems to work with forms at random, Indeedee is somewhat consistent. It probably isn't random.
- Egg moves may or may not be unlocked.

If I had to speculate, due to the error message in console when attempting to CONTINUE corrupted Loading... Runs, there's some sort of desync with forms and their movesets.

Link to Discord's Bug Report
https://discord.com/channels/1125469663833370665/1282211336750764066
 

### Reproduction

Reproduced with Rotom and Indeedee, Cos-Chu 

Might require a slow Computer. Unlock Indeedee M and F, Unlock Rotom and Rotom Forms. **DO NOT DO THIS ON THE MAIN WEBSITE**
- This seems to occur at random, but I might be mistaken.

1. Swap to a Form from Base form
2. Swap the positions of moves in the moveset. (This can only be done once per form)
3. Add the form to your Party 
4. Swap back to Base form
5. Begin a Run without having removed the Pokemon from your Party between steps 3-5

### Expected behavior

1. Pokemon should never enter a run with no moves
2. Pokemon Forms should be able to have the positions of their moves modified freely
3. Corruption should not occur for saves where the Pokemon has no moves.

### Screenshots / Videos

<details><summary>Save Corruption</summary>
<p>
Corrupted Save

Image

Save and Quit on the Beta Website

https://github.com/user-attachments/assets/d4af250d-7b35-4565-9a0d-3e900dbf5ff4

Save and Quit on Local does not corrupt the save

https://github.com/user-attachments/assets/6e75d159-ae4d-45ed-91d9-c5bc025ecd2d

</p>
</details> 

<details><summary>Execution</summary>
Rotom

https://github.com/user-attachments/assets/75595682-e5f5-457c-bec1-86c143d0b32f

Indeedee

https://github.com/user-attachments/assets/5d63f86a-0481-41fc-a737-a6a475ab1f00

</details> 

### Session export file

_No response_

### User data export file

Everything but also has Egg moves

[full_unlocks_sane.txt](https://github.com/user-attachments/files/20934103/full_unlocks_sane.txt)

### Additional context

Indeedee has a small bug where the form doesn't completely reset to Male upon leaving it with the cursor as female, requiring Form to be pressed twice. This is likely unrelated as Cos Chu and Rotom don't have that problem.

[Bug] Form Changing Pokémon that need two pokémon cannot swap current moves with other current moves #3937
Certain Pokemon with forms brick their movesets in a similar way, though do not gen without a moveset.

Necrozma had the issue where it would generate without a moveset upon Run start, but that has been fixed.

Report
https://discord.com/channels/1125469663833370665/1274446254428127254
Commit that fixed the issue
https://github.com/pagefaultgames/pokerogue/pull/3594/files

Overriding Species also spawns Pokemon with no Moves, but cannot be done on online.

SirzBenjie:
Changing forms after swapping a moveset and then starting the game will no longer cause that starter to have an empty moveset (#6037).

Pokemon with formes requiring multiple pokemon (Calyrex, Kyurem, Necrozma, and Hoopa) can now swap the positions of two moves that are in the moveset (#3937).

Fixes #3937 and the rest of what is needed for fixes #6037
Also, giving the SSUI the same treatment I have been giving other files (minor cleanup this time, the rest is too much of a behemoth).

- Code cleanup of SSUI (the "minor refactor" part)
   - Use Phaser method chaning (i.e. replacing `var.method1(); var.method2()` with `obj.method1().method2();`)
   - Condense multiple calls to `Container.add(child)` into a single `Container.add([child1, child2])`
   - Misc cleanup (like using `x ||= y` instead of `x = x || y`
   - Shorten `object.setOrigin(A, A)` to `object.setOrigin(A)`, which behaves the same and uses fewer characters (and is best practice in Phaser)
   - I wanted to use `Phaser.GameObject.Group` to collect elements that are always shown/hidden together, but our testing infrastructure does not have mocks for `globalScene.add.group()` and thus breaks down if we add this. I instead inserted comments in the code with how this would be done with TODO markers.
   - ~~Resisted the urge to burn the whole file to the ground and start from scratch~~
- Ported https://github.com/Despair-Games/poketernity/pull/724, which contains fixes for both issues. See its description.

## Screenshots/Videos
TODO?

## How to test the changes?
See #6037 for reproduction steps.
Also, open up the starter-select-ui (new game) and play around with options to ensure everything is working the same as it always has.

## Checklist

## Additional Notes
I had to be disciplined with this PR. The file is in desperate need of a much closer look.
One thing that really needs to be done is that all of the objects and sprites should be completely unloaded when the starter select ui is closed. Currently, they are lingering around in memory.
We also need to make it so that instead of displaying a container for every single starter pokemon, only 81 containers are displayed at a time, like what @Wlowscha did for the Pokedex.

The `processInput` method needs to be split up and simplified, as it is right now an unmaintainable, unreadable mess.

### DIFF
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -335,6 +335,10 @@ export class StarterSelectUiHandler extends MessageUiHandler {
   private natureLabel: Phaser.GameObjects.Text;
   private teraLabel: Phaser.GameObjects.Text;
   private goFilterLabel: Phaser.GameObjects.Text;
+  /** Group holding the UI elements appearing in the instructionsContainer */
+  /* TODO: Uncomment this once our testing infra supports mocks of `Phaser.GameObject.Group` 
+  private instructionElemGroup: Phaser.GameObjects.Group;
+  */
 
   private starterSelectMessageBox: Phaser.GameObjects.NineSlice;
   private starterSelectMessageBoxContainer: Phaser.GameObjects.Container;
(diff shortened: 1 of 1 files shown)


src/ui/starter-select-ui-handler.ts
```diff
@@ -992,192 +882,143 @@ export default class StarterSelectUiHandler extends MessageUiHandler {
...
     const instructionTextSize = textSettings.instructionTextSize;
 
-    this.instructionsContainer = globalScene.add.container(4, 156);
-    this.instructionsContainer.setVisible(true);
-    this.starterSelectContainer.add(this.instructionsContainer);
+    this.instructionsContainer = globalScene.add.container(4, 156).setVisible(true);
+
+    const iRowX = this.instructionRowX;
+    const iRowY = this.instructionRowY;
+    const iRowXText = iRowX + this.instructionRowTextOffset;
```
Wlowscha: Minor pet peeve but I would call this iRowTextX since it's still a coordinate.

src/ui/starter-select-ui-handler.ts
```diff
@@ -3845,19 +3782,22 @@ export class StarterSelectUiHandler extends MessageUiHandler {
         const assetLoadCancelled = new BooleanHolder(false);
         this.assetLoadCancelled = assetLoadCancelled;
 
+        female ??= false;
         if (shouldUpdateSprite) {
-          species.loadAssets(female!, formIndex, shiny, variant, true).then(() => {
+          species.loadAssets(female, formIndex, shiny, variant, true).then(() => {
             // TODO: is this bang correct?
```
DayKev:
```suggestion
```
Looks like Biome formatted this comment below where it was supposed to be. Since you removed the `!` it was supposed to be applied to, it can be removed now.

DayKev (approved):
Tested and the bug doesn't occur anymore in the scenario I was able to trigger it before. Also fixed editing of form movesets (previously when I triggered the bug with Rotom's Base + Mow forms, Rotom Mow's moveset became un-editable).

PR #6084, opened 2025-07-11
[Balance] Encounter condition adjustments for Legendaries and Random Trainers

damocleas:
Random Trainers WERE blocked 3 waves before and after set trainer battles, this does not include Gym Leaders.
    **This is now only 2 waves.**

Before this change, it looked like this:
image
After, the chart looks like this.
image
Credits to @Jimmybald1 and thank them for the help for the changes in this PR / testing of some of them.

## Random Trainer Encounter Chances changed for many biomes (1/X is the chance of a trainer battle each wave)
- Metropolis: 1/2 -> 1/4
- Slum: 1/4 -> 1/6
- Construction Site: 1/4 -> 1/6
- Beach: 1/4 -> 1/6
- Desert: 1/8 -> 1/6
- Sea: 1/8 -> 1/12
- Graveyard: 1/8 -> 1/12
- Island: 0 -> 1/12
- Seabed: 1/16 -> 0

Didn't want to save some of these for Biome Rework. Might as well get them out of the way now while I'm doing this (mostly wanted to adjust Metropolis with this set battle thing so the start of runs isn't hitting two trainers often before Rival 2).

## Encounter Waves for Legendaries, Sub-Legendaries, and Mythicals
This has been adjusted in a notable way
- Removed the 720+ BST check, if we need it again in the future we can, but this would only apply to Arceus and some forms, that are all impossible to encounter in the wild, and have no plans to. _If we need it again for some sort of randomizer mode, we can simply add it back._
- 670+ BST check has been changed to a 660+ BST check, to account for Kyurem, Zacian, and Zamazenta, it has also been moved from wave 70 to 80
- 580+ BST check has been changed to an *all* BST check for Legends/Sub-Legends/Mythicals, to account for the *many* trios and quartets that are available, and Terapagos and the others that have early stages. It has also been moved from wave 50 to 55, so it is after the third rival battle. This roughly translates to wave 25 in Daily Mode.

To account for the above changes, Legendaries that can evolve (Type:Null, Poipole, and Kubfu) have had their Evolution Delays set to "Very Long" and set to always evolve in the wild upon hitting level 60 or higher, so there's ample opportunity to encounter them pre-evolved and have them count towards your "Seen" stat.

I have also adjusted Eternal Floette to match Bloodmoon Ursaluna in biome rarity, since it was classified as a sub-legend a while ago with them but its' rarity did not match this.

Meltan and Cosmog may be looked at another time, but they require changes outside of the scope of this PR.

I felt like improving the game.
Had a good discussion in the daily run channel over on the discord and thought these would be positive changes for the game, it removes the random Guzzlords and such too early.

Some of these changes are preliminary changes for the Biome Rework as well, and fixing small oversights like Island not having random trainers or floette not being adjusted.

The wave check for Legendaries being able to be encountered has been updated to use the "`getWaveForDifficulty`" check

## How to test the changes?
Overrides.

## Checklist
- [X] Are all unit tests still passing? (`pnpm test:silent`)

~~- [ ] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?~~
~~- [ ] Has the translation team been contacted for proofreading/translation?~~


Mourouh (approved):

src/field/arena.ts
```diff
@@ -168,19 +168,14 @@ export class Arena {
 
       ret = getPokemonSpecies(species!);
 
-      if (ret.subLegendary || ret.legendary || ret.mythical) {
+      if (ret.subLegendary || ret.legendary || ret.mythical || ret) {
+        const waveDifficulty = globalScene.gameMode.getWaveForDifficulty(waveIndex);
         switch (true) {
```
Fontbane: The opportunity should be taken to remove this `switch (true)` and just use a normal if statement

src/field/arena.ts
```diff
@@ -168,19 +168,14 @@ export class Arena {
 
       ret = getPokemonSpecies(species!);
 
-      if (ret.subLegendary || ret.legendary || ret.mythical) {
+      if (ret.subLegendary || ret.legendary || ret.mythical || ret) {
```
Fontbane:
```suggestion
      if (ret.subLegendary || ret.legendary || ret.mythical) {
```
Putting `ret` at the end there makes the other 3 redundant and makes it so _every_ species regens before wave 55 (until it gives up on the 10th roll). Not sure why Biome didn't complain about it.
image

src/field/arena.ts
damocleas: I'm just doing the balance part man e

DayKev: Changed it.

DayKev (dismissed):

src/game-mode.ts
```diff
@@ -164,14 +164,14 @@ export class GameMode implements GameModeConfig {
...
-       * after a X0 full party heal
+       * after a X0 full party heal, this also allows for a smoother biome transition for general gameplay feel
        */
       const trainerChance = arena.getTrainerChance();
       let allowTrainerBattle = true;
       if (trainerChance) {
         const waveBase = Math.floor(waveIndex / 10) * 10;
-        // Stop generic trainers from spawning in within 3 waves of a trainer battle
-        for (let w = Math.max(waveIndex - 3, waveBase + 2); w <= Math.min(waveIndex + 3, waveBase + 9); w++) {
+        // Stop generic trainers from spawning in within 2 waves of a trainer battle
```
Fontbane:
```suggestion
        // Stop generic trainers from spawning in within 2 waves of a fixed trainer battle
```
I'd also like to note this block of code (starting at 164) is kinda wack and should probably get refactored soon

damocleas: dude that could apply to most of this file lmao

damocleas:
making some small changes tonight or tomorrow, will post in the callout thread again when done

src/game-mode.ts
```diff
@@ -164,14 +164,14 @@ export class GameMode implements GameModeConfig {
     if (waveIndex % 10 !== 1 && waveIndex % 10) {
```
Fontbane, line 164:
```suggestion
    if (waveIndex % 10 > 1) {
```
Like what is this lmao

DayKev (approved):

Blitz425 (approved):

PR #6091, opened 2025-07-13
Replace remaining Modifiers with Rewards

Wlowscha:
This PR is contributing to the modifier rework by completing the conversion from "consumables" to "rewards".

These are the main points:
- `ModifierType` has been removed and replaced by `Reward`. The surrounding structure is mostly untouched---so we have `RewardGenerator`, which creates Reward instances from a certain group; `WeightedReward`, which is a Reward with an attached weight or weight function to be used in pools; and `RewardOption`, which is displayed during the select reward phase at the end of each encounter.
- Many instances of `Modifier` have already been replaced by HeldItem and TrainerItem. Now, the rest have been converted too. Rather than creating a "Consumable" class, their properties have been incorporated into the respective rewards. (e.g. previously, we had an AddPokeballModifierType which would generate an AddPokeballModifier with an .apply() method; now, the .apply() method is directly on the AddPokeballReward).
- A bunch of functionality that Modifiers and ModifierTypes used to have, necessary for persistent modifiers, has been removed. This is because rewards are always temporary, and all persistent modifiers are taken care of as held items and trainer items.
- The `tier` property has been moved from Reward to RewardOption. This may be changed again---in fact, it is questionable whether we need RewardOption at all.
- What were previously called `ModifierTier` are now `RarityTier` (since they can apply to rewards or items in different contests).

I have also significantly pruned various functions dealing with the reward pool and rarity tiers. Some of the pruning is probably excessive.

The most significant files are:
 - reward.ts,
 - select-reward-phase.ts (which has tests!),
 - battle-scene.ts,
 - init-reward-pool.ts.

Things to be careful about:
- [ ] Ensure that every Reward is doing the correct thing, in particular the apply() and .shouldApply() methods.
- [ ] Fix typing of Reward subclasses, scope of variables, and especially typing of the .apply() parameters.
- [ ] Go through RewardSelectPhase and see if it can be simplified.
- [ ] Do we need `applyReward()` to be a method of BattleScene?
- [ ] Ideally, have tests for each Reward.
- [ ] Ensure that reward generation from the pool is working as expected.


src/items/reward.ts
```diff
@@ -37,71 +41,49 @@ import {
...
-  }
-
-  constructor(
-    localeKey: string | null,
-    iconImage: string | null,
-    newModifierFunc: NewModifierFunc | null,
-    group?: string,
-    soundName?: string,
-  ) {
+  constructor(localeKey: string | null, iconImage: string | null, group?: string, soundName?: string) {
```
Fontbane, line 107:
```suggestion
  constructor(id: RewardId, localeKey: string | null, iconImage: string | null, group?: string, soundName?: string) {
```

Wouldn't it make sense to just pass the RewardIds to the constructor rather than every subclass handling it themselves?

src/items/reward.ts
```diff
@@ -234,72 +150,180 @@ export class AddPokeballModifierType extends ModifierType {
...
+   */
+  apply(): boolean {
+    const moneyAmount = new NumberHolder(globalScene.getWaveMoneyAmount(this.moneyMultiplier));
+
+    globalScene.applyPlayerItems(TrainerItemEffect.MONEY_MULTIPLIER, { numberHolder: moneyAmount });
+
+    globalScene.addMoney(moneyAmount.value);
+
+    for (const p of globalScene.getPlayerParty()) {
+      if (p.species?.speciesId === SpeciesId.GIMMIGHOUL || p.fusionSpecies?.speciesId === SpeciesId.GIMMIGHOUL) {
```
Fontbane:
```suggestion
      if (p.hasSpecies(SpeciesId.GIMMIGHOUL)) {
```

src/items/reward.ts
```diff
@@ -353,20 +379,23 @@ export class TrainerItemReward extends ModifierType {
...
+  }
+
   apply() {
     globalScene.trainerItems.add(this.itemId, allTrainerItems[this.itemId].getMaxStackCount());
-    console.log("WE GOT HERE WE ADDED IT");
   }
 }
 
-export class TerastallizeModifierType extends PokemonModifierType {
+export class TerastallizeReward extends PokemonReward {
```
Fontbane: `ChangeTeraTypeReward` or something like that might be more accurate

src/items/reward.ts
```diff
@@ -384,50 +413,89 @@ export class TerastallizeModifierType extends PokemonModifierType {
...
+  /**
+   * Checks if {@linkcode TerrastalizeConsumable} should be applied
+   * @param playerPokemon The {@linkcode PlayerPokemon} that consumes the item
+   * @returns `true` if the {@linkcode TerrastalizeConsumable} should be applied
+   */
+  shouldApply({ pokemon }: PokemonRewardParams): boolean {
+    return (
+      [pokemon?.species.speciesId, pokemon?.fusionSpecies?.speciesId].filter(
+        s => s === SpeciesId.TERAPAGOS || s === SpeciesId.OGERPON || s === SpeciesId.SHEDINJA,
+      ).length === 0
```
Fontbane:
```suggestion
      pokemon.teraType !== this.teraType &&
      ![SpeciesId.SHEDINJA, SpeciesId.OGERPON, SpeciesId.TERAPAGOS].some(s => hasSpecies(s))
```

Doesn't have to be changed now but I want to point it out as something that should be changed

SirzBenjie (approved):

PR #6099, opened 2025-07-15
[Bug] Fix return in applyChosenModifier; prevent race condition in modifier-ui-handler

SirzBenjie:
Game will no longer crash when spamming `A` before select modifier phase starts.
Slightly more streamlined animations in the select-modifier phase.

Fixes a crash in select-modifier screen. See https://discord.com/channels/1125469663833370665/1385851080449069177/1385851080449069177

#5993 introduced a fix to have modifier selections pass `-1` for the cost instead of `0`, as other parts of the code relied on cost being -1 to determine whether a reward selected in the modifier phase was a purchase or a reward.

However, this PR missed updating the `applyChosenModifier` method to recognize this. The method's signature was previously using `return !cost`. It needed to have been updated to return a comparison against -1.

<s>Now I'm not entirely sure how this faulty return value ends up causing the phase to crash in the way it does. However, it is a logic bug, and this fix *does* stop the crash.
I suspect it has something to do with the return value flowing into the return value of  `modifierSelectCallback` inside of the `start` method.
Regardless, this fixes it.</s>

The above made it more difficult to trigger the crash (I think?) but did not ultimately stop it. Our UI logic is so 🍝 that it's difficult to answer the question as to _why_ the above has anything to do with the crash.

This crash occurs because the modifier select screen is accepting input before all of the assets are loaded and the cursor has had a chance to update its position to the right spot (see `updateCursorTarget` in the `show` method of `modifier-select-ui-handler`). Because the UI was accepting input before the cursor was set to the right thing, and each cursor defaults to 0 (so row is 0, column is 0, aka the reroll button), holding down `Button.ACTION` (aka `a`, or `enter`) would invoke the reroll logic.
The reroll callback would try to modify sprites that haven't been loaded (because they weren't finished!), which caused the crash.

To stop this, I hooked into the already-existing `awaitingActionInput`.  At the start of the `show` method, we set this to `false.`
Only _after_ all of the animations have been completed (meaning everything is loaded), do we allow this value to be set back to true.

I also fixed up the logic around the animation sequence inside the show method to use promises rather than estimating how long things take and using a `globalScene.time.delayedCall()`. For example, the animation for the shop icons to appear starts 
was previously being delayed like so:
```ts
 globalScene.time.delayedCall(1000 + maxUpgradeCount * 2000, { /* snip */ }
```
In other words, it was estimating roughly 2s for each upgrade count that could happen (upgrade count being the number of times a reward would upgrade to the next reward).  It then added a flat 1000ms. All of this delay would be impacted by game speed.

Instead of this, I kept around the delay of `1000`, but have this call started after the tween has completed (by creating a promise that is resolved when the earlier animation has completed).

## Screenshots/Videos
Kinda hard to demonstrate.

## How to test the changes?
Start a run with a mon, probably using some overrides to set to a high level, and hold A throughout the screen. Observe whether game crashes.

## Checklist


emdeann:
I am still able to reproduce the crash on this branch

SirzBenjie:
> I am still able to reproduce the crash on this branch

That's odd... it fixes it for me? Let me try a bunch more...

SirzBenjie:
> I am still able to reproduce the crash on this branch

Found the underlying issue and fixed it.

src/ui/modifier-select-ui-handler.ts
```diff
@@ -269,13 +272,18 @@ export class ModifierSelectUiHandler extends AwaitableUiHandler {
     globalScene.updateBiomeWaveText();
     globalScene.updateMoneyText();
 
+    this.awaitingActionInput = false;
```
DayKev, line 274: This got moved but the comment didn't.

DayKev: Other than that, 👍.

src/ui/modifier-select-ui-handler.ts
```diff
@@ -269,13 +272,18 @@ export class ModifierSelectUiHandler extends AwaitableUiHandler {
...
     let i = 0;
 
+    // TODO: Replace with `Promise.withResolvers` when possible.
```
DayKev, line 276: Btw I think Vite plugins might be the proper way to use polyfills. Not sure how to do it though.

PR #6100, opened 2025-07-15
[AI] Minor fixup of matchup score

SirzBenjie:
Smarter enemy AI when considering switches

It was pointed out to me that the prior PR, #5981 had a few oversights in its logic. This addresses such oversights.

- If the enemy is tera'd, its original types as well as its tera type are considered, instead of just the tera type (by changing the `forDefend` parameter to `false`)
- The pokemon's offensive types are no longer counted in its attack score, as the procedure added by #5981 which considers the attack score on a per-move basis
- Ensure that `hpDiffRatio` caps at 1, as its comment indicates, by moving the `Math.min()` calculations to the return statement

Added comments discussing some of the deficiencies in the logic.
## Screenshots/Videos
N/A

## How to test the changes?
Play waves?

## Checklist

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -2562,8 +2562,8 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     // Get average attack score of all damaging moves (|| 1 prevents division by zero))
     // TODO: Averaging the attack score is excessively simplistic, and doesn't reflect the AI's move selection logic
     // e.g. if the mon has one 4x effective move and three 0.5x effective moves, this score would be ~1.375
-    // instead of
-    // That is, this atkScore logic does not reflect the AI's move selection logic.
+    // which does not seem fair, given that if the AI were to switch, in all likelihood it would use the 4x move.
+    // We could consider a weighted average...
     atkScore /= moveAtkScoreLength || 1;
     /**
      * Based on this Pokemon's HP ratio compared to that of the opponent.
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -2560,6 +2560,10 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       moveAtkScoreLength++;
     }
     // Get average attack score of all damaging moves (|| 1 prevents division by zero))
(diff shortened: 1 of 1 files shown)


Madmadness65 (dismissed):

src/field/pokemon.ts
```diff
@@ -2520,48 +2520,61 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-    }
     if (enemyTypes.length > 1) {
       defScore *=
         1 / Math.max(this.getAttackTypeEffectiveness(enemyTypes[1], opponent, false, false, undefined, true), 0.25);
     }
-    atkScore *= 1.25; //give more value for the pokemon's typing
+
     const moveset = this.moveset;
+    // Start at 1 to prevent division by zero later on
     let moveAtkScoreLength = 0;
```
DayKev: This is starting at `0` though?

src/field/pokemon.ts
```diff
@@ -2520,48 +2520,61 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      // NOTE: Counter and Mirror Coat are considered as attack moves here
+      if (resolvedMove.category === MoveCategory.STATUS || move.getPpRatio() <= 0) {
+        continue;
       }
+      const moveType = resolvedMove.type;
+      let thisScore = opponent.getAttackTypeEffectiveness(moveType, this, false, true, undefined, true);
+
+      // Add STAB multiplier for attack type effectiveness.
+      // For now, simply don't apply STAB to moves that may change type
+      if (this.getTypes(true).includes(moveType) && move.getMove().hasAttr("VariableMoveTypeAttr")) {
```
DayKev: Shouldn't it be this then?
```suggestion
      // For now, simply don't apply STAB to moves that may change type
      if (this.getTypes(true).includes(moveType) && !move.getMove().hasAttr("VariableMoveTypeAttr")) {
```

SirzBenjie: Whoops, yes, thanks.

DayKev (approved):

damocleas (approved):

Madmadness65 (approved):

PR #6103, opened 2025-07-15
[Dev] Add support for manually rolling polyfills. Add polyfill for `promise.withResolvers`

SirzBenjie:
N/A

I have been wanting to use `Promise.withResolvers` for a while now to simplify some logic. However, it is a feature new to ES2026, which is not yet considered "Baseline - Widely available".

I did a lot of digging into best practices for polyfills. I considered vite plugins, `core-js`, and `es-shims`.

Ultimately, adding dependencies (which aren't even typed!) via `es-shims` did not seem to be the best option; we'd be adding a dependency for each new polyfill we want. `es-shims` also makes larger assumptions.

The polyfills in `core-js` are too heavy, adding too much code.

[Cloudflare's polyfills](https://cdnjs.cloudflare.com/polyfill/) do not yet have es2024, which is what introduced Promise.withResolvers.

I don't know much about polyfills. One of the benefits of using `es-shims` is that it would hopefully be updated if the spec or requirements of the polyfill were to change (we'd just have to bump the polyfill version).
However, if we go that route, then we'd need to add a new dependency for each polyfill into our `package.json` file. This really seems like a headache.

I'm open to a larger discussion about the best approach here; I'm not as comfortable with best practices for polyfills.

However, I would really like to have `Promise.withResolvers` in, as it would make quite a bit of code more ergonomic.

- Created a new file, `polyfills.ts`. for shimming our own polyfills
- Added `typings/polyfills.d.ts` to provide global typescript definitions for these polyfills
- Added an import to this file on the top of `main.ts`.

Note: Only `main.ts` needs to import `polyfills`. No other file should.

## Screenshots/Videos

https://github.com/user-attachments/assets/17ec5ddb-cfbb-4a01-bbc7-fd3305770da5

## How to test the changes?
To ensure that the polyfill works properly, you'll need a browser that does not yet support `Promise.withResolvers`. You can see the version that it was added for each browser [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers#browser_compatibility)

I personally downloaded the last firefox version 120 (support for Promise.withResolvers was added in 121): https://ftp.mozilla.org/pub/firefox/releases/120.0b9/win64/en-US/

Then, I added this dummy code to the beginning of `startGame` inside `main.ts`

```ts
const startGame = async (manifest?: any) => {
  const { promise, resolve, reject } = Promise.withResolvers<void>();
  // Fake a promise with resolvers
  if (confirm("Allow promise to resolve?")) {
    resolve();
  } else {
    reject(new Error("Promise rejected by user"));
  }
```

You can see in the video that the polyfill works as expected.

## Checklist


src/main.ts
```diff
@@ -6,6 +6,7 @@ import BBCodeTextPlugin from "phaser3-rex-plugins/plugins/bbcodetext-plugin";
 import InputTextPlugin from "phaser3-rex-plugins/plugins/inputtext-plugin";
 import TransitionImagePackPlugin from "phaser3-rex-plugins/templates/transitionimagepack/transitionimagepack-plugin";
 import UIPlugin from "phaser3-rex-plugins/templates/ui/ui-plugin";
+import "polyfills";
```
DayKev: Shouldn't this be:
```suggestion
import "#app/polyfills";
```
(also will require running Biome to re-sort the imports)

src/typings/polyfills.d.ts
DayKev, line 1: You can also modify "lib" in `tsconfig.json` instead of manually adding the `.d.ts` entries, e.g.:
```jsonc
    // Modifying this option requires all values to be set manually because the defaults get overridden
    // Values other than "ES2024.Promise" taken from https://github.com/microsoft/TypeScript/blob/main/src/lib/es2020.full.d.ts
    "lib": ["ES2020", "ES2024.Promise", "DOM", "DOM.AsyncIterable", "DOM.Iterable", "ScriptHost", "WebWorker.ImportScripts"],
```

src/main.ts
SirzBenjie: Good catch, thanks.

src/typings/polyfills.d.ts
SirzBenjie: I only shimmed `Promise.withResolvers`. If I do the above, it'll include more than what I polyfilled.

DayKev: `Promise.withResolvers` is the only functionality in `ES2024.Promise` (see https://github.com/microsoft/TypeScript/blob/65cb4bd2d52cd882f2c3a503681479eb2ed291ca/src/lib/es2024.promise.d.ts).

src/main.ts
SirzBenjie: Realized that polyfills should be the first possible thing imported, as they intentionally have side effects.

DayKev: Oh yeah I forgot about that.

tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "target": "ES2020",
     "module": "ES2020",
+    "lib": ["ES2020", "ES2024.Promise"],
```
DayKev:
```suggestion
    // Modifying this option requires all values to be set manually because the defaults get overridden
    // Values other than "ES2024.Promise" taken from https://github.com/microsoft/TypeScript/blob/main/src/lib/es2020.full.d.ts
    "lib": ["ES2020", "ES2024.Promise", "DOM", "DOM.AsyncIterable", "DOM.Iterable", "ScriptHost", "WebWorker.ImportScripts"],
```
You have to re-add all of the default libs because this overwrites the whole thing rather than being additive.

src/main.ts
```diff
@@ -1,3 +1,6 @@
+import "app/#polyfills";
```
DayKev:
```suggestion
import "#app/polyfills";
```
Typo.

src/polyfills.ts
```diff
@@ -0,0 +1,19 @@
+/*
+Manual rolling of polyfills desired by the project.
+
+IMPORTANT: When adding / removing polyfills, make sure to edit `src/typings/polyfills.d.ts`
+accordingly, so that the TypeScript compiler is aware of the polyfills.
```
Bertie690: Am I stupid or does this file (`src/typing/polyfills`) just... not exist?

src/main.ts
```diff
@@ -1,3 +1,6 @@
...
+// Polyfills should always be imported first.
```
Bertie690:
```suggestion
// All manually rolled polyfills MUST be loaded first for side effects
```

src/polyfills.ts
SirzBenjie: It did before kev pointed out that I could pull in definitions for polyfills in tsconfig.

Bertie690 (approved):
Lgtm

DayKev (approved):

PR #6109, opened 2025-07-17
[Misc] Move asset initialization into `preload` instead of `launchBattle`

Bertie690:
Virtually nothing (aside from changing timing of memory usage spikes)
We should not be starting our phase system off of a floating promise.
Also, asset loading should not be in `create` or `launchBattle` as these are notably _not_ where they should go

> [!IMPORTANT] 
> Notably, this is effectively the only reason why `phaseManager.currentPhase` can be nullish (due to the game literally not having started yet), so `null` can likely be removed from its signature in a future PR.

Moved code and added an actual error callback because floating promises

## Screenshots/Videos
TODO

## How to test the changes?
Boot up game, make sure it still works

## Checklist


Madmadness65 (dismissed):

test/testUtils/gameManager.ts
```diff
@@ -124,12 +113,34 @@ export class GameManager {
...
+    this.scene.phaseManager.clearAllPhases();
+    this.scene.phaseManager.unshiftNew("LoginPhase");
+    this.scene.phaseManager.unshiftNew("TitlePhase");
+    this.scene.phaseManager.shiftPhase();
+  }
+
+  /**
+   * Initialize various default overrides for starting tests, typically to alleviate randomness.
+   */
+  // TODO: This should not be here
```
DayKev: Where should it be then?

Bertie690: Inside overrides.ts likely

DayKev: We shouldn't be putting automated testing code outside of `test/`.

Bertie690: Oh wait sorry i meant `overridesHelper.ts`

(the OTHER overrides file lmfao)

SirzBenjie (dismissed):

DayKev (dismissed):

test/test-utils/game-manager.ts
```diff
@@ -283,12 +293,11 @@ export class GameManager {
...
           .getMoveset()
           [movePosition].getMove();
-        if (!move.isMultiTarget()) {
-          handler.setCursor(targetIndex !== undefined ? targetIndex : BattlerIndex.ENEMY);
-        }
+
         if (move.isMultiTarget() && targetIndex !== undefined) {
           expect.fail(`targetIndex was passed to selectMove() but move ("${move.name}") is not targetted`);
         }
+        handler.setCursor(targetIndex ?? BattlerIndex.ENEMY);
```
DayKev: Oh I just noticed, why did you remove the "this isn't a multi-target attack" check from this?

Bertie690: oh on second thought it appears the logic was slightly more complex than i thought - i thought it was a mere case of "simplifying logic@

DayKev (approved):

PR #5932, opened 2025-06-04
[Bug] Reworked BattlerTag/ArenaTag code to prevent breakage on 0 PIDs

Bertie690:
## What are the changes the user will see?
Battler tags which check for the user's ID no longer fails 1 in 4.2 billion times when rolling a natural PID of 0.

## Why am I making these changes?
While it would be nice to migrate to a full UUID system as provided by Phaser and ditch the old PIDs, making sure we don't accidentally break key systems from a random roll is a first step.

## What are the changes from a developer perspective?
Cleaned up battlerTags & ArenaTags code to add logging and not randomly break on 0 PIDs.

## Screenshots/Videos
Not really feasible?

## How to test the changes?
Check the new test in `pokemon-id-checks.test.ts`

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test:silent`)
  - [x] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [x] Have I provided screenshots/videos of the changes (if applicable)?

PR #6110, opened 2025-07-18
[Bug][Refactor] Fix loading arena tags

SirzBenjie:
Wish is now usable, and persists across session saves.

#5932 broke wish due to a minor oversight.
This PR fixes that.
Fixes https://discord.com/channels/1125469663833370665/1348446393261621299/1348446393261621299

It also allows, for the first time, wish to work properly across saves.
Also lets neutralizing gas work properly across saves.

- Added missing implementations for `loadTag` to the SuppressAbilitiesTag and WishTag.
- Added many new types and type helpers for the purposes of type-safe parsing of arena data.
- Added a lengthy description at the top of `ArenaTags` that discusses changes
- Added a new abstract class, `SerializableArenaTag` to mark arena tags that should persist in save data.
- Simplified the body of `ArenaData`'s constructor to be more readable.
  -  When creating an Arena, tags that are not Serializable are now ignored when constructing this data.

## Screenshots/Videos

<details><summary>Demonstration of wish fix, and wish working across saves</summary>

https://github.com/user-attachments/assets/b31124c2-ad03-4b68-8b35-f472cadfd15d
</details>

## How to test the changes?
Load up some old saves and ensure everything works.

## Checklist
  - ~~[ ] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?~~ Not doing this.

### DIFF
--- a/src/@types/arena-tags.ts
+++ b/src/@types/arena-tags.ts
@@ -1,0 +1,65 @@
+import type { ArenaTagTypeMap } from "#data/arena-tag";
+import type { ArenaTagType } from "#enums/arena-tag-type";
+import type { NonFunctionProperties } from "./type-helpers";
+
+/** Subset of {@linkcode ArenaTagType}s that are considered traps */
+export type ArenaTrapTagType =
+  | ArenaTagType.STICKY_WEB
+  | ArenaTagType.SPIKES
+  | ArenaTagType.TOXIC_SPIKES
+  | ArenaTagType.STEALTH_ROCK
+  | ArenaTagType.IMPRISON;
+
+/** Subset of {@linkcode ArenaTagType}s that are considered delayed attacks */
+export type ArenaDelayedAttackTagType = ArenaTagType.FUTURE_SIGHT | ArenaTagType.DOOM_DESIRE;
+
+/** Subset of {@linkcode ArenaTagType}s that work like screens */
+export type ArenaScreenTagType = ArenaTagType.REFLECT | ArenaTagType.LIGHT_SCREEN | ArenaTagType.AURORA_VEIL;
+
+/** Subset of {@linkcode ArenaTagType}s for moves that add protection */
+export type TurnProtectArenaTagType =
+  | ArenaTagType.QUICK_GUARD
+  | ArenaTagType.WIDE_GUARD
+  | ArenaTagType.MAT_BLOCK
+  | ArenaTagType.CRAFTY_SHIELD;
+
+/** Subset of {@linkcode ArenaTagType}s that may persist across turns, and thus may be serialized in `SessionSaveData` */
+export type SerializableArenaTagType =
+  | ArenaDelayedAttackTagType
+  | ArenaScreenTagType
+  | ArenaTrapTagType
+  | ArenaTagType.MUD_SPORT
+  | ArenaTagType.WATER_SPORT
+  | ArenaTagType.MIST
+  | ArenaTagType.WISH
+  | ArenaTagType.GRAVITY
+  | ArenaTagType.TRICK_ROOM
+  | ArenaTagType.SAFEGUARD
+  | ArenaTagType.TAILWIND
+  | ArenaTagType.HAPPY_HOUR
+  | ArenaTagType.FIRE_GRASS_PLEDGE
+  | ArenaTagType.WATER_FIRE_PLEDGE
+  | ArenaTagType.GRASS_WATER_PLEDGE
+  | ArenaTagType.FAIRY_LOCK
+  | ArenaTagType.NEUTRALIZING_GAS
+  | ArenaTagType.NO_CRIT
+  | ArenaTagType.IMPRISON;
+
+/** Subset of {@linkcode ArenaTagType}s that cannot persist across turns, and thus should not be serialized in `SessionSaveData`. */
+export type NonSerializableArenaTagType = ArenaTagType.NONE | TurnProtectArenaTagType | ArenaTagType.ION_DELUGE;
+
+/**
+ * Type-safe representation of the serializable data of an ArenaTag
+ */
+export type ArenaTagTypeData = NonFunctionProperties<
+  ArenaTagTypeMap[keyof {
+    [K in keyof ArenaTagTypeMap as K extends SerializableArenaTagType ? K : never]: ArenaTagTypeMap[K];
+  }]
+>;
+
+/** Dummy, typescript-only declaration to ensure that
+ * {@linkcode ArenaTagTypeMap} has a map for all ArenaTagTypes.
+ *
+ * ⚠️ Does not actually exist at runtime, so it must not be used!
+ */
+declare const EnsureAllArenaTagTypesAreMapped: ArenaTagTypeMap[ArenaTagType] & never;
--- a/src/@types/arena-tags.ts
+++ b/src/@types/arena-tags.ts
@@ -23,30 +23,11 @@ export type TurnProtectArenaTagType =
   | ArenaTagType.MAT_BLOCK
   | ArenaTagType.CRAFTY_SHIELD;
 
-/** Subset of {@linkcode ArenaTagType}s that may persist across turns, and thus may be serialized in `SessionSaveData` */
-export type SerializableArenaTagType =
-  | ArenaDelayedAttackTagType
-  | ArenaScreenTagType
-  | ArenaTrapTagType
-  | ArenaTagType.MUD_SPORT
-  | ArenaTagType.WATER_SPORT
-  | ArenaTagType.MIST
-  | ArenaTagType.WISH
-  | ArenaTagType.GRAVITY
-  | ArenaTagType.TRICK_ROOM
-  | ArenaTagType.SAFEGUARD
-  | ArenaTagType.TAILWIND
-  | ArenaTagType.HAPPY_HOUR
-  | ArenaTagType.FIRE_GRASS_PLEDGE
-  | ArenaTagType.WATER_FIRE_PLEDGE
-  | ArenaTagType.GRASS_WATER_PLEDGE
-  | ArenaTagType.FAIRY_LOCK
-  | ArenaTagType.NEUTRALIZING_GAS
-  | ArenaTagType.NO_CRIT
-  | ArenaTagType.IMPRISON;
-
 /** Subset of {@linkcode ArenaTagType}s that cannot persist across turns, and thus should not be serialized in `SessionSaveData`. */
 export type NonSerializableArenaTagType = ArenaTagType.NONE | TurnProtectArenaTagType | ArenaTagType.ION_DELUGE;
+
+/** Subset of {@linkcode ArenaTagType}s that may persist across turns, and thus may be serialized in `SessionSaveData` */
+export type SerializableArenaTagType = Exclude<ArenaTagType, NonSerializableArenaTagType>;
 
 /**
  * Type-safe representation of the serializable data of an ArenaTag
@@ -59,6 +40,8 @@ export type ArenaTagTypeData = NonFunctionProperties<
 
 /** Dummy, typescript-only declaration to ensure that
  * {@linkcode ArenaTagTypeMap} has a map for all ArenaTagTypes.
+ *
+ * If an arena tag is missing from the map, typescript will throw an error on this statement.
  *
  * ⚠️ Does not actually exist at runtime, so it must not be used!
  */
--- a/src/@types/arena-tags.ts
+++ b/src/@types/arena-tags.ts
@@ -2,7 +2,7 @@ import type { ArenaTagTypeMap } from "#data/arena-tag";
 import type { ArenaTagType } from "#enums/arena-tag-type";
 import type { NonFunctionProperties } from "./type-helpers";
 
-/** Subset of {@linkcode ArenaTagType}s that are considered traps */
+/** Subset of {@linkcode ArenaTagType}s that apply some negative effect to pokemon that switch in ({@link https://bulbapedia.bulbagarden.net/wiki/List_of_moves_that_cause_entry_hazards#List_of_traps | entry hazards} and Imprison. */
 export type ArenaTrapTagType =
   | ArenaTagType.STICKY_WEB
   | ArenaTagType.SPIKES
@@ -13,7 +13,7 @@ export type ArenaTrapTagType =
 /** Subset of {@linkcode ArenaTagType}s that are considered delayed attacks */
 export type ArenaDelayedAttackTagType = ArenaTagType.FUTURE_SIGHT | ArenaTagType.DOOM_DESIRE;
 
-/** Subset of {@linkcode ArenaTagType}s that work like screens */
+/** Subset of {@linkcode ArenaTagType}s that create {@link https://bulbapedia.bulbagarden.net/wiki/Category:Screen-creating_moves | screens}. */
 export type ArenaScreenTagType = ArenaTagType.REFLECT | ArenaTagType.LIGHT_SCREEN | ArenaTagType.AURORA_VEIL;
 
 /** Subset of {@linkcode ArenaTagType}s for moves that add protection */
@@ -23,10 +23,10 @@ export type TurnProtectArenaTagType =
   | ArenaTagType.MAT_BLOCK
   | ArenaTagType.CRAFTY_SHIELD;
 
-/** Subset of {@linkcode ArenaTagType}s that cannot persist across turns, and thus should not be serialized in `SessionSaveData`. */
+/** Subset of {@linkcode ArenaTagType}s that cannot persist across turns, and thus should not be serialized in {@linkcode SessionSaveData}. */
 export type NonSerializableArenaTagType = ArenaTagType.NONE | TurnProtectArenaTagType | ArenaTagType.ION_DELUGE;
 
-/** Subset of {@linkcode ArenaTagType}s that may persist across turns, and thus may be serialized in `SessionSaveData` */
+/** Subset of {@linkcode ArenaTagType}s that may persist across turns, and thus must be serialized in {@linkcode SessionSaveData}. */
 export type SerializableArenaTagType = Exclude<ArenaTagType, NonSerializableArenaTagType>;
 
 /**
--- a/src/@types/type-helpers.ts
+++ b/src/@types/type-helpers.ts
@@ -46,7 +46,7 @@ export type InferKeys<O extends Record<keyof any, unknown>, V extends EnumValues
 }[keyof O];
 
 /**
- * Type helper for `Function` types. Equivalent to `Function`, but will not raise a warning in TypeScript.
+ * Type helper that matches any `Function` type. Equivalent to `Function`, but will not raise a warning from Biome.
  */
 export type AnyFn = (...args: any[]) => any;
 
--- a/src/@types/type-helpers.ts
+++ b/src/@types/type-helpers.ts
@@ -64,7 +64,6 @@ export type NonFunctionProperties<T> = {
 
 /**
  * Type helper to extract out non-function properties from a type, recursively applying to nested properties.
- *
  */
 export type NonFunctionPropertiesRecursive<Class> = {
   [K in keyof Class as Class[K] extends AnyFn ? never : K]: Class[K] extends Array<infer U>
--- a/src/@types/type-helpers.ts
+++ b/src/@types/type-helpers.ts
@@ -57,7 +57,6 @@ export type AnyFn = (...args: any[]) => any;
  * Useful to produce a type that is roughly the same as the type of `{... obj}`, where `obj` is an instance of `T`.
  * A couple of differences:
  * - Private and protected properties are not included.
- * -
  */
 export type NonFunctionProperties<T> = {
   [K in keyof T as T[K] extends AnyFn ? never : K]: T[K];
--- a/src/@types/type-helpers.ts
+++ b/src/@types/type-helpers.ts
@@ -44,3 +44,33 @@ export type Mutable<T> = {
 export type InferKeys<O extends Record<keyof any, unknown>, V extends EnumValues<O>> = {
   [K in keyof O]: O[K] extends V ? K : never;
 }[keyof O];
+
+/**
+ * Type helper for `Function` types. Equivalent to `Function`, but will not raise a warning in TypeScript.
+ */
+export type AnyFn = (...args: any[]) => any;
+
+/**
+ * Type helper to extract non-function properties from a type.
+ *
+ * @remarks
+ * Useful to produce a type that is roughly the same as the type of `{... obj}`, where `obj` is an instance of `T`.
+ * A couple of differences:
+ * - Private and protected properties are not included.
+ * -
+ */
+export type NonFunctionProperties<T> = {
+  [K in keyof T as T[K] extends AnyFn ? never : K]: T[K];
+};
+
+/**
+ * Type helper to extract out non-function properties from a type, recursively applying to nested properties.
+ *
+ */
+export type NonFunctionPropertiesRecursive<Class> = {
+  [K in keyof Class as Class[K] extends AnyFn ? never : K]: Class[K] extends Array<infer U>
+    ? NonFunctionPropertiesRecursive<U>[]
+    : Class[K] extends object
+      ? NonFunctionPropertiesRecursive<Class[K]>
+      : Class[K];
+};
--- a/src/@types/type-helpers.ts
+++ b/src/@types/type-helpers.ts
@@ -57,6 +57,7 @@ export type AnyFn = (...args: any[]) => any;
  * Useful to produce a type that is roughly the same as the type of `{... obj}`, where `obj` is an instance of `T`.
  * A couple of differences:
  * - Private and protected properties are not included.
+ * - Nested properties are not recursively extracted. For this, use {@linkcode NonFunctionPropertiesRecursive}
  */
 export type NonFunctionProperties<T> = {
   [K in keyof T as T[K] extends AnyFn ? never : K]: T[K];
@@ -72,3 +73,5 @@ export type NonFunctionPropertiesRecursive<Class> = {
       ? NonFunctionPropertiesRecursive<Class[K]>
       : Class[K];
 };
+
+export type AbstractConstructor<T> = abstract new (...args: any[]) => T;
--- a/src/data/arena-tag.ts
+++ b/src/data/arena-tag.ts
@@ -61,9 +61,13 @@ If the field should be accessible outside of the class, then a public getter sho
 
 /** Interface of the serializable fields of ArenaTagData */
 interface BaseArenaTag {
-  /** The turns that have passed since the tag was added */
+  /** 
+   * The tag's remaining duration. Setting to any number `<=0` will make the tag's duration effectively infinite.
+   */
   turnCount: number;
-  /** The id of the move that set the tag, or undefined if not set by a move */
+  /** 
+   * The {@linkcode MoveId} that created this tag, or `undefined` if not set by a move.
+   */
   sourceMove?: MoveId;
   /** The source pokemon of the move, or undefined if not set by a pokemon */
   // Note: Intentionally not using `?`, as the property should always exist, but just be undefined if not present.
@@ -625,6 +629,7 @@ class WishTag extends SerializableArenaTag {
     const target = globalScene.getField()[this.battlerIndex];
     if (target?.isActive(true)) {
       globalScene.phaseManager.queueMessage(
+        // TODO: Rename key as it triggers on activation
         i18next.t("arenaTag:wishTagOnAdd", {
           pokemonNameWithAffix: this.sourceName,
         }),
@@ -671,7 +676,7 @@ export abstract class WeakenMoveTypeTag extends SerializableArenaTag {
  */
 class MudSportTag extends WeakenMoveTypeTag {
   public readonly tagType = ArenaTagType.MUD_SPORT;
-  override get weakenedType(): PokemonType {
+  override get weakenedType(): PokemonType.ELECTRIC {
     return PokemonType.ELECTRIC;
   }
   constructor(turnCount: number, sourceId?: number) {
@@ -693,7 +698,7 @@ class MudSportTag extends WeakenMoveTypeTag {
  */
 class WaterSportTag extends WeakenMoveTypeTag {
   public readonly tagType = ArenaTagType.WATER_SPORT;
-  override get weakenedType(): PokemonType {
+  override get weakenedType(): PokemonType.FIRE {
     return PokemonType.FIRE;
   }
   constructor(turnCount: number, sourceId?: number) {
--- a/src/data/arena-tag.ts
+++ b/src/data/arena-tag.ts
@@ -366,7 +366,10 @@ type ProtectConditionFunc = (arena: Arena, moveId: MoveId) => boolean;
 export abstract class ConditionalProtectTag extends ArenaTag {
   /** The condition function to determine which moves are negated */
   protected protectConditionFunc: ProtectConditionFunc;
-  /** Does this apply to all moves, including those that ignore other forms of protection? */
+  /**
+   * Whether this protection effect should apply to _all_ moves, including ones that ignore other forms of protection.
+   * @defaultValue `false`
+   */
   protected ignoresBypass: boolean;
 
   constructor(
--- a/src/data/arena-tag.ts
+++ b/src/data/arena-tag.ts
@@ -19,17 +19,47 @@ import { Stat } from "#enums/stat";
 import { StatusEffect } from "#enums/status-effect";
 import type { Arena } from "#field/arena";
 import type { Pokemon } from "#field/pokemon";
-import type { Mutable } from "#types/type-helpers";
+import type {
+  ArenaDelayedAttackTagType,
+  ArenaScreenTagType,
+  ArenaTagTypeData,
+  ArenaTrapTagType,
+  SerializableArenaTagType,
+} from "#types/arena-tags";
+import type { Mutable, NonFunctionProperties } from "#types/type-helpers";
 import { BooleanHolder, NumberHolder, toDmgValue } from "#utils/common";
 import i18next from "i18next";
 
 /*
-Interface for ArenaTag data.
-Any 
+ArenaTags are are meant for effects that are tied to the arena (as opposed to a specific pokemon).
+Examples include (but are not limited to)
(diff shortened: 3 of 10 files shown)


src/@types/type-helpers.ts
```diff
@@ -44,3 +44,32 @@ export type Mutable<T> = {
 export type InferKeys<O extends Record<keyof any, unknown>, V extends EnumValues<O>> = {
   [K in keyof O]: O[K] extends V ? K : never;
 }[keyof O];
+
+/**
+ * Type helper for `Function` types. Equivalent to `Function`, but will not raise a warning in TypeScript.
+ */
+export type AnyFn = (...args: any[]) => any;
```
DayKev, line 51: We already have this iirc (I think it's called `CommonFn` or something?).

src/@types/type-helpers.ts
```diff
@@ -44,3 +44,32 @@ export type Mutable<T> = {
...
+ * A couple of differences:
+ * - Private and protected properties are not included.
+ */
+export type NonFunctionProperties<T> = {
+  [K in keyof T as T[K] extends AnyFn ? never : K]: T[K];
+};
+
+/**
+ * Type helper to extract out non-function properties from a type, recursively applying to nested properties.
+ *
```
DayKev:
```suggestion
```

src/@types/type-helpers.ts
SirzBenjie: Well it's not in type helpers and a cursory search came up with nothing.

DayKev: Lemme double check, maybe I'm thinking of pkty. I could have sworn it was here too though.

DayKev: Oh, maybe I was thinking of `ConditionFn`. Nevermind then.

src/data/terrain.ts
```diff
@@ -13,6 +13,10 @@ export enum TerrainType {
   PSYCHIC,
 }
 
+export interface SerializedTerrain {
+  terrainType: TerrainType;
+  turnsLeft: number;
+}
```
DayKev, line 19:
```suggestion
}

```

src/@types/arena-tags.ts
```diff
@@ -0,0 +1,65 @@
...
+  | ArenaTagType.FIRE_GRASS_PLEDGE
+  | ArenaTagType.WATER_FIRE_PLEDGE
+  | ArenaTagType.GRASS_WATER_PLEDGE
+  | ArenaTagType.FAIRY_LOCK
+  | ArenaTagType.NEUTRALIZING_GAS
+  | ArenaTagType.NO_CRIT
+  | ArenaTagType.IMPRISON;
+
+/** Subset of {@linkcode ArenaTagType}s that cannot persist across turns, and thus should not be serialized in `SessionSaveData`. */
+export type NonSerializableArenaTagType = ArenaTagType.NONE | TurnProtectArenaTagType | ArenaTagType.ION_DELUGE;
```
emdeann, line 27:
```suggestion
export type NonSerializableArenaTagType = Exclude<ArenaTagType, SerializableArenaTagType>
```
This way we only need to manually specify `SerializableArenaTagType`

src/data/arena-tag.ts
```diff
@@ -19,17 +19,78 @@ import { Stat } from "#enums/stat";
...
+session loader is able to deserialize saved tags correctly.
+
+If the data is static (i.e. it is always the same for all instances of the class, such as the 
+type that is weakened by Mud Sport/Water Sport), then it must not be defined as a field, and must
+instead be defined as a getter.
+A static property is also acceptable, though static properties are less ergonomic with inheritance.
+
+If the data is mutable (i.e. it can change over the course of the tag's lifetime), then it *must*
+be defined as a field, and it must be set in the `loadTag` method.
+Such fields cannot be marked as `private/protected`, as if they were, typescript would complain
```
emdeann: Can you explain the reasoning here in more detail? I tested this and got no typescript errors.

It's not the end of the world but I don't like the idea of defining mutable properties as `readonly` and then pretending they're mutable afterwards (as is the case with something like `WishTag` here)

SirzBenjie: > Can you explain the reasoning here in more detail? I tested this and got no typescript errors.
> 
> It's not the end of the world but I don't like the idea of defining mutable properties as `readonly` and then pretending they're mutable afterwards (as is the case with something like `WishTag` here)

Ah, this specific wording is a remnant of the prior structure I had where the interfaces for each class had to be defined manually. I have since found a better way of doing so, using the mapped type at the bottom. (Now, the `serialized` tag interface is automatically derived and accessible via the type helpers I've created).

While the wording needs to be updated, the constraints remain.
Private/protected types won't appear in mapped types, so our types that use the shape of these classes will not have them as fields, which destroys the type safety I'm looking for.

Observe that a variable of type `WishTagData` now no longer understands that it actually has one of those private properties:
image

This is a huge limitation in typescript. There is no way, (other than with accessors), to have a type that is outwardly immutable, and internally mutable. The closest is with casting away the readonly field inside the method, as I've done with the mutable cast.

src/@types/arena-tags.ts
SirzBenjie: Ah I had tried to use `Exclude` in a slightly different way and figured that the enum was just getting in my way.
This is much better.
Actually, because most tags should be serializable, it'll be less cumbersome to write out `NonSerializable`, and then use exclude the other way.

emdeann: The implication there is that tags will be considered serializable by default (at least by this type). Is that acceptable?

src/data/arena-tag.ts
```diff
@@ -19,17 +19,79 @@ import { Stat } from "#enums/stat";
...
+  /**
+   * The side that the arena tag is on
+   * @defaultValue `ArenaTagSide.BOTH`
+   */
+  side: ArenaTagSide;
+}
+
+/**
+ * Abstract base class for `ArenaTag`s
+ */
```
Bertie690:
```suggestion
/**
 * An {@linkcode ArenaTag} represents a semi-persistent effect affecting a given _side_ of the field.
 * Unlike {@linkcode BattlerTag}s (which are tied to individual {@linkcode Pokemon}), `ArenaTag`s function independently of
 * the Pokemon currently on-field, only cleared on arena reset or through their respective {@linkcode ArenaTag.lapse | lapse} methods.
 */
```

src/data/arena-tag.ts
```diff
@@ -19,17 +19,79 @@ import { Stat } from "#enums/stat";
...
+If the field should be accessible outside of the class, then a public getter should be used.
+*/
+
+/** Interface of the serializable fields of ArenaTagData */
+interface BaseArenaTag {
+  /** The turns that have passed since the tag was added */
+  turnCount: number;
+  /** The id of the move that set the tag, or undefined if not set by a move */
+  sourceMove?: MoveId;
+  /** The source pokemon of the move, or undefined if not set by a pokemon */
```
Bertie690:
```suggestion
  /** 
   * The {@linkcode Pokemon.id | PID} of the {@linkcode Pokemon} having created the tag, or `undefined` if not set by a Pokemon.
   * @todo Implement handling for `ArenaTag`s created by non-pokemon sources (most tags will throw errors without a source)
   */
```

src/data/arena-tag.ts
```diff
@@ -19,17 +19,79 @@ import { Stat } from "#enums/stat";
...
+Such fields cannot be marked as `private/protected`, as if they were, typescript would omit them from
+types that are based off of the class, namely, `ArenaTagTypeData`. It is preferrable to trade the
+type-safety of private/protected fields for the type safety when deserializing arena tags from save data.
+
+For data that is mutable only within a turn (e.g. SuppressAbilitiesTag's beingRemoved field),
+where it does not make sense to be serialized, the field should use ES2020's private field syntax (a `#` prepended to the field name).
+If the field should be accessible outside of the class, then a public getter should be used.
+*/
+
+/** Interface of the serializable fields of ArenaTagData */
```
Bertie690:
```suggestion
/** Interface containing the serializable fields of ArenaTagData. */
```

src/data/arena-tag.ts
```diff
@@ -19,17 +19,79 @@ import { Stat } from "#enums/stat";
...
+interface BaseArenaTag {
+  /** The turns that have passed since the tag was added */
```
Bertie690:
```suggestion
  /** 
   * The tag's current duration.
   * Decremented by 1 at the end of each turn, triggering the tag's {@linkcode ArenaTag.onRemove | onRemove} effects and clearing it upon reaching 0.
   * If set to any number `<=0`, will make the tag's duration effectively infinite.
   */
   // TODO: Do we want this?
```

Personally, I'm not a huge fan of the whole "-1=infinite" shebang, but might as well document it since it does kinda exist atm

src/data/arena-tag.ts
```diff
@@ -19,17 +19,79 @@ import { Stat } from "#enums/stat";
...
+  turnCount: number;
+  /** The id of the move that set the tag, or undefined if not set by a move */
+  sourceMove?: MoveId;
+  /** The source pokemon of the move, or undefined if not set by a pokemon */
+  // Note: Intentionally not using `?`, as the property should always exist, but just be undefined if not present.
+  sourceId: number | undefined;
+  /**
+   * The side that the arena tag is on
```
Bertie690:
```suggestion
   * The {@linkcode ArenaTagSide | side of the field} that this arena tag affects.
```

src/@types/arena-tags.ts
```diff
@@ -0,0 +1,48 @@
...
+export type TurnProtectArenaTagType =
+  | ArenaTagType.QUICK_GUARD
+  | ArenaTagType.WIDE_GUARD
+  | ArenaTagType.MAT_BLOCK
+  | ArenaTagType.CRAFTY_SHIELD;
+
+/** Subset of {@linkcode ArenaTagType}s that cannot persist across turns, and thus should not be serialized in `SessionSaveData`. */
+export type NonSerializableArenaTagType = ArenaTagType.NONE | TurnProtectArenaTagType | ArenaTagType.ION_DELUGE;
+
+/** Subset of {@linkcode ArenaTagType}s that may persist across turns, and thus may be serialized in `SessionSaveData` */
```
Bertie690:
```suggestion
/** Subset of {@linkcode ArenaTagType}s that may persist across turns, and thus must be serialized in {@linkcode SessionSaveData}. */
```

src/@types/arena-tags.ts
```diff
@@ -0,0 +1,48 @@
...
+export type ArenaScreenTagType = ArenaTagType.REFLECT | ArenaTagType.LIGHT_SCREEN | ArenaTagType.AURORA_VEIL;
+
+/** Subset of {@linkcode ArenaTagType}s for moves that add protection */
+export type TurnProtectArenaTagType =
+  | ArenaTagType.QUICK_GUARD
+  | ArenaTagType.WIDE_GUARD
+  | ArenaTagType.MAT_BLOCK
+  | ArenaTagType.CRAFTY_SHIELD;
+
+/** Subset of {@linkcode ArenaTagType}s that cannot persist across turns, and thus should not be serialized in `SessionSaveData`. */
```
Bertie690:
```suggestion
/** Subset of {@linkcode ArenaTagType}s that cannot persist across turns, and thus do not need to be serialized in {@linkcode SessionSaveData}. */
```

I think this is a subtle but important distinction - it's not as if we physically _can't_ serialize them (we could with minor tweaking), we just don't have to since they only last for the current turn.

src/@types/arena-tags.ts
```diff
@@ -0,0 +1,48 @@
...
+  | ArenaTagType.STICKY_WEB
+  | ArenaTagType.SPIKES
+  | ArenaTagType.TOXIC_SPIKES
+  | ArenaTagType.STEALTH_ROCK
+  | ArenaTagType.IMPRISON;
+
+/** Subset of {@linkcode ArenaTagType}s that are considered delayed attacks */
+export type ArenaDelayedAttackTagType = ArenaTagType.FUTURE_SIGHT | ArenaTagType.DOOM_DESIRE;
+
+/** Subset of {@linkcode ArenaTagType}s that work like screens */
```
Bertie690:
```suggestion
/** Subset of {@linkcode ArenaTagType}s that create {@link https://bulbapedia.bulbagarden.net/wiki/Category:Screen-creating_moves | screens}. */
```

src/@types/arena-tags.ts
```diff
@@ -0,0 +1,48 @@
+import type { ArenaTagTypeMap } from "#data/arena-tag";
+import type { ArenaTagType } from "#enums/arena-tag-type";
+import type { NonFunctionProperties } from "./type-helpers";
+
+/** Subset of {@linkcode ArenaTagType}s that are considered traps */
```
Bertie690:
```suggestion
/** Subset of {@linkcode ArenaTagType}s that are considered {@link https://bulbapedia.bulbagarden.net/wiki/List_of_moves_that_cause_entry_hazards#List_of_traps | traps (AKA entry hazards). */
```

src/@types/arena-tags.ts
```diff
@@ -0,0 +1,48 @@
...
+export type ArenaTrapTagType =
+  | ArenaTagType.STICKY_WEB
+  | ArenaTagType.SPIKES
+  | ArenaTagType.TOXIC_SPIKES
+  | ArenaTagType.STEALTH_ROCK
+  | ArenaTagType.IMPRISON;
+
+/** Subset of {@linkcode ArenaTagType}s that are considered delayed attacks */
```
Bertie690, line 13:
```suggestion
/** Subset of {@linkcode ArenaTagType}s that are considered delayed attacks. */
```

src/@types/type-helpers.ts
```diff
@@ -44,3 +44,34 @@ export type Mutable<T> = {
 export type InferKeys<O extends Record<keyof any, unknown>, V extends EnumValues<O>> = {
   [K in keyof O]: O[K] extends V ? K : never;
 }[keyof O];
+
+/**
+ * Type helper for `Function` types. Equivalent to `Function`, but will not raise a warning in TypeScript.
```
Bertie690:
```suggestion
 * Type helper that matches any `Function` type. Equivalent to `Function`, but will not raise a warning from Biome.
```

biome gives the warning, not typescript - tsc couldn't care less

src/@types/type-helpers.ts
```diff
@@ -44,3 +44,34 @@ export type Mutable<T> = {
...
+/**
+ * Type helper to extract out non-function properties from a type, recursively applying to nested properties.
+ */
+export type NonFunctionPropertiesRecursive<Class> = {
+  [K in keyof Class as Class[K] extends AnyFn ? never : K]: Class[K] extends Array<infer U>
+    ? NonFunctionPropertiesRecursive<U>[]
+    : Class[K] extends object
+      ? NonFunctionPropertiesRecursive<Class[K]>
+      : Class[K];
+};
```
Bertie690, line 75: A few nits:
1. Would it make sense to move these into their own file?
2. Please add vitest type checking for these. I will do it by my own bare hands if need be

src/data/arena-tag.ts
```diff
@@ -19,17 +19,79 @@ import { Stat } from "#enums/stat";
...
+For data that is mutable only within a turn (e.g. SuppressAbilitiesTag's beingRemoved field),
+where it does not make sense to be serialized, the field should use ES2020's private field syntax (a `#` prepended to the field name).
+If the field should be accessible outside of the class, then a public getter should be used.
+*/
+
+/** Interface of the serializable fields of ArenaTagData */
+interface BaseArenaTag {
+  /** The turns that have passed since the tag was added */
+  turnCount: number;
+  /** The id of the move that set the tag, or undefined if not set by a move */
```
Bertie690:
```suggestion
  /** 
   * The {@linkcode MoveId} that created this tag, or `undefined` if not set by a move.
   */
```

src/data/arena-tag.ts
```diff
@@ -19,17 +19,79 @@ import { Stat } from "#enums/stat";
...
+  public sourceMove?: MoveId;
+  public sourceId: number | undefined;
+  public side: ArenaTagSide;
+
+  constructor(turnCount: number, sourceMove?: MoveId, sourceId?: number, side?: ArenaTagSide) {
+    this.turnCount = turnCount;
+    this.sourceMove = sourceMove;
+    this.sourceId = sourceId;
+    this.side = side ?? ArenaTagSide.BOTH;
+  }
```
Bertie690:
```suggestion
  constructor(turnCount: number, sourceMove?: MoveId, sourceId?: number, side: ArenaTagSide = ArenaTagSide.BOTH) {
    this.turnCount = turnCount;
    this.sourceMove = sourceMove;
    this.sourceId = sourceId;
    this.side = side;
  }
```

src/data/arena-tag.ts
```diff
@@ -63,7 +125,7 @@ export abstract class ArenaTag {
    * This is meant to be inherited from by any arena tag with custom attributes
    * @param {ArenaTag | any} source An arena tag
```
Bertie690:
```suggestion
   * @param source - The {@linkcode BaseArenaTag} being loaded
```

src/data/arena-tag.ts
```diff
@@ -159,33 +229,11 @@ export class MistTag extends ArenaTag {
...
-    side: ArenaTagSide,
-    weakenedCategories: MoveCategory[],
-  ) {
-    super(tagType, turnCount, sourceMove, sourceId, side);
-
-    this.weakenedCategories = weakenedCategories;
-  }
+export abstract class WeakenMoveScreenTag extends SerializableArenaTag {
+  public abstract readonly tagType: ArenaScreenTagType;
+  // Getter, to avoid unnecessary serialization and prevent modification
```
Bertie690:
```suggestion
  // Getter to avoid unnecessary serialization and prevent modification
```

src/data/arena-tag.ts
```diff
@@ -19,17 +19,79 @@ import { Stat } from "#enums/stat";
...
+If the data is static (i.e. it is always the same for all instances of the class, such as the 
+type that is weakened by Mud Sport/Water Sport), then it must not be defined as a field, and must
+instead be defined as a getter.
+A static property is also acceptable, though static properties are less ergonomic with inheritance.
+
+If the data is mutable (i.e. it can change over the course of the tag's lifetime), then it *must*
+be defined as a field, and it must be set in the `loadTag` method.
+Such fields cannot be marked as `private/protected`, as if they were, typescript would omit them from
+types that are based off of the class, namely, `ArenaTagTypeData`. It is preferrable to trade the
+type-safety of private/protected fields for the type safety when deserializing arena tags from save data.
```
Bertie690, line 55: Maybe we can link to a TS issue somewhere mentioning this?
Smthng like ("see <issue> for more information"

src/data/arena-tag.ts
```diff
@@ -19,17 +19,79 @@ import { Stat } from "#enums/stat";
...
+
+For data that is mutable only within a turn (e.g. SuppressAbilitiesTag's beingRemoved field),
+where it does not make sense to be serialized, the field should use ES2020's private field syntax (a `#` prepended to the field name).
+If the field should be accessible outside of the class, then a public getter should be used.
```
Bertie690: This should _probably_ go inside an `@remarks` or `@privateRemarks` on `ArenaTag` or its abstract children

src/enums/arena-tag-type.ts
```diff
@@ -1,3 +1,12 @@
+import type { ArenaTagTypeMap } from "#data/arena-tag";
+import type { NonSerializableArenaTagType, SerializableArenaTagType } from "#types/arena-tags";
+
+/**
+ * @privateremarks
```
Bertie690:
```suggestion
 * Enum representing all different types of {@linkcode ArenaTag}s.
 * @privateRemarks
```

src/data/arena-tag.ts
```diff
@@ -286,21 +345,20 @@ type ProtectConditionFunc = (arena: Arena, moveId: MoveId) => boolean;
  * Class to implement conditional team protection
  * applies protection based on the attributes of incoming moves
  */
-export class ConditionalProtectTag extends ArenaTag {
+export abstract class ConditionalProtectTag extends ArenaTag {
   /** The condition function to determine which moves are negated */
   protected protectConditionFunc: ProtectConditionFunc;
   /** Does this apply to all moves, including those that ignore other forms of protection? */
```
Bertie690:
```suggestion
  /** 
   * Whether this protection effect should apply to _all_ moves, including ones that ignore other forms of protection.
   * @defaultValue `false`
   */
```

src/data/arena-tag.ts
```diff
@@ -550,45 +607,37 @@ class WishTag extends ArenaTag {
...
+    (this as Mutable<this>).healHp = toDmgValue(source.getMaxHp() / 2);
+    (this as Mutable<this>).battlerIndex = source.getBattlerIndex();
   }
 
   onRemove(_arena: Arena): void {
     const target = globalScene.getField()[this.battlerIndex];
     if (target?.isActive(true)) {
-      globalScene.phaseManager.queueMessage(this.triggerMessage);
+      globalScene.phaseManager.queueMessage(
+        i18next.t("arenaTag:wishTagOnAdd", {
```
Bertie690, line 641:
```suggestion
        // TODO: Rename key as it triggers on activation
        i18next.t("arenaTag:wishTagOnAdd", {
```

src/data/arena-tag.ts
```diff
@@ -612,8 +661,12 @@ export class WeakenMoveTypeTag extends ArenaTag {
  * Weakens Electric type moves for a set amount of turns, usually 5.
  */
 class MudSportTag extends WeakenMoveTypeTag {
-  constructor(turnCount: number, sourceId: number) {
-    super(ArenaTagType.MUD_SPORT, turnCount, PokemonType.ELECTRIC, MoveId.MUD_SPORT, sourceId);
+  public readonly tagType = ArenaTagType.MUD_SPORT;
+  override get weakenedType(): PokemonType {
+    return PokemonType.ELECTRIC;
+  }
```
Bertie690: Is there a reason you didn't specify the exact type being returned in the annotation like for screens?

src/data/arena-tag.ts
```diff
@@ -630,8 +683,12 @@ class MudSportTag extends WeakenMoveTypeTag {
  * Weakens Fire type moves for a set amount of turns, usually 5.
  */
 class WaterSportTag extends WeakenMoveTypeTag {
-  constructor(turnCount: number, sourceId: number) {
-    super(ArenaTagType.WATER_SPORT, turnCount, PokemonType.FIRE, MoveId.WATER_SPORT, sourceId);
+  public readonly tagType = ArenaTagType.WATER_SPORT;
+  override get weakenedType(): PokemonType {
```
Bertie690: See above

src/field/arena.ts
```diff
@@ -666,7 +667,11 @@ export class Arena {
    * @param simulated if `true`, this applies arena tags without changing game state
    * @param args array of parameters that the called upon tags may need
    */
-  applyTags(tagType: ArenaTagType | Constructor<ArenaTag>, simulated: boolean, ...args: unknown[]): void {
+  applyTags(
+    tagType: ArenaTagType | Constructor<ArenaTag> | AbstractConstructor<ArenaTag>,
+    simulated: boolean,
+    ...args: unknown[]
+  ): void {
```
Bertie690, line 674: Can't we get actual typing on these args?

src/system/arena-data.ts
```diff
@@ -1,9 +1,19 @@
...
+
+export interface SerializedArenaData {
+  biome: BiomeId;
+  weather: NonFunctionProperties<Weather> | null;
+  terrain: NonFunctionProperties<Terrain> | null;
+  tags?: ArenaTagTypeData[];
+  playerTerasUsed?: number;
+}
 
 export class ArenaData {
```
Bertie690, line 18: I believe we can make this into a regular POJO with a function to construct it.
If we want to accurately represent save data, we should probably make it an interface instead of a class.

src/data/arena-tag.ts
DayKev: I changed the suggestion slightly (you had "by default" twice in the third line, and I changed it to "effectively infinite" instead).

src/@types/type-helpers.ts
DayKev: > 1. Would it make sense to move these into their own file?

Where/why?

src/system/arena-data.ts
DayKev: As someone who converted a bunch of classes to interfaces for pkty, I approve of this suggestion.

src/system/arena-data.ts
```diff
@@ -12,24 +22,27 @@ export class ArenaData {
...
-      : source.terrain
-        ? new Terrain(source.terrain.terrainType, source.terrain.turnsLeft)
-        : null;
-    this.playerTerasUsed = (sourceArena ? sourceArena.playerTerasUsed : source.playerTerasUsed) ?? 0;
-    this.tags = [];
-
+  constructor(source: Arena | SerializedArenaData) {
     if (source.tags) {
-      this.tags = source.tags.map(t => loadArenaTag(t));
+      // NOTE: The filter has to be done _after_ map, data loaded from `ArenaTagTypeData`
```
Bertie690:
```suggestion
      // Exclude any unserializable tags from the serialized data (such as ones only lasting 1 turn).
      // NOTE: The filter has to be done _after_ map, data loaded from `ArenaTagTypeData`
```

Bertie690 (changes requested):
extra nit i forgot to add

src/@types/type-helpers.ts
Bertie690: because the file is getting quite large. 
I want to avoid creating yet another "generic dump site" like for `utils/common`.

For now - a new file (`@types/function-types`) could be made; whichever of this and #6123 happens to be merged first can move it into the `@types/helpers` directory afterwards.

src/@types/arena-tags.ts
SirzBenjie: Imprison is not considered an entry hazard on that page, though, so the comment is _slightly_ misleading.

SirzBenjie:
```suggestion
/** Subset of {@linkcode ArenaTagType}s that apply some negative effect to pokemon that switch in ({@link https://bulbapedia.bulbagarden.net/wiki/List_of_moves_that_cause_entry_hazards#List_of_traps | entry hazards} and Imprison. */
```

src/@types/arena-tags.ts
SirzBenjie: It's not a complete sentence, so no period.

src/@types/arena-tags.ts
SirzBenjie: Well, we also _shouldn't_ serialize them. If we _couldn't_ without breaking things, then the language would have been _must not_.

SirzBenjie:
```suggestion
/** Subset of {@linkcode ArenaTagType}s that cannot persist across turns, and thus should not be serialized in {@linkcode SessionSaveData}. */
```

src/@types/arena-tags.ts
SirzBenjie: Yeah. Almost all tags are serializable. And it's harmless for tags that don't need to be serialized to end up being serialized (as their turn count would be 0 and end up doing nothing). But if tags that need to be serialized aren't, then that's bad.

src/@types/type-helpers.ts
SirzBenjie: The file isn't that large, though. There's only a few things in it. Sure, there's a lot of comments, but that's _documentation._

src/data/arena-tag.ts
SirzBenjie: This is just a part of typescript though. It isn't a typescript issue. Typescript interfaces based on classes always omit private fields. Interfaces are, well, _interfaces_. An interface is supposed to be the public contract for a class.

src/data/arena-tag.ts
SirzBenjie: I sort of like having the instructions for dealing with this class at the top of the file. It's helpful. Slapping them on the class doc is relevant if that information is valuable when outside of the file.

src/data/arena-tag.ts
SirzBenjie: I think this is a bit misleading, as the mechanism for decrementing this is not handled by the arena tag. Indeed, the arena tag itself won't even call `onRemove` if its duration goes to 0. I'll use this:

```suggestion
  /** 
   * The tag's remaining duration. Setting to any number `<=0` will make the tag's duration effectively infinite.
   */
```

src/data/arena-tag.ts
SirzBenjie: What's the point of having the linkcode here if the type is already part of the signature?

src/data/arena-tag.ts
SirzBenjie: Oversight.

src/data/arena-tag.ts
```diff
@@ -612,8 +661,12 @@ export class WeakenMoveTypeTag extends ArenaTag {
  * Weakens Electric type moves for a set amount of turns, usually 5.
  */
 class MudSportTag extends WeakenMoveTypeTag {
-  constructor(turnCount: number, sourceId: number) {
-    super(ArenaTagType.MUD_SPORT, turnCount, PokemonType.ELECTRIC, MoveId.MUD_SPORT, sourceId);
+  public readonly tagType = ArenaTagType.MUD_SPORT;
+  override get weakenedType(): PokemonType {
```
SirzBenjie:
```suggestion
  override get weakenedType(): PokemonType.ELECTRIC {
```

src/data/arena-tag.ts
SirzBenjie: Ah, so that's why there was that bug.. You were misled by the key's name.

src/data/arena-tag.ts
SirzBenjie:
```suggestion
  override get weakenedType(): PokemonType.FIRE {
```

src/field/arena.ts
SirzBenjie: Probably, though not in this PR.

src/system/arena-data.ts
SirzBenjie: I want to do that anyway: take classes and drop fields that really don't need to be serialized. I'm currently looking at `BattleTags`, and there's a lot of fields like this.

src/data/arena-tag.ts
```diff
@@ -286,21 +354,23 @@ type ProtectConditionFunc = (arena: Arena, moveId: MoveId) => boolean;
  * Class to implement conditional team protection
  * applies protection based on the attributes of incoming moves
  */
-export class ConditionalProtectTag extends ArenaTag {
+export abstract class ConditionalProtectTag extends ArenaTag {
   /** The condition function to determine which moves are negated */
   protected protectConditionFunc: ProtectConditionFunc;
```
DayKev, line 368: Shouldn't this also be changed to a getter?

src/data/arena-tag.ts
```diff
@@ -286,21 +354,23 @@ type ProtectConditionFunc = (arena: Arena, moveId: MoveId) => boolean;
...
-  /** Does this apply to all moves, including those that ignore other forms of protection? */
+  /**
+   * Whether this protection effect should apply to _all_ moves, including ones that ignore other forms of protection.
+   * @defaultValue `false`
+   */
   protected ignoresBypass: boolean;
```
DayKev, line 373: Same question.

src/system/arena-data.ts
```diff
@@ -12,24 +22,28 @@ export class ArenaData {
...
-    this.terrain = sourceArena
-      ? sourceArena.terrain
-      : source.terrain
-        ? new Terrain(source.terrain.terrainType, source.terrain.turnsLeft)
-        : null;
-    this.playerTerasUsed = (sourceArena ? sourceArena.playerTerasUsed : source.playerTerasUsed) ?? 0;
-    this.tags = [];
-
+  constructor(source: Arena | SerializedArenaData) {
     if (source.tags) {
```
DayKev: You could simplify it to this right?
```ts
    // Exclude any unserializable tags from the serialized data (such as ones only lasting 1 turn).
    // NOTE: The filter has to be done _after_ map, data loaded from `ArenaTagTypeData`
    // is not yet an instance of `ArenaTag`
    this.tags = source.tags
      ?.map((t: ArenaTag | ArenaTagTypeData) => loadArenaTag(t))
      ?.filter((tag): tag is SerializableArenaTag => tag instanceof SerializableArenaTag) ?? [];
```

src/system/arena-data.ts
```diff
@@ -12,24 +22,28 @@ export class ArenaData {
...
+        .filter((tag): tag is SerializableArenaTag => tag instanceof SerializableArenaTag);
+    } else {
+      this.tags = [];
     }
+
+    if (source instanceof Arena) {
+      this.biome = source.biomeType;
+      this.weather = source.weather;
+      this.terrain = source.terrain;
+      this.playerTerasUsed = source.playerTerasUsed ?? 0;
```
DayKev: This should be moved above the `if` here since it's common to both input types (`Arena` and `SerializedArenaData`), and currently isn't set if the input is `SerializedArenaData`.

src/data/arena-tag.ts
SirzBenjie: Tag can't be serialized, so it doesn't matter.

src/data/arena-tag.ts
SirzBenjie: Again, can't be serialized, so it doesn't matter.

src/system/arena-data.ts
Bertie690: Caveat - the biome field is named differently in each.
I know it's absolutely baffling, but it exists and we have to dael with it

SirzBenjie: Ah, good catch.

> Caveat - the biome field is named differently in each. I know it's absolutely baffling, but it exists and we have to dael with it

I think Kev specifically meant the `playerTerasUsed` which isn't being set.

Bertie690 (dismissed):
lgtm

src/system/arena-data.ts
SirzBenjie: Yeah

src/data/arena-tag.ts
DayKev: Ah, true.

DayKev (approved):

Bertie690 (approved):

PR #6112, opened 2025-07-18
[Misc] Moved + cleaned up string manipulation functions

Bertie690:
Assuming I'm not dumb, none.

We had a lot of duplicate util functions for string and case manipulation.
I counted:
- 2 ways of changing snake case to camel case
- 2 ways of getting pascal case
- 2 near-identical functions for getting title case

Converted all prior calls to string manipulation functions to use the new functions in `utils/strings`. Whether these should be needed is a different matter altogether (`Pascal_Snake_Case` is jank beyond belief), but I digress...

## Screenshots/Videos
N/A

## How to test the changes?
I guess boot up the game and make sure the sprite paths load ok?

## Checklist

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -177,9 +177,10 @@
       }
     },
 
-    // Overrides to prevent unused import removal inside `overrides.ts` and enums files (for TSDoc linkcodes)
+    // Overrides to prevent unused import removal inside `overrides.ts` and enums files (for TSDoc linkcodes),
+    // as well as in all TS files in `scripts/` (which are assumed to be boilerplate templates).
     {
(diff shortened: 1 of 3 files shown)


SirzBenjie:
> Also to be frank we really shouldn't be making custom utils for functions that can be sourced from npm to not reinvent the wheel.

Yes we should. It's not a good idea to use an npm package for something so trivial.

Bertie690:
> Yes we should. It's not a good idea to use an npm package for something so trivial.

ok i fixed it

Bertie690:
I have not changed package json yet shoot
Will do shortly

SirzBenjie (approved):

xsn34kzx (approved):

PR #6114, opened 2025-07-18
[Refactor/Docs] Added `PhaseManager.toTitleScreen`, fully documented phase manager methods

Bertie690:
None
`toTitlePhase` is a nice shorthand for stuff

Also we need docs badly and might as well

Added `toTitlePhase` and updated callsites

Updated TSDocs for all phase manager functions

## Checklist


emdeann:
This isn't bad in premise, but the phase manager as a whole is being actively overhauled. This would cause a bunch of conflicts for no reason

DayKev:
> This isn't bad in premise, but the phase manager as a whole is being actively overhauled. This would cause a bunch of conflicts for no reason

Due to the upcoming phase manager refactor, I'm gonna close this. The changes that are still relevant after the refactor can be put in a new PR.

Bertie690:
Ok then my phase interceptor branch is probably screwed 
_ughhhhhh_

DayKev:
Maybe, maybe not.

DayKev:
Changed it to draft for now.

src/battle-scene.ts
```diff
@@ -713,16 +711,16 @@ export class BattleScene extends SceneBase {
...
+    const res = await this.cachedFetch("./exp-sprites.json");
+    const keys = await res.json();
+    if (!Array.isArray(keys)) {
+      throw new Error("EXP Sprites were not array when fetched!");
+    }
+
+    // TODO: Optimize this
+    for (const k of keys) {
+      expSpriteKeys.add(k);
+    }
```
emdeann: This looks unrelated?

Bertie690: oops mb - will move into another pr ig

Wlowscha (approved):

src/enums/dynamic-phase-type.ts
```diff
@@ -1,6 +1,7 @@
 /**
- * Enum representation of the phase types held by implementations of {@linkcode PhasePriorityQueue}
+ * Enum representation of the phase types held by implementations of {@linkcode PhasePriorityQueue}.
  */
+// TODO: We currently assume these are in order
```
emdeann, line 4: this is being removed anyway

src/phase-manager.ts
```diff
@@ -240,6 +240,21 @@ export class PhaseManager {
...
   }
 
+  /**
+   * Clear all previously set phases, then add a new {@linkcode TitlePhase} to transition to the title screen.
+   * @param addLogin - Whether to add a new {@linkcode LoginPhase} before the {@linkcode TitlePhase}
+   * (but reset everything else).
+   * Default `false`
+   */
+  public toTitleScreen(addLogin = false): void {
+    this.clearAllPhases();
```
emdeann, line 254: `clearAllPhases` and `clearPhaseQueue` are not the same thing - if any of the callsites you've changed expect something to be present in `phaseQueuePrepend`, for example, this would break that.

I don't expect that to be the case though

Bertie690 (commented):
> I don't expect that to be the case though

No - in fact every case where the boolean is true previously called `clearAllPhases` immediately before adding title phases.
(I made sure of it)

SirzBenjie (approved):

src/ui/starter-select-ui-handler.ts
```diff
@@ -4303,7 +4303,10 @@ export class StarterSelectUiHandler extends MessageUiHandler {
     return true;
   }
 
-  tryExit(): boolean {
+  /**
+   * Attempt to back out of the starter selection screen into the appropriate parent modal
+   */
+  tryExit(): void {
```
DayKev, line 4309: Was the return value never used?

DayKev: Huh it... always returned `true`, and the return value was never used. lmao

DayKev (approved):

PR #6115, opened 2025-07-18
[Dev] Turned on `js-check` in TSConfig; fixed script type errors

Bertie690:
The decrypt save util actually works.

I wanted type checking for my egg scripts.
Along the way, I noticed our existing ones were breaking typescript.

Added `checkJs` in ts config
fixed type errors

## How to test the changes?
run ze scripts

## Checklist

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -177,9 +177,10 @@
       }
     },
 
-    // Overrides to prevent unused import removal inside `overrides.ts` and enums files (for TSDoc linkcodes)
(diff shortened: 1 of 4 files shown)


Bertie690:
OOOOOHHHHHHHHHH.
That might be something to look into after this.
(Only thing is we'd need a way of distinguishing any "boilerplate" files from actual script ones wrt linting...)

xsn34kzx (approved):

DayKev (approved):

PR #6121, opened 2025-07-19
[UI/UX] [Bug] Fix modifier-select-phase animation delay 

SirzBenjie:
Makes the changes in #6099 no longer have a discernable impact on the way the animations play out.
Users will now see the pokeball 'opening' animation.

There's a chance this also fixes the visual bug where items display as open pokeballs, though that was not targeted by this PR.

@damocleas pointed out that after #6099, the animations were oddly delayed when there was an upgrade.

Added comments that try to explain what is going on in the method. Usually, we prefer comments to explain _why_ not _what._ However, in this case, the _what_ is hard to discern, and it's hard to write it in a discernable way without disrupting the existing animations.

I restored the prior timing of the animations that cause the shop items to appear.
I also tweaked the `show` method of the `ModifierOption` to be async, and communicate two promises back to the caller: One that resolves once the "upgrade" / "being release from the pokeball" animations have finished playing, and one that resolves once all animations initiated by the method have completed. There are comments discussing why / how they are meant to work.

I then tweaked the rest of the `show` method of the modifier screen to wait for all the animations to finish _before_ showing the cursor sprite and allowing input to be processed.
This ensures that no input is allowed to be processed while the tweens corresponding to the rewards are still playing, which is what caused the crash #6099 fixes.

The resulting code is still pretty ugly, and a bit hard to follow. It is, however, about the best I could muster given the constraints of wanting to keep the animations playing as they were before, while also preventing the crash.

## Screenshots/Videos
<details><summary>Select-modifier-phase with this PR</summary>

https://github.com/user-attachments/assets/94cf7844-877a-4894-ba2f-a52a7bf35829
</details>

## How to test the changes?
Open the game, play some waves. I recommend ensuring that animations look as expected even when upgrading,
so maybe start with a bunch of t3 shinies to ensure max luck and observe what the animations look like.

## Checklist
  - ~~[] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?~~


damocleas (dismissed):
praying this doesn't break again

damocleas (dismissed):

src/ui/modifier-select-ui-handler.ts
```diff
@@ -882,66 +934,101 @@ class ModifierOption extends Phaser.GameObjects.Container {
               duration: 750,
               ease: "Sine.easeOut",
               onComplete: () => {
+                console.log("pbTint complete");
```
DayKev:
```suggestion
```
Leftover log.

src/ui/modifier-select-ui-handler.ts
```diff
@@ -882,66 +934,101 @@ class ModifierOption extends Phaser.GameObjects.Container {
...
         globalScene.tweens.add({
           targets: this.pb,
           duration: 500,
-          delay: 250,
           ease: "Sine.easeIn",
           alpha: 0,
-          onComplete: () => this.pb.destroy(),
+          onComplete: () => {
+            Promise.allSettled(animPromises).then(() => this.pb.destroy());
+            pbResolve();
```
DayKev, line 967: `pbResolve` doesn't contain any code, does it? What does creating an empty (?) Promise and then resolving (?) it here do?

SirzBenjie: This is how promises work. Until the promise's `resolve` method is called, the promise can not be considered fulfilled.

We are essentially creating a barrier later on that says, "hey, you must wait until this promise has resolved before continuing". This is why the promise is pushed to the `finalPromises` array. Each of the promises in this array are awaited on via the `Promise.allSettled` in line 1028. As a result, the promise pushed to `promiseHolder` will only be considered resolved once all of the tweens' `onComplete` methods have been called (as each tween resolves a promise created for it).

The promise isn't really empty. When a promise is created via the `new Promise` syntax, really, its body is just the set of code that is executed before resolve is called. If resolve is never called, then the promise can never be fulfilled.
Using `Promise.withResolvers` gives you the power to resolve the promise _outside_ of the Promise's body. This makes the surrounding code nicer, as everything doesn't have to be stuffed into the block of code passed as the body argument to the promise's constructor.

You can think of `pbResolve` as a function defined like `pbResolve = () => { promise.fulfilled = true; }`, where anything awaiting the promise is constantly executing a `while (!promise.fulfilled) {}` in the background.

src/ui/modifier-select-ui-handler.ts
SirzBenjie: Thanks, was leftover from debugging

src/ui/modifier-select-ui-handler.ts
DayKev: Ah, gotcha.

DayKev:
This seems to re-introduce (?) the crash bug:
<details><summary>Video of crash</summary>
<p>

https://github.com/user-attachments/assets/2516ee97-973e-4240-ad07-ccd0dfe60d30

</p>
</details> 

Video: Game speed of 3x, just holding A with the default cursor position set to "Reroll".

It still happens with game speed at 5x though, I just had it at 3x while I was testing so that I could more easily see the animations.

SirzBenjie:
> This seems to re-introduce (?) the crash bug:
> 
> Video of crash
> 
>  reroll.crash.mp4 
> Video: Game speed of 3x, just holding A with the default cursor position set to "Reroll".
> 
> It still happens with game speed at 5x though, I just had it at 3x while I was testing so that I could more easily see the animations.

Forgot to make one of the tweens (specifically the tween responsible for the pokeball bounce) also prevent the promise from resolving.

Also, I forgot to use the spread syntax on one of the arrays of promises passed to `Promise.allSettled()`. Though, it seems that, at least on chromium, that method will flatten out nested arrays, since the promises inside were properly being awaited. Strange.

DayKev (dismissed):

src/ui/modifier-select-ui-handler.ts
```diff
@@ -820,14 +844,45 @@ class ModifierOption extends Phaser.GameObjects.Container {
...
-    if (!this.modifierTypeOption.cost) {
+  /**
+   * Start the tweens responsible for animating the option's appearance
+   *
+   * @privateremarks
+   * This method is unusual. It "returns" (one via the actual return, one by via appending to the `promiseHolder`
+   * parameter) two promises. The promise returned by the method resolves once the option's appearance animations have
+   * completed, and is meant to allow callers to synchronize with the completion of the option's appearance animations.
+   * The promise appended to `promiseHolder` resolves once *all* animations started by this method have completed,
+   * ans should be used by callers to ensure that all animations have completed before proceeding.
```
Bertie690:
```suggestion
   * and should be used by callers to ensure that all animations have completed before proceeding.
```

src/ui/modifier-select-ui-handler.ts
```diff
@@ -820,14 +844,45 @@ class ModifierOption extends Phaser.GameObjects.Container {
...
+   *
+   * @param remainingDuration - The duration in milliseconds that the animation can play for
+   * @param upgradeCountOffset - The offset to apply to the upgrade count for options whose rarity is being upgraded
+   * @param promiseHolder - A promise that resolves once all tweens started by this method have completed will be pushed to this array.
+   * @param isReward - Whether the option being shown is a reward, meaning it should show pokeball and upgrade animations.
+   * @returns A promise that resolves once the *option's apperance animations* have completed. This promise will resolve _before_ all
+   *   promises that are initiated in this method complete. Instead, the `promiseHolder` array will contain a new promise
+   *   that will resolve once all animations have completed.
```
Bertie690, line 863:
```suggestion
   * promises that are initiated in this method complete. Instead, the `promiseHolder` array will contain a new promise
   * that will resolve once all animations have completed.
```

Bertie690 (dismissed):
lgtm

We really need to rip apart the ui system lol

src/ui/modifier-select-ui-handler.ts
SirzBenjie: Technically, you're supposed to indent continuations like this.

DayKev (approved):

damocleas (approved):

Bertie690 (approved):

PR #6124, opened 2025-07-21
[Dev] Disabled biome `noAwaitInLoop` rule

Bertie690:
N/A
Benjie mentioned wanting to disable the rule.
We use promises sparingly enough that the places where we _do_ use await in a for loop, there's usually a pretty good reason.
Also it becomes an especially large nuisance in tests (since advancing to phases is asynchronous)

Disabled rule

## Screenshots/Videos
N/A

## How to test the changes?
N/A

## Checklist
- [x] Have I tested the changes manually?x

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -139,7 +139,7 @@
         "useAdjacentGetterSetter": "error",
         "noConstantBinaryExpression": "error",
         "noTsIgnore": "error",
-        "noAwaitInLoop": "warn",
+        "noAwaitInLoop": "off",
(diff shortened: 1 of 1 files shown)


DayKev (approved):

SirzBenjie (approved):

SirzBenjie:
It's less so that we use promises sparingly enough, but rather, there's many times where we use promises in a loop.

I think the rule is more meant for things like fetching content inside of a loop, which is generally undesirable. (Like, fetching some resource over http).
However, for us, we use promises for things like waiting for events to complete. Like the egg gacha ui creates the tweens for the egg drops and awaits their animations to complete before beginning the next egg drop animation.

PR #6129, opened 2025-07-23
[Refactor] Minor refactor of battler tags

SirzBenjie:
Users that had an existing session save with a Commanding Dondozo with a Tatsugiri that was not curly form, will see that Order up causes ATK to be raised instead of whatever stat should be raised from the original Tatsugiri form. (This is so niche, and a one-off that I deemed such "breakage" inconsequential.
Users' session save data will now take marginally less space, as fields in BattlerTags that are not necessary to serialize will no longer be serialized.

Needed in preparation to make save migration type safe.

See #6110; these changes apply the same principals, instead to `BattlerTags`

- Adds a new abstract type, `SerializableBattleTag`. All tags that can persist across waves must extend this class.
- Cleaned up the existing TSDoc in the file
- Removed `loadTag` methods from several BattlerTags that were persisting data that should have always been set by the constructor.
- Added several utility types, similar to #6110, to help distinguish between certain battler tags.
- Added a filter when creating the BattlerTags for pokemon data to ensure that only `Serializable` tags are transferred.
- Made the No Retreat move look at the battler tag type instead of the source move's ID.
- Made `isBatonPassable` and `lapseTypes` true private fields of the `BattlerTag` constructor, and converted the existing field names into a getter. These fields should never be persisted or modified outside of the constructors.
- Renamed `_tatsugiriFormKey` to `tatsugiriFormKey` and made it a public (externally immutable, internally mutable, via readonly) member, as the field needs to be both serialized and visible to typescript outside of the class. Having a public property prepended with a `_` just doesn't make sense. Rather than adjust our save migrator for this change to rename fields of incoming sessions, I left it as is. Order up is the only thing that cares about the tatsugiri form.

Essentially, properties that can be determined solely by the tag type, and never change once the tag was added with a particular type, need not be serialized. We have no way of enforcing this, though in the future, reviews of PRs adding / modifying `BattlerTags` should take special care to consider the visibility given to fields, and whether or not it makes sense to construct it.
Easy test: if a field is only ever set in the constructor (and/or the load tags method), then it should instead be a `#` private field. (This is the most ergonomic way to avoid serialization).

I did not adjust saving code, so nothing should break. This PR was almost entirely just improving the types around battler tags, and then also taking a careful look at what is and isn't being serialized, and adjusting accordingly.

## Screenshots/Videos
N/A

## How to test the changes?
Start up a save, and use moves like substitute, get hit by tar shot, etc. Ensure that tags are loaded properly if you save and quit and then reload.

## Checklist


src/data/battler-tags.ts
```diff
@@ -154,7 +210,7 @@ export abstract class MoveRestrictionBattlerTag extends BattlerTag {
   }
 
   /**
-   * Gets whether this tag is restricting a move.
+   * Determine whether this a move's usage is restricted by this tag
```
Wlowscha:
```suggestion
   * Determine whether a move's usage is restricted by this tag
```

src/data/battler-tags.ts
```diff
@@ -2478,15 +2483,16 @@ export class RoostedTag extends BattlerTag {
...
+  constructor(tagType: BattlerTagType.ICE_FACE | BattlerTagType.DISGUISE) {
     super(tagType, BattlerTagLapseType.CUSTOM, 1);
   }
 
   /**
    * Determines if the tag can be added to the Pokémon.
-   * @param {Pokemon} pokemon The Pokémon to which the tag might be added.
-   * @returns {boolean} True if the tag can be added, false otherwise.
+   * @param pokemon - The Pokémon to which the tag might be added.
+   * @returns True if the tag can be added, false otherwise.
```
Wlowscha:
```suggestion
   * @returns `true` if the tag can be added, `false` otherwise.
```

src/data/battler-tags.ts
```diff
@@ -2508,20 +2514,22 @@ export class FormBlockDamageTag extends BattlerTag {
...
+
 /** Provides the additional weather-based effects of the Ice Face ability */
 export class IceFaceBlockDamageTag extends FormBlockDamageTag {
+  public override readonly tagType = BattlerTagType.ICE_FACE as const;
   /**
    * Determines if the tag can be added to the Pokémon.
-   * @param {Pokemon} pokemon The Pokémon to which the tag might be added.
-   * @returns {boolean} True if the tag can be added, false otherwise.
+   * @param pokemon - The Pokémon to which the tag might be added.
+   * @returns True if the tag can be added, false otherwise.
```
Wlowscha:
```suggestion
   * @returns `true` if the tag can be added, `false` otherwise.
```

src/@types/battler-tags.ts
```diff
@@ -0,0 +1,88 @@
+import type { BattlerTagTypeMap, SerializableBattlerTag } from "#data/battler-tags";
+import type { BattlerTagType } from "#enums/battler-tag-type";
+
+/**
+ * Subset of {@linkcode BattlerTagType}s that restricts the use of moves.
```
Bertie690:
```suggestion
 * Subset of {@linkcode BattlerTagType}s that restrict the use of moves.
```

src/@types/battler-tags.ts
```diff
@@ -0,0 +1,88 @@
...
+  | BattlerTagType.DISABLED
+  | BattlerTagType.GORILLA_TACTICS;
+
+/**
+ * Subset of {@linkcode BattlerTagType}s that block damage from moves.
+ */
+export type FormBlockDamageBattlerTagType = BattlerTagType.ICE_FACE | BattlerTagType.DISGUISE;
+
+/**
+ * Subset of {@linkcode BattlerTagType}s that are related to trapping effects.
```
Bertie690, line 26:
```suggestion
 * Subset of {@linkcode BattlerTagType}s that trap the target on field and prevent switching.
```

src/@types/battler-tags.ts
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Subset of {@linkcode BattlerTagType}s that are not able to persist across waves and should therefore not be serialized
+ */
+export type NonSerializableBattlerTagType = Exclude<BattlerTagType, SerializableBattlerTagType>;
+
+/**
+ * Dummy, typescript-only declaration to ensure that
+ * {@linkcode BattlerTagTypeMap} has an entry for all `BattlerTagType`s.
+ *
+ * If battler tag is missing from the map, typescript will throw an error on this statement.
```
Bertie690:
```suggestion
 * If a battler tag is missing from the map, Typescript will throw an error on this statement.
```

src/data/battler-tags.ts
```diff
@@ -31,41 +32,99 @@ import type { MoveEffectPhase } from "#phases/move-effect-phase";
...
+
 /**
  * A {@linkcode BattlerTag} represents a semi-persistent effect that can be attached to a {@linkcode Pokemon}.
  * Tags can trigger various effects throughout a turn, and are cleared on switching out
  * or through their respective {@linkcode BattlerTag.lapse | lapse} methods.
  */
-export class BattlerTag {
-  public tagType: BattlerTagType;
-  public lapseTypes: BattlerTagLapseType[];
+export class BattlerTag implements BaseBattlerTag {
```
Bertie690, line 89: it might be a bit late on this, but shouldn't `BattlerTag` be abstract or smthng?

src/data/battler-tags.ts
```diff
@@ -154,7 +210,7 @@ export abstract class MoveRestrictionBattlerTag extends BattlerTag {
   }
 
   /**
-   * Gets whether this tag is restricting a move.
+   * Determine whether a move's usage is restricted by this tag
```
Bertie690, line 221:
```suggestion
   * Determine whether a move's usage is restricted by this tag.
```

src/data/battler-tags.ts
```diff
@@ -163,23 +219,23 @@ export abstract class MoveRestrictionBattlerTag extends BattlerTag {
...
 
   /**
-   * Checks if this tag is restricting a move based on a user's decisions during the target selection phase
+   * Check if this tag is restricting a move based on a user's decisions during the target selection phase
    *
-   * @param {MoveId} _move {@linkcode MoveId} move ID to check restriction for
-   * @param {Pokemon} _user {@linkcode Pokemon} the user of the above move
-   * @param {Pokemon} _target {@linkcode Pokemon} the target of the above move
-   * @returns {boolean} `false` unless overridden by the child tag
+   * @param _move - {@linkcode MoveId} to check restriction for
```
Bertie690, line 232:
```suggestion
   * @param _move - The {@linkcode MoveId} being checked
```

src/data/battler-tags.ts
```diff
@@ -154,7 +210,7 @@ export abstract class MoveRestrictionBattlerTag extends BattlerTag {
   }
 
   /**
-   * Gets whether this tag is restricting a move.
+   * Determine whether a move's usage is restricted by this tag
    *
    * @param move - {@linkcode MoveId} ID to check restriction for.
```
```suggestion
   * @param move - The {@linkcode MoveId} being checked
```

src/data/battler-tags.ts
```diff
@@ -163,23 +219,23 @@ export abstract class MoveRestrictionBattlerTag extends BattlerTag {
...
 
   /**
-   * Gets the text to display when the player attempts to select a move that is restricted by this tag.
+   * Get the text to display when the player attempts to select a move that is restricted by this tag.
    *
-   * @param {Pokemon} pokemon {@linkcode Pokemon} for which the player is attempting to select the restricted move
-   * @param {MoveId} move {@linkcode MoveId} ID of the move that is having its selection denied
-   * @returns {string} text to display when the player attempts to select the restricted move
+   * @param pokemon - The pokemon for which the player is attempting to select the restricted move
+   * @param move - {@linkcode MoveId | ID} of the move that is having its selection denied
```
```suggestion
   * @param move - The {@linkcode MoveId | ID} of the Move that is having its selection denied
```

src/data/battler-tags.ts
```diff
@@ -188,9 +244,9 @@ export abstract class MoveRestrictionBattlerTag extends BattlerTag {
    * Because restriction effects also prevent selection of the move, this situation can only arise if a
    * pokemon first selects a move, then gets outsped by a pokemon using a move that restricts the selected move.
    *
-   * @param {Pokemon} _pokemon {@linkcode Pokemon} attempting to use the restricted move
-   * @param {MoveId} _move {@linkcode MoveId} ID of the move being interrupted
-   * @returns {string} text to display when the move is interrupted
+   * @param _pokemon - The pokemon attempting to use the restricted move
+   * @param _move - {@linkcode MoveId | ID} of the move being interrupted
```
```suggestion
   * @param _move - The {@linkcode MoveId | ID} of the move being interrupted
```

src/data/battler-tags.ts
```diff
@@ -309,16 +362,14 @@ export class DisabledTag extends MoveRestrictionBattlerTag {
...
     return i18next.t("battle:moveDisabled", { moveName: allMoves[move].name });
   }
 
   /**
-   * @override
-   * @param {Pokemon} pokemon {@linkcode Pokemon} attempting to use the restricted move
-   * @param {MoveId} move {@linkcode MoveId} ID of the move being interrupted
-   * @returns {string} text to display when the move is interrupted
+   * @param pokemon - {@linkcode Pokemon} attempting to use the restricted move
+   * @param move - {@linkcode MoveId | ID} of the move being interrupted
```
Bertie690, line 379:
```suggestion
   * @param move - The {@linkcode MoveId | ID} of the move being interrupted
```

src/data/battler-tags.ts
```diff
@@ -327,18 +378,21 @@ export class DisabledTag extends MoveRestrictionBattlerTag {
...
+  override loadTag(source: NonFunctionProperties<DisabledTag>): void {
     super.loadTag(source);
-    this.moveId = source.moveId;
+    (this as Mutable<this>).moveId = source.moveId;
   }
 }
 
 /**
  * Tag used by Gorilla Tactics to restrict the user to using only one move.
+ *
```
Bertie690, line 397:
```suggestion
 * @todo Split attack boost into a separate attribute on Gorilla Tactics and remove this tag on ability loss
```

src/data/battler-tags.ts
```diff
@@ -430,6 +484,8 @@ export class RechargingTag extends BattlerTag {
  * @see {@link https://bulbapedia.bulbagarden.net/wiki/Beak_Blast_(move) | Beak Blast}
  */
 export class BeakBlastChargingTag extends BattlerTag {
+  public override readonly tagType = BattlerTagType.BEAK_BLAST_CHARGING as const;
```
```suggestion
  public override readonly tagType = BattlerTagType.BEAK_BLAST_CHARGING;
```

Idk if the const assertion is necessary

src/data/battler-tags.ts
```diff
@@ -591,9 +648,9 @@ export class TrappedTag extends BattlerTag {
  * BattlerTag implementing No Retreat's trapping effect.
  * This is treated separately from other trapping effects to prevent
  * Ghost-type Pokemon from being able to reuse the move.
- * @extends TrappedTag
  */
 class NoRetreatTag extends TrappedTag {
+  public override readonly tagType = BattlerTagType.NO_RETREAT as const;
```
Bertie690: See above

src/data/battler-tags.ts
```diff
@@ -638,7 +696,9 @@ export class FlinchedTag extends BattlerTag {
   }
 }
 
-export class InterruptedTag extends BattlerTag {
+// TODO: This might not need to be serialized
```
Bertie690: @SirzBenjie Actually, this SHOULD not be serialized - it's a hacky, partially functinoning way to cancel Fly/Bounce when smack down or Thousand Waves are used.

src/data/battler-tags.ts
```diff
@@ -213,21 +270,19 @@ export class ThroatChoppedTag extends MoveRestrictionBattlerTag {
   }
 
   /**
-   * Checks if a {@linkcode MoveId | move} is restricted by Throat Chop.
-   * @override
-   * @param {MoveId} move the {@linkcode MoveId | move} to check for sound-based restriction
-   * @returns true if the move is sound-based
+   * Check if a move is restricted by Throat Chop.
+   * @param move - The {@linkcode MoveId | ID } of the move to check for sound-based restriction
+   * @returns Whether the move is sound based
```
Bertie690, line 283: Should we even have this comment at all?

This looks very much identical to the super class's comment...

src/data/battler-tags.ts
```diff
@@ -213,21 +270,19 @@ export class ThroatChoppedTag extends MoveRestrictionBattlerTag {
...
 
   /**
    * Shows a message when the player attempts to select a move that is restricted by Throat Chop.
-   * @override
-   * @param {Pokemon} _pokemon the {@linkcode Pokemon} that is attempting to select the restricted move
-   * @param {MoveId} move the {@linkcode MoveId | move} that is being restricted
-   * @returns the message to display when the player attempts to select the restricted move
+   * @param _pokemon - The {@linkcode Pokemon} that is attempting to select the restricted move
+   * @param move - The {@linkcode MoveId | move} that is being restricted
+   * @returns The message to display when the player attempts to select the restricted move
```
Bertie690, line 293: Same here - there is like 0 value added from this being present

I'd be worried that it's out-of-scope for this PR, but then again you are cleaning up all the tsdocs so...

src/data/battler-tags.ts
```diff
@@ -237,10 +292,9 @@ export class ThroatChoppedTag extends MoveRestrictionBattlerTag {
 
   /**
    * Shows a message when a move is interrupted by Throat Chop.
-   * @override
-   * @param {Pokemon} pokemon the interrupted {@linkcode Pokemon}
-   * @param {MoveId} _move the {@linkcode MoveId | move} that was interrupted
-   * @returns the message to display when the move is interrupted
+   * @param pokemon - The interrupted {@linkcode Pokemon}
+   * @param _move - The {@linkcode MoveId | ID } of the move that was interrupted
+   * @returns The message to display when the move is interrupted
```
Bertie690, line 305: this is just like near identical as well

src/data/battler-tags.ts
```diff
@@ -993,6 +1057,7 @@ export class SeedTag extends BattlerTag {
  * Pokemon takes damage equal to 1/4 of it's maximum HP (rounded down).
```
```suggestion
 * Pokemon takes damage equal to 1/4 of its maximum HP (rounded down).
```

GRAMMAR GAHHH

src/data/battler-tags.ts
```diff
@@ -1245,16 +1310,16 @@ export class HelpingHandTag extends BattlerTag {
...
 export class IngrainTag extends TrappedTag {
+  public override readonly tagType = BattlerTagType.INGRAIN as const;
   constructor(sourceId: number) {
     super(BattlerTagType.INGRAIN, BattlerTagLapseType.TURN_END, 1, MoveId.INGRAIN, sourceId);
   }
 
   /**
    * Check if the Ingrain tag can be added to the pokemon
-   * @param pokemon {@linkcode Pokemon} The pokemon to check if the tag can be added to
+   * @param pokemon - The pokemon to check if the tag can be added to
```
Bertie690, line 1329: Might be a tad bit similar as above

src/data/battler-tags.ts
```diff
@@ -1405,7 +1474,9 @@ export class DrowsyTag extends BattlerTag {
 }
 
 export abstract class DamagingTrapTag extends TrappedTag {
-  private commonAnim: CommonAnim;
+  public override readonly tagType: TrappingBattlerTagType;
+  /** Animation to play when during the damage sequence */
```
```suggestion
  /** The animation to play during the damage sequence */
```

src/data/battler-tags.ts
```diff
@@ -1744,6 +1806,7 @@ export class ContactSetStatusProtectedTag extends DamageProtectedTag {
    */
   override onContact(attacker: Pokemon, user: Pokemon): void {
     attacker.trySetStatus(this.statusEffect, true, user);
+    this.statusEffect;
```
Bertie690: Looks like a mistake

Bertie690 (changes requested):
I realize I had these here and never submitted them.
Oops

src/@types/battler-tags.ts
SirzBenjie: I don't like this one, because those effects are also responsible for more things like DoT.

src/data/battler-tags.ts
SirzBenjie: I thought that, too. But there are apparently tags that use this class itself. I think that's fine -- we don't need to subclass everything.

src/data/battler-tags.ts
SirzBenjie: Not a complete sentence in the first place.

src/data/battler-tags.ts
SirzBenjie: Ehh.. I think it's fine.

src/data/battler-tags.ts
```diff
@@ -638,7 +696,9 @@ export class FlinchedTag extends BattlerTag {
   }
 }
 
-export class InterruptedTag extends BattlerTag {
+// TODO: This might not need to be serialized
+export class InterruptedTag extends SerializableBattlerTag {
```
SirzBenjie:
```suggestion
export class InterruptedTag extends BattlerTag {
```

src/data/battler-tags.ts
SirzBenjie:
```suggestion
```

src/data/battler-tags.ts
SirzBenjie: It is.

src/data/battler-tags.ts
```diff
@@ -31,41 +32,99 @@ import type { MoveEffectPhase } from "#phases/move-effect-phase";
...
     return true;
   }
 
+  /**
+   * Apply effects that occur when the tag is added to a {@linkcode Pokemon}
+   * @param _pokemon - The {@linkcode Pokemon} the tag was added to
+   */
   onAdd(_pokemon: Pokemon): void {}
 
   onRemove(_pokemon: Pokemon): void {}
```
Bertie690, line 138:
```suggestion
  /**
   * Apply effects when this tag is removed from a Pokemon.
   * @param _pokemon - The {@linkcode Pokemon} having previously held this tag
   */
  onRemove(_pokemon: Pokemon): void {}
```

src/data/battler-tags.ts
```diff
@@ -132,8 +188,8 @@ export interface TerrainBattlerTag {
  * match a condition. A restricted move gets cancelled before it is used.
  * Players and enemies should not be allowed to select restricted moves.
  */
-export abstract class MoveRestrictionBattlerTag extends BattlerTag {
-  /** @override */
+export abstract class MoveRestrictionBattlerTag extends SerializableBattlerTag {
+  public override readonly tagType: MoveRestrictionBattlerTagType;
```
```suggestion
  declare public readonly tagType: MoveRestrictionBattlerTagType;
```

Bertie690: Also make it abstract while you're at it

Bertie690 (dismissed):
LLLGGGTTTMMEEEEE

src/data/battler-tags.ts
SirzBenjie: Can't have abstract and declare..

Wlowscha (dismissed):

Bertie690 (dismissed):

src/data/battler-tags.ts
```diff
@@ -430,6 +484,8 @@ export class RechargingTag extends BattlerTag {
  * @see {@link https://bulbapedia.bulbagarden.net/wiki/Beak_Blast_(move) | Beak Blast}
  */
 export class BeakBlastChargingTag extends BattlerTag {
+  public override readonly tagType = BattlerTagType.BEAK_BLAST_CHARGING;
+  override readonly sourceMove: MoveId.BEAK_BLAST;
```
DayKev:
```suggestion
  declare readonly sourceMove: MoveId.BEAK_BLAST;
```

src/data/battler-tags.ts
```diff
@@ -1649,14 +1719,14 @@ export class ProtectedTag extends BattlerTag {
...
-export class ContactProtectedTag extends ProtectedTag {
+export abstract class ContactProtectedTag extends ProtectedTag {
   /**
    * Function to call when a contact move hits the pokemon with this tag.
    * @param _attacker - The pokemon using the contact move
    * @param _user - The pokemon that is being attacked and has the tag
    * @param _move - The move used by the attacker
    */
-  onContact(_attacker: Pokemon, _user: Pokemon) {}
+  abstract onContact(_attacker: Pokemon, _user: Pokemon);
```
DayKev: Missing return type (presumably `void`?).

src/data/battler-tags.ts
```diff
@@ -2065,43 +2110,32 @@ export class HighestStatBoostTag extends AbilityBattlerTag {
...
-  loadTag(source: BattlerTag | any): void {
-    super.loadTag(source);
-    this.terrainTypes = source.terrainTypes.map(w => w as TerrainType);
+    this.#terrainTypes = terrainTypes;
   }
 }
 
-export class SemiInvulnerableTag extends BattlerTag {
+export class SemiInvulnerableTag extends SerializableBattlerTag {
+  override tagType: SemiInvulnerableBattlerTagType;
```
DayKev:
```suggestion
  declare tagType: SemiInvulnerableBattlerTagType;
```

src/data/battler-tags.ts
```diff
@@ -2338,22 +2352,11 @@ export class SaltCuredTag extends BattlerTag {
   }
 }
 
-export class CursedTag extends BattlerTag {
-  private sourceIndex: number;
-
+export class CursedTag extends SerializableBattlerTag {
```
DayKev, line 2379: Was `sourceIndex` redundant lol.

src/data/battler-tags.ts
```diff
@@ -2478,15 +2481,16 @@ export class RoostedTag extends BattlerTag {
 }
 
 /** Common attributes of form change abilities that block damage */
-export class FormBlockDamageTag extends BattlerTag {
-  constructor(tagType: BattlerTagType) {
+export class FormBlockDamageTag extends SerializableBattlerTag {
+  public override readonly tagType: BattlerTagType.ICE_FACE | BattlerTagType.DISGUISE;
```
DayKev:
```suggestion
  public declare readonly tagType: BattlerTagType.ICE_FACE | BattlerTagType.DISGUISE;
```

src/data/battler-tags.ts
```diff
@@ -2687,10 +2691,10 @@ export class StockpilingTag extends BattlerTag {
 
 /**
  * Battler tag for Gulp Missile used by Cramorant.
- * @extends BattlerTag
  */
-export class GulpMissileTag extends BattlerTag {
-  constructor(tagType: BattlerTagType, sourceMove: MoveId) {
+export class GulpMissileTag extends SerializableBattlerTag {
+  public override readonly tagType: BattlerTagType.GULP_MISSILE_ARROKUDA | BattlerTagType.GULP_MISSILE_PIKACHU;
```
DayKev:
```suggestion
  public declare readonly tagType: BattlerTagType.GULP_MISSILE_ARROKUDA | BattlerTagType.GULP_MISSILE_PIKACHU;
```

src/data/battler-tags.ts
```diff
@@ -2755,38 +2760,33 @@ export class GulpMissileTag extends BattlerTag {
...
  *
  * Applied by moves: {@linkcode MoveId.ODOR_SLEUTH | Odor Sleuth},
  * {@linkcode MoveId.MIRACLE_EYE | Miracle Eye} and {@linkcode MoveId.FORESIGHT | Foresight}.
  *
- * @extends BattlerTag
  * @see {@linkcode ignoreImmunity}
  */
-export class ExposedTag extends BattlerTag {
+export class ExposedTag extends SerializableBattlerTag {
+  public override readonly tagType: BattlerTagType.IGNORE_DARK | BattlerTagType.IGNORE_GHOST;
```
DayKev:
```suggestion
  public declare readonly tagType: BattlerTagType.IGNORE_DARK | BattlerTagType.IGNORE_GHOST;
```

src/data/battler-tags.ts
```diff
@@ -2755,38 +2760,33 @@ export class GulpMissileTag extends BattlerTag {
...
   private defenderType: PokemonType;
-  private allowedTypes: PokemonType[];
+  private allowedTypes: readonly PokemonType[];
```
DayKev: Why `x: readonly y[]` and not `readonly x: y[]`?

SirzBenjie: Actually, should be both.

It's a hint that says "hey, you shouldn't be modifying the contents of this array".

src/data/battler-tags.ts
SirzBenjie: Yes, completely.

DayKev (dismissed):

Wlowscha (dismissed):

Bertie690 (dismissed):

src/data/battler-tags.ts
```diff
@@ -31,41 +32,106 @@ import type { MoveEffectPhase } from "#phases/move-effect-phase";
...
+  MoveRestrictionBattlerTagType,
+  ProtectionBattlerTagType,
+  SemiInvulnerableTagType,
+  TrappingBattlerTagType,
+} from "#types/battler-tags";
+import type { Mutable, NonFunctionProperties } from "#types/type-helpers";
 import { BooleanHolder, coerceArray, getFrameMs, isNullOrUndefined, NumberHolder, toDmgValue } from "#utils/common";
 
+/*
+BattlerTags are used to represent semi-persistent effects that can be attached to a Pokemon.
```
Bertie690, line 52:
```suggestion
@module
BattlerTags are used to represent semi-persistent effects that can be attached to a Pokemon.
```

src/data/battler-tags.ts
```diff
@@ -31,41 +32,106 @@ import type { MoveEffectPhase } from "#phases/move-effect-phase";
...
+
+Serializable battler tags have strict requirements for their fields.
+Properties that are not necessary to reconstruct the tag must not be serialized. This can be avoided
+by using a private property. If access to the property is needed outside of the class, then
+a getter (and potentially, a setter) should be used instead.
+
+If a property that is intended to be private must be serialized, then it should instead
+be declared as a public readonly propety. Then, in the `loadTag` method (or any method inside the class that needs to adjust the property)
+use `(this as Mutable<this>).propertyName = value;`
+These rules ensure that typescript is aware of the shape of the serialized version of the class.
```
```suggestion
These rules ensure that Typescript is aware of the shape of the serialized version of the class.
```

PR #5747, opened 2025-05-02
[Refactor] File name changes to kebab-case

pymilkmaiden:
## What are the changes the user will see?
No visible changes from user perspective

## Why am I making these changes?
 Better consistency for file names across the project

## What are the changes from a developer perspective?
File names in src & test will be changed to kebab-case for simpler development. Changes from previous PR now on a branch instead of beta to make managing changes easier.

## Screenshots/Videos
N/A

## How to test the changes?
N/A

## Checklist
- [ ] **I'm using `beta` as my base branch**
- [ ] There is no overlap with another PR?
- [ ] The PR is self-contained and cannot be split into smaller PRs?
- [ ] Have I provided a clear explanation of the changes?
- [ ] Have I tested the changes manually?
- [ ] Are all unit tests still passing? (`npm run test:silent`)

PR #6137, opened 2025-07-24
[Misc] Standardize-file-names

SirzBenjie:
None

Consistency in the codebase. Successor to #5747

- Renames (almost) all files and directories to use kebab-case
- Moved two scripts into the `scripts` folder
- Added a new CI and lefthook lint rule that runs `ls-lint`, a new package, that enforces our filenaming conventions
- Deleted an extraneous json file, `src/ui/settings/shiny_icons.json` that seems leftover (There's an exact copy inside of `public`)

`ls-lint` now ensures consistent file naming conventions inside of our codebase

- directories, along with typescript, javascript, and powershell files, have kebab-case enforced
- python uses snake_case
- `src` is now prohibited from having `js` files (not that someone would do this anyway, but it was an easy rule to add).
- A few exceptions to the above, such as the names of the version migration files, were added.
- Replaced `class` with `interface` in a few of the api files in `@typings`, because these really are just interfaces.

## Screenshots/Videos
N/A

## How to test the changes?
First, use `pnpm i` to install the `ls-lint` package.

Then, add some file that goes against naming conventions, like `src/camelCase.ts`

A few things to test:
`pnpm ls-lint` should error
Attempting to create a commit with that file should error.

## Checklist


.ls-lint.yml
```diff
@@ -0,0 +1,27 @@
+# Base settings to use
+# Note that the `_cfg` key isn't part of ls-lint's configuration, it's just a YAML anchor for reuse.
+_cfg: &cfg
+   .ps1: kebab-case
+   .ts: kebab-case
+   .js: kebab-case
+   .*.ts: kebab-case
+   .*.js: kebab-case
+   .dir: kebab-case
+   .py: snake_case # python files should always use snake_case
```
DayKev, line 10: Why the difference for Python files?

SirzBenjie: because you can't use `import python-module` in valid python code.
You shouldn't ever use kebab-case for a python module.

lefthook.yml
```diff
@@ -6,6 +6,8 @@ pre-commit:
       skip:
         - merge
         - rebase
+    ls-lint:
+      run: pnpm exec ls-lint
```
Bertie690, line 10: IDK if this is affected by the "skip on merge/rebase" part above (wary of breaking merges/rebases)

Bertie690 (dismissed):
lgtm.

let's get this bombshell of a PR over with shall we?

.ls-lint.yml
DayKev: Interesting.

Wlowscha (approved):

.ls-lint.yml
```diff
@@ -0,0 +1,27 @@
...
+
+ls:
+   <<: *cfg
+   src:
+      <<: *cfg
+      .dir: kebab-case | regex:@types
+      .js: exists:0
+   src/system/version-migration/versions:
+      .ts: snake_case
```
DayKev, line 19: Oh huh, what's this for?

Wlowscha: I guess it's because we don't want the version numbers to be 1-9-0

SirzBenjie: Yeah, it looks weird to do that.

DayKev: Oh good point.

Bertie690 (commented):
LGTM BEFORE MY COMPUTER EXPLODES AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA

Bertie690 (approved):

PR #6142, opened 2025-07-25
refactor(ui): clean up arena flyout code

### LINKED ISSUE #5150: [Bug] Training Session Move Flyout Goes Negative when Move has been Affected by PP Up, PP Max
### Describe the bug

The Enemy's Move Flyout in Training Session can display negative PP if the Pokemon selected's move had been affected by a PP Up or PP Max.
-The move initially shows as having as much PP as had already been used from the move.

Link to Discord's Bug Report (Training Session IV Option Bugged (Negative PP))
https://discord.com/channels/1125469663833370665/1290074317358370906

### Reproduction

1. Use the PP of a move whose PP has been improved by a PP Up or PP Max.
2. Select the Pokemon whose PP has been used.
3. Have the selected Pokemon use the Move until that move would run out of PP normally.

### Expected behavior

PP value should not be somewhat retained on the Move Flyout in Training Session, and PP should never be negative.

### Screenshots / Videos

https://github.com/user-attachments/assets/9704b612-acee-4c8a-82ce-17d5687d8c0e

### Session export file

[sessionData1_Marinnawith2n.txt](https://github.com/user-attachments/files/18467549/sessionData1_Marinnawith2n.txt)

### User data export file

_No response_

### Additional context

_No response_

Bertie690:
~~[Locales PR](https://github.com/pagefaultgames/pokerogue-locales/pull/205)~~ :(

Weather and terrain will consistently appear at the top of the arena flyout.

(Local dev only, don't put in changelog)
Overridding weathers now show up on the flyout correctly

I tried to add FS/DD to the flyout, refactored the entire thing to make it easier, then realized the text was STILL too large and scrapped that section of the code.

That said, the previous code was absolutely terrible - we stored weather, terrain and arena tags all in the same place with the same kind of data structure and used the same callback for all 4 event types, resulting in the monstrosity that is
```ts
  switch (fieldEffectInfo.constructor) {
```

1. The unholy mess that was `onFieldEffectChanged` has been broken up into variants for each type of emitted event. These are stored inside private element callbacks to ensure that `this` inside the functions remains as the type of `ArenaFlyout` and not `EventEmitter`.  
2. Weather and terrain have their own info interfaces and are now stored independently of the current arena tags.
   - This lets us order them separately on top of the flyout and remove unused fields only used for culling prior terrain/weather data.
3. 

I also added strong typing on the types of events dispatched by `globalScene.arena.eventTarget` by lieu of a generic interface, `TypedEventHandler`, that overrides the types of event dispatching methods to a subset of allowed methods.

Added many doc comments and encapsulations into methods

It amy be worth revisiting re-adding FS/DD once <https://github.com/pagefaultgames/pokerogue/pull/6042> becomse ready, or potentially once the modal is refactored to become a proper menu.

## Screenshots/Videos

https://github.com/user-attachments/assets/f78e7da6-7ec0-44da-b77f-ab77eab5ae0a

IGNORE THE COURT CHANGE BUGS - THOSE ARE BUGS IN THE MOVE ITSELF
## How to test the changes?
I added a few tests for the thing!

## Checklist


Bertie690:
Just for the record, I am placing the patch with the FS/DD stuff
[here for posterity](https://github.com/user-attachments/files/21783397/diff.patch) in case we want to re-apply it later.

src/@types/typed-event-target.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Interface restricting the events emitted by an {@linkcode EventTarget} to a certain kind of {@linkcode Event}.
+ * @typeParam T - The type to restrict the interface's access; must extend from {@linkcode Event}
+ */
+export interface TypedEventTarget<T extends Event = never> extends EventTarget {
```
SirzBenjie: What does defaulting to `never` accomplish? Does that mean that it prevents any events?

SirzBenjie (dismissed):

src/@types/typed-event-target.ts
Bertie690: It requires an explicit type argument.

Wlowscha (changes requested):
The code looks good, but since this touches UI elements, there should be some pictures or videos attached.

Bertie690:
I tested and it still works
_sigh_ will scrounge u basic footage later

src/@types/typed-event-target.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Interface restricting the events emitted by an {@linkcode EventTarget} to a certain kind of {@linkcode Event},
+ * discriminated by their `type` properties.
+ * @typeParam K - A union of string literal types representing the allowed event names.
+ * @typeParam R - An object type matching event names in `K` to their corresponding `Event` subclasses.
+ */
+// TODO: Should we rework this to use `CustomEvent`s instead of separate subclasses?
+// @ts-expect-error: We are overridding these types in a way TS doesn't enjoy
```
SirzBenjie: Can you explain the issue?

src/enums/arena-event-type.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { WeatherType } from "#enums/weather-type";
+import type { ArenaEvent } from "#events/arena";
+import type { ObjectValues } from "#types/type-helpers";
+
+/**
+ * Doc comment removal prevention block
+ * {@linkcode WeatherType}
+ * {@linkcode TerrainType}
+ * {@linkcode ArenaTag}
+ */
```
SirzBenjie, line 13: What is this for?

src/events/arena.ts
```diff
@@ -1,98 +1,162 @@
...
-  /** The {@linkcode ArenaTagType} being added */
-  public arenaTagType: ArenaTagType;
-  /** The {@linkcode ArenaTagSide} the tag is being placed on */
-  public arenaTagSide: ArenaTagSide;
-  /** The current number of layers of the arena trap. */
-  public arenaTagLayers: number;
-  /** The maximum amount of layers of the arena trap. */
-  public arenaTagMaxLayers: number;
+/**
+ * Container class for {@linkcode ArenaEventType.ARENA_TAG_ADDED} events. \
```
SirzBenjie, line 90:
```suggestion
 * Container class for {@linkcode ArenaEventType.ARENA_TAG_ADDED} events.
 *
```

SirzBenjie (commented):
Has the interface for these weather events changed at all?

src/@types/typed-event-target.ts
Bertie690: Contravariance - TS expects this to work with ANY AND ALL event names since the base class accepts anything as an event name. In reality, we are comstraining the events permitted to a subset of allowed events.

This only started causing errors after `strictFunctionTypes` was introduced, so I really wasn't terribly surprised that it started breaking.

src/enums/arena-event-type.ts
Bertie690: TS doesn't see these as "used" if the only reference is in a JSDoc for a _property_.
Stupid, I know, but c'est la vie.

src/events/arena.ts
Bertie690: imo soft linebreak makes more sense here since the statements are part of 1 complete thought (instead of 2 completely separate paragraphs).

Bertie690:
Not that I can recall.
It's mostly just not using the same interface for 3 very different things lol

SirzBenjie (approved):

PR #6143, opened 2025-07-25
feat: refactor battle move flyout, add support for mid-battle moveset changes

### LINKED ISSUE #5150: [Bug] Training Session Move Flyout Goes Negative when Move has been Affected by PP Up, PP Max

Bertie690:
The battle move flyout now tracks PP ups correctly, as well as mid-battle changes to movesets.
This includes:
 - Sketch
 - Leveling up and replacing a move (currently useless since we don't track player moves used in the flyout, but in theory could be used for a revamped enemy AI)
 - Mimic
 - Transform/Imposter

The last 2 of the above effects (due to being temporary) will reset once the target's `SummonData` is cleared (i.e. once it switches out or faints).

Furthermore, if a Pokemon transforms into an opposing Pokemon, the enemy's current flyout info will be revealed to the player, similar to how Showdown will reveal the moveset of any opponent you successfully Transform into.

Fixes #5150

Not only is this nice QoL and fixes a longstanding bug, the information added by this PR could (in theory) be used by a prospective trainer AI or other system to better gauge the revealed moves of its opponent.

The relevant information would almost certainly be stored somewhere else, but the new event listeners can be hooked into relatively easily if the need arises.

### `BattleFlyout`

1. A new `BattleSceneEvent`, `MovesetChangedEvent` has been added.
This takes a `PokemonMove` (instead of a generic `Move` instance), and is emitted whenever a `Pokemon` has its moveset changed in any way (both from effects like Copycat/etc and PP consumption/restoration). Depending on whether the attached `PokemonMove` is found in its normal or temporary movesets, the info will be added to the normal info or temporary info arrays respectively.
   - This serves as a drop-in replacement for `BerryUsedEvent` (which was _only_ used to track Leppa Berry PP restoration) and `MoveUsedEvent` (which was effectively just a "consume PP" event with a really misleading name).
2. The `BattleFlyout`'s move info array has been split in 2, now consisting of 2 parallel sparse arrays containing data for all moves having been used by or added to the attached Pokemon's moveset. The former tracks the Pokemon's regular moves (Sketch, etc.), while the latter is used for moves found exclusively in the temporary `SummonData` moveset (such as by Transform/Mimic).
3. Added a new event emitter to clear a Pokemon's battle flyout's temporary moveset information when its summon data is cleared. This ensures that temporary move info from Transform and Mimic will be retained only until their effects wear off.
4. Added code to `PokemonTransformPhase` to copy move flyout data upon transforming into an opponent.

## Screenshots/Videos

https://github.com/user-attachments/assets/a1627c0c-4624-4131-aa43-94c44fa9dbea

## How to test the changes?
Use local overrides and verify all of the following work:
1. Normal moves update the flyout and are kept after switching
2. Enemies transforming into player pokemon updates the flyout with the player's current moveset
3. Mimic updates the flyout with the learned move temporarily until leaving the field
4. Sketch updates the _permanent_ flyout

2, 3 and 4 would be easiest to test by using Whirlwind or similar to force the enemy out in a 2 mon trainer battle.

I have not had the time to test this recently since last beta merge, so any help would be greatly recommended

## Checklist


SirzBenjie (dismissed):
Events for UI seem to be a really good pattern.

SirzBenjie:
I assume you tested for the eternatus fight? (Can you?)

Bertie690:
what eternatus fight

src/ui/containers/battle-flyout.ts
```diff
@@ -1,33 +1,41 @@
...
 interface MoveInfo {
-  /** The {@linkcode Move} itself */
-  move: Move;
-
-  /** The maximum PP of the {@linkcode Move} */
-  maxPp: number;
-  /** The amount of PP used by the {@linkcode Move} */
-  ppUsed: number;
+  /** The name of the {@linkcode Move} having been used. */
+  name: string;
```
DayKev: Why not just get this from the stored move?

Bertie690: good idea

fabske0:
The flyout doesn't appear at all after the last commit.

image

</details>

Bertie690:
AFHFEJEJFHFHENEJ

PR #6147, opened 2025-07-26
[Dev] Added typedoc deployments for Beta

### LINKED ISSUE #6049: [Feature] Typedoc docs should use `beta` instead of `main`

Given that beta is the intended target for new PRs, it only makes sense that we target it when creating the typedocs as main can be upwards of a month outdated.

### Describe the Feature

Fix config and sych eh fb fhrheheje

### Additional context

Could likely be made into a versioning system with a plugin

Bertie690:
WE HAVE TYPEDOCS FOR BETA 🎉 
https://pagefaultgames.github.io/pokerogue/beta/index.html

I _wanted_ to test that changing the `deploy-docs` action to work on pushes/commits to beta wouldn't break anything... but apparently `act` just pushed to the branch regardless despite running within a Docker container.
Oops!

Fixes #6049 

Updated the workflow to also trigger on pushes to `beta`, creating docs in the `beta` folder.

## How to test the changes?
LOOK AT THE SIDE

READ IT WITH YOUR EEEYEES

## Checklist

### DIFF
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - beta
   pull_request:
     branches:
       - main


Bertie690:
My guess is that since I had write permissions to the branch, the locally ran github action just took those and ran with it (that or it took whatever secrets granted it write access to said branch).
Either way, it works.

SirzBenjie (approved):

SirzBenjie:
Would like this merged ASAP. I want to test something.

PR #6152, opened 2025-07-26
[Dev] Add `workflow-dispatch` trigger to tests github workflow

Bertie690:
N/A
Adding a `workflow-dispatch` trigger to the tests workflow will allow anyone with repository write permissions to manually schedule a test run on a branch, even if a PR has not been created yet.
While you can always re-run existing workflow failures, that does require you to make a PR.
This _should_ allow for easier testing of tentative changes without the need or presumed commitment of opening one.
(yeah my computer is potato so i'd rather test on CI thx)

> [!IMPORTANT]
> I do not know if this will work with the test path changed filter

Added 1 line of code for test workflow trigger

## Screenshots/Videos
N/A

## How to test the changes?
Run the workflow.
Unfortunately, github will [only let you run it if the dispatch workflow is on the default branch](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manually-run-a-workflow#configuring-a-workflow-to-run-manually), so I can't show pics of it here.

## Checklist


SirzBenjie:
Why tho?

Bertie690:
I said it in the desc - so i can run tests on CI just to "check if smthng works" without needing to make a PR and then close it 5 minutes later if it fails

DayKev (approved):

PR #6159, opened 2025-07-28
[Test] Ported over + augmented remaining test matchers from pkty

Bertie690:
Normal folks: none
Devs: Much much easier test checking

`pkty` has a lot of good test matchers, so I saw it fit to port em over.

I also added another matcher, `toHaveUsedPP`, which searches for a move in a Pokemon's moveset having consumed the given amount of PP.

> [!IMPORTANT]
> **This is not a pure port**. I took the liberty of adding tweaks and improvements to the matchers where needed (such as supporting checking individual properties of a `TurnMove` or `Status`), resulting in the matchers not being _exactly_ identical to their poketernity cousins.
> I also (if I do say so myself) vastly improved the test logging behavior, showing abbreviated inline diffs of non-matching objects in the failure message.
image
image

I removed the basic prototype name from inline diffs ("Object", "Array") and limited the display to a maximum of 30 characters to (hopefully) keep it within a 120 character terminal screen without overflow. 
I'm welcome to tweaking the stringification options if the need arises, though most matchers checking numerical properties can get by with a simple reverse mapping or `getEnumStr` instead. 

Added the matchers to the matchers directory.

Added utility functions to remove repetitive code for reverse mapping/case changing and prettify things a bit

> [!NOTE]
> `toHaveStat` was actively unused in their repo (given it is effectively _only_ used when testing Transform and niche effects that override `summonData.stats`), so I omitted it.
> If we need it later, i can copy over the file and work from there.

## Screenshots/Videos
N/A

## How to test the changes?
I picked `spite.test.ts` because it was a move we had no tests for and could show off the new matchers. :)

## Checklist


src/typings/i18next.d.ts
```diff
@@ -3,6 +3,7 @@ import type { TOptions } from "i18next";
 // Module declared to make referencing keys in the localization files type-safe.
 declare module "i18next" {
   interface TFunction {
+    // biome-ignore lint/style/useShorthandFunctionType: IDK why this breaks everything but it does
```
DayKev:
```suggestion
    // biome-ignore lint/style/useShorthandFunctionType: This needs to be an interface due to interface merging
```

Wlowscha (dismissed):

SirzBenjie (dismissed):

PR #6160, opened 2025-07-28
[Dev] Change `target` to `ES2023` in `tsconfig.json`

DayKev:
🤷 (Most likely no visible changes, maybe better performance)

Ability to use features from ES2021-ES2023.

All features of ES2021-ES2023 are now available for use (such as `Object.hasOwn()`, `Array#toSpliced()`, top level `await`, etc).

## Checklist

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -213,7 +213,6 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * TODO: Stop treating this like a unique ID and stop treating 0 as no pokemon
    */
   public id: number;
-  public name: string;
   public nickname: string;
   public species: PokemonSpecies;
   public formIndex: number;
@@ -5664,7 +5663,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
 }
 
 export class PlayerPokemon extends Pokemon {
-  protected battleInfo: PlayerBattleInfo;
+  protected declare battleInfo: PlayerBattleInfo;
   public compatibleTms: MoveId[];
 
   constructor(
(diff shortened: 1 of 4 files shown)


SirzBenjie (approved):

src/field/trainer.ts
```diff
@@ -30,7 +30,6 @@ export class Trainer extends Phaser.GameObjects.Container {
   public config: TrainerConfig;
   public variant: TrainerVariant;
   public partyTemplateIndex: number;
-  public name: string;
```
SirzBenjie, line 33: `name` is actually always present, though. See line 73. Maybe change to declare?

src/field/pokemon.ts
```diff
@@ -213,7 +213,6 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * TODO: Stop treating this like a unique ID and stop treating 0 as no pokemon
    */
   public id: number;
-  public name: string;
```
SirzBenjie, line 216: Change this to declare, it's always defined in the constructor.

src/field/trainer.ts
DayKev: It's inherited from `Phaser.GameObjects.Container` and is already typed as `string` so we don't need to re-declare it.

src/field/pokemon.ts
DayKev: Same as above.

src/field/trainer.ts
SirzBenjie: Ah of course!

SirzBenjie (approved):

Bertie690 (approved):

damocleas (approved):

PR #6164, opened 2025-07-28
[Bug] Fix behavior of ally-targeting moves during sleep

### LINKED ISSUE #6040: [Bug] The move “Dragon Cheer” can be used even in sleep, with the sleep timer not going on
### Describe the bug

bug screenshot
As what could be seen in the picture, the move “Dragon Cheer” can be used even in sleep, and furthermore:
1.It will be failed;
2.The sleep timer will not go on, just like throwing a pokeball.

### Reproduction

1. Enter the Daily Run (2025/6/27)
2. Archaludon uses Sleep
3. Archaludon uses Dragon Cheer

### Expected behavior

When use this move in sleep, just judge should the pokemon be awaken like any other move.

### Screenshots / Videos

_No response_

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

acelynnzhang:

When using an ally-targeting move with no ally while asleep, the failed move messaged will not appear, just the sleep message.

Fixes #6040 

In `move-phase.ts`, the function call to `canMove` is moved down so that status effects can be applied first and also tick down. Previously `canMove` was returning early, in the beginning of the`start` function itself. I moved it into `resolveFinalPreMoveCancellationChecks` so that `handlePreMoveFailures` will be run for all fails/cancels. I added a test case for sleep turns ticking down correctly with Dragon Cheer.
## Screenshots/Videos

Before:

https://github.com/user-attachments/assets/d531de2c-b601-4c3e-9343-301ac189ae69

After:

https://github.com/user-attachments/assets/51c41c23-7f64-4230-bb4b-9856da785fb5

## How to test the changes?

```typescript
const overrides = {
  STATUS_OVERRIDE : StatusEffect.SLEEP,
  OPP_MOVESET_OVERRIDE: [MoveId.SPLASH],
  MOVESET_OVERRIDE: [MoveId.REST, MoveId.DRAGON_CHEER, MoveId.COACHING, MoveId.TACKLE]
```
Use Dragon Cheer.
## Checklist
  - [ ] ~~Have I made sure that any UI change works for both UI themes (default and legacy)~~?


src/phases/move-phase.ts
```diff
@@ -122,18 +122,6 @@ export class MovePhase extends BattlePhase {
...
-    if (!this.canMove(true)) {
-      if (this.pokemon.isActive(true)) {
-        this.fail();
-        this.showMoveText();
-        this.showFailedText();
-      }
-      this.end();
-      return;
-    }
-
```
Bertie690, line 142: We need to keep the "are they off field" part of the check here - otherwise things like Encore and Truant will trigger when a Pokemon is forced off field via Whirlwind, etc.
The rest _should_ be good to go down lower (tho idk)

src/phases/move-phase.ts
```diff
@@ -157,6 +145,7 @@ export class MovePhase extends BattlePhase {
 
     this.resolveCounterAttackTarget();
 
+    //Decides if move should be canceled due to status effect
```
Bertie690:
```suggestion
    // Check status cancellation from sleep, freeze, etc.
```

test/data/status-effect.test.ts
```diff
@@ -390,6 +390,35 @@ describe("Status Effects", () => {
       expect(player.status).toBeFalsy();
       expect(player.getLastXMoves(1)[0].result).toBe(MoveResult.SUCCESS);
     });
+
+    it("Sleep turns should tick down when failing to use ally-targeting moves", async () => {
+      await game.classicMode.startBattle([SpeciesId.FEEBAS]);
+
+      const player = game.scene.getPlayerPokemon()!;
```
Bertie690:
```suggestion
      const player = game.field.getPlayerPokemon();
```

src/phases/move-phase.ts
acelynnzhang: Okay, that makes sense, so maybe just return early if pokemon is not active? I also realized i`sActive` can be true when canMove is false oops lol

acelynnzhang:
Changes have been made based on suggestions, as well as removing text for when a Pokemon is already off field. For example a Pokemon's move should not show move failed after they have already fled because of Whirlwind.

Bertie690:
PR Review: <https://github.com/acelynnzhang/pokerogue/pull/2>

Mostly just removing unneded duplicated checks and whatnot, lgtm

Bertie690 (dismissed):

src/phases/move-phase.ts
```diff
@@ -183,18 +179,29 @@ export class MovePhase extends PokemonPhase {
...
+    // Check if move is unusable (e.g. running out of PP due to a mid-turn Spite
+    // or the user no longer being on field)
+
+    if (!this.canMove(true)) {
+      if (this.pokemon.isActive(true)) {
+        this.fail();
+        this.showMoveText();
+        this.showFailedText();
+      }
+      return;
```
Bertie690, line 195:
```suggestion
      this.fail();
      this.showMoveText();
      this.showFailedText();
      return;
```

PR #6036, opened 2025-06-26
[Misc/Feature] Add dynamic turn order

emdeann:
## What are the changes the user will see?
Anything controlled by phases (postsummon abilities, moves, switches, etc.) will happen dynamically in speed order instead of in a static order determined at the start of the turn. 

## Why am I making these changes?
Mainline behavior + improving the structure of the phase manager

## What are the changes from a developer perspective?
This is obviously (and necessarily) a titanic PR. There are two main changes to focus on:

1. The phase queue is no longer a jumble of prepend and conditional and command queues. It is instead its own object and is represented as a tree-like structure that Curbio lovingly referred to as the "Bonsai". It supports similar basic operations to the old phase queue, namely `push` and `unshift`. 

The main queue is a 2D array of "levels". Phases are popped from the "highest" level first, with each level running in FIFO order. "unshifted" phases are pushed to the level above the currently running phase. Pushed phases are instead pushed to the lowest level. See `phase-tree.ts`.

2. `DynamicQueueManager` has been added to manage the dynamic ordering of phases. 

From `dynamic-queue-manager.ts`:
> The dynamic queue manager holds priority queues for phases which are queued as dynamic.
>
> Dynamic phases are generally those which hold a pokemon and are unshifted, not pushed. 
> Queues work by sorting their entries in speed order (and possibly with more complex ordering) before each time a phase is 
> popped.
>
> As the holder, this structure is also used to access and modify queued phases. This is mostly used in redirection, cancellation,   
> etc. of MovePhases.

The manager is actually quite simple - all it does is map phases to priority queues and provide an API to access them. The actual work of ordering falls on the queues themselves, which have a defined ordering method which is called whenever an element is to be popped. There is a generic `PokemonPhasePriorityQueue` along with specialized ones when necessary. 

The way it currently works is that a phase is considered dynamic if it is a `PokemonPhase` (or otherwise has `getPokemon()` defined) and is not blacklisted by the DQM. It may be better to do it the other way around and explicitly list which phases should be dynamic.

A couple other notes:
- Quick Draw/Claw have been changed to be a battler tag, similar to https://github.com/Despair-Games/poketernity/pull/262
- Most phase manipulation from outside the phase manager has been eliminated, you will notice that moves that played with ordering in particular (encore, instruct, shell trap, etc) are generally drastically simplified)

Todos:
- [ ] Dynamic ordering tests!!
- [ ] Should DQM use a whitelist or blacklist

## Screenshots/Videos
[...]

PR #6169, opened 2025-07-29
[Balance] End of turn triggers won't occur when you end a biome

DayKev:
When you defeat the final Pokémon at the end of a biome (waves 10, 20, ...), the battle will end immediately, meaning end of turn triggers (weather damage, berries, etc) won't occur. This prevents berries from being wasted when your party is about to be automatically healed anyway.

Requested by balance team.

A new phase (`CheckBiomeEndPhase`) is queued along with the other "end of turn" phases from within `TurnStartPhase` which removes the end of turn effect phases and sets a new field `TurnEndPhase#endOfBiome` to `true` on the currently queued `TurnEndPhase`, which causes most of the triggers to be skipped in that phase.

## How to test the changes?
Defeat the last Pokémon of a biome (wave X0) with end of turn triggers that should occur (opponent's Future Sight, weather damage, Aqua Ring, etc) and observe that they do not occur.

You can also run the automated test with `pnpm test:silent check-biome-end-phase`.

## Checklist

### DIFF
--- a/src/phases/check-biome-end-phase.ts
+++ b/src/phases/check-biome-end-phase.ts
@@ -1,6 +1,5 @@
 import { globalScene } from "#app/global-scene";
 import { Phase } from "#app/phase";
-import type { TurnEndPhase } from "#phases/turn-end-phase";
 
 export class CheckBiomeEndPhase extends Phase {
   public override readonly phaseName = "CheckBiomeEndPhase";
@@ -11,15 +10,7 @@ export class CheckBiomeEndPhase extends Phase {
     const { waveIndex } = globalScene.currentBattle;
 
     if (waveIndex % 10 === 0 && globalScene.getEnemyParty().every(p => p.isFainted())) {
-      const phasesToRemove = ["WeatherEffectPhase", "BerryPhase", "CheckStatusEffectPhase"];
-      while (phaseManager.tryRemovePhase(p => phasesToRemove.includes(p.phaseName))) {
-        // do nothing, this loop just exists to remove all the relevant phases from the queue
-      }
-
-      const turnEndPhase = phaseManager.findPhase<TurnEndPhase>(p => p.phaseName === "TurnEndPhase");
-      if (turnEndPhase) {
-        turnEndPhase.endOfBiome = true;
-      }
+      phaseManager.onBiomeEnd();
     }
 
     this.end();
--- a/src/phases/check-biome-end-phase.ts
+++ b/src/phases/check-biome-end-phase.ts
@@ -1,0 +1,27 @@
+import { globalScene } from "#app/global-scene";
+import { Phase } from "#app/phase";
+import type { TurnEndPhase } from "#phases/turn-end-phase";
+
+export class CheckBiomeEndPhase extends Phase {
+  public override readonly phaseName = "CheckBiomeEndPhase";
+
+  public override start(): void {
+    super.start();
+    const { phaseManager } = globalScene;
+    const { waveIndex } = globalScene.currentBattle;
+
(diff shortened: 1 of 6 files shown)


src/phases/check-biome-end-phase.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  public override start(): void {
+    super.start();
+    const { phaseManager } = globalScene;
+    const { waveIndex } = globalScene.currentBattle;
+
+    if (waveIndex % 10 === 0 && globalScene.getEnemyParty().every(p => p.isFainted())) {
+      const phasesToRemove = ["WeatherEffectPhase", "BerryPhase", "CheckStatusEffectPhase"];
+      while (phaseManager.tryRemovePhase(p => phasesToRemove.includes(p.phaseName))) {
+        // do nothing, this loop just exists to remove all the relevant phases from the queue
```
DayKev: I could have done `phaseManager.phaseQueue.filter(...)` instead of this loop but imo `PhaseManager#phaseQueue` should not be `public` and should only be interacted with via helper functions (or at the very least should not be publicly writable, a public getter would be fine).

src/phases/check-biome-end-phase.ts
```diff
@@ -0,0 +1,27 @@
...
+      }
+
+      const turnEndPhase = phaseManager.findPhase<TurnEndPhase>(p => p.phaseName === "TurnEndPhase");
+      if (turnEndPhase) {
+        turnEndPhase.endOfBiome = true;
+      }
```
emdeann: My opinion (and indeed the direction that #6036 takes) is that we should be pulling and manipulating individual phases outside of the purview of the phase manager itself as little as possible. I would prefer this to be something like
```ts
if (waveIndex % 10 === 0 && globalScene.getEnemyParty().every(p => p.isFainted())) {
  phaseManager.onBiomeEnd()
}
```
and then put the phase logic in the `PhaseManager#onBiomeEnd`. Would also let you use `filter` properly

DayKev: It feels kinda weird to have the logic be moved away from where it's used but I guess so.

emdeann (dismissed):

src/phases/check-biome-end-phase.ts
emdeann: Totally makes sense, but at the same time it feels weird to be picking individual phases from the queue and manipulating them outside of the manager (at least to me). The ideal here is a tcm style thing where the phase manager can just not queue these phases at all if the biome ends - this is conducive to that because it can be implemented with changes only to the manager itself

DayKev: True, for now it's awkward but in the future once things are refactored it will be better.

damocleas (dismissed):

src/phases/turn-end-phase.ts
```diff
@@ -18,6 +18,8 @@ import i18next from "i18next";
 
 export class TurnEndPhase extends FieldPhase {
   public readonly phaseName = "TurnEndPhase";
+  public endOfBiome = false;
```
Wlowscha: I would also suggest to change this, especially if later on we need to make a distinction between end of biome (e.g. to avoid processing boosting berries, while processing healing) and interlude

DayKev: Whoops, missed that.

Wlowscha (approved):

damocleas (approved):

emdeann (approved):

emdeann:
Test file should be renamed at some point technically, but hardly matters

PR #6174, opened 2025-07-30
[Test] `game.move.use` overrides summon data moveset if present

Bertie690:
N/A
Better testing UX

If the user has a summon data moveset, it will be reset and overridden with the corresponding move ID when `MoveHelper.use` is called.
Previously, doing so while affected by a Transform-like effect would usually just fail the test due to it not being in the temp moveset.

Cleaned up tsdocs a bit

## Screenshots/Videos
N/A
## How to test the changes?
N/A - Given this is merely enabling functionality that previously would fail tests, I see no reason this should pose problems.

## Checklist


test/test-utils/helpers/move-helper.ts
```diff
@@ -176,8 +177,11 @@ export class MoveHelper extends GameManagerHelper {
       console.warn("Warning: `MoveHelper.use` overwriting player pokemon moveset and disabling moveset override!");
     }
 
+    // Clear out both the normal and temporary movesets before setting the move.
     const pokemon = this.game.scene.getPlayerField()[pkmIndex];
-    pokemon.moveset = [new PokemonMove(moveId)];
+    pokemon.moveset.splice(0);
+    pokemon.summonData.moveset?.splice(0);
+    pokemon.setMove(0, moveId);
```
DayKev, line 184: This reminds me, `Pokemon#moveset` should be made `protected`.

PR #6188, opened 2025-08-01
[Misc] Fix missing types in battler tags and remove DragonCheerTag

SirzBenjie:
Technically, saves which had an active dragon cheer that was at +2 will now be at +1. Though this is so niche it doesn't matter (and will be resolved once Zod schemas come in).

Apparently, I forgot to mark the `tagType` for some of the battler tags in #6129.
Also, I needed a way to get TypeScript to differentiate between battler tags that were using the generic battler tags for serialization. So, I created an interface for this.

- Adds additional type helpers to the battler tag type
- Adds the missing `tagType` field to all classes that didn't have one.
- Removed `DragonCheerTag`, instead reusing the `CritBoostTag`
   - Rather than adding a save migrator (which has been deemed unnecessary for fleeting tags like this), added a new serialized field to the crit tag that tracks the number of stages that the tag boosts. Defaults to 1 when deserializing.
   - The defaulting to 1 when deserializing can be dropped once Zod schemas are fully implemented.
- Marked the `TypeImmuneTag` as abstract
- Created a generic interface, `GenericSerializedBattlerTagType` used exclusively by the `BattlerTagTypeMap` to allow TypeScript to be able to differentiate between tags that want to just reuse the basic `TagType`.
  - This is only necessary for SerializedTags, so `BYPASS_SLEEP` was left alone.

## Screenshots/Videos
N/A

## How to test the changes?
N/A?
Or, make sure dragon cheer works. I did this manually.

## Checklist

### DIFF
--- a/src/data/battler-tags.ts
+++ b/src/data/battler-tags.ts
@@ -34,15 +34,19 @@ import type { StatStageChangeCallback } from "#phases/stat-stage-change-phase";
 import i18next from "#plugins/i18n";
 import type {
   AbilityBattlerTagType,
+  BattlerTagTypeData,
   ContactSetStatusProtectedTagType,
   ContactStatStageChangeProtectedTagType,
+  CritStageBoostTagType,
   DamageProtectedTagType,
   EndureTagType,
   HighestStatBoostTagType,
   MoveRestrictionBattlerTagType,
   ProtectionBattlerTagType,
+  RemovedTypeTagType,
   SemiInvulnerableTagType,
   TrappingBattlerTagType,
+  TypeBoostTagType,
 } from "#types/battler-tags";
 import type { Mutable } from "#types/type-helpers";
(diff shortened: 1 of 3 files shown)


src/data/battler-tags.ts
```diff
@@ -2325,24 +2355,6 @@ export class CritBoostTag extends SerializableBattlerTag {
   }
 }
 
-/**
- * Tag for the effects of Dragon Cheer, which boosts the critical hit ratio of the user's allies.
- */
-export class DragonCheerTag extends CritBoostTag {
-  /** The types of the user's ally when the tag is added */
-  public typesOnAdd: PokemonType[];
```
DayKev, line 2333: Was this not being used or something?

DayKev: Oh nevermind you forgot about it lol.

SirzBenjie: Yeah I did lol. I got it tho.

DayKev (approved):

PR #6194, opened 2025-08-01
[Misc] More battler tag fixups

SirzBenjie:
Quark Drive and protosynthesis will no longer consider abilities like pure power or an ally's flower gift when determining the stat to boost.
Session saves with the stockpiling battler tag will no longer have a serialized function in them.

Cleaning up battler tags.
Removing the serialized function from stockpiling tag is necessary for zod schemas.

- Added default values to fields in various battler tags
- Cleaned up the calculation of highest effective stat in the `HighestStatBoostTag`
- Made is to that `StockpilingTag`'s `onStatStagesChanged` is defined as a method instead of a function property so that it wouldn't be serialized
  - It appears to have used the function property so that there could be parity with the type of `StatStageChangeCallback`
  - I kept this type safety check with a dummy statement that gets removed during bundling.

## Screenshots/Videos
N/A

## How to test the changes?
Start game with a quark drive / protosynthsis mon and ensure

## Checklist

### DIFF
--- a/src/data/battler-tags.ts
+++ b/src/data/battler-tags.ts
@@ -2722,7 +2722,7 @@ export class StockpilingTag extends SerializableBattlerTag {
         true,
         false,
         true,
-        this.onStatStagesChanged,
+        this.onStatStagesChanged.bind(this),
       );
     }
   }


src/data/battler-tags.ts
```diff
@@ -2128,28 +2128,28 @@ export class HighestStatBoostTag extends AbilityBattlerTag {
...
-    ).reduce((highestValue: number, value: number, i: number) => {
-      if (value > highestValue) {
-        highestStat = EFFECTIVE_STATS[i];
-        return value;
-      }
-      return highestValue;
-    }, 0);
+    const highestStat = EFFECTIVE_STATS.reduce(
+      (curr: [EffectiveStat, number], stat: EffectiveStat) => {
+        const value = pokemon.getEffectiveStat(stat, undefined, undefined, true, true, true, false, true, true);
```
DayKev, line 2140: Ah this still has 500 optional params huh.

SirzBenjie: yup.....

DayKev (approved):

PR #6195, opened 2025-08-01
[Test] Removed deprecated functions from phase interceptor

Bertie690:
- **Removed deprecated functions from phase interceptor**
- **Added minor docs to the phase manager + renamed `shift` to `shiftPhase`**

N/A
Callers should not be able to access `run`, `whenAboutToRun` and similar - these are deprecated and/or internal functions.
Made `phaseInterceptor.run` private and cleaned up a bit

Removed:
- `whenAboutToRun` --> `to(XYZ, false)`
- `setOverlayMode` (literally useless and unused)
- `unlock` (prior uses were _literally_ only in phases where it was being called already by lieu of `endBySetMode`)
- `runFrom`

Renamed `shift` to `shiftPhase`

Added a bit of docs/denesting and type safety (will likely merge conflict)

## Screenshots/Videos
N/A

## How to test the changes?
Check that existing tests aren't breaking horribly???

## Checklist


DayKev:
The tests are timing out.

test/test-utils/game-manager.ts
```diff
@@ -178,9 +176,10 @@ export class GameManager {
    * @returns A promise that resolves when the title phase is reached.
    */
   async runToTitle(): Promise<void> {
-    await this.phaseInterceptor.whenAboutToRun(LoginPhase);
-    this.phaseInterceptor.pop();
-    await this.phaseInterceptor.run(TitlePhase);
+    // Go to login phase and skip past it
+    await this.phaseInterceptor.to("LoginPhase", false);
+    this.phaseInterceptor.shiftPhase();
```
DayKev: This is not the same behavior, previously this did `phaseInterceptor.onHold.pop(); phaseInterceptor.shiftPhase();` but you removed the `onHold.pop()` part.

test/phases/mystery-encounter-phase.test.ts
```diff
@@ -49,9 +49,9 @@ describe("Mystery Encounter Phases", () => {
 
       game.onNextPrompt("MysteryEncounterPhase", UiMode.MYSTERY_ENCOUNTER, () => {
         // End phase early for test
-        game.phaseInterceptor.superEndPhase();
+        game.endPhase();
```
DayKev, line 52: That doesn't seem like the same behavior.

Bertie690: It should get overridden by lieu of the phase interceptor stubbing out `Phase.prototype.end`.

test/test-utils/game-manager.ts
Bertie690: Actually, I just forgot a boolean.
Our phase system is 100% linear, so `onHold` is always between lengths 1 and 0.
As a result, whether we remove it from the start or the end does not matter.

Bertie690:
Fixed it

test/moves/parting-shot.test.ts
```diff
@@ -153,43 +149,24 @@ describe("Moves - Parting Shot", () => {
 
   it.todo(
```
DayKev, line 150: This test is disabled so I guess it doesn't matter atm... but the test doesn't make sense. What's the Splash usage for on turn 1? Why is there `doSelectPartyPokemon(1)` if there's no non-fainted Pokémon for the player to switch to (and it's explicitly not supposed to be switching anyway)?

test/test-utils/phase-interceptor.ts
```diff
@@ -379,7 +356,7 @@ export class PhaseInterceptor {
    * @param args - Additional arguments to pass to the original method.
    */
   setMode(mode: UiMode, ...args: unknown[]): Promise<void> {
-    const currentPhase = this.scene.phaseManager.getCurrentPhase();
+    const currentPhase = this.scene.phaseManager.getCurrentPhase()!;
```
DayKev, line 360:
```suggestion
    // @ts-expect-error: `getCurrentPhase()` won't return `null` here
    const currentPhase = this.scene.phaseManager.getCurrentPhase();
```
This way it'll be flagged for removal when `getCurrentPhase()` is no longer able to return `null`.

test/moves/parting-shot.test.ts
Bertie690: I... don't know
My guess is it makes it fail instead of timing out in the event it does switch out??????????
I'm moving that entire test file anyways in my wimp out pr anyhow IIRC

test/test-utils/phase-interceptor.ts
```diff
@@ -425,7 +390,7 @@ export class PhaseInterceptor {
           currentPhase === actionForNextPrompt.phaseTarget &&
           currentHandler.active &&
           (!actionForNextPrompt.awaitingActionInput ||
-            (actionForNextPrompt.awaitingActionInput && currentHandler.awaitingActionInput))
+            (actionForNextPrompt.awaitingActionInput && (currentHandler as AwaitableUiHandler)["awaitingActionInput"]))
```
DayKev, line 394: Is the `as AwaitableUiHandler` needed when you've changed it to `["awaitingActionInput"]` anyway?

test/moves/parting-shot.test.ts
DayKev: Well you changed it to be that way lol.

test/test-utils/phase-interceptor.ts
Bertie690: No but it is nicer

test/moves/parting-shot.test.ts
Bertie690: I deleted a todo test rather

Bertie690: That was bordering on duplicated

DayKev: Oh so the diff isn't fully accurate then.

test/test-utils/phase-interceptor.ts
DayKev: Oh right, that's not correct lol. One sec.

Bertie690: GRRRRRRRR

test/test-utils/phase-interceptor.ts
```diff
@@ -379,6 +356,7 @@ export class PhaseInterceptor {
    * @param args - Additional arguments to pass to the original method.
    */
   setMode(mode: UiMode, ...args: unknown[]): Promise<void> {
+    // @ts-expect-error: `getCurrentPhase()` won't return `null` here
     const currentPhase = this.scene.phaseManager.getCurrentPhase();
```
DayKev:
```suggestion
    // TODO: remove the `!` in PR 6243 / after PR 6243 is merged
    const currentPhase = this.scene.phaseManager.getCurrentPhase()!;
```
Whoops lol.

DayKev (dismissed):

Wlowscha (approved):

SirzBenjie (approved):

PR #6197, opened 2025-08-02
Show battle end phase in wild battle if the last pokemon fled.

DableUTeeF:
I don't know about you guys but if my enemy fled the field I'd count that as a win.

Simply skipping to next battle is really infuriating when the one pokemon with this ability has a really strong priority move. Especially when it shows up right before an important fight.


src/data/abilities/ability.ts
```diff
@@ -6309,6 +6309,9 @@ class ForceSwitchOutHelper {
         if (globalScene.currentBattle.double && !isNullOrUndefined(allyPokemon)) {
           globalScene.redirectPokemonMoves(switchOutTarget, allyPokemon);
         }
+        else{
```
Bertie690, line 6312: This needs to also be added in `ForceSwitchOutAttr` if it isn't already.
(Yeah, we have 2 300+ LOC duplicated sections for switch out logic. It sucks.)

SirzBenjie (changes requested):
~~PR as is will impact things like roar, whirlwind, etc.~~
We are discussing this with balance team. I'm blocking until we have an answer as to what balance team desires.

SirzBenjie:
Balance team's consesus on this is that if a wild pokemon flees or is forced out, then the game should take the player to the battle end sequence but _not_ permit rewards.
This would allow the player to purchase items in the shop and manage their party, but not select a reward as the battle should not be considered "won".

SirzBenjie:
Based on the above consensus, I'm going to close this PR.
I will open an issue describing the desired implementation in a moment. Feel free to work on it if it is desired.

DableUTeeF:
I'm not sure if you tried the actual PR because the one file I changed in the PR is specifically for Wimp Out and Emergency Exit. Roar/Whirlwind need a lot of duplicated code to be changed as per @Bertie690 mentioned above.

Although currently `BattleEndPhase` only accept a boolean called `isVictory`. And will end the run if it's false. Changing that to accept 3 conditions without breaking tons of things probably need a lot of prayer and some witchcraft.

SirzBenjie:
> I'm not sure if you tried the actual PR because the one file I changed in the PR is specifically for Wimp Out and Emergency Exit. Roar/Whirlwind need a lot of duplicated code to be changed as per @Bertie690 mentioned above.

Ah, you're right. There was a pull request, I think by @Bertie690, where the logic for abilities that force exit and moves that force exit was condensed into one place -- I was mistaken. (I did not try the actual PR here). Apologies.

> Although currently BattleEndPhase only accept a boolean called isVictory. And will end the run if it's false. Changing that to accept 3 conditions without breaking tons of things probably need a lot of prayer and some witchcraft.

I know there's a way to get to the end screen without the player having items to choose from, as this is done in a few Mystery Encounters.
I'll investigate how that is accomplished there -- we'd just need to reuse that logic.

DableUTeeF:
Probably this https://github.com/pagefaultgames/pokerogue/blob/beta/src/phases/mystery-encounter-phases.ts#L606

SirzBenjie:
> Probably this https://github.com/pagefaultgames/pokerogue/blob/beta/src/phases/mystery-encounter-phases.ts#L606

If I'm reading that right, wouldn't that just take the player to the next wave without the shop?

SirzBenjie:
Found it.

https://github.com/pagefaultgames/pokerogue/blob/beta/src/phases/mystery-encounter-phases.ts#L554

PR #6201, opened 2025-08-02
[Refactor] Clean up various battler/arena tag move attributes

Bertie690:
N/A

Mortal spin used a different, near-identical version of another attribute to remove battler tags, so I wanted to fix it.
Along the way, I saw about 5 other instances of the exact same thing occurring...

Removed mortal-spin exclusive `lapseTagAttr` and replaced it with `RemoveArenaTagAttr`

Added a helper function to remove all arena tags matching a given side/type (given spikes/etc might exist on both sides of the battlefield)

Refactored `RemoveScreensAttr` and `RemoveTrapsAttr` to extend off of `RemoveArenaTagsAttr` instead of doing their own thing

Removed `null` from the param list of `AddArenaTagAttr` and replaced it with 0 because jank

## Screenshots/Videos
N/A
## How to test the changes?
Rely on existing tests; will likely clean up arena tags later

## Checklist


src/field/arena.ts
```diff
@@ -834,6 +834,36 @@ export class Arena {
...
 
+  /**
+   * Find and remove all {@linkcode ArenaTag}s with the given tag types on the given side of the field.
+   * @param tagTypes - The {@linkcode ArenaTagType}s to remove
+   * @param side - The {@linkcode ArenaTagSide} to remove the tags from (for side-based tags), or {@linkcode ArenaTagSide.BOTH}
+   * to clear all tags on either side of the field
+   * @param quiet - Whether to suppress removal messages from currently-present tags; default `false`
+   * @todo Review the other tag manipulation functions to see if they can be migrated towards using this (more efficient)
+   */
+  public removeTagsOnSide(
```
DayKev: Why not do this?
```ts
public removeTagsOnSide(...) {
  for (const tagType of tagTypes) {
    this.removeTagOnSide(tagType, side, quiet);
  }
}
```
Or change `removeTagOnSide` to `removeTagsOnSide` and allow it to be passed an array of tag types?

src/data/moves/move.ts
```diff
@@ -1306,15 +1307,13 @@ export class MoveEffectAttr extends MoveAttr {
    * @param args Set of unique arguments needed by this attribute
    * @returns true if basic application of the ability attribute should be possible
    */
-  canApply(user: Pokemon, target: Pokemon, move: Move, args?: any[]) {
-    return !! (this.selfTarget ? user.hp && !user.getTag(BattlerTagType.FRENZY) : target.hp)
-           && (this.selfTarget || !target.getTag(BattlerTagType.PROTECTED) ||
-                move.doesFlagEffectApply({ flag: MoveFlags.IGNORE_PROTECT, user, target }));
+  canApply(user: Pokemon, target: Pokemon, move: Move, _args?: any[]) {
+    return (this.selfTarget ? user : target).hp > 0
```
```suggestion
    return !(this.selfTarget ? user : target).isFainted();
```

src/data/moves/move.ts
```diff
@@ -1306,15 +1307,13 @@ export class MoveEffectAttr extends MoveAttr {
...
   }
 
   /** Applies move effects so long as they are able based on {@linkcode canApply} */
-  apply(user: Pokemon, target: Pokemon, move: Move, args?: any[]): boolean {
-    return this.canApply(user, target, move, args);
+  apply(user: Pokemon, target: Pokemon, move: Move, _args?: any[]): boolean {
+    return this.canApply(user, target, move, _args);
```
```suggestion
  apply(user: Pokemon, target: Pokemon, move: Move, args?: any[]): boolean {
    return this.canApply(user, target, move, args);
```
There's no need to change this.

src/field/arena.ts
Bertie690: I thought so too, but it'll only remove the first tag of each kind if passed `ArenaTagSide.BOTH`, meaning tidy up/etc won't remove tags from both sides of the field if the same kind is present on both

Bertie690: I might do it in the future, (we really need to clean these methods up), but I digress

DayKev:
Looks like there's no test coverage for `RemoveScreensAttr`, should probably put some in the Defog test file.

SirzBenjie (approved):

src/data/moves/move.ts
```diff
@@ -6043,37 +6023,38 @@ export class AddArenaTagAttr extends MoveEffectAttr {
...
+ * Attribute to remove one or more arena tags from the field.
  */
 export class RemoveArenaTagsAttr extends MoveEffectAttr {
-  public tagTypes: ArenaTagType[];
-  public selfSideTarget: boolean;
+  /** An array containing the tags to be removed. */
+  private readonly tagTypes: readonly [ArenaTagType, ...ArenaTagType[]];
+  /**
+   * Whether to remove tags from both sides of the field (`true`) or
+   * the target's side of the field (`false`); default `false`
```
```suggestion
   * the target's side of the field (`false`)
   * @defaultValue `false`
```

src/data/moves/move.ts
```diff
@@ -5986,14 +5965,13 @@ export class RemoveAllSubstitutesAttr extends MoveEffectAttr {
 export class HitsTagAttr extends MoveAttr {
   /** The {@linkcode BattlerTagType} this move hits */
   public tagType: BattlerTagType;
-  /** Should this move deal double damage against {@linkcode HitsTagAttr.tagType}? */
+  /** Should this move deal double damage against {@linkcode this.tagType}? */
```
```suggestion
  /** Should this move deal double damage against {@linkcode tagType}? */
```

src/data/moves/move.ts
```diff
@@ -5986,14 +5965,13 @@ export class RemoveAllSubstitutesAttr extends MoveEffectAttr {
...
   public doubleDamage: boolean;
```
```suggestion
  public doubleDamage = false;
```

src/data/moves/move.ts
```diff
@@ -6119,80 +6101,37 @@ export class AddArenaTrapTagHitAttr extends AddArenaTagAttr {
...
-      globalScene.arena.removeTagOnSide(ArenaTagType.SPIKES, target.isPlayer() ? ArenaTagSide.ENEMY : ArenaTagSide.PLAYER);
-      globalScene.arena.removeTagOnSide(ArenaTagType.TOXIC_SPIKES, target.isPlayer() ? ArenaTagSide.ENEMY : ArenaTagSide.PLAYER);
-      globalScene.arena.removeTagOnSide(ArenaTagType.STEALTH_ROCK, target.isPlayer() ? ArenaTagSide.ENEMY : ArenaTagSide.PLAYER);
-      globalScene.arena.removeTagOnSide(ArenaTagType.STICKY_WEB, target.isPlayer() ? ArenaTagSide.ENEMY : ArenaTagSide.PLAYER);
-    }
-
-    return true;
+export class RemoveArenaTrapAttr extends RemoveArenaTagsAttr {
+  constructor(targetBothSides = false) {
+    // TODO: This triggers at a different time than {@linkcode RemoveArenaTagsAbAttr}...
```
```suggestion
    // TODO: This triggers at a different time than `RemoveArenaTagsAbAttr`...
```

Wlowscha (approved):

DayKev (dismissed):

Wlowscha (approved):

src/field/arena.ts
```diff
@@ -827,6 +827,36 @@ export class Arena {
...
+    tagTypes: ArenaTagType[] | ReadonlyArray<ArenaTagType>,
```
DayKev:
```suggestion
    tagTypes: ArenaTagType[] | readonly ArenaTagType[],
```
I guess we don't have the lint enabled that enforces consistency for this.

DayKev (approved):

DayKev (dismissed):

Wlowscha (dismissed):

PR #6202, opened 2025-08-02
[Move] Added laser focus locales

Bertie690:
Laser focus has text
it has text in mainline

Added a new attribute, `MessageAttr` that literally jsut shows text.
Replaced the prior "do X and show message" moves to use it

RIP `splashAttr`, you will not be missed lmao
## Screenshots/Videos

https://github.com/user-attachments/assets/2dfda49b-efc9-4820-91ec-45141ab5e931

## How to test the changes?
READ with your EYES

## Checklist

  - [x] If so, please leave a link to it here: <https://github.com/pagefaultgames/pokerogue-locales/pull/204>


DayKev (approved):

Madmadness65 (approved):

emdeann:
Can this go in the move attribute instead? My concern is that `BattlerTagType.ALWAYS_CRIT` appears move-agnostic (as it should), so a user wouldn't expect it to queue a laser focus-specific message

Bertie690:
> ? My concern is that `BattlerTagType.ALWAYS_CRIT` appears move-agnostic (as it should)

As it stands, this would require a new move attribute for a tag as we ironically lack a generic "show message during move use" indicator.
(We have like 3 different flavours of "attribute that adds tag and plays message right after", but I digress)

Bertie690:
I'd rather not clutter up `move.ts` with effectively single-use move attributes that literally just show text

emdeann:
> As it stands, this would require a new move attribute for a tag as we ironically lack a generic "show message during move use" indicator.

I saw this too, bit shocking. My point is basically that the type on the left here should never be more generic than the tag it points to on the right. Otherwise it will be used in situations where it shouldn't

image

If you want to avoid messing up `move.ts` you can make a `BattlerTagType.LASER_FOCUS` and point it to `LaserFocusTag` and leave the `BattlerTagType.ALWAYS_CRIT` as-is. Wouldn't be my preferred approach but it would work

DayKev:
I think there's a "messageattr" class, isn't there? Or is that just an "abattr" class?

emdeann:
> I think there's a "messageattr" class, isn't there? Or is that just an "abattr" class?

There appears to be only a `PreMoveMessageAttr` (which obviously does not apply here). Something standardized like `getTriggerMessage` on abilities is probably a reasonable addition at some point

DayKev (approved):

src/data/moves/move.ts
```diff
@@ -1694,20 +1691,34 @@ export class SurviveDamageAttr extends ModifiedDamageAttr {
...
+    // TODO: Consider changing if/when MoveAttr `apply` return values become significant
+    if (message) {
+      globalScene.phaseManager.queueMessage(message, 500);
+      return true;
+    }
+    return false;
   }
 }
 
+
```
xsn34kzx:
```suggestion
```

xsn34kzx (approved):
The smallest of nits, but LGTM.

Bertie690:
@DayKev Requesting bypass merge for this because it already got 3 reviews

PR #6208, opened 2025-08-03
[Modifier Refactor] Modernize reward pool

Wlowscha:
The goal of this PR is to complete the necessary steps for the modifier refactor. In particular, it deals with the following:

* `rewardInitObj` is removed entirely, and replaced by an `allRewards` object which associates `RewardId`s to functions which generate rewards.
* The `WeightedModifier` class has been removed. Instead, `RewardPool` contains entries which look like `{id, weight, maxWeight}`. Importantly, the ids can be HeldItemIds or TrainerItemIds---an appropriate helper function will take care of converting these to the appropriate Reward. This removes the need for duplicate ids for items in RewardIds.
* On a similar vein, the generation of fixed rewards has been overhauled. It is now possible to pass a list of ids, or a { id, args } object for generators if specific parameters are needed.
* The reward override has been updated to follow the new scheme.

This PR most likely makes obsolete passing ids to the Reward instances themselves.


Bertie690:
https://github.com/Wlowscha/pokerogue/pull/2

This can be for now or later

PR #6205, opened 2025-08-03
[Test] Added `toHaveArenaTagMatcher` + fixed prior matchers

Bertie690:
## What are the changes the user will see?
N/A
## Why am I making these changes?
i made an oopsie on my matcher
## What are the changes from a developer perspective?
fixed matcher

added new ones (dw, it works, i used it for other branch)

## Screenshots/Videos
TODO

## How to test the changes?
Check the `mortal-spin` branch where i stole these from

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`pnpm test:silent`)
  - [x] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?

PR #6210, opened 2025-08-04
[Refactor] Cleaned up arena trap moves; prevented serialization of max layers

Bertie690:
Save sizes will be marginally smaller due to lack of unneeded serialized fields

Benjie made a point of not serializing unneeded fields

the oversight in arena traps was what brought me there; the massive code duplication was what kept me there

Revamped entry hazards to use inheritance.
The prior `ArenaTrapTag` has been rendered abstract and had code refactored (read: drag-and-dropped) to reduce duplication and allow for better readability.
All `onAdd` functions (having been various flavours of "show a message key") are hanled by the base class and use a function `getOnAddMessage` to return the message to display. (The one exception is Imprison, which overrides the aforementioned method and _probably_ shouldn't be an arena tag anyways but I digress)

Rather than handing all the logic for condition checks and whatnot in `activateTrap`, all that logic is handled by the superclass's `apply` method. `activateTrap` is now _solely_ used for applying the tag's on-entry effects (like applying poison for TSpikes or reducing speed for webs).

Another abstract subclass, `DamagingTrapTag`, is inherited by Spikes/Stealth Rock for more code condensing. It condenses all the functionality for dealing trap-based damage, using an abstract method `getHpRatio` to allow subclasses to dynamically calcualte the HP% to deal. 

Each tag now overrides the getters `maxLayers` and `groundedOnly` to consolidate logic without requiring serialization.

TESTSZZZ
## Screenshots/Videos
N/A

## How to test the changes?
`pnpm test test/entry-hazards.test.ts`

## Checklist
   - [[Test] Added toHaveArenaTagMatcher + fixed prior matchers #6205](<https://github.com/pagefaultgames/pokerogue/pull/6205>); copied over required matcher to make tests work


SirzBenjie (approved):

test/moves/entry-hazards.test.ts
```diff
@@ -0,0 +1,235 @@
...
+  });
+
+  describe("Stealth Rock", () => {
+    it.each<{ multi: TypeDamageMultiplier; species: SpeciesId }>([
+      { multi: 0.25, species: SpeciesId.LUCARIO },
+      { multi: 0.5, species: SpeciesId.DURALUDON },
+      { multi: 1, species: SpeciesId.LICKILICKY },
+      { multi: 2, species: SpeciesId.DARMANITAN },
+      // TODO: Figure out why quad weak pokemon are taking 33% damage from rocks instead of 50%
+      // { multi: 4, species: SpeciesId.ARTICUNO },
```
DayKev: Is this PR the cause of the damage change? If so, this shouldn't just be left unfixed.

DayKev: Wait, you used Articuno which has a passive of Snow Warning, maybe that's the issue?

DayKev: When you use a Legendary pokemon as the enemy it automatically gets set as a boss with its passive active, you should not use legendaries as enemies unless you need to.

src/data/arena-tag.ts
```diff
@@ -725,197 +725,288 @@ export class IonDelugeTag extends ArenaTag {
 }
 
 /**
- * Abstract class to implement arena traps.
+ * Abstract class to implement [arena traps (AKA entry hazards)](https://bulbapedia.bulbagarden.net/wiki/List_of_moves_that_cause_entry_hazards).
+ * These persistent tags remain on-field across turns and apply effects to any {@linkcode Pokemon} switching in. \
+ * Uniquely, adding a tag multiple times will stack multiple "layers" of the effect, increasing its severity.
  */
 export abstract class ArenaTrapTag extends SerializableArenaTag {
```
DayKev: Tbh this should be renamed to something like `EntryHazardTag`.

Bertie690: They're synonyms tbf - one more prevalent in the JA community, one in English

DayKev: Except there's an ability that's literally called "Arena Trap" which isn't related to these. If that ability didn't exist then it wouldn't matter.

src/data/arena-tag.ts
```diff
@@ -725,197 +725,288 @@ export class IonDelugeTag extends ArenaTag {
...
+
+  // TODO: Add a `canAdd` field to arena tags to remove need for callers to check layer counts
+
+  /**
+   * Display text when this tag is added to the field.
+   * @param _arena - The {@linkcode Arena} at the time of adding this tag
+   * @param quiet - Whether to suppress messages during tag creation; default `false`
+   */
+  override onAdd(_arena: Arena, quiet = false): void {
+    // We assume `quiet=true` means "just add the bloody tag no questions asked"
```
Wlowscha:
```suggestion
    // Here, `quiet=true` means "just add the tag, no questions asked"
```

Wlowscha:
For consistency, `DamagingTrapTag` should be renamed to `DamagingEntryHazardTag`.

Also, how come Imprison is treated as a entry hazard?

Bertie690:
Oops i forgot to rename that one loooool

Bertie690:
Also Imprison is jank lol that's why

DayKev (approved):

DayKev:
> Also, how come Imprison is treated as a entry hazard?

It needs to be changed lol. But that's for a future PR.

PR #6212, opened 2025-08-04
[Bug/Ability] Fixed bugs with Intimidate triggers after reload/initial switch 

### LINKED ISSUE #6219: [Bug] On-summon abilities (Intimidate) trigger twice when sent out after initial switch
### Describe the bug

If a Pokemon with an ability that activates when sent out is sent out after an initial switch prompt is accepted, its ability will activate twice instead of once.

In essence, what happens is the `SummonPhase` that is started to play the ball send out animation also triggers a `PostSummonPhase` to activate the first pokemon's ability, which is not removed when the `CheckSwitchPhase` starts a new `SwitchSummonPhase` to send in the replacement.

This only occurs when the first Pokemon is physically sent out (such as during a singles --> doubles transition or a reload)

### Reproduction

1. Bring any Pokemon and an intimidate user
2. set battle mode to switch
3. switch to intimidate mon
4. drops attack twice 

### Expected behavior

Ideally, the first Pokemon should show the animation _not be considered as entering the field_ until the switch prompt is confirmed.

Only once "Confirm" is hit should the actual gameplay effects (ability trigger, hazard damage, etc etc) occur.

### Screenshots / Videos

_No response_

### Session export file

_No response_

### User data export file

_No response_

### Additional context

Discord Bug Report:
https://discord.com/channels/1125469663833370665/1401769301270335608

I suspect this has lasted for _at least_ since dynamic turn order on post-summon abilities was added, if not even earlier.

Fixing this would likely require a complete decoupling of animations from game logic similar to what [Poketernity](<https://github.com/Despair-Games/poketernity/pull/1236>) is doing with theirs.

### LINKED ISSUE #6249: [Bug] PostSummonAbilities don't activate on reload if you only have 1 party member
### Describe the bug

After the player pokemon is summoned, the phaseQueue is empty ONLY if it was a reload (no enemy summon steps happened), if the player has only 1 party member (no check switch phase) and it can't be on the first wave of a biome.

That scenario leads to the `phase-manager.ts` `shiftPhase()` seeing an empty phaseQueue after the player's pokemon is summoned in SummonPhase. Therefore the conditional PostSummonPhase never happens and no post summon abilities are triggered.

The code thats clearing the conditionalQueue too early:
```ts
if (!this.phaseQueue.length) {
  this.populatePhaseQueue();
  // Clear the conditionalQueue if there are no phases left in the phaseQueue
  this.conditionalQueue = [];
}
```

### Reproduction

```ts
const overrides = {
  STARTING_WAVE_OVERRIDE: 2,
  ABILITY_OVERRIDE: AbilityId.GRASSY_SURGE,
  OPP_ABILITY_OVERRIDE: AbilityId.INTIMIDATE,
```
Start a new classic run, notice how both Grassy Surge and Intimidate activate like normal.
Now press F5 to reload the page.
Press continue.
Now notice how nothing activated.

### Expected behavior

It should activate PostSummonPhase after SummonPhase, even if theres no more phases left in the queue (or there needs to be an extra dummy phase for this scenario?)

### Screenshots / Videos

https://github.com/user-attachments/assets/a4245c94-9969-47a8-83ad-28fc0c9ebf38

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

Bertie690:

Intimidate and simlar abilities now trigger exactly once during an initial switch or page refresh with 1 party member. 

Not twice, not 0 times, ONCE.

Helping fix bugs

Fixes #6219 
Fixes #6249

The implementation is not clean at all, but our switch system really needs to be reworked eventually.

Added helper to start battle with initial switch

Added some emergency code to remove extra `PostSummonPhase`s and `ActivatePriorityQueuePhase`s upon confirming a force switch. \
Is it jank? Maybe. \
Does it work? Yeah

Added intimidate test that mirrors the found information

## Screenshots/Videos
N/A
## How to test the changes?
Enable the intimidate test and see that it does indeed fail instead of timing out

## Checklist

### DIFF
--- a/src/data/phase-priority-queue.ts
+++ b/src/data/phase-priority-queue.ts
@@ -44,6 +44,30 @@ export abstract class PhasePriorityQueue {
   public clear(): void {
     this.queue.splice(0, this.queue.length);
   }
+
+  /**
+   * Attempt to remove one or more Phases from the current queue.
+   * @param phaseFilter - The function to select phases for removal
+   * @param removeCount - The maximum number of phases to remove, or `all` to remove all matching phases;
+   * default `1`
+   * @returns The number of successfully removed phases
+   * @todo Remove this eventually once the patchwork bug this is used for is fixed
+   */
+  public tryRemovePhase(phaseFilter: (phase: Phase) => boolean, removeCount: number | "all" = 1): number {
+    if (typeof removeCount === "string") {
+      removeCount = Number.MAX_SAFE_INTEGER; // For the lulz
+    }
+    let numRemoved = 0;
+    let phaseIndex = this.queue.findIndex(phaseFilter);
+    if (phaseIndex === -1) {
+      return 0;
+    }
+    do {
+      this.queue.splice(phaseIndex, 1);
+      numRemoved++;
+    } while (numRemoved < removeCount || (phaseIndex = this.queue.findIndex(phaseFilter)) !== -1);
+    return removeCount;
+  }
 }
 
 /**
@@ -79,6 +103,7 @@ export class PostSummonPhasePriorityQueue extends PhasePriorityQueue {
   private queueAbilityPhase(phase: PostSummonPhase): void {
     const phasePokemon = phase.getPokemon();
 
+    console.log(phasePokemon.getNameToRender());
     phasePokemon.getAbilityPriorities().forEach((priority, idx) => {
       this.queue.push(new PostSummonActivateAbilityPhase(phasePokemon.getBattlerIndex(), priority, !!idx));
       globalScene.phaseManager.appendToPhase(
--- a/src/data/phase-priority-queue.ts
+++ b/src/data/phase-priority-queue.ts
@@ -55,20 +55,21 @@ export abstract class PhasePriorityQueue {
    */
   public tryRemovePhase(phaseFilter: (phase: Phase) => boolean, removeCount: number | "all" = 1): number {
     if (removeCount === "all") {
-      removeCount = Number.MAX_SAFE_INTEGER;
+      removeCount = this.queue.length;
     } else if (removeCount < 1) {
       return 0;
     }
     let numRemoved = 0;
-    let phaseIndex = this.queue.findIndex(phaseFilter);
-    if (phaseIndex === -1) {
-      return 0;
-    }
-    while (numRemoved < removeCount && phaseIndex !== -1) {
+
+    do {
+      const phaseIndex = this.queue.findIndex(phaseFilter);
+      if (phaseIndex === -1) {
+        break;
+      }
       this.queue.splice(phaseIndex, 1);
       numRemoved++;
-      phaseIndex = this.queue.findIndex(phaseFilter);
-    }
+    } while (numRemoved < removeCount && this.queue.length > 0);
+
     return numRemoved;
   }
 }
--- a/src/data/phase-priority-queue.ts
+++ b/src/data/phase-priority-queue.ts
@@ -54,8 +54,10 @@ export abstract class PhasePriorityQueue {
    * @todo Remove this eventually once the patchwork bug this is used for is fixed
    */
   public tryRemovePhase(phaseFilter: (phase: Phase) => boolean, removeCount: number | "all" = 1): number {
-    if (typeof removeCount === "string") {
-      removeCount = Number.MAX_SAFE_INTEGER; // For the lulz
+    if (removeCount === "all") {
+      removeCount = Number.MAX_SAFE_INTEGER;
+    } else if (removeCount < 1) {
+      return 0;
     }
     let numRemoved = 0;
     let phaseIndex = this.queue.findIndex(phaseFilter);
--- a/src/data/phase-priority-queue.ts
+++ b/src/data/phase-priority-queue.ts
@@ -64,11 +64,12 @@ export abstract class PhasePriorityQueue {
     if (phaseIndex === -1) {
       return 0;
     }
-    do {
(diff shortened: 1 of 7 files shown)


Madmadness65 (approved):

src/data/phase-priority-queue.ts
```diff
@@ -44,6 +44,30 @@ export abstract class PhasePriorityQueue {
...
+   * @param phaseFilter - The function to select phases for removal
+   * @param removeCount - The maximum number of phases to remove, or `all` to remove all matching phases;
+   * default `1`
+   * @returns The number of successfully removed phases
+   * @todo Remove this eventually once the patchwork bug this is used for is fixed
+   */
+  public tryRemovePhase(phaseFilter: (phase: Phase) => boolean, removeCount: number | "all" = 1): number {
+    if (typeof removeCount === "string") {
+      removeCount = Number.MAX_SAFE_INTEGER; // For the lulz
+    }
```
DayKev:
```suggestion
    if (removeCount === "all") {
      removeCount = Number.MAX_SAFE_INTEGER;
    } else if (removeCount < 1) {
      return 0;
    }
```

src/data/phase-priority-queue.ts
```diff
@@ -44,6 +44,30 @@ export abstract class PhasePriorityQueue {
...
+    let numRemoved = 0;
+    let phaseIndex = this.queue.findIndex(phaseFilter);
+    if (phaseIndex === -1) {
+      return 0;
+    }
+    do {
+      this.queue.splice(phaseIndex, 1);
+      numRemoved++;
+    } while (numRemoved < removeCount || (phaseIndex = this.queue.findIndex(phaseFilter)) !== -1);
+    return removeCount;
```
DayKev:
```suggestion
    while (numRemoved < removeCount && phaseIndex !== -1) {
      this.queue.splice(phaseIndex, 1);
      numRemoved++;
      phaseIndex = this.queue.findIndex(phaseFilter);
    }
    return numRemoved;
```
How did this work before with `||` instead of `&&`? You were always removing all of the phases.

src/data/phase-priority-queue.ts
```diff
@@ -79,6 +103,7 @@ export class PostSummonPhasePriorityQueue extends PhasePriorityQueue {
   private queueAbilityPhase(phase: PostSummonPhase): void {
     const phasePokemon = phase.getPokemon();
 
+    console.log(phasePokemon.getNameToRender());
```
DayKev: If you're going to put a log at least make it say what's going on. "Queueing ability phase for \<pokemon\>" or something.

emdeann:
Do you know what the logic error in the original PR was that caused this to start happening? This is a level of janky that would maybe be reasonable for a p1 hotfix but I would really prefer to not be in the codebase at all

src/data/phase-priority-queue.ts
Bertie690: oops

Bertie690:
Basically the `SwitchSummonPhase` from doing the initial send out animation still queues its post summon phase even when the switch dialogue is confirmed and another Pokemon is sent in.

Bertie690:
Meaning it triggers everything twice

emdeann:
My question was getting at that we need to figure out why this happens after the dynamic order PR but not before. We didn't need this janky removing phases stuff before it, so we shouldn't now either

SirzBenjie:
I talked with @emdeann about this PR. I made a few minor changes. But, @Bertie690's solution is literally better than one I came up with independently without looking at this PR. It accounts for things I hadn't considered originally.
This also needs to go into hotfix.

SirzBenjie (approved):

src/field/pokemon.ts
```diff
@@ -2234,8 +2234,16 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     return this.hasPassive() && (!canApply || this.canApplyAbility(true)) && this.getPassiveAbility().hasAttr(attrType);
   }
 
-  public getAbilityPriorities(): [number, number] {
-    return [this.getAbility().postSummonPriority, this.getPassiveAbility().postSummonPriority];
+  /**
+   * Return the ability priorities of the pokemon's ability and, if enabled, its passive ability
+   * @returns A tuple containing the ability priorities of the pokemon
+   */
+  public getAbilityPriorities(): [number] | [activePriority: number, passivePriority: number] {
```
Bertie690, line 2241:
```suggestion
  public getAbilityPriorities(): [activePriority: number, passivePriority?: number] {
```

SirzBenjie: No, that will make typescript think that it can return an array of size 2 whose second element is undefined.

src/phases/switch-phase.ts
```diff
@@ -75,8 +76,10 @@ export class SwitchPhase extends BattlePhase {
         if (slotIndex >= globalScene.currentBattle.getBattlerCount() && slotIndex < 6) {
           // Remove any pre-existing PostSummonPhase under the same field index.
           // Pre-existing PostSummonPhases may occur when this phase is invoked during a prompt to switch at the start of a wave.
-          globalScene.phaseManager.tryRemovePhase(
+          globalScene.phaseManager.tryRemoveDynamicPhase(
```
Bertie690, line 80:
```suggestion
          // TODO: Separate the animations from `SwitchSummonPhase` and co. into another phase and use that on initial switch - this is a band-aid fix
          globalScene.phaseManager.tryRemoveDynamicPhase(
```

src/field/pokemon.ts
Bertie690: seriously
Bruv

PR #6186, opened 2025-08-01
[Challenge] Add Nuzlocke-related Challenges

xsn34kzx:
## What are the changes the user will see?

The addition of several new Challenge Mode modifiers:
- Limited Catch
  - Prevents Pokémon from being caught on any wave than isn't the first wave in a biome (X1 waves).
- Hardcore
  - Prevents fainted Pokémon from being revived. Because of this, revive items are prevented from being rolled as rewards at the end of a wave. Fusion and, although it's not possible to roll them, revive items have no effect on fainted Pokémon as well.
- Limited Support
   - Disables the automatic party heal during biome transitions (after X0 waves) and/or the shop after every wave; that is, you can choose to disable either one or both at the same time.

The challenge modifier list is changed to the following order:
- Fresh Start
- Hardcore
- Limited Catch
- Limited Support
- Mono Gen
- Mono Type
- Inverse Battle
- Flip Stat

An additional achievement, "Nuzlocke", is also added, which can be obtained by finishing a run with Permanent Faint, Fresh Start, and Limited Catch toggled on.

## Why am I making these changes?

Nuzlocke mode has been requested heavily, and its specifically been picked up and dropped several times.

## What are the changes from a developer perspective?

Added 8 new `ChallengeType`s:
- `PARTY_HEAL`, for challenges that want to toggle the automatic party heal during biome transitions.
- `SHOP`, for challenges that want to toggle the shop options after a wave ends.
- `POKEMON_ADD_TO_PARTY`, for challenges that want to validate whether a Pokémon can get added to the player's party.
- `POKEMON_FUSION`, for challenges that want to validate whether a Pokémon can be selected for fusion.
- `SHOP_ITEM`, for challenges that want to validate whether particular items are or are not sold in the shop.
- `WAVE_REWARD`, for challenges that want to validate whether particular items can be given as a reward.
- `PREVENT_REVIVE`, for challenges that want to toggle the ability to revive fainted Pokémon.

Application functions for each `ChallengeTypes` where added (i.e., `PARTY_HEAL` -> `applyPartyHeal`), and `Challenge` classes were made for each new challenge modifier with the relevant apply function overrides. Likewise, overloads were written for `applyChallenges` corresponding to each new application function, and calls to `applyChallenges` with any necessary objects (`BooleanHolders` mostly) were placed around to achieve their effects.

Minor clean-up was done in the areas that the code touched, and a new `challenges` folder was added for unit tests that pertain to challenges.

## Screenshots/Videos

<details>
[...]

PR #6215, opened 2025-08-04
[Bug] Release Fainted Pokémon in Switch Menu

xsn34kzx:

Players should now be able to release any fainted Pokémon in the "POKÉMON" party menu during a battle.

While testing https://github.com/pagefaultgames/pokerogue/pull/6186, @Blitz425 found out that fainted Pokémon couldn't be released from the in-battle party menu, which is cumbersome in a mode like Permanent Faint centered around having fainted Pokémon. The text that displays ("X has no energy left to battle!") also implies that this isn't intended behavior and only meant for trying to switch fainted Pokémon in.

The `if` block handling the release behavior was moved above the filtering code, which was using the `FilterNonFainted` callback preventing fainted Pokémon from being released.

### Before (on Beta)

https://github.com/user-attachments/assets/be2cdf91-d20a-43d7-9531-eb9578c59b50

### After

https://github.com/user-attachments/assets/d9f3c329-02f7-4d1b-b046-33e16d6df429

## How to test the changes?

Start a Classic run with two Pokémon (or catch one), let one faint, and upon switching the next Pokémon in, immediately go to party and try releasing the fainted Pokémon.

## Checklist
  - ~[ ] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?~
  - ~[ ] Have I made sure that any UI change works for both UI themes (default and legacy)?~

- ~[ ] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?~
  - ~[ ] If so, please leave a link to it here:~ 
- ~[ ] Has the translation team been contacted for proofreading/translation?~

### DIFF
--- a/src/ui/party-ui-handler.ts
+++ b/src/ui/party-ui-handler.ts
@@ -827,6 +827,10 @@ export class PartyUiHandler extends MessageUiHandler {
       globalScene.triggerPokemonFormChange(pokemon, SpeciesFormChangeItemTrigger, false, true);
     }
 
+    if (option === PartyOption.RELEASE) {
+      return this.processReleaseOption(pokemon);
+    }
+
     // If the pokemon is filtered out for this option, we cannot continue
     const filterResult = this.getFilterResult(option, pokemon);
     if (filterResult) {
@@ -837,7 +841,6 @@ export class PartyUiHandler extends MessageUiHandler {
 
     // For what modes is a selectCallback needed?
     // PartyUiMode.SELECT (SELECT)
-    // PartyUiMode.RELEASE (RELEASE)
     // PartyUiMode.FAINT_SWITCH (SEND_OUT or PASS_BATON (?))
     // PartyUiMode.REVIVAL_BLESSING (REVIVE)
     // PartyUiMode.MODIFIER_TRANSFER (held items, and ALL)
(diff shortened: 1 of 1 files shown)


Bertie690 (dismissed):

damocleas (approved):

src/ui/party-ui-handler.ts
```diff
@@ -827,6 +827,10 @@ export class PartyUiHandler extends MessageUiHandler {
       globalScene.triggerPokemonFormChange(pokemon, SpeciesFormChangeItemTrigger, false, true);
     }
 
+    if (option === PartyOption.RELEASE) {
```
DayKev, line 831:
```suggestion
    // This needs to be processed before the filter is run because
    // otherwise you aren't able to release fainted pokemon
    if (option === PartyOption.RELEASE) {
```

src/ui/party-ui-handler.ts
```diff
@@ -837,7 +841,6 @@ export class PartyUiHandler extends MessageUiHandler {
 
     // For what modes is a selectCallback needed?
     // PartyUiMode.SELECT (SELECT)
-    // PartyUiMode.RELEASE (RELEASE)
```
DayKev: This is still true, you just had to move it up.

DayKev (dismissed):

src/ui/party-ui-handler.ts
xsn34kzx: Is it actually? While debugging, the `selectCallback` shows up `undefined` and isn't used in `processReleaseOption`, so I figured this was just a mistake.

DayKev: It was in `doRelease` or something, but if it's never defined in this codepath I guess that might be the case then.

xsn34kzx: Ohhh, I see it, I'll throw it back in then.

PR #6218, opened 2025-08-05
[Test] Revamped `MockConsoleLog` with color support

Bertie690:
End users: N/A

Devs:
`console.trace`, `console.timeLog` and many more functions work in tests now!

also our logs don't look shit now

PRETTY LOGS GO BRRRRRR

Also we were pretty badly mangling them prior

Reworked `MockConsoleLog` to be a proper wrapper around the global `console` object, extending off of the `Console.Console` class and wrapping select methods with colorization.
This both cleans up the implementation of calling original logging funcs, and ensures that any methods not stubbed will just fallback to the default implementation.

### Color Inference
In order to preserve original formatting and allow for proper styling of production code, any strings containing `%c` directives are ran through a new file (`mock-console/infer-colors.ts`) to locate and convert color styles into hex codes usable by Chalk[^1].

RGB, HSV and HSL values simply use standard conversion formulae, while CSS color names ("green", "Teal", etc.) are handled via a JSON file (`color-map.json`) matching all globally accepted colors to their hex codes. (W3Schools has all 140 of them; I just had to convert the arrays into JSON blobs.)

[^1]: Node.JS doesn't support the `%c` CSS directive (listing it as a no-op), hence why we explicitly need logic to handle it. 

> [!IMPORTANT]
> I selectively made a few differences over PKTY's log statements:
> - Regular newlines are replaced with spaces within locales as they tend to break up the "flow" of text.
> - Many colors are added/changed from pkty. (Suggestions on recolorations are welcome)

### Vitest stuff
Added a new custom reporter (uncreatively named `custom-default-reporter) that hides the test name annotation from the console logs produced by Vitest.

Additionally, 2 new functions in `test/test-utils/setup/test-end-log.ts` are added to produce nice colored logs at the start and end of each test. These are made as hooks (instead of methods attached to the reporter itself) to force Vitest to wait for the logging to finish before starting the next test (fixing an issue where the "test ended" log would appear partway inside the next test's logs).

Added named colors to a select few functions and logging statements as a demonstration.

Deduped a bit of init logic in `GameWrapper.ts` (it was the exact same shit twice, idk why)

## Screenshots/Videos
image

## How to test the changes?
Run some tests baybeee

## Checklist


test/test-utils/mocks/mock-console/infer-color.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Find the first string with a "color:" CSS directive in an argument list.
+ * @param args - The arguments containing the color directive
+ * @returns The found color, or `"green"` if none were found
+ */
+function findColorPrefix(args: string[]): string {
+  for (const arg of args) {
+    const match = /color:\s*(.+?)(?:;|$)/g.exec(arg);
+    if (match === null) {
+      // no match
```
DayKev:
```suggestion
```
I think we can figure out that `match === null` means "no match" lol.

test/test-utils/mocks/mock-console/mock-console.ts
```diff
@@ -0,0 +1,140 @@
...
+   */
+  public static printPostTestWarnings() {
+    for (const data of MockConsole.queuedWarnings) {
+      console.warn(...data);
+    }
+    MockConsole.queuedWarnings = [];
+  }
+
+  /**
+   * Check whether a given ste of data is in the blacklist to be barred from logging.
```
DayKev:
```suggestion
   * Check whether a given set of data is in the blacklist to be barred from logging.
```

test/vitest.setup.ts
```diff
@@ -30,8 +47,12 @@ vi.mock("i18next", async importOriginal => {
       const filename = req.params[0];
 
       try {
-        const json = await import(`../public/locales/en/${req.params[0]}`);
-        console.log("Loaded locale", filename);
+        const localeFiles = import.meta.glob("../public/locales/en/**/*.json", { eager: true });
+        const json = localeFiles[`../public/locales/en/${filename}`] || {};
+        //
```
DayKev:
```suggestion
```

src/phases/move-phase.ts
```diff
@@ -118,7 +118,10 @@ export class MovePhase extends BattlePhase {
   public start(): void {
     super.start();
 
-    console.log(MoveId[this.move.moveId], enumValueToKey(MoveUseMode, this.useMode));
+    console.log(
+      `%cMove: ${MoveId[this.move.moveId]}; Use Mode: ${enumValueToKey(MoveUseMode, this.useMode)}`,
+      "color:RebeccaPurple",
```
SirzBenjie: I don't like the idea of having hardcoded strings for this. Can we use a global constant and assign colors to certain names? It shouldn't be a free for all when choosing a color for a console log.

test/test-utils/mocks/mock-console/infer-color.ts
SirzBenjie, line 1: I would like some sort of explanation as to the purpose of this file, what it's testing, etc.
We shouldn't be testing that an external library works properly in our own tests.

src/phases/move-phase.ts
Bertie690: These are the standard recognized HTML color names.

I literally ripped them off of the internet - there's ~140 globally recognized ones in total.

test/test-utils/mocks/mock-console/infer-color.ts
Bertie690: That's the file that finds an HTML color string ("color:green;", etc) and converts it to thr corresponding hex color useable by Chalk.

src/phases/move-phase.ts
Bertie690: FWIW, there is the giant `color-map.json` if that's what you're getting at

test/test-utils/mocks/mock-console/color-map.json
SirzBenjie, line 1: Out of curiosity, where did this map come from?

Bertie690: W3schools

Bertie690: > W3Schools has all 140 of them; I just had to convert the arrays into JSON blobs.

📚

SirzBenjie: Can you add a comment linking to it?

SirzBenjie: Oh wait, no, you can't lmao, it's a json file...

Bertie690: https://www.w3schools.com/tags/ref_colornames.asp

src/init/init.ts
```diff
@@ -16,6 +17,9 @@ import { initStatsKeys } from "#ui/game-stats-ui-handler";
 
 /** Initialize the game. */
 export function initializeGame() {
+  if (allMoves.length > 0) {
+    return;
+  }
```
SirzBenjie: Why in this PR?

src/phases/move-phase.ts
SirzBenjie: I more mean, we should have a named constant that indicates _when_ to use _which_ color when outputting to a console.

Like, for info, debug, moves, etc....

Bertie690: ... sure?
I can chuck a bunch of colors in a constants file if you want... as template literal strings...
can i do that later

src/init/init.ts
Bertie690: oppa

Bertie690: I copied over the init code from PKTY, and they had a fallback check over here as well

src/phases/move-phase.ts
Bertie690: so like `MOVE_LOG_COLOR`, etc?
i can do that maybe ish
AAAAAAAAAAAAAAAAa

src/constants/colors.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * @module
+ * A big file storing colors used in logging.
+ * Is minified by Vite during production runs to avoid overhead.
```
SirzBenjie:
```suggestion
 * Should be minified by Terser during production builds, so has no overhead.
```

SirzBenjie (approved):

src/constants/colors.ts
Bertie690:
```suggestion
 * Minified by Terser during production builds, so has no overhead.
```
Saying "should" kinda implies we don't really know if it gets minified lol 
Tbf we can easily check by simply checking the vite outputs for this and the prior commit

src/init/init.ts
SirzBenjie: I saw this myself when first doing it. I don't want to check against allMoves being populated. Each thing should be defined in its own right. If `allMoves`' init method wants to skip it, then it should.

SirzBenjie (approved):

src/init/init.ts
Bertie690: sure ig
I highly doubt it matters since we hopefully shouldn't even _run_ this more than once per shard, but w/e

I do eventually want to overhaul how we mock and stub out things for testing in general.

Bertie690:
Marking as draft for a quick bit while I fix the vitest console log issues (us overwriting console log effectively breaks vitest's silent mode)

SirzBenjie (dismissed):
Thanks for adding in the license stuff!

Wlowscha (approved):

SirzBenjie (approved):

PR #6220, opened 2025-08-06
[Bug] Fixed niche interactions with Encore + Magic Bounce/Magic Coat; Yawn + Stomping Tantrum

Bertie690:
1. Encore will not break when using Magic Coat reflected moves.
2. Yawn will correctly count as failed for Stomping Tantrum & co. when blocked by most status-preventing effects, and _should_ will display the corresponding blockage text upon doing so. These include:
    - Abilities
    - Safeguard
    - Terrain
3. (bad) Yawn likely shows 2 messages for status blockage, instead of only 1 (needa confirm).

Bugfixes

Encore was not correctly cleaned up post #5810, and still used a stray `pokemon.getLastXMoves()` call that broke its interactions with reflection.

Also, I noticed yawn was not checking everything

Fixed the aforementioned bugs (moving the yawn condition into a new attribute for now because i didn't want a 5 mile long condition)

## Screenshots/Videos
TODO
## How to test the changes?
New tests in the magic bounce file

## Checklist


SirzBenjie:
Folks, the 1k lines changed is completely misleading.
The order of the methods inside of the move effect phase was just moved around so they occur roughly in the order they happen during the phase. The movement basically bloats the line count.

SirzBenjie (approved):
This is good

Wlowscha (approved):

src/data/battler-tags.ts
```diff
@@ -1279,35 +1293,55 @@ export class EncoreTag extends MoveRestrictionBattlerTag {
...
-            lastMove.targets ?? [],
-            movesetMove,
-            MoveUseMode.NORMAL,
-          ),
-        );
-      }
+    // If the target has not moved yet,
+    // replace their upcoming move with the encored move against randomized targets
+    const movePhase = globalScene.phaseManager.findPhase(
+      (m): m is MovePhase => m.is("MovePhase") && m.pokemon === pokemon,
```
SirzBenjie: Cheaper checks first; no method calls on LHS of a boolean if the RHS doesn't also have a method call.
```suggestion
      (m): m is MovePhase => m.pokemon === pokemon && m.is("MovePhase"),
```

src/data/moves/invalid-moves.ts
```diff
@@ -280,3 +280,68 @@ export const invalidEncoreMoves: ReadonlySet<MoveId> = new Set([
...
+  MoveId.SLEEP_TALK,
+  MoveId.SNATCH,
+  MoveId.INSTRUCT,
+  // Misc moves
+  MoveId.KINGS_SHIELD,
+  MoveId.SKETCH,
+  MoveId.TRANSFORM,
+  MoveId.MIMIC,
+  MoveId.STRUGGLE,
+  // TODO: Add Max/G-Max/Z-Move blockage if or when they are implemented
```
SirzBenjie: This isn't a true todo item, and I don't want it to appear in my TODO greps
```suggestion
  // NOTE: Add Max/G-Max/Z-Move blockage if or when they are implemented
```

src/data/moves/move.ts
```diff
@@ -5716,6 +5706,34 @@ export class AddBattlerTagAttr extends MoveEffectAttr {
...
+      if (target.status || target.isSafeguarded(user)) {
+        return false;
+      }
+
+      // TODO: This does not display the cause of the "but it failed" message,
+      // but fixing it would require a rework of the move failure system
+      return target.canSetStatus(StatusEffect.SLEEP, true, false, user)
+    }
+  }
+}
```
SirzBenjie: I think the Yawn stuff should be its own PR

src/data/battler-tags.ts
Bertie690: @SirzBenjie 
... m might not be a move phase for the LHS conditional (in which case Typescript complains)

SirzBenjie (dismissed):

Bertie690:
@SirzBenjie I still have to move encore move override proccing to the move phase

src/battle.ts
```diff
@@ -75,7 +75,10 @@ export class Battle {
   public battleSeed: string = randomString(16, true);
   private battleSeedState: string | null = null;
   public moneyScattered = 0;
-  /** Primarily for double battles, keeps track of last enemy and player pokemon that triggered its ability or used a move */
+  /**
+   * Primarily for double battles, keeps track of last enemy and player pokemon that triggered its ability or used a move
+   * @todo THis is only used for Sticky Web and in a really jank way...
```
SirzBenjie:
```suggestion
   * @todo This is only used for Sticky Web and in a really jank way...
```

src/battle.ts
```diff
@@ -74,9 +74,23 @@ export class Battle {
...
+  /**
+   * The amount of coins scattered during the current battle
+   * by moves like {@linkcode MoveId.PAY_DAY | Pay Day}.
+   * Converted into money when the battle ends.
+   */
   public moneyScattered = 0;
-  /** Primarily for double battles, keeps track of last enemy and player pokemon that triggered its ability or used a move */
+  // TODO: These trackers are only used for Sticky Web + Mirror Armor edge cases
+  // and are abhorrently janky.
+  // We
```
DayKev:
```suggestion
```

src/battle.ts
```diff
@@ -80,6 +80,8 @@ export class Battle {
   public battleSeed: string = randomString(16, true);
   private battleSeedState: string | null = null;
   public moneyScattered = 0;
+  // TODO: These trackers are only used for Sticky Web + Mirror Armor edge cases
+  // and are abhorrently janky.
```
Wlowscha, line 85:
```suggestion
```

Bertie690: no im keeping that in

Have you _SEEN_ that phase

SirzBenjie (approved):

PR #6222, opened 2025-08-06
[Bug] [Ability] [Move] Update ability and move logic to use `getMoveCategory` instead of `move.category`

### LINKED ISSUE #5066: [Bug] DefDefattr (Psyshock, Psystrike, Secret Sword) do not Consider Fur Coat in Damage Calculation

### LINKED ISSUE #6045: [Bug] VariableMoveCategoryAttr Moves, (Tera Blast, Shell Side Arm, Starstorm) do not consider Physical-Activating Abilities
### Describe the bug

Tera Blast, Shell Side Arm and Tera Starstorm do not activate Abilities that would activate after being hit by a move in the Physical Category and always act as moves in the Special category.
This causes those moves to always activate moves that use `move.category === MoveCategory.SPECIAL, such as Ice Scales and Flare Boost, and never activate Abilities that use `move.category === MoveCategory.PHYSICAL` such as Ice Face and Weak Armor.

### Reproduction

Start with a Neutral Nature Physical-Leaning Pokemon, such as Spearow or Bagon
-Change Opp Ability to Ice Scales and compare damage pre-Gastro Acid
```
	MOVESET_OVERRIDE: [MoveId.SHELL_SIDE_ARM, MoveId.GASTRO_ACID, MoveId.TERA_BLAST],
	ABILITY_OVERRIDE: AbilityId.SYNCHRONIZE,

	OPP_LEVEL_OVERRIDE: 4,
	OPP_SPECIES_OVERRIDE: SpeciesId.GLALIE,

	OPP_MOVESET_OVERRIDE: [MoveId.SPLASH],
	OPP_IVS_OVERRIDE: [31, 31, 31, 31, 31, 31],
	OPP_ABILITY_OVERRIDE: AbilityId.WEAK_ARMOR,

```

### Expected behavior

Shell Side arm. Tera Blast, and Tera Starstorm should activate on Abilities dependent on the Forecasted Category.

https://replay.pokemonshowdown.com/gen9customgame-2393942726-1toat52c2dll193mvlg60h0i3hh641epw

### Screenshots / Videos

Physical Shell Side Arm Activates Ice Scales

https://github.com/user-attachments/assets/71e22cdb-8df3-44e1-aa71-bbfa1ac9da81

Physical Shell Side Arm does not activate Weak Armor

https://github.com/user-attachments/assets/51861adc-2c0e-4750-9521-aeb56d9bd13c

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

thisPieonFire:
 Fur Coat will now be correctly considered in moves that are Special but target Defense instead of Special Defense, like Psyshock, Psystrike and Secret Sword. 

Variable Moves that can target either Defense or Special Defense will now also be considered correctly.

Fixes #5066 and fixes #6045 

Builds further on the requests made in #5963

Updated the logic used in ability.ts and move.ts to use getMoveCategory instead of move.category. 
Added tests for the Fur Coat ability

## How to test the changes?

pnpm test:silent fur-coat

## Checklist

### DIFF
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -1982,7 +1982,7 @@ export class AllyMoveCategoryPowerBoostAbAttr extends FieldMovePowerBoostAbAttr 
    */
   constructor(boostedCategories: MoveCategory[], powerMultiplier: number) {
     super((_pokemon, _defender, move) => {
-      if (_pokemon === null) {
+      if (isNullOrUndefined(pokemon)) {
         return false;
       }
       return boostedCategories.includes(_pokemon.getMoveCategory(_defender, move));
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -1981,7 +1981,12 @@ export class AllyMoveCategoryPowerBoostAbAttr extends FieldMovePowerBoostAbAttr 
    * @param powerMultiplier - The multiplier to apply to the move's power.
    */
   constructor(boostedCategories: MoveCategory[], powerMultiplier: number) {
-    super((_pokemon, _defender, move) => boostedCategories.includes(move.category), powerMultiplier);
+    super((_pokemon, _defender, move) => {
+      if (_pokemon === null) {
+        return false;
+      }
+      return boostedCategories.includes(_pokemon.getMoveCategory(_defender, move));
+    }, powerMultiplier);
   }
 }
 
@@ -2113,7 +2118,13 @@ export abstract class PostAttackAbAttr extends AbAttr {
 
   /** The default `attackCondition` requires that the selected move is a damaging move */
   constructor(
-    attackCondition: PokemonAttackCondition = (_user, _target, move) => move.category !== MoveCategory.STATUS,
+    attackCondition: PokemonAttackCondition = (_user, _target, move) => {
+      if (!_user) {
+        return false;
+      }
+
+      return _user.getMoveCategory(_target, move) !== MoveCategory.STATUS;
+    },
     showAbility = true,
   ) {
     super(showAbility);
@@ -6908,7 +6919,12 @@ export function initAbilities() {
       .attr(PostSummonAddBattlerTagAbAttr, BattlerTagType.TRUANT, 1, false),
     new Ability(AbilityId.HUSTLE, 3)
       .attr(StatMultiplierAbAttr, Stat.ATK, 1.5)
-      .attr(StatMultiplierAbAttr, Stat.ACC, 0.8, (_user, _target, move) => move.category === MoveCategory.PHYSICAL),
+      .attr(StatMultiplierAbAttr, Stat.ACC, 0.8, (_user, _target, move) => {
+        if (_user === null) {
+          return false
+        }
+          return _user.getMoveCategory(_target, move) === MoveCategory.PHYSICAL
+      }),
     new Ability(AbilityId.CUTE_CHARM, 3)
       .attr(PostDefendContactApplyTagChanceAbAttr, 30, BattlerTagType.INFATUATED),
     new Ability(AbilityId.PLUS, 3)
@@ -7140,8 +7156,8 @@ export function initAbilities() {
       .attr(AlliedFieldDamageReductionAbAttr, 0.75)
       .ignorable(),
     new Ability(AbilityId.WEAK_ARMOR, 5)
-      .attr(PostDefendStatStageChangeAbAttr, (_target, _user, move) => move.category === MoveCategory.PHYSICAL, Stat.DEF, -1)
-      .attr(PostDefendStatStageChangeAbAttr, (_target, _user, move) => move.category === MoveCategory.PHYSICAL, Stat.SPD, 2),
+      .attr(PostDefendStatStageChangeAbAttr, (_target, _user, move) => _user.getMoveCategory(_target, move) === MoveCategory.PHYSICAL, Stat.DEF, -1)
+      .attr(PostDefendStatStageChangeAbAttr, (_target, _user, move) => _user.getMoveCategory(_target, move) === MoveCategory.PHYSICAL, Stat.SPD, 2),
     new Ability(AbilityId.HEAVY_METAL, 5)
       .attr(WeightMultiplierAbAttr, 2)
       .ignorable(),
@@ -7152,9 +7168,15 @@ export function initAbilities() {
       .attr(ReceivedMoveDamageMultiplierAbAttr, (target, _user, _move) => target.isFullHp(), 0.5)
       .ignorable(),
     new Ability(AbilityId.TOXIC_BOOST, 5)
-      .attr(MovePowerBoostAbAttr, (user, _target, move) => move.category === MoveCategory.PHYSICAL && (user?.status?.effect === StatusEffect.POISON || user?.status?.effect === StatusEffect.TOXIC), 1.5),
+      .attr(MovePowerBoostAbAttr, (user, _target, move) => {
+          if (user===null) { return false; }
+          return user.getMoveCategory(_target, move) === MoveCategory.PHYSICAL && (user?.status?.effect === StatusEffect.POISON || user?.status?.effect === StatusEffect.TOXIC)
+      }, 1.5),
     new Ability(AbilityId.FLARE_BOOST, 5)
-      .attr(MovePowerBoostAbAttr, (user, _target, move) => move.category === MoveCategory.SPECIAL && user?.status?.effect === StatusEffect.BURN, 1.5),
+      .attr(MovePowerBoostAbAttr, (user, _target, move) => {
+          if (user===null) { return false; }
+          return user.getMoveCategory(_target, move) === MoveCategory.SPECIAL && user?.status?.effect === StatusEffect.BURN
+      }, 1.5),
     new Ability(AbilityId.HARVEST, 5)
       .attr(
         PostTurnRestoreBerryAbAttr,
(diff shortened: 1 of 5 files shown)


src/data/abilities/ability.ts
```diff
@@ -1982,7 +1982,12 @@ export class AllyMoveCategoryPowerBoostAbAttr extends FieldMovePowerBoostAbAttr
    * @param powerMultiplier - The multiplier to apply to the move's power.
    */
   constructor(boostedCategories: MoveCategory[], powerMultiplier: number) {
-    super((_pokemon, _defender, move) => boostedCategories.includes(move.category), powerMultiplier);
+    super((_pokemon, _defender, move) => {
+      if (_pokemon === null) {
+        return false;
+      }
+      return boostedCategories.includes(_pokemon.getMoveCategory(_defender, move));
+    }, powerMultiplier);
```
emdeann: A `_` prefixed to a variable name is an indicator that it's unused. You are now using the variable holding the pokemon, so you should rename it `_pokemon` -> `pokemon` (along with making similar changes for all other instances of this)

src/data/abilities/ability.ts
```diff
@@ -1982,7 +1982,12 @@ export class AllyMoveCategoryPowerBoostAbAttr extends FieldMovePowerBoostAbAttr
    * @param powerMultiplier - The multiplier to apply to the move's power.
    */
   constructor(boostedCategories: MoveCategory[], powerMultiplier: number) {
-    super((_pokemon, _defender, move) => boostedCategories.includes(move.category), powerMultiplier);
+    super((_pokemon, _defender, move) => {
+      if (_pokemon === null) {
```
emdeann:
```suggestion
      if (isNullOrUndefined(pokemon)) {
```

src/data/abilities/ability.ts
```diff
@@ -2114,7 +2119,13 @@ export abstract class PostAttackAbAttr extends AbAttr {
 
   /** The default `attackCondition` requires that the selected move is a damaging move */
   constructor(
-    attackCondition: PokemonAttackCondition = (_user, _target, move) => move.category !== MoveCategory.STATUS,
+    attackCondition: PokemonAttackCondition = (_user, _target, move) => {
```
emdeann: `_user` -> `user`

src/data/abilities/ability.ts
```diff
@@ -2114,7 +2119,13 @@ export abstract class PostAttackAbAttr extends AbAttr {
...
+      if (!_user) {
```
emdeann:
```suggestion
      if (isNullOrUndefined(user)) {
```

src/data/abilities/ability.ts
```diff
@@ -7311,8 +7333,15 @@ export function initAbilities() {
...
       .edgeCase(),
     new Ability(AbilityId.FUR_COAT, 6)
-      .attr(ReceivedMoveDamageMultiplierAbAttr, (_target, _user, move) => move.category === MoveCategory.PHYSICAL, 0.5)
+         .attr(ReceivedMoveDamageMultiplierAbAttr, (_target, _user, move) => {
+        const isPhysicalMove = _user.getMoveCategory(_target, move) === MoveCategory.PHYSICAL
+        // This is for a theoretical move where Fur Coat shouldn't apply, in case it's implemented later
+        const moveIsPhysicalAndActuallyHitsDef = isPhysicalMove && !move.hasAttr("VariableDefAttr")
+        const moveIsSpecialButHitsPhysicalInstead = move.hasAttr("DefDefAttr")
+        return moveIsPhysicalAndActuallyHitsDef || moveIsSpecialButHitsPhysicalInstead
+      }, 0.5)
```
emdeann: I oversimplified a little bit in my last comment because I didn't consider that something like psyshock is still considered a special move even though it hits physical defense. 

In that light this kind of implementation for Fur Coat would always be janky. It seems to me that the way to do this properly would be to have fur coat be a `StatMultiplierAbAttr` that simply doubles the user's defense instead of needing to consider the properties of every incoming move. That would have to consider confusion damage, though

src/data/abilities/ability.ts
emdeann: I also don't believe that you need to make this check at all, the parameters passed to these functions should not be null or undefined. Worth someone else confirming that though

src/data/abilities/ability.ts
thisPieonFire: That could also work, but wouldn't that solution also have to consider certain arena tags and recoil?

Edit: I read the bulbapedia page and that only mentions confusion. No other page seems to mention anything like that (including confusion), as far as I can see

thisPieonFire:
Oops closed it accidentally, reopening

src/data/abilities/ability.ts
emdeann: Sorry, didn't see this. We might need to investigate ingame to see exactly what (if anything) is inexplicably ignored by fur coat. We could also make the decision that fur coat doubling defense in all but one or two arbitrary scenarios is stupid and not duplicate that behavior - we have done similar things in the past.

My main point here is that an ability should never itself be doing the logic of "does this move use this stat"

DayKev: It seems like the issue is that our implementation of Fur Coat is halving the damage, when it should be doubling the user's Defense. Well, we need some testing to figure out which one it should actually be doing, or we can just trust that Showdown and Bulbapedia have it right.

thisPieonFire: I do agree that doubling the defense instead would get rid of a lot of the weird edgecases, except for confusion damage, of course. We (and by that I probably mean "I") can absolutely change it to that instead if people agree. I'm not sure how to test what the correct option should be though.

DayKev: You'd need to hit an Alolan Persian with a physical move in SV, SwSh or SM/USUM (or Furfrou in XY or SM/USUM) over and over until you map out the possible damage values, then see if those values match halved damage or doubled defense based on the damage formula.

Interestingly, Showdown/Bulbapedia indicate that Ice Scales does halved damage and not doubled SpDef, but Fur Coat does doubled defense and not halved damage. They were introduced in different generations (Fur Coat in 6 and Ice Scales in 8) so it's possible GameFreak was inconsistent here, but it's also possible that Showdown/Bulbapedia have one of them wrong as well.

thisPieonFire: So does it need to be tested? I don't have the stuff to test that out, really.

DayKev: For now change Fur Coat to do 2x defense instead of halving damage, since that's simpler to implement and matches with Showdown & Bulbapedia. Also don't worry about confusion damage.

thisPieonFire: That I can do! Any specific reason I don't have to add the confusion damage check?

DayKev: Now that I'm taking another look, it seems like confusion just needs to ignore abilities when calculating the attack and defense stats to use (that is, pass `true` for the three `ignoreAbility` parameters to `getEffectiveStat`).
https://github.com/pagefaultgames/pokerogue/blob/e25db1632608c03433c7f334f853112f3315c651/src/data/battler-tags.ts#L836-L837
Basically, change that to:
```ts
      const atk = pokemon.getEffectiveStat(Stat.ATK, undefined, undefined, true, true, true);
      const def = pokemon.getEffectiveStat(Stat.DEF, undefined, undefined, true, true, true);
```
So you can make that change too.

DayKev: Well good news, we discovered an easy way to test how it actually works, and both Fur Coat and Ice Scales do half damage rather than stat doubling, so you don't need to change anything after all (at least not to doing stat modification).

The confusion change I mentioned does still need to be made but it's not related to these abilities anymore.

SirzBenjie: @thisPieonFire 

Correct implementation should be to use the ability attribute you were given, but to check which damage category is being used.

Secret sword / psyshock are unaffected by fur coat. They are affected by ice scales.
Those moves care about the stat being used.

Ice scales and Fur Coat care about the damage _category_ being dealt. Like whether it is physical or special. It works exactly like reflect / light screen. So Psyshock and secret sword are reduced by ice scales just like they are by light screen (and not reflect).

PR #6229, opened 2025-08-07
[Misc] Removed `populateAnims` script

Bertie690:
Marginally faster load times

Keeping around random ass scripts we need to use random downgraded deps for is bad

Removed the code

## Screenshots/Videos
N/A

## How to test the changes?
make sure game still boots???

## Checklist


DayKev (approved):
We'll need to remember that the script is here so we can get it back if we need to import more animations in the future (such as when Gen 10 is released).

Bertie690:
If you want I can comment it out instead of deleting, but hopefully by then we'd have figured out some way to store them _other_ than giant freaking JSONs...

damocleas (approved):

PR #6211, opened 2025-08-04
Reward type safety

Bertie690:
## TL;DR

Instead of using `super()` with callbacks (a frankly abhorrent way to go about doing inheritance), all generators now override an abstract `generateReward` function with individual typing on each subclass.
Instead of passing a party in a callback, we access it globally from `globalScene`.

All generators have had their `RewardId`s removed, and proper type safety is enforced through a slew of new types in `#types/reward.ts` and a copy-pasted typing of `allRewards`.

Also, `generateRewardOptionFromSpecs` has been merged into `generateRewardOptionFromId` as an overload as it reduces the number of both separate functions and random type parameters.

## Next steps
- Removing prior references to `allRewards.XYZ` in MEs
- Reworking `allRewards` into a proper `const object` so i don't need to have a duplicate type for it
- Fixing syntax errors
- further refinement
- Possibly renaming `generateRewardOptionFromId` into `generateRewardOption`?

PR #6232, opened 2025-08-07
Remove `RewardFunc`

Wlowscha:
This PR is removing the `RewardFunc` type and making the `allRewards` object directly contain instances of `Reward` and `RewardGenerator`.


Wlowscha:
This will be incorporated into #6211

PR #6233, opened 2025-08-07
Added test utils for held item tests

Bertie690:
Title

### DIFF
--- a/src/items/held-item-manager.ts
+++ b/src/items/held-item-manager.ts
@@ -8,6 +8,7 @@ import {
   isItemInRequested,
 } from "#enums/held-item-id";
 import {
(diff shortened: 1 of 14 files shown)


src/items/held-item-manager.ts
```diff
@@ -219,9 +227,14 @@ export class HeldItemManager {
   }
 
   getFormChangeItems(): FormChangeItem[] {
+    // TODO: Please stop using `map(k => k)`
     return getTypedKeys(this.formChangeItems).map(k => k);
```
DayKev:
```suggestion
    return getTypedKeys(this.formChangeItems);
```
Just remove the `.map` then?

PR #6241, opened 2025-08-09
[Refactor/Test] Cleaned up `getAttackTypeEffectiveness` and removed `VariableMoveTypeMultiplierAttr`

Bertie690:
- **Cleaned up flying press/etc moves; removed `VariableMoveTypeMultiplierAttr`**
- **Made `getAttackTypeEffectiveness` take an object for parameters; added FP tests**

N/A

`VariableMoveTypeMultiplierAttr` and `VariableMoveTypeChartAttr` were used in extremely overlapping ways.

The only difference was the former operated on all types, while the latter operated on each individual type.

Also big function with many args usually like to have object literal

The aforementioned 2 attributes have been combined into 1 attribute that is used for all the prior moves (Freeze-Dry, Flying Press, Sheer Cold, Thousand Arrows and Synchronoise)

Moved params into an object and updated callsites

stopped anticipation from hardcoding Hidden Power's type check

## Screenshots/Videos
N/A
## How to test the changes?
`pnpm test flying-press.test.ts`

## Checklist

### DIFF
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -4188,71 +4188,43 @@ function getWeatherCondition(...weatherTypes: WeatherType[]): AbAttrCondition {
     if (globalScene.arena.weather?.isEffectSuppressed()) {
       return false;
     }
-    const weatherType = globalScene.arena.weather?.weatherType;
-    return !!weatherType && weatherTypes.indexOf(weatherType) > -1;
+    return weatherTypes.includes(globalScene.arena.getWeatherType());
   };
 }
 
-function getAnticipationCondition(): AbAttrCondition {
-  return (pokemon: Pokemon) => {
-    for (const opponent of pokemon.getOpponents()) {
-      for (const move of opponent.moveset) {
-        // ignore null/undefined moves
-        if (!move) {
-          continue;
-        }
-        // the move's base type (not accounting for variable type changes) is super effective
-        if (
-          move.getMove().is("AttackMove") &&
-          pokemon.getAttackTypeEffectiveness(move.getMove().type, opponent, true, undefined, move.getMove()) >= 2
-        ) {
-          return true;
-        }
-        // move is a OHKO
-        if (move.getMove().hasAttr("OneHitKOAttr")) {
-          return true;
-        }
-        // edge case for hidden power, type is computed
-        if (move.getMove().id === MoveId.HIDDEN_POWER) {
-          const iv_val = Math.floor(
-            (((opponent.ivs[Stat.HP] & 1) +
-              (opponent.ivs[Stat.ATK] & 1) * 2 +
-              (opponent.ivs[Stat.DEF] & 1) * 4 +
-              (opponent.ivs[Stat.SPD] & 1) * 8 +
-              (opponent.ivs[Stat.SPATK] & 1) * 16 +
-              (opponent.ivs[Stat.SPDEF] & 1) * 32) *
-              15) /
-              63,
-          );
-
-          const type = [
-            PokemonType.FIGHTING,
-            PokemonType.FLYING,
-            PokemonType.POISON,
-            PokemonType.GROUND,
-            PokemonType.ROCK,
-            PokemonType.BUG,
-            PokemonType.GHOST,
-            PokemonType.STEEL,
-            PokemonType.FIRE,
-            PokemonType.WATER,
-            PokemonType.GRASS,
-            PokemonType.ELECTRIC,
-            PokemonType.PSYCHIC,
-            PokemonType.ICE,
-            PokemonType.DRAGON,
-            PokemonType.DARK,
-          ][iv_val];
-
-          if (pokemon.getAttackTypeEffectiveness(type, opponent) >= 2) {
-            return true;
-          }
-        }
+/**
+ * Condition used by {@linkcode AbilityId.ANTICIPATION} to show a message if any opponent knows a
+ * "dangerous" move.
+ * @param pokemon - The {@linkcode Pokemon} with this ability
+ * @returns Whether the message should be shown
+ */
+const anticipationCondition: AbAttrCondition = (pokemon: Pokemon) =>
+  pokemon.getOpponents().some(opponent =>
+    opponent.moveset.some(movesetMove => {
+      // ignore null/undefined moves or non-attacks
+      const move = movesetMove?.getMove();
+      if (!move?.is("AttackMove")) {
+        return false;
       }
-    }
-    return false;
-  };
-}
+
+      if (move.hasAttr("OneHitKOAttr")) {
+        return true;
+      }
+
+      // Check whether the move's base type (not accounting for variable type changes) is super effective
+      const type = new NumberHolder(
+        pokemon.getAttackTypeEffectiveness(move.type, {
+          source: opponent,
+          ignoreStrongWinds: true,
+          move: move,
+        }),
+      );
+
+      // edge case for hidden power, type is computed
+      applyMoveAttrs("HiddenPowerTypeAttr", opponent, pokemon, move, type);
+      return type.value >= 2;
+    }),
+  );
 
 /**
  * Creates an ability condition that causes the ability to fail if that ability
@@ -7083,7 +7055,7 @@ export function initAbilities() {
       .attr(PostFaintContactDamageAbAttr, 4)
       .bypassFaint(),
     new Ability(AbilityId.ANTICIPATION, 4)
-      .conditionalAttr(getAnticipationCondition(), PostSummonMessageAbAttr, (pokemon: Pokemon) => i18next.t("abilityTriggers:postSummonAnticipation", { pokemonNameWithAffix: getPokemonNameWithAffix(pokemon) })),
+      .conditionalAttr(anticipationCondition, PostSummonMessageAbAttr, (pokemon: Pokemon) => i18next.t("abilityTriggers:postSummonAnticipation", { pokemonNameWithAffix: getPokemonNameWithAffix(pokemon) })),
     new Ability(AbilityId.FOREWARN, 4)
       .attr(ForewarnAbAttr),
     new Ability(AbilityId.UNAWARE, 4)
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -730,9 +730,7 @@ export class AttackTypeImmunityAbAttr extends TypeImmunityAbAttr {
   override canApply(params: TypeMultiplierAbAttrParams): boolean {
     const { move } = params;
     return (
-      move.category !== MoveCategory.STATUS &&
-      !move.hasAttr("NeutralDamageAgainstFlyingTypeMultiplierAttr") &&
-      super.canApply(params)
+      move.category !== MoveCategory.STATUS && !move.hasAttr("VariableMoveTypeChartAttr") && super.canApply(params)
     );
   }
 }
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -67,7 +67,7 @@ import { StatusEffect } from "#enums/status-effect";
 import { SwitchType } from "#enums/switch-type";
 import { WeatherType } from "#enums/weather-type";
 import { MoveUsedEvent } from "#events/battle-scene";
-import type { EnemyPokemon, Pokemon } from "#field/pokemon";
+import { EnemyPokemon, Pokemon } from "#field/pokemon";
 import {
   AttackTypeBoosterModifier,
   BerryModifier,
@@ -5349,48 +5349,60 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
   }
 }
 
-export class VariableMoveTypeMultiplierAttr extends MoveAttr {
-  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    return false;
-  }
-}
-
-export class NeutralDamageAgainstFlyingTypeMultiplierAttr extends VariableMoveTypeMultiplierAttr {
-  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    if (!target.getTag(BattlerTagType.IGNORE_FLYING)) {
-      const multiplier = args[0] as NumberHolder;
-      //When a flying type is hit, the first hit is always 1x multiplier.
-      if (target.isOfType(PokemonType.FLYING)) {
-        multiplier.value = 1;
-      }
-      return true;
-    }
-
-    return false;
-  }
-}
-
-export class IceNoEffectTypeAttr extends VariableMoveTypeMultiplierAttr {
+/**
+ * Attribute for moves which have a custom type chart interaction.
+ */
+export class VariableMoveTypeChartAttr extends MoveAttr {
   /**
-   * Checks to see if the Target is Ice-Type or not. If so, the move will have no effect.
-   * @param user n/a
-   * @param target The {@linkcode Pokemon} targeted by the move
-   * @param move n/a
-   * @param args `[0]` a {@linkcode NumberHolder | NumberHolder} containing a type effectiveness multiplier
-   * @returns `true` if this Ice-type immunity applies; `false` otherwise
+   * @param user - The {@linkcode Pokemon} using the move
+   * @param target - The {@linkcode Pokemon} targeted by the move
+   * @param move - The {@linkcode Move} with this attribute
+   * @param args -
+   * `[0]`: A {@linkcode NumberHolder} holding the type effectiveness
+   * `[1]`: The target's entire defensive type profile.
+   * @returns `true` if application of the attribute succeeds
    */
-  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    const multiplier = args[0] as NumberHolder;
-    if (target.isOfType(PokemonType.ICE)) {
-      multiplier.value = 0;
+  apply(user: Pokemon, target: Pokemon, move: Move, args: [multiplier: NumberHolder, types: PokemonType[]]): boolean {
+    return false;
+  }
+}
+
+/**
+ * Attribute to implement {@linkcode MoveId.FREEZE_DRY}'s guaranteed water type super effectiveness.
+ */
+export class FreezeDryAttr extends VariableMoveTypeChartAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: [NumberHolder, PokemonType[]]): boolean {
+    const [multiplier, types] = args;
+
+    if (types.includes(PokemonType.WATER)) {
+      multiplier.value = Math.max(multiplier.value, 2);
       return true;
     }
     return false;
   }
 }
 
-export class FlyingTypeMultiplierAttr extends VariableMoveTypeMultiplierAttr {
-  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+/**
+ * Attribute used by {@linkcode MoveId.THOUSAND_ARROWS} to cause it to deal a fixed 1x damage
+ * against all ungrounded flying types.
+ */
+export class NeutralDamageAgainstFlyingTypeAttr extends VariableMoveTypeChartAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: [multiplier: NumberHolder, types: PokemonType[]]): boolean {
+    const [multiplier, types] = args;
+    if (target.isGrounded() || !types.includes(PokemonType.FLYING)) {
+      return false;
+    }
+    multiplier.value = 1;
+
+    return true;
+  }
+}
+
+/**
+ * Attribute used by {@linkcode MoveId.FLYING_PRESS} to add the Flying Type to its type effectiveness.
+ */
+export class FlyingTypeMultiplierAttr extends VariableMoveTypeChartAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: [multiplier: NumberHolder, types: PokemonType[]]): boolean {
     const multiplier = args[0] as NumberHolder;
     multiplier.value *= target.getAttackTypeEffectiveness(PokemonType.FLYING, user);
     return true;
@@ -5398,39 +5410,19 @@ export class FlyingTypeMultiplierAttr extends VariableMoveTypeMultiplierAttr {
 }
 
 /**
- * Attribute for moves which have a custom type chart interaction.
+ * Attribute used by {@linkcode MoveId.SHEER_COLD} to implement its Gen VII+ ice ineffectiveness.
  */
-export class VariableMoveTypeChartAttr extends MoveAttr {
-  /**
-   * @param user {@linkcode Pokemon} using the move
-   * @param target {@linkcode Pokemon} target of the move
-   * @param move {@linkcode Move} with this attribute
-   * @param args [0] {@linkcode NumberHolder} holding the type effectiveness
-   * @param args [1] A single defensive type of the target
-   *
-   * @returns true if application of the attribute succeeds
-   */
-  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
+export class IceNoEffectTypeAttr extends VariableMoveTypeChartAttr {
+  apply(user: Pokemon, target: Pokemon, move: Move, args: [multiplier: NumberHolder, types: PokemonType[]]): boolean {
+    const [multiplier, types] = args;
+    if (types.includes(PokemonType.ICE)) {
+      multiplier.value = 0;
+      return true;
+    }
     return false;
   }
 }
 
-/**
- * This class forces Freeze-Dry to be super effective against Water Type.
- */
-export class FreezeDryAttr extends VariableMoveTypeChartAttr {
-  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    const multiplier = args[0] as NumberHolder;
-    const defType = args[1] as PokemonType;
-
-    if (defType === PokemonType.WATER) {
-      multiplier.value = 2;
-      return true;
-    } else {
-      return false;
-    }
-  }
-}
 
 export class OneHitKOAccuracyAttr extends VariableAccuracyAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
@@ -5916,8 +5908,8 @@ export class ProtectAttr extends AddBattlerTagAttr {
       for (const turnMove of user.getLastXMoves(-1).slice()) {
         if (
           // Quick & Wide guard increment the Protect counter without using it for fail chance
-          !(allMoves[turnMove.move].hasAttr("ProtectAttr") || 
-          [MoveId.QUICK_GUARD, MoveId.WIDE_GUARD].includes(turnMove.move)) || 
+          !(allMoves[turnMove.move].hasAttr("ProtectAttr") ||
+          [MoveId.QUICK_GUARD, MoveId.WIDE_GUARD].includes(turnMove.move)) ||
           turnMove.result !== MoveResult.SUCCESS
         ) {
           break;
@@ -8146,7 +8138,9 @@ export class UpperHandCondition extends MoveCondition {
   }
 }
 
-export class HitsSameTypeAttr extends VariableMoveTypeMultiplierAttr {
+// TODO: Does this need to extend from this?
+// The only reason it might is to show ineffectiveness text but w/e
+export class HitsSameTypeAttr extends VariableMoveTypeChartAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
     const multiplier = args[0] as NumberHolder;
     if (!user.getTypes(true).some(type => target.getTypes(true).includes(type))) {
@@ -8417,8 +8411,7 @@ const MoveAttrs = Object.freeze({
   TeraStarstormTypeAttr,
   MatchUserTypeAttr,
   CombinedPledgeTypeAttr,
-  VariableMoveTypeMultiplierAttr,
-  NeutralDamageAgainstFlyingTypeMultiplierAttr,
+  NeutralDamageAgainstFlyingTypeAttr,
   IceNoEffectTypeAttr,
   FlyingTypeMultiplierAttr,
   VariableMoveTypeChartAttr,
@@ -10454,7 +10447,7 @@ export function initMoves() {
       .attr(HitHealAttr, 0.75)
       .triageMove(),
     new AttackMove(MoveId.THOUSAND_ARROWS, PokemonType.GROUND, MoveCategory.PHYSICAL, 90, 100, 10, -1, 0, 6)
-      .attr(NeutralDamageAgainstFlyingTypeMultiplierAttr)
+      .attr(NeutralDamageAgainstFlyingTypeAttr)
       .attr(FallDownAttr)
       .attr(HitsTagAttr, BattlerTagType.FLYING)
       .attr(HitsTagAttr, BattlerTagType.FLOATING)
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -999,7 +999,7 @@ export class AttackMove extends Move {
     const ret = super.getTargetBenefitScore(user, target, move);
     let attackScore = 0;
 
-    const effectiveness = target.getAttackTypeEffectiveness(this.type, user, undefined, undefined, this);
+    const effectiveness = target.getAttackTypeEffectiveness(this.type, {source: user, move: this});
     attackScore = Math.pow(effectiveness - 1, 2) * (effectiveness < 1 ? -2 : 2);
     const [ thisStat, offStat ]: EffectiveStat[] = this.category === MoveCategory.PHYSICAL ? [ Stat.ATK, Stat.SPATK ] : [ Stat.SPATK, Stat.ATK ];
     const statHolder = new NumberHolder(user.getEffectiveStat(thisStat, target));
@@ -1795,7 +1795,7 @@ export class SacrificialAttr extends MoveEffectAttr {
     if (user.isBoss()) {
       return -20;
     }
-    return Math.ceil(((1 - user.getHpRatio()) * 10 - 10) * (target.getAttackTypeEffectiveness(move.type, user) - 0.5));
+    return Math.ceil(((1 - user.getHpRatio()) * 10 - 10) * (target.getAttackTypeEffectiveness(move.type, {source: user}) - 0.5));
   }
 }
 
@@ -1833,7 +1833,7 @@ export class SacrificialAttrOnHit extends MoveEffectAttr {
     if (user.isBoss()) {
       return -20;
     }
-    return Math.ceil(((1 - user.getHpRatio()) * 10 - 10) * (target.getAttackTypeEffectiveness(move.type, user) - 0.5));
+    return Math.ceil(((1 - user.getHpRatio()) * 10 - 10) * (target.getAttackTypeEffectiveness(move.type, {source: user}) - 0.5));
   }
 }
 
@@ -1875,7 +1875,7 @@ export class HalfSacrificialAttr extends MoveEffectAttr {
     if (user.isBoss()) {
       return -10;
     }
-    return Math.ceil(((1 - user.getHpRatio() / 2) * 10 - 10) * (target.getAttackTypeEffectiveness(move.type, user) - 0.5));
+    return Math.ceil(((1 - user.getHpRatio() / 2) * 10 - 10) * (target.getAttackTypeEffectiveness(move.type, {source: user}) - 0.5));
   }
 }
 
@@ -5402,9 +5402,9 @@ export class NeutralDamageAgainstFlyingTypeAttr extends VariableMoveTypeChartAtt
  * Attribute used by {@linkcode MoveId.FLYING_PRESS} to add the Flying Type to its type effectiveness.
  */
 export class FlyingTypeMultiplierAttr extends VariableMoveTypeChartAttr {
-  apply(user: Pokemon, target: Pokemon, move: Move, args: [multiplier: NumberHolder, types: PokemonType[]]): boolean {
-    const multiplier = args[0] as NumberHolder;
-    multiplier.value *= target.getAttackTypeEffectiveness(PokemonType.FLYING, user);
+  apply(user: Pokemon, target: Pokemon, _move: Move, args: [multiplier: NumberHolder, types: PokemonType[]]): boolean {
+    const multiplier = args[0];
+    multiplier.value *= target.getAttackTypeEffectiveness(PokemonType.FLYING, {source: user});
     return true;
   }
 }
@@ -11383,9 +11383,10 @@ export function initMoves() {
     new AttackMove(MoveId.RUINATION, PokemonType.DARK, MoveCategory.SPECIAL, -1, 90, 10, -1, 0, 9)
       .attr(TargetHalfHpDamageAttr),
     new AttackMove(MoveId.COLLISION_COURSE, PokemonType.FIGHTING, MoveCategory.PHYSICAL, 100, 100, 5, -1, 0, 9)
-      .attr(MovePowerMultiplierAttr, (user, target, move) => target.getAttackTypeEffectiveness(move.type, user) >= 2 ? 5461 / 4096 : 1),
+      // TODO: Do we want to change this to 4/3?
+      .attr(MovePowerMultiplierAttr, (user, target, move) => target.getAttackTypeEffectiveness(move.type, {source: user}) >= 2 ? 5461 / 4096 : 1),
     new AttackMove(MoveId.ELECTRO_DRIFT, PokemonType.ELECTRIC, MoveCategory.SPECIAL, 100, 100, 5, -1, 0, 9)
-      .attr(MovePowerMultiplierAttr, (user, target, move) => target.getAttackTypeEffectiveness(move.type, user) >= 2 ? 5461 / 4096 : 1)
+      .attr(MovePowerMultiplierAttr, (user, target, move) => target.getAttackTypeEffectiveness(move.type, {source: user}) >= 2 ? 5461 / 4096 : 1)
       .makesContact(),
     new SelfStatusMove(MoveId.SHED_TAIL, PokemonType.NORMAL, -1, 10, -1, 0, 9)
       .attr(AddSubstituteAttr, 0.5, true)
--- a/src/data/type.ts
+++ b/src/data/type.ts
@@ -2,6 +2,13 @@ import { PokemonType } from "#enums/pokemon-type";
 
 export type TypeDamageMultiplier = 0 | 0.125 | 0.25 | 0.5 | 1 | 2 | 4 | 8;
 
+/**
+ * Get the type effectiveness multiplier of one PokemonType against another.
+ * @param attackType - The {@linkcode PokemonType} of the attacker
+ * @param defType - The {@linkcode PokemonType} of the defender
+ * @returns The type damage multiplier between the two types;
+ * will be either `0`, `0.5`, `1` or `2`.
+ */
 export function getTypeDamageMultiplier(attackType: PokemonType, defType: PokemonType): TypeDamageMultiplier {
   if (attackType === PokemonType.UNKNOWN || defType === PokemonType.UNKNOWN) {
     return 1;
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -2400,7 +2400,6 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
         : 1,
     );
 
-    applyMoveAttrs("VariableMoveTypeMultiplierAttr", source, this, move, typeMultiplier);
     if (this.getTypes(true, true).find(t => move.isTypeImmune(source, this, t))) {
       typeMultiplier.value = 0;
     }
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -129,7 +129,8 @@ import {
   TempStatStageBoosterModifier,
 } from "#modifiers/modifier";
 import { applyMoveAttrs } from "#moves/apply-attrs";
-import type { Move } from "#moves/move";
+// biome-ignore lint/correctness/noUnusedImports: TSDoc
+import type { Move, VariableMoveTypeChartAttr } from "#moves/move";
 import { getMoveTargets } from "#moves/move-utils";
 import { PokemonMove } from "#moves/pokemon-move";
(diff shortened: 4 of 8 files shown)


src/data/abilities/ability.ts
```diff
@@ -4188,71 +4186,43 @@ function getWeatherCondition(...weatherTypes: WeatherType[]): AbAttrCondition {
...
+/**
+ * Condition used by {@linkcode AbilityId.ANTICIPATION} to show a message if any opponent knows a
+ * "dangerous" move.
+ * @param pokemon - The {@linkcode Pokemon} with this ability
+ * @returns Whether the message should be shown
+ */
+const anticipationCondition: AbAttrCondition = (pokemon: Pokemon) =>
+  pokemon.getOpponents().some(opponent =>
+    opponent.moveset.some(movesetMove => {
+      // ignore null/undefined moves or non-attacks
```
DayKev: There should never be `null` in a Pokémon's moveset, that would break everything.

test/battle/inverse-battle.test.ts
```diff
@@ -164,6 +149,7 @@ describe("Inverse Battle", () => {
     expect(enemy.status?.effect).not.toBe(StatusEffect.PARALYSIS);
   });
 
+  // TODO: These should belong to their respective moves' test files, not the inverse battle mechanic itself
```
DayKev: Why would the tests specific to inverse battle not be in the inverse battle test file?

src/data/abilities/ability.ts
Bertie690: ah, probably a leftover from a while ago.

src/@types/damage-params.ts
```diff
@@ -1,9 +1,10 @@
 import type { MoveCategory } from "#enums/move-category";
 import type { Pokemon } from "#field/pokemon";
-import type { Move } from "#types/move-types";
+// biome-ignore lint/correctness/noUnusedImports: TSDoc
```
DayKev:
```suggestion
```

src/@types/damage-params.ts
```diff
@@ -32,13 +33,45 @@ export interface damageParams {
 }
 
 /**
- * Type for the parameters of {@linkcode Pokemon#getBaseDamage | getBaseDamage}
+ * Type for the parameters of {@linkcode Pokemon#=.getBaseDamage | getBaseDamage}
```
DayKev:
```suggestion
 * Type for the parameters of {@linkcode Pokemon.getBaseDamage}
```

src/@types/damage-params.ts
```diff
@@ -32,13 +33,45 @@ export interface damageParams {
...
  * @interface
  */
 export type getBaseDamageParams = Omit<damageParams, "effectiveness">;
 
 /**
- * Type for the parameters of {@linkcode Pokemon#getAttackDamage | getAttackDamage}
+ * Type for the parameters of {@linkcode Pokemon#=.getAttackDamage | getAttackDamage}
```
DayKev:
```suggestion
 * Type for the parameters of {@linkcode Pokemon.getAttackDamage}
```

src/@types/damage-params.ts
```diff
@@ -32,13 +33,45 @@ export interface damageParams {
...
  * @interface
  */
 export type getAttackDamageParams = Omit<damageParams, "moveCategory">;
+
+/**
+ * Type for the parameters of {@linkcode Pokemon.getAttackTypeEffectiveness | getAttackTypeEffectiveness}
```
DayKev:
```suggestion
 * Type for the parameters of {@linkcode Pokemon.getAttackTypeEffectiveness}
```

src/data/abilities/ability.ts
```diff
@@ -4239,71 +4239,42 @@ function getWeatherCondition(...weatherTypes: WeatherType[]): AbAttrCondition {
...
+/**
+ * Condition used by {@linkcode AbilityId.ANTICIPATION} to show a message if any opponent knows a
+ * "dangerous" move.
+ * @param pokemon - The {@linkcode Pokemon} with this ability
+ * @returns Whether the message should be shown
+ */
+const anticipationCondition: AbAttrCondition = (pokemon: Pokemon) =>
+  pokemon.getOpponents().some(opponent =>
+    opponent.moveset.some(movesetMove => {
+      // ignore non-attacks
```
```suggestion
```

src/data/moves/move.ts
```diff
@@ -5999,82 +6002,127 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
...
 }
 
-export class VariableMoveTypeMultiplierAttr extends MoveAttr {
-  apply(_user: Pokemon, _target: Pokemon, _move: Move, _args: any[]): boolean {
-    return false;
-  }
+/**
+ * Attribute for moves which have a custom type chart interaction.
+ */
+export abstract class VariableMoveTypeChartAttr extends MoveAttr {
```
DayKev: I think `MoveTypeChartOverrideAttr` is a better name.

src/data/moves/move.ts
```diff
@@ -5999,82 +6002,127 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
...
+  /**
+   * Apply the attribute to change the move's type effectiveness multiplier.
+   * @param user - The {@linkcode Pokemon} using the move
+   * @param target - The {@linkcode Pokemon} targeted by the move
+   * @param move - The {@linkcode Move} with this attribute
+   * @param args -
+   * - `[0]`: A {@linkcode NumberHolder} holding the current type effectiveness
+   * - `[1]`: The target's entire defensive type profile
```
```suggestion
   * - `[1]`: The target's types
```

src/data/moves/move.ts
```diff
@@ -5999,82 +6002,127 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
...
+   * - `[2]`: The current {@linkcode PokemonType} of the move
+   * @returns `true` if application of the attribute succeeds
```
```suggestion
   * @returns Whether application of the attribute succeeds
```

src/data/moves/move.ts
```diff
@@ -5999,82 +6002,127 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
...
+export class HitsSameTypeAttr extends VariableMoveTypeChartAttr {
+  public override apply(
+    user: Pokemon,
+    _target: Pokemon,
+    _move: Move,
+    args: [multiplier: NumberHolder, types: PokemonType[], moveType: PokemonType],
+  ): boolean {
+    const [multiplier, oppTypes] = args;
+    const userTypes = user.getTypes(true);
+    const sharesType = userTypes.every(type => !oppTypes.includes(type));
```
DayKev: Is there no array/set intersection function you can use?

src/data/moves/move.ts
```diff
@@ -5999,82 +6002,127 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
...
+
+    // Synchronoise is never effective if the user is typeless
+    if (sharesType || userTypes.includes(PokemonType.UNKNOWN)) {
```
DayKev: `sharesType` isn't properly named if it being `true` makes the multiplier 0, since Synchronoise is supposed to fail if the target *doesn't* share a type with the user.

src/data/moves/move.ts
```diff
@@ -5999,82 +6002,127 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
...
+  apply(
+    user: Pokemon,
+    target: Pokemon,
+    _move: Move,
+    args: [multiplier: NumberHolder, types: PokemonType[], moveType: PokemonType],
+  ): boolean {
+    const [multiplier] = args;
+    // Intentionally exclude `move` to not re-trigger the effects of this attribute again
+    // (thus leading to an infinite loop)
+    // TODO: Do we need to pass `useIllusion` here?
```
DayKev: `useIllusion` would only be passed if `simulated` is `true`. This attr doesn't support `simulated` so `useIllusion` should not be passed.

src/data/moves/move.ts
```diff
@@ -8705,25 +8753,15 @@ const attackedByItemMessageFunc = (_user: Pokemon, target: Pokemon, _move: Move)
...
-      return true;
-    }
-    return false;
-  }
-}
-
 /**
  * Attribute used for Conversion 2, to convert the user's type to a random type that resists the target's last used move.
- * Fails if the user already has ALL types that resist the target's last used move.
+ * ~~Fails~~ Does nothing if the user already has ALL types that resist the target's last used move.
```
```suggestion
```

src/data/moves/move.ts
```diff
@@ -8705,25 +8753,15 @@ const attackedByItemMessageFunc = (_user: Pokemon, target: Pokemon, _move: Move)
...
  * Fails if the opponent has not used a move yet
- * Fails if the type is unknown or stellar
+ * ~~Fails~~ Does nothing if the type is unknown or stellar
```
```suggestion
```

src/data/moves/move.ts
```diff
@@ -8705,25 +8753,15 @@ const attackedByItemMessageFunc = (_user: Pokemon, target: Pokemon, _move: Move)
...
-    }
-    return false;
-  }
-}
-
 /**
  * Attribute used for Conversion 2, to convert the user's type to a random type that resists the target's last used move.
- * Fails if the user already has ALL types that resist the target's last used move.
+ * ~~Fails~~ Does nothing if the user already has ALL types that resist the target's last used move.
  * Fails if the opponent has not used a move yet
```
```suggestion
```

src/data/moves/move.ts
```diff
@@ -8705,25 +8753,15 @@ const attackedByItemMessageFunc = (_user: Pokemon, target: Pokemon, _move: Move)
...
- * Fails if the type is unknown or stellar
+ * ~~Fails~~ Does nothing if the type is unknown or stellar
  *
  * TODO:
  * If a move has its type changed (e.g. {@linkcode MoveId.HIDDEN_POWER}), it will check the new type.
+ * Does not fail when it should
  */
```
```suggestion
 * Fails if:
 * - the user already has ALL types that resist the target's last used move
 * - the opponent has not used a move yet
 * - the move type is unknown or stellar
 */
// TODO: If a move has its type changed (e.g. Hidden Power), it should check the new type.
// TODO: Does not fail when it should
```

src/data/moves/move.ts
```diff
@@ -12200,13 +12236,13 @@ export function initMoves() {
       }),
     new AttackMove(MoveId.RUINATION, PokemonType.DARK, MoveCategory.SPECIAL, -1, 90, 10, -1, 0, 9) //
       .attr(TargetHalfHpDamageAttr),
-    new AttackMove(MoveId.COLLISION_COURSE, PokemonType.FIGHTING, MoveCategory.PHYSICAL, 100, 100, 5, -1, 0, 9) //
-      .attr(MovePowerMultiplierAttr, (user, target, move) =>
-        target.getAttackTypeEffectiveness(move.type, user) >= 2 ? 4 / 3 : 1,
-      ),
+    new AttackMove(MoveId.COLLISION_COURSE, PokemonType.FIGHTING, MoveCategory.PHYSICAL, 100, 100, 5, -1, 0, 9).attr(
+      MovePowerMultiplierAttr,
+      (user, target, move) => (target.getAttackTypeEffectiveness(move.type, { source: user }) >= 2 ? 4 / 3 : 1),
```
```suggestion
    new AttackMove(MoveId.COLLISION_COURSE, PokemonType.FIGHTING, MoveCategory.PHYSICAL, 100, 100, 5, -1, 0, 9) //
      .attr(MovePowerMultiplierAttr, (user, target, move) =>
        target.getAttackTypeEffectiveness(move.type, { source: user }) >= 2 ? 4 / 3 : 1,
      ),
```

src/field/pokemon.ts
```diff
@@ -2535,85 +2538,115 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-   * @returns a multiplier for the type effectiveness
+   * Calculate the type effectiveness multiplier of a Move used **against** this Pokemon.
+   * @param moveType - The {@linkcode PokemonType} of the move being used
+   * @param source - The {@linkcode Pokemon} using the move, used to check the user's Scrappy and Mind's Eye abilities
+   * and the effects of Foresight/Odor Sleuth
+   * @param ignoreStrongWinds - If `true`, ignores the effect of strong winds (used by anticipation, forewarn, stealth rocks);
+   * default `false`
+   * @param simulated - If `true`, will prevent changes to game state during calculations; default `false`
+   * @param move - The {@linkcode Move} whose type effectiveness is being checked. Used for applying {@linkcode VariableMoveTypeChartAttr}
+   * @param useIllusion - Whether to consider this Pokemon's {@linkcode IllusionData | illusion} when determining types; default `false`
```
```suggestion
   * @param source - (Optional) The {@linkcode Pokemon} using the move
   * @param ignoreStrongWinds - (Default `false`) Whether to ignore the effect of the Strong Winds weather
   * @param simulated - (Default `false`) Whether to prevent changes to game state during calculations
   * @param move - (Optional) The {@linkcode Move} whose type effectiveness is being checked.
   * @param useIllusion - (Default `false`) Whether to consider this Pokemon's {@linkcode IllusionData | illusion} when determining types
```

src/field/pokemon.ts
```diff
@@ -2535,85 +2538,115 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-          if (exposedTags.some(t => t.ignoreImmunity(defenderType, moveType)) && multiplier.value === 0) {
-            return 1;
-          }
-        }
-        return multiplier.value;
-      })
-      .reduce((acc, cur) => acc * cur, 1) as TypeDamageMultiplier;
+    // Apply any typing changes from Freeze-Dry, etc.
+    if (move) {
+      applyMoveAttrs("VariableMoveTypeChartAttr", source ?? null, this, move, multi, types, moveType);
```
DayKev: `VariableMoveTypeChartAttr`'s first param should accept `undefined`, `?? null` is dumb.

src/field/pokemon.ts
```diff
@@ -2535,85 +2538,115 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-    return multiplier as TypeDamageMultiplier;
+    return multi.value as TypeDamageMultiplier;
+  }
+
+  /**
+   * Sub-method of {@linkcode getAttackTypeEffectiveness} that handles nullifying type immunities.
+   * @param source - The {@linkcode Pokemon} from whom the attack is sourced
+   * @param simulated - If `true`, will prevent displaying messages upon activation
+   * @param moveType - The {@linkcode PokemonType} whose offensive typing is being checked
+   * @param defenderType - The defender's {@linkcode PokemonType} being checked
```
```suggestion
   * @param simulated - Whether to prevent modifications to game state
   * @param moveType - The {@linkcode PokemonType} of the move being checked
   * @param defenderType - The defending Pokemon's type
```

src/field/pokemon.ts
```diff
@@ -2634,10 +2667,15 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    let defScore = 1 / Math.max(this.getAttackTypeEffectiveness(enemyTypes[0], { source: opponent }), 0.25);
     if (enemyTypes.length > 1) {
       defScore *=
-        1 / Math.max(this.getAttackTypeEffectiveness(enemyTypes[1], opponent, false, false, undefined, true), 0.25);
+        // TODO: Shouldn't this pass `simulated=true` here?
+        1
+        / Math.max(
+          this.getAttackTypeEffectiveness(enemyTypes[1], { source: opponent, simulated: false, useIllusion: true }),
+          0.25,
+        );
```
```suggestion
      // TODO: Shouldn't this pass `simulated=true` here?
      const typeEffectiveness = this.getAttackTypeEffectiveness(enemyTypes[1], {
        source: opponent,
        simulated: false,
        useIllusion: true,
      });
      defScore *= 1 / Math.max(typeEffectiveness, 0.25);
```

src/field/pokemon.ts
```diff
@@ -2634,10 +2667,15 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
      * Based on how effectively this Pokemon defends against the opponent's types.
      * This score cannot be higher than 4.
      */
-    let defScore = 1 / Math.max(this.getAttackTypeEffectiveness(enemyTypes[0], opponent), 0.25);
+    let defScore = 1 / Math.max(this.getAttackTypeEffectiveness(enemyTypes[0], { source: opponent }), 0.25);
     if (enemyTypes.length > 1) {
       defScore *=
```
```suggestion
```

src/field/pokemon.ts
```diff
@@ -4154,25 +4196,43 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-   * @returns The first matching tag, or `undefined` if none match
+   * @returns The first matching tag, or `undefined` if none match.
+   * @remarks
+   * Equivalent to `this.summonData.tags.find(tagFilter)`.
    */
-  public findTag(tagFilter: (tag: BattlerTag) => boolean) {
-    return this.summonData.tags.find(tagFilter);
+  findTag(tagFilter: (tag: BattlerTag) => boolean): BattlerTag | undefined;
+  findTag(tagFilter: (tag: BattlerTag) => boolean) {
+    return this.summonData.tags.find(t => tagFilter(t));
```
```suggestion
    return this.summonData.tags.find(tagFilter);
```

src/field/pokemon.ts
```diff
@@ -4154,25 +4196,43 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-   * @returns The filtered list of tags
+   * @returns The filtered list of tags.
+   * @remarks
+   * Equivalent to `this.summonData.tags.filter(tagFilter)`.
    */
-  public findTags(tagFilter: (tag: BattlerTag) => boolean): BattlerTag[] {
-    return this.summonData.tags.filter(tagFilter);
+  findTags(tagFilter: (tag: BattlerTag) => boolean): BattlerTag[];
+  findTags(tagFilter: (tag: BattlerTag) => boolean): BattlerTag[] {
+    return this.summonData.tags.filter(t => tagFilter(t));
```
```suggestion
    return this.summonData.tags.filter(tagFilter);
```

src/field/pokemon.ts
```diff
@@ -4154,25 +4196,43 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * Return all `BattlerTag`s satisfying the given predicate.
+   * @param tagFilter - The predicate to match against
+   * @returns The filtered list of tags.
    * @remarks
    * Equivalent to `this.summonData.tags.filter(tagFilter)`.
+   */
+  findTags<T extends BattlerTag>(tagFilter: (tag: BattlerTag) => tag is T): T[];
+  /**
+   * Return all `BattlerTag`s satisfying the given predicate.
    * @param tagFilter - The predicate to match against
```
```suggestion
```
If the TSDocs are identical, you don't need to repeat them on the overloads.
<p>

image
image
image

</p>
</details>

src/field/pokemon.ts
```diff
@@ -4154,25 +4196,43 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-   * @returns The filtered list of tags
+   * @returns The filtered list of tags.
+   * @remarks
+   * Equivalent to `this.summonData.tags.filter(tagFilter)`.
    */
```
```suggestion
```

src/field/pokemon.ts
```diff
@@ -4154,25 +4196,43 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * Find the first `BattlerTag` matching the specified predicate.
+   * @param tagFilter - The predicate to match against
+   * @returns The first matching tag, or `undefined` if none match.
    * @remarks
    * Equivalent to `this.summonData.tags.find(tagFilter)`.
+   */
+  findTag<T extends BattlerTag>(tagFilter: (tag: BattlerTag) => tag is T): T | undefined;
+  /**
+   * Find the first `BattlerTag` matching the specified predicate.
    * @param tagFilter - The predicate to match against
```
```suggestion
```

src/field/pokemon.ts
```diff
@@ -4154,25 +4196,43 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-   * @returns The first matching tag, or `undefined` if none match
+   * @returns The first matching tag, or `undefined` if none match.
+   * @remarks
+   * Equivalent to `this.summonData.tags.find(tagFilter)`.
    */
```
```suggestion
```

test/moves/freeze-dry.test.ts
```diff
@@ -1,326 +1,136 @@
...
 import { SpeciesId } from "#enums/species-id";
+import type { EnemyPokemon, PlayerPokemon } from "#field/pokemon";
 import { GameManager } from "#test/test-utils/game-manager";
+import { stringifyEnumArray } from "#test/test-utils/string-utils";
 import Phaser from "phaser";
-import { beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeAll, describe, expect, it } from "vitest";
 
-describe("Moves - Freeze-Dry", () => {
+type typesArray = [PokemonType] | [PokemonType, PokemonType] | [PokemonType, PokemonType, PokemonType];
```
DayKev:
```suggestion
type TypesArray = [PokemonType] | [PokemonType, PokemonType] | [PokemonType, PokemonType, PokemonType];
```

test/moves/freeze-dry.test.ts
```diff
@@ -1,326 +1,136 @@
...
+
+    const moveType = feebas.getMoveType(allMoves[MoveId.FREEZE_DRY]);
+    const eff = enemy.getAttackTypeEffectiveness(moveType, { source: feebas, move: allMoves[MoveId.FREEZE_DRY] });
+    expect(
+      eff,
+      `Freeze-dry effectiveness against ${stringifyEnumArray(PokemonType, types)} was ${eff} instead of ${multi}!`,
+    ).toBe(multi);
+  }
+
+  describe("Normal -", () => {
```
DayKev:
```suggestion
  describe("Normal", () => {
```

test/moves/freeze-dry.test.ts
```diff
@@ -1,326 +1,136 @@
...
-    expect(enemy.getMoveEffectiveness).toHaveLastReturnedWith(2);
+      expectEffectiveness([PokemonType.WATER], 2);
+    });
   });
 
-  it("should deal 1x damage to water/flying type during inverse battle under Electrify", async () => {
-    game.override.enemyMoveset([MoveId.ELECTRIFY]).enemySpecies(SpeciesId.GYARADOS);
-
-    game.challengeMode.addChallenge(Challenges.INVERSE_BATTLE, 1, 1);
+  describe("Inverse Battle -", () => {
```
DayKev:
```suggestion
  describe("Inverse Battle", () => {
```

test/moves/synchronoise.test.ts
```diff
@@ -28,16 +29,79 @@ describe("Moves - Synchronoise", () => {
       .enemyMoveset(MoveId.SPLASH);
   });
 
-  it("should consider the user's tera type if it is terastallized", async () => {
+  it("should affect all opponents that share a type with the user", async () => {
+    game.override.battleStyle("double");
+    await game.classicMode.startBattle([SpeciesId.BIBAREL, SpeciesId.STARLY]);
+
+    const [bidoof, starly] = game.scene.getPlayerField();
```
DayKev: Last I checked, Bidoof != Bibarel

test/moves/synchronoise.test.ts
```diff
@@ -28,16 +29,79 @@ describe("Moves - Synchronoise", () => {
...
+    game.field.forceTera(abra, PokemonType.STELLAR);
+    game.field.forceTera(karp, PokemonType.STELLAR);
+    game.move.use(MoveId.SYNCHRONOISE);
+    await game.toEndOfTurn();
+
+    expect(abra).toHaveUsedMove({ move: MoveId.SYNCHRONOISE, result: MoveResult.MISS });
+    expect(karp).toHaveFullHp();
+  });
+
+  it("should count as ineffective if no enemies share types with the user", async () => {
```
DayKev, line 79:
```suggestion
  it("should count as a miss if no enemies share types with the user", async () => {
```

test/test-utils/helpers/challenge-mode-helper.ts
```diff
@@ -56,10 +71,22 @@ export class ChallengeModeHelper extends GameManagerHelper {
...
    * @returns A promise that resolves when the battle is started.
+   * @todo This duplicates all its code with the classic mode variant...
    */
+  async startBattle(species: SpeciesId[]): Promise<void>;
+  /**
+   * Transitions the challenge game to the start of a new battle.
+   * Selects 3 daily run starters with a fixed seed of "test"
+   * (see `DailyRunConfig.getDailyRunStarters` in `daily-run.ts` for more info).
+   * @returns A promise that resolves when the battle is started.
+   * @deprecated - Specifying the starters helps prevent inconsistencies from internal RNG changes.
```
DayKev:
```suggestion
   * @deprecated Specifying the starters helps prevent inconsistencies from internal RNG changes.
```

test/test-utils/helpers/challenge-mode-helper.ts
```diff
@@ -30,11 +32,24 @@ export class ChallengeModeHelper extends GameManagerHelper {
...
+   * @returns A promise that resolves when the summon phase is reached.
+   * @todo This duplicates all but 1 line of code from the classic mode variant...
+   */
+  async runToSummon(speciesIds: SpeciesId[]): Promise<void>;
+  /**
+   * Runs the challenge game to the summon phase.
+   * Selects 3 daily run starters with a fixed seed of "test"
+   * (see `DailyRunConfig.getDailyRunStarters` in `daily-run.ts` for more info).
    * @returns A promise that resolves when the summon phase is reached.
+   * @deprecated - Specifying the starters helps prevent inconsistencies from internal RNG changes.
```
DayKev:
```suggestion
   * @deprecated Specifying the starters helps prevent inconsistencies from internal RNG changes.
```

test/test-utils/helpers/field-helper.ts
```diff
@@ -111,12 +112,14 @@ export class FieldHelper extends GameManagerHelper {
...
-   * Force a given Pokemon to be terastallized to the given type.
+   * Force a given Pokemon to be Terastallized to the given type.
    *
-   * @param pokemon - The pokemon to terastallize.
-   * @param teraType - The {@linkcode PokemonType} to terastallize into; defaults to the pokemon's primary type.
+   * @param pokemon - The pokemon to Terastallize
+   * @param teraType - The {@linkcode PokemonType} to Terastallize into; defaults to `pokemon`'s primary type if not provided
    * @remarks
    * This function only mocks the Pokemon's tera-related variables; it does NOT activate any tera-related abilities.
+   * If activating on-Terastallize effects is desired, use either {@linkcode MoveHelper.use} with `useTera=true`,
```
DayKev:
```suggestion
   *
   * If activating on-Terastallize effects is desired, use either {@linkcode MoveHelper.use} with `useTera=true`,
```

test/moves/flying-press.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      const moveType = hawlucha.getMoveType(allMoves[MoveId.FLYING_PRESS]);
+      const flyingPressEff = enemy.getAttackTypeEffectiveness(moveType, {
+        source: hawlucha,
+        move: allMoves[MoveId.FLYING_PRESS],
+      });
+      expect(flyingPressEff).toBe(8);
+    });
+  });
+
+  describe("Inverse Battle -", () => {
```
DayKev:
```suggestion
  describe("Inverse Battle", () => {
```

test/moves/flying-press.test.ts
```diff
@@ -0,0 +1,127 @@
...
+          flyingPressEff,
+          `Flying Press effectiveness against ${toTitleCase(PokemonType[type])} was incorrect!`
+            + `\nExpected: ${flyingPressEff},`
+            + `\nActual: ${primaryEff * flyingEff} (=${primaryEff} * ${flyingEff})`,
+        )
+        .toBe(primaryEff * flyingEff);
+    }
+  }
+
+  describe("Normal -", () => {
```
DayKev:
```suggestion
  describe("Normal", () => {
```

src/data/moves/move.ts
Bertie690: there is for sets but it's significantly slower than simply using arrays (IIRC from benchmarking).

If some other property is desired, it makes sense to use sets, but here it's just unnecessarily slower.

src/data/moves/move.ts
Bertie690: The reason it doesn't check `simulated` is because _all the attribute does is set a number holder_.
What I was referring to was using the `useIllusion` parameter for the Flying-type portion of the multiplier (as that would use the target's _ACTUAL_ types instead of their illusion's).

src/field/pokemon.ts
Bertie690: This is a limitation of the move attribute system, not the move attribute.
I agree with your view 100%, but this is something we'd have to fix on another day.

src/field/pokemon.ts
Bertie690: image

doesn't work for me 🤷

test/moves/synchronoise.test.ts
Bertie690: I do not want to promote the fact that type immunities count as a "miss" internally, nor imply that they should count as misses.

The fact that `HitResult.MISS` is used for both accuracy and type immmunities is extremely nonsensical and should be revisited later on once the move failure system attains more granularity.

src/data/moves/move.ts
DayKev: > The reason it doesn't check `simulated` is because _all the attribute does is set a number holder_.

That's still a side effect, though they're often used as pseudo-return values... ugh, it would be nice if `simulated = true` meant a function would be "pure".

src/data/moves/move.ts
```diff
@@ -6068,43 +6071,47 @@ export class NeutralDamageAgainstFlyingTypeAttr extends VariableMoveTypeChartAtt
...
     const userTypes = user.getTypes(true);
-    const sharesType = userTypes.every(type => !oppTypes.includes(type));
-
     // Synchronoise is never effective if the user is typeless
-    if (sharesType || userTypes.includes(PokemonType.UNKNOWN)) {
-      multiplier.value = 0;
-      return true;
+    if (userTypes.some(type => type === PokemonType.UNKNOWN || !oppTypes.includes(type))) {
+      return false;
     }
```
DayKev: This would return `false` if the user is Flying + Fire and the opponent is Normal type, because when the user's first type is checked (Flying), it will see that Flying is not included in the opponents type (which is Normal). Which means the multiplier is not set to `0` and the move would be effective even though the Pokémon don't share any types.

image

It would also make the multiplier not set to `0` if the user is typeless.

image

src/data/moves/move.ts
```diff
@@ -8705,26 +8760,11 @@ const attackedByItemMessageFunc = (_user: Pokemon, target: Pokemon, _move: Move)
...
 /**
  * Attribute used for Conversion 2, to convert the user's type to a random type that resists the target's last used move.
- * Fails if the user already has ALL types that resist the target's last used move.
- * Fails if the opponent has not used a move yet
- * Fails if the type is unknown or stellar
- *
- * TODO:
- * If a move has its type changed (e.g. {@linkcode MoveId.HIDDEN_POWER}), it will check the new type.
  */
+// TODO: If a move has its type changed (e.g. {@linkcode MoveId.HIDDEN_POWER}), it will check the new type.
```
```suggestion
// TODO: If a move has its type changed (e.g. {@linkcode MoveId.HIDDEN_POWER}), it should check the new type.
```

src/data/type.ts
```diff
@@ -312,25 +314,23 @@ export function getTypeDamageMultiplierColor(
         return "#52C200";
     }
   }
-  if (side === "defense") {
```
DayKev, line 298: Tbh the `side` param is completely unnecessary, this function is called exactly once and only for the `"offense"` colors. These colors will never be used.

src/field/pokemon.ts
DayKev:
```suggestion
   * @param simulated - Whether to prevent displaying messages upon activation
   * @param moveType - The {@linkcode PokemonType | attacking type} being checked
   * @param defenderType - The defending type being checked
```
The convention is "Whether to x" and not "If `true`, x".

Arguably `moveType` should be renamed to `attackingType` if you want to keep this wording (also `defenderType` renamed to `defendingType`).

Also messages being displayed to the user is indeed game state being changed, and wording should be consistent across the codebase. If you want to keep this wording then the `simulated` param should be renamed to `quiet`.

src/data/type.ts
Bertie690: Later pr issue

src/data/moves/move.ts
Bertie690: I will die on this hill - setting a `NumberHolder` is NOT a meaningful side effect (à la `out` parameters in C#) - the state mutations are contained to the calling code.

If we did consider setting such holders as impure, you'd effectively be making `simulated=true` **disable all ability effects** that are not themselves placeholder stubs, which I heavily disagree with.

src/data/moves/move.ts
```diff
@@ -8708,26 +8764,11 @@ const attackedByItemMessageFunc = (_user: Pokemon, target: Pokemon, _move: Move)
...
 /**
  * Attribute used for Conversion 2, to convert the user's type to a random type that resists the target's last used move.
- * Fails if the user already has ALL types that resist the target's last used move.
- * Fails if the opponent has not used a move yet
- * Fails if the type is unknown or stellar
- *
- * TODO:
- * If a move has its type changed (e.g. {@linkcode MoveId.HIDDEN_POWER}), it will check the new type.
  */
+// TODO: If a move has its type changed (e.g. {@linkcode MoveId.HIDDEN_POWER}), it should check the new type.
```
```suggestion
// TODO: If a move has its type changed (e.g. Hidden Power), it should check the new type.
```
Whoops, missed this the first time around.

src/field/pokemon.ts
```diff
@@ -2542,85 +2545,109 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-   * @param moveType {@linkcode PokemonType} the type of the move being used
-   * @param source {@linkcode Pokemon} the Pokemon using the move
-   * @param ignoreStrongWinds whether or not this ignores strong winds (anticipation, forewarn, stealth rocks)
-   * @param simulated tag to only apply the strong winds effect message when the move is used
-   * @param move (optional) the move whose type effectiveness is to be checked. Used for applying {@linkcode VariableMoveTypeChartAttr}
-   * @param useIllusion - Whether we want the attack type effectiveness on the illusion or not
-   * @returns a multiplier for the type effectiveness
+   * Calculate the type effectiveness multiplier of a Move used **against** this Pokemon.
+   * @param moveType - The {@linkcode PokemonType} of the move being used
+   * @param params - Parameters used to modify the type effectiveness
```
DayKev, line 2550: This needs to be something like `__namedparameters.source` for Typedoc doesn't it?

src/field/pokemon.ts
```diff
@@ -2542,85 +2545,109 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    moveType,
+    defenderType,
+  }: {
+    source: Pokemon;
+    simulated: boolean;
+    moveType: PokemonType;
+    defenderType: PokemonType;
+  }): boolean {
+    // TODO: remove type assertion once method is properly typed
+    const hasExposed = !!this.findTag(tag => tag instanceof ExposedTag && tag.ignoreImmunity(defenderType, moveType));
```
```suggestion
    const hasExposed = !!this.findTag(
      tag =>
        [BattlerTagType.IGNORE_DARK, BattlerTagType.IGNORE_GHOST].includes(tag.tagType)
        && (tag as ExposedTag).ignoreImmunity(defenderType, moveType),
    );
```
This avoids an `instanceof`. Unfortunately there's still one more `instanceof ExposedTag` that needs to be removed later so the import can be changed to a type-only import.

src/field/pokemon.ts
```diff
@@ -3624,7 +3661,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * @param __namedParameters.source - Needed for proper typedoc rendering
    * @returns The {@linkcode DamageCalculationResult}
    */
-  // TODO: Condense various multipliers into a single function
+  // TODO: Condense various multipliers into a single function for easier unit testing
```
```suggestion
  // TODO: Condense various multipliers into a separate function for easier unit testing
```
Presumably you mean "separate" and not "single" since this is already a single function and more functions = easier unit testing.

src/field/pokemon.ts
Bertie690: Idr that being a thing
@SirzBenjie what do you think

DayKev: You were the one doing it lmao

DayKev: image

Bertie690: I forgor

src/ai/rival-team-gen.ts
```diff
@@ -138,6 +134,9 @@ function calcPartyTypings(
   referenceConfig: RivalPoolConfig = RIVAL_6_POOL,
 ): void {
   if (!balanceTypes && !balanceWeaknesses) {
+    console.warn(
+      "At least one of the `balanceTypes` and/or `balanceWeaknesses` params of `calcPartyTypings` should be passed `true`, but neither was!",
+    );
```
SirzBenjie, line 139:
```suggestion
    console.warn(
      "At least one of the `balanceTypes` and/or `balanceWeaknesses` params of `calcPartyTypings` should be passed `true`, but neither was!",
    );
```

Maybe this should use [`console.assert`](https://developer.mozilla.org/en-US/docs/Web/API/console/assert_static) with the condition?

src/data/abilities/init-abilities.ts
```diff
@@ -2075,66 +2075,37 @@ function getTerrainCondition(...terrainTypes: TerrainType[]): AbAttrCondition {
...
-            PokemonType.GRASS,
-            PokemonType.ELECTRIC,
-            PokemonType.PSYCHIC,
-            PokemonType.ICE,
-            PokemonType.DRAGON,
-            PokemonType.DARK,
-          ][iv_val];
+/**
+ * Condition used by {@linkcode AbilityId.ANTICIPATION} to show a message if any opponent knows a
+ * "dangerous" move.
```
SirzBenjie:
```suggestion
 * Condition used by {@link https://bulbapedia.bulbagarden.net/wiki/Anticipation_(Ability) | Anticipation}
 * to show a message if any opponent knows a "dangerous" move.
```
I've been realizing that linking to the enum member is wholly useless. There's no reason to click it. I'm slowly working on phasing out linkcodes to `MoveId`s.

src/data/moves/move.ts
```diff
@@ -6002,82 +6005,135 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
...
-      const multiplier = args[0] as NumberHolder;
-      //When a flying type is hit, the first hit is always 1x multiplier.
-      if (target.isOfType(PokemonType.FLYING)) {
-        multiplier.value = 1;
-      }
-      return true;
+export type { MoveTypeChartOverrideAttr };
+
+/**
+ * Attribute to implement {@linkcode MoveId.FREEZE_DRY}'s guaranteed Water-type super effectiveness.
```
SirzBenjie:
```suggestion
 * Attribute to implement {@link https://bulbapedia.bulbagarden.net/wiki/Freeze-Dry_(move) | Freeze Dry}'s 
 * guaranteed Water-type super effectiveness.
```
See above.

src/data/moves/move.ts
```diff
@@ -6002,82 +6005,135 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
...
-   * @param args `[0]` a {@linkcode NumberHolder | NumberHolder} containing a type effectiveness multiplier
-   * @returns `true` if this Ice-type immunity applies; `false` otherwise
-   */
-  apply(_user: Pokemon, target: Pokemon, _move: Move, args: any[]): boolean {
-    const multiplier = args[0] as NumberHolder;
-    if (target.isOfType(PokemonType.ICE)) {
-      multiplier.value = 0;
-      return true;
+/**
+ * Attribute used by {@linkcode MoveId.THOUSAND_ARROWS} to cause it to deal a fixed 1x damage
```
SirzBenjie:
```suggestion
 * Attribute used by {@link https://bulbapedia.bulbagarden.net/wiki/Thousand_Arrows_(move) | Thousand Arrows}
 * to cause it to deal a fixed 1x damage
```
See above

src/data/moves/move.ts
```diff
@@ -6002,82 +6005,135 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
...
+    return true;
   }
 }
 
-export class FlyingTypeMultiplierAttr extends VariableMoveTypeMultiplierAttr {
-  apply(user: Pokemon, target: Pokemon, _move: Move, args: any[]): boolean {
-    const multiplier = args[0] as NumberHolder;
-    multiplier.value *= target.getAttackTypeEffectiveness(PokemonType.FLYING, user);
+/**
+ * Move attribute used by {@linkcode MoveId.SYNCHRONOISE} to render the move ineffective
```
SirzBenjie:
```suggestion
 * Attribute implementing the effect of @link https://bulbapedia.bulbagarden.net/wiki/Synchronoise_(move) | Synchronoise};
 * rendering the move ineffective if the target does not share a type with the user.
```

src/data/moves/move.ts
```diff
@@ -6002,82 +6005,135 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
...
+    // (thus leading to an infinite loop)
+    // TODO: We may need to propagate `useIllusion` here for correct AI interactions
+    multiplier.value *= target.getAttackTypeEffectiveness(PokemonType.FLYING, { source: user });
+    return true;
   }
 }
 
 /**
- * This class forces Freeze-Dry to be super effective against Water Type.
+ * Attribute used by {@linkcode MoveId.SHEER_COLD} to implement its Gen VII+ ice ineffectiveness.
```
SirzBenjie:
```suggestion
 * Attribute used by {@link https://bulbapedia.bulbagarden.net/wiki/Sheer_Cold_(move) | Sheer Cold}
 * to implement its Gen VII+ ice ineffectiveness.
```

src/data/moves/move.ts
```diff
@@ -8708,26 +8764,11 @@ const attackedByItemMessageFunc = (_user: Pokemon, target: Pokemon, _move: Move)
...
  * Attribute used for Conversion 2, to convert the user's type to a random type that resists the target's last used move.
- * Fails if the user already has ALL types that resist the target's last used move.
- * Fails if the opponent has not used a move yet
- * Fails if the type is unknown or stellar
- *
- * TODO:
- * If a move has its type changed (e.g. {@linkcode MoveId.HIDDEN_POWER}), it will check the new type.
  */
+// TODO: If a move has its type changed (e.g. Hidden Power), it should check the new type.
+// TODO: Does not fail when it should
```
SirzBenjie, line 8774:
```suggestion
// TODO: If a move has its type changed (e.g. Hidden Power), it should check the new type.
// TODO: Does not fail when it should
```

Make a TODO test for this?

src/data/type.ts
```diff
@@ -263,13 +284,11 @@ export function getTypeDamageMultiplier(attackType: PokemonType, defType: Pokemo
...
       return 1;
   }
-
-  return 1;
 }
 
 /**
  * Retrieve the color corresponding to a specific damage multiplier
- * @returns A color or undefined if the default color should be used
+ * @returns A color or `undefined` if the default color should be used
```
SirzBenjie:
```suggestion
 * @returns The color, or `undefined` if the default color should be used
```

src/data/type.ts
```diff
@@ -295,25 +314,23 @@ export function getTypeDamageMultiplierColor(
...
+    case 0.5:
+      return "#0093FF";
+    case 1:
+      return;
+    case 2:
+      return "#FE8E00";
+    case 4:
+      return "#FF7400";
+    case 8:
+      return "#FF5500";
```
SirzBenjie, line 335:
```suggestion
      return "#B1B100";
    case 0.125:
      return "#2DB4FF";
    case 0.25:
      return "#00A4FF";
    case 0.5:
      return "#0093FF";
    case 1:
      return;
    case 2:
      return "#FE8E00";
    case 4:
      return "#FF7400";
    case 8:
      return "#FF5500";
```

Magic strings! These really should exist in constants somewhere. Not this PR, but since you're editing the lines anyway, can you slap down a TODO?

src/enums/pokemon-type.ts
```diff
@@ -1,4 +1,5 @@
 export enum PokemonType {
+  /** Typeless */
```
SirzBenjie:
```suggestion
  /** AKA Typeless */
```

src/ai/rival-team-gen.ts
Bertie690: you didn't change the code...?

src/data/moves/move.ts
Bertie690: making new suggestion with fixed range

src/data/moves/move.ts
```diff
@@ -6002,82 +6005,135 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
...
   }
 }
 
-export class FlyingTypeMultiplierAttr extends VariableMoveTypeMultiplierAttr {
-  apply(user: Pokemon, target: Pokemon, _move: Move, args: any[]): boolean {
-    const multiplier = args[0] as NumberHolder;
-    multiplier.value *= target.getAttackTypeEffectiveness(PokemonType.FLYING, user);
+/**
+ * Move attribute used by {@linkcode MoveId.SYNCHRONOISE} to render the move ineffective
+ * against all targets that do not share a type with the user.
```
```suggestion
 * Attribute implementing the effect of {@link https://bulbapedia.bulbagarden.net/wiki/Synchronoise_(move) | Synchronoise},
 * rendering the move ineffective against all targets that do not share at least 1 type with the user.
```

src/data/moves/move.ts
```diff
@@ -6002,82 +6005,135 @@ export class CombinedPledgeTypeAttr extends VariableMoveTypeAttr {
...
+    }
+
+    multiplier.value = 0;
     return true;
   }
 }
 
 /**
- * Attribute for moves which have a custom type chart interaction.
+ * Attribute used by {@linkcode MoveId.FLYING_PRESS} to add the Flying-type to its type effectiveness.
```
```suggestion
 * Attribute used by {@link https://bulbapedia.bulbagarden.net/wiki/Flying_Press_(move) | Flying Press} 
 * to add the Flying-type to its type effectiveness multiplier.
```

src/enums/pokemon-type.ts
```diff
@@ -1,6 +1,12 @@
+/**
+ * A `PokemonType` represents the type of a Pokemon or its moves.
+ *
+ * @see {@link https://bulbapedia.bulbagarden.net/wiki/Type | Types on Bulbapedia}
+ */
 export enum PokemonType {
-  /** Typeless */
+  /** Also known as typelessness. */
```
DayKev:
```suggestion
  /** Also known as {@link https://bulbapedia.bulbagarden.net/wiki/Type#Typeless | typeless}. */
```

PR #6254, opened 2025-08-11
[Refactor] Remove bgm param from arena constructor

fabske0:

N/A

The arena constructor is only called once and the `bgm` property can easily be set from the biome parameter

Removed `bmg` parameter from `Arena`

## Screenshots/Videos

N/A

## Checklist
~~- [ ] Have I provided screenshots/videos of the changes (if applicable)?~~
  ~~- [ ] Have I made sure that any UI change works for both UI themes (default and legacy)?~~


src/field/arena.ts
```diff
@@ -68,9 +68,9 @@ export class Arena {
 
   public readonly eventTarget: EventTarget = new EventTarget();
 
-  constructor(biome: BiomeId, bgm: string, playerFaints = 0) {
+  constructor(biome: BiomeId, playerFaints = 0) {
     this.biomeType = biome;
-    this.bgm = bgm;
+    this.bgm = BiomeId[biome].toLowerCase();
```
DayKev, line 73:
```suggestion
```
I'd also move this to the definition of the `bgm` field. `public bgm = ...`

fabske0: That wouldn't work tho no? since the biome is only passed in the constructor

DayKev: Oh it uses a value from the constructor. Nevermind then lol.

src/battle-scene.ts
```diff
@@ -1514,7 +1514,7 @@ export class BattleScene extends SceneBase {
   }
 
   newArena(biome: BiomeId, playerFaints?: number): Arena {
```
Bertie690: While you're at it can you make it `playerFaints = 0`?

src/field/arena.ts
```diff
@@ -68,9 +68,9 @@ export class Arena {
...
     this.trainerPool = biomeTrainerPools[biome];
     this.updatePoolsForTimeOfDay();
     this.playerTerasUsed = 0;
```
Bertie690: Can you move this to the field definition?
Thx

Bertie690 (approved):
Minor nits but lgtm

PR #6258, opened 2025-08-13
[Misc] Moved `scrapeTrainerNames` to a script; removed `selfStatLowerMoves`

Bertie690:
[Locales PR](<https://github.com/pagefaultgames/pokerogue-locales/pull/223>)

N/A
Trainer name scraper should not be in prod.

Also, while I was converting the script and cleaning it up, Fabi asked me to make the `trainers-common.json` keys into `camelCase` while I was at it. (I also removed the prefixes since we are literally just picking randomly from a big list loool)

Reworked the script to use JSDOM for parsing due to Node.JS' lack of native DOM support.
Rather than requiring an explicit list of "gender counterpart names" to scrape, the script takes a trainer class as input and scrapes all names for said class, recursing into any gender counterparts listed on the page. The final names array is tallied up into a large Set and subsequently converted into JSON suitable for copy-pasting into the file.

Tweaked the format of the keys to not be abhorrent

Removed `selfStatLowerMoves` because nothing uses it without white herb (and white herb can likely be made to use something other than a giant array) 

## Screenshots/Videos
N/A
## How to test the changes?
n/a
## Checklist


src/data/trainer-names.ts
DayKev, line 1: This whole file is just for the trainer name script? Then you can just leave it alone and add a comment to the top stating its purpose and that it can't be easily moved to a script because Node doesn't have native DOM parsing.

src/data/moves/move.ts
```diff
@@ -11542,10 +11540,5 @@ export function initMoves() {
       .condition(new UpperHandCondition()),
     new AttackMove(MoveId.MALIGNANT_CHAIN, PokemonType.POISON, MoveCategory.SPECIAL, 100, 100, 5, 50, 0, 9)
       .attr(StatusEffectAttr, StatusEffect.TOXIC)
-  );
-  allMoves.map(m => {
-    if (m.getAttrs("StatStageChangeAttr").some(a => a.selfTarget && a.stages < 0)) {
-      selfStatLowerMoves.push(m.id);
-    }
-  });
+  )
```
```suggestion
  );
```

src/field/trainer.ts
```diff
@@ -157,9 +153,9 @@ export class Trainer extends Phaser.GameObjects.Container {
 
   /**
    * Returns the name of the trainer based on the provided trainer slot and the option to include a title.
-   * @param {TrainerSlot} trainerSlot - The slot to determine which name to use. Defaults to TrainerSlot.NONE.
-   * @param {boolean} includeTitle - Whether to include the title in the returned name. Defaults to false.
-   * @returns {string} - The formatted name of the trainer.
+   * @param rainerSlot - The slot to determine which name to use; default `TrainerSlot.NONE`
```
SirzBenjie:
```suggestion
   * @param trainerSlot - The slot to determine which name to use; default `TrainerSlot.NONE`
```

src/data/abilities/ability.ts
```diff
@@ -6765,7 +6765,7 @@ function getPokemonWithWeatherBasedForms() {
     );
 }
 
-// biome-ignore format: prevent biome from removing the newlines (e.g. prevent `new Ability(...).attr(...)`)
+// biome-ignore-start format: prevent biome from removing the newlines (e.g. prevent `new Ability(...).attr(...)`)
```
DayKev: Why did you change this to a range suppression comment?

typedoc.json
```diff
@@ -1,7 +1,7 @@
 {
   "entryPoints": ["./src"],
   "entryPointStrategy": "expand",
-  "exclude": ["**/*+.test.ts"],
+  "exclude": ["**/*+.test.ts", "**/src/data/trainer-names.ts"],
```
DayKev: Why did you add a deleted file to the exclude list?

src/data/abilities/ability.ts
Bertie690: Mainly to see if it still worked

typedoc.json
Bertie690: leftover from when i forgot that JSDOM was a thing

src/data/abilities/ability.ts
DayKev: Range suppression comments for formatting are silently converted to non-range suppressions, they don't work. You need to change it back.

SirzBenjie (dismissed):

xsn34kzx (dismissed):

src/utils/i18n.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Select a random i18n key from all nested keys in the given object.
+ * @param key - The i18n key to retrieve a random value of.
+ * The key's value should be an object containing numerical keys (starting from 1).
+ * @returns A typle containing the key and value pair.
+ * @privateRemarks
+ * The reason such "array-like" keys are not stored as actual arrays is due to the
+ * translation software used by the Translation Team (Mozilla Pontoon)
+ * not supporting arrays in any capacity.
+ */
+export function getRandomLocaleKey(key: string): [key: string, value: string] {
```
DayKev: The name of this function isn't entirely accurate, because it returns both the key and the translation for that key (which... why?).

DayKev (approved):

src/utils/i18n.ts
SirzBenjie: Good point, we can probably rename to `getRandomLocaleEntry`

PR #6276, opened 2025-08-16
[Bug] [Refactor] Add failure conditions and move failures part 1

### LINKED ISSUE #4289: [Bug] Gigaton Hammer and Blood Moon can be Selected and fails, when Chosen Successively.

SirzBenjie:
I originally wanted to address enemy AI using moves that will fail in primal weather. I then realized that the moves are checking this `isUsable` method, which is also what we want to hook into to prevent moves from being selected.
I also found out that the AI can sort of cheat and select moves it isn't supposed to be able to under certain conditions (even conditions where the player is prevented from selecting them). I found it was the perfect opportunity to fix #4289 and similar issues.

However, I _then_ found out that the same `isUsable` method was being used to check for move failures in the move phase and move effect phase. This... is wildly incorrect and causes mismatch with cartridge behavior.
To fix this, I decided to finally fix our move phase's failure sequence to align with mainline behavior.

This is a WIP as I writeup and implement everything.

Changes to `move-phase.ts`
- Refactored the failure checks in move-phase to happen as it does on cartridge
- Made it so that moves aren't selectable

Changes to `battler-tags.ts`
- Removed the `PRE_MOVE` lapse type from tags related to failures, in their constructor.

Changes to `pokemon.ts`
- Added a parameter to `lapseTag` that allows specifying the lapse type to use. This is now the preferred method to manually trigger the effects of a tag, as it will handle lapsing it, etc.

New file: `move-condition.ts`
- Moved several move conditions outside of `moves.ts` and into this file.
- Added a new class, `MoveRestriction` that is similar to `MoveCondition`, except controls whether the move is able to be selected during command phases
- Added a `restrictions` method (and associated field) to the `Move` class to track restrictions, similar to conditions.

Cross-file changes:
- Added a new move flag, `Gravity`, for moves that fail when gravity is under effect.
  - Removed `failOnGravityCondition` and replaced it with the call to `f`
  - Added a new method to `moves`, `affectedByGravity` that sets the flag and also adds the Gravity move restriction

## Screenshots/Videos
TODO

## How to test the changes?
For move selection prevention
Attempt to select gigaton hammer / blood moon twice in a row
Attempt to select belch without having consumed a berry (and after having consumed a berry)
Attempt to select stuff cheeks without holding a berry
Attempt to select a gravity-affected move while gravity is in effect

## Checklist
- [ ] The PR is self-contained and cannot be split into smaller PRs? I did my best but.. these changes should really be all or nothing
- [ ] Have I provided a clear explanation of the changes? WIP

  - [ ] If so, please leave a link to it here: https://discord.com/channels/1125469663833370665/1289320675474931785/1406015868076560616 and https://discord.com/channels/1125469663833370665/1289320675474931785/1406349185300562111


src/data/moves/move-condition.ts
```diff
@@ -0,0 +1,117 @@
...
+  public declare readonly func: MoveConditionFunc;
+
+  /**
+   * @param func - A condition function that determines if the move can be used successfully
+   */
+  constructor(func?: MoveConditionFunc) {
+    if (func) {
+      this.func = func;
+    }
+  }
```
Bertie690: ... why can this be undefined

test/moves/belch.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      !game.field.getPlayerPokemon().isMoveSelectable(MoveId.BELCH),
+      "Belch should not be selectable without a berry",
+    );
+
+    player.battleData.hasEatenBerry = true;
+
+    expect(
+      game.field.getPlayerPokemon().isMoveSelectable(MoveId.BELCH),
+      "Belch should be selectable after eating a berry",
+    );
```
Bertie690, line 47:
```suggestion
    expect(
      game.field.getPlayerPokemon().isMoveSelectable(MoveId.BELCH),
      "Belch should be selectable after eating a berry",
    ).toBe(true);
```

remind me to make matchers for these

src/phases/move-phase.ts
```diff
@@ -24,7 +28,10 @@ import { applyMoveAttrs } from "#moves/apply-attrs";
 import { frenzyMissFunc } from "#moves/move-utils";
 import type { PokemonMove } from "#moves/pokemon-move";
 import { BattlePhase } from "#phases/battle-phase";
-import { NumberHolder } from "#utils/common";
+// biome-ignore lint/correctness/noUnusedImports: Used in a tsdoc comment
```
Bertie690: group these together in a range exclusion imo

src/phases/move-phase.ts
```diff
@@ -115,24 +112,78 @@ export class MovePhase extends BattlePhase {
...
+    if (
+      this.checkSleep() ||
+      this.checkFreeze() ||
+      this.checkPP() ||
+      this.checkValidity() ||
+      this.checkTagCancel(BattlerTagType.TRUANT, true) ||
+      this.checkPreUseInterrupt() ||
+      this.checkTagCancel(BattlerTagType.FLINCHED) ||
+      this.checkTagCancel(BattlerTagType.DISABLED, true) ||
+      this.checkGravity() ||
```
Bertie690: throat chop go here

src/phases/move-phase.ts
SirzBenjie: I had that and Biome was complaining about them not being sorted and also wouldn't fix it.
I'll be doing it once I finish the file.

test/moves/belch.test.ts
SirzBenjie: Why do I use `.toBe(true)`? Expect(bool) is sufficient.. Just to make it explicit?

src/data/moves/move-condition.ts
SirzBenjie: I'm going to find another way of doing it. But, basically, it's so we can call subclasses like `FirstMoveCondition` without requiring parameters, as those methods define their own `func`.

src/data/moves/move.ts
```diff
@@ -116,7 +117,20 @@ export abstract class Move implements Localizable {
...
   public attrs: MoveAttr[] = [];
+  /** Conditions that must be met for the move to succeed when it is used.
+   *
+   * @remarks Different from {@linkcode restrictions}, which is checked when the move is selected
+   */
   private conditions: MoveCondition[] = [];
+  /**
+   * Move failure conditions that occur during the second check (after move message and before )
+   */
+  private conditionsSeq2: MoveCondition[] = [];
```
Bertie690: this will need to be clarified fwiw

src/data/moves/move.ts
```diff
@@ -370,13 +384,57 @@ export abstract class Move implements Localizable {
...
+   */
+  public restriction(restriction: MoveRestriction): this;
+    /**
+   * Adds a restriction condition to this move.
+   * The move will not be selectable if at least 1 of its restrictions is met.
+   * @param restriction - The function or `MoveRestriction` that evaluates to `true` if the move is restricted from
+   *    being selected
+   * @param i18nkey - The i18n key for the restriction text
+   * @param alsoCondition - If `true`, also adds a {@linkcode MoveCondition} that checks the same condition when the
+   *    move is used; default `false`
```
```suggestion
   * @param alsoCondition - If `true`, also adds an equivalent {@linkcode MoveCondition} that will fail the move on use; default `false`
```

src/data/moves/move.ts
```diff
@@ -370,13 +384,57 @@ export abstract class Move implements Localizable {
...
+  /**
+   * Adds a restriction condition to this move.
+   * The move will not be selectable if at least 1 of its restrictions is met.
+   * @param restriction - The function or `MoveRestriction` that evaluates to `true` if the move is restricted from
+   *    being selected
+   * @param i18nkey - The i18n key for the restriction text, ignored if `restriction` is a `MoveRestriction`
+   * @param alsoCondition - If `true`, also adds a {@linkcode MoveCondition} that checks the same condition when the
+   *    move is used; default `false`. Ignored if `restriction` is a `MoveRestriction`.
+   * @returns `this` for method chaining
+   */
```
Bertie690, line 476: Do we need TSDocs on overload implementations?
(ie does it do anything)
Just for my own sanity lol

src/data/moves/move.ts
```diff
@@ -579,6 +637,19 @@ export abstract class Move implements Localizable {
     return this;
   }
 
+  /**
+   * Sets the {@linkcode MoveFlags.GRAVITY} flag for the calling Move and adds {@linkcode GravityUseRestriction} to the
+   * move's restrictions.
+   * 
+   * @see {@linkcode MoveId.GRAVITY}
+   * @returns The {@linkcode Move} that called this function
```
```suggestion
   * @returns `this`
```

src/data/moves/move.ts
```diff
@@ -116,7 +117,20 @@ export abstract class Move implements Localizable {
...
+   * @remarks Different from {@linkcode restrictions}, which is checked when the move is selected
+   */
   private conditions: MoveCondition[] = [];
+  /**
+   * Move failure conditions that occur during the second check (after move message and before )
+   */
+  private conditionsSeq2: MoveCondition[] = [];
+  /** Conditions that must be false for a move to be able to be selected.
+   * 
+   * @remarks Different from {@linkcode conditions}, which is checked when the move is invoked
```
Bertie690, line 175: is `@remarks` inline or multiline?
I forgor

Bertie690 (commented):
partail reviw

src/data/moves/move.ts
SirzBenjie: Well, it does give hover information, sort of....?

I don't know.... I think tsdoc renders it well.

src/data/moves/move.ts
SirzBenjie: It's a block tag, (so multiline)

src/data/moves/move-condition.ts
```diff
@@ -0,0 +1,263 @@
...
+/**
+ * Condition to allow a move's use only on the first turn this Pokemon is sent into battle
+ * (or the start of a new wave, whichever comes first).
+ */
+
+export class FirstMoveCondition extends MoveCondition {
+  constructor() {
+    super(user => {
+      return user.tempSummonData.waveTurnCount === 1;
+    });
```
Bertie690: For simple oneline conditions, i prefer a oneline arrow func
```suggestion
    super(user => user.tempSummonData.waveTurnCount === 1;);
```

src/data/moves/move-condition.ts
```diff
@@ -0,0 +1,263 @@
...
+    super(user => getCounterAttackTarget(user, damageCategory) !== null);
+  }
+}
+
+/** Condition check for counterattacks that proc againt physical moves */
+export const counterAttackCondition_Physical = new CounterAttackConditon(MoveCategory.PHYSICAL);
+/** Condition check for counterattacks that proc against special moves*/
+export const counterAttackCondition_Special = new CounterAttackConditon(MoveCategory.SPECIAL);
+/** Condition check for counterattacks that proc against moves regardless of damage type */
+export const counterAttackCondition_Both = new CounterAttackConditon();
```
Bertie690: mixed casing o

src/data/battler-tags.ts
```diff
@@ -814,8 +820,18 @@ export class ConfusedTag extends SerializableBattlerTag {
...
   }
 
+  /**
+   * Tick down the confusion duration and, if there are remaining turns, activate the confusion effect
+   *
+   * @remarks
+   * Rolls the
+   * @param pokemon - The pokemon with this tag
+   * @param lapseType - The lapse type
+   * @returns
```
DayKev: Some unfinished docs.

src/field/pokemon.ts
```diff
@@ -3295,9 +3297,15 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     }
   }
 
-  public trySelectMove(moveIndex: number, ignorePp?: boolean): boolean {
+  /**
+   * Attempt to select the move at the move index.
+   * @param moveIndex
+   * @param ignorePp
+   * @returns
```
DayKev: Noting another unfinished doc.

src/data/moves/pokemon-move.ts
```diff
@@ -38,33 +39,58 @@ export class PokemonMove {
...
+   * Checks whether this move can be performed by a Pokemon, without consideration for the move's targets.
    * The move is unusable if it is out of PP, restricted by an effect, or unimplemented.
    *
+   * Should not be confused with {@linkcode isSelectable}, which only checks if the move can be selected by a Pokemon.
+   *
    * @param pokemon - The {@linkcode Pokemon} attempting to use this move
    * @param ignorePp - Whether to ignore checking if the move is out of PP; default `false`
-   * @param ignoreRestrictionTags - Whether to skip checks for {@linkcode MoveRestrictionBattlerTag}s; default `false`
+   * @param forSelection - Whether this is being checked for move selection; default `false`
    * @returns Whether this {@linkcode PokemonMove} can be selected by this Pokemon.
```
Bertie690: outdated
```suggestion
   * @returns A 2-length tuple containing:
   * 1. Whether this {@linkcode PokemonMove} can be selected by this Pokemon.
   * 2. The text to be shown upon denying the move's usage or selection, or an empty string if the move is indeed selectable.
```

src/data/moves/pokemon-move.ts
```diff
@@ -38,33 +39,58 @@ export class PokemonMove {
...
    */
-  isUsable(pokemon: Pokemon, ignorePp = false, ignoreRestrictionTags = false): boolean {
+  public isUsable(pokemon: Pokemon, ignorePp = false, forSelection = false): [boolean, string] {
```
Bertie690:
```suggestion
  public isUsable(pokemon: Pokemon, ignorePp = false, forSelection = false): [usable: boolean, preventionText: string] {
```

src/data/moves/pokemon-move.ts
```diff
@@ -38,33 +39,58 @@ export class PokemonMove {
...
     return allMoves[this.moveId];
   }
 
+  /**
+   * Determine whether the move can be selected by the pokemon
+   *
+   * @param pokemon - The Pokemon under consideration
+   * @returns An array containing a boolean indicating whether the move can be selected, and a string with the reason if it cannot
+   */
+  public isSelectable(pokemon: Pokemon): [boolean, string] {
```
Bertie690:
```suggestion
  public isSelectable(pokemon: Pokemon): [selectable: boolean, preventionText: string] {
```

src/enums/battler-tag-lapse-type.ts
```diff
@@ -6,11 +8,23 @@ export enum BattlerTagLapseType {
   // TODO: This is unused...
   FAINT,
   /**
-   * Tag activate before the holder uses a non-virtual move, possibly interrupting its action.
+   * Tag activate before the holder uses a non-virtual move, after passing the first failure check sequence during the
```
Bertie690:
```suggestion
   * Tag activates before the holder uses a non-virtual move, after passing the first failure check sequence during the
```

src/phases/move-phase.ts
```diff
@@ -115,277 +110,683 @@ export class MovePhase extends BattlePhase {
...
+   * Follow up moves need to check a subset of the first failure checks
+   *
+   * @remarks
+   *
+   * Based on smogon battle mechanics research, checks happen in the following order:
+   * 1. Invalid move (skipped in pokerogue)
+   * 2. Move prevented by heal block
+   * 3. Move prevented by throat chop
+   * 4. Gravity
+   * 5. sky battle (unused in Pokerogue)
```
Bertie690: i feel bad for the Sky battle ME lol

src/phases/move-phase.ts
```diff
@@ -115,277 +110,683 @@ export class MovePhase extends BattlePhase {
...
     super.start();
 
+    if (!this.pokemon.isActive(true)) {
+      this.end();
+      return;
+    }
+
+    const user = this.pokemon;
+
+    // Removing gigaton hammer's two flags *always* happens first
```
Bertie690: uhhhh do you mean glaive rush?

src/phases/move-phase.ts
```diff
@@ -115,277 +110,683 @@ export class MovePhase extends BattlePhase {
...
+     * Move conditions assume the move has a single target
+     * TODO: is this sustainable?
+     */
+    const move = this.move.getMove();
+    const targets = this.getActiveTargetPokemon();
+    const arena = globalScene.arena;
+    const user = this.pokemon;
+
+    const failsConditions = !move.applyConditions(user, targets[0], 3);
+    const failedDueToTerrain = arena.isMoveTerrainCancelled(user, this.targets, move);
```
Bertie690, line 282: am i stupid but where di teh weather check go

src/phases/move-phase.ts
```diff
@@ -115,277 +110,683 @@ export class MovePhase extends BattlePhase {
...
+      });
+      failed = cancelled.value;
+    }
+
+    if (failed) {
+      this.failMove(failedDueToTerrain);
+      return true;
+    }
+
+    return false;
```
Bertie690, line 305: early return perchance?

src/phases/move-phase.ts
DayKev: You were probably missing a newline. It has to be
```
// biome-ignore-start ...
import type a
import type b
// biome-ignore-end

import c
import d
```

src/phases/move-phase.ts
SirzBenjie: The second failure check:
```ts
else if (arena.isMoveWeatherCancelled(user, move)) {
      failedText = getWeatherBlockMessage(globalScene.arena.getWeatherType());
      this.failed = true;
```

src/phases/move-phase.ts
SirzBenjie: yes, ty

src/phases/move-phase.ts
SirzBenjie: Oh right, we will have sky battles!! Nice!

src/data/moves/pokemon-move.ts
SirzBenjie: Ah thanks, I had forgotten about that syntax.

src/data/moves/move-condition.ts
SirzBenjie: i... can't believe I did that. thx

src/data/moves/move-condition.ts
SirzBenjie: Yep, same. Thx

src/phases/move-phase.ts
SirzBenjie: Notice line `291`

src/data/moves/move-condition.ts
SirzBenjie: Ah, it was like that from debugging lol.

src/data/moves/pokemon-move.ts
```diff
@@ -38,33 +39,59 @@ export class PokemonMove {
...
   }
 
+  /**
+   * Determine whether the move can be selected by the pokemon based on its own requirements
+   * @remarks
+   * Does not check for PP, moves blocked by challenges, or unimplemented moves, all of which are handled by {@linkcode isUsable}
+   * @param pokemon - The Pokemon under consideration
+   * @returns An array containing a boolean indicating whether the move can be selected, and a string with the reason if it cannot
+   */
+  public isSelectable(pokemon: Pokemon): [selectable: boolean, preventionText: string] {
```
DayKev, line 91: What is this method for, since there's already `Pokemon#isMoveSelectable`?

src/data/moves/move-condition.ts
```diff
@@ -0,0 +1,264 @@
...
+
+  getUserBenefitScore(_user: Pokemon, _target: Pokemon, _move: Move): number {
+    return 0;
+  }
+}
+/**
+ * Condition to allow a move's use only on the first turn this Pokemon is sent into battle
+ * (or the start of a new wave, whichever comes first).
+ */
+
```
Bertie690: NEWLINES GAAAAAAAAAAH

test/moves/belch.test.ts
Bertie690: correct - both explicitness and consistency.

src/data/moves/move.ts
```diff
@@ -1629,28 +1796,70 @@ export class MatchHpAttr extends FixedDamageAttr {
...
   }
+}
 
-  getCondition(): MoveConditionFunc {
-    return (user, target, move) => !!user.turnData.attacksReceived.filter(ar => this.moveFilter(allMoves[ar.move])).length;
+/**
+ * Attribute for counter-like moves to redirect the move to a different target
+ */
+export class CounterRedirectAttr extends MoveAttr {
+  declare private moveFilter?: MoveDamageCategory;
```
DayKev, line 1842: What's the `declare` for? Seems like it should just be a normal `private` field?

src/enums/move-flags.ts
```diff
@@ -47,8 +47,8 @@ export enum MoveFlags {
   CHECK_ALL_HITS = 1 << 16,
   /** Indicates a move is able to bypass its target's Substitute (if the target has one) */
   IGNORE_SUBSTITUTE = 1 << 17,
-  /** Indicates a move is able to be redirected to allies in a double battle if the attacker faints */
-  REDIRECT_COUNTER = 1 << 18,
   /** Indicates a move is able to be reflected by {@linkcode AbilityId.MAGIC_BOUNCE} and {@linkcode MoveId.MAGIC_COAT} */
-  REFLECTABLE = 1 << 19
+  REFLECTABLE = 1 << 18,
+  /** Indicates a move that fails when {@link https://bulbapedia.bulbagarden.net/wiki/Gravity_(move) | Gravity} is in effect */
```
Bertie690:
```suggestion
  /** Indicates a move should fail when {@link https://bulbapedia.bulbagarden.net/wiki/Gravity_(move) | Gravity} is in effect */
```

src/enums/battler-tag-lapse-type.ts
```diff
@@ -32,6 +46,16 @@ export enum BattlerTagLapseType {
...
    */
   AFTER_HIT,
-  /** The tag has some other custom activation or removal condition. */
+  /**
+   * The tag has some other custom activation or removal condition.
+   * @remarks
+   * Tags can use this lapse type to prevent them from being automatically lapsed during automatic lapse instances,
+   * such as before a move is used or at the end of a turn. Note that a tag's lapse method can still make use of
+   * these lapse types, which can be invoked via the `lapseTag` method on {@linkcode Pokemon} with the tag and
+   * lapse type.
```
Bertie690:
```suggestion
   * The tag has some other custom activation or removal condition.
   * @remarks
   * Tags can use this lapse type to prevent them from automatically lapsing during automatic lapse instances,
   * such as before a move is used or at the end of a turn. 
   * Such tags will only trigger upon being specifically lapsed with the tag and lapse type via {@linkcode Pokemon.lapseTag}.
```

Automatic lapse instances is a bit wordy but w/e

test/moves/copycat.test.ts
```diff
@@ -35,14 +35,15 @@ describe("Moves - Copycat", () => {
...
-  it("should copy the last move successfully executed", async () => {
+  it("should copy the last move executed across turns", async () => {
     game.override.enemyMoveset(MoveId.SUCKER_PUNCH);
     await game.classicMode.startBattle([SpeciesId.FEEBAS]);
 
     game.move.select(MoveId.SWORDS_DANCE);
     await game.toNextTurn();
 
     game.move.select(MoveId.COPYCAT); // Last successful move should be Swords Dance
+    await game.move.forceEnemyMove(MoveId.SPLASH);
```
DayKev, line 46: Why was the Splash needed here?

src/data/moves/move.ts
```diff
@@ -370,13 +427,76 @@ export abstract class Move implements Localizable {
...
+   * @param i18nkey - The i18n key for the restriction text, ignored if `restriction` is a `MoveRestriction`
+   * @param alsoCondition - If `true`, also adds a {@linkcode MoveCondition} that checks the same condition when the
+   *    move is used; default `false`. Ignored if `restriction` is a `MoveRestriction`.
+   * @param conditionSeq - The sequence number where the failure check occurs; default `4`. Ignored if `alsoCondition`
+   *    is false
+   * @returns `this` for method chaining
+   */
+  public restriction<T extends UserMoveConditionFunc | MoveRestriction>(
+    restriction: T,
+    i18nkey?: string,
```
DayKev, line 479: `i18nkey` doesn't seem to be used.

src/phases/move-phase.ts
```diff
@@ -115,277 +112,685 @@ export class MovePhase extends BattlePhase {
...
+    }
+
+    // Check if the move will heal
+    const move = this.move.getMove();
+    if (
+      move.findAttr(attr => attr.selfTarget && attr.is("HealStatusEffectAttr") && attr.isOfEffect(StatusEffect.FREEZE))
+    ) {
+      // On cartridge, burn up will not cure if it would fail
+      if (move.id === MoveId.BURN_UP && !this.pokemon.isOfType(PokemonType.FIRE)) {
       }
```
DayKev: The inside of this `if` block is missing.

src/phases/move-phase.ts
```diff
@@ -115,277 +112,685 @@ export class MovePhase extends BattlePhase {
...
+   * The only official move with a pre-use condition is Focus Punch
+   *
+   * @returns Whether the move was cancelled due to a pre-use interruption
+   * @see {@linkcode PreUseInterruptAttr}
+   */
+  private checkPreUseInterrupt(): boolean {
     const move = this.move.getMove();
+    const user = this.pokemon;
+    const target = this.getActiveTargetPokemon()[0];
+    return !!move.getAttrs("PreUseInterruptAttr").some(attr => {
```
DayKev: The `!!` doesn't seem to be necessary?

src/phases/move-phase.ts
```diff
@@ -115,277 +112,685 @@ export class MovePhase extends BattlePhase {
...
+    }
+
+    if (this.thirdFailureCheck()) {
+      console.log("Move failed during third failure check");
+      return;
+    }
+    /*
+    At this point, delayed moves (future sight, wish, doom desire) are issued, and if they occur, are
+    Then, combined pledge moves are checked for. Interestingly, the "wasMoveEffective" flag is set to false if the delay occurs
+    In either case, the phase should end here without proceeding
```
DayKev: > At this point, delayed moves (future sight, wish, doom desire) are issued, and if they occur, are

Unfinished doc

> Then, combined pledge moves are checked for. Interestingly, the "wasMoveEffective" flag is set to false if the delay occurs

"if the delay occurs"? You mean "if this was a delayed attack like Future Sight"?

> In either case, the phase should end here without proceeding 

Is this comment related to the above `if` block?

test/moves/copycat.test.ts
SirzBenjie: The last move used has to not be sucker punch, which will fail, and will actually be copied.
Basically, the test was incorrect. On cartridge, copycat *would* copy the failed sucker punch.

DayKev: Ah, gotcha.

src/phases/move-phase.ts
SirzBenjie: > "if the delay occurs"? You mean "if this was a delayed attack like Future Sight"?

Poor wording. If the delay was initiated.

src/data/moves/move.ts
SirzBenjie: It's because it's only conditionally assigned in the constructor, so we don't need it to exist as a property.

DayKev: What's the advantage of doing it this way, along with the extra `if`, instead of just this:
```ts
class x {
  private moveFilter?: MoveDamageCategory;

  constructor(moveFilter?: MoveDamageCategory) {
    super();
    this.moveFilter = moveFilter;
  }
```

src/data/moves/move-condition.ts
SirzBenjie: blame vscode

src/data/battler-tags.ts
```diff
@@ -709,18 +709,21 @@ class NoRetreatTag extends TrappedTag {
...
    * @param pokemon - The {@linkcode Pokemon} with this tag.
-   * @param lapseType - The {@linkcode BattlerTagLapseType | lapse type} used for this function call.
+   * @param lapseType - The {@linkcode BattlerTagLapseType | lapse type} used for this function call. Must be {@linkcode BattlerTagLapseType.PRE_MOVE} in order to apply the flinch effect.
    * @returns Whether the tag should remain active.
    */
   lapse(pokemon: Pokemon, lapseType: BattlerTagLapseType): boolean {
     if (lapseType === BattlerTagLapseType.PRE_MOVE) {
-      (globalScene.phaseManager.getCurrentPhase() as MovePhase).cancel();
+      const currentPhase = globalScene.phaseManager.getCurrentPhase();
+      if (currentPhase?.is("MovePhase")) {
```
Bertie690: Remove optional chains from all the code plsss
I don't want to need to regex the stuff away again

Bertie690 (changes requested):
Please un-Todo the instruct tests as well thx

src/data/moves/move-condition.ts
```diff
@@ -0,0 +1,264 @@
...
+ * @remarks
+ * For trainer pokemon, just checks if there are any benched pokemon allowed in battle
+ *
+ * Wild pokemon cannot teleport if either:
+ * - The current battle is a double battle
+ * - They are under the effects of a *move-based* trapping effect like and are neither a ghost type nor have an active run away ability
+ */
+export const failTeleportCondition = new MoveCondition(user => {
+  if (user.hasTrainer()) {
+    const party = user.isPlayer() ? globalScene.getPlayerParty() : globalScene.getEnemyParty();
```
Bertie690, line 89: I am making a helper function to check this exact thing for switch out moves (grab all backup Pokemon that are off field and allowed in battle)

src/data/moves/move.ts
SirzBenjie: Whether or not the property exists on the object via `in` etc.

DayKev: Do we not want it to exist for some reason?

test/moves/copycat.test.ts
Bertie690: Does the pr fix it?

src/phases/move-phase.ts
```diff
@@ -97,303 +106,695 @@ export class MovePhase extends PokemonPhase {
...
-    // Skip for follow ups/reflected moves, no status condition or post turn statuses (e.g. Poison/Toxic)
-    if (!this.pokemon.status?.effect || this.pokemon.status.isPostTurn() || isIgnoreStatus(this.useMode)) {
-      return;
+  private cureStatus(effect: StatusEffect, msg?: string): void {
+    const pokemon = this.pokemon;
+    // Freeze healed by move uses its own msg
+    globalScene.phaseManager.queueMessage(msg ?? getStatusEffectHealText(effect, getPokemonNameWithAffix(pokemon)));
+    pokemon.resetStatus(undefined, undefined, undefined, false);
+    pokemon.updateInfo();
+  }
```
emdeann: Would this make more sense to be a method on `Pokemon` rather than on the `MovePhase`? Although the move phase is the only user, in a vacuum there is nothing in the method that is related to moves in particular

src/phases/move-phase.ts
```diff
@@ -97,303 +106,695 @@ export class MovePhase extends PokemonPhase {
...
+
+  /**
+   * Queue the status activation message, play its animation, and cancel the move
+   *
+   * @param effect - The effect being triggered
+   * @param cancel - Whether to cancel the move after triggering the status effect; default `true`;
+   *    pass `false` to only trigger the animation and message without cancelling the move (e.g. sleep talk/snore)
+   */
+  private triggerStatus(effect: StatusEffect, cancel = true): void {
```
emdeann, line 530: Do we need to pass `effect` as a parameter? Surely we are only triggering the status that the pokemon has, and we have access to that through `pokemon` itself. 

This seems to me like it might also be better as a method on `Pokemon` (with the cancellation just happening after the call)

src/phases/move-phase.ts
```diff
@@ -97,303 +106,695 @@ export class MovePhase extends PokemonPhase {
...
+      return false;
+    }
+
+    // For some reason, dancer will immediately thaw its user
+    if (this.useMode === MoveUseMode.INDIRECT) {
+      this.pokemon.resetStatus(false);
+      return false;
+    }
+
+    // Check if move use would heal the user
```
emdeann: This comment is misplaced and redundant with the one on L546

emdeann (commented):
All I could get through tonight is move phase but this is looking sick

src/data/battler-tags.ts
```diff
@@ -699,18 +699,21 @@ class NoRetreatTag extends TrappedTag {
...
   constructor(sourceMove: MoveId) {
-    super(BattlerTagType.FLINCHED, [BattlerTagLapseType.PRE_MOVE, BattlerTagLapseType.TURN_END], 1, sourceMove);
+    super(BattlerTagType.FLINCHED, BattlerTagLapseType.TURN_END, 1, sourceMove);
   }
 
   /**
    * Cancels the flinched Pokemon's currently used move this turn if called mid-execution, or removes the tag at end of turn.
    * @param pokemon - The {@linkcode Pokemon} with this tag.
-   * @param lapseType - The {@linkcode BattlerTagLapseType | lapse type} used for this function call.
+   * @param lapseType - The {@linkcode BattlerTagLapseType | lapse type} used for this function call. Must be {@linkcode BattlerTagLapseType.PRE_MOVE} in order to apply the flinch effect.
```
```suggestion
   * @param lapseType - The {@linkcode BattlerTagLapseType | lapse type} used for this function call. 
   *   Must be {@linkcode BattlerTagLapseType.PRE_MOVE} in order to apply the flinch effect
```

src/phases/move-phase.ts
```diff
@@ -97,303 +106,695 @@ export class MovePhase extends PokemonPhase {
...
+    return false;
   }
 
   public start(): void {
     super.start();
 
-    console.log(
-      `%cMove: ${MoveId[this.move.moveId]}\nUse Mode: ${enumValueToKey(MoveUseMode, this.useMode)}`,
-      `color:${MOVE_COLOR}`,
-    );
```
Bertie690, line 129: This console log got incorrectly transferred over, pls revert

src/phases/move-phase.ts
```diff
@@ -97,303 +106,695 @@ export class MovePhase extends PokemonPhase {
...
+   * 11. Sound move with throat chop
+   * 12. Failure due to gravity
+   * 13. Move lock from choice items / gorilla tactics
+   * 14. Failure from taunt
+   * 15. Failure from imprison
+   * 16. Failure from confusion
+   * 17. Failure from paralysis
+   * 18. Failure from infatuation
+   */
+  protected firstFailureCheck(): boolean {
```
Bertie690, line 147: A minor nit - if you're reworking the file would it make sense to move the methods in order of execution? 
(have the failure checks go after `start`)

src/phases/move-phase.ts
SirzBenjie: The move phase is a bit of a misnomer. It's related to checking if a user can, well, issue its _move_. _That_ is the responsibility of the _move_ phase. So I strongly believe that anything that happens during this phase is actually the purview of it.
This "cureStatus" effect is emphatically only the process for curing status as part of a move interaction. Placing it on the Pokemon class can run the risk of it being misunderstood to be the generic "cure a status" effect.

src/phases/move-phase.ts
SirzBenjie: Regarding the first, I prefer to have it be explicit rather than relying on an assumption that a Pokémon will only ever have one status. While this is true in general, doing so here would bake that logic into the game. Also, it would add an additional, unnecessary check inside of the method; we are already getting the status in the instances where the method is called.

As for whether it should be on the Pokemon class, I don't think so, again, as this is specifically only triggering the status effect as it would occur during the move phase. (Not, e.g., poison / burn's damage).

src/enums/battler-tag-lapse-type.ts
```diff
@@ -6,11 +8,23 @@ export enum BattlerTagLapseType {
   // TODO: This is unused...
   FAINT,
   /**
-   * Tag activate before the holder uses a non-virtual move, possibly interrupting its action.
+   * Tag activates before the holder uses a non-virtual move, after passing the first failure check sequence during the
+   * move phase.
```
DayKev:
```suggestion
   * Tag activates before the holder uses a non-virtual move,
   * after passing the first failure check sequence during the move phase.
```

src/data/moves/move-condition.ts
```diff
@@ -0,0 +1,264 @@
...
+   * @param damageCategory - The category of move to counter (physical or special), or `undefined` to counter both
+   */
+  constructor(damageCategory?: MoveDamageCategory) {
+    super(user => getCounterAttackTarget(user, damageCategory) !== null);
+  }
+}
+
+/** Condition check for counterattacks that proc againt physical moves */
+export const counterAttackConditionPhysical = new CounterAttackConditon(MoveCategory.PHYSICAL);
+/** Condition check for counterattacks that proc against special moves*/
```
Bertie690:
```suggestion
/** Condition check for counterattacks that proc against special moves */
```

src/data/moves/move-condition.ts
```diff
@@ -0,0 +1,264 @@
...
+/**
+ * A restriction that prevents a move from being selected
+ *
+ * @remarks
+ * Only checked when the move is selected, but not when it is attempted to be invoked. To prevent a move from being used,
+ * use a {@linkcode MoveCondition} instead.
+ */
+export class MoveRestriction {
+  public declare readonly func: UserMoveConditionFunc;
+  public declare readonly i18nkey: string;
```
Bertie690:
```suggestion
  /** A lambda function used to determine the move's eligibility. */
  public declare readonly func: UserMoveConditionFunc;
  /** 
   * The `i18n` locales key to be shown when restricting a move. 
   * @defaultValue `battle:moveRestricted`
   */
  public declare readonly i18nkey: string;
```

Bertie690: Also why are these using declare and not just normal public properties?

src/data/moves/move-utils.ts
```diff
@@ -133,3 +135,25 @@ export const frenzyMissFunc: UserMoveConditionFunc = (user: Pokemon, move: Move)
 
   return true;
 };
+
+/**
+ * Determine the target for the `user`'s counter-attack move
+ * @param user - The pokemon using the counter-like move
+ * @param damageCategory - The category of move to counter (physical or special), or `undefined` to counter both
+ * @returns - The battler index of the most recent, non-ally attacker using a move that matches the specified category, or `null` if no such attacker exists
```
Bertie690, line 143:
```suggestion
 * @returns - The `BattlerIndex` of the most recent, non-ally attacker having used a move matching the specified category, or `null` if no such attacker exists
```

Bertie690: Also should this return `undefined`?
just a thought

src/data/moves/pokemon-move.ts
```diff
@@ -38,32 +39,59 @@ export class PokemonMove {
...
+   * Checks whether this move can be performed by a Pokemon, without consideration for the move's targets.
    * The move is unusable if it is out of PP, restricted by an effect, or unimplemented.
    *
+   * Should not be confused with {@linkcode isSelectable}, which only checks if the move can be selected by a Pokemon.
+   *
    * @param pokemon - The {@linkcode Pokemon} attempting to use this move
    * @param ignorePp - Whether to ignore checking if the move is out of PP; default `false`
-   * @param ignoreRestrictionTags - Whether to skip checks for {@linkcode MoveRestrictionBattlerTag}s; default `false`
-   * @returns Whether this {@linkcode PokemonMove} can be selected by this Pokemon.
+   * @param forSelection - Whether this is being checked for move selection; default `false`
```
Bertie690, line 49: I'm still a bit shaky on the `forSelection` parameter here.
If usability and selectability are defined as different things, then why does the former also have an option to check the latter?

src/enums/move-category.ts
```diff
@@ -3,3 +3,6 @@ export enum MoveCategory {
   SPECIAL,
   STATUS,
 }
+
+/** Type of damage categories */
```
Bertie690:
```suggestion
/** Type representing the category of a damaging move (physical/special) */
```

src/enums/move-category.ts
```diff
@@ -3,3 +3,6 @@ export enum MoveCategory {
...
+export type MoveDamageCategory = Exclude<MoveCategory, MoveCategory.STATUS>;
```
Bertie690, line 8: personally i prefer the wording `DamagingMoveCategory` but we can always rename later tbh

src/field/pokemon.ts
```diff
@@ -2410,8 +2410,8 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * modifiers from move and ability attributes
    * @param source - The attacking Pokémon.
    * @param move - The move being used by the attacking Pokémon.
-   * @param ignoreAbility - Whether to ignore abilities that might affect type effectiveness or immunity (defaults to `false`).
-   * @param simulated - Whether to apply abilities via simulated calls (defaults to `true`)
+   * @param ignoreAbility - Whether to ignore abilities that might affect type effectiveness or immunity; default `false`
+   * @param simulated - Whether to apply abilities via simulated calls; default `true`; ⚠️ Should only ever be false during `moveEffect` phase
```
Bertie690: idk about you but 3 semicolons on the same line looks kinda weird ngl

src/phases/command-phase.ts
```diff
@@ -205,39 +201,18 @@ export class CommandPhase extends FieldPhase {
...
-    } else {
-      // TODO: Consider a message that signals a being unusable for an unknown reason
-      cannotSelectKey = "";
-    }
-
-    const moveName = move.getName().replace(" (N)", ""); // Trims off the indicator
-
-    globalScene.ui.showText(
-      i18next.t(cannotSelectKey, { moveName }),
+  private queueFightErrorMessage(msg: string) {
```
Bertie690:
```suggestion
  private queueFightErrorMessage(msg: string): void {
```

src/phases/command-phase.ts
```diff
@@ -205,39 +201,18 @@ export class CommandPhase extends FieldPhase {
...
+    const ui = globalScene.ui;
+    ui.setMode(UiMode.MESSAGE);
```
Bertie690, line 209: Do we need to await this?

src/phases/move-phase.ts
```diff
@@ -27,10 +32,38 @@ import { applyMoveAttrs } from "#moves/apply-attrs";
...
+ * - Applies pre-execution move effects, including ability-based type changes
+ *   and triggering move animations.
+ * - Adds the record to the Pokémon's move history.
+ * - Handles move cancellation and the first 3 failure checks (those that appear
+ *   before the move's animation should play)
+ *
+ * @see {@link PokemonPhase}
+ * @see {@link MoveUseMode}
+ * @see {@link MovePhaseTimingModifier}
+ * @see {@link TurnMove}
```
Bertie690: these should be linkcodes
also imo adding linkcodes for the type of the function parameters is maybe not needed? idk

```suggestion
 * @see {@linkcode PokemonPhase}
 * @see {@linkcode MoveUseMode}
 * @see {@linkcode MovePhaseTimingModifier}
 * @see {@linkcode TurnMove}
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+   * @remarks
+   * - If the user is asleep but can use the move, the sleep animation and message is still shown
+   * - If the user is frozen but is thawed from its move, the user's status is cured and the thaw message is shown
+   */
+  private doThawCheck(): void {
+    const user = this.pokemon;
+
+    if (isIgnoreStatus(this.useMode)) {
+      return;
+    }
```
Bertie690, line 201: swap em

```suggestion
    if (isIgnoreStatus(this.useMode)) {
      return;
    }
    
    const user = this.pokemon;
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+    } else {
+      // Powder *always* happens last
+      // Note: Powder's lapse method handles everything: messages, damage, animation, primal weather interaction,
+      // determining type of type changing moves, etc.
+      // It will set this phase's `failed` flag to true if it procs
+      user.lapseTag(BattlerTagType.POWDER, BattlerTagLapseType.PRE_MOVE);
+      return this.failed;
+    }
+    if (this.failed) {
+      this.showFailedText(failedText);
```
Bertie690, line 252:
```suggestion
      this.showFailedText(failedText);
      return true;
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+    }
+    return this.failed;
```
Bertie690, line 254:
```suggestion
    return false;
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+    user.turnData.acted = true;
+    const useMode = this.useMode;
+    const ignoreStatus = isIgnoreStatus(useMode);
+    const isFollowUp = useMode === MoveUseMode.FOLLOW_UP;
+    if (!ignoreStatus) {
+      this.firstFailureCheck();
+      user.lapseTags(BattlerTagLapseType.PRE_MOVE);
+      // At this point, called moves should be decided.
+      // For now, this comment works as a placeholder until called moves are reworked
+      // For correct alignment with mainline, this SHOULD go here, and this phase SHOULD rewrite its own move
```
Bertie690, line 409: Make this a TODO

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+    }
 
     // Reset hit-related turn data when starting follow-up moves (e.g. Metronomed moves, Dancer repeats)
-    if (isVirtual(this.useMode)) {
-      this.pokemon.turnData.hitsLeft = -1;
-      this.pokemon.turnData.hitCount = 0;
+    if (isVirtual(useMode)) {
+      const turnData = user.turnData;
+      turnData.hitsLeft = -1;
+      turnData.hitCount = 0;
```
Bertie690, line 431:
```suggestion
      user.turnData.hitsLeft = -1;
      user.turnData.hitCount = 0;
```

tbh we should make a sub method for it lol

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
       })
     ) {
-      globalScene.arena.setIgnoreAbilities(true, this.pokemon.getBattlerIndex());
+      globalScene.arena.setIgnoreAbilities(true, user.getBattlerIndex());
     }
 
-    this.resolveRedirectTarget();
+    // At this point, move's type changing and multi-target effects *should* be applied
+    // Pokerogue's current implementation applies these effects during the move effect phase
+    // as there is not (yet) a notion of a move-in-flight for determinations to occur
```
Bertie690, line 449: again a todo would be swell so i can add it to my list

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
-    this.lapsePreMoveAndMoveTags();
+    if (!isFollowUp) {
+      // Gorilla tactics lock in (and choice items if they are ever added)
+      // Stance Change form change
+      // Struggle's "There are no more moves it can use" message
 
-    if (!(this.failed || this.cancelled)) {
-      this.resolveFinalPreMoveCancellationChecks();
+      globalScene.triggerPokemonFormChange(user, SpeciesFormChangePreMoveTrigger);
+      // TODO: apply gorilla tactics
```
Bertie690:
```suggestion
      // TODO: apply gorilla tactics here instead of the move effect phase
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
    */
-  protected resolvePreMoveStatusEffects(): void {
-    // Skip for follow ups/reflected moves, no status condition or post turn statuses (e.g. Poison/Toxic)
-    if (!this.pokemon.status?.effect || this.pokemon.status.isPostTurn() || isIgnoreStatus(this.useMode)) {
-      return;
+  private cureStatus(effect: StatusEffect, msg?: string): void {
+    const pokemon = this.pokemon;
+    // Freeze healed by move uses its own msg
+    globalScene.phaseManager.queueMessage(msg ?? getStatusEffectHealText(effect, getPokemonNameWithAffix(pokemon)));
+    pokemon.resetStatus(undefined, undefined, undefined, false);
```
Bertie690: I would like the comment about `asPhase=false` re-added here thx

it is important enough that changing it will break sleep talk/etc

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+    pokemon.updateInfo();
+  }
+
+  /**
+   * Queue the status activation message, play its animation, and cancel the move
+   *
+   * @param effect - The effect being triggered
+   * @param cancel - Whether to cancel the move after triggering the status effect; default `true`;
+   *    pass `false` to only trigger the animation and message without cancelling the move (e.g. sleep talk/snore)
```
Bertie690: @SirzBenjie also how many spaces do we want to use for indents?
that's like a 3 space indentation relative to the prior line

Bertie690:
```suggestion
   * @param cancel - Whether to cancel the move after triggering the status effect animation & message; default `true`.
   *   Set to `false` for sleep-bypassing moves to avoid cancelling attack
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+   * Handle the sleep check
+   * @returns Whether the move was cancelled due to sleep
+   */
+  protected checkSleep(): boolean {
+    const user = this.pokemon;
+    if (user.status?.effect !== StatusEffect.SLEEP) {
+      return false;
+    }
+
+    // For some reason, dancer will immediately wake its user from sleep when triggering
```
Bertie690, line 554:
```suggestion
    // Dancer will immediately wake its user from sleep when triggering
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+      pokemon: user,
+      statusEffect: user.status.effect,
+      duration: turnsRemaining,
+    });
+
+    user.status.sleepTurnsRemaining = turnsRemaining.value;
+    if (user.status.sleepTurnsRemaining <= 0) {
+      this.cureStatus(StatusEffect.SLEEP);
+      return false;
+    }
```
Bertie690, line 572:
```suggestion
    if (turnsRemaining.value <= 0) {
      this.cureStatus(StatusEffect.SLEEP);
      return false;
    }
    
     user.status.sleepTurnsRemaining = turnsRemaining.value;
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+   *    Thawing from a freeze move is not applied until _after_ all other failure checks.
+   * - Activating the freeze status effect (cancelling the move, playing the message, and displaying the animation)
+   * @returns Whether the move was cancelled due to the pokemon being frozen
+   */
+  protected checkFreeze(): boolean {
+    if (this.pokemon.status?.effect !== StatusEffect.FREEZE) {
+      return false;
+    }
+
+    // For some reason, dancer will immediately thaw its user
```
Bertie690, line 598:
```suggestion
    // Dancer will immediately thaw its user when using a move
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+      move.findAttr(attr => attr.selfTarget && attr.is("HealStatusEffectAttr") && attr.isOfEffect(StatusEffect.FREEZE))
+      && (move.id !== MoveId.BURN_UP || this.pokemon.isOfType(PokemonType.FIRE, true, true))
     ) {
-      // Dancer thaws out or wakes up a frozen/sleeping user prior to use
-      this.pokemon.resetStatus(false);
-      return;
+      this.thaw = true;
+      return false;
+    }
+    if (
```
Bertie690, line 617:
```suggestion

    // Thaw if using a self-thawing move or a 20% roll succeeds
    if (
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+      Overrides.STATUS_ACTIVATION_OVERRIDE === false
```
Bertie690, line 618: Pls revert the ordering change - the overrides check should be after the check for self thawing move attrs

overrides are always meant to defer to guaranteed effects AFAIK (or they _should_ do that anyhow)

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
-  protected lapsePreMoveAndMoveTags(): void {
-    this.pokemon.lapseTags(BattlerTagLapseType.PRE_MOVE);
+  protected checkValidity(): boolean {
+    const move = this.move;
+    const moveId = move.moveId;
+    const moveName = move.getName();
+    let failedText: string | undefined;
+    const usability = new BooleanHolder(false);
+    if (moveName.endsWith(" (N)")) {
+      failedText = i18next.t("battle:moveNotImplemented", { moveName: moveName.replace(" (N)", "") });
```
Bertie690, line 664: may i request this be early return-ed a bit?

(ie copy paste L653-ish in here and after the 2nd to last `else if` block)

I know it's marginally more code but it makes it more readable at least for me

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+  private checkPreUseInterrupt(): boolean {
     const move = this.move.getMove();
+    const user = this.pokemon;
+    const target = this.getActiveTargetPokemon()[0];
+    return move.getAttrs("PreUseInterruptAttr").some(attr => {
+      attr.apply(user, target, move);
+      if (this.cancelled) {
+        return true;
+      }
+      return false;
```
Bertie690, line 710:
```suggestion
      return this.cancelled;
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+      // Play the failure message
+      this.showFailedText(
+        i18next.t("battle:moveDisabledGravity", {
+          pokemonNameWithAffix: getPokemonNameWithAffix(this.pokemon),
+          moveName: move.name,
+        }),
+      );
+      return true;
     }
+    return false;
```
Bertie690:
```suggestion
    if (!globalScene.arena.hasTag(ArenaTagType.GRAVITY) || !move.hasFlag(MoveFlags.GRAVITY)) {
      return false;
    }

    // Play the failure message
    this.showFailedText(
      i18next.t("battle:moveDisabledGravity", {
        pokemonNameWithAffix: getPokemonNameWithAffix(this.pokemon),
        moveName: move.name,
      }),
    );
    return true;

```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+  /**
+   * Handle the paralysis status effect check, cancelling the move and queueing the activation message and animation
+   *
+   * @returns Whether the move was cancelled due to paralysis
+   */
+  private checkPara(): boolean {
+    if (this.pokemon.status?.effect !== StatusEffect.PARALYSIS) {
+      return false;
+    }
+    const proc =
```
Bertie690:
```suggestion

    const proc =
```
i like having newlines between guard clauses and acutal functionality

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+      (this.pokemon.randBattleSeedInt(4) === 0 || Overrides.STATUS_ACTIVATION_OVERRIDE === true)
+      && Overrides.STATUS_ACTIVATION_OVERRIDE !== false;
```
Bertie690:
```suggestion
    const proc = Overrides.STATUS_ACTIVATION_OVERRIDE !== null 
      ? !Overrides.STATUS_ACTIVATION_OVERRIDE
      : this.pokemon.randBattleSeedInt(4) === 0
```

marginally more readable trust

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
 
-    /**
-     * Determine if the move is successful (meaning that its damage/effects can be attempted)
-     * by checking that all of the following are true:
-     * - Conditional attributes of the move are all met
-     * - Weather does not block the move
-     * - Terrain does not block the move
-     */
+  /**
+   * Clear out two turn moves, then schedule the move to be used if it passes the third failure check
```
Bertie690: missing `@param` for charging

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+      (this.pokemon.randBattleSeedInt(4) === 0 || Overrides.STATUS_ACTIVATION_OVERRIDE === true)
+      && Overrides.STATUS_ACTIVATION_OVERRIDE !== false;
+    if (!proc) {
+      return false;
     }
+    this.triggerStatus(StatusEffect.PARALYSIS);
+    return true;
+  }
 
+  protected usePP(): void {
```
Bertie690, line 764: missing doc comment

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
+    */
+
+    // Currently, we only do the libero/protean type change here
+
+    applyAbAttrs("PokemonTypeChangeAbAttr", { pokemon: user, move, opponent });
+
+    // TODO: Move this to the Move effect phase where it belongs.
+    // Fourth failure check happens _after_ protean
+    if (!move.applyConditions(user, opponent, 4)) {
+      console.log("Move failed during fourth failure check");
```
Bertie690, line 823: pls remove log statement - more test clutter

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
-    const targets = this.getActiveTargetPokemon();
-
-    // Update the battle's "last move" pointer unless we're currently mimicking a move or triggering Dancer.
-    if (!move.hasAttr("CopyMoveAttr") && !isReflected(this.useMode)) {
-      globalScene.currentBattle.lastMove = move.id;
-    }
+    const targets = this.targets;
 
     // Trigger ability-based user type changes, display move text and then execute move effects.
     // TODO: Investigate whether PokemonTypeChangeAbAttr can drop the "opponent" parameter
```
Bertie690, line 842: these comments are methinks a bit outdated and should be (re)moved

src/phases/move-phase.ts
```diff
@@ -415,22 +842,21 @@ export class MovePhase extends PokemonPhase {
...
-    const moveHistoryEntry = this.moveHistoryEntry;
-    moveHistoryEntry.result = MoveResult.FAIL;
-    this.pokemon.pushMoveHistory(moveHistoryEntry);
+    this.pokemon.pushMoveHistory({
+      move: this.move.moveId,
+      targets: this.targets,
+      result: MoveResult.FAIL,
+      useMode: this.useMode,
+    });
+    console.log("==========PUSHING MOVE HISTORY WITH FAIL FOR %s=============", MoveId[this.move.moveId]);
```
Bertie690: see above wrt. console log clutter

src/phases/move-phase.ts
```diff
@@ -574,33 +984,19 @@ export class MovePhase extends PokemonPhase {
    * to reflect the actual battler index of the user's last attacker.
    *
    * If there is no last attacker or they are no longer on the field, a message is displayed and the
-   * move is marked for failure.
-   * @todo Make this a feature of the move rather than basing logic on {@linkcode BattlerIndex.ATTACKER}
+   * move is marked for failure
```
Bertie690, line 951:
```suggestion
   * move is marked for failure.
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
     }
 
     this.end();
   }
 
-  /** Check for cancellation edge cases - no targets remaining, or {@linkcode MoveId.NONE} is in the queue */
-  protected resolveFinalPreMoveCancellationChecks(): void {
+  /**
+   * Check for cancellation edge cases - no targets remaining
+   * @returns Whether the move fails
```
Bertie690, line 499:
```suggestion
   * @returns Whether the move failed
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
-          || (!this.pokemon.randBattleSeedInt(5) && Overrides.STATUS_ACTIVATION_OVERRIDE !== true)
-          || Overrides.STATUS_ACTIVATION_OVERRIDE === false;
-
-        activated = !healed;
-        break;
-    }
-
-    if (activated) {
-      // Cancel move activation and play effect
+    // Otherwise, trigger the freeze status effect
```
Bertie690: I think this is a bit self explanatory...

```suggestion
```

src/phases/move-phase.ts
```diff
@@ -97,303 +134,693 @@ export class MovePhase extends PokemonPhase {
...
-   * Lapse {@linkcode BattlerTagLapseType.PRE_MOVE | PRE_MOVE} tags that trigger before a move is used, regardless of whether or not it failed.
-   * Also lapse {@linkcode BattlerTagLapseType.MOVE | MOVE} tags if the move is successful and not called indirectly.
+   * Check if the move is valid and not in an error state
+   *
+   * @remarks
+   * Checks occur in the following order
+   * 1. Move is not implemented
+   * 2. Move is somehow invalid (it is {@linkcode MoveId.NONE} or {@linkcode targets} is somehow empty)
+   * 3. Move cannot be used by the player due to a challenge
+   *
```
Bertie690, line 654:
```suggestion
```

DayKev:
https://github.com/pagefaultgames/pokerogue/pull/6276#discussion_r2286663638
https://github.com/pagefaultgames/pokerogue/pull/6276#discussion_r2346983861

src/field/pokemon.ts
DayKev:
```suggestion
   * @param simulated - (Default `true`) Whether to apply abilities via simulated calls. \
   *   ⚠️ Should only ever be false during `MoveEffectPhase`
```

src/phases/move-phase.ts
DayKev: Should probably be 2 spaces.

src/data/moves/move.ts
SirzBenjie: Why add extra data unnecessarily?

src/data/moves/move-condition.ts
SirzBenjie: Ok, well, when that is in, we can update this spot here.

src/data/moves/move-condition.ts
SirzBenjie: Because of javascript property fields. Not doing this and typescript will leave them in, meaning more code.

src/data/moves/move-utils.ts
SirzBenjie: I prefer returning `null`, not `undefined`, which signals a conscious, deliberate choice. `undefined` is auto returned by a slew of things, like accessing a property on a field, etc.

src/data/moves/pokemon-move.ts
SirzBenjie: > Based on its *own* requirements.

This is supposed to be where things intrinsic to the move are checked, that does not account for things like disabled, encore, heal block, etc. It's a placeholder, for now.

src/phases/command-phase.ts
SirzBenjie: It's not awaited anywhere else. Also that's the responsibility of the UI rework

src/phases/move-phase.ts
SirzBenjie: Ehhh I cba atm to re order. I don't really care much about method ordering. Mostly because I use code jumps.

src/phases/move-phase.ts
SirzBenjie: Nah leave it like this. One fewer statement to parse.

src/phases/move-phase.ts
SirzBenjie: Why suggest this change? I had it like that for a reason.

src/phases/move-phase.ts
SirzBenjie: probably lost during a merge. thx

src/phases/move-phase.ts
SirzBenjie: There was a lint saying that line continuations should use 3 spaces at one point.

src/phases/move-phase.ts
SirzBenjie: This is the logic that was there before. The ability effect is applied, which is the holder for early bird. Moving this before is not correct.

src/phases/move-phase.ts
SirzBenjie: Ah thanks, that was left in.

src/phases/move-phase.ts
SirzBenjie: No, I'm leaving this comment in. It's silly

src/data/moves/move.ts
DayKev: Should we be doing this for all optional class fields, then?

src/phases/move-phase.ts
DayKev: Eh? What lint is that?

src/data/moves/move-condition.ts
DayKev: In this case, aren't these not optional? So shouldn't they indeed exist/be left in?

src/phases/move-phase.ts
SirzBenjie: The logic here is easy to follow and is written in a way where the failures can occur. I don't see a way to make it early return.

src/phases/move-phase.ts
SirzBenjie: Agree, that's good!

src/data/moves/move.ts
SirzBenjie: Maybe, but it's minor enough that I just avoid it when I write it.

src/data/moves/move-condition.ts
SirzBenjie: > In this case, aren't these not optional? So shouldn't they indeed exist/be left in?

Ah right, yes.

src/data/moves/pokemon-move.ts
DayKev: Ah, there should probably be a comment explaining that the original method is planned to be trimmed down and this one added to.

src/phases/move-phase.ts
emdeann: What about this effect is specific to the move phase? I understand the first part of what you're saying to an extent, but this seems about as generic as it can be.

Another way to phrase that is, if this were on the Pokemon class, why would I be "incorrect" to use it for something like an ME or item effect?

src/data/moves/move.ts
DayKev: I don't like adding more inconsistencies into the codebase. Are there circumstances (that can or already do exist in this codebase) where it's measurably better to do it this way? Would we actually notice a memory difference if we converted every optional field to use this pattern? Would the increase in bundle size (from changing `this.x = x;` to `if (x !== undefined) { this.x = x; }` everywhere) be worth it?

src/phases/move-phase.ts
```diff
@@ -38,18 +44,21 @@ export class MovePhase extends PokemonPhase {
   protected _targets: BattlerIndex[];
   public readonly useMode: MoveUseMode; // Made public for quash
   /** The timing modifier of the move (used by Quash and to force called moves to the front of their queue) */
-  public timingModifier: MovePhaseTimingModifier;
+  public readonly timingModifier: MovePhaseTimingModifier;
```
DayKev: This isn't supposed to be `readonly`:
image

SirzBenjie: This is an example of that biome lint going wrong.

src/phases/move-phase.ts
SirzBenjie: I can't find it...

src/data/moves/pokemon-move.ts
SirzBenjie: It was to avoid code churn.

src/phases/move-phase.ts
SirzBenjie: > What about this effect is specific to the move phase?

Well, I wanted to say the message, but I think you end up being right. This does really make sense on the Pokémon class. Will update.

Bertie690 (approved):

PR #6277, opened 2025-08-16
[UI/UX][Challenge] Starter select screen shows reduced information in fresh start

Wlowscha:
When selecting the fresh start challenge mode, only moves, abilities etc which the player can use will be displayed.

Additionally, a small bug relative to the tera icon not recalling preferences has been fixed.

Currently, fresh start presents the player with many options which are simply reset once the run starts.
This is an even bigger issue now since the nuzlocke mode will put much more attention on fresh start.

This is a starter select PR so hang on tight and enjoy the ride.

First of all, a new challenge type has been added. This challenge type is currently only used by fresh start. It takes the `dexData` and the `starterData` of a Pokémon, and removes basically all unlocks. It also takes a speciesId (necessary for forms).
Starter select was full of lines like `const starterData = globalScene.gameData.starterData[this.lastSpecies.speciesId]` and `const dexEntry = globalScene.gameData.dexData[species.speciesId]`. These have been replaced by calls to the new function `getSpeciesData`, which, in addition to returning a copy of the data, also applies the challenge modifier to it.
(In a few places, applying the challenge is not necessary, specifically when evaluating upgrades to buy).

The problem with this approach as pointed out by @Xavion3 is that we don't want to override the player's starter preferences every time that a fresh start challenge is opened. To solve this issue, we load an extra copy of `starterPreferences`, which does not get stripped down by the challenge, and is the one that we save at the end. However, we do want to save preference changes which are valid to fresh start---therefore, whenever the player toggles an option, the change is applied to both objects.
This is distasteful and atrocious, but it is a working solution.

The tera button was malfunctioning. It did save the preference when toggling the tera type. However, after loading the screen again, it did not pass the preference to `setSpeciesDetails` to be displayed.

Opening the starter select screen for a classic run, closing it, then opening it again for fresh start, would show some of the cost values colored in blue. This was because the code to change the color of the cost label was only processed when a difference between original and reduced cost was detected. This is now fixed.

It is very possible that `applyStarterModify` is redundant after this and can be removed. It is also possible that it causes some issues with pokémon with more than 4 moves at lv5, I have not tested this.

Rockruff was removed from the list of mons with restricted forms due to balance decision.

## Screenshots/Videos
<details><summary>Starter Select Screenshot</summary>
<p>

burmy

</p>
</details>
<details><summary>Video 1</summary>
<p>

https://github.com/user-attachments/assets/066fce85-e5bc-43f8-bb04-87ec4745c322

</p>
</details> 
<details><summary>Video 2</summary>
<p>

https://github.com/user-attachments/assets/4931de62-17b0-413f-935f-4ec829b8cf0b

</p>
</details>

## How to test the changes?
Open the starter select screen a few times in different conditions. Check that, when opening fresh starts and then going back to classic, the original preferences are preserved (unless e.g. going from ability 1 to ability 2 in fresh start).

## Checklist


Wlowscha:
There is still some weirdness going on with the tera symbol, plus changes to the moveset are not saved until a run is started. Both of these issues are also on main though, so probably have to be fixed separately.

EDIT: The issue with movesets is addressed in #6287 , while tera has been fixed so it also persists in preferences even after accessing the fresh start menu:

<p>

https://github.com/user-attachments/assets/b0752b75-dfc8-499c-89b7-3d7d6b5c2ecb

</p>
</details>

Wlowscha:
Another change in the same vein as #6287, this would make `saveStarterPreferences` be called at the end of `setSpeciesDetails`, so every time that the ability, nature, tera type etc is changed by the user. This again means that the changes will persist even if the page is reloaded for whatever reason. It also means a lot more calls to the function, which however only interacts with local data.

SirzBenjie:
> However, we do want to save preference changes which are valid to fresh start---therefore, whenever the player toggles an option, the change is applied to both objects.

Do we really? We can just not save preference changes from fresh start at all. I don't see an issue with this.

Wlowscha:
> > However, we do want to save preference changes which are valid to fresh start---therefore, whenever the player toggles an option, the change is applied to both objects.
> 
> Do we really? We can just not save preference changes from fresh start at all. I don't see an issue with this.

I think I already replied to this on discord, but the argument is:
With the upcoming nuzlocke mode, people will be playing fresh start a lot, and very possibly doing multiple runs with the same mons as they try to collect ribbons. I would find it very annoying if I had to reselect preferences for moves, ability etc every time.

Wlowscha:
Yet another small tera update which fixes a bug pointed out by @Jimmybald1 on discord, with the tera button resetting to the first type after switching ability, move order etc. 

<details><summary>Before (note how the tera button resets)</summary>
<p>

https://github.com/user-attachments/assets/a6dfdc1f-d269-4b61-b8d9-b7231d5e6e4a

</p>
</details> 
<p>

https://github.com/user-attachments/assets/00d23443-3726-4a91-bd93-6aed66960e77

</p>
</details>

src/data/challenge.ts
```diff
@@ -797,10 +809,60 @@ export class FreshStartChallenge extends Challenge {
     return true;
   }
 
+  applyStarterSelectModify(speciesId: SpeciesId, dexEntry: DexEntry, starterDataEntry: StarterDataEntry): boolean {
+    // Remove all egg moves
+    starterDataEntry.eggMoves = 0;
+    console.log("I AM APPLYING, ", starterDataEntry.eggMoves);
```
DayKev:
```suggestion
```

src/system/game-data.ts
```diff
@@ -2103,13 +2103,17 @@ export class GameData {
     };
   }
 
-  getStarterSpeciesDefaultAbilityIndex(species: PokemonSpecies): number {
-    const abilityAttr = this.starterData[species.speciesId].abilityAttr;
+  getStarterSpeciesDefaultAbilityIndex(species: PokemonSpecies, abilityAttr?: number): number {
+    if (isNullOrUndefined(abilityAttr)) {
+      abilityAttr = this.starterData[species.speciesId].abilityAttr;
+    }
```
DayKev:
```suggestion
    abilityAttr ??= this.starterData[species.speciesId].abilityAttr;
```
You should be able to simplify it like this. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment

src/system/game-data.ts
```diff
@@ -2103,13 +2103,17 @@ export class GameData {
...
     return abilityAttr & AbilityAttr.ABILITY_1 ? 0 : !species.ability2 || abilityAttr & AbilityAttr.ABILITY_2 ? 1 : 2;
   }
 
-  getSpeciesDefaultNature(species: PokemonSpecies): Nature {
-    const dexEntry = this.dexData[species.speciesId];
+  getSpeciesDefaultNature(species: PokemonSpecies, dexEntry?: DexEntry): Nature {
+    if (isNullOrUndefined(dexEntry)) {
+      dexEntry = this.dexData[species.speciesId];
+    }
```
DayKev:
```suggestion
    dexEntry ??= this.dexData[species.speciesId];
```

src/ui/starter-select-ui-handler.ts
```diff
@@ -1126,25 +1129,32 @@ export class StarterSelectUiHandler extends MessageUiHandler {
...
+
     if (args.length >= 1 && args[0] instanceof Function) {
       super.show(args);
       this.starterSelectCallback = args[0] as StarterSelectCallback;
 
       this.starterSelectContainer.setVisible(true);
 
+      this.starterPreferences = loadStarterPreferences();
+      this.originalStarterPreferences = loadStarterPreferences();
+      console.log("Loaded", this.originalStarterPreferences[1]);
```
DayKev:
```suggestion
```

src/ui/starter-select-ui-handler.ts
```diff
@@ -1154,6 +1164,7 @@ export class StarterSelectUiHandler extends MessageUiHandler {
 
         this.setUpgradeAnimation(icon, species);
       });
+      console.log("Initiated", this.originalStarterPreferences[1]);
```
DayKev:
```suggestion
```

src/ui/starter-select-ui-handler.ts
```diff
@@ -1262,6 +1277,17 @@ export class StarterSelectUiHandler extends MessageUiHandler {
...
 
+    if (starterAttributes.tera !== undefined) {
+      // If somehow we have an illegal tera type, it is reset here
+      if (!(starterAttributes.tera === species.type1 || starterAttributes.tera === species?.type2)) {
+        starterAttributes.tera = species.type1;
+      }
+      // In fresh start challenge, the tera type is always reset to the first one
+      if (globalScene.gameMode.hasChallenge(Challenges.FRESH_START) && !ignoreChallenge) {
+        starterAttributes.tera = species.type1;
+      }
```
DayKev, line 1304: Something like this is supposed to be done by an `applyChallenges()` call.

src/ui/starter-select-ui-handler.ts
```diff
@@ -1714,7 +1740,8 @@ export class StarterSelectUiHandler extends MessageUiHandler {
               globalScene.gameData.getSpeciesDexAttrProps(species, this.getCurrentDexProps(species.speciesId)),
               this.isPartyValid(),
             );
-            const isCaught = globalScene.gameData.dexData[species.speciesId].caughtAttr;
+            const { dexEntry } = this.getSpeciesData(species.speciesId);
+            const isCaught = dexEntry.caughtAttr;
```
DayKev:
```suggestion
            const isCaught = this.getSpeciesData(species.speciesId).dexEntry.caughtAttr;
```

src/ui/starter-select-ui-handler.ts
```diff
@@ -1781,9 +1808,11 @@ export class StarterSelectUiHandler extends MessageUiHandler {
       }
     } else {
       let starterContainer: StarterContainer;
-      const starterData = globalScene.gameData.starterData[this.lastSpecies.speciesId];
+      const { starterDataEntry } = this.getSpeciesData(this.lastSpecies.speciesId);
+      const starterData = starterDataEntry;
```
DayKev: Why do you have both `starterDataEntry` and `starterData` which are the same thing?

src/ui/starter-select-ui-handler.ts
```diff
@@ -1179,10 +1190,14 @@ export class StarterSelectUiHandler extends MessageUiHandler {
...
-    const dexEntry = globalScene.gameData.dexData[species.speciesId];
-    const starterData = globalScene.gameData.starterData[species.speciesId];
+  initStarterPrefs(
+    species: PokemonSpecies,
+    preferences: StarterPreferences,
+    ignoreChallenge = false,
+  ): StarterAttributes {
+    const starterAttributes = preferences[species.speciesId];
+    const { dexEntry, starterDataEntry } = this.getSpeciesData(species.speciesId, !ignoreChallenge);
+    const starterData = starterDataEntry;
```
DayKev:
```suggestion
    const { dexEntry, starterDataEntry: starterData } = this.getSpeciesData(species.speciesId, !ignoreChallenge);
```

src/ui/starter-select-ui-handler.ts
```diff
@@ -2994,8 +3037,9 @@ export class StarterSelectUiHandler extends MessageUiHandler {
       container.cost = globalScene.gameData.getSpeciesStarterValue(container.species.speciesId);
 
       // First, ensure you have the caught attributes for the species else default to bigint 0
-      const caughtAttr = globalScene.gameData.dexData[container.species.speciesId]?.caughtAttr || BigInt(0);
-      const starterData = globalScene.gameData.starterData[container.species.speciesId];
+      const { dexEntry, starterDataEntry } = this.getSpeciesData(container.species.speciesId);
+      const caughtAttr = dexEntry?.caughtAttr || BigInt(0);
+      const starterData = starterDataEntry;
```
DayKev:
```suggestion
      const { dexEntry, starterDataEntry: starterData } = this.getSpeciesData(container.species.speciesId);
      const caughtAttr = dexEntry?.caughtAttr ?? BigInt(0);
```

src/ui/starter-select-ui-handler.ts
```diff
@@ -3663,14 +3712,34 @@ export class StarterSelectUiHandler extends MessageUiHandler {
...
   setSpeciesDetails(species: PokemonSpecies, options: SpeciesDetails = {}): void {
     let { shiny, formIndex, female, variant, abilityIndex, natureIndex, teraType } = options;
     const forSeen: boolean = options.forSeen ?? false;
     const oldProps = species ? globalScene.gameData.getSpeciesDexAttrProps(species, this.dexAttrCursor) : null;
     const oldAbilityIndex =
       this.abilityCursor > -1 ? this.abilityCursor : globalScene.gameData.getStarterSpeciesDefaultAbilityIndex(species);
-    const oldNatureIndex =
-      this.natureCursor > -1 ? this.natureCursor : globalScene.gameData.getSpeciesDefaultNature(species);
+    let oldNatureIndex = -1;
+    if (species) {
```
DayKev, line 3759: ~What is the `if (species)` for? `species` is supposed to always be defined.~

Actually I wonder why there's checks for `species` existing all over this method...

Wlowscha: Because `species` _can_ be undefined actually, that happens when you move the cursor onto a filter. The function is called to remove the sprite and all the various icons.

src/ui/starter-select-ui-handler.ts
Wlowscha: While I completely agree in principle, the current logic is too convoluted for something like that. I would recommend keeping it like this for now, then figure out a better system once we refactor the whole starter select screen (which I'm very motivated to do now...)

src/ui/starter-select-ui-handler.ts
DayKev: That's... so the parameter is marked wrong in the method definition then.
And also this whole thing is jank af. Sigh.

Wlowscha:
After some discussion with balance, it seems like a good idea to cap the ivs at 15 (so if the player only has unlocked 10 atk, they will see 10) rather than setting them all at 15.

arceus_lol

SirzBenjie (dismissed):

src/data/challenge.ts
```diff
@@ -797,10 +809,61 @@ export class FreshStartChallenge extends Challenge {
...
+    // Removes shiny and variants
+    dexEntry.caughtAttr &= ~DexAttr.SHINY;
+    dexEntry.caughtAttr &= ~(DexAttr.VARIANT_2 | DexAttr.VARIANT_3);
+
+    // Remove unlocked forms for specific species
+    if (speciesId === SpeciesId.ZYGARDE) {
+      const formMask = (DexAttr.DEFAULT_FORM << 2n) - 1n; // Sets 10%-PC to 10%-AB and 50%-PC to 50%-AB
+      dexEntry.caughtAttr &= formMask;
+    }
+    if (
```
SirzBenjie:
```suggestion
      dexEntry.caughtAttr &= formMask;
    }
    else if (
```

src/data/challenge.ts
```diff
@@ -797,10 +809,61 @@ export class FreshStartChallenge extends Challenge {
...
+      dexEntry.ivs[i] = Math.min(dexEntry.ivs[i], 15);
+    }
+
+    // Removes shiny and variants
+    dexEntry.caughtAttr &= ~DexAttr.SHINY;
+    dexEntry.caughtAttr &= ~(DexAttr.VARIANT_2 | DexAttr.VARIANT_3);
+
+    // Remove unlocked forms for specific species
+    if (speciesId === SpeciesId.ZYGARDE) {
+      const formMask = (DexAttr.DEFAULT_FORM << 2n) - 1n; // Sets 10%-PC to 10%-AB and 50%-PC to 50%-AB
```
SirzBenjie:
```suggestion
      // Sets ability from power construct to aura break
      const formMask = (DexAttr.DEFAULT_FORM << 2n) - 1n;
```

src/data/challenge.ts
DayKev: Btw Biome is going to complain that the `else if` isn't on the same line as the preceding `}` lol.

Wlowscha:
Hiding completely the passive, the egg move boxes, as well as the passive and cost reduction options from the candy upgrade menu.

image

src/ui/starter-select-ui-handler.ts
```diff
@@ -4079,6 +4160,8 @@ export class StarterSelectUiHandler extends MessageUiHandler {
     this.tryUpdateValue();
 
     this.updateInstructions();
+
+    saveStarterPreferences(this.originalStarterPreferences);
```
SirzBenjie, line 4164: @Wlowscha 

What was the need to add this here?

PR #6296, opened 2025-08-19
[Enhancement] Evolutions and form changes do not add to seen pokémon count

Wlowscha:
Evolving mons or applying a form change does not add the pokémon or form to the seen data anymore.

Was requested by @damocleas .
~Note that evo/form change marking stuff as "seen" has been around for a long time, so more discussion should happen before we decide about the change, or whether this should be considered as a "bug fix".~
Considering that hatching egg also does not contribute to the "seen" stat, this is just a matter of consistency.

Removed two calls to `globalScene.gameData.setPokemonSeen` from `evolve` and `formChange` in the `PlayerPokémon` class.

Ther is another call to this in the `changeForm` method of the `Pokemon` class. My understanding is that this is only used by enemy pokemon, since player pokemon will use the overridden method mentioned above. It should therefore be fine to leave this call here.

Note that this function is called in 4 more places in the code:
- In `encounter-phase.ts`, which is where most seen data should be coming from
- in `encounter-phase-utils.ts` for mystery encounters
- in the `safari-zone-encounter` and the `fun-and-games-encounter` (the Wobbufett one), not sure why these are special cased.

## Checklist


damocleas:
Maybe we should use this PR itself to change "Species Seen" to "Species Fought" or "Species Encountered" since that is what it has turned in to over the past year+

damocleas:
This is Species Fought

Wlowscha:
> This is Species Fought

I don't think so

image

damocleas:
> > This is Species Fought
> 
> I don't think so
> 
> image

yeah that's what I meant I zoned out typing this, should do it in this PR since it's so tiny and just works

SirzBenjie (approved):

Wlowscha:
Made obsolete by #6315

PR #6302, opened 2025-08-20
[UI/UX] Add language selection to login screen

fabske0:

The user will be able to change the language in the login screen

- At the moment the only way to change the language before login is doing it manually via localstorage. (Doesn't even work on mobile)
- Saw someone ask in bug reports channel recently

- added an icon which opens the language selection menu
- moved the language change logic to its own file

## Screenshots/Videos

final_1

final_1_dropdown

## How to test the changes?

Set `VITE_BYPASS_LOGIN` in `.env.development` to `0` and press the icon

## Checklist

  - [x] If so, please leave a link to it here: https://github.com/pagefaultgames/pokerogue-locales/pull/230


src/system/settings/settings.ts
```diff
@@ -916,6 +916,7 @@ export function setSetting(setting: string, value: number): boolean {
             }
           };
           globalScene.ui.setOverlayMode(UiMode.OPTION_SELECT, {
+            // if a new language is added, it also needs to be added in "login-form-ui-handler.ts"
```
DayKev: Instead of duplicating the array, it should be moved into a constant that both files can use.

fabske0: Would you make a new file in `system/settings` for it, or where would you put it?

DayKev: For now you can put it in `src/system/settings/`. Maybe something like `language-list.ts`?

fabske0: Thought about `settings-language.ts`, since i also have to move the handler for the actual lang change not just the array.

DayKev: Sure.

DayKev (approved):

fabske0:
This is waiting for the icon

SirzBenjie (approved):
I tested this with the icon and it looks great!

Wlowscha (dismissed):

Madmadness65 (changes requested):
A duplicate of the icon is necessary in the `ui/legacy` folder.
image

SirzBenjie:
At some point we need to make legacy urls fallback to non-legacy if it doesn't exist...

PR #6317, opened 2025-08-21
[Dev] Ensure `i18n` module is initialized immediately when imported

Bertie690:
N/A
As I found out while debugging some code, we were initializing `i18next` NOT during the `initi18n` function call, but through a series of side effect imports culminating in tabulating names for every single trainer class' configs.

In light of this, I decided to make the side effect-ness explicit by having the module actually initialize itself when imported.

Removed the functions `getIsInitialized` and `initi18n` since (as previously discussed) they were never used as intended

Removed a duplicate korean post processor in the init code

Removed re-export of `i18next` from `#plugins/i18next` and fixed all uses thereof

## Screenshots/Videos
N/A
## How to test the changes?
Make sure locales still work?

## Checklist

### DIFF
--- a/src/plugins/i18n.ts
+++ b/src/plugins/i18n.ts
@@ -179,40 +179,40 @@ await i18next
         "es-419": ["es-ES", "en"],
         default: ["en"],
       },
-    supportedLngs: [
-      "en",
-      "es-ES",
-      "es-419", // LATAM Spanish
-      "fr",
-      "it",
-      "de",
-      "zh-Hans",
-      "zh-Hant",
-      "pt-BR",
-      "ko",
-      "ja",
-      "ca",
-      "da",
-      "tr",
-      "ro",
-      "ru",
-      "tl",
-      "nb-NO",
-    ],
-    backend: {
-      loadPath(lng: string, [ns]: string[]) {
-        // Use namespace maps where required
-        let fileName: string;
-        if (namespaceMap[ns]) {
-          fileName = namespaceMap[ns];
-        } else if (ns.startsWith("mysteryEncounters/")) {
-          fileName = toKebabCase(ns + "-dialogue"); // mystery-encounters/a-trainers-test-dialogue
-        } else {
-          fileName = toKebabCase(ns);
-        }
-        // ex: "./locales/en/move-anims"
-        return `./locales/${lng}/${fileName}.json?v=${pkg.version}`;
-        }
+      supportedLngs: [
+        "en",
+        "es-ES",
+        "es-419", // LATAM Spanish
+        "fr",
+        "it",
+        "de",
+        "zh-Hans",
+        "zh-Hant",
+        "pt-BR",
+        "ko",
+        "ja",
+        "ca",
+        "da",
+        "tr",
+        "ro",
+        "ru",
+        "tl",
+        "nb-NO",
+      ],
+      backend: {
+        loadPath(lng: string, [ns]: string[]) {
+          // Use namespace maps where required
+          let fileName: string;
+          if (namespaceMap[ns]) {
+            fileName = namespaceMap[ns];
+          } else if (ns.startsWith("mysteryEncounters/")) {
+            fileName = toKebabCase(ns + "-dialogue"); // mystery-encounters/a-trainers-test-dialogue
+          } else {
+            fileName = toKebabCase(ns);
+          }
+          // ex: "./locales/en/move-anims"
+          return `./locales/${lng}/${fileName}.json?v=${pkg.version}`;
+        },
       },
       defaultNS: "menu",
       detection: {


src/main.ts
```diff
@@ -49,7 +50,7 @@ let game;
 let manifest;
 
 const startGame = async () => {
-  await initI18n();
+  await import("#plugins/i18n");
```
Bertie690:
```suggestion
```

src/data/trainers/trainer-config.ts
```diff
@@ -164,13 +163,10 @@ export class TrainerConfig {
...
-      // First check if i18n is initialized
-      if (!getIsInitialized()) {
-        initI18n();
-      }
       // This is only the male name, because the female name is handled in a different function (setHasGenders)
-      if (name === "Finn") {
-        name = i18next.t("trainerNames:rival");
+      name = i18next.t("trainerNames:rival");
+      if (!name) {
+        throw new Error("skibidi sigma");
```
SirzBenjie: I can only assume this is a placeholder you meant to adjust later 😅

src/plugins/i18n.ts
```diff
@@ -136,193 +134,179 @@ function i18nMoneyFormatter(amount: any): string {
...
+        "mysteryEncounters/funAndGames",
+        "mysteryEncounters/uncommonBreed",
+        "mysteryEncounters/globalTradeSystem",
+        "mysteryEncounters/theExpertPokemonBreeder",
+        "mysteryEncounterMessages",
+      ],
+      detection: {
+        lookupLocalStorage: "prLang",
+      },
+      debug: Number(import.meta.env.VITE_I18N_DEBUG) === 1,
```
SirzBenjie:
```suggestion
      debug: +(import.meta.env.VITE_I18N_DEBUG) === 1,
```
didn't check if biome wants this to not have parenthesis, but we should never use Number(...). Just use `+`

src/plugins/i18n.ts
```diff
@@ -136,193 +134,179 @@ function i18nMoneyFormatter(amount: any): string {
...
+ *    Each language folder should contain a file for each namespace (ex. menu.ts) with the translations.
+ *    Don't forget to declare new language in `supportedLngs` i18next initializer
+ *
+ * Q: How do I add a new namespace?
+ * A: To add a new namespace, create a new file in each language folder with the translations.
+ *    Then update the config file for that language in its locale directory
+ *    and the CustomTypeOptions interface in the @types/i18next.d.ts file.
+ *
+ * Q: How do I make a language selectable in the settings?
+ * A: In src/system/settings.ts, add a new case to the Setting.Language switch statement.
```
SirzBenjie, line 169: Does it make sense for this to be an `@module` comment maybe?

SirzBenjie (commented):
I will say, I don't love the idea that there are side effects of an import itself.
This is generally considered to be quite bad practice.
I worry that this would prohibit tree shaking for other code that ends up importing `i18n`. I haven't investigated.

src/data/trainers/trainer-config.ts
Bertie690: 💀

src/plugins/i18n.ts
Bertie690: 🤷

src/plugins/i18n.ts
DayKev: The conversion isn't necessary anyway, just do `=== "1"`.

src/init/init.ts
```diff
@@ -14,7 +14,9 @@ import { initAchievements } from "#system/achv";
 import { initVouchers } from "#system/voucher";
 import { initStatsKeys } from "#ui/game-stats-ui-handler";
 
-/** Initialize the game. */
+/**
+ * Initialize various values at the beginning of each testing instance.
```
DayKev: This isn't a test function lol.

src/main.ts
```diff
@@ -1,8 +1,9 @@
 import "#app/polyfills";
 // All polyfills MUST be loaded first for side effects
+import "#app/plugins/i18n";
+// Initializes i18n on import
```
DayKev:
```suggestion
// All polyfills MUST be loaded first for side effects
import "#app/polyfills";
// Initializes i18n on import
import "#app/plugins/i18n";
```

src/init/init.ts
Bertie690: oops

src/plugins/i18n.ts
Bertie690:
```suggestion
      debug: import.meta.env.VITE_I18N_DEBUG === "1",
```

DayKev (dismissed):

src/plugins/i18n.ts
```diff
@@ -136,193 +134,180 @@ function i18nMoneyFormatter(amount: any): string {
   return `@[MONEY]{${i18next.t("common:money", { amount })}}`;
 }
 
-//#region Exports
+//#region Initialization
 
 /**
- * Initialize i18n with fonts
+ * @module
```
DayKev:
```suggestion
```
`@module` only applies if it's the very first comment in the file.

src/plugins/i18n.ts
```diff
@@ -136,193 +134,180 @@ function i18nMoneyFormatter(amount: any): string {
   return `@[MONEY]{${i18next.t("common:money", { amount })}}`;
 }
 
-//#region Exports
+//#region Initialization
 
 /**
```
DayKev:
```suggestion
/*
```

src/plugins/i18n.ts
Bertie690: ugh

DayKev (approved):

DayKev (approved):

SirzBenjie (approved):

SirzBenjie:
I still don't like that we have side effects for an import, but oh well.

src/init/init.ts
```diff
@@ -1,3 +1,4 @@
+import "#plugins/i18n"; // Initializes i18n on import
```
DayKev: Why is this in both `init.ts` and `main.ts`?

src/plugins/i18n.ts
```diff
@@ -123,106 +121,93 @@ function i18nMoneyFormatter(amount: any): string {
...
-    detection: {
-      lookupLocalStorage: "prLang",
-    },
-    debug: Number(import.meta.env.VITE_I18N_DEBUG) === 1,
-    interpolation: {
-      escapeValue: false,
+    async () => {
+      if (i18next.services.formatter) {
+        i18next.services.formatter.add("money", i18nMoneyFormatter);
+      }
```
DayKev:
```suggestion
      i18next.services.formatter?.add("money", i18nMoneyFormatter);
```

src/plugins/i18n.ts
```diff
@@ -123,106 +121,93 @@ function i18nMoneyFormatter(amount: any): string {
...
-  await initFonts(localStorage.getItem("prLang") ?? undefined);
-}
-
-export function getIsInitialized(): boolean {
-  return isInitialized;
-}
+  );
 
-// biome-ignore lint/style/noDefaultExport: necessary for i18next usage
 export default i18next;
```
DayKev: This isn't imported from this file anymore so the re-export is no longer necessary.

src/init/init.ts
Bertie690: Oops should be in the fomer only

src/plugins/i18n.ts
Bertie690: Actually, it is.

It's on my bucket list to regex it away

DayKev: Well, might as well remove any you missed then and remove the re-export completely, since you already removed most of them.

DayKev:
The initialization probably needs to be moved from `init.ts` to `main.ts`.

DayKev:
> The initialization probably needs to be moved from `init.ts` to `main.ts`.

Yeah, this fixed the rival name issue.

DayKev:
Oh the game works but the tests are failing... this is silly.

DayKev (approved):

PR #6325, opened 2025-08-21
Moved Form change items to a separate object

Bertie690:
Better type safety and stuffs.

Currently not going too hard on it bc i am lazy


src/data/pokemon-forms/form-change-triggers.ts
```diff
@@ -78,6 +78,7 @@ export class SpeciesFormChangeItemTrigger extends SpeciesFormChangeTrigger {
   }
 
   canChange(pokemon: Pokemon): boolean {
+    // TODO: can we fix this?
```
Wlowscha, line 81: What is "this" that needs to be fixed?

Bertie690: The fact that it accesses a property that should by all means be private

PR #6338, opened 2025-08-22
[Test] Improved typing on `BattlerTagData` and `ArenaTagData`; improved test matchers

Bertie690:
- **Improved typing on `BattlerTagData` and `ArenaTagData`**
- **Added dragon cheer/focus energy tests**

Dragon Cheer and Focus Energy will now fail if the other one's effect is present on the target.
I was augmenting the `toHaveBattlerTag` test matcher for Dragon Cheer tests when I noticed you could query stuff like `lapseTypes` that are static across all instances and will never change ever.

While I was at it, I noticed some of the arena and battler tag types could be cleaned up.
Cleaned up the types in `#types/arena-tags` and `#types/battler-tags` to do the same stuff as before but using existing type helpers rather than re-inventing the wheel.

Added a new exported type `ArenaTagDataMap` and battler tag equivalent that maps each serialized tag type to the parameter type of their `loadTag` function. This is used in the matchers to ensure only serializable properties (read: different across instances) can be checked by callers.

Improved `toHaveBattlerTag` to accept a partially-filled serialized `BattlerTag` similar to its `ArenaTag` counterpart

Added tests for dragon cheer/focus energy to show off utils working

ended up fixing focus energy bug while making tests because dammit
## Screenshots/Videos
N/A
## How to test the changes?
Go to line `88:41` of `dragon-cheer.test.ts` and open intellisense.

Notice that `lapseTypes` (a non-serialized property) does not appear in intellisense
## Checklist

### DIFF
--- a/src/@types/arena-tags.ts
+++ b/src/@types/arena-tags.ts
@@ -2,6 +2,7 @@ import type { ArenaTagTypeMap } from "#data/arena-tag";
 import type { ArenaTagType } from "#enums/arena-tag-type";
 // biome-ignore lint/correctness/noUnusedImports: TSDocs
 import type { SessionSaveData } from "#system/game-data";
+import type { ObjectValues } from "#types/type-helpers";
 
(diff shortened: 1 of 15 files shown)


SirzBenjie (approved):
I'm going to bypass merge this despite it not being reviewed by anyone other than me, because there are literally 0 changes to client side code (it's all typescript stuff and adding tests).

PR #6340, opened 2025-08-22
[Documentation][Move] Add edge case to helping hand

Wlowscha:
Requested by @Bertie690

### DIFF
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -9413,6 +9413,8 @@ export function initMoves() {
       .ignoresSubstitute()
       .target(MoveTarget.NEAR_ALLY)
       .condition(failIfSingleBattle),
+      // should stack multiplicatively if used multiple times in 1 turn`
+      .edgeCase(),
     new StatusMove(MoveId.TRICK, PokemonType.PSYCHIC, 100, 10, -1, 0, 3)
       .unimplemented(),
     new StatusMove(MoveId.ROLE_PLAY, PokemonType.PSYCHIC, -1, 10, -1, 0, 3)


Bertie690 (dismissed):

src/data/moves/move.ts
```diff
@@ -9412,7 +9412,9 @@ export function initMoves() {
       .attr(AddBattlerTagAttr, BattlerTagType.HELPING_HAND)
       .ignoresSubstitute()
       .target(MoveTarget.NEAR_ALLY)
-      .condition(failIfSingleBattle),
+      .condition(failIfSingleBattle)
+      // should stack multiplicatively if used multiple times in 1 turn`
```
```suggestion
      // should stack multiplicatively if used multiple times in 1 turn
```

Bertie690 (changes requested):

Bertie690 (approved):

src/data/moves/move.ts
```diff
@@ -9412,7 +9412,9 @@ export function initMoves() {
       .attr(AddBattlerTagAttr, BattlerTagType.HELPING_HAND)
       .ignoresSubstitute()
       .target(MoveTarget.NEAR_ALLY)
-      .condition(failIfSingleBattle),
+      .condition(failIfSingleBattle)
+      // should stack multiplicatively if used multiple times in 1 turn
```
SirzBenjie, line 9434: Wait what
How can this happen? Instruct??
@Bertie690

DayKev: I think it would require Instruct, because we don't have triple battles. So it basically doesn't matter unless we add triple battles (which probably won't happen for many years if ever).

Bertie690: Correct
Instruct, triple battles or both

DayKev: So yeah, something that won't ever happen.

🤷 if you want to add a pointless edge case comment go ahead.

PR #6344, opened 2025-08-23
[Challenge][Bug] Available starters in fresh start monotype

Wlowscha:
For example, Charmander will show up as a possible option in mono-flying fresh start.

The challenge code which determines the validity of each starter does so by looping through its evolutions, then checking whether any of them are valid. However, the full fresh start challenge only checked the species being among the default starters, automatically excluding evolutions.

This is solved by creating a new array which contains all default species ids and evolutions, and checking against it.

## Screenshots/Videos
image

## How to test the changes?
Start a run with fresh start and a mono type challenge.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "pokemon-rogue-battle",
   "private": true,
-  "version": "1.10.0",
+  "version": "1.10.1",
   "type": "module",
   "scripts": {
     "start": "vite",


src/constants.ts
```diff
@@ -53,6 +53,12 @@ export const defaultStarterSpecies: SpeciesId[] = [
   SpeciesId.QUAXLY,
 ];
 
+export const defaultStarterSpeciesAndEvolutions: SpeciesId[] = defaultStarterSpecies.flatMap(id => [
+  id,
+  (id + 1) as SpeciesId,
+  (id + 2) as SpeciesId,
+]);
```
SirzBenjie, line 60: Note this assumes that default starters will always have 2 evolutions.

Will this work for Hisuian evolutions?

DayKev: Uh, hm...

DayKev: I don't think so?

damocleas: this can be considered a partial fix until tomorrow

PR #6345, opened 2025-08-23
[Item] Held item implementation rework

Wlowscha:
Reworking HeldItem to have stronger typing on effects, parameters etc

Implementation is @Bertie690 's idea, I'm only doing the dirty work.


src/items/held-item-parameter.ts
```diff
@@ -0,0 +1,215 @@
...
+  pokemon: Pokemon;
+  /** The amount of exp to gain */
+  baseStats: number[];
+}
+
+export interface BaseStatTotalParams {
+  /** The pokemon with the item */
+  pokemon: Pokemon;
+  /** Array of the pokemon's base stat; modified in place after item application */
+  baseStats: number[];
```
Bertie690, line 48: Can this be a 6 length tuple of numbers?
we really should make pokemon stats a tuple loooool

src/items/held-item-parameter.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+export interface BatonParams {
+  /** The pokemon with the item */
```
Bertie690: Can we have all these "pkmn and nothing else) like merged into 1 big "itemBaseParams"

src/items/held-item-parameter.ts
```diff
@@ -0,0 +1,215 @@
...
+export interface InstantReviveParams {
+  /** The pokemon with the item */
+  pokemon: Pokemon;
+}
+
+export interface ItemStealParams {
+  /** The pokemon with the item */
+  pokemon: Pokemon;
+  /** The pokemon to steal from (optional) */
+  // TODO: https://github.com/pagefaultgames/pokerogue/pull/5656#discussion_r2135607083
```
Bertie690, line 124: It is optional because NO

src/items/held-item-parameter.ts
```diff
@@ -0,0 +1,215 @@
...
+  target?: Pokemon;
+}
+
+export interface MultiHitParams {
+  /** The pokemon with the item */
+  pokemon: Pokemon;
+  /** The move being used */
+  moveId: MoveId;
+  /** Holder for the move's hit count for the turn */
+  count?: NumberHolder;
```
Bertie690: Idea: make this a union type - Either the hit count OR the damage will be passed at any given time, but NEVER BOTH
(also pls add that to the description thxxx)

src/items/held-item-parameter.ts
```diff
@@ -0,0 +1,215 @@
...
+  pokemon: Pokemon;
+  /** Holder for the multiplier */
+  multiplier: NumberHolder;
+}
+
+export interface ResetNegativeStatStageParams {
+  /** The pokemon with the item */
+  pokemon: Pokemon;
+  /** Whether the move was used by a player pokemon */
+  isPlayer: boolean;
```
Bertie690: Useless you can check is player

src/items/held-item-parameter.ts
```diff
@@ -0,0 +1,215 @@
...
+  /** The pokemon with the item */
+  pokemon: Pokemon;
+  stat: Stat;
+  statValue: NumberHolder;
+}
+
+export interface SurviveChanceParams {
+  /** The pokemon with the item */
+  pokemon: Pokemon;
+  surviveDamage: BooleanHolder;
```
Bertie690, line 168: Missing a comment

src/items/held-item-parameter.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+export interface TurnEndHealParams {
+  /** The pokemon with the item */
+  pokemon: Pokemon;
+}
+
+export interface TurnEndStatusParams {
```
Bertie690: No bcivhfjvjdjelendhfmsjdhfjdekrj

src/items/held-item-parameter.ts
```diff
@@ -0,0 +1,215 @@
...
+  [HeldItemEffect.DAMAGE_MONEY_REWARD]: DamageMoneyRewardParams;
+  [HeldItemEffect.BATON]: BatonParams;
+  [HeldItemEffect.CONTACT_ITEM_STEAL_CHANCE]: ItemStealParams;
+  [HeldItemEffect.TURN_END_ITEM_STEAL]: ItemStealParams;
+  [HeldItemEffect.BASE_STAT_TOTAL]: BaseStatTotalParams;
+  [HeldItemEffect.BASE_STAT_FLAT]: BaseStatFlatParams;
+  [HeldItemEffect.INCREMENTING_STAT]: IncrementingStatParams;
+  [HeldItemEffect.EVO_TRACKER]: EvoTrackerParams;
+};
+declare const EnsureAllEffectsAreMapped: HeldItemEffectParamMap[HeldItemEffect] & never;
```
Bertie690, line 211:
```suggestion
/** 
 * Dummy, Typescript-only constant to ensure that all {@linkcode HeldItemEffect}s have an entry in {@linkcode HeldItemEffectParamMap}.
 * If any are missing, Typescript will throw an error on this line.
 * @remarks
 * ⚠️ Does not exist at runtime, so it must not be used!
 */
declare const EnsureAllEffectsAreMapped: HeldItemEffectParamMap[HeldItemEffect] & never;
```

src/items/held-item.ts
```diff
@@ -122,15 +100,61 @@ export class HeldItem {
...
+    return this;
+  }
+}
+
+/** Tuple type containing >1 `HeldItemEffect`s */
+export type EffectTuple = Readonly<[HeldItemEffect, ...HeldItemEffect[]]>;
+
+export abstract class HeldItem<T extends EffectTuple> extends HeldItemBase {
+  /**
+   * A readonly tuple containing {@linkcode HeldItemEffect | effects} that this class can apply.
```
Bertie690:
```suggestion
   * A readonly tuple containing all {@linkcode HeldItemEffect | effects} that this class can apply.
```

src/items/held-item.ts
```diff
@@ -122,15 +100,61 @@ export class HeldItem {
...
+  unsuppressable(): this {
+    this.isSuppressable = false;
+    return this;
+  }
+}
+
+/** Tuple type containing >1 `HeldItemEffect`s */
+export type EffectTuple = Readonly<[HeldItemEffect, ...HeldItemEffect[]]>;
+
+export abstract class HeldItem<T extends EffectTuple> extends HeldItemBase {
```
Bertie690, line 126:
```suggestion
/**
 * Abstract class for all non-cosmetic held items (i.e. ones that can have their effects applied).
 */
export abstract class HeldItem<T extends EffectTuple> extends HeldItemBase {
```

src/items/held-item.ts
```diff
@@ -122,15 +100,61 @@ export class HeldItem {
...
+   * Check whether a given effect of this item should apply.
+   * @typeParam E - The type of one of this class' {@linkcode effects}
+   * @param effect - The {@linkcode HeldItemEffect | effect} being applied
+   * @param args - Arguments required for the effect application
+   * @returns Whether the effect should apply.
+   */
+  shouldApply<E extends this["effects"][number]>(_effect: E, _args: HeldItemEffectParamMap[E]): boolean {
+    return true;
+  }
+  /**
```
Bertie690, line 145:
```suggestion

  /**
```

src/items/held-item-parameter.ts
Wlowscha: NO?

src/items/held-item-parameter.ts
Wlowscha: Okay

src/items/held-item-parameter.ts
Wlowscha: Probably best to do that in a separate PR since there's likely a bunch of places that would benefit from it

src/items/held-item-parameter.ts
Bertie690: It should not be optional is what I am getting at

src/items/held-item-parameter.ts
Bertie690: Sorry i meant "collapse these into the core `HeldItemParams` interface"

src/items/held-item-parameter.ts
Wlowscha: Actually isn't this a case where we can have two different effects on the item?

src/items/held-item-parameter.ts
Wlowscha: It's this way because MBH picks a target at random iirc

Wlowscha: We can change that when we review the implementation for stealing effects

src/items/held-item-parameter.ts
```diff
@@ -0,0 +1,211 @@
...
+export interface BaseStatTotalParams {
+  /** The pokemon with the item */
+  pokemon: Pokemon;
+  /** Array of the pokemon's base stat; modified in place after item application */
+  baseStats: number[];
+}
+
+export type BatonParams = DefaultHeldItemParams;
+
+export type BerryParams = DefaultHeldItemParams;
```
Bertie690, line 53: If we really want to be pedantic, these should probably be interfaces
Can be a separate pr thi

src/items/held-items/accuracy-booster.ts
```diff
@@ -25,22 +16,10 @@ export class AccuracyBoosterHeldItem extends HeldItem {
   }
 
   /**
-   * Checks if {@linkcode PokemonMoveAccuracyBoosterHeldItem} should be applied
-   * @param pokemon - The {@linkcode Pokemon} to apply the move accuracy boost to
-   * @param moveAccuracy - {@linkcode NumberHolder} holding the move accuracy boost
-   * @returns `true` if {@linkcode PokemonMoveAccuracyBoosterHeldItem} should be applied
+   * Applies this item
```
Bertie690: Zero value added comment (in fact it actually detracts by overriding the superclass' comment in IDE highlighting)

src/items/held-item.ts
```diff
@@ -122,15 +100,65 @@ export class HeldItem {
...
+   * @param args - Arguments required for the effect application
+   */
+  abstract apply<E extends this["effects"][number]>(effect: E, param: HeldItemEffectParamMap[E]): void;
 }
 
-export class ConsumableHeldItem extends HeldItem {
+export abstract class ConsumableHeldItem<T extends EffectTuple> extends HeldItem<T> {
   // Sometimes berries are not eaten, some stuff may not proc unburden...
-  consume(pokemon: Pokemon, isPlayer: boolean, remove = true, unburden = true): void {
+  consume(pokemon: Pokemon, remove = true, unburden = true): void {
```
Bertie690:
```suggestion
  /**
   * Consume this item and apply relevant effects.
   * Should be overridden by any subclasses with their own on-consume effects.
   * @param pokemon - The {@linkcode Pokemon} consuming the item
   * @param remove - Whether to remove the item during consumption; default `true`
   * @param unburden - Whether to trigger item loss abilities (i.e. Unburden)  when consuming the item; default `true`
   */
  public consume(pokemon: Pokemon, remove = true, unburden = true): void {
```

IDRC if we actually need the unburden parameter, but that is for another day and another time

src/items/held-item.ts
```diff
@@ -122,15 +100,65 @@ export class HeldItem {
...
+   * @typeParam E - The type of one of this class' {@linkcode effects}
+   * @param effect - The {@linkcode HeldItemEffect | effect} being applied
+   * @param args - Arguments required for the effect application
+   */
+  abstract apply<E extends this["effects"][number]>(effect: E, param: HeldItemEffectParamMap[E]): void;
 }
 
-export class ConsumableHeldItem extends HeldItem {
+export abstract class ConsumableHeldItem<T extends EffectTuple> extends HeldItem<T> {
   // Sometimes berries are not eaten, some stuff may not proc unburden...
```
Bertie690: Requesting this line eb deleted

src/items/held-item.ts
```diff
@@ -122,15 +100,65 @@ export class HeldItem {
...
+   * Called if and only if {@linkcode shouldApply} returns `true`
+   * @typeParam E - The type of one of this class' {@linkcode effects}
+   * @param effect - The {@linkcode HeldItemEffect | effect} being applied
+   * @param args - Arguments required for the effect application
+   */
+  abstract apply<E extends this["effects"][number]>(effect: E, param: HeldItemEffectParamMap[E]): void;
 }
 
-export class ConsumableHeldItem extends HeldItem {
+export abstract class ConsumableHeldItem<T extends EffectTuple> extends HeldItem<T> {
```
Bertie690, line 156:
```suggestion
/** Abstract class for all `HeldItem`s that can be consumed during battle. */
export abstract class ConsumableHeldItem<T extends EffectTuple> extends HeldItem<T> {
```

Actually curbio how does applying consumable items work?
Would we want to bake it into the `apply` call?

src/utils/common.ts
```diff
@@ -570,3 +570,18 @@ export function pickWeightedIndex(weights: number[]): number | undefined {
 
   return undefined; // TODO: Change to something more appropriate
 }
+
+type InArray<T, X> = T extends readonly [X, ...infer _Rest]
```
Bertie690, line 579:
```suggestion
/** 
 * Type helper to check if a given item is in a tuple, returning `true` or `false` as appropriate.
 * @typeParam T - The tuple to check
 * @param X - The item whose inclusion is being checked
 */
type InArray<T, X> = T extends readonly [X, ...infer _Rest]
```

src/utils/common.ts
```diff
@@ -570,3 +570,18 @@ export function pickWeightedIndex(weights: number[]): number | undefined {
...
+  ? true
+  : T extends readonly [X]
+    ? true
+    : T extends readonly [infer _, ...infer Rest]
+      ? InArray<Rest, X>
+      : false;
+
+/** Type helper to allow only unique elements in a tuple */
```
Bertie690:
```suggestion
/** 
 * Type helper to allow only unique elements in a tuple (effectively converting it to a Set).
 * Within it, any duplicate elements will be flagged and converted to an error message.
 * @typeParam T - The tuple to render unique
 */
```

Bertie690 (changes requested):

src/items/held-item.ts
Wlowscha: I put `consume` in the abstract class and then the subclasses can call it as part of their apply, because they need different parameters iirc

Bertie690:
Lgtm

Bertie690 (approved):

PR #6402, opened 2025-08-24
[Bug] [Ability] Fix berserk multi proc

### LINKED ISSUE #6364: [Bug] Berserk activating twice when put under half health by Parental Bond / Multi Lens
### Describe the bug

When hitting a Pokemon with the ability/passive Berserk with a Pokemon that has either the ability Parental Bond or is holding the Multi Lens, if the first hit puts the Pokemon's HP under half and activates Berserk, the additional attack from Parental Bond / Multi Lens will activate Berserk again allowing the Pokemon with the ability to reach +2 Special Attack in a single turn or in the case of Multi Lens, +3.

### Reproduction

Have a Pokemon with either Parental Bond as a passive / ability or holding the Multi Lens deal damage to an enemy with Berserk as a passive / ability.

### Expected behavior

The Pokemon with Berserk should only receive one special attack boost

### Screenshots / Videos

_No response_

### Session export file

_No response_

### User data export file

_No response_

### Additional context

Multi Hit moves seemingly don't have this issue

SirzBenjie:
Abilities that have an activation gated by dropping below some HP threshold (anger shell, berserk, etc), will no longer sometimes trigger multiple times when hit by a multi hit move that drops them below half hp (including parental bond / multi-lens).

Fixes #6364 

Added a new parameter to `PostMoveInteractionAbAttrParams` that carries the amount of damage dealt in the interaction.
Then, adds tracking for this information to move effect phase.

Previously, the code was checking the last attack received, and then looking at its damage, which clearly did not always work.

## Screenshots/Videos

https://github.com/user-attachments/assets/ff345c7a-dfbc-4a49-b604-54559d441925
</details> 

## How to test the changes?
```ts
const overrides = {
  ABILITY_OVERRIDE: AbilityId.PARENTAL_BOND,
  MOVESET_OVERRIDE: MoveId.EMBER,
  ENEMY_MOVESET_OVERRIDE: [MoveId.SPLASH],
  ENEMY_SPECIES_OVERRIDE: SpeciesId.GLOOM,
  ENEMY_LEVEL_OVERRIDE: 5,
  ENEMY_ABILITY_OVERRIDE: AbilityId.BERSERK,
```

Use this set of overrides, and choose charmander WITHOUT its passive or a boosted nature. This will ensure that the first hit from parental bond drops the target below half hp.

## Checklist
- [x] **I'm using `hotfix-1.10.3` as my base branch**

### DIFF
--- a/src/phases/move-effect-phase.ts
+++ b/src/phases/move-effect-phase.ts
@@ -402,8 +402,14 @@ export class MoveEffectPhase extends PokemonPhase {
    * @param hitResult - The {@linkcode HitResult} of the attempted move
    * @param wasCritical - `true` if the move was a critical hit
    */
-  protected applyOnGetHitAbEffects(user: Pokemon, target: Pokemon, hitResult: HitResult, wasCritical = false): void {
-    const params = { pokemon: target, opponent: user, move: this.move, hitResult };
+  protected applyOnGetHitAbEffects(
+    user: Pokemon,
+    target: Pokemon,
+    hitResult: HitResult,
+    damage: number,
+    wasCritical = false,
+  ): void {
+    const params = { pokemon: target, opponent: user, move: this.move, hitResult, damage };
     applyAbAttrs("PostDefendAbAttr", params);
 
     if (wasCritical) {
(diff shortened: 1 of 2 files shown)


src/phases/move-effect-phase.ts
```diff
@@ -850,7 +860,7 @@ export class MoveEffectPhase extends PokemonPhase {
     }
 
     if (damage <= 0) {
-      return [result, isCritical];
+      return [result, isCritical, dmg];
```
DayKev:
```suggestion
      return [result, isCritical, damage];
```
Pretty sure everything after the "damage" variable is set should use it and not the pre-calculation "dmg" variable.

src/phases/move-effect-phase.ts
```diff
@@ -879,7 +889,7 @@ export class MoveEffectPhase extends PokemonPhase {
       globalScene.applyModifiers(DamageMoneyRewardModifier, true, user, new NumberHolder(damage));
     }
 
-    return [result, isCritical];
+    return [result, isCritical, dmg];
```
DayKev:
```suggestion
    return [result, isCritical, damage];
```

src/phases/move-effect-phase.ts
SirzBenjie: thanks, sloppy oversight

DayKev: Well it's not the clearest wrt naming... "dmg" and "damage" lol.

DayKev (approved):

src/phases/move-effect-phase.ts
DayKev: I think in the future we should probably change "dmg" to "rawDamage" and "damage" to "finalDamage".

PR #6404, opened 2025-08-25
[UI] Add cyclic navigation in settings menu for LEFT/RIGHT buttons

isaac105:
Provides a more comfortable left/right movement experience in the game settings menu.

When selecting game items, I found it easy to move left and right, but I discovered that this feature was not implemented in the game settings window.

While adding this feature, we changed the existing if statement to a ternary operator.

## Screenshots/Videos
https://github.com/user-attachments/assets/f90175d8-82a3-481e-babc-44ff34215a8c

## How to test the changes?
Just go to the game settings menu and adjust the direction keys.

## Checklist


SirzBenjie (approved):
Oh this feature is nice. Thanks!
Especially nice since you implemented something in our mazelike UI code!

@pagefaultgames/junior-dev-team 
I think this is suitable to go into the mid-version update?

src/ui/settings/abstract-settings-ui-handler.ts
```diff
@@ -318,16 +318,20 @@ export class AbstractSettingsUiHandler extends MessageUiHandler {
...
           break;
         case Button.RIGHT:
-          // Moves the option cursor right, if possible.
-          if (this.optionCursors[cursor] < this.optionValueLabels[cursor].length - 1) {
-            success = this.setOptionCursor(cursor, this.optionCursors[cursor] + 1, true);
-          }
+          // Cycle to the leftmost position when at the rightmost, otherwise move right
+          success = this.setOptionCursor(
+            cursor,
+            this.optionCursors[cursor] < this.optionValueLabels[cursor].length - 1 ? this.optionCursors[cursor] + 1 : 0,
```
SirzBenjie: You can also do `Phaser.Math.Wrap(this.optionCursors[cursor] + 1, 0, this.optionValueLabels[cursor].length - 1)` if that seems more readable. I'm not sure I have a preference myself.

isaac105: Thank you for your advice.
I was thinking of keeping the existing logic as much as possible, but I think your suggestion is better.
I have made the changes and push it.

DayKev (commented):
It works, but it might be a good idea to not have the wrapping behavior for the volume settings (accidentally going from "muted" to "100%" volume could be problematic).

Bertie690 (approved):
lgtm

isaac105:
> It works, but it might be a good idea to not have the wrapping behavior for the volume settings (accidentally going from "muted" to "100%" volume could be problematic).

@DayKev 
Thanks for reply!
I believe this is a matter of value judgment.
Should we add more code to prepare for this unlikely situation, or should we leave it as is and ensure the readability of the code?
I would like to choose the latter for the following two reasons.
1. The user experience when this issue is left unaddressed is merely “surprise,” which users can quickly adjust to on their own.
If it were something that would cause users difficulty in resolving on their own, then modification would certainly be necessary, but this does not seem to be the case.  
2. From the outset, users who play with the sound muted are in a situation where they have no choice but to do so, and they will never touch the volume settings.  
Even a 10-year-old child would not touch the volume if they were afraid of being caught by their mother. haha 😁

DayKev (approved):

SirzBenjie:
> @DayKev Thanks for reply! I believe this is a matter of value judgment. Should we add more code to prepare for this unlikely situation, or should we leave it as is and ensure the readability of the code? I would like to choose the latter for the following two reasons.
> 
> 1. The user experience when this issue is left unaddressed is merely “surprise,” which users can quickly adjust to on their own.
>    If it were something that would cause users difficulty in resolving on their own, then modification would certainly be necessary, but this does not seem to be the case.
> 2. From the outset, users who play with the sound muted are in a situation where they have no choice but to do so, and they will never touch the volume settings.
>    Even a 10-year-old child would not touch the volume if they were afraid of being caught by their mother. haha 😁

I like this reasoning ;)

SirzBenjie (approved):

Bertie690 (approved):
> I like this reasoning ;)

Thirded

DayKev:
> I like this reasoning ;)

I do not. There's no volume slider in any game that lets you wrap around from 0% to 100%, that's bad UX.

> From the outset, users who play with the sound muted are in a situation where they have no choice but to do so, and they will never touch the volume settings.

This is just completely false. There is more than exactly one single reason someone would have the volume muted, temporarily or otherwise.

> The user experience when this issue is left unaddressed is merely “surprise,” which users can quickly adjust to on their own.

If someone is wearing headphones or w/e and we suddenly start blasting incredibly loud sounds at them, that's bad.

PR #6409, opened 2025-08-25
[Balance][ME]Tweak weird dream do a party heal in option 1, random tera, and no longer revive pokemon in hardcore

SirzBenjie:
Weird dream will now do a party heal when selecting the first option, and will not 

Requested by balance team.
Directing it to the hotfix-1.10.4 branch, because the fact that it could revive fainted Pokémon in hardcore is considered a balance bug.

Track the status and hp ratio of the previous Pokémon before transforming it into a new one. After transformation, set its HP ratio to the new one, and carry over its status. If it was fainted, set it to faint.
Note that I add a catch for comatose, since I believe that bad things can happen if comatose somehow ends up with a status, and I didn't want to take that chance.

Also adds a new constant for `MAX_POKEMON_TYPE` and use it to set a random tera type for the transformed Pokémon.

Note: We may want to get rid of the shop items in the select-modifier phase that comes after the heal, this is not done here. (note we _do_ want the party heal to take place before the select modifier phase, so the player is aware of this when choosing the modifier).

## Screenshots/Videos

<details><summary>Non-Hardcore weird dream</summary>

https://github.com/user-attachments/assets/374f2200-95ee-430b-8f68-8814d6c360a5
</details>

## How to test the changes?
```ts
const overrides = {
  MYSTERY_ENCOUNTER_OVERRIDE: MysteryEncounterType.WEIRD_DREAM,
  STARTING_WAVE_OVERRIDE: 21,
  STARTING_LEVEL_OVERRIDE: 13,
  STARTING_MODIFIER_OVERRIDE: [{
    name: "TERA_ORB"
  }]
```

Make sure to take multiple pokemon, and ensure the user takes damage on the first wave so you can see the heal phase work successfully.

## Checklist
- [x] **I'm using `hotfix-1.10.4` as my base branch**

### DIFF
--- a/src/enums/pokemon-type.ts
+++ b/src/enums/pokemon-type.ts
@@ -20,3 +20,6 @@ export enum PokemonType {
   FAIRY,
   STELLAR
 }
+
+/** The largest legal value for a {@linkcode PokemonType} (includes Stellar) */
+export const MAX_POKEMON_TYPE = PokemonType.STELLAR;
--- a/src/data/mystery-encounters/encounters/weird-dream-encounter.ts
+++ b/src/data/mystery-encounters/encounters/weird-dream-encounter.ts
@@ -11,7 +11,7 @@ import { MysteryEncounterType } from "#enums/mystery-encounter-type";
 import { Nature } from "#enums/nature";
 import { PartyMemberStrength } from "#enums/party-member-strength";
 import { PlayerGender } from "#enums/player-gender";
-import { PokemonType } from "#enums/pokemon-type";
+import { MAX_POKEMON_TYPE, PokemonType } from "#enums/pokemon-type";
 import { SpeciesId } from "#enums/species-id";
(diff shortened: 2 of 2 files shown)


damocleas (approved):

xsn34kzx (approved):

src/enums/pokemon-type.ts
```diff
@@ -20,3 +20,6 @@ export enum PokemonType {
   FAIRY,
   STELLAR
 }
+
+/** The largest legal value for a {@linkcode PokemonType} (includes Stellar) */
+export const MAX_POKEMON_TYPE = PokemonType.STELLAR;
```
DayKev, line 25: That name might be a bit misleading, since we don't want any Pokémon to have its actual type as Stellar.

DayKev (approved):

Wlowscha (approved):

PR #5679, opened 2025-04-18
[Refactor] Refactor evo conditions and descriptions

Fontbane:
## Checklist
- [ ] **I'm using `beta` as my base branch**
- [ ] There is no overlap with another PR?
- [ ] The PR is self-contained and cannot be split into smaller PRs?
- [ ] Have I provided a clear explanation of the changes?
- [ ] Have I tested the changes manually?
- [ ] Are all unit tests still passing? (`npm run test`)
  - [ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?
- [ ] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (default and legacy)?

Are there any localization additions or changes? If so:
- [ ] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?
  - [ ] If so, please leave a link to it here: 
- [ ] Has the translation team been contacted for proofreading/translation?

PR #6414, opened 2025-08-25
[Hotfix] Fix tyrogue evo

### LINKED ISSUE #6408: [Bug] Tyrogue won't evolve unless evo move is on slot 1
### Describe the bug

Tyrogue is meant to have three different evolution moves for its three evolutions. When knowing multiple of these moves, it is chosen based on whichever is higher up in its moveset ordering. In the recent update, it won't evolve unless a specific move is in the first slot of its moveset. 

### Reproduction

const overrides = {STARTER_SPECIES_OVERRIDE:SpeciesId.TYROGUE, STARTING_LEVEL_OVERRIDE: 19, ITEM_REWARD_OVERRIDE: [{name: "RARE_CANDY"}, {name: "MEMORY_MUSHROOM"}]} satisfies Partial<InstanceType<OverridesType>>;

Use Memory Mushrooms as needed to get any of Low Sweep, Mach Punch and Rapid Spin in different slots. Any other move in the highest slot means Tyrogue won't evolve

### Expected behavior

Tyrogue evolves even if evo move is on fourth slot

### Screenshots / Videos

https://github.com/user-attachments/assets/e6aeaefe-a0d1-4a70-b197-de459415aaf4

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

Wlowscha:
Fixes #6408 

Evolution condition was simplified too much in #5679 .

Doing `as TyrogueMove` does not actually enforce typing at runtime, so we need to check that the moveId is one of the allowed ones.

## Screenshots/Videos

https://github.com/user-attachments/assets/8b51988a-f491-4767-ae62-ff43b99b0a4d

https://github.com/user-attachments/assets/8803af08-451f-44a3-aad1-3e19bb52482e

## How to test the changes?
Override from the bug report:
```ts
const overrides = {
  STARTER_SPECIES_OVERRIDE:SpeciesId.TYROGUE,
  STARTING_LEVEL_OVERRIDE: 19,
  ITEM_REWARD_OVERRIDE: [{name: "RARE_CANDY"}, {name: "MEMORY_MUSHROOM"}],
```
Note that apparently move overrides can mess with the evo conditions, use memory mushrooms instead.

### DIFF
--- a/src/data/balance/pokemon-evolutions.ts
+++ b/src/data/balance/pokemon-evolutions.ts
@@ -77,6 +77,7 @@ export enum EvolutionItem {
   LEADERS_CREST
 }
 
+const tyrogueMoves = [MoveId.LOW_SWEEP, MoveId.MACH_PUNCH, MoveId.RAPID_SPIN];
 type TyrogueMove = MoveId.LOW_SWEEP | MoveId.MACH_PUNCH | MoveId.RAPID_SPIN;
 
 /**
@@ -192,7 +193,7 @@ export class SpeciesEvolutionCondition {
         case EvoCondKey.WEATHER:
           return cond.weather.includes(globalScene.arena.getWeatherType());
         case EvoCondKey.TYROGUE:
-          return pokemon.getMoveset(true).find(m => m.moveId as TyrogueMove)?.moveId === cond.move;
+          return pokemon.getMoveset(true).find(m => tyrogueMoves.includes(m.moveId))?.moveId === cond.move;
         case EvoCondKey.NATURE:
           return cond.nature.includes(pokemon.getNature());
         case EvoCondKey.RANDOM_FORM: {


DayKev (approved):

src/data/balance/pokemon-evolutions.ts
```diff
@@ -77,6 +77,7 @@ export enum EvolutionItem {
   LEADERS_CREST
 }
 
+const tyrogueMoves = [MoveId.LOW_SWEEP, MoveId.MACH_PUNCH, MoveId.RAPID_SPIN];
 type TyrogueMove = MoveId.LOW_SWEEP | MoveId.MACH_PUNCH | MoveId.RAPID_SPIN;
```
Bertie690:
```suggestion
type TyrogueMove = tyrogueMoves[number];
```

src/data/balance/pokemon-evolutions.ts
```diff
@@ -77,6 +77,7 @@ export enum EvolutionItem {
   LEADERS_CREST
 }
 
+const tyrogueMoves = [MoveId.LOW_SWEEP, MoveId.MACH_PUNCH, MoveId.RAPID_SPIN];
```
Bertie690:
```suggestion
const tyrogueMoves = [MoveId.LOW_SWEEP, MoveId.MACH_PUNCH, MoveId.RAPID_SPIN] as const;
```

probably should make this readonly

src/data/balance/pokemon-evolutions.ts
```diff
@@ -192,7 +193,7 @@ export class SpeciesEvolutionCondition {
         case EvoCondKey.WEATHER:
           return cond.weather.includes(globalScene.arena.getWeatherType());
         case EvoCondKey.TYROGUE:
-          return pokemon.getMoveset(true).find(m => m.moveId as TyrogueMove)?.moveId === cond.move;
+          return pokemon.getMoveset(true).find(m => tyrogueMoves.includes(m.moveId))?.moveId === cond.move;
```
Bertie690:
```suggestion
          return pokemon.getMoveset(true).find(m => (tyrogueMoves as readonly MoveId[]) .includes(m.moveId))?.moveId === cond.move;
```

Bertie690 (approved):
Minor nits for typing but lgtm

Fontbane:
My bad, there should probably be a test for this

src/data/balance/pokemon-evolutions.ts
Wlowscha: This suggestion errors when I put it in my code.

DayKev: You have to also add the other suggestion below that adds the `as readonly MoveId[]`, they go together.

PR #6410, opened 2025-08-25
[Bug] Prevent an empty starterpreferences object from being saved

SirzBenjie:
## What are the changes the user will see?
Starter preferences should no longer be sometimes cleared when exiting starter select UI.

## Why am I making these changes?
Fixes a very annoying bug

## What are the changes from a developer perspective?
Made it so that starter preferences cannot save if the object is `{}`.
The only reason starter preferences could be cleared is if the `originalstarterPrefs` was set to `{}`. This happens when `clearStarterPrefs` is called. It is called inside of SSUI's `clear` method. While the `show` method should repopulate it, it does not if it was called with no args (which in turn should only happen when closing an overlay, and in that case, SSUI should not (?) have had its `clear` method called. And yet here we are. I see no other code path that could cause SSUI to work like it is.

## How to test the changes?
No idea. I haven't been able to reliably replicate it.

## Checklist
- [x] **I'm using `hotfix-1.10.4` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [ ] Have I tested the changes manually? I haven't been able to reliably replicate this. In theory, this should prevent the bug.
- [x] Are all unit tests still passing? (`pnpm test:silent`)
  - [ ] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?
- [ ] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (default and legacy)?

PR #6390, opened 2025-08-24
[UI/UX][Refactor] Starter select UI refactor

Wlowscha:
# More-or-less complete list of methods:

The functions setup, setupFlterBar, setupPartyColumn, setupInstructionButtons are about 500 lines uniquely to set up ui elements and some other stuff like the list of starters.
These should be okay for now.

initStarterPrefs is a function which sanitizes the preferences for a given starter, basically ensures that no invalid values are selected as preferences, although it might be useless or redundant.

resetFilters and resetCaughtDropdown are harmless, filter-related functions.

showText is also harmless.

setUpgradeAnimation and setUpgradeIcon are used to show which starters have candy upgrades available, I have not touched these at all. There are also updateCandyUpgradeDisplay and onCandyUpgradeDisplayChanged.

Then we have processInput, which calls one of processFilterModeInput, processStartCursorInput, processRandomCursorInput, processCycleButtonsInput, processPartyIconInput, openPokemonMenu.
The latter is particularly beefy, altogether these functions are over 1000 lines of code.
In particular, processCycleButtonsInput and openPokemonMenu definitely need a makeover.

switchMoveHandler, updateSelectedStarterMoveset are related to the menu which allows to reorder moves and pick the moveset.

updateButtonIcon, updateFilterButtonIcon, updateInstructions are all related to the various buttons that either cycle through preferences or jump to the filters.

isInParty, addToParty, updatePartyIcon, and later popStarter, updateStarterValueLabel, tryUpdateValue are all functions related to adding or removing starters to the column on the right. I would like to change the way these functions store information.

updateStarters is a huge one, probably needs to be renamed. It's the function that applies challenges and filters, thereby reducing the list of starters displayed.

updateScroll is another big one, also needs to be renamed, and is the one which checks which starters actually need to be displayed on the screen, and updates all the containers.

setSpecies is a function which is used to update various data based on the species currently selected in the box. It calls the .setSpecies function of the starterSummary which updates the sprite, information such as name, groth rate, eggs hatched and so on. There is still some redundancy here which can probably be removed.
I also made a setNoSpecies function which clears the information when the cursor goes somewhere else---although this may not be strictly necessary.

startIconAnimation, stopIconAnimation are utility function to update the bouncing animations of icons.

[...]

PR #6426, opened 2025-08-26
Fix starter preferences resetting (1.10.4-rc2)

SirzBenjie:
See #6410 for the other fields

We should have confidence that the _attempted_ fix in #6410 _did stop this bug_ (though introduced a different one), because after that patch went live, people that were previously impacted by the issue _were seeing nickname persist_ (whereas it wouldn't before).

The reason that nickname would persist while the other primitives weren't, is because our nickname is actually using the `String()` method (for whatever reason) to create a string _object_ instead, which is what made it immune.

Made it so that starter preferences aren't saved when *first opening* SSUI, as this could happen _before_ the `initStarterPrefs` method was called, meaning that the starter preferences could be overwritten!
This is accomplished by making all of the calls to `setSpeciesDetials` in `setStarter` pass the `save = false` value (which is a new parameter to `setSpeciesDetails`.

I also *fixed* my previous attempt, which had a HUGE oversight where primitives were being treated as empty objects.

## Screenshots/Videos

https://github.com/user-attachments/assets/0eca5a9d-8952-4f9e-945b-f7e26a8bc450

## How to test the changes?
`pnpm test test/utils/data.test.ts`

## Checklist
- [x] **I'm using `hotfix-1.10.5` as my base branch**

### DIFF
--- a/src/ui/starter-select-ui-handler.ts
+++ b/src/ui/starter-select-ui-handler.ts
@@ -72,7 +72,7 @@ import {
   rgbHexToRgba,
 } from "#utils/common";
 import type { StarterPreferences } from "#utils/data";
-import { loadStarterPreferences, saveStarterPreferences } from "#utils/data";
+import { deepCopy, loadStarterPreferences, saveStarterPreferences } from "#utils/data";
 import { getPokemonSpeciesForm, getPokerusStarters } from "#utils/pokemon-utils";
 import { toCamelCase, toTitleCase } from "#utils/strings";
 import { argbFromRgba } from "@material/material-color-utilities";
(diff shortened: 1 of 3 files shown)


src/ui/starter-select-ui-handler.ts
```diff
@@ -1828,9 +1831,15 @@ export class StarterSelectUiHandler extends MessageUiHandler {
...
       // The sanitized starter preferences
-      let starterAttributes = this.starterPreferences[this.lastSpecies.speciesId];
-      // The original starter preferences
-      const originalStarterAttributes = this.originalStarterPreferences[this.lastSpecies.speciesId];
+      if (this.starterPreferences[this.lastSpecies.speciesId] === undefined) {
+        this.starterPreferences[this.lastSpecies.speciesId] = {};
+      }
+      if (this.originalStarterPreferences[this.lastSpecies.speciesId] === undefined) {
+        this.originalStarterPreferences[this.lastSpecies.speciesId] = {};
+      }
```
SirzBenjie, line 1839: This set of changes ensures that the object being written to in the remainder of this function _is the same object reference_ as the variables below.

Wlowscha (approved):

Wlowscha:
Tested locally and there seem to be no issue with preferences when opening/closing etc. The changes to ssui all make sense, some of these will make it to #6390 at least in the first stage.

damocleas (approved):

PR #6440, opened 2025-08-27
[Bug] Breaking a boss bar will no longer crash at max stat stages

DayKev:
If an enemy boss pokemon has +6 in all stat stages and a boss bar gets broken, the game will no longer crash.

Crash fix.

If no stat is found to be raised then the function returns early instead of passing `undefined` to `StatStageChangePhase`.

## Checklist


Bertie690 (approved):

SirzBenjie:
Do we know what caused this to trigger in this version?

DayKev:
Nope.

PR #5873, opened 2025-05-26
[Refactor] Refactor command phase

SirzBenjie:
## What are the changes the user will see?
None

## Why am I making these changes?
This is another part of the work to cleanup the phase files and make them more maintainable and easier to understand.
All of the methods in command phase now pass the cognitive complexity lint.

## What are the changes from a developer perspective?

Broke up the mega methods inside of command phase into small, purposeful parts without changing any of the logic.
Added documentation to each part.

## Screenshots/Videos
N/A

## How to test the changes?
Just play the game and ensure that the game runs as before.
I played ~25 waves or so.
I went through all of the logic to ensure everything was called just as it was before.

Brittle mechanics that are sensitive to breaking:
Switching via the baton item works as normal
Attacking / target selection in double battles continues as normal
Fleeing / Switching under the effects of a trapping effect continues to work as normal

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`npm run test:silent`)
  - ~~[ ] Have I created new automated tests (`npm run create-test`) or updated existing tests related to the PR's changes?~~
- ~~[ ] Have I provided screenshots/videos of the changes (if applicable)?~~
  - ~~[ ] Have I made sure that any UI change works for both UI themes (default and legacy)?~~

PR #6454, opened 2025-08-29
[Bug] Pokemon not on the field can't be caught

Wlowscha:
The root of this bug is an "oversight" from #5873 , where the line
`const targetPokemon = globalScene.getEnemyField().find(p => p.isActive(true));`
was replaced by
`const targetPokemon = globalScene.getEnemyPokemon();`

Due to the weirdness surrounding these functions, `isActive` can be `true` even if the mon is not on the field, which is why it was called with the parameter `onField` set to `true`. Similarly, `getEnemyPokemon` must be called with the parameter `false`, otherwise it also considers mons that have switched out.

This is relevant in the case of trying to catch a boss in the second party slot, after phazing out the boss in the first slot. This caused a bug reported in [this discord thread](https://discord.com/channels/1125469663833370665/1409970865415651511): even after the second mon has been weakened down, the game continues checking the first mon, which is not on the field anymore but still has all its health segments.

## Screenshots/Videos

After the fix:

https://github.com/user-attachments/assets/0538f2a6-cb58-40d2-a418-85c1c730a344

## How to test the changes?

Check discord thread for session data and instructions.

### DIFF
--- a/src/phases/command-phase.ts
+++ b/src/phases/command-phase.ts
@@ -456,7 +456,7 @@ export class CommandPhase extends FieldPhase {
 
     const numBallTypes = 5;
     if (cursor < numBallTypes) {
-      const targetPokemon = globalScene.getEnemyPokemon();
+      const targetPokemon = globalScene.getEnemyPokemon(false);
       if (
         targetPokemon?.isBoss() &&
         targetPokemon?.bossSegmentIndex >= 1 &&


Fontbane:
Dragon Tail shouldn't even cause wild mons to flee in Doubles but that might need more consideration

SirzBenjie (approved):

damocleas (approved):

DayKev (approved):

PR #6456, opened 2025-08-29
[Modifier rework] Held item manager uses a map

Wlowscha:
@Bertie690 has been requesting this for some time.

### DIFF
--- a/src/items/held-item-manager.ts
+++ b/src/items/held-item-manager.ts
@@ -14,18 +14,17 @@ import {
   type HeldItemSpecs,
   isHeldItemSpecs,
 } from "#items/held-item-data-types";
-import { getTypedKeys } from "#utils/common";
 
 export class HeldItemManager {
   // TODO: There should be a way of making these private...
   public heldItems: HeldItemDataMap;
 
   constructor() {
-    this.heldItems = {};
+    this.heldItems = new Map();
   }
 
   getItemSpecs(id: HeldItemId): HeldItemSpecs | undefined {
-    const item = this.heldItems[id];
+    const item = this.heldItems.get(id);
     if (item) {
       const itemSpecs: HeldItemSpecs = {
         ...item,
@@ -38,7 +37,7 @@ export class HeldItemManager {
 
   generateHeldItemConfiguration(restrictedIds?: HeldItemId[]): HeldItemConfiguration {
     const config: HeldItemConfiguration = [];
-    for (const [id, item] of this.getHeldItemEntries()) {
+    for (const [id, item] of this.heldItems.entries()) {
       // TODO: `in` breaks with arrays
       if (item && (!restrictedIds || id in restrictedIds)) {
         const specs: HeldItemSpecs = { ...item, id };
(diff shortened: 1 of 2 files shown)


src/items/held-item-manager.ts
```diff
@@ -83,7 +78,7 @@ export class HeldItemManager {
     if (isCategoryId(itemType)) {
       return this.getHeldItems().some(id => isItemInCategory(id, itemType as HeldItemCategoryId));
     }
-    return itemType in this.heldItems;
+    return this.heldItems.has(itemType as HeldItemId);
```
Bertie690: Idk if assertion needed (check pls)

src/items/held-item-manager.ts
```diff
@@ -50,7 +49,7 @@ export class HeldItemManager {
 
   generateSaveData(): HeldItemSaveData {
     const saveData: HeldItemSaveData = [];
-    for (const [id, item] of this.getHeldItemEntries()) {
+    for (const [id, item] of this.heldItems.entries()) {
```
Bertie690:
```suggestion
    for (const [id, item] of this.heldItems) {
```
Map iterator defaults to entries

src/items/held-item-manager.ts
```diff
@@ -38,7 +37,7 @@ export class HeldItemManager {
 
   generateHeldItemConfiguration(restrictedIds?: HeldItemId[]): HeldItemConfiguration {
     const config: HeldItemConfiguration = [];
-    for (const [id, item] of this.getHeldItemEntries()) {
+    for (const [id, item] of this.heldItems.entries()) {
```
Bertie690:
```suggestion
    for (const [id, item] of this.heldItems) {
```

src/items/held-item-manager.ts
```diff
@@ -140,53 +135,56 @@ export class HeldItemManager {
...
     } else {
-      this.heldItems[itemType] = { stack: Math.min(addStack, maxStack) };
+      this.heldItems.set(itemType, { stack: Math.min(addStack, maxStack) });
       return true;
     }
+
     return false;
   }
 
   addItemWithSpecs(itemSpecs: HeldItemSpecs): boolean {
```
Bertie690, line 135: Wait why does this always return `true`...

src/items/held-item-manager.ts
```diff
@@ -92,35 +87,35 @@ export class HeldItemManager {
...
     }
   }
 
   isMaxStack(itemType: HeldItemId): boolean {
-    const item = this.heldItems[itemType];
+    const item = this.heldItems.get(itemType);
     return item ? item.stack >= allHeldItems[itemType].getMaxStackCount() : false;
   }
 
   overrideItems(newItems: HeldItemDataMap) {
```
Bertie690: Ftr, this requires the function to be called with an actual Map...

prob not what we had in mind

tbh i kinda dislike the existence of this method in general, will need a second serious look-over - a test utility attempting to overwrite items would likely make more sense than shoving a method on the manager itself. (If anything else uses it, uhhhhhhhh)

Bertie690 (changes requested):
P much what I had in mind - checking and refining the manager's exterior method list might be a good idea (a lot of these methods seem repetitive or able to be condensed down a bit).

Also return values and yadda yadda

src/items/held-item-manager.ts
Wlowscha: Not always, it returns false if the stack is already full

src/items/held-item-manager.ts
Wlowscha: I don't get it, what is the issue?

Wlowscha: Yeah the method was unused anyway, I just killed it.

PR #6490, opened 2025-09-05
[GitHub] Fix Typedoc workflow

DayKev:
Fixing a small mistake from the Typedoc PR.

Inverted the conditional and re-added the `git add`.

## Checklist

### DIFF
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -47,7 +47,7 @@ jobs:
         with:
           version: 10
 
-      - name: Setup Node 22.14.1
+      - name: Setup Node
         uses: actions/setup-node@v4
         with:
           node-version-file: "pokerogue_docs/.nvmrc"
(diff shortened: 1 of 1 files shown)


.github/workflows/github-pages.yml
```diff
@@ -65,9 +65,6 @@ jobs:
 
       - name: Generate Typedoc docs
         working-directory: ${{env.docs-dir}}
-        env:
-          REF_NAME: ${{github.ref_name}}
-          DRY_RUN: ${{env.DRY_RUN}}
         run: pnpm typedoc
```
Bertie690: This needs to pass the env vars to the script afaik
```suggestion
        run: REF_NAME=${{ env.REF_NAME }}DRY_RUN=${{ env.DRY_RUN }} pnpm typedoc
```

DayKev: Does typedoc actually use those?

DayKev: Oh I see, I didn't realize they were used from within the JS files.

DayKev: In that case it probably does work, I thought they weren't being used in this step at all.

SirzBenjie (approved):

PR #6491, opened 2025-09-05
Add blacklist of moves that can't be forced STAB.

Xavion3:
A few dozen specific moves should no longer be generated as the only damaging move in a moveset.

Balance asked.

The blacklist is currently hardcoded inside the moveset generation code.

## Screenshots/Videos
N/A

## How to test the changes?
It's easiest to see this with a debugger, but just generating movesets for pokemon who get these moves can work.

## Checklist

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -3220,9 +3220,51 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       Math.ceil(Math.pow(m[1], weightMultiplier) * 100),
     ]);
 
+    const STAB_BLACKLIST = [
+      MoveId.SHELL_TRAP,
+      MoveId.FUTURE_SIGHT,
+      MoveId.UPPER_HAND,
+      MoveId.SKY_DROP,
+      MoveId.STEEL_ROLLER,
+      MoveId.BELCH,
+      MoveId.FAKE_OUT,
+      MoveId.LAST_RESORT,
+      MoveId.SYNCHRONOISE,
+      MoveId.DREAM_EATER,
+      MoveId.DOOM_DESIRE,
+      MoveId.FOCUS_PUNCH,
+      MoveId.NIGHT_SHADE,
+      MoveId.DRAGON_RAGE,
+      MoveId.SONIC_BOOM,
+      MoveId.BIDE,
+      MoveId.COMEUPPANCE,
+      MoveId.MIRROR_COAT,
+      MoveId.METAL_BURST,
+      MoveId.COUNTER,
+      MoveId.FLING,
+      MoveId.NATURAL_GIFT,
+      MoveId.SPIT_UP,
+      MoveId.FISSURE,
+      MoveId.SHEER_COLD,
+      MoveId.HORN_DRILL,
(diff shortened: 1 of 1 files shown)


Madmadness65 (approved):

src/field/pokemon.ts
```diff
@@ -3220,9 +3220,51 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      MoveId.GUILLOTINE,
+      MoveId.ENDEAVOR,
+      MoveId.NATURES_MADNESS,
+      MoveId.RUINATION,
+      MoveId.SUPER_FANG,
+      MoveId.SNORE,
+      MoveId.HOLD_BACK,
+      MoveId.BEAT_UP,
+      MoveId.PSYWAVE,
+      MoveId.FIRST_IMPRESSION,
```
SirzBenjie: Add explosion moves? (Or do we already blacklist those elsewhere?)

src/field/pokemon.ts
```diff
@@ -3220,9 +3220,51 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    ];
```
SirzBenjie: Might make more sense to place this in its own file in either balance or moves. This method is already pretty big :sob:

Also, we can and should use a map (and tell typescript it's a readonly map 🙂) as it's only ever being checked for inclusion.

src/field/pokemon.ts
Blitz425: Oh actually this is a very good point, I'm not sure if they are blacklisted somewhere but I commonly run into Misty Explosion all the time on Fairies.

src/field/pokemon.ts
Xavion3: I'd avoided the separate file for now, this is just a temporary fix till we start the AI rework in after all.

Xavion3: I have changed it to a readonly set though.

src/field/pokemon.ts
Xavion3: Done.

SirzBenjie (dismissed):

emdeann (dismissed):
Code is good, needs bal approval before sending

Xavion3:
We've already independently started on separating some of this into its own files, and yeah this is meant to be pointed at 1.10 I just hadn't rebased it yet.

Wlowscha:
Alright, I'll approve once it is rebased then.

Xavion3:
Looks like the next minor update is coming off beta now instead of hotfix so no need to rebase.

Wlowscha (dismissed):

src/field/pokemon.ts
Mourouh: I'd add Final Gambit to this list

Xavion3: Final gambit is already blacklisted from ever generating on any pokemon ever.

Blitz425:
Can we add Steel Beam to this list? Afterwards we can approve and ship this in

SirzBenjie:
Added steel beam blacklist as requested. Going to bypass merge since it had approvals otherwise.

PR #6500, opened 2025-09-06
[Dev] Update github-pages.yml to maybe add the files right

Bertie690:
5 TIMES THE CHARM????

### DIFF
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -77,7 +77,7 @@ jobs:
           cd pokerogue_gh
           git config user.email "github-actions[bot]@users.noreply.github.com"
           git config user.name "github-actions[bot]"
-          mv -f /tmp/docs/* $GITHUB_REF_NAME
+          mv -f /tmp/docs $GITHUB_REF_NAME
           git add $GITHUB_REF_NAME
           git commit -m "[skip ci] Deploy docs"
           git push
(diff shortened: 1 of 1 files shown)


.github/workflows/github-pages.yml
```diff
@@ -79,7 +79,7 @@ jobs:
           git config user.name "github-actions[bot]"
           mkdir -p $GITHUB_REF_NAME
           rm -rf $GITHUB_REF_NAME/*
-          cp -r /tmp/docs $GITHUB_REF_NAME
+          cp -r /tmp/docs/* $GITHUB_REF_NAME
```
SirzBenjie:
```suggestion
          mv /tmp/docs $GITHUB_REF_NAME
```

SirzBenjie:
This would be 6 💀

.github/workflows/github-pages.yml
```diff
@@ -77,9 +77,7 @@ jobs:
           cd pokerogue_gh
           git config user.email "github-actions[bot]@users.noreply.github.com"
           git config user.name "github-actions[bot]"
-          mkdir -p $GITHUB_REF_NAME
-          rm -rf $GITHUB_REF_NAME/*
-          cp -r /tmp/docs $GITHUB_REF_NAME
+          mv -f /tmp/docs/* $GITHUB_REF_NAME
```
SirzBenjie:
```suggestion
          mv -f /tmp/docs $GITHUB_REF_NAME
```

SirzBenjie: We don't want to use globbing. `mv` will move the file path specified into the specified destination. This should work.

PR #6504, opened 2025-09-07
[UI/UX] In settings, clamp volume cursor at boundary

SirzBenjie:
See discussion in #6404: this reverts the feature but only for volume settings.

Kev asked and ended up making a really good point. Plus, I thought of a good way of doing it that doesn't increase complexity in the code very much.

Added a `clamp` property to a setting. If true, then pressing left/right while on the boundary will cause the cursor to stay on the boundary instead of cycling. Useful for volume so you don't go from mute to full blast.

Also adjusted the menu to not play the sound if the clamp occurs, since nothing changed.

Well, I was editing the file, and this time couldn't resist the urge to apply my general tiny cleanup to the abstract settings UI. So those are in here. Sorry/not sorry

## Screenshots/Videos
<details><summary>Post behavior</summary>
Although you can't see it, I am pressing left / right on the audio menu

https://github.com/user-attachments/assets/711e467d-bb62-4f5f-bf36-0b40a8d9ffc9
</details> 

## How to test the changes?
Open settings, go to audio, press left to go to mute, try pressing left again and ensure that the option does not wrap (but other options do).
Game speed will also behave like this, though that can be changed. (I did this because it doesn't make sense that the hotkeys don't cycle, but the setting would).

## Checklist
  - [ ] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes? ui tests? yeah right

### DIFF
--- a/src/system/settings/settings.ts
+++ b/src/system/settings/settings.ts
@@ -120,6 +120,14 @@ export interface Setting {
   default: number;
   type: SettingType;
   requireReload?: boolean;
+  /**
+   * Specifies the behavior when navigating left/right at the boundaries of the option
+   *
+   * - `true`: the cursor will stay on the boundary instead of moving
+   * - `false`: the cursor will wrap to the other end of the options list
+   * @defaultValue `false`
+   */
+  clamp?: boolean;
   /** Whether the setting can be activated or not */
   activatable?: boolean;
   /** Determines whether the setting should be hidden from the UI */
@@ -230,6 +238,7 @@ export const Setting: Array<Setting> = [
     ],
     default: 3,
     type: SettingType.GENERAL,
+    clamp: false,
   },
   {
     key: SettingKeys.HP_Bar_Speed,
@@ -639,6 +648,7 @@ export const Setting: Array<Setting> = [
     options: VOLUME_OPTIONS,
     default: 5,
     type: SettingType.AUDIO,
+    clamp: true,
   },
   {
     key: SettingKeys.BGM_Volume,
@@ -646,6 +656,7 @@ export const Setting: Array<Setting> = [
     options: VOLUME_OPTIONS,
     default: 10,
     type: SettingType.AUDIO,
+    clamp: true,
   },
   {
     key: SettingKeys.Field_Volume,
@@ -653,6 +664,7 @@ export const Setting: Array<Setting> = [
     options: VOLUME_OPTIONS,
     default: 10,
     type: SettingType.AUDIO,
+    clamp: true,
   },
   {
     key: SettingKeys.SE_Volume,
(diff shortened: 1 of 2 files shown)


src/system/settings/settings.ts
```diff
@@ -230,6 +238,7 @@ export const Setting: Array<Setting> = [
     ],
     default: 3,
     type: SettingType.GENERAL,
+    clamp: false,
```
SirzBenjie, line 241: Put this here so that it can be removed / changed to true depending on whether we want game speed to wrap or not.

Bertie690 (dismissed):

src/ui/settings/abstract-settings-ui-handler.ts
```diff
@@ -317,22 +310,20 @@ export class AbstractSettingsUiHandler extends MessageUiHandler {
...
+          dir = -1;
+        case Button.RIGHT: {
+          let boundaryAction = Phaser.Math.Wrap;
+          let upperBound = this.optionValueLabels[cursor].length;
+          if (this.settings[cursor]?.clamp) {
+            boundaryAction = Phaser.Math.Clamp;
+            // clamping is right inclusive; wrapping isn't
+            upperBound -= 1;
+          }
+          success = this.setOptionCursor(cursor, boundaryAction(this.optionCursors[cursor] + dir, 0, upperBound), true);
```
Wlowscha: Can we put this code in a function, and have it be called by both left and right with `dir` as an input?

src/ui/settings/abstract-settings-ui-handler.ts
```diff
@@ -403,18 +395,21 @@ export class AbstractSettingsUiHandler extends MessageUiHandler {
       settingIndex = this.cursor + this.scrollCursor;
     }
     const setting = this.settings[settingIndex];
-
     const lastCursor = this.optionCursors[settingIndex];
+    // do nothing if the option isn't changing
+    if (cursor === lastCursor) {
+      return false;
+    }
```
Wlowscha, line 408: Isn't `cursor` telling us what setting is selected, while `lastCursor` in this case is the option selected for that specific setting?

src/ui/settings/abstract-settings-ui-handler.ts
SirzBenjie: Would love to! Will do

src/ui/settings/abstract-settings-ui-handler.ts
SirzBenjie: That's what I thought originally, but no. They're both the cursor in the option. (Thankfully, this time, the names aren't confusing).

Wlowscha: My bad I was looking at the code above lol.

fabske0 (dismissed):

DayKev (dismissed):

fabske0 (approved):

DayKev (approved):

Wlowscha (approved):

PR #6506, opened 2025-09-07
[UI/UX] Use pointer events instead of touch events in `TouchControl`

Wlowscha:
The changes can potentially make the mobile version of the game more accessible, including to devices which are not capable of sending "touch" events but still use some kind of point-and-click system, e.g. eye-tracking devices.

Difficult to test this (whether it still works as normal on phones, and whether there are any improvements on other devices) until the changes are on beta. So either we merge it and test directly on beta, or this PR can serve as a base to discuss alternative solutions.

### DIFF
--- a/src/ui/settings/move-touch-controls-handler.ts
+++ b/src/ui/settings/move-touch-controls-handler.ts
@@ -9,9 +9,9 @@ export const TOUCH_CONTROL_POSITIONS_PORTRAIT = "touchControlPositionsPortrait";
 type ControlPosition = { id: string; x: number; y: number };
 
 type ConfigurationEventListeners = {
-  touchstart: EventListener[];
-  touchmove: EventListener[];
-  touchend: EventListener[];
+  pointerdown: EventListener[];
+  pointermove: EventListener[];
+  pointerup: EventListener[];
 };
 
 type ToolbarRefs = {
@@ -39,9 +39,9 @@ export class MoveTouchControlsHandler {
    * These are used to remove the event listeners when the configuration mode is disabled.
    */
   private configurationEventListeners: ConfigurationEventListeners = {
-    touchstart: [],
-    touchmove: [],
-    touchend: [],
+    pointerdown: [],
+    pointermove: [],
+    pointerup: [],
   };
 
   private overlay: Phaser.GameObjects.Container;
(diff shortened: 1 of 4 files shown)


SirzBenjie:
We don't have to mark touch related tests as TODO, just have to change the event we dispatch to be `pointerdown` in `test-utils`.
```ts
  touchDown(button: string) {
    const node = document.querySelector(`[data-key][id='${button}']`);
    if (!node) {
      return;
    }
    const event = new Event("touchstart");
    node.dispatchEvent(event);
  }

  touchUp(button: string) {
    const node = document.querySelector(`[data-key][id='${button}']`);
    if (!node) {
      return;
    }
    const event = new Event("touchend");
    node.dispatchEvent(event);
  }
  ```

**Comment by Wlowscha** [comment]
> We don't have to mark touch related tests as TODO, just have to change the event we dispatch to be `pointerdown` in testUtils.

My reasoning was to keep changes at a minimum in case we need to revert them, and if it works we can fix the tests later.

**Comment by SirzBenjie** [comment]
So we can also test this on mobile. Open a server and use the --host option
``pnpm start:dev --host`` and then connect to the IP with mobile, while on the same network as the device that is hosting.

**Comment by SirzBenjie** [comment]
> > We don't have to mark touch related tests as TODO, just have to change the event we dispatch to be `pointerdown` in testUtils.
> 
> My reasoning was to keep changes at a minimum in case we need to revert them, and if it works we can fix the tests later.

I agree to keep changes to a minimum if it's in `src`, but adjusting it in `test-utils` is okay.

**Comment by Wlowscha** [comment]
> So we can also test this on mobile. Open a server and use the --host option `pnpm start:dev --host` and then connect to the IP with mobile, while on the same network as the device that is hosting.

Can confirm that the touch controls still work on mobile, after connecting to the local host from my phone. Played around with the Pokédex, started a run, etc

**Comment by SirzBenjie** [comment]
> > So we can also test this on mobile. Open a server and use the --host option `pnpm start:dev --host` and then connect to the IP with mobile, while on the same network as the device that is hosting.
> 
> Can confirm that the touch controls still work on mobile, after connecting to the local host from my phone. Played around with the Pokédex, started a run, etc

Same!

**Comment by Wlowscha** [comment]
I have also modified the ui which allows to change the position of touch controls, and tested it on mobile

**Review by SirzBenjie — APPROVED** [review]

File: src/ui/settings/move-touch-controls-handler.ts [hunk 1]
```diff
@@ -373,12 +372,12 @@ export class MoveTouchControlsHandler {
     this.draggingElement = null;
 
     // Remove event listeners
-    const { touchstart, touchmove, touchend } = this.configurationEventListeners;
+    const { pointerdown, pointermove, pointerup } = this.configurationEventListeners;
     this.getControlGroupElements().forEach((element, index) =>
-      element.removeEventListener("touchstart", touchstart[index]),
+      element.removeEventListener("touchstart", pointerdown[index]),
```
DayKev:
```suggestion
      element.removeEventListener("pointerdown", pointerdown[index]),
```
Shouldn't these also be changed?

src/ui/settings/move-touch-controls-handler.ts
```diff
@@ -373,12 +372,12 @@ export class MoveTouchControlsHandler {
...
     );
-    touchmove.forEach(listener => window.removeEventListener("touchmove", listener));
-    touchend.forEach(listener => window.removeEventListener("touchend", listener));
+    pointermove.forEach(listener => window.removeEventListener("touchmove", listener));
+    pointerup.forEach(listener => window.removeEventListener("touchend", listener));
```
DayKev:
```suggestion
    pointermove.forEach(listener => window.removeEventListener("pointermove", listener));
    pointerup.forEach(listener => window.removeEventListener("pointerup", listener));
```
Re: other comment

DayKev (approved):

PR #6509, opened 2025-09-07
[Bug] Wrong save slot

SirzBenjie:
Hopefully, users will no longer see the wrong session deleted when clearing a run.

P0 bug.

Added quite a few, possibly redundant (though redundant is *absolutely* ok when dealing with save data).
Added a logging debug statement to the `delete` endpoint that logs the slot that is requested to be deleted, so that if something goes wrong again, we can ask to see console.

Changed `PostGameOverPhase` to take a snapshot of the session slot when creating the phase, as an additional level of protection in case something gets corrupted between the phases.

Cleaned up the `getSession` and `loadSession` endpoints (without changing logic) to use `promise.withResolvers` instead of the `new Promise<...>` ugliness.

## Screenshots/Videos
Kinda long, lots of tests. Ping me if you want

## How to test the changes?
Need rogueserver running.
Play around with various things that cause a session slot to be cleared / loaded. Make sure the right session is being loaded and no error messages occur.
I couldn't replicate the trigger circumstances to guarantee that this fixes it. However, endpoints that _could_ clear a session are now validating against some slot ID that is checked _externally_ from global scene's current slot id, in case that somehow comes out of sync. 
It is my hypothesis that globalScene.sessionSlotId is somehow becoming desynchronized with the user's session.

## Checklist
- [x] **I'm using `1.10.7` as my base branch**


src/system/game-data.ts
```diff
@@ -1028,166 +1034,172 @@ export class GameData {
...
-    // biome-ignore lint/suspicious/noAsyncPromiseExecutor: TODO: fix this
-    return new Promise(async (resolve, reject) => {
-      try {
-        const initSessionFromData = async (sessionData: SessionSaveData) => {
-          console.debug(sessionData);
+  async loadSession(slotId: number, sessionData?: SessionSaveData): Promise<boolean> {
+    const { promise, resolve, reject } = Promise.withResolvers<boolean>();
+    try {
+      const initSessionFromData = (fromSession: SessionSaveData) => {
+        console.debug(fromSession);
```
DayKev:
```suggestion
```
This causes a memory leak, might as well remove it while you're already modifying this. Same for the one on line 425 (`console.debug(systemData);`).

src/system/game-data.ts
```diff
@@ -1028,166 +1034,172 @@ export class GameData {
...
+      if (sessionData) {
+        initSessionFromData(sessionData);
+      } else {
+        this.getSession(slotId)
+          .then(data => {
+            if (data) {
+              initSessionFromData(data);
+            }
+            return data && initSessionFromData(data);
+          })
```
DayKev: What exactly is this change for? If `data` isn't `nil` then the session gets initialized twice, otherwise it gets initialized zero times. I thought the problems were with saving and not loading?

DayKev: ~Uh... GitHub is showing the wrong section for some reason, hold on.~

Oh it's just showing more than I intended. Here's what I wanted it to show:

image

SirzBenjie: Looks like I didn't finish cleaning up the promise body.
Addressed.

src/system/game-data.ts
```diff
@@ -1028,166 +1043,179 @@ export class GameData {
...
+            reject(err);
+            return;
+          });
       }
-    });
+    } catch (err) {
+      reject(err);
+    }
+
+    return promise;
```
SirzBenjie: Can ignore these changes when reviewing:, all I did was un-nest the promise and used Promise.withResolvers

src/system/game-data.ts
```diff
@@ -1028,166 +1043,179 @@ export class GameData {
...
+              this.parseSessionData(
+                JSON.stringify(fromSession, (_, v: any) => (typeof v === "bigint" ? v.toString() : v)),
+              ),
+            );
+          } catch (err) {
+            console.debug("Attempt to log session data failed:", err);
           }
+        }
+        // overwrite slot ID if somehow not set.
+        fromSession.slotId = slotId;
```
SirzBenjie: These are the only changes in this block that aren't just dedenting.

src/system/game-data.ts
```diff
@@ -1268,6 +1296,13 @@ export class GameData {
       result = [true, true];
     } else {
       const sessionData = this.getSessionSaveData();
+      if (sessionData.slotId !== slotId) {
+        console.warn(
+          `Tried to clear session at slot ${slotId}, but current session is at slot ${sessionData.slotId}. Overriding to ${sessionData.slotId}.`,
+        );
+        // Nullish coalescing is here in case of sessionData from before we added slotId
+        slotId = sessionData.slotId ?? slotId;
+      }
```
SirzBenjie: Relevant

src/system/game-data.ts
```diff
@@ -1393,7 +1428,8 @@ export class GameData {
         const request = {
           system: systemData,
           session: sessionData,
-          sessionSlotId: globalScene.sessionSlotId,
+          // Nullish coalescing is set here in case there is somehow sessionData that was loaded but has no slotID (due to transition)
+          sessionSlotId: sessionData.slotId ?? globalScene.sessionSlotId,
```
SirzBenjie: relevant

src/system/game-data.ts
```diff
@@ -1521,6 +1557,7 @@ export class GameData {
               }
               case GameDataType.SESSION: {
                 const sessionData = this.parseSessionData(dataStr);
+                sessionData.slotId = slotId;
```
SirzBenjie: relevant

src/system/game-data.ts
```diff
@@ -138,6 +139,9 @@ export interface SessionSaveData {
    * Counts the amount of pokemon fainted in your party during the current arena encounter.
    */
   playerFaints: number;
+
+  /** The slot number the session was last saved in */
+  slotId: number;
```
SirzBenjie: relevant (context: adding to the `SessionSaveData` interface type)

src/system/game-data.ts
```diff
@@ -419,7 +423,15 @@ export class GameData {
...
-      console.debug(systemData);
+      if (isLocal || isBeta) {
+        try {
+          console.debug(
+            this.parseSystemData(JSON.stringify(systemData, (_, v: any) => (typeof v === "bigint" ? v.toString() : v))),
+          );
+        } catch (err) {
+          console.debug("Attempt to log system data failed:", err);
+        }
+      }
```
SirzBenjie, line 431: relevant

src/system/game-data.ts
```diff
@@ -942,48 +954,51 @@ export class GameData {
       mysteryEncounterType: globalScene.currentBattle.mysteryEncounter?.encounterType ?? -1,
       mysteryEncounterSaveData: globalScene.mysteryEncounterSaveData,
       playerFaints: globalScene.arena.playerFaints,
+      slotId: globalScene.sessionSlotId,
```
SirzBenjie: relevant

src/system/game-data.ts
SirzBenjie, line 1305: relevant

src/phases/title-phase.ts
```diff
@@ -177,6 +177,9 @@ export class TitlePhase extends Phase {
       .then((success: boolean) => {
         if (success) {
           this.loaded = true;
+          if (loggedInUser) {
+            loggedInUser.lastSessionSlot = slotId;
+          }
```
SirzBenjie, line 182: relevant

DayKev (approved):

PR #6529, opened 2025-09-09
[Bug] [Ability] Fix stat boost message timing from quark drive / protosynthesis

SirzBenjie:
Stat boost message for quark drive / protosynthesis will now no longer occur after the effect.

Annoying bug

Pass `false` for defer in the method's invocation.

## How to test the changes?
Activate quark drive in battle via something like electric terrain.

## Checklist
  - [ ] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes? phase ordering with the other stuff going on? nty

### DIFF
--- a/src/data/battler-tags.ts
+++ b/src/data/battler-tags.ts
@@ -2159,7 +2159,7 @@ export class HighestStatBoostTag extends AbilityBattlerTag {
       null,
       false,
       null,
-      true,
+      false,
     );
   }
 


Wlowscha (dismissed):

src/data/battler-tags.ts
```diff
@@ -2171,6 +2171,10 @@ export class HighestStatBoostTag extends AbilityBattlerTag {
         pokemonNameWithAffix: getPokemonNameWithAffix(pokemon),
         abilityName: allAbilities[this.ability].name,
       }),
+      undefined,
+      undefined,
+      undefined,
+      false,
```
SirzBenjie: realize this is unnecessary, since it defaults to `null` which is `false`. So it unshifts by default..

emdeann (approved):

Wlowscha (approved):

PR #6534, opened 2025-09-10
[Refactor] [Ability] Initializing abilities now uses a builder class

SirzBenjie:
None
(A miniscule less amount of data for ability instances due to flags being condensed into one field)

I dislike seeing these methods on abilities that are only meant to be used during initialization whenever I'm looking at them and accessing their parameters. This will clean up the intellisense for methods in 90% of the codebase.

- Renamed the `Ability` class inside of `abilities.ts` to `AbBuilder`, then created a new class, `Ability` to replace it.
- Renamed the is[...] properties to discard the `is`, and use `getters` that check a new `flags` field on the ability class
- Created a `build` method on `AbBuilder` that calls Ability's private constructor
- Removed the `implements localizable` field, which is a silly way of localizing fields when we can really just be using accessors.

## Screenshots/Videos
N/A

## How to test the changes?
`pnpm test:silent`

## Checklist
- [ ] Have I tested the changes manually? Nothing to test
- ~~[ ] Have I provided screenshots/videos of the changes (if applicable)?~~ N/A

### DIFF
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -62,7 +62,6 @@ import type {
   PokemonDefendCondition,
   PokemonStatStageChangeCondition,
 } from "#types/ability-types";
-import type { Localizable } from "#types/locales";
 import type { Closed, Exact } from "#types/type-helpers";
 import type { Constructor } from "#utils/common";
 import {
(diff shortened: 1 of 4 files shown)


src/data/abilities/ability.ts
```diff
@@ -77,43 +76,122 @@ import {
...
-  public generation: number;
-  public readonly postSummonPriority: number;
-  public isBypassFaint: boolean;
-  public isIgnorable: boolean;
-  public isSuppressable = true;
-  public isCopiable = true;
-  public isReplaceable = true;
-  public attrs: AbAttr[];
-  public conditions: AbAttrCondition[];
+    // TODO: Make a `getDisplayName` method (or another getter) that puts the (N) and (P) markers after the name
```
emdeann, line 112: Need to be careful when doing this because the (N) in the name is actually used to determine if an ability is unimplemented (lol)

SirzBenjie: Yeah that's why I marked it as TODO. It's for a different PR.

emdeann (dismissed):

PR #6544, opened 2025-09-11
[Refactor] Made `BattleScene.getPokemonById` return `undefined` instead of `null`

Bertie690:
N/A
Needed to port another pr

Also returning `null` from a field pokemon getter is absolutely BORKED
Made function not return `null` and removed unneeded optional chains

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -863,6 +863,7 @@ export class BattleScene extends SceneBase {
    * @returns The `Pokemon` associated with the given ID,
    * or `undefined` if none is found in either team's party.
    * @see {@linkcode Pokemon.id}
+   * @todo `pokemonId` should not allow `undefined`
    */
   public getPokemonById(pokemonId: number | undefined): Pokemon | undefined {
     if (pokemonId == null) {
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -860,19 +860,18 @@ export class BattleScene extends SceneBase {
 
   /**
    * Return the {@linkcode Pokemon} associated with a given ID.
-   * @param pokemonId - The ID whose Pokemon will be retrieved.
-   * @returns The {@linkcode Pokemon} associated with the given id.
-   * Returns `null` if the ID is `undefined` or not present in either party.
-   * @todo Change the `null` to `undefined` and update callers' signatures -
-   * this is weird and causes a lot of random jank
+   * @param pokemonId - The ID whose Pokemon will be retrieved
+   * @returns The {@linkcode Pokemon} associated with the given id,
+   * or `undefined` if no Pokemon with the given id is found.
    */
-  getPokemonById(pokemonId: number | undefined): Pokemon | null {
+  getPokemonById(pokemonId: number | undefined): Pokemon | undefined {
     if (isNullOrUndefined(pokemonId)) {
-      return null;
+      // biome-ignore lint/nursery/noUselessUndefined: shut up biome for now
+      return undefined;
     }
 
     const party = (this.getPlayerParty() as Pokemon[]).concat(this.getEnemyParty());
-    return party.find(p => p.id === pokemonId) ?? null;
+    return party.find(p => p.id === pokemonId);
   }
 
   addPlayerPokemon(
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -861,12 +861,12 @@ export class BattleScene extends SceneBase {
    * Return the {@linkcode Pokemon} associated with the given ID.
    * @param pokemonId - The PID whose Pokemon will be retrieved
    * @returns The `Pokemon` associated with the given ID,
-   * or `undefined` if no Pokemon with the given ID is found.
+   * or `undefined` if none is found in either team's party.
    * @see {@linkcode Pokemon.id}
    */
-  getPokemonById(pokemonId: number | undefined): Pokemon | undefined {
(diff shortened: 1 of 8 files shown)


src/battle-scene.ts
```diff
@@ -860,19 +860,18 @@ export class BattleScene extends SceneBase {
...
+   * @returns The {@linkcode Pokemon} associated with the given id,
+   * or `undefined` if no Pokemon with the given id is found.
    */
-  getPokemonById(pokemonId: number | undefined): Pokemon | null {
+  getPokemonById(pokemonId: number | undefined): Pokemon | undefined {
     if (isNullOrUndefined(pokemonId)) {
-      return null;
+      // biome-ignore lint/nursery/noUselessUndefined: shut up biome for now
+      return undefined;
     }
```
emdeann: Why is this needed at all? If the id was null or undefined you'd get `undefined` returned anyway

SirzBenjie: It's an optimization. No need to call a bunch of methods that are going to do nothing and never match.

src/phases/obtain-status-effect-phase.ts
```diff
@@ -23,6 +23,7 @@ export class ObtainStatusEffectPhase extends PokemonPhase {
    * @param sourceText - The text to show for the source of the status effect, if any; default `null`.
    * @param statusMessage - A string containing text to be displayed upon status setting;
    * defaults to normal key for status if empty or omitted.
+   * @todo STOP PASSING `null` TO THE PHASE
```
SirzBenjie:
```suggestion
   * @todo stop passing `null` to the phase
```

src/phases/pokemon-phase.ts
```diff
@@ -9,7 +9,9 @@ export abstract class PokemonPhase extends FieldPhase {
    * TODO: Make this either use IDs or `BattlerIndex`es, not a weird mix of both
    */
   protected battlerIndex: BattlerIndex | number;
+  /** TODO: Why is this needed? */
```
SirzBenjie:
```suggestion
  // TODO: Why is this needed?
```

src/battle-scene.ts
emdeann: Sure, needs a better comment than "shut up biome for now" then lol

src/battle-scene.ts
```diff
@@ -860,19 +860,18 @@ export class BattleScene extends SceneBase {
...
-   * Returns `null` if the ID is `undefined` or not present in either party.
-   * @todo Change the `null` to `undefined` and update callers' signatures -
-   * this is weird and causes a lot of random jank
+   * @param pokemonId - The ID whose Pokemon will be retrieved
+   * @returns The {@linkcode Pokemon} associated with the given id,
+   * or `undefined` if no Pokemon with the given id is found.
    */
-  getPokemonById(pokemonId: number | undefined): Pokemon | null {
+  getPokemonById(pokemonId: number | undefined): Pokemon | undefined {
     if (isNullOrUndefined(pokemonId)) {
```
DayKev: `pokemonId` shouldn't be allowed to be `undefined` in the first place.
Fyi here is how pkty has this function now:
```ts
  getPokemonById(pokemonId: number): Pokemon | undefined {
    const findInParty = (party: Pokemon[]) => party.find((p) => p.id === pokemonId);
    return findInParty(this.getPlayerParty()) || findInParty(this.getEnemyParty());
  }
```

Bertie690: Afaik it's syntactic sugar for callers not to have to do `this.sourceId ?? 0`

Bertie690: Can this be a separate discussion and/or a separate PR?
I just want this stuff in so I can start implementing Inner's refactor

src/battle-scene.ts
```diff
@@ -860,19 +860,18 @@ export class BattleScene extends SceneBase {
...
   /**
    * Return the {@linkcode Pokemon} associated with a given ID.
-   * @param pokemonId - The ID whose Pokemon will be retrieved.
-   * @returns The {@linkcode Pokemon} associated with the given id.
-   * Returns `null` if the ID is `undefined` or not present in either party.
-   * @todo Change the `null` to `undefined` and update callers' signatures -
-   * this is weird and causes a lot of random jank
+   * @param pokemonId - The ID whose Pokemon will be retrieved
+   * @returns The {@linkcode Pokemon} associated with the given id,
+   * or `undefined` if no Pokemon with the given id is found.
```
```suggestion
   * or `undefined` if no Pokemon with the given id is found.
   * @todo `pokemonId` shouldn't be able to be `undefined`
```

DayKev (approved):

src/battle-scene.ts
```diff
@@ -858,20 +858,20 @@ export class BattleScene extends SceneBase {
...
+   * @param pokemonId - The PID whose Pokemon will be retrieved
+   * @returns The `Pokemon` associated with the given ID,
+   * or `undefined` if no Pokemon with the given ID is found.
+   * @see {@linkcode Pokemon.id}
    */
-  getPokemonById(pokemonId: number | undefined): Pokemon | null {
+  getPokemonById(pokemonId: number | undefined): Pokemon | undefined {
     if (pokemonId == null) {
-      return null;
+      // biome-ignore lint/nursery/noUselessUndefined: shut up biome for now
```
emdeann: lgtm if you can make this a real comment

emdeann (approved):

PR #6545, opened 2025-09-12
[Refactor] Added `playTween` utility function

Bertie690:
N/A
Callbacks are less wieldly than async and also helps keep track of stuff

Added a utility to play a tween and wait for it to finish.
This is currently unused, but will likely be used in a future PR.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "start:podman": "vite --mode development --host 0.0.0.0 --port $PORT",
     "build": "vite build",
     "build:beta": "vite build --mode beta",
+    "build:dev": "vite build --mode development",
     "preview": "vite preview",
     "test": "vitest run --no-isolate",
     "test:cov": "vitest run --coverage --no-isolate",
@@ -51,8 +52,8 @@
     "typedoc-github-theme": "^0.3.1",
     "typedoc-plugin-coverage": "^4.0.1",
     "typedoc-plugin-mdn-links": "^5.0.9",
-    "typescript": "^5.8.3",
-    "vite": "^7.0.6",
+    "typescript": "^5.9.2",
+    "vite": "^7.0.7",
     "vite-tsconfig-paths": "^5.1.4",
     "vitest": "^3.2.4",
     "vitest-canvas-mock": "^0.3.3"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -19,7 +19,7 @@ importers:
         version: 4.2.0
       i18next:
         specifier: ^24.2.3
-        version: 24.2.3(typescript@5.8.3)
+        version: 24.2.3(typescript@5.9.2)
       i18next-browser-languagedetector:
         specifier: ^8.2.0
         version: 8.2.0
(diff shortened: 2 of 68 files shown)


src/utils/anim-utils.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Plays a Tween animation, resolving once the animation completes.
+ * @param config - The config for a single Tween
+ * @param scene - The {@linkcode SceneBase} on which the Tween plays; default {@linkcode globalScene}
+ * @returns A Promise that resolves once the Tween has been played.
+ *
+ * @privateRemarks
+ * The `config` input should not include an `onComplete` field as that callback is
+ * used to resolve the Promise containing the Tween animation.
+ * However, `config`'s type cannot be changed to something like `Omit<TweenBuilderConfig, "onComplete">`
+ * due to how Phaser interprets `TweenBuilderConfig` inputs.
```
SirzBenjie: Can you explain the "due to how Phaser interprets `TweenBuilderConfig` inputs?

Bertie690: it is a weirdly typed object

Bertie690: Idk maaan that's what innerthunder put

DayKev: If you can suggest better wording then that'd be great.

src/utils/anim-utils.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @param config - The config for a single Tween
+ * @param scene - The {@linkcode SceneBase} on which the Tween plays; default {@linkcode globalScene}
+ * @returns A Promise that resolves once the Tween has been played.
+ *
+ * @privateRemarks
+ * The `config` input should not include an `onComplete` field as that callback is
+ * used to resolve the Promise containing the Tween animation.
+ * However, `config`'s type cannot be changed to something like `Omit<TweenBuilderConfig, "onComplete">`
+ * due to how the type for `TweenBuilderConfig` is defined.
+ */
```
emdeann, line 15: Is there a reason to disallow tweens to use `onComplete` with this helper? Feels like this restriction makes it a lot less useful.

I'm not very familiar with tweens but it seems like you should be able to do something like 
```ts
await new Promise<void>(resolve =>
    scene.tweens.add({
      ...config,
      onComplete: (tween, targets) => {
        config.onComplete?.(tween, targets);
        resolve();
      },
    }),
  );
```

emdeann: I knew I remembered this from somewhere, as a matter of fact I basically implemented this util in `AbilityBar#startTween` like, six months ago

```ts
  public async startTween(config: any, text?: string): Promise<void> {
    this.setVisible(true);
    if (text) {
      this.abilityBarText.setText(text);
    }
    return new Promise(resolve => {
      globalScene.tweens.add({
        ...config,
        onComplete: () => {
          if (config.onComplete) {
            config.onComplete();
          }
          resolve();
        },
      });
    });
  }
```

DayKev: I think the point of doing this is that you're using async/await *instead of* using the callback, so there's no purpose in allowing a callback to be passed in.

emdeann: Ah you're right, async in js still escapes me most of the time. We should update the ability bar to use this structure after merging this

emdeann (approved):

src/utils/anim-utils.ts
DayKev: > I knew I remembered this from somewhere, as a matter of fact I basically implemented this util in `AbilityBar#startTween` like, six months ago
> 
> ```ts
>   public async startTween(config: any, text?: string): Promise<void> {
>     // ...
>   }
> ```

Huh, guess that should be updated then.

emdeann: great minds

PR #6547, opened 2025-09-12
[Misc][Refactor] Add scaffolding for TypedArrays and improve typing on methods related to arrays

SirzBenjie:
None

TypedArrays can be substantially more performant and memory efficient for large arrays.
Their capacity is immutable, and they occupy less memory than a single array with many elements.

There are ways we can leverage typed arrays to reduce the memory size of the codebase, in the future. One example would be to allocate a single large array buffer, and use it for scratchpad. This would reduce allocations while also having the benefit of being a cache hit.

Originally, I tried replacing the IVs and Stats with typedArrays. However, after running a few memory benchmarks on chrome, it seems that it is unwise to use typed arrays for many, small arrays, as the overhead of the object prototype for a typed array is larger than for a normal array, due to it being an object with more elements. However, as the number of elements grows, the size of the elements outweighs this fixed overhead.

- Add many utility types for dealing with typed arrays
- Created `utils/array.ts` for utility methods related to arrays. Moved `coerceArray` into this file.
- Created tests for the new typed array helper methods.
- Added type overrides for the `pick` and `weightedPick` in Phaser's `Math.RND and `method to take an `ArrayLike<T>` instead of `T[]`. This is safe; I looked at [the implementation in Phaser](https://github.com/phaserjs/phaser/blob/964835a3bb2e20bda257e5608474fc99175dd007/src/math/random-data-generator/RandomDataGenerator.js#L361).
- Adjusted some uses of arrays to have readonly to allow for more proper type inferencing and safety.
- Moved `nil` and `Constructor` type exports from `src/utils/common.ts` to `src/@types/common.ts`

## Screenshots/Videos
N/A

## How to test the changes?
Run the new tests.

## Checklist


SirzBenjie:
Ignore the failing lint, will fix when I get a chance.

src/data/abilities/ability.ts
```diff
@@ -2960,7 +2953,7 @@ export class PostSummonHealStatusAbAttr extends PostSummonRemoveEffectAbAttr {
 
   public override canApply({ pokemon }: AbAttrBaseParams): boolean {
     const status = pokemon.status?.effect;
-    return status != null && (this.immuneEffects.length === 0 || this.immuneEffects.includes(status));
+    return status != null && this.immuneEffects.includes(status);
```
DayKev: This isn't the same, if `immuneEffects` has no length that means it should be immune to all status effects.

src/utils/common.ts
```diff
@@ -125,29 +123,25 @@ export function randSeedFloat(): number {
...
- * Shuffle a list using the seeded rng. Utilises the Fisher-Yates algorithm.
+ * Shuffle a list in place using the seeded rng and the Fisher-Yates algorithm.
  * @param items An array of items.
- * @returns A new shuffled array of items.
+ * @returns `items` shuffled in place.
  */
 export function randSeedShuffle<T>(items: T[]): T[] {
-  if (items.length <= 1) {
-    return items;
-  }
```
DayKev, line 146: Why remove the early return?

SirzBenjie: It does nothing; the for loop already won't ever run if length is 0 or 1.

DayKev: Ah, I suppose so.

DayKev (dismissed):

src/@types/typed-arrays.ts
```diff
@@ -0,0 +1,524 @@
...
+ * @privateRemarks
+ * Excludes methods that return a new array (e.g. `subarray`, `slice`, `toReversed`, `toSorted`, `with`) as these cannot be resolved by typescript
+ * @internal
+ */
+interface PartialUint8Array<T extends number> extends Uint8Array {
+  at(index: number): T | undefined; // ES2022
+  entries(): ArrayIterator<[number, T]>;
+  fill(value: T, start?: number, end?: number): this;
+  find(predicate: (value: T, index: number, array: this) => boolean, thisArg?: any): T | undefined;
+  findLast(predicate: (value: number, index: number, array: this) => boolean, thisArg?: any): T | undefined;
```
Bertie690: this should be `value: T` no?

src/@types/typed-arrays.ts
```diff
@@ -0,0 +1,524 @@
...
+}
+
+/**
+ * A partial interface of `Uint32Array` where methods that return the array type have been modified to return a more specific type.
+ *
+ * @privateRemarks
+ * Excludes methods that return a new array (e.g. `subarray`, `slice`, `toReversed`, `toSorted`, `with`) as these cannot be resolved by typescript
+ * @internal
+ */
+interface PartialUint32Array<T extends number> extends Uint32Array {
```
Bertie690, line 274: Can't we do declaration merging or something inside `global.d.ts`?

That's how i added type safe checks for `Function.call`/`apply`

src/battle.ts
```diff
@@ -526,24 +526,25 @@ export class FixedBattleConfig {
...
-    const trainerTypes: TrainerType[] = [];
+    const trainerTypes: TrainerType[] = new Array(trainerPool.length);
 
     globalScene.executeWithSeedOffset(() => {
-      for (const trainerPoolEntry of trainerPool) {
+      for (let i = 0; i < trainerPool.length; i++) {
+        const trainerPoolEntry = trainerPool[i];
         const trainerType = Array.isArray(trainerPoolEntry) ? randSeedItem(trainerPoolEntry) : trainerPoolEntry;
-        trainerTypes.push(trainerType);
+        trainerTypes[i] = trainerType;
```
DayKev, line 547: `for (x; y; z)` is better than `for (x of y)`?

src/phases/battle-phase.ts
```diff
@@ -12,7 +12,7 @@ export abstract class BattlePhase extends Phase {
     const tintSprites = globalScene.currentBattle.trainer.getTintSprites();
     for (let i = 0; i < sprites.length; i++) {
       const visible = !trainerSlot || !i === (trainerSlot === TrainerSlot.TRAINER) || sprites.length < 2;
-      [sprites[i], tintSprites[i]].map(sprite => {
+      [sprites[i], tintSprites[i]].forEach(sprite => {
```
DayKev, line 15: Btw you can remove the redundant `sprites[i].setVisible(visible); ...` that's right below this lol.

src/utils/speed-order.ts
```diff
@@ -16,22 +16,23 @@ interface hasPokemon {
...
   return pokemonList;
 }
 
 /**
+ * Shuffle the list of pokemon *in place*
  * @param pokemonList - The array of Pokemon or objects containing Pokemon
- * @returns The shuffled array
+ * @returns The same array instance that was passed in, shuffled.
  */
 function shufflePokemonList<T extends Pokemon | hasPokemon>(pokemonList: T[]): T[] {
```
DayKev, line 31: If this is in-place now then it shouldn't return anything.

src/utils/speed-order.ts
```diff
@@ -16,22 +16,23 @@ interface hasPokemon {
  * @returns The sorted array of {@linkcode Pokemon}
  */
 export function sortInSpeedOrder<T extends Pokemon | hasPokemon>(pokemonList: T[], shuffleFirst = true): T[] {
```
DayKev, line 18: Similar for this one, it doesn't seem like it needs to return the list anymore since it's an in-place mutation.

src/utils/common.ts
```diff
@@ -125,29 +123,25 @@ export function randSeedFloat(): number {
...
 }
 
 /**
- * Shuffle a list using the seeded rng. Utilises the Fisher-Yates algorithm.
+ * Shuffle a list in place using the seeded rng and the Fisher-Yates algorithm.
  * @param items An array of items.
- * @returns A new shuffled array of items.
+ * @returns `items` shuffled in place.
  */
 export function randSeedShuffle<T>(items: T[]): T[] {
```
DayKev, line 141: Is there a point to returning the array if it's shuffled in-place now?

SirzBenjie: Yea, ease of use. Also so you can directly pass an array while shuffling it at the same time for methods that want it.
No harm.

src/utils/speed-order.ts
SirzBenjie: I disagree. Shuffle methods commonly return the array being operated on. This allows for them to be used at callsites as part of an expression _or_ a standalone statement.

src/battle.ts
SirzBenjie: They're nearly identical. The reason I did this was to use the index so we could pre-allocate the trainerTypes array.

src/utils/common.ts
```diff
@@ -125,29 +123,25 @@ export function randSeedFloat(): number {
...
+export function randSeedItem<T>(items: ArrayLike<T>): T {
   return items.length === 1 ? items[0] : Phaser.Math.RND.pick(items);
 }
 
 /**
- * Shuffle a list using the seeded rng. Utilises the Fisher-Yates algorithm.
+ * Shuffle a list in place using the seeded rng and the Fisher-Yates algorithm.
  * @param items An array of items.
- * @returns A new shuffled array of items.
+ * @returns `items` shuffled in place.
```
Bertie690:
```suggestion
 * @returns A reference to the `items` array, now shuffled in place.
```

Bertie690: These changes will conflict with the priority queue docs pr fwiw

src/utils/speed-order.ts
```diff
@@ -16,22 +16,23 @@ interface hasPokemon {
...
-  pokemonList = shuffleFirst ? shufflePokemonList(pokemonList) : pokemonList;
+  if (shuffleFirst) {
+    shufflePokemonList(pokemonList);
```
Bertie690, line 20: This is (also) gonna conflict with my pr

Also i know you said shuffle methods usually return a  reference to the array but I explicitly don't want callers (i.e. priority queues) to be assigning the result of this function

test/utils/array.test.ts
```diff
@@ -0,0 +1,279 @@
...
+  describe("isTypedArray", () => {
+    it.each([
+      ["Int8Array", Int8Array],
+      ["Uint8ClampedArray", Uint8ClampedArray],
+      ["Uint16Array", Uint16Array],
+      ["Uint32Array", Uint32Array],
+      ["Float64Array", Float64Array],
+      ["Float32Array", Float32Array],
+    ])("returns true for %s", (_, ArrayType) => {
+      if (typeof ArrayType !== "undefined") {
```
Bertie690: Why is this check needed?

Bertie690 (changes requested):
Forgot to submit this a while ago, so here goes

src/@types/typed-arrays.ts
SirzBenjie: Yes.

Bertie690: Also, `thisArg` should not be `any` and should be typed with an inferred type parameter for the function callback's `thisType`

SirzBenjie: I am literally mirroring the signature for `findLast` on the TypedArray itself, which uses `any` here.

src/@types/typed-arrays.ts
SirzBenjie: No, we most certainly shouldn't be doing that. I don't want to mess with core types.

src/@types/typed-arrays.ts
Bertie690: I swear there was something in some `d.ts` files somewhere that did that... 
oh well

PR #6550, opened 2025-09-12
[Deps] Update biome to 2.2.4

SirzBenjie:
None

Biome version bump is nice. Also, the faster `noImportCycles` rule is something I want for local dev

- Bumped biome to 2.2.4
- Set the `useReadonlyClassProperties`'s `checkAllProperties` option to false, as this was causing too many false positives in my editor
- Adjusted the parameter count for the `useMaxParams` lint to be 7, as discussed
- Enabled the `noMisusedPromises` with `info` severity
- Enabled the `noImportCycles`

## Checklist

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.2.3/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.2.4/schema.json",
   "vcs": {
     "enabled": true,
     "clientKind": "git",
@@ -212,10 +212,14 @@
         "noUselessUndefined": "error",
         "useMaxParams": {
           "level": "info", // TODO: Change to "error"... eventually...
-          "options": { "max": 7 } // A lot of stuff has a few params, but
+          "options": { "max": 7 }
         },
         "noShadow": "warn", // TODO: refactor and make "error"
-        "noNonNullAssertedOptionalChain": "warn" // TODO: refactor and make "error"
+        "noNonNullAssertedOptionalChain": "warn", // TODO: refactor and make "error"
+        "noDuplicateDependencies": "error",
+        // TODO:  this if performance does not suffer with biome 2.2.4
+        // "noImportCycles": "error",
+        "noMisusedPromises": "error"
       }
     }
   },
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -96,10 +96,12 @@
         },
         "useThrowOnlyError": "error",
         "useTrimStartEnd": "error",
-        "useReadonlyClassProperties": {
-          "level": "info", // TODO: Graduate to error eventually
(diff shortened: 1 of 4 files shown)


DayKev:
> * Enabled the `noMisusedPromises` (which already passes)

Apparently it doesn't already pass lol.

biome.jsonc
```diff
@@ -96,10 +96,12 @@
...
-        "useReadonlyClassProperties": {
-          "level": "info", // TODO: Graduate to error eventually
-          "options": { "checkAllProperties": true }
-        },
+        // Todo: This rule has issues properly detecting reads across files.
+        // Enable once it becomes more mature
+        // "useReadonlyClassProperties": {
+        //   "level": "info", // TODO: Graduate to error eventually
+        //   "options": { "checkAllProperties": true }
+        // },
```
DayKev: Instead of disabling it entirely, you could just have it stop checking non-`private` fields, since those "can't" be used in other files.

biome.jsonc
```diff
@@ -209,11 +211,15 @@
...
-          "level": "warn", // TODO: Change to "error"... eventually...
-          "options": { "max": 4 } // A lot of stuff has a few params, but
+          "level": "info", // TODO: Change to "error"... eventually...
+          "options": { "max": 7 }
         },
         "noShadow": "warn", // TODO: refactor and make "error"
-        "noNonNullAssertedOptionalChain": "warn" // TODO: refactor and make "error"
+        "noNonNullAssertedOptionalChain": "warn", // TODO: refactor and make "error"
+        "noDuplicateDependencies": "error",
+        // TODO:  this if performance does not suffer with biome 2.2.4
```
DayKev:
```suggestion
        // TODO: enable this if performance does not suffer with biome 2.2.4
```

biome.jsonc
SirzBenjie: Good point!

biome.jsonc
```diff
@@ -209,11 +211,15 @@
       "nursery": {
         "noUselessUndefined": "error",
         "useMaxParams": {
-          "level": "warn", // TODO: Change to "error"... eventually...
-          "options": { "max": 4 } // A lot of stuff has a few params, but
+          "level": "info", // TODO: Change to "error"... eventually...
+          "options": { "max": 7 }
```
Bertie690, line 215:
```suggestion
          "options": { "max": 7 } // 7 is a decent middle ground given we can't specify ">=X optional parameters"
```

Bertie690 (approved):

DayKev (approved):

Madmadness65 (approved):

biome.jsonc
SirzBenjie: Ehh, I don't want to discard reviews to put that in. It's OK as is, we're all privy to the convo

PR #6552, opened 2025-09-13
[Balance] Trainer boss Pokémon no longer gain a stat boost when a boss segment is broken

SirzBenjie:
Bosses that have a trainer will no longer receive a random stat boost when a boss segment breaks.

Requested by balance

Inside `EnemyPokemon`'s damage method, the call to `handleBossSegmentCleared` is additionally predicated on `!this.hasTrainer()`. Note that, despite the method's poor name, it is only responsible for providing the stat boost). The boss segments are still cleared inside the damage method.

## How to test the changes?
Go into a fight against a trainer that has a boss pokemon. Maybe Ivy.

## Checklist


emdeann (approved):
Good from dev (duh)

damocleas (approved):

PR #6554, opened 2025-09-14
Remove `ARENA_TINT_OVERRIDE` + minor cleanup of arena tint code

Bertie690:
N/A
To put it bluntly, having an override to change the _TINT_ that is displayed based on time of day is about as useless as it sounds.

Also `return [98, 48, 73].map(c => Math.round((c + 128) / 2)) as [number, number, number]` is just... no... please

Removed useless override
Cleaned up functions to not have default-only switch cases and mapping constants
Added 0.5 docs

## Screenshots/Videos
N/A

## Checklist

### DIFF
--- a/src/field/arena.ts
+++ b/src/field/arena.ts
@@ -537,22 +537,20 @@ export class Arena {
     }
   }
 
-  getTimeOfDay(): TimeOfDay {
+  public getTimeOfDay(): TimeOfDay {
     switch (this.biomeType) {
       case BiomeId.ABYSS:
         return TimeOfDay.NIGHT;
     }
 
-    const waveCycle = ((globalScene.currentBattle?.waveIndex || 0) + globalScene.waveCycleOffset) % 40;
+    const waveCycle = ((globalScene.currentBattle?.waveIndex ?? 0) + globalScene.waveCycleOffset) % 40;
 
     if (waveCycle < 15) {
       return TimeOfDay.DAY;
     }
-
     if (waveCycle < 20) {
       return TimeOfDay.DUSK;
     }
-
     if (waveCycle < 35) {
       return TimeOfDay.NIGHT;
     }
@@ -560,6 +558,10 @@ export class Arena {
     return TimeOfDay.DAWN;
   }
 
+  /**
+   * @returns Whether the current biome takes place "outdoors"
+   * (for the purposes of time of day tints)
+   */
   isOutside(): boolean {
     switch (this.biomeType) {
       case BiomeId.SEABED:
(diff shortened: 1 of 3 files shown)


Madmadness65:
Would it make more sense to just replace the arena tint override with an actual time of day override?

Bertie690:
> Would it make more sense to just replace the arena tint override with an actual time of day override?

Done

SirzBenjie (dismissed):
Why do we have this override if no tests were even using it?

Fontbane:
> Why do we have this override if no tests were even using it?

It's used for manual testing/playtesting

DayKev:
> Why do we have this override if no tests were even using it?

It was used for viewing sprites in-game / getting screenshots.

Bertie690:
> You still need to fix the override, it doesn't do anything atm.
The weird thing is, IDK how other parts of the code would access time of day other than this method
Will look into it later once my laptop actually works

DayKev (approved):

src/field/arena.ts
```diff
@@ -604,23 +594,14 @@ export class Arena {
...
-    }
     if (!this.isOutside()) {
       return [0, 0, 0];
     }
 
-    switch (this.biomeType) {
-      default:
-        return [98, 48, 73].map(c => Math.round((c + 128) / 2)) as [number, number, number];
-    }
+    return [113, 88, 101];
```
bavalpey, line 601: I did this in my array PR. please revert.

src/pipelines/field-sprite.ts
```diff
@@ -33,9 +34,13 @@ export class FieldSpritePipeline extends Phaser.Renderer.WebGL.Pipelines.MultiPi
...
-    const time = globalScene.currentBattle?.waveIndex
-      ? ((globalScene.currentBattle.waveIndex + globalScene.waveCycleOffset) % 40) / 40 // ((new Date().getSeconds() * 1000 + new Date().getMilliseconds()) % 10000) / 10000
-      : getCurrentTime();
+    // TODO: Refactor this to respect time of day calcs and not hardcode it
+    const time =
+      Overrides.TIME_OF_DAY_OVERRIDE !== null
+        ? Overrides.TIME_OF_DAY_OVERRIDE / 4
+        : globalScene.currentBattle?.waveIndex
+          ? ((globalScene.currentBattle.waveIndex + globalScene.waveCycleOffset) % 40) / 40 // ((new Date().getSeconds() * 1000 + new Date().getMilliseconds()) % 10000) / 10000
+          : getCurrentTime();
```
SirzBenjie: If you're refactoring anyway, why not get rid of the nested ternary?

src/field/arena.ts
SirzBenjie, line 608: I did this in my array PR. please revert.

src/pipelines/field-sprite.ts
Bertie690: That would require changing the GLSL code.
I do not want to touch the GLSL code.

src/field/arena.ts
Bertie690: if 2 prs make the exact same change will it still conflict?

DayKev: Yes.

Bertie690: ok will revrrt

src/pipelines/field-sprite.ts
Bertie690: -# i decided to change the glsl code

SirzBenjie (dismissed):

src/@types/sprite-types.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Tuple type representing a set of RGB color values.
+ * Values should be in integer form (inside the interval `[0, 255]`).
+ */
+export type RGBArray = [r: number, b: number, g: number];
```
DayKev:
```suggestion
export type RGBArray = [r: number, g: number, b: number];
```

DayKev (approved):

PR #6555, opened 2025-09-14
[Refactor] Remove `DEBUG_RNG`

Bertie690:
N/A
I really dislike debug constants.

Yes, seeing the game's RNG rolls could be useful... except it isn't since we already log those to console
Removed debug constant
## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -152,8 +152,6 @@ import Phaser from "phaser";
 import SoundFade from "phaser3-rex-plugins/plugins/soundfade";
 import type UIPlugin from "phaser3-rex-plugins/templates/ui/ui-plugin";
 
-const DEBUG_RNG = false;
-
 export interface PokeballCounts {
   [pb: string]: number;
 }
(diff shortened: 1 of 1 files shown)


DayKev:
> Yes, seeing the game's RNG rolls could be useful... except it isn't since we already log those to console

Wdym? They're not logged to console unless manual console logs are added (or `DEBUG_RNG` is set to `true`).

SirzBenjie:
Hear me out -- we should actually *keep* the function that was declared, remove the `DEBUG_RNG` variable, and replace it with a constant defined in the `.env` files.

SirzBenjie:
What is with the removal of the rngcounter?

Bertie690:
> What is with the removal of the rngcounter?

AFAIK, it was only ever read during `DEBUG_RNG` logging and was solely assigned to otherwise.
Probably was used for something else aeons ago but fell out of favor

Bertie690:
> Hear me out -- we should actually keep the function that was declared, remove the DEBUG_RNG variable, and replace it with a constant defined in the .env files.

Counterpoint: We do a LOT of RNG calls to the point that logging each one will flood the console with log statements.
The function would have to be basically reworked from the ground up just to be _vaguely usable_, and idk if we'd even even use it anyhow

Bertie690:
Like for instance, who actually uses the `VITE_I18N_DEBUG` env var?
I'm not saying we should delete it but like imo the utility is significantly less than worth the effort to maintain.

DayKev:
> Like for instance, who actually uses the `VITE_I18N_DEBUG` env var? I'm not saying we should delete it but like imo the utility is significantly less than worth the effort to maintain.

There's zero effort needed to maintain it atm though.

DayKev (approved):

PR #6556, opened 2025-09-15
refactor: split up `BattleScene.newBattle` into sub-methods

Bertie690:
N/A (maybe slightly faster load times, but i doubt it)

That function was a warcrime and a quarter

100+ cognitive complexity is bad

Split up `BattleScene.newBattle` into a series of 5-ish sub helpers that handle the 3 main cases of initializing data (fixed battle, existing session, new battle) and post processing on arena reset.
Each has helper functions where necessary to make the code easier to parse.

> [!NOTE]
> A few notable things for reviewers:
> - I have tried **very hard** to make this a purely semantic rewrite with no functionality changes.
> - The new code packages and passes values with a POJO that "subs in" for the `newDouble`, `newWaveIndex` and similar variables in the scope of the original function.
> - Certain checks have had their order reversed in code to match the previous imperative style. For instance, the double battle override check (which previously overwrote the value of `newDouble` _after_ everything else had applied) is now evaluated _before_ all other effects for simplicity. 
>   This is equivalent functionality-wise (aside from making things a bit more efficient), but does make the diff a bit (read: a LOT) messier.

## Screenshots/Videos
N/A
## How to test the changes?
check the zillion integration tests - this stuff is so ingrained in our codebase's performance that any deviations will break existing tests. (yes, even the override-related ones)

## Checklist

### DIFF
--- a/src/battle-scene.ts
+++ b/src/battle-scene.ts
@@ -1493,8 +1493,7 @@ export class BattleScene extends SceneBase {
    * @returns Whether the battle should be a double battle.
    */
   private checkIsDouble({ double: forcedDouble, battleType, waveIndex, trainer }: NewBattleConstructedProps): boolean {
-    // Disallow using double battle overrides on trainer waves (need `RANDOM_TRAINER_OVERRIDE` instead)
-    // TODO: Print a warning here if this results in the override being ignored
+    // TODO: enforce using the proper override depending on whether it's a trainer or a wild battle
     const doubleBattleOverride = this.doCheckDoubleOverride(waveIndex);
     if (doubleBattleOverride != null) {
       return doubleBattleOverride;


src/battle-scene.ts
```diff
@@ -1275,239 +1288,323 @@ export class BattleScene extends SceneBase {
...
-  ): Battle {
-    const _startingWave = Overrides.STARTING_WAVE_OVERRIDE || startingWave;
-    const newWaveIndex = waveIndex || (this.currentBattle?.waveIndex || _startingWave - 1) + 1;
-    let newDouble: boolean | undefined;
-    let newBattleType: BattleType;
-    let newTrainer: Trainer | undefined;
+  /**
+   * Create and initialize a new battle.
+   * @param fromSession - The {@linkcode SessionSaveData} being used to seed the battle.
+   * Should be omitted if not loading a new save file.
```
```suggestion
   * Should be omitted if not loading a save file.
```

src/battle-scene.ts
```diff
@@ -1275,239 +1288,323 @@ export class BattleScene extends SceneBase {
...
+      this.handleFixedBattle(resolved);
+    } else if (fromSession) {
+      this.handleSavedBattle(resolved, props);
+    } else {
+      this.handleNonFixedBattle(resolved);
+    }
 
-    let battleConfig: FixedBattleConfig | null = null;
+    if (resolved.battleType == null) {
+      throw new Error("Whoopsie! I guess my type checks were wrong");
```
DayKev: Can we please not have these kinds of error messages, especially if they can possibly be public-facing.

Bertie690: i will change it dw

src/data/balance/misc.ts
```diff
@@ -0,0 +1,2 @@
+/** The default wave that new runs start at. */
+export const startingWave = 1;
```
DayKev: Global constants should be in UPPER_SNAKE_CASE.

Seeing this has also "inspired" me to finally fix this mess, no more random files with a single exported constant in them...

src/utils/common.ts
```diff
@@ -474,13 +474,17 @@ export function getLocalizedSpriteKey(baseKey: string) {
...
+ * @param min - The minimum value
+ * @param max - The maximum value
+ * @param exclusive - Whether to exclusively check boundedness; default `false` (inclusive)
+ * @returns Whether `num` is no less than `min` and no greater than `max`.
  */
-export function isBetween(num: number, min: number, max: number): boolean {
+export function isBetween(num: number, min: number, max: number, exclusive = false): boolean {
+  if (exclusive) {
+    return min < num && num < max;
+  }
```
DayKev: This change is unnecessary/redundant. MEs are already constrained from happening on waves 10 and 180, there's no reason to change a utility function just so it can check something that is already checked for. You could also just do `minWave + 1` `maxWave - 1` if you wanted to have the ME checks be extra redundant, instead of modifying a utility function.

src/overrides.ts
```diff
@@ -67,14 +67,25 @@ class DefaultOverrides {
    *
    * If `"single"`, set every non-trainer battle to be a single battle.
    *
-   * If `"double"`, set every battle (including trainer battles) to be a double battle.
+   * If `"double"`, set every battle (_including_ trainer battles that are normally singles-only_)
+   * to be a double battle.
    *
-   * If `"even-doubles"`, follow the `"double"` rule on even wave numbers, and follow the `"single"` rule on odd wave numbers.
+   * If `"even-doubles"`, follow the `"double"` rule on even wave numbers,
+   * or the `"single"` rule on odd wave numbers.
```
DayKev:
```suggestion
   * and the `"single"` rule on odd wave numbers.
```

src/overrides.ts
```diff
@@ -67,14 +67,25 @@ class DefaultOverrides {
...
    *
-   * If `"odd-doubles"`, follow the `"double"` rule on odd wave numbers, and follow the `"single"` rule on even wave numbers.
+   * If `"odd-doubles"`, follow the `"double"` rule on odd wave numbers,
+   * or the `"single"` rule on even wave numbers.
```
DayKev:
```suggestion
   * and the `"single"` rule on even wave numbers.
```

src/overrides.ts
```diff
@@ -67,14 +67,25 @@ class DefaultOverrides {
...
+   * @defaultValue `null`
+   * @privateRemarks
+   * Does not work on trainer battles - use {@linkcode RANDOM_TRAINER_OVERRIDE} instead
```
DayKev: This is contradicted by the `"(_including_ trainer battles that are normally singles-only_)"` comment above.

src/overrides.ts
```diff
@@ -67,14 +67,25 @@ class DefaultOverrides {
...
    */
   readonly BATTLE_STYLE_OVERRIDE: BattleStyle | null = null;
-  readonly STARTING_WAVE_OVERRIDE: number = 0;
+  /**
+   * If present and non-null, will override the starting wave # when starting a new run.
+   * Should never be set to a negative value.
+   * @defaultValue `null`
+   */
+  readonly STARTING_WAVE_OVERRIDE: number | null = null;
```
DayKev, line 104: If you left this as a `number` only it would be easier to do validation, you could do `const startingWave = isBetween(Overrides.STARTING_WAVE_OVERRIDE, 1, 200) ? Overrides.STARTING_WAVE_OVERRIDE : STARTING_WAVE;`

src/@types/new-battle-props.ts
```diff
@@ -0,0 +1,79 @@
+/**
+ * A collection of types/interfaces used for {@linkcode BattleScene.newBattle} and associated
+ * sub-methods.
+ *
+ * Types are listed in order of appearance in the function.
+ * @module
+ */
+
+// biome-ignore lint/correctness/noUnusedImports: TSDoc
```
DayKev:
```suggestion
```

src/overrides.ts
Bertie690: That would also make disabling it weird - you'd have to do `startingWave = 0`

src/utils/common.ts
Bertie690: Where are the duplicated checks?
Having a hard time finding em

src/battle-scene.ts
```diff
@@ -3512,57 +3663,10 @@ export class BattleScene extends SceneBase {
       && battleType === BattleType.WILD
       && !this.gameMode.isBoss(waveIndex)
       && waveIndex % 10 !== 1
-      && waveIndex < highestMysteryEncounterWave
-      && waveIndex > lowestMysteryEncounterWave
+      && isBetween(waveIndex, lowestMysteryEncounterWave, highestMysteryEncounterWave, true)
```
DayKev: Here, X0 waves (10, 180) are covered by checking if it's a boss wave via `isBoss(waveIndex)`.

src/@types/new-battle-props.ts
```diff
@@ -0,0 +1,78 @@
...
+   */
+  mysteryEncounterType?: MysteryEncounterType;
+  /**
+   * The wave number of the NEW wave to spawn.
+   * Will always be >=1 (barring save data corruption).
+   */
+  waveIndex: number;
+  /**
+   * Whether the battle is a double battle.
+   * Always `false` when an ME is spawned.
```
SirzBenjie:
```suggestion
   * Always `false` when an ME is spawned.
```

Why is this the case? How does clowning around work? A better way of phrasing this, I think, would be something along the lines of "⚠️ Mystery encounters do not rely on this property and are set independently."

src/data/balance/misc.ts
```diff
@@ -0,0 +1,2 @@
+/** The default wave that new runs start at. */
+export const STARTING_WAVE = 1;
```
SirzBenjie, line 3: Does this change somewhere? o.o

Or is it for overrides? Perhaps removing this isn't in the purview of this PR, but I think baking in an assumption that the starting wave is 1 is fine, and anything that cares can just check the override.

src/overrides.ts
```diff
@@ -73,14 +73,26 @@ class DefaultOverrides {
...
+   * If `"even-doubles"`, follow the `"double"` rule on even wave numbers,
+   * and the `"single"` rule on odd wave numbers.
    *
-   * If `"odd-doubles"`, follow the `"double"` rule on odd wave numbers, and follow the `"single"` rule on even wave numbers.
+   * If `"odd-doubles"`, follow the `"double"` rule on odd wave numbers,
+   * and the `"single"` rule on even wave numbers.
+   * @defaultValue `null`
+   * @privateRemarks
+   * Prefer using {@linkcode RANDOM_TRAINER_OVERRIDE} to override trainer battles.
+   * This override's ability to override trainer battles is inconsistent and may be removed in a future PR.
```
SirzBenjie: I didn't realize this was inconsistent. Can you elaborate on the issues with this?

SirzBenjie (dismissed):
This is *just* a code approval, I didn't checkout the code and run it myself.

src/data/balance/misc.ts
Bertie690: I agree but i do digress.

src/overrides.ts
Bertie690: IIRC it forces the BATTLE to be double but doesn't force the TRAINER itself to be a double variant.

src/battle-scene.ts
```diff
@@ -1254,244 +1267,345 @@ export class BattleScene extends SceneBase {
...
-        this.currentBattle = new Battle(this.gameMode, newWaveIndex, newBattleType, newTrainer, newDouble);
-      },
-      newWaveIndex << 3,
-      this.waveSeed,
-    );
-    this.currentBattle.incrementTurn();
+    // Edge cases
+    if (
+      // Wave 1 doubles cause crashes
+      // TODO: Investigate why this occurs and fix it for good - this is a constant PITA inside testing and local dev
```
DayKev: That's only when forcing trainer battles via the override:
```ts
        // Forcing a double battle on wave 1 causes a bug where only one enemy is sent out,
        // making it impossible to complete the fight without a reload
        const overrideVariant =
          Overrides.RANDOM_TRAINER_OVERRIDE?.trainerVariant === TrainerVariant.DOUBLE
          && (!hasDouble || newWaveIndex <= 1)
            ? TrainerVariant.DEFAULT
            : Overrides.RANDOM_TRAINER_OVERRIDE?.trainerVariant;

        const variant =
          overrideVariant
          ?? (doubleTrainer ? TrainerVariant.DOUBLE : randSeedInt(2) ? TrainerVariant.FEMALE : TrainerVariant.DEFAULT);
```

Using the double battle override on a wild battle on wave 1 doesn't crash (which I just confirmed).

test/tests/abilities/intimidate.test.ts
```diff
@@ -63,15 +63,17 @@ describe("Abilities - Intimidate", () => {
...
 
     await game.reload.reloadSession();
 
+    expect(game.field.getPlayerPokemon()).toHaveAbilityApplied(AbilityId.INTIMIDATE);
     expect(game.field.getEnemyPokemon()).toHaveStatStage(Stat.ATK, -1);
   });
 
   it("should lower ATK of all opponents in a double battle", async () => {
-    game.override.battleStyle("double");
+    game.override.startingWave(2).battleStyle("double");
```
DayKev:
```suggestion
    game.override.battleStyle("double");
```
This isn't necessary (because preventing wild double battles on wave 1 isn't necessary, as mentioned in the other comment).

test/utils/game-manager-utils.ts
```diff
@@ -111,5 +111,10 @@ export function initSceneWithoutEncounterPhase(scene: BattleScene, speciesIds?:
     scene.getPlayerParty().push(starterPokemon);
   });
 
-  scene.currentBattle = new Battle(getGameMode(GameModes.CLASSIC), 5, BattleType.WILD, undefined, false);
+  scene.currentBattle = new Battle(getGameMode(GameModes.CLASSIC), {
+    waveIndex: 5,
+    battleType: BattleType.WILD,
+    trainer: undefined,
```
DayKev:
```suggestion
```
That's an optional property.

src/battle-scene.ts
```diff
@@ -164,6 +169,12 @@ export interface InfoToggle {
   isActive(): boolean;
 }
 
+/**
+ * The `BattleScene` is the primary scene for the game.
+ * Despite its name, it handles _everything_ other than initial asset loading,
+ * up to and including title menuing and settings handling.
+ * @todo Breakup into multiple scenes
+ */
```
```suggestion
 */
 // TODO: Breakup into multiple scenes if possible/practical
```

src/battle-scene.ts
```diff
@@ -1254,244 +1267,345 @@ export class BattleScene extends SceneBase {
...
+      !this.gameMode.hasTrainers || Overrides.DISABLE_STANDARD_TRAINERS_OVERRIDE
+        ? BattleType.WILD
+        : (Overrides.BATTLE_TYPE_OVERRIDE
+          ?? (this.gameMode.isWaveTrainer(waveIndex, this.arena) ? BattleType.TRAINER : BattleType.WILD));
+
+    // Check for mystery encounter
+    // Can only occur in place of a standard (non-boss) wild battle, waves 10-180
+    // NB: battle type checks are offloaded to `isWaveMysteryEncounter`
+    // TODO: This means MEs can generate when the override is set to `BattleType.WILD`
+    if (!Overrides.BATTLE_TYPE_OVERRIDE && this.isWaveMysteryEncounter(resolved.battleType, waveIndex)) {
```
DayKev, line 1445: Shouldn't this be `Overrides.BATTLE_TYPE_OVERRIDE == null`?

src/battle-scene.ts
```diff
@@ -1254,244 +1267,345 @@ export class BattleScene extends SceneBase {
...
+  /**
+   * Create and initialize a new battle.
+   * @param fromSession - The {@linkcode SessionSaveData} being used to seed the battle.
+   * Should be omitted if not loading an existing save file.
+   * @returns The newly created `Battle` instance.
+   */
+  public newBattle(fromSession?: SessionSaveData): Battle {
+    const props = this.getNewBattleProps(fromSession);
+    const { waveIndex, mysteryEncounterType } = props;
+    const resolved: NewBattleInitialProps = { waveIndex, mysteryEncounterType };
```
DayKev, line 1280: Why is the variable with name `resolved` typed as `NewBattleInitialProps` and not `NewBattleResolvedProps`?

src/battle-scene.ts
```diff
@@ -1254,244 +1268,345 @@ export class BattleScene extends SceneBase {
...
+      case BattleType.WILD:
+        fixedDouble = fromSession.enemyParty.length > 1;
+        break;
+      case BattleType.TRAINER: {
+        const config = trainerConfigs[trainerData.trainerType];
+        // failsafe for corrupt saves (such as due to enum shifting) - don't force a double battle if the trainer type
+        // cannot appear as one.
+        // The code inside `handleSavedBattle` will override the trainer's variant if this results in incompatibilities,
+        // since this is only a problem for parsed save data (and is ignored for fixed battles)
+        fixedDouble = config.doubleOnly || (config.hasDouble && trainerData.variant === TrainerVariant.DOUBLE);
```
DayKev: You broke the failsafe. For reference, this is currently at the very start of `BattleScene#newBattle` (after Jimmy's change):
```ts
    if (
      trainerData?.variant === TrainerVariant.DOUBLE
      && !trainerConfigs[trainerData.trainerType].hasDouble
      && !trainerConfigs[trainerData.trainerType].doubleOnly
    ) {
      trainerData.variant = TrainerVariant.DEFAULT;
      double = false;
    }
```
Relevant PRs (they have session saves to test with):
- #7089
- #7101
- #7110

src/battle-scene.ts
```diff
@@ -1254,244 +1268,345 @@ export class BattleScene extends SceneBase {
...
+   */
+  private handleSavedBattle(resolved: NewBattleInitialProps, props: NewBattleSavedProps): void {
+    resolved.battleType = props.battleType;
+    resolved.double = props.double;
+    if (!resolved.double && props.trainerData?.variant === TrainerVariant.DOUBLE) {
+      console.warn(
+        "Save data indicates a double battle but this trainer has no doubles variant!\nForcing battle to default variant to avoid crashes.",
+      );
+      props.trainerData.variant = TrainerVariant.DEFAULT;
+    }
```
DayKev: See [other comment](https://github.com/pagefaultgames/pokerogue/pull/6556#discussion_r2852801376) about the failsafe.

src/battle-scene.ts
```diff
@@ -3499,10 +3615,60 @@ export class BattleScene extends SceneBase {
...
   /**
    * Returns if a wave COULD spawn a {@linkcode MysteryEncounter}.
-   * Even if returns `true`, does not guarantee that a wave will actually be a ME.
-   * That check is made in {@linkcode BattleScene.isWaveMysteryEncounter} instead.
+   * @param battleType - The {@linkcode BattleType} of the newly created battle
+   * @param waveIndex - The wave number of the newly spawned wave
+   * @returns Whether an ME can legally spawn on the given wave.
+   * @see {@linkcode BattleScene.isWaveMysteryEncounter} - Function that rolls for ME creation on new wave start
    */
   isMysteryEncounterValidForWave(battleType: BattleType, waveIndex: number): boolean {
```
```suggestion
  public isMysteryEncounterValidForWave(battleType: BattleType, waveIndex: number): boolean {
```
Moving this puts it further away from where it's called lol.

src/battle-scene.ts
Bertie690: Because we continually mutate it throughout the function.
i guess i could rename it but making names is haaard man

src/battle-scene.ts
Bertie690: Did you test that it indeed crashes when loading the aforementioned saves?

src/battle-scene.ts
Bertie690: Yes - read the comment about this working on wild overridden battles.

That said, I _intentionally_ kept this in because I didn't want to change functionality in this PR.

src/battle-scene.ts
Bertie690: ...
wtf this is convoluted

test/utils/game-manager-utils.ts
Bertie690: explicit > implicit?

src/battle-scene.ts
DayKev: Yes, that's why I made the comment.

test/utils/game-manager-utils.ts
DayKev: No. Optional properties are optional. It should not be an optional property if we are expected to provide it at every call site regardless of whether it's relevant, nor should it be inconsistently provided when it's not necessary; either always or never, and in this case I don't see any reason for it to be "always".

src/battle-scene.ts
Bertie690: _groans_ i was trying to make it less overtly jank.

i guess the simple version works *sigh*

DayKev: I would have made a save migrator instead but Benjie's rework isn't done yet.

DayKev: If you can make it self contained at the start of the method still so that it can be easily removed once a save migrator can be made, that'd be perfect.

src/battle-scene.ts
```diff
@@ -1290,29 +1291,6 @@ export class BattleScene extends SceneBase {
...
-
-    // TODO: Is this even needed?
-    if (lastBattle?.double && !resolved.double) {
-      this.phaseManager.tryRemovePhase("SwitchPhase");
-      // TODO: We already do this later in the function
-      for (const p of this.getPlayerField()) {
-        p.lapseTag(BattlerTagType.COMMANDED);
-      }
-    }
-
```
DayKev: I feel like you probably didn't mean to remove this chunk of code? For one, now `checkIsDouble` is never called.

Bertie690: RIP.
Merge conflict demon strikes again...

src/battle-scene.ts
```diff
@@ -1251,248 +1265,345 @@ export class BattleScene extends SceneBase {
...
 
-        const variant =
-          overrideVariant
-          ?? (doubleTrainer ? TrainerVariant.DOUBLE : randSeedInt(2) ? TrainerVariant.FEMALE : TrainerVariant.DEFAULT);
+    this.executeWithSeedOffset(
+      () => {
+        // NB: Type assertion is fine as resolved should always be populated at this point
+        this.currentBattle = new Battle(this.gameMode, resolved as NewBattleResolvedProps);
+      },
+      waveIndex << 3, // TODO: Why use this specific bitshift?
```
DayKev, line 1322: Who knows https://github.com/pagefaultgames/pokerogue/commit/366e3e5120292ea2a96009113f78c95254dda9df
Seems like it was just chosen at random.

src/battle-scene.ts
```diff
@@ -1251,248 +1265,345 @@ export class BattleScene extends SceneBase {
...
-      newWaveIndex << 3,
-      this.waveSeed,
-    );
-    this.currentBattle.incrementTurn();
+    // Edge cases
+    if (
+      // Wave 1 doubles cause crashes
+      // TODO: Investigate why this occurs and fix it for good - this is a constant PITA inside testing and local dev
+      waveIndex === 1
+      || this.gameMode.isWaveFinal(waveIndex) // Endless bosses and classic mode finales are never double battles
```
```suggestion
      this.gameMode.isWaveFinal(waveIndex) // Endless bosses and classic mode finales are never double battles
```

DayKev:
Okay apparently the whole "double trainer softlock" on wave 1 doesn't happen anymore. There was probably some nonsense desync happening. I just tested a bunch with the changes from the suggestions and whether I override the trainer type to be double or just the battle to be a trainer battle without specifying whether it should be doubles or not, there's no softlock if there's a doubles trainer battle on wave 1.

src/battle-scene.ts
```diff
@@ -1251,248 +1265,345 @@ export class BattleScene extends SceneBase {
...
+      // TODO: Review this
+      !config.hasDouble
+      || (config.trainerTypeDouble && ![TrainerType.TATE, TrainerType.LIZA].includes(trainerType))
+    ) {
+      doubleTrainer = false;
+    } else {
+      // Forcing a double battle on wave 1 causes a bug where only one enemy is sent out,
+      // making it impossible to complete the fight without a reload
+      // TODO: We duplicate the "wave 1" check with `checkIsDouble`
+      doubleTrainer = waveIndex > 1 && randSeedInt(this.getDoubleBattleChance(waveIndex)) === 0;
```
```suggestion
      doubleTrainer = randSeedInt(this.getDoubleBattleChance(waveIndex)) === 0;
```

DayKev:
Okay apparently the whole "double trainer softlock" on wave 1 doesn't happen anymore. There was probably some nonsense desync happening. I just tested a bunch with the changes from the suggestions and whether I override the trainer type to be double or just the battle to be a trainer battle without specifying whether it should be doubles or not, there's no softlock if there's a doubles trainer battle on wave 1.

Bertie690:
@DayKev 

<details>

image
Clearly, some bullshit is still happening somewhere.

</details>

DayKev:
I sure love when things work properly in-game but the test suite can't handle it.

PR #6562, opened 2025-09-16
[Tests] Cleanup `getCookie` and add many unit tests

SirzBenjie:
None, though cleans up the getCookie code to before the bug

The original PR that adjusted `getCookie` to clean up its logic had a very hard to spot error where it was subslicing the wrong string. This mismatch occured because `cookieTrimmed` was being used to match name, but then the original `c` object was being subsliced. Which is wrong, and resulted in there being an off-by-one error whenever attempting to slice cookies.
This is why clearing cookies fixed it on beta. The bug couldn't be triggered if there were no cookies set prior to the login threshold. This is why I mistakenly believed it to be an idiosyncracy in Firefox (which I don't use, so I don't have my suite of adblockers on, and thus don't avoid the google analytics cookies that were set), which triggered the bug, while chrome didn't.

Added a bunch of unit tests
Made the `getCookie` use the right code, and expanded out variable names to make it clear.
Folks, this is a tale of why we don't use acronyms for variable names. It obfuscates it and makes it easy to miss which variable corresponds to which thing.

Added a bunch of unit tests for this method.
## Screenshots/Videos
N/A

## How to test the changes?
`pnpm test test/utils/cookie.test.ts`

## Checklist

### DIFF
--- a/test/utils/cookies.test.ts
+++ b/test/utils/cookies.test.ts
@@ -1,0 +1,62 @@
+import { getCookie } from "#utils/cookies";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+describe("Unit Tests - cookies.ts", () => {
+  describe("getCookie", () => {
+    const cookieStart = document.cookie;
+    beforeEach(() => {
+      // clear cookie before each test
+      document.cookie = "";
+    });
+
+    afterEach(() => {
+      // restore original cookie after each test
+      document.cookie = cookieStart;
+    });
+    /**
+     * Spies on `document.cookie` and replaces its value with the provided string.
+     */
+    function setDocumentCookie(value: string) {
+      vi.spyOn(document, "cookie", "get").mockReturnValue(value);
+    }
+    it("returns the value of a single cookie", () => {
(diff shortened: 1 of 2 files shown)


test/utils/cookies.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    const cookieStart = document.cookie;
+    beforeEach(() => {
+      // clear cookie before each test
+      document.cookie = "";
+    });
+
+    afterEach(() => {
+      // restore original cookie after each test
+      document.cookie = cookieStart;
+    });
```
DayKev, line 15: Since you're spying on / mocking `document.cookie`, is there any need to [re]set it here?

SirzBenjie: Yeah, the submethods that are called by `getCookie` actually set the document.cookie property. See `removeCookie`.
And I didn't mock the setter for document.cookie, only its getter.

I need to restore the cookie after each test to avoid cross-state pollution.

DayKev (approved):

Bertie690 (approved):

Madmadness65 (approved):

PR #6563, opened 2025-09-16
[AI][Beta][Bug] Ensure AI always spawns with 4 moves if it can

SirzBenjie:
Pokemon will no longer stop generating moves if the number of moves they are still able to learn is less than the total number of moves in the moveset.

<details><summary>Fixes bug reported by Moro</summary>

> Not the first time I've found this out. 5 moves in moveset but spawns with only 3. Not a big deal in general, but what exactly is causing this?
image
</details> 

- Fixed initial stab generation to use the weighted move pool instead of the pre-weighted one
  - I had thought there was a bug with using the pre-weighted movepool,  but it turns out testing just one moveset generation is not enough. We need to test more. I created a little temporary script that can be injected to replace the stab move generation portion so that we can log a bunch of trials for moves: https://gist.github.com/SirzBenjie/1bc34d560b032d2d047024e7a005c845
- Corrected yet another instance of attempting to index into a number.
- Marked the flaky expert Pokémon breeder test as flaky
- Added a regression test to ensure that this does not happen in the future.

## Screenshots/Videos
N/A

## How to test the changes?
Just make sure movesets generate properly.

## Checklist

### DIFF
--- a/test/ai/ai-moveset-gen.test.ts
+++ b/test/ai/ai-moveset-gen.test.ts
@@ -1,4 +1,4 @@
-import { __INTERNAL_TEST_EXPORTS } from "#app/ai/ai-moveset-gen";
+import { __INTERNAL_TEST_EXPORTS, generateMoveset } from "#app/ai/ai-moveset-gen";
 import {
   COMMON_TIER_TM_LEVEL_REQUIREMENT,
   GREAT_TIER_TM_LEVEL_REQUIREMENT,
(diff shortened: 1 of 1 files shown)


test/ai/ai-moveset-gen.test.ts
```diff
@@ -282,4 +282,28 @@ describe("Regression Tests - ai-moveset-gen.ts", () => {
...
+        [10, MoveId.ROCK_THROW],
+        [13, MoveId.DOUBLE_TEAM],
+      ]);
+
+      // Generate the moveset
+      generateMoveset(pokemon);
+      expect(pokemon.moveset).toHaveLength(4);
+      // Unlike other test suites, phase interceptor is not automatically restored after the tests here,
+      // since most tests in this suite do not need the phase
+      game.phaseInterceptor.restoreOg();
```
Bertie690, line 306: This should be removed after the phase interceptor PR fwiw

damocleas (approved):

PR #6578, opened 2025-09-20
[Bug][Balance] Make type boost items consider moves that have a variable type when rolling one for rewards

SirzBenjie:
Moves with dynamic types whose type change is intrinsic to the move and its user will now resolve to that type for the purposes of determining which type boosting item to spawn
Notably, this impacts these moves:
- Revelation Dance
  - Same type as revelation dance would be when considering the user's real types (e.g. not a temporary one from Protean)
- Hidden Power
- Aura Wheel on Morpeko (will always allow both blackglasses and magnet)
- Techno Blast on Genesect - type of its current form instead of always normal. Normal on other mons
- Ivy Cudgel on Ogerpon
- Raging Bull on Paldean Tauros
- Judgement on Arceus
- Multi Attack on Silvally
- Tera Blast, allows both Tera Type and normal
  - Will only consider tera type on mons that are allowed to Terastallize
  - Player must have Tera Orb

As an additional QoL, Pokémon that are not allowed to be used in the challenge will no longer have their moves considered when rolling for type boosting items.

Also fixed an off-by-one bug related to the RNG used to pick a random attack type boost item.

I asked players what bugs / annoyances they found.

Also, I asked balance about Tera Blast. And I asked some users.

image

image

[source](https://discord.com/channels/1125469663833370665/1225619031156064317/1418955780756865124)

Added a new method to `VariableMoveTypeAttr` called `getTypesForItemSpawn` that returns an array of `PokemonTypes` that the move with the attribute should allow when rolling for type boosting items.
By default, the method will return the move's type (same behavior as before). However, it can (and should) be overridden by subclasses where the move's type is mostly static (like revelation dance).
I special cased Aura Wheel on Morpeko, as well, for additional QoL.

## Screenshots/Videos
<details><summary>Arceus</summary>

https://github.com/user-attachments/assets/eed47b00-7d9e-44f2-bffe-595629c00370

https://github.com/user-attachments/assets/f14c99e7-adbb-45a9-80f3-8d98a746031d
</details> 

<details><summary>Silvally</summary>

https://github.com/user-attachments/assets/98814a86-a5e0-42fa-a0da-f0986c9dfe8c
</details> 

<details><summary>Tera Blast</summary>

https://github.com/user-attachments/assets/7a8303ed-0f3f-49e2-91e6-2bd792fbcf30
</details> 

## How to test the changes?

Use the overrides below, changing the first move override and the species ID.
I threw in status moves to ensure that it doesn't consider other moves in the moveslots.

E.g. for Arceus:
```ts
const overrides = {
  ITEM_REWARD_OVERRIDE: [
    { name: "ATTACK_TYPE_BOOSTER"},
    { name: "FORM_CHANGE_ITEM"},
  ],
  STARTER_SPECIES_OVERRIDE: SpeciesId.ARCEUS,
  MOVESET_OVERRIDE: [MoveId.JUDGMENT, MoveId.GROWL, MoveId.TAIL_WHIP, MoveId.DEFENSE_CURL],
  ENEMY_ABILITY_OVERRIDE: AbilityId.BALL_FETCH,
  ENEMY_MOVESET_OVERRIDE: [MoveId.SPLASH],
  ENEMY_LEVEL_OVERRIDE: 1,
```
See the "what changes will the user see" section for the rest.

For Tera Blast:
```ts
const overrides = {
  STARTING_MODIFIER_OVERRIDE: [
    { name: "TERA_ORB"}
  ],
  ITEM_REWARD_OVERRIDE: [
    { name: "ATTACK_TYPE_BOOSTER"},
  ],
  MOVESET_OVERRIDE: [MoveId.TERA_BLAST, MoveId.GROWL, MoveId.TAIL_WHIP, MoveId.DEFENSE_CURL],
  ENEMY_ABILITY_OVERRIDE: AbilityId.BALL_FETCH,
  ENEMY_SPECIES_OVERRIDE: SpeciesId.MAGIKARP,
  ENEMY_MOVESET_OVERRIDE: [MoveId.SPLASH],
  ENEMY_LEVEL_OVERRIDE: 1,
```

Might need to do a few waves for the tera blast test to see it allowing both items.

## Checklist
  - [ ] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes? No, not until item rework

### DIFF
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -54,7 +54,7 @@ import { MoveResult } from "#enums/move-result";
 import { MoveTarget } from "#enums/move-target";
 import { isVirtual, MoveUseMode } from "#enums/move-use-mode";
 import { MultiHitType } from "#enums/multi-hit-type";
-import { PokemonType } from "#enums/pokemon-type";
+import { MAX_POKEMON_TYPE, PokemonType } from "#enums/pokemon-type";
 import { PositionalTagType } from "#enums/positional-tag-type";
 import { SpeciesId } from "#enums/species-id";
 import {
@@ -5195,6 +5195,16 @@ export class VariableMoveTypeAttr extends MoveAttr {
   apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
     return false;
   }
+
+  /**
+   * Determine the type of the move for the purpose of determining the type-boosting item to spawn
+   * @param user - The Pokémon using the move
+   * @param move - The move being used
+   * @returns An array of types to add to the pool of type-boosting items
+   */
+  getTypesForItemSpawn(user: Pokemon, move: Move): PokemonType[] {
+    return [move.type];
+  }
 }
 
 export class FormChangeItemTypeAttr extends VariableMoveTypeAttr {
@@ -5206,8 +5216,10 @@ export class FormChangeItemTypeAttr extends VariableMoveTypeAttr {
 
     if ([ user.species.speciesId, user.fusionSpecies?.speciesId ].includes(SpeciesId.ARCEUS) || [ user.species.speciesId, user.fusionSpecies?.speciesId ].includes(SpeciesId.SILVALLY)) {
       const form = user.species.speciesId === SpeciesId.ARCEUS || user.species.speciesId === SpeciesId.SILVALLY ? user.formIndex : user.fusionSpecies?.formIndex!;
-
-      moveType.value = PokemonType[PokemonType[form]];
+      if (form >= 0 && form <= MAX_POKEMON_TYPE && form !== PokemonType.STELLAR) {
+        moveType.value = form as PokemonType;
+        return true;
+      }
       return true;
     }
 
@@ -5217,6 +5229,14 @@ export class FormChangeItemTypeAttr extends VariableMoveTypeAttr {
     }
     moveType.value = move.type
     return true;
+  }
+
+  override getTypesForItemSpawn(user: Pokemon, move: Move): PokemonType[] {
+    // Get the type
+    const typeHolder = new NumberHolder(move.type);
+    // Passing user in for target is fine; the parameter is unused anyway
+    this.apply(user, user, move, [ typeHolder ]);
+    return [typeHolder.value];
   }
 }
 
@@ -5252,6 +5272,12 @@ export class TechnoBlastTypeAttr extends VariableMoveTypeAttr {
 
     return false;
   }
+
+  override getTypesForItemSpawn(user: Pokemon, move: Move): PokemonType[] {
+    const typeHolder = new NumberHolder(move.type);
+    this.apply(user, user, move, [ typeHolder ]);
+    return [typeHolder.value];
+  }
 }
 
 export class AuraWheelTypeAttr extends VariableMoveTypeAttr {
@@ -5276,6 +5302,15 @@ export class AuraWheelTypeAttr extends VariableMoveTypeAttr {
     }
 
     return false;
+  }
+
+  override getTypesForItemSpawn(user: Pokemon, move: Move): PokemonType[] {
+    // On Morpeko only, allow this to count for both blackglasses and magnet
+    if (this.apply(user, user, move, [new NumberHolder(move.type)])) {
+      return [PokemonType.DARK, PokemonType.ELECTRIC];
+    }
+
+    return [move.type];
   }
 }
 
@@ -5304,6 +5339,12 @@ export class RagingBullTypeAttr extends VariableMoveTypeAttr {
     }
 
     return false;
+  }
+
+  override getTypesForItemSpawn(user: Pokemon, move: Move): PokemonType[] {
+    const typeHolder = new NumberHolder(move.type);
+    this.apply(user, user, move, [ typeHolder ]);
+    return [ typeHolder.value ];
   }
 }
 
@@ -5339,6 +5380,12 @@ export class IvyCudgelTypeAttr extends VariableMoveTypeAttr {
     }
 
     return false;
+  }
+
+  override getTypesForItemSpawn(user: Pokemon, move: Move): PokemonType[] {
+    const typeHolder = new NumberHolder(move.type);
+    this.apply(user, user, move, [ typeHolder ]);
+    return [ typeHolder.value ];
   }
 }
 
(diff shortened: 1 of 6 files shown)


Madmadness65 (approved):

src/data/moves/move.ts
```diff
@@ -5261,6 +5315,23 @@ export class TeraBlastTypeAttr extends VariableMoveTypeAttr {
 
     return false;
   }
+
+  override getTypesForItemSpawn(user: Pokemon, move: Move): PokemonType[] {
+    const coreType = move.type;
+    const teraType = user.getTeraType();
+    /** Whether the user is allowed to tera. In the case of an enemy Pokémon, whether it *will* tera. */
+    const hasTeraAccess = user.isPlayer() ? canTerastallize(user) : willTerastallize(user);
```
Xavion3: This won't work well, You can't use the canTerastallize function you made like this or it'll create unintuitive results due to it being inconsistently applied.

src/utils/pokemon-utils.ts
```diff
@@ -123,3 +124,39 @@ export function getPokemonSpeciesForm(species: SpeciesId, formIndex: number): Po
...
+ *
+ * Does not check if the Pokémon is allowed to Terastallize (e.g., if it's a mega)
+ * @param pokemon - The Pokémon to check
+ * @returns Whether the Pokémon will Terastallize
+ *
+ * @remarks
+ * Should really only be called with an enemy Pokémon, but will technically work with any Pokémon.
+ *
+ * @privateRemarks
+ * Assumes that Pokémon with no trainer ever tera, so this must be changed if
```
DayKev:
```suggestion
 * Assumes that Pokémon with no trainer will never tera, so this must be changed if
```

src/utils/pokemon-utils.ts
```diff
@@ -123,3 +124,39 @@ export function getPokemonSpeciesForm(species: SpeciesId, formIndex: number): Po
...
+ * a wild Pokémon is allowed to tera, e.g. for a Mystery Encounter.
+ */
+export function willTerastallize(pokemon: Pokemon): boolean {
+  // cast is safe, as if it is just a Pokémon, initialTeamIndex will be undefined triggering the null check
+  const initialTeamIndex = (pokemon as EnemyPokemon).initialTeamIndex;
+  return (
+    initialTeamIndex != null
+    && pokemon.hasTrainer()
+    && (globalScene.currentBattle?.trainer?.config.trainerAI.instantTeras.includes(initialTeamIndex) ?? false)
+  );
```
DayKev, line 169:
```suggestion
export function willTerastallize(pokemon: EnemyPokemon): boolean {
  const initialTeamIndex = pokemon.initialTeamIndex;
  const instantTeras = globalScene.currentBattle?.trainer?.config.trainerAI.instantTeras ?? [];
  return (initialTeamIndex != null && pokemon.hasTrainer() && instantTeras.includes(initialTeamIndex));
```
This is an AI-only function, so it should just take an `EnemyPokemon` from the start.

src/ui/handlers/command-ui-handler.ts
```diff
@@ -198,14 +197,10 @@ export class CommandUiHandler extends UiHandler {
...
-    const hasTeraMod = globalScene.getModifiers(TerastallizeAccessModifier).length > 0;
     const activePokemon = globalScene.getField()[this.fieldIndex];
-    const isBlockedForm =
-      activePokemon.isMega() || activePokemon.isMax() || activePokemon.hasSpecies(SpeciesId.NECROZMA, "ultra");
-    const currentTeras = globalScene.arena.playerTerasUsed;
-    const plannedTera =
-      globalScene.currentBattle.preTurnCommands[0]?.command === Command.TERA && this.fieldIndex > 0 ? 1 : 0;
-    return hasTeraMod && !isBlockedForm && currentTeras + plannedTera < 1;
+    const canTera = activePokemon.isPlayer() && canTerastallize(activePokemon);
+    const plannedTera = globalScene.currentBattle.preTurnCommands[0]?.command === Command.TERA && this.fieldIndex > 0;
```
DayKev: That's not the same behavior. It's currently set up to allow future changes where the player is allowed to tera more than once per biome (hence the `currentTeras + plannedTera < 1` check). This change doesn't allow that to happen fwict.

src/utils/pokemon-utils.ts
```diff
@@ -123,3 +124,39 @@ export function getPokemonSpeciesForm(species: SpeciesId, formIndex: number): Po
...
+/**
+ * Determine whether a player Pokémon can Terastallize
+ * @param pokemon - The Pokémon to check
+ * @returns Whether the Pokémon can Terastallize
+ */
+export function canTerastallize(pokemon: PlayerPokemon): boolean {
+  const hasTeraMod = globalScene.findModifier(modifier => modifier.is("TerastallizeAccessModifier")) != null;
+  const isBlockedForm = pokemon.isMega() || pokemon.isMax() || pokemon.hasSpecies(SpeciesId.NECROZMA, "ultra");
+  const currentTeras = globalScene.arena.playerTerasUsed === 0;
+  return hasTeraMod && !isBlockedForm && currentTeras;
```
DayKev:
```suggestion
  const hasAvailableTeras = globalScene.arena.playerTerasUsed < 1;
  return hasTeraMod && !isBlockedForm && hasAvailableTeras;
```

src/data/moves/move.ts
DayKev: Is this function used by AI pokemon to determine what items they're able to generate with?

DayKev: Like Xavion says, this isn't really correct. Once you use your Tera in a biome, you'll suddenly no longer be able to get type boosting items of your Pokemons' tera types until the next biome when you're able to tera again, which is weird.

SirzBenjie: Can you elaborate?
This method should not check temporary state; it doesn't care if the Pokémon is currently terastallized.

DayKev: Specifically, `canTerastallize()` will return `false` once you've used your Tera because `playerTerasUsed` will then be `1`.

Xavion3: Yeah, the desired logic is the version from before this commit, where it just checks if the player has the tera orb. I think the only utility function we actually need here would probably go on the speciesForm class I guess? For the blacklisted forms. The tera orb check is kinda wordy but that's being refactored away into being much simpler with the upcoming modifier rework anyway.

SirzBenjie: Oh whoops. I had forgotten my _own_ logic on this.

src/utils/pokemon-utils.ts
SirzBenjie: I had it as that. But then, this throws a wrench in the moveset generation, which also takes a pokemon. And it needs to take a Pokemon, as it is used for Daily runs.
(credit @Xavion3 for pointing this out)

src/data/moves/move.ts
SirzBenjie: I addressed this by making another function.

PR #6595, opened 2025-09-27
[Balance] Adjust evolution delay requirements

SirzBenjie:
A much more appropriate evolution curve threshold.
No more Crobats at level 23.

Balance team request.

Yeeted the hard to understand and even harder to maintain evolution delay.
Replaced it with a different system, based on level requirements.
A Pokémon's level requirement is the HIGHEST of the level it evolves at and the evolution threshold.

Evolution thresholds have 3 kinds: Strong, Normal, and Wild.
Strong is for bosses and normal is for other.

Strong trainers will always have their Pokemon evolved by the specified level, though this can be tweaked.
Normal trainers have a 1/(0.1*threshold) * level chance to remain unevolved rather than evolving. This rate is configurable via a constant. Wild pokemon have the same, with the rate at 0.2 instead.

## Checklist

### DIFF
--- a/src/@types/species-gen-types.ts
+++ b/src/@types/species-gen-types.ts
@@ -1,0 +1,25 @@
+// biome-ignore lint/correctness/noUnusedImports: Used in TSDoc comment
+import type { EvoLevelThresholdSort } from "#enums/evo-level-threshold-kind";
+import type { SpeciesId } from "#enums/species-id";
+
+/**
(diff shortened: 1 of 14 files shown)


damocleas:
https://github.com/pagefaultgames/pokerogue/blob/beta/src/data/balance/biomes.ts has levels for almost everything set for when they should evolve in the wild, these would likely need to be removed, as this also helps with simplicity

damocleas:
failing now

PR #6604, opened 2025-09-29
[Dev] Added `noBannedTypes`, `useConsistentArrayType` and `useShorthandAssign` Biome rules

Bertie690:
N/A
I like biome rules

Added linter rules
## Screenshots/Videos
N/A
## How to test the changes?
N/A

## Checklist

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -115,7 +115,17 @@
         },
         "useCollapsedIf": "error",
         "useCollapsedElseIf": "error",
-
+        "useDeprecatedReason": "error",
+        "useConsistentArrayType": {
+          "level": "error",
+          "fix": "unsafe",
+          "options": {}
+        },
+        "useShorthandAssign": {
+          "level": "error",
+          "fix": "safe",
+          "options": {}
+        },
         "noSubstr": "error",
         "noYodaExpression": "error",
         "useForOf": "error",
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -204,7 +204,7 @@
         "noForEach": "off", // Foreach vs for of is not that simple.
         "noUselessSwitchCase": "off", // Explicit > Implicit
         "noUselessConstructor": "error",
-        "noBannedTypes": "warn", // TODO: Refactor and make this an error
+        "noBannedTypes": "error", // TODO: Refactor and make this an error
         "noThisInStatic": "error",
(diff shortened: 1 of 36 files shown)


src/data/pokeball.ts
```diff
@@ -116,7 +116,7 @@ export function doPokeballBounceAnim(
   y2: number,
   baseBounceDuration: number,
   // biome-ignore lint/complexity/noBannedTypes: TODO
```
DayKev:
```suggestion
```

src/field/pokemon.ts
```diff
@@ -1847,21 +1847,21 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    // Override moveset based on arrays specified in overrides.ts
+    const overrideArray = coerceArray(this.isPlayer() ? Overrides.MOVESET_OVERRIDE : Overrides.ENEMY_MOVESET_OVERRIDE);
+    if (overrideArray.length === 0) {
+      return !ignoreOverride && this.summonData.moveset ? this.summonData.moveset : this.moveset;
     }
 
+    if (!this.isPlayer()) {
+      this.moveset = [];
+    }
+    // TODO: Maybe we don't need to manually set PP to max?
```
DayKev: The PP is set like this so that pp used isn't reset by the override every time `getMoveset` is called. Though... I'm not sure how necessary that is. It's not being set to max though.
```suggestion
    // TODO: Preserve PP used while the moveset override is active
```

DayKev (approved):

src/field/pokemon.ts
Bertie690: IMO the answer is "yes" since that would allow manual testing of Spite interactions and similar.

DayKev (dismissed):

biome.jsonc
```diff
@@ -205,7 +215,7 @@
         "noForEach": "off", // Foreach vs for of is not that simple.
         "noUselessSwitchCase": "off", // Explicit > Implicit
         "noUselessConstructor": "error",
-        "noBannedTypes": "warn", // TODO: Refactor and make this an error
+        "noBannedTypes": "error", // TODO: Refactor and make this an error
```
emdeann: Should remove this comment now, right?

emdeann (dismissed):

PR #6611, opened 2025-09-30
refactor: rewrite logic to switch, summon and recall Pokemon

Bertie690:
[Original poketernity PR](https://github.com/Despair-Games/poketernity/pull/1236)

[Locales PR](https://github.com/pagefaultgames/pokerogue-locales/pull/246)
(Removes an unused locale key, low priority)

> [!IMPORTANT]
> The bulk of the changes in this PR are either ported straight from `pkty` or consist of minor callsite changes that must be copied among the numerous duplicated callsites.

* On the prompt to switch Pokemon at the start of a battle, if the Player selects "Yes", then cancels out of the Party menu without selecting a Pokemon, the game will now return to the switch prompt.
* Some minor improvements to animation timing
* The order of post-summon effects will now be consistent between reloads and standard new battles. More specifically, the game will always use the prior ordering from reloading the page:
    1. Enemy Pokemon enter field
    2. Player Pokemon enter field
    3. IV scanners
    4. Switch prompts and/or `PostSummonPhase` queueing

(Previously, continuing from a prior wave would swap steps 2 and 3, displaying the IV scanner results _before_ any required player switch ins occur.)

> [!CAUTION]
> This PR also unfortunately ports a few regressions from poketernity (at least for the time being):
> - Trainers in double battles will not show send-in messages for the 2nd pokemon
> - Challenge-ineligible Pokemon will not be removed from the field upon evolution

### Innerthunder
The current phase hierarchy for recalling, switching, and summoning Pokemon (namely `ReturnPhase`, `SwitchPhase`, `SwitchSummonPhase`, and `SummonPhase`) is extremely convoluted. Logic and data manipulation is woven in between animations composed of hanging Promises and other unmarked async calls.

### Me taking a jackhammer to everything else

While porting the above changes, I noticed several issues with the way poketernity handled switch phases:

1. `SummonPhase`s would unconditionally push a `PostSummonPhase` to the queue after finishing its animation.
This goes directly against the idea of the phase being "purely visual", and would re-introduce the same double Intimidate bugs that were fixed a while ago. 
6. `SwitchPhase` gave callers no ability to customize the queued `SummonPhase`, which would prevent being able to customize displayed messages for force switching moves. 

Ported InnerThunder's refactors to the phase hierarchy:
> 
> * Removed `SwitchSummonPhase` and `ReturnPhase`
> * Added `RecallPhase`, which plays animations for recalling a Pokemon and removes the Pokemon from the field and is essentially a replacement for `ReturnPhase`
> * Rewrote `SwitchPhase` to handle all logical aspects of switching, including:
> * Resolving the index of the Pokemon to switch in
> * Applying all changes to game data as a result of the switch, including pre-switch Ability triggers and data transfers from Baton Pass and Shed Tail.
> * Rewrote `SummonPhase` to only contain visual effects related to summoning a Pokemon
> 
> The end result: Instead of the `SummonPhase` > `SwitchSummonPhase` > `ReturnPhase` hierarchy, each phase in the `RecallPhase` > `SwitchPhase` > `SummonPhase` sequence is independent and can be scheduled independently. In theory, every sequence where Pokemon are recalled, switched, and/or summoned can be implemented by a combination of one or more of the new phases.

Made additional changes to resolve the phase queueing logic and prevent regressions:
- Reworked `CheckSwitchPhase`s into a drop-in replacement for player `PostSummonPhase`s.
   - They are now queued **in place of** the player Pokemon's `PostSummonPhase`s, and directly queue the aforementioned `PostSummonPhase` if the relevant conditions are not met or the switch prompt actively denied. This greatly simplifies logic by removing the need to cull extra phases or check their existence _post hoc_.
   - The phase now incorporates added checks for all conditions previously done at callsites that directly pushed them to the queue.
   - This also removes the need for `SummonPhase`s to queue post summon phases. It is now purely animation-focused (aside from adding the Pokemon to the field and a borked check for challenge eligibility)

- Changed the parameter for `SwitchPhase` to a partial options type used for `SwitchPhase`, preserving the ability to customize switch-in behaviour.

- Player Pokemon summon phases are now queued directly inside `EncounterPhase`/co.** instead of `TitlePhase`/`GameOverPhase`.
   - This was already implemented in the fork at the time of the original PR, and is more a port than a change in functionality.
- Added a large helper function (`queueBattlerEntrances`) that is called by `EncounterPhase` and its ME doppelgänger in lieu of pushing several assorted phases in 5 different places. This ensures proper relative ordering of all combatants' `CheckSwitchPhase`/`PostSummonPhase`s while also removing a large chunk of convoluted, duplicated code. 

## Screenshots/Videos

https://github.com/user-attachments/assets/f08a3262-05c8-4bdc-8e03-ddf25b9f6611

## How to test the changes?
The changes in this PR are mainly visual and therefore require manual playtesting.
Other than that, run the game and make sure things like trainer battles and transitions to/from double battles don't break.

I did add a test in `double-battle.test.ts` to guard against regressions (since we previously lacked any tests that manually verified switches occurred in dynamic speed order.)

## Checklist


src/field/pokemon.ts
```diff
@@ -3272,6 +3259,15 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
     return this.isPlayer() ? globalScene.getPlayerField() : globalScene.getEnemyField();
   }
 
+  /**
+   * Gets the Pokémon on the opposing field.
+   *
+   * @returns An array of Pokémon on the opposite side of the field..
```
Bertie690:
```suggestion
   * @returns An array of Pokémon on the opposite side of the field.
```

Wlowscha (dismissed):
There are various TODOs introduced, but I've played around with this for a while and did not encounter any issues. Plus it is needed for #7369

Wlowscha (dismissed):

src/utils/switch-utils.ts
```diff
@@ -0,0 +1,173 @@
...
+  const { double } = globalScene.currentBattle;
+
+  const availablePlayerPartyMembers = globalScene.getPokemonAllowedInBattle();
+  const availableEnemyPartyMembers = globalScene.getEnemyParty().filter(p => p.isAllowedInBattle());
+
+  const addPlayer2 = double && availablePlayerPartyMembers.length > 1;
+  const addEnemy2 = double && availableEnemyPartyMembers.length > 1;
+
+  const playerMons = globalScene.getPlayerParty().slice(0, addPlayer2 ? 2 : 1);
+  const enemyMons = globalScene.getEnemyParty().slice(0, addEnemy2 ? 2 : 1);
```
DayKev: Shouldn't these be `available*PartyMembers.slice(...)`?

DayKev (changes requested):
If a Pokémon evolves into something that's not challenge-legal, it won't be recalled off the field.
<details><summary>Eevee evolves into Flareon in mono-Normal challege</summary>

Image

</details>

src/data/mystery-encounters/encounters/safari-zone-encounter.ts
```diff
@@ -331,8 +331,7 @@ async function summonSafariPokemon() {
   encounter.misc.pokemon = pokemon;
   encounter.misc.safariPokemonRemaining -= 1;
 
-  globalScene.phaseManager.unshiftNew("SummonPhase", 0, false);
-  globalScene.phaseManager.unshiftNew("PostSummonPhase", BattlerIndex.ENEMY);
+  globalScene.phaseManager.queueBattlerEntrance(BattlerIndex.ENEMY, { when: "delayed" });
```
DayKev, line 334: Why change from `unshift` to `push`?

Bertie690:
@DayKev is that on pkty beta

DayKev:
Seems like it does happen at pkty too.

Bertie690:
Added to the list of "accidentally backported regressions I need to fix"

src/phase-manager.ts
```diff
@@ -235,6 +236,55 @@ const turnEndPhases: readonly PhaseString[] = [
...
 
+interface BattlerEntranceParams extends SummonPhaseOptions {
+  /**
+   * String denoting when to add the phases.
+   * Possible values are:
+   *  - `"eager"`: Adds the phase immediately via {@linkcode PhaseManager.unshiftPhase | unshiftPhase}
+   *  - `"delayed"`: Adds the phase via {@linkcode PhaseManager.pushPhase} to run after all phases finish running.
+   */
+  // TODO: Figure out a default value for this and remove it from existing callsites
+  when: "eager" | "delayed";
```
DayKev, line 247: This one is `"delayed"` and the below one is `"deferred"`; they should be the same.

src/phases/summon-phase.ts
```diff
@@ -1,301 +1,409 @@
...
-    this.preSummon();
+    // If the Pokemon about to be summoned is fainted or illegal under active challenges,
+    // try to reorganize the Pokemon's party such that a legal inactive Pokemon is summoned instead.
+    // If this cannot be done, queue a game over and end the phase.
+    if (!this.getPokemon().isAllowedInBattle() && !this.handleIllegalSummon()) {
+      globalScene.phaseManager.clearAllPhases();
+      globalScene.phaseManager.unshiftNew("GameOverPhase", false);
+      super.end();
+      return;
+    }
```
DayKev, line 75: Wouldn't this queue a game over if somehow the enemy team triggers this block?

Wlowscha:
I'm also noticing other related errors, using this override and normal monotype challenge
const overrides = {
  STARTER_SPECIES_OVERRIDE: SpeciesId.EEVEE,
  STARTING_LEVEL_OVERRIDE: 100,
  ITEM_REWARD_OVERRIDE: [{name: "EVOLUTION_ITEM"}]

For example, if eevee evolves in a double battle, the game overs instead of sending it back and leaving the second pokémon out. We probably want to add a bunch of automatic tests for these cases...

DayKev (commented):
Incomplete review

src/data/mystery-encounters/encounters/safari-zone-encounter.ts
Bertie690: Good catch

src/phase-manager.ts
Bertie690: They're intentionally different because they're doing different things.
We'd be giving them the same name and different purposes, which is bad.

src/phases/summon-phase.ts
Bertie690: Ostensibly. 
I copied it from inner's PR.

Bertie690:
@DayKev I really dislike the method chaining for `getSprite` you added.
Going from returning `this` to `Phaser.GameObjects.Sprite` is a wee bit strange IMO.

src/phase-manager.ts
DayKev: Then at the very least you should swap them, since we already have a method using "defer" to mean "push instead of unshift the phase" but you've done the opposite.

Bertie690: I made it consistent with the terms in `PhaseTree`, fwiw.

Issue #6369, opened 2025-08-23
[Feature] Allow Pecharunt's Toxic Chain Passive to properly activate Poison Puppeteer


Blitz425:
### Is your feature request related to a problem? Please describe.

Pecharunt has Toxic Chain as a passive to combo with its ability Poison Puppeteer, but when Toxic Chain gets activated, Poison Puppeteer doesn't.

### Describe the Feature

If Toxic Chain activates as an ability / passive, it should interact with Poison Puppeteer if its on the same Pokemon as an ability / passive.

PR #6622, opened 2025-10-03
[Feature] Fixes #[6369] where Poison Puppeteer didn't work with Toxic Chain.

SamThePanda1:
Now, whenever a Pokemon has both poison puppeteer and toxic chain, the toxic chain will activate, causing poison, and then poison puppeteer will cause confusion.

Fixes #[6369] 
This PR tries to fix issue #6369 where poison puppeteer doesn't work with toxic chain.

I added a conditional block within the current if statement of obtain-status-effect-phase.ts to check if the thing causing the poison was in fact a pokemon and not from another source. If it was, then applyAbAttrs("ConfusionOnStatusEffectAbAttr", ...) is called where within that, the pokemon is confirmed to have the ability/passive Poison Puppeteer, and if it does the opposing pokemon becomes confused.

## How to test the changes?
To test the change, unlock Pecharunt's passive and get it's Toxic Chain to poison a Pokemon and then it's Poison Puppeteer should activate and cause confusion.

## Checklist

### DIFF
--- a/src/phases/obtain-status-effect-phase.ts
+++ b/src/phases/obtain-status-effect-phase.ts
@@ -6,6 +6,7 @@ import { SpeciesFormChangeStatusEffectTrigger } from "#data/form-change-triggers
 import { getStatusEffectObtainText } from "#data/status-effect";
 import type { BattlerIndex } from "#enums/battler-index";
 import { CommonAnim } from "#enums/move-anims-common";
+import { MoveId } from "#enums/move-id";
 import { StatusEffect } from "#enums/status-effect";
 import type { Pokemon } from "#field/pokemon";
 import { PokemonPhase } from "#phases/pokemon-phase";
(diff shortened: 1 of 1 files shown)


src/phases/obtain-status-effect-phase.ts
```diff
@@ -60,6 +61,16 @@ export class ObtainStatusEffectPhase extends PokemonPhase {
           effect: this.statusEffect,
           sourcePokemon: this.sourcePokemon ?? undefined,
         });
+        // Trigger ConfusionOnStatusEffectAbAttr (e.g., Poison Puppeteer) when status is applied
+        //Before this, Poison Puppeteer didn't work with Toxic Chain.
+        if (this.sourcePokemon) {
+          applyAbAttrs("ConfusionOnStatusEffectAbAttr", {
+            pokemon: this.sourcePokemon,
+            opponent: pokemon,
+            move: { id: MoveId.NONE } as any, // Use NONE move since this is Ability/Passive trigger.
```
emdeann, line 70: It would be preferable to make `move` on `ConfusionOnStatusEffectAbAttrParams` optional instead of this cast to `any` here

Bertie690: Just checked and `ConfusionTag` doesn't even use the `sourceMove` parameter anywhere.
We can (and should) remove it if it isn't used, but there should be no harm in marking it as optional

DayKev:
Closed in favor of a more comprehensive approach in #6836

PR #6635, opened 2025-10-06
[Docs] Update documentation on `Ability` class getters

Bertie690:
N/A
I like docs.
Also it's a bit weird that the docs on `AbBuilder.unsuppressable` are better than those on the exposed `Ability` class

Updated getter docs on the main ability class

made ability flag bitmasks use explicit bit shifts to more clearly indicate their status as er bitmasks

## Screenshots/Videos
N/A
## How to test the changes?
N/A (since this is literally all comments)
## Checklist

### DIFF
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -69,32 +69,40 @@ import { BooleanHolder, NumberHolder, randSeedFloat, randSeedInt, randSeedItem, 
 import { toCamelCase } from "#utils/strings";
 import i18next from "i18next";
 
+//#region Bit sets
 /** Bit set for an ability's `bypass faint` flag */
-const AB_FLAG_BYPASS_FAINT = 1;
+const AB_FLAG_BYPASS_FAINT = 1 << 0;
 /** Bit set for an ability's `ignorable` flag */
-const AB_FLAG_IGNORABLE = 2;
+const AB_FLAG_IGNORABLE = 1 << 1;
 /** Bit set for an ability's `suppressable` flag */
-const AB_FLAG_UNSUPPRESSABLE = 4;
+const AB_FLAG_UNSUPPRESSABLE = 1 << 2;
 /** Bit set for an ability's `uncopiable` flag */
-const AB_FLAG_UNCOPIABLE = 8;
+const AB_FLAG_UNCOPIABLE = 1 << 3;
 /** Bit set for an ability's `unreplaceable` flag */
-const AB_FLAG_UNREPLACEABLE = 16;
+const AB_FLAG_UNREPLACEABLE = 1 << 4;
 /** Bit set for an ability's `unimplemented` flag */
-const AB_FLAG_UNIMPLEMENTED = 32;
+const AB_FLAG_UNIMPLEMENTED = 1 << 5;
 /** Bit set for an ability's `partial` flag */
-const AB_FLAG_PARTIAL = 64;
-
-/** Bits set for a swappable ability */
+const AB_FLAG_PARTIAL = 1 << 6;
+/** Bit set for a swappable ability */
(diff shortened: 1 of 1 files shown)


Madmadness65 (dismissed):

src/data/abilities/ability.ts
```diff
@@ -69,32 +69,40 @@ import { BooleanHolder, NumberHolder, randSeedFloat, randSeedInt, randSeedItem,
...
+const AB_FLAG_UNREPLACEABLE = 1 << 4;
 /** Bit set for an ability's `unimplemented` flag */
-const AB_FLAG_UNIMPLEMENTED = 32;
+const AB_FLAG_UNIMPLEMENTED = 1 << 5;
 /** Bit set for an ability's `partial` flag */
-const AB_FLAG_PARTIAL = 64;
-
-/** Bits set for a swappable ability */
+const AB_FLAG_PARTIAL = 1 << 6;
+/** Bit set for a swappable ability */
```
```suggestion
/** Bit set for an unswappable ability */
```

src/data/abilities/ability.ts
```diff
@@ -69,32 +69,40 @@ import { BooleanHolder, NumberHolder, randSeedFloat, randSeedInt, randSeedItem,
...
-
-/** Bits set for a swappable ability */
+const AB_FLAG_PARTIAL = 1 << 6;
+/** Bit set for an unswappable ability */
 const AB_FLAG_UNSWAPPABLE = AB_FLAG_UNCOPIABLE | AB_FLAG_UNREPLACEABLE;
 
+//#endregion Bit sets
+
+/**
+ * An Ability is a class representing the various Abilities Pokemon may have.
```
```suggestion
 * An Ability is a class representing the various Abilities Pokemon may have. \
```

src/data/abilities/ability.ts
```diff
@@ -69,32 +69,40 @@ import { BooleanHolder, NumberHolder, randSeedFloat, randSeedInt, randSeedItem,
...
+ * Each has one or more {@linkcode AbAttr | attributes} that can apply independently
+ * of one another.
+ */
 export class Ability {
   /** The ability's unique identifier */
   public readonly id: AbilityId;
-  /** Key used to localize the ability's name */
+  /** The locales key used to localize the ability's name */
```
emdeann: If you prefer to say "locales key" just say "The locales key for the name of the ability" or something. This way is redundant so it reads weird

src/data/abilities/ability.ts
```diff
@@ -122,35 +130,65 @@ export class Ability {
     }
     return i18next.t(`ability:${this.i18nKey}.description`);
   }
-  /** Whether the retains its effects through a faint */
+
+  /**
+   * Whether this ability retains its effects after the user faints.
```
```suggestion
   * Whether this ability can activate even if the user faints.
```

src/data/abilities/ability.ts
```diff
@@ -236,8 +274,8 @@ class AbBuilder {
    * @param args - The arguments needed to instantiate the given class.
    * @returns `this`
    */
-  attr<T extends Constructor<AbAttr>>(AttrType: T, ...args: ConstructorParameters<T>): this {
-    const attr = new AttrType(...args);
+  attr<T extends Constructor<AbAttr>>(attrType: T, ...args: ConstructorParameters<T>): this {
+    const attr = new attrType(...args);
```
DayKev, line 280: It doesn't really matter either way, but this doesn't need to be changed since `AttrType` is a class constructor (and thus should be capitalized like all other class constructors).

DayKev (approved):

emdeann (approved):
Good once kev's stuff is addressed

src/data/abilities/ability.ts
Bertie690: i mean we don't do this elsewhere sooo IMO best to be consistent

emdeann (approved):

PR #6638, opened 2025-10-08
dev: enable some options from TypeScript's "strict" mode

SirzBenjie:
None

Improve typing and move towards enabling `strict` in tsconfig

Enables these strict options:
- `strictBindCallApply`
- `strictBuiltinIteratorReturn`
- `strictFunctionTypes`
Removed default exports from controller configuration.
Lots of changes to typing.
Also, I tried enabling `NoImplicitAny` and found a TON of actual bugs in our input configuration. Well, logic bugs. One of them was to do with using `Map` as a type, but not actually utilizing anything from map, and instead accessing it as if it were a python dictionary. I replaced this with just a record.

## Screenshots/Videos
N/A

## How to test the changes?
Do tests pass? Then we good.

## Checklist
- [x] Have I tested the changes manually? Extensively


src/data/balance/starters.ts
```diff
@@ -616,7 +616,7 @@ export const speciesStarterCosts = {
   [SpeciesId.PALDEA_TAUROS]: 5,
   [SpeciesId.PALDEA_WOOPER]: 3,
   [SpeciesId.BLOODMOON_URSALUNA]: 5,
-};
+} as Partial<Record<SpeciesId, number>>;
```
Bertie690: Should this be `as` or `satisfies`?

src/dynamic-queue-manager.ts
```diff
@@ -91,7 +91,7 @@ export class DynamicQueueManager {
    * @returns Whether a matching phase exists
    */
   public exists<T extends PhaseString>(type: T, condition?: PhaseConditionFunc<T>): boolean {
-    return !!this.dynamicPhaseMap.get(type)?.has(condition);
+    return !!this.dynamicPhaseMap.get(type)?.has(condition as (phase: Phase) => boolean);
```
Bertie690: this... does not inspire confidence at all

also shouldn't these default to `() => true`?

src/field/pokemon.ts
```diff
@@ -4093,7 +4093,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
   public getTag(tagType: BattlerTagType.SUBSTITUTE): SubstituteTag | undefined;
   public getTag(tagType: BattlerTagType): BattlerTag | undefined;
   public getTag<T extends BattlerTag>(tagType: Constructor<T>): T | undefined;
-  public getTag(tagType: BattlerTagType | Constructor<BattlerTag>): BattlerTag | undefined {
+  public getTag(tagType: BattlerTagType | typeof BattlerTag): BattlerTag | undefined {
```
Bertie690, line 4129: i actually changed this in another PR of mine (read: _3 separate PRs that I `git checkout`ed the file from) to use the battler tag type map (including Heal attr).

just sayin - this is like really outdated

src/system/game-speed.ts
```diff
@@ -12,14 +12,12 @@ const PROPERTIES = ["delay", "completeDelay", "loopDelay", "duration", "repeatDe
...
-export function initGameSpeed() {
-  const thisArg = this as BattleScene;
-
+export function initGameSpeed(this: BattleScene) {
   const transformValue = (value: number | FixedInt): number => {
     if (value instanceof FixedInt) {
       return (value as FixedInt).value;
     }
-    return thisArg.gameSpeed === 1 ? value : Math.ceil((value /= thisArg.gameSpeed));
+    return this.gameSpeed === 1 ? value : Math.ceil((value /= this.gameSpeed));
```
Bertie690:
```suggestion
    return this.gameSpeed === 1 ? value : Math.ceil(value / this.gameSpeed);
```

idfk why it even used `/=` anyways

test/test-utils/test-file-initialization.ts
```diff
@@ -83,7 +83,12 @@ function setupStubs(): void {
    * @param x - The relative x position
    * @param y - The relative y position
    */
-  const setPositionRelative = function (guideObject: any, x: number, y: number): any {
+  const setPositionRelative = function <T extends Phaser.GameObjects.Components.Transform>(
+    this: T,
+    guideObject: any,
+    x: number,
+    y: number,
+  ): any {
```
Bertie690: Please revert - I typed this properly in my pr updating its signature

Bertie690 (changes requested):
minor but otherwise lftm

src/dynamic-queue-manager.ts
SirzBenjie: It's just too narrow with `Phase`. It ends up being just fine.

src/data/balance/starters.ts
SirzBenjie: `as`.

src/dynamic-queue-manager.ts
DayKev: > also shouldn't these default to `() => true`?

Yes but this isn't the PR to change them.

DayKev (dismissed):
Approved when the couple of changes matt requested are made.

Bertie690 (approved):

PR #6644, opened 2025-10-09
[Bug][Beta] Fix Fun and Games playing in wrong order

emdeann:
The ending to Fun and Games (and probably others) will no longer run backwards (rewards -> message)

Bug

`PhaseTree` worked by pushing an empty level whenever a phase was popped, and unshifted phases were then queued to `levels.length - 1`. This had an oversight - in the case where the queue had just been cleared and the same phase pushed and unshifted phases to the queue, the unshifted ones would not run before the pushed ones. There was no opportunity to add the next level (because nothing had been popped yet), so both unshifting and pushing would push a phase onto level zero of the tree.

This is fixed by instead using `currentLevel` when unshifting - we always unshift to `currentLevel + 1`, adding the extra level first if needed.

## How to test the changes?
Run tests, and play through MEs like Fun and Games

## Checklist


src/phase-tree.ts
```diff
@@ -32,6 +32,9 @@ export class PhaseTree {
    * @throws Error if `level` is out of legal bounds
    */
   private add(phase: Phase, level: number): void {
+    if (level === this.currentLevel + 1 && level === this.levels.length) {
```
Bertie690, line 35: I'd like an explanatory doc comment if possible.

emdeann: I don't feel like that's warranted here? We need to add a new level if we're adding above the current level and the array hasn't already been expanded. I don't think a comment would be any better of an explanation than the code itself; it's very straightforward

damocleas (approved):

PR #6653, opened 2025-10-12
[Test] Cleaned up enum test utils

Bertie690:
N/A
I needed some stuff fixed for the modifier rework branch and Benjie suggested moving it out into its own PR.

Improved `getEnumStr` to accept optional parameters for numeric base and padding on integer values

## Screenshots/Videos
N/A
## How to test the changes?
Ryn the new tests i added to `strings.test.ts`
## Checklist


src/utils/enums.ts
```diff
@@ -56,7 +56,8 @@ export function getEnumValues<E extends EnumOrObject>(enumType: TSNumericEnum<E>
  *   two: 2,
  * } as const;
  * console.log(enumValueToKey(thing, 2)); // output: "two"
- * @throws Error if an invalid enum value is passed to the function
+ * @throws {Error}
+ * Throws if an invalid enum value is passed to the function.
```
DayKev:
```suggestion
 * @throws Throws an {@linkcode Error} if an invalid enum value is passed to the function.
```
Technically that should be `@throws {@link Error}\n...` but VSCode doesn't render it properly because Microsoft doesn't care about its own projects.

test/test-utils/string-utils.ts
```diff
@@ -46,103 +69,134 @@ interface getEnumStrOptions {
...
+    casing = "Preserve",
+    prefix = "",
+    suffix = "",
+    excludeValues = typeof val === "string",
+    base = 10,
+    padding = 0,
+  } = options as getEnumStrOptions;
+
+  const keyPart = excludeValues ? "" : getKeyPart(obj, val, { casing, prefix, suffix });
+  const valuePart = typeof val === "string" ? val.toString() : getValuePart(val, { base, padding });
```
DayKev: Is there any point in doing `.toString()` on something that is already a string?

Bertie690: No
It do be funni tho

DayKev:
```suggestion
  const valuePart = typeof val === "string" ? val : getValuePart(val, { base, padding });
```

SirzBenjie (approved):

PR #6658, opened 2025-10-16
[Docs] Update `battler-tags.ts` module doc comment for clarity

Bertie690:
None, nada, zero, zilch, _il n'y en a pas_

I was GOING to make a PR fixing the mess that is Curse, when i noticed the intro sounded a bit... clunky.

Rather than lump doc changes into another wholly unrelated PR, I made the "responsible" decision of making this a separate PR.

updated le wording
## Screenshots/Videos
N/A

## How to test the changes?
check for grammar/etc
## Checklist

### DIFF
--- a/src/data/battler-tags.ts
+++ b/src/data/battler-tags.ts
@@ -40,7 +40,7 @@
  *     // This signature should be used if the class has no subclasses:
  *     public override loadTag(source: BaseBattlerTag & Pick<ExampleTag, "tagType" | "a" | "b"): void;
  *     // This signature should be used if the class has any subclasses:
- *     public override loadTag<const T extends this>(source: BaseBattlerTag & Pick<T, "tagType" | "a" | "b">): void
+ *     public override loadTag<const T extends this>(source: BaseBattlerTag & Pick<T, "tagType" | "a" | "b">): void {
  *       this.a = source.a;
  *       this.b = source.b;
  *     }


src/data/battler-tags.ts
```diff
@@ -205,15 +209,20 @@ export class BattlerTag implements BaseBattlerTag {
...
   }
 }
 
 export class SerializableBattlerTag extends BattlerTag {
-  /** Nonexistent, dummy field to allow typescript to distinguish this class from `BattlerTag` */
+  /**
+   * Nonexistent, dummy field to allow typescript to distinguish this class from `BattlerTag`.
+   *
+   * @remarks
+   * Does not exist at runtime, so must not be used!
```
DayKev, line 234: It's typed as `never` meaning it can't actually be used for anything anyway or TS will complain, so there's no need to modify the doc comment.
Also it's `private` meaning it would only be possible to "use" it within this class, but the class is empty on purpose already.

Bertie690: 1. Benjie put this on a few other classes with similar features; I'm just continuing the trend
2. bracket notation exists
3. if u want to remove it feel free to make a comment - i'm kinda w/e about it anyhwo

src/data/battler-tags.ts
```diff
@@ -1,39 +1,41 @@
 /**
- * BattlerTags are used to represent semi-persistent effects that can be attached to a Pokemon.
- * Note that before serialization, a new tag object is created, and then `loadTag` is called on the
- * tag with the object that was serialized.
+ * `BattlerTag`s are used to represent semi-persistent effects that can be attached to a Pokemon.
  *
- * This means it is straightforward to avoid serializing fields.
- * Fields that are not set in the constructor and not set in `loadTag` will thus not be serialized.
+ * During serialization, a new blank tag object is created, before its `loadTag` is called
+ * with the object that was serialized. \
```
fabske0, line 5:
```suggestion
 * with the object that was serialized.
```

DayKev: I think the newline is fine.

src/data/battler-tags.ts
```diff
@@ -1,39 +1,41 @@
...
+ * This makes it fairly straightforward to avoid serializing fields — anything not set in the class constructor
+ * or the tag's `loadTag` method will not be serialized.
  *
- * Any battler tag that can persist across sessions must extend SerializableBattlerTag in its class definition signature.
- * Only tags that persist across waves (meaning their effect can last >1 turn) should be considered
- * serializable.
+ * Any battler tag that can persist across waves **must extend `SerializableBattlerTag`** in its class definition signature.
+ * (In most cases, this corresponds to it being able to last multiple turns in battle.)
```
DayKev: Is there any case where this wouldn't be true? I think definitionally all serializable tags must last longer than 1 turn.

src/data/battler-tags.ts
```diff
@@ -1,39 +1,51 @@
...
+ *   ```ts
+ *   class ExampleTag extends SerializableBattlerTag {
+ *     // Example, if we add 2 new fields that should be serialized:
+ *     public a: string;
+ *     public b: number;
+ *     // Then we must also define a loadTag method with one of the following signatures:
+ *     // This signature should be used if the class has no subclasses:
+ *     public override loadTag(source: BaseBattlerTag & Pick<ExampleTag, "tagType" | "a" | "b"): void;
+ *     // This signature should be used if the class has any subclasses:
+ *     public override loadTag<const T extends this>(source: BaseBattlerTag & Pick<T, "tagType" | "a" | "b">): void
```
DayKev:
```suggestion
 *     public override loadTag<const T extends this>(source: BaseBattlerTag & Pick<T, "tagType" | "a" | "b">): void {
```

fabske0 (approved):

SirzBenjie (approved):

PR #6661, opened 2025-10-16
[Beta] Pixilate/etc no longer affect non-Normal type moves

DayKev:
Pixilate and similar abilities will no longer incorrectly change the type of moves that are not Normal type.

Bug fix.

Removed unnecessary (and incorrect) `if` check in the constructor of `MoveTypeChangeAbAttrParams`.

## How to test the changes?
Run the new test with `pnpm test:silent normal-move-type-change`

## Checklist

### DIFF
--- a/test/abilities/normal-move-type-change.test.ts
+++ b/test/abilities/normal-move-type-change.test.ts
@@ -74,7 +74,7 @@ describe.each([
     expect(enemyPokemon.hp).toBeLessThan(enemyPokemon.getMaxHp());
   });
 
-  it("should not affect non-Normal type Moves", async () => {
+  it("should not affect moves that are not Normal type", async () => {
     await game.classicMode.startBattle([SpeciesId.FEEBAS]);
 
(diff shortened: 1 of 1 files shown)


test/abilities/normal-move-type-change.test.ts
```diff
@@ -75,9 +74,27 @@ describe.each([
     expect(enemyPokemon.hp).toBeLessThan(enemyPokemon.getMaxHp());
   });
 
+  it("should not affect non-Normal type Moves", async () => {
```
Bertie690: Would it be "non-Normal type", "non Normal-type" or "non Normal type"?

Bertie690 (dismissed):

damocleas (dismissed):

test/abilities/normal-move-type-change.test.ts
DayKev:
```suggestion
  it("should not affect moves that are not Normal type", async () => {
```
Maybe this, so it's extra clear?

Issue #6667, opened 2025-10-18
[Feature] Increase Catchrate for Shinies


damocleas:
Legends: ZA Updated their catchrate to increase for Shinies, and I think that'd be a good feature to add here with our variants existing and such. Here is all of the [ZA catch changes](https://rotomlabs.net/guide/legends-z-a-capture-mechanics) if you're interested.

2x Catchrate for Shinies, if possible with a number that is adjustable in a Balance File
Increasable chance for Events, maybe default to 3x and have it adjustable in the same place as a default change for events?

This will require updating our catch formula.

PR #6670, opened 2025-10-19
[Balance] Increase shiny catch rate; add event bonus (implements #6667)

ds3919:
- Shiny Pokémon are easier to catch.
  - Base bonus: 2× to the modified catch rate.
  - During shiny-rate events: 3×.
- Non-shiny catches are unchanged.
- Master Ball behavior unchanged; Critical Catch still works the same (it just benefits indirectly from the higher base rate).

Implements request in #6667 to make shinies feel more rewarding to catch and to make shiny-rate events feel special. Values are balance-tunable so the team can adjust without code changes.

- **Modified** `src/phases/attempt-capture-phase.ts`: multiply `modifiedCatchRate` by a shiny multiplier.
- **Added** `src/data/balance/catch.ts`: exports `SHINY_CATCH_MULTIPLIER` and `SHINY_EVENT_CATCH_MULTIPLIER`.
- **Notes**: No tiering; uniform bonus for all shinies. Master Ball/uncatchable checks unchanged. Critical Catch logic untouched.

## How to test the changes?
1. Use the same species at the same HP/status with the same ball:
   - Non-shiny: baseline catch behavior.
   - Shiny (set a shiny target or temporarily force one): noticeably higher success rate.
2. Temporarily make `isShinyEventActive()` return `true` and repeat step 1 → success rate increases further.
3. Verify Master Ball still guarantees capture.
4. Run local checks: `pnpm i && pnpm test` (and any lint/typecheck scripts).

### DIFF
--- a/src/data/balance/catch.ts
+++ b/src/data/balance/catch.ts
@@ -1,0 +1,3 @@
+// Tunable knobs for shiny catch bonus (single baseline, no tiers).
+export const SHINY_CATCH_MULTIPLIER = 2.0;        // default per the issue
+export const SHINY_EVENT_CATCH_MULTIPLIER = 3.0;  // default during shiny events
--- a/src/phases/attempt-capture-phase.ts
+++ b/src/phases/attempt-capture-phase.ts
@@ -1,6 +1,7 @@
 import { PLAYER_PARTY_MAX_SIZE } from "#app/constants";
 import { globalScene } from "#app/global-scene";
 import { getPokemonNameWithAffix } from "#app/messages";
+import { SHINY_CATCH_MULTIPLIER, SHINY_EVENT_CATCH_MULTIPLIER } from "#data/balance/catch";
 import { SubstituteTag } from "#data/battler-tags";
 import { Gender } from "#data/gender";
 import {
@@ -27,6 +28,11 @@ import { SummaryUiMode } from "#ui/summary-ui-handler";
 import { applyChallenges } from "#utils/challenge-utils";
 import { BooleanHolder } from "#utils/common";
 import i18next from "i18next";
+
+// TODO: wire this to your real event system/flag when available.
+function isShinyEventActive(): boolean {
+  return false; // placeholder
+}
 
 // TODO: Refactor and split up to allow for overriding capture chance
 export class AttemptCapturePhase extends PokemonPhase {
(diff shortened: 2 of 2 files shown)


src/phases/attempt-capture-phase.ts
```diff
@@ -64,7 +70,18 @@ export class AttemptCapturePhase extends PokemonPhase {
     const catchRate = pokemon.species.catchRate;
     const pokeballMultiplier = getPokeballCatchMultiplier(this.pokeballType);
     const statusMultiplier = pokemon.status ? getStatusEffectCatchRateMultiplier(pokemon.status.effect) : 1;
-    const modifiedCatchRate = Math.round((((_3m - _2h) * catchRate * pokeballMultiplier) / _3m) * statusMultiplier);
+    const shinyMultiplier = (
+      typeof (pokemon as any).isShiny === "function"
+        ? pokemon.isShiny()
+        : !!(pokemon as any).shiny
+    )
```
Bertie690, line 77: Why are these casts to `any` here?
`Pokemon`doesn't normally have a `shiny` parameter (barring some hackjob type-unsafe shenanigans), and Phaser will almost certainly crash if we pass a non-Pokemon object here.

src/data/balance/catch.ts
```diff
@@ -0,0 +1,3 @@
+// Tunable knobs for shiny catch bonus (single baseline, no tiers).
+export const SHINY_CATCH_MULTIPLIER = 2.0;        // default per the issue
+export const SHINY_EVENT_CATCH_MULTIPLIER = 3.0;  // default during shiny events
```
Bertie690, line 3:
```suggestion
/** Catch rate multiplier for Shiny Pokemon during normal gameplay. */
export const SHINY_CATCH_MULTIPLIER = 2.0
/** Catch rate multiplier for Shiny Pokemon during Shiny events. */
export const SHINY_EVENT_CATCH_MULTIPLIER = 3.0; 
```

Bertie690 (changes requested):
I'd recommend copypasting the example code i gave into `timed-event-manager` and using that function in the conditional.
other than that and the typing issues lgtm

damocleas (approved):
approving since balance is needed, but please address comments

src/phases/attempt-capture-phase.ts
```diff
@@ -28,6 +29,11 @@ import { applyChallenges } from "#utils/challenge-utils";
 import { BooleanHolder } from "#utils/common";
 import i18next from "i18next";
 
+// TODO: wire this to your real event system/flag when available.
+function isShinyEventActive(): boolean {
+  return false; // placeholder
+}
+
```
Bertie690, line 36: OK since GitHub ate my comment here Imma need to redo this:
TL;DR the actual timed event manager class is in `src/timed-event-manager.ts`, the global instance of which is imported from `src/global-event-manager.ts`.

There isn't a stock "is shiny event active" function already pre-existing, but adding something vaguely like this should suffice:
```ts
export class TimedEventManager {
  // ...normal stuff

  /** 
   * @returns Whether a Shiny Event is currently active.
   */
  public isShinyEventActive(): boolean {
    return this.activeEvent()?.eventType === EventType.SHINY;
  }
```

damocleas:
Closed in favor of #6694

PR #6678, opened 2025-10-22
[Beta][Bug] DetermineEnemySpecies was using a very low minimum value for evolution chance

Jimmybald1:
Fair evolution odds, instead of nearly always evolve.

The desired effect is to have a random value between the evo threshold and a value 10-20% higher.

`randSeedInt` has a complicated `min` parameter. 

If you take a wild Primeape, its threshold/choice is 65 and the mult is 20% (13 levels higher).
Old code used `randSeedInt(choice, Math.round(choice * multiplier))` in other words: `randSeedInt(65, 13)` which results in `range(13, 77)`. This means a random value between 13 and 77, but all the values between 13 and 65 will by default be lower and thus guarantee evolve. That only leaves 20% of the values that could actually result in a no evolution. Thus the evolution only scales from 80-100% chance after its threshold (and 0% chance before its threshold).

You could swap the two values from `randSeedInt(65, 13)` to `randSeedInt(13, 65)`, that would also give the desired `range(65, 77)`. But for readability it just makes more sense to me to use `randSeedIntRange(choice, Math.round(choice * multiplier))` which results in `randSeedIntRange(65, 78)` which results in `range(65, 78)`. 

The +1 difference is because randSeedInt has an exclusive range, while randSeedIntRange has an inclusive range. For balance reasons you could still decide on making this value -1, but i dont think it matters that much. 
And i believe `randSeedIntRange(65, 78)` is much more intuitive than `randSeedInt(13, 65)` to see what is going to happen.

For readability it just makes more sense to me to use `randSeedIntRange()`.

## How to test the changes?
Just spawn some wild unevolved pokemon with a level greater than but close to their wild evo threshold.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "biome-ci": "biome ci --diagnostic-level=error --reporter=github --no-errors-on-unmatched",
     "typedoc": "typedoc",
     "depcruise": "depcruise src test",
-    "postinstall": "lefthook install; git config --local recurse.submodules true",
+    "postinstall": "lefthook install; git config --local fetch.recurseSubmodules true",
     "update-version:patch": "pnpm version patch --force --no-git-tag-version",
     "update-version:minor": "pnpm version minor --force --no-git-tag-version",
     "update-locales": "git submodule update --progress --init --recursive --depth 1 locales",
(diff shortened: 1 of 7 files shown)


src/ai/ai-species-gen.ts
```diff
@@ -5,24 +5,24 @@ import type { PokemonSpecies } from "#data/pokemon-species";
...
+const NORMAL_TRAINER_LEVEL_DIFF_PERCENT = 1.1;
 /**
  * Controls the maximum level difference that a Pokémon spawned with
  * {@linkcode EvoLevelThresholdKind.WILD} is allowd to remain unevolved.
  */
-const WILD_LEVEL_DIFF_PERCENT = 0.2;
+const WILD_LEVEL_DIFF_PERCENT = 1.2;
 /**
  * Controls the maximum level difference that a Pokémon spawned with
  * {@linkcode EvoLevelThresholdKind.} is allowd to remain unevolved.
```
fabske0:
```suggestion
 * {@linkcode EvoLevelThresholdKind.STRONG} is allowed to remain unevolved.
```

fabske0 (approved):

Issue #6674, opened 2025-10-20
[Code] Refactor `TimedEventManager` to disallow multiple overlapping events


Bertie690:
About half of `TimedEventManager` appears to presume the ability to start multiple events concurrently, and the other half just... doesn't.

After discussing with balance, it has been determined that we probably don't need the ability to compute shiny odds for overlapping events, and the file could be massively cleaned up as a result.

PR #6679, opened 2025-10-22
[Refactor] Only use first active `timedEvent`

fabske0:

Should be none, other than not having multiple events at once

#6674

- All the `timedEventManager` methods only use the first active event. This is done by using `.activeEvent()`
- Renamed `shinyMultiplier` to `shinyEncounterMultiplier`
- Added a `shinyCatchMultiplier` property to the `TimedEvent` interface and added an associated method to get the multiplier.
- also some comments

## How to test the changes?

Turn on an event and play around with the properties and make sure they work

## Checklist

### DIFF
--- a/src/ui/battle-info/battle-info.ts
+++ b/src/ui/battle-info/battle-info.ts
@@ -595,7 +595,7 @@ export abstract class BattleInfo extends Phaser.GameObjects.Container {
     this.setTypes(pokemon.getTypes(true, false, undefined, true));
 
     if (this.lastHp !== pokemon.hp || this.lastMaxHp !== pokemon.getMaxHp()) {
-      return this.updatePokemonHp(pokemon, resolve, instant);
+      this.updatePokemonHp(pokemon, resolve, instant);
     }
     if (!this.player && this.lastLevel !== pokemon.level) {
       this.setLevel(pokemon.level);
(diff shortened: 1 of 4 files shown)


src/timed-event-manager.ts
```diff
@@ -521,43 +504,25 @@ export class TimedEventManager {
...
-      if (this.isActive(te) && te.mysteryEncounterTierChanges != null) {
-        for (const metc of te.mysteryEncounterTierChanges) {
-          if (metc.mysteryEncounter === encounterType) {
-            ret = metc.tier ?? normal;
-          }
-        }
+    const metChanges = this.activeEvent()?.mysteryEncounterTierChanges ?? [];
+    for (const metc of metChanges) {
+      if (metc.mysteryEncounter === encounterType) {
+        return metc.tier ?? normal;
```
emdeann, line 510: This is technically a behavior change if one event has two different tier changes for the same encounter type, but that shouldn't be an issue unless something incorrect is relied upon

src/timed-event-manager.ts
```diff
@@ -398,99 +408,75 @@ export class TimedEventManager {
...
+    return this.activeEvent()?.shinyEncounterMultiplier ?? 1;
+  }
 
-    return multiplier;
+  /**
+   * Get the multiplier for shiny catches during a shiny {@linkcode TimedEvent}
+   * @returns the shiny catch multiplier
+   */
+  getShinyCatchMultiplier(): number {
+    return this.activeEvent()?.shinyCatchMultiplier ?? 1; // todo: add base constant from #6670 after it gets merged
```
DayKev: I have a feeling we'll need to do that ourselves.
```suggestion
    return this.activeEvent()?.shinyCatchMultiplier ?? 1;
```

DayKev (approved):

PR #6687, opened 2025-10-24
[Github] Improve test workflow to merge blob reports from all shards

Bertie690:
N/A
Better testing UX - instead of needing to check 5 test shards for 5 failing tests (each of which can take a while to load), all shard results will now be packaged into 1 large file for easy viewing.

Also allows for easier scalability of test shards - adding a final "check errors" step lets us scale up or down our shard count as necessary without breaking CI.

Configured the test shard workflow to explicitly override the reporters used and upload the blobs as an artifact.

Added a step to condense the blobs back into 1 large file for viewing.

Removed some unused parameters from the test shard template workflows - `project` was unused in the workflow file and `skip` is now checked with a singular `if` statement

> [!IMPORTANT]
> After this PR is merged, I need a repository admin to change the required status checks from `Run Tests (XXX)` to the singular `Tests / Check Test Results` workflow.

## Screenshots/Videos
image

## How to test the changes?
Run the test workflow, fail a few tests, etc etc

## Checklist

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,7 +9,7 @@
 {
   "name": "Node.js & TypeScript",
   // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-  "image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:4-24-bookworm",
   "features": {
     "ghcr.io/devcontainers/features/github-cli:1": {
       "installDirectlyFromGitHubRelease": true,
@@ -18,6 +18,9 @@
     "ghcr.io/devcontainers-extra/features/pnpm:2": {
       "version": "latest"
     }
+  },
+  "containerEnv": {
+    "PORT": "8000"
   },
   "customizations": {
     "vscode": {
--- a/.ls-lint.yml
+++ b/.ls-lint.yml
@@ -25,6 +25,7 @@ ls:
       <<: *cfg
    test: *src
 ignore:
+   - .pnpm-store
    - node_modules
    - .vscode
    - .github
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,11 +47,18 @@ This Linux environment comes with all required dependencies needed to start work
 [codespaces-link]: <https://github.com/codespaces/new?hide_repo_select=true&repo=620476224&ref=beta>
 [devcontainer-ext]: <https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers>
 
+> [!IMPORTANT]
+> Due to quirks of devcontainer port forwarding, you must use **`pnpm start:podman`** to start a local dev server from within a devcontainer.
+> All other instructions remain the same as local development.
+
+### Podman
+For those who prefer Docker containers, see [this instructions page](./docs/podman.md) for information on how to setup a development environment with Podman.
+
 ### Local Development
 
 #### Prerequisites
 
-- node: >=22.14.0 - [manage with pnpm](https://pnpm.io/cli/env) | [manage with fnm](https://github.com/Schniz/fnm) | [manage with nvm](https://github.com/nvm-sh/nvm) | [manage with volta.sh](https://volta.sh/)
+- node: >=24.9.0 - [manage with pnpm](https://pnpm.io/cli/env) | [manage with fnm](https://github.com/Schniz/fnm) | [manage with nvm](https://github.com/nvm-sh/nvm) | [manage with volta.sh](https://volta.sh/)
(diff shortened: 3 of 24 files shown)


.github/workflows/test-shard-template.yml
```diff
@@ -43,4 +35,20 @@ jobs:
         run: pnpm i
 
       - name: Run tests
-        run: pnpm test:silent --shard=${{ inputs.shard }}/${{ inputs.totalShards }}
+        run: >
+          pnpm test:silent
+          --shard=${{ inputs.shard }}/${{ inputs.totalShards }}
+          --reporter=blob
+          --outputFile=test-results/blob-${{ inputs.shard }}.json || true
```
DayKev, line 42: What's the `|| true` for?

.github/workflows/tests.yml
```diff
@@ -41,14 +41,48 @@ jobs:
...
+
+  check-results:
+    name: Check Test Results
+    timeout-minutes: 8
+    needs: run-tests
+    runs-on: ubuntu-latest
+    # don't run if tests skipped
+    if: ${{ needs.run-tests.result != 'skipped' && needs.run-tests.result != 'cancelled' }}
+    steps:
+      - name: Check out Git repository
```
DayKev, line 62: Why does this need to download the repo?

.github/workflows/test-shard-template.yml
Bertie690: To mask the error value and make the action still count as passing.
That way the report merge will always run even if shards fail (which itself fails if a constituent blob fails)

It doesn't really make sense to report errors on individual shards when all the logs for them say is "blob created at XYZ"

.github/workflows/tests.yml
Bertie690: It needs vitest which needs package json
I think we might be able to do a sparse checkout maybe?

DayKev: Why can't you just do `pnpm add vitest`?

Bertie690: good idea

DayKev: Nevermind, it's too annoying to make work, don't worry about it.

.github/workflows/tests.yml
```diff
@@ -41,14 +41,48 @@ jobs:
...
       shard: ${{ matrix.shard }}
       totalShards: 5
-      skip: ${{ needs.check-path-change-filter.outputs.all != 'true'}}
+
+  check-results:
+    name: Check Test Results
+    timeout-minutes: 8
+    needs: run-tests
+    runs-on: ubuntu-latest
+    # don't run if tests skipped
```
DayKev:
```suggestion
```

.github/workflows/test-shard-template.yml
```diff
@@ -3,27 +3,19 @@ name: Test Template
...
         type: number
-      skip:
-        required: true
-        type: boolean
-        default: false
 
 jobs:
   test:
-    # We can't use dynmically named jobs until https://github.com/orgs/community/discussions/13261 is implemented
+    # We can't use dynamically named jobs until https://github.com/orgs/community/discussions/13261 is implemented
```
DayKev, line 15: Btw that's probably not going to happen anymore, Microsoft laid off the team working on it over 2 years ago (and they had it almost complete too, thanks Microsoft).
https://github.com/actions/runner/pull/2477#issuecomment-1501003600

Bertie690: I was just fixing grammar lol kinda depressing but sure

DayKev: Oh yeah, for some reason I thought you added that lol.

.github/workflows/tests.yml
SirzBenjie: Why does it need to use recursive submodules?
If it just needs package.json, then you can just do a normal checkout.

Bertie690: It needs package json, vitest/vite config files, the custom reporters, `nvmrc`, all the pnpm locks, and maybe `tsconfig.json` as well

I could not get it to work for the life of me

Bertie690: > Why does it need to use recursive submodules?
If it just needs package.json, then you can just do a normal checkout.

the EN namespace locale collection crashes vitest if the folder doesn't exist

.github/workflows/tests.yml
```diff
@@ -41,14 +41,48 @@ jobs:
...
+    needs: run-tests
+    runs-on: ubuntu-latest
+    if: ${{ needs.run-tests.result != 'skipped' && needs.run-tests.result != 'cancelled' }}
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v4.2.2
+        with:
+          submodules: "recursive"
+
+
```
DayKev:
```suggestion
```

PR #6688, opened 2025-10-25
[Dev] Enable linting of `move.ts` and update Biome to 2.3.2

DayKev:
N/A

All files should be linted.

Enabled linting of `move.ts` and applied fixes.

Also updated Biome to get the improved diagnostic for `maxFunctionParams`.

## Checklist


Bertie690:
@DayKev While we're updating Biome, we should bump to 2.3.2 - I sneaked in an upstream hotfix for `noUselessStringConcat` that should make it actually function now

PR #6697, opened 2025-10-26
refactor: enforce usage of `Pokemon.isOfType` over `Pokemon.getTypes`; make default values consistent

Bertie690:
A few moves and abilities will now have correct interactions when Stellar Terastallized.
- Flower Veil
- Flower Shield
- Reflect Type

A few other moves were changed to properly fail if the user is Terastallized:
- Conversion 2
- Conversion (also fails if user is same type as its first move, for consistency)

Finally, Forest's Curse/Trick-or-Treat will be automatically removed when the user Terastallizes (tested on showdown (<https://replay.pokemonshowdown.com/gen9ou-2530172961-8xgqrtwj8acyevf4srfv4m8vyohs7e9pw>). These effects were previously ignored if the user was Terastallized, but this saves a bit of space in save data.

I wanted to review the default parameter state of `Pokemon#getTypes`.
Then I realized its defaults were different from `Pokemon#isOfType` (which is outright confusing), alongside being a 71 cognitive complexity MESS.

> [!NOTE]
> I am keeping this PR small and self-contained as I want to make `forDefend` default to `true` later on as well (and perhaps even remove it due to being useful in exactly one place) 

Cleaned up implementation of `Pokemon.getTypes`, and  changed its `useTeraType` parameter to `true` (matching `isOfType`'s default behavior).

Fixed required spots where it was being passed without any arguments (all that remain now either don't run if the user is Terastallized or explicitly disregard it).

## Screenshots/Videos
N/A? Annoyingly hard to test

## How to test the changes?
`pnpm test flower-shield.test.ts` (a few new tests, including a Tera Stellar case)

## Checklist


src/phases/tera-phase.ts
```diff
@@ -36,6 +36,8 @@ export class TeraPhase extends BattlePhase {
 
   end() {
     this.pokemon.isTerastallized = true;
+    // Remove added type from Forest's Curse/Trick-or-Treat
+    this.pokemon.summonData.addedType = null;
```
Wlowscha, line 41: Is this a functional change? I don't see it addressed in the PR description.

src/field/pokemon.ts
```diff
@@ -2052,6 +1986,78 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       types.splice(0, 1);
     }
 
+    return types as Mutable<NonEmptyTuple<PokemonType>>;
+  }
+
+  /**
+   * Helper to {@linkcode getTypes} that handles computing a Pokemon's normal typing.
+   */
+  private getNormalTyping(ignoreOverride = false, useIllusion = false): PokemonType[] {
```
Wlowscha: Minor nitpick, but I would not call a function `getNormalTyping` considering that Normal is the name of a type. "Basic typing" perhaps, or "base typing"? Also, why is it "typing" here and not "type" or "types"?

src/field/pokemon.ts
```diff
@@ -1935,114 +1938,45 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+        return [teraType];
       }
+      types.push(teraType);
+    } else {
+      types.push(...this.getNormalTyping(ignoreOverride, useIllusion));
     }
 
-    // become UNKNOWN if no types are present
+    // become UNKNOWN if no types are present, or remove it if other types are present
+    // _sighs_... If only there was a clean way to tell TS that this makes `types` always nonempty
```
```suggestion
```

src/field/pokemon.ts
```diff
@@ -1935,114 +1938,45 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * @param ignoreOverride - (Default `false`) Whether to ignore any overrides caused by {@linkcode MoveId.TRANSFORM | Transform} and similar effects; default `false`
+   * @param useIllusion - (Default `false`) Whether to consider an active illusion; default `false`
+   * @returns A non-empty array of {@linkcode PokemonType}s corresponding to this Pokemon's typing (real or perceived).
    */
   public getTypes(
-    includeTeraType = false,
+    includeTeraType = true,
     forDefend = false,
     ignoreOverride = false,
     useIllusion = false,
```
DayKev, line 1952: You're updating the method and not changing all of this into an object?

Wlowscha:
I've been digging around a bit and I have some notes about `getTypes`:

The `forDefend` parameter is only used in relation to tera types. It has no effect otherwise, so for any calculations where `includeTeraType` is `false`, the value of `forDefend` is irrelevant. In fact, there is only a single callsite where `getTypes(false, true,` is used.

When `includeTeraType` is `true`, we have this logic
```if (forDefend) {
  return [teraType];
}
types.push(teraType);
```
However, this is arguably pointless since in the second case, no other types are added afterwards; the original types are on a different conditional branch, and extra types (e.g. from Trick or Treat) should be irrelevant in the case of tera. This means that in practice, the `forDefend` parameter is only relevant in the case of TeraStellar, to determine whether the Stellar type OR the original types should be returned.

In most cases we call `getTypes(true, true,` which returns the original types.

Conversely, `getTypes(true, false,` only appears at three callsites:
1) In `Pokemon.getMatchupScore`, where the score is influenced by the effectiveness of the enemy's types, so we do want `getTypes` to return Stellar;
2) In two methods of the `BattleInfo` class, where again it makes sense to have Stellar.

There are a few more cases however which call `getTypes(true)` (using `false` as the default):
1) When determining the effect of Curse (in `CurseAttr`)
2) In the `RemoveTypeAttr` class, but this looks like a mistake, because there is an explicit check for tera before that;
3) In `genNewPartyMemberSpecies`, where I am not sure how it would be relevant.

As far as I can see, `isOfType` is always called either with `true, true` or `false, false`.

Overall, this can lead to inconsistencies, see e.g. this comment in `move.ts`:
```
    new AttackMove(MoveId.BURN_UP, PokemonType.FIRE, MoveCategory.SPECIAL, 130, 100, 5, -1, 0, 7)
      // Pass `true` to `ForDefend` as it should fail if the user is terastallized to a type that is not FIRE
      .condition(user => user.isOfType(PokemonType.FIRE, true, true), 2)
```
Passing `true` to `forDefend` here means that the user is counted as fire type if it is currently Tera Stellar, which may not be intended.

--------------------

Based on this I would suggest the following changes, which may fit in this PR:
1) Clear up the `if (forDefend)` conditional in `getTypes`
2) Rename the `forDefend` parameter to something like `returnOriginalTypesIfStellar`, since that's the only case where it is relevant
3) Possibly, clean up the callsites indicated above if needed, although this might extend beyond the scope of this PR.

Otherwise, I will put this information in an issue and we can deal with it at a later time.

src/field/pokemon.ts
Bertie690: Correct.

I'm going to remove the `forDefend` parameter in a follow-up PR (for which I'll need to alter all the callsites anyways).

src/phases/tera-phase.ts
Bertie690: oops

Bertie690 (commented):
> When determining the effect of Curse (in CurseAttr)

I kep that as is since one of my other PRs handling Curse explicitly fixes that

src/field/pokemon.ts
```diff
@@ -1943,136 +1946,95 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    if (fusionSpeciesForm) {
+      // Check if the fusion Pokemon also has permanent changes from ME when determining the fusion types
+      const fusionCustomTypes =
+        this.fusionCustomPokemonData?.types.map(t => (t === PokemonType.UNKNOWN ? undefined : t)) ?? [];
+
+      const fusionType1 = fusionCustomTypes[0] ?? fusionSpeciesForm.type1;
+      const fusionType2 = fusionCustomTypes[1] ?? fusionSpeciesForm.type2;
+
+      // Assign second type if the fusion can provide one
+      secondType = fusionType1 ?? fusionType2;
```
Wlowscha:
```suggestion
      if (fusionType2 !== null && fusionType2 !== firstType) {
        secondType = fusionType2;
      } else if (fusionType1 !== firstType) {
        secondType = fusionType1;
      }
```
This check is still necessary to ensure correct second typing (as seen with failing tests).

Wlowscha (dismissed):

Bertie690:
@Wlowscha I have not forgotten about the shadow tag oversight, will checkout branch and fix tmr morning

Bertie690:
The oversight has been fixed

src/field/pokemon.ts
```diff
@@ -3661,19 +3629,35 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+  private getTeraTypeBoost(teraType: PokemonType, moveType: PokemonType, matchesSourceType: boolean): number {
+    // Non-stellar Teras give a 50% boost to their type exclusively
+    if (moveType !== PokemonType.STELLAR) {
+      return teraType === moveType ? 0.5 : 0;
     }
 
-    return Math.min(stabMultiplier.value, 2.25);
+    // TODO: Instead of ignoring terapagos' tera stellar boosts when calling this,
+    // we should avoid pushing its usages to the array altogether to reduce save data size
+    const canBoostStellar = !this.stellarTypesBoosted.includes(moveType) || this.hasSpecies(SpeciesId.TERAPAGOS);
```
DayKev, line 3664: Shouldn't this also be checking the attacker and not "`this`"?
You even made a comment on L3610: `// TODO: This uses nothing from this Pokemon AT ALL, so why is this called on the defender?`

Bertie690: Actually, I realize this is 100% correct - this method is called on `source` (attacker), not the defender (since it uses exactly 0 things from said defender).

I have updated the doc comment to mention this discrepancy for future notice.

DayKev: Oh I see; that's confusing lol. Both methods should probably just be entirely removed from the `Pokemon` class.

DayKev (approved):

PR #6714, opened 2025-10-30
[Misc] Allow setting daily boss variant via custom seed

fabske0:

It's now possible to set the Variant of the daily run boss via a custom seed

Requested for event

Added `daily-run.ts#getDailyEventSeedBossVariant()`, which gets extra number after the already existing `/boss{speciesId}{formIndex}`.
The variant is optional. If it's not in the seed or greater than 2, it treats it as `no variant`

00: Standart variant
01: Rare varriant
02: epic variant

## How to test the changes?

```ts
const overrides = {
  DAILY_RUN_SEED_OVERRIDE: "xxxxxxxxxxxxxboss00940202",
  STARTING_WAVE_OVERRIDE: 49,
```

## Checklist


damocleas (approved):

src/data/daily-run.ts
```diff
@@ -283,6 +284,30 @@ export function getDailyEventSeedBoss(seed: string): PokemonSpeciesForm | null {
...
+ * Only the last 2 digits matter for the variant, and it is clamped to 0-2.
+ * (left padded with `0`s as necessary).
+ * @returns A {@linkcode Variant} to be used for the boss, or `null` if no valid match.
+ */
+export function getDailyEventSeedBossVariant(seed: string): Variant | null {
+  if (!isDailyEventSeed(seed)) {
+    return null;
+  }
+
+  const match = /boss(\d{4})(\d{2})(\d{2})/g.exec(seed);
```
Bertie690: You can use `(?:` to ignite capture groups fyi

src/data/daily-run.ts
```diff
@@ -283,6 +284,30 @@ export function getDailyEventSeedBoss(seed: string): PokemonSpeciesForm | null {
...
+  if (!match || match.length !== 4) {
+    return null;
+  }
+
+  const variant = Number.parseInt(match[3]) as Variant;
+  if (variant > 2) {
+    return null;
+  }
+  return variant;
```
Bertie690:
```suggestion
  const match = /boss\d{6}(\d{2})/g.exec(seed);
  if (!match || match.length !== 2) {
    return null;
  }

  const variant = Number.parseInt(match[1]) as Variant;
  if (variant > 2) {
    return null;
  }

  return variant;
```

PR #6729, opened 2025-10-31
[Misc] Allow setting shiny starters via custom seed

fabske0:

The variant for daily run starters can be set via a custom seed now

Requested by Moro
Also makes the seed slightly easier to read

changed the regex for `daily-run.ts#getDailyEventSeedStarters`.

instead of `starters` plus `\d{4}\d{2}` for each species, it's now:
`s\d{4}` for speciesID followed by optionally:
`f\d{2}` for formIndex and/or
`v\d{2}` for variant.

> [!NOTE]
> I kept the previous `getDailyEventSeedStarters` and renamed it to `getDailyEventSeedStartersLegacy` to not break the 31st October custom seed. After that, it can get removed.

## Screenshots/Videos

<details><summary>Example</summary>

image

</details> 

## How to test the changes?

set a custom daily run seed in `overrided.ts`

example:
`starterss0003f01s0025v1s0150f02v2` should give a mega venusaur, t2 pikachu and t3 mega mewtwo y as starters

## Checklist
- [x] **I'm using `hotfix-1.11.2` as my base branch**

### DIFF
--- a/src/data/abilities/ability.ts
+++ b/src/data/abilities/ability.ts
@@ -34,6 +34,7 @@ import { MoveCategory } from "#enums/move-category";
 import { MoveFlags } from "#enums/move-flags";
 import { MoveId } from "#enums/move-id";
 import { MovePhaseTimingModifier } from "#enums/move-phase-timing-modifier";
+import { MovePriorityInBracket } from "#enums/move-priority-in-bracket";
 import { MoveResult } from "#enums/move-result";
 import { MoveTarget } from "#enums/move-target";
 import { MoveUseMode } from "#enums/move-use-mode";
@@ -4142,6 +4143,25 @@ export class ChangeMovePriorityAbAttr extends AbAttr {
   }
 }
 
+export class ChangeMovePriorityInBracketAbAttr extends AbAttr {
+  private readonly newModifier: MovePriorityInBracket;
+  private readonly moveFunc: (pokemon: Pokemon, move: Move) => boolean;
+
+  constructor(moveFunc: (pokemon: Pokemon, move: Move) => boolean, newModifier: MovePriorityInBracket) {
+    super(false);
+    this.newModifier = newModifier;
+    this.moveFunc = moveFunc;
+  }
+
+  override canApply({ pokemon, move }: ChangeMovePriorityAbAttrParams): boolean {
+    return this.moveFunc(pokemon, move);
+  }
+
+  override apply({ priority }: ChangeMovePriorityAbAttrParams): void {
+    priority.value = this.newModifier;
+  }
+}
+
 export class IgnoreContactAbAttr extends AbAttr {
   private declare readonly _: never;
 }
@@ -6721,6 +6741,7 @@ const AbilityAttrs = Object.freeze({
   BlockStatusDamageAbAttr,
   BlockOneHitKOAbAttr,
   ChangeMovePriorityAbAttr,
+  ChangeMovePriorityInBracketAbAttr,
   IgnoreContactAbAttr,
   PreWeatherEffectAbAttr,
   PreWeatherDamageAbAttr,
@@ -7238,7 +7259,7 @@ export function initAbilities() {
       .attr(DoubleBattleChanceAbAttr)
       .build(),
     new AbBuilder(AbilityId.STALL, 4)
-      .attr(ChangeMovePriorityAbAttr, (_pokemon, _move: Move) => true, -0.2)
+      .attr(ChangeMovePriorityInBracketAbAttr, (_pokemon, _move: Move) => true, MovePriorityInBracket.LAST)
       .build(),
     new AbBuilder(AbilityId.TECHNICIAN, 4)
       .attr(MovePowerBoostAbAttr, (user, target, move) => {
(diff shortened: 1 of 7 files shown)


src/data/daily-run.ts
```diff
@@ -58,11 +58,19 @@ export function getDailyRunStarters(seed: string): StarterTuple {
...
     starterSpeciesForm instanceof PokemonSpecies ? starterSpeciesForm : getPokemonSpecies(starterSpeciesForm.speciesId);
   const formIndex = starterSpeciesForm instanceof PokemonSpecies ? undefined : starterSpeciesForm.formIndex;
-  const pokemon = globalScene.addPlayerPokemon(starterSpecies, startingLevel, undefined, formIndex);
+  const pokemon = globalScene.addPlayerPokemon(
+    starterSpecies,
+    startingLevel,
+    undefined,
+    formIndex,
+    undefined,
+    variant !== undefined ? true : undefined,
```
Bertie690:
```suggestion
    !!variant,
```

src/data/daily-run.ts
```diff
@@ -215,14 +223,69 @@ function setDailyRunEventStarterMovesets(seed: string, starters: StarterTuple):
...
+    `(?<species${i}>s\\d{4})(?:(?<form${i}>f\\d{2})(?<variant${i}>v\\d{2})?|(?<variant${i}>v\\d{2})(?<form${i}>f\\d{2})?)?`;
+
+  const matcher = new RegExp(`starters${speciesRegex(1)}${speciesRegex(2)}${speciesRegex(3)}`);
+
+  const speciesConfigurations = matcher.exec(seed)?.groups;
+
+  if (!speciesConfigurations) {
+    const legacyStarters = getDailyEventSeedStartersLegacy(seed);
+    if (legacyStarters != null) {
+      console.log("Using legacy starter parsing for daily run seed.");
```
Bertie690:
```suggestion
      console.log("Using legacy starter parsing for daily run seed");
```

src/data/daily-run.ts
```diff
@@ -215,14 +223,69 @@ function setDailyRunEventStarterMovesets(seed: string, starters: StarterTuple):
...
+      return legacyStarters;
+    }
+    console.error("Invalid starters used for custom daily run seed!");
```
Bertie690:
```suggestion
    console.error("Invalid starters used for custom daily run seed!", seed);
```

src/data/daily-run.ts
```diff
@@ -215,14 +223,69 @@ function setDailyRunEventStarterMovesets(seed: string, starters: StarterTuple):
...
+    return null;
+  }
+
+  const speciesIds = getEnumValues(SpeciesId);
+
+  const starters: Starter[] = [];
```
Bertie690:
```suggestion
  const starters: Starter[] = [];

```

src/data/daily-run.ts
```diff
@@ -215,14 +223,69 @@ function setDailyRunEventStarterMovesets(seed: string, starters: StarterTuple):
...
+ * @example `/starterss0003f01s0025v01s0150f02v02`
+ * @param seed - The daily run seed
+ * @returns An array of {@linkcode Starter}s, or `null` if no valid match.
+ */
+// TODO: Rework this setup into JSON or similar - this is quite hard to maintain
+function getDailyEventSeedStarters(seed: string): StarterTuple | null {
+  const speciesRegex = i =>
+    `(?<species${i}>s\\d{4})(?:(?<form${i}>f\\d{2})(?<variant${i}>v\\d{2})?|(?<variant${i}>v\\d{2})(?<form${i}>f\\d{2})?)?`;
+
+  const matcher = new RegExp(`starters${speciesRegex(1)}${speciesRegex(2)}${speciesRegex(3)}`);
```
Bertie690: These should ideally be constants defined outside the function

Also, rather than repeating it thrice, I'd recommend using `String.prototype.matchAll()`.

Bertie690: Also also, I'd rather the order of form and variant be well defined (making both optional, but requiring one to precede the other when used together).
This would simplify the logic a ton and also more consistency = good IMO

src/data/daily-run.ts
```diff
@@ -215,14 +223,69 @@ function setDailyRunEventStarterMovesets(seed: string, starters: StarterTuple):
   }
 }
 
+/**
+ * Expects the seed to contain `starters` followed by 3 `s{\d{4}}` for the starters. The 4 digits are the species ID. \
+ * Each starter can optionally be followed by `f{\d{2}}` for the form index and `v{\d{2}}` for the variant. \
+ * The order of `f` and `v` does not matter.
+ * @example `/starterss0003f01s0025v01s0150f02v02`
```
Bertie690: Personally, I'd prefer a starting delimiter that _doesn't_ end with an `s` (`starterss` looks really weird)

Also do we want to consider placing commas between the starters for separation reasons?

src/data/daily-run.ts
fabske0: This has a different behavior, since `!!0` is false but a valid variant

src/data/daily-run.ts
fabske0: I'm fine with that

src/data/daily-run.ts
fabske0: Also thought about the double `s` but couldn't think of a good alternative. And i think commas a fine, also for consitency reasons with the moves matcher

fabske0:
> @fabske0 If you don't mind I am going to make the changes in the review comments shortly

If you want, you can make the changes

PR #6732, opened 2025-10-31
[Hotfix] Fix quick claw issues

emdeann:
Quick claw won't be blocked by priority blocking effects and will not activate on non-fight commands.

I ported the quick claw change from inner's TCM but i neglected re-adding the check for the command to be fight only, so that is added. I also re-added a change from when i was working on dynamic turn order - fractional priority is removed and in its place moves have a `MovePriorityModifier` which is either `LAST_IN_BRACKET`, `NORMAL`, or `FIRST_IN_BRACKET`. Effects there are pretty self-explanatory. There's no issue here as (modern) mainline doesn't have differing levels of fractional priority where it would be necessary to implement it that way.

## How to test the changes?
Play with quick claw against dazzling/QM/psychic terrain, try to run, etc. Going into the modifiers file and changing quick claw to always proc makes this easier

## Checklist


src/data/moves/move.ts
```diff
@@ -1060,17 +1061,19 @@ export abstract class Move implements Localizable {
...
-    }
-
     return priority.value;
   }
 
+  public getPriorityModifier(user: Pokemon, simulated = true) {
+    const modifierHolder = new NumberHolder(MovePriorityModifier.NORMAL);
+    applyAbAttrs("ChangeMovePriorityModifierAbAttr", {pokemon: user, simulated: simulated, move: this, priority: modifierHolder});
+    modifierHolder.value = user.getTag(BattlerTagType.BYPASS_SPEED) ? MovePriorityModifier.FIRST_IN_BRACKET : modifierHolder.value;
+    return modifierHolder.value;
```
```suggestion
    if (user.getTag(BattlerTagType.BYPASS_SPEED)) {
      return MovePriorityModifier.FIRST_IN_BRACKET;
    }
    const modifierHolder = new NumberHolder(MovePriorityModifier.NORMAL);
    applyAbAttrs("ChangeMovePriorityModifierAbAttr", { pokemon: user, simulated, move: this, priority: modifierHolder });
    return modifierHolder.value;
```

fabske0 (approved):

PR #6740, opened 2025-11-01
refactor: clean up `PreLeaveFieldClearWeatherAbAttr`

Bertie690:
N/A
The previous code was ugly, over-engineered and bad, relying on a `switch case` on the weather type to check against HARDCODED `AbilityId`s.
Also repeated `filter`ing is suboptimal makes me sad

Cleaned up `PreLeaveFieldClearWeatherAbAttr` to be much much simpler, taking the weather type as a private parameter to be checked later.
As such, it "should" work on abilities not named Desolate Land, Primordial Sea or Delta Stream lol

## Screenshots/Videos
N/A

## How to test the changes?
use primal weather and stuff

## Checklist

### DIFF
--- a/src/data/abilities/ab-attrs.ts
+++ b/src/data/abilities/ab-attrs.ts
@@ -2,7 +2,7 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
 import { globalScene } from "#app/global-scene";
 import { getPokemonNameWithAffix } from "#app/messages";
 import type { EntryHazardTag, SuppressAbilitiesTag } from "#data/arena-tag";
-import type { BattlerTag } from "#data/battler-tags";
+import { type BattlerTag, CritBoostTag } from "#data/battler-tags";
 import { getBerryEffectFunc } from "#data/berry";
 import { allAbilities, allMoves } from "#data/data-lists";
 import { SpeciesFormChangeAbilityTrigger, SpeciesFormChangeWeatherTrigger } from "#data/form-change-triggers";
@@ -53,7 +53,7 @@ import type {
   PokemonStatStageChangeCondition,
 } from "#types/ability-types";
 import type { Move, StatusEffectAttr } from "#types/move-types";
-import type { Closed, Exact } from "#types/type-helpers";
+import type { Closed, Exact, Mutable } from "#types/type-helpers";
 import { coerceArray } from "#utils/array";
 import { BooleanHolder, NumberHolder, randSeedFloat, randSeedInt, randSeedItem, toDmgValue } from "#utils/common";
 import { inSpeedOrder } from "#utils/speed-order-generator";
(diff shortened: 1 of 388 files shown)


Bertie690:
Needs fixing becayse i created branch off of wrong HEAD

Wlowscha:
The changes involving `formChangeItemModifiers` need to be ported over to the modifier rework branch.

Also, is there a good reason for this to introduce like 5 different TODOs?

src/data/abilities/ability.ts
```diff
@@ -3488,48 +3430,33 @@ export class PreLeaveFieldAbAttr extends AbAttr {
...
-        break;
+    const weatherType = globalScene.arena.getWeatherType();
+    if (weatherType !== this.weatherType) {
+      return false;
     }
-    return false;
+
+    if (!globalScene.arena.weather?.isImmutable()) {
+      return true;
+    }
```
DayKev: This abattr is for Primal Weathers, there's no point in checking if the weather is immutable or not or assigning a weather to the abattr.

Really, this can all be simplified down even further: <https://github.com/Despair-Games/poketernity/blob/beta/src/data/abilities/ab-attrs/pre-leave-field-ab-attrs.ts>

Bertie690: lol

Bertie690: > Really, this can all be simplified down even further: https://github.com/Despair-Games/poketernity/blob/beta/src/data/abilities/ab-attrs/pre-leave-field-ab-attrs.ts

I'm 90% sure that version will remove primal weathers if you override it with another one and then switch the original mon out (unless they added some sort of custom logic there).

Bertie690:
> Also explain in what way the changes to updateSpritePipelineData are relevant to this PR.

Will revert & pull oht into a bigger thing

src/data/abilities/ab-attrs.ts
```diff
@@ -2989,47 +2989,36 @@ export class PreLeaveFieldAbAttr extends AbAttr {
   apply(_params: Closed<AbAttrBaseParams>): void {}
 }
 
-/** Clears Desolate Land/Primordial Sea/Delta Stream upon the Pokemon switching out. */
+/**
+ * Ability attribute to clear a primal {@linkcode WeatherType} upon this Pokemon switching out.
+ */
+// TODO: Keeping a reference to the Pokemon from within the Weather class would make this far cleaner
```
DayKev: In what way?

src/data/abilities/ab-attrs.ts
```diff
@@ -2989,47 +2989,36 @@ export class PreLeaveFieldAbAttr extends AbAttr {
...
-        break;
+    const weatherType = globalScene.arena.getWeatherType();
+    if (weatherType !== this.weatherType) {
+      return false;
     }
-    return false;
+
+    if (!globalScene.arena.weather?.isImmutable()) {
+      return true;
+    }
```
DayKev:
```suggestion
```
This check doesn't make sense. You're already checking that the current weather is the correct type above, so it should never be possible to take this `if` branch.

src/data/abilities/ab-attrs.ts
```diff
@@ -2989,47 +2989,31 @@ export class PreLeaveFieldAbAttr extends AbAttr {
...
 
-/** Clears Desolate Land/Primordial Sea/Delta Stream upon the Pokemon switching out. */
+/**
+ * Ability attribute to clear a primal {@linkcode WeatherType} upon this Pokemon switching out.
+ */
 export class PreLeaveFieldClearWeatherAbAttr extends PreLeaveFieldAbAttr {
+  private readonly weatherType: Exclude<WeatherType, WeatherType.NONE>;
+
+  constructor(weatherType: Exclude<WeatherType, WeatherType.NONE>) {
+    super(false);
```
DayKev, line 3023:
```suggestion
    super(false);

```

PR #6743, opened 2025-11-02
[Bug] Fix Parental Bond reducing damage of spread moves on 2nd pokemon

Bertie690:
Parental Bond will no longer reduce the damage of strikes it did not enhance.

Bugfix and also code cleanliness

Broke up PB's damage modification to re-use Tinted Lens' `DamageBoostAbAttr` (renamed for consistency with `MovePowerBoostAbAttr`), alongside a condition that explicitly checks that the move's initial hit count was >1 (ensuring it doesn't reduce damage on subsequent hits)

## Screenshots/Videos
N/A

## How to test the changes?
`pnpm test parental-bond.test.ts` (1 new test)

## Checklist

### DIFF
--- a/src/data/arena-tag.ts
+++ b/src/data/arena-tag.ts
@@ -319,7 +319,11 @@ export class MistTag extends SerializableArenaTag {
     cancelled.value = true;
 
     if (!simulated) {
-      globalScene.phaseManager.queueMessage(i18next.t("arenaTag:mistApply"));
+      globalScene.phaseManager.queueMessage(
+        i18next.t("arenaTag:mistApply", {
+          pokemonNameWithAffix: getPokemonNameWithAffix(this.getSourcePokemon()),
+        }),
+      );
     }
 
     return true;
--- a/src/data/trainers/trainer-config.ts
+++ b/src/data/trainers/trainer-config.ts
@@ -5475,13 +5475,10 @@ export const trainerConfigs: TrainerConfigs = {
     )
     .setPartyMemberFunc(
       1,
-      getRandomPartyMemberFunc([SpeciesId.SCIZOR, SpeciesId.KLEAVOR], TrainerSlot.TRAINER, true, p => {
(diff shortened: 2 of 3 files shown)


src/data/moves/move.ts
```diff
@@ -1117,27 +1117,42 @@ export abstract class Move implements Localizable {
...
+    
+    // ...cannot enhance moves hitting multiple targets unless specified
     const { targets, multiple } = getMoveTargets(user, this.id);
-    const isMultiTarget = multiple && targets.length > 1;
+    if (restrictSpread && multiple && targets.length > 1) {
+      return false;
+    };
+
+    // ...cannot enhance status moves, including ally-targeting Pollen Puff
+    const target = globalScene.getField()[targets[0]];
```
DayKev: `getMoveTargets()` returns a list of legal targets, not the list of Pokémon the move is actually targeting.
Previously the Pollen Puff check looked to see if the ally was a legal target (rather than whether the Pokémon being targeted was the user's ally), so Pollen Puff could never hit more than once with Parental Bond.
Your change makes it so the first legal target is checked to see if it's the Pokémon's ally, which would probably only be true if `BattlerIndex.ENEMY_2` is targeting `BattlerIndex.ENEMY`.
Both are wrong, but in different ways lol.

src/data/moves/move.ts
```diff
@@ -1139,8 +1138,11 @@ export abstract class Move implements Localizable {
     };
 
     // ...cannot enhance status moves, including ally-targeting Pollen Puff
-    const target = globalScene.getField()[targets[0]];
-    if (user.getMoveCategory(target, this) === MoveCategory.STATUS) {
+    if (target != null) {
+      if (user.getMoveCategory(target, this) === MoveCategory.STATUS) {
+        return false;
+      }
+    } else if (this.category === MoveCategory.STATUS) {
```
Bertie690: why what help ugh

this is so janky

DayKev (approved):

PR #6751, opened 2025-11-04
[Hotfix] Prevent crash with two NG Pokemon fainting at the same time

emdeann:
Game will not crash when two Neutralizing Gas users faint in the same turn

DC: https://discord.com/channels/1125469663833370665/1434896848467923027

If only one NG Pokemon is remaining, the game finds it to activate its passive. This was missing a check for this pokemon to be undefined, which would happen in the case that both NG pokemon were in a fainted state (`sourceCount` is decremented one at a time but both pokemon fainted at once, so for that small time it is desynced). 

This is fixed by just adding a null guard

## How to test the changes?
Test with Precipice Blades on the session data provided in the DC post linked above

## Checklist

### DIFF
--- a/src/data/arena-tag.ts
+++ b/src/data/arena-tag.ts
@@ -1532,6 +1532,11 @@ export class SuppressAbilitiesTag extends SerializableArenaTag {
       const setter = globalScene
         .getField(true)
         .filter(p => p.hasAbilityWithAttr("PreLeaveFieldRemoveSuppressAbilitiesSourceAbAttr", false))[0];
+      // Setter may not exist if both NG Pokemon faint simultaneously
+      if (setter == null) {
+        return;
+      }
+
       applyOnGainAbAttrs({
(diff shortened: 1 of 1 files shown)


Madmadness65 (approved):

src/data/arena-tag.ts
```diff
@@ -1532,6 +1532,11 @@ export class SuppressAbilitiesTag extends SerializableArenaTag {
       const setter = globalScene
         .getField(true)
         .filter(p => p.hasAbilityWithAttr("PreLeaveFieldRemoveSuppressAbilitiesSourceAbAttr", false))[0];
```
Bertie690, line 1534: This should be `find`
```suggestion
        .find(p => p.hasAbilityWithAttr("PreLeaveFieldRemoveSuppressAbilitiesSourceAbAttr", false));

```

fabske0 (approved):

PR #6752, opened 2025-11-04
[Hotfix] Fix Pollen Puff crash and incorrect failure

### LINKED ISSUE #6741: [Bug] Pollen Puff bug
### Describe the bug

The move pollen puff is not operating correctly, either failing when it shouldn't, or outright crashing the game.

### Reproduction

To reproduce, fight a double battle with one pokemon using the move pollen puff (I've been doing this with Ledian) which is slower than its partner (in my case, magmar and miltank).  Target that ally with pollen puff, and have that ally target an enemy pokemon (this might not actually be essential).  Pollen puff will say that it fails (even if that ally is missing HP).

The second case has a similar setup, but you must be finishing off the last enemy pokemon.  In this case, when pollen puff executes the game simply crashes.  The current work around is to reload the game and not use pollen puff for that turn.

### Expected behavior

The move is expected to heal injured allies when targeting them, and it is expected not to crash the game when used as the last move of a battle.

### Screenshots / Videos

_No response_

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

emdeann:
What are the changes the user will see?
Pollen Puff will not crash the game when used with no enemies remaining and will not (almost) always fail when used on an ally.

Fixes #6741 

The condition on Pollen Puff would cause on error if the target was not defined (due to fainting). I added `target == null` as the first condition so the check will short circuit if the target doesn't exist. (This means the condition will pass if the target doesn't exist and the move will fail as normal due to having no targets).

The check also previously was checking that the target (if an ally) *had* a heal block tag, not that it *didn't*.

## How to test the changes?
Use pollen puff after eliminating opponents or on an ally.

## Checklist


fabske0 (approved):

src/data/moves/move.ts
```diff
@@ -10700,7 +10700,7 @@ export function initMoves() {
       .attr(HealOnAllyAttr, 0.5, true, false)
       .ballBombMove()
       // Fail if used against an ally that is affected by heal block, during the second failure check
-      .condition((user, target) => target.isOpponent(user) || !!target.getTag(BattlerTagType.HEAL_BLOCK), 2),
+      .condition((user, target) => target == null || target.isOpponent(user) || !target.getTag(BattlerTagType.HEAL_BLOCK), 2),
```
DayKev, line 10703: Shouldn't it just be changed to `target?.isOpponent(user)`? Making "there is no target" a success condition seems incorrect.

emdeann: My thought process is if there's no target the move shouldn't fail due to pollen puff's "can I heal my ally" condition. It should fail due to the regular "are there any targets" check which comes later. It would also be a bit more involved than just doing `target?.isOpponent(user)` because you'd have to also do `!target?.getTag(BattlerTagType.HEAL_BLOCK)`, but that would actually evaluate `true` with `target` undefined.

This is a bandaid because `target` likely shouldn't ever be undefined to begin with

DayKev: Ah, gotcha.

damocleas (approved):

PR #6761, opened 2025-11-08
Converted various held item data types to interfaces

Bertie690:
N/A
Interfaces are faster than types for TS to parse, most notably when used to extend and create new types (like in the modifier rework branch).
TS is better able to cache results with interface extension than type intersection, plus the extra consistency is good.

Changed a bunch of interfaces to types

## Screenshots/Videos
N/A

## How to test the changes?
N/A

## Checklist

### DIFF
--- a/src/@types/held-item-data-types.ts
+++ b/src/@types/held-item-data-types.ts
@@ -5,7 +5,7 @@ import type { AllHeldItems } from "#items/all-held-items";
 import type { CosmeticHeldItem } from "#items/held-item";
 import type { InferKeys } from "#types/helpers/type-helpers";
 
-export type HeldItemData = {
+export interface HeldItemData {
   /**
    * Number of items in the stack, can also be used to track cooldown
    */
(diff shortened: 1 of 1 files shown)


DayKev:
> Interfaces are faster than types for TS to parse.

They are? Source?

Bertie690:
https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections

DayKev:
> https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections

That's referring to doing `type1 & type2 ...`, but types themselves are not inherently slower than interfaces. So it only matters if you want to merge types together.

Bertie690:
> That's referring to doing type1 & type2 ..., but types themselves are not inherently slower than interfaces. So it only matters if you want to merge types together.

We are literally doing that in the held item branch

Bertie690:
> export type HeldItemSpecs = HeldItemData & {
>  id: HeldItemId;
> };

@DayKev On a side note it might be worth looking into a way of enforcing a consistent style for this (similar to array stuff).
I'd prefer "interface over type where possible" (given you can implement a class with it, etc.), but consistnecy would be ideal either way.

DayKev:
> On a side note it might be worth looking into a way of enforcing a consistent style for this (similar to array stuff). I'd prefer "interface over type where possible" (given you can implement a class with it, etc.), but consistnecy would be ideal either way.

https://biomejs.dev/linter/rules/use-consistent-type-definitions/

DayKev:
> > That's referring to doing type1 & type2 ..., but types themselves are not inherently slower than interfaces. So it only matters if you want to merge types together.
> 
> We are literally doing that in the held item branch

The PR description just says "Interfaces are faster than types for TS to parse."

Bertie690:
> The PR description just says "Interfaces are faster than types for TS to parse."

updated desc

SirzBenjie:
I don't care either way. I don't agree with the biome lint about it having higher cognitive load. I think interfaces are good for things that are really defined as objects. Types are good for other things. I'm not sure if they're faster or not. For type intersection, of course they would be.

Bertie690:
> I don't agree with the biome lint about it having higher cognitive load. 

Yeah in that case I'd argue we should turn the lint on to prefer interfaces for object types.

DayKev (approved):

PR #6779, opened 2025-11-15
refactor: decouple `setOrder` array from `MovePhase`s; make `GameManager.setTurnOrder` synchronous

Bertie690:
N/A

The `setOrder` array that Dean added inside #6036 is tightly coupled to `MovePhases` and the related `MovePhasePriorityQueue`, thus applying exclusively to move uses.

Given it is ostensibly supposed to be useful for future MEs that rely on more gimmicky, turn-order-dependent strategies, this omission results in visible inconsistencies with damage order and similar (which are expected to occur in turn order as well).

(This also renders checking the order of ostensibly speed-dependent effects like Dancer turn order more annoying than is necessary, since you cannot use the override for any other type of dynamic speed order interaction.)
1. Moved the `setOrder` array out into a basic `TurnCommandManager` class  whose current only job is managing said array.
2. Moved checks for it from exclusively being in `MovePhasePriorityQueue` to being checked during `sortInSpeedOrder` (to apply to all non-phase speed order interactions, not just moves)
3. Made `game.setTurnOrder` not `async` anymore since it now effectively sets a single variable (and has done so for over 2 months by now).

> [!NOTE]
> I will bulk-remove `async`s from existing callsites in a subsequent PR (possibly after the modifier rework to avoid shooting myself in the foot).

> [!NOTE]
> I kept in the (admittedly somwwhat questionable) behavior of resetting the `setOrder` array on queue manager clearance.
> If we really want this to be usable for MEs (which was the only reason I agreed to keep it inside `src` in the first place), this will have to be changed in a follow-up PR. (I have made one to that end, which will be rebased/undrafted after this is merged.)

## Screenshots/Videos
N/A

## Checklist

### DIFF
--- a/test/tests/abilities/stall.test.ts
+++ b/test/tests/abilities/stall.test.ts
@@ -27,7 +27,7 @@ describe("Abilities - Stall", () => {
       .moveset([MoveId.QUICK_ATTACK, MoveId.TACKLE]);
   });
 
-  /**
+  /*
    * References:
    * https://bulbapedia.bulbagarden.net/wiki/Stall_(Ability)
    * https://bulbapedia.bulbagarden.net/wiki/Priority


emdeann (dismissed):

Wlowscha (approved):

Wlowscha (dismissed):

SirzBenjie (dismissed):

test/helpers/field-helper.ts
```diff
@@ -64,36 +65,27 @@ export class FieldHelper extends GameManagerHelper {
...
-   * (such as for comparison with other speed order-related mechanisms); default `false`
-   * @returns An array containing all on-field {@linkcode Pokemon} in order of descending Speed. \
-   * Speed ties are returned in increasing order of index.
-   *
-   * @remarks
-   * This does not account for Trick Room as it does not modify the _speed_ of Pokemon on the field,
-   * only their turn order.
+   * @param indices - (Default `false`) Whether to only return {@linkcode BattlerIndex}es instead of full Pokemon objects
+   *   (such as for comparison with other speed order-related mechanisms)
+   * @param ignoreOverride - (Default `true`) Whether to ignore preset turn orders and speed-reversing effects (like Trick Room)
```
DayKev, line 70: I can't imagine a scenario where you'd want to use this to get the overridden order that was set by the test framework.

DayKev (approved):

PR #6781, opened 2025-11-15
misc: persist `setOrder` array between turns; remove duplicate calls in test cases

Bertie690:
> [!NOTE]
> The actual changes in this PR are very minor, consisting of ~10 locs inside `src` and a few rounds of regex on `test` code.
> Moreover, **ALL** of the issues this PR fixes were discussed at length in prior discussions.

N/A

As I discussed in #6779, the `setOrder` array used for modifying turn order in MEs/tests should default to being preserved between turns. 
Currently, any external system that would want to customize the order in which Pokemon act would have to reset the preset turn order every single turn, which is backwards and odd.

Moreover, I noticed several tests were setting the turn order array to its prior value multiple times in a row, which (as of now) is 100% useless.

Removed code that would clear the `setOrder` on turn end. It will now persist until set to a different value, the game is reset or a new wave starts (whichever one comes first).

Removed duplicate calls to `game.setTurnOrder` in the same test case with the exact same arguments. This is probably(?) not comprehensive, but the semi-conservative regex I used is guaranteed to produce no false positives.

REGEX:
`(?<foo>.*setTurnOrder\(.*?\);\n)((?:(?!\}\);).|\n)+)\k<foo>` => `$1$2`

(I might just make the function throw an error on duplicate setting in a later PR to see if any remaining cases exist.)
## Screenshots/Videos
N/A

## How to test the changes?
run the existing tests, which should change next to nothing (or literally nothing after #6779 is merged)

## Checklist


SirzBenjie (dismissed):
This is good, but can you edit the PR description to clarify that turn order is now reset between waves? (I believe that's how it works from what I gathered of `src/battle-scene.ts`)

Bertie690:
> This is good, but can you edit the PR description to clarify that turn order is now reset between waves? (I believe that's how it works from what I gathered of `src/battle-scene.ts`)

Done

PR #6782, opened 2025-11-15
fix(move): save Retaliate faint count between waves

Bertie690:
Retaliate now saves its tracker across waves for player Pokemon.
The AI can only Tera once per battle now (technically it's however many times the player is allowed to Terastallize per battle, currently set to 1), replacing the previous logic that prevented the same Pokémon from re-Terastallizing if somehow revived.

Retaliate's effect did not persist between waves for player pokemon, and was also very inefficient due to storing data about the faint histories of all Pokemon within _every_ battle (whereas Retaliate itself only cares about the immediately preceding turn).

Reworked the `playerFaintHistory` and `enemyFaintHistory` arrays inside `Battle` into pairs of boolean counter variables used to track whether an enemy fainted last turn or not. 1 counter (`enemyFaintedLastTurn`) is the actual serialized value used by Retaliate; the other (`enemyFaintedThisTurn`) is merely a temporary storage value used to avoid resetting the actual counter during the same turn it is set.

Added corresponding counters to `Arena` for player Pokemon, the former of which is saved into save data if set to `true` (and set to `undefined` otherwise to save on space in the 99% of cases where it is not set

Added a new method for `Arena` to perform weather/terrain lapsing alongside the aforementioned boolean counter toggling.

Reworked the singular other place faint history was used - the revival blessing hotfix for enemy tera calcs. Now it just checks the "teras used" counter instead (cleaner and consistent with player calcs)

Tests

## Screenshots/Videos
Test failed before and passes now
image

## How to test the changes?
`pnpm test retaliate.test.ts` (4 new/reworked tests)

## Checklist


SirzBenjie (dismissed):

src/battle.ts
```diff
@@ -1,3 +1,8 @@
+// biome-ignore-start lint/correctness/noUnusedImports: TSDoc imports
+import type { Arena } from "#field/arena";
+
+// biome-ignore-end lint/correctness/noUnusedImports: TSDoc imports
```
DayKev:
```suggestion
import type { Arena } from "#field/arena";
```

src/battle.ts
```diff
@@ -84,14 +84,36 @@ export class Battle {
...
   public enemyFaints = 0;
-  public playerFaintsHistory: FaintLogEntry[] = [];
-  public enemyFaintsHistory: FaintLogEntry[] = [];
+  /**
+   * Whether an {@linkcode EnemyPokemon} fainted last turn in battle.
+   *
+   * Used for Retaliate's "double power" condition.
+   * @defaultValue `false`
+   * @see {@linkcode Arena.playerFaintedLastTurn} \
+   * The corresponding (non-serialized) tracker for player Pokemon (persists between waves)
```
DayKev:
```suggestion
   * The corresponding (serialized) tracker for player Pokemon (persists between waves)
```

DayKev: Also I don't like that this is inconsistent between player and enemy. Either each wave is a new battle or it isn't.

src/phases/turn-end-phase.ts
```diff
@@ -24,6 +23,8 @@ export class TurnEndPhase extends FieldPhase {
     super.start();
 
     globalScene.currentBattle.incrementTurn();
+    globalScene.arena.performTurnEndEffects();
```
DayKev: This changes the timing of weather and terrain ending.

src/system/arena-data.ts
```diff
@@ -15,20 +15,24 @@ export interface SerializedArenaData {
...
   public tags: ArenaTag[];
   public positionalTags: SerializedPositionalTag[] = [];
   public playerTerasUsed: number;
+  public playerFaintedLastTurn: boolean;
 
   constructor(source: Arena | SerializedArenaData) {
     // Exclude any unserializable tags from the serialized data (such as ones only lasting 1 turn).
     // NOTE: The filter has to be done _after_ map, data loaded from `ArenaTagTypeData`
     // is not yet an instance of `ArenaTag`
+    // TODO: Move this to a migrate script
```
DayKev: Wdym? This has nothing to do with migration, this is for loading save data.

DayKev (commented):
This includes other changes that should be documented in the PR description.

src/phases/turn-end-phase.ts
Bertie690: Idt our current impl is even correct

src/battle.ts
Bertie690: They are only inconsistent to the degree that player/enemy data is - we only save player stuff since a new set of enemies are generated each wave.

Bertie690:
> This includes other changes that should be documented in the PR description.

Which ones?

DayKev:
> > This includes other changes that should be documented in the PR description.
> 
> Which ones?

The change to enemy tera and moving the terrain/weather end-of-turn stuff to a helper function.

Bertie690:
done

_sigh_

DayKev:
> Reworked the singular other place faint history was used (enemy tera calcs, wrongly)

That wasn't "wrongly", it was preventing a Pokémon from re-Terastallizing if revived somehow.

Wlowscha (dismissed):

Bertie690:
I will draft this until Benjie rolls out Zod schemas.

PR #6799, opened 2025-11-28
[Dev] Update Biome from `2.3.2` to `2.3.8`

DayKev:
N/A

Keeping deps up-to-date.
This update will also allow the removal of the `// biome-ignore` suppressions used by TSDoc imports. (Useful regex https://regex101.com/r/UEGV68/1)

- Biome updated to `2.3.8`.
- `files.includes` updated to use `!!` syntax for some entries.
- Biome run across the entire repo due to changes in the formatter.

## Checklist

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -38,6 +38,8 @@
       // this file is too big
       "!src/data/balance/tm-species-map.ts",
       // there's some sort of bug when Biome parses this file
+      // relating to recursive variable assignment
+      // cf https://github.com/biomejs/biome/issues/8204
       // TODO: remove this exclusion when the bug is fixed
       "!!test/test-utils/setup/test-end-log.ts"
     ]


biome.jsonc
```diff
@@ -25,18 +25,21 @@
...
-      "!**/*.css",
-      "!**/*.html",
+      "!!*.css",
+      "!!*.html",
       // this file is too big
-      "!**/src/data/balance/tm-species-map.ts"
+      "!src/data/balance/tm-species-map.ts",
+      // there's some sort of bug when Biome parses this file
+      // TODO: remove this exclusion when the bug is fixed
+      "!!test/test-utils/setup/test-end-log.ts"
```
Bertie690, line 44: what bug?

DayKev: https://discord.com/channels/1132231889290285117/1132231889911029825/1443806826809528421

DayKev: The issue appears to be due to recursively assigning to a variable. cf https://github.com/biomejs/biome/issues/8204

Why did you copy a function from Vitest source instead of just importing it? Is it modified in a way that you can't modify it after importing it?

biome.jsonc
```diff
@@ -25,18 +25,21 @@
...
+      "!!typedoc",
       // TODO: lint css and html?
-      "!**/*.css",
-      "!**/*.html",
+      "!!*.css",
+      "!!*.html",
       // this file is too big
-      "!**/src/data/balance/tm-species-map.ts"
+      "!src/data/balance/tm-species-map.ts",
+      // there's some sort of bug when Biome parses this file
```
DayKev, line 40:
```suggestion
      // there's some sort of bug when Biome parses this file
      // relating to variable assignment within `while` loops
      // cf https://github.com/biomejs/biome/issues/8204
```

PR #6817, opened 2025-12-03
[Docs] Minor typofixes/addendums to `.md` files

Bertie690:
N/A
typos + docs

fixed a few typos, notably using "Typescript" in `CONTRIBUTING.md`, an accidental singular in `linting.md` and indent issues with nested bullets

Moved the casing suggestions for `localization.md` to its own section to make it more prominent

## Checklist

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,7 +9,7 @@
 {
   "name": "Node.js & TypeScript",
   // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-  "image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:4-24-bookworm",
   "features": {
     "ghcr.io/devcontainers/features/github-cli:1": {
       "installDirectlyFromGitHubRelease": true,
@@ -18,6 +18,9 @@
     "ghcr.io/devcontainers-extra/features/pnpm:2": {
       "version": "latest"
     }
+  },
+  "containerEnv": {
+    "PORT": "8000"
   },
   "customizations": {
     "vscode": {
--- a/.github/workflows/test-shard-template.yml
+++ b/.github/workflows/test-shard-template.yml
@@ -3,27 +3,19 @@ name: Test Template
 on:
   workflow_call:
     inputs:
-      project:
-        required: true
-        type: string
       shard:
         required: true
         type: number
       totalShards:
         required: true
         type: number
-      skip:
-        required: true
-        type: boolean
-        default: false
 
 jobs:
   test:
-    # We can't use dynmically named jobs until https://github.com/orgs/community/discussions/13261 is implemented
+    # We can't use dynamically named jobs until https://github.com/orgs/community/discussions/13261 is implemented
     name: Shard
     timeout-minutes: 10
     runs-on: ubuntu-latest
-    if: ${{ !inputs.skip }}
     steps:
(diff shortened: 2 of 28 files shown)


Madmadness65 (approved):

docs/localization.md
```diff
@@ -97,17 +97,20 @@ If this feature requires new text, the text should be integrated into the code w
...
 
 [^2]: For those wondering, the reason for choosing English specifically is due to it being the master language set in Pontoon (the program used by the Translation Team to perform locale updates).
 If a key is present in any language _except_ the master language, it won't appear anywhere else in the translation tool, rendering missing English keys quite a hassle.
 
+> [!IMPORTANT]
+> The Dev and Translation teams have strict requirements for ensuring consistency of newly added locales entries.
+>
+> - File names should be in `kebab-case`. Example: `trainer-names.json`
+> - Key names should be in `camelCase`. Example: `aceTrainer`
+>    - If you make use of i18next's inbuilt [context support](https://www.i18next.com/translation-function/context), you need to use `snake_case` for the context key. Example: `aceTrainer_male`
```
fabske0: Might be good to also point to the automated workflow here. As a way to check if they did it correct.
With that, it might also happen that the workflow fails if they use a context which isn't supported by the workflow script and therefore needs to be added to the [supported list](https://github.com/pagefaultgames/pokerogue-locales/blob/7da1bd12676b2641497628062390d30d0479d1c6/.github/scripts/locales-format-checker/constants.js#L30). Otherwise LGTM

docs/linting.md
```diff
@@ -54,8 +54,9 @@ We primarily use Biome's [recommended ruleset](https://biomejs.dev/linter/rules/
 
 Some things to consider:
 
-- We have disabled rules that prioritize style over performance, such as `useTemplate`.
-- Some rules are currently marked as warnings (`warn`) to allow for gradual refactoring without blocking development. **Do not write new code that triggers these rules!**
+- We have disabled rules that prioritize style over performance (such as `useTemplate`),
+as well as ones that risk introducing accidental errors (such as `useExponentiationOperator`).
+- Some rules are currently marked as `info` or `warn` to allow for gradual refactoring without blocking development. **Do not write new code that triggers these rules!**
 - The linter is configured to ignore specific files and folders (such as excessively large files or ones in need of refactoring) to improve performance and focus on actionable areas.
```
DayKev:
```suggestion
- The `biome.jsonc` file contains comments explaining reasons various settings and rules are set as they are.
- Some rules are currently marked as `info` or `warn` to allow for gradual refactoring without blocking development. **Do not write new code that triggers these rules!**
```

The "style over performance" line is incorrect, template literals are basically the same performance as string concat currently (see images below), our stated reason for turning that rule off is invalid.

<details><summary>Benchmarks</summary>
<p>

image
image
image

</p>
</details>

fabske0: The mention of the comments is kinda redundant as it basically says the same as the citation just above this text

DayKev: Ah, that's not showing up in Github's file view for some reason. In that case, this section doesn't need to be here at all.

DayKev:
```suggestion
```

Bertie690: sure but I'm keeping in the "don't write new code triggering the rules" part

Bertie690: > The "style over performance" line is incorrect, template literals are basically the same performance as string concat currently (see images below), our stated reason for turning that rule off is invalid.

yea we should probably turn that thing on sometime lol

fabske0 (approved):

SirzBenjie (approved):

PR #6816, opened 2025-12-03
[Docs] Rewrite `comments.md` for clarity

Bertie690:
## What are the changes the user will see?
N/A

## Why am I making these changes?
I wanted to review `comments.md` after ~7 months as part of the style guide changes, and realized the code it was referencing had gotten out of sync.

## What are the changes from a developer perspective?
Updated `comments.md` with new instructions
## Screenshots/Videos
N/A
## How to test the changes?
N/A
## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`pnpm test:silent`)

PR #6818, opened 2025-12-03
[Dev] Enable type-checking for `typedoc.config.js`

Bertie690:
N/A

the file lacked ts check

Added `// @ts-check`

Added comment to include types from `global.d.ts` to allow interface merging

## Screenshots/Videos
N/A

## Checklist


typedoc.config.js
```diff
@@ -4,12 +4,16 @@
...
+// @ts-check
+/// <reference path="./global.d.ts" />
+
 import { globSync } from "node:fs";
 
 const dryRun = !!process.env.DRY_RUN?.match(/true/gi);
 
 /**
- * <!-- @type {Partial<import("typedoc").TypeDocOptions>} -->
+ * <!-- @satisfies {Partial<import("typedoc").TypeDocOptions>} -->
```
DayKev, line 16: Why are the `<!-- -->` needed?

DayKev:
I'm getting errors when viewing the file is VSCode.
<details><summary>Errors</summary>
<p>

image

</p>
</details> 

There are no errors when running typecheck, but that's probably because of this:
```
PS I:\pokerogue repos\PokeRogue dev> pnpm exec tsc --noEmit typedoc.config.js
error TS6504: File 'typedoc.config.js' is a JavaScript file. Did you mean to enable the 'allowJs' option?
  The file is in the program because:
    Root file specified for compilation


Found 1 error.
```

typedoc.config.js
Bertie690: To avoid showing the comment in the hover.

DayKev: Oh it's because VSCode is dumb? Figures.

Bertie690:
> I'm getting errors when viewing the file is VSCode.

that mighr be due to it not being inside any compilation files? IDK

Bertie690:
@fabske0 It might be due to the file not being included in compilation inside either tsconfig.json or jsconfig.json.

I have changes on #6816 for this, so will move later.

fabske0:
> @fabske0 It might be due to the file not being included in compilation inside either tsconfig.json or jsconfig.json.

Maybe, but if it's a config issue, shouldn't we have the errors too? Also those errors are weird

Bertie690:
_shrugs_

DayKev:
Okay I looked at the source and also tested using the type in a `.ts` file and everything looks correct, it's only within this specific `.js` file that the type is being resolved incorrectly in VSCode. So... whatever, I guess it's fine?

PR #6819, opened 2025-12-03
[Misc] Dedupe + move phaser method stubs into separate file

Bertie690:
N/A
I noticed that `test-file-initialization` had the old version of the `setPositionRelative` functions.

I dislike duplicate code.
Moved the method stubs to a separate file (`#app/extensions`, though feel free to rename if a better name is found) containing global runtime extensions to Phaser types' methods.
The file is imported in `init.ts` since it's imported in both cases, and (being earlier than it was called in both cases) _shouldn't_ heinously break anything.

Moved the `setPositionRelative` declarations from `phaser.index.ts` into said file for locality of behavior and to avoid needless re-declaration.

## Screenshots/Videos
N/A
## How to test the changes?
play the game

## Checklist

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,7 +9,7 @@
 {
   "name": "Node.js & TypeScript",
   // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-  "image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:4-24-bookworm",
   "features": {
     "ghcr.io/devcontainers/features/github-cli:1": {
       "installDirectlyFromGitHubRelease": true,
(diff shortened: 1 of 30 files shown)


test/test-utils/test-file-initialization.ts
```diff
@@ -42,7 +42,8 @@ function initTestFile(): void {
  * @todo Investigate why this resets on new test suite start
  */
 function setupStubs(): void {
-  Object.defineProperties(global, {
+  // TODO: Make this type safe
```
DayKev, line 45: How would you do that?

src/typings/phaser/index.d.ts
DayKev, line 1: Ugh we have `src/typings` and `src/@types`? This needs to be consolidated to `src/@types`.

src/extensions.ts
DayKev, line 1: Pkty named it `phaser-extensions.ts`. 🤷

test/test-utils/test-file-initialization.ts
Bertie690: 🤷

src/extensions.ts
Bertie690: We can easily rename it in another PR

It would be so quick

DayKev: Yeah it doesn't really matter, just adding the suggestion if wanted.

DayKev (approved):

PR #6820, opened 2025-12-03
[Dev] Add lefthook command to update packages on checkout/merge

DayKev:
N/A

Devs won't need to run `pnpm i` manually anymore whenever packages are updated.

`pnpm i` will be run whenever a branch is checked out or `git pull` is used.

## Checklist

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -25,4 +25,6 @@ post-merge:
 post-checkout:
   commands:
     update-packages:
-      run: pnpm i
+      # cf https://git-scm.com/docs/githooks#_post_checkout:
+      # The 3rd argument is 1 for branch checkouts and 0 for file checkouts.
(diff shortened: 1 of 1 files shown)


lefthook.yml
```diff
@@ -18,4 +18,11 @@ pre-commit:
...
-      run: pnpm update-submodules
\ No newline at end of file
+      run: pnpm update-submodules
+    update-packages:
+      run: pnpm i
+
+post-checkout:
+  commands:
+    update-packages:
+      run: pnpm i
```
Bertie690:
```suggestion
post-checkout:
  commands:
    update-packages:
      # cf https://git-scm.com/docs/githooks#_post_checkout:
      # The 3rd argument is 1 for branch checkouts and 0 for file checkouts.
      run: if test {3} -eq "1"; then pnpm i; fi
```

PR #6825, opened 2025-12-04
[GitHub] Update and optimize workflows

SirzBenjie:
N/A

#6687 added a step to combine the shards into one report. However, this was checking out the entire repository, which takes an extra 30s. This delays the test runs by that much.
Running the final job that combines the reports from each shard does not actually need the repo. All it needs is vitest.
(However, we have a custom reporter written, so I've used sparse-checkout to only checkout that portion).

The reason why sparse-checkout failed before is because vitest.config.ts was previously expecting the plugins to exist in `src` and be loaded. The namespace plugin requires `locales/en` folder to exist _while running the plugin_.

- Updated the action dependencies used by our workflows from version 4 to version 6 for `setup-node`, `checkout`, and `download-artifact`
- Modified `vite.config.ts` to not import and run the packages if `MERGE_REPORTS` env var is set to a truthy value
- Added `"test:merge-reports": "MERGE_REPORTS=1 vitest run --merge-reports=test-results --silent=passed-only --configLoader runner"` to `package.json` as a shorthand for running the commmand
- Modified the `test.yml` workflow to use sparse-checkout for the repository, to only checkout files in the root as well as `test/test-utils/reporters/custom-default-reporter.ts` to speedup repsitory checkout time
- Created a [composite action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action#creating-a-composite-action-within-the-same-repository) for DRYness in our workflows

## Screenshots/Videos
image

## How to test the changes?
Just look at the workflow runs on this PR.

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,11 +46,11 @@ jobs:
       # don't stop upon 1 shard failing
       fail-fast: false
       matrix:
-        shard: [1, 2, 3, 4]
+        shard: [1, 2, 3, 4, 5]
     uses: ./.github/workflows/test-shard-template.yml
     with:
       shard: ${{ matrix.shard }}
-      totalShards: 4
+      totalShards: 5
 
   check-results:
     name: Check Test Results


.github/workflows/deploy-beta.yml
```diff
@@ -15,20 +15,12 @@ jobs:
...
-      - name: Install pnpm
-        uses: pnpm/action-setup@v4
-
-      - uses: actions/setup-node@v4
-        with:
-          node-version-file: ".nvmrc"
-
-      - name: Install dependencies
-        run: pnpm i
+      - uses: ./.github/actions/setup-deps
```
Bertie690, line 23: What will this show for name in the action log?

SirzBenjie: Literally that path.

image

damocleas (approved):

vite.config.ts
```diff
@@ -6,11 +6,15 @@
...
-import { minifyJsonPlugin } from "./src/plugins/vite/vite-minify-json-plugin";
 
 export const defaultConfig: UserConfig = {
-  plugins: [tsconfigPaths(), minifyJsonPlugin(["images", "battle-anims"], true), LocaleNamespace()],
+  plugins: process.env.MERGE_REPORTS
+    ? []
+    : [
+        tsconfigPaths(),
+        require("./src/plugins/vite/vite-minify-json-plugin").minifyJsonPlugin(["images", "battle-anims"], true),
+        require("./src/plugins/vite/namespaces-i18n-plugin").LocaleNamespace(),
```
DayKev, line 16: Is using `require()` necessary?

SirzBenjie: `await import` would require this being asynchronous, which is the only other alternative I'm aware of.

DayKev: We should have top level await support, though it's not working when I try for some reason.

image
image

.github/workflows/deploy-beta.yml
Bertie690: Do we want to have a custom name for it?

IDK "setup deps" seems a bit more user-friendly than "run le workflow file"

.github/workflows/tests.yml
```diff
@@ -60,28 +60,22 @@ jobs:
...
+        uses: actions/checkout@v6
         with:
-          submodules: "recursive"
-
-      - name: Install pnpm
-        uses: pnpm/action-setup@v4
+          sparse-checkout: |
+            .github/actions/setup-deps/action.yml
+            test/setup
+            test/test-utils/reporters/custom-default-reporter.ts
```
Bertie690, line 68: IMO we should just checkout the entire folder in case new ones get added later
plus its simpler
```suggestion
            test/test-utils/reporters
```

DayKev (approved):

vite.config.ts
SirzBenjie: > We should have top level await support, though it's not working when I try for some reason.

Yeah I'm not sure. Maybe it's because it's not being compiled? I know this works, so 🤷

.github/workflows/tests.yml
SirzBenjie: If a new one gets added then we can modify this here. Not every reporter will necessarily need to be used for tests.

.github/workflows/deploy-beta.yml
SirzBenjie: I think it's fine. It specifies exactly where that comes from without having to remember. Explicit > implicit, no?

.github/workflows/tests.yml
Bertie690: ... they're in the test folder though

DayKev: We're unlikely to add any more custom reporters, so it's not a big deal.

PR #6852, opened 2025-12-16
[Bug] Near ally target moves now fail in single battles

baekanthony:

Users should see a fail message when using moves that target `NEAR_ALLY` in single battles (PP should be used up too).

Fixes https://github.com/pagefaultgames/pokerogue/issues/6641 and also according to Bulbapedia all of these moves should actually fail in single battles.

Before changes, moves that target `NEAR_ALLY` in single battles never made it past the first failure check because it was failing from having a target count of 0. This led to the move being cancelled and nothing showing up on the UI & PP remaining same.

## How to test the changes?

Use the session data in [discord ](https://discord.com/channels/1125469663833370665/1425582246353571999/1425582246353571999)

Can also manually test with: 
```js
const overrides = {
  MOVESET_OVERRIDE: [MoveId.DRAGON_CHEER, MoveId.HELPING_HAND, MoveId.COACHING],
  ENEMY_MOVESET_OVERRIDE: [MoveId.HELPING_HAND],
  BATTLE_STYLE_OVERRIDE: "single"

```

## Checklist

### DIFF
--- a/src/phases/move-phase.ts
+++ b/src/phases/move-phase.ts
@@ -20,6 +20,7 @@ import { MoveFlags } from "#enums/move-flags";
 import { MoveId } from "#enums/move-id";
 import { MovePhaseTimingModifier } from "#enums/move-phase-timing-modifier";
 import { MoveResult } from "#enums/move-result";
+import { MoveTarget } from "#enums/move-target";
 import { isIgnorePP, isIgnoreStatus, isReflected, isVirtual, MoveUseMode } from "#enums/move-use-mode";
 import { PokemonType } from "#enums/pokemon-type";
 import { StatusEffect } from "#enums/status-effect";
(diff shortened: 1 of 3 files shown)


src/phases/move-phase.ts
```diff
@@ -663,7 +666,10 @@ export class MovePhase extends PokemonPhase {
     const usability = new BooleanHolder(false);
     if (moveName.endsWith(" (N)")) {
       failedText = i18next.t("battle:moveNotImplemented", { moveName: moveName.replace(" (N)", "") });
-    } else if (moveId === MoveId.NONE || this.targets.length === 0) {
+    } else if (
+      moveId === MoveId.NONE
+      || (this.targets.length === 0 && move.getMove().moveTarget !== MoveTarget.NEAR_ALLY)
```
Bertie690, line 671: @PistaCream In reality, this `targets.length` check shouldn't even be here.

This is duplicated from `resolveFinalPreMoveCancellationChecks`, and breaks hazard moves' ability to work on an empty field.
I still need to check back with @SirzBenjie about which check is supposed to go where, but the simpler and more correct approach would likely be removing this entirely.

baekanthony: Ah I see, then yea that would be the cleanest solution

PR #6858, opened 2025-12-18
refactor(move): clean up Move-calling Moves; add Nature Power locales

Bertie690:
Re-creating to hopefully stave off the GitHub merge conflict bogeyman

> [!IMPORTANT] 
> Most of the changes are in `move.ts`, everything else is just comments/tests.
> Also, the bulk of said changes simply boil down to "move stuff to base class, add subclasses"

Nature Power has locales now!
Also move calling moves will no longer randomly "fail" seemingly at random. (This was linked to calling a move that had no valid targets; such occurrences will call the move correctly and then have THAT fail instead.)
Metronome code was made using questionable descisions - a public function used solely for overrides, a `do while` loop etc etc.

Then I discovered the other moves and got even more frustrated.

Refactored `CallMoveAttr` to become an abstract move calling class, handling functionality related to actually using the called move.
It exposes an abstract function (`getMove`) which is used by subclasses to determine the move being called.

Only Nature Power extends off of the base class itself, the rest branching off of `CallMoveAttrWithBanlist` (which implements additional functionality for restricting selectable moves based on a banlist and/or challenges).

Added the aforementioned nature power messages and removed an antiquated fallback to "fail" the move if no targets were present.

Moved all the classes next to one another

Added and fixed up existing tests into a mega-conglomerate file

> [!NOTE]
> This does not make move calling moves properly apply within the `MovePhase`, but DOES pave the way by removing jank and making Nature Power an actual subclass instead of doing its own thing.

## Screenshots/Videos
Nature Power locales
https://github.com/user-attachments/assets/47ecc859-dc59-4aee-b0f3-690537f55695

## How to test the changes?
Automated tests, including the new `nature-power.test.ts` and `move-calling-moves.ts` files

## Checklist


src/data/moves/move.ts
```diff
@@ -9456,8 +9479,11 @@ export function initMoves() {
       .unimplemented(),
     new SelfStatusMove(MoveId.METRONOME, PokemonType.NORMAL, -1, 10, -1, 0, 1) //
       .attr(RandomMoveAttr, invalidMetronomeMoves),
-    new StatusMove(MoveId.MIRROR_MOVE, PokemonType.FLYING, -1, 20, -1, 0, 1) //
-      .attr(CopyMoveAttr, true, invalidMirrorMoveMoves),
+    new StatusMove(MoveId.MIRROR_MOVE, PokemonType.FLYING, -1, 20, -1, 0, 1).attr(
+      CopyMoveAttr,
+      invalidMirrorMoveMoves,
+      true,
+    ),
```
Madmadness65:
```suggestion
    new StatusMove(MoveId.MIRROR_MOVE, PokemonType.FLYING, -1, 20, -1, 0, 1) //
      .attr(CopyMoveAttr, invalidMirrorMoveMoves, true),
```
Removing the blank comment on the end makes the formatting on the attributes ugly, so just correcting these

src/data/moves/move.ts
```diff
@@ -10413,8 +10439,7 @@ export function initMoves() {
       .ignoresSubstitute()
       .target(MoveTarget.NEAR_ENEMY)
       .unimplemented(),
-    new SelfStatusMove(MoveId.COPYCAT, PokemonType.NORMAL, -1, 20, -1, 0, 4) //
-      .attr(CopyMoveAttr, false, invalidCopycatMoves),
+    new SelfStatusMove(MoveId.COPYCAT, PokemonType.NORMAL, -1, 20, -1, 0, 4).attr(CopyMoveAttr, invalidCopycatMoves),
```
Madmadness65:
```suggestion
    new SelfStatusMove(MoveId.COPYCAT, PokemonType.NORMAL, -1, 20, -1, 0, 4) //
      .attr(CopyMoveAttr, invalidCopycatMoves),
```
Same for this one

Madmadness65 (commented):
Just a few biome-ruining-things nitpicks

src/data/moves/move.ts
```diff
@@ -7534,281 +7568,270 @@ class CallMoveAttr extends OverrideMoveEffectAttr {
...
+      case BiomeId.LABORATORY:
+        return MoveId.ZAP_CANNON;
+      case BiomeId.POWER_PLANT:
+        return MoveId.CHARGE_BEAM;
+      case BiomeId.END:
+        return MoveId.ETERNABEAM;
+      default:
+        // Fallback for no match
+        biome satisfies never;
+        console.warn(
```
SirzBenjie: We don't need the `console.warn` if we also have `biome satisfies never`, while _also_ having a fallback. TS should catch this if there's something invalid.

src/data/moves/move.ts
```diff
@@ -7534,281 +7568,270 @@ class CallMoveAttr extends OverrideMoveEffectAttr {
...
  */
-export class RandomMovesetMoveAttr extends CallMoveAttr {
-  private readonly includeParty: boolean;
-  private moveId: number;
-  constructor(invalidMoves: ReadonlySet<MoveId>, includeParty = false) {
-    super();
-    this.includeParty = includeParty;
+abstract class CallMoveAttrWithBanlist extends CallMoveAttr {
+  /**
+   * A {@linkcode ReadonlySet} containing all {@linkcode MoveId | moves} that this attribute cannot copy,
```
SirzBenjie:
```suggestion
   * Contains all {@linkcode MoveId | moves} that this attribute cannot copy,
```

fabske0:
What's the reason behind merging the tests into a single file?

Bertie690:
> What's the reason behind merging the tests into a single file?

Allows condensing common checks together more easily, since move-calling moves share a decent chunk of semantics.
Also fewer loose files

src/data/moves/move.ts
```diff
@@ -7557,273 +7591,262 @@ class CallMoveAttr extends OverrideMoveEffectAttr {
...
-    const moveStatus = new BooleanHolder(true);
-    do {
-      moveId = this.getMoveOverride() ?? moveIds[user.randBattleSeedInt(moveIds.length)];
-      moveStatus.value = moveId !== MoveId.NONE;
-      if (user.isPlayer()) {
-        applyChallenges(ChallengeType.POKEMON_MOVE, moveId, moveStatus);
-      }
-    } while (!moveStatus.value);
-    return super.apply(user, target, allMoves[moveId], args);
+  private getMoveIdForTerrain(terrain: TerrainType, biome: BiomeId): MoveId {
```
fabske0: Can we rename this method since it's not just for `terrain` but also the biome

src/data/moves/move.ts
```diff
@@ -7557,273 +7591,262 @@ class CallMoveAttr extends OverrideMoveEffectAttr {
...
     this.invalidMoves = invalidMoves;
   }
 
   /**
-   * User calls a random moveId selected in {@linkcode getCondition}
-   * @param user Pokemon that used the move and will call a random move
-   * @param target Pokemon that will be targeted by the random move (if single target)
-   * @param move Move being used
-   * @param args Unused
+   * Check whether a given move is selectable by this attribute, based on its currnet banlist.
```
fabske0:
```suggestion
   * Check whether a given move is selectable by this attribute, based on its current banlist.
```

src/phases/move-end-phase.ts
```diff
@@ -7,6 +7,10 @@ import { PokemonPhase } from "#phases/pokemon-phase";
 
 export class MoveEndPhase extends PokemonPhase {
   public readonly phaseName = "MoveEndPhase";
+  /**
+   * Whether the current move was a follow-up attack or not.
+   * Used to prevent ticking down Encore and similar effects when copying moves.
+   */
   private wasFollowUp: boolean;
```
fabske0:
```suggestion
  private readonly wasFollowUp: boolean;
```

also just below this but Gh won't let me select it

test/moves/ability-ignore-moves.test.ts
```diff
@@ -33,58 +32,49 @@ describe("Moves - Ability-Ignoring Moves", () => {
...
-    await game.toEndOfTurn();
-
-    expect(enemy.isFainted()).toBe(false);
-    expect(game.field.getPlayerPokemon().getLastXMoves()[0].move).toBe(MoveId.PHOTON_GEYSER);
-  });
-
-  it("should not ignore enemy abilities when called by Mirror Move", async () => {
-    game.override.moveset(MoveId.MIRROR_MOVE).enemyMoveset(MoveId.SUNSTEEL_STRIKE);
-
+  it("should not ignore enemy abilities when called by move-calling moves", async () => {
```
fabske0:
```suggestion
  it("should not ignore ignorable enemy abilities when called by move-calling moves", async () => {
```
consistency with the change above

test/tests/moves/ability-ignore-moves.test.ts
```diff
@@ -33,58 +32,49 @@ describe("Moves - Ability-Ignoring Moves", () => {
...
+    await game.phaseInterceptor.to("MoveEndPhase", false);
 
-    expect(enemy.isFainted()).toBe(false);
-    expect(game.field.getPlayerPokemon().getLastXMoves()[0].move).toBe(MoveId.SUNSTEEL_STRIKE);
+    expect(game.field.getPlayerPokemon()).toHaveUsedMove(MoveId.PHOTON_GEYSER);
+    expect(game.scene.arena.ignoreAbilities).toBe(false);
   });
 
   // TODO: Verify this behavior on cart
   it("should ignore enemy abilities when called by Instruct", async () => {
```
fabske0, line 84:
```suggestion
  it("should ignore ignorable enemy abilities when called by Instruct", async () => {
```
same here

test/moves/ability-ignore-moves.test.ts
Bertie690: tbh i don't like the semantics of "ignore ignorable"

fabske0: I'm also fine with that. These 2 suggestions were just for consitency since you added `ignoreable` to the test case above

src/data/moves/move.ts
```diff
@@ -7583,39 +7583,77 @@ export class FirstMoveTypeAttr extends MoveEffectAttr {
...
-      return false;
-    }
+  constructor(
+    /**
+     * Whether to target the called move at the move's selected target if possible; default `false`.
+     * If `false`, will unleash non-spread moves against a random eligible target.
+     *
+     * Should never be `true` if the move is set to target the user!
+     */
+    targetEnemy = false,
```
Bertie690: I realize this should arguably be renamed to `preserveTarget` instead.

SirzBenjie (dismissed):

src/data/moves/move.ts
Bertie690: i like it in full sentence format imo.

Madmadness65 (dismissed):

test/tests/moves/move-calling-moves.test.ts
```diff
@@ -0,0 +1,521 @@
...
+        result: MoveResult.SUCCESS,
+        useMode: MoveUseMode.FOLLOW_UP,
+      });
+      expect(enemy).toHaveStatStage(Stat.ATK, 2);
+    });
+  });
+
+  describe.todo("Metronome", () => {
+    // TODO: Figure out a good way to override RNG rolls to force Metronome to use a move
+    // WITHOUT using the override that mocks the method
```
DayKev, line 503: Can't you just mock the return value of `pokemon.randBattleSeedInt`?

DayKev (approved):

test/tests/moves/move-calling-moves.test.ts
Bertie690: I have found that finicky at best.

DayKev: Mocking the RNG functions is literally the only method that doesn't involve mocking Metronome itself. If that's not viable then it's not possible to write this test.

Bertie690: Either way, it isn't a dealbreaker for this PR.

PR #6863, opened 2025-12-19
[Refactor] Add utility method `PokemonMove#isOutOfPP`

Bertie690:
None

The existing way of checking `PokemonMove` PP depletion is frankly quite silly - either manually comparing PP counrs with `getMovePP` or doing `if (!pm.getPpRatio())` (which uses costly division).

Also, I like helper methods.
Added a new helper method, `PokemonMove.isOutOfPP``, that returns whether the current move's PP is depleted.

updated 1-2 docs in the file
## Screenshots/Videos
N/A
## How to test the changes?
N/A

## Checklist


src/data/moves/pokemon-move.ts
```diff
@@ -93,13 +93,14 @@ export class PokemonMove {
   }
 
   /**
-   * Sets {@link ppUsed} for this move and ensures the value does not exceed {@link getMovePp}
-   * @param count Amount of PP to use
+   * Consume up to `count` PP from this move (up to its maximum PP).
+   * @param count - The amount of PP to use; default `1`
```
DayKev:
```suggestion
   * @param count - (Default `1`) The amount of PP to use
```

DayKev (approved):

Issue #6153, opened 2025-07-26
[Bug] Roost ignores secondary types, fails to restore original typing when Pure-flying type gains 3 types


Bertie690:
### Describe the bug

Currently, Roost accomplishes its "pure flying --> normal" check by checking if the user was originally pure flying and does not have currently have 2 types (presumably indicating an added one from Trick-or-Treat or Forest's Curse).

<img width="1710" height="634" alt="Image" src="https://github.com/user-attachments/assets/11c9ba91-cf23-4a7e-9538-6022715a0dd6" />

This, however, overlooks the case of having 3 or more types.
A Pokemon that is:
- Pure flying originally
- Transforms into a Flying/X opponent
- Gains a type from Forest's Curse/etc (making it Flying/X/Grass)
- Uses Roost

Will become Normal/Grass type for the rest of the turn despite not being pure flying, then revert to Grass/Flying until switched out.

### Reproduction

```ts
const overrides = {
  OPP_MOVESET_OVERRIDE: [MoveId.ROOST, MoveId.FORESTS_CURSE],
  OPP_SPECIES_OVERRIDE: SpeciesId.CORVIKNIGHT,
  ABILITY_OVERRIDE: AbilityId.IMPOSTER,
  BATTLE_STYLE_OVERRIDE: "double"
} satisfies Partial<InstanceType<OverridesType>>;
```

### Expected behavior

<img width="1316" height="173" alt="Image" src="https://github.com/user-attachments/assets/4e61a825-0bf5-4334-9d6f-660b8dae8b48" />

https://bulbapedia.bulbagarden.net/wiki/Roost_(move)

> If a Pokémon with another type besides Flying uses Roost, it will lose its Flying type until the end of the turn (but will not have it replaced with the Normal type).

I see no reason this should differ for Transformed and un-transformed pokemon.

### Screenshots / Videos

AFTER ROOST - TORNADUS

<img width="157" height="30" alt="Image" src="https://github.com/user-attachments/assets/8afb6af3-b514-4246-8c5b-92c0e3c12139" />

AFTER TURN END - TORNADUS
<img width="228" height="32" alt="Image" src="https://github.com/user-attachments/assets/2d50718e-f29b-4e32-b486-c9f3eea6e091" />

<img width="388" height="67" alt="Image" src="https://github.com/user-attachments/assets/1f1d350f-73fe-42aa-9ac4-ec6917f5bf8a" />

Full mp4 video
[...]

PR #6871, opened 2025-12-20
refactor: clean up Roost code

Bertie690:
N/A

Chipping at the cursed Roost type change impl before I go insane.
Denested the Roost attribute to just _not add the tag if it wouldn't do anything_, alongside some docs to aid whatever poor soul happens to try and fix this.

## Screenshots/Videos
N/A
## How to test the changes?
3 more roost tests, including 1 TODO case from the roost bug tracker

## Checklist


src/field/pokemon.ts
```diff
@@ -1919,109 +1919,104 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    // Stellar tera does not apply when used defensively
+    const shouldUseTeraType = !(forDefend && teraType === PokemonType.STELLAR);
+    if (includeTeraType && this.isTerastallized && shouldUseTeraType) {
+      // Offensively, add the tera to the original types for STAB;
+      // defensively, become purely the tera type (if not stellar).
+      types.add(teraType);
+      if (forDefend) {
+        return Array.from(types);
       }
+    } else {
```
Bertie690: Here's my reasoning for how I was able to pull this chunk into an `else` block.
1. This code only executes if the `types` array is empty or `ignoreTeraType` is false.
2. The only way the `types` array could have changed at this point is for the Terastallization section before this to have ran (in which case `ignoreTeraType` must have obviously been true).
3. Hence, this runs if and only if the tera type section doesn't.
4. Hence, we can move it to an `else` for clarity.

I'd normally say correct me if i am wrong, but i am not

src/data/pokemon/pokemon-data.ts
```diff
@@ -100,6 +100,7 @@ interface SerializedPokemonSummonData {
  *
  * @sealed
  */
+// TODO: Change these `null`s into `undefined`s to save on storage space
```
emdeann: Are you referring to the various `| null` types here? If I'm interpreting it right I would disagree; the semantic value of null vs. undefined is worth preserving over like 10 bytes of storage saved. Much bigger fish to fry in that department

src/data/battler-tags.ts
```diff
@@ -2485,6 +2485,8 @@ export class CursedTag extends SerializableBattlerTag {
 /**
  * Battler tag for attacks that remove a type post use.
  */
+// TODO: Remove this "tag" - it is solely used to communicate type changes from Burn Up/Double Shock to Roost,
+// and does so extremely poorly to boot
```
emdeann: It would be useful in a todo if you explained why it's bad, not just that it is bad, so someone else can fix it if they want to

src/field/pokemon.ts
```diff
@@ -1919,109 +1919,106 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+      // Offensively, add the tera to the original types for STAB;
+      // defensively, become purely the tera type and nothing else
+      types.add(teraType);
+      if (forDefend) {
+        return Array.from(types);
+      }
+      // TODO: This being in an `else` block is odd and contradicts the "add attack types to original types for STAB" statement
+    } else {
+      // biome-ignore lint/suspicious/useIterableCallbackReturn: benign
+      this.getOriginalTyping(ignoreOverride, useIllusion).forEach(s => types.add(s));
```
emdeann:
```suggestion
      this.getOriginalTyping(ignoreOverride, useIllusion).forEach(s => { types.add(s) });
```

DayKev: That'll cause a lint error due to formatting, would need to be this:
```suggestion
      this.getOriginalTyping(ignoreOverride, useIllusion).forEach(type => {
        types.add(type);
      });
```
or this:
```suggestion
      for (const type of this.getOriginalTyping(ignoreOverride, useIllusion)) {
        types.add(type);
      }
```

src/data/pokemon/pokemon-data.ts
DayKev: Well it does add up, since it's for every single Pokémon in every single session save of every single player.
Though they can be converted from `null` to `undefined` before being serialized, and then back when de-serialized.

It's also inconsistent though, so it should probably be unified anyway, and might as well go with `undefined` in that case.

test/moves/roost.test.ts
```diff
@@ -242,4 +242,52 @@ describe("Moves - Roost", () => {
     expect(playerPokemonTypes.length === 3).toBeTruthy();
     expect(playerPokemon.isGrounded()).toBeFalsy();
   });
+
+  // TODO: This interaction is extremely broken due to a lack of granularity with type querying effects
+  it.todo("should not change Flying type Pokemon changed into having 3 types into Normal-types", async () => {
```
DayKev:
```suggestion
  it.todo("should respect prior type change effects when determining the user's new typing", async () => {
```

src/field/pokemon.ts
Bertie690: Remind me to handroll a variadic `set.add` eventually

Bertie690: this is benign - do we really want to enforce this

src/data/pokemon/pokemon-data.ts
Bertie690: > the semantic value of null vs. undefined is worth preserving over like 10 bytes of storage saved.

Frankly, we treat both the same in 99.95% of cases, so I don't really see the argument.

src/data/battler-tags.ts
Bertie690:
```suggestion
// TODO: Remove this "tag" - it is solely used to communicate type changes from Burn Up/Double Shock to Roost.
// This is a hacky way to add the bare minimum amount of granularity
// when a more comprehensive system would likely be preferred.
```

test/moves/roost.test.ts
Bertie690: that's grammatically incorrect now

src/field/pokemon.ts
emdeann: Does it error? I've done that plenty of times and never had an issue with it, unless it's a new rule. But yes, I'd rather avoid needing to write an ignore comment if it's easily avoidable

test/moves/roost.test.ts
DayKev: Oh I see, I misread it. The way it was written before is kinda confusing.

src/field/pokemon.ts
DayKev: Tbh we should probably disable this lint rule anyway since it has too many false positives.

test/moves/roost.test.ts
Bertie690: yeah - i edited & fixed it 
all good

src/field/pokemon.ts
emdeann: Agree with that too

Wlowscha:
Isn't this mostly #6697?

Wlowscha (approved):
With the rest of the changes reverted, this is now almost just a list of TODOs. However, it does add a (currently inactive) test which makes the PR still valuable. I would recommend opening an issue collecting the various TODOs so it's easier to address the problems later on :)

src/field/pokemon.ts
```diff
@@ -2341,7 +2341,8 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
   }
 
   /**
-   * @returns the pokemon's current tera {@linkcode PokemonType}
+   * Return this {@linkcode Pokemon}'s current Tera Type, accounting for species-based restrictions.
+   * @returns This Pokemon's current Tera {@linkcode PokemonType}
```
```suggestion
   * @returns This Pokemon's current Tera {@linkcode PokemonType | type}, accounting for species-based restrictions
```

Bertie690:
The issues with roost's typing were documented in #6153, long before this PR was made.

SirzBenjie (approved):

PR #6874, opened 2025-12-20
[Bug] Avoid pre-emptively leaving the field when forcibly switching out 

Bertie690:
U-Turn and company no longer crash if they proc a reactive effect (Rough Skin, Aftermath, etc.) that faints the user.
Also they no longer cancel your shell bell healing which is nice.

Fixes a pair of P1 crashes, plus a more minor P2 bug

Avoid prematurely leaving the field when these abilites resolve (read: while we're still processing shit), instead being moved to a fallback inside `switch-summon-phase`.

Furthermore, the implementation of `PhaseManager#queueDeferred` has changed to deferring the phase normally (similar to `queueFaintPhase`) to allow for switch phases to still occur after any `FaintPhase`s their use would otherwise create.

## Screenshots/Videos
N/A
## How to test the changes?
`pnpm test u-turn.test.ts` (1 new)

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "pokemon-rogue-battle",
   "private": true,
-  "version": "1.11.3",
+  "version": "1.11.4",
   "type": "module",
   "scripts": {
     "start:prod": "vite --mode production",
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -4,6 +4,7 @@ import { ArenaTagType } from "#enums/arena-tag-type";
 import { BattleSpec } from "#enums/battle-spec";
 import { BattleType } from "#enums/battle-type";
 import { BattlerIndex } from "#enums/battler-index";
+import { BiomeId } from "#enums/biome-id";
 import type { Command } from "#enums/command";
 import type { MoveId } from "#enums/move-id";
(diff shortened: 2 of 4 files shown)


test/moves/u-turn.test.ts
```diff
@@ -103,4 +104,42 @@ describe("Moves - U-turn", () => {
...
+    game.doSelectPartyPokemon(1);
+    await game.move.forceEnemyMove(MoveId.DESTINY_BOND);
+    await game.setTurnOrder([BattlerIndex.ENEMY, BattlerIndex.PLAYER]);
+    await game.toEndOfTurn();
+
+    expect(karp).toHaveFainted();
+    expect(feebas).toHaveFainted();
+    expect(feebas.isOnField()).toBe(false);
+
+    // Make sure feebas' faint phase occurs before being switched out (or not)
```
DayKev:
```suggestion
    // TODO: Check whether feebas' faint phase occurs before being switched out (or not)
```
Seems like this is a todo comment?

Bertie690: Nope - i just hit a few more keys than needed
```suggestion
    // Make sure feebas' faint phase runs before being switched out (since that was the root cause of the crash)
```

PR #6879, opened 2025-12-20
[Balance] increase ai friendship scaling

fabske0:

Trainers will reach max friendship at wave 145 instead of 200
And will have a min of 50 friendship

requested in balance chat

divide by 145 instead of 200

## Checklist

### DIFF
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -4,6 +4,7 @@ import { ArenaTagType } from "#enums/arena-tag-type";
 import { BattleSpec } from "#enums/battle-spec";
 import { BattleType } from "#enums/battle-type";
 import { BattlerIndex } from "#enums/battler-index";
+import { BiomeId } from "#enums/biome-id";
 import type { Command } from "#enums/command";
 import type { MoveId } from "#enums/move-id";
 import { MysteryEncounterMode } from "#enums/mystery-encounter-mode";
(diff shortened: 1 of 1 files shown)


fabske0:
> We should make this a constant lol

Will add this tomorrow

src/field/pokemon.ts
```diff
@@ -173,7 +173,7 @@ import { calculateBossSegmentDamage } from "#utils/damage";
 import { getEnumValues } from "#utils/enums";
 import { getFusedSpeciesName, getPokemonSpecies, getPokemonSpeciesForm } from "#utils/pokemon-utils";
 import { inSpeedOrder } from "#utils/speed-order-generator";
-import { argbFromRgba, QuantizerCelebi, rgbaFromArgb } from "@material/material-color-utilities";
+import { argbFromRgba, clampInt, QuantizerCelebi, rgbaFromArgb } from "@material/material-color-utilities";
```
DayKev, line 176: Huh, there's a clamp function from that package. I guess that works too lol.

fabske0: ohh. didn't even check where it was importing from. Was also a it confused as i never saw this one used before

DayKev: Can change it to `Phaser.Math.Clamp` or something later, the functionality should be identical so it's not a big deal.

PR #6887, opened 2025-12-22
[UI/UX] Retry login after registration once, improve error messages

DayKev:
If the automatic login after registration fails, it will be retried once after 2 seconds. If that also fails, then the user will be sent to the login form.
The error messages for unaccounted for login/registration failure are updated to say "Unknown [login/registration] error!".

The registration flow sometimes errors out when trying to log in after registration for unclear reasons, and the user isn't told what failed. This should hopefully reduce confusion.

See first section.

## Checklist
- ~[ ] **I'm using `beta` as my base branch**~

### DIFF
--- a/src/ui/handlers/registration-form-ui-handler.ts
+++ b/src/ui/handlers/registration-form-ui-handler.ts
@@ -6,6 +6,7 @@ import type { LoginPhase } from "#phases/login-phase";
 import type { InputFieldConfig } from "#ui/form-modal-ui-handler";
 import type { ModalConfig } from "#ui/modal-ui-handler";
 import { addTextObject } from "#ui/text";
+import { fixedInt } from "#utils/common";
 import i18next from "i18next";
 import { LoginRegisterInfoContainerUiHandler } from "./login-register-info-container-ui-handler";
 
@@ -124,7 +125,7 @@ export class RegistrationFormUiHandler extends LoginRegisterInfoContainerUiHandl
                       }
                     });
                   };
-                  globalScene.time.delayedCall(2000, retryLogin);
+                  globalScene.time.delayedCall(fixedInt(2000), retryLogin);
                 } else {
                   originalRegistrationAction?.();
                 }


src/ui/handlers/registration-form-ui-handler.ts
```diff
@@ -109,18 +109,25 @@ export class RegistrationFormUiHandler extends LoginRegisterInfoContainerUiHandl
...
-                });
+              const username = usernameInput.text;
+              const password = passwordInput.text;
+              pokerogueApi.account.login({ username, password }).then(loginError => {
+                if (loginError) {
+                  // retry once if the first attempt fails
+                  const retryLogin = () => {
+                    pokerogueApi.account.login({ username, password }).then(error => {
+                      if (error) {
+                        onFail(error);
```
fabske0: Since the registration has been succesful if we reach this point, could we show a more detailed error message here if the retry fails? `Unkown login error!` is definitely better than the current error, but I think it can still be a bit confusing since there is no indication that the registration was succesful

DayKev: What should it say?
Actually, it should probably jump straight to the login form instead but I'm not 100% sure how to get that to happen without breaking things.

fabske0: Maybe something like `Registration successful. Try login in.`
Jumping straight to the login form would be even better. But that can be something for a different PR. This PR is already a big improvement.

DayKev: I prepended "Registration successful! " to the error message to make it extra clear that registration isn't what failed.

fabske0:
Something like this could work. [Go to login after registration #24](https://github.com/DayKev/pokerogue/pull/24)
Could be cleaned up a bit. just wanted to try

src/ui/handlers/registration-form-ui-handler.ts
DayKev: Merged in the "go to login" change.

PR #6889, opened 2025-12-22
[Bug] Tar Shot no longer breaks with Normalize

Bertie690:
Tar Shot now correctly considers the user's type-changing abilities instead of the target's for its fire type condition.
As a result, it will work in attacks against opponents with Normalize and distinctly NOT work on Fire-type moves changed to Normal-type from the user's Normalize ability.

I noticed the broken code while attempting to fix Roost-related shenanigans.

Fixed 1 loc

Added/updated the woefully out of date test suite.

Made override to force enemy terastallization
## Screenshots/Videos
N/A
## How to test the changes?
run tests
## Checklist

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -43,16 +43,16 @@ jobs:
         if: ${{ !cancelled() }}
 
       # Validate types with tsc - https://www.typescriptlang.org/docs/handbook/compiler-options.html#using-the-cli
-      - name: Run Typecheck
+      - name: Run Typecheck (TS files)
         run: pnpm typecheck
         id: typecheck
         if: ${{ !cancelled() }}
 
-      # Run tsc again on the scripts folder.
-      # Required in order to use separate rules for script js files and regular ts files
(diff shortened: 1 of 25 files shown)


test/moves/tar-shot.test.ts
```diff
@@ -27,107 +31,128 @@ describe("Moves - Tar Shot", () => {
...
-    expect(enemy.getStatStage(Stat.SPD)).toBe(-1);
+    game.move.use(MoveId.FIRE_PUNCH);
+    await game.toEndOfTurn();
 
-    await game.toNextTurn();
+    expect(spy).toHaveLastReturnedWith(2 * getTypeDamageMultiplier(PokemonType.FIRE, PokemonType.WATER));
+  });
 
-    game.move.select(MoveId.TAR_SHOT);
+  it("should lower the target's Speed by 1 stage and double the effectiveness of incoming Fire-type moves", async () => {
```
DayKev: This test is a duplicate of the first test.

Bertie690: oops

PR #6891, opened 2025-12-23
[Refactor] Add `playNumberTween` utility function

Bertie690:
none
We have async utils for `tweens.add`, and `addNumberTween` seems a logical next step

Made a util that should yield the values produced by a tween in order as they are produced.

## Screenshots/Videos
N/A

## How to test the changes?
proofread the util?

## Checklist

### DIFF
--- a/src/utils/anim-utils.ts
+++ b/src/utils/anim-utils.ts
@@ -42,7 +42,6 @@ export async function* playNumberTween(
   let done = false;
   let waitFunc: (() => void) | null = null;
 
-  // make the tween & wait for it to start, hooking into its update/completion commands
   await new Promise<Phaser.Tweens.Tween>(resolve => {
     scene.tweens.addCounter({
       ...config,


src/utils/anim-utils.ts
```diff
@@ -21,3 +21,53 @@ export async function playTween(config: PlayTweenConfig, scene: SceneBase = glob
...
+      },
+      onComplete: () => {
+        done = true;
+        waitFunc?.();
+        waitFunc = null;
+      },
+    });
+  });
+
+  // wait and repeatedly yield values
```
DayKev:
```suggestion
```

src/utils/anim-utils.ts
```diff
@@ -21,3 +21,53 @@ export async function playTween(config: PlayTweenConfig, scene: SceneBase = glob
...
+ */
+export async function* playNumberTween(
+  config: NumberTweenConfig,
+  scene: SceneBase = globalScene,
+): AsyncGenerator<number> {
+  const queue: number[] = [];
+  let done = false;
+  let waitFunc: (() => void) | null = null;
+
+  // make the tween & wait for it to start, hooking into its update/completion commands
```
DayKev:
```suggestion
```

Bertie690:
@DayKev Does the thing look like it'll work?
I had trouble making unit tests due to us stubbing out the method in question, which doesn't give much confidence...

DayKev:
Not sure, seems to be more complicated than the existing `playTween` function. Maybe test implementing it in a couple places?

Bertie690:
Drafting while I rework this into a "frame counter utility" (since that's basically all we seem to be using this for anyways)

PR #6898, opened 2025-12-25
[Bug][Refactor] `MoveTypeChangeAbAttr` no longer modifies move power

DayKev:
Pixilate et al will apply their power boost before Terastallization sets the move's minimum base power to 60.

- Move type changing should be handled separately from boosting the move's power.
- Pixilate et al were applying at the wrong time, incorrectly boosting a move's power more than they should.

- Move power boosting removed from `MoveTypeChangeAbAttr`.
- Pixilate et al now use `MovePowerBoostAbAttr` for power boosting.
- Conditions extracted from `MoveTypeChangeAbAttr` and moved to reusable functions so they can be passed to `MovePowerBoostAbAttr`.

## Checklist


src/data/abilities/ability.ts
```diff
@@ -1761,53 +1761,63 @@ export interface MoveTypeChangeAbAttrParams extends AugmentMoveInteractionAbAttr
...
+    if (
+      move.id === MoveId.TERA_STARSTORM
+      && user.getTeraType() === PokemonType.STELLAR
+      && user.hasSpecies(SpeciesId.TERAPAGOS)
+    ) {
+      return false;
+    }
+  }
+
+  return true;
```
Bertie690: Denesting would be nice

PR #6909, opened 2025-12-27
[Test] Update `toHaveUsedPP` to accept negative PP counts

Bertie690:
N/A
Seems like a useful extension of a matcher, plus I already had the changes lying around on a feature branch somewhere

Added wraparound behavior for negative PP counts.

## Checklist

### DIFF
--- a/test/test-utils/matchers/to-have-used-pp.ts
+++ b/test/test-utils/matchers/to-have-used-pp.ts
@@ -65,8 +65,8 @@ export function toHaveUsedPP(
       pass,
       message: () =>
         pass
-          ? `Expected ${pkmName}'s ${moveStr} to NOT have used PP, but it did!`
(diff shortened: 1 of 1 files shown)


DayKev (approved):

fabske0 (approved):

SirzBenjie (commented):
I'm confused by what this will be useful for. So `-1` basically means the move has 1 pp remaining? If this is the case, then I think it would be better if instead there was a second matcher named `toHaveRemainingPP`. It would be better than combining the functionality in this method.

Bertie690:
Fair.
I'll close the PR until the need arises I suppose.

SirzBenjie:
@Bertie690 
FWIW, I don't think it necessarily _needs_ to wait until a need arises. Even if there isn't anything that would use it now. Either way, when the need _does_ arise, my comment would still stand.

PR #6910, opened 2025-12-27
[Refactor] Move instruct's banlist into `invalid-moves.ts`

Bertie690:
(minor)
Encore correctly doesn't work on Dynamax Cannon.
During the PR that implemented instruct, someone told me "this should be inside `invalid-moves.ts`".
I agreed with them, but was too lazy to act on it.

Moved the banlist to `invalid-moves.ts`, as well as removing a few redundant doc comments explaining the obvious.

## Screenshots/Videos
N/A
## How to test the changes?
use the move
## Checklist

### DIFF
--- a/src/data/moves/invalid-moves.ts
+++ b/src/data/moves/invalid-moves.ts
@@ -2,7 +2,6 @@ import { MoveId } from "#enums/move-id";
 
 /**
  * Array containing all move-calling moves, used for DRY when writing move banlists
- *
  */
 const moveCallingMoves = [
   MoveId.ASSIST,


src/data/moves/move.ts
```diff
@@ -7876,83 +7876,14 @@ export class RepeatMoveAttr extends MoveEffectAttr {
...
 
-      if (
-        !lastMove?.move // no move to instruct
-        || !movesetMove // called move not in target's moveset (forgetting the move, etc.)
-        || movesetMove.ppUsed === movesetMove.getMovePp() // move out of pp // TODO: This next line is likely redundant as all charging moves are in the above list
-        || allMoves[lastMove.move].isChargingMove() // called move is a charging/recharging move
-        || uninstructableMoves.includes(lastMove.move)
-      ) {
-        // called move is in the banlist
+      if (!lastMove?.move || !movesetMove || movesetMove.isOutOfPp() || invalidInstructMoves.has(lastMove.move)) {
```
Bertie690, line 7883: rip multi-line conditional

src/data/moves/move.ts
```diff
@@ -7876,83 +7876,14 @@ export class RepeatMoveAttr extends MoveEffectAttr {
...
-        MoveId.MIMIC,
-        MoveId.STRUGGLE,
-        // TODO: Add Max/G-Max/Z-Move blockage if or when they are implemented
-      ];
 
-      if (
-        !lastMove?.move // no move to instruct
-        || !movesetMove // called move not in target's moveset (forgetting the move, etc.)
-        || movesetMove.ppUsed === movesetMove.getMovePp() // move out of pp // TODO: This next line is likely redundant as all charging moves are in the above list
-        || allMoves[lastMove.move].isChargingMove() // called move is a charging/recharging move
```
DayKev, line 7950: Tbh it's probably better to check `.isChargingMove()` instead of needing to continually add moves to an array as each new generation releases.

Bertie690: @DayKev we're likely gonna need to do it anyways

DayKev: Unless Gamefreak adds a 2-turn move that is somehow completely incompatible with the existing 2-turn moves in a way that makes it impossible to use `ChargeMove` for them, that shouldn't happen.

Bertie690: I'd still argue the need to check an extra redundant calculation is reason enough for its removal.
@DayKev Also I'm less worried about gen 10 when we can barely implement some interactions in gen 9 correctly

DayKev: A function call that just returns a boolean isn't that expensive.
```ts
  isChargingMove(): this is ChargingMove {
    return false;
  }

    override isChargingMove(): this is ChargingMove {
      return true;
    }
```

Bertie690: I really cannot be bothered to care - my only request is we don't keep redundant checks in the codebase. 
Whether said checks go inside the banlist or a call to `isChargingMove` is really a matter of preference (i prefer the former since it's more explicit), but if you want the latter i give you permission to change it.

DayKev (dismissed):

SirzBenjie (approved):

src/data/moves/invalid-moves.ts
```diff
@@ -1,5 +1,20 @@
 import { MoveId } from "#enums/move-id";
 
+/**
+ * Array containing all move-calling moves, used for DRY when writing move banlists
+ *
+ */
```
fabske0, line 5:
```suggestion
/**
 * Array containing all move-calling moves, used for DRY when writing move banlists
 */
```
Otherwise good

fabske0 (approved):

PR #6914, opened 2025-12-28
[UI] Fallback if legacy image doesn't exist

fabske0:

If an image is mising in the `legacy` folder, the normal image will be used

Annoying to duplicate images and potential to be forgotten

Instead of saving a key to `legacyCompatibleImages` no matter what, it only saves it if the load was successful

## How to test the changes?

delete something from the `ui/legacy` folder and use the legacy ui theme

## Checklist

### DIFF
--- a/src/data/mystery-encounters/encounters/part-timer-encounter.ts
+++ b/src/data/mystery-encounters/encounters/part-timer-encounter.ts
@@ -1,5 +1,7 @@
 import { CLASSIC_MODE_MYSTERY_ENCOUNTER_WAVES } from "#app/constants";
 import { globalScene } from "#app/global-scene";
+import { MoneyMultiplierModifier } from "#app/modifier/modifier";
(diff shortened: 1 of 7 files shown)


damocleas (dismissed):

Bertie690 (dismissed):

Bertie690 (approved):
@fabske0 Remind me to add `init` to the type of `Phaser.Scene` - I'm 80% sure the only reason it isn't there is that you _can't_ do it with just JSDoc

src/scene-base.ts
```diff
@@ -17,7 +17,16 @@ export class SceneBase extends Phaser.Scene {
     height: 180, // (1080 / 6)
   } as const;
 
-  getCachedUrl(url: string): string {
+  init() {
```
DayKev:
```suggestion
  public init(): void {
```

PR #6920, opened 2025-12-29
[Bug] Fix Rapid Spin/Mortal Spin removing hazards from the wrong side

DayKev:
Rapid Spin and Mortal Spin will remove hazards from the user's side of the field again.

Fixing bugs.
- Likely cause: #6201

`RemoveArenaTagsAttr` now takes a function to get the side of the arena to remove tags from, instead of inferring it based on the target's side of the arena.
A [proper refactor](https://github.com/Despair-Games/poketernity/pull/170) should be done later, this is just to get it working for the hotfix.

Added full test suites for all moves that had code altered.

## How to test the changes?
`pnpm test:silent rapid-spin-mortal-spin`
`pnpm test:silent defog`
`pnpm test:silent feint`
`pnpm test:silent brick-break-psychic-fangs`
`pnpm test:silent raging-bull`

## Checklist

### DIFF
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -11188,7 +11188,8 @@ export function initMoves() {
       .target(MoveTarget.ALL_NEAR_ENEMIES),
     new SelfStatusMove(MoveId.CELEBRATE, PokemonType.NORMAL, -1, 40, -1, 0, 6)
       // NB: This needs a lambda function as the user will not be logged in by the time the moves are initialized
-      .attr(MessageAttr, () => i18next.t("moveTriggers:celebrate", { playerName: loggedInUser?.username })),
+      .attr(MessageAttr, () => i18next.t("moveTriggers:celebrate", { playerName: loggedInUser?.username }))
+      .attr(MoneyAttr),
     new StatusMove(MoveId.HOLD_HANDS, PokemonType.NORMAL, -1, 40, -1, 0, 6)
       .ignoresSubstitute()
       .target(MoveTarget.NEAR_ALLY),


src/data/moves/move.ts
```diff
@@ -115,6 +115,8 @@ import { inSpeedOrder } from "#utils/speed-order-generator";
 import { toCamelCase, toTitleCase } from "#utils/strings";
 import i18next from "i18next";
 
+type GetRemoveArenaTagSideFunc = (user: Pokemon) => ArenaTagSide;
```
Bertie690: To be frank, I'd rather this be an enum - "user" | "enemy" | "both" seems a lot easier than making several abstract lambda functions that all need to be updated.
That said, I recognize this is supposed to be a hotfix, so all I'mma ask is you add a comment saying "reminder to refactor this"

test/moves/brick-break-psychic-fangs.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  it("should not remove screens if the target is immune to the move", async () => {
+    game.override.enemySpecies(SpeciesId.SABLEYE);
+    await game.classicMode.startBattle([SpeciesId.FEEBAS]);
+
+    game.scene.arena.addTag(ArenaTagType.REFLECT, 0, undefined, game.field.getEnemyPokemon().id, ArenaTagSide.ENEMY);
+
+    game.move.use(moveId);
+    await game.toEndOfTurn();
+
+    expect(game).toHaveArenaTag({ tagType: ArenaTagType.REFLECT, side: ArenaTagSide.ENEMY });
```
Bertie690, line 93:
```suggestion
    const feebas = game.field.getPlayerPokemon();
    expect(feebas).toHaveUsedMove({ move, result: MoveResult.MISS });
    expect(game).toHaveArenaTag({ tagType: ArenaTagType.REFLECT, side: ArenaTagSide.ENEMY });
```

test/moves/defog.test.ts
```diff
@@ -24,40 +27,98 @@ describe("Moves - Defog", () => {
...
       .criticalHits(false)
       .enemySpecies(SpeciesId.SHUCKLE)
       .enemyAbility(AbilityId.BALL_FETCH)
-      .enemyMoveset([MoveId.DEFOG, MoveId.GROWL]);
+      .enemyMoveset(MoveId.SPLASH);
   });
 
-  // TODO: Refactor these tests they suck ass
-  it("should not allow Safeguard to be active", async () => {
+  it("should remove opponent Safeguard", async () => {
```
Bertie690: This should almost undeniably be a single `it.each`.

test/moves/defog.test.ts
```diff
@@ -24,40 +27,98 @@ describe("Moves - Defog", () => {
...
-    await game.move.selectEnemyMove(MoveId.GROWL);
+    game.move.use(MoveId.SPLASH);
+    await game.move.forceEnemyMove(MoveId.GROWL);
 
     await game.phaseInterceptor.to("BerryPhase");
 
     expect(game.field.getPlayerPokemon()).toHaveStatStage(Stat.ATK, -1);
   });
+
+  it("should remove terrains", async () => {
```
Bertie690, line 38: _sigh_ no terrain overrides :(

test/moves/defog.test.ts
```diff
@@ -24,40 +27,98 @@ describe("Moves - Defog", () => {
...
+
+    expect(game.field.getEnemyPokemon()).toHaveStatStage(Stat.EVA, -1);
+  });
+
+  it.each<{ tagType: ArenaTagType; tagName: string }>([
+    { tagType: ArenaTagType.SPIKES, tagName: "Spikes" },
+    { tagType: ArenaTagType.STEALTH_ROCK, tagName: "Stealth Rocks" },
+    { tagType: ArenaTagType.TOXIC_SPIKES, tagName: "Toxic Spikes" },
+    { tagType: ArenaTagType.STICKY_WEB, tagName: "Sticky Web" },
+  ])("should remove $tagName from the field", async ({ tagType }) => {
```
Bertie690:
```suggestion
  ])("should remove $tagName from both sides of the field", async ({ tagType }) => {
```

test/moves/defog.test.ts
```diff
@@ -24,40 +27,98 @@ describe("Moves - Defog", () => {
...
+
+    expect(game).not.toHaveArenaTag({ tagType, side: ArenaTagSide.PLAYER });
+    expect(game).not.toHaveArenaTag({ tagType, side: ArenaTagSide.ENEMY });
+  });
+
+  it.each<{ tagType: ArenaTagType; tagName: string }>([
+    { tagType: ArenaTagType.REFLECT, tagName: "Reflect" },
+    { tagType: ArenaTagType.LIGHT_SCREEN, tagName: "Light Screen" },
+    { tagType: ArenaTagType.AURORA_VEIL, tagName: "Aurora Veil" },
+  ])("should remove $tagName only from the target's side of the field", async ({ tagType }) => {
```
Bertie690, line 85: just add the safeguard/etc tests here???

test/moves/rapid-spin-mortal-spin.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    game.move.use(moveId);
+    await game.toEndOfTurn();
+
+    expect(game).not.toHaveArenaTag({ tagType, side: ArenaTagSide.PLAYER });
+    expect(game).toHaveArenaTag({ tagType, side: ArenaTagSide.ENEMY });
+  });
+
+  it.each<{ tagType: BattlerTagType; tagName: string }>([
+    { tagType: BattlerTagType.SEEDED, tagName: "Leech Seed" },
+    { tagType: BattlerTagType.FIRE_SPIN, tagName: "Fire Spin" },
```
Bertie690:
```suggestion
    { tagType: BattlerTagType.FIRE_SPIN, tagName: "Partial Trapping effects" },
```

test/moves/rapid-spin-mortal-spin.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    player.addTag(tagType, 0, undefined, game.field.getEnemyPokemon().id);
+
+    game.move.use(moveId);
+    await game.toEndOfTurn();
+
+    expect(player).not.toHaveBattlerTag(tagType);
+  });
+
+  if (moveId === MoveId.RAPID_SPIN) {
+    it("should increase the user's speed stage by 1", async () => {
```
Bertie690, line 82: These 2 should be `it.runIf`s.

test/moves/rapid-spin-mortal-spin.test.ts
```diff
@@ -0,0 +1,97 @@
...
+      game.move.use(MoveId.RAPID_SPIN);
+      await game.toEndOfTurn();
+
+      expect(game.field.getPlayerPokemon()).toHaveStatStage(Stat.SPD, 1);
+    });
+  }
+
+  if (moveId === MoveId.MORTAL_SPIN) {
```
Bertie690, line 90: Also here

Bertie690 (approved):
Minor assorted nits.

test/moves/rapid-spin-mortal-spin.test.ts
DayKev: Ah, makes sense Vitest would have something for this.

test/moves/rapid-spin-mortal-spin.test.ts
DayKev: Went with "Binding effects"
> Rapid Spin inflicts damage and then removes the effects of any binding effects
- https://bulbapedia.bulbagarden.net/wiki/Rapid_Spin_(move)

(https://bulbapedia.bulbagarden.net/wiki/Category:Binding_moves)

src/data/moves/move.ts
```diff
@@ -12223,7 +12222,7 @@ export function initMoves() {
       .danceMove(),
     new AttackMove(MoveId.RAGING_BULL, PokemonType.NORMAL, MoveCategory.PHYSICAL, 90, 100, 10, -1, 0, 9)
       .attr(RagingBullTypeAttr)
-      .attr(RemoveScreensAttr),
+      .attr(RemoveScreensAttr, user => (user.isPlayer() ? ArenaTagSide.ENEMY : ArenaTagSide.PLAYER)),
```
fabske0: Shouldn't this remove your screens when targeting your ally instead of always removing the enemy screens?

fabske0 (commented):
Also something probably for a different PR, but i noticed that rapid spin removes your toxic spikes (not tested with other hazards) but doesn't remove them from the flyout

src/data/moves/move.ts
DayKev: Oh yeah. Sigh, well that's why it needs a proper refactor instead of this bandaid fix.

Bertie690:
@fabske0 That likely indicates an event isn't being triggered for the flyout correctly.

Frankly, I'd say "don't bother & revisit after I refactor the arena flyout"

src/data/moves/move.ts
Scoooom: Does `target` have the `isPlayer` method? An inverse check might be a better default? That way it always applies the effect to the targets side of the field, which unless I am mistaken, is how most of these moves work (The titular move Rapid Spin being a key example of one that does not)
```      .attr(RemoveScreensAttr, target => (target.isPlayer() ? ArenaTagSide.PLAYER : ArenaTagSide.ENEMY)),```

DayKev: I used `user` because Rapid Spin specifically always removes hazards from the user's side of the field.
It would never have been an issue in the first place if all the moves only were based on the target's side of the field. Unfortunately it's a mix (Rapid Spin: based on user; Raging Bull: based on target) which is what's caused this mess.
I can add an extra param so it's `(user, target) => ...` and then switch between them based on the move.

DayKev: Fixed.

DayKev:
> Also something probably for a different PR, but i noticed that rapid spin removes your toxic spikes (not tested with other hazards) but doesn't remove them from the flyout

I wouldn't be surprised if the flyout code was making the same mistake as the move was.

Bertie690 (approved):

Bertie690:
@DayKev It isn't - trust me (I refactored the entire thing in a prior PR).
All the thing cares about is what side of the field the event is being emitted for.

What's likely happening is a removal event isn't being queued for the right side in the calling code.

PR #6925, opened 2025-12-30
[Misc] Cleanup `executeIf` to not use `null`

Bertie690:
N/A

Minor cleanups, since `undefined` is typically more reasonable for a util of that kind than `null`.
Made function use `promise.resolve` and return `undefined` instead of `null`


src/system/game-data.ts
```diff
@@ -1243,7 +1243,7 @@ export class GameData {
   saveAll(skipVerification = false, sync = false, useCachedSession = false, useCachedSystem = false): Promise<boolean> {
     return new Promise<boolean>(resolve => {
       executeIf(!skipVerification, updateUserInfo).then(success => {
-        if (success !== null && !success) {
+        if (success != null) {
```
Bertie690:
```suggestion
        if (success != null && !success) {
```

emdeann (approved):

PR #6933, opened 2026-01-02
[Dev] Strongly type Vite/Vitest configs; fix invalid options

Bertie690:
N/A
1. We were previously using `require` to load our plugins within Vite's config dynamically, which is icky and type-unsafe. (No fault to Benjie for trying, but I have a better solution.)
2. We were actually mis-typing the arguments passed to Vitest - without type annotations, TS did not enforce the return type of our default exports, leading to typos ( `coverage.reporters` should be `coverage.reporter`) and similarly nonexistent properties.

- Reworked the `defaultConfig` export into a `sharedConfig` async function that takes in a `ConfigEnv` object. 
  - This ensures we can add all properties common to both prod and test builds in a single place. (Previously, we duplicated esbuild between both config functions due to depending on the `mode` parameter, but are now free to consolidate them).
- ~~Enabled a few safety related parameters, like `expect.requireAssertions`.~~ Too annoying for now
- Added necessary `satisfies` statements to ensure TypeScript properly steps in and yells at us if we specify incorrect properties.

## How to test the changes?
N/A
## Checklist


Bertie690:
@fabske0 other than that?
I can revert it easy enougj

fabske0:
> @fabske0 other than that?
> I can revert it easy enougj

Looked good otherwise

test/utils/array.test.ts
```diff
@@ -64,8 +64,7 @@ describe("Utils - Array", () => {
       it("typed array input", () => {
         const arr = new Uint8Array([1, 2, 3, 4]);
         const result = subArray(arr, 2);
-        // @ts-expect-error We get a questionable error about Uint8Array not being assignable to Uint8Array...
-        expectTypeOf(result).toEqualTypeOf<Uint8Array>();
+        expectTypeOf(result).toEqualTypeOf<Uint8Array<ArrayBuffer>>();
```
SirzBenjie, line 67: Thanks!

test/test-utils/mocks/mock-console/mock-console.ts
```diff
@@ -5,6 +5,8 @@ import { type InspectOptions, inspect } from "node:util";
 import chalk, { type ChalkInstance } from "chalk";
 
 // Tell chalk we support truecolor
+// TODO: Find a reliable, easy way to determine whether the actual final output is being piped to a file WHILE CIRCUMVENTING VITEST,
+// and disable color if that is determined to be the case
```
SirzBenjie, line 9: This is really hard to tell in general. One would have to check the output stream. I'm sure we could also find a way to set some option for logging to a file somehow.

vitest.config.ts
```diff
@@ -4,64 +4,67 @@
...
+      env: {
+        TZ: "UTC",
       },
+      testTimeout: 20_000,
+      slowTestThreshold: 10_000,
+      // TODO: Vitest's current framework produces spurious errors for type tests with this option enabled.
+      // We should move our type tests to a separate folder not covered by normal tests, and then enable the option.
+      // expect: {
+      //   requireAssertions: true,
+      // },
```
SirzBenjie, line 31: We can do this pretty easily. Just need to have different vitest configs for different test sorts.
Also, if we can detect the options that were passed here, we can set this dynamically, right? I think there's a command line flag for doing the type tests. Can we configure this based on that?

test/utils/array.test.ts
Bertie690: yup - it previously accepts array or shared array buffer

test/test-utils/mocks/mock-console/mock-console.ts
Bertie690: I tried using `util.inspect` on the `console` object and got virtually identical results yeah.
Not undoable (IIRC Vitest themselves has functionality for it) but not easy by any means.

vitest.config.ts
Bertie690: @SirzBenjie We don't even run the type tests atm - it's just using `expectTypeOf` alongside `tsc`.
I do agree but that's a task for another day.

PR #6934, opened 2026-01-04
test: enable v8 coverage instrumentation

Bertie690:
N/A

I always wanted to get some sort of "code coverage badge", and now would be a good time to do it.

Swapped our coverage provider from Istanbul to V8 (which is faster and more accurate), and enabled using the latter on CI runs to output a coverage report.

Added a workflow step (powered by [dynamic-badge-action](https://github.com/schneegans/dynamic-badges-action)) that takes the coverage JSON file produced by combining the blobs and uploads it to a private gist whenever a beta test workflow completes. Failed runs will output a red "Tests: Failing" badge, while successes will show the current coverage %.

> [!NOTE]
> The personal access token I made for the gist upload is set to expire in Jan of next year as a trial.
> If all goes well, I'll likely make the replacement token persist indefinitely (after all, the only thing it can do is upload to gists).

> [!IMPORTANT]
> I cannot test the uploads portion of the workflow from the fork itself (due to not having access to main repo secrets), but have independently verified that the workflow **does work**.
> I copied over the `index.js` file from the action, filled in the inputs used in the workflow and was able to successfully upload a badge.

## Screenshots/Videos
image

Time per test run is mostly unchanged (~25% at max)
image
image

## How to test the changes?
LOOK with your EYES

## Checklist


SirzBenjie:
How useful is it to have coverage reports per PR?

Otherwise, I like this a lot. Great work!

Bertie690:
> How useful is it to have coverage reports per PR?

Not _that_ useful, but still nice to have.
@SirzBenjie Would it make sense to dynamically enable coverage only on pushes to main/beta?

vitest.config.ts
```diff
@@ -47,9 +50,10 @@ export default defineConfig(async config => {
       restoreMocks: true,
       watch: false,
       coverage: {
-        provider: "istanbul",
+        provider: "v8",
         reportsDirectory: "coverage",
-        reporter: ["text-summary", "html"],
+        reporter: process.env.MERGE_REPORTS ? ["text-summary", "json-summary"] : [],
+        exclude: ["public", "assets", "locales"],
```
DayKev: I don't think this does anything. What you should do is add `include: ["src/**/*.ts"]` instead.
cf https://vitest.dev/guide/coverage.html#including-and-excluding-files-from-coverage-report

package.json
```diff
@@ -47,7 +47,7 @@
     "@types/crypto-js": "^4.2.2",
     "@types/jsdom": "^27.0.0",
     "@types/node": "^24.10.2",
-    "@vitest/coverage-istanbul": "^4.0.15",
+    "@vitest/coverage-v8": "^4.0.16",
     "@vitest/expect": "^4.0.15",
     "@vitest/utils": "^4.0.15",
```
DayKev: All `@vitest` deps should match `vitest`'s version.

DayKev:
> > How useful is it to have coverage reports per PR?
> 
> Not _that_ useful, but still nice to have. @SirzBenjie Would it make sense to dynamically enable coverage only on pushes to main/beta?

Yes, there's no need to slow down PR tests to run coverage.

Also "25%" is not "mostly unchanged"...

.github/workflows/tests.yml
```diff
@@ -21,6 +21,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  COVERAGE_DIR: coverage/coverage-summary.json
```
DayKev: A file isn't a directory.

vitest.config.ts
Bertie690: @DayKev I read the coverage report and it included stuff in the submodules.

.github/workflows/tests.yml
Bertie690: fine

.github/workflows/tests.yml
```diff
@@ -132,7 +134,7 @@ jobs:
           forceUpdate: true
 
       - name: Extract coverage percentage
-        run: echo "COVERAGE_PERCENTAGE=$(jq '.total.branches.pct' <<< "${{needs.check-results.outputs.coverage_report}}")" >> $GITHUB_ENV
+        run: echo "COVERAGE_PERCENTAGE="$(jq '.total.branches.pct' <<< "${{needs.check-results.outputs.coverage_report}}")""" >> $GITHUB_ENV
```
DayKev: That looks to be an unbalanced number of `"`s.

package.json
```diff
@@ -13,11 +13,11 @@
...
     "build:app": "vite build --mode app && node scripts/sync-dist-assets.js",
     "preview": "vite preview",
-    "test": "vitest run --no-isolate",
-    "test:cov": "vitest run --coverage --no-isolate",
-    "test:watch": "vitest watch --coverage --no-isolate",
-    "test:silent": "vitest run --silent='passed-only' --no-isolate",
-    "test:merge-reports": "MERGE_REPORTS=1 vitest run --merge-reports=test-results --silent=passed-only --configLoader runner",
+    "test": "vitest run",
+    "test:cov": "vitest run --silent='passed-only' --coverage",
+    "test:watch": "vitest watch--silent='passed-only'",
```
DayKev:
```suggestion
    "test:watch": "vitest watch --silent='passed-only'",
```

.github/workflows/tests.yml
Bertie690: oops 1 too many
```suggestion
        run: echo "COVERAGE_PERCENTAGE="$(jq '.total.branches.pct' <<< "${{needs.check-results.outputs.coverage_report}}")"" >> $GITHUB_ENV
```

PR #6942, opened 2026-01-06
refactor(ui): don't display form name in battle info UI

DayKev:
In the battle info UI, the Pokémon's name will no longer be prepended with "Mega"/etc.

It takes up too much of the limited space available, especially in late endless.

Add `prependFormName` param to `Pokemon#getNameToRender` and pass `false` for the param in `BattleInfo`.
The param causes the method to return `this.species.getName()` with no `formIndex` param passed to `getName()`, instead of `this.name`.

## Screenshots/Videos
<details><summary>Example with Mega Charizard X</summary>
<p>

image

</p>
</details>

## Checklist


damocleas (dismissed):

SirzBenjie:
So, for consistency, I don't like that the prompt still reads "What will *mega* Charizard X do?" I feel like that should match.

DayKev:
> So, for consistency, I don't like that the prompt still reads "What will _mega_ Charizard X do?" I feel like that should match.

Changed.
<p>

image

</p>
</details>

PR #1960, opened 2024-06-08
[Other(qol)]: Lint pr title

OrangeRed:
## What are the changes?
Add workflow to lint the title of PR so it matches the [conventionalcommits](https://conventionalcommits.org) spec

```
Custom Conventional Commits spec

[bug(ui)]: Fix female trainer names
 ^   ^     ^
 |   |     |__ Subject
 |   |________ Scope
 |____________ Prefix
```

## Why am I doing these changes?
This is the first step to needed to auto generate changelogs from commit history

## What did change?
new workflow

### Screenshots/Videos
N/A

## How to test the changes?
make a new branch, make some dummy commits and open a pull request onto `OrangeRed/lint-pr-title` trigger the workflow.
You can test the regex that I'm using here: https://regex101.com/r/42XxHM/1

## Checklist
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes (manually)?
   ~~- [ ] Are all unit tests still passing? (`npm run test`)~~
~~- [ ] Are the changes visual?~~
  ~~- [ ] Have I provided screenshots/videos of the changes?~~

PR #6943, opened 2026-01-07
github: add PR title lint workflow

DayKev:
- Based on https://github.com/pagefaultgames/pokerogue/pull/1960
- Port of https://github.com/Despair-Games/poketernity/pull/1313

N/A

To improve consistency in PR naming.
This will also allow [partially] automated PR labelling to be added as well.

- PR titles must now be in the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format.
  Example: `fix(move): Future Sight no longer crashes`

  Full format details in the newly added "Submitting a Pull Request" section of `CONTRIBUTING.md`.

- The pull request template was updated.

- `@actions/core` and `@actions/github` were added as optional dependencies which can be ignored when installing packages via `pnpm i --no-optional`. This provides type checking when editing the `pr-title.ts` file.

## Checklist

### DIFF
--- a/.github/workflows/lint-pr-title.yml
+++ b/.github/workflows/lint-pr-title.yml
@@ -22,7 +22,6 @@ jobs:
         with:
           submodules: false
           sparse-checkout: |
-            package.json
             pnpm-lock.yaml
             .nvmrc
             .github/scripts/pr-title.ts
@@ -30,6 +29,8 @@ jobs:
 
       - name: Install pnpm
         uses: pnpm/action-setup@v4
+        with:
+          version: 10
 
       - name: Set up Node.js
         uses: actions/setup-node@v6


CONTRIBUTING.md
```diff
@@ -138,8 +145,69 @@ Most non-trivial changes (*especially bug fixes*) should come along with new tes
...
 2. Select [everything.prsv](test/test-utils/saves/everything.prsv) (`test/test-utils/saves/everything.prsv`) and confirm.
+
+
+## ✅ Submitting a Pull Request
+
+Most information related to submitting a pull request is contained in comments within the pull request template that is shown when you open a new pull request,
+however full documentation on the pull request title format is here to best utilize the space available.
+
+The pull request title must follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format with a valid prefix and optionally a valid scope. \
+If a save migrator or version increase is part of the PR, a `!` must be added before the `:`.
```
Bertie690:
```suggestion
If a save migrator, version increase or other breaking change is part of the PR, a `!` must be added before the `:`.
```

DayKev: There isn't really a concept of breaking changes for us (we can pretty much always construct a version migrator) which is why I just mentioned the 2 scenarios where it's relevant but I guess I can just put a catchall in there anyway.

.github/workflows/lint-pr-title.yml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v4
+        with:
+          submodules: false
+          sparse-checkout: |
+            package.json
+            pnpm-lock.yaml
+            tsconfig.json
```
SirzBenjie: If you don't do cone mode, you don't need to specify any files that are in root; they'll all be there.

CONTRIBUTING.md
```diff
@@ -138,8 +145,69 @@ Most non-trivial changes (*especially bug fixes*) should come along with new tes
...
+- "event" - e.g. adding a Christmas event to the game
+- "graphics" - Anything related to art/graphics (adding new sprites, fixing a sprite that isn't displaying properly, etc)
+- "item"
+- "move"
+- "ui" - UI/UX
+
+> [!IMPORTANT]
+> All scopes are valid when using the "feat", "fix", or "refactor" prefixes. \
+> All scopes except "audio", "battle", "graphics", and "ui" are valid when using the "balance" prefix. \
+> No other prefixes have valid scopes.
```
SirzBenjie, line 213: Can we make `docs` have optional valid scopes?

Bertie690: I'd say it should have all valid prefixes yeah.

.github/workflows/lint-pr-title.yml
SirzBenjie: Looking at it, we might _prefer_ to use `sparse-checkout-cone-mode` here, and then modify the workflow steps to install the two `@actions` dependencies.

.github/workflows/lint-pr-title.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: pnpm/action-setup@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: ".nvmrc"
+          cache: "pnpm"
+
+      - name: Install Node.js dependencies
+        run: pnpm add @actions/core @actions/github tsx
```
SirzBenjie: With this line here, why do we also need to add the optional dependencies? (See other comment)
Could change this to `pnpm add @actions/core@2.0.1 @actions/github@6.0.1`

We also don't need `tsx`. We can use node's new ability to run typescript directly, we can just directly execute `.ts` files. This will work as long as the file only has erasable syntax. (which it does, and is easy to keep it that way).

package.json
```diff
@@ -68,20 +83,9 @@
...
-    "i18next-browser-languagedetector": "^8.2.0",
-    "i18next-http-backend": "^3.0.2",
-    "i18next-korean-postposition-processor": "^1.0.0",
-    "json-stable-stringify": "^1.3.0",
-    "jszip": "^3.10.1",
-    "phaser": "^3.90.0",
-    "phaser3-rex-plugins": "^1.80.17"
+  "optionalDependencies": {
+    "@actions/core": "^2.0.1",
+    "@actions/github": "^6.0.1"
```
SirzBenjie, line 87: With my other comment, it might be better to not have these in `package.json` and install a specific version directly from the repo. It seems really weird to me to put workflow dependencies in `package.json` here.

CONTRIBUTING.md
fabske0, line 213: I feel like tests should allow specifying what they are for

package.json
DayKev: Then the deps wouldn't be installed when editing the workflow's `.ts` file (as mentioned in the PR description).

.github/workflows/lint-pr-title.yml
DayKev: Ideally it would only install those 3 deps but I was having trouble getting it to work without grabbing both `package.json` and `pnpm-lock.yaml`.
Though putting them in the optional deps section is for a different reason anyway.

package.json
DayKev: Also isn't that the point of `"optionalDependencies"`? To have packages that aren't necessary for the app to be run, but are still useful in some circumstances to have installed.

DayKev: image
image

`pnpm install --no-optional` can be used by anyone who doesn't want to install the `@actions` deps. Should I update the lefthook command to add `--no-optional`?

.github/workflows/lint-pr-title.yml
DayKev: Realized I didn't need `package.json`, just `pnpm-lock.yaml`.

Now it only installs the `@actions` packages. Also removed `tsx` and minimized the number of files checked out from the repo.

package.json
SirzBenjie: Ah, of course we need the optional deps installed for linting while writing the code. This seems fine, then.

PR #6951, opened 2026-01-10
feat(ui): expose login error messages

### LINKED ISSUE #6769: [Feature] Expose login/registration error messages to the end user

When creating an account, any errors are not exposed to the end user. Only "Unknown error" is seen, which isn't helpful. An error message is exposed in the browser console, however this isn't accessible when trying to play on a mobile browser. 

### Describe the Feature

Request is to

1. For standard/"expected" error messages, handle them with a user-friendly string
2. For unexpected error messages, display unknown error with an "advanced" button showing the error message so people can Google/report it here 

### Additional context

Relevant code looks to be https://github.com/pagefaultgames/pokerogue/blob/beta/src%2Fplugins%2Fapi%2Fpokerogue-account-api.ts#L55

I'm not familiar with this project but I'd be happy to have a crack at implementing this.

HarrySeop:

When server-side errors occur during login, registration, or password change operations, users will now see user-friendly error messages instead of technical error strings or "Unknown error".

- Fixes #6769

### Error Mapping Table

| Backend Error                   | Frontend Mapping                 | Affected Screen              | Backend Source                                                                                                                                                                                                     |
| ------------------------------- | -------------------------------- | ---------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| `failed to generate token`      | `menu:serverErrorGenerateToken`  | Login                        | [login.go:L60](https://github.com/pagefaultgames/rogueserver/blob/master/api/account/login.go#L60), [L81](https://github.com/pagefaultgames/rogueserver/blob/master/api/account/login.go#L81)                      |
| `failed to add account session` | `menu:serverErrorAddSession`     | Login                        | [login.go:L85](https://github.com/pagefaultgames/rogueserver/blob/master/api/account/login.go#L85)                                                                                                                 |
| `failed to generate uuid`       | `menu:serverErrorGenerateUuid`   | Registration                 | [register.go:L39](https://github.com/pagefaultgames/rogueserver/blob/master/api/account/register.go#L39)                                                                                                           |
| `failed to generate salt`       | `menu:serverErrorGenerateSalt`   | Registration/Change Password | [register.go:L45](https://github.com/pagefaultgames/rogueserver/blob/master/api/account/register.go#L45), [changepw.go:L37](https://github.com/pagefaultgames/rogueserver/blob/master/api/account/changepw.go#L37) |
| `failed to remove sessions`     | `menu:serverErrorRemoveSessions` | Change Password              | [changepw.go:L42](https://github.com/pagefaultgames/rogueserver/blob/master/api/account/changepw.go#L42)                                                                                                           |
| `failed to add account record`  | `menu:serverErrorUpdateAccount`  | Change Password              | [changepw.go:L47](https://github.com/pagefaultgames/rogueserver/blob/master/api/account/changepw.go#L47) 

### Backend Coverage

All server-side errors from `rogueserver` are now properly handled:

- ✅ `api/account/login.go` - 6 errors (2 new mappings)
- ✅ `api/account/register.go` - 5 errors (2 new mappings)
- ✅ `api/account/changepw.go` - 4 errors (3 new mappings)

**No server errors are left unmapped.**

## Screenshots/Videos

See pagefaultgames/pokerogue-locales#272

## Checklist

- [x] A locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo
  - [x] Link to locales PR: pagefaultgames/pokerogue-locales#272
- [x] The translation team been contacted for proofreading/translation on Discord

### DIFF
--- a/src/ui/handlers/change-password-form-ui-handler.ts
+++ b/src/ui/handlers/change-password-form-ui-handler.ts
@@ -6,14 +6,14 @@ import type { InputFieldConfig } from "#ui/form-modal-ui-handler";
 import { FormModalUiHandler } from "#ui/form-modal-ui-handler";
 import i18next from "i18next";
 
-export class ChangePasswordFormUiHandler extends FormModalUiHandler {
-  private readonly ERR_PASSWORD: string = "invalid password";
-  private readonly ERR_ACCOUNT_EXIST: string = "account doesn't exist";
-  private readonly ERR_PASSWORD_MISMATCH: string = "password doesn't match";
-  private readonly ERR_GENERATE_SALT: string = "failed to generate salt";
-  private readonly ERR_REMOVE_SESSIONS: string = "failed to remove sessions";
-  private readonly ERR_ADD_RECORD: string = "failed to add account record";
+const ERR_PASSWORD: string = "invalid password";
+const ERR_ACCOUNT_EXIST: string = "account doesn't exist";
+const ERR_PASSWORD_MISMATCH: string = "password doesn't match";
+const ERR_GENERATE_SALT: string = "failed to generate salt";
+const ERR_REMOVE_SESSIONS: string = "failed to remove sessions";
+const ERR_ADD_RECORD: string = "failed to add account record";
 
+export class ChangePasswordFormUiHandler extends FormModalUiHandler {
   setup(): void {
     super.setup();
   }
@@ -40,16 +40,18 @@ export class ChangePasswordFormUiHandler extends FormModalUiHandler {
       error = error.slice(0, colonIndex);
     }
     switch (error) {
-      case this.ERR_PASSWORD:
+      case ERR_PASSWORD:
         return i18next.t("menu:invalidRegisterPassword");
-      case this.ERR_ACCOUNT_EXIST:
+      case ERR_ACCOUNT_EXIST:
         return i18next.t("menu:accountNonExistent");
-      case this.ERR_PASSWORD_MISMATCH:
+      case ERR_PASSWORD_MISMATCH:
         return i18next.t("menu:passwordNotMatchingConfirmPassword");
-      case this.ERR_GENERATE_SALT:
-      case this.ERR_REMOVE_SESSIONS:
-      case this.ERR_ADD_RECORD:
-        return i18next.t("menu:serverError");
+      case ERR_GENERATE_SALT:
+        return i18next.t("menu:serverErrorGenerateSalt");
+      case ERR_REMOVE_SESSIONS:
+        return i18next.t("menu:serverErrorRemoveSessions");
+      case ERR_ADD_RECORD:
+        return i18next.t("menu:serverErrorUpdateAccount");
     }
 
     return super.getReadableErrorMessage(error);
@@ -87,7 +89,7 @@ export class ChangePasswordFormUiHandler extends FormModalUiHandler {
             return onFail(this.getReadableErrorMessage("invalid password"));
           }
           if (passwordInput.text !== confirmPasswordInput.text) {
-            return onFail(this.ERR_PASSWORD_MISMATCH);
+            return onFail(ERR_PASSWORD_MISMATCH);
           }
 
           pokerogueApi.account.changePassword({ password: passwordInput.text }).then(error => {
--- a/src/ui/handlers/change-password-form-ui-handler.ts
+++ b/src/ui/handlers/change-password-form-ui-handler.ts
@@ -10,6 +10,9 @@ export class ChangePasswordFormUiHandler extends FormModalUiHandler {
   private readonly ERR_PASSWORD: string = "invalid password";
   private readonly ERR_ACCOUNT_EXIST: string = "account doesn't exist";
   private readonly ERR_PASSWORD_MISMATCH: string = "password doesn't match";
+  private readonly ERR_GENERATE_SALT: string = "failed to generate salt";
+  private readonly ERR_REMOVE_SESSIONS: string = "failed to remove sessions";
+  private readonly ERR_ADD_RECORD: string = "failed to add account record";
 
   setup(): void {
     super.setup();
@@ -43,6 +46,10 @@ export class ChangePasswordFormUiHandler extends FormModalUiHandler {
         return i18next.t("menu:accountNonExistent");
       case this.ERR_PASSWORD_MISMATCH:
         return i18next.t("menu:passwordNotMatchingConfirmPassword");
+      case this.ERR_GENERATE_SALT:
+      case this.ERR_REMOVE_SESSIONS:
+      case this.ERR_ADD_RECORD:
+        return i18next.t("menu:serverError");
     }
 
     return super.getReadableErrorMessage(error);
--- a/src/ui/handlers/change-password-form-ui-handler.ts
+++ b/src/ui/handlers/change-password-form-ui-handler.ts
@@ -8,12 +8,12 @@ import i18next from "i18next";
 
 // TODO: Consider replacing server error strings with numeric error codes for better maintainability
 // TODO: Centralize server error constants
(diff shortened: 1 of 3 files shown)


src/ui/handlers/change-password-form-ui-handler.ts
```diff
@@ -10,6 +10,9 @@ export class ChangePasswordFormUiHandler extends FormModalUiHandler {
   private readonly ERR_PASSWORD: string = "invalid password";
   private readonly ERR_ACCOUNT_EXIST: string = "account doesn't exist";
   private readonly ERR_PASSWORD_MISMATCH: string = "password doesn't match";
+  private readonly ERR_GENERATE_SALT: string = "failed to generate salt";
```
Bertie690: This isn't a thing at you, but a symptom of the wider context - there is no reason at all for these to be properties on the UI handler itself.
ESBuild won't know that these are constant values and thus cannot minify them at use sites.

Just throwing it out there.

src/ui/handlers/change-password-form-ui-handler.ts
```diff
@@ -47,6 +50,10 @@ export class ChangePasswordFormUiHandler extends FormModalUiHandler {
         return i18next.t("menu:accountNonExistent");
       case this.ERR_PASSWORD_MISMATCH:
         return i18next.t("menu:passwordNotMatchingConfirmPassword");
+      case this.ERR_GENERATE_SALT:
+      case this.ERR_REMOVE_SESSIONS:
+      case this.ERR_ADD_RECORD:
+        return i18next.t("menu:serverError");
```
Bertie690: If we know what error code was produced, we should probably specify as such rather than relying on post hoc server logs and whatnot.

Perhaps the locale key would have an extra field to specify the type of error produced?

src/ui/handlers/registration-form-ui-handler.ts
```diff
@@ -43,6 +43,9 @@ export class RegistrationFormUiHandler extends LoginRegisterInfoContainerUiHandl
         return i18next.t("menu:invalidRegisterPassword");
       case "failed to add account record":
         return i18next.t("menu:usernameAlreadyUsed");
+      case "failed to generate uuid":
```
Bertie690: Should this be UUID?

Then again idk how the backend server messages are formatted

DayKev (dismissed):

Bertie690 (changes requested):
@HarrySeop 
Code looks decent.

~~THAT SAID, since we are literally `switch case`-ing on Rogueserver error text, I'd like to see some info (in the PR itself) showing _where exactly_ in the backend you found those hardcoded strings.

An extra column in the markdown table with some GitHub permalinks would do fine - I just don't particularly want to go around chasing down everything myself.~~

EDIT - turns out checking took way less time than i thought, they're all there.

src/ui/handlers/registration-form-ui-handler.ts
DayKev: I checked and they match the server messages being sent.

src/ui/handlers/change-password-form-ui-handler.ts
DayKev: Ah, I didn't pull these out of the class like I did with the login and registration UI handlers since I didn't touch this file at the time of the refactor.
It should be done here too, though idk if it matters for this PR.

src/ui/handlers/registration-form-ui-handler.ts
Bertie690: welp uhhhh guess that's a good spot for a TODO then!

(yes i care this much about proper acronym capitalization)

src/ui/handlers/change-password-form-ui-handler.ts
Bertie690: My thoughts exactly - this is a "later" problem

src/ui/handlers/registration-form-ui-handler.ts
DayKev: Tbh the server should be sending error codes since that's how these messages are being used. The actual error messages being displayed are the `return i18next.t(...)`s.

HarrySeop: Thank you for pointing this out! I kept it as lowercase `"uuid"` to match the exact server message format from [register.go:L39](https://github.com/pagefaultgames/rogueserver/blob/master/api/account/register.go#L39).

Since these error strings are used as switch-case keys, changing the capitalization here would break the matching logic unless the server is also updated.

Would you prefer I:
1. Add a TODO comment noting this should be `"UUID"` once server-side messages are standardized?
2. Keep it as-is for now since it matches the current server output?

I'm leaning toward option 1, but I'd appreciate your thoughts on this!

src/ui/handlers/change-password-form-ui-handler.ts
HarrySeop: Thank you for the feedback! - providing specific error information would be much more helpful.

I see two possible approaches:

**Option 1: Pass error type as parameter**
```typescript
return i18next.t("menu:serverError", { errorType: "GENERATE_SALT" });
```
- Single locale key, but error type appears in the message

**Option 2: Separate locale keys per error**
```typescript
return i18next.t("menu:serverErrorGenerateSalt");
return i18next.t("menu:serverErrorGenerateToken");
```
- Multiple locale keys, but allows specific user-friendly messages

I'm leaning toward **Option 2** as it provides better UX with specific messages like "Failed to generate authentication token" instead of showing technical error codes to users.

**Which approach would you prefer?** I'm happy to implement either one!

HarrySeop:
> @HarrySeop Code looks decent.
> 
> ~~THAT SAID, since we are literally `switch case`-ing on Rogueserver error text, I'd like to see some info (in the PR itself) showing _where exactly_ in the backend you found those hardcoded strings.
> 
> An extra column in the markdown table with some GitHub permalinks would do fine - I just don't particularly want to go around chasing down everything myself.~~
> 
> EDIT - turns out checking took way less time than i thought, they're all there.

@Bertie690  I apologize for not including that information initially! I've now added a reference column to the table in the PR description with GitHub permalinks to each error string in the rogueserver code.

I've also left a comment on your review about the direction for error-specific messages. I'd appreciate it if you could take a look when you have a chance. Thank you!

src/ui/handlers/registration-form-ui-handler.ts
DayKev: I would add a TODO mentioning that the server's error messages should be updated or replaced with numeric error codes in the future.

src/ui/handlers/change-password-form-ui-handler.ts
HarrySeop: I've implemented **Option 2** with separate locale keys for each error type. This approach provides better UX with specific, user-friendly messages (e.g., "Unable to generate authentication token") rather than exposing technical error codes to users. Each error now has a distinct, identifiable message for easier debugging.

src/ui/handlers/registration-form-ui-handler.ts
HarrySeop: Added a TODO comment noting that server error messages should be standardized with error codes in the future. Thanks for the guidance!

src/ui/handlers/change-password-form-ui-handler.ts
HarrySeop: Addressed this by moving the error constants outside the class to module-level `const` in this PR. Thanks for pointing out the ESBuild minification issue!

DayKev (dismissed):

Bertie690 (dismissed):

Wlowscha (dismissed):

DayKev:
Converted to draft since we need to update the locales, then this is ready.

DayKev (approved):

Wlowscha (approved):

PR #6971, opened 2026-01-14
fix(move): Magnitude power values and thresholds set to correct values

DayKev:
Magnitude will now have the correct move base power values and will output them with the correct frequency.

The move is bugged.

- Removed the extra incorrect entry from the move power array and fixed the incorrect value in the RNG thresholds array.
- Ported the RNG helper from pkty for the test.
- Cleaned up a bunch of docs on the RNG functions in the `Pokemon` class.

## How to test the changes?
`pnpm test:silent magnitude`

## Checklist
  - [x] I have created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes

### DIFF
--- a/test/moves/magnitude.test.ts
+++ b/test/moves/magnitude.test.ts
@@ -1,0 +1,48 @@
+import { allMoves } from "#data/data-lists";
+import { AbilityId } from "#enums/ability-id";
+import { MoveId } from "#enums/move-id";
+import { SpeciesId } from "#enums/species-id";
+import { GameManager } from "#test/test-utils/game-manager";
+import { NumberHolder } from "#utils/common";
+import Phaser from "phaser";
+import { beforeAll, beforeEach, describe, expect, it } from "vitest";
+
+describe("Move - Magnitude", () => {
+  let phaserGame: Phaser.Game;
+  let game: GameManager;
+
+  beforeAll(() => {
+    phaserGame = new Phaser.Game({
+      type: Phaser.HEADLESS,
+    });
+  });
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override
+      .ability(AbilityId.BALL_FETCH)
+      .battleStyle("single")
+      .criticalHits(false)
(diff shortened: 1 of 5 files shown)


test/moves/magnitude.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  it("should have the correct rng thresholds and power values", async () => {
+    await game.classicMode.runToSummon([SpeciesId.FEEBAS]);
+
+    const magnitudeAttr = allMoves[MoveId.MAGNITUDE].getAttrs("MagnitudePowerAttr")[0];
+    const movePower = new NumberHolder(0);
+    const results = { [0]: 0, [10]: 0, [30]: 0, [50]: 0, [70]: 0, [90]: 0, [110]: 0, [150]: 0 };
+
+    await game.rng.equalSample(100, () => {
+      movePower.value = 0;
+      magnitudeAttr.apply(null!, null!, null!, [movePower]);
```
Bertie690, line 42: _screams at the sight of `null!`_

Bertie690 (approved):
@DayKev IIRC there was some other test code that used a similar method for egg probabilities/etc. 
Might be worth testing

test/moves/magnitude.test.ts
DayKev: If move attrs took objects like ability attrs then I could just pass in `{ movePower }` but sadly they don't (yet).

DayKev:
> It might be too early for me but I definitely don't understand how `equalSample` is meant to be used.

@SirzBenjie If you have `randSeedInt(100)` somewhere in production code and do `equalSample(100, ...)` then you'll get each value from `0` to `99` once as the output of that `randSeedInt` call. If you did the same but it was `randSeedInt(10)` then it'd output each value `0` to `9` ten times.
It basically forces the RNG functions to linearly interpolate from their min and max values based on the value you input to `equalSample` (at least that's my understanding of it, PigeonBar was the one who wrote it; you can ask them about it in the pkty Discord for more information if you need to).

PR #6974, opened 2026-01-14
misc: clean up `getLastXMoves` docs; fix callsites

Bertie690:
N/A
I wanted to clarify the difference between `getLastXMoves` and `getLastNonVirtualMove`, as well as optimize them both to use `findLast` and `toReversed` instead of `slice().reverse()`/etc.

Updated the 2 functions

Fixed a few random cases of shallowly cloning a shallowly cloned array
## Screenshots/Videos
N/A
## How to test the changes?
N/A
## hecklist


src/data/moves/move.ts
```diff
@@ -9656,11 +9656,10 @@ export function initMoves() {
       .ignoresProtect()
       .attr(DestinyBondAttr)
       .condition(failAgainstFinalBossCondition, 2)
+      // Destiny Bond fails if it was successfully used last turn
       .condition((user, _target, move) => {
-        const lastTurnMove = user.getLastXMoves(1);
-        return (
-          lastTurnMove.length === 0 || lastTurnMove[0].move !== move.id || lastTurnMove[0].result !== MoveResult.SUCCESS
-        );
+        const lastTurnMove: TurnMove | undefined = user.getLastXMoves(1)[0];
```
```suggestion
        const lastTurnMove = user.getLastXMoves(1).at(0);
```
`arr.at(index)` is type safe `arr[index]`

src/data/pokemon-forms/form-change-triggers.ts
```diff
@@ -182,9 +182,14 @@ export class SpeciesFormChangePreMoveTrigger extends SpeciesFormChangeMoveTrigge
 export class SpeciesFormChangePostMoveTrigger extends SpeciesFormChangeMoveTrigger {
   description = i18next.t("pokemonEvolutions:forms.postMove");
   canChange(pokemon: Pokemon): boolean {
-    return (
-      pokemon.summonData && pokemon.getLastXMoves(1).filter(m => this.movePredicate(m.move)).length > 0 === this.used
-    );
+    if (!pokemon.summonData) {
+      return false;
+    }
```
DayKev:
```suggestion
```
This isn't necessary, `summonData` is always defined now since you changed it a while ago.

src/data/pokemon-forms/form-change-triggers.ts
```diff
@@ -182,9 +182,14 @@ export class SpeciesFormChangePreMoveTrigger extends SpeciesFormChangeMoveTrigge
...
+    const lastMove = pokemon.getLastXMoves(1)[0];
```
DayKev:
```suggestion
    const lastMove = pokemon.getLastXMoves(1).at(0);
```

src/field/pokemon.ts
```diff
@@ -3909,6 +3909,8 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * @param ignoreFaintPhase - Whether to ignore adding a FaintPhase if this damage causes a faint
    * @returns The actual damage dealt
    */
+  // TODO: Rework this to not use a dummy `ignoreSegments` parameter that is only used by a superclass
+  // TODO: This should ostensibly be private, but Pain Split still uses it for whatever reason...
```
```suggestion
  // TODO: Remove uses of this outside of the `Pokemon` class and subclasses (known use: Paint Split) and change to `protected`
```

src/field/pokemon.ts
```diff
@@ -3909,6 +3909,8 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * @param ignoreFaintPhase - Whether to ignore adding a FaintPhase if this damage causes a faint
    * @returns The actual damage dealt
    */
+  // TODO: Rework this to not use a dummy `ignoreSegments` parameter that is only used by a superclass
```
```suggestion
  // TODO: Rework this to use an object for the optional parameters
```

src/field/pokemon.ts
```diff
@@ -4394,19 +4396,23 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * @privateRemarks
+   * Callers that want to obtain the last move actually _executed_ (i.e. selected from the user's moveset)
+   * should use {@linkcode getLastNonVirtualMove} instead.
    */
-  // TODO: Update documentation in dancer PR to mention "getLastNonVirtualMove"
-  public getLastXMoves(moveCount = 1): TurnMove[] {
-    const moveHistory = this.getMoveHistory();
-    if (moveCount > 0) {
-      return moveHistory.slice(Math.max(moveHistory.length - moveCount, 0)).reverse();
+  // TODO: Most of the time, callers either pass 1 or -1; review whether we need to even allow other values
```
```suggestion
```
There's no reason to restrict it. There are callsites that use different numbers, restricting it arbitrarily just forces them to do `(-1).slice(...)` instead of passing the number to the function.

src/field/pokemon.ts
```diff
@@ -4394,19 +4396,23 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+  // TODO: Most moves accessing this can be reworked to use the current "move in flight" once implemented
+  public getLastXMoves(moveCount = 1): readonly TurnMove[] {
+    const hist = this.getMoveHistory().toReversed();
```
DayKev, line 4475: Why did you rename the variable? There was no need to abbreviate it.

src/field/pokemon.ts
```diff
@@ -4394,19 +4396,23 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * Callers that want to obtain the last move actually _executed_ (i.e. selected from the user's moveset)
+   * should use {@linkcode getLastNonVirtualMove} instead.
    */
-  // TODO: Update documentation in dancer PR to mention "getLastNonVirtualMove"
-  public getLastXMoves(moveCount = 1): TurnMove[] {
-    const moveHistory = this.getMoveHistory();
-    if (moveCount > 0) {
-      return moveHistory.slice(Math.max(moveHistory.length - moveCount, 0)).reverse();
+  // TODO: Most of the time, callers either pass 1 or -1; review whether we need to even allow other values
+  // TODO: Most moves accessing this can be reworked to use the current "move in flight" once implemented
```
DayKev, line 4473:
```suggestion
```
We don't need this `TODO` [here].

src/field/pokemon.ts
```diff
@@ -4420,7 +4426,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * or `undefined` if no applicable moves have been used since switching in.
    */
   public getLastNonVirtualMove(ignoreStruggle = false, ignoreFollowUp = true): TurnMove | undefined {
-    return this.getLastXMoves(-1).find(
+    return this.summonData.moveHistory.findLast(
```
```suggestion
    return this.getMoveHistory().findLast(
```
Why did you change this? It should at least be `this.getMoveHistory().findLast`.

src/data/moves/move.ts
Bertie690: Is it as performant?

src/field/pokemon.ts
Bertie690:
```suggestion
  // TODO: Remove uses of this outside of the `Pokemon` class and subclasses and change to `protected`
  // Known violators: Pain Split, Status effect code
```

src/field/pokemon.ts
Bertie690: Find last is faster yeah - doesn't need to shallowly clone the array and reverse it

src/field/pokemon.ts
Bertie690: Can i keep it in as a reminder

src/data/moves/move.ts
DayKev: The difference is predictably irrelevant.
<details>

image

</details>

src/field/pokemon.ts
DayKev: I guess.

src/field/pokemon.ts
Bertie690: because it isn't the result of `moveHistory`anymore - it's a reversed copy of it

I prefer to avoid conflating my FIFO and LIFO arrays

(Also as an aside: does it really matter what we call the variable?)

src/data/battler-tags.ts
```diff
@@ -2071,7 +2071,7 @@ export class TruantTag extends AbilityBattlerTag {
       return super.lapse(pokemon, lapseType);
     }
 
-    const lastMove = pokemon.getLastXMoves()[0];
+    const lastMove = pokemon.getLastXMoves(1)[0];
```
Wlowscha:
```suggestion
    const lastMove = pokemon.getLastXMoves(1).at(0);
```

Wlowscha:
Is there a specific reason why we are passing 1 to `getLastXMoves` even though it's the default parameter? I agree that it's more clear, but then wouldn't it be better to make `moveCount` not optional?

Bertie690:
Good question - imma leave it for another day.

PR #6980, opened 2026-01-16
fix: Prevent double daily starters

fabske0:

Dailies won't be able to have the same starter twice

Happened today

Added an additional filter to filter out already picked speciesIds

## Screenshots/Videos

image

</details>

image

</details>

## How to test the changes?

Check the starters with the below seed before and after the PR
```ts
const overrides = {
  DAILY_RUN_SEED_OVERRIDE: "3JSTxAx7NMOGBbxMLh1csQ=="
```

## Checklist

  - [ ] I have created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes
  - [ ] I have made sure that any UI change works for both the default and legacy UI themes (if applicable)


damocleas (dismissed):

SirzBenjie (commented):
So this is good, but there's a subtle gotcha that means this won't prevent duplicates in all cases. The pokemon being generated can select a lower evolution level, which is then set to a later stage (see `pokemon#getTrainerSpeciesForLevel`). To get around this, you might consider using `getRootSpeciesId`. That would have the impact of preventing, for example, a hitmonchan and hitmonlee on the same team, however. An alternative would be do use a `do while` loop, conditioned on the speciesID not being in the party.

Madmadness65 (dismissed):

PR #6924, opened 2025-12-30
refactor: clean up `Arena.randomSpecies`

Bertie690:
## What are the changes the user will see?
N/A

## Why am I making these changes?
The `Arena#randomSpecies` function is FAR too complex for its own good.

Also strong typing is nice.
## What are the changes from a developer perspective?
Ported a few pkty refactors to that function, among other things making `Arena#PokemonPools` actually correctly typed instead of using generic `number` index signatures. 

## Screenshots/Videos
N/A
## How to test the changes?
play game, see no crash
## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`pnpm test:silent`)
  - [x] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?

PR #6985, opened 2026-01-18
refactor: enable `exactOptionalPropertyTypes` in `tsconfig.json`

Bertie690:
N/A

Works towards the goal of increased type strictness, while also improving the ergonomics of `Object.values` and the `Object.keys` overloads introduced in #6924.

Reasons to care:
1. Both `in` guards and `Object.keys`/`values`/`entries` care about presence vs "present but `undefined`"
2. Allows distinguishing between "absent" vs "present but `undefined` for the purposes of object records. 
   Without the option, there's no way to enforce that a `Record<K, V>` may be _missing_ keys but not have them explicitly set to `undefined`.

```ts
/** Type matching stats to numerical stage amounts. */
type StatRecord = Partial<Record<BattleStat, StatStage>>;

// This is unexpected
const foo: StatRecord = {
  [Stat.ATK]: undefined;
  [Stat.DEF]: undefined;
  [Stat.SPATK]: undefined;
};

// picture this being inside StatStageChangePhase/etc
function queueStatChanges(statRecord: StatRecord) {
  for (const [stat, stage] of statRecord.entries()) {
    // we have to check for `undefined` here since TS doesn't distinguish between "missing" and "present but undefined"
    if (stage == null) {
      continue;
    }

    // ...
  }
}
```

Enabled the compiler option and added `| undefined` to ~~a few~~ around 100 or so callsites.

## Screenshots/Videos
N/A
## How to test the changes?
`pnpm exec tsgo --noEmit` should pass without errors.

## Checklist

### DIFF
--- a/src/data/abilities/ab-attrs.ts
+++ b/src/data/abilities/ab-attrs.ts
@@ -77,7 +77,7 @@ export interface AbAttrBaseParams {
    * Used to prevent message flyouts and other effects from being triggered.
    * @defaultValue `false`
    */
-  readonly simulated?: boolean;
+  readonly simulated?: boolean | undefined;
 
   /**
    * (For callers of {@linkcode applyAbAttrs}): If provided, **only** apply ability attributes of the passive (true) or active (false).
@@ -88,7 +88,7 @@ export interface AbAttrBaseParams {
    * (For implementations of {@linkcode AbAttr}): This will *never* be undefined, and will be `true` if the ability being applied
    * is the pokemon's passive, and `false` otherwise.
    */
-  passive?: boolean;
+  passive?: boolean | undefined;
 }
 
 export interface AbAttrParamsWithCancel extends AbAttrBaseParams {
@@ -1817,8 +1817,8 @@ export abstract class PostAttackAbAttr extends AbAttr {
 }
 
 export class PostAttackStealHeldItemAbAttr extends PostAttackAbAttr {
-  private readonly stealCondition?: PokemonAttackCondition;
-  private stolenItem?: PokemonHeldItemModifier;
+  private readonly stealCondition?: PokemonAttackCondition | undefined;
+  private stolenItem?: PokemonHeldItemModifier | undefined;
 
   constructor(stealCondition?: PokemonAttackCondition) {
     super();
@@ -1964,7 +1964,7 @@ export class PostAttackApplyBattlerTagAbAttr extends PostAttackAbAttr {
 
 export class PostDefendStealHeldItemAbAttr extends PostDefendAbAttr {
   private readonly condition?: PokemonDefendCondition;
-  private stolenItem?: PokemonHeldItemModifier;
+  private stolenItem?: PokemonHeldItemModifier | undefined;
 
   constructor(condition?: PokemonDefendCondition) {
     super();
@@ -2016,7 +2016,7 @@ export class PostDefendStealHeldItemAbAttr extends PostDefendAbAttr {
  */
 export interface PostSetStatusAbAttrParams extends AbAttrBaseParams {
   /** The pokemon that set the status condition, or `undefined` if not set by a pokemon */
-  sourcePokemon?: Pokemon;
+  sourcePokemon?: Pokemon | undefined;
   /** The status effect that was set */
   effect: StatusEffect;
 }
@@ -2309,8 +2309,8 @@ export class PostSummonRemoveArenaTagAbAttr extends PostSummonAbAttr {
 export class PostSummonAddArenaTagAbAttr extends PostSummonAbAttr {
   private readonly tagType: ArenaTagType;
   private readonly turnCount: number;
-  private readonly side?: ArenaTagSide;
-  private readonly quiet?: boolean;
+  private readonly side?: ArenaTagSide | undefined;
+  private readonly quiet?: boolean | undefined;
   // TODO: This should not need to track the source ID in a tempvar
   private sourceId: number;
 
@@ -3078,7 +3078,7 @@ export interface PreStatStageChangeAbAttrParams extends AbAttrBaseParams {
    * @remarks
    * Currently, only used by {@linkcode ReflectStatStageChangeAbAttr} in order to reflect the stat stage change
    */
-  source?: Pokemon;
+  source?: Pokemon | undefined;
   /** Holder that will be set to true if the stat stage change should be cancelled due to the ability */
   cancelled: BooleanHolder;
 }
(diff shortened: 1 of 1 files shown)


src/data/arena-tag.ts
```diff
@@ -150,7 +150,14 @@ export abstract class ArenaTag implements BaseArenaTag {
...
     this.maxDuration = turnCount;
-    this.sourceMove = sourceMove;
+    // TODO: Rework tags passing `MoveId.NONE` to instead pass `undefined` for consistency
+    // TODO: Enforce that all arena tags explicitly declare any used properties to ensure only required properties are serialized
+    if (sourceMove) {
+      this.sourceMove = sourceMove;
+    }
+    if (sourceId !== undefined) {
+      this.sourceId = sourceId;
+    }
```
DayKev, line 160: Hm, it's `sourceMove?: MoveId;` but `sourceId: number | undefined;`, that should also probably be made consistent.

src/data/mystery-encounters/mystery-encounter-option.ts
```diff
@@ -69,6 +69,8 @@ export class MysteryEncounterOption implements IMysteryEncounterOption {
     this.requirements = this.requirements ?? [];
     this.primaryPokemonRequirements = this.primaryPokemonRequirements ?? [];
     this.secondaryPokemonRequirements = this.secondaryPokemonRequirements ?? [];
+    this.primaryPokemon = undefined;
+    this.secondaryPokemon = undefined;
```
DayKev, line 73: Can't all of this be replaced with initializers? `public hasDexProgress = false;`, `public primaryPokemon: PlayerPokemon | undefined = undefined;` etc.

src/data/mystery-encounters/mystery-encounter.ts
```diff
@@ -301,6 +301,8 @@ export class MysteryEncounter implements IMysteryEncounter {
     this.dialogueTokens = {};
     this.enemyPartyConfigs = [];
     this.startOfBattleEffects = [];
+    this.primaryPokemon = undefined;
+    this.secondaryPokemon = undefined;
     this.introVisuals = undefined;
     this.misc = null;
     this.expMultiplier = 1;
```
DayKev, line 308: Same with all of this.

src/phases/common-anim-phase.ts
```diff
@@ -9,8 +9,10 @@ export class CommonAnimPhase extends PokemonPhase {
   // we need to allow phaseName to be a union of the two
   public readonly phaseName: "CommonAnimPhase" | "PokemonHealPhase" | "WeatherEffectPhase" = "CommonAnimPhase";
   private anim: CommonAnim | null;
-  private targetIndex?: BattlerIndex;
+  private targetIndex?: BattlerIndex | undefined;
 
+  // TODO: Why can common anim be null?
```
DayKev, line 14: Who knows, it's probably safe to change that since it's not allowed to be `null` at pkty.

src/phases/message-phase.ts
```diff
@@ -4,10 +4,11 @@ import { Phase } from "#app/phase";
 export class MessagePhase extends Phase {
   public readonly phaseName = "MessagePhase";
   private text: string;
-  private callbackDelay?: number | null;
-  private prompt?: boolean | null;
-  private promptDelay?: number | null;
-  private speaker?: string;
+  // TODO: Remove null from signatures
```
DayKev, line 7: Should also change it to an object param (and `showText`/etc). cf https://github.com/Despair-Games/poketernity/pull/1272 lol (though I actually missed `MessagePhase` in that PR, oops)

src/phases/party-exp-phase.ts
```diff
@@ -8,9 +8,10 @@ import { Phase } from "#app/phase";
 export class PartyExpPhase extends Phase {
   public readonly phaseName = "PartyExpPhase";
   expValue: number;
-  useWaveIndexMultiplier?: boolean;
-  pokemonParticipantIds?: Set<number>;
+  useWaveIndexMultiplier: boolean | undefined;
+  pokemonParticipantIds: Set<number> | undefined;
 
+  // TODO: Fix this
```
DayKev: What's broken about it?

src/plugins/api/api-base.ts
```diff
@@ -1,8 +1,11 @@
 import { SESSION_ID_COOKIE_NAME } from "#app/constants";
 import { getCookie } from "#utils/cookies";
+import type { SetRequired, UndefinedOnPartialDeep } from "type-fest";
 
 type DataType = "json" | "form-urlencoded";
 
+interface doFetchType extends SetRequired<UndefinedOnPartialDeep<RequestInit>, "method"> {}
```
DayKev:
```suggestion
interface DoFetchConfig extends SetRequired<UndefinedOnPartialDeep<RequestInit>, "method"> {}
```

test/test-utils/helpers/prompt-handler.ts
```diff
@@ -55,8 +55,11 @@ export class PromptHandler extends GameManagerHelper {
   /** The original `setModeInternal` function, stored for use in {@linkcode setMode}. */
   private readonly originalSetModeInternal: (typeof this.game.scene.ui)["setModeInternal"];
 
-  /** A {@linkcode NodeJS.Timeout | Timeout} containing an interval used to check prompts. */
-  public static runInterval?: NodeJS.Timeout;
+  /**
+   * A {@linkcode NodeJS.Timeout | Timeout} containing an interval used to check prompts.
+   * Is unused during test execution itself, as pre- and post-test hooks automatically set and clear it to avoid deadlocks.
```
DayKev, line 61:
```suggestion
   *
   * Is unused during test execution itself, as pre- and post-test hooks automatically set and clear it to avoid deadlocks.
```

src/data/mystery-encounters/mystery-encounter-option.ts
Bertie690: probably, but this can be a later thing

src/data/mystery-encounters/mystery-encounter.ts
Bertie690: see above - it probably can but I'd put it off for a larger overhaul/cleanup of ME code at large

src/phases/common-anim-phase.ts
Bertie690: again - separate day kinda thang

src/plugins/api/api-base.ts
```diff
@@ -27,42 +30,50 @@ export abstract class ApiBase {
...
-
     return await this.doFetch(path, { method: "POST", body, headers });
   }
 
   /**
    * A generic request helper.
-   * @param path The path to send the request to.
-   * @param config The request {@linkcode RequestInit | Configuration}.
+   * @param path - The path to send the request to
+   * @param config - The request {@linkcode doFetchType | configuration}
```
Bertie690:
```suggestion
   * @param config - The request {@linkcode DoFetchConfig | configuration}
```

src/data/arena-tag.ts
Bertie690: in function arguments, the former IS the latter
`exactOptionalPropertyTypes` only enforces special `?` behavior for reading from object/class properties.

(IIRC there's an ESlint rule to avoid unneeded `| undefined`s, so might be worth a look.)

DayKev: I mean in the class properties, not the constructor parameters.

PR #6625, opened 2025-10-04
[Bug] Fix Heal Block inconsistencies + add restriction tests

Bertie690:
## What are the changes the user will see?
Heal block:
1. Restricts a list of moves (presumably) more in sync with mainline

## Why am I making these changes?
Was working on #5989 when I noticed heal block's messages were incorrect.
I also noted the entire move impl was incorrect lol

## What are the changes from a developer perspective?
Changed Heal Block's normal restriction method to a banlist instead of checking triage move status

Fixed approximately all the locales inconsistencies with heal block (locales PR WIP)

## Screenshots/Videos
TODO

## How to test the changes?
Use heal block and make sure nothing explodes

## Checklist
- [x] **I'm using `beta` as my base branch**
- [x] There is no overlap with another PR?
- [x] The PR is self-contained and cannot be split into smaller PRs?
- [x] Have I provided a clear explanation of the changes?
- [x] Have I tested the changes manually?
- [x] Are all unit tests still passing? (`pnpm test:silent`)
  - [x] Have I created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes?
- [x] Have I provided screenshots/videos of the changes (if applicable)?
  - [ ] Have I made sure that any UI change works for both UI themes (default and legacy)?

Are there any localization additions or changes? YES
- [ ] Has a locales PR been created on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repo?
  - [ ] If so, please leave a link to it here:
- [ ] Has the translation team been contacted for proofreading/translation?

PR #7007, opened 2026-01-27
misc: remove unused parameters from move restriction functions

Bertie690:
N/A
@DayKev noticed some holdovers inside #6625 after it was merged.

Removed an unused parameter from `getRestrictingTag` and `isMoveSelectable`, since those never were passed `target` parameters in the first place.

## Screenshots/Videos
N/A
## How to test the changes?
N/A
## Checklist
  - [x] The PR title matches the Conventional Commits format (as described in [CONTRIBUTING.md](../CONTRIBUTING.md#pr-title-format))

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -148,7 +148,11 @@ import { RibbonData } from "#system/ribbons/ribbon-data";
 import { awardRibbonsToSpeciesLine } from "#system/ribbons/ribbon-methods";
 import type { AbAttrMap, AbAttrString, TypeMultiplierAbAttrParams } from "#types/ability-types";
 import type { Constructor } from "#types/common";
-import type { GetAttackDamageParams, GetBaseDamageParams } from "#types/damage-params";
+import type {
+  GetAttackDamageParams,
+  GetAttackTypeEffectivenessParams,
+  GetBaseDamageParams,
+} from "#types/damage-params";
 import type { DamageCalculationResult, DamageResult } from "#types/damage-result";
 import type { LevelMoves } from "#types/pokemon-level-moves";
 import type { StarterDataEntry, StarterMoveset } from "#types/save-data";
@@ -1143,7 +1147,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
   }
 
   getSprite(): Phaser.GameObjects.Sprite {
-    return this.getAt(0) as Phaser.GameObjects.Sprite;
+    return this.getAt(0);
   }
 
   getTintSprite(): Phaser.GameObjects.Sprite | null {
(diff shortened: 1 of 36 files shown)


src/field/pokemon.ts
```diff
@@ -4358,22 +4358,32 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    */
   // TODO: Move this behavior into a matcher and expunge it from the codebase - we only use it for tests
   public hasRestrictingTag(moveId: MoveId): boolean {
-    return this.getRestrictingTag(moveId) !== null;
+    return this.getRestrictingTag(moveId) != null;
   }
 
   /**
-   * Determine whether the given move is selectable by this Pokemon and the message to display if it is not.
+   * Get the {@link MoveRestrictionBattlerTag} that is restricting this Pokemon's move usage, if one exists.
```
SirzBenjie:
```suggestion
   * Get the {@linkcode MoveRestrictionBattlerTag} that is restricting this Pokemon's move usage, if one exists.
```

src/field/pokemon.ts
```diff
@@ -4358,22 +4358,32 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+   * @param moveId - The ID of the move to check
+   * @returns The first tag on this Pokemon that restricts the move, or `undefined` if the move is not restricted.
    * @remarks
-   * Checks both the move's own restrictions and any restrictions imposed by battler tags like disable or throat chop.
-   *
+   * Does not consider target-based restrictions from Heal Block, which is done by {@linkcode getTargetRestrictingTag}.
+   */
+  private getRestrictingTag(moveId: MoveId): MoveRestrictionBattlerTag | undefined {
+    return this.findTag(
+      (t): t is MoveRestrictionBattlerTag => t instanceof MoveRestrictionBattlerTag && t.isMoveRestricted(moveId, this),
```
SirzBenjie:
```suggestion
      t => t instanceof MoveRestrictionBattlerTag && t.isMoveRestricted(moveId, this),
```

`t is MoveRestrictionBattlerTag)` conveys a bidirectional contract. That is, the method should return `true` whenever `t` is a `MoveRestrictionBattlerTag` and `false` if it is not. No additional logic or filtering should go on.

I presume the reason this was done was to avoid a typescript error? If so, we should be using a type assertion on the return type.

Bertie690: The way I see it (agree to disagree), `(x): x is T` does not _necessarily_ convey a bidirectional contract, merely that the forwards direction must hold.
Now, is it a good idea to make standalone identity functions (like the ones for the modifier rework) bidirectional? 100% - TS uses them for narrowing inside discriminated unions.
However, IMO requiring bidirectionality simply to use `x is T` inside what is effectively a call to `find` is restricting a perfectly good pattern for technical correctness.

Bertie690: And yes, TS is not smart enough to know that the `find` block is a type predicate without explicitly telling it as such. (A regular `instanceof` without anything else WOULD narrow, but adding anything else inside the conditional will remove it.)

src/field/pokemon.ts
```diff
@@ -4358,22 +4358,32 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    );
```
```suggestion
    return this.findTag(t => t instanceof MoveRestrictionBattlerTag && t.isMoveRestricted(moveId, this)) as
      | MoveRestrictionBattlerTag
      | undefined;
```
As per Benjie's suggestion (yes that's how Biome formatted it lol).

Bertie690: fine - i hate the extra type assertions but iy's w/e

PR #7013, opened 2026-01-31
refactor: clean up `challenge.ts`

DayKev:
N/A

The docs in the `challenge.ts` file were very unhelpful, and some variables were poorly named.

Comments that should have been TSDocs were turned into TSDocs.
Initializer values are now actually used as initializer values.
Variable names were improved.
Various TSDocs were improved.

## Checklist
  - [x] The PR title matches the Conventional Commits format (as described in [CONTRIBUTING.md](../CONTRIBUTING.md#pr-title-format))

### DIFF
--- a/.github/workflows/lint-pr-title.yml
+++ b/.github/workflows/lint-pr-title.yml
@@ -4,6 +4,8 @@ on:
     branches:
       - main
       - beta
+      - release
+      - "hotfix*"
     types: [opened, edited, reopened, synchronize]
 
 concurrency:
--- a/src/@types/move-target-set.ts
+++ b/src/@types/move-target-set.ts
@@ -1,0 +1,6 @@
+import type { BattlerIndex } from "#enums/battler-index";
+
+export interface MoveTargetSet {
+  targets: BattlerIndex[];
+  multiple: boolean;
+}
--- a/src/battle.ts
+++ b/src/battle.ts
@@ -6,7 +6,7 @@ import { BattleType } from "#enums/battle-type";
 import { BattlerIndex } from "#enums/battler-index";
 import { BiomeId } from "#enums/biome-id";
 import type { Command } from "#enums/command";
-import type { MoveId } from "#enums/move-id";
+import { MoveId } from "#enums/move-id";
 import { MysteryEncounterMode } from "#enums/mystery-encounter-mode";
 import type { MysteryEncounterType } from "#enums/mystery-encounter-type";
 import type { PokeballType } from "#enums/pokeball";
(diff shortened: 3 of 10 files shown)


Wlowscha (approved):

Wlowscha:
The item-related challenges will need to be fixed in the modifier rework as well (that might also automatically address the TODOs).

src/data/challenge.ts
```diff
@@ -147,7 +140,7 @@ export abstract class Challenge {
 
   /**
    * Decrease the value of the challenge
-   * @returns Returns true if the value changed
+   * @returns `true` if the value changed
```
Bertie690: IMO "Whether the value changed" sounds better (not to mention you yourself suggested it to me on another occasion)

src/data/challenge.ts
```diff
@@ -157,16 +150,14 @@ export abstract class Challenge {
     return false;
   }
 
-  /**
-   * Whether to allow choosing this challenge's severity.
-   */
+  /** Whether to allow choosing this challenge's severity. */
```
Bertie690: Should this be `@sealed`?

src/data/challenge.ts
```diff
@@ -188,268 +179,268 @@ export abstract class Challenge {
...
-  /**
-   * Gets the minimum difficulty added by this challenge.
-   * @returns The difficulty value.
-   */
+  /** @returns The minimum difficulty value of this challenge. */
   getMinDifficulty(): number {
     return 0;
   }
 
+  // biome-ignore-start lint/correctness/noUnusedFunctionParameters: pseudo-abstract methods
```
Bertie690, line 200:
```suggestion
  // TODO: Refactor the class hierarchy to remove the need for all these methods
  // biome-ignore-start lint/correctness/noUnusedFunctionParameters: pseudo-abstract methods
```

Bertie690 (changes requested):
Just a TODO for later - we should remove the boolean returns at some point.

src/data/challenge.ts
DayKev: Oh I missed that, I was too focused on the wording duplication.

src/data/challenge.ts
DayKev: Maybe many of these should?

DayKev:
> Just a TODO for later - we should remove the boolean returns at some point.

@Bertie690 From what? If you mean `#increaseValue`/etc then we can't, they're used by the UI.

Wlowscha (approved):

Bertie690:
From the apply methods i meant

PR #7017, opened 2026-02-01
fix(battle): Use min `Def` to prevent infinite dmg

fabske0:

It's no longer to deal infinite damage due to the enemy having `0` def. Instead it will always have 1

[Discord](https://discord.com/channels/1125469663833370665/1250836282926436413/1467593208069619956)

Added a `Math.max`  to `#getEffectiveStat` so it never falls below `1`

> Note that prior to [Generation III](https://m.bulbapedia.bulbagarden.net/wiki/Generation_III), no stat can fall below 1 or rise above 999; any further modifiers will be treated as if the stat was capped regardless of whether the stat is at −6 or +6 or not.

via [Bulbapedia](https://bulbapedia.bulbagarden.net/wiki/Stat_modifier)

## Screenshots/Videos

image
</details> 
image
</details> 

## How to test the changes?

```ts
const overrides = {
  ENEMY_HELD_ITEMS_OVERRIDE: [{ name: "SOUL_DEW", count: 10 }],
  ENEMY_NATURE_OVERRIDE: Nature.LONELY,
  MOVESET_OVERRIDE: [MoveId.TACKLE, MoveId.SCREECH]
```

Use `screech` 3 times then `tackle`

## Checklist

### DIFF
--- a/src/field/pokemon.ts
+++ b/src/field/pokemon.ts
@@ -1582,7 +1582,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       ret *= highestStatBoost.multiplier;
     }
 
-    return Math.floor(ret);
+    return Math.max(Math.floor(ret), 1);
   }
 
   calculateStats(): void {


Wlowscha (approved):

damocleas:
need to put this on 1.11.7 when we make it

EscaShark:
Shouldn't this floor at 1 instead? Pretty sure that's what happens in mainline, "0.25 Defense" doesn't exist there even in edge cases like this.

DayKev (dismissed):
I'll make a 1.11.7 branch to point this to

src/field/pokemon.ts
```diff
@@ -1578,7 +1578,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
       ret *= highestStatBoost.multiplier;
     }
 
-    return Math.floor(ret);
+    return Math.max(Math.floor(ret), 0.25);
```
```suggestion
    return Math.max(Math.floor(ret), 1);
```
I think we should stick with integer values here.

fabske0: yeah i guess

PR #7028, opened 2026-02-06
test: disable stacktrace spam in Vitest VSCode extension v1.40.0+

Bertie690:
N/A

Vitest's VS Code extension was updated in vitest-dev/vscode#720 to respect the current reporter's output settings, resulting in actual colored output (yay) alongside a copious amount of stack traces every other line (not yay).

Added a constant to `CustomDefaultReporter` to toggle stack traces.
For whatever reason, these never seemed to show up on terminal anyhow, so this really only affects IDEs for the most part.

## Screenshots/Videos
N/A

## How to test the changes?
Run `tackle.test.ts`
## Checklist

### DIFF
--- a/test/test-utils/reporters/custom-default-reporter.ts
+++ b/test/test-utils/reporters/custom-default-reporter.ts
@@ -5,13 +5,17 @@ import type { UserConsoleLog } from "vitest";
 import type { TestState } from "vitest/node";
 import { DefaultReporter } from "vitest/reporters";
 
+const IGNORE_STACKTRACE: boolean = true;
+
+const F_POINTER = "❯";
+
 /**
- * Custom Vitest reporter to strip the current file names from the output.
+ * Custom Vitest reporter to strip file name headers and `console.log` stack traces from logging output.
  */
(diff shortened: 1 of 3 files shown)


test/test-utils/reporters/custom-default-reporter.ts
```diff
@@ -28,14 +32,17 @@ export default class CustomDefaultReporter extends DefaultReporter {
...
 
-    write(log.content); // this is about the only changed line (that and us skipping a newline)
+    // this is the original version of the following line:
+    // write(c.gray(log.type + c.dim(` | ${headerText}\n`)) + log.content);
+    write(log.content);
 
-    if (!log.origin) {
+    if (!log.origin || IGNORE_STACKTRACE) {
       return;
     }
```
DayKev, line 41: Doesn't this mean the rest of this function will just never run? This is `if (true) { return; }`.

Bertie690: unless someone actively wants the stacktrace, yes.

For whatever reason, `origin` was not populated in terminal runs, so it preserves prior behavior for that stuff.

PR #7036, opened 2026-02-08
dev: add `ValueHolder` + subtyping support for `NumberHolder` et al.

Bertie690:
N/A

I wanted to add subtyping support for `NumberHolder` and company.
Along the way, I ended up creating a system for configurable error messages inside function parameters/etc. What joy!

Added a new class `ValueHolder<Value, Base>` which is fully generic on` Value` (to allow for widening to base types on creation) while still enforcing assignability checks via rest parameters inside the class constructor. 
This is helped by an assortment of type utilities inside `#types/error-types.ts`, which I plan on using for other purposes later.

Re-exported the `NumberHolder`/`BooleanHolder` aliases from `#utils/common` to avoid code churn

## Screenshots/Videos
N/A

## How to test the changes?
Check out the new typing test file (`test/types/valueHolder.test-d.ts`) that checks assignability quirks and vets the error messages produced when passing incompatible types.

Other than that, try creating `ValueHolder`s with things other than what they expect.

## Checklist
  - [x] I have created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes if necessary)


src/@types/error-type.ts
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Internal type helper to convert 2 arbitrary types (or unions thereof) into human-readable strings
+ * suitable for use in error messages and the like.
+ * @internal
+ */
+export type FormatType<T> =
+  // special cases to avoid splitting up into `'true' | 'false'` and `'string' | 'number'` | ...
+  IsEqual<Primitive, T> extends true
+    ? "a primitive"
+    : IsEqual<boolean, T> extends true
```
DayKev, line 24: `boolean` is already part of `Primitive`, can this ever be reached?

Bertie690: Yes - this checks if you pass in the union for primitive VERBATIM.
(`IsEqual<Primitive, boolean` is false`.)

src/@types/error-type.ts
```diff
@@ -0,0 +1,80 @@
...
+      >
+    : never;
+
+type AddPrefix<Str extends string> = `${Str extends `${"a" | "e" | "i" | "o" | "u"}${infer _}` ? "an" : "a"} ${Str}`;
+
+/**
+ * Internal type helper to convert a stringable value into a more descriptive string.
+ * @internal
+ */
+type PrintTypeInternal<T> =
```
SirzBenjie, line 57: Can you explain the purpose of this in more depth in the doc comment?

src/utils/common.ts
```diff
@@ -6,6 +6,10 @@ import type { Variant } from "#sprites/variant";
 import { toCamelCase } from "#utils/strings";
 import i18next from "i18next";
 
+// Re-export the value holder classes for compatibility with existing imports looking over here
```
SirzBenjie, line 9:
```suggestion
// Re-export the value holder classes for compatibility with existing imports looking over here
// TODO: Remove these re-exports and update associated imports
```

src/utils/value-holder.ts
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * The stored value.
+   * @sealed
+   */
+  public value: Value & Base;
+
+  /**
+   * Create a new `ValueHolder` with the provided initial value.
+   * @param value - The initial value to store in the holder
+   * @param MISMATCH - Dummy parameter used to prevent constructing the class with an invalid type for `value`
```
SirzBenjie, line 52: I've not seen this trick before. Can you explain?

test/types/value-holder.test-d.ts
```diff
@@ -0,0 +1,77 @@
...
+  it("should widen literal types to their underlying base types", () => {
+    const holder = new NumberHolder(1);
+    expectTypeOf(holder).toEqualTypeOf<NumberHolder<number>>();
+    expectTypeOf(holder).not.toEqualTypeOf<NumberHolder<1>>();
+    expectTypeOf(holder.value).toEqualTypeOf<number>();
+
+    holder.value = 2;
+    holder.value = 42;
+    holder.value = -1;
+    holder.value = 0;
```
SirzBenjie, line 17: Are these just making sure it doesn't error?

test/types/value-holder.test-d.ts
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  it("should not allow assigning incorrect values", () => {
+    const holder = new NumberHolder(1);
+    // @ts-expect-error should not allow assigning a string to a NumberHolder
```
SirzBenjie, line 22: Will this show up in vitest's reporter if it fails?

SirzBenjie (dismissed):
I just had some requests for clarification. Good PR, though!

src/utils/value-holder.ts
Bertie690: It's the exact same one I showed you a month ago, just cleaned up slightly for VSCode hover doc compatibility.

The key takeaway is that we physically cannot restrict the typing of `Val` in the class or its constructor (that would might break our base type inference), so the next best thing is to make the class unconstructable by tacking on a dummy parameter if the types don't match.

test/types/value-holder.test-d.ts
Bertie690: correct

test/types/value-holder.test-d.ts
Bertie690: No - it will type error instead.
-# we should probably configure vitest to skip these things entirely...

DayKev: Afaict this will show up as a type error when running typecheck if for some reason the `@ts-expect-error`s become invalid, Vitest won't see any difference though.

src/@types/error-type.ts
```diff
@@ -0,0 +1,83 @@
...
+        : // special case boolean
+          boolean extends T
+          ? "boolean"
+          : T extends (...args: never) => unknown
+            ? "function"
+            : T extends object
+              ? "object"
+              : T extends symbol
+                ? "symbol"
+                : string extends T
```
DayKev, line 73: Why are some of these `thing extends T` and others `T extends thing`?

Bertie690: `T extends string` = T can be a string (or a literal, or a union containing `string`)
`string extends T` = T must be `string` or a union containing `string` (but, since we distribute over unions in the code that uses this, this effectively requires that `T` match `string` exactly)

Wlowscha (approved):

PR #6516, opened 2025-09-08
refactor(ui): break up and clean up `StarterSelectUiHandler`

Wlowscha:
## What are the changes the user will see?
Reduced memory usage from the starter select screen.

## Why am I making these changes?
This PR has two main goals:
1) Functionally change the starter select screen to use 81 containers instead of one per each starter,
2) Refactor and split up the various methods in it to reduce complexity and hopefully make it easier to make changes in the future.

## What are the changes from a developer perspective?

(This is a continuation of #6390, see there for older commits).

List of bugs found by @fabske0 here: https://discord.com/channels/1125469663833370665/1412372243492634666/1414529712369176607
(they should all be fixed, hopefully).

This is by necessity a gigantic PR due to the size of the files involved.

Note that, both due to the size of the files involved and their UI nature, manual testing to find bugs is most likely the best approach to review this PR.

Many functions in the starter select ui handler and adjacent files have been updated to use `StarterSpeciedId` in their arguments. This prevents passing a generic `Species` or `SpeciesId` by mistake which may cause issues. In particular, this affects all functions which interact with starter data in the game data.

Cursors for ability, nature etc. which stored a copy of the preferences for the current species have been removed as they were redundant.

### Easy to digest changes

Renamed the `StarterAttributes` type to `StarterPreferences` to better convey its usage, and renamed the `StarterPreferences` interface to `AllStarterPreferences`.

Introduced a `getSpeciesDefaultDexAttrProps` function in `game-data.ts` which returns, for a given species, default choices of preferences for shiny, gender and form based on caught data.

There are some changes to the Pokedex files, but they mostly consist of renaming variables or using some of the new utility functions.

Multiple functions and data structures have been moved from `StarterSelectUiHandler` to the new `starter-select-ui-utils.ts` file. In particular, the definition of `SpeciesDetails` is here, and is now the same in all pokedex files as well.
Language settings are also here, as well as various functions that convert `StarterPreferences` to different data formats (e.g. `SpeciesDetails` or `DexAttrProps`.
Functions such as `isPassiveAvailable` have also been moved here, although some are still wrapped in a helper method when used in the starter select ui class.

The `StarterContainer` now has a `setSpecies` method so that it can be updated (changing species) on the fly.

### Heavier changes

[...]

PR #7039, opened 2026-02-08
refactor(ui): Add strong typing to `UiHandler.show()`

Wlowscha:
N/A

Ui code is a long-standing thorn in Pokerogue's back. One of many features which make it difficult to modify or add new UI elements is the obscure typing of the arguments of the `UiHandler.show()` methods, which in many cases require to sift the code for an untyped list of `args` to be understood.

Many type definitions have been moved from individual `ui-handler.ts` files to the new `ui-types-ts`.

The new type `HandlerOf<M extends UiMode>` maps each `UiMode` to the corresponding ui handler.
This is used in `ui.ts` so that the signatures of the various `setMode` methods assign the correct type to `args`. The methods have been overloaded to deal with handlers whose `show()` method has no arguments, which now show properly at callsites.

A new file, `ui-handler-params.ts`, contains interface definitions for all UI handlers which require parameters.

In `ui.ts`, the `handlers` attribute which used to be an array of UI handler instances has been converted to a `Record`, although currently partial.  The other parameters have also been reorganized and more comments have been added.

Currently, the game runs without issue at first glance, although bugs surely abound. There are two prominent issues:
- some test utils files are broken;
- some functions in the mystery encounter utils file use `globalScene.ui.setMode(modeToSetOnExit)` after previously calling `const modeToSetOnExit = globalScene.ui.getMode();` which is an issue because the mode is then not well determined and it is unclear what parameters to pass. Presumably `modeToSetOnExit` is always the same and this can be adjusted, but deeper understanding is required.

This PR has some overlap with #6516 and with the modifier refactor, but it should be relatively limited.

## Checklist

### DIFF
--- a/src/ui/handlers/pokedex-page-ui-handler.ts
+++ b/src/ui/handlers/pokedex-page-ui-handler.ts
@@ -1771,7 +1771,7 @@ export class PokedexPageUiHandler extends MessageUiHandler {
                             ui.setMode(UiMode.POKEDEX_PAGE, {
                               species: newSpecies,
                               savedStarterAttributes: this.savedStarterAttributes,
-                              filteredIndices: this.filteredIndices ?? [],
+                              filteredIndices: this.filteredIndices ?? undefined,
                             });
                             return true;
                           },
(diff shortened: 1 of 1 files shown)


src/data/mystery-encounters/utils/encounter-pokemon-utils.ts
```diff
@@ -740,45 +737,41 @@ export async function catchPokemon(
...
+                    });
                   },
-                  () => {
-                    globalScene.ui.setMode(
-                      UiMode.PARTY,
-                      PartyUiMode.RELEASE,
-                      0,
-                      (slotIndex: number, _option: PartyOption) => {
+                  //TODO: are these options wrong?
+                  onYes: () => {
```
Bertie690, line 770: IMO these should be named `onConfirm` and `onDeny` instead of `yes/no`

PR #7040, opened 2026-02-09
feat(ability): fix various Dancer mainline inconsistencies

Bertie690:
Dancer will:
- No longer trigger on moves that did nothing (protected, missed, or otherwise failed).
- Trigger at the correct time in battle
- Target the correct Pokemon when targeting an ally in double battles

Dancer had a few bugs that I documented a while back, so this is me tying up loose ends.

Added a new Phase, `DancerPhase` that triggers Dancer (queued at the end of the `MoveEffectPhase` to ensure proper ordering of effects).

Made sure Dancer only triggers if the move hit a target successfully

## Screenshots/Videos

https://github.com/user-attachments/assets/83486bba-49ea-4a0a-8678-2d2a8683a97b

## How to test the changes?
`pnpm test dancer.test.ts` (10+ new tests)
## Checklist


src/phases/move-effect-phase.ts
```diff
@@ -884,9 +884,42 @@ export class MoveEffectPhase extends PokemonPhase {
...
+      source: user,
+      targets,
+    };
+    for (const pokemon of globalScene.getField(true)) {
+      if (pokemon === user) {
+        continue;
+      }
+
+      // Avoid creating unneeded phases if the ability in question cannot apply.
+      // This does duplicate the relevant checks slightly, but the overahead is likely minimal
```
Bertie690:
```suggestion
      // This does duplicate the relevant checks slightly, but the overhead is likely minimal
```

src/data/abilities/ab-attrs.ts
```diff
@@ -4455,80 +4456,88 @@ export class PostBiomeChangeTerrainChangeAbAttr extends PostBiomeChangeAbAttr {
...
-  canApply(_params: Closed<PostMoveUsedAbAttrParams>): boolean {
+/**
+ * Attribute to trigger effects after a move is used by a different Pokémon on the field.
+ * @remarks
+ * This will only trigger on successful, non-Dancer induced and non-reflected move uses, the checks for which are
+ * performed inside the {@linkcode MoveEffectPhase}.
+ */
+abstract class PostMoveUsedAbAttr extends AbAttr {
+  // biome-ignore lint/correctness/noUnusedFunctionParameters: psuedo-abstract method
+  public override canApply(params: Closed<PostMoveUsedAbAttrParams>): boolean {
```
emdeann, line 4479: Biome supports a prefix underscore rather than needing a comment, right?

emdeann (commented):
Generally looks good to me, maybe 1.12.1 though? I do want this in as it's been sitting a while, but I don't think it's worth delaying 1.12 for as it's not related to anything in that update

src/data/abilities/ab-attrs.ts
DayKev: This should let subclasses inherit the docs (probably? VSCode doesn't always do things intelligently...).

emdeann: I don't see why it would be different, the param is still on the method either way. Unless VS Code strips it from the hover docs if it has a prefix underscore? But like... there's no way it would do that right?

DayKev: The way hover docs work in VSCode is inconsistent at best, so tbh I don't know if they work when the param names don't match in the subclass method.

emdeann: Microsoft moment

Wlowscha (approved):

emdeann (approved):

PR #7053, opened 2026-02-13
balance(ability): Battle Bond uses the stat raising mode for fusions

DayKev:
Battle Bond will now activate for fusions unless one of the fusion species is Greninja, applying the stat stage boosts.

Requested by balance team member.

Removed `NoFusionAbilityAbAttr` from Battle Bond and added checks for fusions in the `.conditionalAttr`s.

## How to test the changes?
`pnpm test:silent battle-bond`

## Checklist

### DIFF
--- a/src/data/daily-seed/daily-seed-utils.ts
+++ b/src/data/daily-seed/daily-seed-utils.ts
@@ -1,5 +1,6 @@
 import { globalScene } from "#app/global-scene";
 import { isBeta, isDev } from "#constants/app-constants";
+import { Gender } from "#data/gender";
 import type { PokemonSpecies } from "#data/pokemon-species";
 import { AbilityId } from "#enums/ability-id";
 import { MoveId } from "#enums/move-id";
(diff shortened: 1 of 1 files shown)


damocleas:
what if it is already in battle bond form already? currently that is possible to fuse while it is in battle bond form as that is what many do in endless when using it

DayKev:
@damocleas Hm, previously that would lock it into Ash-Greninja form wouldn't it? Since the ability can no longer activate to do form changes if the Pokémon is fused. I think the only change from this would be that those Pokémon would now get the stat stage boosts; they'd still not form change after being fused. So I guess it would be a buff.

damocleas:
> @damocleas Hm, previously that would lock it into Ash-Greninja form wouldn't it? Since the ability can no longer activate to do form changes if the Pokémon is fused. I think the only change from this would be that those Pokémon would now get the stat stage boosts; they'd still not form change after being fused. So I guess it would be a buff.

wouldn't really want the buff for that since it's already so good and is used, I think fusions can live without the stat boost until we get the form change refactor next update

just have a note in the changelog

DayKev:
Sure.

PR #7078, opened 2026-02-17
refactor: add strong typing to `Object.keys` and related methods

Bertie690:
N/A

I had some changes from #6924 I wanted to pull out.

In essence, we had a lot of record-like objects that were being accessed via `Object.keys`/`Object.entries`. However, these functions solely returning `string` resulted in unclear `string` index signatures (which gave no information about _what_ keys were expected) and frequent type assertions when coercing stringified keys back to their original types.

Overloaded the types of `Object.keys` and `Object.entries` to allow for more specific return types when the input is a record-like object, alongside changing various ways of coercing stringified keys back to their original types to be more type-safe.

Cleaned up callsites that produced type errors due to the new overloads, alongside fixing some related code where necessary.

Converted some object types that previously used `string` indexes into records with proper key types. (This included removing some single-use type aliases like `BiomeTierPokemonPools`.)

> [!IMPORTANT]
> While these type overrides are technically slightly type-unsafe (as TypeScript's structural typing means that object types can be extended to add more properties), the places where these functions is often used (within object types that are intended to be nominal and fixed in shape) makes this not much of an issue in practice.
> I'd love to have a way to brand object types with information saying "i should not and cannot be extended off of", but for now this is the best we can do.

## Screenshots/Videos
N/A
## How to test the changes?
`pnpm exec tsgo`
## Checklist


Wlowscha:
There seem to be other changes which are undocumented, e.g. getting rid of `BiomeTierPokemonPools `.

Bertie690:
Updated

src/@types/type-helpers.ts
```diff
@@ -164,3 +182,15 @@ export type TupleRange<Min extends number, Max extends number, T> = IntClosedRan
...
+ * Internal type helper to forcibly prompt TypeScript's language service to prefer keeping a type opaque during hover expansion.
+ *
+ * Used for clarity of intent when preventing distributive conditional types from expanding into their full definitions
+ * while still allowing them to distribute over unions as normal.
+ * Is otherwise equivalent to {@linkcode NonNullable}.
+ * @internal
+ * @privateRemarks
+ * Any uses of this type should be double-checked to ensure that IDE hover tooltips are actually improved by its addition.
+ */
+export type PreventHoverExpansion<T> = T & {};
```
DayKev, line 157: Can you show an example where this matters? Also this sounds like you're making our types function differently just to work around a bug in tsgo (which is unfinished).

global.d.ts
```diff
@@ -1,5 +1,45 @@
+import type { Stringable, Stringify, Unstringify } from "#types/strings";
+import { ObjectValues, PreventHoverExpansion } from "#types/type-helpers";
```
DayKev:
```suggestion
import type { ObjectValues, PreventHoverExpansion } from "#types/type-helpers";
```
Why isn't Biome complaining about this... 🤔

src/@types/type-helpers.ts
Bertie690: This is the exact same behavior of TypeScript's language service for the regular TS compiler and isn't TSGo exclusive.
[TS Playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBACgThAbhAdsAEge2XAogDzAEMUBnAS0xQB4AVAPigF4paoAyKAbwF8BuAFACA9ACpRAqKKgBBKKEjyAFkWBQAxiSgAjaOqo5gEACZRyaTFCJRSwOOYDmOkK3AQAyuvthgAOkmiwgIQhJhwagrQ7naORNoANtAstvYoTgA+UCgArgC2unBQmdrkDuZqxZiYiVqZOfHxRVDZKMYQAGbmJoIi4gFQAMIGEOFW8m7ylimOUO1huf5SUAACkTBEcES5UO5QALSsStCRkxrDo+WW1tNpAUEhYGERE9GpZe0g1LshRq2kOzE0nFEowWPAkKgMNgRgRiGRKDRJFBhMIoAA5ABCAC4oAB1PQkFCYCJEJzAI5QMBwTDGbLqEw2QFOeLkIybRrASxUiCkEbIKCoPLjSCkAA0Vn+tF2xkwPKyxKsDUwAHczK0QjNMNoAFYQdTAf7K1lKeSkhwMyKkJEoqAACmVSnI6hNysw2Xipm0CCIAGtmrz-ph2lAAAYAeR1er8PogIFIIeEEGA6gAlItkaj3OZ6VBldANvn9eRkPEXAhctgZrQ3J5vGpBbl-u1qdtyXoaRBtEReVYUCBlUQ4+LyeR-qPZkRyHBSzpUKUUOnvvhfsZ-i0fUTlSgoAB+UMAEi47h4IagOJQELgAnoPTEEiWQxQhjGJ05jLes3mi2kqzc6022y7ActAUq+lj6E+IxqJcYw3A4dzBKEownK8jjkB8sIkG0xhfAKy6oKuAJvMCECglA5FLiua4oBuKrbnuIaHsep7npePRCHe-SPs+1xMuh5AWhMb6stRexwfxDL6C0bLEOE37olQYlMpSTwIoGwbWCcBZQIk7RqN2ZikPJv6QP+Ww7PshzHEJlgtMqmxgAhDxPMK0AAKpkHxHxfKCSKUQR-yoWk6GfOY7QjFAYZ4VRUAEOo8TZG0XxMiR4pwfQ9BIuRe5hllZ4WT8AUHlwYURW50VFXFCVJUFDipe+jj0CeeXZVAbktfl7jsTaADECBlFQkWRvqewIPEqiCSKvT3tIcgxi45KqBoWi6A1wWdCYxlUsWE0AEoQLkGw+laSy4AAjtkO2JGgpwhnBTQ5PkIwhsZ5QjCgRDxAhKEpQkEAANKxswRGxH9HCwNSkDhCAgMgLeqI9QR-UIkNuojWNE2mPobRCAIbTxdp7QtEWg25CAEbarDpDUP9FV-CDQJ-bD9C2pgOL-sA5CfdQ+36HAOH-eK66bigGUpji3oyigM61SFNP0AA2gAuoI+PjQgszE5zpPkzqVOYeqAt04RtUkczrPsxsnPc7zYQC0LNEi2LEsQEQUsy15IAG9h8vKz05C5I8yETFwACSaADPEmC8sYu0kOaPCzC2UAAESRHs4W2Cngg2mATrANkGsjoaxpQBWGvzVaavaRBtifmz3BIgrfrmFA4fAJH0cmHHaQQNQACM4r9wADMP9BK+eeQFAIPBCLXahzJgVP98DZMU1TtqLymGbNCgKOt+JHwiKi5EAHo7gI8-11TABMq+65TsakN7Jib1U282g8WEMi0CLH+RUBz4CCAA))

(This was originally more important when I used tagged literals to preserve enum members before realizing that would break index signatures.)

DayKev: Ah, I see. Interesting, I'm surprised the recently added variable type expansion doesn't already cover this in TS5.

global.d.ts
```diff
@@ -8,6 +48,31 @@ declare global {
...
+  interface NumberConstructor {
+    // NB: I would make this produce NaN if an invalid string is passed, but that just becomes `number`...
+    new <N extends number>(value: Stringify<N>): N;
+    <N extends number>(value: Stringify<N>): N;
+    parseInt<N extends number>(string: Stringify<N>, radix?: 10): N;
+  }
+
+  interface Number {
+    toString<T extends number>(this: T, radix?: 10): Stringify<T>;
+  }
```
DayKev, line 73:
```ts
const _test = Number("1") // 1
const test2: string = "3";
const _test3 = Number(test2) // number
const _test4 = Number(4) // number
const _test5 = Number("5").toString() // "5"
const _test6 = Number(test2).toString() // `${number}`
```
Interesting that this means `Number(4)` is `number` but `Number("4")` is `4`, lol.

DayKev (dismissed):

global.d.ts
Bertie690: @DayKev you want me to add that?

DayKev: It's not really necessary (there's no reason to ever do `Number(<number literal here>)`) but if you want to sure.

DayKev:
Okay apparently if you do that it breaks things
```ts
const _test = Number(Stat.HP) // : Stat.HP
const test2: Stat = Stat.HP;
const _test3 = Number(test2); // : Stat
```
I removed the commit, we don't need it anyway lol.

DayKev (dismissed):

DayKev:
Oh, I took a second look at the code that broke and it broke because it's jank. I'll fix it in a bit.

DayKev (dismissed):

src/battle-scene.ts
```diff
@@ -1131,14 +1129,11 @@ export class BattleScene extends SceneBase {
...
-    this.pokeballCounts = Object.fromEntries(
-      getEnumValues(PokeballType)
-        .filter(p => p <= PokeballType.MASTER_BALL)
-        .map(t => [t, 0]),
-    );
-    this.pokeballCounts[PokeballType.POKEBALL] += 5;
     if (Overrides.POKEBALL_OVERRIDE.active) {
       this.pokeballCounts = Overrides.POKEBALL_OVERRIDE.pokeballs;
+    } else {
+      this.pokeballCounts = Object.fromEntries(getEnumValues(PokeballType).map(t => [t, 0] as const));
```
DayKev: Doesn't removing the `filter` mean that there will now be a `PokeballType.LUXURY_BALL` key in the object where there wasn't before?

Edit: GitHub why are you so bad? Manually pasting the diff below since it won't show the selected range automatically in the comment's preview box:
```diff
-     getEnumValues(PokeballType)
-       .filter(p => p <= PokeballType.MASTER_BALL)
-       .map(t => [t, 0]),
-   );
-   this.pokeballCounts[PokeballType.POKEBALL] += 5;
    if (Overrides.POKEBALL_OVERRIDE.active) {
      this.pokeballCounts = Overrides.POKEBALL_OVERRIDE.pokeballs;
+   } else {
+     this.pokeballCounts = Object.fromEntries(getEnumValues(PokeballType).map(t => [t, 0] as const));
```

Bertie690: oh shit right

src/@types/biomes.ts
```diff
@@ -6,23 +6,22 @@ import type { TimeOfDay } from "#enums/time-of-day";
...
   timesOfDay: TimeOfDay[];
-};
+}
 
 export type CatchableSpecies = Readonly<Record<SpeciesId, readonly BiomeTierTimeOfDay[]>>;
 
 export type ArenaPokemonPools = Readonly<Record<BiomePoolTier, readonly SpeciesId[]>>;
 
-export type BiomePokemonPools = ReadonlyDeep<Record<BiomePoolTier, Record<TimeOfDay, SpeciesId[]>>>;
+export type BiomePokemonPools = Readonly<Record<BiomePoolTier, Readonly<Record<TimeOfDay, readonly SpeciesId[]>>>>;
```
DayKev, line 24: Why did you change this?

src/data/pokemon-species.ts
```diff
@@ -1094,7 +1091,7 @@ export class PokemonSpecies extends PokemonSpeciesForm implements Localizable {
       const levelDiff = player ? 0 : forTrainer || isBoss ? (forTrainer && isBoss ? 2.5 : 5) : 10;
       ret.push([prevolutionLevels[0][0], 1]);
       for (let l = 1; l < prevolutionLevels.length; l++) {
-        const evolution = pokemonEvolutions[prevolutionLevels[l - 1][0]].find(
+        const evolution = pokemonEvolutions[prevolutionLevels[l - 1][0]]!.find(
```
DayKev:
```suggestion
        const evolution = pokemonEvolutions[prevolutionLevels[l - 1][0]].find(
```
This evaluates to `any` now, so the `!` doesn't do anything.

src/data/pokemon-species.ts
```diff
@@ -1116,7 +1113,7 @@ export class PokemonSpecies extends PokemonSpeciesForm implements Localizable {
         ]); // TODO: are those bangs correct?
       }
       const lastPrevolutionLevel = ret[prevolutionLevels.length - 1][1];
-      const evolution = pokemonEvolutions[prevolutionLevels.at(-1)![0]].find(e => e.speciesId === this.speciesId);
+      const evolution = pokemonEvolutions[prevolutionLevels.at(-1)![0]]!.find(e => e.speciesId === this.speciesId);
```
DayKev:
```suggestion
      const evolution = pokemonEvolutions[prevolutionLevels.at(-1)![0]].find(e => e.speciesId === this.speciesId);
```
Same here.

src/field/arena.ts
```diff
@@ -500,10 +500,12 @@ export class Arena {
...
       return;
     }
-    this.pokemonPool = Object.entries(allBiomes.get(this.biomeId).pokemonPool).reduce(
+
+    const currBiome = allBiomes.get(this.biomeId);
+    this.pokemonPool = Object.entries(currBiome.pokemonPool).reduce(
       (acc, [tier, pool]) => {
-        // TODO: Remove type assertion after https://github.com/pagefaultgames/pokerogue/pull/7078 is merged
-        acc[tier as `${BiomePoolTier}`] = [...pool[TimeOfDay.ALL], ...pool[timeOfDay]];
+        tier satisfies `${BiomePoolTier}`;
```
DayKev, line 507: What's the point of this `satisfies`?

src/@types/biomes.ts
Bertie690: IIRC, it resulted in geuninely godawful IDE hover information (expanding directly into its implementation in certain places and the half dozen `type-fest` helper types that come with).

src/data/pokemon-species.ts
Bertie690: what - that is not good

src/field/arena.ts
Bertie690: Ensuring that `tier` is a stringified `BiomePoolTier` and can thus be used to index the object

src/@types/biomes.ts
DayKev: Ah, that's unfortunate.

src/field/arena.ts
DayKev: It's already typed as `Stringify<BiomePoolTier>` though?
image

Everything involved is strongly typed and statically defined. I don't think it's possible for this `satisfies` to ever not match unless there are other type errors that would be caught first.

Bertie690: Fair, though I do prefer to keep it in as extra insurance regardless.

src/data/pokemon-species.ts
```diff
@@ -1082,61 +1082,63 @@ export class PokemonSpecies extends PokemonSpeciesForm implements Localizable {
...
-          ),
-        ]); // TODO: are those bangs correct?
-      }
-      const lastPrevolutionLevel = ret[prevolutionLevels.length - 1][1];
-      const evolution = pokemonEvolutions[prevolutionLevels.at(-1)![0]].find(e => e.speciesId === this.speciesId);
+    const prevolutionLevels = this.getPrevolutionLevels().reverse();
+    const levelDiff = player ? 0 : forTrainer || isBoss ? (forTrainer && isBoss ? 2.5 : 5) : 10;
+    ret.push([prevolutionLevels[0][0], 1]);
+    for (let l = 1; l < prevolutionLevels.length; l++) {
+      const evolution = pokemonEvolutions[prevolutionLevels[l - 1][0]]!.find(
```
DayKev:
```suggestion
      const evolution = pokemonEvolutions[prevolutionLevels[l - 1][0] as keyof typeof pokemonEvolutions].find(
```

src/data/pokemon-species.ts
```diff
@@ -579,6 +579,9 @@ export abstract class PokemonSpeciesForm {
     const rootSpeciesId = this.getRootSpeciesId();
     for (const moveId of moveset) {
       if (Object.hasOwn(speciesEggMoves, rootSpeciesId)) {
+        // TODO: Review typing of `speciesEggMoves` - asserting `rootSpeciesId` is `keyof typeof speciesEggMoves` results in `never[]`
+        // due to incompatible tuple intersections
+
```
DayKev:
```suggestion
```

DayKev (dismissed):

PR #7088, opened 2026-02-19
fix(move): Expanding Force considers Terrain status at time of usage 

### LINKED ISSUE #4969: [Bug] Expanding Force does not Consider Terrain Status at Time of Usage
### Describe the bug

Expanding Force does not consider Terrain on the field at the time the move is used, but only at the beginning of the round.
This means that if Psychic Terrain is activated or deactivated before Expanding Force is used, its status as a single/multi-target move becomes incorrect.

-It seems that the animation for hitting multiple Pokémon with Expanding Force does not apply to the Pokémon in the righthand slot.

Link to Discord's Bug Report  (expanding force not working in psychic terrain)
https://discord.com/channels/1125469663833370665/1273795261977854014

### Reproduction

Use Expanding force with a slow Pokemon and make Psychic Terrain active on that turn.
-Switch in Psychic Surge, other Pokemon uses Expanding Force.

### Expected behavior

https://replay.pokemonshowdown.com/gen9doublescustomgame-2181534212 
Expanding Force should only consider Terrains when it is used

### Screenshots / Videos

Expanding Force hits one Target on the turn Psychic Terrain becomes Active

https://github.com/user-attachments/assets/e60ac79a-d22c-4cde-8629-01eb5c3dbc79

Expanding Force hits Two Targets on the turn Electric Terrain replaces Psychic Terrain

https://github.com/user-attachments/assets/66c603cf-b294-464d-ab98-3f22eda0917c

Rising Voltage is affected Correctly by Electric Terrain (It gains a boost in BP [Volt Switch showed 3. Rising Voltage showed 5])

https://github.com/user-attachments/assets/1eca6f5e-d933-409a-a5d6-1d080dae9443

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

YGGDRR:

Expanding Force now should properly target both enemies when Psychic-Terrain is introduced mid turn and user is Grounded, it will also only display single selection tageting in UI to match in the behaviour in he original games, this also allow targeting t be consistent when psychic terrain is removed mid turn (prior to this it defaulted to slot 1)

Fixes #4969

Rename VariableTargetAttr OverrideTargetAttr and let the old logic only aply to Tera-Starstorm (aplies eagerly at ui selection by a call to getMoveTargets).
VariableTargetAttr has its logic now handled directly by a method in move-phase so it can be aplied lazily and doesnt interact with the UI

## Screenshots/Videos

https://github.com/user-attachments/assets/119d1075-9f83-4fb1-b408-7f3aaafc4f00

## How to test the changes?

expanding-force.test.ts has been added to test it

## Checklist

### DIFF
--- a/src/data/moves/move.ts
+++ b/src/data/moves/move.ts
@@ -8862,34 +8862,49 @@ export class StatusIfBoostedAttr extends MoveEffectAttr {
   }
 }
 
-export class VariableTargetAttr extends MoveAttr {
-  private readonly targetChangeFunc: (user: Pokemon, target: Pokemon, move: Move) => number;
+type VariableMoveTargetFunc = (user: Pokemon, target: Pokemon, move: Move) => MoveTarget;
 
-  constructor(targetChange: (user: Pokemon, target: Pokemon, move: Move) => number) {
+/**
+ * Attribute to dynamically modify a move's targets during its execution.
+ *
+ * Used for Expanding Force.
+ * @see {@linkcode OverrideTargetAttr} - Attribute to modify move targets during target selection
+ */
+export class VariableTargetAttr extends MoveAttr {
+  private readonly targetChangeFunc: VariableMoveTargetFunc;
+
+  constructor(targetChangeFunc: VariableMoveTargetFunc) {
     super();
 
-    this.targetChangeFunc = targetChange;
+    this.targetChangeFunc = targetChangeFunc;
   }
 
-  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    const targetVal = args[0] as NumberHolder;
-    targetVal.value = this.targetChangeFunc(user, target, move);
+  apply(user: Pokemon, target: Pokemon, move: Move, args: [NumberHolder, ...any[]]): boolean {
+    const targetHolder = args[0];
+    targetHolder.value = this.targetChangeFunc(user, target, move);
     return true;
   }
 }
 
+/**
+ * Attribute to override the targets of a move during target selection and other related processes.
+ * Used by Tera Starstorm.
+ *
+ * @see {@linkcode VariableTargetAttr} - Attribute to dynamically modify a move targets mid-turn
+ */
+// TODO: Cleanup a lot of the move target code
 export class OverrideTargetAttr extends MoveAttr {
-  private readonly targetChangeFunc: (user: Pokemon, target: Pokemon, move: Move) => number;
+  private readonly targetChangeFunc: VariableMoveTargetFunc;
 
-  constructor(targetChange: (user: Pokemon, target: Pokemon, move: Move) => number) {
+  constructor(targetChangeFunc: VariableMoveTargetFunc) {
     super();
 
-    this.targetChangeFunc = targetChange;
+    this.targetChangeFunc = targetChangeFunc;
   }
 
-  apply(user: Pokemon, target: Pokemon, move: Move, args: any[]): boolean {
-    const targetVal = args[0] as NumberHolder;
-    targetVal.value = this.targetChangeFunc(user, target, move);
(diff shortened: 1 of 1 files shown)


src/phases/move-phase.ts
```diff
@@ -543,6 +546,31 @@ export class MovePhase extends PokemonPhase {
...
+  // recompute targets when VariableTargetAttr may change mid-turn
+  protected resolveChangeMoveTarget(): void {
+    const user = this.pokemon;
+    const moveObj = this.move.getMove();
+
+    if (!moveObj.hasAttr("VariableTargetAttr")) {
+      return;
+    }
+
+    const newTargetSet = getMoveTargets(user, moveObj.id);
```
Bertie690: Is it intentional that `getMoveTargets` still retains a call to `VariableMoveTargetAttr`?

src/phases/move-phase.ts
```diff
@@ -543,6 +546,31 @@ export class MovePhase extends PokemonPhase {
...
+
+    if (newTargetSet.multiple) {
+      // move has become a spread move – adopt the full set of targets.
+      this.targets = newTargetSet.targets;
+    } else {
+      // remain single–target. Preserve whatever the player already chose, but
+      // guard against the (rare) case where the chosen battler is no longer
+      // valid by falling back to the first available option.
+      const current = this.targets[0];
```
Bertie690, line 572: As discussed on Discord, this is inadequate and presents some issues.
EF on mainline should undergo target selection as a normal single-target move, with said selected target being used as a fallback if it no longer becomes a spread move at time of execution.
In current, removing Psychic terrain mid-turn will result in the move hitting both opponents due to the aforementioned faulty logic.

src/phases/move-phase.ts
```diff
@@ -543,6 +546,31 @@ export class MovePhase extends PokemonPhase {
     }
   }
 
+  // recompute targets when VariableTargetAttr may change mid-turn
+  protected resolveChangeMoveTarget(): void {
+    const user = this.pokemon;
+    const moveObj = this.move.getMove();
+
+    if (!moveObj.hasAttr("VariableTargetAttr")) {
```
Bertie690, line 554: This might be a bit of an ask, but it might make sense to rework `VariableTargetAttr` to perform post-processing on the resulting `BattlerIndex` array; this would make handing the target select interactions with EF vs. Tera Starstorm slightly easier while also paving the way for rampaging moves and/or Curse to use this (which are currently both suffering due to this discrepancy).

src/overrides.ts
```diff
@@ -51,7 +51,15 @@ import type { IntClosedRange, TupleOf } from "type-fest";
  * }
  * ```
  */
-const overrides = {} satisfies Partial<InstanceType<OverridesType>>;
+const overrides = {
```
Bertie690, line 54: These should not be committed.

(tbh we should probably make a script or something that does a regex check on the file to ensure no changes to the `overrides` constant are committed)

src/data/moves/move.ts
```diff
@@ -8878,6 +8878,22 @@ export class VariableTargetAttr extends MoveAttr {
   }
```
Bertie690, line 8886: Would this make more sense as modifying the `BattlerIndex`es rather than the move target sets?
That would both set it apart from the other attribute for Tera Starstorm as well as be more extensible for something like Curse (which currently requires its own move target enum value just to handle said targeting).

src/data/moves/move.ts
```diff
@@ -8878,6 +8878,22 @@ export class VariableTargetAttr extends MoveAttr {
...
 }
 
+export class OverrideTargetAttr extends MoveAttr {
```
Bertie690, line 8896: Would it make sense to swap the names?

(Pinging @DayKev for their opinion)

Bertie690 (changes requested):
All in all looks good, though the overrides issues need to be fixed no matter what.

PR #7096, opened 2026-02-21
test: ban uses of chai-style assertions

Bertie690:
N/A

We voted as a team to remove Chai-style assertions from our test code in favor of Jest-style assertions. We mostly use the latter anyways (with the chai style ones being few and far between), and fit nicer with custom `expect` matchers anyhow.

Overrode all 90-odd methods from Chai to return `never` instead of their original types.

Made a type test to double-check that all keys from `Chai.assertion` are effectively blacklisted from use.

Removed existing uses
## How to test the changes?
Look at `vitest.d.ts` and the new type test file (which can also be used to thoroughly type-test existing matchers to avoid future breakages).
Try to use a Chai assertion in the code and see that it results in a type error (which is good).

## Checklist
  - [x] I have made sure that any UI changes work for both the default and legacy UI themes (if applicable)


test/tests/test-framework/matcher-types-test-d.ts
DayKev, line 1: Oh I didn't notice the directory this was in before. This doesn't belong in `@types/`.

Bertie690: it's related to the matchers, and i don't feel good placing it with files that test actual ingame behavior.

as far as I am concerned, that file is an internal detail related to the matcher typing files and should be logically grouped with them.
While it could also go inside `matchers`, that moves it further away from the very thing it is testing (the types inside `vitest.d.ts`).

DayKev: Alright, I'm reorganizing this entire directory structure because I'm tired of tests being in random subdirectories and *nothing* but types should be in `@types/`.

DayKev (approved):

PR #7109, opened 2026-02-23
Change PYROAR ability from DROUGHT to SOUL_HEART

MateusGaioffatto:
Currently, Pyroar has the "Moxie" in which is not very useful to him, because he is primarily a special attack Pokémon (109 base Special Attack against 68 base Physical Attack). 

So, I thought "Soul Heart" would be good for him, as it works like a Moxie but for Special Attacks


damocleas:
We do not take balance changes without discussion and agreement in the public balance channels in the discord (discord.gg/pokerogue)

PR #7128, opened 2026-03-01
refactor: pressure isn't a dummy ability anymore

Bertie690:
N/A

Pressure was previously implemented by a dummy `AbAttr` that did nothing and was checked inside the move phase manually.

Also, it didn't have tests.
Reworked pressure into an (admittedly quite simple) ability to increment a PP number holder by 1.

Added a test file
## How to test the changes?
`pnpm test:silent pressure.test.ts`
## Checklist


src/phases/move-phase.ts
```diff
@@ -641,28 +641,16 @@ export class MovePhase extends PokemonPhase {
...
-  public getPpIncreaseFromPressure(targets: Pokemon[]): number {
-    const foesWithPressure = this.pokemon
-      .getOpponents(true)
-      .filter(opponent => targets.includes(opponent) && opponent.hasAbilityWithAttr("IncreasePpAbAttr"));
-    return foesWithPressure.length;
+    const { move, pokemon: user } = this;
+    const ppHolder = new NumberHolder(1);
+    this.getActiveTargetPokemon().forEach(target => {
+      applyAbAttrs("IncreasePpAbAttr", { pokemon: target, opponent: user, pp: ppHolder });
+    });
```
SirzBenjie, line 662: This will still decrease if the opponent is hidden, correct? (e.g. if they used fly/dig)?

test/@types/vitest.d.ts
```diff
@@ -375,6 +375,7 @@ interface PokemonMatchers {
    * Check whether a {@linkcode Pokemon} has applied a specific {@linkcode AbilityId}.
    * @param expectedAbilityId - The `AbilityId` that should have been applied
    */
+  // TODO: Naming this `toHaveAppliedAbility` would make more sense
```
SirzBenjie, line 391: This TODO should go near the definition of the method, not at a callsite.

test/matchers/to-have-used-pp.ts
```diff
@@ -42,7 +42,7 @@ export function toHaveUsedPP(
   const pkmName = getPokemonNameWithAffix(received);
   const moveStr = getEnumStr(MoveId, moveId);
 
-  const movesetMoves = received.getMoveset().filter(pm => pm.moveId === moveId);
+  const movesetMoves = received.getMoveset().filter(pm => this.equals(pm.moveId, moveId, this.customTesters));
```
SirzBenjie, line 45: Wait what? What is `this` here?

src/phases/move-phase.ts
Bertie690: I'd like to say it would, though we'd want to verify it regardless (and is semi-irrelevant to the current discussion due to the behaviour being literally identical to what it was prior).

test/@types/vitest.d.ts
Bertie690: Putting it here makws more sense since "go to definition" targets here vs the actual file with the function definition.

Either way, I plan on fixing this TODO relatively soon.

test/matchers/to-have-used-pp.ts
Bertie690: `MatcherState`.

TS would warn (and a bunch of tests would actively explode) if this ended up being `undefined`.

src/data/abilities/ab-attrs.ts
```diff
@@ -5066,8 +5066,35 @@ export class FlinchStatStageChangeAbAttr extends FlinchEffectAbAttr {
   }
 }
 
+export interface IncreasePpAbAttrParams extends Omit<AugmentMoveInteractionAbAttrParams, "move"> {
```
macwellbz: Better way than adding the `Omit` type could be to make `move` an optional param in `AugmentMoveInteractionAbAttrParams`.

Though this way might also be a bit less error-prone for new changes.

src/data/moves/pokemon-move.ts
macwellbz, line 97: lmao I think it's fine

src/data/moves/pokemon-move.ts
```diff
@@ -22,6 +22,7 @@ import i18next from "i18next";
  */
 export class PokemonMove {
   public moveId: MoveId;
+  // TODO: Consider making this readonly outside the class (since we provide functions to alter it already)
```
macwellbz, line 31: I think keeping it writable is fine as the `usePp()` function is kinda specific and having to account for current PP count in order to set the specific value you want could be a bit tedious.

src/phases/move-phase.ts
macwellbz: I checked out `getActiveTargetPokemon()` and I'm pretty sure it does as the opponents being hidden doesn't remove them from the `_targets` list.

src/data/abilities/ab-attrs.ts
Bertie690: The issue with marking `move` as optional is that `AugmentMoveInteractionAbAttrParams` should ALWAYS be supplied a move when it is called (since things like Parental Bond and similar need to know what move is used for compatibility checks).

I suppose you could move the shared properties to a common interface, but I don't think it's strictly necessary for just 2 interfaces. (3 maybe, but 2 nah).

src/data/moves/pokemon-move.ts
Bertie690: It is minor but the only real reason to _not_ change it is the modifier rework (which is basically holding up any major codebase-wide refactors for fear of massive merge conflicts).

src/data/moves/pokemon-move.ts
Bertie690: The thing is, there are exactly 3 scenarios where one needs to adjust PP counts:
1. Dock a fixed amount of PP (Spite, normal consumption, etc.)
2. Restore a fixed amount of PP (Leppa Berry, healing items)
3. Drain or restore all PP (Lunar Dance, Grudge, healing items, etc.)

All of these can be handled easily by `usePP` (or an analogous helper to restore PP), and setting a move's PP to a specific fixed value is (to my knowledge) limited to a single mystery encounter.

src/phases/move-phase.ts
Bertie690: Good to know, though we should verify on cart or showdown in any case.

src/data/moves/pokemon-move.ts
macwellbz: I assume you can't change it now because of the other refactor PRs right?

Bertie690: Correct - I'd rather not risk any more issues with 1.12's items rework so I'm just marking it as a "later" sort of thing.

src/@types/save-data.ts
Fontbane, line 1: Unrelated change, was this included on accident?

Bertie690: Oopsq

Wlowscha:
Could we change the name from `IncreasePpAbAttr` to something like `IncreaseUsedPpAbAttr`? The current name seems to indicate that the PP are increased, when in fact the opposite is happening...

Bertie690:
@Wlowscha done

src/data/abilities/ab-attrs.ts
```diff
@@ -5066,8 +5066,35 @@ export class FlinchStatStageChangeAbAttr extends FlinchEffectAbAttr {
...
-  private declare readonly _: never;
+export interface IncreasePpUsedAbAttrParams extends Omit<AugmentMoveInteractionAbAttrParams, "move"> {
+  /** Holder for the amount of PP that will be consumed; can be modified by ability application */
+  readonly pp: NumberHolder;
+}
+
+/**
+ * Attribute for abilities that increase the PP consumption of received attacks.
+ *
+ * Used by {@link https://bulbapedia.bulbagarden.net/wiki/Pressure_(Ability) | Pressure}
```
DayKev:
```suggestion
 * @see {@link https://bulbapedia.bulbagarden.net/wiki/Pressure_(Ability)}
```

src/data/abilities/ab-attrs.ts
```diff
@@ -5066,8 +5066,35 @@ export class FlinchStatStageChangeAbAttr extends FlinchEffectAbAttr {
...
+ */
+export class IncreasePpUsedAbAttr extends AbAttr {
+  /**
+   * The amount of PP to increase.
+   * @defaultValue `1`
+   */
+  private readonly ppIncrease: number;
+
+  constructor(ppIncrease = 1) {
+    super();
```
DayKev, line 5122:
```suggestion
    super();

```

DayKev (dismissed):

DayKev (approved):

Wlowscha (approved):

PR #7136, opened 2026-03-04
refactor: clean up & document game speed-related code

Bertie690:
N/A

The game speed code was not well documented, and also specified many lengthy types that could (and should) be inferred.

I was particularly motivated by `mutateProperties`, which had a parameter type of `any` and a biome ignore that brushed a larger issue aside.

Added documentation to various functions and methods used by the game speed code, removing type specifications where TS was able to infer them.

Changed the scene parameter to be passed in directly instead of via `this`

Altered `obj` inside `mutateProperties` to be `object` for slightly more safety, and moved it out of the enclosing function.

## Screenshots/Videos
N/A
## How to test the changes?
N/A

## Checklist


src/system/game-speed.ts
```diff
@@ -1,112 +1,118 @@
...
-    // We do not mutate Tweens or TweenChain objects themselves.
+  /**
+   * Recursively mutate an object's duration-related properties.
+   * @param obj - The object to mutate
+   * @param allowArray - (Default `false`) Whether to allow mutating arrays of tween configs at the top level.
+   */
+  const mutateProperties: {
+    <O extends object>(obj: O, ...allowArray: O extends readonly any[] ? [true] : []): void;
+    (obj: object, allowArray?: boolean): void;
+    // biome-ignore lint/complexity/noExcessiveCognitiveComplexity: This goes down from 20 to 13 if moved outside of `initGameSpeed` (and is only here to access a single function)
```
DayKev: Then why didn't you move this and `transformValue` outside of `initGameSpeed`?

src/system/game-speed.ts
```diff
@@ -1,112 +1,119 @@
...
-  const originalAddEvent: typeof Phaser.Time.Clock.prototype.addEvent = this.time.addEvent;
-  this.time.addEvent = function (config: Phaser.Time.TimerEvent | Phaser.Types.Time.TimerEventConfig) {
+  // #region Method overrides
+
+  const originalAddEvent = this.time.addEvent;
+  this.time.addEvent = function (this: Phaser.Time.Clock, config) {
     if (!(config instanceof Phaser.Time.TimerEvent) && config.delay) {
       config.delay = transformValue(config.delay);
     }
     return originalAddEvent.apply(this, [config]);
```
DayKev:
```suggestion
    return originalAddEvent.call(this, config);
```
You didn't change this to `.call` like the others.

src/system/game-speed.ts
Bertie690: I can't - relies on `transformValiue` that relies on `this.gameSpeed`

Bertie690: The alternative is moving them both out and either adding another parameter or using `Function.bind`.
(In hindsight, I can just have the function pass the scene as a normal parameter to avoid `this` scoping issues)

DayKev (dismissed):

src/system/game-speed.ts
```diff
@@ -1,112 +1,119 @@
...
+/**
+ * Override and wrap various Phaser methods to alter time-related properties based on the current game speed.
+ * Any duration values passed that are {@linkcode FixedInt}s will be treated as fixed duration values and preserved.
+ * @privateRemarks
+ * While this may appear somewhat heavy-handed, there is effectively no other way to achieve the desired effect within the constraints set by Phaser,
+ * as altering game speed would affect _all_ time-related processes (including ones we want to remain unchanged).
+ */
+export function initGameSpeed(this: BattleScene): void {
+  /** Mutate a duration value based on the current speed. */
+  const transformValue = (duration: number | FixedInt): number => {
```
DayKev: Btw changing this to `duration` isn't correct, sometimes it's a delay as well.

Bertie690: Will fix

Bertie690: It's still better than `value` TBF...

src/system/game-speed.ts
```diff
@@ -1,112 +1,176 @@
...
+  ): Phaser.Sound.BaseSound => {
+    if (typeof args[1] === "number") {
+      // no scene included; duration is at index 1
+      (args[1] as number) = applyGameSpeedMult(scene, args[1] as number);
+    } else {
+      // scene included in parameter list; duration is at index 2
+      (args[2] as number) = applyGameSpeedMult(scene, args[2] as number);
+    }
+    return originalFadeOut(...(args as Parameters<typeof originalFadeOut>));
+  }) satisfies typeof originalFadeOut;
```
DayKev, line 146: I'm not sure why you're doing weird things with this function, the scene needs to be passed as the first parameter to the original `fadeOut`/`fadeIn` functions and that's what was done before (it just ignored the `scene` passed in at callsites and always passed `globalScene` to the original, which it should probably pass in the `scene` param like normal rather than override with `globalScene`).
Also `duration` should be typed as `number | FixedInt` for both of these.

Bertie690: @DayKev The function has an overload that lacks a scene parameter.
We previously did not account for said overload during the method wrapping.

DayKev (approved):

emdeann (approved):

PR #7142, opened 2026-03-07
refactor(ui): change placement of option labels in challenge select

DayKev:
The option labels and left arrow are dynamically placed based on the actual length of the option text, rather than fixed based on the option with the longest text, and the padding between the right arrow and the border was set to a fixed distance for all languages (generally results in a reduction).

The current method was causing unnecessary text overlapping.

- The right arrow is now fixed relative to the container border instead of dynamic based on the longest option text.
- The left arrow is dynamic, placed a fixed spacing to the left of the option text.
- The option text is right-aligned within the left and right arrows.
- The padding between the right arrow and the container border is fixed.
- A console warning is added to notify if there is potential text overlap due to a challenge name + option text combination being too long.

## Screenshots/Videos

image

</details>

image

</details>

## Checklist

### DIFF
--- a/src/@types/biomes.ts
+++ b/src/@types/biomes.ts
@@ -51,8 +51,6 @@ export interface Biome {
   readonly terrainPool: TerrainPool;
   /** A string representing the bgm of the biome. Only needed if the bgm name doesn't match the biome name. */
   readonly bgm?: string;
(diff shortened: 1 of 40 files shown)


SirzBenjie:
It looks better before, but yeah...

PR #7145, opened 2026-03-09
refactor: clean up and denest `AttemptCapturePhase`

Bertie690:
The capture formula will not round intermediate results during computation. Due to this, the values of certain variables may vary by ~1% due to rounding

`AttemptCapturePhase` was an absolute nightmare, with 15+ nesting levels' worth of tween chain-wrapped UI-handler callback code.

Also, it was virtually untestable (and unreadable) due to the extremely convoluted way in which the capture formula was computed. What fun!

Rewrote the file from scratch, moving lots of code into private methods rather than nested callbacks.
Among other things, the shake check count is now pre-determined ahead of time instead of computed on the fly in the middle of the tween chain, greatly simplifying logic and easing the ability for subclasses to override the formula logic.

Removed the unused `PokeballType.LUXURY_BALL` to ease this and remove unused clutter. (Once we do get more balls in 1.13+, we can actually add them in dex order instead of some arbitrary listing.)

## Screenshots/Videos

https://github.com/user-attachments/assets/08d19015-2728-475e-9a26-8ba6f19a52ac

</details>

https://github.com/user-attachments/assets/4a8c1563-a1ae-4407-9121-6c1e7b702ce3

</details>

NB: I was not able to actively record footage of a failed catch attempt, but verified off-camera that it indeed works.

## How to test the changes?
`pnpm test capture-phase.test.ts payback.test.ts`.
Former has had a 7-month TODO test properly filled in, while the latter demonstrates the new capture formula being actually testable instead of **not**.

## Checklist


src/data/pokeball.ts
```diff
@@ -17,36 +17,33 @@ export function getPokeballAtlasKey(type: PokeballType): string {
...
     case PokeballType.MASTER_BALL:
       return "mb";
-    case PokeballType.LUXURY_BALL:
-      return "lb";
+    // case PokeballType.LUXURY_BALL:
+    //   return "lb";
   }
 }
 
 export function getPokeballName(type: PokeballType): string {
```
DayKev, line 24: This would be simpler:
```ts
if (pokeballType === PokeballType.POKEBALL) {
  return i18next.t("pokeball:pokeBall");
}
return i18next.t(`pokeball:${camelCase(pokeballType)}`);
```

Or you could rename the first enum member to `POKE_BALL` and then get rid of the need for the `if`.

src/data/pokeball.ts
```diff
@@ -59,8 +56,8 @@ export function getPokeballCatchMultiplier(type: PokeballType): number {
       return 3;
     case PokeballType.MASTER_BALL:
       return -1;
-    case PokeballType.LUXURY_BALL:
-      return 1;
+    // case PokeballType.LUXURY_BALL:
+    //   return 1;
```
DayKev:
```suggestion
```
This isn't necessary, just remove it. We use Git for a reason.
Same with every other spot like this.

src/utils/time.ts
```diff
@@ -0,0 +1,21 @@
+import { globalScene } from "#app/global-scene";
+import type { SceneBase } from "#app/scene-base";
+
+/**
+ * Create a Timer event and wait for it to conclude.
+ * @param delay - The delay (in milliseconds) to wait for
```
DayKev:
```suggestion
 * @param delay - The delay (in milliseconds) to wait for; will be modified by game speed.
```

test/tests/phases/capture-phase.test.ts
```diff
@@ -25,9 +27,38 @@ describe("Capture Phase", () => {
       .enemyAbility(AbilityId.BALL_FETCH)
       .enemyMoveset(MoveId.SPLASH)
       .startingLevel(100)
-      .enemyLevel(100);
+      .enemyLevel(100)
+      .startingModifier([{ name: "MASTER_BALL", count: 1 }]);
```
DayKev, line 31: This works? You're supposed to use the Pokeball override.

src/battle-scene.ts
```diff
@@ -806,6 +807,23 @@ export class BattleScene extends SceneBase {
...
+   * @remarks
+   * This function is allowed to return inactive (i.e. fainted) Pokemon.
+   */
+  // TODO: Replace prior indexing into `getField` with this abstraction to make an overhaul of this system easier
+  public getPokemonByBattlerIndex(battlerIndex: BattlerIndex.ENEMY | BattlerIndex.ENEMY_2): EnemyPokemon | undefined;
+  public getPokemonByBattlerIndex(battlerIndex: BattlerIndex.PLAYER | BattlerIndex.PLAYER_2): PlayerPokemon | undefined;
+  public getPokemonByBattlerIndex(battlerIndex: BattlerIndex): Pokemon | undefined;
+  public getPokemonByBattlerIndex(battlerIndex: BattlerIndex): Pokemon | undefined {
+    // TODO: `?? undefined` is dumb and a byproduct of us splicing `null`s into `getFIeld`
+    return this.getField()[battlerIndex] ?? undefined;
```
DayKev, line 837:
```suggestion
    return this.getField()[battlerIndex];
```
You can just replace the `null` with `undefined` in `#getField`'s array initialization instead of leaving a TODO for a single-line change (well 2-line since the TSDoc would also need to be updated).

src/battle-scene.ts
```diff
@@ -806,6 +807,23 @@ export class BattleScene extends SceneBase {
...
+   * @returns The {@linkcode Pokemon} with the given battler index,
+   * or `undefined` if no such Pokemon exists.
+   * @remarks
+   * This function is allowed to return inactive (i.e. fainted) Pokemon.
+   */
+  // TODO: Replace prior indexing into `getField` with this abstraction to make an overhaul of this system easier
+  public getPokemonByBattlerIndex(battlerIndex: BattlerIndex.ENEMY | BattlerIndex.ENEMY_2): EnemyPokemon | undefined;
+  public getPokemonByBattlerIndex(battlerIndex: BattlerIndex.PLAYER | BattlerIndex.PLAYER_2): PlayerPokemon | undefined;
+  public getPokemonByBattlerIndex(battlerIndex: BattlerIndex): Pokemon | undefined;
+  public getPokemonByBattlerIndex(battlerIndex: BattlerIndex): Pokemon | undefined {
```
```suggestion
  public getPokemonByBattlerIndex(battlerIndex: Exclude<BattlerIndex, BattlerIndex.ATTACKER>): Pokemon | undefined {
```
That third overload isn't needed, because `BattlerIndex.ATTACKER` should never be passed to this (so one of the first 2 overloads should always be matched).

src/phases/attempt-capture-phase.ts
```diff
@@ -14,41 +15,62 @@ import {
...
-    super(BattlerIndex.ENEMY + targetIndex);
+  constructor(fieldIndex: 0 | 1, pokeballType: PokeballType) {
+    super();
 
     this.pokeballType = pokeballType;
+    const index = (BattlerIndex.ENEMY + fieldIndex) as BattlerIndex.ENEMY | BattlerIndex.ENEMY_2;
+
+    // bang is benign as we immediately abort if nullish
+    this.pokemon = globalScene.getPokemonByBattlerIndex(index)!;
+    if (!this.pokemon || !this.pokemon.isActive(true)) {
```
DayKev:
```suggestion
    if (!this.pokemon?.isActive(true)) {
```

src/phases/attempt-capture-phase.ts
```diff
@@ -57,164 +79,235 @@ export class AttemptCapturePhase extends PokemonPhase {
...
-    this.pokeball.setOrigin(0.5, 0.625);
+    this.pokeball = globalScene.addFieldSprite(16, 80, "pb", pokeballAtlasKey).setOrigin(0.5, 0.625);
     globalScene.field.add(this.pokeball);
 
     globalScene.playSound(isCritical ? "se/crit_throw" : "se/pb_throw");
-    globalScene.time.delayedCall(300, () => {
-      globalScene.field.moveBelow(this.pokeball as Phaser.GameObjects.GameObject, pokemon);
+    waitTime(300).then(() => {
+      globalScene.field.moveBelow<Phaser.GameObjects.GameObject>(this.pokeball, pokemon);
     });
```
DayKev, line 96:
```suggestion
    await waitTime(300);
    globalScene.field.moveBelow<Phaser.GameObjects.GameObject>(this.pokeball, pokemon);
```

src/phases/attempt-capture-phase.ts
```diff
@@ -57,164 +79,235 @@ export class AttemptCapturePhase extends PokemonPhase {
...
+      ease: "Sine.easeIn",
+      scale: 0.25,
+      y: 20,
+    });
+
+    // Ball closes
+    this.pokeball.setTexture("pb", `${pokeballAtlasKey}_opening`);
+    pokemon.setVisible(false);
+    globalScene.playSound("se/pb_catch");
+    waitTime(17).then(() => this.pokeball.setTexture("pb", `${pokeballAtlasKey}`));
```
DayKev:
```suggestion
    await waitTime(17);
    this.pokeball.setTexture("pb", `${pokeballAtlasKey}`);
```

src/phases/attempt-capture-phase.ts
```diff
@@ -57,164 +79,235 @@ export class AttemptCapturePhase extends PokemonPhase {
...
+    // Trailing pause after the final wobble
+    await waitTime(500);
   }
 
-  failCatch(_shakeCount: number) {
-    const pokemon = this.getPokemon();
+  /**
+   * Resolve a failed capture attempt by restoring the Pokemon, replaying the ball opening, and ending the phase.
+   */
+  async failCatch(_shakeCount: number): Promise<void> {
```
DayKev:
```suggestion
  private async failCatch(): Promise<void> {
```

src/phases/attempt-capture-phase.ts
```diff
@@ -57,164 +79,235 @@ export class AttemptCapturePhase extends PokemonPhase {
...
+    await new Promise<void>(resolve =>
+      globalScene.time.delayedCall(250, () => doPokeballBounceAnim(this.pokeball, 16, 72, 350, resolve, isCritical)),
+    );
+
+    // Play the wobble animations for each successful shake check.
+    // Critical captures always show 1 wobble; the result is revealed after.
+    await this.animateShakeChecks(successfulShakes);
+
+    if (!caught) {
+      await this.failCatch(successfulShakes);
```
DayKev:
```suggestion
      await this.failCatch();
```

src/phases/attempt-capture-phase.ts
```diff
@@ -223,32 +316,257 @@ export class AttemptCapturePhase extends PokemonPhase {
...
+      }
+    }
+  }
+
+  /**
+   * Prompt the player to choose whether to keep a caught Pokemon upon their party being full.
+   * @returns A Promise that resolves with the chosen action.
+   */
+  private async promptFullPartyCatchAction(): Promise<FullPartyCatchAction> {
+    return await new Promise(resolve => {
```
DayKev: Benjie doesn't want any more new instances of code like this, you're supposed to use `Promise.withResolvers<T>`.
I think it would be like this? I'm not 100% familiar with async code yet, but there's no type errors with this.
```ts
  private async promptFullPartyCatchAction(): Promise<FullPartyCatchAction> {
    const { promise, resolve } = Promise.withResolvers<FullPartyCatchAction>();
    await globalScene.ui.setMode(
      UiMode.CONFIRM,
      () => resolve("summary"),
      () => resolve("pokedex"),
      () => resolve("release"),
      () => resolve("cancel"),
      "fullParty",
    );
    return promise;
  }
```

src/phases/attempt-capture-phase.ts
```diff
@@ -223,32 +316,257 @@ export class AttemptCapturePhase extends PokemonPhase {
...
+    });
+  }
+
+  /**
+   * Show the summary page for the caught Pokemon, then return the UI to message mode.
+   */
+  private async showCaughtPokemonSummary(): Promise<void> {
+    const { pokemon } = this;
+
+    // TODO: We should need to create a new dummy Pokemon solely to display the summary page
```
DayKev:
```suggestion
    // TODO: We shouldn't need to create a new dummy Pokemon solely to display the summary page
```

src/phases/attempt-capture-phase.ts
```diff
@@ -223,32 +316,257 @@ export class AttemptCapturePhase extends PokemonPhase {
...
+    const newPokemon = globalScene.addPlayerPokemon(
```
DayKev, line 477: This Pokémon object needs to be `.destroy()`ed.

src/phases/attempt-capture-phase.ts
```diff
@@ -223,32 +316,257 @@ export class AttemptCapturePhase extends PokemonPhase {
...
+      shiny: pokemon.shiny,
+      variant: pokemon.variant,
+      form: pokemon.formIndex,
+      female: pokemon.gender === Gender.FEMALE,
+    };
+
+    await new Promise<void>(resolve => {
+      globalScene.ui.setOverlayMode(UiMode.POKEDEX_PAGE, pokemon.species, attributes, null, null, () => {
+        globalScene.ui.setMode(UiMode.MESSAGE).then(resolve);
+      });
```
DayKev, line 523:
```suggestion
    await globalScene.ui.setOverlayMode(UiMode.POKEDEX_PAGE, pokemon.species, attributes);
    await globalScene.ui.setMode(UiMode.MESSAGE);
```
Why not like this?

DayKev (commented):
Partial review

src/data/pokeball.ts
Bertie690: Changing the enum name would be a breaking change.
Either way, not the main focus of the PR IMO.

test/tests/phases/capture-phase.test.ts
Bertie690: That... doesn't exist yet (IIRC `startingPokeballs` was added by the modifier rework?)

fabske0: It exists in Overrides.ts for a while (almost 2 years), but not in overrides-helper.ts

src/battle-scene.ts
Bertie690: That is not how TS overload signatures work.
Whether or not `BattlerIndex.ATTACKER` should be allowed as an argument, TS does not allow passing a value that is a union of 2 known overload types. Even if `BattlerIndex` was restricted to those 4 values, the lack of a case that encompasses both halves would make the function unusable when given a plain `BattlerIndex`.

src/phases/attempt-capture-phase.ts
Bertie690: We didn't wait for this in the prior version, and doing so would delay the other chunks of the sequence by 300ms.

src/phases/attempt-capture-phase.ts
Bertie690: See above

src/phases/attempt-capture-phase.ts
Bertie690: Former passes the function to the handler as a callback, latter does not.
I'm not 100% sure whether that distinction actually matters or not, since the UI isn't my area of expertise.

src/phases/attempt-capture-phase.ts
Bertie690: In my defence, the prior code never actually destroyed the pokemon itself.

src/battle-scene.ts
Bertie690: I agree, but since this isn't the main focus of this PR (and would have a non-zero chance of breaking something), I'd rather err on the side of caution for now and not lump in unrelated changes.

src/data/pokeball.ts
DayKev: How would it be a breaking change? The enum value is stored in save data, not the key.

Bertie690: Right - I forgot that's how `PokeballCounts` worked.

Either way, I will remain adamant about not mixing in unrelated changes and will turn this into a separate PR.

src/phases/attempt-capture-phase.ts
Bertie690: > you're supposed to use `Promise.withResolvers<T>`.

Just giving my viewpoint: from my understanding, `Promise.withResolvers<T>` was mainly used for handling large functions inside save data code (much of which I have manually cleaned up).
For small single-expression functions like this, the two versions are effectively identical, and I actually find the `new Promise` form marginally more readable than the equivalent `Promise.withResolvers` code (mostly because it stays a single statement vs splitting into 3).

An even more ideal version (read: giant pipe dream) would be for `setMode` itself to resolve with the button press result directly, but I do digress as that would require extensive refactors to a broken system.

Either way, we should probably discuss this further to come up with a consensus (ideally somewhere other than GitHub).

Bertie690: @SirzBenjie Any thoughts?

test/tests/phases/capture-phase.test.ts
Bertie690: Either way, if it works it works ig.
(This would need reworking inside the modifier rework basically regardless of what I do lol)

src/data/pokeball.ts
```diff
@@ -17,36 +17,33 @@ export function getPokeballAtlasKey(type: PokeballType): string {
       return "rb";
     case PokeballType.MASTER_BALL:
       return "mb";
-    case PokeballType.LUXURY_BALL:
-      return "lb";
+    // case PokeballType.LUXURY_BALL:
+    //   return "lb";
```
fabske0:
```suggestion
```

src/data/pokeball.ts
```diff
@@ -17,36 +17,33 @@ export function getPokeballAtlasKey(type: PokeballType): string {
...
+      return i18next.t("pokeball:rogueBall");
     case PokeballType.MASTER_BALL:
-      ret = i18next.t("pokeball:masterBall");
-      break;
-    case PokeballType.LUXURY_BALL:
-      ret = i18next.t("pokeball:luxuryBall");
-      break;
+      return i18next.t("pokeball:masterBall");
+    // case PokeballType.LUXURY_BALL:
+    //   return i18next.t("pokeball:luxuryBall");
```
fabske0:
```suggestion
```

src/data/pokeball.ts
```diff
@@ -76,8 +71,8 @@ export function getPokeballTintColor(type: PokeballType): number {
       return 0xd52929;
     case PokeballType.MASTER_BALL:
       return 0xa441bd;
-    case PokeballType.LUXURY_BALL:
-      return 0xffde6a;
+    // case PokeballType.LUXURY_BALL:
+    //   return 0xffde6a;
```
fabske0:
```suggestion
```

src/phases/attempt-capture-phase.ts
```diff
@@ -14,41 +15,62 @@ import {
...
+ */
+export class AttemptCapturePhase extends Phase {
   public readonly phaseName = "AttemptCapturePhase";
-  private pokeballType: PokeballType;
+  /** The Pokemon to attempt to capture. */
+  private readonly pokemon: EnemyPokemon;
+
+  /** The type of Pokeball being thrown. */
+  private readonly pokeballType: PokeballType;
   private pokeball: Phaser.GameObjects.Sprite;
```
fabske0: can we rename this to `pokeballSprite` or something similar

fabske0 (dismissed):

SirzBenjie:
The user facing changes section of the description needs to be rewritten such that users understand how it will impact them.

src/data/mystery-encounters/utils/encounter-pokemon-utils.ts
```diff
@@ -438,186 +435,226 @@ export async function applyModifierTypeToPlayerPokemon(
...
-          globalScene.trainer.setTexture(
-            `trainer_${globalScene.gameData.gender === PlayerGender.FEMALE ? "f" : "m"}_back`,
-          );
-        });
-      });
+  await waitTime(512);
+  audioManager.playSound("se/pb_throw");
+
+  // Trainer throw frames
+  // TODO: There should be a better way of queueing these lol
```
SirzBenjie:
```suggestion
  // TODO: Find a better way of queueing these.
```

One way would be to use a tweenChain with onComplete in each individual part.

SirzBenjie (dismissed):

PR #7159, opened 2026-03-20
fix(move): Baneful Bunker now protects from status effects

### LINKED ISSUE #7008: [Bug] Spore went through Baleful Bunker
### Describe the bug

Was just playing and protected against a parasect.  To my surprise, it spored me though baleful bunker.  I checked bulbapedia, and that does not appear to be intended behavior.

### Reproduction

In theory, just have something use spore while you use protect.  If should put your pokemon to sleep.

### Expected behavior

Protection moves, including Baleful Bunker, should block spore.

### Screenshots / Videos

_No response_

### Session export file

_No response_

### User data export file

_No response_

### Additional context

_No response_

verissimopedro:
Baneful Bunker now correctly blocks status-category moves (for example, Spore), in addition to blocking damaging moves.

Its contact side effect is unchanged: attackers are only poisoned if they make contact.

No UI changes.

Baneful Bunker was routed through damage-only protection handling, so status moves could bypass it.
That behavior is incorrect for Baneful Bunker.

This change fixes that rules mismatch and aligns move behavior with expected in-game mechanics.

Fixes #7008.

- Refactored protect resolution to use `ProtectedTag.blockStatus` in src/phases/move-effect-phase.ts, replacing subclass-based branching in protection checks.
- Added `blockStatus` on protection tags in src/data/battler-tags.ts:
  - `ProtectedTag` defaults to blocking status moves.
  - Damage-only protect tags keep `blockStatus = false`.
- Unified Baneful Bunker and Burning Bulwark under `ContactSetStatusProtectedTag` in src/data/battler-tags.ts.
- Updated battler-tag factory wiring in src/data/battler-tags.ts
- Updated protection-related type unions in src/@types/battler-tags.ts
- Expanded regression coverage in test/tests/moves/baneful-bunker.test.ts and added related protect/burning-bulwark regression tests to verify behavior remains correct after refactor.

## Screenshots/Videos
https://www.youtube.com/watch?v=48dd1TGTlBE

## How to test the changes?
1. Run focused tests:
   - pnpm test:silent baneful-bunker
2. Change the override configurations to give yourself Baneful Bunker and the opponent a status move like Spore.
3. In battle, use Baneful Bunker against the status move (e.g. Spore):
   - Confirm the status move is blocked and the opponent is not poisoned, since status moves don't make contact.

## Checklist


Bertie690:
@verissimopedro I'd like it if you could reformat your PR description to fit the [default PR template](https://github.com/pagefaultgames/pokerogue/blob/beta/.github/pull_request_template.md?plain=1)? 

It's not a big deal in the end, but it makes automatic changelog generation at least slightly easier.

src/data/battler-tags.ts
```diff
@@ -1869,6 +1869,32 @@ export class ContactDamageProtectedTag extends ContactProtectedTag {
...
 
+/**
+ * `BattlerTag` class for moves that block ALL moves (including status moves) and
+ * inflict poison status on the attacker if the attacker's move makes contact.
+ * @see {@link https://bulbapedia.bulbagarden.net/wiki/Baneful_Bunker_(move)}
+ * @remarks
+ * Unlike {@linkcode DamageProtectedTag}, this blocks status moves as well as damaging moves.
+ * @sealed
+ */
+export class ContactSetPoisonProtectedTag extends ContactProtectedTag {
```
Bertie690: I'm gonna be frank here - having 2 separate classes for baneful bunker and burning bulwark is dubious at best. These 2 moves do virtually the same thing aside from the status effect they apply and whether they block status moves or not, which is hardly worth the loss of generality.

I'd presume one way to make this far simpler and easier to maintain would be exposing a public getter (call it `blockStatus`) onto the base `ProtectedTag` class, defaulting to `true` on everything but Burning Bulwark and company. (This would be checked inside `move-effect-phase`, in the same place as the `instanceof` checks for `DamageProtectedTag` go now.)

Bertie690 (commented):
As mentioned in my review, I think this would be a solid opportunity to prune out some jank in the codebase.
It's not terrible _per se_, but I'd rather nip any code debt in the bud (so to speak)

src/data/battler-tags.ts
verissimopedro: I have reformated my PR description and will work on this approach

verissimopedro: Just pushed a new commit with the approach using the new blockStatus.
Now in the move effect phase if we are dealing with a status move, only moves with the blockStatus set to true can attempt to block, and since moves like King’s Shield, Obstruct, Silk Trap and Burning Bulwark which don't protect against status moves have it set to false, they can't attempt to block.
No longer hardcoded by checking specific protection subclasses.

Let me know when I should edit the template to fit this new commit.

src/data/battler-tags.ts
```diff
@@ -1872,20 +1878,42 @@ export class ContactDamageProtectedTag extends ContactProtectedTag {
 /** Base class for `BattlerTag`s that block damaging moves but not status moves */
 export abstract class DamageProtectedTag extends ContactProtectedTag {
   public declare readonly tagType: DamageProtectedTagType;
+
+  public override get blockStatus(): boolean {
+    return false;
+  }
 }
```
Bertie690: With the getter in place, we don't need a separate class for the effect - Burning Bulwark can simply override the getter directly

src/data/battler-tags.ts
```diff
@@ -1769,6 +1769,12 @@ export class InfestationTag extends DamagingTrapTag {
 
 export class ProtectedTag extends BattlerTag {
   public declare readonly tagType: ProtectionBattlerTagType;
+
+  /** Whether this protection effect blocks status moves in addition to damaging moves. */
```
```suggestion
  /** 
   * Whether this protection effect should block status moves.
   * @defaultValue `true`
   * @remarks
   * Damaging moves are always blocked, regardless of this flag's state. 
   */
```

DayKev (dismissed):
Looks good. I added some additional tests as well that confirm the other moves still work correctly after the change.

verissimopedro:
Updated the PR description to reflect these new changes.

SirzBenjie (dismissed):

src/data/battler-tags.ts
```diff
@@ -1872,20 +1878,42 @@ export class ContactDamageProtectedTag extends ContactProtectedTag {
...
 
-export class ContactSetStatusProtectedTag extends DamageProtectedTag {
+/**
+ * `BattlerTag` class for protection effects that set a status condition on contact.
+ * Used by {@linkcode MoveId.BANEFUL_BUNKER} and {@linkcode MoveId.BURNING_BULWARK}.
+ */
+export class ContactSetStatusProtectedTag extends ContactProtectedTag {
   public declare readonly tagType: ContactSetStatusProtectedTagType;
   /** The status effect applied to attackers */
   #statusEffect: StatusEffect;
```
Bertie690: While you're at it, can you make this `readonly`?

PR #7166, opened 2026-03-21
feat: Add weather animations toggle setting

heatbang9:
## Description
Adds a new setting to toggle weather effect animations on/off.

## Changes
- Added `weatherAnimations` property to BattleScene (default: true)
- Added `Weather_Animations` setting key in settings
- Added UI toggle in Display settings (Move Animations 옆에 위치)
- When disabled, skips weather animation but still shows weather lapse message and applies weather effects

## Testing
- Tested that weather effects still apply damage/abilities when animation is disabled
- Setting properly saves and loads from game data
- UI shows correct toggle state

## Related
- This is useful for players who want to speed up gameplay or have performance issues with weather animations


heatbang9:
Duplicate of #7167 - using cleaner branch name

PR #7183, opened 2026-03-23
fix(ui): filtering the Pokédex by biome works again

DayKev:
The Pokédex will show the correct Pokémon that appear in each biome when filtering on the main Pokédex page.

Note: Technically it's mostly broken only on beta than main, but there were still some inconsistencies before the biome structure rework that are also fixed by this.

The Pokédex is bugged and doesn't properly consider evolution lines when displaying which Pokémon appear in which biomes.

The Pokédex will get every species in an evolutionary line (starting with the current Pokémon) and then check the biomes those species appear in, and it will consider the initial species as appearing in any of the biomes from the entire line (but will not hop to non-contiguous evolutionary lines, e.g. it will not check Unovan Lilligant if the biome has Hisuian Lilligant specified).

Some unused private fields were removed from the `PokedexUiHandler` class as well.

## Screenshots/Videos
Demonstrating Petilil (Forest + Tall Grass), Lilligant (Forest + Tall Grass), and Hisuian Lilligant (Meadow):

Due to Petilil being specified in the Forest and Tall Grass biomes, both Lilligant and Hisuian Lilligant should appear in those biomes on later waves due to Petilil being auto-evolved.
However, due to only Hisuian Lilligant being specified in the Meadow biome, only Petilil and Hisuian Lilligant should appear in that biome because when Hisuian Lilligant is auto-devolved, it can't become Unovan Lilligant.

Before:
<details><summary>Grass types in Forest biome (Budew + Roserade missing)</summary>

image

</details>
<details><summary>Grass types in Meadow biome (Budew + Petilil missing)</summary>

image

</details>
<details><summary>"Habitat unknown" filter</summary>

image

</details>

After:
<details><summary>Grass types in Forest biome</summary>

image

</details>
<details><summary>Grass types in Meadow biome</summary>

image

</details>
<details><summary>"Habitat unknown" filter (no random baby pokemon)</summary>

image

</details>

## Checklist


src/ui/handlers/pokedex-ui-handler.ts
```diff
@@ -1453,8 +1437,8 @@ export class PokedexUiHandler extends MessageUiHandler {
 
     this.filteredPokemonData = [];
 
-    allSpecies.forEach(species => {
-      const starterId = this.getStarterSpeciesId(species.speciesId);
+    for (const species of allSpecies) {
+      const starterId: SpeciesId = this.getStarterSpeciesId(species.speciesId);
```
fabske0: Instead of making this `const` implicit `speciesId`, could the method be adjusted to return a speciesId instead of number

DayKev: Probably, I'll take a look.

DayKev: No errors doing that, should be fine. Also added typing for the param it takes since that was missing.

PR #7202, opened 2026-03-28
fix: locales files should now properly utilize the cache

DayKev:
Locales will no longer improperly load old cached files.

A regression in locales file caching was noticed.

The manifest file will now be loaded before the i18n plugin is initialized.

## How to test the changes?
Watch the network tab in the browser devtools and make sure `?t=...` is appended to the locales files (filtering for `dialogue.json` is an easy way to locate one).
Note: You will need a `manifest.json` file in the repo root, you can copy the contents of the current one from live [here](https://pokerogue.net/manifest.json).

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "pokemon-rogue-battle",
   "private": true,
-  "version": "1.11.15",
+  "version": "1.11.16",
   "type": "module",
   "scripts": {
     "start:prod": "vite --mode production",


src/main.ts
```diff
@@ -1,7 +1,7 @@
 import "#app/polyfills"; // All polyfills MUST be loaded first for side effects
-import "#plugins/i18n"; // Initializes i18n on import
+await import("#init/init-manifest"); // initializes the manifest, must be done *before* i18n is initialized due to being used for caching
```
Bertie690, line 2: Can we hoist this and turn it into a regular static side-effect import?

DayKev: I have no idea what that means so 🤷.

DayKev: Sorry, I should clarify: I know *what* you mean (Vitest does it somehow), but I have no idea *how* we would do that nor why adding complexity would be better than just changing the imports to `await import`.

fabske0:
There are no `?t=...` after the locales files for me. Maybe i did something wrong?

image

</details>

src/main.ts
```diff
@@ -15,23 +15,7 @@ if (isBeta || isDev) {
...
-};
-
-// Catch global promise rejections and display them in an alert so users can report the issue.
-window.addEventListener("unhandledrejection", event => {
-  // const errorString = `Received unhandled promise rejection. Open browser console and click OK to see details.\nReason: ${event.reason}`;
-  console.error(event.reason);
-  //alert(errorString);
-});
-
-async function startGame(gameManifest?: Record<string, string>): Promise<void> {
```
fabske0, line 34: I suppose this isn't needed anymore with the alert disabled?

SirzBenjie:
> There are no `?t=...` after the locales files for me. Maybe i did something wrong?
> 
> Image

Presumably because there's no manifest?

DayKev:
> There are no `?t=...` after the locales files for me. Maybe i did something wrong?
> 
> Image

Sorry, forgot to mention that you need to copy the `manifest.json` file from https://pokerogue.net/manifest.json to the repo root when testing for it to work.

src/main.ts
DayKev: Correct, the only reason to have it was due to the alerts but those were disabled ages ago.

PR #7240, opened 2026-04-10
fix(beta): Moves with post-target effects no longer softlock with no valid targets

emdeann:
Moves like explosion, burn up, etc. (ones with a POST_TARGET effect) will no longer freeze the game when used with no valid targets i.e. into protect

https://discord.com/channels/1125469663833370665/1125894949020381285/1491239539736576040

#7189 changed the move animation code to loop through targets instead of always using the first target, but didn't consider the case where no valid targets exist. There is an early return after the animation loop intended to avoid calling `postAnimCallback` multiple times, but this means that if the loop never iterated, the return will be hit without ever running the callback. That breaks the chain and `end()` -> `shiftPhase` is never called, so the game freezes.

It's fixed by simply using `user` as the animation target if none exists. I can't say for certain that this is correct in all cases, but it's equivalent to the previous logic of `target?.getBattlerIndex() ?? BattlerIndex.ATTACKER` (the code still looked like this before my edits, but all the optional chaining was useless as `target` can't be undefined in the for loop).

## How to test the changes?
```ts
const overrides = {
  ENEMY_MOVESET_OVERRIDE: MoveId.EXPLOSION,
  ENEMY_SPECIES_OVERRIDE: SpeciesId.PAWMOT,
  MOVESET_OVERRIDE: MoveId.PROTECT
```

## Checklist

### DIFF
--- a/test/tests/moves/post-apply-effect.test.ts
+++ b/test/tests/moves/post-apply-effect.test.ts
@@ -1,0 +1,45 @@
+/*
+ * SPDX-FileCopyrightText: 2024-2026 Pagefault Games
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+import { AbilityId } from "#enums/ability-id";
+import { MoveId } from "#enums/move-id";
+import { SpeciesId } from "#enums/species-id";
+import { GameManager } from "#test/framework/game-manager";
+import Phaser from "phaser";
+import { beforeAll, beforeEach, describe, it } from "vitest";
+
+describe("Move - Post Apply Effect", () => {
+  let phaserGame: Phaser.Game;
+  let game: GameManager;
+
+  beforeAll(() => {
+    phaserGame = new Phaser.Game({
+      type: Phaser.HEADLESS,
+    });
+  });
+
+  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


test/tests/moves/post-apply-effect.test.ts
```diff
@@ -0,0 +1,45 @@
...
+      .enemyLevel(100);
+  });
+
+  it("should not softlock when failing", async () => {
+    await game.classicMode.startBattle(SpeciesId.FEEBAS);
+
+    game.move.use(MoveId.PROTECT);
+    await game.move.forceEnemyMove(MoveId.EXPLOSION);
+    await game.toEndOfTurn();
+    // If we get here, there was no softlock
```
DayKev, line 44:
```suggestion

    // If we get here, there was no softlock
    expect(game.field.getPlayerPokemon()).toHaveFullHp();
```
An `expect` is needed for tests, even if it's "unnecessary" otherwise.

DayKev (dismissed):

damocleas (approved):

DayKev (approved):

Bertie690 (approved):

PR #7260, opened 2026-04-17
refactor(battle): `StatStageChangePhase` rewrite

emdeann:

- Like champions, simultaneous stat changes happen at the same time with one animation instead of sequentially
  - The exception is that changes on multiple targets at once still happen one at a time (ie intimidate in doubles). I think this PR is big enough as is
- Stat changes from berry phase and boss segments are consolidated into a single phase each 

SSCP was in desperate need of a rewrite and the champions changes provide a good motivation.

The phase itself is rewritten to take an options object and call methods instead of doing significant logic directly in `start`

- Replaced logic to queue a phase for every stat change to instead queue a maximum of two phases - one for all positive changes and one for all negative changes
- Reaction/protection abilities are now triggered once for all stats in a group instead of per stat. This would cause a deviation from mainline if we had an item like Mirror Herb, but it does not apply to Opportunist (because it would end up copying all changes anyway). Balance decided that the improved game speed is worth the technical deviation here.
- Changed the phase to accept an array of `changes` instead of separate `stats` and `stages` to ensure there is a stage for every stat change. Also created the `groupStatChange` utility method to build a full `changes` array when a group of stats all want to change by the same amount, to avoid excessive burden on callsites
- Outside of this, most changes are breaking up/moving things around

I modified `BerryPhase` to "queue" stat changes for each berry, then apply them all at the end of the phase. This is technically a behavior change as it means stat changes will always be the last thing to happen in a berry phase, but I don't think this is significant. The way I implemented this was by adding a `queuedBerryStatChanges` array to the pokemon. There is probably a better way than this, but I'm trying to avoid touching modifiers right now.

The same change is made for boss segments, but it's cleaner as it can be handled entirely in the method. I didn't want to further increase the complexity of the method, so while there I also removed the reimplementation of `weightedPick` and used the utility instead.

Other changes are callsites and can generally be ignored, barring typos.

## Screenshots/Videos
<details><summary>Basic Changes</summary>

https://github.com/user-attachments/assets/ad123ad7-9ae8-4232-ac57-56ba93e656e6

</details>
<details><summary>Boss Segment/Berry Phase</summary>

https://github.com/user-attachments/assets/d5a5b530-c5c4-4f84-865a-5695717715c5

Here the enemies have all stat berries + starf berries, so the changes are +1 to all and then an additional +2 to a random stat.
</details>

## How to test the changes?
Stat changes are very pervasive, pretty much just play the game. I wrote a small amount of automated tests

## Checklist


src/enums/stat-change-source.ts
```diff
@@ -0,0 +1,5 @@
+export enum StatChangeSource {
+  STICKY_WEB,
+  MIRROR_ARMOR,
+  OPPORTUNIST,
+}
```
DayKev:
```suggestion
import type { ValueOf } from "type-fest";

export const StatChangeSource = {
  STICKY_WEB: 1,
  MIRROR_ARMOR: 2,
  OPPORTUNIST: 3,
}

export type StatChangeSource = ValueOf<typeof StatChangeSource>;
```
Please don't add new enums.

src/utils/stat-change.ts
```diff
@@ -0,0 +1,6 @@
+import type { BattleStat } from "#enums/stat";
+import type { StatChange } from "#types/stat-change";
+
+export function groupStatChange(stats: readonly BattleStat[], stages: number): StatChange[] {
```
DayKev, line 10: Instead of making a utility function, shouldn't this be handled by the SSCPhase and the callsites just pass the pre-grouped stats?

DayKev (commented):
A quick skim.

src/enums/stat-change-source.ts
emdeann: Doesn't matthew have a pr open right now to add the inline enums plugin? Or is that not what I think it is

src/utils/stat-change.ts
emdeann: If you handle it that way then you have an overload-esque thing where the phase has to handle being passed an array of stats + a number or a proper array of changes. I dislike constructs like that

src/enums/stat-change-source.ts
DayKev: Yes, but that's a stopgap, not the final solution. The goal is to have zero enums and remove the plugin.

src/enums/stat-change-source.ts
```diff
@@ -0,0 +1,9 @@
+import type { ValueOf } from "type-fest";
+
+export const StatChangeSource = {
+  STICKY_WEB: 1,
+  MIRROR_ARMOR: 2,
+  OPPORTUNIST: 3,
+};
```
Bertie690: you forgot the `as const`

```suggestion
export const StatChangeSource = {
  STICKY_WEB: 1,
  MIRROR_ARMOR: 2,
  OPPORTUNIST: 3,
} as const;
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+  /** If `true`, skip `StatStageChangeMultiplierAbAttr` */
+  ignoreAbilities?: boolean;
+  /** Callback invoked with the applied changes. */
+  onChange?: StatStageChangeCallback;
+  /** The category of effect that produced this change, if relevant */
+  sourceEffect?: StatChangeSource;
+  /**
+   * When `true`, pre-processing (multipliers, protection checks, sign-splitting)
+   * is skipped because it was already performed by the phase that queued this one.
+   */
```
Bertie690:
```suggestion
   * Whether to avoid pre-processing the stat changes (applying multipliers, protection checks, sign-splitting, etc.)
   * if these effects were already performed by the phase that queued this one.
   * @remarks
   * Should not be passed by anything other than this Phase.
   */
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
-    this.comingFromMirrorArmorUser = comingFromMirrorArmorUser;
-    this.comingFromStickyWeb = comingFromStickyWeb;
+  private isIncrease = false;
+
+  constructor(options: StatStageChangePhaseOptions) {
+    super(options.battlerIndex);
+    this.options = { ...options };
+    // Deep copying allows this phase to simplify operations by modifying changes in place
+    this.options.changes = deepCopy(options.changes).filter(c => c.stages !== 0); // Allow changes with 0 stages to be passed as no-ops
+    this.selfTarget = options.sourcePokemon != null && options.sourcePokemon === this.getPokemon();
```
Bertie690, line 43: The nullish check is unnecessary so long as `this.getPokemon` is never nullish itself.

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
-          this.ignoreAbilities,
-          this.canBeCopied,
-          this.onChange,
-          this.comingFromMirrorArmorUser,
-        );
-      }
+    const pokemon = this.getPokemon();
+
+    if (!pokemon.isActive(true)) {
       return this.end();
```
Bertie690: i am surprised Biome doesn't shriek like a banshee for this.

```suggestion
      this.end();
      return;
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
-    this.ignoreAbilities = ignoreAbilities;
-    this.canBeCopied = canBeCopied;
-    this.onChange = onChange;
-    this.comingFromMirrorArmorUser = comingFromMirrorArmorUser;
-    this.comingFromStickyWeb = comingFromStickyWeb;
+  private isIncrease = false;
+
+  constructor(options: StatStageChangePhaseOptions) {
+    super(options.battlerIndex);
+    this.options = { ...options };
```
Bertie690, line 41: Do we need to shallow clone `options` IN ADDITION to deep cloning the array?
I would presume nothing from the outside should attempt to modify it.

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+      this.applyStageMultipliers(pokemon);
+      this.removeCancelledChanges(pokemon);
+      this.splitBySign();
     }
 
-    if (!pokemon.isActive(true)) {
+    this.isIncrease = this.options.changes.some(c => c.stages > 0);
+    if (this.options.changes.length === 0) {
       return this.end();
     }
```
Bertie690:
```suggestion
    if (this.options.changes.length === 0) {
      this.end();
      return;
    }
    this.isIncrease = this.options.changes.some(c => c.stages > 0);
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
-          this.showMessage,
-          this.ignoreAbilities,
-          this.canBeCopied,
-          this.onChange,
-          this.comingFromMirrorArmorUser,
-        );
-      }
+    const pokemon = this.getPokemon();
+
+    if (!pokemon.isActive(true)) {
```
Bertie690, line 50: When would this be true? Is this just a carryover from the past version?

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
 
-  getStatStageChangeMessages(stats: readonly BattleStat[], stages: number, relStages: number[]): string[] {
-    const messages: string[] = [];
+  /**
+   * Play the rising stat change animation, depending on whether there were increases or decreases.
+   *
+   * @param pokemon - The Pokemon to animate
+   * @param onComplete - Callback for after the animation completes
+   */
+  private playStatChangeAnimation(pokemon: Pokemon, onComplete: () => void): void {
```
Bertie690: This callback only ever has 1 value.

Remove the `onComplete` callback, make this function `async` and call `applyStatChangesAndEnd` inside `start`

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+    const tileX = (this.player ? 106 : 236) * scale;
+    const tileY = ((this.player ? 148 : 84) + (this.isIncrease ? 160 : 0)) * scale;
+    const tileWidth = 156 * scale;
+    const tileHeight = 316 * scale;
+
+    // On increase, show the red sprite located at ATK; on decrease, the blue sprite at SPD
+    const spriteColor = this.isIncrease ? Stat[Stat.ATK].toLowerCase() : Stat[Stat.SPD].toLowerCase();
+    const statSprite = globalScene.add.tileSprite(tileX, tileY, tileWidth, tileHeight, "battle_stats", spriteColor);
+    statSprite.setPipeline(globalScene.fieldSpritePipeline);
+    statSprite.setAlpha(0);
```
Bertie690: method chaining pls

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+    statSprite.setScale(6);
+    statSprite.setOrigin(0.5, 1);
+    statSprite.setMask(new Phaser.Display.Masks.BitmapMask(globalScene, pokemon.maskSprite ?? undefined));
+
+    globalScene.playSound(`se/stat_${this.isIncrease ? "up" : "down"}`);
+
+    globalScene.tweens.add({
```
Bertie690: May I request we denest this tween chain (and also wait for it to _end_  before returning)?

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
-    } else {
-      opponentPokemon = globalScene.getPlayerField()[globalScene.currentBattle.lastPlayerInvolved];
+    if (!this.options.processed) {
+      this.applyStageMultipliers(pokemon);
+      this.removeCancelledChanges(pokemon);
+      this.splitBySign();
     }
 
-    if (!pokemon.isActive(true)) {
+    this.isIncrease = this.options.changes.some(c => c.stages > 0);
```
Bertie690: It would probably make more sense to include `isIncrease` as a parameter inside the private constructor signature and pass it along during `splitBySign` (effectively memoizing it).

Then again, not too miffed either way.

src/data/arena-tag.ts
```diff
@@ -318,18 +321,23 @@ export class MistTag extends SerializableArenaTag {
...
-      const bypassed = new BooleanHolder(false);
+      const bypassed = new ValueHolder(false);
       applyAbAttrs("InfiltratorAbAttr", { pokemon: source, simulated, bypassed });
       if (bypassed.value) {
         return false;
       }
     }
 
-    cancelled.value = true;
+    cancelledStats.push(...changes.filter(c => c.stages < 0 && !cancelledStats.includes(c.stat)).map(c => c.stat));
```
Bertie690: Is there a more elegant and/or performant way of doing this than looping over both arrays like this?

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
-          false,
-          pokemon,
-          cancelled,
-          opponentPokemon,
-        );
-      }
+    const negative = this.options.changes.filter(c => c.stages < 0);
+    if (negative.length === 0) {
+      return;
+    }
```
Bertie690, line 111: Just a thought - this is currently hardcoding that stat stage prevention abilities and effects will **only** block stat drops and never stat boosts.

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
-      // If one stat stage decrease is cancelled, simulate the rest of the applications
-      if (cancelled.value) {
-        simulate = true;
-      }
+    if (cancelledStats.length < this.options.changes.length) {
+      this.checkAbilityProtection(pokemon, opponent, negative, cancelledStats);
+    }
+
+    this.options.changes = this.options.changes.filter(c => !cancelledStats.includes(c.stat));
+  }
```
Bertie690: Assuming `cancelledStats` is supposed to not contain duplicate entries (which we should very much SAY SOMETHING ABOUT), perhaps a `Set` would be a more suitable datatype here?

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+    const abAttrParams: PreStatStageChangeAbAttrParams & ConditionalUserFieldProtectStatAbAttrParams = {
+      pokemon,
+      changes,
+      cancelledStats,
+      simulated: false,
+      target: pokemon,
+    };
+
+    // It is the responsibility of the ability to check if it is applicable based on `stats` and `cancelledStats`
+    applyAbAttrs("ProtectStatAbAttr", abAttrParams);
```
Bertie690:
```suggestion
    applyAbAttrs("ProtectStatAbAttr", abAttrParams);
```

I think this somewhat goes without saying (given that's basically how 80% of our ability system works)

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
-  relativeChanges: number[],
-) => void;
+export interface StatStageChangePhaseOptions {
+  battlerIndex: BattlerIndex | number;
+  changes: readonly StatChange[];
+  /** The Pokemon who caused these stat changes (may be the same as the Pokemon this Phase is applied to). */
+  sourcePokemon: Pokemon | undefined;
+  /** If `true`, skip `StatStageChangeMultiplierAbAttr` */
+  ignoreAbilities?: boolean;
+  /** Callback invoked with the applied changes. */
```
Bertie690:
```suggestion
  /** 
   * A callback to invoke with the applied changes.
   * Used exclusively to allow Stockpile to track the stat stages actually applied.
   */
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+  onChange?: StatStageChangeCallback;
+  /** The category of effect that produced this change, if relevant */
+  sourceEffect?: StatChangeSource;
```
Bertie690: Perhaps rename to `effectType`?

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,419 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+  /**
+   * Compute the relative change for each requested change by clamping to [-6, 6].
+   *
+   * @param pokemon - The Pokemon receiving the stat changes
+   * @returns A new array of {@linkcode StatChange}s
+   */
+  private getAppliedChanges(pokemon: Pokemon): StatChange[] {
+    return this.options.changes.map(({ stat, stages }) => {
+      const current = pokemon.getStatStage(stat);
+      const clamped = stages > 0 ? Math.min(current + stages, 6) : Math.max(current + stages, -6);
```
Bertie690: can't we just unconditionally clamp it between -6 and 6?

```suggestion
      const clamped = Phaser.Math.Clamp(current + stages, -6, 6);
```

src/enums/stat-change-source.ts
DayKev:
```suggestion
} as const;
```

src/phases/berry-phase.ts
```diff
@@ -67,7 +67,13 @@ export class BerryPhase extends FieldPhase {
       }
       globalScene.eventTarget.dispatchEvent(new BerryUsedEvent(berryModifier));
     }
+    globalScene.phaseManager.unshiftNew("StatStageChangePhase", {
```
Bertie690: this queues a no-op SSCP on an empty array

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,420 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
-// TODO: Refactor this mess of a phase
+/**
+ * Phase responsible for resolving, animating, and applying one or more
+ * stat-stage changes to a single target Pokemon.
+ *
+ * Changes to multiple stats may be applied at once. If both raises and drops are provided to the same phase,
+ * it will be split into one phase for raises and one phase for drops.
+ */
 export class StatStageChangePhase extends PokemonPhase {
   public readonly phaseName = "StatStageChangePhase";
```
Bertie690:
```suggestion
  public override readonly phaseName = "StatStageChangePhase";
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,420 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
-    this.selfTarget = selfTarget;
-    this.stats = stats;
-    this.stages = stages;
-    this.showMessage = showMessage;
-    this.ignoreAbilities = ignoreAbilities;
-    this.canBeCopied = canBeCopied;
-    this.onChange = onChange;
-    this.comingFromMirrorArmorUser = comingFromMirrorArmorUser;
-    this.comingFromStickyWeb = comingFromStickyWeb;
+  private isIncrease = false;
```
Bertie690, line 36: doc comments?

Bertie690 (changes requested):

test/tests/battler-tags/octolock.test.ts
```diff
@@ -27,8 +27,12 @@ describe("BattlerTag - OctolockTag", () => {
 
       vi.spyOn(game.scene.phaseManager, "unshiftPhase").mockImplementation(phase => {
         expect(phase).toBeInstanceOf(StatStageChangePhase);
-        expect((phase as StatStageChangePhase)["stages"]).toEqual(-1);
-        expect((phase as StatStageChangePhase)["stats"]).toEqual([Stat.DEF, Stat.SPDEF]);
+        expect((phase as StatStageChangePhase)["options"].changes).toEqual(
```
Bertie690, line 30: can we just delete this file 
it is so fragile

test/tests/battler-tags/stockpiling.test.ts
```diff
@@ -34,10 +40,11 @@ describe("BattlerTag - StockpilingTag", () => {
 
       vi.spyOn(game.scene.phaseManager, "unshiftPhase").mockImplementation(phase => {
         expect(phase).toBeInstanceOf(StatStageChangePhase);
-        expect((phase as StatStageChangePhase)["stages"]).toEqual(1);
-        expect((phase as StatStageChangePhase)["stats"]).toEqual(expect.arrayContaining([Stat.DEF, Stat.SPDEF]));
+        const options = (phase as StatStageChangePhase)["options"];
```
Bertie690, line 43: ditto here

test/tests/phases/stat-stage-change-phase.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  beforeEach(() => {
+    game = new GameManager(phaserGame);
+    game.override
+      .criticalHits(false)
+      .battleStyle("single")
+      .enemySpecies(SpeciesId.RATTATA)
+      .enemyAbility(AbilityId.BALL_FETCH)
+      .ability(AbilityId.BALL_FETCH)
+      .passiveAbility(AbilityId.NO_GUARD)
```
Bertie690, line 29: moveset override ;(

test/tests/phases/stat-stage-change-phase.test.ts
```diff
@@ -0,0 +1,202 @@
...
+        onChange,
+      });
+
+      game.move.use(MoveId.SPLASH);
+      await game.toEndOfTurn();
+
+      expect(onChange).toHaveBeenCalledTimes(1);
+      const applied = onChange.mock.calls[0][1];
+      expect(applied).toEqual([{ stat: Stat.ATK, stages: 1 }]);
+      expect(player).toHaveStatStage(Stat.ATK, 6);
```
Bertie690: use `expect().toHaveBeenCalledExactlyOnceWith`

test/tests/phases/stat-stage-change-phase.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+      expect(enemy).toHaveStatStage(Stat.ATK, -6);
+    });
+
+    it("should report the clamped change to the onChange callback", async () => {
+      await game.classicMode.startBattle(SpeciesId.MIGHTYENA);
+      const player = game.field.getPlayerPokemon();
+      player.setStatStage(Stat.ATK, 5);
+
+      const onChange = vi.fn();
```
Bertie690: ANY TYPE PARAM FFS

test/tests/phases/stat-stage-change-phase.test.ts
```diff
@@ -0,0 +1,202 @@
+import { AbilityId } from "#enums/ability-id";
+import { MoveId } from "#enums/move-id";
+import { SpeciesId } from "#enums/species-id";
+import { Stat } from "#enums/stat";
+import { GameManager } from "#test/framework/game-manager";
```
Bertie690, line 5:
```suggestion
import { GameManager } from "#test/framework/game-manager";
import type { StatStageChangeCallback } from "#types/stat-change";
```

test/tests/phases/stat-stage-change-phase.test.ts
Bertie690:
```suggestion
      const onChange = vi.fn<StatStageChangeCallback>();
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,420 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+  processed?: boolean;
+}
 
-// TODO: Refactor this mess of a phase
+/**
+ * Phase responsible for resolving, animating, and applying one or more
+ * stat-stage changes to a single target Pokemon.
+ *
+ * Changes to multiple stats may be applied at once. If both raises and drops are provided to the same phase,
+ * it will be split into one phase for raises and one phase for drops.
```
Bertie690, line 27:
```suggestion
 * they will be split into one phase for raises and one phase for drops.
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,420 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+    if (cancelledStats.length < this.options.changes.length) {
+      this.checkAbilityProtection(pokemon, opponent, negative, cancelledStats);
+    }
 
-      // If one stat stage decrease is cancelled, simulate the rest of the applications
-      if (cancelled.value) {
-        simulate = true;
-      }
+    this.options.changes = this.options.changes.filter(c => !cancelledStats.includes(c.stat));
+  }
```
Bertie690, line 130:
```suggestion
   * @param opponentPokemon - The Pokemon that caused the change, if present; used to resolve the target for Mirror Armor's reflection effect.
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,420 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
 
-      return !cancelled.value;
+  /**
```
Bertie690, line 132:
```suggestion
   * @param cancelledStats - An array containing each {@linkcode BattleStat} whose change was cancelled; will be modified by this method
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,420 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+
+    if (whiteHerb) {
+      pokemon.loseHeldItem(whiteHerb);
+      globalScene.updateModifiers(this.player);
     }
   }
 
-  getStatStageChangeMessages(stats: readonly BattleStat[], stages: number, relStages: number[]): string[] {
-    const messages: string[] = [];
+  /**
```
Bertie690, line 313:
```suggestion
   * Play the stat change animation, depending on whether there were increases or decreases.
```

src/data/abilities/ab-attrs.ts
```diff
@@ -472,13 +475,11 @@ export class TypeImmunityStatStageChangeAbAttr extends TypeImmunityAbAttr {
     super.apply(params);
     cancelled.value = true; // Suppresses "No Effect" message
     if (!simulated) {
```
Bertie690, line 478:
```suggestion
    if (!simulated) {
      const { stat, stages } = this;
```

src/data/abilities/ab-attrs.ts
```diff
@@ -472,13 +475,11 @@ export class TypeImmunityStatStageChangeAbAttr extends TypeImmunityAbAttr {
...
-      globalScene.phaseManager.unshiftNew(
-        "StatStageChangePhase",
-        pokemon.getBattlerIndex(),
-        true,
-        [this.stat],
-        this.stages,
-      );
+      globalScene.phaseManager.unshiftNew("StatStageChangePhase", {
+        battlerIndex: pokemon.getBattlerIndex(),
+        changes: [{ stat: this.stat, stages: this.stages }],
```
Bertie690, line 481:
```suggestion
        changes: [{ stat, stages }],
```

src/data/abilities/ab-attrs.ts
```diff
@@ -644,13 +645,11 @@ export class MoveImmunityStatStageChangeAbAttr extends MoveImmunityAbAttr {
   override apply(params: MoveImmunityAbAttrParams): void {
     super.apply(params);
     // TODO: We probably should not unshift the phase if this is simulated
```
Bertie690, line 648:
```suggestion
    // TODO: We probably should not unshift the phase if this is simulated
    const { stat, stages } = this;

```

src/data/abilities/ab-attrs.ts
```diff
@@ -644,13 +645,11 @@ export class MoveImmunityStatStageChangeAbAttr extends MoveImmunityAbAttr {
...
-    globalScene.phaseManager.unshiftNew(
-      "StatStageChangePhase",
-      params.pokemon.getBattlerIndex(),
-      true,
-      [this.stat],
-      this.stages,
-    );
+    globalScene.phaseManager.unshiftNew("StatStageChangePhase", {
+      battlerIndex: params.pokemon.getBattlerIndex(),
+      changes: [{ stat: this.stat, stages: this.stages }],
```
Bertie690, line 651:
```suggestion
      changes: [{ stat, stages }],
```

src/data/abilities/ab-attrs.ts
```diff
@@ -745,38 +744,32 @@ export class PostDefendStatStageChangeAbAttr extends PostDefendAbAttr {
...
-        globalScene.phaseManager.unshiftNew(
-          "StatStageChangePhase",
-          other.getBattlerIndex(),
-          false,
-          [this.stat],
-          this.stages,
-        );
+        globalScene.phaseManager.unshiftNew("StatStageChangePhase", {
+          battlerIndex: other.getBattlerIndex(),
+          changes: [{ stat: this.stat, stages: this.stages }],
```
Bertie690, line 750:
```suggestion
          changes: [{ stat, stages }],
```

src/data/abilities/ab-attrs.ts
```diff
@@ -745,38 +744,32 @@ export class PostDefendStatStageChangeAbAttr extends PostDefendAbAttr {
...
-      globalScene.phaseManager.unshiftNew(
-        "StatStageChangePhase",
-        (this.selfTarget ? pokemon : attacker).getBattlerIndex(),
-        this.selfTarget,
-        [this.stat],
-        this.stages,
-      );
+      globalScene.phaseManager.unshiftNew("StatStageChangePhase", {
+        battlerIndex: (this.selfTarget ? pokemon : attacker).getBattlerIndex(),
+        changes: [{ stat: this.stat, stages: this.stages }],
```
Bertie690, line 757:
```suggestion
        changes: [{ stat, stages }],
```

src/data/abilities/ab-attrs.ts
```diff
@@ -1017,13 +1008,11 @@ export class PostReceiveCritStatStageChangeAbAttr extends AbAttr {
 
   override apply({ simulated, pokemon }: PostMoveInteractionAbAttrParams): void {
     if (!simulated) {
```
Bertie690, line 1011:
```suggestion
    if (!simulated) {
      const { stat, stages } = this;

```

src/data/abilities/ab-attrs.ts
```diff
@@ -1017,13 +1008,11 @@ export class PostReceiveCritStatStageChangeAbAttr extends AbAttr {
...
-      globalScene.phaseManager.unshiftNew(
-        "StatStageChangePhase",
-        pokemon.getBattlerIndex(),
-        true,
-        [this.stat],
-        this.stages,
-      );
+      globalScene.phaseManager.unshiftNew("StatStageChangePhase", {
+        battlerIndex: pokemon.getBattlerIndex(),
+        changes: [{ stat: this.stat, stages: this.stages }],
```
Bertie690, line 1014:
```suggestion
        changes: [{ stat, stages }],
```

Bertie690 (changes requested):

src/data/abilities/ab-attrs.ts
Bertie690: @emdeann
github didn't let me add a comment too high up
Add this before L743:
```ts
    const { stat, stages } = this;
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,420 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+    this.options.changes = this.options.changes.filter(c => !cancelledStats.includes(c.stat));
+  }
 
-      return !cancelled.value;
+  /**
+   * Invoke abilities that may block or reflect a set of stat drops,
+   * pushing any blocked stats into {@linkcode cancelledStats}.
+   *
+   * @param pokemon - The Pokemon receiving the stat changes
+   * @param opponentPokemon - The Pokemon that caused the change, if not self-inflicted
```
DayKev:
```suggestion
   * @param opponentPokemon - The Pokemon that caused the change, if present; used to resolve the target for Mirror Armor's reflection effect.
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,420 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+   * @param changes - The negative stat changes to evaluate
+   * @param cancelledStats - An array containing each {@linkcode BattleStat} whose change was cancelled; modified by this method
```
DayKev:
```suggestion
   * @param cancelledStats - An array containing each {@linkcode BattleStat} whose change was cancelled; will be modified by this method
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,420 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+    if (whiteHerb) {
+      pokemon.loseHeldItem(whiteHerb);
+      globalScene.updateModifiers(this.player);
     }
   }
 
-  getStatStageChangeMessages(stats: readonly BattleStat[], stages: number, relStages: number[]): string[] {
-    const messages: string[] = [];
+  /**
+   * Play the rising stat change animation, depending on whether there were increases or decreases.
```
DayKev:
```suggestion
   * Play the stat change animation, depending on whether there were increases or decreases.
```

DayKev (commented):
Some of matt's comments got put in the wrong place by VSCode and/or GitHub, these are the actual suggestions.

src/phases/stat-stage-change-phase.ts
DayKev: Because there was no return type annotation for `start()` at the time.

Bertie690:
> Some of matt's comments got put in the wrong place by VSCode and/or GitHub, these are the actual suggestions.

I am never clicking that button again

src/data/arena-tag.ts
emdeann: This is very clear in what it's doing and both of these arrays are max 5 elements (99% of the time, <= 2 elements). If you had a suggestion on something clearer that would be fine, but performance isn't really a concern

test/tests/battler-tags/octolock.test.ts
emdeann: I don't disagree, but I don't want to remove it in this pr

test/tests/phases/stat-stage-change-phase.test.ts
emdeann: deleted it before you commented this :P

test/tests/phases/stat-stage-change-phase.test.ts
emdeann: Remember how you said ai is useful at making unit tests?

test/tests/phases/stat-stage-change-phase.test.ts
emdeann: I'm surprised vitest has such an aggressively nonatomic matcher

src/phases/stat-stage-change-phase.ts
emdeann: Seems more opaque than it is now, you're losing the information about "source"

src/phases/stat-stage-change-phase.ts
emdeann: For what?

src/phases/stat-stage-change-phase.ts
emdeann: It's a bit defensive, but there's no reason not to

src/phases/stat-stage-change-phase.ts
emdeann: Do you trust the battlerIndex/pid mess that PokemonPhase does in `getPokemon`? It probably doesn't matter anyway because the phase would end if the pokemon wasn't active, but while `getPokemon` can return undefined I want to have the check here for clarity

src/phases/stat-stage-change-phase.ts
emdeann: I think it's a carryover, but either way it's not so much a specific scenario as it's just much better to not blow up if some weirdness does happen

src/phases/stat-stage-change-phase.ts
emdeann: I think it's easier to reason about it this way rather than propogating it through phases. If it were a significant computation you're right though

src/phases/stat-stage-change-phase.ts
emdeann: You're right, that was the previous behavior and it saves having to redirect through all the tag and ability code, but it could be an unexpected issue if such an effect were added. Open to either way

src/phases/stat-stage-change-phase.ts
emdeann: I started changing it to a set then forgot lol

src/phases/stat-stage-change-phase.ts
emdeann: If you want to fix tweens then you need to suggest the change yourself, i have no idea how they work

src/phases/stat-stage-change-phase.ts
emdeann: If you think the other comment was unnecessary because it's obvious, this one certainly is

src/phases/stat-stage-change-phase.ts
Bertie690: I wasn't suggesting to change it _per se_, but we should probably document it in any case.

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,423 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+   * @param pokemon - The Pokemon receiving the stat changes
+   */
+  private removeCancelledChanges(pokemon: Pokemon): void {
+    if (this.selfTarget) {
+      return;
+    }
 
-    const filteredStats = this.stats.filter(stat => {
-      const cancelled = new BooleanHolder(false);
+    const negative = this.options.changes.filter(c => c.stages < 0);
```
Bertie690, line 107:
```suggestion
    // NB: This currently hardcodes the fact that abilities and field effects can _only_ respond to stat decreases and not increases.
    // If any effects that can cancel stat stage increases are added, this check should be removed.
    const negative = this.options.changes.filter(c => c.stages < 0);
```

src/phases/stat-stage-change-phase.ts
Bertie690: the phase's properties

github moved the comments off of their targets

src/phases/stat-stage-change-phase.ts
Bertie690: > It's a bit defensive, but there's no reason not to

I can think of several off the top of my head:
1. slight complexity increase
2. Is inconsistent with literally all other phases taking object params (do you see them cloning their parametesr object "for insurance"?)
3. Is wholly unnecessary assuming sensible caller behaviour (thereby implying that certain callsites are themselves irreparably _in_sensible)

I'm not against defensive programming in and of itself, but any instances of it should be justified by more than "no reason not to"

src/phases/stat-stage-change-phase.ts
Bertie690: I wholeheartedly disagree - the former was re-stating what the code was already doing (or a facet thereof), while this is documenting a notable restriction on this phase's call signature.

The way it currently stands, this effectively suggests that an outside consumer may want or need to handle all this pre-processing by themselves for some unknown reason, rather than its current state of being used exclusively by this class for recursive purposes.

src/phases/stat-stage-change-phase.ts
Bertie690: In that case, we should mention as such inside a comment.

src/phases/stat-stage-change-phase.ts
Bertie690: `sourceEffectType`?

test/tests/battler-tags/octolock.test.ts
Bertie690: got you covered :)

test/tests/phases/stat-stage-change-phase.test.ts
Bertie690: it's from jest

test/tests/phases/stat-stage-change-phase.test.ts
Bertie690: I said it was _passable_, not _good_.

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,423 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
 
-        pokemon.setStatStage(s, pokemon.getStatStage(s) + stages.value);
-      }
+  /**
+   * Apply the resolved changes, queue battle messages, trigger reactive abilities/items, and end the phase.
+   *
+   * @param pokemon - The Pokemon receiving the stat changes
+   * @param applied - The clamped per-stat deltas to apply
+   */
+  private applyStatChangesAndEnd(pokemon: Pokemon, applied: readonly StatChange[]): void {
```
Bertie690, line 223: this should ideally be async as well to avoid hanging promises

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,423 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
+  /**
+   * Compute the relative change for each requested change by clamping to `[-6, 6]`.
+   *
+   * @param pokemon - The Pokemon receiving the stat changes
+   * @returns A new array of {@linkcode StatChange}s
+   */
+  private getAppliedChanges(pokemon: Pokemon): StatChange[] {
+    return this.options.changes.map(({ stat, stages }) => {
+      const current = pokemon.getStatStage(stat);
+      const clamped = stages > 0 ? Math.min(current + stages, 6) : Math.max(current + stages, -6);
```
Bertie690:
```suggestion
      const clamped = Phaser.Math.Clamp(current + stages, -6, 6);
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -2,343 +2,423 @@ import { applyAbAttrs } from "#abilities/apply-ab-attrs";
...
-      }
+    if (cancelledStats.size < this.options.changes.length) {
+      this.checkAbilityProtection(pokemon, opponent, negative, cancelledStats);
+    }
 
-      // If one stat stage decrease is cancelled, simulate the rest of the applications
-      if (cancelled.value) {
-        simulate = true;
-      }
+    this.options.changes = this.options.changes.filter(c => !cancelledStats.has(c.stat));
```
Bertie690, line 129: Since the set should ostensibly be a subset of `this.options.changes`, we can presumably simplify this a bit:

```suggestion
    if (cancelledStats.size >= this.options.changes.length) {
      this.options.changes = [];
      return;
    }

    this.options.changes = this.options.changes.filter(c => !cancelledStats.has(c.stat));
    this.checkAbilityProtection(pokemon, opponent, negative, cancelledStats);
```

emdeann: This doesn't work because `checkAbilityProtection` doesn't modify `this.options.changes`, it modified the passed `cancelledStats` array. It doesn't actually save much anyway, besides a call to `filter` on a very small array

src/phases/stat-stage-change-phase.ts
emdeann: You'd have to elaborate more for me on what causes a hanging promise how it is now, I'm not really familiar with how it works

Bertie690: A hanging promise is any promise that isn't `await`-ed or `catch`-ed

src/phases/stat-stage-change-phase.ts
Bertie690: You move every single `onComplete` callback out of the tween and into the main function body in sequence, then replace any calls to `tweens.add` with `await playTween`.

```ts
scene.tweens.add({
  target: foo,
  onComplete(() => {
    console.log("1")
    scene.tweens.add({
      target: bar,
      onComplete(() => {console.log("2")},
    });
  }),
});

// becomes...

await playTween({
  target: foo,
});
console.log("1")
await playTween({
  target: bar,
});
console.log("2")
```

src/phases/stat-stage-change-phase.ts
Bertie690: You misunderstand - I am saying that the abilities should not add anything to the `cancelledStats` array that was not initially inside `this.options.changes`.

src/phases/stat-stage-change-phase.ts
emdeann: Seems like it would make more sense to catch then? I don't see the usefulness of making everything async because you're just kicking the can to a synchronous method regardless

src/phases/stat-stage-change-phase.ts
Bertie690: Then we should document it with a TODO that suggests it should be removed eventually

src/phases/stat-stage-change-phase.ts
Bertie690: > kicking the can to a synchronous method regardless

The ONLY reason we don't have proper async support in our code is Phaser is stuck on 2011 JS and doesn't even use ES6 classes (much less promises).
just humor me

emdeann: I mean, the proverbial can here is being kicked to `start`, which I can't imagine we ever want to make async, let alone do it

DayKev: Actually doing that is fine, see [`LoginPhase`](https://github.com/pagefaultgames/pokerogue/blob/beta/src/phases/login-phase.ts) or [`QuietFormChangePhase`](https://github.com/pagefaultgames/pokerogue/blob/beta/src/phases/quiet-form-change-phase.ts) for example.

`async`/`await` is just a more readable alternative to `method().then((...) => {...})`.

emdeann: I mean, I guess? It feels icky to ever call one of those methods async because the entire basis of the phase system is that it's synchronous (and you're overriding a sync method with an async method which can't be a good thing lol). The phase manager is obviously never awaiting a phase start, but I guess that doesn't matter anyway because that's a pretty loose system to begin with

DayKev: Well there's already a lot of phases running async code, they just use the older method (`async`/`await` is newer).

Also, since the phases handle transitioning to the next phase themselves (via `Phase#end`), it doesn't matter if they're async.

emdeann: That's true, my issue is more that even though it's not a huge issue with only one callsite, I don't like having a polymorphic method that has different return types as a principle. But that can be solved by just marking the base method async at some point

src/phases/stat-stage-change-phase.ts
emdeann: This isn't the same as other phases. I literally ran into an issue while working on this very PR that took me a minute to solve for the precise reason that this phase manipulates some options in place and that has potential to cause issues elsewhere. I'm not going to remove the free safety against those issues to try to match _internal_ behavior with other phases

src/phases/stat-stage-change-phase.ts
emdeann: You don't think it's incredibly, extremely obvious that a consumer shouldn't want to for some reason reimplement the entire preprocessing cycle of this phase? I suppose there's no problem with including a comment if that's the case, but I really don't see it

src/phases/stat-stage-change-phase.ts
emdeann: I don't find it necessary to add a comment to identify that a method which can return undefined, can return undefined. But if you really want to you can suggest it and I won't complain

src/phases/stat-stage-change-phase.ts
emdeann: That works equally well I suppose

src/phases/stat-stage-change-phase.ts
emdeann: Can you make your suggestion the full set of changes you'd like to make then? I'm not sure what you're getting at without that

also btw it would be useful if you unresolve conversations that you reply to, otherwise I won't see them

src/phases/stat-stage-change-phase.ts
Bertie690: no
~~I am paranoid about game freak~~

src/phases/stat-stage-change-phase.ts
Bertie690:
```suggestion
    // TODO: Change this once `getPokemon` is reworked to be sane
    this.selfTarget = options.sourcePokemon != null && options.sourcePokemon === this.getPokemon();
```

src/phases/stat-stage-change-phase.ts
Bertie690: @emdeann am away from laptop but if you could rename it that would be greatly appreciated

emdeann: I'll go through and fix this with conflicts and everything probably tomorrow

src/phases/stat-stage-change-phase.ts
emdeann: It's a little awkward because these were using `delay` for timing and it wasn't exact, so the tween to hide `statSprite` would seemingly start before the tween to animate it had ended. I split it into awaited tweens and it doesn't look different to me, but I think it's technically a behavior change.

Bertie690: Alr - I'll take a look at it in a later PR.

src/phases/stat-stage-change-phase.ts
Bertie690: i was going to do that anyways
i'll do it now if u want

src/phases/stat-stage-change-phase.ts
Bertie690: you know what it's fine as is
I was splitting hairs where it really doesn't matter lol

Bertie690 (dismissed):

src/phases/stat-stage-change-phase.ts
```diff
@@ -3,343 +3,426 @@ import { audioManager } from "#app/global-audio-manager";
...
 import i18next from "i18next";
 
-export type StatStageChangeCallback = (
-  target: Pokemon | null,
-  changed: BattleStat[],
-  relativeChanges: number[],
-) => void;
+export interface StatStageChangePhaseOptions {
+  battlerIndex: BattlerIndex | number;
+  changes: readonly StatChange[];
```
fabske0: Since you moved the callback to `@types` can we also move this interface

src/phases/stat-stage-change-phase.ts
```diff
@@ -3,343 +3,426 @@ import { audioManager } from "#app/global-audio-manager";
...
+   * is skipped because it was already performed by the phase that queued this one.
+   * @remarks
+   * Should not be passed by anything other than this phase.
+   */
+  processed?: boolean;
+}
 
-// TODO: Refactor this mess of a phase
+/**
+ * Phase responsible for resolving, animating, and applying one or more
```
fabske0:
```suggestion
 * Phase responsible for resolving, animating, and applying one or more stat changes.
 * They will be split into one phase for raises and one phase for drops.
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -3,343 +3,426 @@ import { audioManager } from "#app/global-audio-manager";
...
+  private isIncrease = false;
+
+  constructor(options: StatStageChangePhaseOptions) {
+    super(options.battlerIndex);
+
+    this.options = { ...options };
+    // TODO: Change this once `getPokemon`'s return type is fixed
+    this.selfTarget = options.sourcePokemon != null && options.sourcePokemon === this.getPokemon();
+    this.options.changes = deepCopy(options.changes).filter(c => c.stages !== 0); // Allow changes with 0 stages to be passed as no-ops
+    this.selfTarget = options.sourcePokemon != null && options.sourcePokemon === this.getPokemon();
```
fabske0: Why is `this.selfTarget` assigned twice?

src/data/arena-tag.ts
```diff
@@ -318,18 +321,28 @@ export class MistTag extends SerializableArenaTag {
   override apply(
     simulated: boolean,
     defender: Pokemon,
-    cancelled: BooleanHolder,
+    changes: readonly StatChange[],
+    cancelledStats: Set<BattleStat>,
     source: Pokemon | undefined,
```
fabske0, line 328: Please add `changes` to the doc comment and adjust `cancelled` to `cancelledStats` along with it's description

src/phases/stat-stage-change-phase.ts
```diff
@@ -3,343 +3,426 @@ import { audioManager } from "#app/global-audio-manager";
...
+   * Invoke abilities that may block or reflect a set of stat drops,
+   * adding any blocked stats to {@linkcode cancelledStats}.
+   *
+   * @param pokemon - The Pokemon receiving the stat changes
+   * @param opponentPokemon - The Pokemon that caused the change, if not self-inflicted
+   * @param changes - The negative stat changes to evaluate
+   * @param opponentPokemon - The Pokemon that caused the change, if present; used to resolve the target for Mirror Armor's reflection effect.
+   * @param cancelledStats - A set containing each {@linkcode BattleStat} whose change was cancelled; will be modified by this method
+   */
+  private checkAbilityProtection(
```
fabske0, line 141: Remove duplicate param
```suggestion
   *
   * @param pokemon - The Pokemon receiving the stat changes
   * @param opponentPokemon - The Pokemon that caused the change, if present; used to resolve the target for Mirror Armor's reflection effect.
   * @param changes - The negative stat changes to evaluate
   * @param cancelledStats - A set containing each {@linkcode BattleStat} whose change was cancelled; will be modified by this method
   */
  private checkAbilityProtection(
```

src/phases/stat-stage-change-phase.ts
```diff
@@ -3,343 +3,426 @@ import { audioManager } from "#app/global-audio-manager";
...
+    pokemon.disableMask();
+  }
+
+  /**
+   * Build a stat change message for a group of changes that share the same magnitude.
+   *
+   * @param changes - The changes described by this message (all sharing one {@linkcode StatChange.stages | stages} value)
+   * @param isIncrease - Whether the intended direction was positive (used
+   *   when the clamped change is `0` to distinguish "won't go any higher" from "won't go any lower")
+   * @returns The localised message string
```
fabske0, line 366:
```suggestion
   * @param changes - The changes described by this message (all sharing one {@linkcode StatChange.stages | stages} value)
   * @returns The localised message string
```

src/utils/stat-change.ts
```diff
@@ -0,0 +1,6 @@
...
+  return stats.map(stat => ({ stat, stages }));
```
fabske0, line 11: This is a very simple function, but could we still ge some docs

src/data/abilities/ab-attrs.ts
```diff
@@ -3038,19 +3014,16 @@ export class PreLeaveFieldRemoveSuppressAbilitiesSourceAbAttr extends PreLeaveFi
...
   /**
    * The source of the stat stage drop. May be omitted if the source of the stat drop is the user itself.
    *
    * @remarks
    * Currently, only used by {@linkcode ReflectStatStageChangeAbAttr} in order to reflect the stat stage change
    */
   source?: Pokemon | undefined;
-  /** Holder that will be set to true if the stat stage change should be cancelled due to the ability */
-  cancelled: BooleanHolder;
+  /** All cancelled stat changes; a subset of {@linkcode stats}. Any stat change which the ability cancels should be added to this array */
```
fabske0, line 3026:
```suggestion
  /** All cancelled stat changes; a subset of {@linkcode stats}. Any stat change which the ability cancels should be added to this set */
```

src/data/abilities/ab-attrs.ts
```diff
@@ -3303,10 +3282,10 @@ export class ConditionalUserFieldStatusEffectImmunityAbAttr extends UserFieldSta
 }
 
 export interface ConditionalUserFieldProtectStatAbAttrParams extends AbAttrBaseParams {
-  /** The stat being affected by the stat stage change */
-  stat: BattleStat;
-  /** Holds whether the stat stage change is prevented by the ability */
-  cancelled: BooleanHolder;
+  /** The stats being affected by the stat stage change */
+  changes: readonly StatChange[];
+  /** All cancelled sWat changes; a subset of {@linkcode stats}. Any stat change which the ability cancels should be added to this array */
```
fabske0:
```suggestion
  /** All cancelled stat changes; a subset of {@linkcode stats}. Any stat change which the ability cancels should be added to this set */
```

test/tests/phases/stat-stage-change-phase.test.ts
```diff
@@ -0,0 +1,200 @@
...
+  describe("basic application", () => {
+    it("should raise a single stat by the correct amount", async () => {
+      await game.classicMode.startBattle(SpeciesId.MIGHTYENA);
+      const player = game.field.getPlayerPokemon();
+
+      game.move.use(MoveId.SWORDS_DANCE);
+      await game.toNextTurn();
+
+      expect(player).toHaveStatStage(Stat.ATK, 2);
+      expect(player).toHaveStatStage(Stat.ATK, 2);
```
fabske0:
```suggestion
      expect(player).toHaveStatStage(Stat.ATK, 2);
```

fabske0 (commented):
I didn't test in game yet

src/phases/stat-stage-change-phase.ts
emdeann: Probably a suggestion application that got screwed up

Bertie690 (dismissed):

src/data/abilities/ab-attrs.ts
SirzBenjie: This is marked as resolved but still appears incorrect.

src/data/berry.ts
```diff
@@ -126,13 +133,20 @@ export function getBerryEffectFunc(berryType: BerryType): BerryEffectFunc {
...
-            "StatStageChangePhase",
-            consumer.getBattlerIndex(),
-            true,
-            [randStat],
-            stages.value,
-          );
+          if (berryPhase) {
+            const queuedChange = consumer.queuedBerryStatChanges.find(c => c.stat === randStat);
+            if (queuedChange == null) {
+              consumer.queuedBerryStatChanges.push({ stat: randStat, stages: stages.value });
```
SirzBenjie, line 142: Endless players are gonna love you with changes like this.

src/phases/stat-stage-change-phase.ts
```diff
@@ -3,343 +3,398 @@ import { audioManager } from "#app/global-audio-manager";
...
+
+    // On increase, show the red sprite located at ATK; on decrease, the blue sprite at SPD
+    const spriteColor = this.isIncrease ? Stat[Stat.ATK].toLowerCase() : Stat[Stat.SPD].toLowerCase();
+    const statSprite = globalScene.add.tileSprite(tileX, tileY, tileWidth, tileHeight, "battle_stats", spriteColor);
+    statSprite
+      .setPipeline(globalScene.fieldSpritePipeline)
+      .setAlpha(0)
+      .setScale(6)
+      .setOrigin(0.5, 1)
+      .setMask(new Phaser.Display.Masks.BitmapMask(globalScene, pokemon.maskSprite ?? undefined));
```
SirzBenjie, line 337: is `?? undefined` here because `pokemon.maskSprite` can be null? It just seems strange to see `undefined` on the RHS of `??`

SirzBenjie (dismissed):
looks great! I went through the whole thing.

src/phases/stat-stage-change-phase.ts
emdeann: That's correct, yeah. This was just code movement though, it's not something I changed in this PR

src/data/abilities/ab-attrs.ts
emdeann: I'm still curious how I did this.. W is a full 2 keys from t

fabske0 (approved):

damocleas (approved):

SirzBenjie (approved):

PR #7261, opened 2026-04-18
fix: prioritize localstorage for language detection

fabske0:

Prevents an issue where the language selection could be stored in the wrong place, causing it to be unable to be changed.

[Discord](https://discord.com/channels/1125469663833370665/1494829500528201901)

bug where the user had a `i18next` cookie, which took prio over the localstorage and prevented the user from changing languages

added a custom order for i18next to use to detect the language.
Now local storage will be prioritized with using the `navigator` second.
The [default](https://github.com/i18next/i18next-browser-languageDetector/blob/9efebe6ca0271c3797bc09b84babf1ba2d9b4dbb/src/index.js#L13) is prioritizing cookies over local storage

## How to test the changes?

add a cookie via devtools with the key `i18next` and as the value any other language, e.g. `de`

## Checklist


src/plugins/i18n.ts
```diff
@@ -216,6 +216,8 @@ await i18next
       defaultNS: "menu",
       detection: {
         lookupLocalStorage: "prLang",
+        caches: ["localStorage"],
```
DayKev, line 219:
```suggestion
```
That seems to be the [default](https://github.com/i18next/i18next-browser-languageDetector/blob/9efebe6ca0271c3797bc09b84babf1ba2d9b4dbb/src/index.js#L20), so I don't think specifying that will do anything.

DayKev (approved):

src/plugins/i18n.ts
fabske0: Since we don't know how the cookie got there, I added it in the hope no cookie will be made, but I can remove it

DayKev: True, I guess it doesn't matter.

emdeann (approved):

SirzBenjie (approved):

PR #7263, opened 2026-04-20
refactor(audio): Stream BGM with HTML5 Audio

emdeann:
Significantly lower memory usage from bgm, leading to reduced forced reloads, etc.
~Potentially a slightly more noticeable stutter at the loop point of songs.~ (Due to the follow-up PR there shouldn't be any change in looping behavior)

What we currently do is load full bgm tracks into memory and never clear them. Note that it would make a huge difference for memory just to clear tracks after use, even if they are not streamed. However, for loading times and general memory usage, I think it makes much more sense to stream them. This also allows composers liberty to make tracks however long they want - right now, if someone made even a 4-5+ minute song it would be pretty devastating for gameplay.

Added a new `BackgroundMusic` class as a simple wrapper for a Howlerjs HTML audio streamer. This doesn't really *need* to be its own class, but I wanted to pull as much complexity out of BattleScene as possible. 

Replaced `playSoundWithoutBgm` with `replaceBgmUntilEnd`. The former was a weird pseudo-bgm thing that would replace background music for the duration of a sound, but try to not actually consider that sound as background music (even though it was loaded with `loadBgm` lol). `loadBgm` is also removed as no bgm needs to be preloaded.

Made the non-music entries in `assests/bw/` proper sound effects instead of weird half-sfx half-bgm nonsense. See associated assets pr.

<details><summary>Memory Graph</summary>

image

</details>

## How to test the changes?
Play the game, wait at a battle for the loop point of the song. Note that this needs to be tested over network as well as locally to be complete

## Checklist

- [x] I have created an associated PR on the [assets](https://github.com/pagefaultgames/pokerogue-assets) repository
  - If so, include a link to the PR here: https://github.com/pagefaultgames/pokerogue-assets/pull/84


src/audio/background-music.ts
```diff
@@ -0,0 +1,104 @@
...
+  public pause(): void {
+    this.howl.pause();
+  }
+
+  public resume(): void {
+    this.howl.play();
+  }
+
+  public setVolume(value: number): void {
+    this.howl.volume(Math.max(0, Math.min(1, value)));
```
DayKev:
```suggestion
    this.howl.volume(Phaser.Math.Clamp(value, 0, 1));
```

src/audio/background-music.ts
```diff
@@ -0,0 +1,104 @@
...
+ * rather than being decoded into memory all at once. Considering single BGM tracks can exceed
+ * 100 MB decoded, this is extremely significant.
+ *
+ * It is separate because Phaser does not support using both WebAudio and HTML5 audio
+ * in the same project. There is indication that Phaser 4 will support this, at which time
+ * this class could be removed and replaced with Phaser's built in support.
+ */
+export class BackgroundMusic {
+  /** The key for the audio file */
+  readonly key: string;
```
DayKev:
```suggestion
  public readonly key: string;
```

src/battle-scene.ts
```diff
@@ -2480,24 +2446,15 @@ export class BattleScene extends SceneBase {
...
-      this.bgmResumeTimer = this.time.delayedCall(pauseDuration || fixedInt(sound.totalDuration * 1000), () => {
-        this.resumeBgm();
-        this.bgmResumeTimer = null;
-      });
-    }
-    return sound as AnySound;
+  public replaceBgmUntilEnd(bgmName: string, callback?: () => void): BackgroundMusic {
+    const tempBgm = new BackgroundMusic(bgmName, false, 0, () => {
+      callback?.();
+      this.bgm?.resume;
```
```suggestion
      this.bgm?.resume();
```

src/audio/background-music.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  private loopOnEnd(): void {
+    this.howl.stop();
+    this.howl.play("loop");
+  }
+
+  public play(volume?: number): void {
+    if (volume != null) {
+      this.howl.volume(volume);
```
fabske0:
```suggestion
      this.setVolume(volume);
```

src/battle-scene.ts
```diff
@@ -2310,128 +2309,105 @@ export class BattleScene extends SceneBase {
...
-    return false;
+    return this.bgm.fadeOut(duration, destroy);
   }
 
   /**
-   * Fades out current track for `delay` ms, then fades in new track.
-   * @param newBgmKey
-   * @param destroy
-   * @param delay
+   * Fade out the current bgm over `delay` ms, then start a new one.
```
Bertie690:
```suggestion
   * Fade out the current BGM track over `delay` ms, then start `newBgmKey` once it finishes.
```
Also, if you're going to change the doc comment, might as well document the parameters svp

src/audio/background-music.ts
```diff
@@ -0,0 +1,104 @@
+import { getCachedUrl } from "#utils/fetch-utils";
+import { Howl } from "howler";
+
+/**
+ * Stream background music with Howler.
+ *
+ * This class is separate from and unrelated to Phaser's sound system.
```
Bertie690:
```suggestion
 * Class representing a background music track being streamed from HTML5 Audio via Howler.
 * 
 * This class is separate from and unrelated to Phaser's sound system.
```

src/audio/background-music.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  public get isPlaying(): boolean {
+    return this.howl.playing();
+  }
+
+  public get isPaused(): boolean {
+    return !this.howl.playing() && (this.howl.seek() as number) > 0;
+  }
+
+  constructor(key: string, loop: boolean, loopPoint = 0, onEnd?: () => void) {
```
Bertie690: Does this have to use callbacks?
Also, the constructor is entirely undocumented

src/audio/background-music.ts
```diff
@@ -0,0 +1,104 @@
...
+  private readonly howl: Howl;
+  private destroyed = false;
+
+  public get isPlaying(): boolean {
+    return this.howl.playing();
+  }
+
+  public get isPaused(): boolean {
+    return !this.howl.playing() && (this.howl.seek() as number) > 0;
+  }
```
Bertie690, line 34: I'd like comments on these

```suggestion
  /** The underlying {@linkcode Howl} instance used to stream music. */
  private readonly howl: Howl;
  /** Whether this BGM has been evicted from memory. */
  private destroyed = false;

  /** @returns Whether this BGM is currently playing. */
  public get isPlaying(): boolean {
    return this.howl.playing();
  }

  /** @returns Whether this BGM is currently paused mid-playback. */
  public get isPaused(): boolean {
    return !this.howl.playing() && (this.howl.seek() as number) > 0;
  }
```

src/audio/background-music.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  public destroy(): void {
+    if (this.destroyed) {
+      return;
+    }
+    this.destroyed = true;
+    this.howl.unload();
+  }
+
+  public fadeOut(duration: number, destroy: boolean): boolean {
```
Bertie690: ... i hate the fact that this returns a boolean but if that's what callers expect i guess we can slap on a todo

src/audio/background-music.ts
```diff
@@ -0,0 +1,104 @@
...
+    if (!this.isPlaying) {
+      return false;
+    }
+
+    const current = this.howl.volume();
+    this.howl.fade(current, 0, duration);
```
Bertie690:
```suggestion
    const currentVolume = this.howl.volume();
    this.howl.fade(currentVolume, 0, duration);
```

src/audio/background-music.ts
```diff
@@ -0,0 +1,104 @@
...
+      onload: () => {
+        this.howl["_sprite"].loop = [loopPoint * 1000, this.howl.duration() * 1000 - loopPoint * 1000, true];
+      },
+    });
+
+    if (onEnd != null) {
+      this.howl.on("end", onEnd);
+    } else if (loop && loopPoint > 0) {
+      this.howl.on("end", this.loopOnEnd.bind(this));
+    }
```
Bertie690, line 57: @emdeann 
If looping is mutually exclusive with an `onEnd` callback, we should explicitly forbid callers from passing both at once (possibly via an overload).
Also, we should MENTION that they are incompatible in the constructor

src/battle-scene.ts
```diff
@@ -2479,24 +2445,15 @@ export class BattleScene extends SceneBase {
...
-      this.bgmResumeTimer.destroy();
-    }
-    if (resumeBgm) {
-      this.bgmResumeTimer = this.time.delayedCall(pauseDuration || fixedInt(sound.totalDuration * 1000), () => {
-        this.resumeBgm();
-        this.bgmResumeTimer = null;
-      });
-    }
-    return sound as AnySound;
+  public replaceBgmUntilEnd(bgmName: string, callback?: () => void): BackgroundMusic {
```
Bertie690: I'd like to see doc comments on this.

Also, why does this even need a callback argument? (what are callers using it for?)

Bertie690 (dismissed):
LGTM, aside from some minor housekeeping nitpicks.

src/audio/background-music.ts
emdeann: The previous system used callbacks but it seems like they're unused now, so I'm removing most of them. The constructor inherits the class doc though, at least on hover? Almost none of our codebase explicitly documents constructors

src/audio/background-music.ts
emdeann: Similar thing, I don't think it's *wrong* to return a boolean here, but it's unused so we can remove it. yagni and such

src/system/game-data.ts
```diff
@@ -1823,7 +1823,7 @@ export class GameData {
     if (!showMessage) {
       return true;
     }
-    globalScene.playSound("level_up_fanfare");
+    globalScene.playSound("bw/level_up_fanfare");
```
fabske0: image

src/loading-scene.ts
```diff
@@ -358,19 +358,10 @@ export class LoadingScene extends SceneBase {
       .loadSe("gacha_dial")
       .loadSe("gacha_running")
       .loadSe("gacha_dispense")
+      .loadSe("heal", "se", "heal.mp3")
+      .loadSe("item_fanfare", "se", "item_fanfare.mp3")
```
fabske0: image

src/data/mystery-encounters/encounters/the-winstrate-challenge-encounter.ts
```diff
@@ -164,7 +164,7 @@ async function spawnNextTrainerOrEndEncounter() {
     // Give 10x Voucher
     const newModifier = modifierTypes.VOUCHER_PREMIUM().newModifier();
     globalScene.addModifier(newModifier);
-    globalScene.playSound("item_fanfare");
+    globalScene.playSound("se/item_fanfare");
```
fabske0, line 167: image

emdeann:
@fabske0 I just put up the assets pr, you're seeing errors because I moved some stuff around there

fabske0:
After a few reloads there is no BGM playing and I get this in my console:
`background-music.ts:39 HTML5 Audio pool exhausted, returning potentially locked audio object.`

Seems to only be the title music. Going into starter select plays the `menu` bgm and starting the run also plays normally

emdeann:
@fabske0 See if these changes fix it for you

fabske0:
> @fabske0 See if these changes fix it for you

@emdeann Now as soon as i press something it starts playing. Just needs the tests fixed, otherwise LGTM

DayKev:
Btw this is all you need to do to fix the tests:

```patch
From bad13181659c9e3e4fe62c9e76a4a5eea52a9ecf Mon Sep 17 00:00:00 2001
From: NightKev <34855794+DayKev@users.noreply.github.com>
Date: Sun, 3 May 2026 21:20:29 -0700
Subject: [PATCH] Stub out `BattleScene#playBgm` and `#replaceBgmUntilEnd` in
 tests

---
 test/framework/game-wrapper.ts | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/test/framework/game-wrapper.ts b/test/framework/game-wrapper.ts
index 144a46f5df2..3053d30f648 100644
--- a/test/framework/game-wrapper.ts
+++ b/test/framework/game-wrapper.ts
@@ -108,6 +108,9 @@ export class GameWrapper {
       key: "",
     };
 
+    this.scene.playBgm = () => null;
+    this.scene.replaceBgmUntilEnd = () => null!;
+
     this.scene.cameras = {
       main: {
         setPostPipeline: () => null!,
-- 
2.54.0.windows.1

```

```
 Test Files  434 passed | 8 skipped (442)
      Tests  4243 passed | 8 skipped | 112 todo (4363)
   Start at  21:14:47
   Duration  148.57s (transform 97.18s, setup 263.35s, import 11.70s, tests 1914.50s, environment 917.71s)
```

src/battle-scene.ts
```diff
@@ -2310,129 +2307,105 @@ export class BattleScene extends SceneBase {
...
-   * @param delay
+   * Fade out the current BGM track over `delay` ms, then start `newBgmKey` once it finishes.
+   * @param newBgmKey - The key for the next track to start
+   * @param delay - The delay to use before starting the next track
    */
-  fadeAndSwitchBgm(newBgmKey: string, destroy = false, delay = 2000) {
-    this.fadeOutBgm(delay, destroy);
+  fadeAndSwitchBgm(newBgmKey?: string, delay = 2000): void {
+    this.fadeOutBgm(delay);
     this.time.delayedCall(delay, () => {
```
DayKev: Btw this will cause the fade out to ignore game speed while the delay before the next bgm starts playing will be modified by game speed.
Well, it was also an issue with the old version of the method.

src/battle-scene.ts
```diff
@@ -2310,129 +2307,105 @@ export class BattleScene extends SceneBase {
...
+
+    const loopPoint = resolvedName === this.arena?.bgm ? this.arena.bgmLoopPoint : this.getBgmLoopPoint(resolvedName);
+
+    const shouldFadeOut = fadeOutPrevious && this.bgm?.isPlaying;
+
+    if (shouldFadeOut) {
+      const fadeDuration = 500;
+      this.fadeOutBgm(fadeDuration);
+      this.time.delayedCall(fadeDuration + 250, () => {
+        this.playNewBgm(resolvedName, loop, loopPoint);
```
DayKev, line 2377: `fadeOutBgm()` doesn't respect game speed, but the delay before calling `playNewBgm()` will respect game speed.

src/battle-scene.ts
```diff
@@ -2479,24 +2442,18 @@ export class BattleScene extends SceneBase {
...
-    }
-    return sound as AnySound;
+  /**
+   * Replace the current BGM track with `bgmName`, then resume it after `bgmName` finishes.
+   * @param bgmName - The key for the replacement track
+   * @returns The newly-created {@linkcode BackgroundMusic} object
+   */
+  public replaceBgmUntilEnd(bgmName: string): BackgroundMusic {
+    const tempBgm = new BackgroundMusic(bgmName, false);
+    tempBgm.onEnd(() => this.bgm?.resume());
```
DayKev: Shouldn't this also call `.destroy()` on `tempBgm` after it finishes playing?

src/battle-scene.ts
```diff
@@ -3492,7 +3449,7 @@ export class BattleScene extends SceneBase {
       return;
     }
 
-    this.fadeOutBgm(fixedInt(2000), false);
+    this.fadeOutBgm(fixedInt(2000));
```
DayKev: The `BackgroundMusic` class (and Howl) don't support `FixedInt` objects, so calling `fadeOutBgm(fixedInt(2000))` is going to cause the duration to be ignored at best (which seems to be what happens when testing) or break things at worst.

src/audio/background-music.ts
```diff
@@ -0,0 +1,115 @@
...
+  constructor(key: string, loop: boolean, loopPoint = 0) {
+    this.key = key;
+    const url = getCachedUrl(`audio/bgm/${key}.mp3`);
+
+    this.howl = new Howl({
+      src: [url],
+      html5: true,
+      preload: true,
+      onload: () => {
+        this.howl["_sprite"].loop = [loopPoint * 1000, this.howl.duration() * 1000 - loopPoint * 1000, true];
```
DayKev, line 50: Presumably Howl expects the loop point to be in ms, hence the `* 1000`.

Why are you modifying the private `_sprite` field instead of using `sprite: {...}` in the config being passed in to the Howl constructor?

emdeann: `this.howl.duration()` isn't available at config time because the Howl hasn't loaded yet. I'm not very familiar with Howler but I didn't find a better way

DayKev (approved):

PR #7254, opened 2026-04-14
fix: Fix assorted errors

Bertie690:
among other things:
- Fixed most of the type errors
- Did a pass over the held items to fix errors and broken documentation comments

PR #7273, opened 2026-04-25
refactor: rework `HeldItem` into an attribute based system

Bertie690:
## The issue
@DayKev brought up a very good point during discussion of #7254 - the current system of `HeldItem`s directly applying effects (while very minimal infrastructure-wise) is extremely hard to maintain and work with.
Items like Multi Lens effectively have to hack together multiple effect handling with verbose code that was unable to be 100% safely typed (TypeScript's type system is unable to associate the value of `effect` with the shape of the passed `params` object).
Moreover, the process of creating a new item that uses 2 pre-existing attributes together is far more convoluted than it really ought to be, requiring one to either duplicate the code for both attributes, create extra `HeldItem` instances used solely for their `apply` functions, or other similar shenanigans.

## The Rework
`HeldItem` has been reworked to be a generic attribute holder similar to `Ability` and `Move`, except with strong types provided by the `HeldItemBuilder` builder class (inspired very loosely by `@commander-js/extra-typing`'s generic builder chain, albeit a bit simpler).

All existing item classes have been reworked into `HeldITemAttr`s.

Assorted cleanups

### DIFF
--- a/src/@types/common.ts
+++ b/src/@types/common.ts
@@ -7,11 +7,13 @@ export type { Constructor } from "type-fest";
 export type nil = null | undefined;
 
 /**
- * This removes the `| undefined` from `Map#get`'s return type.
+ * A Map that is known to have keys for every key that it can possibly have, and is thus guaranteed
+ * to always return a proper value instead of `undefined`.
  * @remarks
  * Used for maps where we know the entire structure at compile time
  * (but may sometimes only technically be populated at runtime).
  */
+// TODO: Move this to another file
 export interface DataMap<K, V> extends Map<K, V> {
   get(key: K): V;
 }
--- a/src/items/held-item-builder.ts
+++ b/src/items/held-item-builder.ts
@@ -145,11 +145,11 @@ export class HeldItemBuilder<Attrs extends HeldItemAttr = never, ConsumableEffec
    * Build a non-consumable {@linkcode HeldItem} with the stored attributes and flags.
    * @returns A fully-typed `HeldItem` with all registered attributes.
    */
-  public build(): HeldItem<Attrs["effect"]> {
+  public build(): HeldItem<Attrs> {
     // @ts-expect-error - TypeScript doesn't support friend classes, so this is the closest we can get to
     // ensuring `HeldItem` is only constructed by its corresponding builder.
     // NB: The type specifier here is required due to TS resolving this as `any`
-    const item: HeldItem<Attrs["effect"]> = new HeldItem({
+    const item: HeldItem<Attrs> = new HeldItem({
       type: this.id,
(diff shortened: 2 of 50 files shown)


src/items/held-item-builder.ts
```diff
@@ -0,0 +1,207 @@
...
+    this.addAttr(new attrType(...args));
+    return this;
+  }
+
+  /** Internal helper method to add an attribute to the builder. */
+  private addAttr(attr: HeldItemAttr): void {
+    (attr as Mutable<HeldItemAttr>)["type"] = this.id;
+    const { effect } = attr;
+    // bang is safe since the constructor initializes the map with all effects set to empty arrays
+    const existing = this.attrMap.get(effect)!;
```
DayKev, line 93: You could make `attrMap` a `DataMap` type instead of a regular `Map`, like with `allBiomes`. Then the `!` wouldn't be necessary.

DayKev (approved):
I don't see anything obviously wrong

src/items/held-item-builder.ts
Bertie690: ahhh right i forgot you added that.

src/@types/common.ts
```diff
@@ -7,11 +7,13 @@ export type { Constructor } from "type-fest";
 export type nil = null | undefined;
 
 /**
- * This removes the `| undefined` from `Map#get`'s return type.
+ * A Map that is known to have keys for every key that it can possibly have, and is thus guaranteed
```
DayKev, line 10: Should that be "values for every key"?

src/@types/common.ts
```diff
@@ -7,11 +7,13 @@ export type { Constructor } from "type-fest";
...
+ * to always return a proper value instead of `undefined`.
  * @remarks
  * Used for maps where we know the entire structure at compile time
  * (but may sometimes only technically be populated at runtime).
  */
+// TODO: Move this to another file
```
DayKev, line 16: Where should it go instead?

src/@types/common.ts
Bertie690: Oop

src/@types/common.ts
Bertie690: Its own file mayhaps?

DayKev: Why?

Bertie690: In hindsight, I was wrong.
(I removed the todo comment in a followup commit I pushed to the branch, so it's all ok now).

src/@types/common.ts
Bertie690: Fixed in a followup commit

PR #7295, opened 2026-05-06
refactor: Rework trainer items to an attribute-like system

Bertie690:
This PR builds upon the work laid out in #7273 by expanding the attribute treatment towards the `TrainerItem` class.

This is a significantly larger rewrite than the former (due in part to the fact that trainer items were never made generic nor strongly typed to begin with).

### DIFF
--- a/src/data/status-effect.ts
+++ b/src/data/status-effect.ts
@@ -98,6 +98,7 @@ export function getStatusEffectDescriptor(statusEffect: StatusEffect): string {
     return "";
   }
   const i18nKey = `${getStatusEffectMessageKey(statusEffect)}.description` as ParseKeys;
+  // TODO: Use i18next key nesting instead of whatever the fuck this is
   return i18next.t(i18nKey);
 }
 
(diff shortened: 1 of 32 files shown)


src/data/status-effect.ts
```diff
@@ -98,6 +98,7 @@ export function getStatusEffectDescriptor(statusEffect: StatusEffect): string {
     return "";
   }
   const i18nKey = `${getStatusEffectMessageKey(statusEffect)}.description` as ParseKeys;
+  // TODO: Use i18next key nesting instead of whatever the fuck this is
```
SirzBenjie:
```suggestion
  // TODO: Use i18next key nesting instead
```

SirzBenjie (changes requested):

SirzBenjie (approved):

src/data/status-effect.ts
SirzBenjie: Probably shouldn't have swears in our repo 😅

PR #7302, opened 2026-05-10
refactor: Rewrite script files from JS to TS

Bertie690:
N/A

Previously, we wrote all our script files in pure JS.

However, JSDoc type assertions and a large inability to import basically anything from the main repo made this undesireable long-term.
Converted all the scripts into plain TS files, compiled and ran with `vite-node` to enable parsing of TSConfig paths. (This does slightly lengthen the startup time, but the improved DX is worth it.)

Added a `tsconfig.json` to the script directory to enable extra subpath aliases for script-specific folders. This allows us to keep our script files organized without having to use relative imports or allowing said path aliases to bleed into the main repo.

Since the script files' tsconfig now includes the entire main repo alongside it (to allow referencing of main repo types and functions), the command to typecheck the script files has been renamed to `typecheck:all` since it typechecks the entire codebase instead of just the scripts folder.

> [!IMPORTANT]
> **We cannot change the TSConfig settings for our script files to differ from the main repo**.
> Allowing our scripts to reference main repo types and functions causes TS to transitively typecheck every single file being imported, thus defeating the point of having different settings (since they'd have to be compatible with the main repo's settings anyway).
> Attempting to prevent this re-typechecking while still allowing access to shared helpers is similarly infeasible, requiring us to isolate _every single import_ our script files would access to prevent exponential blowup (INCLUDING type imports used solely in documentation comments).

## Screenshots/Videos
N/A
## How to test the changes?
Run the script files to ensure they don't break.

`pnpm exec tsgo --listFilesOnly | grep "scripts"` to ensure that the main repo's TSConfig excludes all script files

## Checklist


tsconfig.json
```diff
@@ -48,6 +49,10 @@
     "useUnknownInCatchVariables": false,
     "rootDir": ".",
     "outDir": "./build",
+    // Note: We actively CANNOT UNDER ANY CIRCUMSTANCES enable declaration emit, as main repo code uses
+    // mixins for classes with private/protected members (causing TS4094 errors).
+    // (This also means we cannot use `composite` or project references in any capacity.)
+    // See: https://github.com/microsoft/TypeScript/issues/17744
```
DayKev, line 54: I mean sure, but we don't build with TS regardless because we build with Vite (which has multiple build plugins that we use as well).

.github/scripts/changelog-reader/main.mts
DayKev, line 1: Why didn't you convert these to `.mts`?
Actually, in addition to converting this script directory to `.mts` it should be moved to `.github/scripts/`.
Also, the `.github` files should have their own stricter tsconfig in the `.github` directory since they should never import from or export to directories outside of `.github` (the existing tsconfig file isn't even downloaded when the `.github` scripts are run) so there shouldn't be any issues with doing that, unlike with the `scripts` directory which has some files that import from `src`.
Another depcruise rule should be added to enforce this.
There's a single function imported in this from the scripts util file outside of this directory that would need to be duplicated (so `.github` files don't cross outside that directory as mentioned above), but copying just a single function is fine.

scripts/daily-seed/prompts/general.ts
```diff
@@ -5,49 +5,45 @@
...
+import { BiomeId } from "#enums/biome-id";
+import { BiomePoolTier as BIOME_POOL_TIERS } from "#enums/biome-pool-tier";
+import { Challenges } from "#enums/challenges";
+import { MysteryEncounterType } from "#enums/mystery-encounter-type";
+import { getEnumKeys } from "#utils/enums";
+import { toTitleCase, toUpperSnakeCase } from "#utils/strings";
 import { confirm, input, number, search, select } from "@inquirer/prompts";
 import { Ajv } from "ajv";
 import chalk from "chalk";
 import customDailyRunSchema from "../../../src/data/daily-seed/schema.json" with { type: "json" };
```
DayKev:
```suggestion
import customDailyRunSchema from "../../../src/data/daily-seed/schema.json";
```

image

scripts/daily-seed/prompts/general.ts
```diff
@@ -87,29 +83,29 @@ export async function promptLuck() {
...
   const biomeName = await search({
     message: "Please enter the starting biome to set.",
     source: term => {
-      const biomes = Object.keys(BIOMES).map(toTitleCase);
+      const biomes = Object.keys(BiomeId).map(toTitleCase);
       if (!term) {
         return biomes;
       }
-      return biomes.filter(id => id.toLowerCase().includes(term.toLowerCase()));
+      return ["1", "2"];
```
DayKev: Huh? Did you forget to revert the temporary testing code when you were making https://github.com/SBoudrias/Inquirer.js/issues/2101?

scripts/daily-seed/prompts/pokemon.ts
```diff
@@ -5,42 +5,45 @@
...
-import { MOVES } from "../../enums/moves.js";
-import { NATURES } from "../../enums/natures.js";
-import { SPECIES } from "../../enums/species.js";
-import { toTitleCase, toUpperSnakeCase } from "../../helpers/casing.js";
 
-/**
- * @typedef {0 | 1 | 2} Variant
- */
+// NB: We cannot import this from #sprites/variant as that file imports `Pokemon` (and 90% of `src` along with it)
+export type Variant = 0 | 1 | 2;
```
DayKev: Wasn't that only an issue when trying to make the more strict `tsconfig.json`, since now it's typechecking everything anyway?

scripts/tsconfig.json
```diff
@@ -0,0 +1,71 @@
...
+        "../src/ui/settings/*.ts",
+        "../src/ui/utils/*.ts",
+        "../src/ui/*.ts"
+      ],
+      "#data/*": ["../src/data/pokemon-forms/*.ts", "../src/data/pokemon/*.ts", "../src/data/*.ts"],
+      "#test/*": ["../test/*.ts"],
+      "#app/*": ["../src/*.ts"],
+
+      // new subpath imports for script subfolders
+      "#utils/*": ["../src/utils/*.ts", "./utils/*.ts"], // add mapping for scripting utilities
```
DayKev: Not sure I like `src` imports being mixed with `scripts` imports. This is the one case where we care which directory an import comes from.

scripts/utils/process.ts
```diff
@@ -10,7 +10,7 @@ import { spawnSync } from "node:child_process";
 /**
  * Check if a given command exists in the system's `PATH`.
  * Equivalent to the `which`/`where` commands on Unix/Windows.
- * @param {string} command - The command to check
+ * @param command - The command to check
  * @returns {boolean} Whether the command exists.
```
DayKev:
```suggestion
 * @returns Whether the command exists.
```

tsconfig.json
```diff
@@ -94,7 +99,17 @@
       "#app/*": ["./src/*.ts"]
     }
   },
-  // Exclude checking for script JS files as those are covered by the folder's `jsconfig.json`
-  "include": ["**/*.ts", "**/*.d.ts", "**/*.mts", ".github/**/*.mts"],
-  "exclude": ["dist", "node_modules"]
+  // Exclude checking for script files as those are covered by the folder's `tsconfig.json`
+  // We do not use explicit `exclude` paths to allow explicitly including selected scripts files
```
DayKev:
```
// We do not use explicit `exclude` paths to allow explicitly including selected scripts files
```
What does this mean?

tsconfig.json
```diff
@@ -24,14 +24,15 @@
...
-    "types": ["@commander-js/extra-typings"],
     // Skip typechecking `.d.ts` files for a ~20% speed boost
     // Won't be necessary if https://github.com/microsoft/TypeScript/issues/30511 is ever resolved
     "skipLibCheck": true,
     "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "esModuleInterop": true,
     "sourceMap": false,
+    "allowJs": true,
+    "checkJs": true,
```
DayKev: What's the point of enabling these here? Shouldn't these be in the `scripts/` directory's `tsconfig.json` (though I don't see why it's added at all)?
No `.js`/`.mjs` files were added to the `"includes"` so I don't think this actually does anything regardless of which file it's put in, though the whole point is not having any of those anymore so...

tsconfig.json
```diff
@@ -24,14 +24,15 @@
       "ESNext.Iterator",
       "ESNext.Collection"
     ],
-    "types": ["@commander-js/extra-typings"],
```
DayKev, line 27: You didn't put this anywhere else, is that intentional?

DayKev (changes requested):

DayKev:
Overview of the [PR review commit](https://github.com/pagefaultgames/pokerogue/pull/7302/commits/d8ec8aefa800574797051d9e11bcae8ee281cea7) changes:

- Fixed `.dependency-cruiser.cjs` (`"test"`/`"scripts"` needed to be `"^test"`/`"^scripts"`)
- Fixed the import in `global.d.ts` (aka reverted it being changed).
  Somehow this happened:
  ```diff
  - import type { PreventHoverExpansion } from "#types/type-helpers";
  + import type { PreventHoverExpansion } from "#types/prevent-hover-expansion";
  ```
  Which... how???
- Added `vite-node`/`.ts` to missed `package.json` commands and removed `pnpm dlx` from `sample-movesets` since `vite-node` was added as a dev dep
- Removed `with { type: "json" }` from import in `scripts/daily-seed/prompts/general.ts` since it's not valid with our `tsconfig` settings (error message shown in the image in [this comment](https://github.com/pagefaultgames/pokerogue/pull/7302#discussion_r3214236141))
- Changed `moveset: number[]` to `moveset: MoveId[]` in `scripts/daily-seed/prompts/pokemon.ts` for more accurate typing
- Added a biome suppression comment to `decrypt-save.ts` on the `for (const k in keys)` (though it should probably be converted to `for (let k = 0; k < keys.length; k++)`)
- Reverted the removal of `pool: "threads"` from the Vitest invocation in `scripts/profiling/profiling.ts` (why did you remove this? it's still necessary)
- Added missed types in `scripts/utils/process.ts` and removed the leftover JSDoc type
- Re-ordered the subpath imports in the new `tsconfig.json` so that more specific paths are before less specific paths (TS parses the list in definition order until it finds a match, meaning if a less specific path matches then auto-imports will use that instead of the more specific path that comes after it)
- Removed obsolete `*.js`/`.cjs`/`.mjs` excludes from `tsconfig.json` files (there are no js files in the includes arrays)
- Removed `*.cts` from `includes` in `tsconfig.json` files; we don't have any of those nor should we ever have any
- Removed redundant entries from `includes` in `tsconfig.json` files (`*.d.ts` is unnecessary, `*.ts` already covers that)
- Changed the `*.cjs` in `jsconfig.json` to `.dependency-cruiser.cjs` since that's the only `.cjs` file we'll ever have

---

Also [added a Biome rule](https://github.com/pagefaultgames/pokerogue/pull/7302/commits/6f28942b532125d28c81186d768319ac0825e57f) to prevent imports from `#test` in `src` so that editors will show errors if someone does it. Depcruiser parses the subpath imports making it more accurate, so it wasn't removed.

Bertie690:
> Which... how???

I initially tried to enforce separation of imports to prevent duplicate typechecking.
i guess i forgot to revert that one file...

tsconfig.json
Bertie690: No you misunderstand - enabling project references requires .d.ts generation.
(TS would still be relegated to a typechecking role.)

.github/scripts/changelog-reader/main.mts
Bertie690: I asked fabi and rhey said it was ok to keep it there 
_shrug_

scripts/daily-seed/prompts/general.ts
Bertie690: 💀

scripts/daily-seed/prompts/pokemon.ts
Bertie690: Oh true dat

scripts/tsconfig.json
Bertie690: so do you want a separate path import JUST for script utilities?

tsconfig.json
Bertie690: Not blanket `exclude`-ing the entire scripts dir so we can check boilerplates under the normal config

tsconfig.json
Bertie690: Yes.
It was useless (`commander-js/extra-typings` does NOT change the types of the normal package, contrary to initial belief; we still gotta import it regardless)

tsconfig.json
DayKev: That would be pointless though, the `scripts/` directory isn't in includes so adding it to excludes would do literally nothing.

image

- https://www.typescriptlang.org/tsconfig/#include

Bertie690: I know what excludes does.
We previously used `**/*.ts` to include every TS file, which is the context in which the comment was written

Will clarify a bit

DayKev: Yes but you removed that, *and* added that comment at the same time.

.github/scripts/changelog-reader/main.mts
fabske0: Since this only works with a GH token, it is basically GH exclusive unless you manually add one every time you want to use it manually.

scripts/tsconfig.json
fabske0: > so do you want a separate path import JUST for script utilities?

i think that would be good

Bertie690: _sigh_ OK

scripts/create-test/create-test.ts
```diff
@@ -0,0 +1,118 @@
...
+import { writeFileSafe } from "#utils/file";
+import { toKebabCase, toTitleCase } from "#utils/strings";
+import fs from "node:fs";
+import { join } from "node:path";
+import { Command, InvalidArgumentError } from "@commander-js/extra-typings";
+import chalk from "chalk";
+
+// #region Constants
+
+const version = "2.1.0";
```
fabske0: The version in the deleted js file was `2.1.1`.
Also should this PR bump the versions of all the scripts?

scripts/daily-seed/main.ts
```diff
@@ -184,7 +151,7 @@ async function handleAnswer(answer) {
   await promptOptions();
 }
 
-const OUTFILE_ALIASES = /** @type {const} */ (["-o", "--outfile", "--outFile"]);
+const OUTFILE_ALIASES = ["-o", "--outfile", "--outFile"] as const;
 
 /**
  * @returns {Promise<void>}
```
fabske0:
```suggestion
/**
 * Finish the daily seed configuration process.
```

scripts/daily-seed/main.ts
```diff
@@ -208,9 +175,9 @@ async function finish() {
 
 /**
  * Write the seed config to a file.
- * @param {string} outFile
+ * @param outFile
```
fabske0:
```suggestion
 * @param outFile - The output file path
```

scripts/create-test/create-test.ts
Bertie690: Oops.

As for version bumping, I'd say no since this doesn't change any functionality at all (and is 90% copy-pasting code).

.github/scripts/changelog-reader/main.mts
DayKev: Please convert this to TS as well and make the GH actions tsconfig for this and the PR title lint script (it currently just uses default tsconfig settings when run by GitHub, while locally being processed by the tsconfig in the root dir; which is a problem since those settings are not the same, and it might as well use the more strict settings always).

I guess you can remove the section that prints the changelog locally (L76-L80 of `main.mjs` atm) since it's not necessary anymore, rather than copy the function from the scripts utils file.

scripts/daily-seed/prompts/pokemon.ts
DayKev: Okay well then you should be able to remove this re-definition and import the original.

tsconfig.json
DayKev: You didn't answer this btw.

.github/scripts/changelog-reader/main.mts
Bertie690: I might have forgotten to push

DayKev (changes requested):
See new comments in previous comment threads.

tsconfig.json
Bertie690: Forgor to push - will fix

Bertie690:
@DayKev i forgor to push

.github/scripts/changelog-reader/main.mts
fabske0:
```suggestion
/**
 * Load the configuration from the environment.
 * @returns Whether the config was loaded successfully.
 */
async function loadConfig(): Promise<boolean> {
```

scripts/create-test/create-test.ts
fabske0: I'd argue it should be bumped at least for the ones where you switched to `commander`

src/ui/handlers/test-dialogue-ui-handler.ts
```diff
@@ -6,6 +6,7 @@ import { FormModalUiHandler } from "#ui/form-modal-ui-handler";
 import type { ModalConfig } from "#ui/modal-ui-handler";
 import i18next from "i18next";
 
+// TODO: This is unused, remove it
```
fabske0: This is not unused

DayKev:
```suggestion
/** Handler for the i18n testing menu used by the localization team. */
```

biome.jsonc
```diff
@@ -381,6 +381,27 @@
           }
         }
       }
+    },
+    // Ban imports from test files in src
```
DayKev: This comment is kinda unnecessary. The `"message"` key / the lint rule itself already explains it.

DayKev (dismissed):

tsconfig.json
```diff
@@ -90,11 +95,17 @@
...
       "#app/*": ["./src/*.ts"]
     }
   },
-  // Exclude checking for script JS files as those are covered by the folder's `jsconfig.json`
-  "include": ["**/*.ts", "**/*.d.ts", "**/*.mts", ".github/**/*.mts"],
+  // We cannot use a blanket "**/*.ts" include here excluding `scripts` because it would bar us from checking any boilerplate files,
+  // which should be valid under _main repo_ settings.
+  // (`scripts` has its own TSConfig with extra subpath aliases, so attempting to typecheck them would cause errors about missing modules.)
+  // We also don't add the `.github/scripts/` directory here as those files
+  // need different settings to ensure they can be run by Node without issue.
```
DayKev:
```suggestion
  // need different settings to ensure they can be run by Node without issues.
```

DayKev (dismissed):

fabske0 (dismissed):

DayKev (approved):

PR #7303, opened 2026-05-11
fix(audio): Switch bgm to webaudio with proper eviction

emdeann:
BGM behaves properly, especially on mobile

Mobile devices are extremely aggressive with HTML5 audio and tend to take over the device's media player, which is not an acceptable behavior for the game. It is somewhat fixable, but the differences between platforms are too large to do it in a satisfactory way

I kept the BGM class, it's now a very thin wrapper but keeping it allows minimizing other changes in this PR. It's not completely unreasonable as it still serves to take some complexity away from the battlescene class. Outside of that, I modified the bgm pipeline to load/evict from phaser and removed Howler.

Note that BGM tracks will take a bit to load - this is unavoidable when downloading the entire file at once. It was less noticeable before because we preloaded 8 or so tracks, but for obvious reasons we're not going to keep doing that.

The effect memory-wise is that the usage stays relatively stable (although it does still slowly leak memory over time). There is a spike whenever a new BGM is loaded, which is generally shortly resolved by GC. Note that it's possible that certain devices/browsers decide to GC less often, but we cannot control that.

I've since updated the BGM class a bit to make it as easy to use as possible considering not-yet-loaded tracks. As requested by Kev I took the opportunity to move audio code into a dedicated manager, so that's where the bloat comes from.

## How to test the changes?
Play runs!

## Checklist


emdeann:
Gonna mark this as draft for a little while because I want to fix up some of the async loading issues. Wasn't an issue before because the HTML5 audio object is created essentially instantly

src/audio-manager.ts
```diff
@@ -0,0 +1,466 @@
...
+    };
+  }
+
+  /**
+   * Get the effective volume for a given setting
+   * (the product of that volume and the main volume).
+   * @param setting
+   * @returns
+   */
+  public getVolume(setting: VolumeSetting): number {
```
fabske0, line 40:
```suggestion
  /**
   * Get the effective volume for a given setting
   * (the product of that volume and the main volume).
   * @param setting - The {@linkcode VolumeSetting} to get the volume for
   * @returns The effective volume for the given setting (between 0 and 1)
   */
  public getVolume(setting: VolumeSetting): number {
```

src/audio-manager.ts
```diff
@@ -0,0 +1,466 @@
...
+  }
+
+  /**
+   * Stops the previously playing bgm (if it exists) and starts playing a new bgm.
+   * @remarks
+   * Helper function used by {@linkcode BattleScene.playBgm | playBgm}.
+   * @param bgmName - The bgm to play
+   * @param loop - Whether to loop the bgm
+   * @param loopPoint - The starting point of the loop, in seconds
+   */
```
fabske0:
```suggestion
  /**
   * Stops the previously playing bgm (if it exists) and starts playing a new bgm.
   * @param bgmName - The bgm to play
   * @param loop - Whether to loop the bgm
   * @param loopPoint - The starting point of the loop, in seconds
   * @remarks
   * Helper function used by {@linkcode BattleScene.playBgm | playBgm}.
  */
```

src/audio-manager.ts
```diff
@@ -0,0 +1,466 @@
...
+
+    return tempBgm;
+  }
+
+  /**
+   * Plays a sound effect (such as a Pokemon cry, UI cursor sfx, etc)
+   * @param sound - The sound effect to play
+   * @param config - (Optional) A `Phaser` {@linkcode Phaser.Types.Sound.SoundConfig | SoundConfig}
+   * or {@linkcode Phaser.Types.Sound.SoundMarker | SoundMarker} object
+   * @returns
```
fabske0:
```suggestion
   * @returns The sound object
```

src/battle-scene.ts
```diff
@@ -2313,442 +2305,6 @@ export class BattleScene extends SceneBase {
   // #region Audio
   // TODO: move audio-related code out of `BattleScene`
```
fabske0:
```suggestion
```

src/battle-scene.ts
```diff
@@ -2313,442 +2305,6 @@ export class BattleScene extends SceneBase {
...
-      case "battle_rival_3_afd": //Andr06 - Final N Battle Remix (AFD)
-        return 49.147;
-      case "battle_trainer_afd": //Andr06 - PokéRogue Trainer Remix (AFD)
-        return 13.686;
-    }
-
-    return 0;
-  }
-
   // #endregion
```
fabske0:
```suggestion
```

fabske0 (dismissed):

damocleas (dismissed):

src/audio-manager.ts
```diff
@@ -0,0 +1,466 @@
...
+      }
+    }
+  }
+
+  /**
+   * Fades out the current bgm over `duration` ms.
+   * @param duration - (Default `500`) The amount of time the fade out should take place over, in ms
+   */
+  public fadeOutBgm(duration = 500): void {
+    this.currentBgm?.fadeOut(duration);
```
DayKev:
```suggestion
   * @param duration - (Default `500`) The amount of time the fade out should take place over, in ms
   * @param fixed - (Default `false`) Whether the duration should ignore game speed
   */
  public fadeOutBgm(duration = 500, fixed = false): void {
    this.currentBgm?.fadeOut(duration, fixed);
```

src/audio-manager.ts
```diff
@@ -0,0 +1,466 @@
...
+    }
+
+    const loopPoint =
+      resolvedName === globalScene.arena?.bgm ? globalScene.arena.bgmLoopPoint : this.getBgmLoopPoint(resolvedName);
+
+    const shouldFadeOut = fadeOutPrevious && this.currentBgm?.isPlaying;
+
+    if (shouldFadeOut) {
+      const fadeDuration = 500;
+      this.fadeOutBgm(fadeDuration);
```
DayKev:
```suggestion
      this.fadeOutBgm(fadeDuration, true);
```

src/audio-manager.ts
```diff
@@ -0,0 +1,466 @@
...
+    this.currentBgm?.fadeOut(duration);
+  }
+
+  /**
+   * Fade out the current BGM track over `delay` ms, then start `newBgmKey` once it finishes.
+   * @param newBgmKey - (Optional) The key for the next track to start
+   * @param delay - (Default `2000`) The delay to use before starting the next track
+   */
+  public fadeAndSwitchBgm(newBgmKey?: string, delay = 2000): void {
+    this.fadeOutBgm(delay);
```
DayKev:
```suggestion
    this.fadeOutBgm(delay, true);
```

src/battle-scene.ts
```diff
@@ -3567,7 +3118,7 @@ export class BattleScene extends SceneBase {
       return;
     }
 
-    this.fadeOutBgm(2000);
+    audioManager.fadeOutBgm(2000);
```
```suggestion
    audioManager.fadeOutBgm(2000, true);
```

src/phases/egg-hatch-phase.ts
```diff
@@ -311,7 +312,7 @@ export class EggHatchPhase extends Phase {
     this.evolutionBgm?.fadeOut(100);
```
DayKev:
```suggestion
    this.evolutionBgm?.fadeOut(100, true);
```

DayKev (commented):
The suggestions restore the usage of `FixedInt`s from before the previous PR (since `howl` didn't take game speed into account and so they were removed in it).

DayKev (dismissed):

DayKev (dismissed):
Uh, apparently the volume controls are broken. Only the master volume does anything, the other settings are ignored and treated as always being at 100%.

fabske0 (changes requested):
LGTM, but `audio-manager` (and maybe the `global-audio-manager`) should be moved to `src/audio`

DayKev:
> LGTM, but `audio-manager` (and maybe the `global-audio-manager`) should be moved to `src/audio`

Moved `audio-manager`, but I don't think the global audio manager file should go in there. Maybe in a `globals` directory along with the global scene file, etc? In a different PR though.

fabske0:
> > LGTM, but `audio-manager` (and maybe the `global-audio-manager`) should be moved to `src/audio`
> 
> Moved `audio-manager`, but I don't think the global audio manager file should go in there. Maybe in a `globals` directory along with the global scene file, etc? In a different PR though.

I'm fine with a `globals` dir

fabske0 (dismissed):

DayKev (dismissed):
Found a bug: transition from wave 4 to wave 5 (for example), the trainer battle music never starts, instead looping the intro forever. Loading wave 5 from save causes it to transition into the battle music the same as before the BGM changes.

The issue starts with the previous BGM PR (7315352e858d8de31401097d28418dde3f239021); switching to the commit before it was merged (c0fe96681a32606ab98f93d67a2d4befc2867b46) and the music transitions from intro to battle (though not necessarily properly, given it doesn't wait for the battle to start before swapping to the battle music; when I would expect it to keep playing the intro music until you dismiss the "X would like to battle!" message, at which point it should *then* start playing the battle music).

Edit: Listening closer, I think it's (sometimes?) playing the BGMs in the wrong order (it starts with the battle bgm, and then transitions into the intro bgm).

Edit 2: Playing at 2x speed instead of 5x, the music transitions normally; starting with intro bgm, then into battle bgm. So maybe there's a timing issue.

Ideally both bugs can be fixed, but at least it should have the same behavior as it used to.

src/audio/audio-manager.ts
```diff
@@ -0,0 +1,450 @@
...
+    const resolvedName = timedEventManager.getEventBgmReplacement(
+      bgmName ?? globalScene.currentBattle?.getBgmOverride() ?? globalScene.arena?.bgm,
+    );
+
+    if (!resolvedName) {
+      return null;
+    }
+
+    if (this.currentBgm && this.currentBgm.key === resolvedName) {
+      this.currentBgm.play(this.getVolume(VolumeSetting.BGM)); // nop if already playing
```
DayKev:
```suggestion
      this.currentBgm.play(this.getVolume(VolumeSetting.BGM)); // no-op if already playing
```

DayKev (approved):

src/audio/audio-manager.ts
emdeann: I'm in architecture rn and we use the `nop` instruction all the time i'm rotted

DayKev: lol, yeah I figured it was a shortening used somewhere.

PR #7307, opened 2026-05-17
feat(fusion): Pokémon Infinite Fusion

Eddydpyl:
- DNA Splicer fusions unlock the resulting pair as a brand-new starter, complete with its own dex entry, IVs, abilities, nature, moveset, eggs, ribbons, and classic win count — tracked separately from the vanilla starters.
- Fusion sprites are loaded from a local Pokemon Infinite Fusion installation, with autogenerated palette-blended sprites as a fallback for pairs the artist set doesn't cover. Shiny variants are baked from the IF palette swap.
- Randomized trainer fusions, rare on unnamed trainers and more common on named ones. Uses pokemons with sprites.
- A settings tab that exposes several knobs: Stat formula (Infinite Fusion / PokéRogue / Maximum), Shiny recolor strength, Sprite source priority (Artist first / Autogen first / Artist only) and Trainer Fusion (Default / None / Always).
  
Just thought it'd be cool, plus this approach saves us from attribution headaches: The user still has to download IF and its sprites from official sources and have them on their own computer. As it is though, I probably wouldn't merge this myself to the main game. This is just a (working) proof of concept that should probably be gated behind a master toggle.

[AI generated] The work spans 8 commits (c5679de → efd6c1b):

### Infinite Fusion starter pool (c5679de)
- Synthetic species IDs (100_000 + head*10_000 + body) installed at runtime into the dex / starter / cost / eggMoves / levelMoves / TM / passive / biome / colors tables, and stripped from SystemSaveData so vanilla orthogonality is preserved.
- Per-pair unlock store in localStorage records shiny variants, IVs, genders, abilities, natures, value reductions, egg moves (2+2 asymmetric inheritance),passive, ribbons, and classic-win counts.
- New src/sprites/ module: artist-first resolver (fusion-sprite-resolver.ts), HTTP loader (fusion-sprite-loader.ts), variant probe, IF id ↔ NatDex id translation map, palette swap for shinies, and a back-mirror with bottom-trim.
- IF folder integration via the File System Access API in src/system/if-folder-handle.ts; community dex flavor text from Data/dex.json surfaces through src/system/if-dex-entries.ts.
- Starter UI rewrites: new Fusion filter on the starter select screen, split-coloured candy (head/body), evolutions/pre-evolutions rewritten to synthetic post-evo ids with eager install.
- New src/ui/settings/settings-fusions-ui-handler.ts and src/system/settings/fusion-settings.ts driving the Fusions settings tab.
- Save export/import carries fusion unlocks via the new fusionStarters field on SystemSaveData.
- Setup docs at docs/fusion-setup.md; English i18n keys for fusion strings (other locales pending in the locales fork).

### Splice-time refinements (2371e68)
 - Pokemon.playAnim now uses sprite.play(customKey) instead of setTexture, so the IF texture's single-frame animation replaces any in-progress vanilla animation. Combined with a post-fuse playerPokemon.loadAssets() chain, the IF sprite appears immediately in-battle after splicing.
- New MAXIMUM stat formula: max(headCost, bodyCost) + (finalBst >= 600 ? 2 : 1), anchored to the pricier parent with a +1 / +2 bump based on the BST of the fully-evolved fused form. Walks each parent's evolution chain to the leaf.
- New FusionUnlockSplashPhase reuses pokemonInfoContainer + showText with an end-callback, so the fresh-splice splash mirrors vanilla's catch popup (IVs, ability, nature, moveset). Pushed via unshift so the splash plays alongside fuse()'s queued LearnMovePhases.
  
### Ability slot derivation in battle (c3ecc25)
- Pokemon.getAbility for fusions previously returned body.getAbility(fusionAbilityIndex) — the body species's ability at the body's pre-fuse index. That contradicted the starter UI / splash popup, which advertise the IF-derived slot mapping.The slot now drives off this.abilityIndex against that map. NONE falls back to head.ability1, matching vanilla's existing fallback shape elsewhere.

### Phase queue fixes (d0d3f68, b335f6d)
- TrainerVictoryPhase now bails out via this.end() when currentBattle.trainer is undefined. Previously the unguarded trainer?.config.modifierRewardFuncs! resolved to undefined and the for-of threw "not iterable", crashing the run.
- VictoryPhase now skips queuing the post-battle chain when a NewBattlePhase is already in the phase queue — the case where two enemy pokemon faint on the same turn and both sibling VictoryPhases observe "no unfainted enemy" and each push BattleEnd / TrainerVictory / EggLapse / SelectModifier / NewBattle. NewBattlePhase deduplicated itself, but the rest ran twice, producing duplicate reward screens and the crash above.

### Trainer fusion mechanism (d7d382e → efd6c1b)
- New FusionSettingKeys.Trainer_Fusion setting (DEFAULT / NONE / ALWAYS).
- Trainer.genPartyMember calls a new shouldFuseTrainerPokemon(config) helper that switches on the setting. In DEFAULT mode the per-slot chance is 25% for trainers with hasCharSprite (gym leaders, E4, champions, evil-team bosses) and 5% otherwise.
- On a hit, the freshly-created EnemyPokemon runs generateFusionSpecies() (filtered by the existing getCompatibleFusionSpeciesFilter, which forces matching legendary tier and matching evolution stage for non-legendaries — so a Bug Catcher's Caterpie cannot pair with Arceus and a Lance Dragonite cannot pair with anything below a final-form non-legendary).
- A new Pokemon.swapFusionRoles() helper exchanges the head/body pair across all fusion fields (species, formIndex, abilityIndex, shiny, variant, gender, luck, customPokemonData). After the fusion is generated, a 50/50 coin flip decides whether the trainer's original mon stays as head or moves to body, so each slot can produce visibly different mixes across runs.

## Screenshots/Videos
Screenshot 2026-05-17 at 19 46 53
Screenshot 2026-05-17 at 19 49 43
Screenshot 2026-05-17 at 12 07 51

## How to test the changes?
1. Follow the new docs/fusion-setup.md to point the Fusions settings tab at a local IF install.
2. Start a run, get a DNA Splicer, splice two starters — confirm the fresh-splice splash mirrors a vanilla catch popup and the new pair appears as its own starter slot with split-coloured candy on the starter screen.
3. In battle with the spliced fusion, verify the in-battle ability slot matches what the starter UI / splash advertise (slot 0 → head.ability1, slot 1 → body.ability2, slot 2 → head.abilityHidden).
4. Open Settings → Fusions and try each stat formula (IF / PokéRogue / Maximum) on a freshly-spliced pair; confirm the Maximum starter cost lands at max(head, body) + (finalBst >= 600 ? 2 : 1).
5. Open Settings → Fusions, set Trainer fusions to Always, start a fresh run, and walk into a few trainer fights — every trainer pokemon should be fused, with the trainer's mon landing as head or body roughly evenly across slots.
6. Set the same setting to None; verify no trainer pokemon are fused, even on gym leaders. Set it to Default and confirm gym leaders / E4 / champions show fusions noticeably more often than Bug Catchers and other unnamed trainers.
7. For the phase fixes, force a double-battle scenario where both enemies faint on the same turn (e.g. multi-hit AoE) — verify only one reward screen / TrainerVictoryPhase runs, with no "not iterable" crash.


fabske0:
Thank you for your interest in contributing to PokéRogue.
But this is not a feature we want at the current moment, as we do not want the game to turn into Infinte fusion 2.0. There are other games that already do that.

We also do not accept AI-generated Pull Requests.

Thanks again for your interest in contributing to PokéRogue

\- PokéRogue team

Eddydpyl:
Aight

PR #7299, opened 2026-05-09
fix(item): MBH will no longer steal items when the holder is fainted

DayKev:
## What are the changes the user will see?
The Mini Black Hole item will no longer steal opponents' items if the holder is fainted.

## Why am I making these changes?
Fixing a bug.

## What are the changes from a developer perspective?
Added an `apply` method to the MBH item's class that has a check to see if the Pokémon is fainted, and returns `false` if so, otherwise returning the superclass's `apply` method.

## How to test the changes?
`pnpm test:silent mini-black-hole`

## Checklist
- The PR content is correctly formatted:
  - [x] **I'm using `beta` as my base branch**
  - [x] **The current branch is not named `beta`, `main` or the name of another long-lived feature branch**
  - [x] I have provided a clear explanation of the changes within the PR description
  - [x] The PR title matches the Conventional Commits format (as described in [CONTRIBUTING.md](https://github.com/pagefaultgames/pokerogue/blob/beta/CONTRIBUTING.md#pr-title-format))
- [x] The PR is self-contained and cannot be split into smaller PRs
- [x] There is no overlap with another open PR
- The PR has been confirmed to work correctly:
  - [x] I have tested the changes manually
  - [x] The full automated test suite still passes (use `pnpm test:silent` to test locally)
  - [x] I have created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes if necessary

PR #7298, opened 2026-05-08
refactor: convert `Pokemon#getTypes` and `#isOfType` to object params

DayKev:
## What are the changes the user will see?
N/A

## Why am I making these changes?
Easier use of these functions.

## What are the changes from a developer perspective?
- Converted `Pokemon#getTypes` and `#isOfType` to use object params.
- Added `ignoreThirdType` param to both, which handles whether to ignore the type added from Forest's Curse or Trick-or-Treat, split from the `bypassSummonData` param.

## Checklist
- The PR content is correctly formatted:
  - [x] **I'm using `beta` as my base branch**
  - [x] **The current branch is not named `beta`, `main` or the name of another long-lived feature branch**
  - [x] I have provided a clear explanation of the changes within the PR description
  - [x] The PR title matches the Conventional Commits format (as described in [CONTRIBUTING.md](https://github.com/pagefaultgames/pokerogue/blob/beta/CONTRIBUTING.md#pr-title-format))
- [x] The PR is self-contained and cannot be split into smaller PRs
- [x] There is no overlap with another open PR

PR #7290, opened 2026-05-03
feat: add more data to `window.gameInfo`

DayKev:
## What are the changes the user will see?
More information is exposed through `window.gameInfo` for plugin devs. Specifics in the [PR description](https://github.com/pagefaultgames/pokerogue/pull/7290).

## Why am I making these changes?
Making it easier for plugin devs.

## What are the changes from a developer perspective?
Added a bunch of stuff to the output of `BattleScene#updateGameInfo` for Pokémon:
- nickname
- gender
- base moveset
- temporary moveset (e.g. from Transform)
- base ability
- temporary ability (e.g. from Transform or Skill Swap)
- passive ability
- whether the passive is enabled
- nature
- base stats
- temporary in-battle stats (e.g. from Transform)
- current stat stages
- whether the Pokémon is shiny
- if the Pokémon is shiny, which shiny variant it is
- temporary in-battle types (e.g. from Transform or Forest's Curse)
- whether the Pokémon is a fusion (note that some of the fields may be inaccurate for fusions)

Fixed the `types` field so it shows the Pokémon's actual/base types as intended, ignoring temporary types such as from Transform or Forest's Curse.

Removed the unnecessary (and basically useless) `modeChain` field.

Added a `gameInfoVersion` field, to be updated whenever the output of `gameInfo` is modified.

Added a helper function to `src/utils/strings.ts` to capitalize only the first letter of a string.

Pulled `decodeNickname` out of `Pokemon#getNameToRender` into an exported utility function in `src/utils/pokemon-utils.ts`.

## Screenshots/Videos
<details><summary>Output Example 1</summary>

image

</details>
<details><summary>Output Example 2</summary>

image

</details>

## How to test the changes?
Load a battle, open the browser devtools, and enter `window.gameInfo`.

## Checklist
- The PR content is correctly formatted:
  - [x] **I'm using `beta` as my base branch**
  - [x] **The current branch is not named `beta`, `main` or the name of another long-lived feature branch**
  - [x] I have provided a clear explanation of the changes within the PR description
  - [x] The PR title matches the Conventional Commits format (as described in [CONTRIBUTING.md](https://github.com/pagefaultgames/pokerogue/blob/beta/CONTRIBUTING.md#pr-title-format))
- [x] The PR is self-contained and cannot be split into smaller PRs
- [x] There is no overlap with another open PR
- The PR has been confirmed to work correctly:
  - [x] I have tested the changes manually

PR #7288, opened 2026-05-02
fix(ai): Sunny Day will generate properly on NPC movesets

DayKev:
## What are the changes the user will see?
NPC Pokémon can have Sunny Day when their moveset supports it (has a damaging fire move) instead of the opposite.

## Why am I making these changes?
Bug noticed.

## What are the changes from a developer perspective?
`shouldRemoveSunnyDay()` returns `false` instead of `true` when the Pokémon has a damaging fire move in its moveset, the same as `shouldRemoveRainDance()` is for damaging water moves.

## Checklist
- The PR content is correctly formatted:
  - [x] **I'm using `beta` as my base branch**
  - [x] **The current branch is not named `beta`, `main` or the name of another long-lived feature branch**
  - [x] I have provided a clear explanation of the changes within the PR description
  - [x] The PR title matches the Conventional Commits format (as described in [CONTRIBUTING.md](https://github.com/pagefaultgames/pokerogue/blob/beta/CONTRIBUTING.md#pr-title-format))
- [x] The PR is self-contained and cannot be split into smaller PRs
- [x] There is no overlap with another open PR

PR #7310, opened 2026-05-18
chore!: Minor update 1.11.19 to main

fabske0:
**Changelog:** hotfix-1.11.19 ---> main
---------------------------
## Bug Fixes

- #7215
  - Life Dew will correctly ignore Substitute.
- #7217
  - Various text / rendering fixes for Legacy UI.
- #7189
  - When using Expanding Force in Psychic Terrain during a double battle, the animation will now correctly play for all affected targets instead of only one. This reflects the move's behavior as a spread move under these conditions.
- #7188
  - When a species has type changing abilities such as Aerilate, Pixilate, Refrigerate or Galvanize, moves that call other moves (Nature Power, Metronome, Mimic, ...) will now be displayed with their original type in the moves menu, even if their type would be changed by the ability during battle.
- #7224
  - The splash text "It's been [#] runs..." will display properly.
- #7218
  - Ability Bars in Legacy UI now have a minimum size and consistent text placement.
- #7187
  - The move "Gravity" will now show a message regarding Pokémon affected by the move.
- #7164
  - Court Change now switches every layer of each entry hazard properly (Spikes and Toxic Spikes)
- #7227
	- Mist will display the correct message when blocking a stat stage decrease.
	- Infiltrator will correctly ignore the effects of Mist.
- #7237
  - Ability Charms will now increase the chance of ME Pokémon having their HA instead of decreasing it, as intended.
- #7236
  - If an opposing Pokémon has Dazzling or Queenly Majesty, the user will no longer be prevented from targeting their ally.
- #7250
  - Biome weather will activate correctly when moving between biomes.
- #7159
  - Baneful Bunker now correctly blocks status-category moves (for example, Spore), in addition to blocking damaging moves.
- #7261
  - Prevents an issue where the language selection could be stored in the wrong place, causing it to be unable to be changed.
- #7284
  - Fused pokemon stats will no longer flip every wave in Flip Stat mode
- #7239
  - The candy count popup during egg hatching and battles will now be synced to the correct value.
- #7299
  - The Mini Black Hole item will no longer steal opponents' items if the holder is fainted.
- #7288
  - NPC Pokémon can have Sunny Day when their moveset supports it (has a damaging fire move) instead of the opposite.
- #7233
  - The game no longer softlocks or causes the replacement Pokémon to immediately faint when a pivot move user (U-Turn, Flip Turn, Volt Switch, etc.) faints from contact damage abilities like Rough Skin or Iron Barbs during a trainer battle.
- #7214
  - The game no longer softlocks when a player's Pokémon is fainted by Cramorant's Gulp Missile ability.
- #7223
  - Add flags to the viewport in index.html to prevent zooming on touch devices with rapidly pressing touch controls
- #7153
  - The user will now see a custom message when Pollen Puff is used on a target which has been affected by Heal Block.
- #7152
	- `Parting Shot` will no longer say "But it failed!" when the user has no eligible Pokémon in their party to switch into. It will correctly drop the target's stats and end the turn.
	- The move will no longer force a switch if the stat-drop is blocked by abilities like `Clear Body`, `Mist`, `Good as Gold`, or if the target's stats are already at `-6/-6`.
	- The move correctly bypasses `Substitute` (due to being sound-based).
- #7266
  - If the first hit of a multi-hit move would OHKO a boss pokemon but the damage is reduced by the boss bar mechanic, Sturdy won't incorrectly trigger on future hit(s).
	The flyout will still incorrectly be shown, but the functionality will be correct.
- #7320
  - The 3rd row of items in the summary will be properly displayed
## Balance

- #7279
  - Trainer Pokemon will generate with more "concise" move pools by adjusting the generation filters further.
## Translation

- #7253
  - Adds the flavor text for when Hero form Palafin is first sent out after changing forms.
- #7231
	- Out-of-date save data will now properly be re-downloaded from the server.
	- Messages related to save data are no longer hardcoded English text and can be translated to other languages.
- #6858
  - Nature Power has locales now!
	Also move calling moves will no longer randomly "fail" seemingly at random. (This was linked to calling a move that had no valid targets; such occurrences will call the move correctly and then have THAT fail instead.)
- #7305
  - Vietnamese & Polish addition
## Miscellaneous

- #7230
  - Remove unused "Consent Preferences" menu item inside Manage Data
- #7245
  - In ssui or the pokedex, the user will see "Reduce cost to x" instead of "Reduce cost"
- #7258
  - The message that appears after you reach the end of endless mode won't be skippable for 3 seconds after appearing, preventing newer players from skipping past it accidentally and then being confused when the game ends unexpectedly.
- #7259
  - Fusion names created by having Rapidash as the base will be better.
- #6697
  - A few moves and abilities will now have correct interactions when Stellar Terastallized.
	- Flower Veil
	- Flower Shield
	- Reflect Type

	A few other moves were changed to properly fail if the user is Terastallized:
	- Conversion 2
	- Conversion (also fails if user is same type as its first move, for consistency)

	Finally, Forest's Curse/Trick-or-Treat will be automatically removed when the user Terastallizes (tested on showdown (<https://replay.pokemonshowdown.com/gen9ou-2530172961-8xgqrtwj8acyevf4srfv4m8vyohs7e9pw>). These effects were previously ignored if the user was Terastallized, but this saves a bit of space in save data.
- #7280
  - The game speed options will change from the current selections to "Slow" (2x), "Normal" (3x), "Faster" (4x), and "Turbo" (5x)
- #7290
  - More information is exposed through `window.gameInfo` for plugin devs. Specifics in the [PR description](https://github.com/pagefaultgames/pokerogue/pull/7290).
- #7291
  - The font used for Japanese text is updated.
- #7207
  - When pressing `V` the flyout will now show the move revealed by forewarn
- #7077
  - Add an option to turn on or off skipping move confirmations.
- #7222
  - iOS users will now see a visible upload button in order to be compliant with iOS security protocols.
- #7263
  - Significantly lower memory usage from bgm, leading to reduced forced reloads, etc.
	~Potentially a slightly more noticeable stutter at the loop point of songs.~ (Due to the follow-up PR there shouldn't be any change in looping behavior)
- #7004
  - Terrain now properly boosts move power instead of attack damage.
- #7278
  - New option to not prioritize `Pass Baton`
- #7287
  - The only user-facing change should be that if exp gain display is set to `SKIP`, there is no longer any EXP bar animation. Instead, the bar and level number jump directly to the final values.
## Unknown

- #7225
- #7234
- #7229
- #7246
- #7252
- #7220
- #7022
- #7208
- #7262
- #7267
- #7148
- #7268
- #7264
- #7271
- #7272
- #7274
- #7276
- #7281
- #7275
- #7289
- #6886
- #7078
- #7294
- #7298
- #7118
- #7302
- #7312
- #7317
## Beta

- #7240
- #7303

---------------------------
**This changelog was auto generated at May 25, 2026 at 11:43 PM UTC.**


fabske0:
Assets are up to date. Locales needs an update before merging

PR #7339, opened 2026-05-29
fix(audio): Make BGM loading a bit more robust

emdeann:
Hopefully, no freezes on the party heal screen

Some reports have come in that people are rarely freezing on the party heal screen, indicating that the `onEnd` callback of the party background music isn't firing. I haven't been able to reproduce this, but I made some changes that should make it less likely regardless. It's questionable whether it's good practice to hinge something like a phase transition on music to begin with, but I'm not sure what the alternative is. We could just add a timeout equal to the length of the heal, but that would be screwy with load times and cut the music off.

Throw an error in the load promise if one happens, catch it in the BGM constructor and destroy it immediately, firing `onEnd` in the process. Also made a couple of general robustness fixes like attempting to play if `resume` is called on a non-started track (which maybe could happen i.e. if a track takes so long to load that it has been "paused" before actually playing)

## How to test the changes?
Run through a x10 wave and see if it works

## Checklist

  - [ ] **The current branch is not named `beta`, `main` or the name of another long-lived feature branch**
  - [ ] The PR title matches the Conventional Commits format (as described in [CONTRIBUTING.md](https://github.com/pagefaultgames/pokerogue/blob/beta/CONTRIBUTING.md#pr-title-format))
- [ ] The PR is self-contained and cannot be split into smaller PRs

### DIFF
--- a/src/audio/background-music.ts
+++ b/src/audio/background-music.ts
@@ -47,22 +47,26 @@ export class BackgroundMusic {
   constructor(key: string, loop: boolean, loopPoint = 0) {
     this.key = key;
 
-    globalScene.loadBgm(key).then(() => {
-      if (this.destroyed) {
-        return;
-      }
-      this.sound = globalScene.sound.add(key, { loop });
-      if (loop) {
-        this.sound.on("looped", () => {
-          if (!this.destroyed) {
-            this.sound?.play({ seek: loopPoint });
-          }
-        });
-      } else {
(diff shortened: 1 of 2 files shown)


damocleas (approved):

fabske0 (approved):
Also couldn't reproduce, but code looks fine.
maybe it's just something that happens online because the track takes too long to load

PR #7353, opened 2026-05-31
fix(audio): BGM no longer plays loudly after some fades even when muted

emdeann:
Certain MEs will no longer loudly play the previous BGM track when muted during the fade out.

I removed `fadeAndSwitchBgm` as it's redundant considering that there's already a `fadeOutPrevious` parameter on `playBgm`. I'm not sure where the actual issue came from, I suspect it was some sort of race condition due to `fadeAndSwitchBgm` using `delayedCall` instead of `playAfterDelay`, etc. But I figured this change (which needed to happen anyway) would fix the issue, and it does.

Note that this is technically a behavior change because the default fade duration is only 500ms rather than 2000, and there is a short delay between the delayed track ending and the new one starting. The 2000ms fade resulted in a pretty long period of just silence, so I don't think that there was much reason behind it and if anything it's better now, but it would be easy to add that delay back if it were desired.

## How to test the changes?
```ts
const overrides = {
  STARTING_WAVE_OVERRIDE: 14,
  MYSTERY_ENCOUNTER_OVERRIDE: MysteryEncounterType.FUN_AND_GAMES,
```
Use these overrides and spam through a few encounters (using "Leave" is fine). Before the changes you'll hear the BGM spike in loudly, after you won't

## Checklist

  - [ ] **The current branch is not named `beta`, `main` or the name of another long-lived feature branch**
  - [ ] The PR title matches the Conventional Commits format (as described in [CONTRIBUTING.md](https://github.com/pagefaultgames/pokerogue/blob/beta/CONTRIBUTING.md#pr-title-format))
- [ ] The PR is self-contained and cannot be split into smaller PRs

### DIFF
--- a/src/audio/audio-manager.ts
+++ b/src/audio/audio-manager.ts
@@ -67,6 +67,7 @@ export class AudioManager {
    * @param fadeDuration - (Default `500`) How long, in ms, the fade out of the previous bgm should take
    * @returns The {@linkcode BackgroundMusic} instance for the new bgm,
    * or `null` if no valid bgm could be played or the input bgm was the same as the currently playing bgm
+   * todo group the fade parameters next to each other
    */
   public playBgm(bgmName?: string, fadeOutPrevious = false, loop = true, fadeDuration = 500): BackgroundMusic | null {
     const resolvedName = timedEventManager.getEventBgmReplacement(
--- a/src/audio/audio-manager.ts
+++ b/src/audio/audio-manager.ts
@@ -64,10 +64,11 @@ export class AudioManager {
    * Can be overridden by a currently running event.
    * @param fadeOutPrevious - (Default `false`) Whether to fade out the previously playing bgm
    * @param loop - (Default `true`) Whether to loop the new bgm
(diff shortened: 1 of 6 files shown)


src/audio/audio-manager.ts
```diff
@@ -64,10 +64,12 @@ export class AudioManager {
    * Can be overridden by a currently running event.
    * @param fadeOutPrevious - (Default `false`) Whether to fade out the previously playing bgm
    * @param loop - (Default `true`) Whether to loop the new bgm
+   * @param fadeDuration - (Default `500`) How long, in ms, the fade out of the previous bgm should take
    * @returns The {@linkcode BackgroundMusic} instance for the new bgm,
    * or `null` if no valid bgm could be played or the input bgm was the same as the currently playing bgm
+   * todo group the fade parameters next to each other
    */
```
DayKev:
```suggestion
   */
  // TODO: use object params
```

DayKev (dismissed):

damocleas (approved):

fabske0 (approved):

damocleas (approved):

PR #7360, opened 2026-06-03
feat(battle): added "Retry Biome" option on Daily Run

### LINKED ISSUE #6541: [Feature] Add "Retry from the start of current biome" as an option in Daily Runs.

Many players enjoyed the puzzle aspect that Daily Runs could provide, however an exploit to facilitate this aspect was patched and left many upset. While we obviously don't want to condone and leave in exploits like that, many of balance and those part of the daily run community feel a "rewind"-like feature would vastly improve Daily Run for players. I personally feel like Daily Run is pretty underdeveloped, and this would be a much welcomed addition to the mode.

### Describe the Feature

[ONLY FOR DAILY RUN] Add the ability to retry from the start of the biome you're currently in, you lose all rewards you gained from that point. This option would be in two places, 

1. When the "Retry Battle" option is on, add a new option that lets you retry from the beginning of the biome you're in, along with retrying the battle and quitting.

2. An option in the shop and/or main menu, allows players to quickly retry without needing to lose intentionally

### Additional context

_No response_

afonsob05:

Upon being defeated in a Daily Run, if a player has the "Enable Retries" setting ON, instead of the standard 2 option menu for retrying the last battle/quitting, there is now a 3rd option for retrying the whole biome. Selecting that option, the player will go back to the first wave of the biome they were in when defeated (a checkpoint), losing all the items accumulated in the waves past that checkpoint.

This feature addresses a concern raised by some fans of the Daily Run mode, as reported in issue #6541
Before, the only options available for a defeat would be : 
1. To retry the just lost battle;
 2. Quit, erasing the save. 

With this, a simple mistake could cost the whole run, and aside from exploits, there was no other way to go back in the run as much as a whole biome.

Now, with the addition of the 3 option for retrying the whole biome, the player now has a little bit more freedom to make mistakes, thus enjoying the game more, and improving playability!

Fixes #6541

`game-data.ts` and `save-data.ts`: the checkpoint data struct was created. This checkpoint is an integral copy of the Session Save the player is in. New functions were also created for saving and loading this checkpoint save, so that when the checkpoint is loaded, the SessionSave is re-written.
(update on `game-data.ts`): the saveAll method was modified, to comprise a new  optional 'reset' flag, that is set to true when used in the loadBiomeCheckpoint method.

`save-data-api.ts`: the updateAll method was updated to comprise the previous changes regarding the 'reset' flag. now, the method can send 2 distinct url when performing doPost:
- [NEW] "/savedata/updateall?reset=true" for the biomeCheckpoint variant;
- "/savedata/updateall".

`encounter-phase.ts`: the logic for saving this checkpoint was added. A simple if-clause was created to ensure that the save only occurs when "enableRetries" is On, the gamemode is Daily Run & the player is in the first wave of that biome.

`game-over-phase.ts`: the new menu was created, alongside new keys to present the options (more on that on the locales PR, link included below). A bit of refactoring was done to ensure that the previously menu still exists, but only on modes like Classic Run. New functions were also created to handle this new way to "end" a game,

 

## Screenshots/Videos

[before screenshot here](https://github.com/user-attachments/assets/9bed3294-5f73-4764-88a0-13c1e74c4aca)

</details>

[after screenshot here](https://github.com/user-attachments/assets/fa5fb392-70a1-4dd1-a843-ef18ea7cfa3e)

</details>

## How to test the changes?

For these changes to be tested, the reviewer should make sure that the "Enable Retries" feature is on, as well as that a Daily Run is being played. Overrides can be used to rig the starting phase, however this can be tested in the first 10 waves.
Force a defeat, and the new menu (with the new options) should appear!

## Checklist

  - [X] **The current branch is not named `beta`, `main` or the name of another long-lived feature branch**
  - [X] I have provided a clear explanation of the changes within the PR description
  - [X] The PR title matches the Conventional Commits format (as described in [CONTRIBUTING.md](https://github.com/pagefaultgames/pokerogue/blob/beta/CONTRIBUTING.md#pr-title-format))
- [X] The PR is self-contained and cannot be split into smaller PRs
- [X] There is no overlap with another open PR
  - [X] I have tested the changes manually
  - [X] The full automated test suite still passes (use `pnpm test:silent` to test locally)
  - [~] I have created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes if necessary
- [X] I have provided screenshots/videos of the changes (if applicable)
  - [X] I have made sure that any UI changes work for both the default and legacy UI themes (if applicable)

- [X] I have created an associated PR on the [locales](https://github.com/pagefaultgames/pokerogue-locales) repository
  - If so, include a link to the PR here: [link](https://github.com/pagefaultgames/pokerogue-locales/pull/322)
- [X] I have contacted the Translation Team on Discord for proofreading/translation

- [~] I have created an associated PR on the [assets](https://github.com/pagefaultgames/pokerogue-assets) repository

### DIFF
--- a/src/@types/save-data.ts
+++ b/src/@types/save-data.ts
@@ -70,6 +70,11 @@ export interface SessionSaveData {
    * Counts the amount of pokemon fainted in your party during the current arena encounter.
    */
   playerFaints: number;
+  /**
+   * Integral copy of a session save, that is ovewritten at the first wave of every biome.
+   * Only used in Daily Runs, if "Retry Battles" is enabled.
+   */
+  biomeCheckpoint?: Omit<SessionSaveData, "biomeCheckpoint">;
 }
 
 export interface Unlocks {
--- a/src/phases/encounter-phase.ts
+++ b/src/phases/encounter-phase.ts
@@ -292,10 +292,8 @@ export class EncounterPhase extends BattlePhase {
           // Set weather and terrain before session gets saved
           this.trySetWeatherIfNewBiome();
           this.trySetTerrainIfNewBiome();
-          // Save biome checkpoint in all X1 waves, if Enable Retries is active
-          if (globalScene.enableRetries && globalScene.gameMode.isDaily && battle.waveIndex % 10 === 1) {
-            globalScene.gameData.saveBiomeCheckpoint();
-          }
+          this.trySaveBiomeCheckpointIfNewBiome();
+
           // Game syncs to server on waves X1 and X6 (As of 1.2.0)
           globalScene.gameData
             .saveAll(true, battle.waveIndex % 5 === 1 || (globalScene.lastSavePlayTime ?? 0) >= 300)
@@ -671,5 +669,18 @@ export class EncounterPhase extends BattlePhase {
    */
   protected trySetTerrainIfNewBiome(): void {
     globalScene.arena.setBiomeTerrain();
+  }
+
+  /**
+   * Save a biome checkpoint if and only if this encounter is the start of a new biome.
+   * @remarks
+   * By using function overrides, this should happen if and only if this phase
+   * is exactly a `NewBiomeEncounterPhase` or an `EncounterPhase` (to account for
+   * Wave 1 of a Daily Run), but NOT `NextEncounterPhase`.
+   */
+  protected trySaveBiomeCheckpointIfNewBiome(): void {
+    if (globalScene.enableRetries && globalScene.gameMode.isDaily) {
+      globalScene.gameData.saveBiomeCheckpoint();
+    }
   }
 }
--- a/src/phases/encounter-phase.ts
+++ b/src/phases/encounter-phase.ts
@@ -292,6 +292,10 @@ export class EncounterPhase extends BattlePhase {
           // Set weather and terrain before session gets saved
           this.trySetWeatherIfNewBiome();
           this.trySetTerrainIfNewBiome();
+          // Save biome checkpoint in all X1 waves, if Enable Retries is active
+          if (globalScene.enableRetries && globalScene.gameMode.isDaily && battle.waveIndex % 10 === 1) {
+            globalScene.gameData.saveBiomeCheckpoint();
+          }
           // Game syncs to server on waves X1 and X6 (As of 1.2.0)
           globalScene.gameData
             .saveAll(true, battle.waveIndex % 5 === 1 || (globalScene.lastSavePlayTime ?? 0) >= 300)
--- a/src/phases/game-over-phase.ts
+++ b/src/phases/game-over-phase.ts
@@ -77,21 +77,21 @@ export class GameOverPhase extends BattlePhase {
       );
     } else if (this.isVictory || !globalScene.enableRetries) {
       this.handleGameOver();
-    } else if (globalScene.gameMode.isDaily) {
+    } else if (globalScene.gameMode.isDaily && globalScene.gameData.biomeCheckpoint) {
       globalScene.ui.showText(i18next.t("battle:retryBattleDaily"), null, () => {
         globalScene.ui.setMode(UiMode.OPTION_SELECT, {
           options: [
             {
               label: i18next.t("battle:retryBattleOption"),
               handler: () => {
-                this.doRetryBattle();
+                this.doRetry();
                 return true;
               },
             },
             {
               label: i18next.t("battle:retryBiomeOption"),
               handler: () => {
-                this.doRetryBiome();
+                this.doRetry(true);
                 return true;
               },
             },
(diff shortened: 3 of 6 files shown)


emdeann:
I'm concerned that this will require server changes. 

```go
existingSave, err := savedata.GetSession(db.Store, uuid, slot)
if err != nil {
	if !errors.Is(err, savedata.ErrSaveNotExist) {
		httpError(w, r, fmt.Errorf("failed to retrieve session save data: %s", err), http.StatusInternalServerError)
		return
	}
} else {
	if existingSave.Seed == session.Seed && existingSave.WaveIndex > session.WaveIndex {
		httpError(w, r, fmt.Errorf("session out of date: existing wave index is greater"), http.StatusBadRequest)
		return
	}
}
```
Retrying from the start of a biome will fail to save to the server until the player clears past where they were before. When testing it with the server running I also had several other issues, where the biome reset button only reset me to the start of the same battle, and a full crash that occurred when saving and quitting after a biome reset. If these don't happen when testing normally they're likely a result of the client being confused due to the mismatch with the server.

You could probably get away with this without server changes by clearing the session before sending the new one (with the biome reset), but that's not an ideal solution for obvious reasons. I think we'd probably want to do something like send a parameter with the `UpdateSession` endpoint to indicate an intentional biome reset, such that the `existingSave.WaveIndex > session.WaveIndex` can be selectively ignored.

src/phases/game-over-phase.ts
```diff
@@ -78,37 +77,41 @@ export class GameOverPhase extends BattlePhase {
...
+          options: [
+            {
+              label: i18next.t("battle:retryBattleOption"),
+              handler: () => {
+                this.doRetryBattle();
+                return true;
+              },
+            },
+            {
+              label: i18next.t("battle:retryBiomeOption"),
```
emdeann, line 92: This should be shown only if there is a biome checkpoint to load. It's confusing behavior to have this silently fall back to a wave reset instead

afonsob05: Now it should be fixed! Now the new dialogue only appears if there is an available checkpoint at the moment of the loss!

src/@types/save-data.ts
```diff
@@ -70,6 +70,11 @@ export interface SessionSaveData {
    * Counts the amount of pokemon fainted in your party during the current arena encounter.
    */
   playerFaints: number;
+  /**
+   * Integral copy of a session save, that is ovewritten at the first wave of every biome.
+   * Only used in Daily Runs, if "Retry Battles" is enabled.
+   */
+  biomeCheckpoint?: Omit<SessionSaveData, "biomeCheckpoint">;
```
emdeann, line 77: It's worth at least discussing the size implications here given that it's essentially doubling the size of daily run sessions. I figure that the percentage of runs which are daily runs with retries enabled is probably low enough that it's fine though

src/phases/encounter-phase.ts
```diff
@@ -293,6 +292,10 @@ export class EncounterPhase extends BattlePhase {
           // Set weather and terrain before session gets saved
           this.trySetWeatherIfNewBiome();
           this.trySetTerrainIfNewBiome();
+          // Save biome checkpoint in all X1 waves, if Enable Retries is active
+          if (globalScene.enableRetries && globalScene.gameMode.isDaily && battle.waveIndex % 10 === 1) {
+            globalScene.gameData.saveBiomeCheckpoint();
+          }
```
emdeann: Better to tie this to the biome change rather than hardcoding it at X1 - see `EncounterPhase.trySetWeatherIfNewBiome` and related methods.

src/phases/game-over-phase.ts
```diff
@@ -120,6 +123,71 @@ export class GameOverPhase extends BattlePhase {
     }
   }
 
+  /**
+   * Handles retry battle, available when the Enable Retries is active.
+   */
+  private doRetryBattle(): void {
```
emdeann: `doRetryBattle` and `doRetryBiome` are heavily duplicated. Even the callback is the same - the only difference is whether `loadSession` or `loadBiomeCheckpoint` is called. You should make this a single function with a boolean parameter instead

src/@types/save-data.ts
afonsob05: "I figure that the percentage of runs which are daily runs with retries enabled is probably low enough that it's fine though" 

Yep, I think so too! Especially given that  @SirzBenjie stated "1:1 copy is fine" on Discord!

src/phases/game-over-phase.ts
afonsob05: Done and refactored!

src/phases/encounter-phase.ts
afonsob05: Fixed!! Created similar a function to ```EncounterPhase.trySetWeatherIfNewBiome``` and ```EncounterPhase.trySetTerrainIfNewBiome```!

PR #7369, opened 2026-06-09
feat: added party switch mode

### LINKED ISSUE #6583: [Feature] Allow rearranging party after seeing a fixed trainer fight, but before seeing their lead Pokémon

Currently, you have to remember the boss that is coming up and prepare the wave before the fixed fight. This is an element of poor design, in my opinion. Especially because, in mainline, you'd always be able to switch your Pokémon beforehand.

### Describe the Feature

When entering a fixed-wave trainer battle, you will be given the opportunity to rearrange the order of your Pokémon. This wouldn't even _need_ to be (though could start out as) being walled behind the "switch / set" in the option.
Basically, this feature would add need a new "rearrange party" phase. This will happen _after_ seeing which gym leader you are fighting, but before seeing their lead.

### Additional context

I think this would be nice for a few reasons. Adding a phase before fighting fixed-battles would open the door for those battle animations that appear, like in gen3 and gen4. I would absolutely _love_ to have that intro animation, as I'd like to make our game more visually appealing in all the ways we can. Having this be the prelude to such an animation would be amazing.

Image

https://github.com/user-attachments/assets/0c39b60b-ef5d-4e8b-a627-b5e107a68829

### LINKED ISSUE #7286: [Feature] Adjustable Party Positions

Sometimes you want to switch your order around before an upcoming trainer fight or just adjust the order of your party in general.

### Describe the Feature

During the shop between waves there should be an option that let's you swap the position of your party. The option should be inside the `check team` menu. The button should probably be renamed to `manage party` or similar to fit the new functionality.

### Additional context

Image

Lord-Claro:

New switch option when selecting a pokemon in the check team menu.
New button for a newly implemented switch mode for easier use of the switch option.
New keybind for faster use of switching option.
New sliding animation for swapping pokemon in double battles.

There currently is no way to adjust the party positions, the only way to switch pokemon is either before a wild encounter or inside a battle but this switch only works to switch for the pokemon in the front of the party.

Resolves #7286 
Resolves #6583 arguably as well

- Implemented the switch option in the party options that allows the user to switch the indexes of the party pokemon.
- Implemented a new keybind to access as a faster switch option and added it's implementation to controller and touch controls.
- Added a new button that when pressed sets partySwitchMode true and makes it so when a pokemon is selected it always is selected with the switch option and the party options don't appear.
- Added a new phase for dealing with the different kind of animations made by the switches in the "check team" menu. This class is used to correct the positions with the new indexes visually. It makes the switching animation before the animation of the new wave and if there is a double battle and one of the party pokemon switched with the other it slides to the side.
- Added test coverage that tests using the new switch option in the menu and the new switch mode accessed by the press of the new swap mode button. The tests verify if the button appears in the correct menu and if the switches happen correctly.

## Screenshots/Videos

Before and after video:
https://youtu.be/NaUN9SF7beo
Doubles animation:
https://youtu.be/aSVUC7MceI8

## How to test the changes?

Run the automated test locally:
```
pnpm test:silent switch-mode-button
pnpm test:silent switch-pokemon
```

## Checklist

  - [X] **The current branch is not named `beta`, `main` or the name of another long-lived feature branch**
  - [X] I have provided a clear explanation of the changes within the PR description
  - [X] The PR title matches the Conventional Commits format (as described in [CONTRIBUTING.md](https://github.com/pagefaultgames/pokerogue/blob/beta/CONTRIBUTING.md#pr-title-format))
- [X] The PR is self-contained and cannot be split into smaller PRs
- [X] There is no overlap with another open PR
  - [X] I have tested the changes manually
  - [X] The full automated test suite still passes (use `pnpm test:silent` to test locally)
  - [X] I have created new automated tests (`pnpm test:create`) or updated existing tests related to the PR's changes if necessary
- [X] I have provided screenshots/videos of the changes (if applicable)


damocleas:
merge conflict

Wlowscha:
I really like this feature, and would like to have it in soon. However, I must question whether we need the new button and keybind at all.

I have found that it takes longer to remember to press the switch mode button than to just switch two Pokémon. Even if one wants to reorganize the entire party, the time spent clicking on a mon and selecting "switch" is so brief that it does not warrant bloating the ui with a new button.

Even if we decide to keep the button on the screen, we can just reuse one of the existing keybinds (which the party ui does not really use) instead of adding a new one.

Overall, I would strongly suggest to leave out the button and keybind for now, and only add them if there are many requests for it from the player base.


That aside the implementation seems good; having a new phase to manage the switching is certainly the way to go. Nice to see that e.g. Primordial Sea correctly stops working, which was the first gotcha I was looking for. Some things can still be made smoother:
- After a trainer battle, when the switch happens, the trainer icon is still on the field---this does not look very nice.
- More of a personal preference, but I would like to see a message for calling back the previous mon, as well as a message for sending out the new one.

src/phases/party-reorder-switch-phase.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    const leavingPokemon = party.filter(pokemon => pokemon.isOnField() && !desiredField.includes(pokemon));
+
+    const enteringIndexes = desiredField.reduce<number[]>((acc, pokemon, index) => {
+      if (!pokemon.isOnField()) {
+        acc.push(index);
+      }
+      return acc;
+    }, []);
```
DayKev: Don't use `.reduce` like this. This looks like it could be a `for` loop of some sort.

src/phases/party-reorder-switch-phase.ts
```diff
@@ -0,0 +1,150 @@
...
+    return slides;
+  }
+
+  /**
+   * Play the "come back" recall animation for the given pokemon, removing them
+   * from the field once the animation completes.
+   * @param leavingPokemon - The pokemon to recall
+   * @param onComplete - Callback invoked once every recall animation has finished
+   */
+  private recallPokemon(leavingPokemon: Pokemon[], onComplete: () => void): void {
```
DayKev: TODO: investigate reusing existing phases instead of duplicating functionality here.

Wlowscha: That looks difficult with the present code, might be easier to do after switch phase refactor. Main thing is that here we need to account for the possibility of recalling two mons at the same time, switching the party around, then sending out up to two mons.

Wlowscha:
Made an improvement to `SummonPhase` so that, in double battles, the mon on the right will always have its sprite on top, even if the mon on the left gets sent out later. In other words, this will not happen anymore (currently on main):

image

Wlowscha:
Tests passing locally:

image

PR #7371, opened 2026-06-09
feat: Allow learning of prevo exclusive moves via Memory Mushroom

Fontbane:
## What are the changes the user will see?

- Mons can use Memory Mushrooms to relearn moves that their pre-evolutions learn but they do not, as long as the mon is at the right level.
- When using Memory Mushrooms, moves can be highlighted in different colors based on their source.
- Adds functionality so enemy mons can access these moves too

TODO:
- Add a prefix in front of the relearning options' names to signify their sources
- Add prevo moves to Dex levelup listing
- Change colors since they're obviously just for demonstration rn
- Add flavor text when a mon reaches a level where they get access to a new prevo move

## Why am I making these changes?

- Desired by balance team for natdex movesets

## What are the changes from a developer perspective?

- Moved the `PartyUiMode` enum into its own file
- Added a `MovesetRetrieveContext` interface to provide more context to moveset lookup functions
- Reworked the `Pokemon.getLearnableLevelMoves` function. It no longer calls `Pokemon.getLevelMoves`
  - It also, rather than returning an array of MoveIds, it returns a tuple of 3 numbers: a level, the move id, and a new `LearnableMoveSource` enum.
  - These are temporarily stored in a `learnableLevelMoves` field in the mon's data which is cleared when it's no longer needed. This way the set can be accessed multiple times without having to regenerate.
- Added functions in Species Data Registry to get prevo chains based on formKey and specifically get sets of PrevoMoves
- Added a function in `pokemon-utils.ts` to determine non-default colors of party menu ui options. Mainly used for relearning, returns two strings, the hex codes for the option's text color and its shadow color. There's a switch statement there that changes colors depending on LearnableMoveSource.

## Screenshots/Videos

Note: This is with Slakoth line's NatDex sets, which are not in yet.

image

## Checklist

- The PR content is correctly formatted:
  - [ ] **I'm using `beta` as my base branch**
  - [ ] **The current branch is not named `beta`, `main` or the name of another long-lived feature branch**
  - [ ] I have provided a clear explanation of the changes within the PR description
  - [ ] The PR title matches the Conventional Commits format (as described in [CONTRIBUTING.md](https://github.com/pagefaultgames/pokerogue/blob/beta/CONTRIBUTING.md#pr-title-format))
- [ ] The PR is self-contained and cannot be split into smaller PRs
- [ ] There is no overlap with another open PR
- The PR has been confirmed to work correctly:
  - [ ] I have tested the changes manually
[...]

PR #7380, opened 2026-06-13
feat: Allow prevo moves during memory mushroom and add prefix

fabske0:

- Prevo level moves will be available during memory mushrooms
- When using a memory mushroom it will show an icon or the level indicating where that move comes from
    - Fusion sources will have a blue text

- When a Pokemon levels to a point where a prevolution would have learned a move, such that that move is now available via mushroom, a new message is displayed. For example, when Breloom gets to level 40, because Shroomish learns spore at 40 (from Dean)

Level Moves the species gets itself later, but the prevo would already have it are not available.
Example: Shroomish gets seed bomb at 36, but because Breloom gets it at 44, a level 40 breloom won't have it available 

Discussion in balance chat

- Turned on `includePrevolutionMoves` for memory mushrooms
- moves the `PartyUiMode` enum to the enums dir and converted it to a const object
- `Pokemon.getLevelMoves()` and `pokemon.getLearnableMoves()` now additionally return the source of the move (e.g. level, prevo, egg...)
- Using the memory mushroom now shows the level/an icon for each move

## Screenshots/Videos

image

</details>

image

</details>

## How to test the changes?

```ts
const overrides = {
  ITEM_REWARD_OVERRIDE: [{ name: "MEMORY_MUSHROOM"}]
```

## Checklist

  - [x] I have made sure that any UI changes work for both the default and legacy UI themes (if applicable)


damocleas (dismissed):

damocleas:
balance approval review ^

src/@types/pokemon-species.ts
```diff
@@ -14,6 +15,7 @@ interface PokemonSpeciesPassives {
 }
 
 export type LevelMoves = [number, MoveId][];
+export type LevelMovesWithSource = [number, MoveId, LearnableMoveSource][];
```
SirzBenjie:
```suggestion
export type LevelMovesWithSource = [level: number, move: MoveId, source: LearnableMoveSource][];
```

src/@types/pokemon-species.ts
```diff
@@ -14,6 +15,7 @@ interface PokemonSpeciesPassives {
 }
 
 export type LevelMoves = [number, MoveId][];
```
SirzBenjie:
```suggestion
export type LevelMoves = [level: number, move: MoveId][];
```

src/ai/ai-moveset-gen.ts
```diff
@@ -75,10 +76,12 @@ import { ValueHolder } from "#utils/value-holder";
...
 function getAndWeightLevelMoves(pokemon: Pokemon): Map<MoveId, number> {
   const movePool = new Map<MoveId, number>();
-  let allLevelMoves: [number, MoveId][];
+  let allLevelMoves: LevelMoves;
   // TODO: Investigate why there needs to be error handling here
   try {
-    allLevelMoves = pokemon.getLevelMoves(1, true, true, pokemon.hasTrainer());
+    allLevelMoves = pokemon
+      .getLevelMoves(1, true, true, pokemon.hasTrainer())
+      .map(([lvl, move, _source]) => [lvl, move] as const);
```
SirzBenjie: Is this necessary? A cursory glance tells me there's no harm in the extra array field.

src/field/pokemon.ts
```diff
@@ -2775,112 +2797,165 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+          const moves: LevelMovesWithSource = [];
+          for (const lm of speciesLevelMoves) {
+            const moveSource =
+              lm[0] === RELEARN_MOVE
+                ? LearnableMoveSource.RELEARN
+                : lm[0] === EVOLVE_MOVE
+                  ? LearnableMoveSource.EVOLUTION
+                  : e === evolutionChain.length - 1
+                    ? LearnableMoveSource.LEVEL
+                    : LearnableMoveSource.PREVO;
```
SirzBenjie: I tend to not like nested ternaries like this one.

src/field/pokemon.ts
```diff
@@ -2775,112 +2797,165 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+        // TODO: Might need to pass specific form index in simulated evolution chain
+        const speciesLevelMoves = getPokemonSpeciesForm(evolutionChain[e][0], this.formIndex).getLevelMoves();
+        if (includeRelearnerMoves) {
+          const moves: LevelMovesWithSource = [];
+          for (const lm of speciesLevelMoves) {
+            const moveSource =
+              lm[0] === RELEARN_MOVE
+                ? LearnableMoveSource.RELEARN
+                : lm[0] === EVOLVE_MOVE
+                  ? LearnableMoveSource.EVOLUTION
```
SirzBenjie: It seems that the order of these checks means that if the pokemon has a prevo that learns the move on evolution, but the mon itself learns it as level up, then it'll keep the source as "evo". iirc an example of this was metagross with metang getting confusion or something like that.

Should determine whether this is correct.

src/field/pokemon.ts
```diff
@@ -1913,17 +1915,37 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-   * @returns An array of {@linkcode MoveId}s, as described above.
+   * @returns A tuple of the Level (or `null` for non level moves), {@linkcode MoveId} and their corresponding {@linkcode LearnableMoveSource}, as described above.
    */
-  public getLearnableLevelMoves(): MoveId[] {
-    let levelMoves = this.getLevelMoves(1, true, false, true).map(lm => lm[1]);
+  public getLearnableLevelMoves(): [number | null, MoveId, LearnableMoveSource][] {
+    let levelMoves: [number | null, MoveId, LearnableMoveSource][] = this.getLevelMoves(1, true, true, true).map(lm => [
+      lm[0],
+      lm[1],
+      lm[2],
```
SirzBenjie: I think the `map` here is unnecessary.
Also, is the first value allowing `null` because of the egg moves which can be learned later?

src/field/pokemon.ts
```diff
@@ -2775,112 +2797,165 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+          || (includeRelearnerMoves && lm[0] === RELEARN_MOVE)
+          || lm[0] > 0
+        ) {
+          const moveSource =
+            lm[0] === RELEARN_MOVE
+              ? LearnableMoveSource.FUSION_RELEARN
+              : lm[0] === EVOLVE_MOVE
+                ? LearnableMoveSource.FUSION_EVOLUTION
+                : LearnableMoveSource.FUSION_LEVEL;
+          levelMoves.push([lm[0], lm[1], moveSource]);
```
SirzBenjie:
```suggestion
      for (const [level, move] of fusionLevelMoves) {
        if (
          (includeEvolutionMoves && level === EVOLVE_MOVE)
          || (includeRelearnerMoves && level === RELEARN_MOVE)
          || level > 0
        ) {
          const moveSource =
            level === RELEARN_MOVE
              ? LearnableMoveSource.FUSION_RELEARN
              : level === EVOLVE_MOVE
                ? LearnableMoveSource.FUSION_EVOLUTION
                : LearnableMoveSource.FUSION_LEVEL;
          levelMoves.push([level, move, moveSource]);
```

src/field/pokemon.ts
```diff
@@ -2775,112 +2797,165 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
    * @param includeRelearnerMoves - Whether to include moves that would require a relearner. Note the move relearner inherently allows evolution moves
    * @returns A list of moves and the levels they can be learned at
    */
   getLevelMoves(
     startingLevel?: number,
     includeEvolutionMoves = false,
-    simulateEvolutionChain = false,
+    includePrevolutionMoves = false,
     includeRelearnerMoves = false,
     learnSituation: LearnMoveSituation = LearnMoveSituation.MISC,
```
SirzBenjie: This method is far too long (it was already before, but this PR exacerbates its already too large size, from 126 LoC to 184).
It would be best to split it up into smaller chunks so that it is more tenable to parse.

src/field/pokemon.ts
```diff
@@ -2775,112 +2797,165 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+          (includeEvolutionMoves && lm[0] === EVOLVE_MOVE)
+          || (includeRelearnerMoves && lm[0] === RELEARN_MOVE)
+          || lm[0] > 0
+        ) {
+          const moveSource =
+            lm[0] === RELEARN_MOVE
+              ? LearnableMoveSource.RELEARN
+              : lm[0] === EVOLVE_MOVE
+                ? LearnableMoveSource.EVOLUTION
+                : LearnableMoveSource.LEVEL;
```
SirzBenjie: You can probably extract this common logic into a utility function. (Would help reduce LoC bloat).

src/field/pokemon.ts
fabske0: I renamed the var to make it more clear. How would you handle it otherwise?

src/field/pokemon.ts
fabske0: I didn't do that because there are like 5 different scenarios. I changed them to a switch case though for better readability.

src/field/pokemon.ts
fabske0: Changed it so it prefers level over other sources

src/field/pokemon.ts
fabske0: Split it up into 6 methods. 1 for each scenario + 1 for filtering and sorting

src/field/pokemon.ts
```diff
@@ -2771,116 +2790,296 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+        ret.push([lm[0], lm[1], moveSource]);
+      }
+    }
+    return ret;
+  }
+
+  /**
+   * Helper method for `getLevelMoves()`
+   *
+   * Get all level moves the species form can learn via its prevolutions.
```
SirzBenjie:
```suggestion
   * Get all level moves the species form and its prevolutions can learn.
```

src/field/pokemon.ts
```diff
@@ -2771,116 +2790,296 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-            lm =>
-              (includeEvolutionMoves && lm[0] === EVOLVE_MOVE)
-              || (includeRelearnerMoves && lm[0] === RELEARN_MOVE)
-              || lm[0] > 0,
-          );
+        for (const lm of speciesLevelMoves) {
+          if (includeEvolutionMoves && lm[0] === EVOLVE_MOVE) {
+            ret.push([lm[0], lm[1], LearnableMoveSource.EVOLUTION]);
+          } else if ((!e || lm[0] > 1) && (e === evolutionChain.length - 1 || lm[0] <= evolutionChain[e + 1][1])) {
+            ret.push([lm[0], lm[1], LearnableMoveSource.PREVO]);
```
SirzBenjie: This is confusing me. Why isn't `LearnableMoveSource.LEVEL` possible in this conditional branch (where `includeRelearnerMoves`  is `false`) but possible in the branch where it is true?

This is seemingly preventing this method from being collapsed into something like:
```ts
 for (const [level, move] of speciesLevelMoves) {
      if (includeRelearnerMoves && level === RELEARN_MOVE) {
        ret.push([level, move, LearnableMoveSource.RELEARN]);
      } else if (includeEvolutionMoves && level === EVOLVE_MOVE) {
        ret.push([level, move, LearnableMoveSource.EVOLUTION]);
      } else if ((!e || level > 1) && (e === evolutionChain.length - 1 || level <= evolutionChain[e + 1][1])) {
        ret.push([level, move, LearnableMoveSource.PREVO]);
      }
    }
```

Also, here

```ts
else if ((!e || level > 1) && (e === evolutionChain.length - 1 || level <= evolutionChain[e + 1][1])) {
```

It's a bit hard to swallow what the conditional is really checking for here...

src/field/pokemon.ts
```diff
@@ -2771,116 +2790,296 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+              break;
+            case EVOLVE_MOVE:
+              moveSource = LearnableMoveSource.EVOLUTION;
+              break;
+            default:
+              moveSource = e === evolutionChain.length - 1 ? LearnableMoveSource.LEVEL : LearnableMoveSource.PREVO;
           }
+          moves.push([lm[0], lm[1], moveSource]);
         }
+        ret.push(...moves);
```
SirzBenjie: Is there a reason to not just do `ret.push([lm[0], lm[1], moveSource]` in line 2902? Why create a temporary array?

SirzBenjie (commented):
I have more comments here.

Given that this is itself around 400 LoC total, I think extracting it out into its own module just... makes sense. Our `Pokemon` class is already untennably large.

I apologize for the nits here. Many of the smells are _not_ introduced by your PR. If_ you would rather this get in and then cleaned up afterwards we can do that.


A few other observations:

The "fusion" version of each helper function is too much duplication. Now, I notice that the value for `LearnableMoveSource` is always 1 higher for the fusion version. You can leverage this to do away with the fusion version of each `getXMoves` source and then adding 1 to the output if it's fusion.

For example:


```ts
function computeLearnableMoveSource(level: number, fromPrevo = false): LearnableMoveSource {
  switch (level) {
    case RELEARN_MOVE:
      return LearnableMoveSource.RELEARN;
    case EVOLVE_MOVE:
      return LearnableMoveSource.EVOLUTION;
    default:
      return isPrevo ? LearnableMoveSource.PREVO : LearnableMoveSource.LEVEL;
  }
}
```

And then when you use it (note this is in my WIP version that extracted out the helpers from the pokemon class).

```ts
/**
 * Helper method for {@linkcode getLevelMoves}
 *
 * Get all level moves the species form can learn on its own.
 * @param includeEvolutionMoves - Whether to include evolution moves
 * @param includeRelearnerMoves - Whether to include moves that would require a relearner. Note the move relearner inherently allows evolution moves
 * @returns A list of moves and the levels they can be learned at, along with the source of the moveexcludes moves from prevolutions, but includes evolution moves and relearner moves
 */
function getRegularLevelMoves(
  pokemon: Pokemon,
  includeEvolutionMoves: boolean,
  includeRelearnerMoves: boolean,
  fromFusion = false
): LevelMovesWithSource {
  const ret: LevelMovesWithSource = [];
  const moves = pokemon.getSpeciesForm(true).getLevelMoves();
  for (const [level, move] of moves) {
    if (
      (includeEvolutionMoves && level === EVOLVE_MOVE)
      || (includeRelearnerMoves && level === RELEARN_MOVE)
      || level > 0
    ) {
      // Cast to LearnableMoveSource is safe as each enum value has a fusion version that is 1 greater
      const moveSource = computeLearnableMoveSource(level) + +fromFusion as LearnableMoveSource;
      ret.push([level, move, moveSource]);
    }
  }
  return ret;
}
```

Another thing is, I recommend replacing the `const lm of learnMoves` with `const [level, move, (source)] of learnMoves`, and then replacing `lm[0]` and `lm[1]` everywhere with it. It's easier to read (and is negligibly more performant as it avoids lookups).


I was working on creating a patch with my suggestions and sending it to you, but I had to stop for the day and wanted to get this review in.
Here's the file I was working on (had to rename extension to .txt to allow github, it's a typescript file). Maybe this is a starting point? (I got stuck because I couldn't understand the reason for the split between the branches in `getPrevolutionMoves`).
[learnsets.txt](https://github.com/user-attachments/files/29074545/learnsets.txt)

src/field/pokemon.ts
fabske0: Some of the stuff I copied from #7371 
But I'm getting more confused every second I think about it. Because in the `includeRelearnerMoves=true` path if a prevo has a relearn/evo move it gets that as it's source, but should it be that way or should it be `prevo`
If it's a prevo it should probably always be source `Prevo`??

fabske0: `Level` should be possible I think, the only diff is that `includeRelearnerMoves=false` excludes level 1 moves from the species itself (not prevo)

fabske0:
@SirzBenjie Applied your file with some minor changes (docs and adding level in the relearner=false path)
From a quick test, the output is the same between this and the old version.
Now we only need to decide whether a relearn/evo move from a prevo should have evo/relearn as source or `prevo`

src/field/learnsets.ts
SirzBenjie, line 1: Maybe best located in `src/data/pokemon/learnsets.ts`. Not that our file structure makes much sense...

fabske0: I think keeping it in constants is fine. Kev wanted to split that up anyways.

#6844

src/field/learnsets.ts
```diff
@@ -0,0 +1,274 @@
...
+  pokemon: Pokemon,
+  includeEvolutionMoves: boolean,
+  includeRelearnerMoves: boolean,
+  fromFusion = false,
+): LevelMovesWithSource {
+  const ret: LevelMovesWithSource = [];
+  // Convey that this method cannot use `fromFusion` if the pokemon is not a fusion and the bang is then safe
+  const speciesBase = fromFusion ? pokemon.fusionSpecies! : pokemon.species;
+  if (!speciesBase && fromFusion) {
+    // TODO: Find a better way to handle fromFusion=true without being a fusion
```
SirzBenjie, line 75: I think this is fine, no?

src/field/learnsets.ts
```diff
@@ -0,0 +1,274 @@
...
+function getPrevolutionMoves(
+  pokemon: Pokemon,
+  includeEvolutionMoves: boolean,
+  includeRelearnerMoves: boolean,
+  fromFusion = false,
+): LevelMovesWithSource {
+  const ret: LevelMovesWithSource = [];
+  // Convey that this method cannot use `fromFusion` if the pokemon is not a fusion and the bang is then safe
+  const speciesBase = fromFusion ? pokemon.fusionSpecies! : pokemon.species;
+  if (!speciesBase && fromFusion) {
```
SirzBenjie, line 74: With the check in line 81 protecting against undefined, you can get rid of the `!` after `pokemon.fusionSpecies`

src/field/learnsets.ts
```diff
@@ -0,0 +1,274 @@
...
+      }
+    }
+  }
+  return ret;
+}
+
+/**
+ * Helper method for {@linkcode getLevelMoves}
+ *
+ * Get all level moves the fusion species form can learn, excluding any moves from the base species' evolution chain.
```
SirzBenjie: > excluding any moves from the base species' evolution chain.

I'm realizing I don't see logic for this. Is this method supposed to do that?

Also, as is, I believe this is redundant with the `getRegularLevelMoves` and passing fusion=true (I didn't finish all the edits in the file I uploaded in my review comment).

src/field/learnsets.ts
```diff
@@ -0,0 +1,274 @@
...
+
+    return (
+      !(level > pokemon.level)
+      && !isOwnMoveFromNonLevelSource
+      && !isLockedPrevoMove
+      && (includeRelearnerMoves || !isRelearner || allowedEvolutionMove)
+    );
+  });
+
+  /**
```
SirzBenjie:
```suggestion
  /*
```

src/field/learnsets.ts
```diff
@@ -0,0 +1,274 @@
...
+
+  // A set of moves the species learns by level.
+  // Used to prefer showing a move as level rather than showing as a prevo/evo move
+  const ownMoves = new Set(
+    levelMoves
+      .filter(lm => lm[2] === LearnableMoveSource.LEVEL || lm[2] === LearnableMoveSource.FUSION_LEVEL)
+      .map(lm => lm[1]),
+  );
+
+  /**
```
SirzBenjie:
```suggestion
  /*
```

Not a doc comment

src/field/learnsets.ts
fabske0: It is fine, I'm just not that happy about nothing stopping you from calling it with fusion while not being a fusion

src/field/learnsets.ts
fabske0: It's complaining about line 86 when i remove the `!` because of the `fromFusion` check in the if

src/field/learnsets.ts
fabske0: You're right. I might have accidentally accepted a wrong copilot suggestion here

fabske0:
Example for the fusion changes

image
</details> 
<details><summary>after</summary>
image
</details>

damocleas (dismissed):

src/field/learnsets.ts
```diff
@@ -0,0 +1,279 @@
...
+    const speciesLevelMoves = getPokemonSpeciesForm(evolutionChain[e][0], pokemon.formIndex).getLevelMoves();
+    for (const [level, move] of speciesLevelMoves) {
+      const includeLevelOne = !e || level > 1 || includeRelearnerMoves;
+      if (includeRelearnerMoves && level === RELEARN_MOVE) {
+        const source = isPrevo ? LearnableMoveSource.PREVO : LearnableMoveSource.RELEARN;
+        ret.push([level, move, (source + +fromFusion) as LearnableMoveSource]);
+      } else if (includeEvolutionMoves && level === EVOLVE_MOVE) {
+        const source = isPrevo ? LearnableMoveSource.PREVO : LearnableMoveSource.EVOLUTION;
+        ret.push([level, move, (source + +fromFusion) as LearnableMoveSource]);
+      } else if (includeLevelOne && (!isPrevo || level <= evolutionChain[e + 1][1])) {
```
emdeann: @fabske0 This `(!isPrevo || level <= evolutionChain[e + 1][1])` is explicitly excluding any moves learned by prevolutions after the evolution level. I talked to balance and this is not the intent. As an example, in the current state of the PR, Breloom will *not* be able to learn Spore by mushroom, because Shroomish learns Spore after its evolution level.

emdeann:
If my above comment is resolved and I have a good enough understanding of how this works, you should be able to resolve #7401 by replacing `LevelUpPhase#end` with this:

```ts
  public override end(): void {
    // this if check feels like an unnecessary optimization
    if (this.lastLevel < 100) {
      const levelMoves = this.getPokemon().getLevelMoves(this.lastLevel + 1, false, true);
      const prevoMoves = levelMoves.filter(([, , source]) => getBaseLearnableMoveSource(source) === LearnableMoveSource.PREVO);
      const newMoves = levelMoves.filter(([, , source]) => getBaseLearnableMoveSource(source) !== LearnableMoveSource.PREVO);

      for (const [, moveId] of newMoves) {
        globalScene.phaseManager.unshiftNew("LearnMovePhase", this.partyMemberIndex, moveId);
      }
      for (const [, moveId] of prevoMoves) {
        globalScene.phaseManager.queueMessage(
          i18next.t("battle:prevoMovePrompt", {
            pokemonName: getPokemonNameWithAffix(this.pokemon),
            moveName: allMoves[moveId].name,
          }),
          null,
          true,
        );
      }
    }
    if (!this.pokemon.pauseEvolutions) {
      const evolution = this.pokemon.getEvolution();
      if (evolution) {
        this.pokemon.breakIllusion();
        globalScene.phaseManager.unshiftNew("EvolutionPhase", this.pokemon, evolution, this.lastLevel);
      }
    }
    super.end();
    return;
  }
```

src/field/learnsets.ts
fabske0: This was just copied over from the previous `getLevelMoves` method.
Changed it to check the current level instead

fabske0:
> If my above comment is resolved and I have a good enough understanding of how this works, you should be able to resolve #7401 by replacing `LevelUpPhase#end` with this:

LGTM
added to the PR

damocleas (dismissed):

src/field/learnsets.ts
```diff
@@ -0,0 +1,279 @@
...
+ * Helper method for {@linkcode getLevelMoves}
+ *
+ * Get all level moves the species form can learn on its own.
+ * @param pokemon - The Pokemon to get the moves for
+ * @param includeEvolutionMoves - Whether to include evolution moves
+ * @param includeRelearnerMoves - Whether to include moves that would require a relearner. Note the move relearner inherently allows evolution moves
+ * @param fromFusion - Whether to get the moves from the fusion species \
+ * @remarks
+ * `fromFusion` should only be used if the pokemon is a fusion
+ * @returns A list of moves and the levels they can be learned at, along with the source of the moveexcludes moves from prevolutions, but includes evolution moves and relearner moves
```
DayKev:
```suggestion
 * @returns A list of moves and the levels they can be learned at, along with the source of the move.
 * Excludes moves from prevolutions, but includes evolution moves and relearner moves.
```

src/field/learnsets.ts
```diff
@@ -0,0 +1,279 @@
...
+import { getPokemonSpeciesForm } from "#utils/pokemon-utils";
+
+/**
+ * Helper method for {@linkcode getLevelMoves}
+ *
+ * Get all level moves the species form can learn on its own.
+ * @param pokemon - The Pokemon to get the moves for
+ * @param includeEvolutionMoves - Whether to include evolution moves
+ * @param includeRelearnerMoves - Whether to include moves that would require a relearner. Note the move relearner inherently allows evolution moves
+ * @param fromFusion - Whether to get the moves from the fusion species \
```
DayKev:
```suggestion
 * @param fromFusion - (Default `false`) Whether to get the moves from the fusion species
```

src/field/learnsets.ts
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/**
+ * Helper method for {@linkcode getLevelMoves}
+ *
+ * Get all level moves the species form and its prevolutions can learn.
+ * @param pokemon - The Pokemon to get the prevolution moves for
+ * @param includeEvolutionMoves - Whether to include evolution moves
+ * @param includeRelearnerMoves - Whether to include moves that would require a relearner. Note the move relearner inherently allows evolution moves
+ * @param fromFusion - Whether to get the prevolution moves from the fusion species \
```
DayKev:
```suggestion
 * @param fromFusion - (Default `false`) Whether to get the prevolution moves from the fusion species
```

src/field/learnsets.ts
```diff
@@ -0,0 +1,279 @@
...
+ * @returns A list of moves and the levels they can be learned at, along with the source of the move
+ */
+function getPrevolutionMoves(
+  pokemon: Pokemon,
+  includeEvolutionMoves: boolean,
+  includeRelearnerMoves: boolean,
+  fromFusion = false,
+): LevelMovesWithSource {
+  const ret: LevelMovesWithSource = [];
+  // Convey that this method cannot use `fromFusion` if the pokemon is not a fusion and the bang is then safe
```
DayKev: The wording on this comment is weird. Is this supposed to be a todo?

src/field/learnsets.ts
```diff
@@ -0,0 +1,279 @@
...
+ * Finds all non-duplicate items from the input, and pushes them into the output.
+ * Two items count as duplicate if they have the same Move, regardless of level.
+ *
+ * @param levelMoves - The input array to search for non-duplicates from
+ * @param ret - The output array to be pushed into.
+ */
+function getUniqueMoves(levelMoves: LevelMovesWithSource, ret: LevelMovesWithSource): void {
+  const uniqueMoves: MoveId[] = [];
+  for (const lm of levelMoves) {
+    if (!uniqueMoves.find(m => m === lm[1])) {
```
DayKev:
```suggestion
    if (!uniqueMoves.some(m => m === lm[1])) {
```

src/field/learnsets.ts
```diff
@@ -0,0 +1,279 @@
...
+
+/**
+ * Helper method for {@linkcode getLevelMoves}
+ *
+ * Filters out moves not within the correct level range(s) and duplicates.
+ * Sorts the moves first by level, then by source.
+ * @param levelMoves - The list of level moves to filter and sort
+ * @param startingLevel - Don't include moves below this level
+ * @param includeEvolutionMoves - Whether to include evolution moves
+ * @param includeRelearnerMoves - Whether to include moves that would require a relearner. Note the move relearner inherently allows evolution moves
```
DayKev:
```suggestion
 * @param includeEvolutionMoves - (Default `false`) Whether to include evolution moves
 * @param includeRelearnerMoves - (Default `false`) Whether to include moves that would require a relearner. Note the move relearner inherently allows evolution moves
```

emdeann (dismissed):
When I tested this it worked fine (aside from my comment which was addressed). I don't have the bandwidth to review this full PR, but others have done that from a code hygiene perspective. This is definitely something we want on beta for as long as possible before the update

src/field/learnsets.ts
```diff
@@ -0,0 +1,279 @@
...
+  }
+  const evolutionChain = speciesBase.getSimulatedEvolutionChain(
+    pokemon.level,
+    pokemon.hasTrainer(),
+    pokemon.isBoss(),
+    pokemon.isPlayer(),
+  );
+  for (let e = 0; e < evolutionChain.length; e++) {
+    const isPrevo = e < evolutionChain.length - 1;
+    // TODO: Might need to pass specific form index in simulated evolution chain
```
DayKev, line 87: Where exactly is this comment referring to? `const evolutionChain` above? or `.getLevelMoves()` here?

src/field/learnsets.ts
```diff
@@ -0,0 +1,279 @@
...
+
+/**
+ * Helper method for {@linkcode getLevelMoves}
+ *
+ * Get all level moves the fusion species form can learn.
+ * @param includeEvolutionMoves - Whether to include evolution moves
+ * @param includeRelearnerMoves - Whether to include moves that would require a relearner. Note the move relearner inherently allows evolution moves
+ * @returns A list of moves and the levels they can be learned at, along with the source of the move
+ */
+function getFusionEvolutionMoves(
```
DayKev: What's the difference between this and `getRegularLevelMoves()` with `fromFusion=true`? It seems like the output would be identical.

src/field/learnsets.ts
```diff
@@ -0,0 +1,279 @@
...
+
+  // A set of moves the species gets by level, but are above the current level
+  const prevoLockedMoves = new Set(
+    levelMoves
+      .filter(
+        lm =>
+          (lm[2] === LearnableMoveSource.LEVEL || lm[2] === LearnableMoveSource.FUSION_LEVEL) && lm[0] > pokemon.level,
+      )
+      .map(lm => lm[1]),
+  );
```
DayKev, line 171: What does this have to do with prevos? It appears to just be filtering the pokemon's normal level up moves for those above the current level.

src/field/learnsets.ts
```diff
@@ -0,0 +1,279 @@
...
+    levelMoves.push(...getRegularLevelMoves(pokemon, includeEvolutionMoves, includeRelearnerMoves));
+  }
+
+  if (pokemon.isFusion() && learnSituation !== LearnMoveSituation.EVOLUTION_FUSED_BASE) {
+    const methodFunc = includePrevolutionMoves ? getPrevolutionMoves : getRegularLevelMoves;
+    levelMoves.push(...methodFunc(pokemon, includeEvolutionMoves, includeRelearnerMoves, true));
+  }
+
+  const ret = filterAndSortLevelMoves(pokemon, levelMoves, startingLevel, includeEvolutionMoves, includeRelearnerMoves);
+  return ret;
```
DayKev:
```suggestion
  return filterAndSortLevelMoves(pokemon, levelMoves, startingLevel, includeEvolutionMoves, includeRelearnerMoves);
```

src/field/pokemon.ts
```diff
@@ -1921,18 +1924,37 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    *
    * Available egg moves are only included if the {@linkcode Pokemon} was
    * in the starting party of the run and if Fresh Start is not active.
-   * @returns An array of {@linkcode MoveId}s, as described above.
+   * @returns A tuple of the Level (or `null` for non level moves), {@linkcode MoveId} and their corresponding {@linkcode LearnableMoveSource}, as described above.
    */
-  public getLearnableLevelMoves(): MoveId[] {
-    let levelMoves = this.getLevelMoves(1, true, false, true).map(lm => lm[1]);
+  public getLearnableLevelMoves(): [number | null, MoveId, LearnableMoveSource][] {
```
DayKev, line 1930: This should be moved into the `#region LevelMoves` block (but not in this PR, otherwise git history would be annoying).
```suggestion
  // TODO: move into `#region LevelMoves`
  public getLearnableLevelMoves(): [number | null, MoveId, LearnableMoveSource][] {
```

src/field/pokemon.ts
```diff
@@ -1921,18 +1924,37 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    let learnableMoves: [number | null, MoveId, LearnableMoveSource][] = [];
+    learnableMoves = this.getLevelMoves(1, true, true, true);
     if (this.metBiome === -1 && !globalScene.gameMode.isFreshStartChallenge() && !globalScene.gameMode.isDaily) {
-      levelMoves = this.getUnlockedEggMoves().concat(levelMoves);
+      const eggMoves: [number | null, MoveId, LearnableMoveSource][] = this.getUnlockedEggMoves().map(em => [
+        null,
+        em,
+        LearnableMoveSource.EGG,
+      ]);
+      learnableMoves = eggMoves.concat(learnableMoves);
```
```suggestion
      learnableMoves.push(...eggMoves);
```

src/field/learnsets.ts
Mourouh: Shroomish gets Seed Bomb at level 36. Breloom gets Seed Bomb at level 44. A level 40 Breloom should not get Prevo access to Seed Bomb.

src/field/pokemon.ts
```diff
@@ -1921,18 +1924,37 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
-      levelMoves = this.usedTMs.filter(m => !levelMoves.includes(m)).concat(levelMoves);
-    }
-    levelMoves = levelMoves.filter(lm => !this.moveset.some(m => m.moveId === lm));
-    return levelMoves;
+      const tmMoves: [number | null, MoveId, LearnableMoveSource][] = this.usedTMs.map(tm => [
+        null,
+        tm,
+        LearnableMoveSource.TM,
+      ]);
+      learnableMoves = tmMoves.filter(tm => !learnableMoves.some(lm => lm[1] === tm[1])).concat(learnableMoves);
```
```suggestion
      learnableMoves.push(...tmMoves.filter(tm => !learnableMoves.some(lm => lm[1] === tm[1])));
```

src/field/pokemon.ts
```diff
@@ -1921,18 +1924,37 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
+    }
+    learnableMoves = learnableMoves.filter(lm => !this.moveset.some(m => m.moveId === lm[1]));
+
+    // Sort by source, so the moves with a species prefix will be at the top
```
```suggestion
    // Sort by source in descending order, so the moves with a species prefix will be at the top
```
This is supposed to be in descending order, right?

src/field/pokemon.ts
```diff
@@ -2777,144 +2799,38 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
...
    * @param startingLevel - Don't include moves below this level
    * @param includeEvolutionMoves - Whether to include evolution moves
-   * @param simulateEvolutionChain - Whether to include moves from prior evolutions
+   * @param includePrevolutionMoves - Whether to include moves from prior evolutions
    * @param includeRelearnerMoves - Whether to include moves that would require a relearner. Note the move relearner inherently allows evolution moves
-   * @returns A list of moves and the levels they can be learned at
+   * @returns A list of moves and the levels they can be learned at, along with the source of the move
    */
-  getLevelMoves(
+  public getLevelMoves(
```
DayKev, line 2814:
```suggestion
  // TODO: convert to use object param
  public getLevelMoves(
```

src/field/pokemon.ts
```diff
@@ -2923,7 +2839,7 @@ export abstract class Pokemon extends Phaser.GameObjects.Container {
    * @param levelMoves - The input array to search for non-duplicates from
    * @param ret - The output array to be pushed into.
    */
-  private static getUniqueMoves(levelMoves: LevelMoves, ret: LevelMoves): void {
+  private static getUniqueMoves(levelMoves: LevelMovesWithSource, ret: LevelMovesWithSource): void {
```
DayKev: This is unused.

image

</details>

src/field/learnsets.ts
DayKev: That's not what this const is though, from what I can see. It's used later for that purpose, but this specific const seems like it should be named `levelMovesAboveCurrentLevel` or something like that.

src/ui/handlers/party-ui-handler.ts
```diff
@@ -775,7 +705,9 @@ export class PartyUiHandler extends MessageUiHandler {
     // show move description
     const option = this.options[this.optionsCursor];
     const pokemon = globalScene.getPlayerParty()[this.cursor];
-    const move = allMoves[pokemon.getLearnableLevelMoves()[option]];
+    const learnableMoves = pokemon.getLearnableLevelMoves();
+    const moveId: MoveId = learnableMoves[option]?.[1];
```
DayKev:
```suggestion
    const moveId: MoveId | undefined = learnableMoves[option]?.[1];
```

src/ui/handlers/party-ui-handler.ts
```diff
@@ -1568,6 +1493,20 @@ export class PartyUiHandler extends MessageUiHandler {
...
     this.updateOptionsWindow();
+    this.updatePartyMessageBox();
+  }
+
+  private updatePartyMessageBox(): void {
+    if (!this.optionsMode || !this.optionsBg) {
+      return;
+    }
+
+    /** Item management mode uses a taller message box due to two-line messages. */
```
DayKev:
```suggestion
    // Item management mode uses a taller message box due to two-line messages.
```

src/field/learnsets.ts
fabske0: It is kinda covered by the todo below. removing this comment

src/field/learnsets.ts
fabske0: renamed

src/field/learnsets.ts
fabske0: I'm not sure. It's copied over from the previous `pokemon#getLevelMoves()`.
But that comment was added 2 years ago so idk.

<details><summary>img</summary>
image
</details>

src/field/learnsets.ts
fabske0: Was a leftover from when I just split the existing method up before benjie refactored it

damocleas (approved):

emdeann (approved):

Madmadness65 (approved):

PR #7381, opened 2026-06-15
feat: add `luck` to `window.gameInfo`, add `@since` docs

DayKev:
The party's luck value can now be retrieved from `window.gameInfo.luck` by plugin devs.

Addition of luck value was requested.

- `window.gameInfo.luck` now returns the party's luck value, or `-1` if there is no active battle (e.g. when on the title screen).
- Added `/** @since x.y.z */` docs to all of the `gameInfo` fields.
- `gameInfoVersion` increased from `2.0.0` to `2.1.0`.

## Screenshots/Videos
<details><summary>On title screen</summary>

image

</details>
<details><summary>In battle</summary>

image

</details>
<details><summary>Full party of T3 shinies</summary>

image

</details> 

## How to test the changes?
Load the game, open the browser devtools console, and query `window.gameInfo` in various situations.

## Checklist


src/battle-scene.ts
```diff
@@ -3083,87 +3083,207 @@ export class BattleScene extends SceneBase {
...
+        acc: number;
+        /** @since 2.0.0 */
+        eva: number;
+      };
+      /** @since 2.0.0 */
+      shiny: boolean;
+      /** @since 2.0.0 */
+      variant: string;
+      /** @since 2.0.0 */
+      isFusion: boolean;
```
SirzBenjie, line 3180: This is a lot of noise; `@since` tags shouldn't be on each individual entry if it's the same as the parent struct. Fields should be assumed to inherit the version of their parent; only sub-fields that were added or updated since the parent was introduced should have it.

EDIT: Messed up the comment range, but you get the idea.

DayKev: So like 09e2c81 + 28ae0b0 then?

src/battle-scene.ts
```diff
@@ -3083,87 +3083,186 @@ export class BattleScene extends SceneBase {
...
-      gameInfoVersion: "2.0.0",
+      gameInfoVersion: "2.1.0",
       playTime: this.sessionPlayTime ?? 0,
       gameMode: this.currentBattle ? this.gameMode.getName() : "Title",
       biome: this.currentBattle ? getBiomeName(this.arena.biomeId) : "",
       wave: this.currentBattle?.waveIndex ?? 0,
+      luck: Math.min(
+        this.currentBattle ? (this.party?.map(p => p.getLuck()).reduce((total, cur) => total + cur, 0) ?? -1) : -1,
+        14,
+      ),
```
fabske0:
```suggestion
      luck: this.currentBattle ? getPartyLuckValue(this.party) : -1,
```

src/battle-scene.ts
fabske0: I think that's even more confusing. Maybe have them grouped so it's like:
```ts
type GameInfo = {
      /** @since 1.0.0 */
      party: PartyInfo[];
      gameMode: string;
      biome: string;
      wave: number;
      /** @since 1.1.0 */
      playTime: number;
      /** @since 2.0.0 */
      gameInfoVersion: string;
      /** @since 2.1.0 */
      luck: number;
    };
```

fabske0: Also I'd prefer if the types get moved out of the method itself and into it's own file. Then having a `@since` doc for every property is fine imo.

Bertie690 (dismissed):

src/battle-scene.ts
DayKev: I think I misunderstood the initial comment. Made another change.

> Also I'd prefer if the types get moved out of the method itself and into it's own file. Then having a `@since` doc for every property is fine imo.

I didn't want to move them to another file since they're only used here, in this one spot.

Also I kept them in the same order in the type as in the object itself, I'd rather not re-order just the type without re-ordering the object too.

PR #7193, opened 2026-03-25
feat!: add Battle Bond Greninja and hisui basculin as starters

macwellbz:
## What are the changes the user will see?

Battle Bond Greninja is now available as a separate starter from regular Froakie who will start generating Mega Stones in the upcoming [Legends: Z-A Mega Evolutions PR]( https://github.com/pagefaultgames/pokerogue/pull/6668). Also removes the Battle Bond Froakie and Frogadier forms which were not originally from mainline.

New starter is 5 cost and players who had the form previously unlocked on the base starter get access to it, as well as having their Froakie nature and shiny / variant unlocks carried over.

Similarly, white-striped Basculin has been split off from the other two forms. Basculin has been moved to 3 cost and this new version is 4 cost. For players that had obtained white-striped basculin, shiny variants and candies transferred over to the new form.

## Why am I making these changes?

#7151

## What are the changes from a developer perspective?

The old forms for the Froakie evolution line have been deleted and replaced with the new Battle Bond Greninja starter.

The migration file `v1_11_17.ts` is incomplete and some comments have been left there. I will try to address them but could use some help as so far I haven't been able to.

### Notes from Benjie

> [!WARNING]
>  **This patch updates the game version to 1.12.0**. This is required in order for save migration to run.
- In order for the session migrator to work, `parseSessionData` in `game-data.ts` was tweaked such that version migrators now operate on POJOs instead of classes by passing them the output from `JSON.parse(dataStr)` instead of running the transforms. I manually inspected the existing system migrators, and all of them _should_ work on raw JSON data.

## Screenshots/Videos

<details><summary>Before</summary>

image

image

image

</details>
<details><summary>After</summary>

image

image

image

</details>

## How to test the changes?

Import `everything.prsv` and view the new starter ingame. 

`/test/tests/abilities/battle-bond.test.ts` has also been updated.

## Checklist

- The PR content is correctly formatted:
  - [x] **I'm using `beta` as my base branch**
  - [x] **The current branch is not named `beta`, `main` or the name of another long-lived feature branch**
  - [x] I have provided a clear explanation of the changes within the PR description
  - [x] The PR title matches the Conventional Commits format (as described in [CONTRIBUTING.md](https://github.com/pagefaultgames/pokerogue/blob/beta/CONTRIBUTING.md#pr-title-format))
- [x] The PR is self-contained and cannot be split into smaller PRs
- [x] There is no overlap with another open PR
- The PR has been confirmed to work correctly:
  - [x] I have tested the changes manually
[...]

PR #7412, opened 2026-06-22
refactor: Add type safety to session migrators

SirzBenjie:

(note: added below the changelog cutoff markdown comment so this shouldn't appear in changelog)
[beta only] Should fix remaining cases of sessions showing up as corrupted / could not load.

#7193  left migrators as using classes like `PokemonData` despite shifting away from migrators running prior to converting into classes. This caused issues with migrators assuming well-formed fields. To alleviate this, type safety has been improved for migrators.

This also fixes some lingering bugs causing sessions to not migrate (due to missing fields that may have been expected).

#### `src/@types/save-migrators.ts`
Defines a new interface, `SessionSaveMigratorIn` for the type of the data that Session migrators operate on.
The interface inherits 

#### `src/@types/type-helpers.ts`
Added CoercePropertiesToUnknown<T> (map of properties to unknown) and StringLiteral<T> helpers. The former is used to improve editor experience by adding the properties of `SessionSaveData` to the interface for autocomplete while marking their type as unknown to sustain type safety.
The latter is added to help with type inferencing for the utility methods that were added

#### `src/utils/migrator-utils.ts` (new)
Three methods to aid in save migrator validation:

`valiateIsArrayOfObjects` - Returns whether the argument is an array, and all of its elements are non-null objects. Alleviates repetition where several migrators needed to check, e.g., that `data.modifiers` or `data.party` were well formed in order to safely operate on the data.

`ensurePropertyIsObject` - Given an object and a string literal corresponding to a property, ensure the property exists on the object as a non-null object, replacing (or adding) the property with an empty object if it does not.

`isPropertyAnObject` - Similar to above. Returns whether the specified property exists as a non-null object.

#### `src/system/version-migration/migration-errors.ts` (new)
Added custom error classes for migration errors (and used them when pre-migration validation fails).

#### migrators (src/system/version-migration/migrators/*.ts)

Adjusted migrators to operate on fields in a type safe matter. If the migrator receives data in a form that is not expected, the migrator will abort or skip working on the field.

Before session migrators are run, validation is done to ensure that `party` is an array of objects (using the aforementioned `validateIsArrayOfObjects`, throwing the new `SessionMigrationError` if it is not.
For `enemyParty`, if it is null (due to being allowed to be null for some mystery encounters), it is replaced with an empty array prior to migration. If `enemyParty` is not null, then the same `validateIsArrayOfObjects` check is done, also throwing on failure.
This allows migrators to operate on `party` and `enemyParty` as arrays of objects, removing validation boilerplate that would otherwise be necessary (as other properties are assumed to be unknown).
The presence of this validation enables the `party` and `enemyParty` fields to be marked as `Record<string, unknown>[]` instead of `unknown` in the `SessionMigratorIn` interface.

#### Misc

Removed several long-obsolete fields in `PokemonData` (as they were only ever there due because of migrators)
 - MysteryEncounterPokemonData
 - FusionMysteryEncounterPokemonData
 - NatureOverride
 - `hitsRecCount` from `customPokemonData`

Removed the unnecessary battle bond hidden ability on battle bond Greninja (that was put incorrectly in #7193).

## Screenshots/Videos
N/A

## How to test the changes?
Take some old saves and ensure migrators work as expected.

## Checklist
- ~~[ ] I have provided screenshots/videos of the changes (if applicable)~~
  - ~~[ ] I have made sure that any UI changes work for both the default and legacy UI themes (if applicable)~~


src/@types/type-helpers.ts
```diff
@@ -155,3 +155,5 @@ export type TupleRange<Min extends number, Max extends number, T = unknown> =
  * Any uses of this type should be double-checked to ensure that IDE hover tooltips are actually improved by its addition.
  */
 export type PreventHoverExpansion<T> = T & {};
+
+export type StringKeyedObject = { [key: string]: unknown };
```
Bertie690: we should literally just use `Record<string, unknown>`

also just FYI, this is incomaptible with any interface that lacks an explicit `string` index signature (not types tho, it's weird)

src/system/pokemon-data.ts
```diff
@@ -69,18 +69,13 @@ export class PokemonData {
...
-  public mysteryEncounterPokemonData: CustomPokemonData | null;
-  public fusionMysteryEncounterPokemonData: CustomPokemonData | null;
-
   /**
    * Construct a new {@linkcode PokemonData} instance out of a {@linkcode Pokemon}
    * or JSON representation thereof.
    * @param source The {@linkcode Pokemon} to convert into data (or a JSON object representing one)
    */
   // TODO: Remove any from type signature in favor of 2 separate method funcs
+  // TODO: change the source to `any` and/or create a method explicitly for converting from raw JSON data
```
DayKev: We should be removing `any` from `source`, not making it only `any`. At the point `new PokemonData(...)` is invoked, it should be valid save data or a `Pokemon` object.

src/system/version-migration/versions/v1_10_0.ts
```diff
@@ -16,6 +16,32 @@ interface OldTurnMove {
...
+    // NOTE: This unfortuately has to mis-classify Dancer and Magic Bounce-induced moves as `FOLLOW_UP`,
+    // given we previously had _no way_ of distinguishing them from follow-up moves post hoc.
+    useMode: tm.virtual ? MoveUseMode.FOLLOW_UP : tm.ignorePP ? MoveUseMode.IGNORE_PP : MoveUseMode.NORMAL,
+  };
+}
+
+function migrateSummonData(summonData: Record<string, unknown>): void {
+  if (typeof summonData !== "object" || summonData === null) {
+    return;
+  }
```
DayKev: This is unnecessary since you call `ensurePropertyIsObject(pkmn, "summonData")` directly before this function is called. You could change the `summonData` param's type to `: object` and remove this `if` statement.

src/system/version-migration/versions/v1_12_0_0.ts
```diff
@@ -190,10 +192,23 @@ function migrateSessionHisuiBasculin(pokemon: PokemonData, replaceSpecies: boole
...
+      // If it isn't a number, the guard below merely fails.
+      monoGenChallenge = challenges.find(c => c.id === 0)?.value as number;
+    }
+
+    if (!validateIsArrayOfObjects(data.party) || !validateIsArrayOfObjects(data.enemyParty)) {
+      console.warn(
+        "Malformed party/enemyParty in save data, skipping battle bond Greninja and hisui basculin migrator",
+      );
+      return;
+    }
```
DayKev, line 243: I think the enemy party being empty is a valid scenario (MEs), so you might have broken this migrator if that gets stored as `null` rather than an empty array. The enemy party should be coerced to an empty array if it's `null`.

src/system/version-migration/versions/v1_7_0.ts
```diff
@@ -36,14 +38,29 @@ const migrateUnselectableForms: SystemSaveMigrator = {
...
+  migrate: data => {
+    if (!validateIsArrayOfObjects(data.modifiers)) {
+      console.warn("Malformed modifiers in save data, skipping tera type migrator");
+      return;
+    }
     for (let i = 0; i < data.modifiers.length; ) {
       if (data.modifiers[i].className === "TerastallizeModifier") {
+        // Assert the modifier has the expected args structure
+        const modifierArgs = data.modifiers[i].args;
+        // Just remove malformed modifier...
```
DayKev: We shouldn't be silently modifying the save data like this if it's [seemingly] corrupt, at bare minimum there should be a warning in the console (though I'd prefer it not happen at all).

src/system/version-migration/versions/v1_0_4.ts
```diff
@@ -126,8 +126,15 @@ export const settingsMigrators: readonly SettingsSaveMigrator[] = [fixRerollTarg
 const migrateModifiers: SessionSaveMigrator = {
   version: "1.0.4",
   // biome-ignore lint/complexity/noExcessiveCognitiveComplexity: necessary?
-  migrate: (data: SessionSaveData): void => {
+  migrate: data => {
+    if (!validateIsArrayOfObjects(data.modifiers)) {
+      console.warn("Malformed modifiers in save data, skipping form change item migrator");
+      return;
+    }
```
DayKev, line 133: This would cause the migrator to not run on the enemy party's items if the player's items were corrupt (or if the player had no items, since `[]` is potentially stored as `null`).

src/system/version-migration/versions/v1_9_0.ts
```diff
@@ -1,40 +1,39 @@
...
-import type { SessionSaveData } from "#types/save-data";
 import type { SessionSaveMigrator } from "#types/save-migrators";
 
+function updatePokemonMoveset(data: Record<string, unknown>): void {
+  if ("moveset" in data && Array.isArray(data.moveset)) {
+    data.moveset = data.moveset.filter(m => !!m).map(m => PokemonMove.loadMove(m));
+  } else {
+    data["moveset"] = [new PokemonMove(MoveId.TACKLE), new PokemonMove(MoveId.GROWL)];
+  }
+
```
DayKev:
```suggestion
```
This migrator was never necessary in the first place, also now that these are pre-parse migrators you shouldn't be creating `PokemonMove` objects here anyway.

src/system/version-migration/versions/v1_9_0.ts
```diff
@@ -1,40 +1,39 @@
...
+  if (typeof data.battleData !== "object" || data.battleData === null) {
+    data.battleData = {};
+  }
+
+  if (data.customPokemonData && typeof data.customPokemonData["hitsRecCount"] === "number") {
+    // cast is safe since we check the presence of hitCount above.
+    (data.battleData as { hitCount?: number }).hitCount = data.customPokemonData["hitsRecCount"];
+    data.customPokemonData["hitsRecCount"] = null;
```
DayKev, line 15:
```suggestion
    // biome-ignore lint/performance/noDelete: intentional, the field doesn't exist anymore
    delete data.customPokemonData["hitsRecCount"];
```

src/system/version-migration/versions/v1_12_0_0.ts
```diff
@@ -230,21 +247,35 @@ const convertCustomPokemonDataTypes: SessionSaveMigrator = {
...
-        }
-        if (typeof modifier.typePregenArgs[0] === "number" && modifier.typePregenArgs[0] >= 50) {
-          modifier.typePregenArgs[0] += shiftAmount;
-        }
-      }
+  migrate: data => {
+    if (validateIsArrayOfObjects(data.modifiers)) {
+      data.modifiers.forEach(shiftFormChangeModifier);
+    } else {
+      console.warn("Malformed modifiers in save data, skipping form change item migrator");
```
DayKev, line 303: Technically I think it's not malformed for `data.modifiers` to be `null`... maybe? There's weird behavior where `[]` is being serialized as `null` (by rogueserver..?).

src/system/pokemon-data.ts
SirzBenjie: It should be `unknown`.

Based on what you are saying, you seem to prefer the latter method: having a `fromRaw` / `fromSerialized` method that calls the constructor with the arguments it needs.

If you mean that migrators should be ensuring that the save data is valid, I think that's what ends up decoupling the logic.

Migrators should only be concerned with addressing the change that was made to save data.

PokemonData is, by definition, a Data Transfer Object (DTO). It emphatically should be concerned with validating data that is being parsed into it.

src/system/version-migration/versions/v1_10_0.ts
SirzBenjie: Thanks, that was missed.

src/system/version-migration/versions/v1_12_0_0.ts
SirzBenjie: I thought MEs store enemy party as an empty array?

src/system/version-migration/versions/v1_7_0.ts
SirzBenjie: You're right, that shouldn't be done in the migrator. It should be done when loading the object.

src/system/version-migration/versions/v1_0_4.ts
SirzBenjie: Thanks

src/system/version-migration/versions/v1_9_0.ts
SirzBenjie: Yeah I left it intact for the draft.

src/system/version-migration/versions/v1_12_0_0.ts
SirzBenjie: Ah, thanks.

Bertie690 (approved):

damocleas (approved):

